Issue #3959, opened 2020-06-25
Bottom bar covers commenting in Check-in Meetings


avivapinchas:
## Issue - Bug

With smaller windows, bottom bar covers the commenting field in the discussion thread: 
![Screen Shot 2020-06-25 at 12 52 52 PM](https://user-images.githubusercontent.com/11621016/85772942-f5261800-b6e2-11ea-888d-3b525bcf44de.png)

- **OS:** macOS
- **Browser:** Chrome Version 83.0.4103.116 (Official Build) 

### Acceptance Criteria (optional)
Users can:
   - Enter comments without being blocked by the bottom bar

mattkrick:
this isn't how it should be. the discussion panel should sense the bottom bar & scootch up.
what are the steps to reproduce? are you changing the window size? is it for the retro or checkin?

avivapinchas:
This is during a check-in meeting, and yes, I was changing the window size - doesn't look to happen if the window size is already small.

jordanh:
Must add window resize listener

Issue #3934, opened 2020-06-16
Snackbar floats about 


avivapinchas:
## Issue - Bug

After fixing #3759, the snackbar no longer covers the bottom bar, but rather floats towards the top of the screen: 
![Screen Shot 2020-06-16 at 3 16 13 PM](https://user-images.githubusercontent.com/11621016/84827773-92e65c80-afea-11ea-8b07-a6941cdd5614.png)
![Screen Shot 2020-06-16 at 3 20 33 PM](https://user-images.githubusercontent.com/11621016/84827781-94b02000-afea-11ea-8777-3984952e3163.png)

I've noticed this for both the conversion prompt and the 'XX wants to join this meeting' notifications. Depending on the size of the window, the prompt can hide behind the browser URL bar, so all you see is a shadow at the top of the window. 

### Acceptance Criteria (optional)
Users can:
   - Snackbar chills in a place that makes sense

mattkrick:
any steps to reproduce reliably?

avivapinchas:
🤔 Maybe it got fixed? It was happening consistently for me - mostly the prompt hiding behind the URL bar - but I can't repro it right now. Will keep an eye out for it again and jot it down 👍

avivapinchas:
Got this again today - 
![floating-prompt](https://user-images.githubusercontent.com/11621016/86058838-1cece700-ba27-11ea-86c7-5937a4fdf432.gif)

This was during a Check-in meeting on the demo team I have. You can see at the start of the GIF that there's just a shadow towards the top - that's actually the prompt hiding. 

This was on my laptop screen, where last week I was using my second monitor and I wasn't seeing this, so not sure if window size is maybe a factor?

mattkrick:
any actions before it happened? resizing the window, etc?

avivapinchas:
Looked out for this during my last meeting - didn't resize. I typically notice the shadow at the top of the window, and then resize so I can actually see the dismiss button. It's happened at each check-in meeting that I've had this week, and this week I've been using my monitor and seeing it there.

nickoferrall:
Hey @avivapinchas, I'm having difficulty reproducing this one. Does the snackbar only float about with the "free tier exceeded" and "XX wants to join this meeting" notifications or for all snackbar notifications? And is this happening on Chrome?

I've taken a look at a few different snackbars on Chrome and Safari and on different screen sizes but they're all behaving nicely so far. 

[...]

PR #4036, opened 2020-07-30
Fixes for bottom bar cover and floating snackbar

nickoferrall:
Fixes for issues #3959 and #3934 

- Resizing the window/meeting content now causes the discussion thread to account for the position of the bottom bar
- Unable to reproduce the floating snackbar bug but have added a `Math.max` which should hopefully catch it

### Test

- [ ] Resize the window in a check-in and retro meeting and check whether the discussion thread position adjusts for the bottom bar
- [ ] Click the burger button and slide the bottom bar. Check whether the discussion thread position adjusts

### DIFF
--- a/packages/client/components/ActionMeetingAgendaItems.tsx
+++ b/packages/client/components/ActionMeetingAgendaItems.tsx
@@ -1,6 +1,6 @@
 import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
-import React from 'react'
+import React, {useRef} from 'react'
 import {createFragmentContainer} from 'react-relay'
 import {ActionMeetingAgendaItems_meeting} from '~/__generated__/ActionMeetingAgendaItems_meeting.graphql'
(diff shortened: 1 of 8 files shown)


mattkrick:
first test passes! clicking the burger button still doesn't account for the offset though. let me know if you can't reproduce & we'll track it down together

nickoferrall:
> first test passes! clicking the burger button still doesn't account for the offset though. let me know if you can't reproduce & we'll track it down together

Hmm interesting - I'm struggling to reproduce the bug so I thought I'd share a quick Loom video to share with you the steps I'm taking. I've tried doing this on Check-in and Retro meetings: https://www.loom.com/share/196eeff357434ccca1d75da9c729a0bf

mattkrick:
great video! i think i found the difference. The resize logic doesn't work for the group/vote phase.

nickoferrall:
Ah thanks for letting me know @mattkrick. I've made the update to the group/vote phase

PR #4042, opened 2020-08-03
add db tables for TemplateDimension and TemplateScale

### LINKED ISSUE #4010: Create sprint poker backend data model for templating v1
#### Acceptance Criteria
- [ ] `Dimension` table created with `name`, `scaleId`, and `teamId` fields.
- [ ] `Scale` table created with `name`, `teamId`, and `values` fields.
  - [ ] `values` can be enum field. each enum value may be a tuple. e.g. `(display_name, numerical_value)`. having both fields allows all scales to be used in math formulas, such as median or more complicated user-defined computations.
- [ ] `Template` table created (or `ReflectTemplate` renamed to be generic) with `name`, `teamId`, `scope` fields.
  - [ ] `scope` can be enum field, with values `team` (default), `org`, or `public`, to define who can see and use the template.
- [ ] `TemplateDimension` table created with `templateId` and `dimensionId` fields to model m2m relation btwn `Template` and `Dimension` table.
- [ ] `Meeting` table has `templateId` field.
- [ ] Indices on `id` fk fields on all tables above, to be added at discretion.

#### Out of Scope for v1
- Backend for "formulas" to calculate weighted scores from multiple dimensions

#### Other considerations
- We can hard code default dimensions, scales, templates, etc. Then clone them for each team. Similar to current behavior for retro templates. Have we encountered any limitations to this approach? How does it compare with keeping default parabol-owned dimensions, scales, and templates in the db along with team-owned ones?

**Estimated effort:** 8 hours

tiffanyhan:
Resolve #4010


packages/server/database/migrations/20200803121128-addTemplateDimension.ts
```diff
@@ -0,0 +1,22 @@
...
+      .indexCreate('templateId')
+      .run()
+  } catch (e) {
+    console.log(e)
+  }
+}
+
+export const down = async function(r: R) {
+  try {
+    await r.tableDrop('TemplateDimension').run()
```
mattkrick, line 18: i'm betting we'll need a `scaleId` for the foreign key, but safe to leave off until we need it!

packages/server/database/types/TemplateDimension.ts
```diff
@@ -0,0 +1,36 @@
...
+  createdAt?: Date
+  updatedAt?: Date
+  deletedAt?: Date
+}
+
+export default class TemplateDimension {
+  id: string
+  name: string
+  teamId: string
+  templateId: string | null
```
mattkrick, line 18: non null?

packages/server/database/types/TemplateDimension.ts
```diff
@@ -0,0 +1,36 @@
...
+  scaleId: string | null
```
mattkrick, line 19: non null?

tiffanyhan: yeah i wasn't 100% sure about these non-nulls. my thinking was the entity should still be able to persist without those fk ids, even if the initial endpoints require those properties. what do you think? we can always make them non null to begin with, then adjust if requirements change.

mattkrick: let's do it! i'm always in favor of having the strictest possible types, helps errors pop up so much sooner.
on the big picture, a dimension can't function without a scale or a template, so I'd say it's gotta have those. We could have a `DimensionInProgress` class that had them as nullable, but i don't think the UI requires that

tiffanyhan:
Er sorry for this diff looking weird. The other changes are already in master, but GitHub thinks they aren't for some reason. Accidentally deleted the branch on my local and created a new one off my master (instead of merging master to the branch). Lmk if this is a blocker and I can create a new PR!

Issue #2219, opened 2018-06-19
Due dates on archived task cards


ackernaut:
## Enhancement

Currently one can change the due date on a card in the archive.

### Acceptance Criteria

- Disallow changing due dates on archived task cards
- [Audit](https://github.com/ParabolInc/action/issues/2219#issuecomment-523516595) and disable all editing and editing state feedback

**Estimated effort:** 10.5 points ([see CONTRIBUTING.md](https://github.com/ParabolInc/action/blob/master/CONTRIBUTING.md#points-and-sizes))

jordanh:
EE: 10.5

jordanh:
@ackernaut, I edited your issue :)

ackernaut:
This is still a thing. Also, in the archive you can simply tap on a card and it shows an editing state. So the best fix whenever we get to this is to disable all editing and editing state feedback. Updating AC.

jefftorcato:
Hi, I would like to complete this task for 10.5 points.

jordanh:
@jefftorcato this one is all yours! Could you write me at jordan@parabol.co and we'll get you all set up to contribute to Parabol?

NikAiyer:
I'll take a look at this one

sinansonmez:
It seems like this issue is closed by #4047, right? If there is still some work to be done, I can take a look

Dschoordsch:
Yes, that was fixed long time ago

PR #4047, opened 2020-08-04
Disabled editing and due date selection for archived cards

NikAiyer:
This PR aims to solve issue #2219: 
It adds disable option to TeamArchive outcome cards so that editing and date selection can be disabled.

To test:
- [ ] Start server
- [ ] Log in to parabol instance with user
- [ ] Go to team section and check archived cards
- [ ] See that the due date button is missing and that clicking on card does not change editing status

### DIFF
--- a/packages/client/modules/outcomeCard/components/OutcomeCard/OutcomeCard.tsx
+++ b/packages/client/modules/outcomeCard/components/OutcomeCard/OutcomeCard.tsx
@@ -60,7 +60,6 @@ interface Props {
   setEditorState: (newEditorState: EditorState) => void
   useTaskChild: UseTaskChild
   dataCy: string
-  disabled?: boolean
 }
 
 const OutcomeCard = memo((props: Props) => {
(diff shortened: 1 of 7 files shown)


packages/client/modules/outcomeCard/components/OutcomeCard/OutcomeCard.tsx
```diff
@@ -74,7 +75,8 @@ const OutcomeCard = memo((props: Props) => {
     task,
     setEditorState,
     useTaskChild,
-    dataCy
+    dataCy,
+    disabled
   } = props
   const isPrivate = isTaskPrivate(task.tags)
   const isArchived = isTaskArchived(task.tags)
```
mattkrick, line 82: we've already got an `isArchived` flag here. let's use that instead to reduce the number of variables.
consider renaming `disabled` to `isArchived`. the task itself really isn't disabled, so it might get confusing when revisiting the code later on

NikAiyer: Yeah that makes sense. For some reason I didn't see isArchived. I'll just use that.

mattkrick:
flawless. nice work!

Issue #4024, opened 2020-07-24
Show who’s typing in a thread…


ackernaut:
### Background

I’d like to see feedback on who’s typing in the thread, similar to who’s adding a task. I think that polishing the multiplayer details of the UI improve folks’ attachment to the product in general.

I think this could be a good candidate for the 20% bucket of a sprint.

### Acceptance criteria
- The UI shows who’s typing. Compare with the copy variations of the `EditingStatus` component on the task cards. Perhaps something like:
  - Dan is typing…
  - Dan and Sara are typing…
  - Several are typing…
- The copy uses the animated `Ellipsis` component
- The copy hangs at the same indention of the input and top-level comment copy
- The copy uses the smaller hint styling (12px, text gray)
- See Figma for [a concept](https://www.figma.com/file/UhuGb8dL15bf3ylXoVqS1m/Retro-Discuss-Thread-v1-Prototype?node-id=714%3A119513) and [a screenshot](https://www.figma.com/file/UhuGb8dL15bf3ylXoVqS1m/Retro-Discuss-Thread-v1-Prototype?node-id=714%3A119584)
- **Stretch goal**: do we add this at the reply level, inline, with additional indention? E.g. ‘Dan is replying…’

### Estimated effort

- 3 hours

PR #4050, opened 2020-08-06
Show who is commenting

nickoferrall:
Issue #4024 

This PR shows who is commenting in the discuss phase of a meeting. 

### Considerations
- If a user is anonymously typing, don't let the team know that they're typing
- Showing "Dave is typing..." causes the existing comments to move up. I prefer this compared with extra padding at the bottom of a discussion thread. It's also comparable to what happens when someone is typing in iMessage
- However, on the empty state screen on the discussion thread there's just whitespace at the bottom so I've adjusted the height of the wrapper so that "Dave is typing" doesn't cause it to jump up
- Added `editCommenting` to the demo as well as `addComment`. I've added one comment to the first reflection group; it would be easy to add additional comments if we'd like. Just need to decide on the text. 

### Test

- [ ] Open two windows and start typing a comment in one of the discussion threads. The other window should show "X is typing"
- [ ] When the user clicks away from the input or waits 5 seconds, "X is typing disappears"
- [ ] When the user submits a comment, "X is typing disappears"
- [ ] Go to `/retrospective-demo` and observe the commenting functionality in the first reflection group

If we're happy with this approach, I can adding commenting feedback to threads too.


ackernaut:
Hey, @nickoferrall I’ll take a card to review this soon. Exciting stuff!

ackernaut:
### Test cycle no. 1

- [ ] Open two windows and start typing a comment in one of the discussion threads. The other window should show "X is typing"
- [ ] When the user clicks away from the input or waits 5 seconds, "X is typing disappears"
- [ ] When the user submits a comment, "X is typing disappears"

ackernaut:
I got sidetracked from reviewing this, but for some reason, the commenting hint is just not showing for me locally. I’ll look more in the AM.

nickoferrall:
One thing I should have mentioned in the description is that the two windows should belong to two different users, i.e. when User A types, only User B will see "User A is typing...". If you're signed into the same account in both windows, there won't be any commenting feedback.

If you're already signed in with two different users, just let me know and I'll investigate!

ackernaut:
Yes, it was 2 different accounts in 2 different Chrome profiles. I must be running or building something improperly.

nickoferrall:
I've checked it a few times on my device and it seems to be working for me. I did previously run into a (painful) problem with the gql executor which was only resolved by restarting my computer. If it's still not working for you, I'll dig into further and see if I can reproduce the bug

ackernaut:
Now I’m seeing the realtime editing, thanks for debugging that. A few things I’m noticing:
- I wonder if the feedback should start when the other user focuses on the input instead of when they start typing. Did you have a certain rationale? 🤔 
- For me, the hint text shows above (z-axis) the last comment and obscures it. I think it should bump up the comments or we leave empty space for it. Since the footprint is small and folks can scroll, the empty footprint might be okay. We could probably tighten up the top gutter as needed.
- There’s a flash in the hint text sometimes when editing. I suspect that when I stop typing briefly it triggers the 5-second timer. As I’m typing again, the 5 seconds are still counting down and the hint text goes away. Since I’m already typing again the hint comes back immediately. The 5-second delay should be cleared and restarted if the user starts typing again before it elapses.
- I haven’t look much more closely at the code, but I’ll give it a second read to see if my eyes spot anything

mattkrick:
We should add this functionality to the demo, too.
code would probably get added somewhere in here: https://github.com/ParabolInc/parabol/blob/revisit-check-in/packages/client/modules/demo/handleCompletedDemoStage.ts#L75

nickoferrall:
Thanks for the helpful feedback @ackernaut 
> I wonder if the feedback should start when the other user focuses on the input instead of when they start typing. Did you have a certain rationale? 🤔
- I was considering this as the edit task functionality is implemented `onFocus`. For me, as a user, I might click an input while I’m still thinking about what I’m going to type. I wouldn’t expect the team to be notified that I’m commenting until I start typing. I may well be an unusual user though and I’m happy to implement it onFocus if you have a preference
- I’ve updated the code so the flashing and thread position bugs should now be fixed

nickoferrall:
> We should add this functionality to the demo, too.
> code would probably get added somewhere in here: https://github.com/ParabolInc/parabol/blob/revisit-check-in/packages/client/modules/demo/handleCompletedDemoStage.ts#L75

thanks @mattkrick. I've added the function to the demo but I'm not sure how to test whether the demo is working? Do we have a demo account that I need to sign-in with?

ackernaut:
Hey @nickoferrall this is looking really good. Just a note, I only took a quick glance at code, so if you want me to look at a pattern more closely let me know I’ll be glad to.

I think we should leave an empty space where the hint text appears so that comments don’t jump up and down. I left a comment in Figma. The block should probably be 24-32px (2-3x 8px grid) between the input bar and the latest comment, and the text is centered in that grid space. It’s a minimal tradeoff to reduce thread change noise but cause scrolling just a little sooner if the thread has many comments.

On the other hand, if we add this multiplayer feedback to the replies I think it’s okay to have some change in layout then. I don’t think we’d want an empty block for hint text for every reply in the thread.

nickoferrall:
hey @ackernaut, thanks for the Figma sketch. I've added the empty space to prevent the jumping so it's ready for another look when you get the chance

nickoferrall:
hey @ackernaut, I've just pushed those lil padding updates and fixed the merge conflicts so it's ready when you are

PR #3984, opened 2020-07-10
Deprecated CustomPhaseItem for ReflectPrompt

NikAiyer:
This PR aims to resolve the following issue: #3962.

This PR renames the CustomPhaseItem table to ReflectPrompt and updates all references,  renames customPhaseItemsByTeamId dataloader to reflectPromptsByTemplateId and updates all references, and adds a templateId index to ReflectPrompt.

PR #4053, opened 2020-08-06
refactor custom phase item to reflect prompt

mattkrick:
this builds on top of #3984. it should work fine without requiring a refresh, but i might have missed something, so i'd recommend deploying this one during off-peak hours since it directly impacts meetings.

the migration will require a follow up to remove the old fields. i only added the new fields because the migration is probably going to take around 10 mins so i don't want the app to look for something that it removes.

items that i did not migrate over:
- Meeting focusedPhaseItemId
- Reflection phaseItem, retroPhaseItemId
- ReflectionGroup phaseItem, retroPhaseItemId
- a handful of graphql fields. can search for `deprecationReason` to snag em on the next version

overview: 
- CustomPhaseItem was an interface with a single concrete type of RetroPhaseItem. both have been deprecated in favor of ReflectPrompt
- creating & editing reflections now take in better named fields. createReflection accepts the old value for legacy purposes

TEST
- [ ] can run a retrospective meeting
- [ ] can run the demo


mattkrick:
this is a blocker for the shareable templates PR.
since we've just released i figured this is the best time to merge something that has such broad changes

PR #4067, opened 2020-08-12
migrate and postdeploy against ci env

tiffanyhan:
The only change now is that deploying to development triggers a build too.


mattkrick:
LGTM! do you have a yaml linter in vscode? i should probably get one of those...

Issue #4038, opened 2020-07-31
Permit re-invitation by email


jordanh:
## Issue - Enhancement

I spoke with [a user](https://app.hubspot.com/contacts/3888472/contact/7270101) 🔒  at a large institution who's in the middle of purchasing a license for us. Among her top requests was a subtle but important friction around inviting folks and starting meetings. After our meeting we had an email exchange.

Me:

> you're correct! We currently will display an error "User X is already on the team" if you double invite them. One clarifying question: Would you want users who've already taken part on the team to get a second invitation email?

User:

> I will probably provide you a longer answer than what you’re asking but… I often rely on copying and pasting from our mailing lists in AD into the invite for a retro. Since we have so much cross over in work and people can be apart of multiple organization ams doing retro’s, its possible they have already been invited into the organization from another team or did another retro under the Parabol team, it’s just too much of an extra lift to cross check across who is in Parabol and who is on the mailing list. So it would be lovely if there was some coding (similar to Slack) when you go to invite someone it checks it against who is already there, goes ahead and sends the email to those who aren’t, and then for those already there flags the users and says “hey this person is already here, do you still want to resend”?
>
> Because a lot of us are using the tool currently under the free version waiting for the enterprise license to come, we’re creating generic “teams” and setting up individual retros based on the different organization teams we’re working with at the moment. So I might have a [redacted] but then invite 9 different organization teams to 9 different retros just so we can use the tool. Unfortunately it mixes data, which is one of the many reasons we want the enterprise license
>

Key to understanding what is going on here is this user is copying emails from an email or from an Outlook calendar appointment. When pasting a large block of emails, it's difficult reconcile who is and who isn't already on the team.

In one design for this, we could pop a modal that says, "I've sent X invitations. These [list of users] are already on the team. Do you wish me to resend?"

### Acceptance Criteria (optional)
When users drop duplicate emails in the invitation dialog:
   - We send emails to users who haven't been invited yet
   - We confirm if we'd like to resend invitations to users who are already on the team

- **Estimated effort:** 12 hours ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

jordanh:
It looks like invitations do get sent today – but validation text may be throwing this user off. Looks like it might need a light copy edit and perhaps a different color than red/hot (something warning-ish?)

PR #4082, opened 2020-08-19
Invitation error feedback

nickoferrall:
This is a PR for issue #4038.

### Updates

- Displaying a warning message with a warning icon if the user enters the email address of an existing teammate
- If the user has submitted the invitation and there's an error, display the error message with the error icon
- Gives the user feedback as to how many of the email addresses that they've entered are already existing teammates
- Fixed a bug in `parseEmailAddressList` where `emailAddresses.parseAddressList()` would return null if any email was invalid. This would cause the error message to flash as the user typed. 

### Test

- [ ] Click invite to team and enter the email address of an existing teammate. A warning message should display.
- [ ] Enter the email addresses of 2/3 existing teammates. The warning message should update.
- [ ] Once the warning message has displayed, type invalid emails before or after and the warning message shouldn't flash


packages/client/utils/parseEmailAddressList.ts
```diff
@@ -11,7 +12,20 @@ const parseEmailAddressList = (rawStr = '') => {
     .trim()
     // remove trailing commas
     .replace(/,$/g, '')
-  return emailAddresses.parseAddressList(commaDelimStr)
+
+  const commaDelimArr = commaDelimStr.split(', ')
```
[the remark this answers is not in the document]
nickoferrall: True! I've removed the spaces and pushed the update. 

For the alternate solution, if we had a raw string of "validEmail1, invalidEmail, validEmail2" and it returned null, if we reduce the raw string by 1 character until it's no longer null, wouldn't we be left with only "validEmail1", i.e. we'd lose "validEmail2"?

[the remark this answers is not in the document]
mattkrick: that's correct, i think it's OK if we say something like "invalid email(s) after validEmail1" because if they goof on the syntax then it's possible that goof is what made validEmail2 valid. e.g. `matt@parabol.co,b@ b@b.co`. we can't guarantee that they meant `b@b.co`, maybe they meant bb@b.co?

ackernaut:
@nickoferrall do you need a review from me? I see Matt has made a few comments. Happy to do so.

nickoferrall:
@ackernaut I've added a new warning message colour, `WARNING_MAIN = '#FF9800'`, and added icons to the warning message. If you get the time, it'd be great to know if you feel it's a good fit for the design

ackernaut:
### Test cycle no. 1

- [ ] Click invite to team and enter the email address of an existing teammate. A warning message should display.
- [ ] Enter the email addresses of 2/3 existing teammates. The warning message should update.
- [ ] Once the warning message has displayed, type invalid emails before or after and the warning message shouldn't flash

packages/client/utils/parseEmailAddressList.ts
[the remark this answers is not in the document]
nickoferrall: Fair point! I've added the `invalid email(s) after validEmail1` feedback while preventing the jumping. It's ready for a re-review when you are

ackernaut:
I pushed 2 small changes:
- IMO there’s not enough of a case to deviate from the brand palette for the warning color, so keeping it simple.
- I scrubbed the `rem` values. The forward-motion deprecation pattern we use is to clean things up as you see them, and not go scrubbing for the sake of scrubbing.

ackernaut:
Thanks for the efforts here, and sorry for the delay on the review 👋

nickoferrall:
Thanks for the review Terry - will stick with pixels over rem from now on 👍 

@mattkrick, when you get the chance, it'd be great to your eyes on the functionality of this one to make sure that there's no red flags

packages/client/components/AddTeamMemberModal.tsx
```diff
@@ -98,34 +117,59 @@ const AddTeamMemberModal = (props: Props) => {
...
   const [rawInvitees, setRawInvitees] = useState('')
   const [invitees, setInvitees] = useState([] as string[])
   const {error, onCompleted, onError, submitMutation, submitting} = useMutationProps()
   const atmosphere = useAtmosphere()
   const onChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
+    if (isSubmitted) setIsSubmitted(false)
     const nextValue = e.target.value
     if (rawInvitees === nextValue) return
-    const parsedInvitees = parseEmailAddressList(nextValue)
+    const {parsedInvitees, invalidEmailExists} = parseEmailAddressList(nextValue) as any
```
mattkrick, line 129: it'd be good to do the typescript casting within the parseEmailAddressList instead of outside of it.
that way you can reuse the function anywhere & you still get type safety

packages/client/components/AddTeamMemberModal.tsx
```diff
@@ -98,34 +117,59 @@ const AddTeamMemberModal = (props: Props) => {
...
       ? (parsedInvitees.map((invitee) => (invitee as any).address) as string[])
       : invitees
     const teamEmailSet = new Set(teamMembers.map(({email}) => email))
     const uniqueInvitees = Array.from(new Set(allInvitees))
+    if (invalidEmailExists) {
+      const lastValidEmail = uniqueInvitees.pop()
+      lastValidEmail
+        ? onError(new Error(`Invalid email(s) after ${lastValidEmail}`))
+        : onError(new Error(`Invalid email(s)`))
+    } else if (!invalidEmailExists) {
```
mattkrick, line 140: can just be else

packages/client/utils/parseEmailAddressList.ts
```diff
@@ -11,7 +19,38 @@ const parseEmailAddressList = (rawStr = '') => {
...
+  if (emailCount === 1 && isAddingNewEmail) {
+    return {
+      parsedInvitees: emailAddresses.parseAddressList(commaDelimStr),
+      invalidEmailExists: false
+    }
+  }
+  const filteredStr = isAddingNewEmail ? filterLastEmail(commaDelimStr) : commaDelimStr
+  if (emailAddresses.parseAddressList(filteredStr)) {
+    return {
+      parsedInvitees: emailAddresses.parseAddressList(filteredStr),
```
mattkrick, line 36: no need to parse twice. parse it & then perform the conditional

packages/client/utils/parseEmailAddressList.ts
```diff
@@ -11,7 +19,38 @@ const parseEmailAddressList = (rawStr = '') => {
...
+      invalidEmailExists: false
+    }
+  } else {
+    for (let i = filteredStr.length; i >= 0; i--) {
+      const slicedStr = filteredStr.slice(0, i)
+      if (emailAddresses.parseAddressList(slicedStr)) {
```
mattkrick, line 42: should only parse once. work performed on keystroke will feel laggy on mobile, so we wanna keep it as cheap as possible

packages/client/utils/parseEmailAddressList.ts
```diff
@@ -11,7 +19,38 @@ const parseEmailAddressList = (rawStr = '') => {
...
+        return {
+          parsedInvitees: emailAddresses.parseAddressList(slicedStr),
+          invalidEmailExists: true
+        }
+      }
+    }
+    return {
+      parsedInvitees: emailAddresses.parseAddressList(filteredStr),
```
mattkrick, line 50: parse once

packages/client/utils/parseEmailAddressList.ts
```diff
@@ -1,7 +1,15 @@
 import emailAddresses from 'email-addresses'
 
+const filterLastEmail = (commaDelimStr: string) => {
+  const trimmedStr = commaDelimStr.trim()
+  const lastEmailIndex = trimmedStr.lastIndexOf(' ' || ',')
+  // get everything up to the final email and remove trailing commas
+  return trimmedStr.slice(0, lastEmailIndex).replace(/,$/g, '')
+}
+
 const parseEmailAddressList = (rawStr = '') => {
```
mattkrick, line 10: define the return type here so we don't have to cast as `any` when we use it

mattkrick:
@nickoferrall reviewed! just a few minor things. nothing to block pushing to prod, but a couple things to clean it up a bit as a 2nd pass

nickoferrall:
Thanks @mattkrick! I'll add these changes when I create a PR for the bulk email bug

Issue #4031, opened 2020-07-28
make enableSAMLForDomain an upsert instead of insert for domain


mattkrick:
AC
- [ ] if enableSAMLForDomain gets called twice for the same domain, the existing record should be updated.

Estimate: 1 hour

PR #4087, opened 2020-08-20
Make enableSAMLForDomain an upsert

nickoferrall:
Issue #4031 

This PR converts make `enableSAMLForDomain` to an upsert instead of an insert for a domain.

### DIFF
--- a/packages/server/database/migrations/20200828161500-samlDomainAsId.ts
+++ b/packages/server/database/migrations/20200828161500-samlDomainAsId.ts
@@ -1,16 +1,23 @@
 import {R} from 'rethinkdb-ts'
 export const up = async function(r: R) {
   try {
+    const updatedSaml = await r
+      .table('SAML')
+      .map((row) => {
+        return {
+          id: row('domain'),
+          domain: row('domain'),
+          url: row('url'),
+          metadata: row('metadata')
+        }
+      })
+      .run()
+
     await r
       .table('SAML')
-      .update((row) => ({
-        id: row('domain')
-      }))
+      .insert(updatedSaml, {conflict: 'replace'})
       .run()
-  } catch (e) {
-    console.log(e)
-  }
-  try {
+
     await r
       .table('SAML')
       .filter((row) => row('id').ne(row('domain')))
@@ -21,6 +28,24 @@ export const up = async function(r: R) {
   }
 }
 
-export const down = async function() {
-  // noop
+export const down = async function(r: R) {
+  try {
+    const updatedSaml = await r
+      .table('SAML')
+      .map((row) => {
+        return {
+          domain: row('domain'),
+          url: row('url'),
+          metadata: row('metadata')
+        }
+      })
+      .run()
+
+    await r
+      .table('SAML')
+      .insert(updatedSaml)
+      .run()
+  } catch (e) {
+    console.log(e)
+  }
 }
--- a/packages/server/database/migrations/20200828161500-samlDomainAsId.ts
+++ b/packages/server/database/migrations/20200828161500-samlDomainAsId.ts
@@ -1,7 +1,7 @@
 import {R} from 'rethinkdb-ts'
 export const up = async function(r: R) {
   try {
-    const updatedSaml = await r
+    const samlWithId = await r
       .table('SAML')
       .map((row) => {
         return {
@@ -15,7 +15,7 @@ export const up = async function(r: R) {
 
     await r
       .table('SAML')
-      .insert(updatedSaml, {conflict: 'replace'})
+      .insert(samlWithId, {conflict: 'replace'})
       .run()
 
     await r
@@ -30,7 +30,7 @@ export const up = async function(r: R) {
 
 export const down = async function(r: R) {
   try {
-    const updatedSaml = await r
+    const samlWithoutId = await r
       .table('SAML')
       .map((row) => {
         return {
@@ -43,7 +43,7 @@ export const down = async function(r: R) {
 
     await r
       .table('SAML')
-      .insert(updatedSaml)
+      .insert(samlWithoutId)
       .run()
   } catch (e) {
     console.log(e)
--- a/packages/server/database/migrations/20200828161500-samlDomainAsId.ts
+++ b/packages/server/database/migrations/20200828161500-samlDomainAsId.ts
@@ -1,0 +1,26 @@
+import {R} from 'rethinkdb-ts'
+
+export const up = async function(r: R) {
+  try {
+    await r
+      .table('SAML')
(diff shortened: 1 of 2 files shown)


packages/server/graphql/intranetSchema/mutations/enableSAMLForDomain.ts
```diff
@@ -19,14 +19,31 @@ const enableSAMLForDomain = {
...
-      })
-      .run()
+      .getAll(normalizedDomain, {index: 'domain'})
+      .count()
+      .run()) as number
+
+    if (samlDomainExists) {
+      await r
+        .table('SAML')
+        .update({
```
mattkrick, line 32: careful, this updates every single row on the table!
you want to first `.get` the row you want to update

packages/server/graphql/intranetSchema/mutations/enableSAMLForDomain.ts
```diff
@@ -19,14 +19,31 @@ const enableSAMLForDomain = {
...
       .table('SAML')
-      .insert({
-        domain: normalizedDomain,
-        url: normalizedUrl,
-        metadata: metadata
-      })
-      .run()
+      .getAll(normalizedDomain, {index: 'domain'})
+      .count()
+      .run()) as number
```
mattkrick, line 27: it looks like a boolean would be more expressive here, so you could add `.ne(0)`
or if you get the first record, then you can use that id to perform the update

packages/server/graphql/intranetSchema/mutations/enableSAMLForDomain.ts
```diff
@@ -19,15 +20,34 @@ const enableSAMLForDomain = {
...
+        .table('SAML')
+        .insert({
+          domain: normalizedDomain,
+          url: normalizedUrl,
+          metadata: metadata
+        })
+        .run()
+    }
+
     return true
```
tiffanyhan, line 51: Just wanted to point out the [difficulty of doing an atomic upsert](http://rethinkdb.1115947.n5.nabble.com/upsert-on-a-secondary-key-td983.html) in RethinkDB on a non primary key. I think it should be all right though since the saml table shouldn't be prone to race conditions.

mattkrick: that's very true.
what if we start fresh & have the `id` be the domain?
then, we could do a `.insert({id: domain, metadata, url}, {conflict: 'replace'})`

tiffanyhan: yup that's the other alternative i was thinking of! if the data needs to be cleaned anyway, which i'm assuming it is, it's probably worth it to start over again to get the guarantee and nicer API.

nickoferrall: Makes sense. I've pushed the update using the domain as the id

mattkrick: i like it! could you write a migration that:
- clones the existing rows in the DB that changes the ID to match the domain
- delete the rows where ID !== domain

nickoferrall: Sure! I've just pushed an update with a new migration.

A couple of small questions/considerations I'd like to ask:
1. With my current implementation, there will be a dormant `domain` column for existing data. SAML is a relatively small table so perhaps that's not an issue? Otherwise, I could drop the column.
2. I could add a migrate down function that would clear / remove the `id` column but I'm not sure that that would be particularly useful? My understanding is that if we were renaming a column, for example, we'd want the option to migrate back to the previous name in case it caused bugs in the app. As we're simply adding data to an empty column, migrating back doesn't seem necessary to me.

packages/server/database/migrations/20200828161500-samlDomainAsId.ts
```diff
@@ -0,0 +1,26 @@
+import {R} from 'rethinkdb-ts'
+export const up = async function(r: R) {
+  try {
+    await r
+      .table('SAML')
+      .update((row) => ({
+        id: row('domain')
```
mattkrick, line 7: did this work? 
this should produce an error, the primary key cannot be updated.
you'll need to clone the rows

packages/server/database/migrations/20200828161500-samlDomainAsId.ts
```diff
@@ -0,0 +1,26 @@
...
+      }))
+      .run()
+  } catch (e) {
+    console.log(e)
+  }
+  try {
+    await r
+      .table('SAML')
+      .filter((row) => row('id').ne(row('domain')))
```
mattkrick, line 16: we use the field `domain` so let's keep it for now. IMO a little duplicate data is worth the expressiveness

packages/server/database/migrations/20200828161500-samlDomainAsId.ts
```diff
@@ -0,0 +1,26 @@
...
+      .delete()
+      .run()
+  } catch (e) {
+    console.log(e)
+  }
+}
+
+export const down = async function() {
+  // noop
```
mattkrick, line 25: to migrate down, just clone the rows but don't set an ID so rethinkdb generates an ID on its own

packages/server/database/migrations/20200828161500-samlDomainAsId.ts
nickoferrall: Doesn't this only remove the rows where the ID !== domain? It doesn't remove the domain field?

packages/server/graphql/intranetSchema/mutations/enableSAMLForDomain.ts
```diff
@@ -19,13 +19,18 @@ const enableSAMLForDomain = {
...
       .table('SAML')
-      .insert({
-        domain: normalizedDomain,
-        url: normalizedUrl,
-        metadata: metadata
-      })
+      .insert(
+        {
+          id: normalizedDomain,
+          domain: normalizedDomain,
```
nickoferrall, line 28: @mattkrick as you mentioned that we're still using `domain`, I've re-added it to the mutation insert. This way we can still do an atomic upsert while keeping the domain field. Is this what you had in mind?
I've just pushed an update to the mutation and migration following your feedback

packages/server/database/migrations/20200828161500-samlDomainAsId.ts
mattkrick: ahh i think i imagined a `.without`. you're right

PR #4089, opened 2020-08-20
backupOrganization

### LINKED ISSUE #4083: slice DB by orgId
For local development purposes, we need a way to get some real data, but for security & privacy reasons (and the fact that it's freaking huge), we don't want devs to clone the entire DB.

I think the best way to do this is to have a function that takes a list of orgIds and returns the DB that only contains info pertinent to those orgIds. As a standard rule of practice, we can instruct devs to use the parabol orgId so they are only working with our own data. However, cloning other orgIds may be useful when debugging 3rd party issues, e.g. stripe.

so, i envision a JS script that is run inside the container. it creates a `backup` db. then, it goes through the production data & grabs all the records it needs & clones them to the backup db. then, it exports that backup db to a tarball. then, it deletes the backup db, leaving just the tarball.

AC
- [ ] i can run a remote script that takes in a list of orgIds and outputs a tarball containing a DB for just those orgIds. This includes slicing out data that references teams/users/etc that are not a part of the desired orgIds. For example, User.tms, which is an array of teamIds, should only contain teamIds that are under the desired orgIds. 
- [ ] can import the tarball locally & it works

Estimate: 16 hours

mattkrick:
For local dev, we don't need to clone the whole DB. we just wanna clone the records under the parabol org so we have some real data to play with.

fixes #4083 

documentation: https://www.notion.so/parabol/How-to-Get-Mock-Data-To-Develop-Locally-fc0b90425a3049a3aa21abeb7ca88f23

### DIFF
--- a/packages/server/graphql/intranetSchema/mutations/backupOrganization.ts
+++ b/packages/server/graphql/intranetSchema/mutations/backupOrganization.ts
@@ -1,0 +1,365 @@
+import {GraphQLID, GraphQLList, GraphQLNonNull, GraphQLString} from 'graphql'
+import getRethink from '../../../database/rethinkDriver'
+import {requireSU} from '../../../utils/authorization'
+import {GQLContext} from '../../graphql'
+
+const backupOrganization = {
+  type: GraphQLNonNull(GraphQLString),
+  description: 'copies all the records from RethinkDB for a list of organizations',
+  args: {
+    orgIds: {
+      type: GraphQLNonNull(GraphQLList(GraphQLNonNull(GraphQLID)))
+    }
+  },
+  resolve: async (_source, {orgIds}, {authToken}: GQLContext) => {
+    // AUTH
+    requireSU(authToken)
+
+    // RESOLUTION
+    const r = await getRethink()
+    const DESTINATION = 'orgBackup'
+
+    // create the DB
+    try {
+      await r.dbDrop(DESTINATION).run()
+    } catch (e) {
+      // db never existed. all good
+    }
+    await r.dbCreate(DESTINATION).run()
+    // create all the tables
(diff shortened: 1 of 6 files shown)


packages/server/graphql/intranetSchema/mutations/backupOrganization.ts
```diff
@@ -0,0 +1,365 @@
...
+      type: GraphQLNonNull(GraphQLList(GraphQLNonNull(GraphQLID)))
+    }
+  },
+  resolve: async (_source, {orgIds}, {authToken}: GQLContext) => {
+    // AUTH
+    requireSU(authToken)
+
+    // RESOLUTION
+    const r = await getRethink()
+    const DESTINATION = 'orgBackup'
```
mattkrick, line 20: hardcoding this name to reduce the chance that the prod DB gets mangled

scripts/toolboxSrc/postDeploy.ts
```diff
@@ -13,9 +13,11 @@ const flushSocketConnections = async () => {
 const storePersistedQueries = async () => {
   const queryMap = JSON.parse(fs.readFileSync(path.join(PROJECT_ROOT, 'queryMap.json')).toString())
   const hashes = Object.keys(queryMap)
+  const now = new Date()
   const records = hashes.map((hash) => ({
     id: hash,
-    query: queryMap[hash]
+    query: queryMap[hash],
+    createdAt: now
```
mattkrick, line 20: in the future, it'd be nice to purge these if they haven't been seen in awhile

mattkrick:
@nickoferrall this is live on staging. do you mind following the instructions in the notion doc & letting me know if it works for you?

PR #4103, opened 2020-08-24
Docker update

jordanh:
Continuing on from @ray-man 

Things this needs before merging:

- [ ] Dockerfile needs to build assets, not pull them from local filesystem
- [ ] Postdeploy hook needs to be managed

(maybe something else, too)


docker-compose.yml
```diff
@@ -1,9 +1,39 @@
...
+    ports:
+      - "6379:6379"
+    volumes:
+      - redis-data:/data
+    networks:
+      - parabol-network
+  app:
+    build:
+      context: .
+      dockerfile: ./docker/Dockerfile.dev
```
mattkrick, line 25: this doesn't look right, is this an old version? i thought we made this exclusively for prod

jordanh: I didn't, I left it as dev but editable for prod. At this phase, I don't really mind which way we go for now. This'll need one more refactor before it's really widely usable I think. Don't want to lose the work tho.

I'll update it to prod to make the intent more clear, then let's merge as to not lose the work?

Issue #4057, opened 2020-08-07
Sprint Poker template & scale UI implemented


ackernaut:
See #4184 for backend implementation

For concepts and notes, see #4015 and [this update in Notion](https://www.notion.so/parabol/Meeting-template-editing-807b69835881449e953708f2d9982a68) (w/ Loom and links to Figma).

#### Templates
- Teams can use or clone a single starting template
- Teams can create a new template
- The template has a single dimension
- The default name of the template and the single dimension is ‘Story Points’
- Folks can edit this value
- Folks can set the level of sharing of team-owned templates: team, org, public
- For mobile, the details view goes back to the list view
#### Scales
- Folks can select a default scale: Fibonacci, T-shirt Sizes, Five Fingers
- Default scales
  - Fibonacci: 0, 1, 2, 3, 5, 8, 13, 21, 34, ?, Pass (see default colors in Figma)
  - T-shirt Sizes: XS, S, M, L, XL, ?, Pass
  - Five Fingers, 1, 2, 3, 4, 5, ?, Pass
- ❓ and 🚫  are included in scales by default:
  - Folks can tap to exclude (boolean toggle)
  - Folks can tap the palette select to change the color of the card. Maybe disable this if the value is toggled off
- Folks can create their own scale
- A scale value option has a value field and a color
- A scale value option can be added, edited, and deleted
- The scale editing view goes back to the template details view
- TBD: can edit or clone existing scales?
#### In the future
- We’ll most likely add multiple dimensions
- We may add numerical value fields for semantic scale values
- We may add a description field for scale values: e.g. Fibonacci `3` means `2-3 days`
- We may add an optional template with a weighted formula for value-based and effort-based dimensions

mattkrick:
@tianrunhe 
On a scale, the `values` property should be an array of objects where each object is: `{color, label, value}`for example, on tshirt sizes, we'd have values = `[{color: 'blue', label: 'S', value: 1}, {color: 'green', label: 'M', value: 2}]`

ackernaut:
>❓ and 🚫  are included in scales by default:
>  - Folks can tap to exclude (boolean toggle)
>  - Folks can tap the palette select to change the color of the card. Maybe disable this if the value is toggled off

I just updated to this. I believe this is where we landed @mattkrick cc @tianrunhe 

Unlike the other values that have a char max of 2, these are simple booleans but folks can change the color

mattkrick:
lgtm!
question card value = 2^31 - 1
[...]

PR #4104, opened 2020-08-24
add poker to carousel

mattkrick:
This PR does the following:
- refactors `ReflectTemplate` table to `MeetingTemplate` since it is now shared with poker templates
- adds 2 public templates for the new poker meeting
- adds a `poker` featureFlag
- adds the poker meeting type if the feature flag is turned on

TEST
turn the feature on
```gql
mutation {
  addFeatureFlag(email: "@parabol.co", flag:poker) {
    error {
      message
    }
  }
}
``` 

- [x] Go to New meeting, scroll to the poker meeting. it looks good


mattkrick:
@ackernaut this does surprisingly little, but it lays some good groundwork

ackernaut:
Taking a card :)

mattkrick:
gonna start building some PRs on top of this. if there are any glaring bugs lemme know!

packages/server/database/migrations/20200824105351-pokerTemplate.ts
```diff
@@ -0,0 +1,200 @@
...
+    name: 'T-Shirt Sizes',
+    values: ['XS', 'SM', 'M', 'L', 'XL'],
+    teamId: 'aGhostTeam'
+  },
+  {
+    id: 'fiveFingersScale',
+    createdAt,
+    updatedAt: createdAt,
+    name: 'Five Fingers',
+    values: ['1', '2', '3', '4', '5'],
```
ackernaut, line 80: @mattkrick do we include `?` and `Pass` as dumb values so that they can remove or edit the color? Do we add a color palette value yet? Or is this really just stubbing it out?

mattkrick: mmmmm, both valid points.
i think we do need the ? and Pass. that way folks can remove them if they want.
we do need colors attached to those values.
we may need a "value" attached to the "label". so if e.g. someone renames "Pass" to their native language we still treat it as the pass card & it gets sorted to the bottom of the list.
maybe values is `{label, value, color}`
i'm tempted to keep it denormalized & just have this be an array of objects to avoid a whole new table just for values.

ackernaut: That sounds good, maybe this gets reworked in #4057

PR #4105, opened 2020-08-25
New user task view

### LINKED ISSUE #3933: Implement User filter for My Tasks
Acceptance Criteria:
- [ ] The Tasks view has a User filter that defaults to the current User (just like what the Team view shows today)
- [ ] The list of users is populated by all the users on all the teams that the viewer is a part of
- [ ] the filter is persisted in the URL for search params
- [ ] If the User filter is set & then the team filter changes, if the User filter includes a user that is not a part of the team filter, the user filter is reset

URL query param example: /tasks?users=user123&teams=foobarz&archived=true

Estimate: 30 hours

tianrunhe:
This resolves #3933.

*   [ ] By default, the list of team members in the user filter dropdown menu is populated by all the users on all the teams that the viewer is a part of
*   [ ] When user goes to `/me/tasks`, the user filter is selected as the viewer by default
*   [ ] When a team is selected in the team filter, the user filter only shows the team members in that team
*   [ ] When a user is selected in the user filter, the team filter only shows teams the user is in
*   [ ] Tasks in the kanban view are filtered according to team & user filter
*   [ ] Team filter, user filter and archived checkbox are persisted in the URL in the form of `/tasks?userId=user123&teamId=foobar&archived=true`
*   [ ] Task card footer displays team name only when a user is selected in user filter
*   [ ] Adding a new task corresponds to the user filter:
    *   [ ] If a user is selected, the new task will be created for that user
    *   [ ] If a user is selected, adding a new task will show team selector for the teams the user is in
    *   [ ] If a team is selected, the new task will be created for the viewer in that team
*   [ ] Moving/Updating tasks will be reflected for other users in real time
*   [ ] Moving/Updating tasks in team view will be reflected in user tasks view for other users


tianrunhe:
Hey @ackernaut , I'd like you to review/test the features first before I send it to Matt for implementation details. Thanks for your time!

ackernaut:
Good deal, taking a card

tianrunhe:
Known bugs:

*   Checking/unchecking the archived checkbox does not render the view until a force refresh

ackernaut:
### Test cycle no. 1

*   [ ] By default, the list of team members in the user filter dropdown menu is populated by all the users on all the teams that the viewer is a part of
*   [ ] When user goes to `/me/tasks`, the user filter is selected as the viewer by default
*   [ ] When a team is selected in the team filter, the user filter only shows the team members in that team
*   [ ] When a user is selected in the user filter, the team filter only shows teams the user is in
*   [ ] Tasks in the kanban view are filtered according to team & user filter
*   [ ] Team filter, user filter and archived checkbox are persisted in the URL in the form of `/tasks?userId=user123&teamId=foobar&archived=true`
*   [ ] Task card footer displays team name only when a user is selected in user filter
*   [ ] Adding a new task corresponds to the user filter:
    *   [ ] If a user is selected, the new task will be created for that user
    *   [ ] If a user is selected, adding a new task will show team selector for the teams the user is in
    *   [ ] If a team is selected, the new task will be created for the viewer in that team
*   [ ] Moving/Updating tasks will be reflected for other users in real time
*   [ ] Moving/Updating tasks in team view will be reflected in user tasks view for other users

ackernaut:
Hey @tianrunhe I made a first pass. This is going to be such a big win! I do have a few things that need attention:

- The 3 controls should be aligned `flext-start` in a 100% width container **_instead of_** using `space-between` in a fixed-width container. The container should allow for long names. The content is dynamic so it doesn’t matter if those controls change widths—there’s no need for a fixed layout. For laptop+ viewports, we could use more gutter between the 3 controls, maybe 3-4x grid (24-32px), but on mobile maybe it’s 2x (16px). The `flex-wrap` and `flex-shrink: 0` properties should help with wrapping on mobile if the controls are too wide. It’s okay for a control to wrap to a new line when necessary as long as it pushes down the column container below.

#### On desktop it pushes out of the fixed-width container
Screen Shot 2020-08-31 at 17 12 11

#### On mobile, controls are hidden if they don’t fit
Screen Shot 2020-08-31 at 17 12 38


### Changing teams on a card, the owner’s not on all my teams

Zach’s not on Marketing, but I am, and all of my teams are showing up in the bottom-left menu on the task card. It should only give me options for Zach’s teams that I am also on.



### Adding to any team with a single team filtered

The menu gives me the option to add to both teams I’m on but since I’m already filtered to a team I don’t see the task card when it’s created. The task should be set to the single team in the filter with me as the owner (easiest) or give me a menu of members of that team (I’d defer).

tianrunhe:
@ackernaut Thanks for the review. Fixes are in!

packages/client/components/DashFilterToggle/DashFilterToggle.tsx
```diff
@@ -21,8 +25,12 @@ const StyledIcon = styled(Icon)({
 const StyledLinkButton = styled(LinkButton)({
   color: PALETTE.TEXT_GRAY,
   fontWeight: 600,
+  marginRight: '16px',
```
ackernaut: While adding margins (which affect layout) here is safe because of the narrow scope of this component, I like to sometimes separate styles for a control from styles for the layout, so that they control can be used in many layouts. I’ll handle the layout in the containing component and context. One way is to make this component styleable by adding the className property (which I did in a separate commit). Another way is to use whatever containing element or wrapper to handle layout.

packages/client/modules/userDashboard/components/UserTasksHeader/UserTasksHeader.tsx
```diff
@@ -50,6 +50,8 @@ const StyledCheckbox = styled(Checkbox)({
 
 const UserTasksHeaderDashSectionControls = styled(DashSectionControls)({
   justifyContent: 'flex-start',
+  flexWrap: 'wrap',
+  flexShrink: 0,
```
ackernaut: `flexShrink` goes on the individual components that you don’t want to shrink, not their container. I moved this in a separate commit.

ackernaut:
The 2 fixes look great:

- [x] Changing teams on a card, the owner’s not on all my teams
- [x] Adding to any team with a single team filtered

I pushed the styles a little further and commented on why in the previous commit.

One thing to consider. If my teammate and I are only on 1 team together, then I can’t reassign their card to another team. We should make that control disabled* unless there are options:



Disabled:
- The cursor doesn’t change to a pointer
- The control doesn’t have a hover state or tooltip
- Clicking does nothing

ackernaut:
Adding @mattkrick for code review

tianrunhe:
> The 2 fixes look great:
> 
> *   [x] Changing teams on a card, the owner’s not on all my teams
> *   [x] Adding to any team with a single team filtered
> 
> I pushed the styles a little further and commented on why in the previous commit.
> 
> One thing to consider. If my teammate and I are only on 1 team together, then I can’t reassign their card to another team. We should make that control disabled\* unless there are options:
> 
> 
> 
> Disabled:
> 
> *   The cursor doesn’t change to a pointer
> *   The control doesn’t have a hover state or tooltip
> *   Clicking does nothing

Yeah I noticed this too. Given this is the behavior today in production, I didn't change it. It requires some effort from implementation perspective though hence I created #4129 to track it.

mattkrick:
so stoked to see this! will  code review tomorrow AM

tianrunhe:
> so stoked to see this! will  code review tomorrow AM 

Yeah I imagined we would go through a couple of iterations so thanks for your time in advance!

packages/client/modules/userDashboard/components/UserTasksHeader/UserTasksHeader.tsx
```diff
@@ -42,37 +65,83 @@ const StyledCheckbox = styled(Checkbox)({
...
+    originRef: teamMemberFilterOriginRef,
+    menuProps: teamMemberFilterMenuProps
+  } = useMenu(MenuPosition.UPPER_RIGHT, {
     isDropdown: true
   })
-  const {teamFilter, showArchivedTasksCheckbox} = viewer
-  const teamFilterName = (teamFilter && teamFilter.name) || 'All teams'
+  const {teams} = viewer
+  const teamMembers = teams.map(({teamMembers}) => teamMembers).flat()
+  const users = [...new Set(teamMembers.map(({user}) => user).flat())]
```
mattkrick, line 97: `Set` only works to dedupe things are are shallowly equal (e.g. `{} !== {}`). i don't think relay guarantees that it reuses the same user object, so you'll have to dedupe on ID instead of just object reference.

packages/client/modules/userDashboard/containers/UserColumns/UserColumnsContainer.tsx
```diff
@@ -24,33 +30,51 @@ const UserColumnsContainer = (props: Props) => {
...
   {
     const {
       viewer: {teams}
     } = props
 
-    return <TaskColumns area={AreaEnum.userDash} tasks={filteredTasks} teams={teams} />
+    // iff 1 user is selected, we show team names at the footer; otherwise we show task owner name
+    const areaForTaskCard = userIds && userIds.length === 1 ? AreaEnum.userDash : AreaEnum.teamDash
+    const filteredTeams = userIds ? teams.filter(({teamMembers, id: teamId}) => {
+      const inTeam = teamMembers.find(({userId}) => userIds.includes(userId)) != undefined
```
mattkrick, line 54: interesting that the linter didn't pick this up, we may have to revisit those rules.
`!=` works here for `undefined` and `null`, but in general, it's a pretty big footgun, so the JS community as a whole has moved away from it & only uses strict unequals `!==`. in this case, i'd probably just prefix with `!!`, but that's personal preference

packages/client/mutations/connections/getUserTasksConn.ts
```diff
@@ -1,6 +1,6 @@
 import {ConnectionHandler, RecordProxy} from 'relay-runtime'
 
-const getUserTasksConn = (viewer: RecordProxy, userIds: string[] | null, teamIds: string[] | null) =>
+const getUserTasksConn = (viewer: RecordProxy, userIds?: string[] | null, teamIds?: string[] | null) =>
```
mattkrick, line 3: this is tricky. i'm not sure if relay treats null as being equivalent to undefined. i don't think it does, which means it may not find the right connection

packages/client/utils/parseUserTaskFilters.ts
```diff
@@ -0,0 +1,14 @@
+import * as queryString from "query-string"
```
mattkrick, line 1: let's use the browser native handlers instead to keep the payload small

mattkrick (changes requested):
code review complete! will start digging into QA now

packages/client/components/MyDashboardTasks.tsx
```diff
@@ -17,20 +17,18 @@ interface Props {
...
+  const {userIds, teamIds, showArchived} = parseUserTaskFilters(atmosphere.viewerId)
+
   useStoreQueryRetry(retry)
   useDocumentTitle('My Tasks | Parabol', 'My Tasks')
-  useEffect(() => {
-    return () => filterTeam(atmosphere, null)
-  }, [])
   return (
     <>
       <UserTasksHeader viewer={viewer} />
```
mattkrick, line 28: if you promote this to a parent component then it won't flicker when you change the filter.
Alternatively, you can memoize the component, which should have the same effect

mattkrick (changes requested):
Things to fix:
- [ ] The filters flicker when they're clicked. We need to fix this
- [ ] I see this error in the console `react_devtools_backend.js:2273 Warning: createFragmentSpecResolver: Expected prop `team` to be supplied to `Relay(TeamArchive)`, but got `undefined`. Pass an explicit `null` if this is intentional. `
- [ ] The archive view does not paginate. When i scroll to the bottom, i expect to see the next batch load. Note: this bug made it into production already, so we do't strictly need to fix it here, but it'd be nice to get it working again.
- [ ] when i goto tasks & click the archive checkbox & then uncheck the archive checkbox a new network request is made. this request should be cached (that's how it works in prod today)

tianrunhe:
@mattkrick I'm not sure how to make the filters not flicker? When any filter is changed, it will trigger a redraw of both of the filters, right?

packages/client/components/ArchiveTaskRoot.tsx
```diff
@@ -4,18 +4,34 @@ import {QueryRenderer} from 'react-relay'
...
+  if (error) {
+    return <ErrorComponent error={error} eventId={''} />
+  }
+  if (!props) {
+    return <UserTasksHeader viewer={null} />
+  }
+  return (
+    <>
+      <UserTasksHeader viewer={props.viewer} />
+      <TeamArchive viewer={props?.viewer ?? null} returnToTeamId={props?.returnToTeamId} team={props?.team} />
```
mattkrick, line 30: under what condition is `team` not undefined here?
hint: move this inside the functional component, you're not passing team in where you should be

packages/client/types/constEnums.ts
```diff
@@ -129,7 +129,7 @@ export const enum Layout {
   LAYOUT_GRID = 8, // 1x
   ROW_GUTTER = 16, // 2x
   SETTINGS_MAX_WIDTH = 768,
-  TASK_COLUMNS_MAX_WIDTH = 1304 // (4 x 296 card max-width) + (5 x 24 - gutters around cols)
+  TASK_COLUMNS_MAX_WIDTH = 1360
```
mattkrick, line 132: curious, what's the reason for the change? what's the calculation?

mattkrick (changes requested):

packages/client/types/constEnums.ts
tianrunhe: When the width was 1304, I found the Done column did not fit:



The number 1360 was calculated as follows:

1.  Card width: 296
2.  Left and right padding of 12px defined on [DraggaleTask](https://github.com/ParabolInc/parabol/blob/master/packages/client/containers/TaskCard/DraggableTask.tsx#L13)
3.  Somehow the column width has 15px addition which I couldn't find reference.

Therefore (296 + 12 + 12 + 15) \* 4 = 1340 for 4 columns. Finally 10px left/right padding defined on [TaskColumns](https://github.com/ParabolInc/parabol/blob/master/packages/client/components/TaskColumns/TaskColumns.tsx#L26). Hence 1360.

Issue #4090, opened 2020-08-20
Bulk email invitation


jordanh:
## Issue - Bug

[User](https://app.hubspot.com/contacts/3888472/contact/7270101) 🔒 :

1. Opens a new Outlook message
2. Types in an email list alias
3. Expands the alias to (51) contacts
4. Select all and copy, paste into Parabol invite form
5. Manually edits to change format from "Lname, Fname <user@domain.com>; ..." to "user@domain.com, ..."
6. Hits send
7. Sees error text under the "Send 51 invitations" button

Steps to try and reproduce on our side:

1. Generate a list of 51-ish users in the form "user+bugtest1@parabol.co, user+bugtest2@parabol.co, ..."
2. Paste and try to send

My guess is we're getting throttled by mailgun and presently don't have a way to queue/retry requests.

### Acceptance Criteria (optional)
   - Triage and open new issue to fix
   - Consider enhancing parsing to allow for Outlook's "Lname, Fname <user@domain.com>;..." semi-colon delimited format
   - Communicate with user on findings an expections

Estimated effort: 4 hours

tiffanyhan:
Copy and pasted directly from email list in Outlook. Send button is disabled at first because of formatting. She manually removes last name, first name, semicolons, so that input only contains a list of comma separated emails. After that, the send button is enabled. Upon clicking "Send" she gets an error message. However, emails actually went through which she discovered later from talking to invitees.

mattkrick:
i don't think this is a mailgun throttling issue, if it were, we'd see an error in sentry.
@jordanh if i'm wrong, please include an email so we can dig into it further!

PR #4131, opened 2020-09-03
Bulk email invite

nickoferrall:
This PR is for issue #4090.

As discussed with Matt, I initially tested sending 50+ emails and while they were being sent successfully, the frontend wasn't updating - it looked like the `onComplete` function wasn't being triggered in the client. 

Having made no changes to the functionality, it is now working.. Perhaps it was a problem on Mailgun's end. I've tested sending 50+ & 100+ emails several times and it's now working without any problems.

This PR includes:

- Adding overflow for the successful model to handle bulk invites
- Updated warning colour
- Cleaned up the `parseEmailAddressList` function, updating the types
- Fixed bug where removing all emails wouldn't remove them from state and button would still say "Send 3 Invitations"
- Using `parseOneAddress` on the last email to check if the user is still typing. This is less performant but I think it's the most reliable way of checking whether the user is still typing and therefore we shouldn't show a warning message

@ackernaut you mentioned [here](https://github.com/ParabolInc/parabol/pull/4082#issuecomment-682843374) that you didn't want to deviate from the brand palette for the warning colour. My only concern is that #FD6157 is a reddish colour and there was a user tension #4038 around differentiating between a warning vs error. In this PR I've switched the warning colour to brand palette orange instead - what do you think?

@mattkrick it would be great to get your eyes on the functionality. The issue mentions "Communicate with user on findings". Shall I send the user an email updating them?


packages/client/utils/parseEmailAddressList.ts
```diff
@@ -21,33 +13,45 @@ const parseEmailAddressList = (rawStr = '') => {
     .replace(/,$/g, '')
 
   const emailCount = commaDelimStr.split(',').length
-  const lastChar = rawStr.charAt(rawStr.length - 1)
-  const isAddingNewEmail = lastChar !== ' ' && lastChar !== ','
+  const lastEmailIndex = commaDelimStr.lastIndexOf(' ' || ',')
```
mattkrick, line 16: this doesn't do what i think you want it to. `' '` is always truthy so it'll never fall back to `','`

mattkrick (changes requested):
small code bug, other than that i'll let @ackernaut give the thumbs up on style & UX

packages/client/utils/parseEmailAddressList.ts
nickoferrall: Ah yes, thanks - just pushed an update

ackernaut (approved):
That looks like it’s the retro palette color, but that works for me to distinguish warnings from error.

nickoferrall:
OK great, thanks @ackernaut.

@mattkrick, Terry's given it the thumbs up. Happy for me to merge it into master or would you like to review the update?

packages/client/utils/parseEmailAddressList.ts
```diff
@@ -13,7 +13,7 @@ const parseEmailAddressList = (rawStr = ''): {parsedInvitees: any; invalidEmailE
     .replace(/,$/g, '')
 
   const emailCount = commaDelimStr.split(',').length
-  const lastEmailIndex = commaDelimStr.lastIndexOf(' ' || ',')
+  const lastEmailIndex = Math.max(commaDelimStr.lastIndexOf(' '), commaDelimStr.lastIndexOf(','))
```
mattkrick, line 16: are you sure this works? if both are -1 then it will slice from -1, which is the last char of the string

nickoferrall: good point! I've updated the Math.max to make sure that we slice from the beginning in that case

PR #4141, opened 2020-09-08
Add empty state to ScopePhase AddJira 

mattkrick:
TEST
- [x] can start a poker meeting (behind feature flag)
- [x] When the viewer hits the scope phase, the "Add Jira" tab is selected
- [ ] When the viewer clicks "Import stories from Jira" it opens an Atlassian OAuth window
- [ ] After successful OAuth, the button opens a menu of projects available to the viewer. That list excludes any projects that already have their own tab
- [ ] on the "Add Jira" tab, if OAuth has already been completed, then it just opens the project list menu 
- [ ] Clicking a project creates a new tab for that project & autofocuses it. There is no content (todo in next PR)
- [ ] a more_vert icon exists on each tab. if it's clicked, a menu opens. that menu has a "remove project" item. when clicked, the tab is removed


mattkrick:
merging early to fix the `EndDraggingReflectionPayload` bug

Issue #4080, opened 2020-08-18
Implement multi-connection User Presence


mattkrick:
This issue is to implement the work as speced in #4060

AC
- [ ] The architecture designed by #4060 is implemented
- [ ] User can join a meeting, open a 2nd tab on the dashboard & still show up as present in the meeting
- [ ] When a new version launches, the User still appears to be connected to other team members. When they log off shortly after the new version launch, they show up as offline
- [ ] autopausing users still works (lastSeenAt still lives in rethinkdb)

Estimate: 16 hours

jordanh:
Groomed

Issue #4060, opened 2020-08-07
Architect new User Presence


mattkrick:
User Presence is a hugely importance piece to our app.
- It's the logic that shows your avatar in the meeting when you're online
- it prevents you from getting a notification that someone created a task for you when you're in the meeting with them
- it's how we know you've been active in the app & we shouldn't put your account on pause

Today, we handle user presence by storing 3 fields on the `User` object in RethinkDB: `connectedSockets, lastSeenAt, lastSeenAtURL`. When a user connects, a socketId is added to the `connectedSockets` array. When they disconnect, that socketId is removed. As a user moves throughout the app, `lastSeenAt, lastSeenAtURL` is updated.

This strategy is bad for the following reasons:
- Storing these fields in RethinkDB means the User object is constantly changing & busting the cache. This is inefficient
- `lastSeenAtURL` is a single value, which means if the user has 2 tabs open, it is last-write-wins. This means if tab A is in the meeting & tab B is on the dashboard, the user will appear to not be in the meeting because tab B set `lastSeenAtURL` to the dashboard
- It doesn't scale horizontally. When we have n servers & we scale down due to a crash or a deployment of a new version, we'll want the connectedSockets array to be updated. For example, when 1 server deploys the newest version, it will have to restart & kick off all connected clients. Those clients will automatically reconnect to a new server, but how do we modify the connectedSockets array?

A good solution will have the following characteristics.

- [ ] When a user has multiple tabs open & 1 is the meeting, it looks like they are in that meeting
- [ ] The `User` document in RethinkDB does not contain lastSeenAtURL or connectedSockets (lastSeenAt is probably still necessary to determine if a user is inactive)
- [ ] An eviction policy is created to guarantee that server downscaling does not leave some users marked as online. 
- [ ] When a user logs in, we can cheaply notify all team members that they are now online

Possible solutions for some inspiration:
- Today, we cache the User object in redis. maybe before each cache we add in these fields?
- Maybe we have a UserPresence key/value lookup in redis that just keeps these fields?
- Maybe redis stores a list of userIds that are connected to a particular node. When that node goes does, we can loop over each userId & remove that socketId from their connectedSockets list?
- Maybe we just use a TTL on a UserPresence record in redis that gets reset every time the client replies to the server ping? (would that be too expensive?)
- Maybe each team has an array of online userIds? See #3259

Acceptance Criteria:
- [ ] A plan with a path forward on what to build & the pros/cons of the choices made

Estimate: 20 hours

[...]

PR #4147, opened 2020-09-10
Multi-connection user presence with Redis

### LINKED ISSUE #3637: support multiple lastSeenAtURL locations in case 2 tabs are open
Repro:
- Join a meeting
- in a 2nd tab, go to dash
- see that you are not part of the meeting 

AC
- above repro fixed
- notifications for new tasks not created for users who are in the meeting

Effort: 11 points

nickoferrall:
Issue #4080 

This PR stores user presence, online users within a team and online users in Redis.

This also fixes #3637 

### AC

- [ ]  The architecture designed by #4060 is implemented
- [ ]  User can join a meeting, open a 2nd tab on the dashboard & still show up as present in the meeting
- [ ]  When a new version launches, the User still appears to be connected to other team members. When they log off shortly after the new version launch, they show up as offline
- [ ]  autopausing users still works (lastSeenAt still lives in rethinkdb)


scripts/toolboxSrc/postDeploy.ts
```diff
@@ -6,10 +6,6 @@ import getProjectRoot from '../webpack/utils/getProjectRoot'
 
 const PROJECT_ROOT = getProjectRoot()
 
-const flushSocketConnections = async () => {
```
nickoferrall, line 9: As we're no longer storing `connectedSockets` in User, I've removed this. We could flush all connected sockets in Redis but as this is already happening in disconnect, I don't think that's necessary?

packages/client/components/MeetingRoot.tsx
```diff
@@ -7,7 +7,6 @@ import renderQuery from '../utils/relay/renderQuery'
 import MeetingSelector from './MeetingSelector'
 import SetAppLocationMutation from '~/mutations/SetAppLocationMutation'
 
-// Changing the name here requires a change to getLastSeenAtURL.ts
```
nickoferrall, line 10: We have a few of these comments in the app but it looks like the `getLastSeenAtURL.ts` isn't being used anywhere so I've removed these comments and the file

packages/server/graphql/intranetSchema/mutations/addNewFeature.ts
```diff
@@ -43,14 +45,7 @@ const addNewFeature = {
...
 
-    const onlineUserIds = await r
-      .table('User')
-      .filter((user) =>
-        user('connectedSockets')
-          .count()
-          .ge(1)
-      )('id')
-      .run()
+    const onlineUserIds = await redis.smembers('onlineUserIds')
```
nickoferrall, line 48: The original proposal in Notion was to include User Presence and online users in a team in Redis but I thought it'd also be helpful to store all online users so that we can access them in `O(N)` here. 

Alternate approaches I considered were:
1. Getting all of the `teamIds` from RethinkDb and then getting all of the online users from the `team:${teamId}` but this hits both the db and Redis
2. Use [KEYS](https://redis.io/commands/keys) to get all online user keys matching a pattern, i.e. `presence`. This would be a bit of a hack and Redis don't advise it in production. 

Taking up a bit more space in Redis seems like the right approach to me, particularly as we might want to access all `onlineUserIds` elsewhere in the app in the future.

packages/server/graphql/intranetSchema/mutations/disconnectSocket.ts
```diff
@@ -1,50 +1,50 @@
...
+      throw new Error('Called disconnect without a valid socket')
+    }
+    await redis.lrem(`presence:${userId}`, 0, JSON.stringify(disconnectingSocket))
+
+    // If this is the last socket, tell everyone they're offline
+    if (userPresence.length === 1) {
+      const listeningUserIds = new Set()
+      await redis.srem('onlineUserIds', userId)
+      for (const teamId of tms) {
+        const teamMembers = await redis.smembers(`team:${teamId}`)
```
mattkrick, line 38: multi.exec here, too.
could refactor into a helper function like `getListeningUserIds(tms, commands)` where it calls a couple redis commands and returns an array of listeningUserIds

packages/server/graphql/intranetSchema/mutations/disconnectSocket.ts
```diff
@@ -58,11 +58,11 @@ export default {
       userId,
       event: 'Disconnect WebSocket',
       properties: {
-        connectedSockets,
+        connectedSockets: userPresence.map((socket) => JSON.parse(socket).socketId),
```
mattkrick, line 61: same as connect, let's just return a count

packages/server/graphql/mutations/helpers/getUsersToIgnore.ts
```diff
@@ -1,24 +1,23 @@
...
-        .get(teamMember('userId'))('lastSeenAtURL')
-        .default('~')
-    }))
-    .filter({lastSeenAtURL})('userId')
-    .run() as unknown) as string[]
+const getUsersToIgnore = async (viewerId: string, teamId: string) => {
+  const redis = getRedis()
+  const userPresence = await redis.lrange(`presence:${viewerId}`, 0, -1)
+  const userLastSeenAtURLs = userPresence.map((socket) => JSON.parse(socket).lastSeenAtURL)
+  const activeTeamMemberIds = await redis.smembers(`team:${teamId}`)
```
mattkrick, line 8: can join both redis calls into 1 async request

packages/server/graphql/mutations/helpers/getUsersToIgnore.ts
```diff
@@ -1,24 +1,23 @@
...
-        .table('User')
-        .get(teamMember('userId'))('lastSeenAtURL')
-        .default('~')
-    }))
-    .filter({lastSeenAtURL})('userId')
-    .run() as unknown) as string[]
+const getUsersToIgnore = async (viewerId: string, teamId: string) => {
+  const redis = getRedis()
+  const userPresence = await redis.lrange(`presence:${viewerId}`, 0, -1)
+  const userLastSeenAtURLs = userPresence.map((socket) => JSON.parse(socket).lastSeenAtURL)
```
mattkrick, line 7: call JSON.parse once & then this & L13 can both map over that parsed array

packages/server/graphql/mutations/setAppLocation.ts
```diff
@@ -22,27 +24,39 @@ export default {
...
+    }
 
     // RESOLUTION
-    const {lastSeenAtURL} = viewer
-    const lastSeenAt = new Date()
+    const parsedConnectedSocket = JSON.parse(connectedSocket) as UserPresence
+    const {lastSeenAtURL} = parsedConnectedSocket
+    const {lastSeenAt} = viewer
+    const now = new Date()
+    const datesAreOnSameDay = now.toDateString() === lastSeenAt.toDateString()
```
mattkrick, line 46: i like this thought!
i think doing this once in `connectSocket` probably suffices.

packages/server/graphql/mutations/setAppLocation.ts
```diff
@@ -22,27 +24,39 @@ export default {
...
+    if (!datesAreOnSameDay) {
+      await db.write('User', viewerId, {lastSeenAt: now})
+    }
     const data = {userId: viewerId}
     if (lastSeenAtURL !== location) {
-      await db.write('User', viewerId, {lastSeenAt, lastSeenAtURL: location})
+      parsedConnectedSocket.lastSeenAtURL = location
+      await redis.lrem(`presence:${viewerId}`, 0, connectedSocket)
+      await redis.rpush(`presence:${viewerId}`, JSON.stringify(parsedConnectedSocket))
```
mattkrick, line 54: can do both in parallel

packages/server/graphql/types/User.ts
```diff
@@ -148,9 +145,10 @@ const User = new GraphQLObjectType<any, GQLContext, any>({
     isConnected: {
       type: GraphQLBoolean,
       description: 'true if the user is currently online',
-      resolve: (source) => {
-        const {connectedSockets} = source
-        return Array.isArray(connectedSockets) && connectedSockets.length > 0
+      resolve: async ({id: userId}) => {
+        const redis = getRedis()
+        const userPresence = await redis.lrange(`presence:${userId}`, 0, -1)
```
mattkrick, line 150: `LLEN` is a little cheaper since we don't need to read the values, just the length.

scripts/toolboxSrc/postDeploy.ts
mattkrick: what happens when we deploy a new version? the old server shuts down without callings disconnect.

mattkrick (changes requested):

packages/server/graphql/intranetSchema/mutations/connectSocket.ts
```diff
@@ -50,26 +45,40 @@ export default {
       // TODO: re-identify
     }
 
-    if (connectedSockets.length === 1) {
+    const datesAreOnSameDay = lastSeenAt && now.toDateString() === lastSeenAt.toDateString()
```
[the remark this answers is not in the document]
nickoferrall: Interesting! Thanks for the repro steps. Do we need to rehydrate all `Date` fields in the `User` object if we're just using lastSeenAt? Also, if we update the User object directly, do we bust the cache or is that just if we update RethinkDB? 

I've pushed an update converting `lastSeenAt` to a `Date` but I can update all `Date` fields in a handler if that makes more sense

packages/server/graphql/mutations/helpers/getUsersToIgnore.ts
```diff
@@ -1,24 +1,42 @@
...
+  let userPresence
+  let activeTeamMemberIds
+  await redis
+    .multi()
+    .lrange(`presence:${viewerId}`, 0, -1)
+    .smembers(`team:${teamId}`)
+    .exec((execErr, results) => {
+      if (execErr) throw new Error(`Failed to execute redis command: ${execErr}`)
+      results.forEach((res, index) => {
+        if (index === 0 && !res[0]) {
```
nickoferrall, line 16: This looks a bit ugly to me but `results` are an array of `[err, response]` so I think these checks are necessary. I haven't found any cleaner examples of reading from `exec` after many a Google.

packages/server/utils/getListeningUserIds.ts
```diff
@@ -0,0 +1,33 @@
...
+const getListeningUserIds = async (command: RedisCommand, tms: string[], userId: string) => {
+  const redis = getRedis()
+  const listeningUserIds = new Set()
+  for (const teamId of tms) {
+    const commands = [
+      [command, `team:${teamId}`, userId],
+      ['smembers', `team:${teamId}`]
+    ]
+    let teamMembers
+    await redis.multi(commands).exec((execErr, results) => {
```
mattkrick, line 17: let's get that `await` outside of the synchronous loop so we only make 1 call

scripts/toolboxSrc/postDeploy.ts
```diff
@@ -1,10 +1,35 @@
 import fs from 'fs'
 import path from 'path'
 import getRethink from '../../packages/server/database/rethinkDriver'
-import db from '../../packages/server/db'
 import getProjectRoot from '../webpack/utils/getProjectRoot'
+import getRedis from '../../packages/server/utils/getRedis'
 
 const PROJECT_ROOT = getProjectRoot()
+const redis = getRedis()
```
mattkrick, line 8: nit: doing this here opens the connection pool when the file is imported vs. when a function is called.
ideally, files are side-effect free, but functions do things (the exception being singletons)

packages/server/graphql/intranetSchema/mutations/connectSocket.ts
```diff
@@ -24,20 +33,10 @@ export default {
...
-        .default([])
-        .append(socketId)
-    })
-    await db.write('User', userId, reqlUpdater)
-    const {inactive, tms} = user
-    const connectedSockets = user.connectedSockets || []
-    connectedSockets.push(socketId)
+    const userDb = await db.read('User', userId)
+    // hydrate user as lastSeenAt could be a string if we're getting it from Redis
+    const user = hydrateRedisDoc(JSON.stringify(userDb), 'user')
```
mattkrick, line 38: change hydrateRedisDocs to a `User` instead of `user` and we get this for free

mattkrick: can remove this explicit call

packages/server/graphql/mutations/setAppLocation.ts
```diff
@@ -22,27 +23,35 @@ export default {
   })(async (_source, {location}, {authToken, dataLoader, socketId: mutatorId}) => {
     const operationId = dataLoader.share()
     const subOptions = {mutatorId, operationId}
+    const redis = getRedis()
```
mattkrick, line 26:
```js
if (!mutatorId) {
  sendToSentry(new Error('No mut id on setAppLoc'))
}
```

packages/server/utils/getListeningUserIds.ts
```diff
@@ -0,0 +1,33 @@
...
+    commands.push([command, `team:${teamId}`, userId], ['smembers', `team:${teamId}`])
+  }
+  const responses = await redis.multi(commands).exec((execErr) => {
+    if (execErr) {
+      throw new Error(`Failed to execute redis command: ${execErr}`)
+    }
+  })
+  const membersOfAllTeams = [] as string[][]
+  responses.forEach((res, index) => {
+    if (index % 2 !== 0) {
```
mattkrick, line 22: add comments showing that this is going after the `smembers` comamnd results

packages/server/utils/getListeningUserIds.ts
```diff
@@ -0,0 +1,33 @@
...
+      const teamMembers = res[1] as string[]
+      membersOfAllTeams.push(teamMembers)
```
mattkrick, line 24: membersOfAllTeams.push(...teamMembers)

packages/server/utils/getListeningUserIds.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+  })
+  for (const teamMemberUserId of membersOfAllTeams.flat()) {
```
mattkrick, line 27: can remove flat() here if above is implemented

scripts/toolboxSrc/postDeploy.ts
```diff
@@ -1,13 +1,34 @@
...
-  return db.writeTable('User', {connectedSockets: []})
+  const redis = getRedis()
+  const userPresenceStream = redis.scanStream({match: 'presence:*'})
+  userPresenceStream.on('data', async (keys) => {
+    if (!keys?.length) return
+    userPresenceStream.pause()
+    const writes = keys.map((key) => {
+      return ['del', key]
+    })
+    await redis.multi(writes).exec()
```
mattkrick, line 18: lets get rid of the await as well as the pause/resume. we'll know it finished up when the stream `end` fires.

nickoferrall:
thanks for the review @mattkrick, I've pushed those updates 👍

mattkrick:
code looks good! @ackernaut would you mind kicking the tires on this one when it lands on staging?
things to look for:
- connection is correct when we deploy a new version (postDeploy hook)
- i can be in a meeting & on the dashboard & i still show up in the meeting (my avatar)

ackernaut:
@mattkrick when will this land on staging? For release this week?

mattkrick:
yep! i think this & the user task filters are the big ones to land this week

ackernaut:
@mattkrick 
>connection is correct when we deploy a new version (postDeploy hook)

I assume I’d have to be connected to a view when the new version lands to test this?

>i can be in a meeting & on the dashboard & i still show up in the meeting (my avatar)

This seems to be working as intended on the RC for v5.20.0

mattkrick:
yep! best way to test it is to be connected to staging with 2 clients
then, ssh into staging & run `dokku ps:restart web`
then watch for each of the 2 clients connectivities. they should still both be logged in. When one disconnects, it should go gray for the other

ackernaut:
@mattkrick I ran `dokku ps:restart web` on staging and both users stayed green light 🟢  on the team dash in their respective browser clients

ackernaut:
Yes, tried the gray light just now

Issue #4004, opened 2020-07-21
Build Sprint Poker Jira Backend integration v1


mattkrick:
Sprint poker is only useful if teams can get their stories into parabol to be processed. We think most stories live in Jira, so we start there.

Jira's IA begins with an `Issue`. An issue can be a story, bug, task (or subtask). Let's start there for now, and probably default to story when we have to choose. 

In Parabol, Tasks live in Teams which live in Organizations
In Jira, Issues live in Projects which live in Clouds

AC
- [ ] in `AtlassianManager.tsx` there exists a `getIssuesByProjectId(cloudId: string, projectKey: string, issueTypes?: string[]): {issues: JiraIssue[], hasNextPage: boolean}`. The array is sorted by most recent first. Note this is not natively supported by Jira, so you'll need to play with it (see https://community.atlassian.com/t5/Jira-questions/REST-Client-get-all-issues-of-project/qaq-p/494826). This should only return 100 issues tops. if more than 100 are returned, then hasNextPage should be true. we will use that flag to either trigger another fetch or to say "too many issues! please refine your query"
- [ ] can accept JQL and return a list of issues, e.g. `getJQL(cloudId: string, projectKey: string, jql: string)`. Ideally, this only performs a query, no mutations.
- [ ] there's a method called `updateIssueStoryPoints(cloudId: string, projectKey: string, issueKey: string, storyPoints: number)` This includes as much metadata & references to parabol as reasonable. See https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issues/#api-rest-api-3-issue-issueidorkey-put
- [ ] ~~there is a method called `getJQLAutocomplete(cloudId: string, projectKey: string, jql: string): JQL[]` it returns up to 5 JQL autocomplete options that we will use to put in a dropdown helper menu. On the frontend, it will be throttled every 50ms, _do not_ throttle it in the AtlassianManager. See https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-jql/#api-rest-api-3-jql-autocompletedata-get~~ this is out of scope for now

Effort: 30 hours

mattkrick:
assigning myself, this is blocking the scope phase work

mattkrick:
goodness gracious the Jira API isn't friendly.
the only way to get more than 6 results is to transform a query into JQL.
unfortunately, JQL doesn't do per-character search, it _only_ returns results per word (e.g. foo won't return results for food)
[...]

PR #4158, opened 2020-09-14
Poker jira front-end

mattkrick:
This wires up the front end of the Jira Integration issue #4004. 
This involved a large-ish refactor of the integration auth from the `User` to the `TeamMember`.
This is because when fetching a meeting, the `teamId` is indeterminate because the teamId is not part of the route.
To avoid a 2nd round trip, we put the auth on `TeamMember` object.
This allows us to write graphql like `meetingMember { teamMember { atlassianAuth { accessToken } } }`.
Instead of: `User { atlassianAuth(teamId) { accessToken} }`, which would require a `teamId`.
Lesson learned: when deciding where an entity lives in the graphql schema, pick the place where it'll have the fewest args.

AC:
- [ ] When the viewer clicks "Import stories from Jira" it opens an Atlassian OAuth window
- [ ] After successful OAuth, the button the button is replaced with a list of issues that is populated based on the search query
- [ ] JiraScopingSearchResults has some hardcoded variables. if loading is set to true, you see some mock items
- [ ] if results has 0 lengrth, you see a message saying no results
- [ ] if results is length 2, you see a select all
- [ ] if results is length 1 (default) you see a single item & no select all
- [ ] typing is a search query on the search input makes an X appear
- [ ] clicking the x on the search query input clears the input

Regression tests:
- [ ] can push a task to jira on the team kanban
- [ ] can push a task to github on the team kanban

### DIFF
--- a/packages/client/clientSchema.graphql
+++ b/packages/client/clientSchema.graphql
@@ -16,6 +16,7 @@ extend type PokerMeeting {
   localPhase: NewMeetingPhase!
   localStage: NewMeetingStage!
   showSidebar: Boolean!
+  jiraSearchQuery: String
 }
 
 extend type RetrospectiveMeeting {
@@ -82,5 +83,9 @@ extend type UpdatesStage {
 }
 
 extend type AgendaItemsStage {
+  localScheduledEndTime: String
+}
+
+extend type EstimateStage {
   localScheduledEndTime: String
 }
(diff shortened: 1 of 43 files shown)


mattkrick:
merging. no big concerns for testing because this is all behind the poker feature flag

Issue #4115, opened 2020-08-27
Support tasks without assignees


mattkrick:
This is blocking sprint poker SCOPE phase

A story is just a task without an assignee.
to keep our IA nice & tight, we should reflect that.

questions
- should a task without an assignee appear on the kanban?
- if so, we need a "UNASSIGNED" magic keyword as a filter criterion
- if so, do those stories get added after they've been added to the estimate phase? or after they have been assigned a point value?
- if we include integrated tasks (eg jira) then when the story gets assigned in jira, we'll have to update the assignee in parabol. if we do that, we need to create a jira user -> parabol user mapper & have a fallback if no parabol user exists for an assignation

this seems right, but it seems like a lot of work.
a simpler approach is to just exclude tasks w/o assignees from the kanban

AC
- [ ] the `Task` object is refactored to support null userId
- [ ] all task mutations are updated to support a Task with a nullable assignee (e.g. createTask, updateTask, deleteTask, etc etc)
- [ ] A Task card shows a question mark as the assignee if it doesn't have an assignee
- [ ] The Task assignee dropdown in the Task card footer includes the current assignee's name. if already selected & clicked again, the task becomes unassigned
- [ ] the `tasks` query returns only tasks that have assignees

Estimate: 16 hours

ackernaut:
Some of my (non-clarifying, doh) questions:
- How often folks are estimating something that has been assigned in Jira or not?
- Do folks really expect Jira tasks to be tasks matched to the Parabol team and show up in the kanban query? It seems like the most likely trend is for folks to handle team meta work as Parabol cards but wouldn’t be inclined to create and/or sync tasks that really live somewhere else. I think the value add is more around showing a test of integrated tasks in the context of an activity rather than cluttering up a one-size-fits-all board for the team.
- I’m wondering if we hide the owner in this context
- I’m wondering if we separate the inclusion of integrated tasks in certain activities from inclusion on the kanban. For example, with the Hot Take #4107 idea, would we expect all those integrated objects referenced in a post to be added to a kanban?

jordanh:
> How often folks are estimating something that has been assigned in Jira or not?

Quite often, it's fairly common behavior for assignments to be delayed until the work is "accepted" by a team member. It'll be a behavior we'll need to design for

> Do folks really expect Jira tasks to be tasks matched to the Parabol team and show up in the kanban query?

[...]

Issue #4129, opened 2020-09-02
Do not show task assignee dropdown menu if there are no options


tianrunhe:
## Issue - Enhancement

From #4105:

> If my teammate and I are only on 1 team together, then I can’t reassign their card to another team. We should make that control disabled* unless there are options:
![92017222-0e69c700-ed19-11ea-83a5-7917bbeea98f](https://user-images.githubusercontent.com/1879975/92036129-9f926b00-ed24-11ea-887e-fd5d25b6adb8.gif)

This is the behavior today in production already and it's the same but smaller issue for user assignee (i.e., solo team)

### Acceptance Criteria
When there are no other alternative options for user/team assignee on the task card:
- [ ] The cursor doesn’t change to a pointer
- [ ] The control doesn’t have a hover state or tooltip
- [ ] Clicking does nothing

- **Estimated effort:** 12 points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

jordanh:
Raising this menu causes a fetch. We don't know there aren't teams until the menu is raised. How about we give them a link to a create a new team? Something like, "no teams, go make one?"

jordanh:
The other thing we could do is show a better zero state. Iceboxing as this came up during scrub

PR #4164, opened 2020-09-15
Tasks without assignees

nickoferrall:
PR for issue #4115 

### Consideration

- The `/team` and `/me/tasks` pages do not show unassigned tasks. However, if a user unassigns a task, they will see the unassigned task (a task with a question mark instead of an avatar) until they refresh the page. I think this is preferable compared with immediately removing the task from `/team` or `/me/tasks` as a user may accidentally unassign a task and want to reassign it.

### AC

- [ ]  the Task object is refactored to support null userId
- [ ]  all task mutations are updated to support a Task with a nullable assignee (e.g. createTask, updateTask, deleteTask, etc etc)
- [ ]  A Task card shows a question mark as the assignee if it doesn't have an assignee
- [ ]  The Task assignee dropdown in the Task card footer includes the current assignee's name. if already selected & clicked again, the task becomes unassigned
- [ ]  the tasks query returns only tasks that have assignees


packages/server/graphql/types/Team.ts
```diff
@@ -228,7 +228,7 @@ const Team = new GraphQLObjectType<ITeam, GQLContext>({
         const viewerId = getUserId(authToken)
         const allTasks = await dataLoader.get('tasksByTeamId').load(teamId)
         const tasks = allTasks.filter((task) => {
-          if (isTaskPrivate(task.tags) && task.userId !== viewerId) return false
+          if (!task.userId || (isTaskPrivate(task.tags) && task.userId !== viewerId)) return false
```
nickoferrall, line 231: We could use the tasks query here instead of the resolve function. As both could work, I thought I'd keep things as they are and filter from here.

packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterTeamAssigneeMenu.tsx
```diff
@@ -33,9 +33,13 @@ const TaskFooterTeamAssigneeMenu = (props: Props) => {
...
   return (
-    <Menu {...menuProps} ariaLabel={'Assign this task to another team'}>
+    <Menu
+      {...menuProps}
+      defaultActiveIdx={taskTeamIdx}
+      ariaLabel={'Assign this task to another team'}
+    >
       <DropdownMenuLabel>Move to:</DropdownMenuLabel>
-      {assignableTeams.map((team) => {
+      {teams.map((team) => {
```
nickoferrall, line 42: Instead of filtering `teams`, I thought we could include the team the task is currently assigned to. In production it currently shows "Move to" menu with no items in the menu if there are no other teams.

This PR shows this currently assigned team. But, unlike the `/team` page, clicking the currently assigned team doesn't unassign it; the update function doesn't fire.

packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterUserAssigneeMenu.tsx
```diff
@@ -25,33 +25,36 @@ const TaskFooterUserAssigneeMenu = (props: Props) => {
   const {area, menuProps, task, viewer} = props
   const {userId, id: taskId} = task
   const {team} = viewer
-  const {teamMembers} = team || {teamMembers: []}
-  const assignees = useMemo(
-    () => teamMembers.filter((teamMember) => teamMember.userId !== userId),
-    [userId, teamMembers]
-  )
+  const {teamMembers}: any = team || {teamMembers: []}
```
nickoferrall, line 28: TypeScript complains if we try to map through `teamMembers` as it says it's readOnly. I tried to cast it to an interface but it wouldn't accept that. I tried a few solutions from [here](https://stackoverflow.com/questions/46634876/how-can-i-change-a-readonly-property-in-typescript) such as `type Writeable<T> = {-readonly [P in keyof T]-?: T[P]}` with no joy.

I'm reluctantly using an `any` at the moment, (TypeScript was previously recognising `assignees` as `any` too) but I'd love to know if there's a better way

packages/client/mutations/CreateJiraIssueMutation.ts
```diff
@@ -59,6 +59,7 @@ const CreateJiraIssueMutation = (
       const userId = task.getValue('userId')
       const teamId = task.getValue('teamId')
       const integration = task.getLinkedRecord('integration')
+      if (!userId) return
       const user = store.get(userId)
       if (!user || !integration) return
```
nickoferrall, line 64: It looks Jira integrations require a `user` so I'm removing the integration button from the unassigned tasks

mattkrick:
pinging @tianrunhe for a review!

tianrunhe:
Apologize for not responding! I checked this last week and thought it had a couple of conflicts (from the files and also functionality wise) with #4105. I feel like we should merge my changes in first and revisit this one. @nickoferrall if you need this sooner, definitely let me know and we can work together to resolve the conflicts.

nickoferrall:
No problem, I'm happy to wait until your PR has been merged

mattkrick:
@nickoferrall looks like a couple merge conflicts, other than that g2g!

packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterTeamAssigneeMenu.tsx
```diff
@@ -23,14 +23,19 @@ const TaskFooterTeamAssigneeMenu = (props: Props) => {
...
   const {team, id: taskId} = task
   const {id: teamId} = team
-
+  const {teams} = viewer
   const assignableTeams = useMemo(() => {
-    const {teams} = viewer
-    const filteredTeams = userIds ? teams.filter(({teamMembers}) =>
-      !!teamMembers.find(({userId}) => userIds.includes(userId))
-    ) : (teamIds ? teams.filter(({id}) => teamIds.includes(id)) : teams)
-    return filteredTeams.filter((team) => team.id !== teamId)
```
nickoferrall, line 32: Also here @tianrunhe & @mattkrick , by filtering the currently assigned team from the dropdown, we get an empty menu if the user is only in one team:

empty-menu

We have an issue open for this [here](https://github.com/ParabolInc/parabol/issues/4129). In this file, I'm including the `teamId` in the menu so it always shows something. If the user clicks the currently assigned team, the mutation doesn't fire.

nickoferrall (commented):
Merge conflicts fixed (although it looks like there's a problem with the typechecks in `master`?). I've added two comments about updates involving the recent user task filter changes

tianrunhe:
@nickoferrall Does this PR need update against the latest master branch?

nickoferrall:
yep - thanks @tianrunhe, just fixed the merged conflicts

tianrunhe:
I find the un-assignment not that intuitive for our user to figure it out. Maybe a ❌ before the current task owner name? @ackernaut what do you think?

ackernaut:
@tianrunhe I think that’s okay. We don’t really want many unassigned tasks. Each item can be selected or not. What about [this look for that avatar and label](https://www.figma.com/file/AcSOOEtScsOqcOB4iqowMZ/Unassigned?node-id=1%3A10)?

tianrunhe (changes requested):
@nickoferrall  If you could update to the icon Terry provided, that'd be super cool. Otherwise the code LGTM!

nickoferrall:
@tianrunhe sure thing, I like the look of that unassigned icon. Just pushed the update

ackernaut:
@nickoferrall I like the idea of using the gray text color for the ‘Unassigned’ label. What do you think?

nickoferrall:
@ackernaut I like it - it helps to differentiate the label from the input text. I've pushed an update using the gray label and `PALETTE.TEXT_MAIN` on hover to be consistent with the menu in `TaskFooterTeamAssignee`

ackernaut:
`TEXT_MAIN_HOVER` is slightly darker than `TEXT_MAIN` but I’m okay with the change. Thanks for making that change, this polish is nice :)

ackernaut:
Nah, `TEXT_MAIN_HOVER` usage is probably not consistent enough so I’d say ship `TEXT_MAIN` and I can clean up anything that gives me tension in a UI polish PR, never.

Issue #3954, opened 2020-06-24
Mass-invitation link duration too short


jordanh:
This is a feedback thread to spool up user feedback relating to the duration our mass-invitation links are good for.

This thread will be closed when either we have consensus on a road forward and a link to a new issue, or that this thread has will be iceboxed due to 6 months of inactivity

jordanh:
[A user](https://app.hubspot.com/contacts/3888472/contact/11345501) writes 🔒 :
> Meeting link is only good for one day (when shared with others)

jordanh:
[A user (soon to be enterprise customer)](https://app.hubspot.com/contacts/3888472/contact/7270101) 🔒 writes:

> Most importantly, could you make it so the invitation link lasts longer than 24 hours? We ask folks to spool up feedback over a week or longer, and today I have to send out a new link each morning

jordanh:
[A user](https://app.hubspot.com/contacts/3888472/contact/10810851) 🔒  said:

> Could you make the invitation links last the duration of the sprint? Ours are 2 weeks

jordanh:
Today links are 13 characters in length.

@mattkrick out of curiosity, what napkin math do you use to calculate the trade offs between security and link length?

Alternatively posed, what would the link length be for:

* 7 days?
* 14 days?
* 21 days?
*28 days or a month?

mattkrick:
related issue: https://github.com/ParabolInc/parabol/issues/3359

the dictionary is 36 chars (A-Z0-9).
So 36 ^ 12 = 4.7E18 combinations
safe in terms of days isn't really relevant because the problem is human, e.g. who left their slack open/got pwned & now that attacker is cruising all the invitation links. the probability of a user getting pwned for bad security practices is far greater than a collision.

jordanh:
I've been thinking about this some and I wonder if this friction is really an opportunity.

First, let's talk about pwnage. The case we're trying to guard against is:

1. A user has a breach of information and/or leaks a mass-invitation link
2. A malicious user joins the team/org and gains access to company-sensitive information and/or opens a door for social-engineering attacks

Let's assume #1 happens, how might we reduce the risk for #2?

I think the key is to minimize the risk from "malicious users." Today, if is mass-invitation link is long-lived the risk is somewhat high on free accounts as when a link is accessed, a new user may join the team without email validation. I'd like to assume that a member validated as owning a user account with the same email domain is not malicious.

We could consider changing the mass-invitation user flow such that:

[...]

PR #4168, opened 2020-09-15
Fix 4123 invite UI copy

### LINKED ISSUE #4123: Descriptive copy added to invitation UI
## Issue - Enhancement

Some users are caught off guard by the 1-day expiration of mass-invitation links. Today there is a tooltip informing them that the link expires after 1 day, but we suspect it is insufficient.

Note: some know/learn that the mass-invitation links expire after 24 hours, but continue to use them anyway. This issue does not address that tension

### Acceptance Criteria (optional):
   - a design for clear copy that explains the tradeoffs and expiration times between the 2 invitation methods
   - an implementation of this design or an opened implementation issue if the below effort is exceeded

- **Estimated effort:** 6 hours ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

ackernaut:
Fixes #4123 

### Test
- [ ] Check out the invitation dialog. The copy is explicit, helpful, and well-written.

Screen Shot 2020-09-15 at 14 40 22

### DIFF
--- a/packages/client/components/AddTeamMemberModal.tsx
+++ b/packages/client/components/AddTeamMemberModal.tsx
@@ -5,9 +5,7 @@ import useAtmosphere from '~/hooks/useAtmosphere'
 import useMutationProps from '~/hooks/useMutationProps'
 import {PALETTE} from '~/styles/paletteV2'
 import {ICON_SIZE} from '~/styles/typographyV2'
-
 import styled from '@emotion/styled'
-
 import {AddTeamMemberModal_teamMembers} from '../__generated__/AddTeamMemberModal_teamMembers.graphql'
 import useBreakpoint from '../hooks/useBreakpoint'
 import InviteToTeamMutation from '../mutations/InviteToTeamMutation'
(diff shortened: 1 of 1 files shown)


nickoferrall:
LGTM - although I know Jordan mentioned extending the duration of emails from 7 days to 30 days [here](https://github.com/ParabolInc/parabol/issues/3954#issuecomment-683895804).

If we want to hold off on updating the duration until that discussion thread is concluded then I think it's good to merge. Otherwise, it looks like we can update the duration [here](https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/server/utils/serverConstants.ts#L4).

ackernaut:
I pondered that discussion and the idea of changing that to 30 days. I saw the constant for it. It may me wonder if it’s really as simple as changing that one value. I’m curious how often folks will read this text, if it’ll affect behaviors, and if it’ll help with any of the dissonance with user expectations. It’s simple enough of a change that it’s fast and safe. Let’s see if it causes any tensions or helps.

PR #4175, opened 2020-09-18
remove trailing slash from intranet graphql url

tiffanyhan:
**AC:**

- [x] Go to `/admin/graphql` and verify doing any query works


tianrunhe (approved):
Great find!

Issue #3682, opened 2020-03-23
Timeline event not published on end meeting


mattkrick:
after ending a meeting & going to the timeline view i don't see my most recent meeting

AC
- [ ] above untrue

Effort: 11

PR #4179, opened 2020-09-22
Publish timeline event on end meeting

nickoferrall:
PR for issue #3682.

### Test

- [x] End a meeting, go to the Timeline and view the newly created timeline event without refreshing the page

### DIFF
--- a/packages/client/mutations/EndNewMeetingMutation.ts
+++ b/packages/client/mutations/EndNewMeetingMutation.ts
@@ -16,6 +16,7 @@ import {EndNewMeetingMutation as TEndNewMeetingMutation} from '../__generated__/
 import handleRemoveSuggestedActions from './handlers/handleRemoveSuggestedActions'
 import handleRemoveTasks from './handlers/handleRemoveTasks'
 import handleUpsertTasks from './handlers/handleUpsertTasks'
+import {ConnectionHandler, RecordProxy} from 'relay-runtime'
 
 graphql`
   fragment EndNewMeetingMutation_team on EndNewMeetingPayload {
@@ -25,6 +26,7 @@ graphql`
       endedAt
       teamId
     }
+    removedTaskIds
     team {
       id
       activeMeetings {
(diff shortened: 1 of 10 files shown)


mattkrick:
tagging @tianrunhe as reviewer

packages/server/graphql/mutations/endNewMeeting.ts
```diff
@@ -386,7 +383,8 @@ export default {
       teamId,
       isKill: getIsKill(meetingType, phase),
       updatedTaskIds,
-      removedTaskIds
+      removedTaskIds,
+      timelineEventId
```
tianrunhe, line 387: Why do you need to pass id this along?

packages/client/mutations/EndNewMeetingMutation.ts
```diff
@@ -24,7 +26,19 @@ graphql`
...
+        agendaItemCount
+        commentCount
+        taskCount
+      }
+      ... on RetrospectiveMeeting {
+        commentCount
+        reflectionCount
+        taskCount
+        topicCount
+      }
```
tianrunhe, line 39: Where do you need these fields?

tianrunhe (commented):
Feature tested and worked! Just two minor comments below:

packages/server/graphql/mutations/endNewMeeting.ts
nickoferrall: By passing `timelineEventId` here, `EndNewMeetingPayload` can use it with the dataLoader to grab the `timelineEvent`. If we don't pass the ID along, we won't have access to the `timelineEvent`.

packages/client/mutations/EndNewMeetingMutation.ts
nickoferrall: They're being used in `TimelineEventCompletedActionMeeting.tsx` & `...RetroMeeting.tsx`. In the `endNewMeetingTeamUpdater` function, I'm setting the linkedRecord of `meeting` on to `timelineEvent` so that we have access to these fields in the frontend.

nickoferrall:
Thanks for the review @tianrunhe! If you're happy with my responses, I think we're good to merge

tianrunhe (approved):
LGTM! Thanks for the expanataion!

Issue #3636, opened 2020-03-09
Toggle timestamp bug: click sets isEditing


mattkrick:
AC
- [ ] can toggle timestamp without "editing" a task

Effort: 10.5 points

mattkrick:
still valid

PR #4187, opened 2020-09-22
Fix for toggle timestamp bug

nickoferrall:
PR for issue #3636.

### Test

- [ ] Click on timestamp in the header of a task card and see that it does not say "editing". It toggles between "Created X hours ago" and "Updated X hours ago"
- [ ] "editing" still displays when the user clicks on the editable input text

### DIFF
--- a/packages/client/modules/outcomeCard/components/OutcomeCard/OutcomeCard.tsx
+++ b/packages/client/modules/outcomeCard/components/OutcomeCard/OutcomeCard.tsx
@@ -8,7 +8,7 @@ import {OutcomeCard_task} from '~/__generated__/OutcomeCard_task.graphql'
 import TaskEditor from '../../../../components/TaskEditor/TaskEditor'
 import TaskIntegrationLink from '../../../../components/TaskIntegrationLink'
 import TaskWatermark from '../../../../components/TaskWatermark'
-import {UseTaskChild} from '../../../../hooks/useTaskChildFocus'
+import useTaskChildFocus, {UseTaskChild} from '../../../../hooks/useTaskChildFocus'
 import {cardFocusShadow, cardHoverShadow, cardShadow, Elevation} from '../../../../styles/elevation'
 import cardRootStyles from '../../../../styles/helpers/cardRootStyles'
 import {Card} from '../../../../types/constEnums'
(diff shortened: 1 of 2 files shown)


nickoferrall:
Hey @ackernaut, it would be great to get your eyes on this one. One thing to note, when clicking the assignee/team dropdown menu in the task card footer, it also triggers `isEditing`. Do you think it should? 

I think it should only trigger when the user is editing the input text. If you feel the same way, I can update the PR.

ackernaut:
@mattkrick can you help me remember our reasoning for adding the `isEditing` state to task cards when interacting with the menus opposed to only when editing the description?

mattkrick:
interrupting chicken for local state. (if i'm selecting a date & the meeting advances, i shouldn't advance).
for remote state, it's just nice to know that someone is doing something, somewhere out in the great void

ackernaut:
### Re: when to apply `isEditing`

> interrupting chicken for local state. (if i'm selecting a date & the meeting advances, i shouldn't advance).
> for remote state, it's just nice to know that someone is doing something, somewhere out in the great void

That’s one very good reason (thanks for the reminder, Matt). In general I think it’s a good idea to consider any update to the task a form of editing. Is it the word ‘editing’ that seems strange? I can see how that verb seems more specific, about editing text. I wonder if ‘updating’ sounds any better as a general verb? Is it too much more of a visual footprint in the small hint text ‘_Terry is editing_…’ vs ‘_Terry is updating_…’ Probably an okay change?

My vote is that we do not change the behavior, especially as it helps solve for the problem of the view state going away while I’m interacting with an object. I’d entertain updating the verb to be more general, but we may need to scrub other places that uses ‘editing’ in a similar sense.

### Re: PR review

This LGTM. Do you need a more in-depth code review? I quickly read through it.

ackernaut (approved):

nickoferrall:
Thank you both for the explanation. I agree with you Terry that I think it makes sense to keep things as they are.

In terms of a more in-depth code review, I'm largely just moving the following code from the parent component to the child so that `editing` isn't triggered when clicking the timestamp: 
```
          onBlur={() => {
            removeTaskChild('root')
            setTimeout(handleCardUpdate)
          }}
          onFocus={() => addTaskChild('root')}
```

I think it's safe to merge but I'll tag @mattkrick in case you fancy a quick scan of the functionality.

PR #4204, opened 2020-09-28
SFU media server

tiffanyhan:


packages/client/hooks/useMedia.ts
```diff
@@ -0,0 +1,25 @@
...
+
+const useMedia = ({mediaRef, mediaSource}: useMediaSignature) => {
+  useEffect(() => {
+    if (mediaSource?.track) {
+      const stream = new MediaStream()
+      stream.addTrack(mediaSource.track)
+      const el = mediaRef.current!
+      if (el.srcObject !== stream) el.srcObject = stream! // conditional is required to remove flickering video on update
+    }
+  })
```
mattkrick, line 17: may be worth passing in a list of deps here so we aren't creating a new mediastream on every re-render

packages/client/hooks/useMediaRoom.ts
```diff
@@ -0,0 +1,37 @@
...
+    deviceInfo: null
+  },
+  peers: {},
+  producers: {},
+  consumers: {}
+}
+
+const useMediaRoom = (roomId: string, teamId: string) => {
+  const atmosphere = useAtmosphere()
+  const {viewerId: peerId, authToken} = atmosphere
```
mattkrick, line 24: is it safe to use the viewerId as the peerId? what if someone is logged in on 2 devices or 2 tabs?

packages/client/hooks/useMediaRoom.ts
```diff
@@ -0,0 +1,37 @@
...
+  const [state, dispatch] = useReducer(reducerMediaRoom, initialState)
+  const disposable = useRef<(() => void) | null>()
+  useEffect(() => {
+    initMediaRoom(authToken, teamId, roomId, peerId, disposable, dispatch)
+    return () => {
+      disposable.current && disposable.current()
```
mattkrick, line 30: super nit: `disposable.current?.()` might be a hair cleaner

packages/client/utils/mediaRoom/Logger.ts
```diff
@@ -0,0 +1,21 @@
+import debug from 'debug'
```
mattkrick, line 1: may be worthwhile (or at least fun to try) to see if you can only import `debug` if we're in development. that way in prod it won't get bundled.

packages/client/utils/mediaRoom/urlFactory.ts
```diff
@@ -0,0 +1,23 @@
...
+  authToken,
+  teamId
+}: getSignalingServerUrlSignature) {
+  const url = new URL(`wss://${signalingServerHostname}`)
+  url.port = String(process.env.PROTOO_LISTEN_PORT)
+  url.searchParams.append('roomId', roomId)
+  url.searchParams.append('peerId', peerId)
+  url.searchParams.append('authToken', authToken)
+  url.searchParams.append('teamId', teamId)
+  return url.toString()
```
mattkrick, line 22: this returns a string, but the interface is an object
not sure what is calling this function. still tbd?

packages/sfu/server.ts
```diff
@@ -0,0 +1,121 @@
...
+const logger = new Logger()
+const mediasoupWorkers = []
+const rooms = new Map<string, Room>()
+const queue = new AwaitQueue()
+
+async function getOrCreateRoom(roomId: string) {
+  if (!rooms.get(roomId)) {
+    const worker = getMediaSoupWorker()
+    const room = await Room.create(roomId, worker)
+    rooms.set(roomId, room)
```
mattkrick, line 23: there exists a race condition here.
server A & B both call `rooms.get`, it returns false for both
server A & B both create the room & call `rooms.set`
whichever returns last (e.g. B) would then be the only one that exists in the the `rooms` map.

packages/sfu/server.ts
```diff
@@ -0,0 +1,121 @@
...
+    }
+    const {sub: userId, iat} = decodedAuthToken
+    const isBlacklistedJWT = await checkBlacklistJWT(userId, iat)
+    logger.debug('Is blacklisted:', isBlacklistedJWT)
+    if (isBlacklistedJWT) {
+      reject(401, 'Your authentication has expired')
+      return
+    }
+    if (!isTeamMember(decodedAuthToken, teamId as string)) {
+      reject(403, "Oops! You're not authorized to be here")
```
mattkrick, line 103: we auth the teamId, but we don't auth the roomId.
if an attacker has the roomId could they join the chat?
we could prefix the `roomId` with the `teamId`, e.g. `roomId = `${teamId}:foo123`
alternatively, we could let the server generate the roomId itself.
i'm sure other folks have solved this & the nuances (e.g. having different clients race to create or get the room). it'd be fun to check out what those solutions are.

mattkrick (commented):
so far so good! i didn't see the part where the client is initiating the request to the server, so stopped there

packages/sfu/server.ts
tiffanyhan: @mattkrick did you see where the function is called from? it gets put onto an in-memory queue, where each task is processed sequentially one by one.

packages/client/utils/mediaRoom/urlFactory.ts
tiffanyhan: Ah the interface is an object, but it's an interface for the fn signature not the return value. I didn't specify typing for the return value, which I should probably do. This fn is called by `MediaRoom` right before it creates a web socket connection to the media server.

packages/client/utils/mediaRoom/Logger.ts
tiffanyhan: Hmm I want to be able to switch on debug logs on a development server environment. So far it's been really helpful to surface bugs that only come up in a deployed environment. Is there any env var that is specific only to production? Or maybe I can try to import it only if `DEBUG=mediasoup*` env var is set...

packages/client/hooks/useMedia.ts
tiffanyhan: that sounds like a good idea!

packages/client/hooks/useMediaRoom.ts
tiffanyhan: That's a good question. Assuming they log into the same Parabol account, what will currently happen is only one of the devices or tabs will be able to join the mediaroom. following the example of other video products, it seems we should let each of those clients join the room.

Any ideas on how to achieve it? The first thing that comes to mind is using a randomly generated id and prepending it to the viewerId.

packages/sfu/server.ts
tiffanyhan: this is very true. it's currently only doing authentication but not authorization for the room. i like the idea of prefixing the roomId with the teamId.

tiffanyhan: the `roomId` is now is composite of of `teamId` and `meetingId`

packages/client/hooks/useMediaRoom.ts
tiffanyhan: changed so that `peerId` is a composite of `clientId` and `viewerId`, where `clientId` is a randomly generated string for a single browser/device.

packages/client/utils/mediaRoom/Logger.ts
tiffanyhan: `LoggerFactory` now instantiates `NoOpLogger` or `DebugLogger` depending on value of `process.env.DEBUG`. `debug` module is dynamically imported in `DebugLogger` creation.

mattkrick:
ack! sorry i haven't touched this in awhile. @tiffanyhan is this good to merge?

tiffanyhan:
> ack! sorry i haven't touched this in awhile. @tiffanyhan is this good to merge?

yea i think so! i've been keeping it updated with master. just wasn't sure if you wanted to review it further.

PR #4205, opened 2020-09-29
Undo grouping

### LINKED ISSUE #4151: Undo button for completed grouping phase
## Issue - Enhancement

This issue defines AC resulting from the discussion #3701, "Allow for regrouping of cards until vote phase completed" addressing the following JTBD:

> When a user has accidentally advanced the meeting, or wishes to add new reflections, or wishes to divide or combine two groups after having advanced to the Vote phase, they want an easy way to Undo the completion state of the Group phase, so that they can adjust the meeting content to their liking

@ackernaut please let me know if this would be blocked be design, or if you think the below AC are sufficient

A design for when users wish to to undo the completion status of the Retro Group phase:
   - The 'Phase Completed' badge of the Group base is updated to read 'Phase Completed Undo' for the Facilitator, where the word Undo is a clickable link
   - Undoing the phase allows the Grouping phase to continue
   - Any non-facilitator users navigated to the Vote phase are navigated backward to the Group phase

Plus:
   - Opened an implementation issue to implement the resulting design

- **Estimated effort:** 12 hours ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

tiffanyhan:
#4151

**AC:**
- [x] after starting a new meeting, adding reflections, grouping them, optionally voting on them, and optionally adding discussion comments and tasks-- the facilitator can navigate back to the completed group phase and type in `ididntmeanto`
- [x] any votes on all reflection groups are removed
- [x] any discussion items, including comments, tasks, and replies, in the meeting are deleted
- [x] all discussion stages are reset so there's only 1 placeholder discussion stage in the discussion phase. 
- [x] all stages and phases after GROUP are reset (basically, the `Meeting` object looks identical to how it looked before the meeting was advanced from GROUP to VOTE)
- [x] as a facilitator, the group phase no longer looks completed
- [x] as a participants, they are transported back to the group phase


mattkrick:
requesting review from @nickoferrall

packages/client/components/PhaseCompleted.tsx
```diff
@@ -0,0 +1,26 @@
...
+  meetingId?: string
+  resetToStageId?: string
+}
+
+const PhaseCompleted = (props: Props) => {
+  const {isComplete, canUndo, meetingId, resetToStageId} = props
+  if (!isComplete) return null
+  return (
+    <div>
+      {canUndo ? (
```
nickoferrall, line 17: `canUndo` is always true at the moment and `Wrapper` in `UndoablePhaseCompleted` seems to be identical to the `PhaseCompleteTag` styled component. Perhaps we could remove this file, use `UndoablePhaseCompleted` in the parent and use the `PhaseCompleteTag` in `UndoablePhaseCompleted` to reduce duplication?

nickoferrall (approved):
LGTM! I love the cheat code. Once the merge conflicts are fixed, it looks good to go

packages/client/components/PhaseCompleted.tsx
tiffanyhan: @nickoferrall that is true! i just pushed a change to pass in `canUndo=true` only from `retroGroupPhase`. Once Terry has the design, `UndoablePhaseCompleted` should have different styling to include a clickable button to undo (instead of using the hotkey).  I thinkk it's good to hold off on further refactoring until the UI design is settled on.

packages/client/mutations/ResetMeetingToStageMutation.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+  }
+`
+
+export const resetMeetingToStageUpdater: SharedUpdater<ResetMeetingToStageMutation_meeting> = (
+  payload,
+  {store}
+) => {
+  const meetingId = safeProxy(payload)
+    .getLinkedRecord('meeting')
```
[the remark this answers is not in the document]
tiffanyhan: good question! I was simply following the pattern of other mutations, and didn't look too much into what `safeProxy` does. looks like this is something that @mattkrick wrote, and my first guess is that it's a thin wrapper to ensure stricter type safety.

packages/client/mutations/ResetMeetingToStageMutation.ts
```diff
@@ -0,0 +1,94 @@
...
+    .getValue('id')!
+  const meeting = store.get<ClientRetrospectiveMeeting>(meetingId)
+  if (!meeting) return
+  const reflectionGroups = meeting.getLinkedRecords('reflectionGroups')
+  if (!reflectionGroups) return
+  reflectionGroups.forEach((rg) => rg.setValue(0, 'viewerVoteCount'))
```
mattkrick, line 77: `viewerVoteCount` is already requested & relay should magically merge in the new value since the reflectionGroupIds match. try to comment out all these lines & see if it still works. if it does, we can get rid of this entire updater (yay less code!)

packages/server/graphql/mutations/resetMeetingToStage.ts
```diff
@@ -0,0 +1,134 @@
...
+      if (viewerId !== defaultFacilitatorUserId)
+        return standardError(new Error('Not meeting facilitator'), {userId: viewerId})
+      return standardError(new Error('Not meeting facilitator anymore'), {userId: viewerId})
+    }
+
+    // VALIDATION
+    const foundResponse = findStageById(phases, stageId)
+    if (!foundResponse)
+      return standardError(new Error('Meeting stage not found'), {userId: viewerId})
+    const {stage: resetToStage} = foundResponse!
```
mattkrick, line 45: `foundResponse` is guaranteed to be truthy since the above line guarantees that it exists

packages/client/mutations/ResetMeetingToStageMutation.ts
tiffanyhan: yea i tried that and unfortunately relay doesn't automagically update the viewer vote count on reflection groups. was gonna ask you if you had any clues about why that would bwe.

packages/server/graphql/mutations/resetMeetingToStage.ts
```diff
@@ -0,0 +1,134 @@
...
+      return standardError(new Error('Stage has not finished'), {userId: viewerId})
+    if (resetToStage.phaseType !== NewMeetingPhaseTypeEnum.group)
+      return standardError(new Error('Resetting to this stage type is not supported'), {
+        userId: viewerId
+      })
+    if (meeting.endedAt)
+      return standardError(new Error('The meeting has already ended'), {userId: viewerId})
+
+    // RESOLUTION
+    const meetingCount = await r
```
mattkrick, line 58: you can grab this direct from the meeting & reduce a hit to the DB

packages/server/graphql/mutations/resetMeetingToStage.ts
```diff
@@ -0,0 +1,134 @@
...
+        .table('RetroReflectionGroup')
+        .getAll(r.args(reflectionGroupIds))
+        .update({voterIds: []})
+        .run(),
+      r
+        .table('NewMeeting')
+        .get(meetingId)
+        .update({phases: newPhases})
+        .run(),
+      (r.table('MeetingMember').getAll(meetingId, {index: 'meetingId'}) as any)
```
mattkrick, line 122: nit: can cast to `RetroMeetingMember` to get free typechecking of the votesRemaining field

packages/client/mutations/ResetMeetingToStageMutation.ts
mattkrick: thaaat is interesting. let me pull a copy & find out why

packages/server/graphql/mutations/resetMeetingToStage.ts
tiffanyhan: ah wasn't aware we stored it on the meeting, thanks!

packages/server/graphql/mutations/resetMeetingToStage.ts
tiffanyhan: oh good catch

packages/client/components/PhaseCompleted.tsx
nickoferrall: Makes sense! Sounds good to me

packages/server/graphql/mutations/resetMeetingToStage.ts
```diff
@@ -0,0 +1,134 @@
...
+        const resettedStage = createdPhases[phaseIndex]?.stages[stageIndex]
+        if (!resettedStage) continue
+        newStages.push(Object.assign(resettedStage, {id: stage.id}))
+      }
+      phase.stages = newStages
+      newPhases.push(phase)
+    }
+    primePhases(newPhases, resetToPhaseIndex)
+
+    const reflectionGroups = await dataLoader
```
mattkrick, line 96: warning that the dataloader only makes 1 fetch per request. when requested again, it'll use the cached version.
since you need to mutate the `voterIds` field, you either need to not use the dataloader, or mutate the cached value in addition to mutating the record in the DB.

e.g.
```js
reflectionGroups.forEach((group) => group.voterIds = [])
```

mattkrick:
Screen Shot 2020-09-30 at 9 23 50 AM
here's how i debugged:
- was the bad value on client or server? i saw the server is still sending down the wrong number, so i know the problem is on the server

this led me to find the bug was in the server mutation (it was used a cached value). see other comment

packages/server/graphql/mutations/resetMeetingToStage.ts
tiffanyhan: true! out of curiosity and if anything comes top of mind, how would you handle `Property 'update' does not exist on type 'RetroMeetingMember'.`?

tiffanyhan:
Thanks for your reviews @nickoferrall and @mattkrick !

Issue #4157, opened 2020-09-14
Summary topics out of order


gcrickman:
## Issue - Bug

[A user](https://app.hubspot.com/contacts/3888472/contact/6576801) 🔒  writes:

> In the summary the topics are no longer sorted in ascending order (the most voted first) but just randomly.

ackernaut:
### AC
- Triage and fix if LTE 3 hours
### EE
- 3 hours

jordanh:
10.5 points

jordanh:
Was fixed in #4206

PR #4206, opened 2020-09-29
Summary Topics

NikAiyer:
This PR attempts to solve the following issue: #4157 
Changed the sort by variable in RetroTopic and QuickStats from stageOrder to voteCount so that topics are sorted in ascending order.


mattkrick:
looks like only issues that have more than 0 votes make it into the summary.

here's some additional tests:
- [ ] all the reflection groups make it into the summary, sorted by vote count
- [ ] if the meeting is ended before the vote phase is reached, all groups appear in the summary

nickoferrall (approved):
LGTM!

PR #4220, opened 2020-09-30
add isStarter to starter templates

mattkrick:
@tianrunhe here's a start to the starter templates work.
since we want these templates to live in on-premise deployments, we probably want to always guarantee that the new templates are added via a migration.

Alternatively, we could have a template master file (e.g. retroTemplates.json) and guarantee that the master file always gets pushed to the DB. but tbh i think that's probably too much for now.

give it a test & if it looks good feel free to merge & then add the additional templates for @avivapinchas in a new migration

### DIFF
--- a/packages/server/database/migrations/20200930085630-isStarter.ts
+++ b/packages/server/database/migrations/20200930085630-isStarter.ts
@@ -1,0 +1,38 @@
+import {R} from 'rethinkdb-ts'
+export const up = async function(r: R) {
+  const starters = [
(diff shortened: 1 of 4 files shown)


tianrunhe:
@mattkrick Thanks for this! LGTM. I'll pull down this branch and add the data in. Is the new test tarball an intentional change or it was an older version when we iterate through the issue?

mattkrick:
ah shoot, unintentional. feel free to cherry pick the good bits!

PR #4222, opened 2020-09-30
filter for isActive on reflect prompts

### LINKED ISSUE #4171: Deleted prompts in retro template are still shown after refresh and cannot be deleted again
## Issue - Bug

To reproduce:

1.  Start a new meeting
2.  Select retrospective meeting and click the retro template dropdown
3.  Create a new template and see a "New Prompt" is already created
4.  Add another prompt, "New prompt #2"
5.  Delete "New prompt #2" and see it's gone
6.  Now refresh the page
7.  Go to Retro Meeting and click the template dropdown again
8.  See the "New prompt #2" shows up again in the "\* New Template"
9.  Try to delete that prompt again and it won't disappear

Users can:

*   [ ] create and delete prompt at will

mattkrick:
fixes #4171 

@nickoferrall to take it home!

### DIFF
--- a/packages/server/graphql/mutations/addReflectTemplate.ts
+++ b/packages/server/graphql/mutations/addReflectTemplate.ts
@@ -76,7 +76,8 @@ const addReflectTemplate = {
         parentTemplateId
       })
       const prompts = await dataLoader.get('reflectPromptsByTemplateId').load(parentTemplate.id)
-      const newTemplatePrompts = prompts.map((prompt) => {
+      const activePrompts = prompts.filter(({isActive}) => isActive)
+      const newTemplatePrompts = activePrompts.map((prompt) => {
         return new RetrospectivePrompt({
           ...prompt,
           teamId,
--- a/packages/server/graphql/mutations/updateTemplateScope.ts
+++ b/packages/server/graphql/mutations/updateTemplateScope.ts
@@ -79,8 +79,9 @@ const updateTemplateScope = {
       })
       clonedTemplateId = clonedTemplate.id
       const prompts = await dataLoader.get('reflectPromptsByTemplateId').load(templateId)
-      const promptIds = prompts.map(({id}) => id)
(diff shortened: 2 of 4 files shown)


nickoferrall:
@mattkrick it looks like the prompts and the meeting are actually connected. To repro the bug:
1. Create a retro and go to the Group phase
2. In a second window, remove one of the prompts
3. The retro Group/Vote/Discuss phase in the first window will now crash

We can filter `isActive` on the frontend but then removing a prompt will update previously completed retros. 

Perhaps I should go ahead with the `removedAt` idea and create a custom dataLoader? 

If you agree, I have a small question about the custom dataLoader:
```
export const reactables = (parent: RethinkDataLoader) => {
  return new DataLoader<ReactablesKey, Reactable, string>(
    async (keys) => {
```
It looks like we pass in `ReactablesKey`, we return `Reactable` but what is the third `string` - is it the caching key?

mattkrick:
that's good to know, the meeting should be robust enough to handle that change. could you track down the error & see if you can fix it in the case that a prompt goes away? 
which one of the 4 files causes the crash? ReflectPhase?

how would the `removedAt` logic go? include the prompt if `!removedAt || removedAt > meeting.startAt`? seems smart to me

nickoferrall:
For sure - I'll dig into that in the morning! 

Yep, that check plus `prompt.createdAt < meeting.createdAt` in the custom dataLoader. 

> It looks like we pass in ReactablesKey, we return Reactable but what is the third string - is it the caching key?

Just bumping this question as I was working with the custom data loader before and it wasn't playing nicely so I just want to check my understanding of how it's working

mattkrick:
> Just bumping this question as I was working with the custom data loader before and it wasn't playing nicely so I just want to check my understanding of how it's working

yep! that's it. to verify you can cmd + click the `DataLoader` to get to the types files & then follow the `C` generic var down to the cacheKeyFn definition

PR #4231, opened 2020-10-01
Proxy jira through parabol

### LINKED ISSUE #4004: Build Sprint Poker Jira Backend integration v1

mattkrick:
fixes #4004

TEST
- [ ] can integrate against jira in poker
- [ ] when the jira tab is loaded, all the jira issues are initially fetched
- [ ] as the search input changes, the list of issues updates
- [ ] can update an issue with story points (UI not built yet)

### DIFF
--- a/packages/client/components/JiraScopingNoResults.tsx
+++ b/packages/client/components/JiraScopingNoResults.tsx
@@ -21,11 +21,15 @@ const Info = styled(Icon)({
   paddingRight: 16
 })
 
-const JiraScopingNoResults = () => {
+interface Props {
+  error?: string | null
+}
+const JiraScopingNoResults = (props: Props) => {
+  const {error} = props
   return (
     <Message>
-      <Info>info</Info>
-      No issues match that query
+      <Info>{error ? 'warn' : 'info'}</Info>
+      {error || 'No issues match that query'}
     </Message>
   )
 }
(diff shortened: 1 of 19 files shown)


mattkrick:
merging because it's all back-end stuff behind a feature flag

PR #4232, opened 2020-10-02
Sprint poker template scale backend

### LINKED ISSUE #4184: Build Sprint Poker template backend
Child issue of #4057

## Issue

Build server side mutations to support operations of sprint poker template.

In `/admin/graphql` user can:

### Poker Template:

*   [ ] Create a sprint poker template:
    *   [ ] A new dimension is created by default
    *   [ ] 3 set of default scales are created for the template (T-shirts, Fibonacci, 5 fingers)
*   [ ] Remove a sprint poker template
*   [ ] Rename a sprint poker template

### Poker Template Dimension:

*   [ ] Create a new dimension for a  sprint poker template
*   [ ] Remove a dimension from a sprint poker template
*   [ ] Rename a dimension for a sprint poker template
*   [ ] Change the scale for a dimension in a sprint poker template

### Poker Template Scale:

*   [ ] Create a new scale for a sprint poker template
*   [ ] Remove a scale for a sprint poker template
*   [ ] Rename a scale for a sprint poker template

### Poker Template Scale Value:

*   [ ] Create a new scale value for a scale in a sprint poker template
*   [ ] Remove a scale value from a scale in a sprint poker template
*   [ ] Update the scale (value, label or color) in a scale in a sprint poker template

*   **Estimated effort:** 20 hours

tianrunhe:
Resolves #4184 

In `/admin/graphql` user can:

### Poker Template:

*   [ ] Create a sprint poker template:
    *   [ ] A new dimension is created by default
    *   [ ] 3 set of default scales are created for the template (T-shirts, Fibonacci, 5 fingers)
*   [ ] Remove a sprint poker template
*   [ ] Rename a sprint poker template

### Poker Template Dimension:

*   [ ] Create a new dimension for a  sprint poker template
*   [ ] Remove a dimension from a sprint poker template
*   [ ] Rename a dimension for a sprint poker template
*   [ ] Change the scale for a dimension in a sprint poker template

### Poker Template Scale:

*   [ ] Create a new scale for a sprint poker template
*   [ ] Remove a scale for a sprint poker template
*   [ ] Rename a scale for a sprint poker template

### Poker Template Scale Value:

*   [ ] Create a new scale value for a scale in a sprint poker template
*   [ ] Remove a scale value from a scale in a sprint poker template
*   [ ] Update the scale (value, label or color) in a scale in a sprint poker template


mattkrick:
this looks fantastic. still only about 20 files into the code review, will finish it up on monday. stoked to see this land in the ship next week!

packages/server/graphql/mutations/helpers/makePokerTemplateScales.ts
```diff
@@ -0,0 +1,50 @@
+import TemplateScale from '../../../database/types/TemplateScale'
```
mattkrick, line 1: if scales can be re-used across templates on the same team we can get rid of this

packages/server/graphql/mutations/removePokerTemplateDimension.ts
```diff
@@ -0,0 +1,72 @@
...
+      return standardError(new Error('Team not found'), {userId: viewerId})
+    }
+
+    // VALIDATION
+    const {teamId, templateId} = dimension
+    const dimensionCount = await r
+      .table('TemplateDimension')
+      .getAll(teamId, {index: 'teamId'})
+      .filter({
+        isActive: true,
```
mattkrick, line 39: @nickoferrall came up with a solid pattern which we may want to implement here.

the problem is a dimension might get removed while a meeting is in progress or after a meeting has ended.
if i re-visit an ended meeting, i should still be able to see that removed dimension.

this can be accomplished by replacing `isActive` with a `removedAt: Date` field. 
The logic is pretty simple: https://github.com/ParabolInc/parabol/blob/7c207c7651c38003ef1e3a88c05ce2314b0a4001/packages/server/utils/isRecordActiveForMeeting.ts

packages/server/graphql/types/RemovePokerTemplateScaleValuePayload.ts
```diff
@@ -0,0 +1,22 @@
+import {GraphQLObjectType} from 'graphql'
+import {GQLContext} from '../graphql'
+import TemplateScale from './TemplateScale'
+import StandardMutationError from './StandardMutationError'
+
+const RemovePokerTemplateScaleValuePayload = new GraphQLObjectType<any, GQLContext>({
+  name: 'RemovePokerTemplateScaleValuePayload',
+  fields: () => ({
+    error: {
```
mattkrick, line 9: nit: I found a slightly cleaner pattern for mutation payloads--
I make the payload a union of an Error & a Success.
That way, if `error` is present, we are guaranteed an NonNull error
if it's not present, then every other field (e.g. `scale` in this case) can be guaranteed to be non-null
Definitely not something you have to go through & redo! Just something for your own taste if you want to try it out moving forward

example here: https://github.com/ParabolInc/parabol/blob/jira-filter-menu/packages/server/graphql/types/AddReactjiToReactablePayload.ts#L1

packages/server/graphql/types/TemplateScale.ts
```diff
@@ -0,0 +1,67 @@
...
+  fields: () => ({
+    id: {
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'shortid'
+    },
+    createdAt: {
+      type: new GraphQLNonNull(GraphQLISO8601Type)
+    },
+    isActive: {
+      type: GraphQLBoolean,
```
mattkrick, line 29: nit, if you do a `resolve: ({isActive}) => !!isActive` then we can make the type NonNull.

packages/server/graphql/types/TemplateScale.ts
```diff
@@ -0,0 +1,67 @@
...
+      description: 'true if the scale is currently used by the team, else false'
+    },
+    teamId: {
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'foreign key. use the team field'
+    },
+    team: {
+      type: Team,
```
mattkrick, line 37: nonnull?

packages/server/graphql/types/SharableTemplate.ts
```diff
@@ -0,0 +1,68 @@
...
+    type: new GraphQLNonNull(GraphQLISO8601Type)
+  }
+})
+
+const SharableTemplate = new GraphQLInterfaceType({
+  name: 'SharableTemplate',
+  description: 'A meeting template that can be shared across team, orgnization and public',
+  fields: sharableTemplateFields,
+  resolveType: (type) => {
+    const templateType = type.prompts ? SharableTemplateEnum.RETROSPECTIVE : SharableTemplateEnum.SPRINT_POKER
```
mattkrick, line 59: suggestions:
- the `MeetingTemplate` object has a `type` value. we can use that to ascertain the type. that also means we won't have to create a SharableTemplateEnum
- I've found that `resolveType` is somewhat brittle because it means that this interface has to know EVERY single kind of child type it represents. when we add new types, we also have to remember to come back here & update the resolver. The new way of doing it is adding an `isTypeOf` field to the new type itself. By using the new way, that means whenever you add a new type, you just have to create the new type, you don't have to also edit this resolver. an example is here: https://github.com/ParabolInc/parabol/blob/jira-filter-menu/packages/server/graphql/types/TimelineEventCompletedRetroMeeting.ts#L11

mattkrick (changes requested):
absolutely fantastic work! this is a whole bunch of CRUD operations that you've handled really well.

the only big change I'd suggest is decoupling scale from template.
Doing that means:
- users can create new templates faster by using scales already created by their team
- we don't create a bunch of duplicate scales that aren't being used

everything else is just small stuff!

packages/server/graphql/mutations/addPokerTemplateScaleValue.ts
```diff
@@ -0,0 +1,71 @@
...
+      .get(templateId)
+      .run()
+    const viewerId = getUserId(authToken)
+
+    // AUTH
+    if (!template || !isTeamMember(authToken, template.teamId) || !template.isActive) {
+      return standardError(new Error('Team not found'), {userId: viewerId})
+    }
+
+    // VALIDATION
```
mattkrick, line 43: we need some validation for the scaleValue here.

- the label should be a max of 2 chars
- the value should be a number 0-99
- the color should start with a # and be followed by 6 hex numbers normalized to uppercase to keep it all clean)

tianrunhe:
@mattkrick In case you thought I'm finished and started to review it... it is still WIP. I'll request another review when I'm ready. Thank you!

packages/server/graphql/mutations/addPokerTemplateScaleValue.ts
mattkrick: one thing to note:
there exist some special cards like ❓  and 🚫 .
❓ should have a value of 2^31 -1
🚫  should have a value of -1
we give them those values for the sort order.
we should make exceptions for these 2 values.

packages/server/graphql/mutations/addPokerTemplateScaleValue.ts
```diff
@@ -0,0 +1,92 @@
...
+    const endIndex = scale.values.length - 1
+    if (index > endIndex + 1 || index < 0) {
+      return standardError(new Error('Invalid index'), {userId: viewerId})
+    }
+    const {color, label, value, isSpecial} = scaleValue
+    if (isSpecial) {
```
mattkrick, line 58: we may allow users to remove the ? card.
if that's the case, we should allow them to add it back in.
therefore, we should allow `isSpecial`, but make sure that the value is -1 or 2**31-1

mattkrick (changes requested):
SO CLOSE!
a bunch of comments left, but pretty tiny stuff. great work!

packages/server/graphql/mutations/addPokerTemplateScale.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    // RESOLUTION
+    let newScale
+    const sortOrder =
+      activeScales.length > 0
+        ? Math.max(...activeScales.map((scale) => scale.sortOrder)) + 1 + dndNoise()
+        : 0
+    if (parentScaleId) {
+      const parentScale = await dataLoader.get('TemplateScales').load(parentScaleId)
+      if (!parentScale) {
```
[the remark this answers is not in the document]
tianrunhe: I added this check with one exceptional condition: the parent scale is a starter scale.

packages/server/graphql/mutations/updatePokerTemplateScaleValue.ts
```diff
@@ -0,0 +1,94 @@
...
+  description: 'Update a scale value for a scale in a poker template',
+  type: UpdatePokerTemplateScaleValuePayload,
+  args: {
+    scaleId: {
+      type: new GraphQLNonNull(GraphQLID)
+    },
+    scaleValue: {
+      type: new GraphQLNonNull(TemplateScaleInput)
+    },
+    index: {
```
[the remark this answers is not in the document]
mattkrick: we can use `beforeValue` instead of index to handle drag & drop behavior.

for non-numeric scales, we should update the values going from least to greatest.
e.g. `[{value: 0, label: 'S'}, {value: 1, label: 'M'}]` if i drag the M above the S, then it should read:
`[{value: 0, label: 'M', {value: 1, label: 'S'}]`
the args would be:
```js
scaleId: 'foo',
scaleValue: 1,
beforeValue: 0
```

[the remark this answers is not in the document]
mattkrick: beforeValue is also susceptible to race conditions, so index works here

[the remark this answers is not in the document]
mattkrick: @tianrunhe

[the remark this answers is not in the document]
tianrunhe: @mattkrick I have deeper thinking about this today. I just felt that our data model is fragile. So our current thinking is that a scale value has a label and a numerical value. We give the user the ability to edit the label and then we "infer" the numerical value on our end. Think about the following cases, because hey users can put whatever they want:  
  
Edge case 1: user creates three scale values, in the order, "1", "S", "M", "2". How do we assign numerical value for each label here?

Even more extreme cases: user creates three scale values, in the order, "2", "3", "1".  Do we wanna "be smart" and say that oh I think you actually mean 1, 2, 3, or do we respect the ordering user put in, and then weirdly map them to the ordering?

So one solution is that when create a scale, we ask would you like a label only scale or a numerical scale? Then for label only scale, we respect the ordering they put in. For numerical scale, we put a client side check for sth like "hey you cannot move a 3 in the middle of 1 and 2".

Feels like more work either way. Any simple solutions here?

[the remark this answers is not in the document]
mattkrick: idea 1: forced sort. if when they add a new value, we pick where it goes. we sort all numeric values from least to greatest. all non-numeric values come after the numeric values. they have free rein over how non-numeric values are sorted. this is bad because it may feel too magical.

idea 2:  when the label is non-numeric, a numeric "value" input field appears. basically, we have a hidden "value" column & when 1 or more labels are non-numeric, that column gets unhidden @ackernaut for input & design. i think this is probably the best.

idea 3: throw an error. if they enter a non-numeric value in a numeric set, we say they can't do that. this limits functionality, so it's not great.

packages/server/graphql/mutations/removePokerTemplateScale.ts
```diff
@@ -0,0 +1,50 @@
...
+    const {teamId} = scale
+    if (!isTeamMember(authToken, teamId)) {
+      return standardError(new Error('Team not found'), {userId: viewerId})
+    }
+    if (!scale || scale.removedAt) {
+      return standardError(new Error('Scale not found'), {userId: viewerId})
+    }
+
+    // RESOLUTION
+    await r
```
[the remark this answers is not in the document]
tianrunhe: 1. It's okay to delete the last **team** scale as we can fall back to default scale (fibonacci, t-shirt, etc)
2. I made the change for every active dimension to use the next available scale after a team scale is deleted
3. I modified MeetingPoker DB type to also persist a list of mapping from `dimensionId` to `scaleId` so we can revisit an ended meeting with the available information

packages/server/dataloader/customLoaderMakers.ts
```diff
@@ -323,3 +329,37 @@ export const meetingSettingsByType = (parent: RethinkDataLoader) => {
...
+    async (keys) => {
+      const r = await getRethink()
+      const types = {} as {[meetingType: string]: string[]}
+      keys.forEach((key) => {
+        const {meetingType} = key
+        types[meetingType] = types[meetingType] || []
+        types[meetingType].push(key.teamId)
+      })
+      const entries = Object.entries(types)
+      const resultsByType = await Promise.all(
```
mattkrick, line 344: very efficiently done! 💪

packages/server/dataloader/customRedisQueries.ts
```diff
@@ -35,6 +35,26 @@ const customRedisQueries = {
...
     return publicTemplatesByType
+  },
+  starterScales: async (teamIds: string[]) => {
+    const r = await getRethink()
+
+    const starterScales = await Promise.all(
+      teamIds.map((teamId) => {
+        return r
+          .table('TemplateScale')
+          .getAll(teamId, {index: 'teamId'})
```
mattkrick, line 46: clever way to avoid another index. well done!

packages/server/graphql/mutations/addPokerTemplateDimension.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    const teamScales = await r
+      .table('TemplateScale')
+      .filter({teamId})
+      .filter((row) =>
+        row('removedAt')
+          .default(null)
+          .eq(null)
+      )
+      .orderBy(r.desc('updatedAt'))
```
mattkrick, line 62: nice!

packages/server/graphql/mutations/addPokerTemplateScale.ts
```diff
@@ -0,0 +1,102 @@
...
+    let newScale
+    const sortOrder = Math.max(0, ...activeScales.map((scale) => scale.sortOrder)) + 1 + dndNoise()
+    if (parentScaleId) {
+      const parentScale = (await dataLoader
+        .get('TemplateScales')
+        .load(parentScaleId)) as TemplateScale
+      if (!parentScale) {
+        return standardError(new Error('Parent scale not found'), {userId: viewerId})
+      }
+      if (!!parentScale.isStarter && parentScale.teamId !== teamId) {
```
mattkrick, line 57: need to remove a `!`?

packages/server/graphql/mutations/addPokerTemplateScaleValue.ts
```diff
@@ -0,0 +1,101 @@
...
+    await r
+      .table('TemplateScale')
+      .get(scaleId)
+      .update(
+        (row) => ({
+          values: row('values').append(scaleValue),
+          updatedAt: now
+        }),
+        {returnChanges: true}
+      )
```
mattkrick, line 67: i see you started returning changes but then broke it into 2.

on thing line, you can add `('changes')(0)('new_val').default(null)`
if it's null, then that means the update didn't occur.

since atomicity is at play here, it's probably worth doing it all at once

packages/server/graphql/mutations/addPokerTemplateScaleValue.ts
```diff
@@ -0,0 +1,101 @@
...
+      .run()
+    const updatedScale = await r
```
mattkrick, line 69: can merge with above update

packages/server/graphql/mutations/addPokerTemplateScaleValue.ts
```diff
@@ -0,0 +1,101 @@
...
+      .table('TemplateScale')
+      .get(scaleId)
+      .run()
+
+    if (validateScaleLabelValueUniqueness(updatedScale.values)) {
```
mattkrick, line 74: nit: if a function called `validate` returns true, I'd assume it is `valid`, but in this case, returning true means it is invalid.

could inverse the logic so it returns `false` or perhaps name it like `scaleValueHasDuplicates` or `scaleValueIsInvalid`?

packages/server/graphql/mutations/helpers/validateScaleValue.ts
```diff
@@ -0,0 +1,30 @@
...
+const validateColorValue = (color: string) => {
+  return /^#[0-9A-F]{6}$/i.test(color)
+}
+
+const validateScaleLabel = (label: string) => {
+  return 0 < label.length && label.length <= 2
+}
+
+const validateScaleValue = (value: number, isSpecial: boolean) => {
+  return isSpecial ? -1 <= value && value <= Math.pow(2, 31) - 1 : 0 <= value && value <= 99
```
mattkrick, line 12: for special values we probably shouldn't accept a range, we should only accept those 2 values.

otherwise, some a-hole could add a scaleValue like `{value: 10000000, isSpecial: true}` and then the card would look real bad

packages/server/graphql/mutations/removePokerTemplateScale.ts
```diff
@@ -0,0 +1,78 @@
...
+      .table('TemplateScale')
+      .get(scaleId)
+      .update({removedAt: now, updatedAt: now})
+      .run()
+
+    const activeTeamScales = await r
+      .table('TemplateScale')
+      .getAll(teamId, {index: 'teamId'})
+      .filter((row) =>
```
mattkrick, line 47: let's make this deterministic by adding an `orderBy` (updatedAt? createdAt?)

packages/server/graphql/types/MeetingTemplate.ts
```diff
@@ -0,0 +1,67 @@
...
+  GraphQLNonNull,
+  GraphQLString,
+  GraphQLInterfaceType,
+  GraphQLBoolean
+} from 'graphql'
+import GraphQLISO8601Type from './GraphQLISO8601Type'
+import SharingScopeEnum from './SharingScopeEnum'
+import Team from './Team'
+
+export const sharableTemplateFields = () => ({
```
mattkrick, line 12: rename to `meetingTemplateFields`?

packages/server/graphql/types/SharableTemplateEnum.ts
```diff
@@ -0,0 +1,12 @@
+import {GraphQLEnumType} from 'graphql'
+
+const SharableTemplateEnum = new GraphQLEnumType({
```
mattkrick, line 3: can delete this here & from rootSchema

mattkrick (changes requested):
fantastic work.
just a few minor tweaks, shouldn't take more than 30 mins.
merge when ready!

packages/server/graphql/mutations/addPokerTemplateScale.ts
tianrunhe: Actually, meant to be triple ! to handle both `false` and `undefined`

tianrunhe:
@mattkrick Thanks for the review! Addressed the feedback. What a journey😎! Really learnt a lot. Feeling awesome!

packages/server/graphql/mutations/addPokerTemplateScale.ts
```diff
@@ -49,12 +49,12 @@ const addPokerTemplateScale = {
...
     if (parentScaleId) {
       const parentScale = (await dataLoader
-        .get('TemplateScales')
+        .get('templateScales')
         .load(parentScaleId)) as TemplateScale
       if (!parentScale) {
         return standardError(new Error('Parent scale not found'), {userId: viewerId})
       }
-      if (!!parentScale.isStarter && parentScale.teamId !== teamId) {
+      if (!!!parentScale.isStarter && parentScale.teamId !== teamId) {
```
mattkrick, line 57: just 1 will do, no need to go crazy 😜

packages/server/graphql/types/TemplateDimension.ts
```diff
@@ -70,12 +72,23 @@ const TemplateDimension = new GraphQLObjectType<any, GQLContext>({
...
       }
+    },
+    availableScales: {
+      type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(TemplateScale))),
+      description: 'The list of scales can be set for this dimension',
+      resolve: async ({teamId}, _args, {dataLoader}) => {
+        const activeTeamScales = await dataLoader.get('scalesByTeamId').load(teamId)
+        const publicScales = await db.read('starterScales', 'aGhostTeam')
+        const activeScales = [...activeTeamScales, ...publicScales]
+        activeScales.slice().sort((a, b) => (a.sortOrder < b.sortOrder ? -1 : 1))
```
mattkrick, line 89: you'll want to remove `slice` here, otherwise you're copying an array, sorting it, and then returning the original.

Issue #4240, opened 2020-10-05
Incorrect Check-in agenda state after last item


jordanh:
## Issue - Bug

Here's a Loom of what the bug looks like: https://www.loom.com/share/cf5a40ba89d24768b430c487ac52e2da

No clean reproduction, yet. We'll see if this reproduces during our next check-in meetings.

Possible repro and symptoms:

   - Add several agenda items
   - Process all items
   - On last call, add another item
   - Attempt to navigate back to new item, note:
      - you cannot navigate there
      - all completed items now incorrectly show that they weren't completed

- **OS:** macOS
- **Browser:** Chrome
- **Estimated effort:** 5 hours (https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

### Acceptance Criteria (optional)
Users can:
   - add new agenda item at Last Call, and navigate to it
   - see correct agenda item completion state

jordanh:
Possible dup of #4884 ?

jordanh:
Iceboxing

PR #4242, opened 2020-10-06
Check-in agenda state

NikAiyer:
This PR attempts to solve the following issue: #4240 

If there are no stages present on the current phase, the onClick handler in AgendaItem gets returned with a undefined value, and the completion status gets set to false. Because the other phases do not have stages, when the user is on any other phase, clicking on agenda items becomes disabled and the items themselves are not shown as complete. In order to access the agenda items, you must click on Team Agenda first.

This PR fixes this issue by doing the following:
* adds phases variable to keep track of all phases in meeting, as well as the stages for agendaitems
* uses the stages from the agendaitems phase whenever the current phase is not agendaitems

Users can:

*   add new agenda item at Last Call, and navigate to it
*   see correct agenda item completion state

### DIFF
--- a/packages/client/modules/teamDashboard/components/AgendaItem/AgendaItem.tsx
+++ b/packages/client/modules/teamDashboard/components/AgendaItem/AgendaItem.tsx
@@ -17,6 +17,7 @@ import UpdateAgendaItemMutation from '../../../../mutations/UpdateAgendaItemMuta
 import pinIcon from '../../../../styles/theme/images/icons/pin.svg'
 import unpinIcon from '../../../../styles/theme/images/icons/unpin.svg'
 import {ICON_SIZE} from '../../../../styles/typographyV2'
+import {NewMeetingPhaseTypeEnum} from '~/types/graphql'
 
 const AgendaItemStyles = styled('div')({
   position: 'relative',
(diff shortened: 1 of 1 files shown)


mattkrick:
adding @nickoferrall to test

packages/client/modules/teamDashboard/components/AgendaItem/AgendaItem.tsx
```diff
@@ -69,11 +70,14 @@ const getItemProps = (
     isComplete: false,
     isUnsyncedFacilitatorStage: false
   }
-
   if (!meeting) return fallback
-  const {facilitatorUserId, facilitatorStageId, localStage, localPhase} = meeting
+  const {facilitatorUserId, facilitatorStageId, localStage, localPhase, phases} = meeting
+  const agendaItemsPhase = phases!.find(
```
nickoferrall, line 75: Looks like we don't need both `!` here, we can just use `phases.find(...`

packages/client/modules/teamDashboard/components/AgendaItem/AgendaItem.tsx
```diff
@@ -220,7 +241,12 @@ export default createFragmentContainer(AgendaItem, {
       endedAt
       facilitatorStageId
       facilitatorUserId
+      phases {
+        phaseType
+        ...AgendaItemAllMeetingPhases @relay(mask: false)
```
nickoferrall, line 246: We could pass in `...AgendaItemPhase` instead here and then we'd be able to remove `...AgendaItemAllMeetingPhases` from the file

nickoferrall (changes requested):
Nice work tracking down and then solving this nasty looking bug. Small changes requested and then it looks good to go

nickoferrall (approved):
LGTM!

PR #4245, opened 2020-10-06
Mock Sprint Poker estimate phase

nickoferrall:
PR for issue #4185 

### Considerations
- I've included 3 mock swipeable columns so that we can see how the UI looks with a column before and after
- I've added temporary logic that lets the user click "Next" and go from Scope to Estimate. The user can't click Estimate in the sidebar. Once the backend is up and running, we can replace the temporary logic

### AC

- [ ]  In a sprint poker meeting, can click "next" from the SCOPE phase & advance to the estimate phase
- [ ]  If there are no estimate stages, a screen tells them that there are no items to estimate & it provides a link to the SCOPE phase (build this, but no need to use the logic to point here, just point to the mock by default)
- [ ]  The header component has the Estmate title & subtitle
- [ ]  A mock wide Task card exists as seen here: https://www.figma.com/file/csERjrh67KZKgnT5VahFSb/Sprint-Poker?node-id=2689%3A118425
- [ ]  2 mock Swipeable columns exists for 2 mock dimensions (https://www.figma.com/file/csERjrh67KZKgnT5VahFSb/Sprint-Poker?node-id=2689%3A118425)


mattkrick:
lgtm!

PR #4256, opened 2020-10-08
Jira Search Menus

mattkrick:
This PR does 3 things:
### Refactors integrations in the GraphQL schema.

We want to be more inviting to others creating their own integrations, so i wanted to encapsulate integrations as much as possible under the service's namespace.

They now look like this:
```gql
TeamMember {
  integrations {
    slack {
      notifications
    }
    atlassian {
      issues {}
      projects {}
    }
  }
}
```

### Adds a Jira Search Options menu

Now, you can decide whether you want to search jira issues using JQL or just a basic search string + project filter.
The project filter is a multi-select. if you have more than 10 projects, then a search box appears so you can search faster.

### Adds a Jira Search History menu

When you add an issue to the meeting, the query used to find that issue gets persisted to your top 5 so you can re-use it


mattkrick:
95% of this is behind a feature flag so i don't feel too much tension merging.
may be worthwhile to test a slack or jira integration on staging due to the schema changes, but that's the only thing outside the poker meeting that changed

PR #4270, opened 2020-10-13
Add CLI to create new mutations

### LINKED ISSUE #3960: CLI to create new mutations
today, new mutations take ~15 mins to bootstrap. We can probably shave 10 mins off & reduce errors by automating it.

i imagine a CLI like `yarn createMutation updateScope Template`. We'll probably use jscodeshift 

Acceptance criteria
- [ ] a CLI takes in: the mutation `name` and the primary object `type`
- [ ] it creates a `server/grahql/mutations/{Name}Payload.ts` file that returns a single `type` in the payload and an empty resolve function
- [ ] it adds the mutation to `rootMutation.ts`
- [ ] it creates a `server/graphql/mutations/{name}.ts` that has a return type of `type`
- [ ] it creates `client/mutations/{Name}Mutation.ts`

Estimate: 8 hours

mattkrick:
fixes #3960

creating a new mutation isn't hard, but there's a lot of ugly boilerplate.
we should get paid to solve hard problems, not do mindless junk.
So, I made a CLI to generate the boilerplate.
we've got about 10 mutations to develop for the next sprint, so this should help.

How to use:
- call the CLI: e.g. `yarn newMutation addSmilesToFriends --subscription team`

How it works:
- this will generate a mutation called `addSmilesToFriends` and connect it to the Team subscription
- if this mutation is going to be hooked up to a subscription, you can add that with a `--subscription` (or `-s`) flag
- it will generate a mutation file on the client & server
- it will generate a mutation payload file on the server
- it will update the rootMutation & rootSubscription on the server
- it will update the client subscription with the mutation fragment

@tianrunhe @tiffanyhan @nickoferrall let me know if this is useful!

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "dev": "node scripts/dev.js",
     "lint": "lerna run --parallel lint",
     "test": "start-server-and-test start http://localhost:3000 cypress",
+    "newMutation": "yarn sucrase-node scripts/codeshift/newMutation.ts",
     "cypress": "yarn workspace parabol-cypress start",
     "typecheck": "lerna run --parallel typecheck",
     "ultrahook": "export $(grep ^ULTRAHOOK_API_KEY .env | tr -d \"'\") && ultrahook -k $ULTRAHOOK_API_KEY dev 3000"
--- a/package.json
+++ b/package.json
@@ -53,6 +53,7 @@
     "husky": "^3.0.1",
     "jscodeshift": "^0.7.0",
     "lerna": "^3.19.0",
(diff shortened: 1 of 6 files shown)


mattkrick:
merging so we can get started with the grip of mutations for the estimate phase

nickoferrall:
@mattkrick amazing! I just gave it a try - this will be really helpful

tiffanyhan:
so cool, thanks for doing this and sharing!

tianrunhe:
oh boy this will make my life much easier!

PR #4278, opened 2020-10-14
Poker Estimate Phase: Card Deck

### LINKED ISSUE #4257: Poker: Estimate: Deck of cards
Given an array of scale values: `[{label: 0, value: 0, color: 'blue'}]`, create a deck of cards that looks like this: https://www.figma.com/file/csERjrh67KZKgnT5VahFSb/Sprint-Poker?node-id=2689%3A118425

Acceptance criteria:
- cards look like the figma design
- clicking a card puts it in a "selected" state which causes it to lift out of the deck
- the deck has a collapsed state (can be triggered via a cheat code or similar for now) where the deck collapses down to the width of a single card & the selected card is on top of the deck. this animation is smooth and abides by material design guidelines (ie the cards don't do anything that real life playing cards can't do)
- per the figma designs, for a deck of 11 cards, the resting state of the cards in the center are 12px higher than the cards on the edges. as the deck size increase & decreases, this delta should change, too. 
- for decks with fewer cards, they are more spread out, but they are always overlapping (e.g. overlap in figma is 100px for 11 cards, for 3 cards maybe it's 75px)
- BONUS: when hovering a card, it becomes the center of the curve of card heights. that is to say, given 11 cards, the card being hovered is 12 pixels higher than the card furthest from it. the cards next to the one being hovered are a few pixels lower (e.g. 9 pixels). the cards around those are slightly less (e.g. 5px)

Guidance:
- Achieving the slight curve that the cards have is a bit difficult because it has to look good whether there are 3 cards, 10, or 30. A naive approach would be to derive the max height from the number of cards and some coefficient (e.g. 11 cards * 1.1 = 12px) and set the height based on the index delta (e.g. if the tallest card is at index 3, then `heightOfIndex5 = 12px - (5 - 3)*3px = 6px`). this works, but implies a linear falloff. another approach might be to create a normal distribution & then match the centroid of each card to the curve of that distribution. beta distributions can look good, too.
- animating the deck from collapse/expand is difficult because we want the selected card to end up on the top of the deck, but it's somewhere in the middle. Just like real cards, the selected card should not be able to float through other cards like a ghost through a wall. it should be like real life. so, that means the selected card has to come out of the deck & then get put back in on top. should take ~300ms using the DECELERATE bezier curve. see ReflectionStacks in the group phase for inspiration. here are some options, not sure which would look best:
  - once the deck is collapsed, the selected card animates up from 60px to 175px (full height), box shadow increases to show elevation & then it slides down to 0px to sit on top of the stack
  - once the deck is collapsed, the selected card animates right 125px (full width), box shadow increases to show elevation & then it slides left to 0px to sit on top of the stack
  - while the deck is collapsing, the selected card moves to the right more slowly than the rest. once the selected card is at least 125px to the left of any other card, it can then go directly to the top (this is probably best because it has the least movement)

mattkrick:
fixes #4257 

TEST
- [ ] a deck of cards overlays the estimate phase dimension
- [ ] clicking a card makes it rise up
- [ ] clicking another card lowers the old one down & the new one up
- [ ] hitting "c" collapses/expands the deck 
- [ ] deck looks good with 3 cards
- [ ] deck looks good with 20 cards
- [ ] resizing the window & then collapsing still makes a stack at the bottom right of the screen
- [ ] toggling the left nav or resizing the screen keeps the cards in the center

### DIFF
--- a/packages/client/components/EstimatePhaseArea.tsx
+++ b/packages/client/components/EstimatePhaseArea.tsx
@@ -4,6 +4,7 @@ import {PALETTE} from '~/styles/paletteV2'
 import SwipeableViews from 'react-swipeable-views'
 import useBreakpoint from '~/hooks/useBreakpoint'
 import {Breakpoint} from '~/types/constEnums'
+import PokerCardDeck from './PokerCardDeck'
 
 const EstimateArea = styled('div')({
   display: 'flex',
@@ -67,6 +68,7 @@ const EstimatePhaseArea = () => {
           return <StepperDot key={idx} isActive={idx === activeIdx} />
         })}
       </StepperDots>
+      <PokerCardDeck />
       <SwipeableViews
         containerStyle={containerStyle}
         enableMouseEvents
@@ -79,6 +81,7 @@ const EstimatePhaseArea = () => {
           <SwipableEstimateItem key={idx} />
         ))}
       </SwipeableViews>
+
     </EstimateArea>
   )
 }
--- a/packages/client/components/PokerCard.tsx
+++ b/packages/client/components/PokerCard.tsx
@@ -1,37 +1,67 @@
 import styled from '@emotion/styled'
-import React from 'react'
+import React, {RefObject, useEffect, useRef} from 'react'
+import usePokerZIndexOverride from '../hooks/usePokerZIndexOverride'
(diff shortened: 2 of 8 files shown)


mattkrick:
everything is behind the feature flag, so good to merge

Issue #4139, opened 2020-09-08
Add Create Jira Issue to Scope Phase


mattkrick:
Per [the scope phase designs](https://www.figma.com/file/csERjrh67KZKgnT5VahFSb/Sprint-Poker?node-id=2696%3A123437), we need a button that lets folks add a Jira Story directly.

AC
- [ ] there's a button that matches the designs in figma to add a new jira story
- [ ] clicking that button scrolls to the top of the list, adds a new item & then focuses it so the user is ready to type. The checkbox is checked by default to be included in the scope.
- [ ] if that story is blurred & no text exists, then the story is removed
- [ ] a mutation called `jiraCreateIssue(content: string, cloudId: string, projectKey: string, teamId: string, meetingId?: string)` is built that is very similar to `createJiraIssue`. the difference is that it does not create a Task in our system, it only creates the Issue in Jira. Instead of returning a `Task` in the payload, it returns a `JiraIssue`
- [ ] `JiraCreateIssueMutation` is built on the frontend to call `jiraCreateIssue` the returned JiraIssue is prepended to the jira issue connection so it stays at the top of the list 
- [ ] after the story is persisted successfully, `UpdatePokerScopeMutation(service: 'jira', taskId: issueKey, action: 'add')` is called to add it to the scope

Estimate: 16 hours

ackernaut:
### Estimated effort
- 8 hours

ackernaut:
@nickoferrall the [main view is seen here](https://www.figma.com/file/csERjrh67KZKgnT5VahFSb/Sprint-Poker?node-id=2696%3A123437) but I should get together the UI states (focused, editing, etc.) for you. These states will be similar to some of [the concepts seen on this page for archived iterations](https://www.figma.com/file/csERjrh67KZKgnT5VahFSb/?node-id=2846%3A3). I’ll ping you in Figma when it’s ready.

ackernaut:
>`JiraCreateIssueMutation` is built on the frontend to call `jiraCreateIssue` the returned JiraIssue is prepended to the jira issue connection so it stays at the top of the list 

Is it worth mentioning that the new issue added appears at the top of the list despite what the current query may be? It would be possible to create an issue that doesn’t match the query and it would immediately go away which would probably feel broken. The last detail I remember discussing is that the new issue would remain until the query is updated, then would go away if it doesn’t match.

ackernaut:
@nickoferrall I’ve got a few states here
[...]

PR #4282, opened 2020-10-19
Add Jira Issue to Scope Phase

nickoferrall:
PR for issue #4139.

### AC

- [ ]  there's a button that matches the designs in figma to add a new jira story
- [ ]  clicking that button scrolls to the top of the list, adds a new item & then focuses it so the user is ready to type. The checkbox is checked by default to be included in the scope.
- [ ]  if that story is blurred & no text exists, then the story is removed
- [ ]  a mutation called jiraCreateIssue(content: string, cloudId: string, projectKey: string, teamId: string, meetingId?: string) is built that is very similar to createJiraIssue. the difference is that it does not create a Task in our system, it only creates the Issue in Jira. Instead of returning a Task in the payload, it returns a JiraIssue
- [ ]  JiraCreateIssueMutation is built on the frontend to call jiraCreateIssue the returned JiraIssue is prepended to the jira issue connection so it stays at the top of the list
- [ ]  after the story is persisted successfully, UpdatePokerScopeMutation(service: 'jira', taskId: issueKey, action: 'add') is called to add it to the scope

### Addiitionally

- [ ] If filtering the Jira Issues, a newly added Jira issue should go to the top of the list, even if it doesn't match the query
- [ ] Let the user select which Jira project they want to add the new issue to. Show a suggested project key by default


packages/client/mutations/handlers/handleJiraCreateIssue.ts
```diff
@@ -0,0 +1,49 @@
...
+  const now = new Date().toISOString()
+  const newEdge = ConnectionHandler.createEdge(
+    store,
+    jiraIssuesConn,
+    jiraIssueProxy,
+    'JiraIssueEdge'
+  )
+  newEdge.setValue(now, 'cursor')
+  const node = newEdge.getLinkedRecord('node')
+  const newJiraIssueId = jiraIssue.getValue('id')
```
nickoferrall, line 44: Strangely, if I add `id: newJiraIssueId` to the `newJiraIssue`, I get the following error: `Cannot create a record with idffdd81d9-58c8-43af-8a23-2f6d1181cbd0:TES-315, this record already exists.` Assigning it here feels a bit hacky but this is the only way I've found of getting the correct `id` so that `UpdatePokerScopeMutation` can fire and tick the checkbox

packages/server/graphql/mutations/jiraCreateIssue.ts
```diff
@@ -0,0 +1,123 @@
...
+    const operationId = dataLoader.share()
+    const subOptions = {mutatorId, operationId}
+    const viewerId = getUserId(authToken)
+
+    // AUTH
+    if (!isTeamMember(authToken, teamId)) {
+      return standardError(new Error('Team not found'), {userId: viewerId})
+    }
+
+    // VALIDATION
```
mattkrick, line 52: to validate:
- does a meetingMember for the viewer with the provided meetingId exist?
- if meetingId is provided, is the meeting owned by the provided teamId?

packages/server/graphql/resolvers.ts
```diff
@@ -145,3 +147,37 @@ export const resolveForBillingLeaders = (fieldName) => async (
...
+    sendToSentry(new Error('No atlassian access token exists for team'), {userId: viewerId})
+    return null
+  }
+  const {userId} = teamAuth
+  const accessToken = await dataLoader.get('freshAtlassianAccessToken').load({teamId, userId})
+  const manager = new AtlassianServerManager(accessToken)
+  const issueRes = await manager.getIssue(cloudId, issueKey)
+  if ('message' in issueRes) {
+    sendToSentry(new Error(issueRes.message), {userId: viewerId})
+    return null
```
mattkrick, line 170: returning null isn't great because if the client hits an error, they should see what the error is.

packages/server/graphql/types/JiraCreateIssuePayload.ts
```diff
@@ -0,0 +1,29 @@
...
+const JiraCreateIssuePayload = new GraphQLObjectType<any, GQLContext>({
+  name: 'JiraCreateIssuePayload',
+  fields: () => ({
+    error: {
+      type: StandardMutationError
+    },
+    jiraIssue: {
+      type: JiraIssue,
+      description: 'The issue straight from Jira',
+      resolve: resolveJiraIssue
```
mattkrick, line 16: is the `source` argument in `resolveJiraIssue` the same as the `data` returned in `jiraCreateIssue`?
Where is the serviceTaskId?

packages/client/mutations/handlers/handleJiraCreateIssue.ts
mattkrick: i should do a better job of documenting this pattern.
createProxyRecord should only be called inside an optimisticUpdater.
is there a reason you can't just use `const jiraIssue = payload.getLinkedRecord('jiraIssue')` instead of extracting out all the values & then creating a new proxy?

packages/server/graphql/types/JiraCreateIssuePayload.ts
nickoferrall: The `source` arguments are different but `serviceTaskId` is only being used to grab the `cloudId` and `key` which are included in the `JiraCreateIssuePayload`. So in the resolver, I've added: 
```  
const [cloudId, issueKey] = serviceTaskId
    ? serviceTaskId.split(':')
    : [source.cloudId, source.key]
```
Perhaps this causes more confusion than it's worth but I thought the resolver could help reduce some duplicate code

packages/client/components/JiraScopingSearchResults.tsx
```diff
@@ -72,25 +93,56 @@ const JiraScopingSearchResults = (props: Props) => {
...
+                  isSelected={usedJiraIssueIds.has(node.id)}
+                  meetingId={meetingId}
+                  persistQuery={persistQuery}
+                />
+              )
+            })}
+          </ResultScroller>
+        </>
+      )}
+      <Button onClick={() => setIsEditing(true)} palette='blue'>
```
mattkrick, line 130: i think we should make this button hidden when you're editing a new issue.
alternatively, clicking it again should re-focus the new issue.

packages/client/components/NewJiraIssueInput.tsx
```diff
@@ -0,0 +1,195 @@
...
+              onBlur={handleCreateNewIssue}
+              onChange={(e) => setNewIssueText(e.target.value)}
+              placeholder='New issue summary'
+              ref={ref}
+              type='text'
+            />
+          </Form>
+          <StyledButton ref={originRef} onMouseDown={togglePortal}>
+            <StyledLink>{selectedProjectKey}</StyledLink>
+            <StyledIcon>expand_more</StyledIcon>
```
mattkrick, line 153: i think this is making the line height too tall. when the icon goes away, the component should be the same height so it feels like only the copy has changed

packages/server/graphql/mutations/jiraCreateIssue.ts
```diff
@@ -52,13 +55,33 @@ export default {
...
     const viewerAuth = viewerAuthRes.find((auth) => auth.teamId === teamId)
-
     if (!viewerAuth || !viewerAuth.isActive) {
       return standardError(new Error('The viewer does not have access to Jira'), {
         userId: viewerId
       })
     }
 
+    if (meetingId) {
+      const meeting = (await r
```
mattkrick, line 65: `teamId` is stored on the meeting member, so you could remove this DB hit & just go off the meetingMember

mattkrick:
just a little conflict to resolve (i think from Tiffany's PR), other than that it looks great!

PR #4304, opened 2020-11-02
add voteForPokerStory

### LINKED ISSUE #4263: Poker: Estimate: voteForPokerStory mutation
When folks get to the estimate stage, they should be able to vote for a particular story

AC
- [ ] a server mutation exists `voteForPokerStory(meetingId: string, stageId: string, score?: string)`
- [ ] validation is performed to make sure they have access to the meeting & the stageId points to an estimate stage in the meeting & the score matches a label on the current template dimension (stage.dimensionId)
- [ ] a `?` has a score of 2^31 -1 (2147483647) because we always want it to come last
- [ ] a 🚫  has a score of -1 because we always want it to come first
- [ ] if score is null, then remove the user's vote from the array
- [ ] before the update, stage.scores is checked to make sure that the user has not voted previous
- [ ] an update is performed like stage.scores.push({userId, score})
- [ ] after the update is performed, the DB record is checked to ensure that stage.scores only includes 1 record for that userId (a user could trigger this mutation twice back to back. we must protect against that). if 2 records are found, a rollback is performed.
- [ ] if isRemove is true, then that user's vote is removed from the array
- [ ] the update is published to the meeting channel
- [ ] a mutation payload type exists that includes a `stage` field
- [ ] a client mutation exists that requests the `stage { scores { userId, score } }`
- [ ] other members of the meeting see the change, too (the query is added to the meeting subscription on client & server)

Estimate: 4 hours

mattkrick:
fixes #4263. all behind feature flag

### DIFF
--- a/packages/server/graphql/mutations/removePokerTemplateScaleValue.ts
+++ b/packages/server/graphql/mutations/removePokerTemplateScaleValue.ts
@@ -1,11 +1,10 @@
-import {GraphQLID, GraphQLNonNull} from 'graphql'
+import {GraphQLFloat, GraphQLID, GraphQLNonNull} from 'graphql'
+import {SubscriptionChannel} from 'parabol-client/types/constEnums'
 import getRethink from '../../database/rethinkDriver'
 import {getUserId, isTeamMember} from '../../utils/authorization'
 import publish from '../../utils/publish'
 import standardError from '../../utils/standardError'
 import RemovePokerTemplateScaleValuePayload from '../types/RemovePokerTemplateScaleValuePayload'
-import {SubscriptionChannel} from 'parabol-client/types/constEnums'
-import TemplateScaleInput from '../types/TemplateScaleInput'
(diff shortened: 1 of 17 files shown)


packages/server/graphql/mutations/removePokerTemplateScaleValue.ts
```diff
@@ -54,7 +50,11 @@ const removePokerTemplateScaleValue = {
       .table('TemplateScale')
       .get(scaleId)
       .update((row) => ({
-        values: row('values').deleteAt(oldScaleValueIndex),
+        values: row('values').deleteAt(
+          row('values')
+            .offsetsOf(oldScaleValue)
```
mattkrick, line 55: @tianrunhe i missed this in the code review, but here's a change that ensures the operation is atomic. kind of a strange pattern, but useful if you ever have to modify a value inside an array

PR #4312, opened 2020-11-04
refactor atlassianAuth, fix ts problems

### LINKED ISSUE #4250: refactor freshAtlassianAccessToken and atlassianAuthByUserId to freshAtlassianAuth
now that usage for atlassian is pretty well established we can do away with overly verbose patterns

AC
- [ ] as the title says

Estimate: 4 hours

mattkrick:
fixes #4250.

for whatever reason, fixing that make typescript catch a bunch of errors it wasn't catching before, so i had to clean those up, too. 😞 

@nickoferrall some of the bugs that typescript caught involved CRUD mutations for comments where meetingId isn't guaranteed

### DIFF
--- a/packages/server/graphql/types/DragEstimatingTaskPayload.ts
+++ b/packages/server/graphql/types/DragEstimatingTaskPayload.ts
@@ -23,7 +23,7 @@ export const DragEstimatingTaskSuccess = new GraphQLObjectType<any, GQLContext>(
     },
     stage: {
       type: GraphQLNonNull(EstimateStage),
-      resovle: resolveStage(NewMeetingPhaseTypeEnum.ESTIMATE)
+      resolve: resolveStage(NewMeetingPhaseTypeEnum.ESTIMATE)
     }
   })
 })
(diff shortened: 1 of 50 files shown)


nickoferrall:
Interesting @mattkrick! I see that the `meetingId` isn't guaranteed in the addReactji mutation which is one I'm working on. I guess we'll be able to remove that check when my PR is ready as `meetingId` will be a non-nullable argument

packages/server/graphql/types/DragEstimatingTaskPayload.ts
```diff
@@ -21,13 +23,7 @@ export const DragEstimatingTaskSuccess = new GraphQLObjectType<any, GQLContext>(
...
     stage: {
       type: GraphQLNonNull(EstimateStage),
-      resolve: async ({meetingId, stageId}, _args, {dataLoader}) => {
-        const meeting = await dataLoader.get('newMeetings').load(meetingId)
-        const {phases} = meeting
-        const estimatePhase = phases.find((phase) => phase.phaseType === 'ESTIMATE')
-        const {stages} = estimatePhase
-        return stages.find((stage) => stage.id === stageId)
-      }
+      resovle: resolveStage(NewMeetingPhaseTypeEnum.ESTIMATE)
```
nickoferrall, line 26: Just noticed a tiny typo here

mattkrick: nice catch! i've been thinking about adding typings to the newMutation generator. i'll get that in

Issue #4260, opened 2020-10-09
Poker: Estimate: Discussion Thread in Right Drawer


mattkrick:
For the poker meeting, we want to put the discussion thread in a drawer.
this drawer should work exactly the same as the left drawer on the dash or in the meeting (can be swiped & peeked on when on mobile). 

AC
- drawer looks like the figma spec: https://www.figma.com/file/csERjrh67KZKgnT5VahFSb/Sprint-Poker?node-id=2689%3A118425
- on desktop, there is no hamburger menu to close it, so it's always open
- drawer works the same as the dash or meeting left nav. on mobile, it can be swiped open, longpress to peek, etc.

Note-- we'll probably have to change some physics in `SwipeableDashSidebar` to first detect if the user is attempting to open the left drawer or the right drawer! shouldn't be hard, just detect if the press is on the left or right side.

Estimate: 16 hours

ackernaut:
Does this include wiring up the thread itself or shall that be in another issue?

nickoferrall:
@ackernaut yep, this PR includes wiring up the thread

PR #4317, opened 2020-11-06
Poker discussion thread drawer

nickoferrall:
PR for issue #4260 

### AC

- [ ] drawer looks like the figma spec: https://www.figma.com/file/csERjrh67KZKgnT5VahFSb/Sprint-Poker?node-id=2689%3A118425
- [ ] on desktop, there is no hamburger menu to close it, so it's always open
- [ ] drawer works the same as the dash or meeting left nav. on mobile, it can be swiped open, longpress to peek, etc. #


packages/client/components/SwipeableDashSidebar.tsx
```diff
@@ -63,20 +63,28 @@ const updateSpeed = (clientX: number) => {
...
   const dy = Math.abs(movementY)
   const rads = -Math.atan(movementX / dy)
   if (dx > UNCERTAINTY_THRESHOLD || dy > UNCERTAINTY_THRESHOLD) {
-    // if it's open & it's a swipe to the left || it's closed & it's a swipe to the right
-    swipe.isSwipe = swipe.isOpen ? rads <= -MIN_ARC_RADS : rads >= MIN_ARC_RADS
+    const swipingLeft = rads <= -MIN_ARC_RADS
+    const swipingRight = rads >= MIN_ARC_RADS
+    // if it's open & sidebar is on the right & it's a swipe to the right || it's closed
+    // & sidebar is on the right & it's a swipe to the left. Vice versa for left sidebar
+    swipe.isSwipe = swipe.isOpen
```
nickoferrall, line 77: Any preference/thoughts on the nested ternary vs a function / lookup? I don't love the look of it but in this example I find it slightly cleaner & easier to read. I think this guy has [a point](https://medium.com/javascript-scene/nested-ternaries-are-great-361bddd0f340
)

mattkrick: darn combinatorics. fwiw i think you wrote it as cleanly as it could be written!

packages/client/components/SwipeableDashSidebar.tsx
```diff
@@ -98,16 +106,20 @@ const swipe = {
...
 const SwipeableDashSidebar = (props: Props) => {
-  const {children, isOpen, onToggle} = props
+  const {children, isOpen, isRightSidebar = false, onToggle} = props
   const {portal, openPortal} = usePortal({
     allowScroll: true,
     noClose: true
   })
   const [xRef, setX] = useRefState(0)
+  const sidebarWidth: number = isRightSidebar ? DiscussionThreadEnum.WIDTH : NavSidebar.WIDTH
+  const HYSTERESIS_THRESH = HYSTERESIS * sidebarWidth
```
mattkrick, line 121: nit: i like to reserve all caps for things that are constant at build time (i've probably screwed up & set a bad example somewhere, but that's my method to madness)

packages/client/components/SwipeableDashSidebar.tsx
```diff
@@ -6,46 +6,46 @@ import useRefState from '../hooks/useRefState'
...
+    transform: `translateX(${isRightSidebar ? -x : x}px)`,
+    transition: `transform 200ms ${DECELERATE}`,
+    zIndex: ZIndex.SIDEBAR
+  })
+)
+const Sidebar = styled('div')<{x: number; hysteresisThresh: number}>(({x, hysteresisThresh}) => ({
   boxShadow: x > 0 ? navDrawerShadow : undefined,
-  pointerEvents: x > HYSTERESIS_THRESH ? undefined : 'none'
+  height: '100vh',
+  pointerEvents: hysteresisThresh !== undefined && x > hysteresisThresh ? undefined : 'none'
```
mattkrick, line 48: under what condition is hysteresisThread undefined?

packages/client/hooks/useScrollThreadList.ts
```diff
@@ -43,7 +46,8 @@ const useScrollThreadList = (
         // the delay is required for new task cards, not sure why height is determined async
       }, 50)
     }
-  }, [isInit, threadables, preferredNames])
+    setOldClientHeight(clientHeight)
+  }, [isInit, threadables, preferredNames, el?.clientHeight])
```
mattkrick, line 50: oh now we're getting into fun stuff!
accessing the `clientHeight` seems really harmless because it's just a property. In reality, it triggers a DOM repaint.
in other words, it's like
```ts
class Element {
  get clientHeight() {
    return this.blockTheThreadAndTalktoTheDOM()
  }
}
```

by doing this on every re-render we can no longer leverage the speed of the VDOM.

perhaps you need `useLayoutEffect`?

packages/server/graphql/mutations/addComment.ts
```diff
@@ -64,7 +67,8 @@ const addComment = {
     const containsThreadablePhase = phases.find(
       (phase) =>
         phase.phaseType === NewMeetingPhaseTypeEnum.discuss ||
-        phase.phaseType === NewMeetingPhaseTypeEnum.agendaitems
+        phase.phaseType === NewMeetingPhaseTypeEnum.agendaitems ||
```
mattkrick, line 70: probably time to refactor
```ts
const threadablePhases = [NewMeetingPHaseTypeEnum.discuss, NewMeetingPHaseTypeEnum.agendaItems, NewMeetingPHaseTypeEnum.ESTIMATE]
({phaseType}) => threadablePhases.includes(phaseType)
```

packages/server/utils/getJiraCloudIdAndKey.ts
```diff
@@ -0,0 +1,2 @@
+const getJiraCloudIdAndKey = (id: string) => id.split(':')
```
mattkrick, line 1: i like this! add some typings so we know what we get as a return value

packages/server/graphql/types/PokerMeeting.ts
```diff
@@ -39,6 +43,31 @@ const PokerMeeting = new GraphQLObjectType<any, GQLContext>({
...
+      args: {
+        storyId: {
+          type: GraphQLNonNull(GraphQLID)
+        }
+      },
+      resolve: async ({teamId}, {storyId}, {authToken, dataLoader}) => {
+        const isJiraId = await isValidJiraId(storyId, teamId, {authToken, dataLoader})
+        if (isJiraId) {
+          const [cloudId, key] = getJiraCloudIdAndKey(storyId)
+          return {
```
mattkrick, line 58: remember, whatever the return value is must fulfill the `Story` contract.
For example, here we're returning a JiraIssue, but a JiraIssue has a nonnull field called cloudName.
that means the following would throw an error
```gql
story(storyId) {
  id
  cloudName
}
```

try to fulfill the Story contract by resolving all required fields.
see https://github.com/ParabolInc/parabol/blob/de8d951f698f01dd1a1decfe0e3e1b605d4520ae/packages/server/graphql/types/EstimateStage.ts#L91 for a hint

packages/server/graphql/types/PokerMeeting.ts
```diff
@@ -39,6 +43,31 @@ const PokerMeeting = new GraphQLObjectType<any, GQLContext>({
...
+            id: storyId,
+            cloudId,
+            key,
+            type: StoryTypeEnum.JIRA_ISSUE
```
mattkrick, line 62: after the above is implemented, `type` can go away

mattkrick (changes requested):
great work! just a couple bugfixes & we're good to merge

packages/server/utils/getJiraCloudIdAndKey.ts
```diff
@@ -1,2 +1,5 @@
-const getJiraCloudIdAndKey = (id: string) => id.split(':')
+const getJiraCloudIdAndKey = (id: string): [string, string] => {
+  const [cloudId, key] = id.split(':')
+  return [cloudId, key]
```
mattkrick, line 3: `return [cloudId, key] as [cloudId: string, issueKey: string]` takes it a step further. then, the person calling this will get a helpful little popup tells them which is which

nickoferrall: ooh, nice touch. Updated!

nickoferrall: Hmm it looks like this update causes a Typecheck error. TypeScript isn't showing any errors in the file, but I'm getting a build failure. Looking at [the docs](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-0.html#labeled-tuple-elements), it looks like the syntax is correct and we're using TypeScript 4.0. 
Perhaps there's a problem with how the build is set-up? I'll keep digging

mattkrick: can you bump all typescript versions to latest in the package.jsons? (can't trust the `^`). running `yarn outdated` shows that some are referencing v3

mattkrick:
CR looks good! After checking out the UI here are a final couple things:
- [ ] since we won't be supporting video yet, let's remove that stuff for now. make it just a discussion thread
- [ ] replace the video button with a collapse button that collapses the drawer
- [ ] the desktop version should have a collapse/expand button just like the mobile view
- [ ] the collapse & expand buttons should be the same size in the exact same location (just like the expand/collapse hamburger on the left side)

nickoferrall:
Hey @ackernaut, it'd be great to get your thoughts on how to approach the top of the right drawer.

The original [figma design](https://www.figma.com/file/csERjrh67KZKgnT5VahFSb/Sprint-Poker?node-id=2689%3A118425) has a video section at the top which we've removed until the video is ready. @mattkrick and I spoke yesterday and we thought that reducing the height of the drawer would enable users to still see the attendees avatar group & the burger icon. Now that I'm looking at it though, it feels a bit chopped-off to me and lacks symmetry with the left sidebar.

Instead, I prefer the 2nd screenshot which has a section at the top with the avatars of the users that have commented and a burger icon. It'd train users to expect to see who's discussing in the thread which would help when we launch video. The attendees avatar group could be hidden or move to the left. 

Other ideas:

- Until the video is ready, we could display the regular discussion thread instead of a drawer and position it at the far right
- Remove the top section and prevent the drawer from being collapsable
- 2nd screenshot without the avatar group - just the burger icon

Once we've agreed on an approach, I'll fix the merge conflict and there's a bit of functionality with the control bar that I'll update too.

reduced-height
initial-drawer

ackernaut:
@nickoferrall here are my thoughts on the layout and behaviors:

- The left and right sidebars have the full height of the viewport
- The thread sidebar is collapsible
- The meeting top bar including the avatars and invite control fills the center, main view
- Until we have the video component, the top bar of the thread sidebar can be the simple label and control to close

I’ll make an updated concept in Figma

PR #4318, opened 2020-11-06
replace polymorphic EstimateStage with Story union

### LINKED ISSUE #4316: estimate stages in sidebar only show 1 per serviceTaskId
each stage is a dimension. with multiple dimensions per serviceTaskId, this could look ugly.
instead, we should only show 1 for each serviceTaskId.

AC
- [ ] On a meeting with 2 dimensions per template, adding an issue to the scope only adds 1 item to the left sidebar
- [ ] clicking that item on the left sidebar will take you 1 of the stages that pertain to that issue
- [ ] dragging the stages around adjusts the sort order for all the stages, including the hidden ones
- which stage you go to is determined by the following:
- [ ] if the facilitatorStageId is a one of the possible stages, go there
- [ ] go to the first stage that you haven't cast your vote for
- [ ]  if you voted for all of them, then it takes you to the last one

Estimate: 4 hours

mattkrick:
fix #4316

The sidebar didn't support multiple dimensions & this was making it hard to build out other pieces of the estimate phase.
I also really didn't like the bifurcated EstimateStage, so instead of having e.g. EstimateStageJira, we now have a single EstimateStage with a `story` and that story can be a jira issue or a task.


packages/client/hooks/useRecordIdsWithStages.ts
```diff
@@ -1,25 +1,31 @@
...
-    | ParabolScopingSearchResults_meeting['phases']
-    | JiraScopingSearchResults_meeting['phases'],
-  phaseType: NewMeetingPhaseTypeEnum,
-  recordName: 'issue' | 'task'
-): Set<string> => {
-  const targetPhase = (phases as ReadonlyArray<any>).find((phase) => phase.phaseType === phaseType)!
-  const {stages} = targetPhase
+const useRecordIdsWithStages = (phaseRef: any) => {
   return useMemo(() => {
+    const estimatePhase = readInlineData<useRecordIdsWithStages_phase>(
```
mattkrick, line 8: hey @tiffanyhan here's a little-known pattern that makes this function a little more generic. Relay has this `readInlineData` function where you can declare the fields you want inline. This is great for functions that don't live in the component file because it uncouples them & the data requirements are co-located with the logic. No right or wrong, just another something for the tool belt!

packages/server/graphql/types/Story.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export const storyFields = () => ({
+  id: {
+    type: GraphQLNonNull(GraphQLID),
+    description: 'serviceTaskId'
+  }
+})
+
+const Story = new GraphQLInterfaceType({
+  name: 'Story',
```
mattkrick, line 11: @nickoferrall you probably beat me to it, but I needed a `Story` type for this, may be a merge conflict in our future

Issue #4258, opened 2020-10-09
Poker: Estimate: Reveal votes animation


mattkrick:
https://www.figma.com/file/csERjrh67KZKgnT5VahFSb/Sprint-Poker?node-id=2838%3A147623

AC
- When the votes are being cast, all the avatars exist on a single row (the question mark card row).
- When the votes are revealed, there will exist multiple rows, 1 for each unique value.
- the avatars from the single row should animate to the row according to their vote (e.g. if Bruce voted for 8, he should be on the 8 row)

Estimate: 16 hours

ackernaut:
I see that an estimate dimension stage is shaped like this:
```graphql
stage {
  isVoting
  scores {
    userId
    value
    label
  }
}
```
So I can use the `scores` to batch users e.g. showing who voted or showing who voted per value row and `isVoting` to swap modes?

Out of scope is how to handle the group of `hoveringUsers` when they have or have not voted? If they haven‘t voted they start peeking in the view when they are hovering? If they have voted but are hovering (movement, as opposed to just resting over the cards) maybe we start to shimmy their avatar to indicate they may be changing their mind?

Am I missing anything @mattkrick ?

mattkrick:
you got it! the hoveringUsers is in my draft PR

PR #4320, opened 2020-11-06
Poker Estimate voting UI

ackernaut:
Working on #4258 
### Voting layout
- [x] Dimension name with conditional team revote control
- [x] Active voting row (empty or with voters)
- [x] Tip banner for empty states
- [x] Reveal votes control
- [x] Dimension final value control or static final value
- [x] Voting result rows
- [x] Avatar overflow badge
- [x] Avatar overflow is dynamic
- [x] Dimension columns have correct sizing based on breakpoints
- [x] Layout accounts for thread drawer
### Deck
- [x] Made choices on default deck layout settings from checklist #4337
### Data and mutations
- [x] The facilitator can press a button to reveal votes for a single dimension 
- [x] The facilitator can press a ‘Team Revote’ button to reset the voting for a single dimension
- [x] Participants can select a card, can remove vote, can change vote during discussion
- [x] When votes are revealed, a Pass vote is added to scores for any participant who didn’t vote
- [x] Facilitator can enter a final value in the primary control
- [ ] Facilitator can tap on a revealed voting row to set the final value (@ackernaut does this set the pending value or does it submit it, too?)

### Column navigation
- [x] Pressing next goes to the next column or the next story
- [x] Swiping a column goes to the correct stage URL
### Animation
- [ ] Avatars and overflow badge transition in and out of rows
- [ ] Peeking avatars animate into a preview or revealed voting row when in flight and the `isVoting` state changes
- [ ] Avatars wiggle if folks voted but start changing their votes again
- [ ] Avatars fly from their place in the preview voting row to their respective revealed voting row
- [ ] Avatars transition in and out of revealed rows when changing votes
- [ ] Avatars transition into the Pass 🚫  row for folks who didn’t vote
- [ ] Revealed voting rows transition in or out if folks change their vote during discussion causing a new row or removing a row
- [ ] Final result card does a little bounce or dance every time it changes based on input from the facilitator


mattkrick:
@ackernaut merged master & resolved conflicts

packages/client/components/PokerEstimatePhase.tsx
```diff
@@ -57,19 +59,22 @@ const ShowDiscussionButton = styled(PlainButton)({
   padding: 8
 })
 
+const StyledMeetingHeaderAndPhase = styled(MeetingHeaderAndPhase)<{isDesktop: boolean}>(({isDesktop}) => ({
+  width: isDesktop ? `calc(100% - ${DiscussionThreadEnum.WIDTH}px)` : '100%'
```
ackernaut, line 63: @nickoferrall hey there, I moved the conditional width here to account for the drawer. If we set it on this container, then the children can just fill the container and don’t need to worry about the drawer layout.

mattkrick:
burndown:
- [ ] active card is tied to the stage
- [ ] deck collapse state is tied to the stage

packages/client/hooks/userPokerAvatarOverflow.ts
```diff
@@ -0,0 +1,21 @@
+import {RefObject, useLayoutEffect, useState} from 'react'
+import useResizeObserver from './useResizeObserver'
+
+const usePokerAvatarOverflow = (rowRef: RefObject<HTMLDivElement>) => {
```
ackernaut, line 4: @mattkrick I was thinking of adding this to the meeting top bar, where the participant avatars and invite control are. Right now we have 2 fixed overflow thresholds at the 2 main breakpoints. Safe to try?

ackernaut: I.e. adding a similar overflow hook

mattkrick: sure, that sounds good!

mattkrick: you'll need to refactor to support different icon sizes though, right?

ackernaut: I was thinking about setting the circles to the 40px sizing across breakpoints. At any rate, I’d use a new hook if I had to separate concerns of meeting avatars vs poker row avatars.

packages/client/components/PokerEstimatePhase.tsx
nickoferrall: Nice, that makes sense

packages/client/types/constEnums.ts
```diff
@@ -170,8 +170,16 @@ export const enum MeetingLabels {
 export const enum PokerCards {
   HEIGHT = 175,
   WIDTH = 125,
-  OVERLAP = 96
-
+  OVERLAP = 96,
+  MAX_VALUE = 2147483647,
+  // the angle in degrees from the middle of the deck to a deck edge (0-90)
+  TILT = 8,
```
ackernaut, line 176: @mattkrick I set the radius and tilt so that the Fibonacci deck looked good on mobile (~360px). I considered making these responsive so that the cards could breathe a little better at desktop breakpoints. Worth pursuing?

mattkrick: absolutely.
could you just fill in the logic to this function?
```ts
calculateTiltAndRadius = (totalCards: number, maxWidth: number) => {
  // do stuff
  return {tilt, radius}
}
```

ackernaut: Oh, something like this, where we could play with 2 const enums based on a breakpoint?
```js
const calculateTiltAndRadius = (totalCards: number, maxWidth: number) => {
    // RADIUS = 900
    // RADIUS_MOBILE = 1200
    // TILT = 12
    // TILT_MOBILE = 8
    const tilt = isDesktop ? PokerCards.TILT : PokerCards.TILT_MOBILE
    const radius = isDesktop ? PokerCards.RADIUS : PokerCards.RADIUS_MOBILE
    // calculate based on const enums
    return {tilt, radius}
  }
```
The fanning effect is nice when there’s room. 😄

mattkrick: ah gotcha, so the number of cards doesn't affect anything?

ackernaut: ah, good point, we could play with the width if the number of cards is high or low, which could save space at the mobile breakpoint, I’ll simmer on that, but I did look at that cheap trick of just setting 2 settings based on the breakpoint

mattkrick: something that might be fun is to make a table, maybe in excel?
you could do something like:

| width| totalCards | tilt | radius |
| 300 | 10 | 8  | 1000 |
| 400 | 10 | 10 | 1200

and after a few samplings (maybe 20?), we could derive a function from that

mattkrick:
this PR is getting a little big, merging now & then will pick up animations in a subsequent PR

packages/client/types/constEnums.ts
ackernaut: I made a side branch and backed up to the commit before I yanked the hotkeys so that I can play with it easier. I’ll see if I can find a pattern.

Issue #3940, opened 2020-06-17
make startNewMeeting, endNewMeeting meeting type specific


mattkrick:
starting and ending a meeting both have a bunch of logic that is very specific to the meeting type. while down the road we'll have custom meeting types & need a generic `startCustomMeeting` mutation, for now the logic is a little spaghetti-ish by combining both types in the same mutation.

Acceptance Criteria
- [ ] `startCheckinMeeting` mutation created
- [ ] `startRetrospectiveMeeting` mutation created
- [ ] `startNewMeeting` is deprecated (but not removed, we'll do that later)
- [ ] `endCheckinMeeting` mutation created
- [ ] `endRetrospectiveMeeting` mutation created
- [ ] `endNewMeeting` is deprecated (but not removed, we'll do that later)

Estimate: 6 hours

abhiaiyer91:
i read the `startNewMeeting` API and it looks like the only specific logic is around `retrospectives`.

https://github.com/ParabolInc/parabol/blob/master/packages/server/graphql/mutations/startNewMeeting.ts#L76

Do you want to extract some the shared logic between these mutations?

mattkrick:
there's a fair bit of stuff for the check-in meetings, too: https://github.com/ParabolInc/parabol/blob/master/packages/server/graphql/mutations/startNewMeeting.ts#L91-L121

the agendaItem stuff should get removed for retrospective meetings, too (that's a current bug).

up to you on how to extract, it's really more art than science

PR #4329, opened 2020-11-10
Split-up start & end meeting mutations

nickoferrall:
PR for issue #3940 

### AC

- [ ]  startCheckinMeeting mutation created
- [ ]  startRetrospectiveMeeting mutation created
- [ ]  startNewMeeting is deprecated (but not removed, we'll do that later)
- [ ]  endCheckinMeeting mutation created
- [ ]  endRetrospectiveMeeting mutation created
- [ ]  endNewMeeting is deprecated (but not removed, we'll do that later)

### Considerations

- To be consistent with the new `startSprintPoker` mutation and to keep things short & sweet, I've removed the word `meeting` from the name of the functions, i.e. `startCheckIn` instead of `startCheckInMeeting`
- I've added a deprecation flag to the `startNewMeeting` and `endNewMeeting` mutations but I haven't removed them from the subscriptions, root mutations, etc


packages/client/mutations/EndCheckInMutation.ts
```diff
@@ -0,0 +1,155 @@
...
+import {RecordProxy} from 'relay-runtime'
+
+graphql`
+  fragment EndCheckInMutation_team on EndCheckInSuccess {
+    isKill
+    meeting {
+      id
+      endedAt
+      teamId
+      ... on ActionMeeting {
```
mattkrick, line 28: the meeting is already of type ActionMeeting so this inline fragment isn't necessary

packages/client/mutations/EndRetrospectiveMutation.ts
```diff
@@ -0,0 +1,145 @@
...
+import {RecordProxy} from 'relay-runtime'
+
+graphql`
+  fragment EndRetrospectiveMutation_team on EndRetrospectiveSuccess {
+    isKill
+    meeting {
+      id
+      endedAt
+      teamId
+      ... on RetrospectiveMeeting {
```
mattkrick, line 27: shouldn't be necessary

packages/client/mutations/EndRetrospectiveMutation.ts
```diff
@@ -0,0 +1,145 @@
...
+          message
+        }
+      }
+      ...EndRetrospectiveMutation_notification @relay(mask: false)
+      ...EndRetrospectiveMutation_team @relay(mask: false)
+    }
+  }
+`
+
+const popEndRetrospectiveToast = (atmosphere: Atmosphere, meetingId: string) => {
```
mattkrick, line 71: i think we can refactor this toast to be used by all meetings

packages/client/subscriptions/NotificationSubscription.ts
```diff
@@ -66,7 +68,9 @@ const subscription = graphql`
       ...ArchiveTimelineEventMutation_notification @relay(mask: false)
       ...SetNotificationStatusMutation_notification @relay(mask: false)
       ...CreateTaskMutation_notification @relay(mask: false)
+      ...EndCheckInMutation_notification @relay(mask: false)
       ...EndNewMeetingMutation_notification @relay(mask: false)
```
mattkrick, line 72: can we remove this now, or should we wait a month or 2 before we remove?

packages/server/graphql/mutations/endCheckIn.ts
```diff
@@ -0,0 +1,303 @@
...
+    const completedCheckIn = ((await r
+      .table('NewMeeting')
+      .get(meetingId)
+      .update(
+        {
+          endedAt: now,
+          phases
+        },
+        {returnChanges: true}
+      )('changes')(0)('new_val')
```
mattkrick, line 216: it's possible (not probable, though) that this gets called 2x at the exact same time.
in that case, we'll wait to handle the case that an update did not result in a `new_val`.
easiest thing to do is add a `.default(null)` and then see `if (!completedCheckIn)...`

packages/server/graphql/mutations/endRetrospective.ts
```diff
@@ -0,0 +1,192 @@
...
+    )
+    const presentMemberUserIds = presentMembers.map(({userId}) => userId)
+    endSlackMeeting(meetingId, teamId, dataLoader).catch(console.log)
+    finishRetroMeeting(completedRetrospective, dataLoader)
+    const {facilitatorUserId, templateId} = completedRetrospective
+    const template = await dataLoader.get('meetingTemplates').load(templateId)
+    const {name: meetingTemplateName} = template
+    presentMemberUserIds.forEach((userId) => {
+      const wasFacilitator = userId === facilitatorUserId
+      segmentIo.track({
```
mattkrick, line 126: duplicate code 3 times & it's probably time to refactor...

can we refactor this so the majority of the logic is shared across meetings? maybe we pass in a meeting object to a function called `sendMeetingEndToSegment` or similar

packages/server/graphql/types/EndCheckInPayload.ts
```diff
@@ -0,0 +1,66 @@
...
+import makeMutationPayload from './makeMutationPayload'
+import ActionMeeting from './ActionMeeting'
+import {resolveNewMeeting} from '../resolvers'
+import errorFilter from '../errorFilter'
+
+export const EndCheckInSuccess = new GraphQLObjectType<any, GQLContext>({
+  name: 'EndCheckInSuccess',
+  fields: () => ({
+    isKill: {
+      type: GraphQLBoolean,
```
mattkrick, line 17: can this be nonnull?

packages/server/graphql/types/EndRetrospectivePayload.ts
```diff
@@ -0,0 +1,48 @@
...
+import {GQLContext} from '../graphql'
+import TimelineEvent from './TimelineEvent'
+import makeMutationPayload from './makeMutationPayload'
+import RetrospectiveMeeting from './RetrospectiveMeeting'
+
+export const EndRetrospectiveSuccess = new GraphQLObjectType<any, GQLContext>({
+  name: 'EndRetrospectiveSuccess',
+  fields: () => ({
+    isKill: {
+      type: GraphQLBoolean,
```
mattkrick, line 13: nonnull

mattkrick (changes requested):
looking good! just a few little things

mattkrick:
@tiffanyhan can you give this a quick test?

One test to be sure of that is broken today:

- start a check in meeting, get to the agenda items
- start a retro meeting
- add an agenda item to the check in meeting
- discuss that agenda item (make sure the comments get added!)

strategy: try to break it by running 2 different meetings as 2 different users (in 2 browsers) at the same time without hitting the browser refresh button.

packages/client/subscriptions/NotificationSubscription.ts
nickoferrall: As we're keeping the old mutations, I was thinking that we may as well keep the references to the old mutations too. Then we can remove everything once the new mutations are up and running. I'd be happy to remove these references though if you think it'd be cleaner

mattkrick: i like it! can you create a new issue like "remove old checkin mutation" that we can process in a month or 2?

nickoferrall: Sure thing 👍

tiffanyhan (approved):
> @tiffanyhan can you give this a quick test?
> 
> One test to be sure of that is broken today:
> 
> * start a check in meeting, get to the agenda items
> * start a retro meeting
> * add an agenda item to the check in meeting
> * discuss that agenda item (make sure the comments get added!)
> 
> strategy: try to break it by running 2 different meetings as 2 different users (in 2 browsers) at the same time without hitting the browser refresh button.



Looks good to me! I was able to replicate the bug of not being able to add any discussion topics to existing agenda items when starting a retro while a checkin meeting is in progress. Works on the new branch.

packages/server/graphql/mutations/endCheckIn.ts
nickoferrall: Interesting - it could get called twice because of something going wrong in the codebase or would the user be able to trigger it twice?

mattkrick: user could trigger it twice. a good pen-tester will always rapid fire the same query over & over hoping to break something. few years back, a guy figured out he could get free starbucks giftcards doing this. albeit ours isn't as exciting as that...

nickoferrall: haha good to know!

PR #4331, opened 2020-11-10
S3 images

tiffanyhan:


tiffanyhan:
**To replicate UI bug:**
- set `FILE_STORE_PROVIDER='s3'` in `.env`file
- go to user profile and upload a new avatar image
- in devtools console network tab, observe that the new avatar url is returned to the client
- observe that the client does not load the new image automatically. a 403 fetch is logged in the browser console.
- if the page is refreshed, the avatar now loads successfully without a 403.

tiffanyhan:
**To replicate Static Server singleton problem:**
- put a console log in `StaticServer` constructor in `packages/server/utils/StaticServer.ts`
- start the server and see that the above log outputs twice
- comment out `import {staticServer} from '../utils/serveStatic'` in `packages/server/fileStorage/LocalFileSystemManager.ts`; observe that log outputs only once now.

Tried: Using an object literal, `Object.freeze`, creating a separate file whose only job is to instantiate the static server, importing using `require`, turning off hmr, building for production and running with `yarn start`.

Apart from the singleton problem, how should we change the behavior of `StaticServer` to account for dynamically generated user content?

package.json
```diff
@@ -58,6 +58,7 @@
     "prettier": "^1.19.1",
     "relay-compiler": "^8.0.0",
     "relay-config": "^8.0.0",
+    "sharp": "^0.26.2",
```
tiffanyhan, line 61: oops, move to server package.json

packages/server/utils/validateSVGUpload.ts
```diff
@@ -0,0 +1,17 @@
...
+  if (!safeBuffer) throw new Error('Attempted Stored XSS attack')
+}
+
+const validateSVGUpload = async (buffer: Buffer): Promise<Buffer> => {
+  await validateSafeSVG(buffer)
+  const pngBuffer = await sharp(buffer).png().toBuffer()
+  return pngBuffer
+}
+
+export default validateSVGUpload
```
tiffanyhan, line 17: move to two separate functions with no default export in this module. i don't like that it does two things at once.

tiffanyhan:
**still todos:**

- [ ] deprecate old mutations
- [ ] move relevant file to `/fileStorage` folder
- [ ] PR `graphql-trebuchet-client` package to expect uploadables in `OperationPayload`?

packages/client/Atmosphere.ts
```diff
@@ -115,17 +123,31 @@ export default class Atmosphere extends Environment {
...
         accept: 'application/json',
         Authorization: this.authToken ? `Bearer ${this.authToken}` : '',
-        'content-type': 'application/json',
         'x-correlation-id': connectionId || ''
       },
-      body: JSON.stringify(body)
-    })
+    }
+    const uploadables = body.payload.variables?.uploadables
+    delete body.payload.variables?.uploadables
```
tiffanyhan, line 136: hacky way around `graphql-trebuchet-client` not expecting `uploadables`. I forked and cloned the repo, but was unable to run `yarn build` successfully even with no changes to the code.

packages/client/validation/makeUserServerSchema.js
```diff
@@ -1,13 +1,9 @@
-import {urlRegex} from './regex'
+import { url } from 'parabol-client/validation/templates'
 import legitify from './legitify'
 
 export default function makeUserServerSchema() {
   return legitify({
-    picture: (value) =>
-      value
-        .trim()
-        .matches(urlRegex, 'that picture url doesn’t look quite right')
```
tiffanyhan, line 9: this `urlRegex` wasn't recognizing a url with `localhost` in it as a proper url. based on my reading, it seems using a regex for url validation is error prone. several people recommended using a built-in URL parser for validation, which seems a simpler and more robust solution.

packages/client/validation/templates.js
```diff
@@ -72,4 +72,18 @@ export const makeTeamNameSchema = (teamNames) => (value) =>
...
     .test((val) => teamNames.includes(val) && 'That name is already taken')
 
-export const url = (value) => value.matches(urlRegex)
+export const url = (value) => 
+  value
+    .trim()
+    .satisfies(
+      (value) => {
+        try {
+          new URL(value)
```
tiffanyhan, line 81: see my above comment

packages/server/fileStorage/normalizeAvatarUpload.ts
```diff
@@ -0,0 +1,15 @@
+import sharp from 'sharp'
+
+const normalizeAvatarUpload = async (
+  ext: string,
+  buffer: Buffer
+): Promise<[string, Buffer]> => {
+  const resizedBuffer = await sharp(buffer).resize(200, 200).toBuffer()
```
tiffanyhan, line 7: maybe need to check if size is bigger than 200x200 before resizing

packages/server/graphql/httpGraphQLHandler.ts
```diff
@@ -64,21 +65,28 @@ const httpGraphQLBodyHandler = async (
...
-    await httpGraphQLBodyHandler(res, body, authToken, connectionId, ip)
-  } else {
+  const contentTypeHeader = req.getHeader('content-type')
+  const ct = Object.keys(contentTypeBodyParserMap).find((key) => contentTypeHeader.startsWith(key))
+  if (!ct) {
     res.writeStatus('415').end()
+    return
+  }
+  const parseFn = contentTypeBodyParserMap[ct]
+  const body = await parseFn({res, req})
```
tiffanyhan, line 84: call signature of `parseBody` changed so that we can have a unified call to whichever fn here.

packages/server/graphql/mutations/createOrgPicturePutUrl.js
```diff
@@ -35,7 +35,7 @@ const createOrgPicturePutUrl = {
 
     // RESOLUTION
     const partialPath = `Organization/${orgId}/picture/${shortid.generate()}.${ext}`
-    const url = await getS3PutUrl(contentType, contentLength, partialPath)
+    const url = await getS3SignedPutUrl(contentType, contentLength, partialPath)
```
tiffanyhan, line 38: this will all be deprecated

packages/server/graphql/mutations/createUserPicturePutUrl.js
```diff
@@ -35,11 +35,11 @@ const createUserPicturePutUrl = {
     // RESOLUTION
     const imgId = shortid.generate()
     const partialPath = `User/${userId}/picture/${imgId}.${ext}`
-    const url = await getS3PutUrl(contentType, contentLength, partialPath)
+    const url = await getS3SignedPutUrl(contentType, contentLength, partialPath)
     let pngUrl
     if (pngVersion) {
       const partialPath = `User/${userId}/picture/${imgId}.png`
-      pngUrl = await getS3PutUrl(contentType, contentLength, partialPath)
+      pngUrl = await getS3SignedPutUrl(contentType, contentLength, partialPath)
```
tiffanyhan, line 42: this will all be deprecated

packages/server/graphql/mutations/helpers/updateUserProfile.ts
```diff
@@ -0,0 +1,104 @@
...
+    })
+    segmentIo.identify({
+      userId,
+      traits: {
+        name: validUpdatedUser.preferredName
+      }
+    })
+  }
+
+  const teamIds = teamMembers.map(({teamId}) => teamId)
```
tiffanyhan, line 96: resolvers moved to its own helper file, so it can be called from `upload` mutations

packages/server/graphql/mutations/uploadUserImage.ts
```diff
@@ -0,0 +1,67 @@
...
+import getFileStoreManager from '../../fileStorage/getFileStoreManager'
+import FileStoreManager from '../../fileStorage/FileStoreManager'
+import updateUserProfile from './helpers/updateUserProfile'
+import UpdateUserProfilePayload from '../types/UpdateUserProfilePayload'
+
+export default {
+  type: UpdateUserProfilePayload,
+  description: 'Upload an image for a user avatar',
+  args: {
+    file: {
```
tiffanyhan, line 17: only file is required, bc we don't trust client supplied metadata anyway

packages/server/package.json
```diff
@@ -129,6 +130,7 @@
     "rethinkdbdash": "^2.3.31",
     "samlify": "^2.6.1",
     "secure-compare": "^3.0.1",
+    "sharp": "^0.26.2",
```
tiffanyhan, line 133: sharp: 

> Resizing an image is typically 4x-5x faster than using the quickest ImageMagick and GraphicsMagick settings due to its use of libvips.

packages/server/package.json
```diff
@@ -99,6 +99,7 @@
     "aws-sdk": "^2.99.0",
     "base64url": "^3.0.1",
     "bcrypt": "^5.0.0",
+    "busboy": "^0.3.1",
```
tiffanyhan, line 102: busboy is used internally in `graphql-upload` and several other packages. supports streams as a first-class citizen.

packages/server/parseFormBody.ts
```diff
@@ -0,0 +1,80 @@
...
+  })
+}
+
+const parseFormBody = ({res, req}: {res: HttpResponse; req: HttpRequest}): Promise<JSON | null> => {
+  const parser = new Busboy({
+    headers: reqHeaders(req),
+    limits: {
+      fields: 1,
+      files: 1,
+      fileSize: APP_MAX_AVATAR_FILE_SIZE
```
tiffanyhan, line 52: safest limits for now

packages/server/parseFormBody.ts
```diff
@@ -0,0 +1,80 @@
...
+    })
+    fileStream.on('end', () => {
+      const parsedFile = {
+        buffer,
+        contentType
+      } as {
+        contentType: string
+        buffer: Buffer
+      }
+      resolve(parsedFile)
```
tiffanyhan, line 38: parses the entire file before sending to => redis => graphql executor. i think we could technically use [Redis Streams](https://redis.io/topics/streams-intro) for the optimal solution. this was simple and works, so i'll see what tensions come up.

packages/server/utils/getS3SignedPutUrl.js
```diff
@@ -0,0 +1,7 @@
+import { s3SignPutObject } from './s3'
+import getS3PutUrl from './getS3PutUrl'
+
+export default function getS3SignedPutUrl(contentType, contentLength, partialPath) {
+  const pathname = getS3PutUrl(partialPath)
+  return s3SignPutObject(pathname, contentType, contentLength, 'public-read')
+}
```
tiffanyhan, line 7: will be deprecated

packages/server/utils/getS3PutUrl.ts
```diff
@@ -0,0 +1,8 @@
+import protocolRelativeUrl from './protocolRelativeUrl'
+import path from 'path'
+import {APP_CDN_USER_ASSET_SUBDIR} from 'parabol-client/utils/constants'
+
+export default function getS3PutUrl(partialPath) {
+  const baseUrl = protocolRelativeUrl.parse(process.env.CDN_BASE_URL)
+  return path.join(baseUrl.pathname, APP_CDN_USER_ASSET_SUBDIR, partialPath)
+}
```
tiffanyhan, line 8: will be deprecated

packages/client/Atmosphere.ts
mattkrick: well this is interesting! what was the error?

packages/server/parseFormBody.ts
mattkrick: i'm so glad you brought this up! i've been thinking about this too--- do we send the whole thing to graphql? do we stream it to redis? do we store it to disk here & then just return a file path to graphql? @adaniels-parabol probably has better thoughts than me on this.

packages/server/utils/getS3PutUrl.ts
mattkrick: feel free to put deprecated notes as comments into the code itself! that way we're more likely to remove it in the future

packages/server/parseFormBody.ts
```diff
@@ -0,0 +1,80 @@
...
+  const headers = {} as {[key: string]: string}
+  req.forEach((k, v) => (headers[k] = v))
+  return headers
+}
+
+const parseFile = (fileStream, contentType) => {
+  return new Promise((resolve, reject) => {
+    let buffer: Buffer
+    fileStream.on('data', (curBuf) => {
+      buffer = buffer ? Buffer.concat([buffer, curBuf]) : Buffer.concat([curBuf])
```
mattkrick, line 28: this is the real steak & potatoes of the PR! this piece concerns me a bit because we already copied the buffer once as it came in & here we're copying it again.
i think we can entirely get rid of busboy & use uWS. Alex just released a version with multipart handling (https://github.com/uNetworking/uWebSockets.js/releases/tag/v18.7.0). 

Note that upgrading has a breaking change so you'll have to refactor `handleOpen` to use an upgrade event. He has examples included in the repo (https://github.com/uNetworking/uWebSockets.js/releases/tag/v18.0.0). You can break that into its own PR if you like

mattkrick (changes requested):
fantastic work! code is very clean & readable, most of the comments are just nits.
Only big piece is upgrading uWS to the latest so we can take advantage of its native `getParts` so we don't have to double clone the buffer

packages/client/Atmosphere.ts
tiffanyhan: I saw that you upgraded already but I was just getting a bunch of type errors previously. not sure if i was missing s step before building.

```
clean-webpack-plugin: /Users/tiffanyhan/resources/graphql-trebuchet-client/dist/**/* has been removed.
ℹ ｢atl｣: Using typescript@3.9.7 from typescript
ℹ ｢atl｣: Using tsconfig.json from /Users/tiffanyhan/resources/graphql-trebuchet-client/tsconfig.json
ℹ ｢atl｣: Checking started in a separate process...
✖ ｢atl｣: Checking finished with 7 errors
Hash: e7386da1a7fd1ad8d6fd
Version: webpack 4.16.1
Time: 2711ms
Built at: 11/17/2020 11:37:42 AM
                        Asset       Size  Chunks             Chunk Names
                     index.js   8.38 KiB   index  [emitted]  index
                 index.js.map   11.1 KiB   index  [emitted]  index
             types/index.d.ts  181 bytes          [emitted]  
     types/GQLHTTPClient.d.ts  491 bytes          [emitted]  
types/GQLTrebuchetClient.d.ts   1.89 KiB          [emitted]  
Entrypoint index = index.js index.js.map
[./src/GQLHTTPClient.ts] 553 bytes {index} [built]
[./src/GQLTrebuchetClient.ts] 2.52 KiB {index} [built]
[./src/index.ts] 146 bytes {index} [built]

ERROR in [at-loader] ./src/GQLTrebuchetClient.ts:76:18 
    TS2769: No overload matches this call.
  Overload 1 of 2, '(event: Events.DATA | Events.CLOSE | Events.TRANSPORT_SUPPORTED, listener: (m: string | boolean | object | ArrayBuffer | SharedArrayBuffer | Blob | ArrayBufferView | ClosePayload, ...args: any[]) => void): any', gave the following error.
    Argument of type '"data"' is not assignable to parameter of type 'Events.DATA | Events.CLOSE | Events.TRANSPORT_SUPPORTED'.
  Overload 2 of 2, '(event: Events.KEEP_ALIVE | Events.TRANSPORT_CONNECTED | Events.TRANSPORT_RECONNECTED | Events.TRANSPORT_DISCONNECTED, listener: () => void): any', gave the following error.
    Argument of type '"data"' is not assignable to parameter of type 'Events.KEEP_ALIVE | Events.TRANSPORT_CONNECTED | Events.TRANSPORT_RECONNECTED | Events.TRANSPORT_DISCONNECTED'.

ERROR in [at-loader] ./src/GQLTrebuchetClient.ts:79:18 
    TS2769: No overload matches this call.
  Overload 1 of 2, '(event: Events.DATA | Events.CLOSE | Events.TRANSPORT_SUPPORTED, listener: (m: string | boolean | object | ArrayBuffer | SharedArrayBuffer | Blob | ArrayBufferView | ClosePayload, ...args: any[]) => void): any', gave the following error.
    Argument of type '"reconnected"' is not assignable to parameter of type 'Events.DATA | Events.CLOSE | Events.TRANSPORT_SUPPORTED'.
  Overload 2 of 2, '(event: Events.KEEP_ALIVE | Events.TRANSPORT_CONNECTED | Events.TRANSPORT_RECONNECTED | Events.TRANSPORT_DISCONNECTED, listener: () => void): any', gave the following error.
    Argument of type '"reconnected"' is not assignable to parameter of type 'Events.KEEP_ALIVE | Events.TRANSPORT_CONNECTED | Events.TRANSPORT_RECONNECTED | Events.TRANSPORT_DISCONNECTED'.

ERROR in [at-loader] ./src/GQLTrebuchetClient.ts:82:11 
    TS2345: Argument of type '{ id: string; type: string; payload: OperationPayload; }' is not assignable to parameter of type 'Data'.
  Object literal may only specify known properties, and 'id' does not exist in type 'ArrayBuffer | SharedArrayBuffer | Blob | ArrayBufferView'.

ERROR in [at-loader] ./src/GQLTrebuchetClient.ts:121:28 
    TS2345: Argument of type '{ id: string; type: string; }' is not assignable to parameter of type 'Data'.
  Object literal may only specify known properties, and 'id' does not exist in type 'ArrayBuffer | SharedArrayBuffer | Blob | ArrayBufferView'.

ERROR in [at-loader] ./src/GQLTrebuchetClient.ts:140:28 
    TS2345: Argument of type '{ id: string; type: string; payload: OperationPayload; }' is not assignable to parameter of type 'Data'.
  Object literal may only specify known properties, and 'id' does not exist in type 'ArrayBuffer | SharedArrayBuffer | Blob | ArrayBufferView'.

ERROR in [at-loader] ./src/GQLTrebuchetClient.ts:142:43 
    TS2345: Argument of type '{ type: string; payload: OperationPayload; }' is not assignable to parameter of type 'Data'.
  Object literal may only specify known properties, and 'payload' does not exist in type 'ArrayBuffer | SharedArrayBuffer | Blob | ArrayBufferView'.

ERROR in [at-loader] ./src/GQLTrebuchetClient.ts:153:26 
    TS2345: Argument of type '{ id: string; type: string; payload: OperationPayload; }' is not assignable to parameter of type 'Data'.
  Object literal may only specify known properties, and 'id' does not exist in type 'ArrayBuffer | SharedArrayBuffer | Blob | ArrayBufferView'.
error Command failed with exit code 2.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.

```

packages/server/graphql/mutations/createOrgPicturePutUrl.js
tiffanyhan: oh i didn't know about this field! i'll add

packages/server/parseFormBody.ts
tiffanyhan: > do we store it to disk here & then just return a file path to graphql

^^ i think that would work for the on-premise local file system scenario, but not for Parabol Cloud S3 scenario. The reason being because the main server and gqlexecutor may not always be on the same machine.

i think part of the difficulty is we're really trying to solve the problem with a unified solution for different deployment scenarios.

> do we send the whole thing to graphql?

^^ do you mean directly to gqlexecutor? how would this work?

the one thing i like about sending it all to redis is that we don't make any assumptions about what we want to do with the contents until we hit the graphql resolver.

packages/server/fileStorage/getFileStoreManager.ts
```diff
@@ -0,0 +1,20 @@
...
+let fileStoreManager: FileStoreManager
+const managers = {
+  s3: S3Manager,
+  local: LocalFileSystemManager
+}
+
+const getFileStoreManager = () => {
+  if (!fileStoreManager) {
+    const fileStoreProvider = process.env.FILE_STORE_PROVIDER!
+    const Manager = managers[fileStoreProvider] ?? LocalFileSystemManager
```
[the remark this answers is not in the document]
tiffanyhan: yea i agree with this. i'll change to throw an error if the env var isn't defined. always reminds me i should update the `action-devops` repo too!

packages/client/validation/Legitity.ts
```diff
@@ -28,6 +28,13 @@ class Legitity {
     return this
   }
 
+  satisfies(cb: (any) => boolean, msg?: string) {
```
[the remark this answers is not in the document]
tiffanyhan: yes it would! looks like i missed that method. i'll refactor to use it and get rid of the extra `satisfies`.

packages/server/fileStorage/FileStoreManager.ts
```diff
@@ -0,0 +1,31 @@
+import shortid from 'shortid'
+
+export interface PutFileOptions {
+  partialPath: string
+  buffer: Buffer
+}
+
+export default abstract class FileStoreManager {
+  _putFileCb: undefined | ((fullPath: string) => void)
```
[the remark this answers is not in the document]
tiffanyhan: i had previously tried `private` but that didn't work with `abstract` which i guess makes sense, but `protected` does!

packages/server/graphql/httpGraphQLHandler.ts
```diff
@@ -64,21 +65,28 @@ const httpGraphQLBodyHandler = async (
...
-  if (contentType.startsWith('application/json')) {
-    const body = await parseBody(res)
-    if (!body) {
-      res.writeStatus('422').end()
-      return
-    }
-    await httpGraphQLBodyHandler(res, body, authToken, connectionId, ip)
-  } else {
+  const contentTypeHeader = req.getHeader('content-type')
+  const ct = Object.keys(contentTypeBodyParserMap).find((key) => contentTypeHeader.startsWith(key))
```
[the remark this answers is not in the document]
tiffanyhan: ah i think it's just because the header can have more content to it, e.g.:

```
Content-Type: text/html; charset=UTF-8
Content-Type: multipart/form-data; boundary=something
```

packages/server/graphql/httpGraphQLHandler.ts
tiffanyhan: that is very very good to know. so any data from the req we want to hold onto for async things we do via closures right? i'll hold off on fixing this until we get the uwebsockets upgraded

tiffanyhan: on second thought, i'm probably in agreement with:

> it might be more expressive to use a different function call/signature for each

just for the fact of getting deterministic autocompletions for each function.

packages/client/Atmosphere.ts
```diff
@@ -115,17 +123,31 @@ export default class Atmosphere extends Environment {
...
+    if (uploadables) {
+      if (!window.FormData) throw new Error('Uploading files without `FormData` not supported')
+      const formData = new FormData()
+      formData.append('body', JSON.stringify(body))
+      Object.keys(uploadables).forEach(key => {
+        if (!Object.prototype.hasOwnProperty.call(uploadables, key)) return
+        formData.append(key, uploadables[key])
```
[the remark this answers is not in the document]
tiffanyhan: resolved via #4346

packages/client/Atmosphere.ts
```diff
@@ -115,17 +123,31 @@ export default class Atmosphere extends Environment {
...
-    })
+    }
+    const uploadables = body.payload.variables?.uploadables
+    delete body.payload.variables?.uploadables
+    if (uploadables) {
+      if (!window.FormData) throw new Error('Uploading files without `FormData` not supported')
+      const formData = new FormData()
+      formData.append('body', JSON.stringify(body))
+      Object.keys(uploadables).forEach(key => {
+        if (!Object.prototype.hasOwnProperty.call(uploadables, key)) return
```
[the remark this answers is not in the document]
tiffanyhan: Resolved via #4346

mattkrick:
uWS updated!
i really like your idea of redis streams. feel free to implement here if you want, but if scope creep is a concern, i completely understand just passing it all as 1 large chunk.

packages/client/Atmosphere.ts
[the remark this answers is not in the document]
tiffanyhan: hey @mattkrick , after attempting to work with your changes in #4346 , I think there may have been miscommunication. the way the code was written originally, there was expected to be only two fields: (1) `"body"`; and (2) `"file"`.  "body" contains a json stringified version of `StartMessage` containing `OperationPayload`. The only way an uploadable could overwrite anything was if the uploadable was named "body".

in my most recent changes, "body" is kept as-is, but each uploadable is prefixed with "uploadables" key. An uploadable can be named anything. E.g. `uploadables.file` or `uploadables.image`. This way, there's no way an uploadable can overwrite another field when constructing `formData`.*

i prefer this way because we avoid having to manually deserialize and reconstruct the request message. 

Curious to know your thoughts and if you still have any concerns.

*(Later in `parseFormBody`, the uploadable overwrites its same key name in `body.payload.variables`, but this is by design, so we can simply access the uploadable as a variable in the mutation resolver.)

packages/server/parseFormBody.ts
tiffanyhan: upgraded to v18.10.0 bc [the interface of `MultipartField[]` changed from a map to an object data structure](https://github.com/uNetworking/uWebSockets.js/issues/375), and it is easier to work with objects in typescript.

tiffanyhan: also forked the repo bc [the type definition file in the binaries](https://github.com/uNetworking/uWebSockets.js/tree/v18.10.0) is out-of-date than [the one on master](https://github.com/uNetworking/uWebSockets.js/blob/master/docs/index.d.ts), and the out of date one doesn't include the new api for `getParts`.

tiffanyhan:
> uWS updated!
> i really like your idea of redis streams. feel free to implement here if you want, but if scope creep is a concern, i completely understand just passing it all as 1 large chunk.

thanks for upgrading! i'm okay with send it all as 1 large chunk for now

packages/client/Atmosphere.ts
```diff
@@ -115,16 +134,19 @@ export default class Atmosphere extends Environment {
     })
   }
 
-  fetchHTTP = async (body: object, connectionId?: string) => {
+  fetchHTTP = async (body: FetchHTTPData, connectionId?: string) => {
+    const uploadables = body.payload.uploadables
+    delete body.payload.uploadables
```
mattkrick, line 139: nit: let's fully encapsulate uploadables into a single function. e.g.

```ts
body.payload.uploadables ? toFormData(data) : JSON.stringify(body)
```

that way everything that has to do with uploadables is all done in 1 function that's inside a conditional

packages/server/parseFormBody.ts
```diff
@@ -0,0 +1,52 @@
...
+}
+
+const parseFormBody = async ({res, contentType}: ParseFormBodySignature): Promise<JSON | null> => {
+  const resBuffer = await parseRes(res)
+  if (!resBuffer) return null
+  const parts = getParts(resBuffer, contentType) as MultipartField[]
+  if (!parts) return null
+
+  try {
+    let parsedBody
```
mattkrick, line 33: `let parsedBody: unknown`
this is super important because the client could send us garbage & we want to be sure that their garbage doesn't make our server throw an error

packages/server/parseFormBody.ts
```diff
@@ -0,0 +1,52 @@
...
+    parts.forEach(({name, data, type}) => {
+      if (name === 'body') {
+        parsedBody = JSON.parse(Buffer.from(data).toString())
```
mattkrick, line 36: parsing the string could throw an error

packages/server/parseFormBody.ts
```diff
@@ -0,0 +1,52 @@
...
+        return
+      }
+      if (!name.startsWith('uploadables')) return
+      const [, key] = name.split('.')
+      parsedBody.payload.variables[key] = {
```
mattkrick, line 41: we can't be sure that `parsedBody.payload` exists, or that it includes variables or that it comes before the `uploadable` key does.
we want to get all the keys & put them in an object & then after the loop, put that object inside the parsedBody iif it exists.

mattkrick (changes requested):
looking good! just a couple extra items

packages/client/Atmosphere.ts
tiffanyhan: that's a good idea, changed!

packages/server/parseFormBody.ts
tiffanyhan: this is true. the entire `forEach` is currently contained within a try-catch block. do you think a nested try-catch block inside the loop should be added?

my thinking is that we expect a parsable body field, so if the client passes an un-parseable body field, we should immediately throw an error which will ultimate return 422 client error.

packages/server/parseFormBody.ts
tiffanyhan: @mattkrick is it wrong to assume that all http requests going to this handler (httpGraphQLHandler) will be the type below? i'm assuming all requests should conform to this interface, which would mean that `body.payload` should exist. if not, the server will return 422 client error.

```
interface FetchHTTPData {
  type: 'start' | 'stop'
  payload: OperationPayload
}
```

i am also targeting `variables` instead of `parsedBody` so that the uploadable will be made available to the mutation resolver as a variable. the other option was putting it in context, but i thought this way was cleaner. There is the possibility a variable might accidentally get overwritten, but only if the developer passed an uploadable under the same name key as another passed variable. i think the mistake would probably throw an error bc of type checking, which seems okay imo. what do you think?

packages/server/parseFormBody.ts
mattkrick: good thinking, yeah, a single try/catch should work, if the client fails that they're probably up to no good anyways.

packages/server/parseFormBody.ts
```diff
@@ -24,26 +34,34 @@ const parseRes = (res: HttpResponse) => {
...
-      const [, key] = name.split('.')
-      parsedBody.payload.variables[key] = {
-        contentType: type,
-        buffer: Buffer.from(data)
-      } as Uploadable
     })
-    return parsedBody
+    if (!parsedBody || typeof parsedBody !== 'object') return null
+    if (Object.keys(parsedUploadables).length) {
+      const validParsedBody = parsedBody as FetchHTTPData
```
mattkrick, line 58: what if the body is `'{}'`?
for all parsed JSON coming from an untrusted client we've gotta do runtime validation

packages/server/parseFormBody.ts
```diff
@@ -24,26 +34,34 @@ const parseRes = (res: HttpResponse) => {
...
+      validParsedBody.payload.variables = {
+        ...validParsedBody.payload.variables,
+        ...parsedUploadables
+      }
+    }
+    return parsedBody as JSON
```
mattkrick, line 64: after validation, we know the type is going to be a superset of `FetchHTTPData`, so we can use that since it's more specific than JSON

packages/server/parseFormBody.ts
tiffanyhan: yea that's a good point. i was thinking to just throw the error and catch if the expected properties didn't exist in body. doing a runtime check is much more explicit though. i just added a user defined type guard, then type cast after the json is validated to be the correct shape.

it wasn't clear to me before the difference btwn `any` vs `unknown`, but now i see how it's useful using `unknown` in this case.

mattkrick:
LGTM! only thing left to do is to update the servers to support the new env var. 
@tiffanyhan would you mind adding `FILE_STORE_PROVIDER` to the env on the dev, staging, and prod servers? 
after you ssh into each server, it'll look like `dokku config:set --no-restart web FILE_STORE_PROVIDER='s3'`
http://dokku.viewdocs.io/dokku/configuration/environment-variables/

tiffanyhan:
yes, env var added to dev, staging and prod! thanks for the thorough review and testing on this one @mattkrick

PR #4332, opened 2020-11-11
add-new-route

mattkrick:
@tiffanyhan here's how i'd handle that case of self-hosted content.
the static server has a lot of sharp edges to prevent folks from doing rude things.
if the customer is gonna be self-hosted, we don't really have to worry about that.

basically, we store everything that they upload in `/self-hosted`.
Then, when they want to retrieve that, we look in `/self-hosted`
If we don't find it, we send em a 404.
If we do, we pipe it to them (this has a slight overhead vs. sending the whole thing at once, but means it can support ver large files, if the need shall arise)

note: untested!

### DIFF
--- a/packages/server/selfHostedHandler.ts
+++ b/packages/server/selfHostedHandler.ts
@@ -1,0 +1,36 @@
+import fs from 'fs'
+import mime from 'mime-types'
+import path from 'path'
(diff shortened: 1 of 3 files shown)


tiffanyhan:
oh i think that sounds reasonable. thanks for looking into this!

PR #4334, opened 2020-11-12
Add Deck Activity Peekers

mattkrick:
When you hover a card in the deck, everyone else in the meeting is notified that you are thinking. this is shown by animating your avatar in from the right side. When you cast your vote, the avatar will animate to the votedUsers row. Otherwise, if you leave the deck without casting a vote, your avatar just hides on the right side.

### DIFF
--- a/packages/client/components/EstimatePhaseArea.tsx
+++ b/packages/client/components/EstimatePhaseArea.tsx
@@ -6,6 +6,9 @@ import useBreakpoint from '~/hooks/useBreakpoint'
 import {Breakpoint} from '~/types/constEnums'
 import PokerCardDeck from './PokerCardDeck'
 import DeckActivityAvatars from './DeckActivityAvatars'
+import {createFragmentContainer} from 'react-relay'
+import graphql from 'babel-plugin-relay/macro'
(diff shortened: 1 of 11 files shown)


packages/client/components/EstimatePhaseArea.tsx
```diff
@@ -52,14 +55,28 @@ const containerStyle = {
...
   const isDesktop = useBreakpoint(Breakpoint.SIDEBAR_LEFT)
 
   const onChangeIdx = (idx) => {
     setActiveIdx(idx)
   }
 
+  const getVotedUserEl = (_userId: string) => {
+    // mock Element
+    return {
+      getBoundingClientRect() {
```
mattkrick, line 74: @ackernaut here's the missing piece where our branches will have to merge. basically, i need this function to return the element of that user's avatar in the voted users array.

packages/client/components/PokerCard.tsx
```diff
@@ -51,6 +51,17 @@ interface CardBaseProps {
   isSelected: boolean,
   totalCards: number
 }
+
+// const getRotation = (totalCards: number, idx: number) => {
```
mattkrick, line 55: coming back to this, i think a _very_ subtle arch in the hand will look good & allow us to fit in an extra card or 2

PR #4337, opened 2020-11-13
add deck curve (in debug mode)

mattkrick:
hey @ackernaut i've gotten as far as i can get for this. Can you take it from here?
If you like, I can wrangle those other animations

Burndown
- [ ] The deck is centered on desktop/mobile
- [ ] the collapsed deck offset is correct on desktop/mobile (hit c to collapse)
- [ ] a max radius/tilt is determined for desktop
- [ ] the radius & tilt are determined based on window width (hit q/w to change radius and e/r to change tilt)
- [ ] the MAX_HIDDEN variable is deemed acceptable (The % of the outermost cards that are hidden below the fold)
- [ ] an initial ENTER transition occurs for the whole deck (should only happen once, when the deck is first rendered)


ackernaut:
- [ ] The deck is centered on desktop/mobile
- [ ] the collapsed deck offset is correct on desktop/mobile (hit c to collapse)
- [ ] a max radius/tilt is determined for desktop
- [ ] the radius & tilt are determined based on window width (hit q/w to change radius and e/r to change tilt)
- [ ] the MAX_HIDDEN variable is deemed acceptable (The % of the outermost cards that are hidden below the fold)
- [ ] an initial ENTER transition occurs for the whole deck (should only happen once, when the deck is first rendered)

ackernaut:
I’m going to merge so I can start adding votes via the deck. We can clean up anything later.

ackernaut (commented):
lgtm

PR #4346, opened 2020-11-16
support nested formdata pieces

mattkrick:
this PR adds a convention to convert nested objects into a file FormData schema.
Nested fields are store as dot separated values (e.g. `{query, 'payload.uploadables.file': File}`)

@tiffanyhan this cleans up a few of the rough edges for file uploads from the client. after migrating to the new uWS we can use this same convention to reconstruct the object on the server


tiffanyhan (approved):
thanks for upgrading `graphql-trebuchet-client` and improving the formData creation! i'm happy to see how much cleaner the code is now. @mattkrick

PR #4347, opened 2020-11-16
Add migration for modified Poker Templates

### LINKED ISSUE #4303: Update default poker template
looks like i didn't check the migrations well enough.
The default template should only have a single dimension, because almost all users report only voting on 1 dimension.

AC
- [ ] There is an "Estimated Effort" template that has 1 dimension: story points. It is the default template
- [ ] There is a "weighted shortest job first" template that has 2 dimensions: story points & story value
- [ ] A migration exists that adds these templates & removes the other ones

Effort: 4 hours

tianrunhe:
Resolves #4303 .


mattkrick:
LGTM!

PR #4354, opened 2020-11-18
Fixes fatal client-side bugs found in sentry

mattkrick:
In sentry, we found 3 fatal client side bugs that are rooted in objects not existing in the client cache where they are guaranteed to exist. This leads me to believe the bug is in Relay, likely as a race condition in how it garbage collects. It is possible that the Relay team has fixed this bug in a newer version (even if they didn't know they fixed it or it wasn't in the changelog).

For this reason, we should up the priority of bumping the relay version. There will be breaking changes & we'll likely have to re-fork it again with our fixes that won't get accepted to their master branch.

@nickoferrall


jordanh:
🎊

PR #4357, opened 2020-11-18
upgrade uws

mattkrick:
to handle multipart form data, we need to upgrade uWS.
There are a few breaking changes:
- uWS v18 no longer passes the `request` into the open handler. instead, it introduces an upgrade handler. Once auth has been performed in the upgrade handler, we can pass in the authToken & IP address to the open handler.
- it added a native way to get proxied IP addresses that always returns IPv6, so we use that instead. we'll need to verify this in prod to make sure we don't rate limit folks unfairly


mattkrick:
merging to master for this ship because if something sockety goes wrong, it'd be nice to make sure it's going wrong here vs. in the next

PR #4359, opened 2020-11-19
Scope add task

tiffanyhan:


packages/client/mutations/handlers/handleJiraCreateIssue.ts
```diff
@@ -13,19 +13,15 @@ const handleJiraCreateIssue = (payload: RecordProxy<any>, store: RecordSourceSel
   const teamMember = store.get(teamMemberId)
   const integrations = teamMember?.getLinkedRecord('integrations')
   const atlassian = integrations?.getLinkedRecord('atlassian')
-  const jiraSearchQueryId = toJiraSearchQueryId(meetingId)
-  const jiraSearchQuery = store.get(jiraSearchQueryId)
+  const meeting = store.get(meetingId)
+  const jiraSearchQuery = meeting?.getLinkedRecord('jiraSearchQuery')
```
tiffanyhan, line 17: fixes a bug in which `const jiraSearchQuery = store.get(jiraSearchQueryId)` never returns anything

packages/client/utils/relay/insertNode.ts
```diff
@@ -0,0 +1,13 @@
...
+  conn: RecordProxy,
+  node: RecordProxy,
+  store: RecordSourceSelectorProxy,
+  type: string,
+  cursorValue: string = new Date().toISOString()
+) => {
+  const newEdge = ConnectionHandler.createEdge(store, conn, node, type)
+  newEdge.setValue(cursorValue, 'cursor')
+  ConnectionHandler.insertEdgeBefore(conn, newEdge)
+}
```
tiffanyhan, line 13: noticed duplicate logic in 4 different places, so i pulled it out into this fn

packages/client/utils/relay/toJiraSearchQueryId.ts
```diff
@@ -1,3 +1,4 @@
+/* to deprecate */
 const toJiraSearchQueryId = (meetingId: string) => `jiraSearchQuery:${meetingId}`
 
 export default toJiraSearchQueryId
```
tiffanyhan, line 4: i believe this should be removed bc from what i've seen, using `jiraSearchQuery:${meetingId}` as a store id doesn't return anything

packages/client/components/NewParabolTaskButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+import Icon from './Icon'
+import {ZIndex} from '~/types/constEnums'
+import FloatingActionButton from './FloatingActionButton'
+
+const Button = styled(FloatingActionButton)({
+  color: '#fff',
+  padding: '10px 12px',
+  width: '150px',
+  top: '85%',
+  left: '174%',
```
tiffanyhan, line 12: the only difference btwn this and `NewJiraIssueButton` is this `left` value and the `<StyledLabel>` text. i think making an absolute position an argument is not so helpful , but this is might be ripe for DRY in the future.

packages/client/components/NewParabolTaskInput.tsx
```diff
@@ -0,0 +1,131 @@
...
+  }
+
+  if (!isEditing) return null
+  return (
+    <>
+      <Item>
+        <Checkbox active />
+        <Task>
+          <Form onSubmit={handleCreateNewTask}>
+            <SearchInput
```
mattkrick, line 110: can you take a crack at using a DraftJS input instead of a plaintext input here? I think it might be weird if folks could add richtext content in a task card, but they couldn't do it here

packages/client/components/ParabolScopingSearchResults.tsx
```diff
@@ -9,18 +10,28 @@ import useLoadMoreOnScrollBottom from '~/hooks/useLoadMoreOnScrollBottom'
...
+  overflow: 'auto'
+})
 interface Props {
   relay: RelayPaginationProp
   viewer: ParabolScopingSearchResults_viewer | null
   meeting: ParabolScopingSearchResults_meeting
 }
 
+const StyledNewIntegrationRecordButton = styled(NewIntegrationRecordButton)({
+  left: '174%',
```
mattkrick, line 26: numbers like this scare me when it comes to being mobile responsive (looks like i missed this on the last go-around for the jira one). 
is there a way you could position this (and the underlying NewIntegrationRecordButton) differently to avoid the percentages & get it on a 4px grid scale?

for example, make the Button a child of a `position: relative` parent that takes up the whole section & then add `bottom: 8, right: 8` or similar

packages/client/mutations/handlers/handleJiraCreateIssue.ts
mattkrick: _this_ is interesting! do you have a step to reproduce? I'd be curious to dig in & see why we can't access it directly

packages/client/mutations/handlers/handleUpsertTasks.ts
```diff
@@ -68,6 +70,12 @@ const handleUpsertTask = (task: Task | null, store: RecordSourceSelectorProxy<an
       }
     }
   }
+  const parabolSearchQuery = meeting?.getLinkedRecord('parabolSearchQuery')
+  const queryString = parabolSearchQuery?.getValue('queryString') as string | undefined
```
mattkrick, line 74: nit: sometimes a more verbose conditional makes for faster reading. for example, if i stick all this stuff in a closure where `parabolSearchQuery` is guaranteed to exist, then the function exits early, typescript coercion isn't necessary & if i'm cracking open this file in a couple months, i know which pieces I can immediately skip over
```ts
const parabolSearchQuery = meeting?.getLinkedRecord<ParabolSearchQuery>('jiraSearchQuery')
if (parabolSearchQuery) {
  const queryString = parabolSearchQuery.getValue('queryString')
  const statusFilters = parabolSearchQuery.getValue('statusFilters')
  const scopingTasksConn = getScopingTasksConn(viewer, [teamId], statusFilters, queryString)
  if (!scopingTasksConn) return
  insertNodeBeforeInConn(scopingTasksConn, task, store, 'ParabolTaskEdge')
}
```

packages/client/utils/relay/toJiraSearchQueryId.ts
mattkrick: ahhhh explains it.
I think somewhere in the shuffle we stopped assigning this object a deterministic Id (see https://github.com/ParabolInc/parabol/blob/d457c22f969a1aea8df77ff3fdb10c722dde6745/packages/client/utils/relay/LocalPokerHandler.ts#L32)

Could you use this function in that LocalPokerHandler to generate a deterministic ID for this object?
doing so should fix that case above where you had to build a workaround in handleJiraCreateIssue.ts

mattkrick (changes requested):

packages/client/utils/relay/toJiraSearchQueryId.ts
tiffanyhan: d'oh. you are totally right, will do!

packages/client/mutations/handlers/handleUpsertTasks.ts
tiffanyhan: oh that's an interesting point. i'm discovering new things about typescript all the time!

packages/client/components/ParabolScopingSearchResults.tsx
tiffanyhan: great point, changed it!

packages/client/components/NewParabolTaskInput.tsx
```diff
@@ -0,0 +1,131 @@
...
+}
+
+const NewParabolTaskInput = (props: Props) => {
+  const {isEditing, setIsEditing, meeting} = props
+  const {id: meetingId, teamId} = meeting
+  const [newTaskContent, setNewTaskContent] = useState('')
+  const atmosphere = useAtmosphere()
+  const {onError, onCompleted} = useMutationProps()
+
+  const updatePokerScope = (res) => {
```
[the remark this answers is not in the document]
tiffanyhan: ah that is interesting. when do you think a field in a mutation payload should be marked using `GraphQLNonNull` vs letting it remain nullable by default?

packages/client/components/ParabolScopingSearchResultItem.tsx
```diff
@@ -1,61 +1,160 @@
...
 const ParabolScopingSearchResultItem = (props: Props) => {
-  const {task, meetingId, isSelected} = props
-  const {id: serviceTaskId, plaintextContent} = task
-  const snippet = plaintextContent.split('\n')[0]
+  const {task, meetingId, isSelected, teamId, setIsEditing} = props
+  const {id: taskId, content, plaintextContent} = task
   const atmosphere = useAtmosphere()
   const {onCompleted, onError, submitMutation, submitting} = useMutationProps()
-  const onClick = () => {
+  const contentState = useMemo(() => convertFromRaw(JSON.parse(content)), [content])
```
mattkrick, line 59: could you use `useEditorState` here to DRY it out?

packages/client/mutations/CreateTaskMutation.ts
```diff
@@ -121,7 +121,8 @@ const CreateTaskMutation: StandardMutation<TCreateTaskMutation, OptionalHandlers
         createdBy: viewerId,
         updatedAt: now,
         tags: [],
-        content: newTask.content || makeEmptyStr()
+        content: newTask.content || makeEmptyStr(),
+        plaintextContent: newTask.plaintextContent || makeEmptyStr()
```
mattkrick, line 125: `makeEmptyStr` isn't plaintext, it's a JSON string. probably best to just use `extractTextFromDraftString` on whatever the content is

mattkrick (changes requested):
fantastic! just 2 little nits in the CR, will get into UI testing after our 1 on 1

PR #4368, opened 2020-11-21
Poker Final Score

mattkrick:
This allows the user to set the final score & push it to an arbitrary field in jira.

TEST
- [ ] clicking the issue field dropdown shows a list of jira fields for the given issue that accept either strings or numbers.
- [ ] selecting a field that only accepts numbers (like story points) will throw an error immediately if you try to type in a letter
- [ ] you can type in a float (e.g. 1.25) as a value & the font size will reduce until it fits, down to size 12px.
- [ ] you can remove the score & submit & have it remove the score in jira


packages/server/graphql/types/PokerMeeting.ts
```diff
@@ -51,28 +50,25 @@ const PokerMeeting = new GraphQLObjectType<any, GQLContext>({
           type: GraphQLNonNull(GraphQLID)
         }
       },
-      resolve: async ({teamId}, {storyId}, {authToken, dataLoader}) => {
-        const userId = getUserId(authToken)
-        const isJiraId = await isValidJiraId(storyId, teamId, {authToken, dataLoader})
-        if (isJiraId) {
-          const [cloudId, issueKey] = getJiraCloudIdAndKey(storyId)
-          const jiraIssue = await resolveJiraIssue(cloudId, issueKey, teamId, userId, dataLoader)
```
mattkrick, line 59: @nickoferrall this was a fun one that i missed in the code review.

`isValidJiraId` is a little dangerous because it determines that by reaching out to jira. as the number of integrations increases, it won't be possible to ask every single service if they have the id we want. instead, we can get the service from the meeting stage itself.

another fun one is the use of `userId` will fail here if the viewer isn't integrated with jira. instead, we want to use the auth token of the person who originally added the issue, they're (almost) guaranteed to have access to the issue that they added. so, we grab it from `stage.creatorUserId`.

nickoferrall: Ah yes, that makes sense. Thanks for letting me know!

Issue #3900, opened 2020-06-08
Migrate slack before Dec 4


mattkrick:
Slack integration needs to be migrated before December 4.
https://api.slack.com/authentication/migration?utm_medium=referral&utm_source=apiblog&utm_campaign=fy20-Q401-granular-permissions

Estimate: 20 hours

mattkrick:
we've kicked this can down the road as long as we could.
we need to do this in the next sprint to ensure we hit the Dec 4 deadline

avivapinchas:
Copy recommendation [here](https://docs.google.com/document/d/13At7-pbn-zBrSWcuFYJ245mSYq9dY4SmyDgSbz_Q3oQ/edit) 🔒

mattkrick:
Next steps:
- submit to slack for approval
- as soon as approved, we will push the updated code to prod & alert everyone via slack
- the update via slack will tell folks that their integration has gone away & they need to re-integrate
- the updated code will wide out everyone's current slack integration (so sending the slack message must come first)

PR #4373, opened 2020-11-24
Slack migration

nickoferrall:
PR to resolve issue #3900 

Once we're happy with the code, the next step is to resubmit the app to Slack. Once approved, we can push the code to production and notify our users to upgrade their app.

I noticed that there's a bug on the `settings/integrations` page where the user needs to click a Team Notifications menu item twice before it updates. I've resolved this by splitting up the `setSlackNotification` logic with a new mutation called `setDefaultSlackChannel`.

### To test

- [ ] Go to the [new Slack app](https://api.slack.com/apps/A01EGHDU7PZ) that I've created called Parabol-staging. Update the SLACK_CLIENT_ID and SLACK_CLIENT_SECRET in your `.env`
- [ ] Authorise the Slack app in your Parabol app
- [ ] Confirm Meeting Start, End, Time Limit Started & Time Limit Ended notifications are sent to your selected channel 
- [ ] Change the default Slack channel by selecting a different channel in Team Notifications dropdown. Check that the notifications are being sent to the new channel
- [ ] Toggle notifications settings, e.g. switch off Meeting Start
- [ ] Remove Slack integration & confirm that notifications are no longer being sent


packages/client/hooks/useSlackChannels.ts
```diff
@@ -13,18 +14,20 @@ const useSlackChannels = (
...
+        userManager.getConversationList(),
         botManager.getConversationList(['im'])
       ])
       if (!isMounted) return
       if (!channelResponse.ok) {
         console.error(channelResponse.error)
         return
       }
-      const {channels: publicChannels} = channelResponse
+      const {channels} = channelResponse
```
mattkrick, line 25: we should change the return type of `getConversationList` so we don't have to recast its response.

packages/server/graphql/mutations/setDefaultSlackChannel.ts
```diff
@@ -0,0 +1,74 @@
...
+    // RESOLUTION
+    if (slackChannelId !== defaultTeamChannelId) {
+      await r
+        .table('SlackAuth')
+        .get(slackAuthId)
+        .update({defaultTeamChannelId: slackChannelId})
+        .run()
+    }
+    const data = {teamId, userId: viewerId}
+    publish(SubscriptionChannel.TEAM, teamId, 'SetDefaultSlackChannelSuccess', data, subOptions)
```
mattkrick, line 69: does this mutation affect the team, or just the team member?
if there is no benefit to the rest of the team receiving this update, we don't need it to be a subscription.
alternatively, we could publish it to the user's NOTIFICATION channel (which would be beneficial if they have multiple tabs open).

packages/client/mutations/SetDefaultSlackChannelMutation.ts
```diff
@@ -0,0 +1,77 @@
...
+      ... on ErrorPayload {
+        error {
+          message
+        }
+      }
+      ...SetDefaultSlackChannelMutation_team @relay(mask: false)
+    }
+  }
+`
+export const setDefaultSlackChannelUpdater = (payload) => {
```
mattkrick, line 33: nit: I like the suffix this with the subscription name + Updater. (e.g. setDefaultChannelSlackChannelTeamUpdater). it's verbose, but useful because on occasion, a mutation could push to 2 different subscription channels & each could have their own updater. (e.g. TeamUpdater, MeetingUpdater).
It's also a healthy sanity check because when you look in the TeamSubscription.ts file, you should be guaranteed to see that all updaters end in `TeamUpdater` 
(that said, i think it's probably best to not have a subscription, or change this to a notification subscription)

packages/client/mutations/SetDefaultSlackChannelMutation.ts
```diff
@@ -0,0 +1,77 @@
...
+) => {
+  return commitMutation<TSetDefaultSlackChannelMutation>(atmosphere, {
+    mutation,
+    variables,
+    updater: (store) => {
+      const {slackChannelId, teamId} = variables
+      const payload = store.getRootField('setDefaultSlackChannel')
+      if (!payload) return
+      payload.setValue(teamId, 'teamId')
+      payload.setValue(slackChannelId, 'slackChannelId')
```
mattkrick, line 57: remember, no setting values in the updater. that breaks subscriptions.

packages/client/subscriptions/TeamSubscription.ts
```diff
@@ -40,6 +40,7 @@ const subscription = graphql`
   subscription TeamSubscription {
     teamSubscription {
       __typename
+      ...SetDefaultSlackChannelMutation_team @relay(mask: false)
```
mattkrick, line 43: remember, if you've published to a subscription channel, you have to include the updater here
(the newMutation CLI script does this for you automatically if you use it 😉 )

packages/client/utils/SlackManager.ts
```diff
@@ -52,14 +69,15 @@ export interface SlackPublicConversation {
...
-interface SlackChannelInfo {
+export interface SlackChannelInfo {
   id: string
   name: string
   is_channel: boolean
   created: number
   is_archived: boolean
   is_general: boolean
   name_normalized: string
+  is_im: boolean
```
mattkrick, line 80: this is a perfect place to test out writing your first discriminated union in typescript.
do a little digging on the subject & try it out.

packages/server/graphql/mutations/helpers/notifySlack.ts
```diff
@@ -127,13 +127,14 @@ const upsertSlackMessage = async (
   if (!channelId) return
   const manager = new SlackServerManager(accessToken)
   const botManager = new SlackServerManager(botAccessToken)
-  const channelInfo = await manager.getChannelInfo(channelId)
-  if (channelInfo.ok) {
-    const {channel} = channelInfo
-    const {latest} = channel
+  const convoInfo = await manager.getConversationInfo(channelId)
```
mattkrick, line 130: nit: if `getConversationInfo` isn't returning the right type, we should fix it

mattkrick (changes requested):
looking good! little changes on the subscription piece & some typescript things. after that it's ready for a test!

packages/client/mutations/SetDefaultSlackChannelMutation.ts
nickoferrall: Ah yes, thanks! I created this PR before your previous comment about this technique. I've moved it to the payload!

packages/client/subscriptions/TeamSubscription.ts
nickoferrall: The newMutation CLI is super helpful - I used it for this mutation but it doesn't look like it adds an updater function by default. I could take a crack at adding it to the script if you like?

packages/client/utils/SlackManager.ts
nickoferrall: I realised that this interface was no longer being used so I've removed it but I got the chance to play with the discriminated unions with `SlackCoversation` and `ConversationListResponse.channels`! I've added type guard checks [here](https://github.com/ParabolInc/parabol/blob/slack-migration/packages/client/hooks/useSlackChannels.ts#L30) and [here](https://github.com/ParabolInc/parabol/blob/slack-migration/packages/server/graphql/mutations/helpers/notifySlack.ts#L129) to prevent recasting

packages/server/graphql/mutations/removeAllSlackAuths.ts
```diff
@@ -0,0 +1,42 @@
...
+        .getAll(r.args(allSlackAuthIds))
+        .update({botAccessToken: null, isActive: false, updatedAt: now})
+        .run(),
+      r
+        .table('SlackNotification')
+        .delete()
+        .run()
+    ])
+    const data = {
+      slackAuthRes: JSON.stringify(slackAuthRes),
```
nickoferrall, line 35: As this function will only be used internally, I thought it'd be best to show the stringified response so we can see how many Slack Auths / notifications have been updated / removed

packages/server/graphql/mutations/messageAllSlackUsers.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    const delay = 100
+    const messagedUserIds = [] as string[]
+    for (const slackAuth of allSlackAuths) {
+      const {botAccessToken, defaultTeamChannelId, userId} = slackAuth
+      const manager = new SlackServerManager(botAccessToken)
+      const postMessageRes = await manager.postMessage(defaultTeamChannelId, message)
+      if (!postMessageRes.ok) {
+        return standardError(new Error(postMessageRes.error), {userId})
```
mattkrick, line 37: we probably don't want to error the whole thing if only 1 fails

packages/server/graphql/mutations/messageAllSlackUsers.ts
```diff
@@ -0,0 +1,52 @@
...
+      }
+      messagedUserIds.push(userId)
+      if (messagedUserIds.length % 100 === 0) {
```
mattkrick, line 40: this won't do what you want it to do. the for loop is synchronous

packages/server/graphql/mutations/removeAllSlackAuths.ts
mattkrick: good thinking! since this will be private, can you move this to the intranetSchema?

packages/client/mutations/MessageAllSlackUsersMutation.ts
```diff
@@ -0,0 +1,34 @@
...
+        }
+      }
+      ... on MessageAllSlackUsersSuccess {
+        messagedUserIds
+      }
+    }
+  }
+`
+
+const MessageAllSlackUsersMutation: StandardMutation<TMessageAllSlackUsersMutation> = (
```
mattkrick, line 21: why the client mutation?

packages/client/mutations/RemoveAllSlackAuthsMutation.ts
```diff
@@ -0,0 +1,35 @@
...
+      }
+      ... on RemoveAllSlackAuthsSuccess {
+        slackAuthRes
+        slackNotificationRes
+      }
+    }
+  }
+`
+
+const RemoveAllSlackAuthsMutation: StandardMutation<TRemoveAllSlackAuthsMutation> = (
```
mattkrick, line 22: for internal mutations, we don't need to write a mutation for the client

packages/server/graphql/mutations/messageAllSlackUsers.ts
```diff
@@ -0,0 +1,52 @@
+import {GraphQLNonNull, GraphQLString} from 'graphql'
+import getRethink from '../../database/rethinkDriver'
+import {requireSU} from '../../utils/authorization'
+import MessageAllSlackUsersPayload from '../types/MessageAllSlackUsersPayload'
+import {GQLContext} from '../graphql'
+import SlackServerManager from '../../utils/SlackServerManager'
+import standardError from '../../utils/standardError'
+
+const messageAllSlackUsers = {
```
mattkrick, line 9: can you move this to the private intranet schema?

packages/server/database/migrations/20201130124200-removeSlackUserAccessToken.ts
```diff
@@ -0,0 +1,23 @@
+import {R} from 'rethinkdb-ts'
```
mattkrick, line 1: can you add another migration for removing all the slack auths?
to determine if something should be a migration vs. mutation, i like to ask what would happen to self-hosted customers. i don't want to have to ask all of them to run a mutation, they should just be able to upgrade to the latest version & have everything work for them. so, in that case, removing the old bad auths should be a migration.

same goes for deploying to dev/staging/prod. when we upgrade, everything should (ideally) work without having to remember to run a mutation

mattkrick (changes requested):
looking good! just a few more

packages/server/graphql/mutations/messageAllSlackUsers.ts
```diff
@@ -0,0 +1,56 @@
+import {GraphQLNonNull, GraphQLString} from 'graphql'
+import getRethink from '../../database/rethinkDriver'
+import {requireSU} from '../../utils/authorization'
+import MessageAllSlackUsersPayload from '../types/MessageAllSlackUsersPayload'
+import {GQLContext} from '../graphql'
+import SlackServerManager from '../../utils/SlackServerManager'
+import standardError from '../../utils/standardError'
+
+interface MessageSlackUserError {
```
nickoferrall, line 9: I don't like repeating the fields here and in `types/MessageSlackUserError.ts`. I looked at importing `IMessageSlackUserError` but it requires a `__typename`. Creating a `GraphQLInputObjectType` works in this file as we're inputting an error object but we'd need to return a `GraphQLObjectType` in the payload as it's no longer an input. 

One option I looked at was creating a `fields` function and creating both a `GraphQLInputObjectType` and a `GraphQLObjectType`, using the same `fields`. This would reduce replication but add its own complexity of having two different object types.

packages/server/graphql/mutations/messageAllSlackUsers.ts
```diff
@@ -0,0 +1,56 @@
...
+    message: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'The slack message that will be sent to all Slack users'
+    }
+  },
+  resolve: async (_source, {message}, {authToken}: GQLContext) => {
+    const r = await getRethink()
+
+    //AUTH
+    requireSU(authToken)
```
nickoferrall, line 27: On the private schema, are all users with access also super users? If so, and if there are no plans to change that, I can remove this

packages/server/graphql/mutations/messageAllSlackUsers.ts
nickoferrall: Doh! It looks like [these options](https://blog.logrocket.com/rate-limiting-node-js/) would work but after digging into the rate limits in the [Slack API](https://api.slack.com/methods/chat.postMessage), I see that they only limit sending several hundred messages per minute per individual workspace. So we should be fine as we're sending a relatively low volume to different workspaces

packages/server/database/migrations/20201130124200-removeSlackUserAccessToken.ts
nickoferrall: Nice, that's a helpful way of thinking about it. Migration added!

packages/server/graphql/rootMutation.ts
```diff
@@ -269,6 +273,8 @@ export default new GraphQLObjectType<any, Context>({
       pokerResetDimension,
       pokerAnnounceDeckHover,
       pokerSetFinalScore,
-      updateJiraDimensionField
+      updateJiraDimensionField,
+      messageAllSlackUsers,
```
mattkrick, line 277: mutations should only live on 1 schema. can you remove these so they are just on the private one?

packages/server/graphql/types/MessageAllSlackUsersPayload.ts
```diff
@@ -0,0 +1,25 @@
+import {GraphQLID, GraphQLList, GraphQLNonNull, GraphQLObjectType} from 'graphql'
```
mattkrick, line 1: can you move these to the intranetSchema types directory?

mattkrick (changes requested):

packages/server/graphql/mutations/messageAllSlackUsers.ts
mattkrick: today that's true, barring any bugs. for things this important, it's a good idea to have every mutation be responsible for its own perms so when we do get hacked, it is harder for them to cause havoc

mattkrick:
hey @ackernaut can you give this an integration test? make sure you have a slack integration set up before changing to this branch to test the migration

ackernaut:
@mattkrick sure thing!

ackernaut (commented):
I tried this locally. I saw the 2 migrations removing tokens and auths. The local team that had Slack no longer has Slack. Is the intention? That they’ll have to go back and add it again? When I tried adding it again I got this window:

Screen Shot 2020-12-04 at 11 40 45

mattkrick:
hmmm, was this with the new client id & client secret for testing?

mattkrick:
@ackernaut were you able to get this up & running?

ackernaut:
@mattkrick let me look for the new secret variables

ackernaut:
With the new app creds, I can add Slack locally and it works!

I saw that the migration removes tokens and auths. Users will have to manually add Slack again? If that’s the case then it seems to be working as expected.

mattkrick:
yeah. super sucks, but i don't see a way around it. we need new perms, only way to get new perms is to have them go through the oauth flow again. i suppose a graceful way to do it would just be to pop up the slack oauth flow as soon as they log in, but aint nobody got time for that

mattkrick:
i think the last check i wanna do on this is send a message to our slack channel saying, "hey, reintegrate please" via the mutation. if that works, then i feel comfy wiping out current integrations

mattkrick:
We want to message all slack users before the migration to delete their credentials runs.
So, we need to remove the migration from this PR & put it in a subsequent one.
Then, we push this PR to prod & call "messageAllSlackusers"
Then, we merge that migration that gets rid of everything.

packages/server/graphql/intranetSchema/mutations/messageAllSlackUsers.ts
```diff
@@ -0,0 +1,56 @@
...
+    // RESOLUTION
+    const allSlackAuths = await r.table('SlackAuth').filter({isActive: true}).run()
+    if (!allSlackAuths || !allSlackAuths.length) {
+      return standardError(new Error('No authorised Slack users'))
+    }
+
+    const messagedUserIds = [] as string[]
+    const errors = [] as MessageSlackUserError[]
+    for (const slackAuth of allSlackAuths) {
+      const {botAccessToken, defaultTeamChannelId, userId} = slackAuth
```
mattkrick, line 38: need to de-dupe on slackChannelId

packages/server/database/migrations/20201130124200-removeSlackUserAccessToken.ts
```diff
@@ -0,0 +1,23 @@
...
+export const up = async function(r: R) {
```
mattkrick, line 2: these migrations need to live in another PR otherwise if we called messageAllSlackUsers, we wouldn't message anyone.

mattkrick (changes requested):

PR #4381, opened 2020-11-30
Poker fixups

mattkrick:
fixing up some last minute stuff:
- [x] add optimistic updater for adding items to scope
- [x] add a progress bar in the left nav instead of the final score for multi-dimension templates
- [x] put a progress bar on reveal votes


ackernaut:
@mattkrick do we want to add a tooltip (I’m open to formating) that reveals what these progress bars represent? See: https://www.figma.com/file/csERjrh67KZKgnT5VahFSb/Sprint-Poker?node-id=3914%3A4485

mattkrick:
i like the tooltip, but i don't think we should include dimension names. The default dimension name is "Estimated Effort".
"Estimated Effort: 5 | Estimated Value: 10" feels really verbose to me, I'd much rather opt for something small & if they don't remember the order they can click into it

PR #4396, opened 2020-12-02
Final score control styles

ackernaut:
- [ ] Looks good for participants for both Jira and Parabol items
  - Participants have read-only styles (no cursor changes, no hover styles, etc.)
  - It looks good on mobile
- [ ] Facilitators have an explicit secondary control to ‘Edit Score’ or ‘Update’ for both Jira and Parabol items
- [ ] Mini poker cards have hover styles when they can be interacted with by the facilitator
  - Hovering the input to add or edit the final score
  - Tapping a card to set the score from a revealed voting row
- [ ] The Jira field dropdown has a cleaner look
  - The label is ‘Jira:’ on mobile
- [ ] The inner gutter is 16px (2x grid) for the control and revealed rows, aligning everything to the right of the mini poker card or input. We have to account for the camouflaged `2px` border of the avatar in some places resulting in values like `14px` or `2px` in some spots.
- [ ] A quick pass at a mobile scope tab panel


ackernaut:
@mattkrick I think this is good enough for v1 in order to timebox the effort, unless you have some suggested changes

ackernaut:
@mattkrick after stepping away from this late day commit I have some second thoughts…I’ll revisit in the AM.

ackernaut:
@mattkrick this is ready for review

mattkrick:
LGTM!

PR #4405, opened 2020-12-04
Poker Edge cases

### LINKED ISSUE #4369: Poker Edge Case Tests
Here's a list of things that will probably break poker that we should fix before shipping: 
- [x] viewing the meeting as someone not integrated with jira
- [x] switching facilitators halfway through & the 2nd facilitator isn't integrated with Jira
- [ ] switching facilitators halfway through & the 2nd facilitator is integrated with jira but doesn't have access to the project
- [x] Jira doesn't respond. We need to put a timeout on it & make sure the response looks exactly like we want & if not, it returns null. If null, the client should handle it gracefully
- [x] updating issue/adding comment fails (or jira hangs). an error should be thrown & the client told that jira is down
- [ ] integrated with 2 jira clouds. the field "story points estimate" has a. different ID for each cloud, but it still works
- [x] for each method in `AtlassianServerManager` go through the poker meeting & make fail via timeout & fail by returning an error. make sure we handle everything gracefully
- [ ] de-authenticate with the jira integration after setting the scoping phase. it fails gracefully
- [x] join a team after a few rounds of voting. what does my vote look like?
- [x] limit to 50 tasks max
- [ ] change the template after the meeting ended. change the dimension name, delete a dimension. delete a few scale values. change their colors. after each change, check out the ended meeting. it shouldn't bork.
- [x] no announce hover when deck is collapsed

mattkrick:
Fixes #4369
This PR makes sure the poker meeting doesn't break when things go wrong.
- [ ] Handles Jira stories when Jira goes down
- [ ] when you hit refresh, the entry animations don't apply (exception is the expanded poker deck)
- [ ] when your avatar is broken, it uses a fallback
- [ ] a poker meeting can only have 50 stories


mattkrick:
i'll get this merged into master before we push, but it won't include fixes for the 2 bugs that terry found

PR #4416, opened 2020-12-08
Poker UI refinements

ackernaut:
- The poker avatar footprint is 40px plus a thicker border (3px) for crisp visual separation
- The poker stage progress background is lighter


packages/client/components/PokerVotingAvatarGroup.tsx
```diff
@@ -14,14 +14,14 @@ const NoVotesHeaderLabel = styled('div')({
   fontSize: 14,
   fontWeight: 600,
   lineHeight: '24px',
-  paddingLeft: 2 // same as avatar border
+  paddingLeft: 3 // same as avatar border
```
mattkrick, line 17: could you make a constEnum like `PokerCards.PokerAvatarBorderWidth`?
then, we can use that instead of comments

packages/client/components/PokerVotingAvatarGroup.tsx
```diff
@@ -14,14 +14,14 @@ const NoVotesHeaderLabel = styled('div')({
...
 })
 
 const Wrapper = styled('div')({
   alignItems: 'center',
   display: 'flex',
   position: 'relative',
-  marginLeft: 14, // plus 2 background border of avatar = 16
+  marginLeft: 13, // plus 3 background border of avatar = 16
```
mattkrick, line 24: `16 - PokerAvatarBorderWidth` would remove the need for a comment

mattkrick:
things i need to test but can't until tomorrow:

does this also handle the overflow avatars? that had a preset HxW of 40.
do the peeking avatars still look good if they only peek 64px of the total 46?
are there any janky pieces in the animations when doing a peek, reveal, refresh, etc?

ackernaut:
@mattkrick I’ll take a closer look at the items you mentioned, and thanks for looking

ackernaut:
- [ ] overflow badge uses avatar size const
- [ ] use avatar border const

packages/client/components/PokerVotingAvatarGroup.tsx
```diff
@@ -8,20 +8,21 @@ import {PALETTE} from '../styles/paletteV2'
...
   fontWeight: 600,
   lineHeight: '24px',
-  paddingLeft: 2 // same as avatar border
+  paddingLeft: PokerCards.AVATAR_BORDER
 })
 
 const Wrapper = styled('div')({
   alignItems: 'center',
   display: 'flex',
+  marginLeft: 16 - PokerCards.AVATAR_BORDER,
```
mattkrick, line 24: i spent way too long verifying that 13px was correct. nice detail!

ackernaut: Once upon a time, waaaay back, I spent a lot more effort tweaking pixels, haha

PR #4420, opened 2020-12-08
Parabol scope story

### LINKED ISSUE #4261: Poker: Estimate: Parabol Story Card
Once the scoping phase supports Parabol tasks, we need a way to display them in the estimate phase.
We've already got it wired up for Jira (https://github.com/ParabolInc/parabol/blob/estimate-vote/packages/client/components/PokerEstimateHeaderCardJira.tsx#L1), we just need a parabol equivalent

AC
- [ ] A story card component exists that displays a Parabol Task in the estimate stage

Estimate: 4 hours

tiffanyhan:
Resolve #4261

AC
- [ ] A story card component exists that displays a Parabol Task in the estimate stage


mattkrick:
CR complete! 
@ackernaut can you give this a test & merge when ready?

ackernaut:
Looking at this now…

ackernaut (changes requested):
Here’s what I saw:

- When the description is long the icon buttons on the right are positioned over the copy and it doesn’t look good. I made a quick commit so that the content fills the space allowed. It looks good on mobile. We’re collapsing long text to control the height anyway, so a few words won’t make much of a difference. Feel free to push back if you don’t like it.

- When I have an integrated card (Jira, GitHub) and add it to the Scope from the Parabol tab after I’ve reached the Estimate phase, the new stage and card appear but the integration link is missing. It only shows if I refresh.

- When I tried integrating a plain Parabol task in the Estimate view 2 things happened: I only saw my Jira integration, not my GitHub integration, in the menu. Since I could only choose the Jira project I tapped it, and the integration link shows a question mark, then goes away as if it failed.

#### Things to fix:
- [ ] When adding an integrated Parabol task from the Scope, it loads correctly in the Estimate phase including the integration link
- [ ] When I press to integrate a task, the menu shows me all of my integrations like on the task kanban (if feasible, I’m not sure how the integration context is different for this meeting)
- [ ] When I select an integration available to me in the push menu from the header card, the integration succeeds and the card now has the integration link on it (or I see some useful error if feasible)

mattkrick:
cool, i'll dig into this & wrap it up

ackernaut:
@mattkrick I think we should allow the Parabol task to be editable. Per the feedback from yesterday, if folks are grooming their stories sometimes they update it with better information. It would be silly to have to go to the kanban to edit the task. Is it as simple as removing the read-only boolean?

mattkrick:
I'm OK with editing, pushing to an integration feels weird though. the CTA is to estimate value. i think editing can help clarify that, but pushing to an integration is something totally different

ackernaut:
That sounds right, we really want folks adding new items from the appropriate tab in the Scope phase right?

ackernaut:
Another small nit I noticed and scope-creepy. When adding new on the Scope tab it seems to scroll correctly for Jira but scrolls slightly out of view on the Parabol tab:



Screen Shot 2020-12-10 at 14 00 11

ackernaut:
>If they could edit it [the Parabol task], it should probably expand when they tap into the description.

PR #4433, opened 2020-12-11
Poker meeting control bar

nickoferrall:
Looks like the meeting control bar is currently off-centre in the Icebreaker & Scope phase. Perhaps a regression after merge conflicts.

This lil fix should resolve that.


mattkrick:
this is a great catch!
the fix for it feels a little hacky though. 
big picture, `isRightDrawerOpen` is a property of the meeting object. we shouldn't be passing that in as its own separate prop because the meeting object already includes that (also, it can get confusing because `isRightDrawerOpen` should ALWAYS mean "the right drawer is open" but here it means "the right drawer is open & the current phase is estimate"). Instead, why not include a react hook that change the values when the estimate phase loads/unloads?

Using both `left` and `right` in css can also be at fault. since the widths of each sidebar are deterministic, could you devise some css that uses just `left`?

nickoferrall:
I like it! I've added a hook to calculate the left.

I've also got the poker deck transitioning smoothly when the right drawer closes. Using the same hook, we can calculate the poker deck `left` and add a transition to get it moving smoothly when the sidebar / drawer opens or closes.

packages/client/components/MeetingControlBar.tsx
```diff
@@ -109,7 +109,12 @@ const MeetingControlBar = (props: Props) => {
   const [confirmingButton, setConfirmingButton] = useClickConfirmation()
   const cancelConfirm = confirmingButton ? () => setConfirmingButton('') : undefined
   const tranChildren = useTransition(buttons)
-  const {onMouseDown, onClickCapture} = useDraggableFixture(isLeftSidebarOpen, isRightDrawerOpen)
+  const showRightDrawer = isRightDrawerOpen && meetingType === MeetingTypeEnum.poker && phaseType === NewMeetingPhaseTypeEnum.ESTIMATE || false
```
mattkrick, line 112: big picture,  if `isRightDrawerOpen` is true, I'd probably expect to see a right drawer somewhere. instead of managing 2 different pieces of state, let's just make this one always be the single source of truth.

in the estimate phase, add a hook that sets `isRightDrawerOpen` to true when the component mounts & sets it to false when the component unmounts. if entry animations are a pain, we can default it to `true` somewhere up higher if needed.

mattkrick:
test complete! only thing i'd change is the 1 comment up above. after that's fixed it's good to merge!

nickoferrall:
Great! I've added a hook called `useRightDrawer` and resolved that issue. Merging now!

PR #4435, opened 2020-12-11
Scale UI polish

mattkrick:
Burndown:
- [x] remove isSpecial
- [x] remove scaleValue.value
- [x] validate the color on server side is one of the pre-defined options 
- [x] refactor "add new dimension" to work the same as "add new scale value". have the button turn into a new dimension input, auto-scroll to the new input, etc. @tianrunhe would you like to take this?
- [x] there's a bug in the scale menu when you have the max # of scales you can't click the edit button
- [x] migrate to remove isSpecial & also change Pass card value to match the constEnum


packages/client/modules/meeting/components/ScaleDropdownMenuItem.tsx
```diff
@@ -5,20 +5,21 @@ import {createFragmentContainer} from 'react-relay'
...
 import {ScaleDropdownMenuItem_dimension} from '../../../__generated__/ScaleDropdownMenuItem_dimension.graphql'
 import {ScaleDropdownMenuItem_scale} from '../../../__generated__/ScaleDropdownMenuItem_scale.graphql'
 import ScaleActions from './ScaleActions'
 
 interface Props {
   scale: ScaleDropdownMenuItem_scale
+  scaleCount: number
   dimension: ScaleDropdownMenuItem_dimension
-  menuProps: MenuProps
+  closePortal: () => void
```
mattkrick, line 22: never give a component more than it needs. i try to make a game of passing as few props as possible & if the prop is an object, as small an object as possible

packages/client/modules/meeting/components/ScaleDropdownMenuItem.tsx
```diff
@@ -61,42 +63,50 @@ const ScaleActionButtonGroup = styled('div')({
...
     submitMutation()
-    UpdatePokerTemplateDimensionScaleMutation(atmosphere, {dimensionId: dimension.id, scaleId}, {onError, onCompleted})
+    UpdatePokerTemplateDimensionScaleMutation(atmosphere, {dimensionId, scaleId}, {onError, onCompleted})
     closePortal()
   }
 
   return (
     <MenuItem
       ref={ref}
-      key={scale.id}
```
mattkrick, line 82: key is only needed if `map` is called in the same component. you already passed a key to this component in the parent, so this is good to go

packages/client/modules/meeting/components/SelectScaleDropdown.tsx
```diff
@@ -43,21 +43,21 @@ const StyledMenu = styled(Menu)({
...
+  const {id: seletedScaleId} = selectedScale
+  const {id: teamId, scales} = team
+  const defaultActiveIdx = useMemo(() => scales.findIndex(({id}) => id === seletedScaleId), [dimension])
 
   const atmosphere = useAtmosphere()
   const {onError, onCompleted, submitting, submitMutation} = useMutationProps()
 
   const addScale = () => {
-    const {menuProps} = props
-    const {closePortal} = menuProps
```
mattkrick, line 55: nit: always try to do all the destructuring at the top, that way you do it once & know that all handlers are referencing the same thing. this will help with bugs as you write more things with `useRef` where the object contents is mutable

packages/client/modules/meeting/components/TemplateScaleValueItem.tsx
```diff
@@ -66,11 +64,12 @@ const ScaleAndDescription = styled('div')({
 
 const TemplateScaleValueItem = (props: Props) => {
   const {dragProvided, isDragging, isOwner, scale, scaleValue} = props
+  const {id: scaleId} = scale
+  const {label, color} = scaleValue
   const [isHover, setIsHover] = useState(false)
   const [isEditingScaleValueLabel] = useState(false)
   const {submitting, submitMutation, onError, onCompleted} = useMutationProps()
   const atmosphere = useAtmosphere()
-  const canRemove = !scaleValue.isSpecial
```
mattkrick, line 73: i think this was an artifact of some old logic. since ? and Pass are always there, we can allow folks to go down to 0 user-defined values

packages/client/modules/meeting/components/TemplateScaleValueList.tsx
```diff
@@ -7,23 +7,25 @@ import {TemplateScaleValueList_scale} from '~/__generated__/TemplateScaleValueLi
...
 import MovePokerTemplateScaleValueMutation from '../../../mutations/MovePokerTemplateScaleValueMutation'
+import AddScaleValueButtonInput from './AddScaleValueButtonInput'
 import TemplateScaleValueItem from './TemplateScaleValueItem'
 
 interface Props {
   scale: TemplateScaleValueList_scale
 }
 
 const ScaleList = styled('div')({
-  margin: 0,
```
mattkrick, line 17: divs will never have margin or padding by default, so unless you're inheriting another styled component you won't need these

packages/client/modules/meeting/components/TemplateScaleValueList.tsx
```diff
@@ -76,16 +78,18 @@ const TemplateScaleValueList = (props: Props) => {
             )
           }}
         </Droppable>
-      </ScaleList>
-    </DragDropContext>
+      </DragDropContext>
+      <AddScaleValueButtonInput scale={scale} />
```
mattkrick, line 82: Putting this button/input in the list means it feels like an item when it's an input

packages/client/modules/meeting/components/PokerTemplateScaleDetails.tsx
```diff
@@ -96,102 +87,67 @@ const ScaleValues = styled('div')({
...
-        </ScaleHeader>
-        <TemplateScaleValueList scale={scale} />
-        <NewTemplateScaleValueLabelInput
-          isOwner={isOwner}
-          isHover={true}
-          isEditing={isEditingScaleValueLabel}
-          setIsEditing={setIsEditingScaleValueLabel}
-          scale={scale}
-        />
-        {!isEditingScaleValueLabel && scale.values.length < Threshold.MAX_POKER_TEMPLATE_SCALES &&
```
mattkrick, line 167: note that `MAX_POKER_TEMPLATE_SCALES` is referring to the number of scales for a team.
what we want is the number of scale values a particular scale can have (added `MAX_POKER_SCALE_VALUES`)

packages/client/modules/meeting/components/ScaleDropdownMenuItem.tsx
```diff
@@ -61,42 +63,50 @@ const ScaleActionButtonGroup = styled('div')({
...
                 [...values.filter(({isSpecial}) => !isSpecial).map(({label}) => label), PokerCards.QUESTION_CARD, PokerCards.PASS_CARD].join(", ")
               }
             </ScaleValues>
           </ScaleNameAndValues>
-          <ScaleActionButtonGroup key={`scale_edit_${scale.id}`} >
+          <ScaleActionButtonGroup>
             <ScaleActions
-              scaleId={scale.id}
-              isStarter={scale.isStarter}
-              scaleCount={0}
```
mattkrick, line 98: these things are always hard to catch on big PRs.
typically, what i like to do is:

```js
// FIXME
const scaleCount = 0
...
<ScaleActions scaleCount={scaleCount}/>
```

by declaring the variable up above & putting a fixme on it it makes it a little easier to spot before committing, but there's no perfect method

mattkrick (commented):
@tianrunhe just a quick overview of the final touches. most of them are pretty minor, but i thought it'd be good to go through some of the sticky stuff. let me know if you have any questions!

packages/client/components/EditableText.tsx
```diff
@@ -87,62 +87,34 @@ interface Props {
...
-    const {initialValue} = nextProps
-    if (prevState.isEditing || initialValue === prevState.value) return null
-    return {value: initialValue}
-  }
-
-  state = {
-    autoFocus: false,
-    isEditing: false,
-    value: this.props.initialValue
+const EditableText = forwardRef((props: Props, ref: any) => {
```
mattkrick, line 90: @tianrunhe @tiffanyhan @nickoferrall 
in the past, I know we've talked about functional vs. class components & how we're almost entirely using functional components now. This is probably the best example we'll have of converting a legacy `Class` component to a functional component-- thought it'd be fun to compare side to side (~50 LOCs shorter with hooks)

packages/client/mutations/AddPokerTemplateDimensionMutation.ts
```diff
@@ -61,15 +61,18 @@ const AddPokerTemplateDimensionMutation: StandardMutation<
...
-        scaleId: SprintPokerDefaults.DEFAULT_SCALE_ID,
+        createdAt: nowISO,
+        name: `*New Dimension #${dimensionCount + 1}`,
         description: '',
+        scaleId: SprintPokerDefaults.DEFAULT_SCALE_ID,
         sortOrder,
-        name: `*New Dimension #${dimensionCount + 1}`,
-        createdAt: nowISO,
         templateId,
-        scale: defaultScale
```
mattkrick, line 71: @tianrunhe this was the source of the "update inside an update" bug.
1. the field `scale` doesn't exist on a dimension, so that was making it short circuit
2. createProxyRecord only accepts an object where the values are not linked records (ie primitives), for linked records, you have to call `setLinkedRecord` like below
3. make sure to include all linked records you can! for example, if a component tries to grab the `dimension.template.name` and `template` isn't provided in the optimistic updater, it's gonna crash

tianrunhe: Oh I see! Thanks for sharing!

PR #4439, opened 2020-12-14
Template editor FAB changes

### LINKED ISSUE #4128: Reflect Template Modifications
Chatted with a possible enterprise client (my girlfriend). @ackernaut eager for your thoughts

Acceptance Criteria:
- [ ] The top list item is a control, [Create New Template](https://www.figma.com/file/wYDUg5qQMjbEllgxBY37sT/Template-Editor-Dialogs?node-id=591%3A40). This control is hidden when there’s a template called ‘[*New Template](https://www.figma.com/file/wYDUg5qQMjbEllgxBY37sT/Template-Editor-Dialogs?node-id=591%3A472)’
  - The "+" FAB is confusing. I'm told people think it means "Select & Start Meeting" based on its location. Let's get rid of it. On the Team tab, we have a "Create New Template" list item. clicking it turns it into a new item. this makes it impossible to create 2, which eliminates a possible error state.
- [ ] The floating button ‘Use Template’ is an explicit control to use the selected template for the meeting. If folks navigate to other templates (e.g. just browsing) but close the dialog instead of tapping the Use Template button, the template selection doesn’t change.
  - **Edited**: a new template can only be selected by using the Select button. Navigating to view other templates, and exiting the modal without selecting, doesn’t change the template for the meeting. ~Consider adding a "Select" button somewhere (maybe to replace the fab?). The only thing it does is close the modal. I'm told it feels weird that you select something & don't click anything after that.~ 
- ~Clone & Edit Template Icon is closer to the Template title (maybe 32px left padding vs. 200px+ away)~

Estimate: 8 hours

tianrunhe:
Resolves #4128 

Tests:

- [ ] The top list item is a control, `Create New Template`. This control is hidden when there’s a template called ‘*New Template’.
- [ ] The floating button ‘Use Template’ is an explicit control to use the selected template for the meeting. 
- [ ] If folks navigate to other templates (e.g. just browsing) but close the dialog instead of tapping the Use Template button, the template selection doesn’t change.
- [ ] Creating a new template & cloning a template should select the active template
- [ ] Removing a template should activate the next available template


tianrunhe:
@ackernaut Ready for UI/UX tests. Thanks!

ackernaut:
### Test cycle no. 1

- [ ] The top list item is a control, `Create New Template`. This control is hidden when there’s a template called ‘*New Template’.
- [ ] The floating button ‘Use Template’ is an explicit control to use the selected template for the meeting. 
- [ ] If folks navigate to other templates (e.g. just browsing) but close the dialog instead of tapping the Use Template button, the template selection doesn’t change.
- [ ] Creating a new template & cloning a template should select the active template
- [ ] Removing a template should activate the next available template

ackernaut (approved):
Hey @tianrunhe this looks good! Thanks for knocking it out so quickly. A few things to consider at your discretion:
- I pushed a change to the new and select controls, feel free to push back. I think the hanging icon of the new template control doesn’t work by itself in a list and the select button was too large when the template had many components
- When I close the dialog while browsing to another template, but without selecting it, and then open it again it’s still on that other template and my instinct was that the selected template should load when I open the dialog again — low tension, no need for change if you don’t want

tianrunhe:
@ackernaut 
> * I pushed a change to the new and select controls, feel free to push back. I think the hanging icon of the new template control doesn’t work by itself in a list and the select button was too large when the template had many components

Looking great!

> * When I close the dialog while browsing to another template, but without selecting it, and then open it again it’s still on that other template and my instinct was that the selected template should load when I open the dialog again — low tension, no need for change if you don’t want

I fixed this. Simple changes.

tianrunhe:
@mattkrick Ready for code review. Thanks!

packages/client/modules/meeting/components/SelectTemplate.tsx
```diff
@@ -0,0 +1,75 @@
...
+  template: SelectTemplate_template
+  teamId: string
+}
+
+const SelectTemplate = (props: Props) => {
+  const {template, closePortal, teamId} = props
+  const {id: templateId} = template
+  const atmosphere = useAtmosphere()
+  const {submitting, error} = useMutationProps()
+  const errorTimerId = useRef<undefined | number>()
```
mattkrick, line 47: when does this get set?

packages/client/utils/relay/setTemplateId.ts
```diff
@@ -0,0 +1,19 @@
+import {commitLocalUpdate} from 'react-relay'
+import {ITeam, MeetingTypeEnum} from '../../types/graphql'
+import Atmosphere from '../../Atmosphere'
+
+const setTemplateId = (atmosphere: Atmosphere, teamId: string, templateId: string | null, meetingType: MeetingTypeEnum) => {
```
mattkrick, line 5: nit: lets rename this to show it's changing the active template.
maybe `activateTemplate` or `setActiveTemplate`?

packages/client/hooks/useSelectTopTemplate.ts
```diff
@@ -1,10 +1,11 @@
 import {useEffect, useRef} from 'react'
+import {MeetingTypeEnum} from '../types/graphql'
 import {commitLocalUpdate} from 'react-relay'
-import SelectTemplate from '../mutations/SelectTemplateMutation'
 import isTempId from '../utils/relay/isTempId'
+import setTemplateId from '../utils/relay/setTemplateId'
 import useAtmosphere from './useAtmosphere'
 
-const useSelectTopTemplate = (edges: readonly {node: {id: string}}[], selectedTemplateId: string, teamId: string, isActive: boolean) => {
+const useSelectTopTemplate = (edges: readonly {node: {id: string}}[], selectedTemplateId: string, teamId: string, isActive: boolean, meetingType: MeetingTypeEnum) => {
```
mattkrick, line 8: lets rename this since we're changing the active template & not the selected one.
maybe `useActivateTopTemplate`?

mattkrick (changes requested):
CR complete! all pretty minor stuff, fantastic work!

packages/client/modules/meeting/components/PokerTemplatePicker.tsx
```diff
@@ -24,10 +27,17 @@ const Dropdown = styled(NewMeetingDropdown)({
...
+  const atmosphere = useAtmosphere()
   const {settings} = props
-  const {selectedTemplate} = settings
+  const {teamId, selectedTemplate} = settings
   const {name: templateName} = selectedTemplate
-  const {togglePortal, modalPortal} = useModal({id: 'templateModal'})
+  const {togglePortal, modalPortal, closePortal} = useModal({
+    id: 'templateModal',
+    onClose: () => {
+      setTemplateId(atmosphere, teamId, selectedTemplate.id, MeetingTypeEnum.poker)
```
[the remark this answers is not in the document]
tianrunhe: This is to address Terry's feedback:

> When I close the dialog while browsing to another template, but without selecting it, and then open it again it’s still on that other template and my instinct was that the selected template should load when I open the dialog again

But I guess setting the activeTemplate to be the selectedTemplate on portal mount fixes this. Will remove this piece.

mattkrick:
Oh gosh this is so darn good!
Just a couple changes to the UI:
- if selected template === active template, the "Select Template" button should be hidden. that way, it is hidden when they open the modal. that means when they click another template, it will appear (bonus points for animating the opacity) & draw their attention to it.
- the "Select Template" FAB should be floating in the bottom right (maybe right: 16, bottom: 16) instead of just being at the bottom of the list. to reproduce, check out the FAB position on a template with only 1 or 2 prompts. it's much higher up
-  IMO, "Create new template" doesn't belong on the org or public tabs, it should only be on the team tabs, but if you feel strongly otherwise I'm OK with leaving it

ackernaut:
>if selected template === active template, the "Select Template" button should be hidden. that way, it is hidden when they open the modal. that means when they click another template, it will appear (bonus points for animating the opacity) & draw their attention to it.

+1, a nice touch

>the "Select Template" FAB should be floating in the bottom right (maybe right: 16, bottom: 16) instead of just being at the bottom of the list. to reproduce, check out the FAB position on a template with only 1 or 2 prompts. it's much higher up

I do like the floating aspect of it. Perhaps we need to ensure some minimal bottom gutter so controls (e.g. sharing settings) can be scrolled out from under fully?

>IMO, "Create new template" doesn't belong on the org or public tabs, it should only be on the team tabs, but if you feel strongly otherwise I'm OK with leaving it

I agree with the pure logic. However, TMK folks start off on the Public tab and it may not be apparent where to create a new template. Maybe if they want it bad enough they’ll find it? I wonder if folks would start cloning templates just to get their own template and have to totally overhaul the cloned template to get to their end goal (assuming they don’t look on the Team tab, which honestly, might not be apparent)?

tianrunhe:
I updated the button appear/disappear piece and left the rest untouched. Let me know if that looks good! @mattkrick

packages/client/modules/meeting/components/SelectTemplate.tsx
```diff
@@ -19,7 +25,9 @@ const ButtonBlock = styled('div')({
   pointerEvents: 'none',
   right: 0,
   bottom: 0,
-  width: '100%'
+  width: '100%',
+  animation: `${fadein} 2s`,
+  WebkitAnimation: `${fadein} 2s`
```
mattkrick, line 30: emotion does vendor prefixes for us automatically! no webkit needed

mattkrick:
This is gonna make a lot of first time users VERY happy. fantastic work!

PR #4440, opened 2020-12-14
Saml mutation improvements

### LINKED ISSUE #4383: improvements to `enableSAMLForDomain`
## Issue - Enhancement

For various reasons, setting up SAML still requires manual developer intervention. We should robustify the mutation so that non-technical folks can reliably set up sso for clients.

`enableSAMLForDomain` mutation should:
- [ ] before setting `id` and `domain` fields, top level domain should be stripped out of input value if present
- [ ] `url` field should not normalize case, or should only normalize parts of url which are case insensitive i.e. query parameters should preserve case
- [ ] if the `url` value contains `"microsoftonline"`, we should add the deflate encoded `?SAMLRequest=` query param value to the `url` field.
- [ ] add `createdAt`, `updatedAt` and `deletedAt` ts fields
- [ ] check if valid xml for `metadata` field

- **Estimated effort:** 8 hours ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

tiffanyhan:
Resolves #4383

**AC**

Do `enableSAMLForDomain` mutation in admin graphiql:

- [ ] if the value for `domain` field contains any `.`, an error msg will be returned. only values without any `.` (e.g. philips) should be accepted. if any uppercase values are present, they will be normalized to lowcase.
- [ ] if the value for `metadata` field isn't valid xml, an error msg will be returned. only valid xml will be accepted.
- [ ] if the value for `url` field isn't a valid url, an error msg will be returned. only valid urls will be accepted.
- [ ] if the value for `url` field contains `"microsoftonline"`, the mutation will automatically add a URI encoded `?SAMLRequest=<:long string with both low and upcase chars>` query param to the saved `url`.
- [ ] `updatedAt` field will always be set, and `createdAt` field will be set if creating


mattkrick:
oh this is so great. thank you so much for digging into this!
for the first test, we'll need to make sure we support subdomains, per https://github.com/ParabolInc/parabol/issues/4442.
I think the way forward is a SAML record like: `{id: 'parabol', domains: ['parabol.co', 'external.parabol.co', 'parabolpals.com']}`, but i haven't thought it through yet.
Would you be open for integrating that issue into this PR as well?

tiffanyhan:
@mattkrick you want to include TLD in `domains` field right? i think this is the correct thing to do. it would just require a bit more manual work for the migration... thinking to search for users with enterprise accounts and look at their full domain including TLD, then create a hash map. I think it's okay since we only have a couple rows in the table.

mattkrick:
yep! that's definitely the right thing to do... i goofed up big time when i built it without including the tlds, so this is a fortunate way for us to fix that!

for the migration, we could probably just suffix with `.com` and then if there are any non-coms we can update that manually directly in the DB

tiffanyhan:
cool just checked and everything is `.com` except for one `.pl`. added to the mutation!

tiffanyhan:
@mattkrick what do you think of having a separate table `SAMLDomain` consisting of `domain`, `verifiedAt`, `verifyToken`, and `samlId` fields. if `verifyToken`is nonnull, a background or cron job picks it up and uses `dns.resolveTxt(hostname, callback)` to check for the txt record. there can also be a sync way to trigger this check. once verified, we set `verifiedAt` ts and delete `verifyToken` value.

i think it's possible to keep `SAMLDomain` as a nested field, but i think it's better to use a separate table bc of migration to postgres plans.

packages/server/database/migrations/20201217140026-addDomainsToSAML.ts
```diff
@@ -3,12 +3,14 @@ import {R} from 'rethinkdb-ts'
...
-    .map((row) => ({
-      id: row('id'),
-      domains: [row('domain')],
-      url: row('url'),
-      metadata: row('metadata')
-    }))
+    .map((row) => {
+      return {
+        id: row('id'),
+        domains: [row('id').add(r.branch(row('id').eq('allegro'), '.pl', '.com'))],
```
mattkrick, line 9: careful! don't want client information to make it to our git commits

mattkrick:
we've got to ship this out to a client this week so i'm gonna go ahead & take over on this one

mattkrick:
had to ship this out so i finished it up in #4478

Issue #3614, opened 2020-03-03
Make grouping columns responsive


jordanh:
## Issue - Enhancement

A user 🔒 writes:

> We used “Working and Stuck” which only has two lists. It would be cool to have these columns be responsive so that you could widen the list to 2 columns.

Let's collect more feedback before defining acceptance criteria and estimated effort.

mattkrick:
good idea, would like to see it pop up again.

Issue #4438, opened 2020-12-14
Adding reflection in Group stage triggers infinite loop


nickoferrall:
### Test

- [ ] Repeatedly clicking + "Add reflection" icon in Group stage adds a reflection without freezing the page

PR #4444, opened 2020-12-15
Responsive group columns

nickoferrall:
This PR resolves issue #3614 and resolves a bug captured in issue #4438.

### Considerations

- There's a bug in prod where repeatedly clicking the "Add reflection" icon triggers an infinite loop. This PR includes a fix for this and it also disables the icon onClick rather than removing it, which is a personal preference
- When the column is expanded, it's possible to drag & drop the cards onto one side, leaving the other sub-column empty. We could auto-minimize the column here but I think the user should decide as I implemented auto-minimize and found it a bit irritating
- Columns are maximised by default if there's space. However, if a user closes the sidebar meaning there is now space to maximise the columns, the columns do not maximise automatically. A user most likely closes the sidebar when they want more real estate and may not want the columns to expand

### AC

- [ ]  each column has a "maximize" icon (I opted to use the same icon for maximise and minimize)
- [ ]  when clicked, the column grows to the width of 2 cards
- [ ]  a column is maximized by default if space allows (e.g. if there are only 2 columns on a desktop width)
- [ ]  icon controls (add reflection, expand/collapse column) have a tooltip since they don’t have a text label


nickoferrall:
Hey @ackernaut, it'd be great to get your eyes on this one please. I've added a few considerations in the description.

Once you're happy with the UI/UX, I'll request a code review from Matt

ackernaut:
### Test cycle no. 1 (in progress…)
- [ ]  each column has a "maximize" icon (I opted to use the same icon for maximise and minimize)
- [ ]  when clicked, the column grows to the width of 2 cards
- [ ]  a column is maximized by default if space allows (e.g. if there are only 2 columns on a desktop width)
- [ ]  icon controls (add reflection, expand/collapse column) have a tooltip since they don’t have a text label

ackernaut:
### Later than usual review

> each column has a "maximize" icon

I created [a PR with some suggested style changes](https://github.com/ParabolInc/parabol/pull/4449). The custom SVG icon didn’t match MD closely enough IMO. I provided something to try in the PR. I opted for 2 icon states. I noticed the use of ‘maximise’ 🇬🇧   and I like it since we are a global team. I quickly checked to see if we use ‘-ize’ or ‘-ise’ elsewhere in UI copy. The closest I got was ‘Tap to customize the Icebreaker’. I thought maybe we try to be consistent with ‘z’ or ‘s’ but then again I like the diverse touch here — we’re built by many developers.

>a column is maximized by default if space allows

I didn’t see this behavior. Maybe I missed something.

>icon controls (add reflection, expand/collapse column) have a tooltip since they don’t have a text label

You’re my hero. I’ve wanted to add that tooltip for a minute.

>There's a bug in prod where repeatedly clicking the "Add reflection" icon triggers an infinite loop. This PR includes a fix for this and it also disables the icon onClick rather than removing it, which is a personal preference

🙏   nice catch! gracias, compañero

>When the column is expanded, it's possible to drag & drop the cards onto one side, leaving the other sub-column empty. We could auto-minimize the column here but I think the user should decide as I implemented auto-minimize and found it a bit irritating

I think this is okay. I noticed that when I collapsed and expanded again it divided them across the 2 sides. I think in order to divide them up on the fly there would be some magic, some complex magic that we had once upon a time in a previous version involving a group grid

>Columns are maximised by default if there's space. However, if a user closes the sidebar meaning there is now space to maximise the columns, the columns do not maximise automatically. A user most likely closes the sidebar when they want more real estate and may not want the columns to expand

I still think I missed this behavior. I think that’s a good call on the automatic behavior. We don’t want UI components changing shape or moving around too much on their own.

ackernaut (changes requested):
See comment

nickoferrall:
Nice, thanks for the review and for making those changes! 

I like the tighter padding and I originally went with the twisted MD icons before adding the SVG. I'll stick with MD where possible from now on.

> I noticed the use of ‘maximise’ 🇬🇧  and I like it since we are a global team

Ha, good catch! This wasn't intentional and I'd be happy to change it to keep things consistent. It starts with "maximise" but next thing you know "colour", tea & cricket have invaded the codebase..

> I didn’t see this behavior. Maybe I missed something.

The columns are only expanded by default if there are only two prompts and you're using a larger monitor or there is only one prompt and you're using a laptop. If you create a template with just 1 prompt, is it initially expanded when you land on the Group stage?

ackernaut:
Hey @nickoferrall I tried a template with 2 prompts and yes the columns were expanded. Nice work. I noticed they're expanded even with a few comments. I think this is good as it shows the increased adaptability of the UI.

ackernaut (approved):

ackernaut:
And…I noticed it **_works well_** in the demo. I forget to check there sometimes.

nickoferrall:
Excellent! I'll move onto the code review then and tag Mr Krick

packages/client/hooks/useSortSubColumns.ts
```diff
@@ -0,0 +1,48 @@
...
+  reflectionGroups: GroupingKanbanColumn_reflectionGroups
+) => {
+  const atmosphere = useAtmosphere()
+  useLayoutEffect(() => {
+    if (!reflectionGroups) return
+    commitLocalUpdate(atmosphere, (store) => {
+      reflectionGroups.forEach((group, index) => {
+        const reflectionGroup = store.get(group.id)
+        if (!reflectionGroup) return
+        const subColumn = index % 2 === 0 ? SubColumn.LEFT : SubColumn.RIGHT
```
mattkrick, line 18: what if the retro is a single column? can we allow them to expand the column to be 3+ columns?

packages/client/hooks/useSortSubColumns.ts
```diff
@@ -0,0 +1,48 @@
...
+          reflectionGroup.setValue(subColumn, 'subColumn')
+        } else {
+          currentSubColumn === SubColumn.LEFT
+            ? (leftSubColumnCount += 1)
+            : (rightSubColumnCount += 1)
+        }
+      })
+    })
+  }
+  useEffect(() => updateReflectionGroups(), [reflectionGroups])
```
mattkrick, line 45: this gets a little spicy because the dependency is `reflectionGroups` and the function actually changes `reflectionGroups` which means we're guaranteed to trigger at least 2 re-renders.

big picture, the question is, "is the subColumn state or a calculation?" If it's a calculation, then we can accomplish this without requiring the relay store & doing any re-rendering:

- determine a `columnCount` like how many columns we can support (1, 2, 3, etc.)
- for each column, do something like `reflectionGroups.map((group, index) =>  if (index %  columnCount !== 0 return null...`

mattkrick (changes requested):
looking good!
only big thing i'd change is swap out the `Left Right` with a numeric index so we can support more than 2 columns & then we can do all the magic without telling the relay story about which column the group is in

packages/client/hooks/useSortSubColumns.ts
nickoferrall: Initially, I used a calculation instead of state but this caused reflection groups that the user isn't updating to move around. For example, in the recording below I move group 3 into group 2. This removes group 3 which updates the indexes of subsequent groups meaning `idx % columnCount` logic causes group 5 to move. 

So I think we have to use state to keep track of the original position of the group. To support more than two columns, I could keep track of the subColumn index rather than `LEFT` `RIGHT`?


https://user-images.githubusercontent.com/39854876/103781633-fda1c880-500c-11eb-918c-8936d7de67ba.mp4

mattkrick: valid point! 
what if we moved the handling outside of the `useEffect` and put it next to where `setIsWidthExpanded` is called?

So big picture:
- you click the expand button
- a calculation is performed to see how many extra columns we can add (minimum 2 cols, because we want expand to do _something_)
- a `subColumnCount` state is set with an integer (1,2,3, etc) 
- a `subColumn` integer (0, 1, 2) value is assigned to each child in that column

doing it on click should allow us to set state outside of the `useEffect` hook

mattkrick:
modified AC
- [ ] by default, columns are expanded equally (e.g. if there is space for 5, and only 2 prompts, then each prompt gets a width of 2 because 2 * 2 = 4 which is <= 5.
- [ ] if a column has a width of > 1, then the toggle should be "collapse" so, if by default each column has a width of 2, then the toggle will collapse. this means that if the default width is 2 and you want to expand to 3, you have to click twice :-(.
- [ ] when expand is clicked, it is expanded to a minimum of 2 columns a a maximum of n columns where n gets rid of the vertical scrollbar and n + m other columns < meeting area width (if the meeting area can support 5 columns and there are 3 prompts, then expanding the first column would yield at most 3 sub columns because 3 + 1 + 1 = 5)

packages/client/hooks/useSortSubColumns.ts
nickoferrall: I've pushed an update where it calculates the subColumn index when clicking the expand button so we can now have 3+ subColumns 🎉 
There is a still a `useEffect` listening to `reflectionGroup` changes. I think we still need this as it sorts a new reflection group that doesn't have a subColumn index. This happens when a user adds a new reflection or removes a reflection from a group.

packages/client/components/GroupingKanbanColumn.tsx
```diff
@@ -116,51 +131,56 @@ const GroupingKanbanColumn = (props: Props) => {
     submitMutation()
     CreateReflectionMutation(atmosphere, {input}, {onError, onCompleted})
   }
-  const ref = useRef<HTMLDivElement>(null)
-  const canAdd = phaseType === NewMeetingPhaseTypeEnum.group && !isComplete && !isAnyEditing
-  const isExpanded =
-    useCoverable(promptId, ref, MeetingControlBarEnum.HEIGHT, phaseRef) || !!endedAt
+
+  const toggle = (e: MouseEvent<Element>) => {
+    e.stopPropagation()
```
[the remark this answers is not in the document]
nickoferrall: Ah this was initially used in the `Reflect` stage and it prevented highlighting the column. I've removed this now as the responsive columns only exist in Group & Vote stages

packages/client/hooks/useCallbackRef.ts
```diff
@@ -0,0 +1,17 @@
+import {MutableRefObject, useCallback, useRef} from 'react'
+import useForceUpdate from './useForceUpdate'
+
+const useCallbackRef = (): [(node) => void, MutableRefObject<HTMLDivElement | null>] => {
```
nickoferrall, line 4: We need to use the `phaseRef` width to calculate the initial number of subColumns when the page renders. It looks like a callback ref is the best way of doing this but I don't love the solution as reassigning a ref and using `forceUpdate` feels a lil hacky

mattkrick:
After some UI testing found 2 little bugs:
- The X scrollbar appears when it shouldn't (usually this is fixed by making the x scrollbar hidden, sometimes that takes a nested div). See the first pic
- vertical scrolling is generally better than horizontal scrolling (more natural to scroll down on mobile, on desktop, a scrollwheel is pretty common, but not all scrollwheels can scroll horizontally). with that in mind, i'd hope that clicking "expand" would guarantee that the total width would be less than the width of the meeting area (see video)
Screenshot from 2021-01-13 09-13-18

https://user-images.githubusercontent.com/5514175/104486063-174c8e00-5580-11eb-954a-e79b5fe9efee.mp4

packages/client/components/ReflectionCard/ReflectionCardRoot.tsx
```diff
@@ -11,8 +10,8 @@ const ReflectionCardRoot = styled('div')({
   display: 'block',
   maxWidth: '100%',
   position: 'relative',
-  transition: `box-shadow 2000ms ${DECELERATE}`,
-  width: ElementWidth.REFLECTION_CARD
+  transition: `all 100ms ${BezierCurve.DECELERATE}`,
```
mattkrick, line 13: box-shadow should still animate in at 2s. a transition can support different speeds & curves for different props

mattkrick:
- on `!isDesktop` (where the width is < 2 reflections) there should be no expand button

optimization goal: maximize screen real estate & prefer vertical scrolling.

to achieve that goal:
- on desktop, there is always an "expand button", no matter how many columns, or the `isExpanded` state of other columns
- clicking "expand" should give a width of at least 2
- if there currently isn't a horizontal scrollbar & expand is clicked, it should be expanded to as many columns as possible while total width <= meeting area width
- if expanding to 2 would cause a scrollbar, go ahead & expand to `MAX`
- if a scrollbar is already present, then clicking expand should expand to `MAX`
- `MAX` should never be greater than an entire meeting areas width (e.g. if a meeting area is 4 columns wide, then MAX Math.min(4, `columnsRequiredToRemoveVerticalScrollbar`). In other words, a meeting area must should at least 1 entire prompt

packages/client/hooks/useSubColumns.ts
```diff
@@ -0,0 +1,86 @@
...
+    }
+  }, [phaseRef?.current])
+
+  useResizeObserver(() => {
+    if (!isDesktop && subColumnCount > DEFAULT_SUB_COLUMNS) {
+      setSubColumnCount(DEFAULT_SUB_COLUMNS)
+    }
+  }, phaseRef)
+
+  return [subColumnCount > DEFAULT_SUB_COLUMNS, subColumnCount, subColumnIndexes, toggleWidth]
```
mattkrick, line 83: adding an `as const` suffix here means you don't have to explicitly declare the return type at the top. that's nice because if the return type signature changes, it will change automatically for you

PR #4445, opened 2020-12-15
Pre-validate jira field mappers

### LINKED ISSUE #4423: Handle Jira De-authentication mid-meeting
This shouldn't block a beta release, but we should intercept those errors & make them friendlier.

Errors & repro here: https://github.com/ParabolInc/parabol/issues/4369#issuecomment-740758151

AC
- [ ] clear steps forward if a jira de-auth happens mid meeting

Estimate: 8 hours

mattkrick:
TEST
- [ ] if the jira project supports Story Points, then that field is set as the first dimension default
- [ ] if the jira project supports Story point estimate, then that one is used
- [ ] else, the comment field is used

Also fixes #4423


ackernaut:
🔥

Issue #4370, opened 2020-11-21
Improving build time & stability


mattkrick:
Doing a bunch of front-end coding lately, I'm really frustrated with the DX. HMR is spotty on both the server & client, relay-compiler will seemingly restart for no reason, etc.

Investing time here in DX will yield huge savings as every developer will be able to code faster, so we should prioritize this.

Thoughts:
- My conclusion is HMR on the server is unreliable. We should get rid of it (for the server). If we are thoughtful, a server restart should be almost as fast as HMR.
- having a co-dependency between server & client packages is wrong. The client should have no dependencies on parabol-server & all shared assets should live in the client.
- We could probably shortcut the GQL Executor in development & just go direct to executeGraphQL
- There should be a way to cache the server bundle & the client bundle independently, although I'm not sure how to do this without some HMR on the server
- We should get those `.graphql` files living on the client compiled & hardcoded so we don't need that babel transformer. in the future, we'll just attach the GH schema to ours & all requests will go through parabol-server
- although it looks like sucrase is faster than swc (https://github.com/alangpierce/sucrase/blob/d8928b27be3d83729bff48fdcb00d9cde5eed3fc/README.md#try-it-out) there are some other tools that we should explore:
  - https://github.com/evanw/esbuild/blob/master/docs/architecture.md
  - https://www.snowpack.dev/ (looks like webpack 5 would be a better call than this)
- we should see if there's a new react HMR tool to improve client HMR

@tianrunhe @tiffanyhan @nickoferrall any input here is welcome! let's get a list of what we want the perfect DX to look like & start building

mattkrick:
i wanna bite this off in chunks:
- upgrade to webpack 5 to reduce build times
- remove all dependencies on parabol-server from parabol-client
- remove HMR on the server & instead use nodemon or similar that watches for changes on the server folder
- work on caching server & client webpack builds

ackernaut:
Can we address the need to restart when changing const enums in some burndown?

mattkrick:
yep! that's a big one.

PR #4455, opened 2020-12-17
Upgrade to Webpack 5

mattkrick:
webpack is at v5.10, so it's probably stable enough to upgrade.
This should be worthwhile since it promises faster build times, better caching & fixed long-term caches. That last one means users won't have to download the whole app on every upgrade, just the changed modules.

This is just the first piece of #4370 but i want to keep this isolated because there are going to be a lot of moving pieces

things to consider:
- does server side HMR work now? it got rebuilt...
- webpack can natively support service workers. should we switch to that? 
- can i move back to node 14.10? there was a bug in it...

What this does:
- bumps a BUNCH of packages. everything from babel to typescript to webpack plugins to emoji-mart (now with new emojis!)
- replaces webpack-hot-client with webpack-hot-middleware. the former wasn't reliably invalidating webpack builds. now at least when you change client code the browser refreshes. it's a start.
- added react-refresh, but commented out because it DOUBLES build time (https://github.com/pmmmwh/react-refresh-webpack-plugin/issues/255). that's not worth it unless we can guarantee that we won't have to restart the HMR server. 
- removes storybook stuff. we weren't using it & it had a bunch of old dependencies


mattkrick:
...aaaand after all that:
Webpack v4 Build time: 11.5s
Webpack v5 Build time: 11.8s

Screen_Shot_2018-12-31_at_1 19 26_PM

ackernaut:
🤦

PR #4458, opened 2020-12-17
Fix a bug where user cannot clone a public scale

### LINKED ISSUE #4457: User cannot clone a Parabol pre-defined scale
## Issue - Bug

When clicking the copy button by a Parabol pre-defined scale (Fibonacci, T-Shirts or Five Fingers), nothing happened. From server side, error:

```
SEND TO SENTRY Error: Team not found
    at Object.resolve [as MutationaddPokerTemplateScaleResolver] (webpack-internal:///./packages/server/graphql/mutations/addPokerTemplateScale.ts:41:83)
    at processTicksAndRejections (internal/process/task_queues.js:93:5) undefined
```

- [ ] Users can clone a public scale

tianrunhe:
Resolves #4457 

The bug is that when we clone a scale, we should not use the `teamId` for that scale. For pre-defined the scale, the `teamId` is _aGhostTeam_ so not matching with the viewer from server side, hence the request was rejected. The fix is to use the `teamId` for the dimension.

Tests:
- [ ] User can clone a pre-defined scale
- [ ] User can still create a new scale

### DIFF
--- a/packages/client/modules/meeting/components/ScaleActions.tsx
+++ b/packages/client/modules/meeting/components/ScaleActions.tsx
@@ -19,14 +19,16 @@ const CloneAndDelete = styled('div')({
 interface Props {
   scale: ScaleActions_scale
   scaleCount: number
+  teamId: string
 }
 
 const ScaleActions = (props: Props) => {
   const {
     scale,
     scaleCount,
+    teamId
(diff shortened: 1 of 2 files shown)


ackernaut:
LGTM

Issue #4345, opened 2020-11-16
Links in Discussion Thread comments should be links 


avivapinchas:
##  Issue - Bug

When posting a link in a Discussion Thread comment, the link presents as text. It should be a link that I can click. 

Example: 
![Screen Shot 2020-11-16 at 1 47 33 PM](https://user-images.githubusercontent.com/11621016/99300497-4c539300-2812-11eb-8bfd-02d022a9208f.png)

- **Estimated effort:** X points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

### Acceptance Criteria (optional)
Users can:
   - Paste links into a comment in a Discussion Thread and have them link to whatever the link is pointing to

PR #4483, opened 2020-12-23
Discussion thread links

nickoferrall:
This PR resolves issue #4345.

### Consideration

Copying and pasting between Draft JS editors doesn't work very well - font styles and hyperlinks aren't always maintained. This is a [known issue](https://github.com/facebook/draft-js/issues/787) in Draft JS. Someone from the community has created a PR with a suggested solution that hasn't been approved but I'd be happy to look into it. 

However, I can't imagine that many of our users are copying & pasting text that they've just pasted so I suggest that I create an issue for this and prioritise it if any user tension pops up. 

### AC

- [ ] Users can paste links into the comment editor
- [ ] Users can paste links into the task editor
- [ ] Users can paste a combination of text and hyperlinks and the hyperlinks will still work


packages/client/components/TaskEditor/CommentEditor.tsx
```diff
@@ -164,6 +166,15 @@ const CommentEditor = (props: Props) => {
         }
       }
     }
+    const links = linkify.match(text)
```
nickoferrall, line 169: Initially I included this code in `useLinks` to keep things DRY but as it's only being repeated twice and as we need the above `onPasteText` code with the `ref` in this file, I felt it'd be best to keep it in the same function

mattkrick:
this isn't working for me:
1. in a thread, add a comment & type `http://www.google.com`
2. hit space, see it turn into a hyperlink, copy the text
3. submit
4. hit ctrl+v to add a second, duplicate comment
5. clicking the first comment works as a hyperlink, but the 2nd does not

nickoferrall:
Interesting - I couldn't replicate that behaviour in the comment editor but if I followed steps 1-3 and then did ctrl+v into the task editor, the link was underlined but didn't work. 

I'm now trimming the whitespace which has resolved the task editor issue and I think it should resolve the comment editor issue too.

mattkrick:
success!

PR #4487, opened 2020-12-28
guarantee unique ID for redis jobs

mattkrick:
fixes all those sentry errors by guaranteeing a unique id in redis

requires all servers have a SERVER_ID env var.
this var should be an int between 0 and 1023.


mattkrick:
all servers updated with new env var

PR #4488, opened 2020-12-28
replace shortid with snowflake

mattkrick:
We've been using shortid to generate unique IDs. The authors deemed this architecture as unsafe to use in a distributed environment, so we need a new solution.
Requirements are:
- unique IDs across the entire app (not global)
- do not need to be random
- need to be nice slugs since we use these in URLs

This is more or less a solved problem. twitter snowflake came out 11 years ago that creates IDs based on the timestamp, machine ID, and a sequence. this is my JS implementation of it.

40 bits of a timestamp (using Jan 1, 2021 as the epoch) 

\+ 10 bits for the machine ID (all servers must have an ID of 0 - 1023) 

\+ 12 bits for a sequential integer

---------
That's 62 bits, but Javascript only supports 53-bit float64s, so we convert it to a base62 string (a-zA-Z0-9). 

note: must add SERVER_ID to the .env for all servers!


packages/server/generateUID.ts
```diff
@@ -0,0 +1,45 @@
...
+const MACHINE_ID_BIT_LEN = 10
+const SEQ_BIT_LEN = 12
+const TS_OFFSET = BigInt(MACHINE_ID_BIT_LEN + SEQ_BIT_LEN)
+const MID_OFFSET = BigInt(SEQ_BIT_LEN)
+const BIG_ZERO = BigInt(0)
+const MAX_SEQ = 2 ** SEQ_BIT_LEN - 1
+
+let seq = 0
+let lastTime = Date.now()
+const generateUID = () => {
```
mattkrick, line 15: hey @tianrunhe can you review this little chunk of code?

mattkrick:
looks like this tops off around 500 IDs/ms. 
I think that's plenty, but I'll still leave the check to make sure we don't exceed 4096.

PR #4497, opened 2021-01-05
Remove Slack auths migrations

nickoferrall:
This PR implements the migrations that remove Slack auths. Once we've messaged all Slack users using #4373, we can use these migrations to remove their old auths.


mattkrick:
@nickoferrall when checking this out, i noticed the channel list looks pretty small. this is probably because we're using a bot token instead of a user token. since a bot isn't a member of many (or any) channels, the majority of the list gets filtered out. see https://github.com/ParabolInc/parabol/blob/26a7d763a9fbd02ac828b46ce7f5758154c79ca2/packages/client/hooks/useSlackChannels.ts#L29.

could you fix this so the list shows all the channels the user could install the app with?

nickoferrall:
Huh, that's strange. I've pushed an update so that the channel list includes all public channels and the Parabol bot. The bot only becomes a member of a public channel when the user selects it.

I noticed that the channels listed in the Slack oath page include some public channels and some DMs. I think it'd be safer to just show public channels and the Parabol bot. 

The oath logic [is here](https://github.com/ParabolInc/parabol/blob/c5a7084e335af5e84a2a5ba4d718b0eb3116751b/packages/client/utils/SlackClientManager.ts) but I'm not sure how the Slack oath app decides which channels to display - I suspect it uses the scopes? We could remove the private message scope option and just show public channels or the Parabol bot?

slack-channels

mattkrick:
the incoming:webhook scope let's you pick ANY channel, there's no way around that.
because the user _could_ pick any channel (even an IM) we wanted to support public & IMs because it would suck to be able to initially do something & then not do it again. (also, I imagine sometimes PMs might just want a notification for themselves, but not a whole public channel)

nickoferrall:
> the incoming:webhook scope let's you pick ANY channel, there's no way around that.

ahh that's annoying! I'll push an update adding DMs to our dropdown menu options

nickoferrall:
Currently, when the user selects a DM on the oath page, e.g. @nick, that person then gets messaged privately from the Parabol bot, not from @nick. 

Is that the behaviour you're expecting? Or when the user clicks @nick in the Slack oath page, are you expecting the bot sends a direct message to @nick and not to @ Parabol?

The update I pushed yesterday does the former, not the later. I think that makes sense as my other Slack apps message me from the bot, not from @nick. 

All public channels and the Parabol bot are now visible in the dropdown menu.

nickoferrall:
Jordan & Grayson mentioned that customers would like to use private channels too so I've added them to the dropdown. 

Strangely, Slack's API returns 61 channels when requesting public channels, i.e. `botManager.getConversationList()` or `botManager.getConversationList(['public_channels'])`, but 51 channels when requesting public & private channels, i.e. `botManager.getConversationList(['public_channel', 'private_channel'])`. 

I suspect this is a bug on Slack's end. In the meantime, I've made two requests so that we get all of the channels and have merged the responses.

PR #4500, opened 2021-01-06
Fix json2csv Parser import after upgrading

### LINKED ISSUE #4489: CSV export malformed
## Issue - Bug

[A user](https://app.hubspot.com/contacts/3888472/contact/16890601/) 🔐  (see 🎟️  [#272804875](https://app.hubspot.com/contacts/3888472/ticket/272804875/)) reports an export of their meeting data is corrupt:

> The .csv file I get from Parabol does not seem to be formatted correctly and is not useful for examining meeting summaries in Excel. It appears to be one line (that does load into Excel, but not in a meaningful way). Is this expected?

Users can:
   - Triaged and fixed or,
   - Triaged and new implementation issue opened

tianrunhe:
Fixes #4489

### DIFF
--- a/packages/client/modules/email/components/SummaryEmail/ExportToCSV.tsx
+++ b/packages/client/modules/email/components/SummaryEmail/ExportToCSV.tsx
@@ -1,5 +1,5 @@
 import graphql from 'babel-plugin-relay/macro'
-import {Parser} from 'json2csv' // only grab the sync parser
+import Parser from 'json2csv/lib/JSON2CSVParser' // only grab the sync parser
 import withAtmosphere, {
   WithAtmosphereProps
 } from 'parabol-client/decorators/withAtmosphere/withAtmosphere'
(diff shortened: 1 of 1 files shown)


tianrunhe:
[Webpack bundle analyzer](https://github.com/webpack-contrib/webpack-bundle-analyzer) result:

before change:
Screen Shot 2021-01-06 at 1 27 33 PM

after change:
Screen Shot 2021-01-06 at 1 25 18 PM

mattkrick:
hmmm, 42B smells funky.

I think i told ya wrong. you gotta run that on the production webpack config!
the development webpack config keeps that in the DLL, so it won't be readily apparent.

the production config will also minify it, so we'll see a more accurate payload size

mattkrick:
@tianrunhe 
(@nickoferrall @tiffanyhan this is kinda neat stuff, so may be worthwhile to read it)

how webpack tree shaking works

When you import something in JS like `import foo from 'foo'`, how does it know what file to read? It grabs the file that's listed as `main` in your package.json. So in that case: https://github.com/zemirco/json2csv/blob/master/package.json#L22

Now, let's say that `index.js` file imported 2 other things:

```js
// index.js
import Parser from './parser'
import AsyncParser from './asyncParser'

export Parser
export AsyncParser
```

Since that file has 2 dependencies, webpack must include them both. Why? Because what if AsyncParser looked like this?

```js
// asyncParser.js
import Parser from './parser'
Parser.foo = 'hi there'
```

As a side effect, AsyncParser is mutating Parser! Gross!

To tell webpack that there are so side effects, the package has to declare itself as "pure" meaning it doesn't do any of that side effect junk, but not every package does that.

The alternative is to simply import _just_ the file you need:

```js
import Parser from 'json2csv/lib/parser'
```

PR #4522, opened 2021-01-13
Add pg

### LINKED ISSUE #4496: Include Postgresql in our bundle
AC
- [ ] add postgresql to the dockerfile 
- [ ] add instructions to the readme on how to get postgresql setup as a docker contianer
- [ ] add `pg` to the server package so we can start writing to the postgresql database
- [ ] add a singleton driver similar to the rethinkdb and redis drivers that opens a connection when first called & reuses that connection after that
- [ ] DevOps backlog issue created to add Postgres to our other environments
- [ ] 
Estimate: 16 hours

tiffanyhan:
Resolves #4496
- [ ] devs can follow `README.md` instructions to get a dockerized instance of postgres up for local development


packages/server/postgres/getPostgres.ts
```diff
@@ -0,0 +1,11 @@
+import { Pool } from 'pg'
+
+let pool
+const getPostgres = () => {
+  if (!pool) {
+    pool = new Pool({ max: 1 })
```
mattkrick, line 6: we'll wanna step this up for production. i think for rethinkdb we have 50 conns in prod & 3 in dev.

packages/server/postgres/init.sql
```diff
@@ -0,0 +1,2 @@
+DROP DATABASE IF EXISTS development;
```
mattkrick, line 1: if it's easier to use the same name for dev & prod, let's do it.
looking back, having `actionDevelopment` and `actionProduction` in rethinkdb was probably a mistake. plus they're really long names!

packages/server/postgres/Dockerfile
```diff
@@ -0,0 +1,4 @@
+FROM postgres:13.1
+ENV POSTGRES_PASSWORD password
```
mattkrick, line 2: i'm always scared of bad defaults.
if we write `ENV POSTGRES_PASSWORD` it will let docker know we expect that var, but then we don't have a bad default hardcoded in there

mattkrick (commented):
super work! excited to see what you do with this for the organization user ledger

packages/server/postgres/Dockerfile
tiffanyhan: @mattkrick can you elaborate on what you mean by this comment?

tiffanyhan: keep env vars in env file, not in code

tiffanyhan:
@mattkrick thanks for the review, this should be good to go now

summary of changes:
- removed `POSTGRES_PASSWORD` from `Dockerfile`, only set in `.env`
- postgres image version changed from 13.1 to 12.5 per @adaniels-parabol 
- reverted creation of separate metrics database
- changed default database name to `web`, which should be consistent across all envs
- renamed `getPostgres` to `getPgPool` for more specificity

.env.example
```diff
@@ -26,6 +26,8 @@ MAIL_GOOGLE_USER='key_MAIL_GOOGLE_USER'
 MAIL_GOOGLE_PASS='key_MAIL_GOOGLE_PASS'
 MIN_LOG_DURATION='200'
 PORT='3000'
+POSTGRES_URL='postgresql://postgres:password@localhost:5432/web'
```
mattkrick, line 29: let's have a single source of truth for the password. If it's already in the POSTGRES_URL, then no need for an additional env var

tiffanyhan: @mattkrick great idea, however after looking into it it seems that the two references different env vars and so the value may not always coincide. See note 2 below. 

given this info, i'm comfortable with leaving the 2 as separate. what do you think?

from https://hub.docker.com/_/postgres
```
POSTGRES_PASSWORD
This environment variable is required for you to use the PostgreSQL image. It must not be empty or undefined. This environment variable sets the superuser password for PostgreSQL. The default superuser is defined by the POSTGRES_USER environment variable.

Note 1: The PostgreSQL image sets up trust authentication locally so you may notice a password is not required when connecting from localhost (inside the same container). However, a password will be required if connecting from a different host/container.

Note 2: This variable defines the superuser password in the PostgreSQL instance, as set by the initdb script during initial container startup. It has no effect on the PGPASSWORD environment variable that may be used by the psql client at runtime, as described at https://www.postgresql.org/docs/current/libpq-envars.html. PGPASSWORD, if used, will be specified as a separate environment variable.
```

mattkrick: ah, that's an interesting thing they're doing with the docker image.
so the info we need is:
- SU user
- SU password
- user
- password
- host
- database
- port

from the docs, it looks like if we don't provide a `POSTGRES_USER` or `POSTGRES_DB`, then it'll create a database named `postgres`. 
does that mean if we include something like `POSTGRES_DB='web'` then we can get rid of the init.sql?  that may be cleaner

packages/server/postgres/init.sql
```diff
@@ -0,0 +1,2 @@
+DROP DATABASE IF EXISTS web;
```
mattkrick, line 1: will this drop the DB if the image gets rebuilt? i think if we specify a POSTGRES_DB env var then we can get rid of this file & have it make the DB for us

mattkrick:
almost done!
could you also add the env vars to the action-devops repo? we'll need a different cryptographically secure password for each env

packages/server/postgres/init.sql
tiffanyhan: yeah good idea

.env.example
tiffanyhan: > to DRY things out a bit, perhaps we just create these env vars instead of the URL? i don't have any tension sharing the SU user/pass for the app since the app will need SU permissions anyways.

@mattkrick by DRY, do you mean specifying the env vars `PGUSER`, `PGHOST`, `PGDATABASE`, `PGPORT` separately save `PGPASSWORD` and then when instantiating the pg client, programmatically specifying that `PGPASSWORD` equals `POSTGRES_PASSWORD`?

tiffanyhan:
> could you also add the env vars to the action-devops repo? we'll need a different cryptographically secure password for each env

hmm i'm not sure what exactly we're doing in each other env... if i remember correctly i think we're probably going to use a managed instance in either Digital Ocean or AWS? perhaps @adaniels-parabol could provide guidance?

references:
https://www.digitalocean.com/products/managed-databases-postgresql/

.env.example
mattkrick: the docker image will need POSTGRES_USER and POSTGRE_DB, too. we can just pass those in as params at runtime

mattkrick:
we'll have a managed instance in DO, but if we get the env vars set up now we can handle that setup outside of this PR

.env.example
tiffanyhan: just wanted to clarify: by `pass those in as params at runtime`, do you mean setting the docker postgres image env vars not in `.env` file but either in the Dockerfile, docker-compose file, or injecting them by always calling the docker-compose command with exported env vars?

wanted to make sure bc the comment seems to contradict an earlier suggestion to take the env vars out of Dockerfile and only set them in `.env` file.

also: not sure abt you but i'm finding it challenging to talk abt dry setup config stuff in comments, lmk if you prefer to take this sync.

tiffanyhan:
```
we'll have a managed instance in DO, but if we get the env vars set up now we can handle that setup outside of this PR
```

makes sense, i'll kick it off. note: i'll probably still include a `POSTGRES_URL` type env var, in addition to the `PGUSER`, `PGHOST`, `PGDATABASE`, `PGPORT`, and `PGPASSWORD` env vars. Reason being is bc it seems from the docs that `pgm` supports specifying an env var for a url, but not the separate `PGUSER`, `PGHOST`, `PGDATABASE`, `PGPORT`, and `PGPASSWORD` env vars. 

Reference: https://salsita.github.io/node-pg-migrate/#/cli

mattkrick:
duplicating those vars isn't great. could you just use a typescript file? https://github.com/lorenwest/node-config/wiki/Configuration-Files#typescript---ts

mattkrick:
closed in favor of #4580

Issue #4459, opened 2020-12-17
User replicated 3x


nickoferrall:
[An email](https://app.hubspot.com/contacts/3888472/contact/9033151) 🔒  from a paying customer:

> We have 3 [email redacted] accounts in parabol. The screenshot only shows 2, but there's another same one much lower. When I deactivated one of them (coz I don't want to pay triple for one user), it deactivated all three at the same time.

### AC

- [ ] Triage bug
- [ ] Remove replicated users from paying customer account

gcrickman:
This may be a separate issue but for the same client still occurring:

> We still notice the bug occurs whenever we add a new person to the team in the middle of a meeting. We have a lot of interns (about 20 at any given time) and there’s always someone leaving and someone new joining the company every week. 
> 
> So it’s a regular scenario for someone new to join the meeting as they get sent the meeting link. We see the “Mr. new person is requesting to join the meeting. Accept. Deny” (or something like that) while in the middle of the meeting. When one of us (or many of us) press “Accept”, errors start happening. 
>

nickoferrall:
I've created a PR for this replicated user bug (#4541) but the page freezing is likely a separate issue so I've created a new issue for that (#4542)

PR #4541, opened 2021-01-19
Fix replicated user

nickoferrall:
This PR fixes issue #4459 

A paying user reported that one of their users has been replicated three times. Deactivating one of the users deactivates all three.

It looks like multiple users must have clicked "Accept" request to join at the same time.

### Steps to reproduce:

- Copy & paste the `AcceptTeamInvitationMutation` in `InviteToTeamMutation` several times 
- Add a new user to an org. Some of the time, you will see that that user appears multiple times in org members
- If the user is added to an active meeting, they will also appear several times in the meeting 

### AC

- [ ] If multiple users click "Accept" request to join, a new user isn't replicated in the org 
- [ ] A new user isn't replicated in the team either
- [ ] Existing replicated users are removed by the migration


packages/server/database/migrations/20210118184615-removeDuplicateOrgUsers.ts
```diff
@@ -0,0 +1,51 @@
...
+        'orgId'
+      )('id')
+      .run()
+
+    const duplicatesToRemove = duplicateOrgUsers
+      .map((user) => user['reduction'].slice(1, user['reduction'].length))
+      .filter((user) => user.length)
+      .flat()
+
+    await r
```
mattkrick, line 31: how do we account for billing? i assume some of these duplicates are pro/enterprise?

packages/server/database/types/OrganizationUser.ts
```diff
@@ -26,7 +24,7 @@ export default class OrganizationUser {
 
   constructor(input: Input) {
     const {userId, id, removedAt, inactive, orgId, joinedAt, newUserUntil, role, tier} = input
-    this.id = id || generateUID()
+    this.id = id || `${userId}::${orgId}`
```
mattkrick, line 27: we can't do this because we need to keep track of people who were in an org, left, and then came back

packages/server/graphql/mutations/helpers/addTeamMemberToMeetings.ts
```diff
@@ -13,6 +13,10 @@ import createMeetingMembers from './createMeetingMembers'
  */
 
 const setInPhase = (phase: CheckInPhase | UpdatesPhase, newStage: CheckInStage | UpdatesStage) => {
+  const isMemberInMeeting = phase.stages.find(
```
mattkrick, line 16: ideally, the code would never reach this function if the team member is already in the meeting.
what is the condition where this function would get called when a team member is possible already in the meeting? we should identify that & exit a few functions earlier vs. waiting until this function

mattkrick (changes requested):

packages/server/billing/helpers/adjustUserCount.ts
```diff
@@ -84,10 +81,7 @@ const addUser = async (orgIds: string[], userId: string) => {
       new Date()
     return new OrganizationUser({orgId, userId, newUserUntil, tier: organization.tier})
   })
-  await r
-    .table('OrganizationUser')
-    .insert(docs)
-    .run()
+  await r.table('OrganizationUser').insert(docs, {conflict: 'replace'}).run()
```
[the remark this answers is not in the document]
nickoferrall: Makes sense, thanks. I just wanted to give you an update. I've added checks to see if the user exists in the org to prevent replication and it kinda works but not perfectly. 

When replicating the `AcceptTeamInvitationMutation` 5x, sometimes the 2nd or 3rd (or 5th) function invocation fires and reads the `OrganizationUser` table before the first `addUser` insert has completed. `userInOrg` is therefore false and it replicates the same user.

I've pushed an update that checks if the user already exists in the org just before the insert in `adjustUserCounts`. This helps and moving the `userInOrg` check below the `Promise.all` also gives it more time and reduces replication. 

But it's not a perfect solution. Occasionally, the user is still replicated. Strangely, it seems to happen with more frequency if I remove the user from the org & re-add them.

A few potential solutions I'm thinking of:

1. I can't see any code that checks if a user joins/leaves/joins/leaves. If that's the case, perhaps we could make `removedAt` and `joinedAt` and array instead and keep track of joining & leaving that way? Then we could use `userId::orgId` as the id
2. Try to limit the number of times the client mutation can be called
3. Adding timeouts or a function that removes duplicates at the end of `acceptTeamInvitation`. This seems very hacky

Do you have any other ideas I could explore?

packages/server/database/migrations/20210118184615-removeDuplicateOrgUsers.ts
nickoferrall: When removing an org user, I'm adding a `removedReason` field. Once the migration has been shipped, I was thinking I could do a query to check which users have the `removedReason` 'duplicate' . I imagine the list will be small and I could share that list with the sales team who can organise a refund.

packages/server/billing/helpers/adjustUserCount.ts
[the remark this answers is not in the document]
mattkrick: can you point to the race condition in the code? i'd think the easiest solution would be to make sure that the user has been inserted before continuing on

packages/server/database/migrations/20210118184615-removeDuplicateOrgUsers.ts
mattkrick: that's good for things in the past, but we'll still need to tell stripe that their seat count has decreased, otherwise stripe will keep charging the same price

packages/server/billing/helpers/adjustUserCount.ts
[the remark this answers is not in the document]
nickoferrall: Sure thing, the new `OrganizationUser` is being inserted [here](https://github.com/ParabolInc/parabol/blob/fix-replicated-user/packages/server/billing/helpers/adjustUserCount.ts#L96).

Do you mean check whether the user has not been inserted before inserting? If so, I'm doing this [here](https://github.com/ParabolInc/parabol/blob/fix-replicated-user/packages/server/billing/helpers/adjustUserCount.ts#L85) and [here](https://github.com/ParabolInc/parabol/blob/fix-replicated-user/packages/server/safeMutations/acceptTeamInvitation.ts#L109). 

The inserting is happening successfully so I'm not sure we'd need to check if the user has been inserted before continuing. The problem is the user is being inserted when it already exists. 

I also tried adding a `branch` statement to check if the user exists at the final moment and only insert if it doesn't but no dice. It seems that the 2nd/3rd function invocation reads from `OrganizationUser` before the 1st insert has happened so the checks return false

packages/server/database/migrations/20210118184615-removeDuplicateOrgUsers.ts
nickoferrall: Update pushed with Redis lock and updating Stripe! Two questions:

1. Do we only need to update Stripe and not the `Invoice` or `InvoiceItemHook` db tables? It looks like the `Invoice` fields are generated from info from Stripe [here](https://github.com/ParabolInc/parabol/blob/fix-replicated-user/packages/server/billing/helpers/generateInvoice.ts#L394) but I wanted to check

2. Can I access our Stripe account? I haven't been able to test whether the Stripe has actually updated following the migration

packages/server/database/migrations/20210126171515-removeDuplicateOrgUsers.ts
```diff
@@ -0,0 +1,82 @@
...
+    })
+
+    const orgs = (await r
+      .table('Organization')
+      .getAll(r.args(orgIds))
+      .hasFields('stripeSubscriptionId')
+      .pluck('id', 'stripeSubscriptionId')
+      .run()) as Organization[]
+
+    orgs.forEach(async (org) => {
```
mattkrick, line 52: remember forEach is a sync construct, so you'll probably want to use `map` and return an array of promises & then await the Promise.all of them

packages/server/database/migrations/20210126171515-removeDuplicateOrgUsers.ts
```diff
@@ -0,0 +1,82 @@
...
+      const stripeQty = stripeSubscription.quantity || 0
+      const dupsCount = orgDupsCount[orgId]
+      const newStripeQty = Math.max(1, stripeQty - dupsCount)
+      await manager.updateSubscriptionQuantity(stripeSubscriptionId, newStripeQty)
+    })
+
+    await r
+      .table('OrganizationUser')
+      .getAll(r.args(duplicateOrgUserIds))
+      .update({removedAt: now, removalReason: 'duplicate'})
```
mattkrick, line 65: do we have a rough idea of how many folks this affects?
it'd be nice to do a dry run of this on the prod DB first to get a rough idea just in case things go awry

nickoferrall: My guess is it's a super small number of paying orgs that have duplicate org users. I could copy the production db and run the query locally?

packages/server/database/migrations/20210127153000-removeDupOrgUsers.ts
```diff
@@ -0,0 +1,82 @@
+import {R} from 'rethinkdb-ts'
+import StripeManager from '../../utils/StripeManager'
+import Organization from '../types/Organization'
+export const up = async function(r: R) {
+  const now = new Date()
+  const manager = new StripeManager()
+  try {
+    const duplicateUserIds = await r
+      .table('OrganizationUser')
```
mattkrick, line 9: this query doesn't work on prod as there are more than 100k rows. we'll need a more efficient query to continue

Issue #4545, opened 2021-01-19
updateTask mutation not broadcasting changes


jordanh:
## Issue - Bug

Watch 🎞:  : [this Loom reproduction video](https://www.loom.com/share/09334ef7249349969e596e1a2b14f8d3)

Reproduction:
   - User A creates task card during Check-In meeting
   - User A edits and blurs task card, triggering updateTask
   - Other users in meeting do not see task card content update
      -  Other users do see task editing status change, and other broadcast events

Speculation: may be related to edit of check-in question, and not seeing those changes broadcast either

- **OS:** macOS
- **Browser:** Chrome
- **Estimated effort:** 16 hours ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

### Acceptance Criteria (optional)
Users can:
   - Edit task cards and have their changes broadcast to others

mattkrick:
this looks like a regression from code added around here: https://github.com/ParabolInc/parabol/blob/bd5e3d599d280e638f37b75bb1a61c3a5b3a9a67/packages/client/hooks/useEditorState.ts#L30

possible venues may be making sure that the logic there is firing & triaging from there

PR #4556, opened 2021-01-19
Fix task not updating bug

nickoferrall:
Fix for issue #4545

### AC:

- [ ] Edit task cards and have their changes broadcast to others
- [ ] Sentry error on line 52 isn't triggered on initial render


packages/client/hooks/useEditorState.ts
```diff
@@ -13,7 +13,7 @@ const useEditorState = (content?: string | null | undefined) => {
   const lastFiredRef = useRef<Date | null>(null)
   const initialRender = useRef(true)
   useEffect(() => {
-    if (initialRender) {
+    if (initialRender.current) {
```
mattkrick, line 16: omfg :see_no_evil:
nice catch

PR #4580, opened 2021-01-22
Add postgres

mattkrick:
Getting up & running with our repo has been a challenge in the past.
Now, as long as the dev has docker-compose, we'll take care of redis, rethinkdb, and postgres for them.

What this does:
- Sets up all the databases with a single command `db:start`
- Adds pgadmin4 as a db admin dashboard (see readme for how-to)
- Sets up node-pg-migrate to handle postgres migrations
- runs pg migrate when you call `yarn dev`

To install:
- [ ] add the new env vars to your .env
- [ ] uninstall all currently running databases. those are all managed for you now!
- [ ] install docker-compose if you haven't already
- [ ] call `yarn db:start` to get all databases running
- [ ] to add a new pg migration, call `yarn pg:migrate create FOO`
- [ ] you can run a migration with `yarn pg:migrate up` or just call `yarn dev` & it'll do it automatically


packages/server/postgres/pgmConfig.js
```diff
@@ -0,0 +1,12 @@
+
+const pgmConfig = {
```
mattkrick, line 2: @tiffanyhan the pgm docs are...not good. basically, i just debugged their code until i figured out how it was parsing this file & now it's working!

scripts/dev.js
```diff
@@ -41,14 +44,19 @@ const dev = async (maybeInit) => {
 
   const buildDLL = require('./buildDll')()
   const clearRedis = redis.flushall()
-  const migrateDB = require('./migrate')()
+  const migrateRethinkDB = require('./migrate')()
+  const pgConfig = {
```
mattkrick, line 48: @tiffanyhan these prop names were totally undocumented in the package, can't blame ya for not seeing them! should be good to go now though. this sets up the pattern to call pgm from code

docker/dev.yml
```diff
@@ -0,0 +1,55 @@
...
+    networks:
+      - parabol-network
+  postgres:
+    image: postgres:12.5
+    restart: unless-stopped
+    env_file: ../.env
+    ports:
+      - "5432:5432"
+    volumes:
+      - "../packages/server/postgres/postgres.conf:/usr/local/etc/postgres/postgres.conf"
```
mattkrick, line 23: @tiffanyhan i wanted to get rid of that DOCKERFILE so a fun little trick is to mount config files as volumes. doing this means we don't have to copy it into the container

docker/dev.yml
```diff
@@ -0,0 +1,55 @@
+
+version: "3.7"
```
mattkrick, line 2: i could see how running the prod docker-compose was probably causing some heartache, so i went ahead & made a dev one. this will be nice for folks moving forward, so now none of us are responsible for managing the DBs on new computers

mattkrick (commented):
@tiffanyhan i wanted to get you unblocked on the OrgUser stuff so i went ahead & fixed up the postgres stuff. take a look & let me know if you have any questions!

tiffanyhan:
this is... so much nicer. thank you for all the help with getting this set up!

Issue #4554, opened 2021-01-19
Add User table to postgres


mattkrick:
let's start to move to postgres. see #4530

AC
- [ ] Every time a User is created, they are also created in postgres
- [ ] Every time a User is updated, the update is propagated to postgres
- [ ] a postdeploy hook is written to copy all Users to postgres

Estimate: 16 hours

tiffanyhan:
Ran into the following problems with pgtyped:
- no support for postgres domains, ran into this while trying to implement a [custom domain for email address field](https://dba.stackexchange.com/questions/68266/what-is-the-best-way-to-store-an-email-address-in-postgresql)
- inconsistent casting of json types. this is apparently a [longstanding problem with node-postgres](https://github.com/brianc/node-postgres/issues/2012). only object values were correctly interpolated. non-empty array values couldn't not be saved into json field. empty array values weirdly get turned into `{}`. pgtyped doesn't do anything to address this problem.
- [incomplete support for postgres types](https://github.com/adelsz/pgtyped/issues/45) and no hook provided for custom type mappings

As an experiment, i tried installing zapatos to achieve this task. What I found:
- custom domain types are supported
- custom types are supported as a first-class citizen, with accessible hooks for user defined transformations of types
- addresses the json casting problem, by setting `castArrayParamsToJson` optional runtime config to true

Other differences:
- zapatos uses a sql template tag and doesn't seem to support `.sql` files. i think this is fine, since we don't already have a bunch of `.sql` files we have to work with.
- zapatos generates ALL crud input params and output types into one file from the database; not dependent on written code. pgtyped generates relevant input params and output types based on the database and the written queries in the codebase; generated types go in a separate file corresponding to the src query file. i appreciate the simplicity of the zapatos generation and the fact that we don't have to watch the code to generate the types.
- zapatos comes with helpers for basic crud queries. the helpers reduce a lot of boilerplate and are basically what i would write myself if using pgtyped.
- zapatos supports transactions, personally untested.
- zapatos supports joins returned as nested json. untested, but the nice interface with json gives me more confidence that this would offer a more convenient API for much of our json structured data.

mattkrick:
> no support for postgres domains, ran into this while trying to implement a custom domain for email address field

citext is the right answer. emails don't always have tlds, they just need an at sign. custom domains can often do more harm than good

> inconsistent casting of json types. 

[...]

PR #4582, opened 2021-01-25
Pgtyped orguser audit

### LINKED ISSUE #4081: OrganizationUser audit trail
## Issue - Enhancement

One day, it'd be nice to have an `OrganizationUserJournal` table that is updated each time something important happens to an Org User. Today, I think this should be when a user is created, marked active, marked inactive, or removed from an organization.

Today we can create a plot from backend data of the number of active users over time:

image

What I really want to be able to do is create analyses for ourselves and our users of how their organization's activity has changed over time. Because we only store the current state, this kind of historical analysis is currently impossible.

   - Migration to add `OrganizationUserAudit` table in POSTGRES
   - Row schema designed that include 'eventDate' timestamp
   - Rows added when a user is created, marked active, marked inactive, or removed from an organization

- **Estimated effort:** X hours ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

tiffanyhan:
Resolve #4081

**AC:**
Audit row(s) added when:
- [ ] a new user and org is bootstrapped
- [ ] a new org is created
- [ ]  a user joins a new org
- [ ] a user leaves an org
- [ ] a user is paused
- [ ] a user in unpaused
- [ ] a user deletes their account, thereby leaving all joined organizations


tiffanyhan:
Working with pgtyped was pretty delightful. i especially appreciated its enablement of named parameters and parameter expansions. Next steps would be:

- [ ] adding pgtyped watch mode to `dev.js`
- [ ] creating wrapper functions to manage pg pool connections and transactions. so we don't forget to release the client back to the pool, etc. this is in a lot of other libs that do a lot of other things, but we can probably just right it ourself.

packages/server/postgres/queries/OrgUserAudit.ts
```diff
@@ -0,0 +1,57 @@
+/** Types generated for queries found in "packages/server/postgres/queries/OrgUserAudit.sql" */
```
mattkrick, line 1: can we keep generated files in their own directory? it's always nice to be able to easily blow away a generated directory when something goes wrong or exclude from git if we need to down the road

package.json
```diff
@@ -24,6 +24,7 @@
     "start": "pm2-runtime start pm2.config.js",
     "heroku-postbuild": "echo 'Skipping build'",
     "build": "node scripts/prod.js",
+    "pg:build": "pgtyped -c ./packages/server/postgres/pgtypedConfig.js",
```
mattkrick, line 27: what do you recommend as a pattern for calling this? should we call it on ever server restart?
perhaps we could use pm2 to run a dedicated instance that watches a directory & runs it whenever that dir changes?
we can leave that piece out of scope for this PR, but it'll be good to think about for a future PR

packages/server/postgres/getPgClient.ts
```diff
@@ -0,0 +1,10 @@
+import getPgPool from './getPgPool'
```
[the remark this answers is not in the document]
tiffanyhan: i see the two as distinct. `getPgPool` gives access to the pool. you can then call `pool.query` which will get any available client from the pool then automatically release it after you call a single query.

`pool.connect` returns a single client which you can then use to do multiple queries. useful for transactions or when you need to ensure multiple queries go on the same connection in general. if using this pattern with vanilla `pg`, you have to remember to release the client manually.

`getPgPool` would be useful for usage of the first type, and `getPgClient` would be useful for usage of the second type. 

in `pgTyped`, the API requires passing a client into the generated function, so that's why `getPgClient` was used in this instance. i kept the two getters as separate as we're setting up the postgres API in general, but it could be that we only end up using the client getter in practice.

packages/server/postgres/migrations/1611342655964_addOrganizationUserAuditTable.ts
```diff
@@ -0,0 +1,25 @@
...
+export const shorthands: ColumnDefinitions | undefined = undefined;
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.sql(`
+    CREATE TABLE "OrganizationUserAudit" (
+      id SERIAL PRIMARY KEY,
+      "orgId" VARCHAR(100) NOT NULL,
+      "userId" VARCHAR(100) NOT NULL,
+      "eventDate" TIMESTAMP WITH TIME ZONE NOT NULL,
+      "eventType" INT NOT NULL
```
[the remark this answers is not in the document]
tiffanyhan: yea that is definitely a downside to keeping enums in the app. 

i think other pros of enums besides saving space is its easier for devs to change the enum without creating a migration, and it keeps business logic separate from data.

> in a future where we might have a data analytics team that works off of our DB, what's the recommended pattern for them to match the numeric value to the typescript definition?

that's a great question. i'd expect anyone working on data to be able to find the mapping in the codebase, but that may or may not be a reasonable expectation. i suppose there could also be some hook that updates the enum definitions to a wiki, so there's still a single source of truth in addition to an easy place for them to find it.

packages/server/postgres/types/OrgUserAuditEventTypeEnum.ts
```diff
@@ -0,0 +1,8 @@
+enum OrgUserAuditEventTypeEnum {
+  added = 1,
+  activated,
```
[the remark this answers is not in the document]
tiffanyhan: that's a great point. what did you have specifically in mind by, being "explicit with our namings"?

packages/server/postgres/queries/OrgUserAudit.ts
tiffanyhan: yeah sure!

package.json
tiffanyhan: pgtyped cli has watch mode, and i was thinking we could call this from `dev.js`. disclaimer: i haven't tried invoking watch mode manually yet, so i can't yet vouch for its reliability.

https://pgtyped.now.sh/docs/cli

packages/server/postgres/migrations/1611342655964_addOrganizationUserAuditTable.ts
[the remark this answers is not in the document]
tiffanyhan: just to be fair, other pros of keeping enums in the db are we get more semantic meaning when querying the db directly plus database-level validation of enums.

packages/server/postgres/helpers/OrgUserAudit.ts
```diff
@@ -0,0 +1,16 @@
...
+  orgIds: string[],
+  userId: string,
+  eventType: OrgUserAuditEventTypeEnum,
+  eventDate: Date = new Date()
+) => {
+  const pgClient = await getPgClient()
+  const auditRows = orgIds.map((orgId) => ({orgId, userId, eventType, eventDate}))
+  const parameters = {auditRows}
+  await insertQuery.run(parameters, pgClient)
+  pgClient.release()
```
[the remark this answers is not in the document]
tiffanyhan: i'm not sure i understand the question. what would be the purpose of having a pool of available connections if the client is a singleton?

packages/server/postgres/helpers/OrgUserAudit.ts
```diff
@@ -0,0 +1,16 @@
+import {insertOrgUserAudit as insertQuery} from '../queries/OrgUserAudit'
```
mattkrick, line 1: can we keep the file name == function name & export it as the default? doing this means we don't have to rename generated query when we import it

packages/server/postgres/getPgClient.ts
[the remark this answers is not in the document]
mattkrick: that'd be a dealbreaker for pgtyped if true!
can you show me where you're seeing it requires a client & not a pool? from what i see, the param def looks like:
```ts
export interface IDatabaseConnection {
    query: (query: string, bindings: any[]) => Promise<{
        rows: any[];
    }>;
}
```

lets get rid of `getPgClient` all together, it really only does 1 thing & that 1 thing is dangerous (provide an unterminating connection). when the time for transactions comes, those transactions can grab a connection in their own functions.

packages/server/postgres/helpers/OrgUserAudit.ts
tiffanyhan: @mattkrick hmm yeah that would be nice. if i'm not mistaken, i don't think there's any way to configure pgtyped to export a generated fn as the default. i think the intention of the author was for each sql file to contain multiple queries pertaining to one table.

packages/server/postgres/getPgClient.ts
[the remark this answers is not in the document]
tiffanyhan: yeah you're right. you can pass in either a pool or client. i was just going by the docs instead of looking at the types and source code.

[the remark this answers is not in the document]
tiffanyhan: > lets get rid of getPgClient all together, it really only does 1 thing & that 1 thing is dangerous (provide an unterminating connection). when the time for transactions comes, those transactions can grab a connection in their own functions.

i was seeing the client getter as more placeholder for work in progress. that's a good point though, somebody looking at the code as-is could be encouraged to do bad behavior.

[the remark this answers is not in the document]
tiffanyhan: client getter deleted

packages/server/postgres/migrations/1611342655964_addOrganizationUserAuditTable.ts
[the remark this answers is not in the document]
mattkrick: > i think other pros of enums besides saving space is its easier for devs to change the enum without creating a migration, and it keeps business logic separate from data.

if the type was a varchar would a migration still be needed?

> just to be fair, other pros of keeping enums in the db are we get more semantic meaning when querying the db directly plus database-level validation of enums.

could you tell me more about the semantic meaning?
how does the DB validate enums? is there something preventing me from passing in `249` when that enum has no reference?

packages/server/postgres/helpers/OrgUserAudit.ts
```diff
@@ -0,0 +1,15 @@
+import {insertOrgUserAudit as insertQuery} from '../queries/generated/OrgUserAudit'
+import getPgPool from '../getPgPool'
+import OrgUserAuditEventTypeEnum from '../types/OrgUserAuditEventTypeEnum'
+
+export const insertOrgUserAudit = async (
```
mattkrick, line 5: can we rename this file to `insertOrgUserAudit`?

packages/server/postgres/queries/src/OrgUserAudit.sql
```diff
@@ -0,0 +1,10 @@
+/* 
+  @name insertOrgUserAudit 
+  @param auditRows -> ((orgId, userId, eventDate, eventType)...)
+*/
+INSERT INTO "OrganizationUserAudit" (
```
mattkrick, line 5: let's rename this to `insertOrgUserAuditRow`

packages/server/postgres/helpers/OrgUserAudit.ts
```diff
@@ -0,0 +1,15 @@
+import {insertOrgUserAudit as insertQuery} from '../queries/generated/OrgUserAudit'
```
mattkrick, line 1: if we change the sql file name to `insertOrgUserAuditRow` then we can import that without having to rename it up here

packages/server/postgres/queries/src/OrgUserAudit.sql
```diff
@@ -0,0 +1,10 @@
+/* 
+  @name insertOrgUserAudit
```
mattkrick, line 2: lets have the name match the filename: insertOrgUserAuditRow

packages/server/postgres/getPgPool.ts
```diff
@@ -2,7 +2,7 @@ import {Pool} from 'pg'
 import PROD from '../PROD'
 
 let pool: Pool
-const getPG = () => {
+const getPgPool = () => {
```
mattkrick, line 5: lets rename to getPg since we'll be using this a lot!

packages/server/postgres/migrations/1611342655964_addOrganizationUserAuditTable.ts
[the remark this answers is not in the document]
mattkrick: after chatting, let's use the postgres enum type & use pg-typed to generate the enum in typescript land. then, we'll get checking for free in typescript PLUS validation at the DB level. win win!

packages/server/postgres/migrations/1611342655964_addOrganizationUserAuditTable.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export const shorthands: ColumnDefinitions | undefined = undefined;
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.sql(`
+    CREATE TABLE "OrganizationUserAudit" (
+      id SERIAL PRIMARY KEY,
+      "orgId" VARCHAR(100) NOT NULL,
+      "userId" VARCHAR(100) NOT NULL,
+      "eventDate" TIMESTAMP WITH TIME ZONE NOT NULL,
```
mattkrick, line 12: no timezone needed here, our DB & servers are all in UTC & if we need to convert it to a specific TZ we can, but we don't need to know the source tz.

packages/server/postgres/queries/generated/insertOrgUserAuditQuery.ts
```diff
@@ -0,0 +1,40 @@
+/** Types generated for queries found in "packages/server/postgres/queries/src/insertOrgUserAuditQuery.sql" */
+import { PreparedQuery } from '@pgtyped/query';
+
+export type OrganizationUserAuditEventTypeEnum = 'added' | 'activated' | 'inactivated' | 'removed';
```
mattkrick, line 4: :fire:

packages/server/graphql/mutations/helpers/createTeamAndLeader.ts
```diff
@@ -69,4 +70,5 @@ export default async function createTeamAndLeader(userId: string, newTeam: Valid
   if (!organizationUser) {
     await adjustUserCount(userId, orgId, InvoiceItemType.ADD_USER)
   }
+  await insertOrgUserAudit([orgId], userId, 'added')
```
mattkrick, line 73: why is this necessary? isn't this handled already in `adjustUserCount`? will calling it here create a dupe?

tiffanyhan: it's not handled in `adjustUserCount` bc an `OrganizationUser` is inserted separately on line 44. Can we get rid of that insertion in the string of promises, and just call `adjustUserCount` to make it consistent?

mattkrick: shoot i'm not seeing it. can you link to the link where `OrganizationUser` is getting inserted on L44?

tiffanyhan: oops sorry, on line 44 it just fetches to check if an `OrganizationUser` was already made. `adjustUserCount` never gets called bc the check always finds an `OrganizationUser` already. 

it looks like `bootstrapNewUser` and `addOrg` both call `createOrg` which creates an `Organization` as well as `OrganizationUser` before calling `createTeamAndLeader`.

tiffanyhan: maybe we can change `createNewOrg` to only create an `Organization` and not an `OrganizationUser`? it looks like `createNewOrg` is only called by `bootstrapNewUser` and `addOrg` and nothing else, so should be a safe change.

mattkrick: ah, makes sense.
we don't want to do org user audit things in `createTeamAndLeader` because that can be called by `addTeam`, which would result in multiple records.
instead, let's
- remove L70-L73 here because that conditional branch is never true
- add the `insertOrgUserAudit` call to `createNewOrg`

packages/server/graphql/mutations/helpers/createTeamAndLeader.ts
```diff
@@ -38,7 +36,7 @@ export default async function createTeamAndLeader(userId: string, newTeam: Valid
     orgId
   })
 
-  const [organizationUser] = await Promise.all([
+  await Promise.all([
     r
```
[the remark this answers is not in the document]
tiffanyhan: oh true true

tiffanyhan:
https://github.com/ParabolInc/parabol/issues/4554#issuecomment-770314089

^^ i'm beginning to have second thoughts about pgtyped, gonna try and make an alt PR for this with zapatos tmrw.

mattkrick:
for updating a field to null it should be like
@tiffanyhan 
```sql
UPDATE foo
SET bar = NULL
WHERE baz = 'hi';
```

PR #4588, opened 2021-01-26
Fix the issue where timeline events do not scroll

### LINKED ISSUE #4470: Timeline does not allow scrolling
## Issue - Timeline Bug

I have never been able to scroll through my timeline. It loads the 10 most recent meetings and doesn't allow me to see anything older than that. The scroll bar is showing as all the way at the end of the page and it looks as though the last meeting card is slightly cut off but there is no indication that there is anything further or that there should be the ability to scroll further. 

If I zoom out to 25% it shows 30 meetings on my timeline but the text is so small I can't read the details on the cards. I also don't know if there is more meeting history that still isn't showing.

- **OS:** macOS 
- **Browser:** Chrome Version 87.0.4280.88 (Official Build) (x86_64)
(as well as previous versions, this has always been the case for me)

tianrunhe:
Fixes #4470.

This works with one caveat: sometimes when you scroll down, more contents loaded but the scroll bar "bounces" back all the way to the top. @mattkrick Would like to have your insights.

### DIFF
--- a/packages/client/components/TimelineFeedList.tsx
+++ b/packages/client/components/TimelineFeedList.tsx
@@ -1,9 +1,14 @@
+import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
 import React from 'react'
 import {createPaginationContainer, RelayPaginationProp} from 'react-relay'
 import useLoadMoreOnScrollBottom from '~/hooks/useLoadMoreOnScrollBottom'
 import {TimelineFeedList_viewer} from '../__generated__/TimelineFeedList_viewer.graphql'
 import TimelineEvent from './TimelineEvent'
+
+const ResultScroller = styled('div')({
+  overflow: 'auto'
+})
 
 interface Props {
   viewer: TimelineFeedList_viewer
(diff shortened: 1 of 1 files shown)


packages/client/components/MyDashboardTimeline.tsx
```diff
@@ -47,10 +46,8 @@ const MyDashboardTimeline = (props: Props) => {
       <TimelineFeed>
         <TimelineFeedItems>
           <ErrorBoundary>
-            <Suspense fallback={<TimelineLoadingEvents />}>
```
tianrunhe, line 50: @mattkrick I fixed the bouncing issue by removing this block. Not quite sure why that's the cause though.

mattkrick: this tells me when a timeline event type is requested but the component hasn't been fetched yet, it will fallback to showing the mock for the entire list, not just that item!

to fix, instead of removing it here, add another suspense with fallback here: https://github.com/ParabolInc/parabol/blob/5a92a4605b066eb0bffb68f130b9ac10f8c3acf3/packages/client/components/TimelineEvent.tsx

PR #4604, opened 2021-01-27
make the poker deck slideable

mattkrick:
from early feedback, we've learned that the cards were overlapping too much, we need to support more than 20 values, & we need it to work on mobile.

this does those things. now, you can click & drag the deck to slide it. i opted for this slide behavior over a left/right button that deals things to minimize animations that aren't really critical to the experience.

@tianrunhe can you give this a test?

TEST
- [ ] start a meeting with 5 dimensions. each dimension has a different scale. the scales should have e.g. 2, 8, 15, 20, 25, 30 scale values, respectively
- [ ] the deck in each dimension looks good on desktop
- [ ] the deck also looks good on mobile
- [ ] if the width of the deck is greater than the width of the estimate phase area, you can drags it to the left & right


tianrunhe:
A fun one minor issue. To reproduce:
1. Vote a card
2. Reveal the card
3. Move the card you voted at bottom right away
4. Click "Team Revote"
Observe that the deck of the cards did not reset back to the middle position.

mattkrick:
fixed by disabling the slide when the deck is collapsed

Issue #4386, opened 2020-12-01
Robust MQ Protocol


mattkrick:
see #4343. we need a way to guarantee SOME subscription messages arrive at the client, otherwise, bad things can happen (like missing reflection cards).

maybe redis, maybe zmq or rabbitmq. needs exploration

AC
- [ ] a comparison of robust MQs is made & a decision is rendered

Estimate: 16 hours

mattkrick:
for the short term, i think redis fills the need. we can use https://redis.io/commands/rpoplpush for reliable messaging.
note not every message should be reliable! there should be some pattern matching. perhaps on the __typename?

AC
- [ ] every subscription message that must be robust is generated. the __typename for each payload robust payload is stored in a Set
- [ ] before a subscription payload is sent from the server, the __typename is checked against the Set above. if it matches, the payload is augmented with an `syn` flag
- [ ] trebuchet-client is patched to look for the `syn` flag. if found, it replies with an ack.
- [ ] trebuchet-client does not accept messages that come out of order (ie it either rejects out of order payloads or it keeps them in a queue until the one it's waiting on comes through. implementation TBD)
- [ ] when the server receives an ack the process is complete.
- [ ] if the server doesn't get an ack, it resends the syn payload using an exponential fallback.
- [ ] after X seconds without an ack, the server disconnects the client as it is unreachable
- [ ] the server uses redis' rpoplush to achieve robust messaging

PR #4606, opened 2021-01-28
Robust Message Queue Protocol

tianrunhe:
Implements #4386 .  See the corresponding [PR](https://github.com/mattkrick/trebuchet-client/pull/1) from trebuchet-client.


packages/server/socketHelpers/sendEncodedMessage.ts
```diff
@@ -1,16 +1,56 @@
...
+  const {socket, reliableQueue} = context
+  if (timeout > MAX_TIMEOUT) {
+    isHttpResponse(socket) ? socket.close() : handleClose(socket)
     return
   }
-  transport.send(str, false, str.length > ESTIMATED_MTU)
+  sendEncodedMessageBasedOnSocket(socket, message)
+  const timer = setTimeout(() => {
+    const newTimeout = timeout * TIME_OUT_COEFFICIENT
+    sendAndPushToReliableQueue(context, synId, message, newTimeout)
```
mattkrick, line 26: if this message failed to ack twice, this synId could have been created up to 8 seconds ago!
is it possible that a client gets 128 messages within 8 seconds? you bet!
instead of reusing an 8-second old synId, we should get a new one.
in other words, this function should call synId = context.synId++.

I'd suggest adding an API to the connectionContext like this:

```ts
getSynId() {
  this.synId++
  if (this.synId > MAX_ID) this.synId = 0;
  return this.synId
}
```
then, this function can just call context.getSynId()

mattkrick (changes requested):
solid first pass! making fresh synIds on each retry is the only big bugaboo i could find

packages/server/socketHelpers/ConnectionContext.ts
```diff
@@ -37,6 +42,14 @@ class ConnectionContext<T = WebSocket | HttpResponse> {
     this.readyQueue.forEach((thunk) => thunk())
     this.readyQueue.length = 0
   }
+
+  getMid() {
```
mattkrick, line 46: if a single connection sends 2B messages, our server will crash before we overflow :sweat_smile: 
this is good though, better safe than sorry. it'd be real hard to debug an overflow like this

packages/server/socketHelpers/sendEncodedMessage.ts
```diff
@@ -1,16 +1,57 @@
...
   }
-  transport.send(str, false, str.length > ESTIMATED_MTU)
+  const message = JSON.stringify([object, mid])
+  sendEncodedMessageBasedOnSocket(socket, message)
+  const timer = setTimeout(() => {
+    const newTimeout = timeout * TIMEOUT_COEFFICIENT + Math.random() * TIMEOUT_JITTER
+    sendAndPushToReliableQueue(context, mid, object, newTimeout)
+  }, timeout)
+
+  if (mid in reliableQueue) {
```
mattkrick, line 31: itty bitty super nitty:
- try to avoid using `in`. it's dangerous. in your browser console, type `'toString' in {}`. notice that it returns true. that's because `in` refers to all methods, not just the enumerable ones. for that reason, you have to follow it with `foo.hasOwnProperty('toString')`, unless `foo` was instantiated with `new Object(null)`, in which case you have to write `Object.prototype.call(foo, 'toString')` :vomiting_face: 
- as an alternative, you can just write `message = reliableQueue[mid]`. the property access cleaner & actually _faster_ than checking if the property exists!

packages/server/socketHandlers/handleMessage.ts
```diff
@@ -29,19 +30,45 @@ const handleParsedMessage = async (
...
+  message: ArrayBuffer
+) => {
   const {connectionContext} = websocket
   if (isPong(message)) {
     keepAlive(connectionContext)
     return
   }
+  if (isAck(message)) {
+    const mid = Buffer.from(message).readUInt32LE() >> 1
+    const timer = connectionContext.reliableQueue[mid].timer
```
mattkrick, line 59: never trust the client!

maybe i'm a sneaky & i tell you the `mid` is 2198759
reliableQueue[mid] will be undefined & your server will crash.

packages/server/socketHelpers/sendEncodedMessage.ts
```diff
@@ -1,16 +1,57 @@
...
-  if (transport.done) return
-  const str = JSON.stringify(message)
-  if (isHttpResponse(transport)) {
-    sendSSEMessage(transport as HttpResponse, str)
+const TIMEOUT_COEFFICIENT = 2.0
+const INITIAL_TIMEOUT = 2000 // 2 seconds
+const MAX_TIMEOUT = 10000 // 10 seconds
+const TIMEOUT_JITTER = 20 // 20 ms
+
+const sendAndPushToReliableQueue = (
```
mattkrick, line 13: if this function took in the stringified payload like `JSON.stringify([object, mid])` then we wouldn't have to re-stringify it every time we resent it

packages/server/socketHelpers/sendEncodedMessage.ts
```diff
@@ -1,16 +1,57 @@
...
+    const newTimeout = timeout * TIMEOUT_COEFFICIENT + Math.random() * TIMEOUT_JITTER
+    sendAndPushToReliableQueue(context, mid, object, newTimeout)
+  }, timeout)
+
+  if (mid in reliableQueue) {
+    const existingTimer = reliableQueue[mid].timer
+    clearTimeout(existingTimer)
+    delete reliableQueue[mid]
+  }
+  reliableQueue[mid] = {timer, object}
```
mattkrick, line 36: nit: the word `object` is pretty vague, maybe `payload`, `messageBody`, etc?

packages/server/socketHelpers/sendEncodedMessage.ts
```diff
@@ -1,16 +1,57 @@
...
+) => {
+  const {socket, reliableQueue} = context
+  if (timeout > MAX_TIMEOUT) {
+    handleDisconnect(context)
     return
   }
-  transport.send(str, false, str.length > ESTIMATED_MTU)
+  const message = JSON.stringify([object, mid])
+  sendEncodedMessageBasedOnSocket(socket, message)
+  const timer = setTimeout(() => {
```
mattkrick, line 26: we've already got a client where it asks for messages it missed (pull queue).
if the server also resends messages (push queue) then that means it's possible that the server could send the same message twice even if the client didn't request it.
that means trebuchet-client can't have just a `requestedMids` array, but it'll need to keep track of EVERY message that comes in, cause any message could be a duplicate! 

as a first pass, i'd suggest we just get rid of the `setTimeout` on the server. it'll reduce complexity & give us about the same quality, although i could be convinced either way

mattkrick (changes requested):
good stuff! all pretty minor, the only major thing is if we want the server to maintain a push queue via setTimeout. happy to discuss outside of GH!

mattkrick:
protocol looks good!
can you go ahead & add back in the blacklist for certain payload types & then bump to the latest trebuchet-client?

tianrunhe:
@mattkrick Ready for review for SSE handlers. The Typecheck failure should go away after we push the trebuchet-client changes.

mattkrick:
reviewing now...

packages/server/graphql/subscribeGraphQL.ts
```diff
@@ -93,15 +97,18 @@ const subscribeGraphQL = async (req: SubscribeRequest) => {
         reason: TrebuchetCloseReason.SESSION_INVALIDATED
       })
     }
-    sendGQLMessage(socket, 'data', payload, opId)
+    const subscriptionName = Object.keys(data)[0]
```
mattkrick, line 100: data is not guaranteed. could just be an error.

packages/server/parseBody.ts
```diff
@@ -9,7 +9,7 @@ const parseBody = ({
   res: HttpResponse
   parser?: (buffer: Buffer) => any
 }) => {
-  return new Promise<JSON | null>((resolve) => {
+  return new Promise<JSON | Buffer | null>((resolve) => {
```
mattkrick, line 12: this is a little brittle. the return type shall always be the return type of `parser` or `null`.

### Typescript Challenge for @tianrunhe @nickoferrall @tiffanyhan 

 try to rewrite this signature using a generic so whatever parser you pass into it, this function returns the return value of that generic or `null`.

for example:
```ts
const parser = (buffer: Buffer) => buffer.length === 0
const foo = await parseBody({res, parser}) // foo is boolean | null
```
 
The answer is here (no peeking until you're ready!): https://gist.github.com/mattkrick/c816bf3428b2ba47aa4d586bc55c8a4e

packages/server/socketHelpers/ConnectionContext.ts
```diff
@@ -37,6 +42,21 @@ class ConnectionContext<T = WebSocket | HttpResponse> {
...
+    this.mid++
+    if (this.mid >= MAX_MID) {
+      this.mid = 0
+    }
+    return this.mid
+  }
+
+  clearEntryForReliableQueue(mid: number) {
+    const entry = this.reliableQueue[mid]
+    if (entry) {
```
mattkrick, line 56: no need for the conditional, just calling `delete` is fine

packages/server/sse/SSEPingHandler.ts
```diff
@@ -12,7 +15,14 @@ const SSEPingHandler = (res: HttpResponse, req: HttpRequest) => {
       connectionContext.isAlive = true
     }
   }
+
+  const parser = (buffer: Buffer) => buffer
+  const messageBuffer = (await parseBody({res, parser})) as Buffer
```
mattkrick, line 20: after completing the typescript challenge, you can remove this cast

mattkrick (commented):
Just 1 typescript challenge away from merging!

mattkrick:
bumped to latest trebuchet-client, gonna give it a good kick tomorrow to make sure we're not running any mem leaks

mattkrick:
gave it a good kick locally, let's see how it holds up in prod!

PR #4611, opened 2021-01-28
removed isCheckedIn

### LINKED ISSUE #4351: Meetings should only include participants
​

AC

- [ ] MeetingMember is created lazily when people join

- [ ] summary is still sent to all team members

- [ ] event still gets sent to all team members

​

Estimate: 16 hours

mattkrick:
fix #4351 

@nickoferrall you've got some really good experience testing race conditions with joining meetings. mind trying to break this?

TEST
- [x] when a meeting is started, the only meeting member is the person who created the meeting
- [x] if 2 people join at the exact same time, everything works as intended (each person is a meeting member, 1 checkin stage each in a deterministic order)
- [x] if 1 person joins twice at the exact same time, everythign works as intended (1 meeting member, 1 stage per checkin, 1 stage per updates if check-in meeting) 

In the past, we built meetings so they started with every team member. 
We did that because we assumed absence would be the exception, not the norm.
If someone was absent from the meeting, we expected another team member to look at their board & discuss what they were up to.

...we were wrong.

A person shouldn't be a part of a meeting until they join the meeting.
When they join the meeting, a meeting member is created for them.

Notes:
- isCheckedIn is now gone. instead of keeping it around & always setting it to true, i decided to just nuke it. that means old meeting summaries can't tell how many were present/absent. it also means lastMetAt & monthlyStreak may be a bit more optimistic because it will assume someone being on a team with a meeting is the same as that person being in the meeting
-  when a user joins a meeting, instead of automatically checking them in, we automatically create a meeting member for them. this opens the door to spectators in the future
- viewerMeetingMember is now nullable. that's because the meeting is fetched before the viewer checks into the meeting. this isn't great because it requires 2 round trips to get into a meeting for the first time, but the alternative is a query that causes a side effect (joining a meeting). that feels worse.


packages/client/components/MeetingSelector.tsx
```diff
@@ -56,7 +55,6 @@ const MeetingSelector = (props: Props) => {
   useSubscription('MeetingSelector', OrganizationSubscription)
   useSubscription('MeetingSelector', TaskSubscription)
   useSubscription('MeetingSelector', TeamSubscription)
-  useSubscription('MeetingSelector', MeetingSubscription, {meetingId})
```
mattkrick, line 59: this occurs before the meeting member is created, so the subscription would fail. moved this to useAutoCheckin

packages/client/components/Snackbar.tsx
```diff
@@ -158,7 +152,7 @@ const Snackbar = React.memo(() => {
   }, [showSnack, transitionChildren])
 
   return portal(
-    <Modal snackbarOffset={snackbarOffset}>
+    <Modal isMeetingRoute={isMeetingRoute} isDesktop={isDesktop}>
```
mattkrick, line 155: the meeting control was getting measured before it had a chance to get relocated in the DOM, which caused a bad value for snackbarOffset, which means the snackbar was placed off the screen.

also, the snackbar was lined up in the center of the screen, but now it's offset by the left nav so it's directly above the control bar

also, when navigated to/from the meeting, now the y position updates correctly

packages/client/hooks/useAutoCheckIn.ts
```diff
@@ -2,30 +2,48 @@ import graphql from 'babel-plugin-relay/macro'
...
-    const {isCheckedIn} = viewerMeetingMember
-    if (!isCheckedIn) {
-      NewMeetingCheckInMutation(atmosphere, {meetingId, userId: viewerId, isCheckedIn: true})
+    const subscribeToMeeting = () => {
+      if (atmosphere.registerQuery) {
+        atmosphere.registerQuery(queryKey, MeetingSubscription, {meetingId}, router).catch()
+      }
+    }
+    if (viewerMeetingMember) {
+      subscribeToMeeting()
```
mattkrick, line 35: if they've previously joined the meeting, subscribe to the meeting

packages/client/hooks/useAutoCheckIn.ts
```diff
@@ -2,30 +2,48 @@ import graphql from 'babel-plugin-relay/macro'
...
+    } else if (!endedAt) {
+      JoinMeetingMutation(
+        atmosphere,
+        {meetingId},
+        {onCompleted: subscribeToMeeting, onError: () => { }}
```
mattkrick, line 40: if this is their first time in the meeting, subscribe to meeting events after they join

packages/client/hooks/useCallbackRef.ts
```diff
@@ -1,8 +1,8 @@
 import {useCallback, useRef} from 'react'
 import useForceUpdate from './useForceUpdate'
 
-const useCallbackRef = () => {
-  const ref = useRef<HTMLDivElement | null>(null)
+const useCallbackRef = <T = HTMLDivElement>() => {
```
mattkrick, line 4: @nickoferrall just a little typescript sugar so we can use this for more than just div elements

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/makeActionStats.ts
```diff
@@ -27,17 +26,12 @@ const makeActionStats = (meeting: Meeting) => {
...
-    meetingMembersPresentCount >= 10
-      ? meetingMembersPresentCount
-      : `${meetingMembersPresentCount}/${meetingMembersCount}`
+
   return [
     {value: doneTaskCount, label: `${plural(doneTaskCount, 'Task')} Done`},
     {value: agendaItemsCompleted, label: plural(agendaItemsCompleted, AGENDA_ITEM_LABEL)},
     {value: newTaskCount, label: plural(newTaskCount, 'New Task')},
-    {value: memberCount, label: 'Present'}
+    {value: meetingMembersCount, label: 'Participants'}
```
mattkrick, line 34: i chose not to be backwards compatible here because the value of this metric isn't that high compared to the extra complexity to maintain it.

packages/server/graphql/mutations/joinMeeting.ts
```diff
@@ -0,0 +1,149 @@
...
+    }
+    const {endedAt, teamId, phases} = meeting
+    if (endedAt) {
+      return {error: {message: 'Meeting has already ended'}}
+    }
+    if (!isTeamMember(authToken, teamId)) {
+      return {error: {message: 'Not on the team'}}
+    }
+    const meetingMember = createMeetingMember(meeting, teamId, viewerId)
+    const {errors} = await r
```
mattkrick, line 74: since the id is a composite of the userId & meetingId, calling this concurrently will result in only 1 meeting member & failing on the 2nd call

packages/server/safeMutations/acceptTeamInvitation.ts
```diff
@@ -112,9 +106,6 @@ const acceptTeamInvitation = async (
     await setUserTierForUserIds([userId])
   }
 
-  // if a meeting is going on right now, add them
-  await addTeamMemberToMeetings(teamMember, teamId, dataLoader)
```
mattkrick, line 116: joining a meeting when joining a team always felt bad. feels good doing this lazily, when the user joins the meeting

scripts/dev.js
```diff
@@ -37,9 +37,10 @@ const removeArtifacts = async () => {
 const dev = async (maybeInit) => {
   const isInit = !fs.existsSync(path.join(TOOLBOX_ROOT, 'updateSchema.js')) || maybeInit
   const redis = new Redis(process.env.REDIS_URL)
+  const toolboxPromise = compileToolbox()
```
mattkrick, line 40: always rebuild the toolbox in order to get a fresh schema in order to rebuild the schema. takes a little longer to start up, but guarantees that the graphql schema is correct

nickoferrall (approved):
Tests are all passing!

One tiny thing I noticed was that when a new user joins a meeting, their team presence icon at the top right of the page & the team member icon in the sidebar become visible at the exact same time. But when a user that is already on the team joins a meeting, their team presence icon displays before the team member icon shows up in the sidebar. 

I don't think that's a big deal at all but thought I'd mention it.

Also, if a user leaves a meeting and signs out, their team member icon remains. If they leave a meeting and leave the team/org, their icon is removed. I imagine that's the desired behaviour as the user presence icon keeps track of whether they're in the meeting or not.

mattkrick:
that's good insight! that's because today, the top right avatars are populated from the Team's TeamMembers instead of MeetingMembers. It's probably worth refactoring it so it pulls from the meeting members. I'll backlog a new issue with some possible design considerations (eg maybe we bring back the online/offline badges?)

Issue #4531, opened 2021-01-15
Fix Remote Drags when subcolumns are expanded


mattkrick:
Subcolumns are currently showing really buggy behavior for remote drags, when the card is dropped, it fades out (which is the behavior we expect when the destination is not visible on the screen).
To fix, it may be as easy as recalculating the bounding boxes 

AC
- [ ] remote drag & drops look good on sub column view

Estimate: 8 hours

nickoferrall:
@mattkrick if you get the chance, it'd be great to get a gif or loom video to see the faded out behaviour that you're seeing. 

I've tested it on my Mac, a remote windows device and my iPhone and I'm consistently getting the behaviour in this loom video. 

I'll work on the draggable logic but seeing the bug would be a big help!

https://www.loom.com/share/0c70a63ac0204cf6adc7eeccceedf68e

mattkrick:
i caught it in the last retro, couldn't catch it in dev. i'm OK with deprioritizing this for now until we can make it do it again. 

FWIW, the buggy code is here: https://github.com/ParabolInc/parabol/blob/cac1a6ce31b763831c5e981ed5638104b133b021/packages/client/utils/retroGroup/updateClonePosition.ts#L25

the problem is `isClipped` is coming up true when it shouldn't be. probably a problem with the bbox, which means probably a problem with the element. maybe the element is stale or something bailed? since it only happens sometimes, i bet it's something like setTimeout or rAF is getting called with a stale element. perhaps you could open the chrome dev tools & set the CPU to be 6x slower than normal to try to throttle the CPU, which might make skipping frames more common?

nickoferrall:
I've set the CPU to be 6x slower and while the cards move less smoothly, I haven't replicated the fading out behaviour yet. 

If the viewer quickly drags & drops a reflection card, `UpdateDragLocationMutation` doesn't fire so the card disappears from the drag position and appears at the drop position for both viewer and other members. It's more of a flash than a fade though and it's not remote drag specific.

`dropZoneEl` is the other area of the codebase I'm looking at as there used to just be one `dropZoneEl` per column but now there can be several.

I'll focus on other work until today's Product Retro and if the bug pops up again, I'll return to debugging this.

mattkrick:
ah! the dropzone seems interesting. we may need to give them all IDs that include the subCol number.

nickoferrall:
[...]

PR #4612, opened 2021-01-29
Retro columns fixes

nickoferrall:
While working on issue #4531, I noticed two UI bugs that this PR resolves:

1. On Windows, removing a reflection from a reflection group can cause `ColumnScrollContainer` to show a vertical scrollbar
2. If a column has two sub columns and you remove all of the reflections from one of the sub columns, the column will automatically minimise. By adding a `minWidth`, the column will only expand or minimise when the user decides.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,7 +48,7 @@ $ yarn db:start
 $ yarn dev
 ```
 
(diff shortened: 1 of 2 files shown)


mattkrick:
LGTM!

Issue #4619, opened 2021-01-29
Meeting settings in setActiveTemplate can be undefined


nickoferrall:
Issue reported by Sentry. Affected 15 users since release v5.31.3.

https://sentry.io/organizations/parabol/issues/2154982180/?referrer=github_plugin

```
TypeError: Cannot read property 'setLinkedRecord' of undefined
  at setActiveTemplateInRelayStore (./packages/client/utils/relay/setActiveTemplate.ts:12:21)
  at handler (./packages/client/mutations/AddReflectTemplateMutation.ts:41:3)
  at updater (./packages/client/subscriptions/TeamSubscription.ts:132:9)
  at Set.forEach (<anonymous>)
  at d (./packages/client/hooks/useDocumentTitle.ts:12:5)
...
(14 additional frame(s) were not displayed)
```

Issue #4618, opened 2021-01-29
Notifications can be undefined


nickoferrall:
Sentry issue that has affected 17 users over the past 9 months. 

Viewer can be undefined which we're already checking for. It looks like the viewer is occasionally defined but doesn't have any notifications. 

https://sentry.io/organizations/parabol/issues/1648269052/?referrer=github_plugin

```
TypeError: Cannot destructure property 'edges' of 'n' as it is undefined.
  at c (./packages/client/components/TopBarNotifications.tsx:25:10)
  at Array.forEach (<anonymous>)
  at Set.forEach (<anonymous>)
  at setConnectedStatus (./packages/client/hooks/useTrebuchetEvents.ts:19:7)
  at call (./packages/client/hooks/useTrebuchetEvents.ts:86:7)
...
(26 additional frame(s) were not displayed)
```

PR #4620, opened 2021-01-29
Fix for undefined meetingSettings & notifications

nickoferrall:
Fixes for issues #4618 and #4619 that were caught in Sentry.

### DIFF
--- a/packages/client/components/TopBarNotifications.tsx
+++ b/packages/client/components/TopBarNotifications.tsx
@@ -21,7 +21,7 @@ interface Props {
 
 const TopBarNotifications = (props: Props) => {
   const {viewer} = props
-  const {notifications} = viewer || {notifications: {edges: []}}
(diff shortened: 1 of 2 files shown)


packages/client/components/TopBarNotifications.tsx
```diff
@@ -21,7 +21,7 @@ interface Props {
 
 const TopBarNotifications = (props: Props) => {
   const {viewer} = props
-  const {notifications} = viewer || {notifications: {edges: []}}
+  const notifications = viewer?.notifications || {edges: []}
```
mattkrick, line 24: any ideas why viewer would be null?

nickoferrall: The viewer can be null on the initial render before the query is finished but I'm not sure why the viewer can exist but notifications can be null / undefined

Issue #4542, opened 2021-01-19
Page freezes when accepting request to join meeting


nickoferrall:
A [paying user](https://app.hubspot.com/contacts/3888472/contact/9033151) has reported that the page sometimes freezes for their team when they "Accept" a request to join a meeting.

> We still notice the bug occurs whenever we add a new person to the team in the middle of a meeting. We have a lot of interns (about 20 at any given time) and there’s always someone leaving and someone new joining the company every week.
> 
> So it’s a regular scenario for someone new to join the meeting as they get sent the meeting link. We see the “Mr. new person is requesting to join the meeting. Accept. Deny” (or something like that) while in the middle of the meeting. When one of us (or many of us) press “Accept”, errors start happening.

I've created a PR (#4541) for an issue where a user was replicated 3x (#4459) when multiple team members clicked "Accept" at the same moment.

I suspect that this bug is in the same family of errors although I haven't been able to reproduce it yet. It's possible that #4541 has fixed this bug too but I think we should investigate. 

### AC

- [ ] If multiple users accept a request to join a meeting, the page reliably works and doesn't freeze

nickoferrall:
Sentry hasn't been picking up this error due a bug that has been fixed in #4601 and I haven't been able to reproduce it.

The next time this bug occurs, Sentry should pick it up and help to debug it.

nickoferrall:
Infinite loop triggered when a new user tries to join a meeting. Possibly a related bug.

Sentry issue: [ACTION-PRODUCTION-3ST](https://sentry.io/organizations/parabol/issues/1640701522/?referrer=github_plugin)

PR #4625, opened 2021-02-01
Fix new joiner infinite loop 

nickoferrall:
This PR _may_ resolve issue #4542. 

[This](https://sentry.io/organizations/parabol/issues/1641789488/?project=107196&query=level%3Afatal&statsPeriod=14d) and [this](https://sentry.io/organizations/parabol/issues/2109105416/?project=107196&query=level%3Afatal&statsPeriod=14d) Sentry issue suggest that the new joiner can infinitely bounce between the [ViewerNotOnTeam](https://github.com/ParabolInc/parabol/blob/91aff6e5c28bf344521db5701b0113401a2bc1dc/packages/client/components/ViewerNotOnTeam.tsx#L44) and the [MeetingSelector](https://github.com/ParabolInc/parabol/blob/01bfad9f162ef8f69c034a956bfbce9db3b1e9b9/packages/client/components/MeetingSelector.tsx#L35) components.

I haven't been able to reproduce the issue but I suspect that the `teamId` may be added to Relay's `authObj` before the server updates with the meeting. `isOnTeam` is therefore true in `ViewerNotOnTeam` but the `MeetingSelector` sends the viewer back to `ViewerNotOnTeam` as it doesn't have the `meeting` yet.

### DIFF
--- a/packages/client/components/ViewerNotOnTeam.tsx
+++ b/packages/client/components/ViewerNotOnTeam.tsx
@@ -24,6 +24,7 @@ interface Props {
 const ViewerNotOnTeam = (props: Props) => {
   const {viewer} = props
   const {
+    meeting,
     teamInvitation: {teamInvitation, meetingId, teamId}
   } = viewer
   const atmosphere = useAtmosphere()
@@ -41,10 +42,15 @@ const ViewerNotOnTeam = (props: Props) => {
           {invitationToken: teamInvitation.token},
           {history, meetingId}
         )
-      } else if (isOnTeam) {
-        // if already on the team, goto team dash
-        const redirectTo = getValidRedirectParam()
-        const nextRoute = redirectTo || `/team/${teamId}`
+        return
+      }
+      const redirectTo = getValidRedirectParam()
+      // if already on the team, go to team dash
+      const nextRoute = redirectTo || `/team/${teamId}`
+      const isRoutingToMeeting = nextRoute === `/meet/${meetingId}`
+      const isRoutingToTeam = nextRoute === `/team/${teamId}`
+      const isNextRoutePossible = (isOnTeam && isRoutingToTeam) || (meeting && isRoutingToMeeting)
+      if (isNextRoutePossible) {
         history.replace(nextRoute)
       } else if (teamId) {
         PushInvitationMutation(atmosphere, {meetingId, teamId})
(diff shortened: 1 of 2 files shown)


mattkrick:
these are always fun ones. if your assumption is correct, then we should just ignore what the authObj says. i'll add a couple comments

packages/client/components/ViewerNotOnTeam.tsx
```diff
@@ -41,10 +42,15 @@ const ViewerNotOnTeam = (props: Props) => {
...
-        const redirectTo = getValidRedirectParam()
-        const nextRoute = redirectTo || `/team/${teamId}`
+        return
+      }
+      const redirectTo = getValidRedirectParam()
+      // if already on the team, go to team dash
+      const nextRoute = redirectTo || `/team/${teamId}`
+      const isRoutingToMeeting = nextRoute === `/meet/${meetingId}`
+      const isRoutingToTeam = nextRoute === `/team/${teamId}`
+      const isNextRoutePossible = (isOnTeam && isRoutingToTeam) || (meeting && isRoutingToMeeting)
```
mattkrick, line 52: if the problem is with the auth token updating before the query updates, I think we can simply this:
```
if (teamInvitation) acceptTeamInvitation...
else if (meeting) history.replace(/meet/meetingId)
else if (team) history.replace('/team/teamId')
else PushInvitation

nickoferrall: makes sense! I've removed the `authObj` check here and in the TeamContainer

mattkrick:
TESTS
- [ ] invite user b to the team. User b goes to the meeting route. the team invitation is auto accepted & they go to the meeting
- [ ] invite user b to the team. User b goes to the team route. the team invitation is auto accepted & they go to the team dash
- [ ] while user b is on the invitation-required route, user a invites user b from a meeting via email. they get redirected to the meeting
- [ ] same as above, but invite comes from the team dash. user b lands on the team dash
- [ ] user b goes to team dash without an invite. they go to invitation-required & user a gets a snackbar asking to accept
- [ ] user b goes to meeting without an invite. they go to invitatino-required & user a gets a snackbar asking to accept
- [ ] user b is on the team & goes to the meeting route, they go to the meeting

Issue #4627, opened 2021-02-02
TaskFooterUserAssigneeMenu fatal bug


nickoferrall:
Bug affecting 6 users over the past two months. Can't destructure userId from teamMembers.

https://sentry.io/organizations/parabol/issues/2059791893/?referrer=github_plugin

```
TypeError: Cannot destructure property 'userId' of 'undefined' as it is undefined.
  at None (./packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterUserAssigneeMenu.tsx:28:55)
  at Array.map (<anonymous>)
  at a (./packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterUserAssigneeMenu.tsx:28:49)
  at c (./packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterUserAssigneeMenu.tsx:28:23)
...
(14 additional frame(s) were not displayed)
```

PR #4628, opened 2021-02-02
Fix task footer menu fatal error

nickoferrall:
PR to fix issue #4627

### DIFF
--- a/packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterUserAssigneeMenu.tsx
+++ b/packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterUserAssigneeMenu.tsx
@@ -24,13 +24,13 @@ const TaskFooterUserAssigneeMenu = (props: Props) => {
   const {area, menuProps, task, viewer} = props
   const {userId, id: taskId} = task
   const {team} = viewer
-  const {teamMembers}: any = team || {teamMembers: []}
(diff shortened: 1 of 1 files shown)


packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterUserAssigneeMenu.tsx
```diff
@@ -24,13 +24,13 @@ const TaskFooterUserAssigneeMenu = (props: Props) => {
   const {area, menuProps, task, viewer} = props
   const {userId, id: taskId} = task
   const {team} = viewer
-  const {teamMembers}: any = team || {teamMembers: []}
-  const taskUserIdx = useMemo(() => teamMembers.map(({userId}) => userId).indexOf(userId) + 1, [
+  if (!team) return null
```
mattkrick, line 27: since this component could re-render, it may be worth it to move this check to L43 after all the hooks have run so react doesn't get a hook count mismatch across renders

nickoferrall: Makes sense! Strange thing with this one is [Sentry suggests](https://sentry.io/organizations/parabol/issues/2059791893/?project=107196&referrer=github_plugin) that `team` can exist with `teamMembers` being `null` even though it's [non-nullable](https://github.com/ParabolInc/parabol/blob/a9cc0145d12379e863ab6e12a74498d7cc82036c/packages/server/graphql/types/Team.ts#L277).

PR #4641, opened 2021-02-03
Fix pinned items removed on meeting end

### LINKED ISSUE #4352: Pinned items removed on meeting end
After ending our Product check-in meeting, the pinned agenda items were removed from the team. During the meeting, we were unable to remove an agenda item (see #4348) - perhaps the bugs are related. 

### AC

- [ ] Able to end a check-in meeting without removing the pinned agenda items

EE: 4 hours

nickoferrall:
PR to fix #4352 

`PromiseAll` guarantees the order on the client but not on the backend. It looks like the agenda items could occasionally be cloned and then cleared rather than cleared and then cloned.


mattkrick:
fantastic find!

PR #4643, opened 2021-02-03
Fix collapsing stack

nickoferrall:
This PR fixes the collapsing stack issue where the reflection group doesn't land precisely in the target after collapsing the stack. This will hopefully resolve the remote reflection buggyness outlined in issue #4531.


mattkrick:
you found it! nice

Issue #4428, opened 2020-12-10
Updated palette explored


ackernaut:
We’re working on an updated palette as part of a brand refresh.

- First order of business is to make sure it looks good in the UI (this issue)
- Then we can build out the palette #4372 
- Then we can apply the palette in the app (TK issue)

### Acceptance criteria
Made some static visuals that show:
- A target meeting concept in light mode
- A target meeting concept in dark mode
- A target dashboard concept in light mode
- A target dashboard concept in dark mode

### Estimated effort
- 12 hours

ackernaut:
### Burndown
Working [in Figma](https://www.figma.com/file/OA9NkpSTlHVqqRL9IE9KsF/Palette-v3?node-id=15%3A366):
- [ ] Checked green secondary usage in the UI
- [ ] Checked Estimate phase w/ various colors on cards etc.
- [ ] Checked Reflect phase w/ various prompt colors
- [ ] Checked Dashboard view with tasks
- [ ] Audited shades of gray in app and resolved to 9 variants (100-900)
- [ ] Created subset of 9 variants for blue, yellow, any other UI shades
- [ ] Stretch: spot checked a dark mode
- [ ] Started a draft of updated usage guidelines
- [ ] Updated #4372 with latest proposal

ackernaut:
Moving this back to in progress. I’m planning to capture the progress, takeaways, and next steps so I can be done with the exploration. We’re real close, but there are a few colors to suss out in order to implement #4372

Issue #4372, opened 2020-11-23
Refactor color palette


mattkrick:
We've got way, way too many colors in the palette.
After reading this article (https://refactoringui.com/previews/building-your-color-palette/) i think we should follow suit.
Pick a primary, blue, green, red, gray & get 9 shades, call them 100-900.
as far as naming, i think it should be simple as RED_100, BLUE_500, PURPLE_900, etc. 
i think we have to decouple what they are from how they're used (e.g. replace GRAY_BACKGROUND with GRAY_300). _how_ they're used can live in the comments, but they should only serve as guidelines which can be broken.

AC
- [ ] PALETTE is replaced with PALETTE_V3
- [ ] PALETTE_V3 has a max of 9 shades per color
- [ ] [Burndown](https://github.com/ParabolInc/parabol/issues/4372#issuecomment-789284623)

Stuck on: Branding complete

Estimate: 16 hours

ackernaut:
That’s basically what we used to do. It’ll take some effort to resolve what we get from the branding exercise, as you noted. I’m optimistic if we can get the variants right. Thanks for capturing!

ackernaut:
I think one thing we should consider when refactoring the palette is if we could potentially build a light and dark theme with it. I’m okay kicking a dark theme down the road but I wonder how it might increase attachment with agile engineering teams. Here’s an example of a product from the wild: https://linear.app/

mattkrick:
absolutely we want dark theme.
in a perfect work, we could use e.g. RED_500 on both light & dark, but i fear contrast might not always be enough.
in an almost perfect world, we could create lookup rules a la i18n & say use RED_500 for light and RED_300 for dark. 
in a bad world, we use RED_300 in places, RED_200 in others, and BLUE_300 in others. this is what we want to avoid.

ackernaut:
Inspiration: [https://dribbble.com/search/dark%20mode](https://dribbble.com/search/dark%20mode)

ackernaut:
@mattkrick should we scope a pattern for dark mode in this issue or separate out that work? I could stand to look at how other folks are handling it, but I wonder if we use some sort of theme mapper for only the colors that diverge. Drunken code:

```javascript
export const enum DarkTheme {
  CARD_SURFACE = PALETTE.GRAY_800
}

export const enum LightTheme {
  CARD_SURFACE = PALETTE.LIGHT_100
}
```

We could resolve the theme at some higher context and pass that to the needy components: 

```javascript
[...]

PR #4658, opened 2021-02-09
Palette refactoring part 1

ackernaut:
Working on #4428 aiming toward #4372


ackernaut:
@mattkrick would you mind taking a look at this?
- I removed some unused components (and their palette variables)
- I updated the values in palette v2 to the new values, with comments to show which value it is
- I added palette v3 with the used values. Note those const enums are not used yet but the values are in palette v2

My proposal is that we only include values used, not all weights in our expanded palette. We can add new ones from that palettes if and when we use them in design.

For gradients, we use palette values and name them as such.

I also think we only use opacity when truly needed, e.g. dialog scrims. All other color values are effectively opaque so we should use a weighted value.

Shall I do a find and replace and get rid of palette v2 entirely now that the existing colors are mapped?

ackernaut:
This will solve for #4372 since the new palette has been created

Issue #4495, opened 2021-01-05
Sort reflections between Reflect and Group phase


mattkrick:
between the reflect & group phase, we can do some analysis on the content of the reflections & take a best guess at which ones belong with other ones.

- We know magically grouping them doesn't work. people don't like it.
- team retro suggests groups & makes people accept/deny the magical groups. this is better.
- my suggestion is to simply put similar cards next to each other. that way, folks feel like they did it, it was their idea. It'll be one of those subtle things why parabol just makes meetings better, but you don't know why.

AC
- [x] When the facilitator navigates to the grouping phase for the very first time, the navigateMeeting mutation calls handleCompletedStage
- [x] in handleCopletedStage, when the stage completed is the reflect stage, we update the `sortOrder` for all reflection groups after removing the empty reflections (L19)
- [x]  The ReflectPhaseCompletePayload includes a `reflectionGroups` field
- [x] On the client, the navigatemeetingMutation reqeusts `reflectionGroups { sortOrder}` from the reflect phase complete payload
- [x] when a reflection is ungrouped, the sort order of its old group is identified & the ungrouped reflection is put into its own group. The sort order of that new group is determined based off the `(old group's sortOrder + the sort order of the group after it) / 2 + noise()` unless the old group is at the bottom. if that's the case, then it is `old group's sortOrder + 1 + noise()`
- [x] when a new reflection is created, it's put at the bottom of the shortest sub column. To accomplish this, the reflection group's sortOrder is calculated by finding the subColumnIdx with the fewest number of reflections & then grabbing the sort order of the bottom reflection group & adding 1 + noise() to that.

Estimate: 8 hours

PR #4661, opened 2021-02-10
Implement grouping sorting, using the autoGroupReflection method

tianrunhe:
implements #4495 

Pretty simple & lightweight implementation, uses the already well-written `groupReflections` method. Algorithm:
1. Group the reflections behind the scene
2. Sort the reflections based on their group membership
3. Reflections that belong to the same group sit together

### DIFF
--- a/packages/client/components/GroupingKanbanColumn.tsx
+++ b/packages/client/components/GroupingKanbanColumn.tsx
@@ -22,6 +22,7 @@ import {
 } from '../types/constEnums'
 import {NewMeetingPhaseTypeEnum} from '../types/graphql'
 import getNextSortOrder from '../utils/getNextSortOrder'
+import groupReflections from '../utils/smartGroup/groupReflections'
 import {SwipeColumn} from './GroupingKanban'
 import GroupingKanbanColumnHeader from './GroupingKanbanColumnHeader'
 import ReflectionGroup from './ReflectionGroup/ReflectionGroup'
@@ -85,6 +86,8 @@ interface Props {
   swipeColumn?: SwipeColumn
 }
 
+const AUTO_GROUPING_THRESHOLD = 0.5
+
 const GroupingKanbanColumn = (props: Props) => {
   const {
     columnsRef,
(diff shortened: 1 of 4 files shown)


packages/client/components/GroupingKanbanColumn.tsx
```diff
@@ -115,6 +118,18 @@ const GroupingKanbanColumn = (props: Props) => {
     () => groups.filter((group) => group.reflections.length > 0),
     [groups]
   )
+
+  const allReflections = filteredReflectionGroups.flatMap(({reflections}) => reflections)
+  // only do magical sorting before users create groups
+  if (allReflections.length === filteredReflectionGroups.length) {
```
mattkrick, line 124: this feels a little brittle. what if i group 1 & then ungroup it?
instead, let's put this in a hook that runs on initial render AND when nothing is grouped. that way the only time it'll run is if they refresh the initial state

packages/client/components/GroupingKanbanColumn.tsx
```diff
@@ -85,6 +86,8 @@ interface Props {
   swipeColumn?: SwipeColumn
 }
 
+const AUTO_GROUPING_THRESHOLD = 0.5
```
mattkrick, line 89: 0.5 feels a bit to high for an initial coefficient, is that what we had it as before? getGroupMatrix will increment this up to 5 times in order to find a good fit, no harm in starting low

packages/client/components/GroupingKanbanColumn.tsx
tianrunhe: Let's say before sorting, the reflections are `A, B, A', B'`. Now you enter the group phase, we sort the cards for you so it looks like `A, A', B, B'`. You grabbed card `B`, group it with `B'` and ungroup it immediately. Don't you want the state resets to what it look like **before** you grabbing the card, i.e., `A, A', B, B'`? Wouldn't it feel bad if it goes back to `A, B, A', B'`?

mattkrick: if the drop location of `B` is deterministic, you're absolutely right!
With the new subcolumns feature, when you ungroup `B` it could wind up in the first or 2nd column. if it got dropped & went to the 2nd column, it'd feel bad if it then magically teleported to the first column.

alternatively, let's say you group all 4 together, then ungroup all 4 in a random order. that ungrouping order might be `B', A', A, B`. It'd feel bad if they all got resorted once the last card got ungrouped.

tianrunhe: Alright makes sense! Technical question though: how do I achieve the effect of "running sth only **before** first render"? I cannot use `useEffect` as it will run **after** the render. `useState` feels weird too. It looks like something that should be inside `componentWillMount` but it seems there's no equivalent hooks for it.

mattkrick: big picture, a functional component is the same as the `render` method of a class component, so everything inside the function gets called on render, never before. componentWillMount has been deprecated because in the future, react be doing some work, determine it won't finish that work within 14ms, and put it on pause, which could cause a whole lot of sadness if that hook got called but the component then had to wait for the next animation frame to run.

for this case, i'd suggest `useMemo` with no deps so it just runs once.

tianrunhe:
Close in favor of #4696

PR #4680, opened 2021-02-17
(bug): make SAML Issuer and AssertionConsumerServiceURL dynamic

### LINKED ISSUE #4679: Make saml:Issuer and AssertionConsumerServiceURL to be dynamic
## Issue - Bug

Make saml:Issuer and AssertionConsumerServiceURL to be dynamic. Currently action.parabol.co is hardcoded. Lines of code referenced below.

https://github.com/ParabolInc/parabol/blob/master/packages/server/graphql/intranetSchema/mutations/enableSAMLForDomain.ts#L17-L18

This causes issue with SAML SSO setup on private managed instances at different URLs

adaniels-parabol:
Makes the domain name used in SAML template dynamic.

closes #4679


mattkrick:
i totally forgot we had that env var already. fantastic!

PR #4682, opened 2021-02-18
Fix the issue where only meeting member get summary email

### LINKED ISSUE #4681: Meeting summary gets sent to all team members
When a meeting is completed, all team members should see the summary.

AC
- [ ] https://github.com/ParabolInc/parabol/blob/240950a38ffae1eae0b6293f06babfdb3ae9f74e/packages/server/graphql/mutations/helpers/endMeeting/sendNewMeetingSummary.ts#L23 should get the list of userIds from the active team members, not just meeting members.
- [ ] https://github.com/ParabolInc/parabol/blob/592b2f5b70302b091dfd5b7c46b0ed7111d412bc/packages/server/graphql/mutations/endRetrospective.ts#L127 uses team members instead of meeting members

Estimate: 1 hour

tianrunhe:
Fixes #4681 

Tests:

- [ ] All team members should get meeting summary
- [ ] All team members should have meetings available in their timeline

### DIFF
--- a/packages/server/graphql/mutations/endCheckIn.ts
+++ b/packages/server/graphql/mutations/endCheckIn.ts
@@ -223,9 +223,10 @@ export default {
     }
 
     // remove any empty tasks
-    const [meetingMembers, team, removedTaskIds, result] = await Promise.all([
+    const [meetingMembers, team, teamMembers, removedTaskIds, result] = await Promise.all([
       dataLoader.get('meetingMembersByMeetingId').load(meetingId),
       dataLoader.get('teams').load(teamId),
+      dataLoader.get('teamMembersByTeamId').load(teamId),
       removeEmptyTasks(meetingId),
       finishCheckInMeeting(completedCheckIn, dataLoader)
     ])
@@ -233,7 +234,7 @@ export default {
     const updatedTaskIds = (result && result.updatedTaskIds) || []
     sendMeetingEndToSegment(completedCheckIn, meetingMembers as MeetingMember[])
     sendNewMeetingSummary(completedCheckIn, context).catch(console.log)
-    const events = meetingMembers.map(
+    const events = teamMembers.map(
       (meetingMember) =>
         new TimelineEventCheckinComplete({
           userId: meetingMember.userId,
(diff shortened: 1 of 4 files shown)


packages/server/graphql/mutations/endCheckIn.ts
```diff
@@ -223,17 +223,18 @@ export default {
...
+      dataLoader.get('teamMembersByTeamId').load(teamId),
       removeEmptyTasks(meetingId),
       finishCheckInMeeting(completedCheckIn, dataLoader)
     ])
     endSlackMeeting(meetingId, teamId, dataLoader).catch(console.log)
     const updatedTaskIds = (result && result.updatedTaskIds) || []
     sendMeetingEndToSegment(completedCheckIn, meetingMembers as MeetingMember[])
     sendNewMeetingSummary(completedCheckIn, context).catch(console.log)
-    const events = meetingMembers.map(
+    const events = teamMembers.map(
```
mattkrick, line 237: the predicate on the next line should be `teamMember` not `meetingMember`

packages/server/graphql/mutations/endRetrospective.ts
```diff
@@ -114,17 +114,18 @@ export default {
...
+      dataLoader.get('teamMembersByTeamId').load(teamId),
       removeEmptyTasks(meetingId),
       dataLoader.get('meetingTemplates').load(templateId)
     ])
     endSlackMeeting(meetingId, teamId, dataLoader).catch(console.log)
     finishRetroMeeting(completedRetrospective, dataLoader)
     sendMeetingEndToSegment(completedRetrospective, meetingMembers as MeetingMember[], template)
     sendNewMeetingSummary(completedRetrospective, context).catch(console.log)
-    const events = meetingMembers.map(
+    const events = teamMembers.map(
```
mattkrick, line 128: next line should call it a teamMember

mattkrick (commented):
2 little nits, then LGTM! can you also test it to make sure that someone who was _not_ in the meeting sees the event in their timeline & when then click it, they see the summary but they are not added as part of the meeting?

tianrunhe:
I can confirm that clicking an event in the timeline will not add the user as part of the meeting. However found a new issue here: user who were not part of the meeting cannot comment to the discussion thread:



Not sure if we wanna fix this too in this PR

mattkrick:
great find! let's save that for another PR, i'm excited to get this one out to prod

PR #4686, opened 2021-02-18
Fix the bug where reflect prompts and dimensions cannot be deleted

tianrunhe:
Fixes support issue reported [here](https://parabol.slack.com/archives/C4JAUUZ9P/p1613668617087100). Notice that the user reported that he could not rename a reflect but that was actually a side effect of not being able to remove it from front end. Hence the "Prompt not found" error. I looked at in our DB, from server side, the prompt was removed already.

Tests:

- [ ] User can remove prompts from retro templates
- [ ] User can remove dimensions from poker templates
- [ ] User can rename prompts
- [ ] User can rename dimensions

### DIFF
--- a/packages/client/mutations/handlers/handleRemovePokerTemplateDimension.ts
+++ b/packages/client/mutations/handlers/handleRemovePokerTemplateDimension.ts
@@ -13,9 +13,9 @@ const handleRemovePokerTemplateDimension = (
     meetingType: MeetingTypeEnum.poker
   })
   if (!settings) return
-  const selectedTemplate = settings.getLinkedRecord('selectedTemplate')
-  if (!selectedTemplate) return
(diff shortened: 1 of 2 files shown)


ackernaut:
### Tests:

- [ ] User can remove prompts from retro templates
- [ ] User can remove dimensions from poker templates
- [ ] User can rename prompts
- [ ] User can rename dimensions

mattkrick:
CR looks good! did you happen to do a search across the codebase for other things that might still use `selected` instead of `active`?

ackernaut (approved):
All tests passed! Nice work, and a quick turn-around!

tianrunhe:
> CR looks good! did you happen to do a search across the codebase for other things that might still use `selected` instead of `active`?

Yup, I did exactly that! That's how I also found the dimensions from poker template having the same issue as prompts

PR #4696, opened 2021-02-22
Sort the reflection goups on server side after REFLECT phase

### LINKED ISSUE #4495: Sort reflections between Reflect and Group phase

tianrunhe:
Resolves #4495

Tests:

- [ ] When user navigates from reflect phase to group phase, the reflection cards are sorted within each column
- [ ] If user creates reflection groups, the column won't sort again
- [ ] When user ungroups, the left-alone card stays close with the old group
- [ ] Expanding sub-columns won't break any of the above-mentioned behaviors

### DIFF
--- a/packages/client/mutations/NavigateMeetingMutation.ts
+++ b/packages/client/mutations/NavigateMeetingMutation.ts
@@ -6,13 +6,14 @@ import Atmosphere from '../Atmosphere'
 import {ClientRetrospectiveMeeting} from '../types/clientSchema'
 import {IReflectPhase} from '../types/graphql'
 import {SharedUpdater} from '../types/relayMutations'
-import {VOTE} from '../utils/constants'
+import {AUTO_GROUPING_THRESHOLD, REFLECT, VOTE} from '../utils/constants'
 import isInterruptingChickenPhase from '../utils/isInterruptingChickenPhase'
 import isViewerTyping from '../utils/isViewerTyping'
 import getBaseRecord from '../utils/relay/getBaseRecord'
 import getInProxy from '../utils/relay/getInProxy'
 import safeProxy from '../utils/relay/safeProxy'
 import {setLocalStageAndPhase} from '../utils/relay/updateLocalStage'
+import groupReflections from '../utils/smartGroup/groupReflections'
 import {
   NavigateMeetingMutation as TNavigateMeetingMutation,
   NavigateMeetingMutationVariables
@@ -33,6 +34,9 @@ graphql`
     phaseComplete {
       reflect {
         emptyReflectionGroupIds
+        reflectionGroups {
+          sortOrder
+        }
       }
       group {
         emptyReflectionGroupIds
@@ -142,6 +146,11 @@ export const navigateMeetingTeamUpdater: SharedUpdater<NavigateMeetingMutation_t
       reflectPrompt?.setValue([], 'editorIds')
     })
   }
+
+  const reflectionGroups = meeting.getLinkedRecords('reflectionGroups')
+  const sortedReflectionGroups = reflectionGroups.sort((a, b) =>
+    a.getValue('sortOrder') < b.getValue('sortOrder') ? -1 : 1)
+  meeting.setLinkedRecords(sortedReflectionGroups, 'reflectionGroups')
 }
 
 const NavigateMeetingMutation = (
@@ -182,6 +191,38 @@ const NavigateMeetingMutation = (
             if (completedStageId) {
               setLocalStageAndPhase(store, meetingId, completedStageId)
             }
+          }
+          if (phaseType === REFLECT) {
+            const reflectionGroups = meeting.getLinkedRecords('reflectionGroups')
+            if (!reflectionGroups) return
+            const reflections = reflectionGroups.reduce((filteredReflections, reflectionGroup) => {
+              const reflections = reflectionGroup.getLinkedRecords('reflections')
+              if (reflections && reflections.length !== 0) {
+                const entitiesRecords = reflections[0].getLinkedRecords('entities')
+                if (!entitiesRecords) return filteredReflections
+                const entities = entitiesRecords.map((entitiesRecord) => {
+                  return {
+                    lemma: entitiesRecord.getValue('lemma'),
+                    name: entitiesRecord.getValue('name'),
+                    salience: entitiesRecord.getValue('salience')
+                  }
+                })
+                const reflection = {
+                  entities: entities,
+                  reflectionGroupId: reflections[0].getValue('reflectionGroupId')
+                }
+                filteredReflections.push(reflection)
+              }
+              return filteredReflections
+            }, [] as any[])
+            const {reflectionGroupMapping} = groupReflections(reflections, AUTO_GROUPING_THRESHOLD)
+
+            const sortedReflectionGroups = reflectionGroups.sort((groupA, groupB) => {
+              const newGroupIdA = reflectionGroupMapping[groupA.getDataID()]
+              const newGroupIdB = reflectionGroupMapping[groupB.getDataID()]
+              return newGroupIdA.localeCompare(newGroupIdB)
+            })
+            meeting.setLinkedRecords(sortedReflectionGroups, 'reflectionGroups')
           }
         }
       }
--- a/packages/client/mutations/NavigateMeetingMutation.ts
+++ b/packages/client/mutations/NavigateMeetingMutation.ts
@@ -148,7 +148,7 @@ export const navigateMeetingTeamUpdater: SharedUpdater<NavigateMeetingMutation_t
   }
 
   const reflectionGroups = meeting.getLinkedRecords('reflectionGroups')
-  const sortedReflectionGroups = reflectionGroups.sort((a, b) =>
+  const sortedReflectionGroups = reflectionGroups.slice().sort((a, b) =>
     a.getValue('sortOrder') < b.getValue('sortOrder') ? -1 : 1)
   meeting.setLinkedRecords(sortedReflectionGroups, 'reflectionGroups')
 }
--- a/packages/client/mutations/NavigateMeetingMutation.ts
+++ b/packages/client/mutations/NavigateMeetingMutation.ts
@@ -6,14 +6,13 @@ import Atmosphere from '../Atmosphere'
 import {ClientRetrospectiveMeeting} from '../types/clientSchema'
 import {IReflectPhase} from '../types/graphql'
 import {SharedUpdater} from '../types/relayMutations'
-import {AUTO_GROUPING_THRESHOLD, REFLECT, VOTE} from '../utils/constants'
+import {REFLECT, VOTE} from '../utils/constants'
 import isInterruptingChickenPhase from '../utils/isInterruptingChickenPhase'
 import isViewerTyping from '../utils/isViewerTyping'
 import getBaseRecord from '../utils/relay/getBaseRecord'
 import getInProxy from '../utils/relay/getInProxy'
 import safeProxy from '../utils/relay/safeProxy'
 import {setLocalStageAndPhase} from '../utils/relay/updateLocalStage'
-import groupReflections from '../utils/smartGroup/groupReflections'
 import {
   NavigateMeetingMutation as TNavigateMeetingMutation,
   NavigateMeetingMutationVariables
@@ -148,8 +147,9 @@ export const navigateMeetingTeamUpdater: SharedUpdater<NavigateMeetingMutation_t
   }
 
   const reflectionGroups = meeting.getLinkedRecords('reflectionGroups')
-  const sortedReflectionGroups = reflectionGroups.slice().sort((a, b) =>
-    a.getValue('sortOrder') < b.getValue('sortOrder') ? -1 : 1)
+  const sortedReflectionGroups = reflectionGroups
+    .slice()
+    .sort((a, b) => (a.getValue('sortOrder') < b.getValue('sortOrder') ? -1 : 1))
   meeting.setLinkedRecords(sortedReflectionGroups, 'reflectionGroups')
 }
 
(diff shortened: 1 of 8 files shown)


packages/client/mutations/NavigateMeetingMutation.ts
```diff
@@ -142,6 +146,11 @@ export const navigateMeetingTeamUpdater: SharedUpdater<NavigateMeetingMutation_t
       reflectPrompt?.setValue([], 'editorIds')
     })
   }
+
+  const reflectionGroups = meeting.getLinkedRecords('reflectionGroups')
+  const sortedReflectionGroups = reflectionGroups.sort((a, b) =>
```
mattkrick, line 151: `reflectionGroups.slice().sort`, sort mutates the current array instead of doing a shallow clone.
relay requires a new object when you call `setLinkedRecords` because internally it does shallow object equality comparisons

packages/client/mutations/NavigateMeetingMutation.ts
```diff
@@ -183,6 +192,38 @@ const NavigateMeetingMutation = (
               setLocalStageAndPhase(store, meetingId, completedStageId)
             }
           }
+          if (phaseType === REFLECT) {
```
mattkrick, line 195: whoaaaa, double duty!
why not just wait for the sort order to come back from the server?

packages/server/graphql/mutations/helpers/handleCompletedStage.ts
```diff
@@ -16,7 +24,47 @@ const handleCompletedRetrospectiveStage = async (
   dataLoader: DataLoaderWorker
 ) => {
   if (stage.phaseType === REFLECT || stage.phaseType === GROUP) {
-    const data = await removeEmptyReflections(meeting)
+    const data: Record<string, any> = await removeEmptyReflections(meeting)
+
+    if (stage.phaseType === REFLECT) {
+      const r = await getRethink()
+      const now = new Date()
+      const reflectionGroups = await dataLoader
```
mattkrick, line 32: nit: can combine these 2 in a `Promise.all` so we can return faster

packages/server/graphql/mutations/helpers/handleCompletedStage.ts
```diff
@@ -16,7 +24,47 @@ const handleCompletedRetrospectiveStage = async (
...
+        .table('RetroReflection')
+        .getAll(meeting.id, {index: 'meetingId'})
+        .filter({isActive: true})
+        .orderBy('createdAt')
+        .run()
+      const {reflectionGroupMapping, autoGroupThreshold, nextThresh} = groupReflections(
+        reflections,
+        AUTO_GROUPING_THRESHOLD
+      )
+      console.log(`autoGroupThreshold = ${autoGroupThreshold}; nextThresh = ${nextThresh}`)
```
mattkrick, line 45: remove log

packages/server/graphql/mutations/helpers/handleCompletedStage.ts
```diff
@@ -16,7 +24,47 @@ const handleCompletedRetrospectiveStage = async (
...
+
+      reflectionGroups.sort((groupA, groupB) => {
```
mattkrick, line 47: slice first! here you're mutating an array provided by a dataloader. that means ANY other function that is using that same dataloader is going to get mutated & depending on when this runs, that means it'll be nondeterministic

packages/server/graphql/mutations/helpers/handleCompletedStage.ts
```diff
@@ -16,7 +24,47 @@ const handleCompletedRetrospectiveStage = async (
...
+        const newGroupIdA = reflectionGroupMapping[groupA.id]
+        const newGroupIdB = reflectionGroupMapping[groupB.id]
+        return newGroupIdA.localeCompare(newGroupIdB)
```
mattkrick, line 50: why `localeCompare` instead of just a < b ? -1 : 1?

packages/server/graphql/mutations/helpers/handleCompletedStage.ts
```diff
@@ -16,7 +24,47 @@ const handleCompletedRetrospectiveStage = async (
...
+      })
+
+      reflectionGroups.forEach(async (group, index) => {
+        group.sortOrder = index
+        await r
```
mattkrick, line 55: don't use async/await in a `forEach`, it doesn't act how you think it acts!

instead, try `await Promise.all(reflectionGroups.map(() => return r.table(...))`

packages/server/graphql/mutations/helpers/updateReflectionLocation/removeReflectionFromGroup.ts
```diff
@@ -12,10 +12,22 @@ const removeReflectionFromGroup = async (reflectionId, {dataLoader}) => {
...
   if (!reflection) throw new Error('Reflection not found')
   const {reflectionGroupId: oldReflectionGroupId, meetingId, promptId} = reflection
+  const oldReflectionGroup = await r
+    .table('RetroReflectionGroup')
+    .get(oldReflectionGroupId)
+    .run()
+  const {sortOrder: oldSortOrder} = oldReflectionGroup
+  const maxSortOrder = await r
+    .table('RetroReflectionGroup')
+    .getAll("1kHnE9uN5C", {index: 'meetingId'})('sortOrder')
```
mattkrick, line 22: this ain't good

packages/server/graphql/mutations/helpers/updateReflectionLocation/removeReflectionFromGroup.ts
```diff
@@ -12,10 +12,22 @@ const removeReflectionFromGroup = async (reflectionId, {dataLoader}) => {
...
+    .max()
+    .default(0)
+    .run()
   const meeting = await dataLoader.get('newMeetings').load(meetingId)
 
   // RESOLUTION
-  const reflectionGroup = new ReflectionGroup({meetingId, promptId})
+  const newSortOrder = oldSortOrder === maxSortOrder ? oldSortOrder + 1 : (oldSortOrder + (oldSortOrder + 1)) / 2
```
mattkrick, line 29: that last piece is just `oldSortOrder + .5`. that doesn't work because what if the next 5 reflection groups all have sort orders less than .5 greater?
you need to grab the sort order of the next item

mattkrick (changes requested):

packages/client/mutations/NavigateMeetingMutation.ts
tianrunhe: I saw flickering without this part in optimistic updater

mattkrick: hmmm, lemme see if i can find a way around that.
my concern is that we're requiring a whole lot more code (the clustering algo, etc), all the reflection entities, and performing a whole lot of extra work & it only benefits the facilitator for a split second.

mattkrick: @tianrunhe check out L190 above. for the vote phase, we undo changing the facilitatorStageId. that avoids the flicker!

packages/server/graphql/mutations/helpers/updateReflectionLocation/removeReflectionFromGroup.ts
```diff
@@ -12,10 +12,30 @@ const removeReflectionFromGroup = async (reflectionId, {dataLoader}) => {
     .run()
   if (!reflection) throw new Error('Reflection not found')
   const {reflectionGroupId: oldReflectionGroupId, meetingId, promptId} = reflection
+  const oldReflectionGroup = await r
```
mattkrick, line 15: is it safe to use a dataloader to get?
can we combine the next 3 async calls?

packages/server/graphql/mutations/helpers/updateReflectionLocation/removeReflectionFromGroup.ts
```diff
@@ -12,10 +12,30 @@ const removeReflectionFromGroup = async (reflectionId, {dataLoader}) => {
...
+    .table('RetroReflectionGroup')
+    .getAll(meetingId, {index: 'meetingId'})
+    .filter({isActive: true})
+    .orderBy('sortOrder')
+    .run()
   const meeting = await dataLoader.get('newMeetings').load(meetingId)
 
   // RESOLUTION
-  const reflectionGroup = new ReflectionGroup({meetingId, promptId})
+  const nextReflectionGroupIndex =
```
mattkrick, line 29: if the client lags, the oldReflectionGroup may already be inactive, which would result in `-1 + 1`, which would put this at the top (if the column has at least 1 card in it) instead of the bottom

packages/server/graphql/mutations/helpers/updateReflectionLocation/removeReflectionFromGroup.ts
```diff
@@ -12,10 +12,30 @@ const removeReflectionFromGroup = async (reflectionId, {dataLoader}) => {
...
+    reflectionGroups.findIndex(({id}) => id === oldReflectionGroup.id) + 1
+  let newSortOrder
+  if (nextReflectionGroupIndex >= reflectionGroups.length) {
+    newSortOrder = oldSortOrder + 1
+  } else {
+    const nextSortOrder = reflectionGroups[nextReflectionGroupIndex].sortOrder
```
mattkrick, line 35: reflectionGroups[nextReflectionGroupIndex] is not guaranteed. what if no cards are in the column?

packages/server/graphql/mutations/helpers/updateReflectionLocation/removeReflectionFromGroup.ts
```diff
@@ -12,10 +12,30 @@ const removeReflectionFromGroup = async (reflectionId, {dataLoader}) => {
...
+    .run()
   const meeting = await dataLoader.get('newMeetings').load(meetingId)
 
   // RESOLUTION
-  const reflectionGroup = new ReflectionGroup({meetingId, promptId})
+  const nextReflectionGroupIndex =
+    reflectionGroups.findIndex(({id}) => id === oldReflectionGroup.id) + 1
+  let newSortOrder
+  if (nextReflectionGroupIndex >= reflectionGroups.length) {
+    newSortOrder = oldSortOrder + 1
```
mattkrick, line 33: needs noise. it's possible this function gets called by 2 different folks at the (almost) exact moment.

packages/server/graphql/mutations/helpers/updateReflectionLocation/removeReflectionFromGroup.ts
```diff
@@ -12,10 +12,30 @@ const removeReflectionFromGroup = async (reflectionId, {dataLoader}) => {
...
+  } else {
+    const nextSortOrder = reflectionGroups[nextReflectionGroupIndex].sortOrder
+    newSortOrder = (oldSortOrder + nextSortOrder) / 2
```
mattkrick, line 36: needs noise, 2 people could call this at the same time

packages/server/graphql/mutations/helpers/updateReflectionLocation/removeReflectionFromGroup.ts
```diff
@@ -12,10 +12,30 @@ const removeReflectionFromGroup = async (reflectionId, {dataLoader}) => {
...
+  }
+  const reflectionGroup = new ReflectionGroup({meetingId, promptId, sortOrder: newSortOrder})
   await r
```
mattkrick, line 39: could we combine a few of these await clauses? less awaiting means less of a chance of race conditions

packages/server/graphql/mutations/helpers/updateReflectionLocation/removeReflectionFromGroup.ts
tianrunhe: I don't quite follow on this one. Can you elaborate more? Appreciate it!

packages/server/graphql/mutations/helpers/updateReflectionLocation/removeReflectionFromGroup.ts
tianrunhe: If there are no cards, `reflectionGroups.length === 0` and `nextReflectionGroupIndex === 1` so it will execute the if block, right?

packages/server/graphql/mutations/helpers/updateReflectionLocation/removeReflectionFromGroup.ts
mattkrick: `findIndex` returns -1 if no match is found.
that's possible because `oldReflectionGroupId` could refer to a reflection group that has already gone away

packages/server/graphql/mutations/helpers/updateReflectionLocation/removeReflectionFromGroup.ts
mattkrick: yep, you're right!

mattkrick: for things like this, it's often nicer to do conditional access, that way in a couple weeks when we forget the business logic, we can quickly rule out an undefined error.
e.g. `reflectionGroups[next]?.sortOrder ?? 0`. when i read that single line, i don't have to re-read L29-32 & i know typescript will protect me for the rest of it.

mattkrick:
### Business logic:

- When I ungroup a card from a group in column X, if that card belongs in column X, it should go directly beneath its old group in column X
- When I ungroup a card from a group in column X, if that card belongs in column Y, it should go to the bottom of column Y

```ts
const {promptId} = reflection
const cardsInColumn = await r.table('RetroReflectionGroup').getAll(meetingId, {index: 'meetingId'}).filter({isActive: true, promptId})
let sortOrder = 1e6
if (reflection.promptId === oldReflectionGroup.promptId) {
  // case 1
  const oldGroupIdx = cardsInColumn.findIndex(group => group.id === oldReflectionGroup.id)
  if (oldGroupIdx === -1) {
    // should never happen, but we play it safe
    cardsInColumn[cardsInColumn.length - 1]?.sortOrder + 1 + dndNoise() ?? 1e6
  } else if (oldGroupIdx === cardsInColumn.length - 1) {
    // it's at the bottom!
    sortOrder = cardsInColumn[cardsInColumn.length - 1].sortOrder + 1 + dndNoise()
  } else {
    // stick it directly after the old group
    const oldSortOrder = cardsInColumn[oldGroupIdx].sortOrder
    const afterSortOrder = cardsInColumnn[oldGroupIdx + 1].sortOrder
    sortOrder = (oldSortOrder + afterSortOrder) / 2 + dndNoise()
  }
} else {
   // case 2
   sortOrder = cardsInColumn[cardsInColumn.length - 1]?.sortOrder + 1 + dndNoise() ?? 1e6
}  
```

Issue #4652, opened 2021-02-05
Meeting avatars are populated from MeetingMembers


mattkrick:
We need to pull meeting member avatars from meeting members instead of team members. this will be less fetching & less work for really large teams that only have a few members present in a meeting.

AC
- [ ] In a meeting, the meeting avatars are sourced from the meetingMembers instead of TeamMembers

Estimate: 4 hours

PR #4697, opened 2021-02-23
Meeting avatars are populated from MeetingMembers

nickoferrall:
PR to resolve issue #4652


packages/client/modules/meeting/components/MeetingAvatarGroup/NewMeetingAvatarGroup.tsx
```diff
@@ -185,15 +186,21 @@ export default createFragmentContainer(NewMeetingAvatarGroup, {
       team {
         id
         teamMembers {
```
nickoferrall, line 188: Passing `teamMembers` down to `AddTeamMemberAvatarButton` and then to `AddTeamMemberModal` as we want to check whether the user is inviting a new team member rather than a new meeting member

mattkrick:
nice! did this happen to get rid of that delayed flickering you mentioned?

nickoferrall:
It did indeed!

Issue #4028, opened 2020-07-27
Security issue #46


jordanh:
## Issue - Bug

See: https://github.com/ParabolInc/action-devops/issues/46

Lowish priority, but should probably fix. 2x reports now.

- **Estimated effort:** 4 hours ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

jordanh:
Still a thing. Keeping this on the backlog

PR #46, opened 2016-04-16
Eslint Refactor

jrwells:
Largely the same as the first 'linting' pull request.

A (relatively) major refactor making our source ESLint compliant based on our custom rule set (derived mainly from [airbnb's rules](https://github.com/airbnb/javascript)).

Things done:

1. ESLint refactor
2. Update ESLint to be ESLint 2.0+ compliant
3. Update npm scripts
4. Remove a single debugger statement

⛵

PR #4698, opened 2021-02-23
Fix security issue #46

nickoferrall:
PR to fix issue #4028 

This PR also:
- Fixes a bug where the `AddOrgMutation` wasn't handling errors in the `onComplete` function
- Fixes several bugs where error objects were being used in the jsx instead of a string
- Refactored a few class components to functional components 

I'll leave the testing criteria out of this public PR description as it's a security issue.

### DIFF
--- a/packages/client/modules/newTeam/components/NewTeamForm/NewTeamForm.tsx
+++ b/packages/client/modules/newTeam/components/NewTeamForm/NewTeamForm.tsx
@@ -1,20 +1,15 @@
 import {NewTeamForm_organizations} from '../../../../__generated__/NewTeamForm_organizations.graphql'
-import React, {useEffect, useState, ChangeEvent, FormEvent, FocusEvent} from 'react'
+import React, {useState, ChangeEvent, FormEvent} from 'react'
 import styled from '@emotion/styled'
 import {createFragmentContainer} from 'react-relay'
 import graphql from 'babel-plugin-relay/macro'
-import {RouteComponentProps, withRouter} from 'react-router-dom'
 import FieldLabel from '../../../../components/FieldLabel/FieldLabel'
 import Panel from '../../../../components/Panel/Panel'
 import PrimaryButton from '../../../../components/PrimaryButton'
 import Radio from '../../../../components/Radio/Radio'
-import withAtmosphere, {
-  WithAtmosphereProps
-} from '../../../../decorators/withAtmosphere/withAtmosphere'
 import NewTeamOrgPicker from '../../../team/components/NewTeamOrgPicker'
 import AddOrgMutation from '../../../../mutations/AddOrgMutation'
 import AddTeamMutation from '../../../../mutations/AddTeamMutation'
-import withMutationProps, {WithMutationProps} from '../../../../utils/relay/withMutationProps'
 import Legitity from '../../../../validation/Legitity'
 import teamNameValidation from '../../../../validation/teamNameValidation'
 import NewTeamFormBlock from './NewTeamFormBlock'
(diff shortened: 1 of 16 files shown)


packages/server/graphql/mutations/addTeam.ts
```diff
@@ -45,7 +45,10 @@ export default {
...
         .run()
 
-      const orgTeamNames = orgTeams.map((team) => !team.isArchived && team.name)
+      const orgTeamNames = [] as string[]
+      orgTeams.forEach((team) => {
+        if (!team.isArchived) orgTeamNames.push(team.name)
+      })
       const {
         data: {newTeam},
         errors
```
nickoferrall, line 54: Why is `addTeamValidation` a thunk out of interest?

packages/client/modules/newTeam/components/NewTeamForm/NewTeamForm.tsx
```diff
@@ -88,53 +86,33 @@ interface Fields {
...
-class NewTeamForm extends Component<Props, State> {
-  state = {
-    isNewOrg: this.props.isNewOrganization,
-    orgId: '',
-    fields: {
-      orgName: {...DEFAULT_FIELD},
-      teamName: {...DEFAULT_FIELD}
-    }
-  }
+const NewTeamForm = (props: Props) => {
```
mattkrick, line 89: nice refactor!

packages/client/modules/newTeam/components/NewTeamForm/NewTeamForm.tsx
```diff
@@ -88,53 +86,33 @@ interface Fields {
...
+  const {isNewOrganization, organizations} = props
```
mattkrick, line 90: nit: maybe call this `isInitiallyNewOrg`? something so we don't confuse it with what we're using for state

packages/client/modules/newTeam/components/NewTeamForm/NewTeamForm.tsx
```diff
@@ -146,96 +124,88 @@ class NewTeamForm extends Component<Props, State> {
...
-      .max(100, 'Maybe just the legal name?')
+  const validate = (name: FieldName) => {
+    const validators = {
+      teamName: validateTeamName,
+      orgName: validateOrgName
+    }
+    const res: Legitity = validators[name]()
+    const field = fields[name]
+    if (res.error !== field.error) {
+      setFields({
```
mattkrick, line 135: nit: we've got a nice hook called `useForm` that takes care of all this boilerplate logic for you if you'd like to use that

packages/server/graphql/mutations/addTeam.ts
mattkrick: me from years ago was a terrible programmer

mattkrick (commented):
looks really good!
only big suggestion is to try using `useForm` & see if that cleans up some of the boilerplate of using a form.

forms are so hard & it seems like everyone loves building their own form packages. thankfully, it's super rare for us to have large forms. i think this is our second largest form in the whole app!

let me know if you find `useForm` helpful, we could still make some fixes to it to solve our simple use cases. 

or if time is tight just let me know & we can push this through without. your call!

nickoferrall:
Refactored to use `useForm` - it was very soothing removing that code!

mattkrick:
linkify is a great tool because it very aggressively makes things links.
for example,  `foo.goo.com` and `project: github.com`
i wonder if this is too aggressive? 

also, it looks like you found a bug!
type `goo.co` character by character. notice that it doesn't throw an error. that's because `validate`  is referencing a stale value (onChange only dispatches a change to update the value, the value isn't updated until the next cycle).

suggestions:
- L112 in useForm should return `{error: undefined, value}`
- validateOrgName & validateTeamName should use the supplied value instead of grabbing the value from the stale state (e.g. in L77)

nickoferrall:
> i wonder if this is too aggressive?

I think we should throw an error for both `foo.goo.com` and `project: github.com` as they both appear as hyperlinks in invitation emails which could be a security risk.

I've implemented your suggestions and I've also removed the `dirty` check so that we show the error message as the user types rather than `onBlur`.

mattkrick:
I'm cool with that! LGTM

PR #4700, opened 2021-02-24
Generate a new id for user who has signed in with Google and deleted their account before

### LINKED ISSUE #4654: google oauth deleted account recreation email missing
## Issue - Bug (non critical)

when deleting a user then creating a new account, locally created accounts get recreated with a new row and id. but when using google auth, the newly created account reuses the same db row and id and the email stays "DELETED".

Screen Shot 2021-02-08 at 12 33 30 PM

Users can:
- sign up with google oauth, delete their account, sign up again with google oauth and see their email populated when they sign in

- **Estimated effort:** 2 hours ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

tianrunhe:
Fixes #4654 .

### DIFF
--- a/packages/server/graphql/mutations/loginWithGoogle.ts
+++ b/packages/server/graphql/mutations/loginWithGoogle.ts
@@ -101,14 +101,14 @@ const loginWithGoogle = {
         isEmailVerified: email_verified !== 'false'
       })
       const newUser = new User({
+        // if a deleted user with the same id exists, generate a new id
+        id: deletedUser?.isRemoved ? `google-oauth2|${generateUID()}` : userId,
         preferredName,
         picture,
         email,
         identities: [identity],
         segmentId
       })
-      // if a deleted user with the same id exists, generate a new id
-      newUser.id = deletedUser && deletedUser.isRemoved ? `google-oauth2|${generateUID()}` : userId
       context.authToken = await bootstrapNewUser(newUser, !invitationToken)
       return {
         userId,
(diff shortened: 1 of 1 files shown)


packages/server/graphql/mutations/loginWithGoogle.ts
```diff
@@ -86,22 +87,28 @@ const loginWithGoogle = {
...
       // it's a new user!
       const nickname = name || email.substring(0, email.indexOf('@'))
       const preferredName = nickname.length === 1 ? nickname.repeat(2) : nickname
-      const userId = `google-oauth2|${sub}`
       const identity = new AuthIdentityGoogle({
         id: sub,
         isEmailVerified: email_verified !== 'false'
       })
       const newUser = new User({
-        id: userId,
```
mattkrick, line 98: let's do the work up here so we can avoid mutating the object down below.
that way future us won't have to know that `newUser.id` is invalid when the object is instantiated
 
```
id: deletedUser?.isRemoved ? `google-oauth2|${generateUID()}` : userId
```

packages/server/graphql/mutations/loginWithGoogle.ts
```diff
@@ -86,22 +87,28 @@ const loginWithGoogle = {
...
         preferredName,
         picture,
         email,
         identities: [identity],
         segmentId
       })
+      // if a deleted user with the same id exists, generate a new id
+      newUser.id = deletedUser && deletedUser.isRemoved ? `google-oauth2|${generateUID()}` : userId
```
mattkrick, line 111: nit: instantiate the newUser object with the correct ID instead of mutating the object down below

mattkrick:
LGTM!

PR #4739, opened 2021-03-03
Fix broken navigation

nickoferrall:
`reflectionGroups` is `null` in non-retro meetings


nickoferrall:
I'll work on refactoring some of the logic in NavigateMeetingMutation as a lot of the code is retro focused and doesn't need to be run in non-retro meetings. Have pushed this temporary fix to master to help with the ship 🚢

PR #4743, opened 2021-03-04
Prevent Poker drawer from jumping up

nickoferrall:
Poker drawer now behaves in the same way as the left sidebar, i.e. when the meeting control bar is dragged on top of it, it doesn't jump up


packages/client/hooks/useControlBarCovers.ts
```diff
@@ -42,6 +42,7 @@ export const useCoverable = (
   parentRef?: RefObject<HTMLDivElement>,
   columnsRef?: RefObject<HTMLDivElement>
 ) => {
+  if (id === 'poker-thread') return true
```
mattkrick, line 45: hooks must always run in the same order every single time.
if we conditionally return early, that isn't true (i mean, with our current business logic it is because you can't get from 1 meeting type to another without unmounting this component, but as a best practice, we should follow the rules of hooks).


 

the magic string feels a little brittle (when i call this hook, i don't know that the string has to match a certain template).
i'd suggest the hook takes a `meetingType` variable to be a little more explicit.

packages/client/hooks/useControlBarCovers.ts
```diff
@@ -40,9 +41,11 @@ export const useCoverable = (
   ref: RefObject<HTMLDivElement>,
   height: number,
   parentRef?: RefObject<HTMLDivElement>,
-  columnsRef?: RefObject<HTMLDivElement>
+  columnsRef?: RefObject<HTMLDivElement>,
+  meetingType?: MeetingTypeEnum
 ) => {
   const updateCoverables = () => {
+    if (meetingType === 'poker') return
```
mattkrick, line 48: can remove?

nickoferrall: No, we need both `meetingType` checks here. Without this line, the ref's height will be adjusted and without line 84, `isExpanded` will be false meaning the height of the thread wrapper won't be 100%

mattkrick: ahh gotcha LGTM!

Issue #4543, opened 2021-01-19
Unable to upload images >50 KB


nickoferrall:
When uploading a profile photo JPEG image greater than 50 KB, I get the error `File is too large (1MB Max)`.

### AC

- [ ] Able to upload images up to 1MB in size
- [ ] `File is too large (1MB Max)` error only displays when uploading images greater than 1MB

PR #4746, opened 2021-03-05
Fix user img upload

nickoferrall:
PR to resolve issue #4543

jpgWithoutEXIF would increase the size of jpgs by ~10x. As we're not removing the EXIF information from the user org images, I assume it's safe to remove.

Questions:

1. Why did we want to remove the EXIF info from user images?
2. For user imgs, we convert svgs to pngs. I'd love to know why and whether not doing this in the org image upload was intentional?

### AC

- [ ]  Able to upload images up to 1MB in size
- [ ]  File is too large (1MB Max) error only displays when uploading images greater than 1MB


mattkrick:
exif data is a security concern. we (should) be removing exif from all jpgs. reason being is the metadata could contain personal info. for example, the creator name, when it was created, etc. if someone stumbles across that file, they shouldn't get info about a person.

if writing to the canvas is causing the file to bloat, then we need to figure out a way to recompress it or stripping it without writing to canvas. i'm sure this is a solved problem as just about every company that accepts jpg uploads does it.

svgs are not displayed in emails because of security implications, so we creator a rastor of it so the emailed meeting summary can show the avatar correctly. i suppose we should do the same for org avatars if we ever decide to email invoices.

nickoferrall:
Makes sense! Thanks for the explanation.

When creating the [blob object](https://developer.mozilla.org/en-US/docs/Web/API/HTMLCanvasElement/toBlob), the jpeg was being converted to a png. There was still some bloat when specifying the image type as a jpeg so I've compressed it to 75% quality. I've also added the same security logic to the org avatars.

packages/client/utils/jpgWithoutEXIF.ts
```diff
@@ -7,9 +7,13 @@ const jpgWithoutEXIF = (file: File) => {
...
       context.drawImage(img, 0, 0, canvas.width, canvas.height)
-      canvas.toBlob((blob) => {
-        resolve(blob!)
-      })
+      canvas.toBlob(
+        (blob) => {
+          resolve(blob!)
+        },
+        'image/jpeg',
+        0.75
```
mattkrick, line 15: ahhh that's the ticket!
curious, did you compare the input img size vs. output image size after this change?

nickoferrall: Yep! Input vs output size of files around the 1mb limit would be very similar. Some of the smaller files would vary a bit more but still be pretty similar

Issue #4751, opened 2021-03-05
Provide feedback for flatMap bug


nickoferrall:
Issue #4490 mentions that users using old browsers (Chrome v68 or earlier) are unable to use `flatMap`. As outlined in discussion #4701, we should let users know why the app crashes when they encounter this bug and ask them to upgrade their browser.

### Acceptance Criteria

- [ ] When a user encounters the `flatMap` error, don't render the default error message
- [ ] Instead, explain why they encountered the bug & ask them to upgrade their browser
- [ ] Don't send the error message to Sentry as it's not a bug we intend to resolve

PR #4752, opened 2021-03-05
Provide feedback for old browser bugs

nickoferrall:
This PR resolves issue #4751 

Rather than just showing a message asking the user to upgrade, I thought it'd be helpful to provide a link. 

We could write a bunch of conditionals ([like so](https://stackoverflow.com/questions/9847580/how-to-detect-safari-chrome-ie-firefox-and-opera-browser)) to detect the browser and render a link based on the browser they're using. But browsers change, the conditionals may break and we would probably never know if Opera v8 is no longer showing the correct error message.

So I suggest that we show [this link](https://browser-update.org/update-browser.html) which explains why they should upgrade their browser & provides links to popular browsers. 

### Acceptance Criteria

- [ ]  When a user encounters the flatMap error, don't render the default error message
- [ ]  Instead, explain why they encountered the bug & ask them to upgrade their browser
- [ ]  Don't send the error message to Sentry as it's not a bug we intend to resolve

### DIFF
--- a/packages/client/components/ErrorComponent/ErrorComponent.tsx
+++ b/packages/client/components/ErrorComponent/ErrorComponent.tsx
@@ -27,6 +27,21 @@ const ErrorComponent = (props: Props) => {
   const {error, eventId} = props
   console.error(error)
   const {modalPortal, openPortal, closePortal} = useModal()
+  const oldBrowserErrs = ['flatMap is not a function']
+
+  const handleClick = () => {
+    const url = 'https://browser-update.org/update-browser.html'
+    window.open(url)
+  }
+  const isOldBrowserErr = oldBrowserErrs.find((err) => error.message.includes(err))
+  if (isOldBrowserErr) {
+    return (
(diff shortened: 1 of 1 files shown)


packages/client/components/ErrorComponent/ErrorComponent.tsx
```diff
@@ -27,6 +27,21 @@ const ErrorComponent = (props: Props) => {
   const {error, eventId} = props
   console.error(error)
   const {modalPortal, openPortal, closePortal} = useModal()
+  const oldBrowserErrs = ['flatMap is not a function']
+
+  const handleClick = () => {
+    const url = 'https://browser-update.org/update-browser.html'
+    window.open(url)
```
mattkrick, line 34: must use `_blank, noreferrer` for security reasons.

consider using `<a>` instead of onClick to be a little more declarative

mattkrick:
LGTM!

PR #4756, opened 2021-03-08
Migrate postgres team

### LINKED ISSUE #4629: Replicate Team to PG (Phase 1 of 3)
After User, Team is next to go to PG. Doing this will allow us to start caching teams, which will let us get rid of tms on the authToken #4071 

let's start to move to postgres. see #4530

AC
- [ ] Every time a Team is created, they are also created in postgres
- [ ] Every time a Team is updated, the update is propagated to postgres
- [ ] a postdeploy hook is written to copy all Team to postgres

Estimate: 16 hours

tiffanyhan:
Resolve #4629


packages/server/postgres/migrations/1614900793895_addTeamTable.ts
```diff
@@ -0,0 +1,38 @@
...
+      'action',
+      'retrospective',
+      'poker'
+    );
+
+    CREATE TABLE "Team" (
+      id VARCHAR(100) PRIMARY KEY,
+      name VARCHAR(100) NOT NULL,
+      "createdAt" TIMESTAMP NOT NULL,
+      "createdBy" VARCHAR(100) NOT NULL,
```
tiffanyhan, line 18: there are many orphaned teams in rethink. i.e. teams whose "createdBy" points to a nonexistent user in rethink. do we want to keep these records? i've taken the foreign key constraint off `createdBy` for now. the alternative is to keep the constraint but skip backfilling the orphaned records.

packages/server/database/migrations/20210307124052-prepareBackfillTeams.ts
```diff
@@ -0,0 +1,33 @@
+export const up = async function (r) {
+  try {
+    await Promise.all([
+      r.db('actionProduction')
+        .table('Team')
+        .indexCreate('createdAt')
```
mattkrick, line 6: are these indices purely for the postMigration?
Might be worth just doing `updatedAt` and then setting updatedAt === createdAt if it's null.
that way we only need to reference updatedAt

packages/server/database/migrations/20210307124052-prepareBackfillTeams.ts
```diff
@@ -0,0 +1,33 @@
...
+        .run(),
+      r.db('actionProduction')
+        .table('Team')
+        .indexCreate('updatedAt')
```
mattkrick, line 10: also need an indexWait call here otherwise the post migration could run before the index is ready (usually this isn't a problem, but we're making a back to back call on a pretty large table)

packages/server/postgres/migrations/1614900793895_addTeamTable.ts
tiffanyhan: hmmm i'll add the FK constraint later. i think the first pass should just be mirroring what we have in rethink. don't wanna get too bogged down with adding the extra data integrity when what we really need is the performance.

packages/server/database/migrations/20210307124052-prepareBackfillTeams.ts
tiffanyhan: i love this. also changed in user migration PR

packages/server/postgres/migrations/1614900793895_addTeamTable.ts
```diff
@@ -0,0 +1,38 @@
...
+    CREATE TYPE "MeetingTypeEnum" AS ENUM (
+      'action',
+      'retrospective',
+      'poker'
+    );
+
+    CREATE TABLE "Team" (
+      id VARCHAR(100) PRIMARY KEY,
+      name VARCHAR(100) NOT NULL,
+      "createdAt" TIMESTAMP NOT NULL,
```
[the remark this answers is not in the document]
tiffanyhan: also changed in user migration PR

mattkrick:
LGTM!

PR #4776, opened 2021-03-17
Palette refactor

ackernaut:
Our default deprecation pattern is to avoid touching all the files and only update to use the new ‘patterns’ as we move forward.

However, in the case of the color values, we would probably be haunted by them for a long time, and probably see more inconsistencies in the UI color usage defeating the point of the recent brand palette efforts.

This is a big one that touches most UI files.

- [x] Swaps all `paletteV2` for `paletteV3`
- [x] Cleans up miscellaneous color usage (e.g. Date Picker, other CSS files, etc.)

### Deferred
- Migrates all reflection prompt colors to new colors
- Updates default public poker scale templates to use new colors


ackernaut:
> hmmm, it's gonna iterate through that entire table 12 times, which might be a minute.
> something that may be a little safe:
> 
> 1. create a lookup table
>    const colors = {'#5CA0E5': PALETTE.SKY_500};
> 2. reference it
>    r.table('ReflectPrompt').update((row) => ({groupColor: r(colors)(row('groupColor')).default('#ffffff')}))

I tried it and it didn’t appear to work. I migrated back down, and then tried the row update query you wrote above. Is it missing something? The only thing I have left are these migrations. I’d like to get this PR merged before there are too many conflicts so I may separate out the migrations into separate tickets.

ackernaut:
I opened #4784 and #4783

PR #4777, opened 2021-03-17
first pass at GH integration frontend

mattkrick:
@nickoferrall here's a PR you can build the GH create FAB on. Feel free to tear apart any of the components, it was basically a copy & paste job


mattkrick:
gonna get this back into master to keep it from going stale again.
IMO, the coming soon for github isn't bad, but if we want we can completely remove the gh tab

Issue #4767, opened 2021-03-16
"Just started" snackbar should say "You" if you're the defaultFacilitator, not just the facilitator


mattkrick:
Repro
- someone else starts a retro
- you become facilitator
- you go back to dashboard
- it says "You just started". wtf no i didn't, don't put that evil on me

AC
- [ ] the person who started it is the starter

Estimate: 4 hours

PR #4780, opened 2021-03-17
"Just started" snackbar uses createdBy

nickoferrall:
PR to resolve issue #4767 

This PR:

- Adds `createdBy` and createdByUser` fields to `NewMeeting` so the snackbar can accurately state who started the meeting
- Removes `defaultFacilitatorUserId` and implements a migration 
- Removes unnecessary query fields in new meeting mutations
- Removes `useResumeFacilitation`

What's the purpose of `useResumeFacilitation`? It looks like it ensures that if the facilitator disconnects, a new tator is crowned and the original facilitator then rejoins, the original tator automatically takes the crown back. 

If that is its purpose, I think we could remove it and let the users decide. If the original tator leaves the meeting for 20 minutes and the meeting is going swimmingly with the new tator, it might feel a bit disruptive to auto change the tator again.

### AC:

- [ ] the person who started it is the starter


packages/server/database/migrations/20210317155500-swapTatorIdWithCreatedBy.ts
```diff
@@ -0,0 +1,33 @@
+import {R} from 'rethinkdb-ts'
+export const up = async function(r: R) {
+  const now = new Date()
+  try {
+    await r
```
mattkrick, line 5: we can combine both of these queries into 1 by using `replace, merge, without`. try it out!

packages/client/components/NewMeetingActionsCurrentMeetings.tsx
```diff
@@ -77,8 +75,8 @@ export default createFragmentContainer(NewMeetingActionsCurrentMeetings, {
         meetingType
         id
         createdAt
-        facilitator {
-          id
+        createdBy
+        createdByUser {
```
mattkrick, line 79: optionally, try using `readInlineData` in `useSnacksForNewMeetings` & then you can replace this stuff with a fragment spread. that way, the data is being used in the same file that it is requested in

(`readInlineData` is a new-ish function that lets you request things in functions outside of a component. it's kinda nice in cases like this

packages/client/mutations/StartRetrospectiveMutation.ts
```diff
@@ -7,18 +7,6 @@ graphql`
...
-      name
-      meetingMembers {
-        user {
-          id
-          preferredName
-        }
-      }
-    }
-    team {
-      ...TopBarMeetingsActiveMeetings @relay(mask: false)
```
mattkrick, line 20: if we remove this stuff, how do other team members know a meeting has started?

mattkrick (commented):
nice changes! i think the mutation stuff has to be added back in though.
i honestly can't remember if the resume facilitation was a customer request or not, if it feels bad i'm cool with taking it out

packages/client/mutations/StartRetrospectiveMutation.ts
nickoferrall: As we're just using the `meetingId` in this file, I didn't think we needed the other fields. The fragment is being used in the meeting subscription so I thought that when the `meetingId` updates, i.e. when the meeting is created, the team subscription updates and lets the other members know it's started

mattkrick: not sure i follow 100%, but i think the gotcha here is that the user isn't subscribed to a meeting until they enter the meeting, so if someone is on the dashboard, they only have the team subscription going

mattkrick:
LGTM! funny how these little things get so big, but you cleaned up a lot of old tech debt here!

Issue #4785, opened 2021-03-18
Story can be null in PokerEstimateHeaderCardParabol


nickoferrall:
Affected 48 users in 16 days

https://sentry.io/organizations/parabol/issues/2247547498/?referrer=github_plugin

```
TypeError: Cannot destructure property 'content' of 'l' as it is null.
  at c (./packages/client/components/PokerEstimateHeaderCardParabol.tsx:96:10)
...
(7 additional frame(s) were not displayed)
```

mattkrick:
looking here: https://github.com/ParabolInc/parabol/blob/774c4cf2dfbceaa2d56910583e325080c55deae0/packages/server/graphql/types/EstimateStage.ts#L154

if you create a task, add it to the poker meeting, then go back to the dashboard to delete the task (empty out the contents) then the task gets deleted, thus causing a null story.

i'll let you decide how to solve this :wink:

Issue #4708, opened 2021-02-25
TypeError: Cannot read property 'id' of undefined


tiffanyhan:
https://sentry.io/organizations/parabol/issues/1790265980/?referrer=github_plugin

```
TypeError: Cannot read property 'id' of undefined
  at None (./packages/client/hooks/useSnacksForNewMeetings.ts:22:84)
  at Array.filter (<anonymous>)
  at d (./packages/client/hooks/useSnacksForNewMeetings.ts:17:8)
...
(6 additional frame(s) were not displayed)
```

mattkrick:
@nickoferrall you can't map over a meetingRef, your fragment needs to be `(plural: true)`

nickoferrall:
@mattkrick it looks like there's a [known relay issue](https://github.com/facebook/relay/issues/3041) when using `(plural: true)` with `readInLineData` so I used the solution suggested in that issue. Perhaps I should add a comment in the code explaining why?

mattkrick:
ahh that's a bummer.
just pass in a team instead & then you can access `team { activeMeetings }`

nickoferrall:
Unfortunately, `useSnacksForNewMeetings` is being used in [TopBarMeetings](https://github.com/ParabolInc/parabol/blob/73f281fc23fd30966f67c464108c80e361ad59ea/packages/client/components/TopBarMeetings.tsx#L27) which uses `teams` and [NewMeetingActionsCurrentMeetings](https://github.com/ParabolInc/parabol/blob/60c789134d79a97bcfbb868895c4bec68816d017/packages/client/components/NewMeetingActionsCurrentMeetings.tsx#L46) which uses `team`. If we pass in `teams`, we'll have the same `plural` `readInLineData` issue.

Is the issue with mapping through `meetingRef` aesthetic of functional? The Sentry issue that Tiffany shared here occurred before the mapping through `meetingRef` made it to prod. 

An alternate option could be scrapping `readInLineData` and going back to using a regular Relay query in the components using the hook?

jordanh:
I can no longer find this in Sentry. Closing for now

PR #4786, opened 2021-03-18
Check if story and meeting exist

nickoferrall:
PR to resolve issues #4785 & #4708


mattkrick:
any idea about the root cause here?

nickoferrall:
For #4785, `Story` is nullable but we don't have a check to see if it exists (although the sidebar and control bar should prevent the user from advancing if there isn't as story). 

For #4708, I'm not too sure how an active meeting could be an empty array / not have an `id`

mattkrick:
is a buggy optimistic updater possibly at fault? would love to reproduce!

nickoferrall:
Interesting - could be, I'll take a look! Tomorrow morning I'll work on the GH issue but once that's ready or when I feel like switching things up, I'll dig into this!

mattkrick:
@nickoferrall any update on this? would love to understand the root cause before merging the fix

nickoferrall:
@mattkrick I haven't looked into this yet but I'll take a look now!

nickoferrall:
I haven't been able to repro this yet but here's a few things that I've learnt so far:

1. As you mentioned, this could be a buggy optimistic updater issue, [like this one](https://github.com/ParabolInc/parabol/commit/af6c7ea717852cf402b996481896be51932d5ab5). I suspected that [this is where the problem lies](https://github.com/ParabolInc/parabol/blob/f092a703bd675aaaa5129d3e155c6d4cb51474e2/packages/client/mutations/UpdatePokerScopeMutation.ts#L152) but we can remove `nextEstimateStage.setLinkedRecord(story, 'story')` and story will still exist so it may not be the problem.

2. The Sentry events for the undefined `Story` happen in `PokerEstimateHeaderCardParabol` and not `PokerEstimateHeaderCardJira` because we check whether `Story` exists in the Jira header [here](https://github.com/ParabolInc/parabol/blob/22328ab58233bb34fb25a162c19ae29a5bc08f74/packages/client/components/PokerEstimateHeaderCardJira.tsx#L92). We've probably been showing the incorrect error message that Jira is down when actually it might be the optimistic updater at fault. 

3. This issue first appeared in release 6.1.0. In that release, we had two large PRs: [Freeze templates for meeting durarion](https://github.com/ParabolInc/parabol/pull/4695/files) and [Robust Meeting Queue Protocal](https://github.com/ParabolInc/parabol/pull/4606/files). I haven't noticed any changes in these PRs, or the other smaller ones, that could have caused this bug. The Freeze Templates PR touches the `UpdatePokerScopeMutation` with `dimensionRef`s but that shouldn't affect `Story`.

I'll keep looking into the optimistic updater in `UpdatePokerScopeMutation`, consider other mutations that might be at fault and see if I can repro the nullable Story.

packages/client/components/PokerEstimateHeaderCardJira.tsx
```diff
@@ -83,8 +84,6 @@ const PokerEstimateHeaderCardJira = (props: Props) => {
   const {stage} = props
   const {serviceTaskId, story} = stage
   const [isExpanded, setIsExpanded] = useState(false)
-  const descriptionRef = useRef<HTMLDivElement>(null)
-  const maxHeight = Math.min(descriptionRef.current?.scrollHeight ?? 300, 300)
```
nickoferrall, line 87: If the user clicks expand, I think we can show 100% of the height of the card. Calculating height like this didn't show the full error description

packages/client/components/PokerEstimateHeaderCardParabol.tsx
```diff
@@ -87,31 +99,65 @@ const Content = styled('div')({
...
   const maxHeight = Math.min(descriptionRef.current?.scrollHeight ?? 300, 300)
   useEffect(
     () => () => {
       setIsExpanded(false)
     },
-    [taskId]
+    [story?.id]
   )
-  const {useTaskChild} = useTaskChildFocus(taskId)
+  const {useTaskChild} = useTaskChildFocus(story?.id || '')
```
nickoferrall, line 144: This feels ugly but if we check whether `story` exists beforehand and return the error card before `useEffect` or `useTaskChildFocus` are called, we won't be following the rules of hooks

mattkrick: i _really_ wish the rule of hooks didn't apply to returning `null`.
the least ugly pattern i've found is `const storyId = story?.id ?? ''` at the top of the component

nickoferrall: Makes sense! I've moved it to the top which feels slightly less ugly

packages/client/components/PokerEstimateHeaderCardParabol.tsx
```diff
@@ -87,17 +98,37 @@ const Content = styled('div')({
...
 }
 
-type Story = Required<NonNullable<NonNullable<PokerEstimateHeaderCardParabol_stage>['story']>>
-
 const PokerEstimateHeaderCardParabol = (props: Props) => {
   const {stage} = props
-  const {story} = stage
-  const {content, id: taskId, teamId} = story as Story
-  const integration = story!.integration
+  const story = stage.story as Extract<typeof stage['story'], {__typename: 'Task'}> | null
```
mattkrick, line 129: @nickoferrall i know we weren't happy with how we were doing this, so i found a neat little hack. if you request the `__typename` field explicitly in the fragment, then relay gives you a discriminated union. you can then `Extract` the type you want, which means nullability for the story fields will be preserved.

packages/client/hooks/useSnacksForNewMeetings.ts
```diff
@@ -32,7 +32,7 @@ const useSnacksForNewMeetings = (meetingsRef: any) => {
     const fiveSecsAgo = new Date(Date.now() - ms('5s'))
     const sortedMeetings = meetings
       .filter((meeting) => {
-        if (dismissedMeetingIds.has(meeting.id)) return false
+        if (dismissedMeetingIds.has(meeting?.id)) return false
```
mattkrick, line 35: the next line will throw if meeting is undefined

packages/client/components/PokerEstimateHeaderCardParabol.tsx
nickoferrall: Nice! That's good to know

packages/client/hooks/useSnacksForNewMeetings.ts
nickoferrall: Updated!

mattkrick:
Screenshot from 2021-03-31 08-20-46

when collapsed, i see the expand button as well as scrollbars.

I'd suggest:
- hide the scroll bars when collapsed
- show the scroll bars when expanded
- set a max height when expanded so the dimension is still above the fold

mattkrick:
great work! everything looks like it's working as intended!

Issue #4789, opened 2021-03-19
Address Slack feedback 


mattkrick:
@nickoferrall the below are the 2 pieces of feedback we got from Slack that is blocking them from re-approving our app. Could you take a look & address these?

If using the [App Home "Messages" tab](https://api.slack.com/surfaces/tabs), my bot responds to users’ direct messages. It also uses the app_home_opened event to send a welcome message the first time any user opens this view.

> The App Home "Messages" tab is in use but does not respond to users' direct messages, nor does it send a welcome message the first time a user opens it.

Permission Scopes

> We noticed you added groups:read to support private channels and wanted to check which method you're calling with that scope? If displaying a list of private channels, you'll need to make sure they're accessible to both the bot and user. You can do that with the users.conversations method, documented here: https://api.slack.com/methods/users.conversations
Additionally, in our testing we were not able to configure the staging app to post to a private channel or a DM at all, even after reinstalling. Could you please take a look at this?

PR #4792, opened 2021-03-19
Implement Slack feedback

nickoferrall:
PR to resolve issue #4789 

The issue included 3 pieces of feedback.

1. 
> The App Home "Messages" tab is in use but does not respond to users' direct messages, nor does it send a welcome message the first time a user opens it.

We're not sending a welcome message so it looks like we can remove this. I've flicked the "Messages Tab" switch in the staging app.

app-home

2. 
> We noticed you added groups:read to support private channels and wanted to check which method you're calling with that scope? If displaying a list of private channels, you'll need to make sure they're accessible to both the bot and user. You can do that with the users.conversations method, documented here: https://api.slack.com/methods/users.conversations

It looks like the `groups:read` scope was unnecessary. I've removed this from the staging app and in `SlackManager` in this PR.

3. 
> Additionally, in our testing we were not able to configure the staging app to post to a private channel or a DM at all, even after reinstalling. Could you please take a look at this?

This one still worries me. 
Initially, I wasn't able to integrate with Slack at all as the localhost URI wasn't included in the staging app which I've now added (not sure how it was working without this previously - perhaps Slack updated this).

I've also added the private channel functionality that was included in PR #4497 but not in our initial submission. 

However, it's not possible to DM an individual as discussed previously - it is only possible to message someone through the @ parabol bot. This functionality was working previously so I'm not sure why that wasn't working for them.

I haven't updated the Parabol app but if you're happy with the Parabol-staging changes (i.e. removing `groups:read` scope & flicking the "Messages" tab), I can update it.


mattkrick:
these changes seem reasonable. i already removed the localhost:3000 callback uri because they force us to have https everything (even localhost :facepalm: ). 

could you cherry-pick the bot DM piece from the other PR & push to dev? i think once we can DM via `@parabol` we could be good to go for another test

nickoferrall:
> could you cherry-pick the bot DM piece from the other PR & push to dev? 

Sure - this PR already includes the message private channels and DM from `@parabol` bits. I branched out of #4497 and deleted the migrations to create this PR. 

If you're happy with it, it's ready to merge to dev.

PR #4794, opened 2021-03-20
add issueKey to left nav subtitle

### LINKED ISSUE #4671: Poker: Integrated issue number part of agenda item
## Issue - Enhancement

[A user](https://app.hubspot.com/contacts/3888472/contact/2531) 🔒  writes:

> Jira issue number could be displayed in the left nav list along with title. We often refer to the issues by their number and I know other teams do that as well.

Users can:
   - if issue imported into poker, issue number prepended to agenda item

- **Estimated effort:** 2 hours

mattkrick:
Folks want the jira issue key in the left nav. 
So, here it is:
Screenshot from 2021-03-19 17-28-23

@ackernaut to confirm if this looks good. maybe we don't put a strikethrough the subtitle?

TEST
- [ ] if the issue can't be found, it uses the issue key as the title
- [ ] if the issue is found, it uses the title as title & issue key as subtitle
- [ ] if its a parabol issue, it just shows the title

 fix #4671

### DIFF
--- a/packages/client/components/JiraFieldMenu.tsx
+++ b/packages/client/components/JiraFieldMenu.tsx
@@ -4,8 +4,8 @@ import {createFragmentContainer} from 'react-relay'
 import useAtmosphere from '../hooks/useAtmosphere'
 import {MenuProps} from '../hooks/useMenu'
 import UpdateJiraDimensionFieldMutation from '../mutations/UpdateJiraDimensionFieldMutation'
+import JiraServiceTaskId from '../shared/gqlIds/JiraServiceTaskId'
 import {SprintPokerDefaults} from '../types/constEnums'
-import getJiraCloudIdAndKey from '../utils/getJiraCloudIdAndKey'
 import {JiraFieldMenu_stage} from '../__generated__/JiraFieldMenu_stage.graphql'
 import {JiraFieldMenu_viewer} from '../__generated__/JiraFieldMenu_viewer.graphql'
 import Menu from './Menu'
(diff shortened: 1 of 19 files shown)


ackernaut:
@mattkrick if we can show progress on the score, folks could use that meta to know if the issue has been groomed (instead of relying on the strikethrough). I still like the strikethrough for other meetings, so if these nav items are unique to poker, then yeah I’d be okay dropping it here. Of course our default is to make common patterns consistent, but when we have good cases, like poker, we can branch safely imo. These items feel more like a working session than progress through topics. So adding points is the clear indication that the team has at least partially addressed it. Striking through just because of curious forward navigation is bad in this case.

Testing…

### TEST
- [ ] if the issue can't be found, it uses the issue key as the title
- [ ] if the issue is found, it uses the title as title & issue key as subtitle
- [ ] if its a parabol issue, it just shows the title

ackernaut:
@mattkrick this looks good to me. I didn’t dig into how to trigger ‘issue key as title’ state.

If you like these styles, please merge: https://github.com/ParabolInc/parabol/pull/4798

ackernaut (approved):

ackernaut:
I saw a fatal error when testing this branch, and I’m not sure that it was caused by code changes here:

### Steps
- [ ] Start a poker session using the ‘weighted job’ template with 2 local viewers (browsers)
- [ ] Add 3 Jira issues and 1 Parabol issue
- [ ] Advance to Estimate
- [ ] The facilitator votes first and then sees a fatal view ‘Cannot destructure property 'label' of 'scaleValue' as it is undefined.’

mattkrick:
working on fatal error now, will merge once fixed

mattkrick:
@ackernaut couldn't reproduce the error, i took a guess where it might occur & I guess we'll just wait & see if someone can catch it in prod for more info

PR #4795, opened 2021-03-20
atomic update poker scope

### LINKED ISSUE #4719: Bug: Selecting Jira issue fast enough will cause item missing
## Issue - Bug

Repro:
1. Start a poker meeting
2. In Scope phase, wait for Jira issues load
3. Quickly select two issues
4. Observe that both of the issues appear on the navigation column, but then the first one is removed

If you select issues slowly, it does not seem to have this issue.

- **Estimated effort:** X points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

Users can:
   - [ ] Pick Jira issues from the list and see them show up on nav

mattkrick:
fix #4719 

I decided to go with a document lock instead of a nested in-place update.
Reason being, that nested update would have to operate on 1 update at a time, and in the case of a large select-all, it could actually take longer than using a lock.

Also updated the lock API so it's easier to just wait until we have a lock

TEST
- [x] follow the repro steps in #4719

### DIFF
--- a/packages/server/billing/helpers/processInvoiceItemHook.ts
+++ b/packages/server/billing/helpers/processInvoiceItemHook.ts
@@ -40,7 +40,7 @@ const getSafeProrationDate = async (
 
 const processInvoiceItemHook = async (stripeSubscriptionId: string) => {
   const redisLock = new RedisLock(stripeSubscriptionId, 3000)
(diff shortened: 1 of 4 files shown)


tianrunhe (approved):
Good stuff!

PR #4798, opened 2021-03-22
poker subnav item styles (merge into branch ‘fix-4671’)

ackernaut:
¡**Ojo**! This merges into branch `fix-4671` **not** master

- [ ] Adjusts alignment and spacing (e.g. Parabol task label wasn’t aligned vertically; Jira issue number was too cramped IMO)
- [ ] `isComplete` styles are now optional; the poker subnav items don’t need it per UX case

### DIFF
--- a/packages/client/components/PokerSidebarEstimateSection.tsx
+++ b/packages/client/components/PokerSidebarEstimateSection.tsx
@@ -43,10 +43,10 @@ const Title = styled('div')({
 })
 
 const Subtitle = styled('div')({
+  color: PALETTE.SLATE_500,
   fontSize: 11,
   fontWeight: 600,
-  color: PALETTE.SLATE_500,
-  lineHeight: '8px'
+  lineHeight: '12px'
 })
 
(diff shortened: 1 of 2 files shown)


packages/client/components/PokerSidebarEstimateSection.tsx
```diff
@@ -43,10 +43,10 @@ const Title = styled('div')({
 })
 
 const Subtitle = styled('div')({
+  color: PALETTE.SLATE_500,
   fontSize: 11,
   fontWeight: 600,
-  color: PALETTE.SLATE_500,
-  lineHeight: '8px'
+  lineHeight: '12px'
```
ackernaut, line 49: This only increase the subnav item height from `44` to `48` pixels, but I think it gives it room to breath. I don’t want to start getting cramped in the UI because then we’ll do it everywhere.

mattkrick:
i'm a little weary of the `isComplete` thing because checking things off a todo list is a really satisfying feeling, but if you're cool with it so am i

Issue #4715, opened 2021-02-25
Sprint Poker: GitHub: Scope: Create new issue


mattkrick:
AC
- [ ] Just like jira, a FAB exists. when clicked, it allows you to create a new issue & you can select the github repo for that issue

Estimate: 16 hours

nickoferrall:
Hey @mattkrick, it looks like there's no ScopePhaseArea for GitHub so I'll create an empty `ScopePhaseAreaGitHub` component with a FAB unless there's an existing branch with that UI?

mattkrick:
@nickoferrall in PR at #4777

nickoferrall:
Hey @mattkrick, the existing [createGitHubIssue mutation](https://github.com/ParabolInc/parabol/blob/1cab023452901d164045bc38905a211b413fcfd7/packages/server/graphql/mutations/createGitHubIssue.ts#L22) uses a `taskId` to link the GH issue to the Parabol task. The rich text editor in the task also means we can add markdown to the body of the GH issue.

In our 1-1 I think we mentioned that I'd use the existing mutation but I'm now thinking it'd be best to create a new mutation, e.g. `createGitHubIssuePoker`, as it wouldn't be linked to a Parabol task. `nameWithOwner` and a `content` string would be the args. A GH issue title but no body would be created and it'd look like this (prob "New issue title" instead of "summary"):

<img width="654" alt="new-gh-issue" src="https://user-images.githubusercontent.com/39854876/111693396-910e2b80-87fe-11eb-9b8b-8b0cf2ca8e58.png">

Alternatively, I could use DraftJS and increase the height of the new issue to provide space for a body but it wouldn't be consistent with the other integrations and might not look so pretty. Still, I think it'd make sense to create a new mutation.

What do you think?

mattkrick:
Correct me if I'm wrong, I think this is the same case as how jira is today?

- https://github.com/ParabolInc/parabol/blob/d5633f04cd1c8761232da7c132d5e63e7968efa9/packages/client/mutations/CreateJiraTaskIntegrationMutation.ts#L0-L1
- https://github.com/ParabolInc/parabol/blob/d5633f04cd1c8761232da7c132d5e63e7968efa9/packages/client/mutations/JiraCreateIssueMutation.ts#L39

one mutation is to create a new task & the other is to push an existing task to jira.
i think we could use the same strategy for github, but i'd also be open to architectural changes. for example:
- always create a parabol task? pros/cons?
[...]

Issue #4713, opened 2021-02-25
Sprint Poker: GitHub: Scope: Search Results


mattkrick:
AC
- [ ] by default, a reasonable list of items is populated (using the last successful query, if one exists)
- [ ] as the search bar is updated, the query is sent to our server, and then forwarded to github. the results are returned & populated, just like atlassian

Estimate: 16 hours

PR #4799, opened 2021-03-22
Create new GH issue in Poker

nickoferrall:
PR to resolve issue #4715 

This PR: 

- displays a FAB 
- lets the user add a new issue to a GitHub repo
- renames `CreateGitHubIssueMutation` to `GitHubTaskIntegrationMutation` to be consistent with the Jira integration

As the functionality for rendering the GitHub issue results isn't ready yet, the new issue disappears once it's created. I've commented out the code related to displaying & searching for issues as I assume it's out of scope for this PR.

Once rendering the Scope Results is ready (#4713), the `GitHubCreateIssueMutation` that I've created in this PR can be updated so that the new issue displays.


packages/server/graphql/mutations/gitHubCreateIssue.ts
```diff
@@ -0,0 +1,98 @@
...
+import GitHubServerManager from '../../utils/GitHubServerManager'
+import standardError from '../../utils/standardError'
+
+const gitHubCreateIssue = {
+  type: GraphQLNonNull(GitHubCreateIssuePayload),
+  description: 'Create an issue in GitHub',
+  args: {
+    meetingId: {
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'The id of the meeting where the Jira issue is being created'
```
mattkrick, line 16: GH issue, not jira

packages/server/graphql/mutations/gitHubCreateIssue.ts
```diff
@@ -0,0 +1,98 @@
...
+
+    //AUTH
+    if (!isTeamMember(authToken, teamId)) {
+      return standardError(new Error('Team not found'), {userId: viewerId})
+    }
+
+    // VALIDATION
+    const [viewerAuth, assigneeAuth] = await dataLoader.get('githubAuth').loadMany([
+      {teamId, userId: viewerId},
+      {teamId, userId: viewerId}
```
mattkrick, line 53: you're loading the same thing here

packages/client/mutations/GitHubCreateIssueMutation.ts
```diff
@@ -0,0 +1,48 @@
+import graphql from 'babel-plugin-relay/macro'
+import {commitMutation} from 'react-relay'
+import {StandardMutation} from '../types/relayMutations'
+import {GitHubCreateIssueMutation as TGitHubCreateIssueMutation} from '../__generated__/GitHubCreateIssueMutation.graphql'
+
+graphql`
+  fragment GitHubCreateIssueMutation_meeting on GitHubCreateIssueSuccess {
+    meetingId
```
mattkrick, line 8: what's this fragment accomplish? if i create a new github issue do my team members see it show up?

nickoferrall: This is a placeholder. As the results aren't rendering (an issue I can work on but assume is outside the scope here), the issue will be created in GH but the viewer & team members won't see anything in Parabol.

I thought it'd be worth keeping this fragment here as we'll need it once the results are rendering but I'd be happy to remove it too

mattkrick: let's treat i just like the jira version: https://github.com/ParabolInc/parabol/blob/d5633f04cd1c8761232da7c132d5e63e7968efa9/packages/client/mutations/JiraCreateIssueMutation.ts#L68

Issue #4566, opened 2021-01-21
Limit number of avatars displayed in team dashboard


nickoferrall:
If there are many team members in a team, the team member avatars can scroll on endlessly creating a horizontal scrollbar and all sorts of ugliness.

### AC

- [ ] Limit the number of avatars that can be displayed in the dashboard so that a horizontal scrollbar doesn't appear
- [ ] If there's not enough space to display all avatars, the final avatar on the right says: "+7" with 7 being the number of avatars that are not on display
- [ ] When hovering over the "+7" avatar, a dropdown menu appears with the missing avatars
- **Note**: in Sprint Poker we measured the space available so that the overflow was dynamic; previously in the meeting UI it was a fix number based on breakpoints. Going the dynamic path is an improvement over the fixed number

![many-team-members](https://user-images.githubusercontent.com/39854876/105359743-2f4e8e00-5bce-11eb-8bb1-f9c30dfb7b5b.png)

### EE
- 8 hours

jordanh:
Related to #4493

PR #4803, opened 2021-03-23
Limit number of dashboard avatars

nickoferrall:
PR to resolve issue #4566 

This PR: 

- dynamically calculates the max number of avatars that should be displayed on the dashboard
- if there's not enough space for all avatars, an overflow avatar is displayed 

In the issue I mentioned that we could hover over the overflow avatar which says +7 and a dropdown menu would show the 7 missing avatars. This dropdown behaviour doesn't exist in the new meeting view or poker view so, for consistency, I haven't added a dropdown menu to the dashboard.

### To test

Create a large number of meeting members. I did this by adding a new line below 59, adding `const test = [...teamMembers, ...teamMembers, ...]` and replacing `teamMembers` in lines 61-62 with `test`.

- [ ] When there's not enough space for all of the team member avatars, show an overflow avatar
- [ ] There's no horizontal scrollbar
- [ ] Overflow avatar shows the correct number of missing avatars


mattkrick:
Screenshot from 2021-03-31 11-52-20

AC
- [ ] no scrollbar appears, we just show the overflow avatar
- [ ] the online badge is not clipped for the right-most avatar
- [ ] the add team member icon is always on the left
- [ ] settings & integrations is on the same line as the org

nickoferrall:
Thanks, I've just pushed an update. 

I haven't been able to reproduce the add team member icon not being on the left - the latest update has the icon on the left on all breakpoints in Chrome & Safari. 

I think the clipped right-most avatar is caused by the horizontal-scroll. Now that I've removed that, it should no longer be clipped.

ackernaut:
Looking now…

ackernaut (approved):
Sorry for the delay here! I got pulled away obviously. Everything looks as outline above, including the fixes.

Feel free to merge when you’re ready @nickoferrall and excellent work! 👏

nickoferrall:
Thanks Terry!

PR #4806, opened 2021-03-23
always generate random id for google signup

### LINKED ISSUE #4726: Domain change in Google causing problems
## Issue - Bug

A [user](https://app.hubspot.com/contacts/3888472/contact/27384701) 🔒 writes:
 

> For a while now, me and my team colleagues have had issues to join meetings on Parabol. Typically we can log in, but then we can't join the meeting proper. It takes us several attempts, trying several ways to reach the meeting, until we finally get there.
> 
> I think this may have started when our employer changed email addresses from `@xxxx.eu` to `@xxxx.com`. We log in through Google as `@xxxx.com`, but when we land in the app our profiles still display as `@xxxx.eu`.
> 
> Another strange effect that I have noticed: sometimes (I think upon login, but not 100% sure), Parabol creates an organization and team for myself, regardless of already existing ones. As a result, right now I am a member of 3 different orgs with name "xxxxx's Org", which are unintended. I have deleted similar ones in the past.

> Can this be fixed so that my team doesn't have such a bad experience trying to get into our retrospectives?

tiffanyhan:
Resolve #4726

### DIFF
--- a/packages/server/graphql/mutations/loginWithGoogle.ts
+++ b/packages/server/graphql/mutations/loginWithGoogle.ts
@@ -91,13 +91,8 @@ const loginWithGoogle = {
         }
       }
 
-      const userId = `google-oauth2|${sub}`
-      const deletedUser = await r
-        .table('User')
-        .get(userId)
-        .run()
-
       // it's a new user!
+      const userId = `google-oauth2|${generateUID()}`
       const nickname = name || email.substring(0, email.indexOf('@'))
       const preferredName = nickname.length === 1 ? nickname.repeat(2) : nickname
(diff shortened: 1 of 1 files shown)


tiffanyhan:
@tianrunhe added you as reviewer for this small change. context is that when a domain change happens in gmail, the google id stays the same. we then try to create a new user with the same id, but it silently fails with a return value like `{"deleted": 0 ,
"errors": 1 ,
"first_error": "Duplicate primary key 'id'}` causing weird behavior. The PR gets around the issue by always generating a random id.

tianrunhe (approved):
Great! Cleaner solution now. I like it!

Issue #4624, opened 2021-02-01
Meetings view implemented


ackernaut:
Output from #4321 #4546 #4547

[See Figma](https://www.figma.com/file/dfTQxhoKzohkCpxlztBbhS/Meetings-view?node-id=74%3A9238)

### First pass criteria
- [See Figma](https://www.figma.com/file/dfTQxhoKzohkCpxlztBbhS/Meetings-view?node-id=221%3A0)
- The URL of the Meetings view is `[domain]/meetings`
- Folks land here as the default dashboard view instead of the timeline when logging in etc.
- Any open meeting appears here as a meeting card (card specs via #4623)
- Viewers tap on a card and go to that meeting
- Removes the meeting menu in the upper right of the dashboard top bar
- The FAB (Floating Action Button) says ‘Add Meeting’ instead of ‘Start Meeting’ for all dashboard views
- There’s an empty state ([see Figma](https://www.figma.com/file/dfTQxhoKzohkCpxlztBbhS/Meetings-view?node-id=99%3A6774)) — **note**: the wiggly lines are a stretch and would only show in the Meetings view when empty, they are probably anchored with absolute positioning from the button container, not the layout container, so that they are always relative to the button

### Expanded acceptance criteria (deferred, needs more design)
- Any open meeting or series shows up here as a card #4623 (all meetings may end up being a series of 1)
- Viewers can tap on any meeting card to go to the meeting, take a required action (pick a template), or edit the meeting (from the menu)
- There’s a team filter #4634 
- There’s a meeting type filter (let’s see if this is a tension if folks have a team filter #4634)
- There’s a ‘See completed meetings’ link that takes folks to the timeline. I’d rather see how many folks ask where their summaries are. Also the timeline may have a filter one day, but today if most folks go to their timeline there’s a good chance their last completed meeting is at the top.
- New folks have a demo card #4635
- There’s an empty state #4636

### Estimated effort
- 16 hours

ackernaut:
This issue is ready!

Issue #4623, opened 2021-02-01
Meeting card implemented


ackernaut:
Output from #4321 #4546 #4547

[See Figma](https://www.figma.com/file/dfTQxhoKzohkCpxlztBbhS/Meetings-view?node-id=74%3A9238)

### First pass criteria

 - [See Figma](https://www.figma.com/file/dfTQxhoKzohkCpxlztBbhS/Meetings-view?node-id=221%3A0)
- A meeting card exists if a meeting is open
- A meeting card has:
  - A type label
  - An illustration
  - The meeting name
  - The meeting phase meta
  - The team meta
 - The cards are read-only (e.g. you can’t edit the meeting name from this view)
 - Viewers tap on a card and go to that meeting
 - There’s a menu with 2 options: 1) Share invite link 2) Cancel meeting
 - Stretch: an avatar group is present for active viewers

### Expanded criteria (deferred, needs more design)

The meeting card…
- Represents [a meeting series](https://github.com/ParabolInc/parabol/issues/4584#issuecomment-772705804)
- Uses the meeting name as the title e.g. `Retro 1`
  - Folks can tap the name to edit (no need for an icon). Blurring or pressing enter submits the change.
- Has a meta line, the easy first version being `Stage • Team`
- Has a meeting type label in the upper left
- Has an illustration per meeting type, roughly 16x9
- Has a more menu in the upper right with options:
  - Copy invite link
  - Copy meeting URL
  - Cancel meeting series
- Active viewers appear as avatars with an overflow badge (tapping the badge can reveal the hidden list in a menu)
- Optional: activity meta represented by a `chat_bubble` and a count shows when there are no active viewers. Hovering the bubble shows a tooltip:
  - Retro e.g. `12 Reflections` or `5 Groups` or `7 Topics`
  - Poker e.g. `12 Stories`
  - Check-In e.g. `12 Agenda Items`
- Looks like a stack when it’s a series (deferred until stack UX is resolved)
- Can trigger dialogs to take actions (e.g. choose a template, edit calendar)
- Can be tapped. Tapping the card takes you to the current phase, or triggers the required dialog then to the first phase (e.g. pick a template -> first phase)
  - Deferred: has a semibold, blue action link in the bottom right when there’s a required task like ‘Choose a Template’ (we could just let them discover what’s blocking them by tapping the card with no explicit option. They choose a template if they must, and then are forwarded on. I like this. TA)
  
 ### Estimated effort
 - 30 hours

[...]

PR #4814, opened 2021-03-24
Meetings view v1

ackernaut:
Addresses #4624 and #4623 

### Tests
- [x] In the dashboard, the left navigation has a Meetings item
- [x] The meetings view shows my active meetings if any
- [x] The meetings view has an empty state
- [x] The meetings view has the ‘+ Add Meeting’ FAB
- [x] The meeting card for each active meeting:
  - Shows the meeting type, name, team name, and phase
  - Tapping the card takes you to the meeting
- [x] When someone else starts a meeting, I see the new card on my Meetings view immediately
- [x] When the meeting is ended, I see the card go away on my Meetings view immediately
- [x] Replaces `/me` with `/meetings` as the default view where folks land when (signing up, logging in, authenticated and coming from the demo, when archiving a team, when coming back from the New Meeting view, Cypress tests, etc.)
- [x] The Meetings view looks good on mobile
- [x] The meetings menu in the top bar has been removed
- [x] Added `useDocumentTitle` to say `My Meetings | Parabol` in the browser tab
- **Deferred**: there’s an active viewer avatar group on the meeting card (see #4835)
- **Deferred**: the meeting cards are sorted, probably by most active viewers (see #4835)
- **Deferred**: there’s a menu to share the invite link and maybe to kill the meeting (see #4836)


packages/client/components/MeetingsDashRoot.tsx
```diff
@@ -0,0 +1,35 @@
...
+  query MeetingsDashRootQuery {
+    viewer {
+      ...MeetingsDash_viewer
+    }
+  }
+`
+
+interface Props extends WithAtmosphereProps { }
+
+const MeetingsDashRoot = (props: Props) => {
```
ackernaut, line 21: Looking for feedback from a friendly Parabol developer. Is this a good way to go about this? To have a root query for the meetings view?

mattkrick: for now, I'd say it's not necessary & just use a fragment.
reason being:
- an extra query means more waterfall loading (the previous query has to load before this query can be fetched, so it'll take longer)
- we're already fetching most of these fields for the active meetings notification, so the amount of extra info isn't too great

ackernaut: The problem that I have is that I’m setting up a new route in `Dashboard.tsx` for `/meetings` — all of the routes have a route component. I’m not sure how to pass along the `viewer` through the router.

ackernaut:
@mattkrick an update here:
- The UI is in order, the adaptive layout is good enough for a first pass
- I deferred adding active viewer avatars #4835 and an options menu #4836 in the spirit of breaking down the work
- I need help working out the query fragments:
  - From what I saw all routes end up pointing to a root query, so I followed suit
  - The query `TopBarMeetingsActiveMeetings` in `TopBarMeetings.tsx` is included in many mutations. I didn’t get that far. I just included a basic query without any options to get the UI in place. I left `TopBarMeetings` for reference.

mattkrick:
@ackernaut this is looking good!
 
i see this fixes all the redirects to go from `/me` to `/meetings`, which means we'll have to hold off merging until we get the whole thing built out. I think that's fine for now, but if things start to get stale I might rip out those redirects just so we can get all the other good stuff into master

ackernaut:
@mattkrick why do we have to wait? re: `/me` to `/meetings`

packages/client/components/MeetingCard.tsx
```diff
@@ -0,0 +1,141 @@
...
+  overflow: 'hidden',
+  width: '100%'
+})
+
+interface Props {
+  meeting: MeetingCard_meeting
+}
+
+const ILLUSTRATIONS = {
+  retrospective: `${__STATIC_IMAGES__}/illustrations/retro-mtg-color-bg.svg`,
```
mattkrick, line 76: this reminds me we should move away from this pattern to support airgapped deployments.
fine for now! i'll backlog an issue

packages/client/components/StartMeetingFAB.tsx
```diff
@@ -75,8 +75,8 @@ const StartMeetingFAB = (props: Props) => {
   return (
     <Block>
       <Button onClick={onClick} onMouseEnter={onMouseEnter} onMouseLeave={onMouseLeave}>
-        <MeetingIcon isExpanded={isBig}>{'forum'}</MeetingIcon>
-        <MeetingLabel isExpanded={isBig}>{'Start New Meeting'}</MeetingLabel>
```
mattkrick, line 79: I like removing the word `Start`
i think we need to reduce the width of the button though, because currently it looks like there's just a lot of empty space on the right

mattkrick:
looks absolutely amazing! 
the only thing that looks a little off is the button. this is probably because i had to use a fixed width button to achieve the animations & we removed a whole word without reducing the width:
Screenshot from 2021-04-01 10-10-53

in other places, the icon is part of the body that is centered: 
Screenshot from 2021-04-01 10-10-39

old version for context: 
Screenshot from 2021-04-01 10-11-14

packages/client/components/Dashboard.tsx
```diff
@@ -45,6 +50,16 @@ const getShowFAB = (location: NonNullable<RouteProps['location']>) => {
       exact: true,
       strict: true
     }) ||
+    !!matchPath(pathname, {
+      path: '/meetings/',
+      exact: true,
+      strict: true
```
mattkrick, line 56: if you disable `strict`, you allow the trailing slash

ackernaut:
@mattkrick the original fragment `TopBarMeetingsActiveMeetings` was used in many mutations. I assume we should update that for the meetings view? The fragment `TopBarMeetings_teams` had some parameters like `plural: true` and `mask: false`. Do we need that extra magic?  I’m thinking about the new `MeetingCard_meeting` fragment.

mattkrick:
hmmmm yeah we should get rid of TopBarMeetings assets & see where typescript & relay yell at us.
definitely should capture the required fields from MeetingCard since we'll be adding a bunch of extra stuff in the near future

packages/client/mutations/AcceptTeamInvitationMutation.ts
```diff
@@ -40,7 +40,6 @@ graphql`
     # this is just for the user that accepted the invitation
     team {
       ...DashNavListTeam
-      ...TopBarMeetingsActiveMeetings
```
ackernaut, line 43: @mattkrick I yanked the TopBarMeetings component and all the places where the fragment was used. I’m not exactly sure the logic behind reusing the fragments everywhere, did we want to update active meetings when these events occurred, like accepting an invite to a new team that may have active meetings? I did not replicate this with the MeetingsDash active meetings fragment. Does this break expected behaviors?

ackernaut:
IMO this would be safe to ship. At any rate, I don’t want this PR to get stale. I’ve pulled in latest master and yanked the old TopBarMeetings file and fragment usage.

ackernaut:
We need to add `useDocumentTitle` so that it says ‘My Meetings | Parabol’ on the browser tab.

ackernaut:
@tianrunhe I’m going to merge this as I completed a review of UX and code with Matt. Would you mind kicking the tires on the Meetings view when you test release?

tianrunhe:
> @tianrunhe I’m going to merge this as I completed a review of UX and code with Matt. Would you mind kicking the tires on the Meetings view when you test release?

Will do!

PR #4817, opened 2021-03-25
Migrate team reads

### LINKED ISSUE #4630: Read Team from PG instead of RethinkDB (Phase 2 of 3)
After #4629 is working is prod, we need to start reading from pg instead of RethinkDB. See #4530 

AC
- [ ] All reads for the `Team` table come from pg, not RethinkDB
- [ ] if you nuke the data in the Team table in rethinkdb, nothing bad happens

Estimate: 16 hours

tiffanyhan:
Resolve #4630


tiffanyhan:
@mattkrick any idea what the problem could be? the [typecheck is complaining](https://github.com/ParabolInc/parabol/runs/2204397001) it can't find type `Json`, even though that issue was previously solved by adding the type to `packages/server/types/index.d.ts`.

https://github.com/ParabolInc/parabol/runs/2204397001

mattkrick:
all fixed! 
it was the client package that failed the typecheck, which is weird because I don't know why the client package traversed all the way into server generated assets :shrug:

tiffanyhan:
thanks @mattkrick ! i'll keep this as draft for now, as this is still blocked from releasing until we can run the equality checker in prod (and resolve any issues that might come up).

tiffanyhan:
closing in favor of https://github.com/ParabolInc/parabol/pull/5201

Issue #4765, opened 2021-03-16
Poker: point value of revealed votes sorted 


jordanh:
## Issue - Enhancement

![image](https://user-images.githubusercontent.com/103000/111351349-52cb0d80-8640-11eb-8c33-eee705970947.png)

### Acceptance Criteria (optional)
Sprint poker ssers:
   - See revealed votes sorted (chef's choice for asc or desc order)

- **Estimated effort:** 2 hours

PR #4821, opened 2021-03-26
Sort revealed scores in Poker

nickoferrall:
PR to resolve issue #4765

### DIFF
--- a/packages/client/components/PokerDiscussVoting.tsx
+++ b/packages/client/components/PokerDiscussVoting.tsx
@@ -49,16 +49,16 @@ const PokerDiscussVoting = (props: Props) => {
         topLabel = label
       }
     })
-    const scoreLabels = Object.keys(scoreObj).sort((a, b) => {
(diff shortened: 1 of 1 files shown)


packages/client/components/PokerDiscussVoting.tsx
```diff
@@ -49,16 +49,16 @@ const PokerDiscussVoting = (props: Props) => {
...
-    const rows = scoreLabels.map((label) => {
-      return {
-        key: label,
-        scaleValue: scaleValues.find((scaleValue) => scaleValue.label === label)!,
-        scores: scoreObj[label]
-      }
-    })
+    const scoreLabels = Object.keys(scoreObj)
+    const rows = scaleValues
+      .filter((scaleValue) => scoreLabels.find((label) => label === scaleValue.label))
```
mattkrick, line 54: nice change! i like that you're using their defined sortOrder.
Only 1 change: let's make sure the QUESTION_CARD is at the top & the PASS_CARD is at the bottom. reason being, if someone has a question, that should be the first item of discussion

mattkrick (changes requested):
just 1 little change, then feel free to merge

Issue #4825, opened 2021-03-26
Add type safety to deleteUser


nickoferrall:
@gcrickman [mentioned](https://parabol.slack.com/archives/C4JAUUZ9P/p1616680996039700) that deleting a user returns an error message.

This is because `reason` is an optional argument but we're slicing it in `deleteUser`. 

### AC

- [ ] Add type safety to `deleteUser` 
- [ ] If the user doesn't add a reason, the mutation doesn't return an error

PR #4826, opened 2021-03-26
Add type safety to deleteUser

nickoferrall:
PR to resolve issue #4825 

This PR defines the argument types & checks if `reason` exists so that an error isn't returned when it's not provided.


mattkrick:
LGTM! nice find

PR #4831, opened 2021-03-29
Validate new poker issues

nickoferrall:
This PR:

- Validates the new GitHub or Jira issue 
- Increases the width of the new issue input
- Gives the user feedback if they enter an invalid new issue title

I thought I'd create a separate PR for this one to avoid scope creep in the Create GH Issue PR.

### DIFF
--- a/packages/client/components/NewGitHubIssueInput.tsx
+++ b/packages/client/components/NewGitHubIssueInput.tsx
@@ -100,8 +100,8 @@ interface Props {
 const validateIssue = (issue: string) => {
   return new Legitity(issue)
     .trim()
-    .min(2, 'C’mon, you call that an issue?')
-    .max(255, "Who's gonna read all that? Short and sweet please.")
+    .min(2, `C’mon, you call that an issue?`)
+    .max(255, `Who's gonna read all that? Short and sweet please.`)
 }
 
 const NewGitHubIssueInput = (props: Props) => {
--- a/packages/client/components/NewGitHubIssueInput.tsx
+++ b/packages/client/components/NewGitHubIssueInput.tsx
@@ -79,7 +79,7 @@ const NewIssueInput = styled('input')({
   color: PALETTE.TEXT_MAIN,
   fontSize: 16,
   margin: 0,
-  padding: 0,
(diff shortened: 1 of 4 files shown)


packages/client/components/NewGitHubIssueInput.tsx
```diff
@@ -79,30 +84,49 @@ const SearchInput = styled('input')({
...
   meeting: NewGitHubIssueInput_meeting
   setIsEditing: (isEditing: boolean) => void
   viewer: NewGitHubIssueInput_viewer
 }
 
+const validateIssue = (issue: string) => {
+  return new Legitity(issue)
+    .trim()
+    .min(2, 'C’mon, you call that an issue?')
+    .max(255, "Who's gonna read all that? Short and sweet please.")
```
mattkrick, line 104: :rofl: nice
you can always use a smart quote `’` to avoid using an escape (plus it gives it a nice polished look)

nickoferrall: Nice, added!

packages/client/components/NewGitHubIssueInput.tsx
```diff
@@ -131,14 +164,16 @@ const NewGitHubIssueInput = (props: Props) => {
...
           <Form onSubmit={handleCreateNewIssue}>
-            <SearchInput
+            <NewIssueInput
               autoFocus
               onBlur={handleCreateNewIssue}
-              onChange={(e) => setNewIssueText(e.target.value)}
-              placeholder='New issue summary'
+              onChange={onChange}
+              name='newIssue'
+              placeholder='New issue title'
```
mattkrick, line 172: What are your thoughts on using a `maxLength` here? doing so could prevent an error state from occurring, which might be nice

nickoferrall: I do enjoy the snarky error messages but I agree that `maxLength` is a better UX! I've pushed the update but have kept the validation check for min length

mattkrick:
TBH I don't know why I didn't merge this sooner. LGTM!
Nice work

PR #4843, opened 2021-03-30
Enable Datadog APM tracer

### LINKED ISSUE #4298: Add GraphQL Logger for long running queries
To improve debugging, I'd like a GraphQL Logger that logs long running queries (> 1 second?) broken down by showing how long it took each field to resolve.
Sampling is OK (if performance is a problem)
Where the data lives is TBD (RethinkDB, datadog, ???)
There are already products that do this today, but last time we tried doing it with datadog there were severe performance penalties. Perhaps we should try again? I think this is probably better to buy vs. build

AC
- [ ] Perform a market analysis of all products that do this today. Compare them & present the comparison to @mattkrick so we can decide which path to go down. Also include how you'd build it if that looks favorable 
- [ ] long running queries are logged somewhere, showing the time it took to resolve each field & the variables passed in
- [ ] built in such a way that we may be able to add logging for payloads that throw an error (or return an error in a successful payload)

Estimate: 40 hours

tianrunhe:
Closes #4298 . This PR enables Datadog APM for our app. GraphQL, Redis and Postgres will all be enabled for free.

For this approach,
Pros:
1. Simple and fast, almost no code change
2. Env variable toggled switch to turn it on and off as we wish. No full-cycle of release deployment is needed.
3. "Safe to try". We could do sth like: turning it on for a week, identifying the GraphQL bottleneck, turning it off, optimizing the queries and turning it on again to see the numbers.

Cons:
1. Potential performance give back (need to be tested in production)
2. Guaranteed performance lost in GraphQL execution because in order for the instrument to work, we have to use the vanilla GraphQL.js library, instead of GraphQL-JIT.

Learnings while I'm testing on dev server (@jordanh):
1. As mentioned in cons 2., we need to reroute all of the GraphQL queries to the vanilla GraphQL.js executor. In the beginning, I didn't realize one of the major diffs between desktop vs. production server is that on desktop we have the full query docs in the request, whereas in production we only include the `docId` in the request. In production we use the `docId` to fetch the compiled and optimized GraphQL-JIT queries. However when I redirect those queries to the vanilla GQL executor, it needs the entire document hence it breaks all GraphQL queries in production environment.
2. After I fixed the above issue, I quickly encountered another one where some GraphQL traces are missing. This one is a quick fix though: in addition to the main `parabol-action` process, we also need to enable a separate DD tracer for the `gql-executor` process.


packages/server/graphql/executeGraphQL.ts
```diff
@@ -55,6 +58,9 @@ const executeGraphQL = async (req: GQLRequest) => {
   let response: FormattedExecutionResult
   if (isAdHoc) {
     response = await graphql({schema, source, variableValues, contextValue})
+  } else if (docId && process.env.DD_TRACE_ENABLED === 'true') {
+    const document = await documentCache.fromID(docId)
+    response = await graphql({schema, source: print(document!), variableValues, contextValue})
```
mattkrick, line 63: sub out graphql for `graphql.execute` and we can pass in the document directly

packages/server/tracer.js
```diff
@@ -1,3 +1,14 @@
 import tracer from 'dd-trace'
-tracer.init() // initialized in a different file to avoid hoisting.
+tracer.init({
+  enabled: process.env.DD_TRACE_ENABLED === 'true',
+  env: process.env.DD_ENV ?? 'test'
+})
+
+tracer.use('graphql', {
```
mattkrick, line 7: can we get rid of this `trace.use`? graphql doesn't exist on the webserver, so it should be safe to remove

packages/server/server.ts
```diff
@@ -1,5 +1,4 @@
-// Datadog APM, must be first import (disabled for now)
-// import './tracer'
+import 'dd-trace/init'
```
mattkrick, line 1: side effects on imports are the root of all evil :wink:. 

plus, we probably don't wanna do much tracing in development.

if we initialize it here in our own code, it should be more understandable to our future selves
```ts
import tracer from 'dd-trace'
if (PROD) {
  tracer.init()
}
```

tianrunhe: I agreed that we don't need tracing on local so the PROD condition makes sense. Curious about the import stuff though, what's your take on the [suggestion from DD documentation](https://docs.datadoghq.com/tracing/setup_overview/setup/nodejs/?tab=containers#adding-the-tracer-in-code) where they suggest to put initialization in a different file?

mattkrick:
looking good! last step: can you add the env vars to:
- development, staging, and production server
- some mock values to `.env.example` for our friends playing at home

packages/server/server.ts
tianrunhe: @mattkrick

tianrunhe:
action-devops repo was already updated [here](https://github.com/ParabolInc/action-devops/commit/b41d677f234f95b38561974551759378f4b212a3)

mattkrick:
ah, did you update the server via dokku?

tianrunhe:
Wait... I thought during deployment the config is pulled from the devops repo but it does not update the serve automatically?🤯

tianrunhe:
@mattkrick  OK I've updated the servers:

> root@action-dev-nyc1-01:~# dokku config:show web | grep DD
DD_AGENT_HOST:                   172.17.0.1
DD_API_KEY:                      77216f30f96818f3b294ef788c40e618
DD_APM_ENABLED:                  true
DD_ENV:                          development
DD_LOGS_INJECTION:               true
DD_RUNTIME_METRICS_ENABLED:      true
DD_TRACE_ANALYTICS_ENABLED:      true
DD_TRACE_ENABLED:                false

> root@action-staging-web-nyc1-02:~# dokku config:show web | grep DD
DD_AGENT_HOST:                    172.17.0.1
DD_API_KEY:                       77216f30f96818f3b294ef788c40e618
DD_APM_ENABLED:                   true
DD_ENV:                           staging
DD_LOGS_INJECTION:                true
DD_RUNTIME_METRICS_ENABLED:       true
DD_TRACE_ANALYTICS_ENABLED:       true
DD_TRACE_ENABLED:                 false

> root@action-production-web-nyc1-02:~# dokku config:show web | grep DD
DD_AGENT_HOST:                    172.17.0.1
DD_API_KEY:                       77216f30f96818f3b294ef788c40e618
DD_APM_ENABLED:                   true
DD_ENV:                           production
DD_LOGS_INJECTION:                true
DD_RUNTIME_METRICS_ENABLED:       true
DD_TRACE_ANALYTICS_ENABLED:       true
DD_TRACE_ENABLED:                 false

PR #4844, opened 2021-03-30
Remove IM scopes in Slack

nickoferrall:
To send a DM from the `@Parabol` bot, we were using the [conversations.open](https://api.slack.com/methods/conversations.open) method and the [conversations.list](https://api.slack.com/methods/conversations.list) method to get the bot's channel id. Both methods required the IM scope.

The [postMessage](https://api.slack.com/methods/chat.postMessage) method requires a Slack channel id as an argument. However, instead of using `conversations.list` to get the bot's channel id (using the IM scope), the fine print explains that you can pass the user's id as the channel id which will then send a notification from `@Parabol`. 

This PR:

- Uses the user's id as the channel id when using the `@Parabol` bot
- Removes references to IM

In the `Parabol-staging` app, I've removed the IM scopes but I haven't removed the `localhost` redirect URL in case you want to test the functionality.

### DIFF
--- a/packages/client/hooks/useSlackChannels.ts
+++ b/packages/client/hooks/useSlackChannels.ts
@@ -1,7 +1,6 @@
 import {useEffect, useState} from 'react'
 import {SlackChannelDropdownChannels} from '../components/SlackChannelDropdown'
 import SlackClientManager from '../utils/SlackClientManager'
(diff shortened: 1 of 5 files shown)


mattkrick:
well done! tested IMs, group chats & regular channels, all work as intended!

PR #4849, opened 2021-04-01
Meetings view mk

mattkrick:
@ackernaut for your consideration. i was winging it with the 120px width, i could be off by a couple


packages/client/components/Dashboard.tsx
```diff
@@ -120,7 +108,10 @@ const Dashboard = (props: Props) => {
         )}
         <DashMain>
           <Switch>
-            <Route path='/meetings' component={MeetingsDashRoot} />
+            <Route
+              path='/meetings'
+              render={(routeProps) => <MeetingsDash {...routeProps} viewer={viewer} />}
```
mattkrick, line 113: not exactly intuitive, but this is how you pass in the viewer

ackernaut: I see, I see

ackernaut:
Looking at this now…

PR #4852, opened 2021-04-01
Fixed meeting control bar width

BartoszJarocki:
There's an issue with `MeetingControlBar` width and `fit-content` usage. In this case `width: auto` fixes the issue. 
### Before
CleanShot 2021-04-01 at 22 20 00
### After
CleanShot 2021-04-01 at 22 20 39


mattkrick:
looks good!

@tianrunhe can you do a quick review? things to check for:
- [ ] animations across phases still work
- anything else you can think of

tianrunhe (approved):
LGTM! Great work!

PR #4856, opened 2021-04-03
Pg rethink eq checker for User & Team table

### LINKED ISSUE #4818: Implement postgres <=> rethink equality checker for User table
## Issue - Enhancement
Between the writes and reads stages of table migration, we need to periodically audit equality of prod data, and resolve any differences.

- [ ] a private graphql mutation exists that takes table name as arg, and compares equality btwn records in rethinkdb and postgres
- [ ] should use loose equality when relevant, e.g. null or empty string in pg is equivalent to undefined in rethink
- [ ] any unequal records are written to an output file, which can then be used to resolve differences
- [ ] User table comparison implemented

- **Estimated effort:** 8 hours ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

tiffanyhan:
Resolve #4818 #4857


tianrunhe:
Hey @tiffanyhan, I did a local test and it worked well. I have a few "challenging" questions below:

* How do you plan to use/interpret the output to fix the inequality? Do we really need a nested JSON output? Depend on how we will use the output, would other format be easier to work with?
* Do we really need to output for **every** records in one run? I imagine that if we have a bug in our code which makes inequality between pg and rethink, it's highly likely that it will be present in all records so probably a slice of the data is quite enough for us to troubleshoot and fix. WDYT?
* If you agree with last question, then my next challenging question is: can we then fit the output in the mutation result directly, rather than outputting to a file in the server. I think:
  1. I have to log into the server to grab the output file anyway, so why bother having this one-off tool in the admin console? It could just be a simple node script running on the server for example so I don't need to go to two places. It's not like that growth team will ever use it.
  2. If we agree that we don't need to output for every record, then we can have a `limit/offset` argument in the mutation. Looks like you already have that in `checkTableEq` so we can just expose that. This way it makes the tool run faster, on smaller data sets and result directly shown on the admin console page.

tiffanyhan:
thanks for the review @tianrunhe!

i like the nested json format bc i want to be able to easily visually scan the results and differentiate btwn false positives and true positives. e.g. false positive would be rethink email is `"DELETED"` and pg email is `"DELETED:<user_id>:<:timestamp>"`. did you have another format in mind?

i think that's a really good point, most of the time only a small slice of the data is needed. i still want to be able to do the entire dataset, once the small slices are resolved. there might be some queries in private mutations that are run infrequently, and so only affect a few enterprise customers for example. 

i'm thinking to expose two options: one sync option where a small slice of the dataset is taken and results returned to the graphiql; and another async option where the entire dataset is compared and results written to file. that way we can use mostly the first option and it will speed up iterations greatly, but still have the peace of mind of running the entire dataset when needed.

lmk if you have followup thoughts!

tianrunhe:
@tiffanyhan yeah I like having two options so I imagine I will look like:

```
mutation {
  checkRethinkPgEquality(tableName: "User", first: 1000, outputToFile: true)
}
```
Let's do it!

tiffanyhan:
@tianrunhe changes made, thanks for looking into this!

packages/server/postgres/utils/checkTeamEq.ts
```diff
@@ -0,0 +1,72 @@
+import Team from '../../database/types/Team'
+import getTeamsById, {IGetTeamsByIdResult} from '../../postgres/queries/getTeamsById'
+import getRethink from '../../database/rethinkDriver'
+import lodash from 'lodash'
+import {CustomResolver, checkTableEq} from './checkEqBase'
+
+const namesAreEqual: CustomResolver = (rethinkName, pgName) =>
+  lodash.isEqual(rethinkName, pgName) || rethinkName.slice(0, 100) === pgName
```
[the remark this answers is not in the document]
tiffanyhan: yup! i tried to use varchars unless it's for a url or email field. we always have the server limit on the payload, but the extra db constraint on char limit is an extra safety measure.

https://github.com/ParabolInc/parabol/blob/master/packages/server/postgres/migrations/1616623815919_addTeamTable.ts

packages/server/postgres/utils/checkTeamEq.ts
```diff
@@ -0,0 +1,72 @@
...
+}
+
+const maybeUndefinedFieldsDefaultValues = {
+  jiraDimensionFields: [],
+  lastMeetingType: 'retrospective'
+} as {
+  [Property in keyof Partial<Team>]: any
+}
+
+const getPairNeFields = (rethinkTeam: Team, pgTeam: IGetTeamsByIdResult): string[] => {
```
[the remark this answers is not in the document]
tiffanyhan: yup! i was holding off on refactoring this part, but i think the API should be stable now. the only thing I changed in the generic function is the shape of `maybeUndefinedFields` now has tuple with `[<:CustomResolver>, <:defaultPgValue>]` as values, since a field can both maybe be undefined and maybe need a custom resolver.

https://github.com/ParabolInc/parabol/pull/4856/files#diff-9f822689a32c27c0e504373b650b2638acee971898aa27f930dec63e2f86b0d2R38-R42

packages/server/graphql/intranetSchema/mutations/checkRethinkPgEquality.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    const fileName = `${tableName}-${new Date()}`
+    const fileLocation = path.join(
+      process.cwd(),
+      'packages/server/graphql/intranetSchema/mutations',
+      '__output__',
+      fileName
+    )
+    await fs.promises.mkdir(path.dirname(fileLocation), {recursive: true})
```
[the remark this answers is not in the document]
tiffanyhan: i assume that since the directory is in the project and not in the system root dir or elsewhere outside of the project, the node user will have permissions to write. am i missing something important?

tiffanyhan:
@tianrunhe simplified the mutation api, hopefully it makes it more intuitive to interact with

tianrunhe (approved):
LGTM!

tiffanyhan:
requesting final review from @mattkrick

packages/server/postgres/utils/checkEqBase.ts
```diff
@@ -0,0 +1,125 @@
...
+    .run()) as rethinkType[]
+
+  const ids = rethinkRows.map((t) => (t as WithId<rethinkType>).id) as string[]
+  const pgRows = await pgQuery(ids)
+  const pgRowsById = {} as {[key: string]: pgType}
+  pgRows.forEach((pgRow) => {
+    pgRowsById[(pgRow as WithId<pgType>).id] = pgRow
+  })
+
+  for (const rethinkRow of rethinkRows) {
```
mattkrick, line 104: implicitly, this is assuming that PG is a subset of rethinkDB.
it's possible to introduce a bug where there are more PG docs than rethinkDB docs.
as a final check, we need to get a count of all the docs in rethinkdb and all the docs in PG and make sure the 2 numbers match.

mattkrick (changes requested):
great first pass! looking forward to your thoughts & the next round

packages/server/postgres/utils/checkEqBase.ts
```diff
@@ -0,0 +1,125 @@
...
+): string[] {
+  const neFields = [] as string[]
+
+  for (const [f, customResolver] of Object.entries(alwaysDefinedFieldsCustomResolvers)) {
+    const [rethinkValue, pgValue] = [rethinkUser[f], pgUser[f]]
+    if (!lodash.isEqualWith(rethinkValue, pgValue, customResolver as CustomResolver)) {
+      neFields.push(f)
+    }
+  }
+  for (const [f, tuple] of Object.entries(maybeUndefinedFieldsCustomResolversDefaultValues)) {
```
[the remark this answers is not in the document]
tiffanyhan: yes i was a little iffy on the tuple here so agree with using an object instead

packages/server/postgres/utils/checkEqBase.ts
```diff
@@ -0,0 +1,125 @@
...
+        error: string | string[]
+        rethinkRow?: Partial<rethinkType>
+        pgRow?: Partial<pgType>
+      }
+    | number
+  recordsCompared: number
+}
+
+function getPairNeFields<rethinkType, pgType>(
+  rethinkUser: rethinkType,
```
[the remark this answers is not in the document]
tiffanyhan: oh that's totally an oversight, will change!

packages/server/graphql/intranetSchema/mutations/checkRethinkPgEquality.ts
```diff
@@ -0,0 +1,84 @@
...
+  },
+  resolve: async (_source, {tableName, numberOfRecords, startPage, writeToFile}, {authToken}) => {
+    // AUTH
+    requireSU(authToken)
+
+    // VALIDATION
+    if (!tableResolvers.hasOwnProperty(tableName)) {
+      return `That table name either doesn't exist or hasn't yet been implemented.`
+    }
+    if (numberOfRecords && numberOfRecords > 3000) {
```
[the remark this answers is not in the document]
tiffanyhan: hmmm i like this! if the graphiql request times out bc the server keeps paginating and hasn't found 10 errors yet, then whoever's calling the mutation can just rerun the request and set `writeToFile=true` to see the results.

packages/server/postgres/utils/checkEqBase.ts
```diff
@@ -0,0 +1,125 @@
...
+    const [rethinkValue, pgValue] = [rethinkUser[f], pgUser[f]]
+    if (!lodash.isEqualWith(rethinkValue, pgValue, customResolver as CustomResolver)) {
+      neFields.push(f)
+    }
+  }
+  for (const [f, tuple] of Object.entries(maybeUndefinedFieldsCustomResolversDefaultValues)) {
+    const [customResolver, defaultValue] = tuple as [CustomResolver, any]
+    const [rethinkValue, pgValue] = [rethinkUser[f], pgUser[f]]
+    if (!lodash.isUndefined(rethinkValue)) {
+      if (!lodash.isEqualWith(rethinkValue, pgValue, customResolver)) {
```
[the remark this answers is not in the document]
tiffanyhan: agree with the `preferredNames`

the story is a bit more complicated with `email` field. in pg, we have unique constraint on `email` so writing `DELETED` for any deleted user will violate the constraint and throw an error. 

if we wanted to, we can do the migration without any constraints in pg, then add the constraint later and deal with any violations in the data then. i'm open to changing to doing it this way after seeing how applying constraints now can complicate things.

@mattkrick what do you think?

packages/server/postgres/utils/checkEqBase.ts
```diff
@@ -0,0 +1,125 @@
...
+  for (const [f, customResolver] of Object.entries(alwaysDefinedFieldsCustomResolvers)) {
+    const [rethinkValue, pgValue] = [rethinkUser[f], pgUser[f]]
+    if (!lodash.isEqualWith(rethinkValue, pgValue, customResolver as CustomResolver)) {
+      neFields.push(f)
+    }
+  }
+  for (const [f, tuple] of Object.entries(maybeUndefinedFieldsCustomResolversDefaultValues)) {
+    const [customResolver, defaultValue] = tuple as [CustomResolver, any]
+    const [rethinkValue, pgValue] = [rethinkUser[f], pgUser[f]]
+    if (!lodash.isUndefined(rethinkValue)) {
```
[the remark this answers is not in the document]
tiffanyhan: ah good to know, i'll look into those alternatives!

packages/server/postgres/utils/checkEqBase.ts
tiffanyhan: the user `email` field also complicates this too. in pg, we set a limit of 500 chars for email field. that caused some rows to violate the unique constraint bc there were some bad actors who inserted users with ~1mb email addresses, and the first 500 chars of these were the same chars. in the backfill, i got around that problem by skipping over any users whose email had more than 500 chars after verifying that all of them looked like nonsense.

so in the case of users, the final count of both tables in rethink and pg would be different. ways around this are to add in extra logic to the fn to filter rethink rows to the valid ones. alternatively, we could drop the constraint until the migration is complete and add it before we switch over to using postgres.

more to consider on whether to defer having pg constraints @mattkrick

tiffanyhan:
@mattkrick wanted to give an update on what i was thinking. lmk if you have any thoughts:

i think we should start with not putting **any** constraints in pg table (maybe char limit is okay). this way we have a bit of separation of concerns, and the equality checker logic can be greatly simplified to truly compare docs 1to1. once we've confirmed equality, there should be a separate step to add pg constraints into the tables. pg will let us know what records are in violation, so any cleanup can happen at this point. the step to add constraint can happen before or after migrating reads, but not before confirming equality.

to add foreign key constraint, the referenced table must also be migrated, so there will be some staggering here depending on the order of table migrations. in the initial backfill, we can put a `// todo: add constraint once `xx` table migrated` to more easily find the relevant fields again.

packages/server/postgres/utils/checkEqBase.ts
tiffanyhan: added `extraPgRows`, `missingPgRows`, `rethinkRowCount`, and `pgRowCount` to the error resp. it's possible to have `rethinkRowCount === pgRowCount` and still have errors (e.g. if there's one missing pgRow and one extra pgRow), so any missing and extra rows are returned explicitly.

packages/server/postgres/utils/checkEqBase.ts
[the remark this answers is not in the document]
tiffanyhan: migrations added to drop pg constraints and make data values equal

packages/server/postgres/utils/checkEqBase.ts
```diff
@@ -0,0 +1,125 @@
...
+    pgNeFields[neField] = pgRow[neField]
+  }
+  errors[rowId] = {
+    error: neFields,
+    rethinkRow: rethinkNeFields,
+    pgRow: pgNeFields
+  }
+}
+
+export async function checkTableEq<rethinkType, pgType>(
```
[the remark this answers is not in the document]
tiffanyhan: thanks for calling this out. i wanted to give more semantic meaning using the types, but they weren't actually getting us any type safety.  using `DBDoc` interface now which gives us explicit typing, and noop extension of the interface gives us better semantics.

tiffanyhan:
@mattkrick re-requesting review. i did a thing, adding an extra temp db field `eqChecked: date`, which allows us to surface any extra pg rows without needing to do a seq scan or bring a huge payload back to the app server. we can delete the field after migration is complete. looking forward to your thoughts on 2nd iteration

packages/server/postgres/migrations/postDeploy/1619383917863_updateDeletedUserEmails.ts
```diff
@@ -0,0 +1,58 @@
...
+  reasonRemoved: null,
+  rol: null,
+  // app doesn't allow following fields to be undefined, but found bad data anyway
+  inactive: false,
+  payLaterClickCount: 0,
+  featureFlags: []
+}
+
+const cleanUsers = (users: User[]): IBackupUserQueryParams['users'] => {
+  const cleanedUsers = [] as any
```
mattkrick, line 26: nit: when the input & output arrays are the same size, try to use `map` to get free type checking & express that it's (probably) a pure function.

e.g. the function could be replaced with:

```ts
const skippedPgUsers =  skippedRethinkUsers.map((user) => ({
  ...undefinedUserFieldsAndTheirDefaultPgValues,
  ...user
}))

File: packages/server/postgres/queries/src/updateTeamByTeamIdQuery.sql [hunk 12]
```diff
@@ -9,5 +10,7 @@ UPDATE "Team" SET
   "lastMeetingType" = COALESCE(:lastMeetingType, "lastMeetingType"),
   "tier" = COALESCE(:tier, "tier"),
   "orgId" = COALESCE(:orgId, "orgId"),
-  "updatedAt" = COALESCE(:updatedAt, "updatedAt")
-WHERE id = :id;
+  "updatedAt" = COALESCE(:updatedAt, "updatedAt"),
+  "eqChecked" = COALESCE(:eqChecked, "eqChecked")
+WHERE id IN :ids
+RETURNING *;
```
mattkrick, line 16: I've always gone back & forth on returning the doc.
On the 1 hand, it's great because it's a single hit to the DB to get a write + a fresh read.
On the other hand, if I don't need that read, it's a little wasteful.

No changes needed, but if you come across some strong opinions either way would love to hear em!

packages/server/postgres/queries/updateTeamByTeamId.ts
```diff
@@ -1,19 +1,21 @@
...
 } from '../../postgres/queries/generated/updateTeamByTeamIdQuery'
 import getPg from '../../postgres/getPg'
 
 const updateTeamByTeamId = async (
   teamUpdates: Partial<IUpdateTeamByTeamIdQueryParams>,
-  teamId: string
-) => {
-  await catchAndLog(() =>
+  teamIds: string | string[]
+): Promise<IUpdateTeamByTeamIdQueryResult[]> => {
```
mattkrick, line 12: Let's have some fun with typescript!
Declaring the return type can be brittle because if the underlying query changed (like it did here) you have to update it. There's also the possibility that this will return `Promise<null>` if the write to PG fails.

Instead, let's let TS do the work for us by rewriting catchAndLog:

```ts
const catchAndLog = async <T extends (...args: any[]) => any>(riskyFn: T) => {
  try {
    return (await riskyFn()) as ReturnType<T>
  } catch (e) {
    sendToSentry(e, {tags: {migration: 'postgres'}})
  }
  return null
}
```

with that change, you can get rid declaring the return type here & everywhere else :tada:

packages/server/postgres/queries/updateUser.ts
```diff
@@ -10,9 +14,9 @@ interface UpdateUserQueryParams extends Omit<IUpdateUserQueryParams, 'identities
 const updateUser = async (
   update: Partial<UpdateUserQueryParams>,
   userIds: string | string[]
-): Promise<void> => {
+): Promise<IUpdateUserQueryResult[]> => {
```
mattkrick, line 17: can remove after updating catchAndLog

packages/server/postgres/utils/checkEqBase.ts
```diff
@@ -0,0 +1,164 @@
...
+  const errors: IError = {
+    rethinkRecordsCompared: 0,
+    foundErrors: 0,
+    missingPgRows: [],
+    extraPgRows: []
+  }
+  const batchSize = 3000
+
+  for (let i = 0; i < 1e5; i++) {
+    if (errors.foundErrors === maxErrors) {
```
mattkrick, line 112: nit: i always try to write `>=` in these cases because gremlins will creep into my code, cause a float error & increment by 1.00000001 when i least expect it (doesn't seem to be the case here)

packages/server/postgres/utils/checkEqBase.ts
```diff
@@ -0,0 +1,164 @@
...
+    pgNeFields[neField] = pgRow[neField]
+  }
+  errors[rowId] = {
+    error: neFields,
+    rethinkRow: rethinkNeFields,
+    pgRow: pgNeFields
+  }
+}
+
+const getExtraPgRowIds = async (tableName: string): Promise<string[]> => {
```
mattkrick, line 80: not sure i understand this part!

making sure rethinkdb & PG have the same number of rows is probably the final test (after making sure there are no missing rows in pg & all the checked rows are equal). if any of those other errors are present, we don't need to check equality.


if that's the case, could we just write `SELECT COUNT(*)` for PG & `r.table('Team').count()` for rethinkdb & compare the number? if we do that could we remove the `eqChecked` column?

mattkrick (changes requested):
looking real good! a couple typescript things & 1 question on the eqChecked date. I'm betting this'll ship next week!

packages/server/graphql/intranetSchema/mutations/checkRethinkPgEquality.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+    const fileName = `${tableName}-${new Date()}`
+    const fileLocation = path.join(
+      process.cwd(),
+      'packages/server/graphql/intranetSchema/mutations',
+      '__output__',
+      fileName
+    )
+    await fs.promises.mkdir(path.dirname(fileLocation), {recursive: true})
+    checkEqAndWriteOutput(tableName, fileLocation, maxErrors)
```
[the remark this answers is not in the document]
tiffanyhan: yes it's possible this fn while check the entire table and it will almost certainly time out. in that case, we have the option of setting `writeToFile=true` and the returned graphql msg will tell the caller where to find the file.

packages/server/postgres/utils/checkEqBase.ts
tiffanyhan: we could just do an aggregate count on both tables. what i was trying to avoid was a scenario in which we happened to have more pg rows than rethink rows, and then it's up to the caller to figure out what these extra rows are from there.

having the extra field is more efficient in terms of time and i think it's a more complete solution to return any possible extra pg rows straight away, but i also agree it may be premature optimization.

packages/server/graphql/mutations/deleteUser.ts
```diff
@@ -74,20 +74,14 @@ export default {
       })
     }
     // do this after 30 seconds so any segment API calls can still get the email
+    const update = {
+      isRemoved: true,
+      email: 'DELETED',
```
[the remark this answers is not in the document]
tiffanyhan: yes, we should definitely have unique constraint on the email field. however: i came to the conclusion (and thought you agreed, but it may have been miscommunicated) that using pg constraints upfront made the equality checking part of the migration process more complicated than it needs to be. i think we should leave pg constraints off until we've verified that both tables are equal.

i also like tabling the constraint part of it until later, because it forces us to be more deliberate about why we want to migrate to postgres. it seems like the primary objective is stability and performance, and better data integrity is also very important but secondary.

packages/server/graphql/intranetSchema/mutations/flagOverLimit.ts
```diff
@@ -29,6 +30,7 @@ const flagOverLimit = {
     if (organizationUsers.length === 0) return {error: {message: 'OrgId has no members'}}
 
     // RESOLUTION
+    copy = copy.trim().slice(0, USER_OVERLIMIT_COPY_LIMIT)
```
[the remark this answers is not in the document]
tiffanyhan: yeah good point!

packages/server/postgres/queries/updateTeamByTeamId.ts
tiffanyhan: oh that's a good point. my thinking was that explicit is usually better, but i really like the suggestion of going with implicit return types to offload all of that upkeep to typescript.

packages/server/postgres/migrations/postDeploy/1619383917863_updateDeletedUserEmails.ts
tiffanyhan: oooh that's a good catch. i think i was going back and forth on the implementation of the mapping fn, but since this one returns equal length it is a lot more expressive to use map

packages/server/postgres/utils/checkEqBase.ts
tiffanyhan: yea this is a good point

packages/server/graphql/mutations/deleteUser.ts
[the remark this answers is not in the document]
mattkrick: ahhh, i was thinking we were only treating foreign keys as strings until both tables were ready in pg. 
sure, let's the email field for the next PR, but definitely should get it in before we remove the write to rethinkdb.

packages/server/postgres/migrations/1619586448010_addEqCheckedFieldToUserAndTeam.ts
```diff
@@ -0,0 +1,27 @@
...
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.sql(`
+    ALTER TABLE "User"
+      ADD COLUMN "eqChecked" TIMESTAMPTZ NOT NULL DEFAULT '-infinity'::timestamptz;
+  
+    ALTER TABLE "Team"
+      ADD COLUMN "eqChecked" TIMESTAMPTZ NOT NULL DEFAULT '-infinity'::timestamptz;
+  
+    CREATE INDEX IF NOT EXISTS "idx_User_eqChecked" ON "User"("eqChecked");
+    CREATE INDEX IF NOT EXISTS "idx_Team_eqChecked" ON "User"("eqChecked");
```
mattkrick, line 15: AH! looks like I missed this. (Should be "Team"?)
With the new plan of attack, let's just remove this migration & we can get by with a `COUNT`

packages/server/postgres/utils/checkEqBase.ts
mattkrick: ahhh i see!
performance is definitely important, but since we'll only run these queries a handful of times, I'd opt for a pattern that we can repeat as quickly as possible for the remaining tables over query performance. 

If I understand the tradeoff, if we _just_ do a count, we could avoid creating & destroying an index & the query would only have to read instead of upsert. On the negative, if we wanted to see which rows only existed in PG, we'd have fetch all the ids in rethinkdb and perform a `NOT IN (...)` scan that would approach O(n^2).

for the sake of speed, I'd say we:
- remove the eqChecked index
- convert the upsert to a read query
- perform a count on both DBs & if they are different, add an error line that shows the difference

if we do come across that dreadful case where we have more docs in pg than rethinkdb we can always write a new PR!

tiffanyhan: > I'd opt for a pattern that we can repeat as quickly as possible for the remaining tables over query performance.

yeah that's a heuristic i wasn't thinking of in terms of tradeoffs, i see what you mean

tiffanyhan: done!

mattkrick:
what a wild ride! LGTM :tada:  
haven't tested it with prod data yet, let's cut it loose on staging & see what happens!

PR #4862, opened 2021-04-06
fix a sentry bug & log another

mattkrick:
@tianrunhe

### DIFF
--- a/packages/client/components/MeetingSidebarTeamMemberStageItems.tsx
+++ b/packages/client/components/MeetingSidebarTeamMemberStageItems.tsx
@@ -1,4 +1,5 @@
 import styled from '@emotion/styled'
+import * as Sentry from '@sentry/browser'
 import graphql from 'babel-plugin-relay/macro'
 import React from 'react'
 import {createFragmentContainer} from 'react-relay'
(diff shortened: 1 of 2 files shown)


packages/client/components/MeetingSidebarTeamMemberStageItems.tsx
```diff
@@ -108,6 +118,7 @@ graphql`
 export default createFragmentContainer(MeetingSidebarTeamMemberStageItems, {
   meeting: graphql`
     fragment MeetingSidebarTeamMemberStageItems_meeting on NewMeeting {
+      id
```
tianrunhe, line 121: duplicate `id` field

packages/client/components/ViewerNotOnTeam.tsx
```diff
@@ -77,10 +77,10 @@ const ViewerNotOnTeam = (props: Props) => {
 export default createFragmentContainer(ViewerNotOnTeam, {
   viewer: graphql`
     fragment ViewerNotOnTeam_viewer on User {
-      meeting(meetingId: $meetingId) {
+      meeting(meetingId: $meetingId) @include(if: $meetingId) {
```
tianrunhe, line 80: Looks like we need a boolean after `if`

mattkrick: ah that's fine, just create a boolean in the top query based off the presence of meetingId?

Issue #4830, opened 2021-03-29
Fetch list of GitHub repositories for user


mattkrick:
As we build our GitHub poker integration, it's time we build some strong patterns. 
The first pattern is how to fetch data from GitHub using their graphql API. we've done this in the past using graphql-typed (https://github.com/ParabolInc/parabol/blob/ebf3a3c55be197cf7c43d9cc4e3c372246317d0b/packages/client/utils/GitHubManager.ts#L6) but that was years ago & there are better ways of doing it (ie graphql-codegen)

Some sources of inspiration:
- https://twitter.com/ReardenQL/status/1371907617929453571
- https://benlimmer.com/2020/05/16/adding-typescript-types-github-graphql-api/

AC
- [x] Regard `GitHubManager#getRepos` as deprecated. This method uses the old way & we'll remove it in a subsequent PR
- [ ] Add a `getRepositories` method to https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/server/utils/GitHubServerManager.ts#L11 that fetches all the repositories for the github user (see https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/client/utils/githubQueries/getRepos.graphql#L0-L1 for inspiration) no need to worry about pagination for now.
- [ ]  Using the sources of inspiration for guidance, vscode's typescript intellisense should offer autocomplete suggestions when writing the graphql query for the github API. e.g. when i type `viewer {` it should suggest `login` as a possible field to select. Ideally, this query could be written in a `.ts` file, but keeping it in a separate `.graphql` field is acceptable if the first isn't easy.
- [ ] the return type of `GitHubServer#getRepositories` should return a strongly typed result based on the graphql query (use graphql-codegen)

Estimate: 16 hours

PR #4870, opened 2021-04-08
Fetch GitHub repos

nickoferrall:
To resolve issue #4830 

This PR adds a `getRepositories` method to the `GitHubServerManager` and uses graphql-codegen to auto-generate types ensuring strongly typed results.

For intellisense, I think our best option is this [GraphQL](https://marketplace.visualstudio.com/items?itemName=GraphQL.vscode-graphql) VSCode extension. I checked out a couple of other extensions and [GraphQL-Zeus](https://github.com/graphql-editor/graphql-zeus) but I think it's best to avoid using an additional package when the extension works well for intellisense:

gql-extension

If you're happy with this PR, I could implement / create an issue to remove the `GitHubManager` file and move all of the methods to the server.


scripts/dev.js
```diff
@@ -38,11 +39,13 @@ const dev = async (maybeInit) => {
   const isInit = !fs.existsSync(path.join(TOOLBOX_ROOT, 'updateSchema.js')) || maybeInit
   const redis = new Redis(process.env.REDIS_URL)
   const toolboxPromise = compileToolbox()
+  exec('yarn graphql-codegen', (err) => {
```
nickoferrall, line 42: I wasn't too sure what the best approach would be here. While it works, it seems inconsistent with the rest of the file. Also, should this be included in the staging and prod builds too?

mattkrick: dig into the graphql-codegen package & you should be able to figure out what function the bin is calling. then, you can call that directly

packages/server/utils/githubQueries/getRepositories.ts
```diff
@@ -0,0 +1,29 @@
+import gql from 'graphql-tag'
```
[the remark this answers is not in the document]
nickoferrall: As in a `.graphql` file? The issue mentioned that the query would ideally be written in a TypeScript file so I went with this approach but I don't think we get any additional benefits from doing this

scripts/dev.js
nickoferrall: updated!

packages/server/utils/githubQueries/getRepositories.ts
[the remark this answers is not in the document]
mattkrick: ah no, graphql-tag turns a string into an AST & then down below you're calling print to turn that AST back into a string.

can we build this in such a way that gets the benefits of intellisense without the runtime penalty of going from string -> AST -> string?

scripts/dev.js
```diff
@@ -13,6 +13,8 @@ const PROJECT_ROOT = getProjectRoot()
 const TOOLBOX_ROOT = path.join(PROJECT_ROOT, 'scripts', 'toolbox')
 const pgMigrate = require('node-pg-migrate').default
 const cliPgmConfig = require('../packages/server/postgres/pgmConfig')
+const {generate} = require('@graphql-codegen/cli')
+const yaml = require('js-yaml')
```
mattkrick, line 17: we don't have js-yaml installed in the app, we'll either need to install it (if you want to keep the yaml) or we should build the config as a JS object (might be cleaner?). the reason this works today is because we have a dependency that already includes it, but if that dependency is removed, then this call will mysteriously break.

we'll also need a way to call this from CI. perhaps add the same call to prod.js?

mattkrick (changes requested):
couldn't get intellisense to work out of the box, might be worth adding a how-to in the README so new devs can get it up & running!

packages/server/utils/GitHubServerManager.ts
```diff
@@ -1,13 +1,24 @@
...
   access_token: string
   error: any
   scope: string
 }
 
+interface GQLResponse<TData> {
+  data?: TData
+  errors?: any[]
+}
+interface GitHubCredentialError {
```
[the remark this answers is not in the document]
nickoferrall: I copied this from the `GitHubManager` file - I have found a reference to this in the [GitHub docs](https://docs.github.com/en/developers/apps/authenticating-with-github-apps#authenticating-as-a-github-app) but it's not referenced as a type in the [schema](https://docs.github.com/en/graphql/overview/public-schema). I've added a comment

packages/server/utils/GitHubServerManager.ts
```diff
@@ -1,13 +1,24 @@
...
+import {getRepositories} from './githubQueries/getRepositories'
+import {GetRepositoriesQuery} from '../../server/types/typed-document-nodes'
 
 interface OAuth2Response {
   access_token: string
   error: any
   scope: string
 }
 
+interface GQLResponse<TData> {
```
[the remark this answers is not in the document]
nickoferrall: Strangely, the [additional type](https://www.npmjs.com/package/@octokit/graphql#additional-types) exported by octokit is just: 
```
export declare type GraphQlQueryResponseData = {
    [key: string]: any;
};
```
and the `GraphQlQueryResponse` is not exposed. Instead, I've copied the `GraphQlQueryResponse` from the package  into this file to give us extra type safety with errors.

.graphqlconfig
```diff
@@ -15,7 +15,8 @@
       "schemaPath": "./packages/server/utils/githubSchema.graphql",
       "name": "GitHub GraphQL Schema",
       "includes": [
-        "./packages/client/utils/githubQueries/*.graphql"
+        "./packages/client/utils/githubQueries/*.graphql",
+        "./packages/server/utils/githubQueries/*"
```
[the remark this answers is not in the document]
nickoferrall: Ah I got it working with a `graphql.config.js` file too - I'm not sure why it wasn't committed. Sorry for sending you on a wild goose chase! I've added a section to the Readme where we can continue to add VSCode extensions that we think could be helpful

packages/server/package.json
```diff
@@ -18,7 +18,7 @@
     "precommit": "lint-staged",
     "lint": "eslint --fix ./**/*.ts",
     "postversion": "git push --follow-tags",
-    "tsgen:gh": "gql2ts utils/githubSchema.json -e $(pwd)/../client/types/configGH.js -o packages/client/types/githubGraphql.d.ts && eslint --fix 'packages/client/types/githubGraphql.d.ts'",
+    "tsgen:gh": "gql2ts utils/githubSchema.ts -e $(pwd)/../client/types/configGH.js -o packages/client/types/githubGraphql.d.ts && eslint --fix 'packages/client/types/githubGraphql.d.ts'",
```
mattkrick, line 21: i think we can remove this entirely now, but i'll scope that for another issue

packages/server/utils/updateGitHubSchema.ts
```diff
@@ -4,12 +4,12 @@ import fetch from 'node-fetch'
 
 const updateGitHubSchema = async () => {
   const schemaPath = './githubSchema.graphql'
-  const schemaJSONPath = './githubSchema.json'
+  const schemaJSONPath = './githubSchema.ts'
```
mattkrick, line 7: i think we can delete this entire file, too. this was for the old way of doing things

Issue #4811, opened 2021-03-24
Teammate can kick you out of Poker


nickoferrall:
Steps to reproduce:

- User A and User B go to a Story in the Estimate Phase
- User A goes back to the Scope phase and removes the Story that User B is currently on
- User B will be kicked out of the meeting and won't be able to go back to the meeting unless they change the URL from the Estimate to the Scope phase

### AC

- [ ] If the active Story in the Estimate Phase is removed, users are moved to the previous stage in the Estimate Phase

jordanh:
This is a good one

PR #4873, opened 2021-04-09
Fix tator being kicked out of Poker

nickoferrall:
PR to resolve issue #4811 

Steps to reproduce:

- User A, the facilitator, goes to a Story in the Estimate Phase
- User B goes back to the Scope phase and removes the Story that User A is currently on
- User A will be kicked out of the meeting and won't be able to go back to the meeting unless they change the phase in the URL

One solution would be to prevent users from removing Stories that are active in the Estimate Phase but I think users should have that power and we should accommodate.

The solution in this PR doesn't update the `facilitatorStageId`. I think that's OK as the local stage updates and the `facilitatorStageId` will change when the facilitator makes their next move. If that seems problematic, I could update the `facilitatorStageId` in `updatePokerScope`. 

### AC

- [ ] Follow steps to reproduce but instead of being kicked out the meeting, the facilitator goes back to the previous stage

### DIFF
--- a/packages/client/mutations/UpdatePokerScopeMutation.ts
+++ b/packages/client/mutations/UpdatePokerScopeMutation.ts
@@ -12,6 +12,7 @@ import {
 graphql`
   fragment UpdatePokerScopeMutation_meeting on UpdatePokerScopeSuccess {
     meeting {
+      facilitatorStageId
       phases {
         ...useMakeStageSummaries_phase
         ... on EstimatePhase {
--- a/packages/client/hooks/useUpdatedSafeRoute.ts
+++ b/packages/client/hooks/useUpdatedSafeRoute.ts
@@ -2,7 +2,6 @@ import graphql from 'babel-plugin-relay/macro'
 import {Dispatch, SetStateAction, useEffect, useRef} from 'react'
 import {readInlineData} from 'relay-runtime'
 import {useUpdatedSafeRoute_meeting} from '~/__generated__/useUpdatedSafeRoute_meeting.graphql'
-import findStageBeforeId from '../utils/meetings/findStageBeforeId'
 import findStageById from '../utils/meetings/findStageById'
 import fromStageIdToUrl from '../utils/meetings/fromStageIdToUrl'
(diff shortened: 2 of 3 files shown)


mattkrick:
great find! it looks like this PR gracefully handles the case where facilitatorStageId is invalid. 
This is a good approach! But if the facilitatorStageId could be invalid, then we'll have to make sure we perform a similar check every time the facilitatorStageId is referenced.

A cleaner approach might be to guarantee that facilitatorStageId is valid, and as you mentioned, we can do this by updating it in the DB in `updatePokerScope`.

We do something similar when a team member or agenda item gets removed from a meeting. Since `updatePokerScope` already handles most of that stuff, I think it should be as simple as:

- in `updatePokerScope`, make a log of all the stageIds that are going to be removed.
- check if `meeting.facilitatorStageId` is one of the `removedStageIds`
- if so, calculate what the next stage should be & update the stage start time & the meeting.facilitatorStageId. An example of that logic is here: https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/server/graphql/mutations/helpers/removeStagesFromMeetings.ts#L27

packages/client/mutations/UpdatePokerScopeMutation.ts
```diff
@@ -12,6 +12,7 @@ import {
 graphql`
   fragment UpdatePokerScopeMutation_meeting on UpdatePokerScopeSuccess {
     meeting {
+      facilitatorStageId
```
nickoferrall, line 15: Am I right in saying that by including `facilitatorStageId` in this query, we let Relay know that it's being updated in the mutation and should be re-fetched? Even though we're not using it in the optimistic updater

mattkrick: that is correct! for moving between stages I don't much bother with optimistic updates because sometimes we have to build the next stage (e.g. between voting & discuss).

mattkrick:
LGTM! nice work

PR #4877, opened 2021-04-11
Fix team invitation

nickoferrall:
PR to resolve issue #4744 

Current problems:

1. If a user tries to join a team, the team members in that team may see a React Hooks error.
2. The `team` and `meeting` queries in `ViewerNotOnTeam` are invalid as `teamId` and `meetingId` can be null.

Solutions:

1. This is being caused by `PushInvitationMutation` breaking the rules of hooks which explains why the bug only happens in development and not production.
2. A bug in the Relay compiler prevented it from detecting the error (if we use a regular query instead of a fragment, it does detect the error). It looks like we don't need to check whether `team` or `meeting` exist in `ViewerNotOnTeam` as [MeetingSelector](https://github.com/ParabolInc/parabol/blob/235b57d1790f74b6170f8c3ebd5acf30839a472c/packages/client/components/MeetingSelector.tsx#L34) and [TeamContainer](https://github.com/ParabolInc/parabol/blob/f9693de81dd203521b9d1862b744c186936d636c/packages/client/modules/teamDashboard/containers/Team/TeamContainer.tsx#L34) already do this for us. If they don't exist, the viewer will be sent back to `ViewerNotOnTeam`.

### AC

- [ ] If a user tries to join a team, the team members of that team don't see an error in development
- [ ] If a user tries to join a team, the server doesn't complain that the query is invalid
- [ ] A user can successfully join a team by going to `/team`, an active meeting or using the invitation link

PR #4878, opened 2021-04-12
fix push invitation

mattkrick:
my bad for letting this one slip through.
FYI @nickoferrall can't use hooks outside of react components


nickoferrall:
Thanks, my mistake. I spotted this in yesterday's PR too: https://github.com/ParabolInc/parabol/pull/4877

mattkrick:
ah, yours is better! will merge

Issue #4829, opened 2021-03-29
X is editing too... bug


mattkrick:
Repro:

- Go to team kanban
- filter by another team member (not you)
- add a task
- see that it says "X is editing too..."

AC
- [ ] when creating a new task for someone else in a filtered kanban view, it should say "Editing..." instead of "X is editing too..."

Estimate: 4 hours

PR #4879, opened 2021-04-12
Fix editing too bug

nickoferrall:
PR to resolve issue #4829 

Repro the issue:

- Go to team kanban
- filter by another team member (not you)
- add a task
- see that it says "X is editing too..."

The `userId` field in `task` tells us who owns the task but the task owner can be different from the task editor. Using the `viewer` instead of the task `user` resolves this bug.

### AC

- [ ] Filter by another team member, add a task and see "Editing..." rather than "X is editing too..."


ackernaut:
### Test
- [ ] Filter by another team member, add a task and see "Editing..." rather than "X is editing too..."

ackernaut (approved):
LGTM!

Issue #4788, opened 2021-03-19
Add Story Points to required screen from within parabol


mattkrick:
Today, if the user tries to update their story points from a poker meeting, it's possible that they hit an error that asks them to go into Jira & fix the problem.
With Jira's wonderful UI, sometimes it even looks like it should be working from within Jira!
It'd be better if we could do this automatically.
In order to do so, we must have the Oauth scope `manage:jira-project`

There are 2 ways to accomplish this:
- We include this scope from the get go when someone integrates
- When the user encounters this error, the client requests the additional scope

The first is easy, the 2nd is more respectful wrt permissions. Worth noting that the 2nd would also provide a nice pattern to follow for folks who have already integrated, so let's go with the 2nd option.

AC
- [ ] When Jira returns an error that the desired field is not on the appropriate screen (https://github.com/ParabolInc/parabol/blob/8e44b32c0002e44f21e241dc59b3bacc36eea10f/packages/client/utils/AtlassianManager.ts#L610) then the client pops open a modal that tells the user the problem
- [ ] That modal has 2 options: "Fix it for me" and "nevermind"
- [ ] If the user clicks "Fix it for me" we check the current Jira scope. if it doesn't include "manage:jira-project" then we request that scope via oauth from the user
- [ ] if the user has the necessary scope, then we add the desired field to the necessary screen(s) in jira. It may be as easy as adding it to the default screen (https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-screens/#api-rest-api-3-screens-addtodefault-fieldid-post) but if that doesn't work, you may have to sniff out what the jira app does & copy that REST call.
- [ ] 
Estimate: 20 hours

mattkrick:
This will be the BP for @BartoszJarocki

PR #4880, opened 2021-04-12
Add Story Points to required screen from within Parabol

BartoszJarocki:
Related issue: https://github.com/ParabolInc/parabol/issues/4788 

## Changes
1. I've added a new mutation which is supposed to add a missing Jira field to a screen used by the Jira project
2. I've added a new modal to handle the missing field error
3. I've changed a bit the way Atlassian OAuth is called to be able to listen to a completion

## Things to consider
1. I haven't found the way to request only additional scope from Atlassian (when user clicks `Fix it for me`) so I'm requesting all the scopes + `jira:manage-project`. Also, already requested scopes probably should be stored in DB.

### Adding missing Jira field to a screen configuration
Seems like there’s no way to fetch the screen configuration for a project/issue.

With current Jira REST API the flow should be the following
1. Get screen configuration for a project/issue to get the screen ids which need to be updated (missing)
2. Get screen tabs: https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-screen-tabs/#api-rest-api-3-screens-screenid-tabs-get 
3. Add screen tab field: https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-screen-tab-fields/#api-rest-api-3-screens-screenid-tabs-tabid-fields-post 

What I’ve tried:
1. Checking how Jira app gets the screens for a project. Unfortunately, Jira app doesn’t use REST API, for example: https://bjarocki.atlassian.net/plugins/servlet/project-config/PAR/screens is server side rendered so it’s not useful 
2. I thought maybe every project created will have a default screen with ID 10000 but it’s not the case 
3. Googling, seems like there’s no way of currently doing that
4. As a last resort I can ask a question on Atlassian Developer Community unless someone has a better idea

## How it looks like?
CleanShot 2021-04-12 at 19 02 17


mattkrick (changes requested):
This is a great first pass!
Assuming there's no way to set the issue on a screen, what should we do in order to help users out?

packages/client/components/PokerDimensionValueControl.tsx
```diff
@@ -76,9 +78,42 @@ interface Props {
...
 
+const useHandleFinalScoreError = ({error, serviceField, stageId}) => {
+  const atmosphere = useAtmosphere()
+  const {closePortal, openPortal, modalPortal} = useModal()
+
+  useEffect(() => {
+    const nextError = error?.message ?? ''
+
+    if (
+      nextError &&
```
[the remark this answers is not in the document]
BartoszJarocki: Right, fixed 👍

packages/client/components/AddMissingJiraFieldModal.tsx
```diff
@@ -0,0 +1,164 @@
...
+      AddAtlassianAuthMutation(
+        atmosphere,
+        {code, teamId},
+        {onError: mutationProps.onError, onCompleted: onAddAtlassianAuthCompleted}
+      )
+    },
+    [mutationProps.submitMutation, teamId, onAddAtlassianAuthCompleted]
+  )
+
+  const askForManageProjectPermission = useCallback(() => {
```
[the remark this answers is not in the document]
BartoszJarocki: I tend to use `useCallback` because if the variable is recreated on every render and gets passed somewhere in a dependency array it's causing a hook to fire every single time. I agree though, it looks way better this way and it can be safely done that way. TIL about V8 and the way it behaves with closures. Interesting! 

Fixed. 🙏

packages/client/components/PokerDimensionValueControl.tsx
```diff
@@ -201,6 +247,15 @@ const PokerDimensionValueControl = (props: Props) => {
           </>
         )}
       </Control>
+      {addMissingJiraFieldModalPortal(
+        <AddMissingJiraFieldModal
+          finalScore={finalScore}
```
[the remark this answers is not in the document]
BartoszJarocki: Actually, this was a bad choice as the majority of this data can be managed by Relay. I've changed it to look this way
```
        <AddMissingJiraFieldModal
          pendingScore={pendingScore}
          stage={stage}
          closePortal={closeAddMissingJiraFieldModal}
        />
```
where `stage` is just a Relay fragment
```
  stage: graphql`
    fragment AddMissingJiraFieldModal_stage on EstimateStage {
      id
      meetingId
      teamId
    }
  `
```

Regarding object vs primitives, there's no easy answer here. I personally think it's better to pass objects when the number of props starts growing (and group the props if possible) just to make the code more readable. When passing objects which will be used as hooks dependencies it's critical to remember to _not_ recreate them on every render because it might cause infinite loops. It's not the case when using Relay 🤗

packages/client/components/PokerDimensionValueControl.tsx
```diff
@@ -90,19 +125,30 @@ const PokerDimensionValueControl = (props: Props) => {
...
+
   useEffect(() => {
     if (error) {
       // we want this for remote errors but not local errors, so we keep the 2 in different vars
       setPendingScore(finalScore)
     }
   }, [error])
 
+  useEffect(() => {
+    if (finalScore) {
```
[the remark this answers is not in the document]
BartoszJarocki: Good point. I was thinking about this scenario but it's hard to decide what is better. There are two options:
1. Close the modal after `AddMissingJiraFieldMutation` finishes successfully, reuse the `submitScore` from `PokerDimensionValueControl` and let that component handle the errors of `PokerSetFinalScoreMutation` 
2. Show the errors within `AddMissingJiraFieldModal` the way it's done now

Now, as we're talking about it I think it's better to do 1. because `PokerSetFinalScoreMutation` are not really related to the `AddMissingJiraFieldsModal`. Any thoughts @mattkrick?

packages/client/components/PokerDimensionValueControl.tsx
```diff
@@ -76,9 +78,42 @@ interface Props {
...
+  useEffect(() => {
+    const nextError = error?.message ?? ''
+
+    if (
+      nextError &&
+      nextError.includes(
+        `Update failed! In Jira, add the field "${serviceField.name}" to the Issue screen.`
+      )
+    ) {
+      openPortal()
```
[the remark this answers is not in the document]
BartoszJarocki: With the current solution, if the same error happens after attempting to add Jira field it won't show the modal because `useEffect` will not fire (error string, the dependency will be the same).

If you're asking in general about _useEffect_ here vs callback, the _useEffect_ might not be the best choice if `openPortal` is mutating the dom (and now that I see, it is) and the user might see some flickering as it doesn't block the browser painting. In this case, `useLayoutEffect` might be a better choice. I'm not an expert there so correct me If I'm wrong, please.

If we let the `PokerDimensionValueControl` component handle the errors as I described in the comment below the best way would be to put it in `onCompleted` handler of `PokerSetFinalScoreMutation` like that:
```
    PokerSetFinalScoreMutation(
      atmosphere,
      {finalScore: pendingScore, meetingId, stageId},
      {
        onError,
        onCompleted: (res, errors) => {
          onCompleted(res, errors)

          const mutationError = res.pokerSetFinalScore?.error?.message ?? ''
          if (
            mutationError.includes(
              `Update failed! In Jira, add the field "${serviceField.name}" to the Issue screen.`
            )
          ) {
          }
        }
      }
    )
```

BartoszJarocki:
@mattkrick if we won't be able to add a Jira field programmatically the best way to help users would be to add some wizard with exact steps (some gifs, video, or maybe just illustrations) how to do it. It could be something like https://material-ui.com/components/steppers/ with steps:
1. Open your Jira project settings page (video, gif, link if possible)
2. Go to the Screen settings (video, gif, link if possible)
3. Add a given field to this screen, hit save (video, gif, link if possible)
4. Once that done, hit retry (when retry clicked, attempt to set the value again)

That'd be the best if we can't do it automatically.

packages/client/components/AddMissingJiraFieldModal.tsx
```diff
@@ -0,0 +1,163 @@
...
+    AtlassianClientManager.openOAuth(onAtlassianOAuthCompleted, AtlassianManager.MANAGE_SCOPE)
+  }
+
+  return (
+    <StyledDialogContainer>
+      <DialogTitle>{'Oops!'}</DialogTitle>
+      <DialogContent>
+        <div>
+          <StyledTip>
+            {'You do not have this field configured in Jira, do you want us to fix add it for you?'}
```
tianrunhe, line 121: typo: "fix add"

packages/server/graphql/mutations/addMissingJiraField.ts
```diff
@@ -0,0 +1,116 @@
...
+    if (!isTeamMember(authToken, teamId)) {
+      return {error: {message: 'Not on the team'}}
+    }
+    if (endedAt) {
+      return {error: {message: 'Meeting has ended'}}
+    }
+    if (meetingType !== 'poker') {
+      return {error: {message: 'Not a poker meeting'}}
+    }
+    if (isPhaseComplete('ESTIMATE', phases)) {
```
tianrunhe, line 58: Nit: if I pass in the `stageId` for the SCOPE phase, the error message seems confusing as I haven't started the ESTIMATE phase, how can it already be completed?

packages/server/graphql/mutations/addMissingJiraField.ts
```diff
@@ -0,0 +1,116 @@
...
+      return {error: {message: 'Estimate phase is already complete'}}
+    }
+    if (viewerId !== facilitatorUserId) {
+      if (viewerId !== createdBy) {
```
tianrunhe, line 62: During any time of the meeting, anyone can opt in to be the facilitator, so at ESTIMATE phase the facilitator can be different from `createdBy`. We don't need to error out on that case I think.

packages/client/mutations/AddMissingJiraFieldMutation.ts
```diff
@@ -0,0 +1,35 @@
+import graphql from 'babel-plugin-relay/macro'
+import {commitMutation} from 'react-relay'
+import {StandardMutation} from '../types/relayMutations'
+import {AddMissingJiraFieldMutation as TAddMissingJiraFieldMutation} from '../__generated__/AddMissingJiraFieldMutation.graphql'
+
+graphql`
+  fragment AddMissingJiraFieldMutation_part on AddMissingJiraFieldSuccess {
```
tianrunhe, line 7: A pattern we use in the code is to suffix the name of a fragment with the name of the it's resolved field. So in this case it should be `AddMissingJiraFieldMutation_dimensionField`

tianrunhe (commented):
Providing some tiny comments in addition to Matt's here.

Also, a feature improvement suggestion:
Can we choose the site for the user on "Fix it for me" modal pop up? We already know which site the Jira issue comes from, we should save the user one less step from choosing the site:
Screen Shot 2021-04-16 at 11 17 46 AM

packages/client/mutations/AddMissingJiraFieldMutation.ts
BartoszJarocki: Oh, I see. Will fix it.

packages/client/components/AddMissingJiraFieldModal.tsx
BartoszJarocki: Fixed.

packages/server/graphql/mutations/addMissingJiraField.ts
BartoszJarocki: I'm not sure about the exact flow but I took the validation steps from `pokerSetFinalScore` which I believe should be similar. I'm happy to improve if that's not needed though.

packages/server/graphql/mutations/addMissingJiraField.ts
BartoszJarocki: See the comment above 🙏

BartoszJarocki:
@tianrunhe Thank for the review! Regarding choosing the site I haven't found the way to do this. All possible parameters are shown here: https://developer.atlassian.com/cloud/jira/platform/oauth-2-3lo-apps/#1--direct-the-user-to-the-authorization-url-to-get-an-authorization-code. I tried to pass site param different ways and it's not doing anything, unfortunately.

packages/client/components/PokerDimensionValueControl.tsx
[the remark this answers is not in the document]
mattkrick: beautifully explained!

Functionally, there's no real difference between putting the logic in a hook vs. the callback. My reason for preferring the `openPortal` logic inside the callback is because it's closer to the triggering event, which makes it easier to debug. Also fun, while it doesn't much matter here, it could save us a re-render because it's setting the error & opening the portal in the same call. 

`useLayoutEffect` isn't necessary because the modal is going to take over the entire screen, so we don't need to do any DOM measurements, but you're correct if we were going to do something that was going to both move an element & pop a modal based off the element position.

packages/client/components/PokerDimensionValueControl.tsx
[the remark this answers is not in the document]
mattkrick: i like option 1! Always nice to re-use patterns as much as possible

packages/client/components/AddMissingJiraFieldModal.tsx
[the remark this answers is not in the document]
mattkrick: true true! if performance ever becomes an issue we could write `useEventCallback(onClick)`, which will guarantee a persisted object and only use 1 hook!

packages/client/mutations/AddMissingJiraFieldMutation.ts
mattkrick: that's true for queries! for mutations, the suffix is the name of the subscription channel (that way it's easy to spot if we change the channel the subscription is published on: https://github.com/ParabolInc/parabol/blob/0d9a3bde5900aedd8fb49c7437f8694e88b486c1/packages/client/subscriptions/MeetingSubscription.ts)

mattkrick:
Great changes! I'll hold off on merging it until we decide on a good UI for the fallback pattern

BartoszJarocki:
I've asked a question on Atlassian Community, maybe there's something I've missed.
https://community.developer.atlassian.com/t/adding-a-field-to-a-screen-configuration-using-jira-rest-api/47283

enriquesanchez:
@BartoszJarocki @mattkrick 

>if we won't be able to add a Jira field programmatically the best way to help users would be to add some wizard with exact steps 

To clarify, the wizard we are proposing only serves as instructions for the user on what they need to do on Jira to setup the field, correct?

mattkrick:
That's correct! We may be able to piece together a URL to get them to a page close to where they need to do stuff, but that's about it

enriquesanchez:
@mattkrick @BartoszJarocki 

Thank you.

If that's the case then, I'd recommend we keep it simple and avoid a more involved UI. A stepper requires the user to click "next" multiple times and it's just overall more UI to deal with.

I think a short numbered guide could work well here, something like this:

Screen Shot 2021-05-03 at 3 22 55 p m

This could be a modal that appears centered on screen, on top of everything else and has a darkened background. If we can, instead of {name of field} we should display the real name of the field the user is trying to add, that is, if we know it.

I tried following [the steps from this comment above](https://github.com/ParabolInc/parabol/pull/4880#issuecomment-820205553) and had a hard time finding my way around. Jira really is complicated. If I could only get some help with the exact steps and names of menus then I can update the mockup.

>We may be able to piece together a URL to get them to a page close to where they need to do stuff, but that's about it

If this is a possibility then the experience would be much better I think. We could still provide the step by step instructions, but also provide something like a "You can also click here" link. After attempting to find my way around Jira and being unsuccessful, I think a link like that will come in handy.

And lastly, when the user clicks "Done", we should automatically try to retry setting the value for them. One less click we make them go through. Hitting the X (close button) or <kbd>esc</kbd> should cancel and close the modal.

What do y'all think?

BartoszJarocki:
@enriquesanchez I'll play with URLs and see if we can have one for each step. Looks nice!

enriquesanchez:
Thanks @BartoszJarocki!

Once we have more clarity around the URLs and the step-by-step instructions I can create a design issue for implementation.

mattkrick:
closing in favor of designing the workaround since it seems this isn't possible (although if it _is_ possible, we should definitely do this)

BartoszJarocki:
@mattkrick if you have any ideas let me know, please. I'd love to finish that myself 👍

enriquesanchez:
Ah, was just finishing an update to the modal:

Screen Shot 2021-05-12 at 1 50 48 p m

As for the steps, this is as clear as I was able to get them, but man Jira is a mess and I'm not sure it'll be easy for everyone to find their way around. If we can provide a direct URL to the last screen they need to get to, that would be ideal.

1. On Jira, go to your Project Settings and open the Screens settings page
2. Click the Edit button of the Screen Scheme you want to add the {name_of_field} field to
3. Select again the Screen Scheme you wish to configure
4. Use the field picker at the botton of the page to add the new field

@mattkrick what do you think=

ackernaut:
@enriquesanchez the quick steps on the modal are good. It keeps the dialog light and easy. What if we linked to a support doc that showed screenshots for each step with any extra details and a Loom of a Parabol support person at the top of the page demoing how to do it? From the dialog maybe there’s a link ‘watch a video’ on how to add the field? Just a random idea looking at this from outside the working conversation.

mattkrick:
It works!
@BartoszJarocki I'll let you take it from here :wink: 

https://user-images.githubusercontent.com/5514175/118060154-f7b74d80-b346-11eb-9c18-19678bd03283.mp4

BartoszJarocki:
@mattkrick omg, that's neat, thank you! I feel bad I haven't figured it out myself. 
I'll cleanup the server-side code, add some interfaces for Jira API responses and finish that. 

I think the last question is, do we need a fallback mechanism? Technically, the user can rename the screen, and then, if the screen name doesn't start with a project key the solution you've found will not work. 
Screenshot 2021-05-13 at 16 01 02

mattkrick:
there are 2 ways to do it:
- we can request an _even more powerful_ scope & call this endpoint to determine which screen is associated with the given project: https://developer.atlassian.com/cloud/jira/platform/rest/v2/api-group-issue-type-screen-schemes/#api-group-issue-type-screen-schemes. i don't like this approach because the scope is called `manage:jira-configuration` and the only users that can give us that scope are jira cloud admins. we're essentially asking for the root user to give us root access. chances are, the folks using us are _not_ root.
- trial & error:
  - we try to turn it on for the "Project default" screen. if that's not found then...
  - we gather a list of ALL the screens. we sort them so the screens that start with a different project key are on the bottom. then, one by one, we add the field to that screen & then try to update the field. if it works, great! if it doesn't, we undo our change & move onto the next screen.

BartoszJarocki:
@mattkrick I've added the guessing algorithm and it works great. I'd be great if you could take a look and see if there's anything missing. 

About things that might cause issues, I've tested the worst case scenario (I've simply reversed the sorting, and started from the least probable screen) and I'm getting 
```
SEND TO SENTRY Error: Redis took too long to respond
    at Timeout.eval [as _onTimeout] (webpack-internal:///./packages/server/utils/PubSubPromise.ts:53:16)
    at listOnTimeout (internal/timers.js:554:17)
    at processTimers (internal/timers.js:497:7) undefined
```
It's because the whole operation can take a while to finish if our guesses are wrong (some weird edge case but still). So just heads up. 
Screenshot 2021-05-17 at 08 28 16

packages/client/components/PokerDimensionValueControl.tsx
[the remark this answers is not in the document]
BartoszJarocki: @mattkrick for now I've left it in the same place. I didn't like checking 
```
mutationError.includes(
              `Update failed! In Jira, add the field "${serviceField.name}" to the Issue screen.`
            )
```
two times and I couldn't think of a better solution. if you'd rather do it differently let me know, happy to improve!

packages/server/graphql/mutations/addMissingJiraField.ts
```diff
@@ -101,10 +102,58 @@ const addMissingJiraField = {
...
+          cloudId,
+          currentScreen.id
+        )) as JiraScreenTab[]
+        screenTabIds.set(currentScreen, tabId)
+        await manager.addFieldToScreenTab(cloudId, currentScreen.id, tabId, dimensionField.fieldId)
+        // if we can update the field that was previously missing it means we've added it to the right screen
+        await manager.updateStoryPoints(
+          cloudId,
+          issueKey,
+          0, // dummy value
```
mattkrick, line 147: the field in question _could_ take a string instead of a number & then this would bork.
`const {fieldType} = jiraDimensionField` to determine what dummy value to use.

packages/server/graphql/mutations/addMissingJiraField.ts
```diff
@@ -101,10 +102,58 @@ const addMissingJiraField = {
...
+      .map(({screen}) => screen)
+
+    // stores screen tabs so we won't need to refetch it when reverting from previous operation
+    const screenTabIds = new Map()
+    // iterate over all the screens sorted by probability, try to update the given field
+    for (let i = 0; i < possibleScreens.length; i++) {
+      try {
+        const currentScreen = possibleScreens[i]
+        const previousScreen = i > 0 ? possibleScreens[i - 1] : null
+        if (previousScreen) {
```
mattkrick, line 128: we were getting timeouts on this part, so let's run some stuff in parallel. we could also do some batch work (e.g. grab all the tabIds at once), up to you!

```ts
const addFieldToScreen = async () => {
  await manager.getScreenTabs(...)
  await manager.addFieldToScreenTab(...)
}  

await Promise.all([
  previousScreen && manage.removeFieldFromScreenTab(...),
  addFieldToScreen
])
```

packages/server/graphql/mutations/addMissingJiraField.ts
```diff
@@ -101,10 +102,58 @@ const addMissingJiraField = {
...
+          await manager.removeFieldFromScreenTab(
+            cloudId,
+            previousScreen.id,
+            screenTabIds.get(previousScreen),
```
mattkrick, line 132: it's possible that this returns undefined if something went wrong inside jira (maybe a screen has 0 tabs? jira was down for a second? etc). if we can't get a value, let's just ignore this call

packages/server/graphql/mutations/addMissingJiraField.ts
```diff
@@ -101,10 +102,58 @@ const addMissingJiraField = {
...
+            dimensionField.fieldId
+          )
+        }
+
+        const [{id: tabId}] = (await manager.getScreenTabs(
+          cloudId,
+          currentScreen.id
+        )) as JiraScreenTab[]
```
mattkrick, line 140: lets include some error handling here so we don't have to cast
(also, jira throws errors. a lot. for any reason. we have to code _really_ defensively against their API)

packages/server/graphql/mutations/addMissingJiraField.ts
```diff
@@ -101,10 +102,58 @@ const addMissingJiraField = {
...
+        screenTabIds.set(currentScreen, tabId)
+        await manager.addFieldToScreenTab(cloudId, currentScreen.id, tabId, dimensionField.fieldId)
+        // if we can update the field that was previously missing it means we've added it to the right screen
+        await manager.updateStoryPoints(
```
mattkrick, line 144: let's limit the try/catch to just this one. if any of the other async calls throw, we want to know about it!

packages/client/components/PokerDimensionValueControl.tsx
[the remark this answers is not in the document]
mattkrick: i'll PR this PR with an alternate

mattkrick (changes requested):
looking _great_! couple nits just on code style, only big thing is just more defensive coding against JIRA's API. It's a pain trying to catch any & all the errors they throw at us, but that API throws a _ton_

packages/client/components/AddMissingJiraFieldModal.tsx
```diff
@@ -66,33 +65,32 @@ const Label = styled('div')({
   fontWeight: 600
 })
 
-const AddMissingJiraFieldModal = ({stage, pendingScore, closePortal}: Props) => {
+const AddMissingJiraFieldModal = ({stage, closePortal, submitScore}: Props) => {
```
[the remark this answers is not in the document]
BartoszJarocki: got it!

packages/client/components/AddMissingJiraFieldModal.tsx
```diff
@@ -66,33 +65,32 @@ const Label = styled('div')({
...
   const {meetingId, id: stageId, teamId} = stage
   const atmosphere = useAtmosphere()
   const mutationProps = useMutationProps()
 
   const onFixItForMeClicked = () => {
     if (mutationProps.submitting) return
```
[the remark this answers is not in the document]
BartoszJarocki: make sense. fixed 👍

packages/server/graphql/mutations/addMissingJiraField.ts
BartoszJarocki: This check is actually the same as in `pokerSetFinalScore`.  From what I see, the check 
```
      if (viewerId !== createdBy) {
        return {
          error: {message: 'Not meeting facilitator'}
        }
      }
```
is not really needed here as we already know the the viewer isn't the facilitator. I'd remove it (probably from both places, `addMissingJiraField` and `pokerSetFinalScore`), any thoughts @mattkrick @tianrunhe ?

packages/server/graphql/mutations/addMissingJiraField.ts
```diff
@@ -0,0 +1,165 @@
...
+      return {error: {message: 'Invalid stageId provided'}}
+    }
+
+    // RESOLUTION
+    const {creatorUserId, dimensionRefIdx, service, serviceTaskId} = stage
+    const templateRef = await getTemplateRefById(templateRefId)
+    const {dimensions} = templateRef
+    const dimensionRef = dimensions[dimensionRefIdx]
+    const {name: dimensionName} = dimensionRef
+    if ((service as TaskServiceEnum) !== 'jira') {
```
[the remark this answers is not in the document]
BartoszJarocki: done 👍

packages/server/graphql/mutations/addMissingJiraField.ts
```diff
@@ -0,0 +1,165 @@
...
+      return {error: {message: 'Non Jira service'}}
+    }
+    const auth = await dataLoader.get('freshAtlassianAuth').load({teamId, userId: creatorUserId})
```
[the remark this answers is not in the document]
BartoszJarocki: I think it'd be enough to just use viewerId because just before calling this mutation a facilitator was asked to authorize with Jira, any thoughts @mattkrick?

packages/server/graphql/mutations/addMissingJiraField.ts
BartoszJarocki: good point, fixed!

packages/server/graphql/mutations/addMissingJiraField.ts
BartoszJarocki: @mattkrick I've pushed some changes: https://github.com/ParabolInc/parabol/pull/4880/commits/d5376c0258bd9c314eb0b9b9340728166305842c and now it's not failing even in the worst case scenario. The most probable scenario is a bit slower now because I'm fetching all the screen ids now at the beginning. I think it's a good compromise though. I'd appreciate if you could take a look, thanks!

packages/client/utils/AtlassianManager.ts
```diff
@@ -697,8 +755,28 @@ export default abstract class AtlassianManager {
...
+    const res = (await this.delete(
       `https://api.atlassian.com/ex/jira/${cloudId}/rest/api/3/screens/${screenId}/tabs/${tabId}/fields/${fieldId}`
     )) as null | AtlassianError | JiraError
+
+    if (res === null) {
+      return null
+    }
+
+    if (isAtlassianError(res)) {
+      throw new Error(res.message)
```
mattkrick, line 767: if API errors get thrown, then you'll have to wrap any method call in a try/catch and you can't use Promise.all because that will throw if any of the promises throw.

to keep things clean, i'd treat an error message as a normal return value. By returning a discriminated union, typescript is really good at determining the specific type by doing something like `if ('errors' in res) ...`

BartoszJarocki: @mattkrick, I see. Would you rather see error handling directly in the function that calls these methods, i.e., `addMissingJiraField` in this case? I didn't want to return the error messages from the new `AtlassianManager` methods to keep things consistent. 

Just to clarify, I'd change that to
```
async addFieldToScreenTab(cloudId: string, screenId: string, tabId: string, fieldId: string) {
    return (await this.post(
      `https://api.atlassian.com/ex/jira/${cloudId}/rest/api/3/screens/${screenId}/tabs/${tabId}/fields/`,
      {fieldId}
    )) as JiraAddScreenFieldResponse | AtlassianError | JiraError
```

and then use helpers like `isAtlassianError` in `addMissingJiraField` function. Any thoughts? 

And btw, what do you think about helpers like 
```
function isJiraNoAccessError<T>(response: T | JiraNoAccessError): response is JiraNoAccessError {
  return 'errorMessages' in response
}
```
Should I leave it or remove it and use check like `'errorMessages' in response` directly? Thanks!

BartoszJarocki:
@mattkrick not sure where to ask so I figured this is the best place ;) 
Do we have any rules what we want to see in Sentry? I see 
```
const standardError = (error: Error, options: SentryOptions = {}) => {
  const {message} = error
  sendToSentry(error, options)
  return {error: {message}}
}
```
in API code but there's also a lot of places with simply returning the error in the same format without sending it. I assume we should use our own judgement what's important, could be useful and decide?

packages/client/utils/AtlassianManager.ts
BartoszJarocki: https://github.com/ParabolInc/parabol/pull/4880/commits/6a841e0c667b10b2ee98293c221113741cf01acc @mattkrick I think what I think works best here I'd appreciate if you could take a look 🙏

mattkrick: big picture, if a call to a REST endpoint returns an error, should we throw it or return it?
I like returning errors when possible. That's because typescript can tell you the return value is e.g. `JiraError | JiraResult` but if you throw, then it becomes `never | JiraResult` and the caught error is of type `unknown`. type safety for the win!

> And btw, what do you think about helpers like

i don't have a strong opinion either way. They both offer the same type deduction in 1 line (if you write `if ('errorMessages' in response)` typescript is still going to be able to infer the type because `errorMessages` is a discriminator). The helper is more expressive, but at the expense of a little extra mental overhead because the debugger will have to see what the function does.

since we're likely going to handle AtlassianError and JiraError in the same way, you could format one so it looks like the other (e.g. `if ('errorMessages' in res) return {message: [res.errorMessages[0]}`) and that way the calling code would be less verbose because it'd only have to check for `JiraError | JiraResult` instead of `AtlassianError | JiraError | JiraResult`.

mattkrick:
> Do we have any rules what we want to see in Sentry?

The razor is, "if this error occurs 100 times, would that cause us to change something?"

At first i would send every error to sentry with `standardError`. Sentry was filling up with a bunch of errors that were handled exactly as they should be. For example, if someone tries to view a team they're not a member of & we tell them "you're not on this team", that's an error to the user, but should that be an error in sentry? You could argue that the system is working as intended & no need to send the error to sentry because there's nothing to fix. Conversely, you could argue that if 1000 people got that error, maybe we have a problem with our UI/UX?

packages/client/utils/AtlassianManager.ts
BartoszJarocki: Thanks! 

Regarding the error formatting, that'd make code way more readable. As Jira API errors can be under different properties and formatting one as another could be unnecessarily cumbersome I think the best way is to create one more type like
```
/**
 * Jira API throws a lot of different errors that could be placed in a different properties.
 * This type is used to expose the error under a simple interface for others to consume.
 */
export interface JiraApiError {
  errorMessage: string
}
```
This would make our `AtlassianManager` responsible to format error that way. This way handling errors in resolvers is as easy as
```
    const screensResponse = await manager.getScreens(cloudId)
    if (isJiraApiError(screensResponse)) {
      return {error: {message: screensResponse.errorMessage}}
    }
```
See the: https://github.com/ParabolInc/parabol/pull/4880/commits/520318ee9bb02c7483e3358eb839e57094aa7edb

mattkrick:
@BartoszJarocki is this ready for another review? feel free to ping me here or on slack when ready

BartoszJarocki:
Here are the things you should be aware of

1. https://github.com/ParabolInc/parabol/pull/4880/files#diff-baea95f99f42d5aeb5fb3239b5180131c801a0bf091b24d8093a3a2deb892dc3L95 I had to remove setting pending score here because right now, we still need access to the pending score after the Jira error happened. Optimistic update changes the final score and reverts it when `pokerSetFinalScore` mutation fails causing this hook to run and resets the final score as a result. If you think that might cause some side effects that I’m not aware of, please let me know. 
2. https://github.com/ParabolInc/parabol/pull/4880/files#diff-baea95f99f42d5aeb5fb3239b5180131c801a0bf091b24d8093a3a2deb892dc3R106 here I also had to add a check to avoid resetting the pending score. 
3. Currently, the `addMissingJiraField` mutation is allowed to run only by the meeting facilitator. I believe that’s the correct approach as the facilitator can update the final score. Having that in mind, we can’t expect anyone else to have the `manage:jira-project` permission. If I’m wrong here, I’m happy to improve.
4. Regarding the Jira API error handling, I’ve added: https://github.com/ParabolInc/parabol/pull/4880/files#diff-404948a861825f12e7e2bf83ef9e11ecd73ea510c140c77cc4e86a0d73dfaea9R8 which is returned from `AtlassianManager` and simplifies the error handling. I haven’t reused the existing types as it’d require adding some fake attributes (like `code` when using `AtlassianError` etc). After these changes, calling the `AtlassianManager` methods looks simple and all the heavy lifting related to error handling is kept inside of the `AtlassianManager`. For example

```
  async getScreenTabs(cloudId: string, screenId: string) {
    const res = (await this.get(
      `https://api.atlassian.com/ex/jira/${cloudId}/rest/api/3/screens/${screenId}/tabs`
    )) as JiraScreenTab[] | AtlassianError | JiraNoAccessError

    if (isJiraNoAccessError(res)) {
      return {errorMessage: res.errorMessages[0]}
    }

    if (isAtlassianError(res)) {
      return {errorMessage: res.message}
    }

    if (!Array.isArray(res)) {
      return {errorMessage: `Couldn't fetch screen: ${screenId} tabs!`}
    }

    return res
  }
```

And that’s the way it’s called

```
   const screenTabsResponse = await manager.getScreenTabs(cloudId, screen.id)
   if (isJiraApiError(screenTabsResponse)) {
     return null
   }
```

And yes, it’s ready to be reviewed again. There’s a lot of comments here so it might be easier to have a 15 min meeting just to make sure we’re on the same page with everything. I tried to list all the important parts above so @mattkrick please let me know if there’s anything else I can do, thank you!

packages/server/utils/predicates.ts
```diff
@@ -0,0 +1 @@
+export const isNotNull = <T>(item: T | null): item is T => item !== null
```
mattkrick, line 1: :clap: we could use a lot more of these!

mattkrick (changes requested):
looking really good!
only suggestions are to DRY out the atlassian manager error handling & the openOAuth call signature

packages/client/utils/AtlassianClientManager.ts
```diff
@@ -1,22 +1,22 @@
...
-import AddAtlassianAuthMutation from '../mutations/AddAtlassianAuthMutation'
-import AtlassianManager from './AtlassianManager'
+import AtlassianManager, {JiraPermissionScope} from './AtlassianManager'
 
 class AtlassianClientManager extends AtlassianManager {
   fetch = window.fetch.bind(window)
-  static openOAuth(atmosphere: Atmosphere, teamId: string, mutationProps: MenuMutationProps) {
-    const {submitting, onError, onCompleted, submitMutation} = mutationProps
+
+  static openOAuth(
```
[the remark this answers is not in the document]
BartoszJarocki: Agreed! This was a pretty dumb move on my end, fixed! 👌

packages/client/utils/AtlassianManager.ts
```diff
@@ -629,4 +708,96 @@ export default abstract class AtlassianManager {
       throw new Error('Cannot update field in Jira')
     }
   }
+
+  async getScreens(cloudId: string) {
+    const res = (await this.get(
+      `https://api.atlassian.com/ex/jira/${cloudId}/rest/api/3/screens`
+    )) as JiraScreensResponse | AtlassianError | JiraNoAccessError
+
+    if (isJiraNoAccessError(res)) {
```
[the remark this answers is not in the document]
BartoszJarocki: @mattkrick I think that's a really good idea! Do you think it should be done in this PR or rather have a separate issue to refactor the `AtlassianManager`?  Or simply add a couple of new methods and use it only in places changed/added in this PR?

mattkrick:
TEST
- [ ] join a project in jira
- [ ] turn off story points for that project (or story point estimate if nextgen)
- [ ] run a poker meeting in parabol & try to assign points via jira

First pass: got
```
  error: `{"data":null,"errors":[{"message":"Cannot use 'in' operator to search for 'errorMessages' in null","locations":[{"line":5,"column":3}],"path":["addMissingJiraField"]}]}`
```

BartoszJarocki:
@mattkrick checking! 🙈

mattkrick:
merged in #5133 :tada:

Issue #4748, opened 2021-03-05
Poker: Hover over avatar shows name during voting


gcrickman:
## Issue - Enhancement

A [user](https://app.hubspot.com/contacts/3888472/contact/13053751) 🔒 writes:

> When people vote, their icons will show on the screen, which is great. However, I am struggling to get my team to set up profile pictures so I can more clearly see who is who (right now all but 2 people have letters for an icon). Being able to hover my mouse over an icon and then having the name display would make things easier

PR #4882, opened 2021-04-13
Poker avatars tooltip

nickoferrall:
PR to resolve issue #4748 

This PR adds a tooltip to the Poker voting avatars.

I also noticed that the tooltip's position on the task card is slightly off so have fixed that too.

task-card-tooltip


ackernaut (approved):
LGTM!

PR #4886, opened 2021-04-14
Add LogRocket

nickoferrall:
This PR adds LogRocket to the codebase to help with debugging and performance monitoring. 

The credit card modal and the comment editor are not recorded for privacy reasons. 

Note: we should wait until this week's release is complete before merging this PR as we don't want to run this LogRocket experiment at the same time we run @tianrunhe's DataDog experiment.


packages/client/components/AnalyticsPage.tsx
```diff
@@ -61,8 +62,24 @@ const AnalyticsPage = () => {
...
+  const [isSegmentLoaded] = useScript(
+    `https://cdn.segment.com/analytics.js/v1/${key}/analytics.min.js`,
+    {
+      crossOrigin: true
+    }
+  )
   const atmosphere = useAtmosphere()
+  useEffect(() => {
+    const email = window.localStorage.getItem(LocalStorageKey.EMAIL)
+    LogRocket.init(window.__ACTION__.logRocket, {
```
tianrunhe, line 74: Can we follow our pattern in DataDog where you gate this initialization with another env variable, e.g., `LOGROCKET_ENABLED`? That way if this thing is harming our app performance, we can turn it off on the fly by setting the env variable, without the need of an emergency deployment?

.env.example
```diff
@@ -22,6 +22,7 @@ GRAPHQL_HOST='localhost:3000'
 GRAPHQL_PROTOCOL='http'
 HOST='localhost'
 INVITATION_SHORTLINK='key_INVITATION_SHORTLINK'
+LOG_ROCKET_ID='id/parabol'
```
tianrunhe, line 25: Kind reminders:
1. Update this key & value in action-devops/environments
2. Manually set the value with `dokku` on different stages

tianrunhe (changes requested):

packages/client/components/AnalyticsPage.tsx
nickoferrall: I like it! Just pushed an update

tianrunhe (approved):
LGTM!

packages/client/components/AnalyticsPage.tsx
```diff
@@ -61,8 +62,26 @@ const AnalyticsPage = () => {
...
+    `https://cdn.segment.com/analytics.js/v1/${key}/analytics.min.js`,
+    {
+      crossOrigin: true
+    }
+  )
   const atmosphere = useAtmosphere()
+  useEffect(() => {
+    if (window.__ACTION__.logRocketEnabled) {
+      const email = window.localStorage.getItem(LocalStorageKey.EMAIL)
+      LogRocket.init(window.__ACTION__.logRocketId, {
```
mattkrick, line 75: it's possible that `logRocketEnabled` is true but `logRocketId` is null.
let's simplify by doing what we do with segment & sentry-- just send an optional key to the client. if the key is present, it's enabled, else disabled

packages/client/components/AnalyticsPage.tsx
```diff
@@ -61,8 +62,26 @@ const AnalyticsPage = () => {
...
+        release: __APP_VERSION__
+      })
+      if (email) {
```
mattkrick, line 78: this hook only runs once, which means it's possible that `email` could be undefined if we haven't cached it yet.
may also want to include the identify call in https://github.com/ParabolInc/parabol/blob/b7259570dd3be40fc48cd02715fa9a422deb12c1/packages/client/utils/handleSuccessfulLogin.ts

mattkrick (changes requested):
looking forward to this!
just 2 fixes:
- remove 1 env var
- call identify on login

packages/client/components/AnalyticsPage.tsx
nickoferrall: Makes sense! So it might not be cached when the user logs in but by including the identifier there and here, we should always have the email?

mattkrick: yep!

PR #4893, opened 2021-04-15
add meeting card overflow menu

### LINKED ISSUE #4836: Adds an option menu to the meeting card
Continuation of #4623 
### Acceptance criteria
- [See Figma](https://www.figma.com/file/dfTQxhoKzohkCpxlztBbhS/Meetings-view?node-id=221%3A0)
- A menu appears in the top right
  - A 32px control is inset 4px. The icon is 24px centered. Use `more_vert` icon
  - The hover state may use a transparent white background (`#FFFFFF26` = [15% transparency](https://gist.github.com/lopspower/03fb1cc0ac9f32ef38f4)) since we want to play off the illustration background color. (We are avoiding transparency and trying to stick to palette variants only, but I think this is a good case for it)
- The menu item is:
  - (`person_add` icon) Copy invite link
- When the person taps ‘Copy invite link’
  - The menu closes
  - A snackbar appears ‘Meeting link copied!’ with `autoDismiss: 5`
  
### Deferred
Let’s wait until folks ask for a way to remove test meetings from the meetings view.

- The menu item below ‘Copy invite link’ is (`flag` icon) ‘End meeting’
- When the person taps ‘End meeting’
  - The menu closes
  - TBD: they get a dialog that asks them to confirm (or do you have a better idea? e.g. only show this option if the user has a role like facilitator? team lead?)
  - A snackbar appears ‘The meeting has been terminated’ (`popEndMeetingToast.ts`)  

 ### Estimated effort
 - 12 hours

mattkrick:
@ackernaut for your eyes, i don't love how it is right now, i'm thinking about moving the overflow menu icon to the bottom right per https://material.io/components/cards#usage. then, we can linkify both the image  & the meeting title 

fix #4836

TEST
- [ ] right click the bottom of the card & you can open the meeting in a new tab
- [ ] open the overflow menu & click copy invite & the invite now lives in your clipboard
- [ ] a "Copied!" tooltip pops up after you click copy

TEST User avatars
- [ ] meetings are sorted by createdAt descending (for the real-time aspect, it's nice to sort by something immutable. sorting by initial # of participants felt bad because if a new meeting was just created, it wouldn't be on top)
- [ ] on the meeting card, a user avatar shows up for every participant in the meeting
- [ ] when a user joins the meeting, their avatar is added to the list
- [ ] when a user leaves the meeting, their avatar is removed
- [ ] the poker active voting rows look the same (refactored this component to be the same)
- [ ] hovering the avatars shows their name


ackernaut:
### Tests
- [ ] right click the bottom of the card & you can open the meeting in a new tab
- [ ] open the overflow menu & click copy invite & the invite now lives in your clipboard
- [ ] a "Copied!" tooltip pops up after you click copy

ackernaut:
Here are some ideas on where we could move the options menu: https://www.figma.com/file/dfTQxhoKzohkCpxlztBbhS/Meetings-view?node-id=332%3A72920

ackernaut (approved):
It works per design. Let’s decide if we want to change the design here (looking at the new iterations in Figma) and pick a new pattern if we want. Then we can test UI again.

packages/client/components/AvatarList.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+const Wrapper = styled('div')({
+  alignItems: 'center',
+  display: 'flex',
+  position: 'relative',
+  width: '100%',
+  height: 32
+})
+
+const widthToOverlap = {
```
mattkrick, line 18: not sure if the overlap should change as avatar size changes? If not, we can remove this

packages/client/components/AvatarList.tsx
```diff
@@ -0,0 +1,87 @@
...
+  24: 10,
+  46: 10
+}
+
+interface Props {
+  users: AvatarList_users
+  onUserClick?: (userId: string) => void
+  onOverflowClick?: () => void
+  className?: string
+  size: 24 | 46
```
mattkrick, line 28: only reason this is specific is for the widthToOverlap lookup

mattkrick:
@ackernaut one thing i didn't realize is that in order for the avatar list to update in real time, each team member has to subscribe to the `lastSeenAtURLs` of everyone else. currently, this lives on the meeting subscription, but we'll have to move it somewhere else before we merge this

ackernaut:
@mattkrick I see. Is this much of a blocker, or just more work?

mattkrick:
eh, slight performance ding, but i think it enriches the UI so much that we should do it anyways.
is a little creepy that we're telling all your team members what app URLs you're at, but today we only say "meeting X or not", so i think it's probably more helpful than creepy

ackernaut:
TEST
- [ ] right click the bottom of the card & you can open the meeting in a new tab
- [ ] open the overflow menu & click copy invite & the invite now lives in your clipboard
- [ ] a "Copied!" tooltip pops up after you click copy

TEST User avatars
- [ ] meetings are sorted by createdAt descending (for the real-time aspect, it's nice to sort by something immutable. sorting by initial # of participants felt bad because if a new meeting was just created, it wouldn't be on top)
- [ ] on the meeting card, a user avatar shows up for every participant in the meeting
- [ ] when a user joins the meeting, their avatar is added to the list
- [ ] when a user leaves the meeting, their avatar is removed
- [ ] the poker active voting rows look the same (refactored this component to be the same)
- [ ] hovering the avatars shows their name

ackernaut:
@mattkrick this looks great. Stellar work!

I did some size wrangling here: https://github.com/ParabolInc/parabol/pull/4918 so that we get avatars that are the intended size apart from borders (24px, 40px) and fall on the grid when hanging the border and overlapping.

Lots of inspecting and Figma zoom!

mattkrick:
LGTM! thank you for the fine tuning @ackernaut!

Issue #4712, opened 2021-02-25
Sprint Poker: GitHub: Scope: Search bar


mattkrick:
AC
- [ ] The search bar looks just like the atlassian one
- [ ] The search query is sent directly to github. presumably, ideally, this means they can use the same GH search syntax

Estimate: 8 hours

Issue #4801, opened 2021-03-22
Integrate the GitHub GraphQL API into our current schema


mattkrick:
Given our current GraphQL schema, I'd like to extend it with the GitHub API.
This is probably best to do with graphql-tools.

The end result should allow us to make a GraphQL call like this:
```graphql
viewer {
  teamMember(teamId) {
    integrations {
      github {
        api {
          # everything in the api object should live in the GitHub schema
          viewer {
            login
          }
        }
     }
  }
}
```

Acceptance Criteria:
- [ ] using graphql-tools schema stitching, stitch the graphql schema into our own schema at build time
- [ ] the calls to `api` should use the token that exist in the `GitHubAuth` table in postgres (can set up GH auth in Team > Settings > Integrations)
- [ ] come up with an elegant solution to for handling errors in that part of the schema (eg GH is down, the user is rate limited, etc)

Estimate: 20 hours

mattkrick:
building this should simplify #4901 because we won't have to build anything special within our own schema, we can just forward on the request to GH

mattkrick:
Burndown:
- [ ] handle top-level errors like bad credentials
- [ ] handle other errors like rate limiting
- [ ] handle standard graphql errors
- [ ] support fragments
- [ ] support arguments

PR #4901, opened 2021-04-19
GitHub Scope Search Bar

nickoferrall:
WIP PR to resolve issue #4712 

I'm currently working on improving the types in the GitHubIntegration. 
To get the issues from the GitHub API, we need to use two different queries: `getIssues` and `searchIssues`. If we pass an empty query string into `searchIssues` it returns an empty array rather than all issues.

The `node` on getIssues is always of type `Issue` but `searchIssues` can be an `Issue`, PR, etc. If we guarantee that the type in `searchIssues` is `Issue`, we can get the types on the `node`, i.e. `id`, `title` etc.

However, it looks like the graphql-codegen package has a few open issues related to nested types when using unions. While I can get the types on the node when using `getIssues` or `searchIssues`, when I use a union, everything is type `any`. 

A few possible solutions:

1. I'm currently casting the node to the fragment type. Not ideal as we don't really get type safety
2. Use an if / else statement rather than creating a union of types `searchIssues` and `getIssues`. We get type safety but the code isn't DRY
3. Ideally, create generic that enables type safety and DRY code.


packages/client/components/ScopePhaseAreaGitHub.tsx
```diff
@@ -28,8 +28,8 @@ const ScopePhaseAreaGitHub = (props: Props) => {
   console.log({teamMember})
   const hasAuth = integrations?.github?.isActive ?? false
   if (!hasAuth) return <ScopePhaseAreaAddGitHub gotoParabol={gotoParabol} meeting={meeting} />
-  return <ComingSoon>Coming Soon!</ComingSoon>
-  // return <ScopePhaseAreaGitHubScoping meeting={meeting} />
+  // return <ComingSoon>Coming Soon!</ComingSoon>
```
nickoferrall, line 31: Need to re-add this before merging to master

packages/server/graphql/types/GitHubIntegration.ts
```diff
@@ -88,45 +89,82 @@ const GitHubIntegration = new GraphQLObjectType<any, GQLContext>({
...
-          projectKeyFilters.forEach((globalProjectKey) => {
-            const [cloudId, projectKey] = globalProjectKey.split(':')
-            projectKeyFiltersByCloudId[cloudId] = projectKeyFiltersByCloudId[cloudId] || []
-            projectKeyFiltersByCloudId[cloudId].push(projectKey)
+        type GitHubIssue = Pick<Issue, 'title' | 'id' | 'url' | 'updatedAt'> &
+          Record<'nameWithOwner', string>
+        const gitHubIssues = [] as GitHubIssue[]
+        let gitHubErrors: GQLResponse<GetIssuesQuery | SearchIssuesQuery>['errors']
+
+        if (queryString) {
```
nickoferrall, line 110: The repeated code here makes me sad but we do now get type safety

packages/server/graphql/queries/helpers/connectionFromGitHubIssues.ts
```diff
@@ -0,0 +1,29 @@
+import {GetIssuesNodeFragment} from '../../../types/typed-document-nodes'
+
+interface GitHubConnIssue extends GetIssuesNodeFragment {
+  updatedAt: Date
+}
+
+const connectionFromGitHubIssues = <T extends {updatedAt: Date} = GitHubConnIssue>(
```
mattkrick, line 7: i think we can get rid of this whole thing, we should just return the connection exactly as github gives it to us

packages/client/utils/githubQueries/searchIssues.graphql
```diff
@@ -0,0 +1,16 @@
+query searchIssues($queryString: String!) {
+  search(query: $queryString, type: ISSUE, first: 10) {
```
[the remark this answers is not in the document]
mattkrick: we can probably get rid of either these .graphql queries or the tagged template literals

mattkrick (changes requested):

mattkrick:
it looks like we might be able to forward the query directly to github. if that's so, it might be worthwhile to complete https://github.com/ParabolInc/parabol/issues/4801 first & then we can avoid creating our own `issues` field and GitHubIssue type

nickoferrall:
So instead of querying our server and getting the issues, we'd keep the query parameters in Relay. When a component needs the issues (or repos etc), we'd send a query directly to GitHub which would include a query string that looks like `repo:ParabolInc/parabol involves:@me sort:updated is:issue is:open testa`.

If that's correct, we save a trip to our server but wouldn't we miss out on error handling and caching? 

I can pause work on this PR until #4801 is ready

nickoferrall:
Updates pushed! It might be easier to review #4870 first before this one as this includes those changes.


> when a repo is added to nameWIthOwnerFilters, then repo:${nameWithOwner} is appended to the search string
> if nameWithOwnersFilters.length was 0 and is now > 0, then involves:@me is removed from the search string

I've included `involves:@me` in the query string and plan on adding this functionality in the filter PR when users can select repos. 

As we're including `nameWithOwners` directly in the query string, I don't think we need `nameWithOwnerFilters` anymore.

Also, as we mentioned we'd send the string directly to GH, I'm not sure we need to parse the query string as mentioned here: `the input search string is parsed to look for qualifiers `?

packages/server/graphql/types/GitHubIssue.ts
```diff
@@ -1,55 +1,31 @@
...
       type: GraphQLNonNull(GraphQLURLType),
       description: 'The url to access the issue'
-      // resolve: ({cloudName, key}) => {
-      //   return `https://${cloudName}.atlassian.net/browse/${key}`
-      // }
     },
-    nameWithOwner: {
-      type: GraphQLNonNull(GraphQLID),
-      description: 'The owner / repo of the issue as found in GitHub'
+    repository: {
```
nickoferrall, line 21: GitHub returns the `repository` which contains the `nameWithOwner`. Perhaps I should update [SuggestedIntegrationGitHub](https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/server/graphql/types/SuggestedIntegrationGitHub.ts) so that it follows this same format?

packages/client/components/GitHubScopingSearchInput.tsx
```diff
@@ -51,17 +51,22 @@ const GitHubScopingSearchInput = (props: Props) => {
...
     setSearch(atmosphere, meetingId, value)
   }
   const clearSearch = () => {
     setSearch(atmosphere, meetingId, '')
   }
+
   return (
     <Wrapper>
-      <SearchInput value={queryString} placeholder={placeholder} onChange={onChange} />
+      <SearchInput autoFocus value={`is:issue${queryString.slice(8)}`} onChange={onChange} />
```
nickoferrall, line 69: The GH GraphQL query requires a search type of issue, repo or user, so I'm using `type:ISSUE`. If the user replaced `is:issue` in the query string with `is:user` while we're using `type:ISSUE`, the results won't be very helpful. We could parse the string and update the search type but rating users or repos is an unlikely use case, so I think we should stick with `type:ISSUE`. 

I think we could either:
1. Always show `is:issue` in the search input and prevent the user from deleting it
2. Let the user remove `is:issue` and write `is:user`, returning unhelpful results

I prefer option 1. It might be a bit annoying that the user can't remove the text but always showing `is:issue` acts as a placeholder and makes it more likely that the results are useful

packages/server/graphql/types/GitHubIntegration.ts
```diff
@@ -69,64 +68,45 @@ const GitHubIntegration = new GraphQLObjectType<any, GQLContext>({
...
-          projectKeyFilters.forEach((globalProjectKey) => {
-            const [cloudId, projectKey] = globalProjectKey.split(':')
-            projectKeyFiltersByCloudId[cloudId] = projectKeyFiltersByCloudId[cloudId] || []
-            projectKeyFiltersByCloudId[cloudId].push(projectKey)
-          })
-        } else {
-          cloudIds.forEach((cloudId) => {
-            projectKeyFiltersByCloudId[cloudId] = []
-          })
+        const validQueryStr = queryString.replace(/is:pr/g, '')
```
nickoferrall, line 97: If a user enters `is:pr` and `is:issue`, the nodes returned from the query are empty. This is a problem as the `issueCount` includes the empty nodes

packages/client/components/GitHubScopingSearchInput.tsx
mattkrick: this is a fun find!
one problem is what if they change `is:issue` to `is:pr`? while it's probably not a best practice, i imagine some folks would say "let's scope how much effort it takes to finish this PR"

in that case, we probably want to let them edit `is:issue`. but, like you say, we probably don't want to let them shoot themselves in the foot by searching for users in the issues search.

some ideas:
- [ ] we validate the query string & search for `is:user`. if we find it, we remove it
- [ ] we pop a toast asking them why they want to see the world burn
- [ ] if the search returns 0 results, we see if the have `is:user` if they do, instead of saying "No results found" we say "Why are you searching for users you dingbat"

mattkrick: i don't much like the first option because we're doing something magical without explaining why (even though it should be obvious). After enough failed credit card & phone number forms that add parenthesis or remove letters, i've learned editing input fields is hard to get right.
 
i rather like the 3rd option, but it's your call!

packages/server/graphql/types/GitHubIntegration.ts
mattkrick: this smells like a problem with the inline fragment that you might be using?

mattkrick:
GH schema is now in master! details & example here: https://github.com/ParabolInc/parabol/pull/4920

packages/client/components/GitHubScopingSearchInput.tsx
nickoferrall: >  we pop a toast asking them why they want to see the world burn
🤣 

I like the third option too - it's less intrusive and works well with our existing UI. Updated!

nickoferrall:
The schema stitching is very impressive and helpful! It's working well when I hardcode the queryString but if I add the variable `$queryString` to the fragment and root query, it returns the error: `Parse error on ")"`. 

I can see the queryString variable in the network request so I think it's a parsing issue in the stitching, perhaps when there's more than one variable?

mattkrick:
hmmm, let me do some debugging

mattkrick:
fixed in latest commit

Issue #4730, opened 2021-03-03
'highestTier' field in Hubspot seems like it's not updating sometimes


avivapinchas:
## Issue - Bug

To send the appropriate communications to users and customers, we include a `highestTier` field on Contact records. It seems like this field isn't getting updated. 

Some example users: 
- User is showing as `personal` in [Hubspot](https://app.hubspot.com/contacts/3888472/contact/5206551) 🔒 , but `'pro` in [Amplitude](https://analytics.amplitude.com/parabol/chart/g32lmvx/edit/yp645n8) 🔒 This user hasn't met since December, and we did launch some update to the Hubspot data sync in December. 
- User is showing as `personal` in [Hubspot](https://app.hubspot.com/contacts/3888472/contact/11018651) 🔒 , but `Pro` in [Amplitude](https://analytics.amplitude.com/parabol/chart/g32lmvx/edit/yp645n8)🔒 They did meet in January. 
- User is showing as `enterprise` in both [Hubspot](https://app.hubspot.com/contacts/3888472/contact/10154751) 🔒  and [Amplitude](https://analytics.amplitude.com/parabol/chart/g32lmvx/edit/yp645n8) 🔒 , and they upgraded recently, and then met recently (Feb). 

### Acceptance Criteria (optional)
Parabol team can utilize Hubspot to send unique comms to customers, based on their highest tier. 

Triage is performed to find the root cause of the bug, timeboxed to ~1 hour.

tianrunhe:
This one is the same victim of the issue described in https://github.com/ParabolInc/parabol/issues/4842:

* Both of the first two users had their last meeting completed in the interval when we had HubSpot issue so they did not get chances to update the properties
* The third user had an `Enterprise Invoice Drafted` event on 02/09/2021, i.e., when we already fixed the HubSpot issue, so his `tier` property was properly updated.

Similar to #4842, we will backfill those contacts.

tianrunhe:
There is no easy way to find out whose tiers changed during the HubSpot outage time window so I just took all of the pro and enterprise users from our DB and backfill their tiers on HubSpot.

There could be some users who downgraded during that time but we don't have that tracking information and I assume there weren't many.

PR #4904, opened 2021-04-21
A few HubSpot update fixes + backfill script

### LINKED ISSUE #4842: HubSpot parabolId not set reliably
## Issue - Bug

Without the `parabolId` field in HubSpot we can't tell reliably who is, and who is not a user. This may cause us to spam contacts. This is bad.

This affects hundreds (thousands?) of users today. For a full report, see: https://app.hubspot.com/reports-list/3888472/77172779/  🔒  (created by @avivapinchas)

- **How serious is the bug?** P2
- **Estimated effort:** 6 hours

   - When HubSpot integration fires, the parabolId is updated
      - this should handle cases where the contact record exists in HubSpot, but `parabolId` is unset

tianrunhe:
Related to #4842 and #4730

### DIFF
--- a/packages/server/graphql/mutations/deleteUser.ts
+++ b/packages/server/graphql/mutations/deleteUser.ts
@@ -64,8 +64,12 @@ export default {
       orgIds.map((orgId) => removeFromOrg(userIdToDelete, orgId, undefined, dataLoader))
     )
     const validReason = reason?.trim().slice(0, 2000) || 'No reason provided'
-    const cachedUser = await dataLoader.get('users').load(user.id)
-    cachedUser.isRemoved = true
+
+    db.write('User', userIdToDelete, {
+      isRemoved: true,
+      reasonRemoved: validReason
+    })
+
     if (userId) {
       segmentIo.track({
         userId,
@@ -78,9 +82,7 @@ export default {
     // do this after 30 seconds so any segment API calls can still get the email
     setTimeout(() => {
       db.write('User', userIdToDelete, {
-        isRemoved: true,
-        email: 'DELETED',
-        reasonRemoved: validReason
+        email: 'DELETED'
       })
       updateUser(
         {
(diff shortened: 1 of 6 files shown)


packages/server/graphql/mutations/deleteUser.ts
```diff
@@ -64,6 +64,8 @@ export default {
       orgIds.map((orgId) => removeFromOrg(userIdToDelete, orgId, undefined, dataLoader))
     )
     const validReason = reason?.trim().slice(0, 2000) || 'No reason provided'
+    const cachedUser = await dataLoader.get('users').load(user.id)
```
mattkrick, line 67: how does this work?
`segmentIo.track` doesn't use a dataloader, so I'm not sure how changing this would affect the call.

let's just avoid caches by:
1. set `isRemoved,reasonRemoved` in the DBs immediately
2. after a 30-second timeout, set `email: getDeletedEmail(userId)`

packages/server/hubspot/backfillHubSpot.ts
```diff
@@ -0,0 +1,76 @@
+// call with yarn sucrase-node hubspot/backfillHubSpot.ts
```
mattkrick, line 1: if i do this, we can get rid of `debug.babel.js` :wink:

packages/server/hubspot/debug.babel.js
```diff
@@ -0,0 +1,3 @@
+require('../babelRegister')
```
mattkrick, line 1: let's delete this & call backfillHubspot.ts directly w/ sucrase-node

tianrunhe:
@mattkrick I noticed that this backfill script is not applicable for toggling `isRemoved` property in HubSpot because we don't save the email for deleted users. For this issue I'll just:
1. Get a list of users with `isRemoved = false` from HubSpot
2. Call HubSpot API directly to toggle the property

Hence I modified the script to only focus on backfilling `user Id` and `tier`

packages/server/graphql/mutations/deleteUser.ts
```diff
@@ -64,6 +64,12 @@ export default {
       orgIds.map((orgId) => removeFromOrg(userIdToDelete, orgId, undefined, dataLoader))
     )
     const validReason = reason?.trim().slice(0, 2000) || 'No reason provided'
+
+    db.write('User', userIdToDelete, {
+      isRemoved: true,
+      reasonRemoved: validReason
+    })
+
```
mattkrick, line 72: let's call `updateUser` up here, too.

mattkrick:
just 1 change. merge when ready!

PR #4918, opened 2021-04-27
refines avatar group layout

ackernaut:
Having an avatar with a hanging border, and overlapping other avatars, is a challenge on the 8px grid.

- For smaller avatars let’s use a 2px border, for larger (GTE 40px) let’s use 3px. It looks crisp for the larger sizes.
- If we want gutters and edges to fall on the 8px grid we have to offset for the 2 or 3 pixels. So you may see partial padding sizes like 5px (+3) or 4px (+2 +2 vertically)
- To get an avatar of 24 we have to call it 28 (24 + 2 + 2), and for 40 we measure 46 (40 + 3 + 3).
- There’s not a good solution yet for the growing whitespace + any optional vertical gutter so I hard-coded offsets to account for a row of 24px avatars


mattkrick:
_that's_ the Terry touch! looking clean

Issue #4711, opened 2021-02-25
Sprint Poker: GitHub: Scope: Repo filter menu


mattkrick:
We can add value over github by allowing users to easily search multiple repos at the same time.
To do this, we'll use a repo filter menu, similar to the project filter menu in the atlassian integration.

AC
- [ ] There's a filter button on the GH scoping view, just like atlassian
- [ ] Clicking the filter menu opens us a multiselect dropdown of repos that the user has access to
- [ ] it is sorted in a sensible, deterministic fashion (whatever you think is best that the github API supports)
- [ ] if more than ~10 repos, include a search box just like atlassian
- [ ] the function to populate the list lives in the GitHubServerManager. we're going to deprecate the GitHubClientManager because having the client talk directly to GH can be painful

Estimate; 16 hours

PR #4930, opened 2021-05-04
GitHub Scope Repo Filter

nickoferrall:
PR to resolve issue #4711 

Note: it's best to review #4901 first as this PR includes some of those changes.

I feel that the placeholder, `Search your GitHub repos`, in the menu search input also works well as a title. If we only show the search input when there's more than 10 repos, we should prob show a title when the search input isn't displayed. A title + a search input looks a bit busy to me so I'd prefer just showing a search input. 

What do you think? I could update the Jira menu to always show the search input instead of a title too if you agree.

### AC

- [ ]  There's a filter button on the GH scoping view, just like atlassian
- [ ]  Clicking the filter menu opens us a multiselect dropdown of repos that the user has access to
- [ ]  it is sorted in a sensible, deterministic fashion (whatever you think is best that the github API supports)
- [ ]  if more than ~10 repos, include a search box just like atlassian
- [ ]  ~~the function to populate the list lives in the GitHubServerManager. we're going to deprecate the GitHubClientManager because having the client talk directly to GH can be painful~~


packages/client/components/GitHubScopingSearchFilterMenu.tsx
```diff
@@ -75,109 +67,90 @@ type GitHubSearchQuery = NonNullable<
...
-  const {portalStatus, isDropdown} = menuProps
-  const toggleJQL = () => {
-    commitLocalUpdate(atmosphere, (store) => {
-      const searchQueryId = SearchQueryId.join('github', meetingId)
-      const githubSearchQuery = store.get(searchQueryId)
-      // this might bork if the checkbox is ticked before the full query loads
-      if (!githubSearchQuery) return
-      githubSearchQuery.setValue([], 'nameWithOwnerFilters')
-    })
+  const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
```
nickoferrall, line 90: Could clear the search input here when the menu closes but a user that has searched, selected and closed the menu, may wish to quickly open it again and deselect the issue. Having to search again to deselect would be a pain. Alternatively, we could clear the search but show all selected repos at the top of the list

mattkrick: yeaaaah showing the selected things at the top gets my vote! is that what we do for atlassian? i can't remember

nickoferrall:
Thanks for the feedback and the AC!

> when the queryString is updated, it is parsed & the list of project filters is updated

I'm now grabbing the 100 most recently updated repos and only displaying 10 at a time. If the user types into the menu filter search input, the string is parsed and the 10 repos update. Hopefully this is what you had in mind?

gh-menu-filter

packages/client/Atmosphere.ts
```diff
@@ -280,7 +280,7 @@ export default class Atmosphere extends Environment {
         uploadables: uploadables || undefined
       },
       // if sink is nully, then the server doesn't send a response
-      sink
+      sink!
```
mattkrick, line 283: rebase?

packages/client/components/GitHubScopingSearchFilterMenu.tsx
```diff
@@ -75,19 +69,26 @@ type GitHubSearchQuery = NonNullable<
...
 >
 
-const getValue = (item: {name: string}) => item.name.toLowerCase()
+type Repo = Pick<IXGitHubRepository, 'id' | 'nameWithOwner'>
 
-const MAX_PROJECTS = 10
+const MAX_REPOS = 10
+
+const getValue = (item: {nameWithOwner?: string}) => {
+  const repoName = item.nameWithOwner || 'Unknown Project'
```
mattkrick, line 77: nit -- let's call this "Unknown Repository" to be more githubby

mattkrick:
i think the sort is still a little off.
Here's what it looks like for me:
Screenshot from 2021-05-13 09-42-58

we probably don't care when the repository was last updated.
we care about when the user last did something with that repo (ie pushed a commit, made a comment, opened an issue)

packages/client/components/GitHubScopingSearchFilterMenu.tsx
```diff
@@ -96,88 +97,67 @@ const GitHubScopingSearchFilterMenu = (props: Props) => {
...
-              nameWithOwnerFiltersProxy.splice(keyIdx, 1)
-            } else {
-              nameWithOwnerFiltersProxy.push(globalProjectKey)
-            }
-            githubSearchQuery.setValue(nameWithOwnerFiltersProxy, 'nameWithOwnerFilters')
+            const newFilters = isSelected
+              ? nameWithOwnerFilters.filter((name) => name !== nameWithOwner)
+              : nameWithOwnerFilters.concat(nameWithOwner)
+            const queryString = githubSearchQuery.getValue('queryString')
+            githubSearchQuery.setValue(newFilters, 'nameWithOwnerFilters')
```
mattkrick, line 137: now that we know the queryString includes all the `repository:FOO` qualifiers, `nameWithOwnerFilters` is just duplicate information & we should remove that field from `githubSearchQuery`. That way, the queryString is the single source of truth & we have fewer variables to keep track of! I'll do this in another PR, just FYI that it'll be going away

mattkrick:
pushed a little commit that got rid of `nameWithOwnerFilters`. Instead of working hard to maintain that 2nd source of truth, I figure we could just parse them out of the queryString here: https://github.com/ParabolInc/parabol/blob/7fd64af087612d152962d9f4395df43988279595/packages/client/components/GitHubScopingSearchFilterMenu.tsx#L103

packages/client/components/GitHubScopingSearchFilterMenu.tsx
nickoferrall: Makes sense! I considered getting rid of it but thought it might be useful incase we wanted to grab the repos without needing to parse the query string. But I agree that it's cleaner to remove it!

nickoferrall:
Thanks for the screenshot - yours looks pretty different to mine as you've contributed to a bunch of popular public repos 😅 
I've pushed an update filtering by the repos that the user has most recently pushed a commit to!

packages/client/utils/GitHubManager.ts
```diff
@@ -22,7 +22,7 @@ type DocVariables = any
 // type DocVariables<T> = T extends DocumentNode<any, infer V> ? V : never
 
 abstract class GitHubManager {
-  static SCOPE = 'admin:org_hook,read:org,repo,user:email,write:repo_hook'
+  static SCOPE = 'admin:org_hook,read:org,repo,user,write:repo_hook'
```
nickoferrall, line 25: Updated the scope so that users can see contributions to public & private repos. The old scopes would only show contributions to public repos. I think this means that we need to remove users' GitHub auth so that they accept the new scope?

packages/client/utils/GitHubManager.ts
```diff
@@ -23,6 +23,8 @@ type DocVariables = any
 
 abstract class GitHubManager {
   static SCOPE = 'admin:org_hook,read:org,repo,user:email,write:repo_hook'
+  // static SCOPE = 'admin:org_hook,read:org,repo,user,write:repo_hook'
```
mattkrick: how excited are you that we didn't have to change scopes to make this request :pray:

nickoferrall: 😬   I did have to change the scope unfortunately! I think you're looking at a previous commit. I left a comment in the `GitHubManager` file mentioning that I had to change the `user` scope so that we have access to the private contributions

packages/client/utils/GitHubManager.ts
mattkrick: oof, yessir.
we've been talking about it for years, but i think with `GitHubAuth` living in postgres, we should probably use it as an example:
the `GitHubAuth` object should contain a `scope`. that way, we can compare the scope we have to the scope that the app needs & determine when the scope is stale. I'll make an issue & see if we can sneak it in in a timely manner

mattkrick:
LGTM! 
Only blocker is the scope change. I'll write a PR for #4970 & submit the scope change to github.
we can merge as soon as #4970 is in, no need to wait for GH since this feature isn't live yet.

PR #4931, opened 2021-05-04
explicitly send call stack to sentry from server

tiffanyhan:
Resolve #4931

AC
- [ ] server errors sent to sentry contain a call stack as a first-class citizen

### DIFF
--- a/packages/server/graphql/intranetSchema/mutations/addNewFeature.ts
+++ b/packages/server/graphql/intranetSchema/mutations/addNewFeature.ts
@@ -26,6 +26,8 @@ const addNewFeature = {
     }
   },
   resolve: async (_source, {copy, url}, {authToken, dataLoader}) => {
+    sendToSentry(new Error('something bad happened'))
+
     const r = await getRethink()
     const redis = getRedis()
 
--- a/packages/server/graphql/intranetSchema/mutations/addNewFeature.ts
+++ b/packages/server/graphql/intranetSchema/mutations/addNewFeature.ts
@@ -26,8 +26,6 @@ const addNewFeature = {
     }
   },
   resolve: async (_source, {copy, url}, {authToken, dataLoader}) => {
-    sendToSentry(new Error('something bad happened'))
-
     const r = await getRethink()
     const redis = getRedis()
 
--- a/packages/server/utils/sendToSentry.ts
+++ b/packages/server/utils/sendToSentry.ts
@@ -22,8 +22,11 @@ const sendToSentry = async (error: Error, options: SentryOptions = {}) => {
   if (user && ip) {
     ;(user as any).ip_address = ip
   }
+  const stack = error.stack
+
   Sentry.withScope((scope) => {
     user && scope.setUser(user)
+    stack && scope.setExtra('stack', stack)
     if (tags) {
       Object.keys(tags).forEach((tag) => {
         scope.setTag(tag, String(tags[tag]))


tiffanyhan:
while working on this issue, i realized a stack trace is included in sentry errors because `sendToSentry` always logs errors to the console before talking to sentry. it's included automatically as a breadcrumb with the "console" label. however, it's not very obvious and can often be buried amongst lots of other breadcrumbs. it may also be a "happy accident", because that logging statement is contained inside a commented out conditional.

this PR explicitly sends a stack to sentry, so in the UI it will be graphically highlighted as the first thing the viewer sees. it's also useful to see a snippet of the offending code-- sentry has access to client and server source maps, so it should now display this automatically as it does for client errors.

mattkrick:
oh this looks nice! i was worried that it was a sourcemaps issue!
could you push this to the dev server & trigger an error? would love to see what it looks like now

tiffanyhan:
> oh this looks nice! i was worried that it was a sourcemaps issue!
> could you push this to the dev server & trigger an error? would love to see what it looks like now

so i pushed to dev server and triggered an error, but afaik sentry got nothing. it's pretty odd bc i ssh'ed into and verified that dev and staging both share the same `SENTRY_DSN`; also when i set that `SENTRY_DSN` value in my local `.env` file i could trigger an error locally and see it in sentry. any idea what is up? perhaps it's a networking issue that @adaniels-parabol  knows about?

mattkrick:
not sure what's going on!
locally, if i use the action-staging DSN .env var, it works.
BUT, if i run `yarn build && yarn start` locally, then it doesn't work.
so, something about the production build is keeping Sentry from running?
could be worthwhile seeing outbound network traffic to see if it's actually being sent to sentry. if so, then we can compare payloads to sentry of dev vs. prod
if it's not being sent to sentry at all, then we'll need to try & change the production vars to development until it works & then see what's breaking it based on NODE_ENV

tiffanyhan:
i used wireshark to capture traffic in the success vs failure case with filter `ip.addr == 35.188.42.15/32` for sentry's ip address space. in both cases, `captureException` returns an `eventId` and `initSentry` is called with the exact same args.

in the success case, i saw a tcp/tls handshake, application data being sent, then a connection termination. in the failure case, no packets were seen in wireshark until a good while later. then i saw a TCP ack sent from node.js app to sentry, with sentry responding with `TCP ACKed unseen segment`.

locally, it looks like the difference might be in how TLS is handled with dev vs prod env vars? in deployed envs, the failure case only happens in dev server so we maybe need to double check TLS on dev server.

https://docs.sentry.io/product/security/ip-ranges/
https://github.com/getsentry/sentry-javascript/blob/master/packages/node/src/backend.ts

tiffanyhan:
setting `debug=true` doesn't yield any extra info when sending to sentry...

tiffanyhan:
there are definitely some server errors which have been captured in the past in sentry, and all communication with sentry is encapsulated in `sendToSentry` util. so this is possibly a recent regression.

tiffanyhan:
me and jordan used proxyman to intercept the request to sentry. we saw the request and payload in dev mode. however, once the app was built, no request appeared in proxyman. next step suggested by jordan is to monkey patch http (called by node sentry) and bisect our way up the stack trace to find where it goes bad.

tiffanyhan:
found the problem: in dev mode, `initSentry` is called "in scope" for both web server and graphql exec. in prod mode, `initSentry` is only called from the context of the web server. it also needs to be initialized from the context of graphql executor.

e.g. `initSentry` is only called in `server.ts`, which is the webpack entry point for the web server.

mattkrick:
nice find!!

tiffanyhan:
@tianrunhe @mattkrick 

this is how server stack traces thrown from `gqlExecutor` currently look like in sentry: https://sentry.io/organizations/parabol/issues/2396196206/events/91c8fb3b1f3f4a2691b525c15609d130/?project=107915

Since graphql requests get resolved by executing given a `DocumentNode` or `CompiledQuery` obj, do you have any thoughts on how to achieve a nice code snippet display such as this one here: https://sentry.io/organizations/parabol/issues/2082696002/?project=107196&query=is%3Aunresolved , or at least something more helpful than what it is currently?

i haven't dug in to graphql internals, so i'm not sure what the options would be here.

mattkrick:
I think the problem is here: https://github.com/ParabolInc/parabol/blob/da627c505d5cdf7d90f9a8a26da63f77e0675e9e/scripts/webpack/prod.servers.config.js#L55

that line causes the gqlExecutor.js file to end with `//# sourceMappingURL=/static/gqlExecutor_308f123949e8fb896ac01edbe6d987fc.js.map`
I'm betting sentry reads that line but can't find it because it's a relative path.

here are some possible fixes:
- tell webpack to stop minifying our server files. this is turned on by default when `mode: 'production'`, we can turn it off manually & then we won't need source maps! (I think this is my favorite option)
- fix `publicPath` in prod.servers.config so it matches prod.client.config. not sure why the 2 are different!
- try removing the `SourceMapDevToolPlugin` in the server config. TBH I'm not even sure why we upload our server assets to S3
- if all else fails, use the sentry webpack loader & then go into sentry to see what the file looks like (https://sentry.io/settings/parabol/projects/action-production/source-maps/6.11.0/?query=) then, knowing what it should look like, we can work backwards from there

PR #4932, opened 2021-05-04
EmojiMart

### LINKED ISSUE #3639: emoji mart fixups
AC
- [ ] autofocus on open
- [ ] persist frequently used in the local storage
- [ ] consider allowing folks to tab through options in the menu (currently pressing tab closes the menu altogether)

ee 10.5

mikennaji:
Fixes #3639
- [x] autofocus on open
- [x] persist frequently used in the local storage

AutoFocus on Emoji Search Bar on Mount 

By default, EmojiMart will store user-chosen skin and frequently used emojis in localStorage.In passing the " recent" array we give it preassigned values. Setting enableFrequentEmojiSort to true will Instantly sort the “Frequently Used” category based on user selection

### DIFF
--- a/packages/client/components/ReactjiPicker.tsx
+++ b/packages/client/components/ReactjiPicker.tsx
@@ -16,17 +16,6 @@ interface Props {
   menuProps: MenuProps
 }
 
-const recent = [
-  'heart',
-  'tada',
-  'smile',
-  'rocket',
-  'fire',
-  'white_check_mark',
-  'confused',
-  'cry',
-  'x'
-]
 const ReactjiPicker = (props: Props) => {
   const {menuProps, onClick} = props
   const {closePortal} = menuProps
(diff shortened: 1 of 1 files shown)


mattkrick:
Thanks for the PR @mikennaji!
I wonder, is there a way to show the `recent` emojis by default & then as folks use it, the `recent` emojis are replaced with their frequently used ones? I'm pretty sure most folks don't want to use :kissing_heart: with their coworkers :wink: 

Maybe something like...
- if there are no frequently used emojis in local storage, the frequently used row shows the `recent` emojis
- when an emoji gets used more frequently, it gets put into the frequently used row

mikennaji:
By default EmojiMart will store user chosen skin and frequently used emojis in localStorage. That can however be overwritten should you want to store these in your own storage.

import { store } from 'emoji-mart'

store.setHandlers({
  getter: (key) => {
    // Get from your own storage (sync)
  },

  setter: (key, value) => {
    // Persist in your own storage (can be async)
  }
})
The "recent" array can be passed if  "frequently" key  returns nothing instead of an array objects  where the key is the emoji name and the value is the usage count as such  { 'astonished': 11, '+1': 22 } which can be used to create a list of the most used emojis for each user.

mattkrick:
closing for inactivity. if you'd like to take another shot feel free to reopen!

PR #4941, opened 2021-05-07
Prod fixups

### LINKED ISSUE #4940: Handle non-durable analytics calls via trebuchet
Not all messages to the server need a response. For example, analytics calls.

These are messages that our server doesn't send responses for: https://github.com/ParabolInc/parabol/blob/bfed594b1f8f053ea628a3c541502ddd9b47b374/packages/server/graphql/handleGraphQLTrebuchetRequest.ts

These messages get logged in the graphql-trebuchet-client operations, but never get cleared because a `completed` message is never sent.

That means, when a person reconnects, the operation is re-sent: https://github.com/mattkrick/graphql-trebuchet-client/blob/master/src/GQLTrebuchetClient.ts#L83

There are a few ways to fix this:
1. When we send a message, we include a flag like `nonDurable: true`. that tells the client & the server that no response is needed.
2. we always send a response, no matter what
3. we change how the operations cache works (e.g. we only cache subscriptions, not mutations or queries)

This is a bug that is affecting all of our client-side segment events as well as setLocation, so we should fix it soon.

AC
- [ ] one of the 3 options is chosen & built

Estimate: 4 hours

mattkrick:
fix #4940

fixes segment events being sent on repeat when a client reconnects a bunch

also fixes a few errors found in the server logs:
- [ ] treEnd being called after end
- [ ] agendaItem not being found (we were still hard deleting them)


packages/server/graphql/mutations/removeAgendaItem.ts
```diff
@@ -34,7 +34,10 @@ export default {
     const agendaItem = await r
       .table('AgendaItem')
       .get(agendaItemId)
-      .delete({returnChanges: true})('changes')(0)('old_val')
```
mattkrick, line 37: no more hard deletes on agenda items, this was causing dataloaders to fail the lookup

packages/server/graphql/handleGraphQLTrebuchetRequest.ts
```diff
@@ -51,16 +52,17 @@ const handleGraphQLTrebuchetRequest = async (
         safeError.stack = firstError.stack
         sendToSentry(safeError)
       }
-      if (result.data && IGNORE_MUTATIONS.includes(Object.keys(result.data)[0])) return
+      // no client sink = no opId = no response expected
+      if (!opId) return
```
mattkrick, line 56: this is the crux of the file change. client now tells server what responses to ignore :tada:

packages/server/pipeStreamOverResponse.ts
```diff
@@ -12,6 +12,10 @@ const pipeStreamOverResponse = (
   })
   readStream
     .on('data', (chunk: Buffer) => {
+      if (res.done) {
```
mattkrick, line 15: tryEnd has been getting called after `end` has been called. this is somewhat repeatable when the rateLimiter gets hit (e.g. setAppLocation gets called too many times).

i'm not sure if this is the right spot to exit early, but a stream is asynchronous so it seems like it could be

packages/server/socketHelpers/sendGQLMessage.ts
```diff
@@ -6,19 +6,18 @@ interface Message {
   type: string
   syn: boolean
   payload?: Record<string, any>
-  id?: string
+  id: string
 }
 
 const sendGQLMessage = (
   context: ConnectionContext,
+  opId: string,
```
mattkrick, line 14: now that opId is guaranteed in all responses, we can move it up the parameter list

packages/client/Atmosphere.ts
```diff
@@ -279,11 +273,15 @@ export default class Atmosphere extends Environment {
...
-      uploadables: uploadables || undefined
-    }, sink || noopSink)
+    return transport.fetch(
+      {
+        [field]: data,
+        variables,
+        uploadables: uploadables || undefined
+      },
+      // if sink is nully, then the server doesn't send a response
+      sink
```
mattkrick, line 283: graphql-trebuchet-client now handles an optional sink.
if no sink is passed it, the operation isn't cached, which means the server doesn't have to respond

packages/server/socketHandlers/handleMessage.ts
```diff
@@ -28,9 +28,10 @@ const handleParsedMessage = async (
       return
     }
     const response = await handleGraphQLTrebuchetRequest(msg, connectionContext)
-    if (response) {
+    // only reply if an opId was included. no opId = no sink on client = ignored
+    if (response?.id) {
```
mattkrick, line 32: this is basically the whole PR.
if it came in via websocket, don't reply if the client didn't give you an opId

PR #4948, opened 2021-05-11
Add missing segmentId field in insertUserQuery SQL statement

tianrunhe:
`segmentId` was missing in the SQL insertion statement.


tiffanyhan:
looks good! can you run `yarn pg:build` and then check in the re-generated `insertUserQuery.ts` file? will this PR include a backfill for the `segmentId` mismatches as well?

tianrunhe:
@tiffanyhan I run `yarn pg:build` and it looks all fine. However if I'd commit the generated type files (which I'm not sure if I should have), I got this ts type error:

```
parabol-server: graphql/mutations/helpers/bootstrapNewUser.ts:51:16 - error TS2345: Argument of type 'User' is not assignable to parameter of type 'InsertUserQueryParams'.
parabol-server: Property 'segmentId' is optional in type 'User' but required in type 'InsertUserQueryParams'.
parabol-server: 51     insertUser(newUser)
parabol-server:                   ~~~~~~~
parabol-server: Found 1 error.
```

I'm not a PG master so I'm not sure how to instruct the insertion SQL to accept optional value?

For the backfill script, I committed one here but again I'm not sure if this is the right way to do. In particular I'm not sure how to write a SQL to do sth like:
```
UPDATE "User"
SET 
  "segmentId" = 1 where id = '123',
  "segmentId" = 2 where id = '567'
  ... 
```

Thanks for your help!

tiffanyhan:
> @tiffanyhan I run yarn pg:build and it looks all fine. However if I'd commit the generated type files (which I'm not sure if I should have), I got this ts type error:

we should commit! the typescript file gives us a typesafe API to interact with PG from the app. 

the sql statement already does what we want. e.g. if `segmentId` isn't passed, it will insert and set `segmentId` to the default value of null.

you can go to[ `insertUser.ts`](https://github.com/ParabolInc/parabol/blob/master/packages/server/postgres/queries/insertUser.ts) file and mess with the type of the fn args. E.g. if we set the type to `Partial<InsertUserQueryParams>` similar to how we do for [`updateUser.ts`](https://github.com/ParabolInc/parabol/blob/master/packages/server/postgres/queries/updateUser.ts), it will work. In this case, it doesn't give the best type safety, so maybe you can [try to define a more specific type by specifying what is optional and what's required](https://stackoverflow.com/questions/52703321/make-some-properties-optional-in-a-typescript-type). What's required is [what fields are non nullable in PG that have no default values set](https://github.com/ParabolInc/parabol/blob/master/packages/server/postgres/migrations/1616623767058_addUserTable.ts). can also look directly at postgres schema for this.

in general, the files at the top level of [`postgres/queries`](https://github.com/ParabolInc/parabol/blob/master/packages/server/postgres/queries/README.md) has been useful for making tweaks to the types wherever we find that pgtyped falls short.

tiffanyhan:
> For the backfill script, I committed one here but again I'm not sure if this is the right way to do. 

The backfill you committed looks good! Depending on how many `segmentId`s have been affected, the other thing you were trying to do might be preferable though, since the committed version requires a full rethink table scan whereas the pg only one uses an index:

```
UPDATE "User"
SET 
  "segmentId" = 1 where id = '123',
  "segmentId" = 2 where id = '567'
  ... 
```

I think you should be able to do it by [using `FROM` clause with a mapping table and aliases](https://stackoverflow.com/questions/18797608/update-multiple-rows-in-same-query-using-postgresql).

tianrunhe:
@tiffanyhan Thanks for the info! Reeeally helpful!

* I made the change to the type in `insertUser`
* I modified the backfill script to utilize PG's batch insertion

When I tried to run/test my backfill script, I did `yarn pg:postmigrate up backfillSegmentId` I got the following error:
```
Error: Can't get migration files: TSError: ⨯ Unable to compile TypeScript:
packages/server/postgres/migrations/postDeploy/1613857272859_backfillUserTable.ts:29:25 - error TS2352: Conversion of type '{ newFeatureId: any; overLimitCopy: any; isRemoved: boolean; segmentId: any; reasonRemoved: any; rol: any; inactive: boolean; payLaterClickCount: number; featureFlags: any[]; } & User & { ...; }' to type '{ id: string | void; email: string | void; createdAt: void | Date; updatedAt: void | Date; inactive: boolean | void; lastSeenAt: void | Date; preferredName: string | void; ... 12 more ...; payLaterClickCount: number | void; }' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Types of property 'identities' are incompatible.
    Type 'AuthIdentity[]' is not comparable to type 'void | JsonArray'.
      Type 'AuthIdentity[]' is not comparable to type 'JsonArray'.
        Type 'AuthIdentity' is not comparable to type 'string | number | boolean | Json[] | { [key: string]: Json; }'.
          Type 'AuthIdentity' is not comparable to type '{ [key: string]: Json; }'.
            Index signature is missing in type 'AuthIdentity'.

29     const cleanedUser = Object.assign({}, undefinedUserFieldsAndTheirDefaultPgValues, user, {
                           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
30       email: user.email === 'DELETED' ? getDeletedEmail(user.id) : user.email,
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
31       preferredName: user.preferredName.slice(0, USER_PREFERRED_NAME_LIMIT)
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
32     }) as IBackupUserQueryParams['users'][0]
   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

    at createTSError (/Users/brucetian/Projects/parabol/node_modules/ts-node/src/index.ts:421:12)
    at reportTSError (/Users/brucetian/Projects/parabol/node_modules/ts-node/src/index.ts:425:19)
    at getOutput (/Users/brucetian/Projects/parabol/node_modules/ts-node/src/index.ts:552:36)
    at Object.compile (/Users/brucetian/Projects/parabol/node_modules/ts-node/src/index.ts:757:32)
    at Module.m._compile (/Users/brucetian/Projects/parabol/node_modules/ts-node/src/index.ts:836:43)
    at Module._extensions..js (internal/modules/cjs/loader.js:1097:10)
    at Object.require.extensions.<computed> [as .ts] (/Users/brucetian/Projects/parabol/node_modules/ts-node/src/index.ts:839:12)
    at Module.load (internal/modules/cjs/loader.js:941:32)
    at Function.Module._load (internal/modules/cjs/loader.js:782:14)
    at Module.require (internal/modules/cjs/loader.js:965:19)
    at loadMigrations (/Users/brucetian/Projects/parabol/node_modules/node-pg-migrate/dist/runner.js:30:15)
    at async Promise.all (index 0)
    at exports.default (/Users/brucetian/Projects/parabol/node_modules/node-pg-migrate/dist/runner.js:144:40)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```
Which is not my backfill script so I'm a little bit confused here. Any idea?

tiffanyhan:
> When I tried to run/test my backfill script, I did yarn pg:postmigrate up backfillSegmentId I got the following error:

Ahh, long story short is i don't know why `yarn pg:migrate up` works but `yarn pg:postmigrate up` has troubles with typescript files. what i've been doing is using `yarn postmigrate` command to run the migrations programmatically. in `postDeployMigrate`, the thing that makes it not complain about typescript files is [this one-line import](https://github.com/ParabolInc/parabol/blob/master/scripts/postDeployMigrate.js#L7). i remember asking matt why this is so, but he forgot what he wrote in that package that would solve it lol. 

lmk if you have any hunches! but currently, `yarn pg:postmigrate` is basically only useful for creating the migration files. should probably be documented, or change the command to `yarn pg:migrate create` accordingly to prevent confusion.

packages/server/postgres/migrations/postDeploy/1620887961308_backfillSegmentId.ts
```diff
@@ -0,0 +1,50 @@
...
+        .pluck('id', 'segmentId')
+        .skip(offset)
+        .limit(batchSize)
+        .run()
+      if (!rethinkUsers.length) {
+        break
+      }
+      foundUsers = true
+      const pgUsers = rethinkUsers.map(({id, segmentId}) => ({id, segmentId}))
+      await catchAndLog(() => backfillSegmentIdQuery.run({users: pgUsers}, getPg()))
```
tiffanyhan, line 32: Do you have an idea around how many `segmentId`s have been affected? If it's just a couple, it might be worth hard coding the `segmentId <=> id` mapping, so we can avoid a full table scan in rethink.

packages/server/postgres/queries/src/backfillSegmentIdQuery.sql
```diff
@@ -0,0 +1,8 @@
+/*
+  @name backfillSegmentIdQuery
+  @param users -> ((segmentId, id)...)
+*/
+UPDATE "User" AS u SET
+  "segmentId" = c."segmentId"
+FROM (VALUES :users) AS c("segmentId", "id") 
+WHERE c."id" = u."id";
```
tiffanyhan, line 8: looks good!

one nitpick i have is that for one-off queries like this, the need for a defined type-safe api in the app isn't as great, so going through the whole pgtyped build process may be overkill. it might make more sense to use the passed `pgm` client or `getPg` here. pgm uses the node-postgres driver under the hood, so either one should be equivalent.

packages/server/postgres/migrations/postDeploy/1620887961308_backfillSegmentId.ts
tianrunhe: Yeah there will be more than 10K and it's growing every day with newly created accounts 😂

So a full table scan is not avoidable unfortunately.

packages/server/postgres/queries/src/backfillSegmentIdQuery.sql
tianrunhe: I like simpler code but I'm not sure a good way to do [argument expansion](https://pgtyped.vercel.app/docs/sql-file#array-spread-and-pick) without pgtyped? This is sth I tried:

```typescript
      const users_expanded = pgUsers.map(({segmentId, id}) => `('${segmentId}', '${id}')`).join(',')
      pgm.sql(`
        UPDATE "User" AS u SET
        "segmentId" = c."segmentId"
        FROM (VALUES ${users_expanded}) AS c("segmentId", "id") 
        WHERE c."id" = u."id";
      `)
```

tianrunhe:
> lmk if you have any hunches! but currently, `yarn pg:postmigrate` is basically only useful for creating the migration files. should probably be documented, or change the command to `yarn pg:migrate create` accordingly to prevent confusion.

I did two things:
1. Moved the script from `migrations/postDeploy` to `migrations/`: `mv packages/server/postgres/migrations/postDeploy/1620887961308_backfillSegmentId.ts packages/server/postgres/migrations/`
2. Run `yarn pg:migrate up backfillSegmentId`

and now I've got errors for RehtinkDB config:
```
Error: Can't get migration files: TSError: ⨯ Unable to compile TypeScript:
packages/server/database/getRethinkConfig.ts:4:8 - error TS1192: Module '"/Users/brucetian/Projects/parabol/node_modules/@types/dotenv/index"' has no default export.

4 import dotenv from 'dotenv'
         ~~~~~~
packages/server/database/getRethinkConfig.ts:5:8 - error TS1259: Module '"path"' can only be default-imported using the 'esModuleInterop' flag

5 import path from 'path'
         ~~~~

  node_modules/@types/node/path.d.ts:152:5
    152     export = path;
            ~~~~~~~~~~~~~~
    This module is declared with using 'export =', and can only be used with a default import when using the 'esModuleInterop' flag.

    at createTSError (/Users/brucetian/Projects/parabol/node_modules/ts-node/src/index.ts:421:12)
    at reportTSError (/Users/brucetian/Projects/parabol/node_modules/ts-node/src/index.ts:425:19)
    at getOutput (/Users/brucetian/Projects/parabol/node_modules/ts-node/src/index.ts:552:36)
    at Object.compile (/Users/brucetian/Projects/parabol/node_modules/ts-node/src/index.ts:757:32)
    at Module.m._compile (/Users/brucetian/Projects/parabol/node_modules/ts-node/src/index.ts:836:43)
    at Module._extensions..js (internal/modules/cjs/loader.js:1097:10)
    at Object.require.extensions.<computed> [as .ts] (/Users/brucetian/Projects/parabol/node_modules/ts-node/src/index.ts:839:12)
    at Module.load (internal/modules/cjs/loader.js:941:32)
    at Function.Module._load (internal/modules/cjs/loader.js:782:14)
    at Module.require (internal/modules/cjs/loader.js:965:19)
    at loadMigrations (/Users/brucetian/Projects/parabol/node_modules/node-pg-migrate/dist/runner.js:30:15)
    at async Promise.all (index 0)
    at exports.default (/Users/brucetian/Projects/parabol/node_modules/node-pg-migrate/dist/runner.js:144:40)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

packages/server/postgres/queries/src/backfillSegmentIdQuery.sql
tiffanyhan: nice job! that's what i would have done. i prefer the version without pgtyped, but don't have too much tension here.

tiffanyhan (approved):

packages/server/postgres/migrations/postDeploy/1620887961308_backfillSegmentId.ts
```diff
@@ -0,0 +1,50 @@
...
+        .filter((row) => row.hasFields('segmentId'))
+        .pluck('id', 'segmentId')
+        .skip(offset)
+        .limit(batchSize)
+        .run()
+      if (!rethinkUsers.length) {
+        break
+      }
+      foundUsers = true
+      const pgUsers = rethinkUsers.map(({id, segmentId}) => ({id, segmentId}))
```
mattkrick, line 31: nit: you're already plucking just these 2 fields, can you get rid of the map?

packages/server/postgres/migrations/postDeploy/1620887961308_backfillSegmentId.ts
```diff
@@ -0,0 +1,50 @@
...
+export const shorthands: ColumnDefinitions | undefined = undefined
+
+export async function up(): Promise<void> {
+  const r = await getRethink()
+  const batchSize = 3000
+  const doBackfill = async (usersAfterTs?: Date) => {
+    let foundUsers = false
+
+    for (let i = 0; i < 1e5; i++) {
+      console.log('i:', i)
```
mattkrick, line 16: let's remove this & the other log now that we know it's working

PR #4953, opened 2021-05-12
add private mutation lockTeams

mattkrick:
TEST
- [ ] as a superuser, on the private schema, can run a mutation like this:
```gql
mutation {
  lockTeams(teamIds:["2INgjqSy9W"], isPaid:false)
}
```

The result if you're a billing leader:
Screenshot from 2021-05-11 16-55-01

The result if not a billing leader: 
Screenshot from 2021-05-11 16-53-50

We could support custom messaging, but chances are we'd want that messaging to be raw HTML otherwise it'll look really bad.

@jordanh


jordanh:
You are my hero 🦾 

My dream scenario would be optional embedded html (but I realize that will also require a DB migration to implement). With embedded html I could write a message that is, "please request Parabol by doing [specific company X at your peculiar company], they don't think there is enough demand! Prove them wrong."

Would you accept my scope creep?

mattkrick:
get me some test HTML that you wanna see & i'll try it out!

jordanh:
Looking at this now!

jordanh:
@mattkrick I'm thinking about something like this (image not pasted for anonymity): https://www.figma.com/file/VM8xFhm3K6hILz3NxW9UclPo/Jordan-Playground?node-id=9%3A272

Note: dialog text is left-justified (not center). Note CTA is overridden too. In this case we do _not_ want folks to go to the billing page because we know there is 0 authority to make a credit card purchase. Instead, we want to be able to override the CTA to be a custom a href destination

AC is something like:
   * Can override dialog content with custom embedded HTML
   * Can override CTA with custom href
   * Toggling `isPaid` to true clears both customizations (if set), and unlocks team

Doable?

mattkrick:
IMO we should just make it all HTML-- the title, body, and button. 

what do you want the button to do?

mattkrick:
Here's a screenshot
Screenshot from 2021-05-12 14-20-30

Here's the HTML I used:
```
<h2 style="
    color: #444258;
    font-size: 20px;
    font-weight: 600;
    line-height: 1.5;
    margin: 0;
    padding: 20px 24px 0;
">Oh dear…</h2><div style="
    color: #444258;
    font-size: 14px;
    line-height: 1.5;
    padding: 16px 24px 24px;
    text-align: left;
">Your Agile Enablement team didn't see enough requests from users to continuing using Parabol. 
<br><br><div style="font-weight: 600;">If you'd like your team unlocked, you may contact us</div>
</div>
```

jordanh:
I guess there doesn't need to be any button to dismiss this modal. LGTM sir!

packages/server/graphql/intranetSchema/mutations/lockTeams.ts
```diff
@@ -0,0 +1,42 @@
+import {GraphQLBoolean, GraphQLID, GraphQLList, GraphQLNonNull} from 'graphql'
+import getRethink from '../../../database/rethinkDriver'
+import updateTeamByTeamId from '../../../postgres/queries/updateTeamByTeamId'
+import {requireSU} from '../../../utils/authorization'
+
+const lockTeams = {
```
jordanh, line 6: How do I set the HTML?

packages/client/modules/teamDashboard/components/UnpaidTeamModal/UnpaidTeamModal.tsx
```diff
@@ -1,40 +1,42 @@
...
+  const [firstBillingLeader] = billingLeaders
+  const billingLeaderName = firstBillingLeader?.preferredName ?? 'Unknown'
+  const email = firstBillingLeader?.email ?? 'Unknown'
   const isALeader = billingLeaders.findIndex((leader) => leader.id === viewerId) !== -1
   const handleClick = () => history.push(`/me/organizations/${orgId}`)
   const problem = `There in an unpaid invoice for ${teamName}.`
   const solution = isALeader
     ? `Head over to ${orgName} Settings to add a payment method`
-    : `Try reaching out to ${billingLeaderName}`
+    : `Try reaching out to ${billingLeaderName} at ${email}`
```
jordanh, line 39: If the HTML is set, can it override this logic and prefer to show the HTML over either of these templates?

packages/server/graphql/intranetSchema/mutations/lockTeams.ts
mattkrick: added a `message` arg

packages/client/modules/teamDashboard/components/UnpaidTeamModal/UnpaidTeamModal.tsx
mattkrick: ahhhh i think this comment is on an older commit, should work as you expect now

packages/client/modules/teamDashboard/components/UnpaidTeamModal/UnpaidTeamModal.tsx
```diff
@@ -1,40 +1,50 @@
...
+  const atmosphere = useAtmosphere()
+  const {history} = useRouter()
   const {viewerId} = atmosphere
   const {team} = viewer
   if (!team) return null
-  const {teamName, organization} = team
-  const {orgId, billingLeaders, orgName} = organization
-  const billingLeaderName = billingLeaders[0].preferredName
+  const {name: teamName, organization, lockMessageHTML} = team
+  if (lockMessageHTML) {
```
mattkrick, line 30: @jordanh i think this is what you were after?

jordanh: ah dope!

PR #4959, opened 2021-05-13
init sentry in gqlExecutor

tiffanyhan:
**AC:**
- run `yarn build` then `yarn start`
- go to graphiql and do the mutation below, which will trigger a server error
- go to [sentry staging](https://sentry.io/organizations/parabol/issues/?project=107915) and see that the error shows up

```
mutation {
  addAgendaItem(newAgendaItem: {
    content: "content",
    pinned: false,
    teamId: "teamid",
    teamMemberId: "team member id",
    sortOrder: 1,
    meetingId: "meeting id"
  }) {
    error {
      title
      message
    }
  }
}
```

### DIFF
--- a/packages/gql-executor/gqlExecutor.ts
+++ b/packages/gql-executor/gqlExecutor.ts
@@ -1,5 +1,5 @@
 import "./tracer"
-
+import '../server/initSentry'
 import Redis from 'ioredis'
 import {ServerChannel} from 'parabol-client/types/constEnums'
 


tianrunhe:
This fix makes all the sense to me and I tested it working. However I'm not sure if directly importing the settings from `server` is a good practice. That being said I don't have enough tension at this point. @mattkrick WDYT?

tianrunhe (approved):

mattkrick:
i mean, the _right_ way to do this is to have a `SharedServer` package with scripts & variables that are shared across all our microservices (all 2 of them :rofl:)

that seems like overkill for now, so I'm happy with this! wonderful find @tiffanyhan!

tiffanyhan:
> i mean, the right way to do this is to have a SharedServer package with scripts & variables that are shared across all our microservices (all 2 of them 🤣)

i was curious about the "right" way to do the import but i didn't find anything currently existing in our repo which would make sense to move the shared init into. thanks raising/resolving the tension! @tianrunhe @mattkrick

Issue #4963, opened 2021-05-13
Slack not sending notifications to private channels in prod


nickoferrall:
In development, I can send a notification to a private channel by doing the following:

1. Go to a private channel in Slack and click "Add Slack app"
2. In Parabol, add the Slack integration. Select a private channel
3. See that the default channel in Parabol is not the private channel but `@Parabol`
4. Change the default channel to the private channel
5. Create a meeting and see that the notification has been sent to the private channel

In production, step 4 is not possible as the private channel does not appear in the dropdown menu.

### AC

- [ ] Able to select a private channel in Parabol's dropdown menu
- [ ] Selecting a private channel in step 2 results in the default channel being the private channel rather than `@Parabol`

PR #4964, opened 2021-05-13
Send notifications to private channels

nickoferrall:
PR to resolve issue #4963 

In development, it is possible to send notifications to private Slack channels following the steps mentioned in #4963 but in production it is not. The Parabol-staging and the Parabol apps look identical so I'm not sure why this is.

However, [conversations.list](https://api.slack.com/methods/conversations.list) requires the groups:read scope to read private channels which is not activated. Hopefully, this is the issue. I've added the scope to the app and, if you agree, I'll resubmit the Slack app with the `groups:read` scope too.

Additionally, with this PR, users that select a private channel in the Slack pop-up, will see the private Slack channel in Parabol's dropdown immediately rather than needing to change the channel from `@Parabol`.  See below:

slack-private-channel

Required updates to Slack app:

- Add `groups:read` scope for private channels
- Add dummy interactivity request URL to remove warning when clicking on buttons with links
- Possibly add `im:read` scope to get Slackbot IM `channelId`, depending on #4988 

### AC

- [ ]  Able to select a private channel in Parabol's dropdown menu
- [ ]  Selecting a private channel in step 2 results in the default channel being the private channel rather than @Parabol


mattkrick:
LGTM! go ahead & submit the additional scope to slack & we'll merge this as soon as we get approved.
you can use the dev server if they need a test app, or maybe just link to this PR since the change is so minimal?

mattkrick:
TEST:
- [ ] have 2 groups setup in slack. call them `group A` `group B`
- [ ] integrate slack & when you choose a channel in the oauth window, set it to `group A`
- [ ] inside the team settings, change the channel to `group B`
- [ ] start a meeting. the message is sent to `group B`

nickoferrall:
I suggest that we wait to see if [my prediction here](https://github.com/ParabolInc/parabol/pull/4988/files#r636993699) is correct before resubmitting the app, i.e. updating notifications works in dev but not prod as we need the `im:read` scope. 

If it does work in prod, we can resubmit the app without the `im:read` scope. Otherwise, we should add it. 

I suggest waiting as it can take 4 weeks for the submission to be reviewed. If we adjust the scopes after the submission, it could prolong the process even further.

Issue #4848, opened 2021-04-01
add animations to meetings view


mattkrick:
it can be a little jarring if you're looking at the meetings view & a meeting gets added or removed.

AC
- [ ] when a meeting is removed, there is a nice transition
- [ ] when a meeting is added, there is a nice transition
- [ ] when opening the left nav causes a shift, there is a nice transition

Estimate: 8 hours

PR #4975, opened 2021-05-18
Meetings view animation

nickoferrall:
PR to resolve issue #4848 

This PR shows an animation when a meeting card is added, removed or the component size changes, e.g. when clicking the sidebar or resizing the browser. There is also a subtle opacity transition when the page initially loads.

Previously, when one meeting card in a row showed member avatars, the height of all the other cards in the row would also increase. In this PR, only cards with member avatars have an increased height to draw more attention to active meetings. See images below:

current-version
new-cards

### AC

- [ ]  when a meeting is removed, there is a nice transition
- [ ]  when a meeting is added, there is a nice transition
- [ ]  when opening the left nav causes a shift, there is a nice transition


packages/client/components/Dashboard/DashSidebar.tsx
```diff
@@ -23,7 +23,6 @@ const Nav = styled('nav')<{isOpen: boolean}>(({isOpen}) => ({
   [makeMinWidthMediaQuery(Breakpoint.DASH_BREAKPOINT_WIDEST)]: {
     bottom: 0,
     left: 0,
-    position: 'fixed',
```
nickoferrall, line 26: Dashboard content doesn't take into account the sidebar's position when it's `fixed`

enriquesanchez (approved):
Can't speak about the code, but as far as the animation goes, I really like it. It's subtle enough. 👍

enriquesanchez:
As for the card's height, a part of me feels that the unevenness in the grid introduces some awkward gaps and alignment. That said, the current behavior is not ideal either, as seen in the screenshots you shared. 

I'd like to ping @ackernaut and get his opinion on it, since he designed this view he may have some additional thoughts?

nickoferrall:
Thanks for the review! Tagging in @mattkrick to checkout the code

ackernaut:
As for the varying card heights, our options were:
1) Put the avatars on the name or meta rows. The problem here is that you have overflow sooner either on the avatar group or the copy when they share a row and 1 or both are wide.
2) Leave an empty space for avatars on all cards on the grid regardless of row. It uses a lot of space.
3) Make all the cards the same height per row if 1 or many cards have avatars on that row. Cards may not be consistent on the whole grid.
4) Let cards with avatars grow taller.

You can see some of the recent variants [here](https://www.figma.com/file/dfTQxhoKzohkCpxlztBbhS/Meetings-view?node-id=265%3A72135) and [here](https://www.figma.com/file/dfTQxhoKzohkCpxlztBbhS/Meetings-view?node-id=332%3A72920) in Figma.

Here we’re going from 3 to 4. I think either 3 or 4 are safe to try. My vote is option 4.

I like the first option though the cards get busy esp. with overflow. The layout is nuanced.

I don’t like the big cards anymore.

I kinda like having the cards with avatar break the grid a little, as my favorite form of poison.

FWIW I think the avatars make a single card stand out regardless of how we do it and having avatars on the cards seems to be well received so far, and useful. So I don’t think there’s a less wrong answer except maybe option 1 is not worth the headache to us or users.

ackernaut:
Bad idea: we could put the avatar group over the illustration in the top right. I think they might get missed in terms of visual impression when looking at this view quickly to get to the ‘right’ meeting, although with a healthy, light outline maybe there would be enough contrast?

nickoferrall:
Nice, my vote is for option 4 too! I like the quick concept and I think they do still stand out but I prefer the taller cards with avatars breaking the grid structure - my eyes are immediately drawn to the odd card(s) in the row. Plus it gives the avatars more room to breathe

packages/client/components/AvatarList.tsx
```diff
@@ -56,11 +67,14 @@ const AvatarList = (props: Props) => {
       {transitionChildren.length === 0 && emptyEl}
       {transitionChildren.map(({onTransitionEnd, child, status, displayIdx}) => {
         const {id: userId} = child
+        const transitionSpeed =
```
nickoferrall, line 70: This feels a bit fiddly but without it, the meeting card shrinks a bit too slowly when the AvatarList is removed, i.e. when a connected user leaves the meeting

packages/client/hooks/useOverflowAvatars.ts
```diff
@@ -40,7 +42,7 @@ const useOverflowAvatars = <T extends {id: string}>(
     displayIdx
   })) as ((T & {key: string; displayIdx: number}) | OverflowAvatar)[]
 
-  if (overflowCount > 0) {
+  if (overflowCount > 0 && !isInit) {
```
nickoferrall, line 45: This fixes a bug where the meeting card includes the `AvatarList` space even though there aren't any connected users. To reproduce:
- add a user to a meeting
- see the user in the `AvatarList` in the meeting card
- refresh the page
- remove user from the meeting
- see that the user avatar has not been removed from the meeting card. This is because the `transitionChildren` includes an overflow avatar

mattkrick: when you find edge cases like these it makes my heart swell with pride. seriously, great find!

i want you to dig a little deeper into the root cause of this
why is overflow avatar showing up at all? why is maxAvatars incorrect & how can we fix it?
when an overflow avatar is no longer needed, how can we guarantee that it cleans itself up instead of staying stuck on `EXITING`?

mattkrick (changes requested):
animations look great! few little code fixes & exercises & then it's good to go

packages/client/hooks/useOverflowAvatars.ts
nickoferrall: Thanks! 😄 

In `useTransition`, we store the previous items in a ref. This way, when an item is removed from an array, we can keep track of which item has been removed so that we can add animations.

In `useOverflowAvatars`, `totalItems` is calculated on the initial render whereas `maxAvatars` is counted on the second render. If there is a connected user and we refresh the page, `totalItems` is 1 but `maxAvatars` is the default 0 so an overflow avatar is added. On the second render, `maxAvatars` is updated, `overflowCount` is 0, the `overflowAvatar` is removed from `visibleAvatars` and `useTransition` updates the removed `overflowAvatar`'s status to Exiting. It still exists in `transitioningChildren` meaning the `minHeight` of the card is extended but we can't see the overflow avatar as its status is Exiting.

By cleaning up exiting items in `useTransition` when the component unmounts, the `overflowAvatar` is removed on the second render. Also, by calculating the `minHeight` of the meeting card using transitioning children that are not exiting, I can remove the `exitSpeed` prop! Also, I've added an initial render check to `overflowCount` to make sure this calculation is accurate.

packages/client/components/MeetingsDash.tsx
```diff
@@ -61,26 +49,59 @@ const Flash = styled('img')({
...
+      ...meeting,
+      key: meeting.createdAt,
+      displayIdx
+    }))
+  }, [teams])
+  const transitioningMeetings = useTransition(activeMeetings)
   const maybeBigDisplay = useBreakpoint(1900)
+  const isInit = useInitialRender()
+  const cardsPerRow = useCardsPerRow(meetingsDashRef)
+  const memodMeetings = useDeepEqual(activeMeetings)
```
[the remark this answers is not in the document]
nickoferrall: Without the `deepEqual` `activeMeetings` will trigger an infinite loop

packages/client/components/AvatarList.tsx
nickoferrall: Thanks, I really appreciate getting detailed explanations like this where you go through your thought pattern!

packages/client/components/MeetingsDash.tsx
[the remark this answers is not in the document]
mattkrick: oh! this is worth digging into. activeMeetings should change whenever the `teams` object changes. can you figure out what prop is changing every render? we'll want to figure this out.

hint: create a ref to hold the old team & your last hook is a useEffect that updates the old value to the new value. then, in the useMemo on L58 you can compare the old teams ref value to the new teams to see what is

packages/client/hooks/useTransition.ts
```diff
@@ -22,6 +22,16 @@ const useTransition = <T extends {key: Key}>(children: T[]) => {
   const previousTransitionChildrenRef = useRef<TransitionChild<T>[]>([])
   const forceUpdate = useForceUpdate()
 
+  useEffect(() => {
```
mattkrick, line 25: can you give me a test scenario so i can put this to the test? not sure what it does

nickoferrall: For sure! 

1. Comment out the useEffect hook I've added here in `useTransition` 
2. Remove the `isInit` check [here](https://github.com/ParabolInc/parabol/blob/e04f3fb9ed3fff3b4fb78d44ee9b818397c156c8/packages/client/hooks/useOverflowAvatars.ts#L36) 
3. Console log the transitioning avatars [here](https://github.com/ParabolInc/parabol/blob/e04f3fb9ed3fff3b4fb78d44ee9b818397c156c8/packages/client/hooks/useOverflowAvatars.ts#L49)
4. Open two windows. In window A, join an active meeting and see that the connected user is added to the `AvatarList` on the meeting card in window B
5. In window B, refresh the page and see that the transitioning avatars console log includes both the connected user and the overflow avatar
6. Replace `activeTChildren` with `transitioningChildren` in AvatarList [here](https://github.com/ParabolInc/parabol/blob/f21a82d9be54f1cbb2c6412e77b2ae45651306d9/packages/client/components/AvatarList.tsx#L57) and leave the meeting in window A. Even though there are no longer any connected users, you'll see that the meeting card height doesn't shrink as it recognises the overflow avatar with status `EXITING` as a connected user

By adding this clean-up function, I'm removing items that have a status of `EXITING` when the user refreshes the page.  The `isInit` check in `useOverflowAvatars` ensures that the `overflowCount` calculation is accurate even on the first render.

packages/client/components/MeetingsDash.tsx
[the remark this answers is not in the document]
nickoferrall: The infinite loop was being triggered by comparing the initial empty array of `meetings` in `useTopPerRow`. By adding a return statement when `meetings` is empty, I can use `activeMeetings` and remove `useDeepEqual`

packages/client/hooks/useTransition.ts
mattkrick: if i understand correctly, the root cause is that the overflow avatar is showing up invisibly when it shouldn't.

if useOverflowAvatars uses the following line, can you remove this hook?
`  const overflowCount = maxAvatars > 0 && totalItems > maxAvatars ? totalItems - maxAvatars + 1 : 0`

packages/client/components/MeetingsDash.tsx
```diff
@@ -62,25 +52,62 @@ const Flash = styled('img')({
...
   const maybeTabletPlus = useBreakpoint(Breakpoint.FUZZY_TABLET)
-  const maybeBigDisplay = useBreakpoint(1900)
+  const isInit = useInitialRender()
+  const cardsPerRow = useCardsPerRow(meetingsDashRef)
+  const topByRow = useTopPerRow(cardsPerRow, activeMeetings)
+  const hasMeetings = activeMeetings.length > 0
+  const totalRows = !activeMeetings.length ? 0 : Math.ceil(activeMeetings.length / cardsPerRow)
   useDocumentTitle('Meetings | Parabol', 'Meetings')
-  if (!viewer) return null
+  if (!viewer || isInit) return null
```
mattkrick, line 77: this is a good catch if you refresh the page, but the cards all animate into place when they shouldn't under the following scenario:

1. goto /meetings
2. click the first meeting
3. hit the browser back button
4. see the cards animate when they shouldn't

mattkrick (changes requested):
i know this thing is old, just a couple little things and good to merge!

packages/client/hooks/useTransition.ts
nickoferrall: Yep, that'll work for this use case but if a transitioningChild has a status of `EXITING` and is no longer needed, wouldn't we want to clean it up (based on [this suggestion](https://github.com/ParabolInc/parabol/pull/4975#discussion_r643562397))?

packages/client/components/MeetingsDash.tsx
nickoferrall: When going back, `cardsPerRow` could be calculated on the third render. On the second render before it was calculated it would return 1, the default value. By changing the default value to null and only showing the meeting cards once it's calculated, I can remove `isInit` and fix this bug

packages/client/hooks/useTransition.ts
nickoferrall: While the clean-up function might be useful in the future, it doesn't solve any existing issue right now so I've removed it

mattkrick:
clean changes! i think this is a strong improvement over what we have today & if there are any edge cases that crop up we can fix those in a later PR.
nice work animation master!

PR #4980, opened 2021-05-19
Tier update pg

tianrunhe:
Changes:

1. Update `setUserTierForUserIds` to update the `tier` column to PG. As the `Organization` table has not been migrated to PG, we need to join with RethinkDB today.
2. A backfill script for the `tier` column for `User` table in PG.

Tests:
- [ ] Run the backfill script with `yarn postmigrate` and observes that `tier` columns in PG are backfilled with non `personal` tiers
- [ ] Upgrade a local user to `pro` tier and observes the `tier` column in PG has been updated to `pro` as well

### DIFF
--- a/packages/server/utils/setUserTierForUserIds.ts
+++ b/packages/server/utils/setUserTierForUserIds.ts
@@ -1,7 +1,9 @@
 import getRethink from '../database/rethinkDriver'
 import db from '../db'
+import getPg from '../postgres/getPg'
 import {TierEnum} from '../postgres/queries/generated/updateUserQuery'
-import updateUser from '../postgres/queries/updateUser'
+import {updateUserTiersQuery} from '../postgres//queries/generated/updateUserTiersQuery'
+import catchAndLog from '../postgres/utils/catchAndLog'
 import segmentIo from './segmentIo'
 
 const setUserTierForUserIds = async (userIds: string[]) => {
@@ -58,10 +60,12 @@ const setUserTierForUserIds = async (userIds: string[]) => {
         })
         .default('personal')
     }))
-    .run()) as {group: string; reduction: string[]; tier: TierEnum}[]
-  userTiers.forEach(({group: userId, tier}) => {
-    updateUser({tier}, userId)
-  })
+    .map((row) => ({
+      id: row('group'),
+      tier: row('tier')
+    }))
+    .run()) as {id: string; tier: TierEnum}[]
+  await catchAndLog(() => updateUserTiersQuery.run({users: userTiers}, getPg()))
 
   await Promise.all(userIds.map((userId) => db.clear('User', userId)))
   const users = await db.readMany('User', userIds)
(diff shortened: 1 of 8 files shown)


packages/server/utils/setUserTierForUserIds.ts
```diff
@@ -35,6 +39,34 @@ const setUserTierForUserIds = async (userIds: string[]) => {
...
+        .reduce((_left, right) => {
+          return r.branch(right.eq('enterprise'), 'enterprise', right.eq('pro'), 'pro', 'personal')
+        })
+        .default('personal')
+    }))
+    .map((row) => ({
+      id: row('group'),
+      tier: row('tier')
+    }))
+    .run()) as {id: string; tier: TierEnum}[]
```
tiffanyhan, line 67: i noticed that this query might not be equivalent to the original, which i confirmed by testing locally.

to see the difference:
- login as a new user locally
- create many new teams off of many new orgs, e.g. (org2, team2), (org3, team3), etc...
- go to one of the random orgs and upgrade to pro
- check in rethinkdb that the user row will always have the "highest" tier of any orgs they're a part of
- check in pg and see that the tier will *probably* still be "personal". (it's possible by luck the tier could be "pro" too, but the point being that it's not deterministic)

i think it will be fine to just keep the original code, and only change getting `OrganizationUser`s by directly passing in the `userIds` as you did here. can you make those changes?

tiffanyhan (changes requested):
looking good! the only thing was the query in `setUserTierForUserIds` isn't quite logically equivalent to the original. see my comment for how to validate and test

packages/server/utils/setUserTierForUserIds.ts
tianrunhe: That's a great find, thank you!!! I think the non-deterministic logic come from this part:

```typescript
row('reduction')
  .reduce((_left, right) => {
    return r.branch(right.eq('enterprise'), 'enterprise', right.eq('pro'), 'pro', 'personal')
  })
  .default('personal')
```

So `['pro', 'personal']` would just yield `'personal'` whereas `['personal', 'pro']` would give you `'pro'`...

Anyway I took your advice and just copied the original logic directly and modified a little bit to make it work with PG.

tiffanyhan: oh no! it looks like i missed something. for the query to get `userTiers`, can you remove the dependency on rethink `User` table? the goal is that we want to be able to drop rethink User table, and we can't do that if any lingering queries use the table. Let's just get `OrganizationUser`s from the passed `userIds` arg using the `OrganizationUser.userId` index (i think this is what you were doing at first), and same thing for the `id` property in the mapped object.

tiffanyhan (changes requested):

packages/server/utils/setUserTierForUserIds.ts
tianrunhe: You are so right! Guess I was in hurry fixing the bug in the logic and then completely forgot why we needed to rewrite this lol.

tiffanyhan: cool, tested and looks good!

tiffanyhan (approved):

mattkrick:
:fire:

Issue #4911, opened 2021-04-26
Improves context and layout of Slack notifications


enriquesanchez:
This is an implementation issue addressing #3764.

### Background

>Right now, our Slack notifications are pretty plain and not very attractive [...]

Instead, users should:

> - Understand what organization, team and meeting they're being invited to from a Slack notification
> - Join the Slack meeting with one click
> - See Slack notifications for meetings as distinct from other Slack messages

### Acceptance criteria

The goal is to make our notifications more informative, more engaging, and overall friendlier. We also want to display a different format depending on the meeting stage (start, phase initiated, end).

The new notifications should follow the format from [the Slack API](https://api.slack.com/):

- Start with a short message stating one of the following:
  - `Meeting started 👋`
  - `Meeting completed 🎉`
  - `[PHASE NAME] has begun ⏳`
 - Follow with the meeting info: `Team` name and `Meeting` name

If the **meeting is starting**:
- display the shortened `link` for the meeting 
- display a CTA button with the label `Join meeting` in Slack's default green (`#007a5a`)

![](https://user-images.githubusercontent.com/3276087/114430839-77a19a80-9b84-11eb-92b4-c8fa4062f597.png)

If the **meeting has ended**:
- display the summary of the meeting (e.g. `It lasted 54 minutes and generated 13 tasks, 9 agenda items
and 7 comments.`)
- Follow with two actions: `See discussion` and `Review summary`. These two actions should point to the same places as the current meeting summary we show in the users' timeline in Parabol.

![](https://user-images.githubusercontent.com/3276087/114430862-7c664e80-9b84-11eb-961a-151364cea34f.png)

If a **meeting phase has begun**:
- display the phase deadline (e.g. `You have until April 24 at 12:50 PM to complete it.`)
- provide the shortened link for the meeting
- display a green CTA with the label `Open meeting`
- followed by the secondary action `Set reminder`
- `Set reminder` should make use of Slack's own `/remind` feature

![](https://user-images.githubusercontent.com/3276087/114430915-88521080-9b84-11eb-8b66-a851eec67b88.png)

![](https://user-images.githubusercontent.com/3276087/114431513-2f36ac80-9b85-11eb-870d-b4d375674096.png)

### Estimated effort:
12 hrs

nickoferrall:
[...]

Issue #4983, opened 2021-05-19
Comment count doesn't exist in Poker meeting


nickoferrall:
In the Timeline, we show a summary sentence that includes the comment count: `You voted on 2 stories and added 0 comments.`

In `endSprintPoker`, we're not currently saving the comment count to the db. In [PokerMeeting](https://github.com/ParabolInc/parabol/blob/774c4cf2dfbceaa2d56910583e325080c55deae0/packages/server/graphql/types/PokerMeeting.ts#L21), if the comment count is undefined, we return 0, so comment count always appears as 0 in the Timeline.

### AC

- [ ] Save the `commentCount` in the db when ending a Poker meeting
- [ ] Write a migration that adds the comment count to previous Poker meetings
- [ ] Update Slack end poker meeting notification with the comment count

### EE

- 3 hours

PR #4988, opened 2021-05-21
Make Slack notifications pretty

nickoferrall:
PR to resolve issue #4911. 

It also resolves an undocumented bug where timer adjustments send new notifications rather than update the previous one. I've commented inline about this. 

A few additional notes:

- The issue mentioned using Slack's default green `#007a5a` but with Slack's blocks, you choose whether it's a primary button, regular or danger button rather than specify the colour. On a dark mode theme, we get the darker button and in a light mode, it's a lighter button
- I haven't come across a good, free tool for creating short links so I've included the regular link without utm params which is still shortish
- I've added a dummy interactivity request URL to the [Slack app](https://api.slack.com/apps/A01EGHDU7PZ/interactive-messages?). Otherwise, a warning sign shows after clicking the button. See Slack issue [here](https://github.com/CircleCI-Public/slack-orb/issues/188#issuecomment-708988009)
- I noticed that the comment count is currently `undefined` when ending poker meetings. I've created an issue to fix this [here](https://github.com/ParabolInc/parabol/issues/4983). When it's fixed, we can add comment count to the poker notifications

### To test

- Go to the `.env` file and change the `SLACK_CLIENT_ID` & `SLACK_CLIENT_SECRET` to the client ID and secret found in the `Parabol-staging` bot [here](https://api.slack.com/apps/A01EGHDU7PZ)
- Kill and restart the server with `yarn dev`
- You may need to install the Parabol-staging Slackbot in Slack
- Integrate Slack in Parabol (remove it first if it's already integrated),
- See the new notifications styles when you create, set a timer & end a meeting
- If the notifications are sent to `@Parabol-staging` the previous notification should be updated when you adjust the timer rather than sending a new notification

### DIFF
--- a/packages/client/utils/SlackManager.ts
+++ b/packages/client/utils/SlackManager.ts
@@ -57,10 +57,21 @@ export interface SlackConversation {
   locale: string
 }
 
+export interface SlackIM {
+  created: number
+  id: string
+  is_im: true
+  is_org_shared: boolean
+  is_user_deleted: boolean
+  priority: number
+  user: string
+}
+
 interface ConversationListResponse {
   ok: true
-  channels: SlackConversation[]
+  channels: SlackConversation[] | SlackIM[]
 }
+
 interface ConversationJoinResponse {
   ok: true
   channel: {
@@ -130,7 +141,7 @@ interface ConversationInfoResponse {
   channel: SlackConversation
 }
 
-type ConversationType = 'public_channel' | 'private_channel'
+type ConversationType = 'public_channel' | 'private_channel' | 'im'
 
 abstract class SlackManager {
   static SCOPE = 'incoming-webhook,channels:read,channels:join,chat:write,users:read'
@@ -213,9 +224,9 @@ abstract class SlackManager {
     return this.post<PostMessageResponse>('https://slack.com/api/chat.postMessage', payload)
   }
 
-  updateMessage(channelId: string, text: string, ts: string) {
+  updateMessage(channelId: string, blocks: string | Array<{type: string}>, ts: string) {
     return this.get<UpdateMessageResponse>(
-      `https://slack.com/api/chat.update?token=${this.token}&channel=${channelId}&text=${text}&ts=${ts}`
+      `https://slack.com/api/chat.update?token=${this.token}&channel=${channelId}&blocks=${blocks}&ts=${ts}`
     )
   }
 }
--- a/packages/client/utils/SlackManager.ts
+++ b/packages/client/utils/SlackManager.ts
@@ -225,8 +225,9 @@ abstract class SlackManager {
   }
 
   updateMessage(channelId: string, blocks: string | Array<{type: string}>, ts: string) {
+    const stringifiedBlocks = JSON.stringify(blocks)
     return this.get<UpdateMessageResponse>(
-      `https://slack.com/api/chat.update?token=${this.token}&channel=${channelId}&blocks=${blocks}&ts=${ts}`
+      `https://slack.com/api/chat.update?token=${this.token}&channel=${channelId}&blocks=${stringifiedBlocks}&ts=${ts}`
     )
   }
 }
--- a/packages/client/utils/SlackManager.ts
+++ b/packages/client/utils/SlackManager.ts
@@ -225,9 +225,9 @@ abstract class SlackManager {
   }
 
   updateMessage(channelId: string, blocks: string | Array<{type: string}>, ts: string) {
-    const stringifiedBlocks = JSON.stringify(blocks)
+    const newBlocks = typeof blocks === 'string' ? blocks : JSON.stringify(blocks)
     return this.get<UpdateMessageResponse>(
-      `https://slack.com/api/chat.update?token=${this.token}&channel=${channelId}&blocks=${stringifiedBlocks}&ts=${ts}`
+      `https://slack.com/api/chat.update?token=${this.token}&channel=${channelId}&blocks=${newBlocks}&ts=${ts}`
     )
   }
 }
--- a/packages/server/graphql/mutations/endRetrospective.ts
+++ b/packages/server/graphql/mutations/endRetrospective.ts
@@ -120,9 +120,11 @@ export default {
       dataLoader.get('teams').load(teamId),
       dataLoader.get('teamMembersByTeamId').load(teamId),
       removeEmptyTasks(meetingId),
-      dataLoader.get('meetingTemplates').load(templateId),
-      finishRetroMeeting(completedRetrospective, dataLoader)
+      dataLoader.get('meetingTemplates').load(templateId)
     ])
+    // wait for removeEmptyTasks before finishRetroMeeting & wait for meeting stats
+    // to be generated in finishRetroMeeting before sending Slack notifications
+    await finishRetroMeeting(completedRetrospective, dataLoader)
     endSlackMeeting(meetingId, teamId, dataLoader).catch(console.log)
     sendMeetingEndToSegment(completedRetrospective, meetingMembers as MeetingMember[], template)
     sendNewMeetingSummary(completedRetrospective, context).catch(console.log)
--- a/packages/server/graphql/mutations/endRetrospective.ts
+++ b/packages/server/graphql/mutations/endRetrospective.ts
@@ -120,10 +120,10 @@ export default {
       dataLoader.get('teams').load(teamId),
       dataLoader.get('teamMembersByTeamId').load(teamId),
       removeEmptyTasks(meetingId),
-      dataLoader.get('meetingTemplates').load(templateId)
+      dataLoader.get('meetingTemplates').load(templateId),
+      finishRetroMeeting(completedRetrospective, dataLoader)
     ])
     endSlackMeeting(meetingId, teamId, dataLoader).catch(console.log)
-    finishRetroMeeting(completedRetrospective, dataLoader)
     sendMeetingEndToSegment(completedRetrospective, meetingMembers as MeetingMember[], template)
     sendNewMeetingSummary(completedRetrospective, context).catch(console.log)
     const events = teamMembers.map(
--- a/packages/server/graphql/mutations/helpers/makeSlackBlocks.ts
+++ b/packages/server/graphql/mutations/helpers/makeSlackBlocks.ts
@@ -1,3 +1,5 @@
+import generateUID from '../../../generateUID'
+
 export const makeSection = (text: string) => ({
   type: 'section',
   text: {
@@ -22,7 +24,7 @@ type ButtonInput = {
 
 export const makeButtons = (buttons: ButtonInput[]) => ({
   type: 'actions',
-  block_id: `actionblock${new Date().toISOString()}`,
+  block_id: `actionblock${generateUID()}`,
   elements: buttons.map((button) => ({
     type: 'button',
     text: {
--- a/packages/server/graphql/mutations/helpers/makeSlackBlocks.ts
+++ b/packages/server/graphql/mutations/helpers/makeSlackBlocks.ts
@@ -1,3 +1,11 @@
+export const makeSection = (text: string) => ({
+  type: 'section',
+  text: {
+    type: 'mrkdwn',
+    text
+  }
+})
+
 export const makeSections = (fields: string[]) => ({
   type: 'section',
   fields: fields.map((field) => ({
--- a/packages/server/graphql/mutations/helpers/makeSlackBlocks.ts
+++ b/packages/server/graphql/mutations/helpers/makeSlackBlocks.ts
@@ -1,0 +1,27 @@
+export const makeSections = (fields: string[]) => ({
+  type: 'section',
+  fields: fields.map((field) => ({
+    type: 'mrkdwn',
+    text: field
+  }))
+})
+
+type ButtonInput = {
+  text: string
+  url: string
+  type?: 'primary' | 'danger'
+}
+
+export const makeButtons = (buttons: ButtonInput[]) => ({
+  type: 'actions',
+  block_id: `actionblock${new Date().toISOString()}`,
+  elements: buttons.map((button) => ({
+    type: 'button',
+    text: {
+      type: 'plain_text',
+      text: button.text
+    },
+    style: button.type,
+    url: button.url
+  }))
+})
--- a/packages/server/graphql/mutations/helpers/notifySlack.ts
+++ b/packages/server/graphql/mutations/helpers/notifySlack.ts
@@ -121,12 +121,7 @@ export const startSlackMeeting = async (
 
 const getSummaryText = (meeting: MeetingRetrospective | MeetingAction | MeetingPoker) => {
   if (meeting.meetingType === 'retrospective') {
-    const {
-      commentCount = 0,
-      reflectionCount = 0,
-      topicCount = 0,
-      taskCount = 0
-    } = meeting as MeetingRetrospective
+    const {commentCount = 0, reflectionCount = 0, topicCount = 0, taskCount = 0} = meeting
     return `Your team shared ${reflectionCount} ${plural(
       reflectionCount,
       'reflection'
@@ -135,13 +130,7 @@ const getSummaryText = (meeting: MeetingRetrospective | MeetingAction | MeetingP
       'comment'
     )} and created ${taskCount} ${plural(taskCount, 'task')}.`
   } else if (meeting.meetingType === 'action') {
-    const {
-      createdAt,
-      endedAt,
-      agendaItemCount = 0,
-      commentCount = 0,
-      taskCount = 0
-    } = meeting as MeetingAction
+    const {createdAt, endedAt, agendaItemCount = 0, commentCount = 0, taskCount = 0} = meeting
     const meetingDuration = relativeDate(createdAt, {
       now: endedAt,
       max: 2,
@@ -156,7 +145,7 @@ const getSummaryText = (meeting: MeetingRetrospective | MeetingAction | MeetingP
       'comment'
     )}.`
   } else {
-    const estimatePhase = (meeting as MeetingPoker).phases.find(
+    const estimatePhase = meeting.phases.find(
       (phase) => phase.phaseType === 'ESTIMATE'
     ) as EstimatePhase
     const stages = estimatePhase.stages
--- a/packages/server/graphql/mutations/helpers/notifySlack.ts
+++ b/packages/server/graphql/mutations/helpers/notifySlack.ts
@@ -21,6 +21,7 @@ import MeetingAction from '../../../database/types/MeetingAction'
 import MeetingPoker from '../../../database/types/MeetingPoker'
 import plural from 'parabol-client/utils/plural'
 import EstimatePhase from '../../../database/types/EstimatePhase'
+import {makeSections, makeButtons} from './makeSlackBlocks'
 
 const getSlackDetails = async (
   event: SlackNotificationEvent,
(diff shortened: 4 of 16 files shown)


packages/server/graphql/mutations/helpers/notifySlack.ts
```diff
@@ -97,11 +104,55 @@ export const startSlackMeeting = async (
     utm_campaign: 'invitations'
   }
   const options = {searchParams}
-  const team = await dataLoader.get('teams').load(teamId)
+  const [team, meeting] = await Promise.all([
+    dataLoader.get('teams').load(teamId),
+    dataLoader.get('newMeetings').load(meetingId)
+  ])
+  const meetingUrlWithParams = makeAppURL(appOrigin, `meet/${meetingId}`, options)
+  const meetingUrl = makeAppURL(appOrigin, `meet/${meetingId}`)
```
nickoferrall, line 112: To keep the link shortish, I've removed the utm params. Hopefully, this isn't too problematic for our analytics?

packages/server/graphql/mutations/helpers/notifySlack.ts
```diff
@@ -131,25 +203,25 @@ const upsertSlackMessage = async (
     const {channel} = convoInfo
     const {latest} = channel
     if (latest) {
-      const {ts, bot_profile} = latest
-      const {name} = bot_profile
-      if (name === 'Parabol') {
```
nickoferrall, line 136: Removed the check for the `Parabol` bot name as `latest` only exists in the DM and we handle these checks server side. Small advantages being it's slightly cleaner and we don't need to change the name to `Parabol-staging` in development

packages/server/graphql/mutations/helpers/notifySlack.ts
```diff
@@ -131,25 +203,25 @@ const upsertSlackMessage = async (
...
+      const {ts} = latest
+      const timestamp = new Date(Number.parseFloat(ts) * 1000)
+      const ageThresh = new Date(Date.now() - ms('5m'))
+      if (timestamp >= ageThresh) {
+        // blocks need to be stringified when updating but not posting
+        const stringifiedBlocks = JSON.stringify(blocks)
+        const res = await manager.updateMessage(channelId, stringifiedBlocks, ts)
+        if (!res.ok) {
+          console.error(res.error)
+          const postRes = await manager.postMessage(channelId, blocks)
```
nickoferrall, line 215: I noticed that when changing the timer, a new notification is created rather than the previous one being edited. The reason is in `addSlackAuth`, we're using the Slack user id as the `channelId` so that we can send messages from the `@Parabol`. However, to be able to update a message, we need the DM `channelId`. 

Requesting user info on the bot or bot info doesn't return this so I've grabbed all the IMs and found the IM `channelId` with the Slackbot's user id (see [here](https://github.com/ParabolInc/parabol/blob/a78aa53259e1c303e5f0cd82ec393bf46d346219/packages/server/graphql/mutations/addSlackAuth.ts#L119)). 

To be able to read all the IM channels, I think we should need the `im:read` [scope](https://api.slack.com/scopes/im:read). While this works in development without the scope, I suspect this is the same bug as the private channels one where the dev & prod bots require different scopes.

I haven't adjusted the scopes in this PR. If the `im:read` scope is required, it will simply post new notifications when the timer is adjusted rather than update the old one.

Instead, I'll adjust the scope in [this PR](https://github.com/ParabolInc/parabol/pull/4964) so that we don't need to wait until the Slack app is approved to merge this PR.

enriquesanchez:
@nickoferrall I was not able to get this to work on my end, but I'm seeing the notifications you created on the `test-slack-bot` channel. It looks much better 🎉

>The issue mentioned using Slack's default green #007a5a but with Slack's blocks, you choose whether it's a primary button, regular or danger button rather than specify the colour. On a dark mode theme, we get the darker button and in a light mode, it's a lighter button

This is totally OK. Thanks for looking into it.

>I haven't come across a good, free tool for creating short links so I've included the regular link without utm params which is still shortish

We already use our own shortlinks, such as `https://prbl.in/J12QXX12T4IZ` when you click on "Invite to team". Can we use that for the notifications as well?

enriquesanchez:
One more thing:

Screen Shot 2021-05-21 at 3 26 12 p m


Can we move the `Meeting` value a bit more to the left, so it's closer to `Team`? Something like this:

Screen Shot 2021-05-21 at 3 26 12 p m

nickoferrall:
> We already use our own shortlinks when you click on "Invite to team"

Ah thanks! I checked this in dev where we show the regular link but I see in production we have the short link. I've pushed an update with this

> Can we move the Meeting value a bit more to the left, so it's closer to Team?

Slack blocks aren't really built for customisation so this is tricky. You define fields that you want to use and they apply their own [styling](https://app.slack.com/block-kit-builder/T08FL6336#%7B%22blocks%22:%5B%7B%22type%22:%22section%22,%22text%22:%7B%22type%22:%22mrkdwn%22,%22text%22:%22You%20have%20a%20new%20request:%5Cn*%3CfakeLink.toEmployeeProfile.com%7CFred%20Enriquez%20-%20New%20device%20request%3E*%22%7D%7D,%7B%22type%22:%22section%22,%22fields%22:%5B%7B%22type%22:%22mrkdwn%22,%22text%22:%22*Type:*%5CnComputer%20(laptop)%22%7D,%7B%22type%22:%22mrkdwn%22,%22text%22:%22*When:*%5CnSubmitted%20Aut%2010%22%7D,%7B%22type%22:%22mrkdwn%22,%22text%22:%22*Last%20Update:*%5CnMar%2010,%202015%20(3%20years,%205%20months)%22%7D,%7B%22type%22:%22mrkdwn%22,%22text%22:%22*Reason:*%5CnAll%20vowel%20keys%20aren't%20working.%22%7D,%7B%22type%22:%22mrkdwn%22,%22text%22:%22*Specs:*%5Cn%5C%22Cheetah%20Pro%2015%5C%22%20-%20Fast,%20really%20fast%5C%22%22%7D%5D%7D,%7B%22type%22:%22actions%22,%22elements%22:%5B%7B%22type%22:%22button%22,%22text%22:%7B%22type%22:%22plain_text%22,%22emoji%22:true,%22text%22:%22Approve%22%7D,%22style%22:%22primary%22,%22value%22:%22click_me_123%22%7D,%7B%22type%22:%22button%22,%22text%22:%7B%22type%22:%22plain_text%22,%22emoji%22:true,%22text%22:%22Deny%22%7D,%22style%22:%22danger%22,%22value%22:%22click_me_123%22%7D%5D%7D%5D%7D). Rather than show two fields, i.e. Team and Meeting, we could show one field with Team and Meeting separated by a bunch of spaces but this would be kinda hacky and difficult to maintain on desktop and mobile. Unless you have a lot of tension with this one, I suggest we stick with Slack's format

enriquesanchez:
@nickoferrall Thank you!

>we could show one field with Team and Meeting separated by a bunch of spaces but this would be kinda hacky and difficult to maintain on desktop and mobile. Unless you have a lot of tension with this one, I suggest we stick with Slack's format

Agreed, let's go with the defaults Slack allows. Thank you for looking into it.

enriquesanchez (approved):

packages/server/graphql/mutations/endRetrospective.ts
```diff
@@ -120,10 +120,10 @@ export default {
       dataLoader.get('teams').load(teamId),
       dataLoader.get('teamMembersByTeamId').load(teamId),
       removeEmptyTasks(meetingId),
-      dataLoader.get('meetingTemplates').load(templateId)
+      dataLoader.get('meetingTemplates').load(templateId),
+      finishRetroMeeting(completedRetrospective, dataLoader)
```
mattkrick, line 124: this isn't safe because we need to wait for `removeEmptyTasks` probably worth reverting the change & adding a comment as to why it's done in serial

packages/client/utils/SlackManager.ts
```diff
@@ -213,9 +224,9 @@ abstract class SlackManager {
     return this.post<PostMessageResponse>('https://slack.com/api/chat.postMessage', payload)
   }
 
-  updateMessage(channelId: string, text: string, ts: string) {
+  updateMessage(channelId: string, blocks: string | Array<{type: string}>, ts: string) {
```
mattkrick, line 227: blocks must be a string (you pass it in correctly, but the type can't be an array)

packages/server/graphql/mutations/helpers/makeSlackBlocks.ts
```diff
@@ -0,0 +1,35 @@
...
+
+type ButtonInput = {
+  text: string
+  url: string
+  type?: 'primary' | 'danger'
+}
+
+export const makeButtons = (buttons: ButtonInput[]) => ({
+  type: 'actions',
+  block_id: `actionblock${new Date().toISOString()}`,
```
mattkrick, line 25: does this have to be a UID? if so, it won't be unique enough because this could execute twice in the same millisecond.
i'd just use `generateUID`

packages/server/graphql/mutations/helpers/notifySlack.ts
```diff
@@ -131,25 +202,25 @@ const upsertSlackMessage = async (
...
-          const res = await manager.updateMessage(channelId, slackText, ts)
-          if (!res.ok) {
-            console.error(res.error)
-          }
-          return
+      const {ts} = latest
+      const timestamp = new Date(Number.parseFloat(ts) * 1000)
+      const ageThresh = new Date(Date.now() - ms('5m'))
+      if (timestamp >= ageThresh) {
+        // blocks need to be stringified when updating but not posting
```
mattkrick, line 209: alternatively, you could make `updateMessage` take the same type as postMessage so they have the same call signature

packages/server/graphql/mutations/helpers/notifySlack.ts
mattkrick: just outta curiosity, have you tried querying that one specific channel with something like conversations.info? it'd be absolutely amazing to avoid an extra scope if possible

mattkrick (changes requested):
couple light changes, shouldn't take long! if this misses the ship window we can always do a 2nd release just for this

nickoferrall:
> just outta curiosity, have you tried querying that one specific channel with something like conversations.info? it'd be absolutely amazing to avoid an extra scope if possible

Yup, I've tried [conversations.info](https://api.slack.com/methods/conversations.info) (just tested it again to be sure) but it requires a `channelId` which we don't have for the Slackbot. The incoming webhook response doesn't provide the Slackbot `channelId` and nor does [auth.test](https://api.slack.com/methods/auth.test), [bots.info](https://api.slack.com/methods/bots.info) or [users.info](https://api.slack.com/methods/users.info). It looks like we have to search through the IMs to get the Slackbot `channelId`. 

Alternatively, if we want to avoid requesting the `im:read` scope, we could scrap the update message logic. The update message logic happens when a user updates the timer and the default Slack channel is the `@Parabol` bot. If it's a default Slack channel is a group channel, we send a new message as we don't get the timestamp of the latest message. Sparing a user a notification from `@Parabol` is nice but the value add is quite small IMO

packages/client/utils/SlackManager.ts
nickoferrall: Thanks! I've moved the stringify to `updateMessage` so that we can have the same call signature here and keep the array type

packages/server/graphql/mutations/endRetrospective.ts
nickoferrall: Ahh yes! I guess this is the same for [check-ins too](https://github.com/ParabolInc/parabol/blob/8d3826bfaedeb8d17b08afbd9bda56866eace4f6/packages/server/graphql/mutations/endCheckIn.ts#L231) where we're removing empty tasks and finishing the meeting in the same `Promise.all`. I've moved `finishCheckInMeeting` too and added comments in both files 👍

packages/client/utils/SlackManager.ts
```diff
@@ -213,9 +224,10 @@ abstract class SlackManager {
...
+    const stringifiedBlocks = JSON.stringify(blocks)
```
mattkrick, line 228: only stringify if array, otherwise we'll get double quotes

packages/server/graphql/mutations/helpers/notifySlack.ts
```diff
@@ -97,11 +104,54 @@ export const startSlackMeeting = async (
...
+    dataLoader.get('newMeetings').load(meetingId)
+  ])
   const meetingUrl = makeAppURL(appOrigin, `meet/${meetingId}`, options)
-  const slackText = `${team.name} has started a meeting!\n To join, click here: ${meetingUrl}`
-  notifySlack('meetingStart', dataLoader, teamId, slackText).catch(console.log)
+  const button = {text: 'Join meeting', url: meetingUrl, type: 'primary'} as const
+  const blocks = [
+    makeSection('Meeting started :wave: '),
+    makeSections([`*Team:*\n${team.name}`, `*Meeting:*\n${meeting.name}`]),
+    makeSection(`*Link:*\n<${meetingUrl}|https:/prbl.in/${meetingId}>`),
```
mattkrick, line 116: is `https:/prbl.in/${meetingId}` valid?

packages/server/graphql/mutations/helpers/notifySlack.ts
```diff
@@ -97,11 +104,54 @@ export const startSlackMeeting = async (
...
+    makeButtons([button])
+  ]
+  notifySlack('meetingStart', dataLoader, teamId, blocks).catch(console.log)
+}
+
+const getSummaryText = (meeting: MeetingRetrospective | MeetingAction | MeetingPoker) => {
+  if ('reflectionCount' in meeting) {
```
mattkrick, line 123: for discriminated unions like this it's a little cleaner to switch on the type
you can use a switch or a condition, both give full type safety as long as you use object access & that way if reflectionCount goes away, we don't have to change the logic:

```ts
  switch (meeting.meetingType) {
    case 'retrospective':
      
  }
  if (meeting.meetingType === 'retrospective') {

  }

File: packages/server/graphql/mutations/helpers/notifySlack.ts [hunk 11]
```diff
@@ -111,15 +161,36 @@ export const endSlackMeeting = async (meetingId, teamId, dataLoader: DataLoaderW
...
   const options = {searchParams}
-  const team = await dataLoader.get('teams').load(teamId)
+  const [team, meeting] = await Promise.all([
+    dataLoader.get('teams').load(teamId),
+    dataLoader.get('newMeetings').load(meetingId)
+  ])
   const summaryUrl = makeAppURL(appOrigin, `new-summary/${meetingId}`, options)
-  const slackText = `The meeting for ${team.name} has ended!\n Check out the summary here: ${summaryUrl}`
-  notifySlack('meetingEnd', dataLoader, teamId, slackText).catch(console.log)
+  const meetingUrl = makeAppURL(appOrigin, `meet/${meetingId}/discuss/1`)
```
mattkrick, line 169: discuss only exists on retros, what about check-ins?

i see a lot of conditional logic based on meeting type. might be worthwhile branching this logic based on meetingType.

e.g. `const blocks = makeBlocksForMeeting()`
makeBlocksForMeeting = () => {
  switch (meeting.meetingType) {
    case 'retro': 
    case 'action':
      return foo
  }
}
```

File: packages/server/graphql/mutations/helpers/notifySlack.ts [hunk 12]
```diff
@@ -177,15 +250,17 @@ export const notifySlackTimeLimitStart = async (
...
     )}^{date_short_pretty} at {time}|${fallback}>* to complete it.`
-    const cta = `Check it out: ${meetingUrl}`
-    const slackText = [situation, constraint, cta].join('\n')
-    upsertSlackMessage(slackDetail, slackText).catch(console.error)
+    const button = {text: 'Open meeting', url: meetingUrl, type: 'primary'} as const
+    const blocks = [
+      makeSection(`The *${phaseLabel} Phase* has begun :hourglass_flowing_sand:`),
+      makeSections([`*Team:*\n${teamName}`, `*Meeting:*\n${meetingName}`]),
+      makeSection(constraint),
+      makeSection(`*Link:*\n<${meetingUrl}|https:/prbl.in/${meetingId}>`),
```
mattkrick, line 261: https:/prbl.in/${meetingId} i dont think this is valid

packages/server/graphql/mutations/helpers/notifySlack.ts
nickoferrall: The url isn't valid but it's just a hyperlink. When the user clicks on it, they'll be redirected to the regular longer link with UTM params. The url could be anything but I kept it in the same format as our team invite link. 

If we want to use a valid short url using our existing format of `https:/prbl.in/${token}`, we'd run into a security issue as the token works as an invite to a team and a user could send the Slack notification to a public channel

packages/server/graphql/mutations/helpers/notifySlack.ts
nickoferrall: I tried to use the meeting type initially but it means we have to cast the type (which we do in [joinMeeting](https://github.com/ParabolInc/parabol/blob/4ce15d9ba712b4004d6361be3c5daee5714e0ad9/packages/server/graphql/mutations/joinMeeting.ts#L29)). We don't need to cast the type when using  `reflectionCount` but I definitely agree that using a field that could be removed at some point isn't ideal

packages/server/graphql/mutations/helpers/notifySlack.ts
nickoferrall: I responded to this in the PR thread but adding it here incase you didn't see it: 

Yup, I've tried [conversations.info](https://api.slack.com/methods/conversations.info) (just tested it again to be sure) but it requires a `channelId` which we don't have for the Slackbot. The incoming webhook response doesn't provide the Slackbot `channelId` and nor does [auth.test](https://api.slack.com/methods/auth.test), [bots.info](https://api.slack.com/methods/bots.info) or [users.info](https://api.slack.com/methods/users.info). It looks like we have to search through the IMs to get the Slackbot `channelId`. 

Alternatively, if we want to avoid requesting the `im:read` scope, we could scrap the update message logic. The update message logic happens when a user updates the timer and the default Slack channel is the `@Parabol` bot. If it's a default Slack channel is a group channel, we send a new message as we don't get the timestamp of the latest message. Sparing a user a notification from `@Parabol` is nice but the value add is quite small IMO

packages/server/graphql/mutations/helpers/notifySlack.ts
nickoferrall: True, it looks like most of the conditional logic is in the end meeting notification. I've created a couple of functions to make it a bit more readable

packages/server/graphql/mutations/helpers/notifySlack.ts
mattkrick: feels a little weird that the link label is an invalid url. could we instead write something like: [Take me there](https://www.google.com)?

packages/server/graphql/mutations/helpers/notifySlack.ts
mattkrick: well then i made those classes poorly.
For each meeting class (e.g. MeetingRetrospective) could you add the meeting type as a string const?
for example

```ts
export default class MeetingRetrospective extends Meeting {
  meetingType!: 'retrospective'
}
```

after you've done that for all 3 you'll be able to discriminate on the value of `meetingType`

packages/server/graphql/mutations/helpers/notifySlack.ts
nickoferrall: For sure! I thought this was the approach your were suggesting [here](https://parabol.slack.com/archives/D017D764FDY/p1621887268010200).

The "Join meeting" button already takes us to the link so perhaps we just drop the "Link" section?

notification

packages/server/graphql/mutations/helpers/notifySlack.ts
nickoferrall: Cool thanks - I've pushed that update and it's working nicely

packages/server/graphql/mutations/helpers/notifySlack.ts
enriquesanchez: >The "Join meeting" button already takes us to the link so perhaps we just drop the "Link" section?

@nickoferrall I added the link section on purpose because I think having the link there could be handy for cases when you need to share the link with someone, for example on a text message, or an email. You can't copy the link from the "Joing meeting" button, unless you join the meeting first and then copy the link from within Parabol, which I thought would be a bit annoying if joining the meeting is not what you want to do at the moment.

mattkrick:
LGTM! so stoked to see these new notifications in our channel!

Issue #4069, opened 2020-08-12
Add additional retrospective templates - DAKI, lean coffee, starfish, what went well 


avivapinchas:
## Issue - Enhancement

We currently offer 5 retrospective templates. Our competitors sometimes offer more templates. Looking at SEO opportunities, users are directly searching for some of these templates. The opportunities look to me like: 

1. DAKI - drop, add, keep, improve
2. Lean Coffee
3. Starfish
4. What went well

### Acceptance Criteria (optional)
Users can:
   - Utilize any of these templates in a retrospective 

- **Estimated effort:** X points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

gcrickman:
[A user](https://app.hubspot.com/contacts/3888472/contact/11132301) 🔒  mentions:

> We would like the option for a Lean coffee type of retrospective, we do this back in the office and they tend to be very productive.

gcrickman:
[A user](https://app.hubspot.com/contacts/3888472/contact/17236501/) 🔒  writes:

> Is there an option for 3 or more rounds of reflections?

jordanh:
Removing stale from this, as new retro content is never really stale

jordanh:
Linking issues that would be good to do around the same time: #4704, #4784, and #4703

gcrickman:
A [user](https://app.hubspot.com/contacts/3888472/contact/19849851) 🔒  says:

> There's a lot of different methods, you could offer more templates, different techniques so those that aren't experienced have more possibilities.

avivapinchas:
Copy draft for templates ( cc @garethaledavies for a second pair of 👀 )

**Drop Add Keep Improve (DAKI)**
- Drop: What should we no longer do? 
- Add: What should we try or adopt? 
- Keep: What should we continue doing or retain in our practice? 
- Improve: What should we evolve or revise? 

**Lean Coffee ☕  **
@jordanh 🤔 Is it worth offering this format within retros? Lean Coffee is kinda it's own _thing_, but most other retro tools include it in some way. With the linear nature of our app, I'm not sure it _works_, but here are three ways we could approach it: 

Option 1: 
- To Discuss: What should we talk about? 

Option 2: 
- To Discuss: What needs to be discussed but hasn't been talked about? 
- Discussing: What existing conversations should we pick up?

Option 3: 
- To Do: What's up next on your list? 
[...]

PR #4999, opened 2021-05-27
Add retro templates

nickoferrall:
PR to resolve issue #4069 

After deleting the templates and prompts in the down migration, the templates would still show up in the retro template modal but without any prompts.

template-modal

I think this is because we grab the prompts with the dataLoader but we get the templates with `db.read` [here](https://github.com/ParabolInc/parabol/blob/a47534ce10f9b18b2568ab2ee9ad4b41354196fa/packages/server/graphql/types/RetrospectiveMeetingSettings.ts#L95) which initially checks Redis. The templates and prompts were successfully deleted but the templates lived on in the cache.

I'm now clearing the Redis cache in the migrations. After running either migration and then restarting the server, zombie templates no longer show up. Not sure if it's strictly necessary as the cache gets cleared when running `yarn dev` anyway but it saves someone else the surprise of running a down migration and finding empty templates. 

Alternatively / additionally, we could check whether a template contains any prompts but that means requesting all prompts for each template.

Requesting an initial review from @tiffanyhan

### DIFF
--- a/packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
+++ b/packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
@@ -1,6 +1,8 @@
 import {R} from 'rethinkdb-ts'
+const Redis = require('ioredis')
 
 const createdAt = new Date()
+const redis = new Redis(process.env.REDIS_URL)
 
 const nameToId = (name: string, isTemplate: boolean) => {
   const cleanedName = name
@@ -160,16 +162,18 @@ export const up = async function(r: R) {
       r
         .table('ReflectPrompt')
         .insert(reflectPrompts)
-        .run()
+        .run(),
+      // we cache the templates but get the prompts from the db
+      // clear the cache to prevent empty templates
+      redis.flushall()
     ])
+    redis.disconnect()
   } catch (e) {
     console.log(e)
   }
 }
 
 export const down = async function(r: R) {
-  const Redis = require('ioredis')
-  const redis = new Redis(process.env.REDIS_URL)
   const templateIds = templates.map(({id}) => id)
   const promptIds = reflectPrompts.map(({id}) => id)
   try {
--- a/packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
+++ b/packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
@@ -50,7 +50,7 @@ const makePrompt = (promptInfo: PromptInfo) => {
     sortOrder,
     teamId: 'aGhostTeam',
     templateId,
-    title: `New prompt #${sortOrder + 1}`,
+    title: question,
     updatedAt: createdAt
   }
 }
--- a/packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
+++ b/packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
@@ -1,8 +1,8 @@
 import {R} from 'rethinkdb-ts'
-const Redis = require('ioredis')
+import getRedis from '../../utils/getRedis'
 
 const createdAt = new Date()
-const redis = new Redis(process.env.REDIS_URL)
+const redis = getRedis()
 
 const nameToId = (name: string, isTemplate: boolean) => {
   const cleanedName = name
@@ -164,8 +164,8 @@ export const up = async function(r: R) {
         .insert(reflectPrompts)
         .run(),
       // we cache the templates but get the prompts from the db
-      // clear the cache to prevent empty templates
-      redis.flushall()
+      // clear the cache to grab templates from the updated db
+      redis.del('publicTemplates:retrospective')
     ])
     redis.disconnect()
   } catch (e) {
@@ -189,8 +189,8 @@ export const down = async function(r: R) {
         .delete()
         .run(),
       // we cache the templates but get the prompts from the db
-      // clear the cache to prevent empty templates
-      redis.flushall()
+      // clear the cache to grab templates from the updated db
+      redis.del('publicTemplates:retrospective')
     ])
     redis.disconnect()
   } catch (e) {
--- a/packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
+++ b/packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
@@ -1,0 +1,195 @@
+import {R} from 'rethinkdb-ts'
+
+const createdAt = new Date()
+
+const nameToId = (name: string, isTemplate: boolean) => {
+  const cleanedName = name
+    .replace(/[^0-9a-z-A-Z ]/g, '') // remove emojis and apostrophes
+    .split(' ')
+    .map(
+      (name, idx) =>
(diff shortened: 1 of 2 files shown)


packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
```diff
@@ -0,0 +1,199 @@
...
+    createdAt,
+    description,
+    groupColor,
+    id: nameToId(question, false),
+    isActive: true,
+    question,
+    sortOrder,
+    teamId: 'aGhostTeam',
+    templateId,
+    title: question,
```
nickoferrall, line 53: It doesn't look like we're making use of `title` in prompts - I can't see it being requested anywhere. I've followed [this format](https://github.com/ParabolInc/parabol/blob/a6f7d3153947b4f5cd70ef48912bbdf00b60bb94/packages/server/database/migrations/20180904093953-addRetroTemplates.js#L54) of using the question as the title based on the [type description](https://github.com/ParabolInc/parabol/blob/03facdd6bf959cc0e8319adc61d078ccbcdb1f6b/packages/server/graphql/types/ReflectPrompt.ts#L50)

packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
```diff
@@ -0,0 +1,199 @@
...
+        .table('MeetingTemplate')
+        .insert(templates)
+        .run(),
+      r
+        .table('ReflectPrompt')
+        .insert(reflectPrompts)
+        .run(),
+      // we cache the templates but get the prompts from the db
+      // clear the cache to prevent empty templates
+      redis.flushall()
```
tiffanyhan, line 168: I see the reasoning behind doing this in the down migration, but why do we need to flush on the up migration?

packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
```diff
@@ -0,0 +1,199 @@
...
+    sortOrder: 0
+  },
+  {
+    question: `What didn't go well 😞`,
+    description: `What didn't go as planned or desired this sprint?`,
+    groupColor: '#E55C5C',
+    templateId: 'whatWentWellTemplate',
+    sortOrder: 1
+  }
+] as const
```
tiffanyhan, line 150: oh i didn't know about this trick, cool!

packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
```diff
@@ -0,0 +1,199 @@
...
+        .delete()
+        .run(),
+      r
+        .table('ReflectPrompt')
+        .getAll(r.args(promptIds))
+        .delete()
+        .run(),
+      // we cache the templates but get the prompts from the db
+      // clear the cache to prevent empty templates
+      redis.flushall()
```
tiffanyhan, line 193: Nice catch! If we ever did do a down migration, i don't think we'd want to flush the *entire* cache, since we store other tables (e.g. User table) in there.

Can you make it so that we're only deleting the redis keys corresponding to the templates created here? An example is in [`endSprintPoker`](https://github.com/ParabolInc/parabol/blob/master/packages/server/graphql/mutations/endSprintPoker.ts#L58-L61). It looks like we just [prefix keys with table name](https://github.com/ParabolInc/parabol/blob/master/packages/server/dataloader/LocalCache.ts#L126).

tiffanyhan (changes requested):
tested and looks good! only this is i think we should only delete specific keys on the down migration.

packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
nickoferrall: Good point! I've pushed an update removing the public templates key

packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
nickoferrall: If we don't do it in the up migration, `publicTemplates` will be read from the cache rather than the updated db, so it won't have the new retro templates. I could add the new template to Redis but I'd prefer to clear the key to make sure there isn't a mismatch between the db and Redis. I've updated the comment to make it clearer

tiffanyhan: ahhhh i see, i assumed we were cacheing individual templates but from your code link i see we're cacheing the entire batch of retro templates. that makes sense!

tiffanyhan (approved):
this was a bit tricky to test bc we have the 3 stage cache which checks the server before checking redis. i tested and verified the changes by disabling `redis.flushall()` in `dev.js`, making sure the server restarts with each down and up migration respectively, and looking at the keys thru redis cli.

nice job!

packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
mattkrick: true, i think it's an artifact that we never refactored away, we should probably do that...

packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
```diff
@@ -0,0 +1,199 @@
+import {R} from 'rethinkdb-ts'
+import getRedis from '../../utils/getRedis'
+
+const createdAt = new Date()
+const redis = getRedis()
```
mattkrick, line 5: `getRedis` is going to create a connection to redis whenever this file gets imported. side effects are gross! what if up/down never get called? 
to fix that, you can put this inside the up/down function so it only creates that connection

OR, even better:

just get rid of redis calls all together. the server will flush redis when it calls `postdeploy`. you can do that in local dev, too. just call `yarn postdeploy` and it'll flush redis for you

packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
```diff
@@ -0,0 +1,199 @@
...
+        .run(),
+      r
+        .table('ReflectPrompt')
+        .insert(reflectPrompts)
+        .run(),
+      // we cache the templates but get the prompts from the db
+      // clear the cache to grab templates from the updated db
+      redis.del('publicTemplates:retrospective')
+    ])
+    redis.disconnect()
```
mattkrick, line 170: let's get rid of this (see above comment).
if we _did_ do this, we'd want it in a `finally` because Promise.all will throw on the first error & if that first call to the DB failed then this wouldn't get called

packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
nickoferrall: Makes sense! I figured it might not be necessary. I've removed the redis calls and pushed the update!

packages/server/database/migrations/20210527170505-addAdditionalRetroTemplates.ts
nickoferrall: Cool, I've created an [issue](https://github.com/ParabolInc/parabol/issues/5009)

PR #5020, opened 2021-06-04
Handle null overLimitCopy in RethinkDB vs. PG

tianrunhe:
We have the following "inequality" in the comparison result:
```
“xxxxxxxx”: {
    “error”: [
      “overLimitCopy”
    ],
    “rethinkRow”: {
      “overLimitCopy”: null
    },
    “pgRow”: {
      “overLimitCopy”: “”
    }
  },
```
Field `overLimitCopy` in RethinkDB is used by sales to pop messages like "hey your team are over free tier limit, please contact XXX". When the team is indeed upgraded, we set its value to `null` in RethinkDB. In other words, `undefined` for this field represents "we never bothered this user at all", non-empty string represents a popup message, `null` represents "user is upgraded hence the pop up message is cleared".

In RethinkDB, we use `undefined` to represent non-existence and explicit `null` to represent emptiness. In PostgreSQL, `null` is equivalent as undefined. Hence I propose:

* `undefined` in RethinkDB === `null` in PG
* `null` in RethinkDB === `''` in PG

This PR:
1. Change the update logic to write `''` in PG when the value is explicit `null` in RethinkDB
2. Backfill values in PG according to the logic above
3. Update the logic for equality comparison to handle explicit `null` in RethinkDB

### DIFF
--- a/packages/server/postgres/utils/checkEqBase.ts
+++ b/packages/server/postgres/utils/checkEqBase.ts
@@ -32,7 +32,8 @@ function getPairNeFields(
   rethinkRow: RethinkDoc,
   pgRow: PGDoc,
   alwaysDefinedFields: string[],
-  maybeUndefinedFieldsDefaultValues: {[key: string]: any}
+  maybeUndefinedFieldsDefaultValues: {[key: string]: any},
+  maybeNullFieldsDefaultValues?: {[key: string]: any}
 ): string[] {
   const neFields = [] as string[]
 
@@ -45,7 +46,15 @@ function getPairNeFields(
   for (const [f, defaultValue] of Object.entries(maybeUndefinedFieldsDefaultValues)) {
     const [rethinkValue, pgValue] = [rethinkRow[f], pgRow[f]]
     if (rethinkValue !== undefined) {
-      if (!areEqual(rethinkValue, pgValue)) {
+      if (
+        rethinkValue === null &&
+        maybeNullFieldsDefaultValues &&
+        maybeNullFieldsDefaultValues[f] !== undefined
+      ) {
+        if (!areEqual(maybeNullFieldsDefaultValues[f], pgValue)) {
+          neFields.push(f)
+        }
+      } else if (!areEqual(rethinkValue, pgValue)) {
         neFields.push(f)
       }
     } else {
@@ -92,6 +101,7 @@ export async function checkTableEq(
   pgQuery: (ids: string[]) => Promise<PGDoc[] | null>,
   alwaysDefinedFields: string[],
   maybeUndefinedFieldsDefaultValues: {[key: string]: any},
+  maybeNullFieldsDefaultValues: {[key: string]: any},
   maxErrors = 10
 ): Promise<IError> {
   const errors: IError = {
@@ -139,7 +149,8 @@ export async function checkTableEq(
         rethinkRow,
         pgRow,
         alwaysDefinedFields,
-        maybeUndefinedFieldsDefaultValues
+        maybeUndefinedFieldsDefaultValues,
+        maybeNullFieldsDefaultValues
       )
 
       if (neFields.length) {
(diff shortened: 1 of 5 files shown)


tiffanyhan:
> undefined in RethinkDB === null in PG
null in RethinkDB === '' in PG

I think this totally makes sense for string values. Do you think this edge case would ever come up with non char fields, such as int for example? i feel like it's theoretically possible, but don't know if it will be a thing in practical usage in our code 🤔

packages/server/postgres/utils/checkEqBase.ts
```diff
@@ -45,7 +46,15 @@ function getPairNeFields(
...
+        maybeNullFieldsDefaultValues &&
+        maybeNullFieldsDefaultValues[f] !== undefined
+      ) {
+        if (!areEqual(maybeNullFieldsDefaultValues[f], pgValue)) {
+          neFields.push(f)
+        }
+      } else if (!areEqual(rethinkValue, pgValue)) {
         neFields.push(f)
       }
     } else {
```
tiffanyhan, line 60: nitpick: it feels like this conditional statement is getting a bit unwieldy and hard to parse. what if we did something like:

```
if (rethinkValue === undefined) {
  # logic for when rethinkValue is undefined
} else if (rethinkValue === null && maybeNullFieldDefaultValue) {
  # logic for when rethinkValue is null and maybeNullFieldDefaultValue exists
} else # rethinkValue is defined and non-null, or null but !maybeNullFieldDefaultValue {
  # logic for when rethinkValue is defined and non-null, or null but !maybeNullFieldDefaultValue
}
```

tianrunhe:
> undefined in RethinkDB === null in PG  
> null in RethinkDB === '' in PG
> 
> I think this totally makes sense for string values. Do you think this edge case would ever come up with non char fields, such as int for example? i feel like it's theoretically possible, but don't know if it will be a thing in practical usage in our code 🤔

We can just add expected value in:  
```typescript  
const maybeNullFieldsDefaultValues: {[Property in keyof Partial<User>]: any} = {  
 overLimitCopy: '' ,
 intValField: 0
}  
```
Not sure if this is what you were thinking?

packages/server/postgres/utils/checkEqBase.ts
tianrunhe: I rewrite this part from the other side: "under what circumstance should we consider the two equal (hence continue to the next field)?"

tiffanyhan: > > undefined in RethinkDB === null in PG
> > null in RethinkDB === '' in PG
> > I think this totally makes sense for string values. Do you think this edge case would ever come up with non char fields, such as int for example? i feel like it's theoretically possible, but don't know if it will be a thing in practical usage in our code 🤔
> 
> We can just add expected value in:
> 
> ```ts
> const maybeNullFieldsDefaultValues: {[Property in keyof Partial<User>]: any} = {  
>  overLimitCopy: '' ,
>  intValField: 0
> }  
> ```
> 
> Not sure if this is what you were thinking?

yes i was worried about 0 still being a value whereas null is the absence of a value. but you can say the exact same thing for '' anyway so it's a nonstarter. i can't think of any better alternative, so probably the best we can do is just deal with any edge cases as they come up from this handling.

tiffanyhan:
> > undefined in RethinkDB === null in PG
> > null in RethinkDB === '' in PG
> > I think this totally makes sense for string values. Do you think this edge case would ever come up with non char fields, such as int for example? i feel like it's theoretically possible, but don't know if it will be a thing in practical usage in our code 🤔
> 
> We can just add expected value in:
> 
> ```ts
> const maybeNullFieldsDefaultValues: {[Property in keyof Partial<User>]: any} = {  
>  overLimitCopy: '' ,
>  intValField: 0
> }  
> ```
> 
> Not sure if this is what you were thinking?

yes i was worried about 0 still being a value whereas null is the absence of a value. but you can say the exact same thing for '' anyway so it's a nonstarter. i can't think of any better alternative, so probably the best we can do is just deal with any edge cases as they come up from this handling.

tiffanyhan (approved):
we were definitely missing logic for rethink null <=> pg alt null value. thanks for adding this!

Issue #5005, opened 2021-06-01
Change email address mutation


mattkrick:
AC
- [ ] a mutation exists on the intranet schema that lets you enter oldEmail & newEmail
- [ ] User & TeamMember objects are updated with newEmail

Estimate: 4 hours

ackernaut:
@BartoszJarocki that’s it. I assume a UI design issue would be started later.

PR #5031, opened 2021-06-09
Added update email mutation to intranet schema

BartoszJarocki:
Related issue: https://github.com/ParabolInc/parabol/issues/5005 

One note, `User` is also updated in PG, I'm not sure about `TeamMember` as I don't see what's the relation between `User` and `Team` in PG so I haven't done anything related to that.

### DIFF
--- a/packages/server/graphql/intranetSchema/README.md
+++ b/packages/server/graphql/intranetSchema/README.md
@@ -1,11 +1,12 @@
 ## Intranet/Admin interface 
 
 This part of Parabol's API is available only for admins. To become an admin user has to have `su` role set.
-
 `su` role can be set via RethinkDB’s data explorer, on local environment it’s available at: http://localhost:8080/#dataexplorer
+
+To set the `su` role run the following query
 ```
 r.db('actionDevelopment').table('User').filter({email: 'set-your-email-here'}).update({rol:'su'})
 ```
 
-Once the role is set, admin interface can be accessed locally at http://localhost:3000/admin/graphql and allows you to execute available queries and mutations. 
+Once the role is set, the admin interface can be accessed locally at http://localhost:3000/admin/graphql and allows you to execute available queries and mutations. 
 More info: https://github.com/graphql/graphiql 
--- a/packages/server/graphql/intranetSchema/README.md
+++ b/packages/server/graphql/intranetSchema/README.md
@@ -13,5 +13,5 @@ To set the `su` role in PG open pgAdmin (credentials are specified in .env as PG
 UPDATE "User" SET "rol" = 'su' WHERE "id" = 'set-your-user-id-here'; 
 ```
 
-Once the role is set, the admin interface can be accessed locally at http://localhost:3000/admin/graphql and allows you to execute available queries and mutations. 
-More info: https://github.com/graphql/graphiql 
+Once the role is set, the admin interface can be accessed locally (logout and login again is required) at http://localhost:3000/admin/graphql and allows you to execute available queries and mutations. 
+More info: https://github.com/graphql/graphiql
--- a/packages/server/graphql/intranetSchema/README.md
+++ b/packages/server/graphql/intranetSchema/README.md
@@ -1,0 +1,11 @@
+## Intranet/Admin interface 
+
+This part of Parabol's API is available only for admins. To become an admin user has to have `su` role set.
+
+`su` role can be set via RethinkDB’s data explorer, on local environment it’s available at: http://localhost:8080/#dataexplorer
+```
+r.db('actionDevelopment').table('User').filter({email: 'set-your-email-here'}).update({rol:'su'})
+```
+
+Once the role is set, admin interface can be accessed locally at http://localhost:3000/admin/graphql and allows you to execute available queries and mutations. 
+More info: https://github.com/graphql/graphiql 
--- a/packages/server/graphql/intranetSchema/README.md
+++ b/packages/server/graphql/intranetSchema/README.md
@@ -1,11 +1,16 @@
 ## Intranet/Admin interface 
 
 This part of Parabol's API is available only for admins. To become an admin user has to have `su` role set.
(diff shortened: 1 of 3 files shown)


packages/server/graphql/intranetSchema/mutations/updateEmail.ts
```diff
@@ -0,0 +1,74 @@
...
+  },
+  resolve: async (_source, {oldEmail, newEmail}, {authToken}: InternalContext) => {
+    const r = await getRethink()
+
+    // AUTH
+    requireSU(authToken)
+
+    // VALIDATION
+    if (!newEmail) {
+      throw new Error('New email can not be empty')
```
mattkrick, line 28: `type: GraphQLNonNull(GraphQLString),` already validates that the email is nonnull so no need to repeat here

packages/server/graphql/intranetSchema/mutations/updateEmail.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    if (oldEmail === newEmail) {
+      throw new Error('New email is the same as the old one')
+    }
+
+    const email = newEmail.toLowerCase().trim()
```
mattkrick, line 35: let's normalize both emails before the validation begins

packages/server/graphql/intranetSchema/mutations/updateEmail.ts
```diff
@@ -0,0 +1,74 @@
...
+      .getAll(oldEmail, {index: 'email'})
+      .nth(0)
+      .default(null)
+      .run()
+    if (!user) {
+      throw new Error(`User with ${oldEmail} not found`)
+    }
+
+    // RESOLUTION
+    const userId = user.id
```
mattkrick, line 51: this is gonna be the absolutely most nittiest of requests & if you don't like it i'm totally OK with that :see_no_evil: 
when possible, could you destructure like `const {id: userId} = user`? 
Even though it's a little bit longer, it means fewer line diffs so in the future if we need to get more props out of that object e.g. `const {id: userId, createdAt, updatedAt} = user` is just gonna look like a partial line diff

packages/server/graphql/intranetSchema/mutations/updateEmail.ts
```diff
@@ -0,0 +1,74 @@
...
+import {requireSU} from '../../../utils/authorization'
+import {InternalContext} from '../../graphql'
+import updateUser from '../../../postgres/queries/updateUser'
+
+const updateEmail = {
+  type: GraphQLNonNull(GraphQLBoolean),
+  description: `Updates the user email`,
+  args: {
+    oldEmail: {
+      type: GraphQLNonNull(GraphQLString),
```
mattkrick, line 12: if you change this to a GraphQLID then graphql will validate that it doesn't contain any spaces

mattkrick (commented):
looking almost perfect! just a couple nits

packages/server/graphql/intranetSchema/mutations/updateEmail.ts
BartoszJarocki: I actually think that's a very valid request, fixed 😅

packages/server/graphql/intranetSchema/mutations/updateEmail.ts
BartoszJarocki: @mattkrick I see we have https://github.com/ParabolInc/parabol/blob/8dbff9b4c6d19467c96b323be63f61e2926dc0a5/packages/server/graphql/types/GraphQLEmailType.ts, maybe that's the best of both worlds? I see it's not trimming the value but I can add that and then it'd be consistent across the codebase. Also, all the checks would be explicit which I think is a bit more readable. What do you think?

packages/server/graphql/intranetSchema/README.md
```diff
@@ -0,0 +1,11 @@
+## Intranet/Admin interface 
+
+This part of Parabol's API is available only for admins. To become an admin user has to have `su` role set.
+
+`su` role can be set via RethinkDB’s data explorer, on local environment it’s available at: http://localhost:8080/#dataexplorer
+```
+r.db('actionDevelopment').table('User').filter({email: 'set-your-email-here'}).update({rol:'su'})
```
mattkrick, line 7: could you add the SQL here, too? want to keep the User data across both DBs aligned

packages/server/graphql/intranetSchema/mutations/updateEmail.ts
mattkrick: oh yeaaah forgot about that type! no strong preference either way, whatever you think looks best

packages/server/graphql/intranetSchema/mutations/updateEmail.ts
BartoszJarocki: Fixed 👍

packages/server/graphql/intranetSchema/mutations/updateEmail.ts
BartoszJarocki: Fixed by using `GraphQLEmailType` which doesn't allow spaces and makes the value lowercase.

packages/server/graphql/intranetSchema/mutations/updateEmail.ts
BartoszJarocki: Changed to `GraphQLEmailType`. I think it should be consistent across the codebase.

packages/server/graphql/intranetSchema/README.md
BartoszJarocki: 🙈  I forgot about the PG, and sure, added!

BartoszJarocki:
@mattkrick I'd appreciate if you could take a look again, thank you!

tianrunhe (approved):
LGTM!

mattkrick:
nice one! this will help the service folks a bunch!

Issue #4883, opened 2021-04-13
Change password link tells you if you have a non-email account


jordanh:
## Issue - Enhancement

Several users have forgot how they signed up for Parabol and become confused when a change password email never arrives if they've signed up via Oauth

### Acceptance Criteria (optional)
Users can:
   - Error message displayed when "Send email" is clicked if user has non-email account

- **Estimated effort:** 4 hours ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

tianrunhe:
Duplicate as #4299? @jordanh I can close mine in favor of yours or we can "resolve the merge conflict"

mattkrick:
OWASP tells us that we shouldn't give away ANY information about our userbase.
We should do this by sending a "complete your signup via email" flow, so if someone signs up via email, we say "check your inbox" they click the link in the email & then they get an account.
This is safer because if i try to signup with an email that's not mine, the app doesn't tell me if it exists or not. it only says "check your email"

We opt out of this safe flow for the sake of UX.

We also do something else for the sake of UX-- when a user tries to login with username/password & they already signed up with google oauth, well tell them "psst, use google oauth to login". 

So, we're already leaking 1) that they have an account and 2) their AuthIdentity (gmail, password, or sso)

So, by leaking this in the forgot password page, we're not exposing any extra info, but as we become more secure, we'll probably have to undo these nice UX flows.

We also leak SAML info today because if you try to login with a SAML domain, then it'll open the SAML login popup

All this to say, what we're doing isn't secure, but by implementing this PR we're not being MORE insecure.


AC Flow:
- Look for google auth identity. if found, tell them to login with google
- Look for local identity. if not found, then see if the domain matches a SAML domain. if so, tell them to login with SSO
- else if local identity found, tell them an email was sent if one exists in the system

Issue #4299, opened 2020-10-30
Inform user to signin with Google if they try to reset password with their OAuth enabled GMail email address


tianrunhe:
## Issue - Enhancement

Scenario/tension:

*   User enabled OAuth with their GMail account (i.e., Signed in with Google before)
*   User somehow forgot they signed in with Google before and thought they created an account with their GMail and set up a password
*   User tried to reset their password by clicking "Forgot your password" link
*   User was notified they will get an password recovery email but in fact they will never get one, so they are confused

![](https://user-images.githubusercontent.com/1879975/97739053-81cf5100-1a9c-11eb-849d-e6009d77475d.gif)

### Acceptance Criteria (optional)

*   User should be notified that they had signed in with Google before so they should do that, instead of resetting password

**Estimated effort:** X points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

jordanh:
Iceboxing

PR #5033, opened 2021-06-09
New forgot password flow

nickoferrall:
PR to resolve issues #4883 & #4299 

In the auth flow, we're currently prioritising UX over security e.g. auto opening the SAML pop-up if you login with a SAML domain. Building on the AC Matt left in #4883, I think we should only tell the user that we've sent them an email when we actually have. 

Currently, if the user enters an email address in the forgot password flow that does or doesn't belong to one of our users, we tell them an email was sent. This means that we conceal whether the email address belongs to a user which is good from a security perspective.

However, as we're prioritising UX and as we're already leaking email info by opening the SAML pop-up, for example, I think we should give the user feedback if the email address doesn't exist. Other companies do this, e.g. [Stripe](https://dashboard.stripe.com/reset), which makes me feel less nervous.

### AC

- [ ] Look for google auth identity. if found, tell them to login with google
- [ ] Look for local identity. if not found, then see if the domain matches a SAML domain. if so, tell them to login with SSO
- [ ] If a user doesn't exist with the submitted email address, tell the user to try again
- [ ] If the email wasn't sent for a different reason (e.g. exceeded daily limit), tell them something went wrong 
- [ ] Else if local identity found and the email was successfully sent, let them know


nickoferrall:
@tianrunhe mind giving this one a review when you get the chance?

packages/server/graphql/mutations/emailPasswordReset.ts
```diff
@@ -53,12 +55,29 @@ const emailPasswordReset = {
           .count()
           .ge(Threshold.MAX_DAILY_PASSWORD_RESETS) as unknown) as boolean
       }).run()
-      if (failOnAccount || failOnTime || !user) return true
+      if (failOnAccount || failOnTime) {
+        return {error: {message: AuthenticationError.EXCEEDED_RESET_THRESHOLD}}
+      }
+      if (!user) return {error: {message: AuthenticationError.USER_NOT_FOUND}}
```
mattkrick, line 61: let's do the SAML validation first:
- it's possible that the person hasn't created an account yet, but their domain is SAML, in which case we want them to see the SAML instructions
- it's possible the person used to log in with password or google but then their org upgraded to SAML, in which case they should go through the SAML flow

Additional tests:
- [ ] domain `foo.co` is a registered SAML domain
- [ ] when `foo@foo.co` who doesn't exist in our system clicks forget password, they get the SAML error
- [ ] when `bar@foo.co` who DOES exist in our system, they get the SAML error
- [ ] when `baz@foo.co` who has a google identity clicks, they get a SAML error

mattkrick (changes requested):
1 quick change & this is good to go!
feel free to merge after the change

Issue #5028, opened 2021-06-08
Jira stories, can’t add points as a comment


ackernaut:
I looked and can’t see if this is on our radar:

### Reproduction steps
- Estimate a Jira issue in Sprint Poker
- Try selecting _As Comment_ or _Do Not Update_
- Nothing happens

![Screen Shot 2021-06-08 at 10 51 37](https://user-images.githubusercontent.com/307286/121217519-8b494480-c847-11eb-9681-eea8d5e5957e.png)

### Acceptance criteria
- The options _As Comment_ and _Do Not Update_ work as expected

### Estimated effort
- 8 hours

BartoszJarocki:
Selecting these two fields doesn't work because the `UpdateJiraDimensionFieldMutation` fails with `Invalid field name error`.
<img width="1010" alt="Screenshot 2021-06-09 at 17 20 33" src="https://user-images.githubusercontent.com/1017620/121391435-a81c6f80-c94e-11eb-8339-e15571fc3e04.png">

I see these two fields are treated differently and use predefined values. `As Comment` uses `__comment` value as field name, where in Jira this field has `Comment` name and as a result it can't be found. @mattkrick do you recall why the field name for `Comment` is set to `__comment`? 

Also, what's the reason for having 'Do not update' menu option and what should happen when someone actually selects it and clicks `Update`?

mattkrick:
@BartoszJarocki if you dig around the repo for `JIRA_FIELD_COMMENT` then that should help clear some things up.
basically, if they choose to push the new score as a comment, then we don't update a field, we just post a comment to the issue...so if we're still calling jira to try & update a field with that name then the logic broke somewhere.

the DO NOT UPDATE means they get a story estimate within paraobl but we don't push that change to jira

BartoszJarocki:
PR: https://github.com/ParabolInc/parabol/pull/5041

PR #5041, opened 2021-06-10
Jira stories, can’t add points as a comment

BartoszJarocki:
Related issue: https://github.com/ParabolInc/parabol/issues/5028 

The reason it doesn't work is that `UpdateJiraDimensionField` mutation fails when `As Comment` or `Do Not Update` is selected. See the screenshot below. 
Screenshot 2021-06-09 at 17 20 33

@mattkrick I went through the commit history and I can't find any changes that could break that functionality. Before this fix, `UpdateJiraDimensionField` just iterates through Jira fields and throws an error if the field is not found. As `As Comment` or `Do Not Update` aren't normal Jira fields they can't be found in the list of standard Jira fields. This PR checks if any of these custom fields is selected and then just skips the Jira field array search. Otherwise, it just works as before.

### DIFF
--- a/packages/server/graphql/mutations/updateJiraDimensionField.ts
+++ b/packages/server/graphql/mutations/updateJiraDimensionField.ts
@@ -2,6 +2,7 @@ import {GraphQLID, GraphQLNonNull, GraphQLString} from 'graphql'
 import {SprintPokerDefaults, SubscriptionChannel} from 'parabol-client/types/constEnums'
 import getRethink from '../../database/rethinkDriver'
 import JiraDimensionField from '../../database/types/JiraDimensionField'
+import {FreshAtlassianAuth} from '../../dataloader/customLoaderMakers'
 import getTemplateRefById from '../../postgres/queries/getTemplateRefById'
 import updateTeamByTeamId from '../../postgres/queries/updateTeamByTeamId'
 import AtlassianServerManager from '../../utils/AtlassianServerManager'
@@ -9,6 +10,23 @@ import {getUserId, isTeamMember} from '../../utils/authorization'
 import publish from '../../utils/publish'
 import {GQLContext} from '../graphql'
 import UpdateJiraDimensionFieldPayload from '../types/UpdateJiraDimensionFieldPayload'
+
+const getJiraField = async (fieldName: string, cloudId: string, auth: FreshAtlassianAuth) => {
+  // we have 2 special treatment fields, JIRA_FIELD_COMMENT and JIRA_FIELD_NULL which are handled
+  // differently and can't be found on Jira fields list
+  const customFields = [SprintPokerDefaults.JIRA_FIELD_COMMENT, SprintPokerDefaults.JIRA_FIELD_NULL]
+  if (customFields.includes(fieldName as any)) {
+    return {fieldId: fieldName, type: 'string' as const}
+  }
+  // a regular Jira field
+  const {accessToken} = auth
+  const manager = new AtlassianServerManager(accessToken)
+  const fields = await manager.getFields(cloudId)
+  const selectedField = fields.find((field) => field.name === fieldName)
+  if (!selectedField) return null
+  const {id: fieldId, schema} = selectedField
+  return {fieldId, type: schema.type as 'string' | 'number'}
+}
 
 const updateJiraDimensionField = {
   type: GraphQLNonNull(UpdateJiraDimensionFieldPayload),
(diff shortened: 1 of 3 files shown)


BartoszJarocki:
@tianrunhe @nickoferrall as @mattkrick is out, maybe you could take a look and do an initial review? 🙏 
Thanks!

packages/server/graphql/mutations/updateJiraDimensionField.ts
```diff
@@ -77,12 +77,25 @@ const updateJiraDimensionField = {
...
-    const {id: fieldId, schema} = field
+
+    let selectedField
+    // we have 2 special treatment fields, JIRA_FIELD_COMMENT and JIRA_FIELD_NULL which are handled
+    // differently and can't be found on Jira fields list
+    const customFields = [
+      SprintPokerDefaults.JIRA_FIELD_COMMENT,
+      SprintPokerDefaults.JIRA_FIELD_NULL
+    ]
+    if (customFields.includes(fieldName)) selectedField = {id: fieldName, schema: {type: 'string'}}
```
nickoferrall, line 88: How would you feel about combining the custom fields and regular fields into one array? I.e.
```
const allFields = [...fields, ...customFields]
```
That way we could map through it once and avoid the let & if else logic?

nickoferrall (changes requested):
The fix works nicely! Just one suggestion/consideration from me and then I'll give it the green light

packages/server/graphql/mutations/updateJiraDimensionField.ts
BartoszJarocki: @nickoferrall Thank you for checking the PR 🙏 . 
That'd definitely simplify the code. The question is, do we prefer an additional `if` statement and save Jira API call or simpler code with unnecessary Jira API call in case of custom fields usage? 😅

nickoferrall: True! If you'd prefer to save the Jira hit then it LGTM 🎉

nickoferrall (approved):

tianrunhe:
Do we need to do anything if the field is `SprintPokerDefaults.JIRA_FIELD_NULL`? Should we return early in this case?

BartoszJarocki:
@tianrunhe inside `UpdateJiraDimensionFieldMutation` we only save the selected field. The logic of handling the selected field is here: https://github.com/ParabolInc/parabol/blob/55e733946623f8150b49d9cb6c610dbaa9388e99/packages/server/graphql/mutations/pokerSetFinalScore.ts#L101 and as you can see, if the field is `SprintPokerDefaults.JIRA_FIELD_NULL` it'll do nothing.

tianrunhe:
> @tianrunhe inside `UpdateJiraDimensionFieldMutation` we only save the selected field. The logic of handling the selected field is here:
> 
> https://github.com/ParabolInc/parabol/blob/55e733946623f8150b49d9cb6c610dbaa9388e99/packages/server/graphql/mutations/pokerSetFinalScore.ts#L101
> 
> and as you can see, if the field is `SprintPokerDefaults.JIRA_FIELD_NULL` it'll do nothing.

Sorry maybe I need to read the code more: I'm questioning that if we do nothing for this case in `pokerSetFinalScore` then why do we still need to store it in the DB?

BartoszJarocki:
@tianrunhe only because we won't have the information about what field was chosen by the user 😄 

The flow is the following
1. User selects the field via dropdown and `UpdateJiraDimensionFieldMutation` is fired with the selected field, the field is saved in DB
2. User clicks the `Update` button to actually save the estimate in the previously selected field and `PokerSetFinalScoreMutation` is fired (there's no information about the selected field passed here); on the server-side `PokerSetFinalScoreMutation` reads the selected field from the DB, updates the estimate (if anything else than null field is selected). 

Before this fix, in 1. these two custom fields were not saved in the database, and as a result, UI didn't change and it looked like nothing is happening when selecting one of them. Please let me know if that makes more sense now, thank you!

tianrunhe:
> @tianrunhe only because we won't have the information about what field was chosen by the user 😄
> 
> The flow is the following
> 
> 1. User selects the field via dropdown and `UpdateJiraDimensionFieldMutation` is fired with the selected field, the field is saved in DB
> 2. User clicks the `Update` button to actually save the estimate in the previously selected field and `PokerSetFinalScoreMutation` is fired (there's no information about the selected field passed here); on the server-side `PokerSetFinalScoreMutation` reads the selected field from the DB, updates the estimate (if anything else than null field is selected).
> 
> Before this fix, in 1. these two custom fields were not saved in the database, and as a result, UI didn't change and it looked like nothing is happening when selecting one of them. Please let me know if that makes more sense now, thank you!

It makes great sense now! Thanks so much for the explanation!

tianrunhe (approved):

BartoszJarocki:
@tianrunhe no problem 👍 Thank you for taking the time to review it! 🙏

packages/server/graphql/mutations/updateJiraDimensionField.ts
```diff
@@ -77,12 +77,25 @@ const updateJiraDimensionField = {
...
       return {error: {message: 'Not authenticated with Jira'}}
     }
-    const {accessToken} = auth
-    const manager = new AtlassianServerManager(accessToken)
-    const fields = await manager.getFields(cloudId)
-    const field = fields.find((field) => field.name === fieldName)
-    if (!field) return {error: {message: 'Invalid field name'}}
-    const {id: fieldId, schema} = field
+
+    let selectedField
```
mattkrick, line 81: great fix! the only nit is that `selectedField` has an implicit type of `any`, which is pretty common among let statements.
we could cast it to a `Partial<JiraField>`, but partials mean you always have to check nullability.

instead, i'd propose just making a little helper function that keeps the return values strict. that way if anything changes (e.g. the JiraField interface) then we'll get errors for free!

example refactor here: https://github.com/ParabolInc/parabol/pull/5069/files

mattkrick (changes requested):
one little nit on the implicit let statement, ready to merge after that

packages/server/graphql/mutations/updateJiraDimensionField.ts
BartoszJarocki: right, that's a good point! 🙏

BartoszJarocki:
@mattkrick thank you for submitting the fix for implicit let, I've just merged it 👍

Issue #4889, opened 2021-04-14
Adds clarity between comments and task creation


enriquesanchez:
Continuation of #4812 

<details>
<summary>~Old AC~</summary>
### Acceptance criteria

- [See Figma](https://www.figma.com/file/Ia7SoJcudXUM3nvOYXOldR/Adding-comments-and-tasks?node-id=103%3A0) 
- New "action bar":
  - A new "action bar" is displayed above the comment input field. This bar has a `White #FFFFFF` background and is 32px tall and full width. This new bar increases the overall height of `discuss-input-wrapper`.
  - The intended use of the action bar is to group actions and content objects, such as 'Add task', 'Add poll', or 'Schedule event', etc. For now, we will only support the 'Add task' object.
  - 'Add task' should be a blue text button preceded by the `add` MD icon at 18px. It should be centered both vertically and horizontally. I suggest using `flexbox` for alignment. This text button should inherit the same hover and focus styles we already use on other text buttons and links (`Sky 600 #14649E`).
- The comment area will also receive some updates:
  - The 'Add a task' button is removed from `discuss-input-wrapper` (it now lives in the new action bar described above).
  - The comment input field is now followed by a submit icon button. This new button is 32px by 32px, has fully rounded corners and is right aligned. It has a `Slate 200 #F1F0FA` background color when disabled.
  - The icon button uses the MD icon `north` at 18px. The icon is set to `Slate 500 #A7A3C2`.
  - Once text has been entered in the comment field, the new submit button becomes active and will have a background color of `Sky 500 #329AE5` and the MD icon is set to `White #FFFFFF`.
  - On `hover` and `focus` the button's background is set to `Sky 600 #14649E`. The `north` icon stays on `White #FFFFFF`.
  
### Deferred

We will eventually support other content object besides tasks, and we'd want their input buttons to be grouped together inside the new action bar. Once the number of supported objects grows, we might look at moving the action bar to the bottom of  `discuss-input-wrapper`. [See this Figma file for additional explorations](https://www.figma.com/file/D25gVH1t2alUiOdYNs5Owo/Multiple-content-input-types?node-id=0%3A1).
</details>

AC
- [ ] Discussion thread looks like figma: https://www.figma.com/file/D25gVH1t2alUiOdYNs5Owo/Multiple-content-input-types?node-id=56%3A968
- [ ] Send button uses up arrow (not paper airplane)
- [ ] Send button is gray when there is no text in the input
[...]

PR #5049, opened 2021-06-15
Adds clarity between comments and task creation

BartoszJarocki:
Related issue: https://github.com/ParabolInc/parabol/issues/4889 

All the states, sizing have been changed based on design: https://www.figma.com/file/D25gVH1t2alUiOdYNs5Owo/Multiple-content-input-types?node-id=56%3A968

### How does it look like?
CleanShot 2021-06-16 at 16 55 11


BartoszJarocki:
@enriquesanchez I'd appreciate if you could take a look, thank you!

enriquesanchez (changes requested):
This is looking great 🙌

A couple of additions I think we need: 

1. The buttons, when disabled, should have the `disabled` attribute added to them in the HTML. More info here: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled
2. The buttons, when disabled, and hovered, should use `cursor: not-allowed;` for the mouse pointer, so it shows the disabled cursor. More info here: https://developer.mozilla.org/en-US/docs/Web/CSS/cursor

Other than that, I think this is pretty good 🎉

BartoszJarocki:
@enriquesanchez thank you for checking it! 

The `Add task` button already uses that attribute. I've added it to the `Send comment` button as well. I do have one question about it though. By default, that kind of button has set 50% opacity and it looks like that in the disabled state: 
Screenshot 2021-06-17 at 07 16 15
for comparison, this is how it looks per design:
Screenshot 2021-06-17 at 07 33 40

The question is, do we want to keep things consistent in the disabled state, or should we make an exception here and make it look like on the design? It looks nice both ways.

enriquesanchez:
Thanks @BartoszJarocki 🙌

> By default, that kind of button has set 50% opacity and it looks like that in the disabled state:

Oh, is this a default we use across Parabol? If that is the case, then let's go with defaults for now. Technically, disabled buttons don't have any contrast requirements when it comes to accessibility. So while I'd rather it wasn't too transparent, it's not a huge deal right now and I prefer we keep our styles consistent.

BartoszJarocki:
@enriquesanchez I'm not sure in how many places it is used at this point but that kind of style is definitely implemented and used by default. If you're ok with using it that way, I'm gonna push the changes and if needed we can revisit anytime. 🙏

enriquesanchez:
>  If you're ok with using it that way, I'm gonna push the changes and if needed we can revisit anytime

Sounds good! Thank you 🙌

BartoszJarocki:
@enriquesanchez I double-checked and while it was looking ok on my external monitor, it was actually barely visible on my laptop. I overrode the styles and now it looks like on the designs. I'd appreciate if you could do another round and make sure it's all pixel perfect, thank you! 🙏

enriquesanchez:
@BartoszJarocki Looks good to me 👍

packages/client/clientSchema.graphql
```diff
@@ -105,6 +105,7 @@ extend type EstimateStage {
 }
 
 extend type Discussion {
+  isCreatingNewTask: Boolean
```
mattkrick, line 108: I think we can get away with removing this if we use https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/client/utils/isViewerTyping.js to determine if a task is being edited.

the only difference is `isViewerTyping` will also return true if someone is editing a task, not just creating one (which is a feature, not a bug IMO :wink: )

PS feel free to turn that file into a .ts if you like. some older files are still .js and as we touch them we convert them to typescript

mattkrick (changes requested):
great work!
only note is to use the DOM as the source of truth for `isCreatingTask`, which should make a lot of the logic a lot simpler.

I noticed that adding comments & tasks do _not_ work when the meeting is completed, but the UI looks a little buggy. I think I probably wrote the bad logic, but this is a good PR to fix it on!

I propose the following:
- [ ] if the meeting is complete, `allowedThreadables` is an empty array
- [ ] if `allowedThreadables` does not contain `tasks`, then the "add a task" button is hidden (i think this logic is already in place, see poker meetings to confirm)
- [ ] if `allowedThreadables` does not contain "comments" then the comment button is gray & the input field is `readOnly`

packages/client/clientSchema.graphql
BartoszJarocki: @mattkrick that'd be awesome. Any ideas on how can we observe the state of `isViewerTypeing` so we can force React to rerender when it changes? 

Using it that way
```
          <AddTaskButton dataCy={`${dataCy}-add`} onClick={addTask} disabled={isViewerTyping()} />
```
isn't enough unfortunately as when the user is actually adding a new task, the component responsible for showing the `AddTaskButton` isn't rerendering. I googled a bit and I haven't found any obvious answer. I will search more.

BartoszJarocki: @mattkrick I've created a new branch with mutation observer implementation: https://github.com/ParabolInc/parabol/compare/feature/4889/task-creation-design...feature/4889/task-creation-design-mutation-observer?expand=1 
Regarding focus handling, I simply reset the state in the `onBlur` comment input handler. Please let me know what do you think 🙏

mattkrick: if that works let's go with that!

my cheap & easy alternative would be:
```ts
document.addEventListener('focus', function() {
  setCanCreateTask(!isViewerTyping())
}, true);
```

you could put that in a useEffect & get the same observability without having to monitor the whole DOM tree.

BartoszJarocki: @mattkrick ah, thanks for the tip!

I think the cleanest solution is to do the following then
```
  useEffect(() => {
    const focusListener = () => {
      setCanCreateTask(!isViewerTyping())
    }

    document.addEventListener('blur', focusListener, true)
    document.addEventListener('focus', focusListener, true)
    return () => {
      document.removeEventListener('blur', focusListener, true)
      document.removeEventListener('focus', focusListener, true)
    }
  }, [])
```

I'm not sure about the UX because the user will have to click outside of the comment editor to be able to create a task but maybe it's not a big deal.

packages/client/utils/isViewerTyping.ts
```diff
@@ -0,0 +1,6 @@
+const isViewerTyping = () => {
+  const activeElement = document.activeElement as HTMLElement
+  return activeElement.contentEditable === 'true'
+}
```
BartoszJarocki, line 4: I'm not sure about this casting. I haven't found a better way thought so I'd appreciate suggestions if there's a way to avoid casting 🙏

BartoszJarocki:
@mattkrick I've updated the PR, thank you for the review! I couple of questions related to improving the functionality.
I'm not sure how to test the 'completed' meeting use case, any suggestions here?

1. `if the meeting is complete, allowedThreadables is an empty array` - is that already true, or does it need to be implemented?
2. `if allowedThreadables does not contain tasks, then the "add a task" button is hidden (i think this logic is already in place, see poker meetings to confirm)` - already implemented
3. `if allowedThreadables does not contain "comments" then the comment button is gray & the input field is readOnly` - updated the PR with that functionality

packages/client/utils/isViewerTyping.ts
mattkrick: yep, having to cast Element to HTMLElement is the cleanest way i've found

mattkrick:
> if the meeting is complete, allowedThreadables is an empty array - is that already true, or does it need to be implemented?

needs to be implemented. looking at `DiscussionThreadRoot` i think we could simplify the API by removing `isReadOnly` because `isReadOnly = allowedThreadables.length === 0`.

For example, in `EstimatePhaseDiscussionDrawer`

```ts
  const allowedThreadables = [  ...endedAt ? [] : ['comment']]
```

BartoszJarocki:
@mattkrick thank you for the explanation, all clear now! 
I've removed the `isReadOnly` and refactored `allowedThreadables` usage. The completed meeting looks good now. 

Screenshot 2021-06-29 at 15 20 23

It would be great if you could take a look 🙏

mattkrick:
agreed that the behavior when typing a comment is a little weird. "Add a task" looks disabled, but it's not!
let's replace `isViewerTyping` with `isVieweringTypingInTask`
that function should also support android devices

1. let's add `ariaLabel={AriaLabels.TASK_EDITOR}` to the `Editor` in `TaskEditor` (in constEnums, add `AriaLabel { TaskEditor = 'Task Editor'}`
2. also add that label to the android fallback in TaskEditor
3. write a function that targets just the task

```ts
activeElement.ariaLabel === AriaLabels.TASK_EDITOR && (activeElement.contentEditable === 'true' || activeElement.tagName === 'TEXTAREA')
```

BartoszJarocki:
@mattkrick thank you for such detailed suggestions! I've just pushed the changes. I'll double-check on Android and iOS as I haven't done that before 🙈

packages/client/mutations/NavigateMeetingMutation.ts
```diff
@@ -110,7 +110,8 @@ export const navigateMeetingTeamUpdater: SharedUpdater<NavigateMeetingMutation_t
   }
   if (viewerStageId === oldMeeting.facilitatorStageId) {
     const viewerPhaseType = getInProxy(meeting, 'localPhase', 'phaseType')
-    if (!isInterruptingChickenPhase(viewerPhaseType) || !isViewerTyping()) {
+    const isViewerTyping = isViewerTypingInTask() || isViewerTypingInComment()
+    if (!isInterruptingChickenPhase(viewerPhaseType) || !isViewerTyping) {
```
BartoszJarocki, line 114: To not break the existing functionality I've added both checks here.

mattkrick:
looks good!

PR #5050, opened 2021-06-15
Backfill script for the missing teams with same createdAt timestamp

tianrunhe:
Just a quick & dirty backfill script to backfill the ~1,700 teams missed in last round of migrations to PG. I borrowed a lot of code from [1615148741896_backfillTeamTable.ts](https://github.com/ParabolInc/parabol/blob/master/packages/server/postgres/migrations/postDeploy/1615148741896_backfillTeamTable.ts).


tianrunhe:
@tiffanyhan Please take a look. Thanks!

tiffanyhan (changes requested):
@tianrunhe can you add in the `doBackfillAccountingForRaceConditions` logic too? bc there are ~1700 rows to backfill, i'd feel better if that was included in case someone updates one of those teams while the migration is running. probably unlikely, but better safe than sorry imo.

tiffanyhan (approved):

Issue #4691, opened 2021-02-19
Remove refresh for new version snackbar


mattkrick:
Word on the street is people don't think our "New version available" snackbar is as cute as I think it is :shrug: 

Instead, when the client connects to the server & the server is at a different version than the client, we should assume that the client will just work.
If it breaks (e.g. the client calls a server function that no longer exists) then that will open the ErrorComponent. That error component should tell them that they hit a bug because they are using an old version & they need to refresh.

Acceptance Criteria
- [ ] don't pop a snackbar asking them to refresh. Instead, just pop one that goes away after 5 seconds telling them the new version of parabol has been downloaded
- [ ] whenever an error occurs, in the error component, compare the app version with the server version. if there's a mismatch, change the message to "You hit a bug because you're using an old version of parabol. hit refresh & it should work"

Estimate: 8 hours

PR #5051, opened 2021-06-16
Update new version snackbar

nickoferrall:
PR to resolve issue #4691. I'm having difficulty comparing the app version and the server version in the error component but I've implemented a possible solution, written up my understanding of how things are working and left a couple of questions.

I can see that the `__APP_VERSION__` is set in the webpack config [here](https://github.com/ParabolInc/parabol/blob/0ea032bfa3d57a52212278d47cee0bbe3d1972e2/scripts/webpack/prod.client.config.js#L127) and that it looks at the package json file for the latest version.

Using service workers, we create a cache in [sw.ts](https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/client/serviceWorker/sw.ts) with `__APP_VERSION__` as the suffix.

In [useTrebuchetEvents](https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/client/hooks/useTrebuchetEvents.ts#L66), the `onData` function is triggered when the page loads and when the user visits a page that isn't stored in the cache. When the page loads, [handleOpen](https://github.com/ParabolInc/parabol/blob/cf62b042bae754836bf58275144eed5b089af165/packages/server/socketHandlers/handleOpen.ts) checks the version in the package json and sends it to the `onData` function. 

To add service workers to my local environment, I added [InjectManifest](https://github.com/ParabolInc/parabol/blob/0ea032bfa3d57a52212278d47cee0bbe3d1972e2/scripts/webpack/prod.client.config.js#L139) to dev.client.config. I can see that the `__APP_VERSION__` is set when running `yarn dev` whereas `handleOpen` fires when refreshing the page.

`__APP_VERSION__` could therefore be outdated so [we check](https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/client/hooks/useTrebuchetEvents.ts#L66) whether the version is the same as the more recent `handleOpen` version. If it isn't, we update the registration which triggers the [controllerChange](https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/client/hooks/useServiceWorkerUpdater.ts#L31) event listener function which emits the "Refresh to upgrade" snackbar.

As we get the most recent version of the app in `onData` thanks to `handleOpen`, we can save that version to the atmosphere. In the error component, we can compare whether the atmosphere version is the same as the potentially outdated `__APP_VERSION__` and return a different error message if so.

Questions:

1. Why does the user need to refresh the page? If there's a mismatch between the app versions, the cache typically updates without the user doing anything. I guess we ask them as a precaution? Refreshing the page ensures the `onInstall` function fires in the `sw.ts` file and creates a new cache?
2. In dev, `__APP_VERSION__` can be outdated as it's updated when building the app on `yarn dev` whereas `onData` gets the latest version when refreshing the page. In prod, I imagine `__APP_VERSION__` is updated when we push to production which means it may actually be more up-to-date than the version in `onData`?

P.S. I could create a section in Notion that explains how parts of the codebase work and jot down notes on how this works there.

### DIFF
--- a/packages/client/hooks/useServiceWorkerUpdater.ts
+++ b/packages/client/hooks/useServiceWorkerUpdater.ts
@@ -16,12 +16,13 @@ const useServiceWorkerUpdater = () => {
       }
       atmosphere.eventEmitter.emit('addSnackbar', {
         key: 'newVersion',
-        autoDismiss: 0,
-        message: 'A new version of Parabol is available',
+        autoDismiss: 5,
+        message: 'A new version of Parabol is available 🎉',
         action: {
-          label: 'Refresh to upgrade',
+          label: `See what's changed`,
           callback: () => {
-            window.location.reload()
+            const url = `https://www.google.com/`
+            window.open(url, '_blank', 'noopener')?.focus()
           }
         }
       })
--- a/packages/client/hooks/useServiceWorkerUpdater.ts
+++ b/packages/client/hooks/useServiceWorkerUpdater.ts
@@ -21,7 +21,8 @@ const useServiceWorkerUpdater = () => {
         action: {
           label: `See what's changed`,
           callback: () => {
-            const url = 'https://github.com/ParabolInc/parabol/blob/master/CHANGELOG.md#parabol-change-log'
+            const url =
+              'https://github.com/ParabolInc/parabol/blob/production/CHANGELOG.md#parabol-change-log'
             window.open(url, '_blank', 'noopener')?.focus()
           }
         }
--- a/packages/client/hooks/useServiceWorkerUpdater.ts
+++ b/packages/client/hooks/useServiceWorkerUpdater.ts
@@ -21,7 +21,7 @@ const useServiceWorkerUpdater = () => {
         action: {
           label: `See what's changed`,
           callback: () => {
-            const url = `https://www.google.com/`
+            const url = 'https://github.com/ParabolInc/parabol/blob/master/CHANGELOG.md#parabol-change-log'
             window.open(url, '_blank', 'noopener')?.focus()
           }
         }
--- a/packages/client/hooks/useTrebuchetEvents.ts
+++ b/packages/client/hooks/useTrebuchetEvents.ts
@@ -63,6 +63,7 @@ const useTrebuchetEvents = () => {
       // hacky but that way we don't have to double parse huge json payloads. SSE graphql payloads are pre-parsed
       if (typeof payload !== 'string' || !payload.startsWith('{"version":')) return
       const obj = JSON.parse(payload)
+      if (obj.version) atmosphere.version = obj.version
       if (obj.version !== __APP_VERSION__ && 'serviceWorker' in navigator) {
         const registration = await navigator.serviceWorker.getRegistration()
         registration?.update().catch()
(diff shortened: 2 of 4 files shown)


packages/client/hooks/useServiceWorkerUpdater.ts
```diff
@@ -16,12 +16,13 @@ const useServiceWorkerUpdater = () => {
...
-        autoDismiss: 0,
-        message: 'A new version of Parabol is available',
+        autoDismiss: 5,
+        message: 'A new version of Parabol is available 🎉',
         action: {
-          label: 'Refresh to upgrade',
+          label: `See what's changed`,
           callback: () => {
-            window.location.reload()
+            const url = 'https://github.com/ParabolInc/parabol/blob/master/CHANGELOG.md#parabol-change-log'
```
nickoferrall, line 24: Maybe including the changelog will make users more aware that we're open source and attract devs 🤔

packages/client/hooks/useTrebuchetEvents.ts
```diff
@@ -63,6 +63,7 @@ const useTrebuchetEvents = () => {
       // hacky but that way we don't have to double parse huge json payloads. SSE graphql payloads are pre-parsed
       if (typeof payload !== 'string' || !payload.startsWith('{"version":')) return
```
nickoferrall, line 64: Payload is always an object rather than a string for me but it does contain the version. Perhaps this is the wrong place to update atmosphere (if at all) or perhaps there's a difference in my set-up in dev vs prod so that I'm receiving an object rather than a string

mattkrick: looks like this bit of code is outdated.

the service worker gets updated in 1 of 2 ways:
- in a window onload event when `register` gets called
- in this bit of code, which as you've found out is currently broken!

to trigger the `controllerchange` event, the sw.js file fetched from the server has to be different than what is cached. 

i'd suggest we change this to something like
```ts
      // should always be of type object
      const obj = typeof payload === 'string' ? JSON.parse(payload) : payload
      if (obj.version !== __APP_VERSION__ && 'serviceWorker' in navigator) {
```

packages/client/hooks/useTrebuchetEvents.ts
```diff
@@ -63,6 +63,7 @@ const useTrebuchetEvents = () => {
...
       const obj = JSON.parse(payload)
+      if (obj.version) atmosphere.version = obj.version
```
mattkrick, line 66: nit: i'd name this `serverVersion`
then again, if we fix the service worker, with the above code change, the code someone runs should never be stale which means we don't have to do any of that error handling!

can you verify?

here's how i'd test it (after making the above fix):
1. remove any existing serviceWorkers from the browser under localhost
2. in the server mutation "createTask" write `console.log(newTask.threadSortOrder)` 
3. run `yarn build` and then `yarn start`
4.  create a new task & verify that the server is logging the thread sort order
5. change the CreateTaskMutation client mutation so that `variables.newTask.threadSortOrder = 999`
6. change the version in `handleOpen` to something else
7. run `yarn build` and then `yarn start`. without refreshing the page, the client should reconnect & pop the snackbar.
8. create a task & verify that the server receives 999 as the threadSortOrder

if all of those work, then no error case is needed!

mattkrick (changes requested):
nice find on the borked onData!

packages/client/hooks/useTrebuchetEvents.ts
nickoferrall: That worked! I've pushed the update and removed the error handling from the `ErrorComponent`

mattkrick:
this is soooo stinking cool. how cool is this? we can change the client code, push an update & they'll start using that new client code without a refresh. :tada:

Issue #4962, opened 2021-05-13
Poker: Scope: Jira: Support removing persisted queries


mattkrick:
PersistJiraSearchQueryMutation supports an `isRemove` flag on its input. All that's left to do is call it.

AC
- [ ] When hovering a persisted query, a small X appears on the right side, just like an agenda item
- [ ] hitting the Delete key will delete the persisted query
- [ ] clicking the X will delete the persisted query (no way to do it for mobile. that's OK! it's not a critical flow)

Estimate: 4 hours

PR #5053, opened 2021-06-18
Support removing Jira persisted queries

BartoszJarocki:
Related issue: https://github.com/ParabolInc/parabol/issues/4962 

CleanShot 2021-06-17 at 16 53 53

### DIFF
--- a/packages/client/components/JiraScopingSearchHistoryMenu.tsx
+++ b/packages/client/components/JiraScopingSearchHistoryMenu.tsx
@@ -1,6 +1,6 @@
 import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
-import React from 'react'
+import React, {MouseEvent} from 'react'
 import {commitLocalUpdate, createFragmentContainer} from 'react-relay'
 import useAtmosphere from '../hooks/useAtmosphere'
 import {MenuProps} from '../hooks/useMenu'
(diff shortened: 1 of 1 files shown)


packages/client/components/JiraScopingSearchHistoryMenu.tsx
```diff
@@ -72,13 +98,25 @@ const JiraScopingSearchHistoryMenu = (props: Props) => {
         const projectFilters = projectKeyFilters
           .map((filter) => filter.slice(filter.indexOf(':') + 1))
           .join(', ')
+        const handleRemoveJiraSearchQueryClick = (event) => {
```
mattkrick, line 101: nit: `event: React.MouseEvent`
since we're a legacy codebase we don't enforce `noImplicitAny`, but it's nice to provide a type when we can!

mattkrick: if this is the last item in the list, let's close it for them automatically because there's nothing more for them to do with the open menu.

e.g. 
```ts
if (jiraSearchQueries.length > 1) {
  event.stopPropagation()
}
```

mattkrick (changes requested):
fantastic work! just 1 itty bitty change. feel free to merge after that

packages/client/components/JiraScopingSearchHistoryMenu.tsx
BartoszJarocki: @mattkrick right, I'll pay more attention to adding proper types 🤗 
Also, I've added closing the menu when there's nothing left. 👍

Issue #5026, opened 2021-06-07
Slack App sending notifications for teams that users have left


gcrickman:
## Issue - Bug

A [user](https://app.hubspot.com/contacts/3888472/contact/8644551) 🔒  reports:

> I keep getting these notifications even after I've left the team in question. How can I stop that?
> 
> Clarification: I'm not in that channel anymore and I'm not in that team on Parabol. I'm getting notifications in the Parabol channel, under Apps. 

It seems the Parabol App on Slack is still sending notifications for teams that a user has left.

AC
- [ ] when a person leaves a team, their SlackAuth for that team is removed

Estimate: 4 hours

PR #5055, opened 2021-06-18
Leaving team removes Slack auth

nickoferrall:
PR to resolve issue #5026 

Current problem: if a user integrates Slack and leaves that team, they will still receive notifications when former teammates create meetings.

If a user rejoins the team that they left, it'd be nice if their Slack integration was automatically re-added to save them time and to encourage them to use the integration. This is consistent with how the GitHub & Jira integrations work (as we don't actually remove these integrations when removing a user from a team).

AC:

- [ ] User A and B are in a team. User A integrates Slack and leaves the team. User B creates a meeting. User A does not receive a notification
- [ ] Same happens if User B removes User A from the team
- [ ] Same again if User B removes User A from the organization
- [ ] Same again if User A deletes themself as a user
- [ ] If User A leaves a team, is removed from a team or org and then rejoins the same team, their previous Slack auth is up and running without them needing to reintegrate
- [ ] If User A removes their Slack auth, leaves a team and rejoins that same team, their Slack auth is not automatically integrated

@BartoszJarocki mind reviewing this one when you get the chance?


packages/server/graphql/mutations/deleteUser.ts
```diff
@@ -57,7 +59,9 @@ export default {
     } else if (!user) {
       return {error: {message: 'User not found'}}
     }
-    const {id: userIdToDelete} = user
+    const {id: userIdToDelete, tms} = user
+    const teamMemberIds = tms.map((teamId) => toTeamMemberId(teamId, userIdToDelete))
+    removeSlackAuths(teamMemberIds, true)
```
nickoferrall, line 64: Shouldn't need to `await` as the Slack logic isn't reliant on the subsequent logic

BartoszJarocki (commented):
Looks great! I added a few minor comments to consider 👍

packages/server/graphql/mutations/helpers/activatePrevSlackAuth.ts
```diff
@@ -0,0 +1,46 @@
...
+      return
+    }
+
+    await r({
+      auth: r
+        .table('SlackAuth')
+        .get(authId)
+        .update({isActive: true, updatedAt: now}),
+      notifications: upsertNotifications(userId, teamId, defaultTeamChannelId, slackUserId)
+    }).run()
```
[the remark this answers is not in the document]
nickoferrall: It looks like we don't need to as only the User & Team tables are in flux

nickoferrall:
Thanks for the review! Just pushed a lil update

BartoszJarocki (approved):
Looks good to me! 👍

packages/server/graphql/mutations/helpers/removeSlackAuths.ts
```diff
@@ -0,0 +1,46 @@
+import fromTeamMemberId from 'parabol-client/utils/relay/fromTeamMemberId'
+import getRethink from '../../../database/rethinkDriver'
+
+const removeSlackAuths = async (teamMemberIds: string | string[], removeToken: boolean) => {
+  const r = await getRethink()
+  const now = new Date()
+  const teamMemberIdsArr = Array.isArray(teamMemberIds) ? teamMemberIds : [teamMemberIds]
+  const {userId} = fromTeamMemberId(teamMemberIdsArr[0])
```
mattkrick, line 8: this looks a little hairy because i'd assume I could call this function with a list of any teamMemberIds, but in fact I have to call the function with an array of teamMemberIds that all have the same userId.

to clean up, let's make the function signature `(userId: string, teamIds: string[], removeToken?: boolean)`

packages/server/graphql/mutations/helpers/removeSlackAuths.ts
```diff
@@ -0,0 +1,46 @@
...
+    return {authId: null, error}
+  }
+
+  const authIds = existingAuths.map(({id}) => id)
+  await r({
+    auth: r
+      .table('SlackAuth')
+      .getAll(r.args(authIds))
+      .update((row) => ({
+        botAccessToken: r.branch(removeToken, null, row('botAccessToken')),
```
mattkrick, line 31: since `removeToken` is known before calling rethinkdb you don't need conditional logic inside the DB query.
just `botAccessToken: removeToken ? null : undefined` (rethinkdb ignores any props that have `undefined` as a value)

packages/server/graphql/mutations/helpers/removeSlackAuths.ts
```diff
@@ -0,0 +1,46 @@
...
+      })),
+    notifications: r
+      .table('SlackNotification')
+      .getAll(r.args(teamIds), {index: 'teamId'})
+      .filter({userId})
+      .delete()
+  }).run()
+
+  const authId = authIds[0]
+  return {authId, error: null}
```
mattkrick, line 43: we're skimping on the rest of the authIds!
let's return `authIds: string[]` and let the calling function decide to pick the first one.

if you really wanted to get fancy, you could use a typescript overload like:
```ts
function removeSlackAuth(teamIds: string): {authId: string}
function removeSlackAuth(teamIds: string[]): {authId: string[]}
```

but those are mainly just for showing off https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads

packages/server/graphql/mutations/removeTeamMember.ts
```diff
@@ -31,6 +32,7 @@ export default {
     }
 
     // RESOLUTION
+    removeSlackAuths(teamMemberId, false)
```
mattkrick, line 35: i feel like this call belongs in `removeTeamMember`. that way you won't have to modify removeOrgUser,

mattkrick (changes requested):
looking good! 
you did a great job of refactoring the logic so removing the slack auth stuff wasn't repeated.
just a couple suggestions on function APIs

nickoferrall:
Thanks for the review @mattkrick - updated!

packages/server/graphql/mutations/removeSlackAuth.ts
```diff
@@ -28,11 +27,11 @@ export default {
     }
 
     // RESOLUTION
-    const teamMemberId = toTeamMemberId(teamId, viewerId)
-    const {authId, error} = await removeSlackAuths(teamMemberId, true)
+    const {authIds, error} = await removeSlackAuths(viewerId, teamId, true)
     if (error) {
       return standardError(error, {userId: viewerId})
     }
+    const authId = authIds![0]
```
mattkrick, line 34: since `removeSlackAuths` returns a discriminated union on error being null, if you don't destructure immediately, it will give you type safety:

```ts
    const res = await removeSlackAuths(viewerId, teamId, true)
    if (res.error) {
      return standardError(res.error, {userId: viewerId})
    }
    const {authIds} = res
    const authId = authIds[0]
```

mattkrick:
LGTM! 1 itty bitty explanation on why typescript wasn't giving you the typesafety you were expecting. feel free to merge it with or without that change

packages/server/graphql/mutations/removeSlackAuth.ts
nickoferrall: Ooh well that's wonderful knowledge. Updated & will merge now!

PR #5062, opened 2021-06-22
Add poker comment & story count

nickoferrall:
PR to resolve issue #4983 

This PR saves the story count and comment count to the db when a poker meeting ends. It also adds a migration to update previous meeting stats.

### AC:

- [ ] The story count and comment count for a Poker meeting in the timeline feed are accurate
- [ ] Story count and comment count are accurate in Slack notifications

### DIFF
--- a/packages/server/database/migrations/20210622170418-addPokerStats.ts
+++ b/packages/server/database/migrations/20210622170418-addPokerStats.ts
@@ -1,0 +1,54 @@
+import {R} from 'rethinkdb-ts'
+import {MeetingTypeEnum, NewMeetingPhaseTypeEnum} from 'parabol-client/types/graphql'
+
+export const up = async function(r: R) {
+  try {
+    const getDiscussionIds = (meetingRow) => {
+      return meetingRow('phases')
+        .filter({phaseType: NewMeetingPhaseTypeEnum.ESTIMATE})('stages')
+        .concatMap((row) => row('discussionId'))
+    }
+    const getStories = (meetingRow) => {
+      return meetingRow('phases')
+        .filter({phaseType: NewMeetingPhaseTypeEnum.ESTIMATE})('stages')
+        .concatMap((row) => row('serviceTaskId'))
+    }
+
+    await r
+      .table('NewMeeting')
+      .filter({meetingType: MeetingTypeEnum.poker})
+      .filter((row) =>
+        row('endedAt')
+          .default(null)
+          .ne(null)
+      )
+      .update(
+        (row) => ({
+          commentCount: r
+            .table('Comment')
+            .getAll(r.args(getDiscussionIds(row)), {index: 'discussionId'})
+            .count()
+            .default(0),
+          storyCount: getStories(row)
+            .count()
+            .default(0)
+        }),
+        {nonAtomic: true}
+      )
+      .run()
+  } catch (e) {
+    console.log(e)
+  }
+}
+
+export const down = async function(r: R) {
+  try {
+    await r
+      .table('NewMeeting')
+      .filter({meetingType: MeetingTypeEnum.poker})
+      .replace((row) => row.without('commentCount', 'storyCount'))
+      .run()
+  } catch (e) {
+    console.log(e)
+  }
+}
--- a/packages/server/database/migrations/20210624120419-addPokerStats.ts
+++ b/packages/server/database/migrations/20210624120419-addPokerStats.ts
@@ -8,10 +8,12 @@ export const up = async function(r: R) {
         .filter({phaseType: NewMeetingPhaseTypeEnum.ESTIMATE})('stages')
         .concatMap((row) => row('discussionId'))
     }
-    const getStories = (meetingRow) => {
+    const getCompletedStories = (meetingRow) => {
       return meetingRow('phases')
         .filter({phaseType: NewMeetingPhaseTypeEnum.ESTIMATE})('stages')
-        .concatMap((row) => row('serviceTaskId'))
+        .nth(0)
+        .filter({isComplete: true})
+        .map((row) => row('serviceTaskId'))
     }
 
     await r
@@ -29,7 +31,7 @@ export const up = async function(r: R) {
             .getAll(r.args(getDiscussionIds(row)), {index: 'discussionId'})
             .count()
             .default(0),
-          storyCount: getStories(row)
+          storyCount: getCompletedStories(row)
             .count()
             .default(0)
         }),
--- a/packages/server/graphql/mutations/endSprintPoker.ts
+++ b/packages/server/graphql/mutations/endSprintPoker.ts
@@ -65,7 +65,9 @@ export default {
     if (!currentStageRes) {
       return standardError(new Error('Cannot find facilitator stage'), {userId: viewerId})
     }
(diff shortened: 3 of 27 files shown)


packages/server/graphql/mutations/endSprintPoker.ts
```diff
@@ -52,7 +53,7 @@ export default {
 
     // RESOLUTION
     // remove hovering data from redis
-    const estimatePhase = phases.find((phase) => phase.phaseType === 'ESTIMATE')!
+    const estimatePhase = phases.find((phase) => phase.phaseType === 'ESTIMATE')! as EstimatePhase
```
nickoferrall, line 56: It looks like we always need to cast the phase types even when filtering by `phaseType`, (see [example](https://github.com/ParabolInc/parabol/blob/027bb8d7f51d4b958af29569ec181f86d347bb5b/packages/server/graphql/mutations/updateNewCheckInQuestion.ts#L55)). 
I tried adding `!phaseType` to each db phase type such as the `EstimatePhase` (as we do in [CheckInPhase](https://github.com/ParabolInc/parabol/blob/db45e3b5a4c27e917dcc267d33175bae8a04970b/packages/server/database/types/CheckInPhase.ts#L15)) but TypeScript still reads it as the `GenericMeetingPhase`. 

In meeting types we added the `meetingType` field (see [MeetingPoker](https://github.com/ParabolInc/parabol/blob/6b83c0a272bfa0e4cad607ed3b20970b3f2a4784/packages/server/database/types/MeetingPoker.ts#L19)) and TypeScript can successfully discriminate. The only difference between meeting types and meeting phases that I see if in the meeting types, we include `meetingType` in `super`. 

I tried adding the `phaseType` in `super` in EstimatePhase but get the error `Type '{ phaseType: string; }' is not assignable to type '"vote"'`. Any ideas on how I can discriminate phase types without casting?

BartoszJarocki: @nickoferrall the only way I can think of is doing this:
```
    const estimatePhase = phases.find(
      (phase: GenericMeetingPhase): phase is EstimatePhase => phase.phaseType === 'ESTIMATE'
    )!
```
It's called type predicate, more about it here: https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

nickoferrall: Yep, that's a good option too but I was hoping I could just use the `phaseType` like we do with `meetingType` [here](https://github.com/ParabolInc/parabol/blob/af9366fc995ec0cae4ac5017bf28358ba4f3554c/packages/server/graphql/mutations/helpers/notifySlack.ts#L124). By checking the `meetingType`, TypeScript knows which `meeting` type it is without using type predicates or casting but I think this is only possible with if/else or switch statments

mattkrick: the built-in find method isn't too helpful here. 

might be worth it to make a helper function that does a single cast so everything that calls it doesn't have to do a cast

```ts
    // if we have a lookup table somewhere you can use that instead
    interface PhaseTypeLookup {
      ESTIMATE: EstimatePhase
      updates: UpdatesPhase
    }
    const getPhase = <T extends keyof PhaseTypeLookup>(phases: GenericMeetingPhase[], phaseType: T) => {

      return phases.find((phase) => phase.phaseType === phaseType) as unknown as PhaseTypeLookup[T]
    }


    const estimatePhase = getPhase(phases, 'ESTIMATE')
```

packages/server/graphql/mutations/endSprintPoker.ts
```diff
@@ -64,6 +65,8 @@ export default {
     if (!currentStageRes) {
       return standardError(new Error('Cannot find facilitator stage'), {userId: viewerId})
     }
+    const storyCount = new Set(estimateStages.map(({serviceTaskId}) => serviceTaskId)).size
```
mattkrick, line 68: let's only count completed stages

mattkrick (changes requested):
itty bitty edge case of counting completed stories vs. stories added. other than that, good to merge!

packages/server/graphql/mutations/endSprintPoker.ts
nickoferrall: I like it! I've refactored the phase casting to use this new `getPhase` function

packages/server/database/migrations/20210622170418-addPokerStats.ts
```diff
@@ -0,0 +1,54 @@
...
+export const up = async function(r: R) {
+  try {
+    const getDiscussionIds = (meetingRow) => {
+      return meetingRow('phases')
+        .filter({phaseType: NewMeetingPhaseTypeEnum.ESTIMATE})('stages')
+        .concatMap((row) => row('discussionId'))
+    }
+    const getStories = (meetingRow) => {
+      return meetingRow('phases')
+        .filter({phaseType: NewMeetingPhaseTypeEnum.ESTIMATE})('stages')
```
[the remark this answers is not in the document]
nickoferrall: Makes sense. I've pushed an update so that the story count is calculated using complete stages but the commentCount is calculated using complete & incomplete stages

packages/server/database/migrations/20210624120419-addPokerStats.ts
```diff
@@ -8,10 +8,12 @@ export const up = async function(r: R) {
         .filter({phaseType: NewMeetingPhaseTypeEnum.ESTIMATE})('stages')
         .concatMap((row) => row('discussionId'))
     }
-    const getStories = (meetingRow) => {
+    const getCompletedStories = (meetingRow) => {
       return meetingRow('phases')
         .filter({phaseType: NewMeetingPhaseTypeEnum.ESTIMATE})('stages')
-        .concatMap((row) => row('serviceTaskId'))
+        .nth(0)
+        .filter({isComplete: true})
```
mattkrick, line 15: @nickoferrall it looks like this is filtering the 0th element. do we want to run the filter 1 line above?

packages/server/utils/getPhase.ts
```diff
@@ -0,0 +1,22 @@
...
+interface PhaseTypeLookup {
+  agendaitems: AgendaItemsPhase
+  checkin: CheckInPhase
+  discuss: DiscussPhase
+  ESTIMATE: EstimatePhase
+  reflect: ReflectPhase
+  updates: UpdatesPhase
+}
+
+const getPhase = <T extends keyof PhaseTypeLookup>(phases: GenericMeetingPhase[], phaseType: T) => {
```
mattkrick, line 18: :fire:

packages/server/database/migrations/20210624120419-addPokerStats.ts
nickoferrall: nested-stages

`stages` is a nested array so I'm using `nth(0)` to access it. If I try to filter on `stages` without `nth(0)`, I get an error as `isComplete` isn't accessible. I tested the migration locally to make sure the stats are accurate in the timeline and it seems to be doing the business

mattkrick: Ohhhh! i see what it's doing.
the filter is returning an array of estimate phases & you're grabbing the stages from each one & then limiting it to the first phase. good deal :+1:

PR #5067, opened 2021-06-23
Pw protect staging

tiffanyhan:
#### AC:
   - [ ] when deployed to staging, staging server is password protected and the app functions normally
   - [x] when deployed to development, dev server is password protected and the app functions normally
   - [ ] can sign into dev or staging server with username and password found in [action-devops repo](https://github.com/ParabolInc/action-devops/blob/master/environments/staging#L35-L36)
   - [x] locally, the app functions normally


tiffanyhan:
[Dokku plugin](https://github.com/dokku/dokku-http-auth) was installed on dev and staging servers to turn on pw protection

tianrunhe (approved):
LGTM! Please update any related doc😉

PR #5070, opened 2021-06-24
Sort jiraDimensionFields before updates & backfill script

tianrunhe:
1. Sort `jiraDimensionFields` by JSON strings before updates to DB
2. Backfill scripts for RethinkDB

### DIFF
--- a/packages/server/graphql/mutations/updateJiraDimensionField.ts
+++ b/packages/server/graphql/mutations/updateJiraDimensionField.ts
@@ -9,6 +9,7 @@ import {getUserId, isTeamMember} from '../../utils/authorization'
 import publish from '../../utils/publish'
 import {GQLContext} from '../graphql'
 import UpdateJiraDimensionFieldPayload from '../types/UpdateJiraDimensionFieldPayload'
+import stringify from 'fast-json-stable-stringify'
 
 const updateJiraDimensionField = {
   type: GraphQLNonNull(UpdateJiraDimensionFieldPayload),
@@ -102,7 +103,7 @@ const updateJiraDimensionField = {
     const MAX_JIRA_DIMENSION_FIELDS = 100 // prevent a-holes from unbounded growth of the Team object
     const sortedJiraDimensionFields = jiraDimensionFields
       .slice(jiraDimensionFields.length - MAX_JIRA_DIMENSION_FIELDS)
-      .sort((a, b) => (JSON.stringify(a) < JSON.stringify(b) ? -1 : 1))
+      .sort((a, b) => (stringify(a) < stringify(b) ? -1 : 1))
     await Promise.all([
       r
         .table('Team')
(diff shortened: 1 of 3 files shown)


packages/server/graphql/mutations/updateJiraDimensionField.ts
```diff
@@ -100,21 +100,20 @@ const updateJiraDimensionField = {
     }
 
     const MAX_JIRA_DIMENSION_FIELDS = 100 // prevent a-holes from unbounded growth of the Team object
+    const sortedJiraDimensionFields = jiraDimensionFields
+      .slice(jiraDimensionFields.length - MAX_JIRA_DIMENSION_FIELDS)
+      .sort((a, b) => (JSON.stringify(a) < JSON.stringify(b) ? -1 : 1))
```
mattkrick, line 105: `JSON.stringify` is not a stable stringifier, meaning the order of the objects is non-deterministic.
while it will _probably_ work for this node version & future versions of V8, it's not guaranteed & if the order ever does change, that'll be a tricky bug to find.
instead, use fast-json-stable-stringify

packages/server/utils/ensureJiraDimensionField.ts
```diff
@@ -64,6 +64,7 @@ const ensureJiraDimensionField = async (
         jiraDimensionFields: team('jiraDimensionFields')
           .default([])
           .union(fieldsToAdd)
+          .orderBy((jiraDimensionField) => jiraDimensionField.coerceTo('string'))
```
mattkrick, line 67: can you verify if this stringifier is stable in the docs or github issue? if not, we might want to union `[...newJiraDimensionFields, ...fieldsToAdd]` in JS land to guarantee order

mattkrick (changes requested):
minor concern over rethinkdb's stringify stability, i know the JS version is bad. if you can verify that we're good to go!

packages/server/utils/ensureJiraDimensionField.ts
```diff
@@ -64,6 +64,7 @@ const ensureJiraDimensionField = async (
         jiraDimensionFields: team('jiraDimensionFields')
           .default([])
           .union(fieldsToAdd)
```
mattkrick, line 66: ohhhhh! this is probably the culprit.
i think rethinkdb's union method is non-deterministic

tianrunhe: That's right! [Doc](https://rethinkdb.com/api/javascript/union): 
```
results will be mixed together; this is the fastest setting, but ordering of elements is not guaranteed. (This is the default.)
```

packages/server/utils/ensureJiraDimensionField.ts
tianrunhe: I couldn't find anything quick so I did the union & ordering in JS land. This is also consistent with `updateJiraDimensionField.ts ` so I like it better.

Issue #5056, opened 2021-06-21
Agenda Item Sort Order Bug


mattkrick:
Sometimes changing the sort order doesn't propagate to other team members.
Same problem with the optimistic updater.
Refreshing first & then changing order works for the mutator.
Found in meetingId: 4B3UsKdy8g

AC
- [ ] can resort agenda items across multiple clients & achieve eventual consistency

Estimate: 4 hours

PR #5071, opened 2021-06-24
Fixed agenda items sorting and deleting

BartoszJarocki:
Related issue: https://github.com/ParabolInc/parabol/issues/5056 

Agenda items related relay store updaters were broken because they were reading non-exsiting `agendaItemId` property from `AgendaItemsStage`. 
Screenshot 2021-06-24 at 18 11 29
 
Now, both sorting and deletion work without any issues (currently, on prod sorting and deletion are broken).

### DIFF
--- a/packages/client/mutations/UpdateAgendaItemMutation.ts
+++ b/packages/client/mutations/UpdateAgendaItemMutation.ts
@@ -7,6 +7,7 @@ import updateProxyRecord from '../utils/relay/updateProxyRecord'
 import {UpdateAgendaItemMutation as TUpdateAgendaItemMutation} from '../__generated__/UpdateAgendaItemMutation.graphql'
 import {UpdateAgendaItemMutation_team} from '~/__generated__/UpdateAgendaItemMutation_team.graphql'
 import {ActionMeeting_meeting} from '~/__generated__/ActionMeeting_meeting.graphql'
+import {AgendaItem_agendaItem} from '../__generated__/AgendaItem_agendaItem.graphql'
 
 graphql`
   fragment UpdateAgendaItemMutation_team on UpdateAgendaItemPayload {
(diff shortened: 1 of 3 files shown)


packages/server/graphql/mutations/removeAgendaItem.ts
```diff
@@ -44,10 +44,8 @@ export default {
       return standardError(new Error('Agenda item not found'), {userId: viewerId})
     }
     const filterFn = (stage: AgendaItemsStage) => stage.agendaItemId === agendaItemId
-    const meetingIds = await removeStagesFromMeetings(filterFn, teamId, dataLoader)
-    // safe to do so because we guarantee only 1 action meeting at the same time
-    const [meetingId] = meetingIds
-    const data = {agendaItem, meetingId}
+    await removeStagesFromMeetings(filterFn, teamId, dataLoader)
+    const data = {agendaItem, meetingId: agendaItem.meetingId}
```
BartoszJarocki, line 48: `const [meetingId] = meetingIds` this was causing to return wrong meeting id when there's more than one meeting active. I've changed it to return `meetingId` strictly related to `agendaItem`

packages/client/mutations/UpdateAgendaItemMutation.ts
```diff
@@ -44,9 +44,8 @@ const handleUpdateAgendaPhase = (
       phase ? phase.getValue('phaseType') === 'agendaitems' : false
     )
     if (!agendaPhase) return
-    const getSortOrder = (stage) => {
-      const agendaItemId = stage.getValue('agendaItemId')
-      const agendaItem = store.get<AgendaItem_agendaItem>(agendaItemId)
+    const getSortOrder = (stage: RecordProxy) => {
+      const agendaItem = stage.getLinkedRecord<AgendaItem_agendaItem>('agendaItem')
```
mattkrick, line 48: :facepalm: great find!

packages/server/graphql/mutations/removeAgendaItem.ts
mattkrick: this makes a lot more sense :+1:

mattkrick (approved):
fantastic work!

Issue #4972, opened 2021-05-17
Move AtlassianAuth to Postgres


mattkrick:
things like AtlassianAuth don't need to follow the same thoughtful process like User, Team, etc. to get migrated because it's not real-time & it's not joined in many places.

AC
- [ ] A postgres migration exists that creates the AtlassianAuth table in postgres. it's the same as the RethinkDB version, but also includes scope (see #4971 for inspiration)
- [ ] all references to the rethinkdb AtlassianAuth now point to postgres

Estimate: 8 hours

PR #5085, opened 2021-06-29
Migrating AtlassianAuth to Postgres

BartoszJarocki:
Related issue: https://github.com/ParabolInc/parabol/issues/4972

I've replaced all reads and writes I could find. It includes:
1. `addAtlassianAuth` mutation
2. `removeAtlassianAuth` mutation
3. `persistJiraSearchQuery` mutation
4. all resolvers in `AtlassianIntegration` 
5. `customLoaderMakers` loaders 

Few notes
1. `backupOrganization` isn't updated (I think we have to update the whole mutation to use PG 
2. `foreignLoaderMakers` contains the following loaders
```
export const atlassianAuthByUserId = new LoaderMakerForeign(
  'atlassianAuths',
  'userId',
  async (userIds) => {
    const r = await getRethink()
    return r
      .table('AtlassianAuth')
      .run()
  }
)
export const atlassianAuthByTeamId = new LoaderMakerForeign(
  'atlassianAuths',
  'teamId',
  async (teamIds) => {
    const r = await getRethink()
    return r
      .table('AtlassianAuth')
      .getAll(r.args(teamIds), {index: 'teamId'})
      .run()
  }
)
```
I haven't found any usages so I haven't touched it at all. 
Other than that, I pointed few things in the comments. I tested all the scenarios, all the reads and writes work correctly.


packages/server/graphql/types/AtlassianIntegration.ts
```diff
@@ -182,20 +184,18 @@ const AtlassianIntegration = new GraphQLObjectType<any, GQLContext>({
...
-          const r = await getRethink()
-          await r
-            .table('AtlassianAuth')
-            .get(atlassianAuthId)
-            .update({
-              jiraSearchQueries: unexpiredQueries
-            })
-            .run()
+        const unexpiredQueries = jiraSearchQueries.filter(
+          (query) => new Date(query.lastUsedAt) > thresh
```
BartoszJarocki, line 191: beware, previously with data that came from rethink `lastUsedAt` was `Date`, with PG now it's a string so the old comparison didn't work.

packages/server/postgres/migrations/1624862201216_addAtlassianAuthTable.ts
```diff
@@ -0,0 +1,84 @@
...
+      teamId,
+      userId,
+      jiraSearchQueries,
+      cloudIds,
+      scope: AtlassianManager.SCOPE
+    })
+  )
+  if (auths.length) {
+    await insertAtlassianAuthsQuery.run({auths}, client)
+  }
```
BartoszJarocki, line 77: Should this be a part of the migration or a separate post-deploy script?

packages/server/postgres/migrations/1624862201216_addAtlassianAuthTable.ts
```diff
@@ -0,0 +1,84 @@
...
+  await client.connect()
+
+  await client.query(`
+    CREATE TABLE IF NOT EXISTS "AtlassianAuth" (
+      "accessToken" VARCHAR(2600) NOT NULL,
+      "refreshToken" VARCHAR(2600) NOT NULL,
+      "createdAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+      "updatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+      "isActive" BOOLEAN DEFAULT TRUE NOT NULL,
+      "jiraSearchQueries" JSONB[] NOT NULL DEFAULT '{}',
```
BartoszJarocki, line 31: Should we create Postgres custom types for that kind of data? As a result, `pgtyped` gives us the `JsonArray` type and it makes it hard to use strongly typed params in resolvers, in this case, `AltassianAuth`.

packages/server/graphql/types/AtlassianIntegration.ts
```diff
@@ -182,20 +184,18 @@ const AtlassianIntegration = new GraphQLObjectType<any, GQLContext>({
       type: GraphQLNonNull(GraphQLList(GraphQLNonNull(JiraSearchQuery))),
       description:
         'the list of suggested search queries, sorted by most recent. Guaranteed to be < 60 days old',
-      resolve: async ({id: atlassianAuthId, jiraSearchQueries}: AtlassianAuth) => {
+      resolve: async ({teamId, userId, jiraSearchQueries}) => {
```
BartoszJarocki, line 187: I removed Param type here because `AtlassianAuth` has `jiraSearchQueries` as `JiraSearchQuery` type and newly generated PG methods expect `JsonArray`. We could use `IGetAtlassianAuthByUserIdTeamIdQueryResult` here but I'm not sure if it's the best idea. Any thoughts?

mattkrick:
reviewing now...

packages/server/postgres/migrations/1624862201216_addAtlassianAuthTable.ts
```diff
@@ -0,0 +1,84 @@
...
+    }) => ({
+      accessToken,
+      refreshToken,
+      createdAt,
+      updatedAt,
+      isActive,
+      accountId,
+      teamId,
+      userId,
+      jiraSearchQueries,
```
mattkrick, line 70: passing in null will not trigger a default value in pg
maybe
`jiraSearchQueries: jiraSearchQueries || []`

packages/server/postgres/migrations/1624862201216_addAtlassianAuthTable.ts
mattkrick: here is perfect! i'm all for trying to keep as many migrations in the pre-deploy as possible

packages/server/postgres/migrations/1624862201216_addAtlassianAuthTable.ts
```diff
@@ -0,0 +1,84 @@
...
+      cloudIds,
+      scope: AtlassianManager.SCOPE
+    })
+  )
+  if (auths.length) {
+    await insertAtlassianAuthsQuery.run({auths}, client)
```
mattkrick, line 76: could we use the upsert query here?
no worries if not, just nice to keep as few queries as possible

packages/server/dataloader/customLoaderMakers.ts
```diff
@@ -262,6 +268,26 @@ export const freshAtlassianAuth = (parent: RethinkDataLoader) => {
   )
 }
 
+export const atlassianAuth = (parent: RethinkDataLoader) => {
```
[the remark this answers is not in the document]
BartoszJarocki: Removed 🙏

packages/server/graphql/types/AtlassianIntegration.ts
BartoszJarocki: I'm pretty sure it's because underlying PG type is simply `JSONB` not the `Date`

packages/server/dataloader/customLoaderMakers.ts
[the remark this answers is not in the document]
BartoszJarocki: removed 👍

packages/server/graphql/types/PersistJiraSearchQueryPayload.ts
```diff
@@ -9,8 +9,8 @@ export const PersistJiraSearchQuerySuccess = new GraphQLObjectType<any, GQLConte
     atlassianIntegration: {
       type: AtlassianIntegration,
       description: 'The newly created auth',
-      resolve: async ({atlassianAuthId}, _args, {dataLoader}) => {
-        return dataLoader.get('atlassianAuths').load(atlassianAuthId)
+      resolve: async ({teamId, userId}, _args, {dataLoader}) => {
+        return await dataLoader.get('atlassianAuth').load({teamId, userId})
```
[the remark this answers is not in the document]
BartoszJarocki: done 👍

packages/server/postgres/migrations/1624862201216_addAtlassianAuthTable.ts
BartoszJarocki: That's a really good point, I've tested it only with `jiraSearchQueries` filled in the DB 😳 🙏

packages/server/postgres/migrations/1624862201216_addAtlassianAuthTable.ts
BartoszJarocki: They slightly differ from each other, ie. `insertAtlassianAuthsQuery` contains params that are not present in upsert query, like `createdAt`, `updatedAt`, `jiraSearchQueries`, `isActive`. So I'd have to provide these when using upsert query. Do you think it's worth the change?

mattkrick: those 4 fields all have default values in pg so it _should_ work as long as we're not passing in an unacceptable value like `null` for them

mattkrick:
### Re: `JSONB[]`

I wasn't sure about this, so here's how i solved it:
- Get some data in a JSONB[] column: `UPDATE "GitHubAuth" SET "githubSearchQueries" = ARRAY['{"foo": "bar"}'::jsonb]`
- Hit up the docs regarding arrays, didn't find much
- Hit up the docs regarding json, found this: https://www.postgresql.org/docs/12/functions-json.html#FUNCTIONS-JSON-CREATION-TABLE
- 3rd function down is called `array_to_json` which does the trick :tada:

packages/server/postgres/queries/updateJiraSearchQueries.ts
```diff
@@ -0,0 +1,20 @@
...
+  jiraSearchQueries: {
+    queryString: string
+    projectKeyFilters?: string[]
+    lastUsedAt?: string
+    isJQL: boolean
+  }[]
+}
+
+const updateJiraSearchQueries = async (params: UpdateJiraSearchQueryParams) => {
+  await updateJiraSearchQueriesQuery.run(params as IUpdateJiraSearchQueriesQueryParams, getPg())
```
BartoszJarocki, line 18: TS complains without an explicit cast 😱 
```
parabol-server: postgres/queries/updateJiraSearchQueries.ts:18:42 - error TS2345: Argument of type 'UpdateJiraSearchQueryParams' is not assignable to parameter of type 'IUpdateJiraSearchQueriesQueryParams'.
parabol-server: Types of property 'jiraSearchQueries' are incompatible.
parabol-server:   Type '{ queryString: string; projectKeyFilters?: string[] | undefined; lastUsedAt?: string | undefined; isJQL: boolean; }[]' is not assignable to type 'void | JsonArray | null'.
parabol-server:     Type '{ queryString: string; projectKeyFilters?: string[] | undefined; lastUsedAt?: string | undefined; isJQL: boolean; }[]' is not assignable to type 'JsonArray'.
parabol-server:       Type '{ queryString: string; projectKeyFilters?: string[] | undefined; lastUsedAt?: string | undefined; isJQL: boolean; }' is not assignable to type 'string | number | boolean | { [key: string]: Json; } | Json[] | null'.
parabol-server:         Type '{ queryString: string; projectKeyFilters?: string[] | undefined; lastUsedAt?: string | undefined; isJQL: boolean; }' is not assignable to type '{ [key: string]: Json; }'.
parabol-server:           Property 'projectKeyFilters' is incompatible with index signature.
parabol-server:             Type 'string[] | undefined' is not assignable to type 'Json'.
parabol-server:               Type 'undefined' is not assignable to type 'Json'.
parabol-server: 18   await updateJiraSearchQueriesQuery.run(params, getPg())
```

packages/server/postgres/queries/getAtlassianAuthByUserIdTeamId.ts
```diff
@@ -0,0 +1,56 @@
...
+    refreshToken,
+    createdAt,
+    updatedAt,
+    isActive,
+    cloudIds,
+    scope,
+    accountId,
+    teamId,
+    userId,
+    jiraSearchQueries: jiraSearchQueries.map((jsq: any) => {
```
BartoszJarocki, line 44: I haven't found any better way than using `any` here otherwise, TS complains

packages/server/postgres/queries/getAtlassianAuthByUserIdTeamId.ts
```diff
@@ -0,0 +1,56 @@
...
+      return {
+        id: jsq.id,
+        queryString: jsq.queryString,
+        projectKeyFilters: jsq.projectKeyFilters,
+        lastUsedAt: jsq.lastUsedAt ? new Date(jsq.lastUsedAt) : undefined,
+        isJQL: jsq.isJQL
+      }
+    })
+  } as AtlassianAuth
```
BartoszJarocki, line 53: @mattkrick I know it's a bit verbose but it allows us to shape the object exactly as we want (maybe validate in the future all the props?). Please let me know if that's acceptable or if there's some cleaner solution 🙏

packages/server/postgres/migrations/1624862201216_addAtlassianAuthTable.ts
BartoszJarocki: @mattkrick there's one more issue, `insertAtlassianAuthsQuery` takes an array, upsert takes single value, so we'd have to insert in the loop which will probably be slower. What do you think? 🤔

mattkrick: ahhh good point. sure, let's keep it

packages/server/postgres/queries/getAtlassianAuthByUserIdTeamId.ts
mattkrick: remember the null case too!

mattkrick:
added a check for `null` & merged. great work!

Issue #4902, opened 2021-04-20
Add accordion to check-in left nav phase change


mattkrick:
​

AC
- [ ] in the meeting left nav, when a phase becomes active & the phase children show (e.g. users during check-in) the height grows like an according.
- [ ] when the phase goes inactive, the accordion closes

Estimate: 8 hours

PR #5086, opened 2021-06-29
Add accordion animation to sidebar

nickoferrall:
PR to resolve issue #4902

### DIFF
--- a/packages/client/components/ActionSidebarAgendaItemsSection.tsx
+++ b/packages/client/components/ActionSidebarAgendaItemsSection.tsx
@@ -42,7 +42,7 @@ const ActionSidebarAgendaItemsSection = (props: Props) => {
   const childHeight = Math.min(maxAgendaItemsHeight, maxHeight)
 
   return (
-    <MeetingSidebarPhaseItemChild isActive height={childHeight}>
+    <MeetingSidebarPhaseItemChild height={childHeight}>
       <AgendaListAndInput
         gotoStageId={handleClick}
         isDisabled={!isUpdatesNavigable}
--- a/packages/client/components/ActionSidebarAgendaItemsSection.tsx
+++ b/packages/client/components/ActionSidebarAgendaItemsSection.tsx
@@ -1,12 +1,10 @@
 import graphql from 'babel-plugin-relay/macro'
-import React from 'react'
+import React, {useMemo} from 'react'
 import {createFragmentContainer} from 'react-relay'
 
 import {ActionSidebarAgendaItemsSection_meeting} from '../__generated__/ActionSidebarAgendaItemsSection_meeting.graphql'
 import useGotoStageId from '../hooks/useGotoStageId'
-import AgendaListAndInput, {
-  getAgendaItems
-} from '../modules/teamDashboard/components/AgendaListAndInput/AgendaListAndInput'
+import AgendaListAndInput from '../modules/teamDashboard/components/AgendaListAndInput/AgendaListAndInput'
 import MeetingSidebarPhaseItemChild from './MeetingSidebarPhaseItemChild'
 import {NavSidebar} from '../types/constEnums'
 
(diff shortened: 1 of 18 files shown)


mattkrick:
when you animate `maxHeight` from 0 to 1000 it's gonna go _real_ fast. Let's say you use a 1000ms duration, that's 1px per ms.
if it's `height` is really only 100px, then it'll complete in 100ms. when you shrink it, it'll go from that 100 to 0, which will be slower. 
in other words, if you use maxHeight, you have to calculate _exactly_ how how it will be in order to get the effect you want.

i see MUI uses a different approach, but I haven't dug into it yet. 
the accordion is here: https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Accordion/Accordion.js
but more interesting is the Collapse component that it uses: https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Collapse/Collapse.js

you can see how they're measuring the height here: https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Collapse/Collapse.js#L167

whenever i need to understand something that MUI wrote, I like to copy the whole file, and then remove all the extra variables that aren't relevant to my usecase. then i can see what it's _really_ doing & try to emulate.
Let me know if this is enough to move forward or if you want to dig in together

nickoferrall:
Thanks! I played around with the `maxHeight` but ran into the same problem although I can now see the difference when using super large or small numbers. I assumed it would calculate the speed up using the actual height (which is capped by max height) rather than using the max.

I'll use the MUI method for calculating the height and dig into their code. Appreciate the help!

packages/client/components/MeetingSidebarTeamMemberStageItems.tsx
```diff
@@ -9,47 +9,66 @@ import MeetingSubnavItem from '../components/MeetingSubnavItem'
...
-  const {viewerId} = atmosphere
-  const isViewerFacilitator = viewerId === facilitatorUserId
+
   return (
-    <MeetingSidebarPhaseItemChild>
-      <ScrollStageItems>
-        {localPhase.stages.map((stage) => {
+    <MeetingSidebarPhaseItemChild height={childHeight}>
+      <ScrollStageItems showScrollbar={showScrollbar}>
+        {stages!.map((stage) => {
```
nickoferrall, line 71: When transitioning from checkin or updates phase to agendaitems phase, the localPhase will be agendaitems during the 300ms transition when the phase is closing. As we still want to show the checkin or updates children during those 300ms, I'm using the stages instead of the localPhase

packages/client/components/ActionSidebarAgendaItemsSection.tsx
```diff
@@ -1,32 +1,48 @@
...
   const handleClick = async (stageId: string) => {
     gotoStageId(stageId).catch()
     handleMenuClick()
   }
   // show agenda (no blur) at all times if the updates phase isNavigable
   // facilitator can click on updates nav item before completing all check-in stages
-  const updatesPhase = meeting.phases!.find((phase) => phase.phaseType === 'updates')!
+  const updatesPhase = phases.find((phase) => phase.phaseType === 'updates')
   const isUpdatesNavigable = updatesPhase && updatesPhase.stages![0].isNavigable
+  const agendaItemsCount = useMemo(() => {
```
nickoferrall, line 29: Fetching and filtering the `agendaItems` here and a few children below doesn't feel great but as `agendaItems` can live in `teams` or the `meeting`, I think it's necessary

nickoferrall:
Hey @tianrunhe, mind reviewing this one when you get the chance?

tianrunhe (approved):
LGTM from functional perspective. Didn't do code review as I'm lack of front end experience.

tianrunhe:
I tested on my local and it looks slick!

Unfortunately I might not be the best person for code review: I have little front end experience before...

Sorry about the long wait!!!

nickoferrall:
Thanks for checking it out! And all good - you've been on holiday so thanks for quickly reviewing it on your return

mattkrick:
This works fantastically well, great work!
Reading through the code, the additional complexity cost for the accordion is looking a little high. I see you're having to pre-compute a lot of heights that the DOM would be doing for us & some of the parent elements & peer elements like the agenda item list are having to figure out their height even though we're not animating them (or did I miss an agenda item list animation? let me know!).

Big picture, we want each component to be as self-contained as possible & let the DOM do the calculations whenever it can. Is there a way we could build this to reduce complexity? I took a quick crack at it in #5143 to make sure I'm not sending you on an impossible mission. I _think_ that PR fulfills the same AC. Could you take a look at it & see if we could make this a little simpler?

nickoferrall:
Thanks for the review and creating that draft PR! 

I'm calculating the height of overall sidebar & each phase so that when the agendaItems phase has an overflow and the checkin/updates phase is active, the checkin/updates phase shows as much of the child as possible and the agendaItems shrink (I should prob have added comments explaining that). 

Here's a gif of the draft PR when there are many agenda items:

sidebar-overflow

If I add the new `useAnimatedPhaseListChildren` hook to the agendaItems phase so that the height is the scroll height, we also run into problems:

with-hook

I'm currently seeing if I can find a way of accounting for the agendaItems overflow without measuring the sidebar & each phase. Will keep you posted!

PR #5093, opened 2021-07-01
Create unique email for deleted user

### LINKED ISSUE #5057: Consolidate user records in PG
## Issue - Enhancement

We have finally achieved equality between RethinkDB & Postgres for `User` table. We just need to enforce the unique email constraint in PG before we can read from it (#4867 ). Today there are 15 emails in PG that have more than 1 records associate with it.

### Acceptance Criteria
- [ ] For each of the email, determine the unique user record, consolidate the data if necessary
- [ ] Remove the duplicate user records from RethinkDB
- [ ] Remove data from other tables in RethinkDB for the duplicate users (see #5035 )
- [ ] Remove the duplicate user records from PG
- [ ] Add unique constraint back to User email field
- [ ] Run equality checker again to verify two DBs are in-sync after above operations

**Estimated effort:** 12 points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

tianrunhe:
Fixes #5057 

1. When an user is deleted, his/her email field gets changed to `DELETED:${userId}:${Date.now()}`.
2. Migration script for PG to update records with `"email" = 'DELETED'` to match with the format above
3. Migration script for PG to impose unique email constrain on User table


tiffanyhan:
i'm curious, what happened to just hard deleting the users?

packages/server/postgres/migrations/1625124096757_addUserEmailUniqueConstraint.ts
```diff
@@ -0,0 +1,18 @@
+import {MigrationBuilder, ColumnDefinitions} from 'node-pg-migrate'
+
+export const shorthands: ColumnDefinitions | undefined = undefined
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.sql(`
+     ALTER TABLE "User"
+       ADD CONSTRAINT "User_email_key"
+       UNIQUE ("email");
```
tiffanyhan, line 9: In postgres, adding a unique constraint is one of many operations that will lock the entire table which is no good. Can you try using the [following workaround](https://leopard.in.ua/2016/09/20/safe-and-unsafe-operations-postgresql#.YN4S-hNKi-8) to get around it?

> As workaround, you can add column, add unique index concurrently, and then add the constraint onto the table:

```
CREATE UNIQUE INDEX CONCURRENTLY token_is_unique ON large_table(token);
ALTER TABLE large_table ADD CONSTRAINT token UNIQUE USING INDEX token_is_unique;
```

packages/server/postgres/migrations/1625120349057_changeDeletedUserEmail.ts
```diff
@@ -0,0 +1,19 @@
...
+   `)
+}
+
+export async function down(pgm: MigrationBuilder): Promise<void> {
+  pgm.sql(`
+     UPDATE "User"
+       SET "email" = 'DELETED'
+       WHERE "email" LIKE 'DELETED:%';
+   `)
+}
```
tiffanyhan, line 19: can you put these as postdeploy migrations instead of predeploy migrations? If we do predeploy, we (1) backfill existing records (2) deploy code with updated write. Whereas with postdeploy, we (1) deploy code with updated write (2) backfill existing records. i think the latter is safer bc there may be a delay btwn predeploy migrations and deployed code running. with dokku, there's a brief period where both old and new container is running before traffic is switched to new.

tiffanyhan (changes requested):

tianrunhe:
> i'm curious, what happened to just hard deleting the users?

I'm still doing that. This change should be deployed after I clean up the DB so that no two users share the same email.

tianrunhe:
> i'm curious, what happened to just hard deleting the users?

I'm still doing that. This change should be deployed after I clean up the DB so that no two users share the same email.

packages/server/postgres/migrations/1625120349057_changeDeletedUserEmail.ts
tianrunhe: But this has to run before `addUserEmailUniqueConstraint`, right?

tiffanyhan: yes! i meant to put both migration in the postdeploy folder

packages/server/postgres/migrations/1625124096757_addUserEmailUniqueConstraint.ts
```diff
@@ -0,0 +1,20 @@
+import {MigrationBuilder, ColumnDefinitions} from 'node-pg-migrate'
+
+export const shorthands: ColumnDefinitions | undefined = undefined
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.createIndex('User', ['email'], {
+    name: 'User_email_index',
+    unique: true
+  })
```
tiffanyhan, line 9: I wasn't sure if this pgm helper function creates indexes concurrently by default, and it looks like it doesn't. But you can pass it in as an option! can you add that? doing it will prevent the table from locking while the index is being created.

tiffanyhan (changes requested):

tiffanyhan (approved):
thanks for making the changes, lgtm!

packages/server/postgres/migrations/postDeploy/1625124096757_addUserEmailUniqueConstraint.ts
```diff
@@ -0,0 +1,21 @@
...
+  })
+  pgm.sql(`
+      ALTER TABLE "User" ADD CONSTRAINT "User_email_key" UNIQUE USING INDEX "User_email_index";
+   `)
+}
+
+export async function down(pgm: MigrationBuilder): Promise<void> {
+  pgm.sql(`
+     ALTER TABLE "User"
+       DROP CONSTRAINT "User_email_key";
```
mattkrick, line 19: drop index too?

packages/server/postgres/migrations/postDeploy/1625124096757_addUserEmailUniqueConstraint.ts
```diff
@@ -0,0 +1,21 @@
+import {MigrationBuilder, ColumnDefinitions} from 'node-pg-migrate'
+
+export const shorthands: ColumnDefinitions | undefined = undefined
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.createIndex('User', ['email'], {
+    name: 'User_email_index',
+    unique: true,
+    concurrently: true
```
mattkrick, line 9: concurrent indexes scare me a bit in the off chance they go invalid (e.g. a dupe email sneaks in while the index is building)

just outta curiosity did you happen to benchmark the build time of concurrent vs. locking for the User table? it'd be neat to see the difference for ~150k rows

tiffanyhan: yes i was thinking it would be ideal if we could run the migration in an off-peak time. in that scenario, it seems the recommended course of action is to delete the invalid index and try to build the index concurrently again.

i really like the benchmarking idea. i'm biased towards creating the index concurrently, so we build the knowledge in case we have to do it again in the future with more data.

tianrunhe: 1. I tested that with `concurrently: true` the migration would actually fail and the error reported was `CREATE INDEX CONCURRENTLY cannot run inside a transaction block`. I couldn't find a way around it.
2. I removed the `concurrently: true` option and it run just fine.
3. I did a performance test by creating a test table with 150K rows that looks like this:

Screen Shot 2021-07-08 at 4 42 46 PM

and then I run a similar migration script:
```javascript
  pgm.createIndex('Test', 'email', {
    name: 'Test_email_index',
    unique: true
  })
  pgm.createConstraint('Test', 'Test_email_key', 'UNIQUE USING INDEX "Test_email_index"')
```
and it didn't take too long:
```
> time yarn postmigrate 
yarn run v1.22.5
$ node scripts/postDeployMigrate.js
> Migrating files:
> - 1625124096757_addUserEmailUniqueConstraint
> - 1625733417640_testCreateIndexNonCurrently
### MIGRATION 1625124096757_addUserEmailUniqueConstraint (UP) ###
### MIGRATION 1625733417640_testCreateIndexNonCurrently (UP) ###
✨  Done in 1.88s.
yarn postmigrate  1.19s user 0.28s system 70% cpu 2.067 total
```

Let me know what you guys think!

mattkrick: careful! call time !== index creation time.
the index is still building after the call is issued.
 
we'd have to query the index like
```sql
SELECT 
  now()::TIME(0), 
  a.query, 
  p.phase, 
  p.blocks_total, 
  p.blocks_done, 
  p.tuples_total, 
  p.tuples_done
FROM pg_stat_progress_create_index p 
JOIN pg_stat_activity a ON p.pid = a.pid;
```
see: https://www.depesz.com/2019/04/18/waiting-for-postgresql-12-report-progress-of-create-index-operations/

that said, i can't imagine it's more than a minute or 2, so i'm fine with moving forward.

for future reference: the node-pg driver wraps every call inside a transaction by default, so if we do want to build a concurrent index we'd just have to use the `client` strategy and manage the connection ourselves.
if we did that, we'd still have to check for an invalid index because we're deploying this same code to more than just the cloud offering, so we've gotta be sure it runs correctly on private instances, too.

```sql
SELECT * FROM pg_class, pg_index WHERE pg_index.indisvalid = false AND pg_index.indexrelid = pg_class.oid;
```

mattkrick:
LGTM! what are your thoughts on the pgm helpers vs. raw sql? any preference?

PR #5095, opened 2021-07-01
Jira Task sync

mattkrick:
In the database, a Task has an `integration` object that is a subclass of `TaskIntegration`
That type resolves to JiraIssue | GitHub issue depending on the integraiton.service 
The task also has an `integrationHash` which is an indexed field so we can guarantee a single source of truth (multiple parabol tasks on the same team cannot point to the same jira issue)

Burndown:
- [x] ~it feels wrong to have the Task.integration be 1 shape & resolve to a graphql object that is completely different. instead of Task -> JiraIssue i'll change it to Task -> TaskIntegrationJira -> JiraIssue.~ ehhhh, but then again it feels right
- [x] write the same thing for github tasks so we don't break existing functionality
- [x] Nullable task checks for GitHubIssue
- [ ] make helper function to break draft-js content into summary & description
- [x] gql Task.integration works for github
- [ ] 
TEST
- [x] can push a task to jira from the kanban
- [x] can edit a task in jira & then refresh parabol & the content is updated
- [x] can run a poker meeting with jira stories (regression test)
- [x] can run the demo & push a a task to jira


BartoszJarocki:
@mattkrick in my PR related to moving `AtlassianAuth` I've changed some loaders which were extracted here as separate ones: https://github.com/ParabolInc/parabol/pull/5085/files#diff-2e10e056a23bb19b15909712a7dc116686fb39afa12b40009d27a2da67f4dbda so either this one or my PR will have to be updated, just heads up

mattkrick:
LGTM!

BartoszJarocki:
@mattkrick you picked up all the changes 👏   sorry for making it so hard to see what was the change, what was simply the merge, I should have done it in separate commits 🙏 . 

There's one thing about adding missing Jira field functionality. You created a nice enum to make it easier to check the error on the client/server but the client shows the `Fix it for me` dialog according to the error thrown by [updateStoryPoints](https://github.com/ParabolInc/parabol/pull/5095/files#diff-404948a861825f12e7e2bf83ef9e11ecd73ea510c140c77cc4e86a0d73dfaea9R696) method. So [this part](https://github.com/ParabolInc/parabol/pull/5095/files#diff-72a0f2d404bd665fa791eeb9fcd731aac60fe129689ff6fbb2fbbc5584305dc5R131) won't work currently as an error in case of the missing field will be `Update failed! In Jira, add the field "${fieldName}" to the Issue screen.` 

The solution is:
1. Change `SprintPokerDefaults.JIRA_FIELD_UPDATE_ERROR` to be `Update failed! In Jira, add the field ${fieldName} to the Issue screen.`
2. Use `Couldn’t fix the missing field! In Jira, use "Find my field" to determine the error` as a return value of `addMissingJiraField` when _we_ fail to fix it for the user so we can show a human-readable error. 

I'll submit a small PR.

mattkrick:
Feels like an easy fix, I'll go ahead & fix it up here

mattkrick:
alrighty i think this beast is finally ready for master.
the only thing this is missing is images from jira issues, but we'll get to that this sprint or next, so i think a couple broken jira images is safe

PR #5108, opened 2021-07-07
Add a new migration script to sort jiraDimensionFields for both RethinkDB & PG

tianrunhe:
Migration script to:

* Sort `jiraDimensionFields` array field in RethinkDB
* Sort `jiraDimensionFields` array field in PG

@mattkrick I got the error: `Module '"fast-json-stable-stringify"' can only be default-imported using the 'esModuleInterop' flag` if I don't enable the `esModuleInterop` option in tsconfig. Not sure what's a better solution.

### DIFF
--- a/packages/server/tsconfig.json
+++ b/packages/server/tsconfig.json
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
+    "esModuleInterop": true,
     "baseUrl": "../",
     "paths": {
       "server/*": [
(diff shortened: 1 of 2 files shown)


packages/server/tsconfig.json
```diff
@@ -1,6 +1,7 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
+    "esModuleInterop": true,
```
mattkrick, line 4: can remove? this is in the tsconfig.base

tianrunhe: If I remove this, it doesn't work. It looks like the option in `tsconfig.base.json` is not in effect in the sub folder?🤯

mattkrick: can you describe how it doesn't work?

mattkrick: ahhh it's the stringifier.
bigger mystery is why the tsconfig isn't honoring the extends, but we can deal with that later

PR #5114, opened 2021-07-08
Task sync GitHub

mattkrick:
The jira PR is getting too big.
Finishing up the work in a GH PR.
Will merge both to master when ready

Burndown
- [x] delegate Task.integration to the github api automatically
- [x] before forwarding to GH API, remove non-GH fragments
- [x] before forwarding to GH API, add in extra variables
- [x] ~build GitHubCreateIssueSuccess to return _xGitHubIssue~ this is only used in creating tasks inside the GH poker scope phase. we'll do this later
- [x] replace our GitHubIssue with GH's _xGitHubIssue (the GH issue type is completely owned by GitHub)  
- [x] replace GitHubServerManager with our nested github endpoint manager
- [x] optimistic UI for creating GH tasks
- [x] break out GH fragments used in task & reference them in the mutation
- [x] create helper to break draftJS content into {title, body}
- [x] put maxHeight on task cards
- [x] Working demo


mattkrick:
to put the github schema in our schema, we prefix every github object with a prefix, e.g. Issue ->_xGitHubIssue 
githubRequest supports passing in arbitrary queries that get forwarded directly to github.
Should those queries be prefixed or not?
Their nested content will be prefixed because it's coming from our system, so it'd be weird to have a prefixed piece wrapped by a n unprefixed piece.

Then again, that wrapper is written explicitly for the integration & by not using prefixes it'd make i nice because that whole string could be pasted into the integration's graphql explorer.
a good example is getRepositories.graphql
gonna wait until monday to decide

Issue #5107, opened 2021-07-07
Long meeting names throw off meetings view layout


nickoferrall:
If the meeting name or the org name is relatively long, there isn't enough space between the meeting cards. 

![meetings-view](https://user-images.githubusercontent.com/39854876/124682797-48ec4500-de91-11eb-8bf2-418634cdef42.png)

This is because the meeting cards are positioned absolutely and their top is calculated with a fixed [meeting card height](https://github.com/ParabolInc/parabol/blob/c8d057bc6861eaa9d102e8a53966fb79e55a3272/packages/client/components/MeetingsDash.tsx#L98).

Possible solutions are:

1. Pass a ref to each meeting card and use the client height to calculate the top
2. Use the meeting and org name to calculate the height of the card, e.g. if the meeting name > 22 characters, we know it's 2 lines, 44 characters = 3 lines.
3. Make each meeting card in each row the same height as the tallest card. This will be the easiest to calculate, ensures consistency but means that some cards could have a lot of whitespace. If one card on the row has a meeting name that's 3 lines long, a team name that's 2 lines long and has connected users in the `AvatarList`, there'll be 80px of whitespace

I'm working on solution 1 at the moment.

enriquesanchez:
A few more instances of this I noticed today:

<img width="1258" alt="Screen Shot 2021-07-07 at 11 01 03 a m" src="https://user-images.githubusercontent.com/3276087/124810167-9ebefc80-df26-11eb-929c-5cde92dd025e.png">
<img width="1256" alt="Screen Shot 2021-07-07 at 11 01 29 a m" src="https://user-images.githubusercontent.com/3276087/124810171-a088c000-df26-11eb-84c4-0ea26c69a9ef.png">

enriquesanchez:
Would it be a good idea to use CSS Grid for this view?

mattkrick:
i don't think grids animate too well (https://blog.bitsrc.io/animating-css-grid-rows-and-columns-4b3b0997d06a), which is a huge bummer because if they did this would be a LOT easier

mattkrick:
maybe we could absolutely position all the cards in the corner to measure them, then create a grid layout using their sizes, & then absolutely position them on top of their pseudo elements? my gut tells me there'd be a flicker between measurement & animation, but i haven't tested it

nickoferrall:
[...]

PR #5118, opened 2021-07-08
Fix long meeting names throwing off meetings view

nickoferrall:
PR to resolve issue #5107 

I also noticed that long words can overflow on the meeting card and in the meeting. See screenshots:

long-name-1

long-name-2

### AC:

- [ ] if the meeting name or team name takes up multiple lines or if there's a user in the meeting, the rows below that meeting card adjust their position accordingly
- [ ] long meeting or team names don't overflow on the card or in the meeting

Would be interested to get your thoughts on the design @enriquesanchez


enriquesanchez:
Screen Shot 2021-07-09 at 3 23 15 p m

@nickoferrall The long names issue looks fixed on my end, but I think I am missing a step or something for testing this PR because the layout is still broken

nickoferrall:
@enriquesanchez huh that's strange - I just checked out the branch, ran a `yarn` and `yarn dev` but the layout is working on my end on Chrome & Safari. Mind trying again and, if you get the same result, we could Zoom together to reproduce it?

enriquesanchez:
@nickoferrall Ah, I forgot to run `yarn dev`... again 🤦‍♂️
Sorry about that. Here's a screen grab on Firefox:

Screen Shot 2021-07-12 at 12 19 52 p m

👍 It looks good!

Did you also work on adding the text "No one has joined" ?

nickoferrall:
Ahh that's good to hear! I haven't added the "No one has joined" text yet. I think it's a helpful way of keeping the meeting cards the same height when displaying the avatar list but I also kinda like that active meetings are larger and draw the user's attention. Plus the card heights will be different anyway if there's a long meeting / team name.

Do you have any strong opinions on this one?

enriquesanchez:
>Do you have any strong opinions on this one?

@nickoferrall No strong opinions, you're are right it doesn't fully solve the issue.

enriquesanchez (approved):

mattkrick:
you did it! this PR brings back some feels. many years ago our retrospectives used a masonry grid instead of the columns we have today. i had to use many of the complicated patterns that this PR uses, like the variable length array of refs. eventually, the logic got to be too complex. For this case, I'd ask what happens if a meeting gets renamed to be really long? or short? or everyone leaves the meeting? or 2 meetings get created during the same render cycle? or 1 meeting ends & 1 gets created in the same render cycle?)

we're sacrificing a simple DX for animations, and if that's the price we pay, so be it because animations are awesome. buuuuut, i still gotta wonder if we can make it simpler. for example, https://github.com/ParabolInc/parabol/blob/f092a703bd675aaaa5129d3e155c6d4cb51474e2/packages/client/modules/teamDashboard/components/TeamArchive/TeamArchive.tsx uses react-virtualized which accomplishes the same thing. basically, they give you a `cellCache` and whenever something _could_ change the height, then you call `cellCache.clearAll()` and it recalcs everything. 

i'll spend an hour or 2 tonight to see if we can simplify the pattern. if not, LGTM! truly a great feat of engineering & really well done

mattkrick:
LGTM! If you're ready to merge, merge away!

Issue #5009, opened 2021-06-02
Deprecate title in reflectPrompt


nickoferrall:
[Title](https://github.com/ParabolInc/parabol/blob/03facdd6bf959cc0e8319adc61d078ccbcdb1f6b/packages/server/graphql/types/ReflectPrompt.ts#L47) in `reflectPrompt` isn't being used in the codebase and is typically the same string as the `question`. This is probably an artifact that we never refactored away. 

### AC

- [ ] Deprecate `title` in `reflectPrompt` 

### EE

- 1 hour

PR #5120, opened 2021-07-09
Deprecate prompt title

nickoferrall:
PR to resolve issue #5009 

I know that [we sometimes](https://github.com/ParabolInc/parabol/blob/db45e3b5a4c27e917dcc267d33175bae8a04970b/packages/server/graphql/types/RetroReflectionGroup.ts#L32) keep the field with a deprecation reason. I assume we do this when deprecating a previously important field and we want to be careful in case we break something? The prompt title isn't being used anywhere so I assume it's safe to remove it entirely.

### DIFF
--- a/packages/client/types/graphql.ts
+++ b/packages/client/types/graphql.ts
@@ -45686,11 +45686,6 @@ export interface IReflectPrompt {
   template: IReflectTemplate;
 
   /**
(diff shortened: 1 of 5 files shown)


mattkrick:
correct! these are the wonderful days where are API is only used by us :sweat_smile:
LGTM!

PR #5127, opened 2021-07-12
Implement LogRocket watchlist

nickoferrall:
I've added a `isWatched` field to the User object and a private mutation that lets us toggle the boolean. With `isWatched`, we can create a watch list of the users where we want to record all of their sessions in LogRocket.

### DIFF
--- a/packages/client/components/AnalyticsPage.tsx
+++ b/packages/client/components/AnalyticsPage.tsx
@@ -74,7 +74,7 @@ const AnalyticsPage = () => {
     const logRocketId = window.__ACTION__.logRocket
     const errorProneAt = window.localStorage.getItem(LocalStorageKey.ERROR_PRONE_AT)
     const expiredErrorProne =
-      errorProneAt && new Date(parseInt(errorProneAt)) < new Date(Date.now() - ms('30d'))
+      errorProneAt && new Date(parseInt(errorProneAt)) < new Date(Date.now() - ms('14d'))
     if (expiredErrorProne) {
       window.localStorage.deleteItem(LocalStorageKey.ERROR_PRONE_AT)
     } else if (logRocketId && errorProneAt) {
--- a/packages/client/components/AnalyticsPage.tsx
+++ b/packages/client/components/AnalyticsPage.tsx
@@ -76,7 +76,7 @@ const AnalyticsPage = () => {
     const logRocketId = window.__ACTION__.logRocket
     const errorProneAt = window.localStorage.getItem(LocalStorageKey.ERROR_PRONE_AT)
     const expiredErrorProne =
-      errorProneAt && new Date(parseInt(errorProneAt)) < new Date(Date.now() - ms('14d'))
+      errorProneAt && new Date(parseInt(errorProneAt)) < new Date(Date.now() - ms('30d'))
     const email = window.localStorage.getItem(LocalStorageKey.EMAIL)
     const res = await fetchQuery<AnalyticsPageQuery>(atmosphere, query, {})
     const isWatched = res?.viewer?.isWatched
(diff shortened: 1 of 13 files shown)


packages/client/components/AnalyticsPage.tsx
```diff
@@ -70,15 +71,18 @@ const AnalyticsPage = () => {
...
+  const initLogRocket = async () => {
     const logRocketId = window.__ACTION__.logRocket
     const errorProneAt = window.localStorage.getItem(LocalStorageKey.ERROR_PRONE_AT)
     const expiredErrorProne =
-      errorProneAt && new Date(parseInt(errorProneAt)) < new Date(Date.now() - ms('30d'))
-    if (expiredErrorProne) {
+      errorProneAt && new Date(parseInt(errorProneAt)) < new Date(Date.now() - ms('14d'))
+    const email = window.localStorage.getItem(LocalStorageKey.EMAIL)
+    const res = await fetchQuery<AnalyticsPageQuery>(atmosphere, query, {})
+    const isWatched = res?.viewer?.isWatched
```
nickoferrall, line 82: I noticed that in dev, after updating `isWatched` in the db, it seems to require a `yarn dev` before updating on the client. I updated the caching to `force: true` which didn't do the trick. Perhaps this is expected?

mattkrick: ahhh, _just_ for the User table we do some light caching in redis. that'll go away once we go full postgres

packages/server/postgres/migrations/1626296172790_addIsWatched.ts
```diff
@@ -0,0 +1,17 @@
+import {MigrationBuilder, ColumnDefinitions} from 'node-pg-migrate'
+
+export const shorthands: ColumnDefinitions | undefined = undefined
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  await pgm.db.query(`
+  ALTER TABLE "User"
+  ADD COLUMN IF NOT EXISTS "isWatched" BOOLEAN DEFAULT FALSE
```
mattkrick, line 8: let's add a NOT NULL  to this guy

packages/server/graphql/types/User.ts
```diff
@@ -132,6 +132,11 @@ const User = new GraphQLObjectType<any, GQLContext>({
       description: 'true if the user was removed from parabol, else false',
       resolve: ({isRemoved}) => !!isRemoved
     },
+    isWatched: {
+      type: GraphQLNonNull(GraphQLBoolean),
+      description: 'true if all user sessions are being recorded in LogRocket, else false',
+      resolve: ({isWatched}) => !!isWatched
```
mattkrick, line 138: can remove resolve if not null is set in pg

packages/server/graphql/intranetSchema/mutations/updateWatchlist.ts
```diff
@@ -0,0 +1,49 @@
+import {GraphQLBoolean, GraphQLID, GraphQLNonNull} from 'graphql'
+import getRethink from '../../../database/rethinkDriver'
+import updateUser from '../../../postgres/queries/updateUser'
+import {requireSU} from '../../../utils/authorization'
+
+const updateWatchlist = {
+  type: GraphQLBoolean,
+  description: 'add a user to the LogRocket watchlist so that their sessions are recorded',
+  args: {
+    userId: {
```
mattkrick, line 10: maybe accept emails instead of userId? we might want to log entire domains, too (in the case of symph).

API might end up looking like: https://github.com/ParabolInc/parabol/blob/f377b7f9ba79fa11174ad0a6ab7037e6c6a1b3d6/packages/server/graphql/mutations/addFeatureFlag.ts

mattkrick:
just a couple suggestions, other than that merge when ready!

packages/server/graphql/intranetSchema/mutations/updateWatchlist.ts
nickoferrall: Thanks, that makes sense! I'm popping out to dinner now so will merge this so that it's in the ship. Then I can refactor this tomorrow so we can use domains and emails in the future

Issue #5117, opened 2021-07-08
Timeline & team off-centre on large screen


nickoferrall:
When using Chrome/Brave on a 27" monitor, the page isn't correctly centred on the Timeline or Teams page. This doesn't happen on smaller screen sizes / devices.  

This is a screenshot from the current prod so the bug must have been introduced in v6.19 or earlier:

![padding-off](https://user-images.githubusercontent.com/39854876/124964201-59fa9a80-dfe6-11eb-89d8-1d886d1de6dd.png)

PR #5129, opened 2021-07-13
Fix dashboard padding

nickoferrall:
PR to resolve issue: #5117 

Issue was caused by removing `position: fixed` [here](https://github.com/ParabolInc/parabol/blob/c8d057bc6861eaa9d102e8a53966fb79e55a3272/packages/client/components/Dashboard/DashSidebar.tsx#L23) in #4975. 

However, it looks like the [DashMain](https://github.com/ParabolInc/parabol/blob/bdbb60c65d52f48698268aed46f9e095947ee3f3/packages/client/components/Dashboard.tsx#L114) div previously included the nav sidebar on large screens and we adjusted the content of `DashMain` with padding. 

Instead, by removing `position: fixed` and removing the extra padding, we can correctly centre the dashboards and the `DashMain` div won't include the sidebar.

### AC:

- [ ] dashboards look beautiful on all screensizes

### DIFF
--- a/packages/client/components/Dashboard/DashSidebar.tsx
+++ b/packages/client/components/Dashboard/DashSidebar.tsx
@@ -2,9 +2,8 @@ import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
 import React from 'react'
 import {createFragmentContainer} from 'react-relay'
-import makeMinWidthMediaQuery from '~/utils/makeMinWidthMediaQuery'
 import {PALETTE} from '../../styles/paletteV3'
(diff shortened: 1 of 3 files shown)


BartoszJarocki (approved):
@nickoferrall great work! works nicely on all my devices. 

Not related to this issue but I've noticed the shadow is cut off on the timeline view card corners (very very minor detail but still 😅). Should I create an issue for it?
Screenshot 2021-07-15 at 15 47 50

nickoferrall:
Great, thanks for the review!

That's a keen eye. An issue sounds good 👍

enriquesanchez (approved):
Fix works for me 👍 (tested on FF and Chrome).

Issue #5134, opened 2021-07-14
Add scope handling to `addAtlassianAuth` mutation


BartoszJarocki:
## Issue - Enhancement

In a sprint poker meeting, when a user tries to update the estimate using a field that isn't configured for the estimated issue in Jira, we offer to fix that automatically. If the user wants us to fix it, we ask him for additional scope `jira:manage-project`. 
After migrating `AtlassianAuth` to PG we added an additional column `scope` which is populated with default scope. After requesting additional scope, the scope should be updated with the correct value so we can avoid asking again.

### Acceptance Criteria (optional)
- [ ] `addAtlassianAuth` mutation handles the requested scope correctly instead of always using default value
- [ ] `AddMissingJiraFieldModal` doesn't ask for additional scope if user already provided it 

**Estimated effort:** 16 hours ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

PR #5135, opened 2021-07-14
Fixed atlassian auth migration

BartoszJarocki:
After merging https://github.com/ParabolInc/parabol/pull/4880 `scope` is no longer kept as a string but as an array of strings to allow easier manipulation. This PR fixes the migration which assumed it was a string. 

I've also added an improvement to existing functionality: https://github.com/ParabolInc/parabol/issues/5134


mattkrick:
nice catch!

PR #5136, opened 2021-07-14
Jira task sync update

BartoszJarocki:
This PR updates https://github.com/ParabolInc/parabol/pull/5095 with the newest changes from master and aligns the implementation of `AtlassianManager` to the new way of error handling. 

Along with updating the changes, I had to fix a few issues:
1. When checking if `errorMessages` exist in response, we also have to check [if the array is empty](https://github.com/ParabolInc/parabol/pull/5133/commits/81deca19fd48f9a242b0eb779a39f04205c93ce7#diff-baea95f99f42d5aeb5fb3239b5180131c801a0bf091b24d8093a3a2deb892dc3R82). Jira started returning empty array and broke our error handling flow that way. 
2. I reverted the [`MissingJiraFieldError` change introduced yesterday](https://github.com/ParabolInc/parabol/pull/5133/commits/81deca19fd48f9a242b0eb779a39f04205c93ce7#diff-baea95f99f42d5aeb5fb3239b5180131c801a0bf091b24d8093a3a2deb892dc3R82) - it's used to check if we should display the `Fix it for me` popup. The new error message will be displayed when our attempt to add the field fails.
3. I added additional checks in [put](https://github.com/ParabolInc/parabol/pull/5133/commits/81deca19fd48f9a242b0eb779a39f04205c93ce7#diff-baea95f99f42d5aeb5fb3239b5180131c801a0bf091b24d8093a3a2deb892dc3R82) and added [delete](https://github.com/ParabolInc/parabol/pull/5136/files#diff-404948a861825f12e7e2bf83ef9e11ecd73ea510c140c77cc4e86a0d73dfaea9R359) method


packages/client/utils/AtlassianManager.ts
```diff
@@ -264,7 +305,7 @@ export default abstract class AtlassianManager {
       }
       return new Error(json.message)
     }
-    if ('errorMessages' in json) {
+    if ('errorMessages' in json && json.errorMessages.length > 0) {
```
BartoszJarocki, line 308: This additional checks causes TS errors, I'm still figuring what's the best way of handling that.

BartoszJarocki: fixed by adding explicit type guard

mattkrick:
I may break this a bit, i'm gonna merge master into mine so i can see a cleaner diff

packages/client/utils/AtlassianManager.ts
mattkrick: if errorMessages is empty does that mean it's not an error?

mattkrick:
i can't for the life of me figure out how to get this to diff properly.
any tips or can you point me to the individual changes?

mattkrick:
> I reverted the MissingJiraFieldError change introduced yesterday

I see you reverted it here: https://github.com/ParabolInc/parabol/pull/5136/files#diff-baea95f99f42d5aeb5fb3239b5180131c801a0bf091b24d8093a3a2deb892dc3

but it looks like you updated it here: https://github.com/ParabolInc/parabol/pull/5136/files#diff-8d4dd152e0b27203f1cc141dca3881392bf21adfef55644c7204f2cc19d9694a

I'm gonna turn that into a constEnum so the server & client are always in sync

mattkrick:
Alrighty, I _think_ I picked up all the changes!
Closing in favor of https://github.com/ParabolInc/parabol/pull/5095/commits/38e415ebf3af01c39af15fe01a6adf8b35fc75c7
If you've got the time could you check that commit & see if I missed anything?

packages/client/utils/AtlassianManager.ts
BartoszJarocki: Yes, at least now. I don't know what's wrong with Jira but I wasn't even able to make my functionality work so I debugged and this is how the response looked like
```
{
  errorMessages: [],
  errors: {
    customfield_10030: "Field 'customfield_10030' cannot be set. It is not on the appropriate screen, or unknown."
  }
}
```
🤦

Issue #5099, opened 2021-07-02
summary review not working properly 


drewhousman1010:
A [user ](https://app.hubspot.com/contacts/3888472/contact/21973751) 🔒 writes:

> The "Review a summary" isn't working anymore. Click to the link will be redirected back to the Meeting page, not the retro itself. Refer attached for the details.

https://user-images.githubusercontent.com/80418360/124321837-fb47a380-db43-11eb-8d0e-4167392cd704.mp4

Hubspot ticket:

https://app.hubspot.com/contacts/3888472/ticket/466585504/

mattkrick:
@BartoszJarocki this will be a fun one we can triage for our one-on-one

tianrunhe:
Duplicate issue with #5078. Bug is on empty discussion thread.

BartoszJarocki:
@tianrunhe is right here. Related: https://sentry.io/organizations/parabol/issues/2449818299/?project=107196
One note: I can only reproduce that with Retro meetings.

mattkrick:
ahh that makes sense.
for retro meetings we create a dummy discussion topic as a placeholder & then replace that with the first discussion topic after they advance from vote -> discuss.
if they never get that far, then i assume the meeting is left with a dummy topic that doesn't have certain necessary fields (e.g. a discussionId)

maybe we:
- create a helper that checks if no discussion topics were created & if so 
- the timeline event takes them to the group phase instead of the discuss phase 
- the summary fails gracefully by checking if it's a dummy & if so, treats the discussion topics as an empty array

BartoszJarocki:
Sentry issue: [ACTION-PRODUCTION-80J](https://sentry.io/organizations/parabol/issues/2449818299/?referrer=github_plugin)

PR #5137, opened 2021-07-14
Fixed summary review not working properly

BartoszJarocki:
Related issue: https://github.com/ParabolInc/parabol/issues/5099

I've noticed one more thing though. If the meeting ends before going to the discussion phase, the reflections fetched in the meeting summary will be an empty array. I'm investigating this but if someone has any quick fixes for that, please let me know! For now, I've added a check for empty data in CSV export to prevent it from crashing.

AC:
* Meeting summary works when there's no discussion

### DIFF
--- a/packages/server/graphql/types/RetroDiscussStage.ts
+++ b/packages/server/graphql/types/RetroDiscussStage.ts
@@ -7,8 +7,8 @@ import DiscussionThreadStage, {discussionThreadStageFields} from './DiscussionTh
 import NewMeetingStage, {newMeetingStageFields} from './NewMeetingStage'
 import RetroReflectionGroup from './RetroReflectionGroup'
 
-const createEmptyDiscussion = (id) => ({
-  id,
+const DUMMY_DISCUSSION = {
+  id: 'dummy-discussion-id',
   teamId: '',
   meetingId: '',
   createdAt: '',
@@ -18,7 +18,7 @@ const createEmptyDiscussion = (id) => ({
   thread: {
     edges: []
   }
-})
+}
 
 const RetroDiscussStage = new GraphQLObjectType<any, GQLContext>({
   name: 'RetroDiscussStage',
@@ -31,9 +31,9 @@ const RetroDiscussStage = new GraphQLObjectType<any, GQLContext>({
     discussion: {
       type: GraphQLNonNull(Discussion),
       description: 'The discussion about the stage or a dummy data when there is no disscussion',
-      resolve: async ({discussionId}, _args, {dataLoader}) => {
-        const discussion = await dataLoader.get('discussions').load(discussionId)
-        return discussion ?? createEmptyDiscussion(discussionId)
+      resolve: async ({discussionId, reflectionGroupId}, _args, {dataLoader}) => {
+        const isDummy = reflectionGroupId === ''
+        return isDummy ? DUMMY_DISCUSSION : dataLoader.get('discussions').load(discussionId)
       }
     },
     reflectionGroupId: {
--- a/packages/server/graphql/types/RetroDiscussStage.ts
+++ b/packages/server/graphql/types/RetroDiscussStage.ts
@@ -2,9 +2,23 @@ import {GraphQLFloat, GraphQLID, GraphQLNonNull, GraphQLObjectType} from 'graphq
 import {NewMeetingPhaseTypeEnum} from '../../database/types/GenericMeetingPhase'
 import ReflectionGroup from '../../database/types/ReflectionGroup'
 import {GQLContext} from '../graphql'
+import Discussion from './Discussion'
 import DiscussionThreadStage, {discussionThreadStageFields} from './DiscussionThreadStage'
 import NewMeetingStage, {newMeetingStageFields} from './NewMeetingStage'
 import RetroReflectionGroup from './RetroReflectionGroup'
+
+const createEmptyDiscussion = (id) => ({
+  id,
+  teamId: '',
+  meetingId: '',
+  createdAt: '',
+  discussionTopicId: '',
+  discussionTopicType: '',
+  commentCount: 0,
+  thread: {
+    edges: []
+  }
+})
 
 const RetroDiscussStage = new GraphQLObjectType<any, GQLContext>({
   name: 'RetroDiscussStage',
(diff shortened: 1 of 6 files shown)


packages/server/graphql/types/RetroDiscussStage.ts
```diff
@@ -2,10 +2,24 @@ import {GraphQLFloat, GraphQLID, GraphQLNonNull, GraphQLObjectType} from 'graphq
 import {NewMeetingPhaseTypeEnum} from '../../database/types/GenericMeetingPhase'
 import ReflectionGroup from '../../database/types/ReflectionGroup'
 import {GQLContext} from '../graphql'
+import Discussion from './Discussion'
 import DiscussionThreadStage, {discussionThreadStageFields} from './DiscussionThreadStage'
 import NewMeetingStage, {newMeetingStageFields} from './NewMeetingStage'
 import RetroReflectionGroup from './RetroReflectionGroup'
 
+const createEmptyDiscussion = (id) => ({
```
mattkrick, line 10: the ID being `''` feels a little weird. could we just hardcode in a value like `id: 'dummy'`?

packages/server/graphql/types/RetroDiscussStage.ts
```diff
@@ -14,6 +28,14 @@ const RetroDiscussStage = new GraphQLObjectType<any, GQLContext>({
   fields: () => ({
     ...newMeetingStageFields(),
     ...discussionThreadStageFields(),
+    discussion: {
+      type: GraphQLNonNull(Discussion),
+      description: 'The discussion about the stage or a dummy data when there is no disscussion',
+      resolve: async ({discussionId}, _args, {dataLoader}) => {
+        const discussion = await dataLoader.get('discussions').load(discussionId)
```
mattkrick, line 35: the dataloader is gonna cause a hit to the database, which is avoidable if we know the discussionId is `''`.
can we do something like `return discussionId ? dataLoader.get(...) : DUMMY_DISCUSSION`

mattkrick (changes requested):
looks great! just a couple little changes

packages/server/graphql/types/RetroDiscussStage.ts
BartoszJarocki: that was my initial idea but... that's how 100% empty meeting looks like
```
{
  "id": "7",
  "type": "complete",
  "payload": {
    "data": {
      "viewer": {
        "newMeeting": {
          "__typename": "RetrospectiveMeeting",
          "id": "5d0pVmo7xC",
          "createdAt": "2021-07-14T14:42:28.197Z",
          "name": "Retro #1",
          "team": {
            "name": "bartosz’s Team",
            "id": "5coX2pqNRD"
          },
          "meetingType": "retrospective",
          "reflectionGroups": [],
          "meetingMembers": [
            {
              "tasks": [],
              "id": "local|5coX2kBCEw::5d0pVmo7xC",
              "__typename": "RetrospectiveMeetingMember",
              "user": {
                "rasterPicture": "https://action-files.parabol.co/static/avatars/ba.png",
                "preferredName": "bartosz",
                "id": "local|5coX2kBCEw"
              }
            }
          ],
          "phases": [
            {
              "__typename": "CheckInPhase",
              "phaseType": "checkin",
              "id": "5d0pVoEUA9"
            },
            {
              "__typename": "ReflectPhase",
              "phaseType": "reflect",
              "id": "5d0pVqmvmw"
            },
            {
              "__typename": "GenericMeetingPhase",
              "phaseType": "group",
              "id": "5d0pVqmvmz"
            },
            {
              "__typename": "GenericMeetingPhase",
              "phaseType": "vote",
              "id": "5d0pVqmvmB"
            },
            {
              "__typename": "DiscussPhase",
              "phaseType": "discuss",
              "stages": [
                {
                  "id": "5d0pVqmvmF",
                  "reflectionGroupId": "",
                  "reflectionGroup": {
                    "title": "Unknown",
                    "voteCount": 0,
                    "reflections": [],
                    "id": "5d0pVmo7xC:dummyGroup"
                  },
                  "discussion": {
                    "commentCount": 0,
                    "id": "5d0pVqmvmG"
                  }
                }
              ],
              "id": "5d0pVqmvmD"
            }
          ]
        },
        "id": "local|5coX2kBCEw"
      }
    }
  }
}
```
I'll dig into this and see if where how discussionId is created.

BartoszJarocki: This `discussionId` is somehow taken from rethinkdb, but honestly, I can't find which resolver fetches that 🙈

mattkrick: Ohhhh that's right!
Ugh, placeholders are just the worst.
Basically, we create that dummy with a UID.
Then, when the real data comes in, we swap out the dummy date for the new data but keep the UID the same.
You can see the stage getting created here: https://github.com/ParabolInc/parabol/blob/ba2fdeadc7da023f188e2765096810ebdaf59ccd/packages/server/database/types/DiscussPhase.ts
and if you check out the DiscussStage constructor, you'll see it just creates a new UID.

So how do we now it's dummy data?

```ts
const isDummy = stages[0].length === 1 && stages[0].reflectionGroupId === ''
return isDummy ? DUMMY_DATA : dataLoader.get(...)`
```

BartoszJarocki: Thank you, Matt! In this particular resolver, we can use `reflectionGroupId` directly so it's even easier. 
```
      resolve: async ({discussionId, reflectionGroupId}, _args, {dataLoader}) => {
        const isDummy = reflectionGroupId === ''
```
Works like a charm!

BartoszJarocki:
@mattkrick this is ready and it's showing the meeting summary correctly. I do have a bit different question though.
This is how the summary looks like when there are some reflections added but the meeting was finished _before_ discuss phase:
Screenshot 2021-07-16 at 15 05 09

As a result, export to CSV is not working, because it's reading the reflections from
```
            ... on RetroDiscussStage {
              reflectionGroup {
                title
                voteCount
                reflections {
                  content
                  createdAt
                  prompt {
                    question
                  }
                }
              }
            }
```
which is empty at this point. 

I think we still should show the reflections in the summary (quicks stats show the correct number so I think it's even more confusing now).

`QuickStats` are reading the reflections that way
```
      ... on RetrospectiveMeeting {
        reflectionGroups(sortBy: voteCount) {
          voteCount
          reflections {
            id
          }
        }
        meetingMembers {
          tasks {
            id
          }
        }
      }
```
And I think that's the correct way. Also, changing the implementation (or simply adding a new section) to use this data, would allow CSV export to work and the summary would make more sense. Any thoughts? 🙏

mattkrick:
that's good insight. i think that's a regression that i introduced when refactoring the discussions https://github.com/ParabolInc/parabol/blob/91f51f6318d3a02efd588072e15cf3408dd92360/packages/client/modules/email/components/SummaryEmail/ExportToCSV.tsx

we can either revert back to `meeting.reflectionGroups` or when a meeting is terminated we create the discussion stages for them. I'm OK with either one. Can you create a new issue & reference this PR from it so we can prioritize it on the next sprint?

BartoszJarocki:
@mattkrick sure! https://github.com/ParabolInc/parabol/issues/5191

PR #5138, opened 2021-07-15
add FLIP animations to meeting cards

mattkrick:
@nickoferrall 

I noticed that changing the name didn't update the grid, so I figured it was probably time to simplify the strategy.
I remember we chatted about FLIP animations a few months ago. This seemed like a good time to use them!
I tried to comment as much as I could.
The overall strategy is:
- Let flexbox do it's thing. flex-wrap will handle people joining/leaving the meeting & meeting name changes because it's not position absolute.
- When a meeting goes away, turn that meeting card into a position absolute & keep it right where it is, but fade it out over 300ms. use a FLIP animation for the rest of the cards.
- When a new meeting occurs, use a FLIP animation to move the cards to their new home

if you like it feel free to merge to your PR!


packages/client/hooks/useAnimatedMeetingCard.ts
```diff
@@ -0,0 +1,46 @@
...
+    // offset measurements are taken after a DOM paint
+    const {offsetTop, offsetLeft, style} = el
+    if (idx !== displayIdx) {
+      // interpolate last - first in FLIP
+      const diffLeft = rect.left - offsetLeft
+      const diffTop = rect.top - offsetTop
+      const duration = 350
+      // move it back where it was
+      style.transform = `translate(${diffLeft}px,${diffTop}px)`
+      requestDoubleAnimationFrame(() => {
```
nickoferrall, line 25: `requestAnimationFrame` tells the browser that we want to perform an animation before the next repaint. `requestDoubleAnimationFrame` is required here because it calls any animations that might block the main thread which would lead to jankiness?

nickoferrall (approved):
This is wonderfully impressive - there are far fewer calculations we need to maintain now 🎉  It's great to see FLIP in action!

packages/client/hooks/useAnimatedMeetingCard.ts
mattkrick: it's _almost_ the same as a setTimeout of 0. when you assign a value to `style.transform` it queues up that change to be done on the next frame, but sometimes it can happen in the same frame. then to reset the value we want to wait another frame so it doesn't undo the change before it has started.

PR #5140, opened 2021-07-15
Fixed handling update jira field errors

BartoszJarocki:
Related conversation: https://github.com/ParabolInc/parabol/pull/5095#issuecomment-880406015

### DIFF
--- a/packages/client/types/errors.ts
+++ b/packages/client/types/errors.ts
@@ -1,0 +1,6 @@
+export const SprintPokerErrors = {
+  fixMissingFieldError:
+    'Couldn’t fix the missing field! In Jira, use "Find my field" to determine the error',
+  updateFieldError: (fieldName: string) =>
+    `Update failed! In Jira, add the field "${fieldName}" to the Issue screen.`
+}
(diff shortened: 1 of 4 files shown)


packages/client/types/errors.ts
```diff
@@ -0,0 +1,6 @@
+export const SprintPokerErrors = {
+  fixMissingFieldError:
+    'Couldn’t fix the missing field! In Jira, use "Find my field" to determine the error',
+  updateFieldError: (fieldName: string) =>
+    `Update failed! In Jira, add the field "${fieldName}" to the Issue screen.`
+}
```
BartoszJarocki, line 6: I don't particularly like this solution, but I can't think of anything better 🤔

mattkrick: ahhh good catch!

because jira's "context" could also be a cause, I think the following error message is no longer true: `Update failed! In Jira, add the field "${fieldName}" to the Issue screen.` because it could exist on the issue screen & the context could be wrong.

since both error messages have the same root cause (misconfigured jira) and same solution (goto find my field & see the error), let's use a single error message for them both.

Maybe something like: `Jira rejected the field update. In Jira, use "Find my field" to determine the cause`

what do you think?

BartoszJarocki: Sounds great! I'll update the places with a single error message and it'll be 👌

mattkrick:
fixed up in task-sync to keep the merge train moving. thanks for finding this!

PR #5143, opened 2021-07-15
sidebar animations - rough draft

mattkrick:
tried to simplify the animations in #5086.
This just handles the check-in meeting, need to add the hook to other meetings, too.


nickoferrall:
For the reasons mentioned in the [issue here](https://github.com/ParabolInc/parabol/pull/5086#issuecomment-880989425), it looks like we do need to calculate the `maxHeight` so that the checkin/updates phase aren't squished when there are many agenda items. 

I've pushed an update using the hook in this PR and some of the `maxHeight` logic in the other PR. The main difference in this PR is that we're not calculating the height of the agendaItems/discuss/estimate phases which should help with performance & make it more maintainable.

mattkrick:
good find!
all you gotta do is animate the min-height.

```ts
const MeetingSidebarPhaseItemChild = styled('div')<{height: string | number}>(({height}) => ({
  display: 'flex',
  flexDirection: 'column',
  transition: `height 300ms ${BezierCurve.DECELERATE}, min-height 300ms ${BezierCurve.DECELERATE}`,
  height,
  minHeight: height,
  overflow: 'hidden'
}))
```

POC

https://user-images.githubusercontent.com/5514175/126836010-d1df9e47-a0f3-405a-ad23-426dd9c1ed27.mp4

nickoferrall:
Ah well that sounds great! 

I checked out your previous commit, copied & pasted the `MeetingSidebarPhaseItemChild` component that you've shared but I get the following:

sidebar-minHeight

In your video, it looks like there's too much space in the check-in phase and the Team Agenda is moving when switching between the check-in & updates phases. 

I can keep playing around with it to see if I can iron out the final small issues while using `minHeight`. Apologies for all the back and forth on this one!

nickoferrall:
I've played around with this a bit more but I think we do need to calculate the `maxHeight` due to the flexbox bug [here](https://github.com/ParabolInc/parabol/blob/f092a703bd675aaaa5129d3e155c6d4cb51474e2/packages/client/components/NewMeetingSidebarPhaseListItem.tsx#L17).

I was able to fix the behaviour I shared in my last gif by removing `minHeight` from `MeetingSidebarPhaseItemChild` and changing 40 to `fit-content` [here](https://github.com/ParabolInc/parabol/blob/f092a703bd675aaaa5129d3e155c6d4cb51474e2/packages/client/components/NewMeetingSidebarPhaseListItem.tsx#L17).

But the problem is if there's not enough space for the check-in / updates phase, it overflows as we're setting a height of 360px when there's only 300px available, for example:

flexbox-bug

In the solution I pushed to this branch, I calculated the `maxChildHeight` [here](packages/client/hooks/useMaxChildHeight.ts) and passed it to the sidebar child item. An alternate approach could be passing the `maxHeight` to the navItem instead. 

If it'd be more convenient for you to tackle this sync, I'd be happy to Zoom whenever you're free to get this one over the line

mattkrick:
closing in favor of #5240 (but there are a few commits here that we should probably cherrypick)

Issue #5100, opened 2021-07-02
Proxy jira images through parabol


mattkrick:
Discussion in #5098
Jira gives us image URLs that we can't put in img tags because they need an Authorization header. 

To get around this, we'll replace their URLs with a parabol URL that will retrieve an image stored in redis.
For ultimate security, we'd give the client a UUID that expires in an hour & then the server trades that UUID for a filename. then, to keep from saving the same image twice, it trades that filename for a file.
we can get away replacing the UUID with something deterministic as long as it's signed with a server secret.

AC
- [ ] request for a JiraIssue returns a `descriptionHTML` that looks through all img tags & replaces the `src` with `/jira-attachment/filename.ext`. it uses cheerio to do this. see https://github.com/ParabolInc/parabol/blob/a6f7d3153947b4f5cd70ef48912bbdf00b60bb94/packages/server/email/inlineImages.ts#L42-L50 for inspiration
- [ ] filename is the original src hashed with our server secret (e.g. ` crypto
    .createHmac(algo, SERVER_SECRET)
    .update(url)
    .digest('base64')`)
- [ ] the server writes a sentinel value to redis like `{[filename]: 'X'}` to signify that an image will be there soon
- [ ] the server fetches all images from jira, but does not await the result (ie `descriptionHTML returns before jira responds to the image fetches)
- [ ] when the result comes back, the server writes the image as a buffer to redis `{[fileame]: <Buffer>}`
- [ ] an http endpoint exists like `/jira-attachment/filename.ext`
- [ ] the endpoint handler grabs the buffer from redis
- [ ] if it has the sentinel value ('X') then it wait 500ms and tries again.
- [ ] if it still has the sentinel value, it times out & sends a default sad panda image (@ackernaut do you have an image that we could use as a "image failed to load" message?)
- [ ] if it has the image, then it returns that image
- [ ] the image has a ttl of 2 hours in redis (if you want to do a multi-set, here's some inspiration: https://github.com/ParabolInc/parabol/blob/c64e145ab51465a0c8487460c3c76f9c98186cb2/packages/server/dataloader/RedisCache.ts)

Estimate: 8 hours

mattkrick:
@ackernaut @enriquesanchez do we have an image to show when an image fails to load? maybe a sad panda?

enriquesanchez:
@mattkrick I'm not aware of us having anything already designed, but I can come up with something. Do you know the approximate dimensions we need?

mattkrick:
[...]

PR #5190, opened 2021-07-20
Proxy Jira images through Parabol

BartoszJarocki:
Related issue: https://github.com/ParabolInc/parabol/issues/5100 

AC
- [ ] Jira description images are visible in Parabol  

Screenshot 2021-07-20 at 15 05 16


packages/server/utils/atlassian/jiraImages.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+export const downloadAndCacheImages = async (
+  authToken: string,
+  imageUrlToHash: Map<string, string>
+) => {
+  const redis = getRedis()
+  const promises: Promise<void>[] = []
+  for (const [imageUrl, hashedImageUrl] of imageUrlToHash) {
+    const cachedImage = await redis.get(hashedImageUrl)
```
mattkrick, line 54: use `exists` instead of `get` because the buffer will be large & we aren't doing anything with it

packages/server/utils/atlassian/jiraImages.ts
```diff
@@ -0,0 +1,111 @@
...
+    const imageBuffer = await fetchImage(authToken, imageUrl)
+    if (!imageBuffer) return
+    redis.setBuffer(hashedImageUrl, imageBuffer, 'PX', IMAGE_TTL_MS)
+  } catch (error) {
+    console.error(error)
+  }
+}
+
+const fetchImage = async (authToken: string, url: string): Promise<Buffer | null> => {
+  const response = await fetch(url, {
```
mattkrick, line 87: fetch needs a timeout & abort controller.
consider moving this logic into the AtlassianServerManager so it can use fetchWithTimeout

packages/server/utils/atlassian/jiraImages.ts
```diff
@@ -0,0 +1,111 @@
...
+const downloadAndCacheImage = async (
+  redis: Redis.Redis,
+  authToken: string,
+  hashedImageUrl: string,
+  imageUrl: string
+) => {
+  try {
+    redis.setBuffer(hashedImageUrl, NO_IMAGE_BUFFER)
+    const imageBuffer = await fetchImage(authToken, imageUrl)
+    if (!imageBuffer) return
```
mattkrick, line 79: if the fetch failed to return an image buffer let's remove the sentinel from redis so it tries again next time

packages/server/utils/atlassian/jiraImages.ts
```diff
@@ -0,0 +1,111 @@
...
+}
+
+const downloadAndCacheImage = async (
+  redis: Redis.Redis,
+  authToken: string,
+  hashedImageUrl: string,
+  imageUrl: string
+) => {
+  try {
+    redis.setBuffer(hashedImageUrl, NO_IMAGE_BUFFER)
```
mattkrick, line 77: let's put a TTL on this, too, just in case

packages/server/utils/atlassian/jiraImages.ts
```diff
@@ -0,0 +1,111 @@
...
+  }${JIRA_IMAGES_ENDPOINT}/${hashedImageUrl}`
+}
+
+const downloadAndCacheImage = async (
+  redis: Redis.Redis,
+  authToken: string,
+  hashedImageUrl: string,
+  imageUrl: string
+) => {
+  try {
```
mattkrick, line 76: no try/catch needed anymore

mattkrick (changes requested):
lots of comments, but they're all very tiny. great first pass!

packages/server/jiraImagesHandler.ts
```diff
@@ -0,0 +1,37 @@
+import {HttpRequest, HttpResponse} from 'uWebSockets.js'
+import uWSAsyncHandler from './graphql/uWSAsyncHandler'
+import {NO_IMAGE_BUFFER, JIRA_IMAGES_ENDPOINT} from './utils/atlassian/jiraImages'
+import getRedis from './utils/getRedis'
+
+const WAIT_MS = 500
+
+function wait(delayMS: number) {
```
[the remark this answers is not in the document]
BartoszJarocki: oh, right! fixed. 👍

packages/server/jiraImagesHandler.ts
```diff
@@ -0,0 +1,37 @@
...
+
+const ROUTE = `${JIRA_IMAGES_ENDPOINT}/`
+const jiraImagesHandler = uWSAsyncHandler(async (res: HttpResponse, req: HttpRequest) => {
+  const fileName = req.getUrl().slice(ROUTE.length)
+  const imageBuffer = await getImageFromCache(fileName)
+  if (imageBuffer.compare(NO_IMAGE_BUFFER) === 0) {
+    //TODO: send placeholder image
+  }
+
+  res
```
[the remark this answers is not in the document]
BartoszJarocki: fixed!

packages/server/server.ts
```diff
@@ -45,6 +47,7 @@ uws
   .get('/sse/*', SSEConnectionHandler)
   .get('/sse-ping', SSEPingHandler)
   .get('/self-hosted/*', selfHostedHandler)
+  .get(`${JIRA_IMAGES_ENDPOINT}/*`, jiraImagesHandler)
```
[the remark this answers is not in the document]
BartoszJarocki: TIL, fixed!

packages/server/utils/atlassian/jiraImages.ts
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Parses a JIRA issue description and replaces the image urls with Parabol's image urls
+ * @param {string} descriptionHTML - HTML string of related Jira issue description
+ * @returns {UpdateJiraImagesResult} - Map of orginal image urls to hashed image urls and updated description
+ */
+export const updateJiraImageUrls = (
+  cloudId: string,
+  descriptionHTML: string
+): UpdateJiraImagesResult => {
```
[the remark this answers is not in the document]
BartoszJarocki: @mattkrick sure, fixed! 

Re types, I've already added them. I also updated `cheerio` to the latest version.

packages/server/utils/atlassian/jiraImages.ts
```diff
@@ -0,0 +1,111 @@
...
+  if (response.status !== 200) {
+    console.warn('Fetching image failed', response.status)
+    return null
+  }
+
+  return response.buffer()
+}
+
+const createImageUrlHash = (imageUrl: string): string => {
+  const serverSecret = process.env.SERVER_SECRET
```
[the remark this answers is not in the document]
BartoszJarocki: oh, good to know, fixed. 👍

packages/server/utils/atlassian/jiraImages.ts
```diff
@@ -0,0 +1,111 @@
...
+
+    promises.push(downloadAndCacheImage(redis, authToken, hashedImageUrl, imageUrl))
+  }
+
+  await Promise.all(promises)
+}
+
+const createParabolImageUrl = (hashedImageUrl: string): string => {
+  //locally API lives on different port than the frontend application
+  return `${
```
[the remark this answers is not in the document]
BartoszJarocki: That's awesome. I was looking for how other endpoints are handled in the project but I couldn't find it. Fixed 👍

packages/server/utils/atlassian/jiraImages.ts
```diff
@@ -0,0 +1,111 @@
...
+/**
+ * Parses a JIRA issue description and replaces the image urls with Parabol's image urls
+ * @param {string} descriptionHTML - HTML string of related Jira issue description
+ * @returns {UpdateJiraImagesResult} - Map of orginal image urls to hashed image urls and updated description
+ */
+export const updateJiraImageUrls = (
+  cloudId: string,
+  descriptionHTML: string
+): UpdateJiraImagesResult => {
+  const imageUrlToHash: Map<string, string> = new Map()
```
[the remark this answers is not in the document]
BartoszJarocki: oh, sure. I need to do my homework with TS utility types. Fixed!
BTW, the map prints nicely too 😆

packages/server/jiraImagesHandler.ts
```diff
@@ -0,0 +1,37 @@
...
+    await wait(WAIT_MS) // try again after 500 ms
+    imageBuffer = await redis.getBuffer(fileName)
+  }
+  return imageBuffer
+}
+
+const ROUTE = `${JIRA_IMAGES_ENDPOINT}/`
+const jiraImagesHandler = uWSAsyncHandler(async (res: HttpResponse, req: HttpRequest) => {
+  const fileName = req.getUrl().slice(ROUTE.length)
+  const imageBuffer = await getImageFromCache(fileName)
```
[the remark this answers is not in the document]
BartoszJarocki: Makes sense! Fixed 🙏

packages/server/utils/atlassian/jiraImages.ts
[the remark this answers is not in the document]
BartoszJarocki: Ooooh. Interesting. After adding this to `dev.client.config`:
```
      '/jira-attachements': {
        target: `http://localhost:${SOCKET_PORT}`
      }
```
Image request couldn't reach the handler and I started digging what could be the issue.
Seems like hash created that way:
```
const createImageUrlHash = (imageUrl: string) => {
  return crypto
    .createHmac('sha256', serverSecret)
    .update(imageUrl)
    .digest('base64')
}
```
Isn't safe to use in URLs 🤯 . It creates from`https://api.atlassian.com/ex/jira/5bf7b2f4-cec7-4f8a-85f3-5643792a9052/secure/attachment/10000/10000_Screenshot+2021-07-19+at+08.25.20.png` ->`/redZ5TBVrI8sW7FgP79FTY+utdAdaAR/o0raZUkfXc=` and because the hash contains `/` at the beginning it has to be URL encoded. 
More reading: https://www.lytzen.name/2019/09/25/can-a-hash-have-a-slash.html and https://en.wikipedia.org/wiki/Base64#The_URL_applications 

I ended up creating a little helper:
```
const toBase64url = (base64: string) => {
  return base64
    .replace(/=/g, '')
    .replace(/\+/g, '-')
    .replace(/\//g, '_')
}
```
And I simply fix the hashed value
```
const createImageUrlHash = (imageUrl: string) => {
  return toBase64url(
    crypto
      .createHmac('sha256', serverSecret)
      .update(imageUrl)
      .digest('base64')
  )
}
```
After that, hash looks like this `_redZ5TBVrI8sW7FgP79FTY-utdAdaAR_o0raZUkfXc` instead of `/redZ5TBVrI8sW7FgP79FTY+utdAdaAR/o0raZUkfXc=`

packages/server/jiraImagesHandler.ts
```diff
@@ -0,0 +1,37 @@
...
+  if (imageBuffer.compare(NO_IMAGE_BUFFER) === 0) {
+    //TODO: send placeholder image
+  }
+
+  res
+    .writeStatus('200')
+    .writeHeader('Content-Type', 'image/png')
+    .writeHeader('Content-Length', imageBuffer.length.toString())
```
BartoszJarocki, line 33: @mattkrick after enabling proxy on dev server I was getting this
```
[HPM] Error occurred while trying to proxy request /jira-attachements/_redZ5TBVrI8sW7FgP79FTY-utdAdaAR_o0raZUkfXc from localhost:3000 to http://localhost:3001 (HPE_UNEXPECTED_CONTENT_LENGTH) (https://nodejs.org/api/errors.html#errors_common_system_errors)
```
I removed this and it works correctly. I'm not sure about prod env, is `Content-Length` going to be set automatically there too?

packages/server/utils/atlassian/jiraImages.ts
BartoszJarocki: When I add a method that needs to return a `Buffer` to `AtlassianManager` it complains the `response.buffer()` doesn't exist on type `fetch`. It's most probably because `AtlassianManager` has `abstract fetch: typeof fetch` which uses libdom type. 

I see the following solutions
1. extract `fetchWithTimeout` as a separate function that'd have to be bound to a proper `fetch` 
2. instead of `node-fetch` and native browser fetch use something like this https://github.com/matthew-andrews/isomorphic-fetch 
3. Simply copy-paste `fetchWithTimeout` as a separate function that uses `node-fetch` 

Unless there's some simpler solution that'd require changing types in `AtlassianManager` and `AtlassianServerManager`. Any thoughts @mattkrick? 🙏

BartoszJarocki:
@mattkrick thank you for a such detailed review! I fixed almost everything, there are two outstanding questions:
1. How are we want to handle `fetchWithTimeout` - https://github.com/ParabolInc/parabol/pull/5190#discussion_r673714197 
2. How about `Content-Length` header - https://github.com/ParabolInc/parabol/pull/5190#discussion_r673693605 

I'd appreciate your input there, thank you!

packages/server/utils/atlassian/jiraImages.ts
BartoszJarocki: ah, right! fixed 👍

packages/server/utils/atlassian/jiraImages.ts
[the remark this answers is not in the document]
mattkrick: ahh good catch!
we have a package for exactly that purpose. you can see it in action here: https://github.com/ParabolInc/parabol/blob/8e68c8c8106a24333522b25d5ad3306c4de34d5c/packages/server/email/createEmailVerification.ts#L21

so we can rewrite this to something like: 
```ts
base64url.encode(createHmac(...).update(imageUrl))
```

packages/server/utils/atlassian/jiraImages.ts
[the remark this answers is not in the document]
mattkrick: omg you're right! ohh that's so much nicer than it used to be. colors and all!

packages/server/utils/atlassian/jiraImages.ts
mattkrick: buffer doesn't exist, but arrayBuffer does :wink: 

see https://github.com/ParabolInc/parabol/blob/de68ce6722085bb6ca0ac9b27a08733188dc2b6c/packages/client/utils/AtlassianManager.ts#L438

packages/server/jiraImagesHandler.ts
mattkrick: here's what i'd try:
- remove the content-length header & on the browser open dev tools & check the response headers to verify that content-length is received. i think it should be
- if that doesn't work, i think you want byteLength, see https://nodejs.org/api/buffer.html#buffer_static_method_buffer_bytelength_string_encoding. try String(imageBuffer.byteLength)

BartoszJarocki: yeah, it is received. I just don't know if the prod env will behave the same as dev. I think it should 🤔

packages/server/utils/atlassian/jiraImages.ts
BartoszJarocki: I wanted to avoid the conversion ;) will move the logic to atlassian manager then!

packages/server/utils/atlassian/jiraImages.ts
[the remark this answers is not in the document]
BartoszJarocki: perfect! fixed 👍

packages/server/utils/atlassian/jiraImages.ts
BartoszJarocki: Now I realized I'm not handling `content-type` correctly and I always assume it's `png` 😭 
@mattkrick do you have any thoughts about how to store content types along with buffers in Redis?
What about storing it as a string with key like `content-type|imageHash` and just retrieve it endpoint handler and return the correct value? 🤔

mattkrick: here's what i'd do:
- the jira URL is /:fileName.[ext]
- remove the .[ext] before hashing the fileName
- then put the extension on the hashed fileName
- content-type can be determined from the extension using the package mime-types https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/server/utils/StaticServer.ts#L2

BartoszJarocki: that's a great idea! 🙏   added.

packages/server/jiraImagesHandler.ts
mattkrick: oh for sure. uWS probably measures the payload before sending & attaches the necessary headers to keep the request from hanging, nginx just sits in front of it adding it's own headers so we're all good

packages/server/jiraImagesHandler.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (imageBuffer.length > 1) return imageBuffer
+  if (tryAgain) {
+    await sleep(500)
+    return getImageFromCache(fileName, false)
+  }
+  return NO_IMAGE_BUFFER
+}
+
+const jiraImagesHandler = uWSAsyncHandler(async (res: HttpResponse, req: HttpRequest) => {
+  const fileName = req.getParameter(0)
```
mattkrick, line 21: let's first validate fileName before we perform some async work:
- make sure it's not an empty string
- get the mimeType
- make sure the mimeType is an image. if it's not, return a placeholder image
- now that we know the fileName at least looks like an image's file name, call getImageFromCache

packages/server/utils/atlassian/jiraImages.ts
```diff
@@ -0,0 +1,87 @@
...
+  const $ = cheerio.load(descriptionHTML)
+  $('body')
+    .find('img')
+    .each((_i, img) => {
+      const imageUrl = $(img).attr('src')
+      if (!imageUrl) return
+
+      const absoluteImageUrl = `${projectBaseUrl}${imageUrl}`
+      const extname = path.extname(absoluteImageUrl)
+      const imageUrlWithoutExt = absoluteImageUrl.replace(extname, '')
```
mattkrick, line 36: foo.png.png would fail here.
imageUrlWithoutExtension = absoluteImageUrl.slice(0, absoluteImageUrl.length - extname.length)

packages/server/utils/atlassian/jiraImages.ts
```diff
@@ -0,0 +1,87 @@
...
+  const isImageAlreadyCached = await redis.exists(hashedImageUrl)
+  if (isImageAlreadyCached) return
+
+  redis.setBuffer(hashedImageUrl, NO_IMAGE_BUFFER, 'PX', IMAGE_TTL_MS)
+  const imageBuffer = await manager.getImage(imageUrl)
+  if (!imageBuffer) {
+    await redis.del(hashedImageUrl)
+    return
+  }
+  redis.setBuffer(hashedImageUrl, imageBuffer, 'PX', IMAGE_TTL_MS)
```
mattkrick, line 77: nit: add `return` so in the future if anything wants to await this function it will work correctly

packages/server/utils/atlassian/jiraImages.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+const createImageUrlHash = (imageUrl: string) => {
+  return base64url.encode(
+    crypto
+      .createHmac('sha256', serverSecret)
+      .update(imageUrl)
+      .digest('base64')
```
mattkrick, line 85: `digest()` instead of `digest('base64')` to shrink your payload

```sh
> b6.encode(crypto.createHmac('sha256', 'foo').update('bar').digest('base64')).length
59
> b6.encode(crypto.createHmac('sha256', 'foo').update('bar').digest('')).length
43
> 
```

packages/server/jiraImagesHandler.ts
```diff
@@ -0,0 +1,34 @@
...
+import mime from 'mime-types'
+import sleep from '../client/utils/sleep'
+import uWSAsyncHandler from './graphql/uWSAsyncHandler'
+import {NO_IMAGE_BUFFER} from './utils/atlassian/jiraImages'
+import getRedis from './utils/getRedis'
+
+const getImageFromCache = async (fileName: string, tryAgain: boolean) => {
+  const redis = getRedis()
+  const imageBuffer = await redis.getBuffer(fileName)
+  if (imageBuffer === null || imageBuffer.length === 0) return NO_IMAGE_BUFFER
```
mattkrick, line 11: we still need a placeholder for a graceful fallback.
lets do the following:
- put a .png of a jira logo in the static images directory
- import it at the top of this file (you may need to add a webpack file-loader to the dev & prod server configs)
- if the image isn't found, return that placeholder (to test in prod, just run `yarn build` then `yarn start`

packages/server/utils/atlassian/jiraImages.ts
BartoszJarocki: 💯  I def need to be less optimistic when coding!

packages/server/utils/atlassian/jiraImages.ts
BartoszJarocki: ahh, right!

packages/server/jiraImagesHandler.ts
```diff
@@ -1,32 +1,57 @@
 import {HttpRequest, HttpResponse} from 'uWebSockets.js'
+import fs from 'fs'
 import mime from 'mime-types'
 import sleep from '../client/utils/sleep'
 import uWSAsyncHandler from './graphql/uWSAsyncHandler'
-import {NO_IMAGE_BUFFER} from './utils/atlassian/jiraImages'
 import getRedis from './utils/getRedis'
+import jiraPlaceholder from '../../static/images/jira/placeholder.png'
+
+const jiraPlaceholderBuffer = fs.readFileSync(jiraPlaceholder)
```
BartoszJarocki, line 9: I'm not sure what's the best way of reading the file in a node so @mattkrick I'd appreciate if you could take a look here! Thanks!

BartoszJarocki:
@mattkrick thank you for the review! I fixed all the feedback but I'd really appreciate if you could take a look at webpack stuff and file loading, thank you! 🙏

packages/server/server.ts
```diff
@@ -45,6 +46,7 @@ uws
   .get('/sse/*', SSEConnectionHandler)
   .get('/sse-ping', SSEPingHandler)
   .get('/self-hosted/*', selfHostedHandler)
+  .get('/jira-attachements/:fileName', jiraImagesHandler)
```
mattkrick, line 49: *attachments (remove the extra e)

packages/server/jiraImagesHandler.ts
mattkrick: let's do a lazy + async read.
we'll make it async so it doesn't block the event loop
we'll make it lazy so the server doesn't have to read the file every time it starts up

```ts
import {promises as fsp} from 'fs'
let jiraPlaceholderBuffer: Buffer | undefined

const servePlaceholderImage = async (res: HttpResponse) => {
  if (!jiraPlaceholderBuffer) {
    jiraPlaceholderBuffer = await fsp.readFile(jiraPlaceholder)
  }
  res.cork(() => {
    res
      .writeStatus('200')
      .writeHeader('Content-Type', 'image/png')
      .end(jiraPlaceholderBuffer)
  })
}
```

packages/server/server.ts
BartoszJarocki: 🤦 I've just installed https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker so hopefully that won't happen again! 😂

packages/server/jiraImagesHandler.ts
BartoszJarocki: Thank you, all noted!

mattkrick:
added an `await` to serving the placeholders, otherwise it'd return before responding & that causes all kinds of bad.
aside from that all good!

Issue #4813, opened 2021-03-24
Improved empty state for discussion thread


enriquesanchez:
## Issue - Enhancement

Follow up of #3901 and #4812. We want to improve the clarity and visuals of the empty/initial state of the discussion and task thread.

This proposal adds an illustration from Parabol's illustrations library and improves the clarity of the message displayed.

### Current vs. proposed

<img width="816" alt="Screen Shot 2021-03-24 at 12 17 05 p m" src="https://user-images.githubusercontent.com/3276087/112363232-e5038f00-8c9a-11eb-8745-e7f863d57800.png">

[Please refer to the Figma file for a more detailed view of the design](https://www.figma.com/file/Ia7SoJcudXUM3nvOYXOldR/comments-and-tasks?node-id=130%3A0).

### Acceptance Criteria
Users can:
   - Easily understand that the component is in an empty/initial state
   - Use the proposed design as a guide for adding a new comment or creating a new task

**Estimated effort:**
- 10 hours.

jordanh:
Brilliant. Adding this to the next sprint, sir

PR #5192, opened 2021-07-20
Improved discussion thread empty state

BartoszJarocki:
Related issue: https://github.com/ParabolInc/parabol/issues/4813 

This is how it looks like on a big screen:
Screenshot 2021-07-20 at 17 34 03

This is how it looks like on a smaller screen - the empty state along with the header scrolls. I'm not sure if that's the best solution so if someone has better ideas I'd love to improve! 
Screenshot 2021-07-20 at 17 34 28


packages/client/components/DiscussionThreadListEmptyState.tsx
```diff
@@ -23,10 +40,23 @@ const DiscussionThreadListEmptyState = (props: Props) => {
...
+    <DiscussionThreadEmptyStateRoot>
+      <EmptyDiscussionContainer>
+        <EmptyDiscussionSVG />
+      </EmptyDiscussionContainer>
+      <Message>{isReadOnly ? readOnlyMessage : message}</Message>
+      {!isReadOnly && (
+        <TangledArrowContainer>
+          <TangledArrowSVG />
+        </TangledArrowContainer>
+      )}
```
BartoszJarocki, line 57: It felt weird to have this arrow pointing to disabled input so I removed it in this case. Is that acceptable @enriquesanchez?

enriquesanchez: Oh, interesting! In what scenarios would this happen?

enriquesanchez:
This is looking great, @BartoszJarocki 🙌

>This is how it looks like on a smaller screen - the empty state along with the header scrolls. I'm not sure if that's the best solution so if someone has better ideas I'd love to improve!

I think in this scenario we can remove the `TangledArrowSVG`. This way we keep the illustration and the text, and shouldn't create any unnecessary scrolling.

BartoszJarocki:
@enriquesanchez Thank you for the review! I've added hiding arrow on smaller devices. 
Screenshot 2021-07-21 at 15 15 08

mattkrick:
nice attention to detail, LGTM!

Issue #5097, opened 2021-07-02
Comment Status Text is buggy 


mattkrick:
Looks like the comment text doesn't go away. Ah!
This is probably due to refactoring the comment onto the discussion thread instead of on the meeting, but maybe not!
Seems like the status appears easily enough, but when someone stops typing, then it doesn't go away

AC
- [ ] Comment status text works correctly
![Screenshot from 2021-07-01 21-12-38](https://user-images.githubusercontent.com/5514175/124219401-27feab00-dab1-11eb-8814-c53cc2bb6153.png)

PR #5197, opened 2021-07-21
Fixed comment status text

BartoszJarocki:
Related issue: https://github.com/ParabolInc/parabol/issues/5097 

The main problem was with ids comparison.


packages/client/hooks/useBeforeUnload.ts
```diff
@@ -0,0 +1,13 @@
+import {useRef, useEffect} from 'react'
+
+export const useBeforeUnload = (callback: () => any) => {
```
mattkrick, line 3: oooo i like where this is headed!
this solution works right now, but what if in the future discussionId changes from "foo" to "bar"? the callback is persisted & points to the old value of discussionId. That'd be a tricky bug to track down.

easy fix:

`const cb = useEventCallback(callback)`

then you can pass cb to window event listeners

BartoszJarocki: oh, that's 100% right 🤦

BartoszJarocki:
@mattkrick fixed, I'd appreciate if you could take a look 🙏

mattkrick:
:ok_hand:

PR #5201, opened 2021-07-22
Team reads migration

### LINKED ISSUE #4630: Read Team from PG instead of RethinkDB (Phase 2 of 3)

tianrunhe:
Resolves #4630.

I brought the changes from Tiffany's last PR #4817 but also resolves the merge conflicts and update a lot of the code.

Tests:
- [x] Run through the [release test checklist](https://www.notion.so/parabol/Release-Tests-Checklist-a4146ee7b24f455681e6ec58225974df)
- [x] Manually test the following mutations in `/admin/graphql`
  - [x] `archiveOrganization`
  - [x] `downgradeToPersonal`
  - [x] `updateCreditCard`
  - [x] `updateTeamName`
  - [x] `upgradeToPro`
  - [x] `hideConversionModal`
  - [x]  `moveTeamToOrg`
  - [x] `startRetrospective`
  - [x] `verifiedInvitation`
  - [x] `acceptTeamInvitation`


packages/server/postgres/queries/getTeamByTeamId.ts
```diff
@@ -0,0 +1,10 @@
+import {getTeamByTeamIdQuery, IGetTeamByTeamIdQueryResult} from './generated/getTeamByTeamIdQuery'
+import catchAndLog from '../utils/catchAndLog'
+import getPg from '../getPg'
+
+const getTeamByTeamId = async (teamId: string): Promise<IGetTeamByTeamIdQueryResult> => {
+  const teams = await catchAndLog(() => getTeamByTeamIdQuery.run({id: teamId}, getPg()))
+  return teams![0]
```
tiffanyhan, line 7: curious why add the bang operator?  what happens if `catchAndLog` returns null? should we add `teams === null ? [] : teams` before returning the first value, similarly to `packages/server/postgres/queries/getTeamsByOrgId.ts`?

tiffanyhan (commented):
wow, i'm glad you were able to use work from that old PR! i hope it wasn't too hairy to update 😬 thanks for making these changes!

packages/server/postgres/queries/getTeamByTeamId.ts
tianrunhe: This method is expected to return a single team object, isn't it? I did have difficulty in handling the nullify here. Any better idea?

tiffanyhan: i think we could do 

```
teams === null ? [] : teams
return teams[0]
```

if the query returns null, then this wrapper will return undefined instead of raising an error

packages/server/utils/authorization.ts
```diff
@@ -100,6 +100,7 @@ export const isOrgLeaderOfUser = async (authToken, userId) => {
   return uniques.size < total
 }
 
+// this function is not used anywhere
 export const isPaidTier = async (teamId) => {
```
mattkrick, line 104: let's nuke this thing!

packages/server/postgres/queries/getTeamsByOrgId.ts
```diff
@@ -0,0 +1,26 @@
...
+const getTeamsByOrgId = async (
+  orgIds: string | string[],
+  options: Partial<Omit<IGetTeamsByOrgIdQueryParams, 'orgId'>> = {}
+): Promise<IGetTeamsByOrgIdQueryResult[]> => {
+  orgIds = typeof orgIds === 'string' ? [orgIds] : orgIds
+  const orgs = await catchAndLog(() =>
+    getTeamsByOrgIdQuery.run(
+      {
+        orgIds,
+        ...options
```
mattkrick, line 18: let's coerce `isArchived` to a boolean here. then we can get rid of the CAST + condition is the SQL.

packages/server/postgres/queries/getTeamByTeamId.ts
mattkrick: let's nuke this file & just use `getTeamsById`

packages/server/graphql/mutations/moveTeamToOrg.ts
```diff
@@ -10,16 +10,19 @@ import safeArchiveEmptyPersonalOrganization from '../../safeMutations/safeArchiv
...
   const r = await getRethink()
   // AUTH
   const su = isSuperUser(authToken)
   // VALIDATION
-  const {team, org} = await r({
-    team: (r.table('Team').get(teamId) as unknown) as Team,
-    org: (r.table('Organization').get(orgId) as unknown) as Organization
-  }).run()
+  const [{org}, team] = await Promise.all([
+    r({
```
mattkrick, line 21: remove `r({org: ` wrapper

packages/server/graphql/mutations/helpers/resolveDowngradeToPersonal.ts
```diff
@@ -19,7 +19,7 @@ const resolveDowngradeToPersonal = async (
     console.log(e)
   }
 
-  const [rethinkResult] = await Promise.all([
+  const [, teamIds] = await Promise.all([
```
mattkrick, line 22: let's not return `teamIds` here so we can remove the RETURN from our write query.
then, in https://github.com/ParabolInc/parabol/blob/36a79f894e9a78622a4fa95c8aa308591c8f1e6c/packages/server/graphql/mutations/downgradeToPersonal.ts#L44 we can just get the teamIds via the new dataLoader

packages/server/graphql/mutations/addTeam.ts
```diff
@@ -41,15 +40,8 @@ export default {
...
-        .table('Team')
-        .getAll(orgId, {index: 'orgId'})
-        .run()
-
-      const orgTeamNames = [] as string[]
-      orgTeams.forEach((team) => {
-        if (!team.isArchived) orgTeamNames.push(team.name)
-      })
+      const orgTeams = await getTeamsByOrgId(orgId, {isArchived: false})
+      const orgTeamNames = orgTeams.map((team) => team.name)
```
mattkrick, line 44: so much cleaner! :tada:

mattkrick (changes requested):
great work! mostly just little things

packages/server/graphql/mutations/startRetrospective.ts
```diff
@@ -61,7 +61,7 @@ export default {
       meetingType,
       dataLoader
     )
-    const team = await getTeamByTeamId(teamId)
+    const team = await getTeamsById([teamId])[0]
```
mattkrick, line 64: be defensive! if they pass in "foo" as the teamId this will be undefined & throw.

packages/server/graphql/mutations/downgradeToPersonal.ts
```diff
@@ -41,7 +41,8 @@ export default {
 
     // RESOLUTION
     // if they downgrade & are re-upgrading, they'll already have a stripeId
-    const {teamIds} = await resolveDowngradeToPersonal(orgId, stripeSubscriptionId!, viewerId)
+    await resolveDowngradeToPersonal(orgId, stripeSubscriptionId!, viewerId)
+    const {teamIds} = await dataLoader.get('teamsByOrgId').load(orgId)
```
mattkrick, line 45: this code was not tested! this dataloader will not return an object containing `teamIds`.
can you please:
- add a `GQLContext` type to the context object at the top so we get some type safety
- add a checklist of tests that you performed (see https://github.com/ParabolInc/parabol/pull/5205 as an example)

mattkrick:
tagging @tiffanyhan to perform the code review

tiffanyhan:
@bruce i ran into this unexpected behavior while going through the tests:

- [ ] start with a team on a free org
- [ ] start retrospective meeting, but don't end it or go to discuss yet
- [ ] set `showConversionModal=true` on the meeting object in rethinkdb
- [ ] now upgrade the org to pro
- [ ] see in rethink db that `showConversionModal=true` even though it should be false by `hideConversionModal`

Can you please run through the reproduction steps and see if it's behaving similarly for you? If so, I can just open up a new issue that won't block this PR, since I see this PR doesn't have any changes with regard to the file `packages/server/graphql/mutations/helpers/hideConversionModal.ts`.

packages/server/graphql/intranetSchema/mutations/backupOrganization.ts
```diff
@@ -52,10 +53,7 @@ const backupOrganization = {
       .run()
 
     // get all the teams for the orgIds
-    const team = await r
-      .table('Team')
-      .getAll(r.args(orgIds), {index: 'orgId'})
-      .run()
+    const team = await getTeamsByOrgIds(orgIds)
```
tiffanyhan, line 56: can we make this plural? i.e. team => teams

tiffanyhan (changes requested):

tianrunhe:
> @tianrunhe i ran into this unexpected behavior while going through the tests:
> 
> *   [ ] start with a team on a free org
> *   [ ] start retrospective meeting, but don't end it or go to discuss yet
> *   [ ] set `showConversionModal=true` on the meeting object in rethinkdb
> *   [ ] now upgrade the org to pro
> *   [ ] see in rethink db that `showConversionModal=true` even though it should be false by `hideConversionModal`
> 
> Can you please run through the reproduction steps and see if it's behaving similarly for you? If so, I can just open up a new issue that won't block this PR, since I see this PR doesn't have any changes with regard to the file `packages/server/graphql/mutations/helpers/hideConversionModal.ts`.

@tiffanyhan I can reproduce this and I think it's due to these two lines:

[https://github.com/ParabolInc/parabol/blob/master/packages/server/graphql/mutations/helpers/hideConversionModal.ts#L7-L8](https://github.com/ParabolInc/parabol/blob/master/packages/server/graphql/mutations/helpers/hideConversionModal.ts#L7-L8)

Which makes me question if this is a valid test case, i.e., this could never happen in real world:
1. `showConversionModal` can be attributes on both `Organization` and `Meeting` object
2. For `Organization` object, its value can only be mutated through the **private** mutation `flagConversionModal`
3. The `showConversionModal` attribute for a `Meeting` object is set when the meeting starts, and it checks the [`showConversionModal` for the organization the team is in](https://github.com/ParabolInc/parabol/blob/master/packages/server/graphql/mutations/startRetrospective.ts#L69)

Therefore there won't be a case where the attribute for the org is false whereas a meeting in that org has the attribute value to be true, as specified in the test case.

tiffanyhan (approved):
lgtm! ran through all specified mutations and all worked well

packages/server/postgres/queries/getTeamsByOrgIds.ts
```diff
@@ -0,0 +1,22 @@
+import {
+  getTeamsByOrgIdQuery,
+  IGetTeamsByOrgIdQueryParams,
+  IGetTeamsByOrgIdQueryResult
+} from './generated/getTeamsByOrgIdQuery'
+import getPg from '../getPg'
+
+const getTeamsByOrgIds = async (
+  orgIds: string[],
+  options: Partial<Omit<IGetTeamsByOrgIdQueryParams, 'orgId'>> = {}
```
mattkrick, line 10: i think you want to omit `orgIds` here?
might just be simpler to write: options: {isArchived?: boolean} = {}

packages/server/postgres/queries/updateTeamByOrgId.ts
```diff
@@ -9,7 +9,7 @@ const updateTeamByOrgId = async (
   teamUpdates: Partial<IUpdateTeamByOrgIdQueryParams>,
   orgId: string
 ) => {
-  await catchAndLog(() =>
+  return await catchAndLog(() =>
```
mattkrick, line 12: no `return await` (i'll turn on a rule after we get this merged)
remove catchAndLog.

packages/server/safeMutations/acceptTeamInvitation.ts
```diff
@@ -61,8 +57,11 @@ const acceptTeamInvitation = async (teamId: string, userId: string) => {
           .getAll(userId, {index: 'userId'})
           .filter({removedAt: null})
           .coerceTo('array')
-      }) as unknown) as User & {organizationUsers: OrganizationUser[]}
-  }).run()
+      })
+      .run(),
+    getTeamsById([teamId])
+  ])
+  const team = teams[0] ?? null
```
mattkrick, line 64: here you're saying team could be null, but in the next line you're accessing team. that'll cause a nasty runtime error if it really is null. (there are a handful of other places this happens)

packages/server/safeMutations/safeArchiveTeam.ts
```diff
@@ -22,7 +22,7 @@ const safeArchiveTeam = async (teamId: string) => {
     isArchived: true,
     updatedAt: new Date()
   }
-  const [rethinkResult] = await Promise.all([
+  const [rethinkResult, pgResult] = await Promise.all([
```
mattkrick, line 25: i'm not sure this was checked, `pgResult` is going to be an empty array...

packages/server/safeMutations/safeArchiveTeam.ts
```diff
@@ -49,7 +49,7 @@ const safeArchiveTeam = async (teamId: string) => {
     }).run(),
     updateTeamByTeamId(updates, teamId)
```
mattkrick, line 50: `updateTeamByTeamId` still has a catchAndLog in it (as do a few other files related to the Team table)
could you clean up those remaining files?

packages/server/graphql/mutations/addTeam.ts
```diff
@@ -41,15 +40,8 @@ export default {
...
-      const orgTeams = await r
-        .table('Team')
-        .getAll(orgId, {index: 'orgId'})
-        .run()
-
-      const orgTeamNames = [] as string[]
-      orgTeams.forEach((team) => {
-        if (!team.isArchived) orgTeamNames.push(team.name)
-      })
+      const orgTeams = await getTeamsByOrgIds(orgId, {isArchived: false})
```
mattkrick, line 43: So...I can't create a team on this PR :astonished: 

calling `addTeam` throws an error because getTeamsByOrgIds takes an array as the first param. 
While the fix is easy, I'm a little concerned that this broken code went through 2 reviews approvals.
In the future can we be sure to test touched files? Some things are sure to slip through the cracks, and that's OK! But if 3 people test all the changed files we should be able to avoid broken stuff landing in prod.

@tianrunhe @tiffanyhan

tianrunhe: This is my bad. Apologize! Here's the retro:  
  
I checked the references for `getTeamsByOrgId` and saw two patterns  
1. places where `getTeamsByOrgIds` gets called directly
2. it's used in `teamsByOrgId` as a foreign data loader and the data loader is then used in a lot of places.

I tested everything in 2. and that's how I build the test cases. I did not go through all use cases in 1. and that's how `addTeam` get slipped.

In the future I think we do not only need to test all touched files, but also references to changed methods.

mattkrick:
Next steps for dataloaders:
- remove `teams` primary loader
- remove all foreignLoaders that depend on the `teams` primary loader ('teamsByOrgId')
- add a `teamsByOrgId` customLoader

tianrunhe:
@mattkrick I've updated the loaders, please review. Thanks!

packages/server/safeMutations/safeArchiveTeam.ts
```diff
@@ -49,7 +49,7 @@ const safeArchiveTeam = async (teamId: string) => {
...
   ])
-  return {...rethinkResult, users}
+  return {...rethinkResult, team: pgResult[0] ?? null, users}
```
mattkrick, line 52: The `team` object that comes from RethinkDB is null if the object wasn't changed.
`RETURNING *` will always return a team object as long as it exists, so the error message in `archiveTeam` would be wrong.
This is also the only place that `RETURNING *` is used, so it feels a little bad that we're always performing an extra read just for this 1 place.

Options:
- write an `archiveTeam` query that includes `WHERE "isArchived" = false`
- query the team at the top of this function. if `isArchived == false` then return early with {team: null}. else, query the team again (maybe via dataLoader). There's a small chance of a race condition with this, but it's probably OK.
-  ???

mattkrick:
Just one query to go!

mattkrick:
LGTM!

PR #5209, opened 2021-07-23
updating meeting cards styles for improved contrast and HTML semantics

acressall:
## Description
In efforts to slowly improve accessibility @enriquesanchez and I have identified a small element on the meeting view cards where colour contrast could be improved. 

### Visual Changes:
- MeetingTypeLabel: 11px -> 12px
- Name: GRAPE_700 -> GRAPE_800; 20px -> 22px
- Meta: GRAPE_600 -> GRAPE_700

Left: updates | Right: Original
Screen Shot 2021-07-23 at 12 48 22 PM

### DIFF
--- a/packages/client/components/MeetingCard.tsx
+++ b/packages/client/components/MeetingCard.tsx
@@ -52,18 +52,18 @@ const MeetingInfo = styled('div')({
   padding: '8px 8px 8px 16px'
 })
 
-const Name = styled('div')({
-  color: PALETTE.SLATE_700,
-  fontSize: 20,
-  lineHeight: '24px',
+const Name = styled('span')({
+  color: PALETTE.SLATE_800,
+  fontSize: 22,
+  lineHeight: '26px',
   // add right padding to keep a long name from falling under the options button
   // add top and bottom padding to keep a single line at 32px to match the options button
   padding: '4px 32px 4px 0',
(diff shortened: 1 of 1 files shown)


ackernaut (changes requested):
- Our grid value is **8n give or take 4**. For flexibility give or take 4 sparingly. This is a sub-grid value that MD mentions helps alignment of icons so I borrowed that to help align other things
- I like going from 11 to 12 for the meta lines
- I don’t see a need to go from 20 to 22 on the name line. I think it’s too big. The ratio from 20/12 is still pretty strong at 1.67. 20  at Slate 700 || 800 is accessible. Currently in our components, not all type is 8n give or take 4 but I’m trying to move toward that. We would have 16, 20, 24 — 24 is too big here
- I like sticking to a grid footprint for line height as much as possible (so here 24, 28, 32)
- The meeting name needs to be centered with the more menu icon
- The gutters around the text and between are too tight now. The left and right gutter of the card should be 16 (2*8). We can help the rest by increasing gutters around and between by a grid value. Give it some room to breath. Most people probably only have 1 card or 1 row of cards.
- I’ve been experimenting with the lighter Slate for larger text, so here the name line would be Slate 700 and the meta line would be Slate 800

ackernaut (approved):

Issue #5189, opened 2021-07-20
Gracefully handle slow responding GQL Executor


mattkrick:
Triaged in #5157

Sometimes our GQL Executor service might go down or take awhile to respond.
If this happens, the client should fail gracefully.

AC
- [ ] In the PubSubPromise https://github.com/ParabolInc/parabol/blob/bfed594b1f8f053ea628a3c541502ddd9b47b374/packages/server/utils/PubSubPromise.ts#L45 instead of throwing an error about Redis taking too long, it returns a result that looks like a GraphQL payload, e.g. `{data: null, errors: [{message: 'GQL Executor took too long to respond'}]}`
- [ ] whenever this occurs, `sendToSentry` is called from the PubSubPromise and it includes the authToken, docId, query, variables from the request
- [ ] on the frontend, verify that the app doesn't crash

Estimate: 6 hours

mattkrick:
this is terrifying.
the next step is to see if there are specific fields (a la datadog apm), but i don't think that's the case because this happens across a bunch of different mutations & queries.

stiil, we'll want to see which fields are slowing something down. maybe 1 table or 1 query type is doing more harm than the others.

jordanh:
@mattkrick @BartoszJarocki @nickoferrall what do we think would be sensible next steps here? I think were we last left this off was starting with #5331 then with #5333

mattkrick:
yep, #5333 is the next step

jordanh:
**Story Points: 8**
  [See the discussion](https://action.parabol.co/meet/7By0K1Cj6w/estimate/1) in “Sprint `#89` Scoping”

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

PR #5210, opened 2021-07-23
Handle slow responding gql executor

nickoferrall:
PR to resolve issue #5189 

With these changes, the app shows the loading icon rather than crashes when the GQL Executor doesn't respond within the timeout limit.

### DIFF
--- a/packages/server/utils/PubSubPromise.ts
+++ b/packages/server/utils/PubSubPromise.ts
@@ -1,4 +1,5 @@
 import Redis from 'ioredis'
+import {GQLRequest} from '../graphql/executeGraphQL'
 import numToBase64 from './numToBase64'
 import sendToSentry from './sendToSentry'
 
@@ -9,6 +10,7 @@ interface Job {
 }
 
 const {SERVER_ID, REDIS_URL} = process.env
+
 export default class PubSubPromise<Request, Response> {
   jobs = {} as {[jobId: string]: Job}
   publisher = new Redis(REDIS_URL)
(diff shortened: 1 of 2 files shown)


packages/server/utils/PubSubPromise.ts
```diff
@@ -37,12 +39,27 @@ export default class PubSubPromise<Request, Response> {
...
       const timeoutId = setTimeout(() => {
         delete this.jobs[jobId]
-        reject(new Error('Redis took too long to respond'))
+        const errorMessage = `GQL Executor took more than ${MAX_TIMEOUT}ms to respond`
+        const {
+          authToken,
+          query = '',
+          variables = '',
+          docId = ''
+        } = (request as unknown) as GQLRequest
```
nickoferrall, line 53: In [getGraphQLExecutor](https://github.com/ParabolInc/parabol/blob/bfed594b1f8f053ea628a3c541502ddd9b47b374/packages/server/utils/getGraphQLExecutor.ts#L6), we specify that the first arg should be `GQLRequest` but the request here is unknown as it could theoretically be anything. I'm sure there's a better way that I'm forgetting but I've also spent more time than I prob should on this before calling for help. Do you have any suggestions for how we can destructure the req properties without casting?

mattkrick: ahhh this is why i love typescript! i definitely did you a disservice pointing you to this file in the issue.

the quick & dirty is to stringify the whole request `tags: {payload: JSON.stringify(request)}` and hope that it doesn't exceed 255 chars

the _right_ way is to move the handling of the error to the correct layer.

Basically we've got this "transport" layer that uses redis & only cares about redis pub/sub stuff. it doesn't care what it sends, just give it a channel & it'll send it.

Next layer up is the graphql or "data transport" layer. if tomorrow we replace redis with rabbitmq, aws, etc, it wouldn't change a thing. all it cares about is the gql fields. You correctly found this to be `getGraphQLExecutor`.

To fix, it'd look like this:
- put the `reject` back in, maybe `reject('TIMEOUT')`
- in the try/catch of `handleGraphQLTrebuchetRequest` go into the `catch` statement
- check if e?.message === 'TIMEOUT', if so, then send the fields you want to sentry & return the standard graphql response `{data: null, errors:[{message: 'The request took too long'}]}`

nickoferrall: Ah thanks, that makes sense. Updated! 

Previously, I changed the timeout limit to 100ms instead of 10000ms and I could see that new reflections were disappearing and the fatal error message would appear, particularly at 10/20 ms timeout. When I added the return statement, the user would see the loading spinner instead of an error message. Strangely, now when I try to reproduce that behaviour, when I do or don't include the return statement in `PubSubPromise` or `handleGraphQLTrebuchetRequest`, I don't see any difference in behaviour. It typically just shows the loading spinner. Still, I was able to repro the fix previously and this change seems safe to try.

mattkrick:
LGTM! we should probably gracefully handle this for things in the intranetGraphQLHandler, too, but that can wait for another PR

Issue #5106, opened 2021-07-06
Spotlight: UI Skeleton 


mattkrick:
To get moving on spotlight, we'll build the UI first. See #5105 for the full list of UI specs

AC
- [ ] in the reflection card component, there is a static variable `const showSpotlight = true`
- [ ] if `showSpotlight` is true, then the the search icon appears
- [ ] On desktop, the search icon is hidden until hovered
- [ ] Hovering the icon shows a tooltip that says "Find similar" (or whatever copy you like)
- [ ] Clicking the icon opens up the "Spotlight Modal", which is a large modal that takes up the entire meeting area (excluding left nav & right drawer)
- [ ] At the top of the spotlight modal is the card that the user clicked (it will animated into place, just like expanded groups do today)
- [ ] below the card is an autoFocused search input field with the search icon on the left
- [ ] instead of search results, it just shows the following
![124508014-a3ec4200-dd94-11eb-9295-d89df0c228d6](https://user-images.githubusercontent.com/5514175/124647704-2c2d1e80-de4b-11eb-8185-346b3a3fbfca.png)

Estimate: 24 hours

nickoferrall:
Hey @mattkrick & @enriquesanchez, the AC mentions that the modal should take up `the entire meeting area (excluding left nav & right drawer)`. Enrique mentioned in the Loom that we want to use as much space as we can while letting the user know that they're still in the Group phase.

A few questions:

1. We only show the right drawer in the Poker meeting so I assume we just want to account for the left sidebar?
2. Typically, we add opacity to the app around the modal to focus the user's attention on the modal. Do we want to do that here too? The designs in #5105 don't add the background opacity
3. If we want to add opacity around the modal, should we still leave space for the left sidebar? The second image shows what that would roughly look like. If we don't want to add opacity, I think it makes sense to leave space for the left sidebar as the modal feels like it just fits in place. If we do want to add opacity, I think we should leave space for the group title at the top but we probably don't need to leave space for the left sidebar as the user can't click the sidebar, it takes up real estate and it fits more naturally in the center imo. 

![image](https://user-images.githubusercontent.com/39854876/126240651-c2bdaaf0-51e5-4290-8013-0479fca1affa.png)

![right](https://user-images.githubusercontent.com/39854876/126240761-48acb326-3631-42eb-8ef8-d4db57953694.png)

[...]

PR #5214, opened 2021-07-23
Spotlight UI skeleton

nickoferrall:
PR to resolve issue #5106


packages/client/components/SpotlightModal.tsx
```diff
@@ -0,0 +1,207 @@
...
+  const {closeSpotlight, meeting, reflection} = props
+  const isDesktop = useBreakpoint(Breakpoint.NEW_MEETING_SELECTOR)
+  const handleKeyDown = (e: KeyboardEvent<HTMLDivElement>) => {
+    if (e.key === 'Escape') closeSpotlight()
+  }
+  const spotlightReflection = {...reflection, inSpotlight: true}
+  const reflectionGroupsCount = 0
+  return (
+    <ModalContainer onKeyDown={handleKeyDown} isDesktop={isDesktop}>
+      <SelectedReflection>
```
nickoferrall, line 133: Can move `SelectedReflection` and `SimilarReflectionGroups` to their own components once the logic is built up a bit more

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -55,16 +83,29 @@ const getReadOnly = (
...
+  const isComplete = meeting?.localStage?.isComplete
   const phases = meeting ? meeting.phases : null
-  const {id: reflectionId, content, promptId, isViewerCreator} = reflection
   const atmosphere = useAtmosphere()
   const {onCompleted, submitting, submitMutation, error, onError} = useMutationProps()
   const editorRef = useRef<HTMLTextAreaElement>(null)
   const [editorState, setEditorState] = useEditorState(content)
-
+  const [isHovering, setIsHovering] = useState(false)
+  const isDesktop = useBreakpoint(Breakpoint.SIDEBAR_LEFT)
```
nickoferrall, line 104: I find some of the breakpoints a bit confusing, e.g. why `SIDEBAR_LEFT` and `NEW_MEETING_SELECTOR` instead of say XS, S, M?

mattkrick: fair point!
there are 2 schools of thought:
- bins by device size (S,M, L or maybe phone, tablet, desktop, TV)
- bins by functionality (hide the left nav, show 2 columns instead of 3, etc)

binning by device size was originally really popular, but becomes a headache if the sizes of devices vary widely. you're basically building your app to a set number of devices. it's still pretty popular for UI frameworks!

binning by functionality means the layout is determined by your components, not the hardware specs of your users. at least that's the thought.

that said, we've done a real bad job of sticking with 1 approach :see_no_evil:

enriquesanchez (changes requested):
This is looking amazing @nickoferrall 🙌

Here's my feedback:

1. The spotlight modal should have a dropshadow. Let's use the same value we use for other modals/popups: 

`box-shadow: rgba(0, 0, 0, 0.2) 0px 5px 5px -3px, rgba(0, 0, 0, 0.14) 0px 8px 10px 1px, rgba(0, 0, 0, 0.12) 0px 3px 14px 2px;`

2. Is there a way to vertically align the title and the `X` button?
Screen Shot 2021-07-26 at 5 10 18 p m

3. We are going to have some reflection cards that are small (one line of text), and others that are taller. Ideally we can vertically center these instead of having them aligned to the top. Would this be possible?

Screen Shot 2021-07-26 at 5 11 21 p m

4. Any chance we can increase the color contrast of the search field's placeholder text? In the mockups I'm using `Slate 600 #82809A`.

nickoferrall:
Thanks for the review! 

I've implemented those changes. I was umming and arring about aligning the card at the top or center but I think you're right about center alignment. The placeholder text in that screenshot is lighter than how it looks for me - I've added `Slate 600` which looks similar to how it was before so please let me know if it needs to be any darker 👍 

As mentioned in the description, I'm working on fixing the TypeScript error. I've noticed that when you click the spotlight on a reflection and close the modal, the search icon no longer appears for that reflection. To fix, I think I need to either clone the reflection or move the position of that reflection to the modal. I'll look into how the reflection stack logic works in more detail. Either way, it's ready for a re-review @enriquesanchez when you get the chance

enriquesanchez:
@nickoferrall Thanks! This is looking :chef_kiss:

>I was umming and arring about aligning the card at the top or center but I think you're right about center alignment. 

I also like the center alignment. Looks great on big and small viewports:

image

image

>The placeholder text in that screenshot is lighter than how it looks for me

Ah, this is a Firefox thing. By default it lightens the placeholder text. Tested on Chrome and the color contrast is great.

enriquesanchez (approved):

packages/client/components/SpotlightModal.tsx
```diff
@@ -0,0 +1,214 @@
...
+            <CloseIcon>close</CloseIcon>
+          </StyledCloseButton>
+        </TopRow>
+        <Content>
+          <DraggableReflectionCard
+            inSpotlight
+            isReadOnly
+            staticIdx={0}
+            reflection={reflection as any}
+            meeting={meeting as any}
```
nickoferrall, line 152: I've been having problems getting the right types for the meeting and reflection. It looks like we may have had a similar issue in [ExpandedReflection](https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/client/components/RetroReflectPhase/ExpandedReflection.tsx#L15). Ideally, I'd like to just query the bare minimum fields we need in the file, the reflection & meeting ids, but I'm having difficulty as TypeScript complains about missing refs. I think this is because there's a circular reference: the reflection card contains the spotlight modal which contains the reflection card

packages/client/components/ReflectionGroup/ReflectionGroup.tsx
```diff
@@ -116,8 +116,10 @@ const ReflectionGroup = (props: Props) => {
   })
   const onClick = () => {
     if (isEditing) return
-    const wasDrag = staticReflections.some((reflection) => reflection.isDropping)
-    if (wasDrag) return
+    const isNotDraggable = staticReflections.some(
+      (reflection) => reflection.isDropping || reflection.inSpotlight
```
nickoferrall, line 120: If you have a stack of reflections, click the `showSpotlight` icon and close the spotlight modal, the reflection stack modal will then open up when we don't want it to. Using `stopPropagation` when clicking the `showSpotlight` icon doesn't prevent it from happening so it looks like we need to prevent it in the `onClick` here. 

To do this, I've added an `inSpotlight` field to the Relay client. This seems like overkill & is confusing as we also need to pass an `inSpotlight` prop to the `draggableReflection` in the modal so the `showSpotlight` icon doesn't show (Relay doesn't update in time to work in both). I'll work on finding a better solution but any suggestions would be great!

packages/client/components/ReflectionCard/ReflectionCard.tsx
nickoferrall: Makes sense - appreciate the explanation!

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -179,41 +224,89 @@ const ReflectionCard = (props: Props) => {
...
+          onClick={openSpotlight}
+          onMouseEnter={openTooltip}
+          onMouseLeave={closeTooltip}
+          showSearch={showSearch}
+        >
+          <SearchIcon ref={tooltipRef} icon='search' />
+        </SearchButton>
+        {tooltipPortal('Find similar')}
+      </ReflectionCardRoot>
+      {modalPortal(
```
mattkrick, line 298: let's move this logic to the RetroGroupPhase or GroupingKanban. then we can pass down something like:
```ts
const openSpotlight = (reflection: Reflection) => {
  commitLocalUpdate(atmosphere, (store) => store.get(meetingId).setLinkedRecord(reflection, 'spotlightReflection')
})
```
then we only open the modal if it was falsy & now it's a reflection object.
that'll express that there can only be 1 spotlight open & a spotlight takes in a reflection.

packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx
```diff
@@ -88,7 +92,7 @@ const DraggableReflectionCard = (props: Props) => {
     staticReflectionCount,
     swipeColumn
   )
-  const isDragPhase = phaseType === 'group' && !isComplete
+  const isDragPhase = phaseType === 'group' && !isComplete && !inSpotlight
```
mattkrick, line 95: we'll want the cards to be draggable inside the spotlight & if someone else is dragging the card it should be moving around in the spotlight, too

nickoferrall: Makes sense. Just to make sure we're on the same page with how the spotlight dragging behaviour will work: 

- if the viewer opens the spotlight modal, the spotlight modal only opens for the viewer, not for everyone in the meeting
- if someone that is not the viewer drags the reflection that's in the spotlight, the reflection would drag as normal for the non-viewer
- if someone that is not the viewer drags the reflection that's in the spotlight into a new reflection group, nothing would change in the spotlight for the viewer. If the viewer then drags the spotlight reflection into one of the reflection groups in the spotlight, the reflection would move to the new reflection group
- if the viewer moves the spotlight reflection to a reflection group in the spotlight, non-viewers would see the remote reflection move from it's original position on the kanban to its new destination

mattkrick: > if someone that is not the viewer drags the reflection that's in the spotlight into a new reflection group, nothing would change in the spotlight for the viewer.

the opposite for this one. if you've got spotlight open and see the card "Cats", I can drag that around. if i hover a group in your spotlight, it you see that. if i hover a group not in your spotlight, you just see it floating somewhere. if i release the mouse, you see the card go back to its spot. if i drop it on a group in your spotlight, you see that group grow in size.

nickoferrall:
@mattkrick as #5271 builds on top of the work here (and makes a few small changes to the code in this PR) would you be happy for me to close this PR? Once Enrique approves #5271, you could just review the code in that PR instead of both this PR and that one?

mattkrick:
sounds like a good plan!

Issue #5102, opened 2021-07-02
TypeError: window.localStorage.deleteItem is not a function


tiffanyhan:
Since 4 hours ago, affected 5 users in 5 events.  Looks like the function should be `removeItem`.

https://sentry.io/organizations/parabol/issues/2480476448/?referrer=github_plugin

```
TypeError: window.localStorage.deleteItem is not a function
  at d (./packages/client/components/AnalyticsPage.tsx:79:27)
...
(4 additional frame(s) were not displayed)
```

PR #5223, opened 2021-07-26
fix nonexistent method

tiffanyhan:
Resolves https://github.com/ParabolInc/parabol/issues/5102


tiffanyhan:
merging immediately as it's trivial

PR #5226, opened 2021-07-27
deprecate old estimate stage

### LINKED ISSUE #5163: EstimateStage: deprecate service, serviceTaskId
Predecessors: #5161
Now, the EstimateStage should no longer use `service, serviceTaskId, creatorUserId`

AC
- [x] keep creatorUserId but don't use it to determine access rights to the task. that should live in task.integration.accessUserId
- [x] deprecate EstimateStage.service & clear all references to it
- [x] EstimateStage.serviceTaskId is deprecated, too
- [x] on the frontend, nothing uses the 3 fields listed above (just comment out the fields the deprecated fields on the server & then whe re-running the relay compiler it will show you all the places it is used)
- [x] EstimateStage.story is replaced with .task
- [x] the Story GraphQL type is removed
- [x] title field is removed from Task
- [x] JiraCreateIssue & its payload are deleted
- [x] ~updatePokerScope doesn't accept service as an UpdateInput param~ will do this in next PR
- [x] estimate stage taskId is nonnull
Estimate: 16 hours

mattkrick:
merge with #5225

fix #5163

Now that all meetings old & new use the new fields, we can remove the legacy stuff

Burndown:
- [ ] clean up serviceTaskId
- [ ] refactor serviceField (i think this lives on the task.integration JIraIssue


mattkrick:
closing in favor of #5235

PR #5227, opened 2021-07-27
Ironbank mitigation fixes for @ Parabol version 6.22.0

adaniels-parabol:
Resolves node security issues related to https://github.com/ParabolInc/action-devops/issues/64#issuecomment-887325682 (see attachment)

Note for reviewer: there are three typescript errors currently, that may need to be addressed.


packages/server/package.json
```diff
@@ -56,11 +56,11 @@
     "jsdom": "^15.1.1",
     "json-loader": "^0.5.7",
     "lint-staged": "^10.1.7",
-    "mocha": "^5.2.0",
+    "mocha": "^9.0.3",
     "mockdate": "^2.0.1",
     "prettier": "^1.19.1",
-    "relay-compiler": "^8.0.0",
-    "relay-compiler-language-typescript": "^13.0.2",
+    "relay-compiler": "^11.0.2",
```
mattkrick, line 62: oooo this might break stuff. i'll have to test to make sure everything compiles correctly still

adaniels-parabol: Everything seem'ed to still work as long as I didn't also upgrade relay-config. (I know we talked about this yesterday, but wanted to add it to the comments)

adaniels-parabol: Disregard, you were right it did break when I upgraded relay-compiler.

adaniels-parabol: I upgraded all relay-compiler, to include relay-config and relay-runtime and this resolved the issues.

adaniels-parabol:
@mattkrick this is ready again for review.

packages/server/package.json
```diff
@@ -56,13 +56,13 @@
...
     "mockdate": "^2.0.1",
     "prettier": "^1.19.1",
-    "relay-compiler": "^8.0.0",
-    "relay-compiler-language-typescript": "^13.0.2",
-    "relay-config": "^8.0.0",
-    "relay-runtime": "https://github.com/mattkrick/relay/tarball/ea411fa5aa507b9c1082f12e6111bdea7186c817",
+    "relay-compiler": "^11.0.2",
+    "relay-compiler-language-typescript": "^14.0.0",
+    "relay-config": "^11.0.2",
+    "relay-runtime": "^11.0.2",
```
mattkrick, line 65: this will break our subscription caching & possible a few other things, we'll probably have to fork again to upgrade

adaniels-parabol: Is this a simple fix, if you could point me at what needs to change, I may be able to help?

packages/server/package.json
```diff
@@ -137,7 +137,7 @@
     "string-similarity": "^3.0.0",
     "stripe": "^4.24.0",
     "tslib": "2.0.3",
-    "uWebSockets.js": "https://github.com/tiffanyhan/uWebSockets.js/tarball/2e733751ce4765c23ca96e3e57265069536ac173"
+    "uWebSockets.js": "https://github.com/adaniels-parabol/uWebSockets.js/tarball/binaries"
```
mattkrick, line 140: yarn is too aggressive in their caching so if we install the next version from a tarball it'll still use the old version this way.
i think we only used a fork because the version we had was missing the right typings.

i bet we could revert back to master now: `yarn add uWebSockets.js@uNetworking/uWebSockets.js#v19.3.0`

packages/server/email/MailManagerMailgun.ts
```diff
@@ -1,15 +1,22 @@
...
 import MailManager, {MailManagerOptions} from './MailManager'
 
 export default class MailManagerMailgun extends MailManager {
-  mailgun = new Mailgun({apiKey: process.env.MAILGUN_API_KEY, domain: process.env.MAILGUN_DOMAIN})
+  //mailgun = new Mailgun({apiKey: process.env.MAILGUN_API_KEY, domain: process.env.MAILGUN_DOMAIN})
+  mailgun = new Mailgun(FormData);
+  mailgunClient = this.mailgun.client({
+    username: 'api',
+    key: process.env.MAILGUN_API_KEY || '',
+    public_key: process.env.MAILGUN_PUBLIC_KEY
```
mattkrick, line 12: looks like we only need a public_key if we use `parse` or `validate` (https://github.com/mailgun/mailgun-js#setup-client)
we don't use either of those, so we're good!

we could either add in the public key to all our .env files in action-devops and all our dokku instances, or we could just leave this out. i'd opt for leaving it out to keep it simple

packages/server/package.json
```diff
@@ -56,13 +56,13 @@
     "jsdom": "^15.1.1",
     "json-loader": "^0.5.7",
     "lint-staged": "^10.1.7",
-    "mocha": "^5.2.0",
+    "mocha": "^9.0.3",
```
mattkrick, line 59: we don't use this anymore, so we could remove it entirely

packages/server/package.json
adaniels-parabol: @mattkrick I reverted per your guidance, thank you sir!

packages/server/email/MailManagerMailgun.ts
adaniels-parabol: @mattkrick made the change per your guidance, thank you sir!

packages/server/package.json
adaniels-parabol: @mattkrick will take a card to setup a time to prune old/not-needed modules in the future. Leaving this for now.

packages/client/package.json
```diff
@@ -101,7 +101,7 @@
     "react-transition-group": "^4.3.0",
     "react-virtualized": "^9.21.1",
     "regenerator-runtime": "^0.13.5",
-    "relay-runtime": "https://github.com/mattkrick/relay/tarball/ea411fa5aa507b9c1082f12e6111bdea7186c817F",
+    "relay-runtime": "https://github.com/mattkrick/relay/tarball/ea411fa5aa507b9c1082f12e6111bdea7186c817",
```
mattkrick, line 104: missing F in the hash?

adaniels-parabol: @mattkrick I did a check, looks like without F is the proper URL

https://codeload.github.com/mattkrick/relay/legacy.tar.gz/ea411fa5aa507b9c1082f12e6111bdea7186c817F 404 Not Found

https://codeload.github.com/mattkrick/relay/legacy.tar.gz/ea411fa5aa507b9c1082f12e6111bdea7186c817 Downloads tarball

mattkrick: well this is interesting! i must've goofed on this & then the resolution in https://github.com/ParabolInc/parabol/blob/master/package.json saved us :sweat_smile:

mattkrick:
@adaniels-parabol looks like a merge conflict with the lockfile. i know you did some magic to it so I won't just rebuild it.

I imagine this issue might come up in the future where we'll want to add a package & then have to re-prune the lockfile.
Could you share the steps you did to prune it? We could add that package as a dependency & then write an npm script so we can perform it ad-hoc

mattkrick:
Reminder: @mattkrick fix the typecheck issues on this

Issue #5228, opened 2021-07-27
Unable to archive poker meetings


nickoferrall:
Poker meetings don't currently display the archive meeting button; only retro and action meetings show this button

@drewhousman1010 mentioned this bug [here](https://parabol.slack.com/archives/C4JAUUZ9P/p1627392064111800)

PR #5229, opened 2021-07-27
Add archive button to poker meeting timeline events

nickoferrall:
PR to resolve issue #5228


nickoferrall:
Merging as the fix is trivial

Issue #5203, opened 2021-07-22
Polls implementation: Database schema


BartoszJarocki:
This issue is part of Polls implementation: [EPIC #5147](https://github.com/ParabolInc/parabol/issues/5147)

Create a PG database schema for polls implementation

PR #5231, opened 2021-07-28
Polls DB schema

BartoszJarocki:
Related issue: https://github.com/ParabolInc/parabol/issues/5203

I'm also wondering if we should merge this before the feature is ready. I'm concerned this schema could change a bit and it'll require migrations after merging it.


packages/server/postgres/migrations/1627394777775_addPollsTable.ts
```diff
@@ -0,0 +1,58 @@
...
+          'active',
+          'editing',
+          'ended'
+        );
+      END IF;
+    END
+    $$;
+
+    CREATE TABLE IF NOT EXISTS "Poll" (
+      "id" VARCHAR(100) PRIMARY KEY NOT NULL,
```
BartoszJarocki, line 27: What if we use https://www.postgresql.org/docs/current/datatype-uuid.html here?

packages/server/postgres/migrations/1627394777775_addPollsTable.ts
```diff
@@ -0,0 +1,58 @@
...
+      "createdAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+      "updatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+      "deletedAt" TIMESTAMP WITH TIME ZONE DEFAULT NULL, 
+      "createdById" VARCHAR(100) NOT NULL REFERENCES "User"("id"),
+      "discussionId" VARCHAR(100) NOT NULL REFERENCES "Discussion"("id"),
+      "winningOptionId" VARCHAR(100),
```
BartoszJarocki, line 33: This probably should have `REFERENCES "PollOption"("id")` but the only option to do so is to alter the table after its creation. @mattkrick any thoughts on this?

packages/server/postgres/migrations/1627394777775_addPollsTable.ts
mattkrick: today we just use a snowflake as it's shorter than a UUID.
UUIDs work, they're just a little longer, so it's not very pretty if the entity is in the URL (or if we have to debug something).

The norm is `"id" INT GENERATED BY DEFAULT AS IDENTITY` and then just prefix the GraphQL ID with the name of the table. sequences do give away a little bit of data about how many records we have though. For example, a competitor could create a new poll, look at the ID & figure out approximately how popular the feature is. we could avoid that if we encoded it with a secret, but then we're back to the long IDs again

packages/server/postgres/migrations/1627394777775_addPollsTable.ts
mattkrick: i'd just remove this field for now as it's a 2nd source of truth. 
if we wanted to, we could add `winningOptionId` to the graphql schema & compute it by fetching all the poll options & returning the one with the most voteUserIds

packages/server/postgres/migrations/1627394777775_addPollsTable.ts
```diff
@@ -0,0 +1,58 @@
...
+      "title" VARCHAR(3000),
```
mattkrick, line 34: this feels a little long. maybe 300? I'm AOK with the DB limits being bigger than application limits, but i can't imagine application limits being much more than ~200 or so chars

packages/server/postgres/migrations/1627394777775_addPollsTable.ts
```diff
@@ -0,0 +1,58 @@
...
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  const client = new Client(getPgConfig())
+  await client.connect()
+
+  await client.query(`
+    DO $$
+    BEGIN
+      IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'PollStatusEnum') THEN
+        CREATE TYPE "PollStatusEnum" AS ENUM (
+          'creating',
```
mattkrick, line 17: i hadn't considered keeping the status on the poll in the DB!
thinking through it, I'm a little shy of putting editing state in our persisted DB because if i start to edit & then my browser crashes there's no TTL to reset the state. Years ago we did this with meeting facilitators disconnecting-- when the socket disconnected, we looked at all their meetings & reset the state. That's a lot of querying to do on every disconnect!

Here's an imperfect alternative that we could compare it with:
- replaced `status` with `endedAt` on the poll. a poll is either active or ended.
- refactor `EditCommentingMutation(isEditing, discussionId)` to something more generic like `EditDiscussion(isEditing, discussionId, editType: 'comment' | 'poll')`. 
- Georg's PR will persist that editing status in redis & we'll put a TTL on it to make sure nothing gets stuck
- in the clientSchema, we'll have a `newPoll: Poll` type on the discussion thread. then, when you click "create" it'll call `createPoll(newPoll)` that includes all the poll options

i may have forgotten some stuff. let me know your thoughts

mattkrick: maybe instead of editType we have `threadableId`
then on the client we could determine the threadable type from the threadableId (store.get('threadableId').getType())
that way if someone is creating a new poll, we could just say `threableId: null` and that would mean "creating" instead of "editing"

mattkrick:
just a couple thoughts. for sure let's get this merged to master as soon as it's ready (we may already have to change the name of the migration)

packages/server/postgres/migrations/1627394777775_addPollsTable.ts
BartoszJarocki: Makes sense. Thank you for all the details! I'll change it to `"id" INT GENERATED BY DEFAULT AS IDENTITY` 👍

packages/server/postgres/migrations/1627394777775_addPollsTable.ts
BartoszJarocki: The initial idea was to calculate the final result once, when the poll is ended, to avoid querying and calculating that for every user, every time the discussion thread is fetched. I was especially worried if that'd cause some performance issues when there's a lot of people in the meeting. Any thoughts on this @mattkrick? If that's ok, then I'm all in for removing this field! 👍

packages/server/postgres/migrations/1627394777775_addPollsTable.ts
BartoszJarocki: Good catch! I'll lower both. Poll title to 300 and the poll option title to 100 (designs show it can't be longer than 30, but just to keep it safe) 👍

BartoszJarocki:
@mattkrick I've pushed changes related to ids, removed the status, and limited the titles len. When I was going through the implementation of tasks and comments I've noticed they have a few more fields like `meetingId`, `teamId`, `threadSortOrder`. I'm not exactly sure how `threadSortOrder` is used, but I def think `meetingId` and `teamId` would be nice to have in `Polls` table, any thoughts on this? Thank you!

packages/server/postgres/migrations/1627394777775_addPollsTable.ts
BartoszJarocki: Yeah, good point! 
For now, I've added `endedAt` just to be able to _finish_ the poll. 

Until the user clicks `start` I _think_ everything should be local, e.g. poll should be added to the database only when a user added all the options and clicked `start`. When a user starts adding a poll it's best to use `EditCommentingMutation` and notify everybody that `user X creates a poll`. Same with editing.

BartoszJarocki:
@tiffanyhan I'd appreciate the review! Thank you! 🙏

BartoszJarocki:
I've added `teamId` and `meetingId` to the polls table. `meetingId` can't be the foreign key as we don't have a meetings table yet.

tiffanyhan (approved):
lgtm!

BartoszJarocki:
thank you @tiffanyhan! Moving to maintainer review, @mattkrick it'd be great if you could take a look 🙏

mattkrick:
LGTM!

PR #5236, opened 2021-07-29
Persist commenters in redis

### LINKED ISSUE #5022: Redis: Persist discussion commentors
Today, discussion thread commentors are all managed via local updates. that means someone joining a discussion late won't actually see folks actively commenting until they stop & start again.

This is a good PR for someone who would like to learn more about redis & subscriptions

AC
- [ ] commentors is persisted in redis,  just like `pokerHover` is used today
- [ ] the editCommenting mutation logic is removed because it's all handled magically now

Estimate: 8 hours

Dschoordsch:
Persist the commenters in redis and add the commenters as field to
Discussion. The field is also part of the result of
EditCommentingSuccess now, so it's updated automatically in the client.

Resolves #5022


mattkrick:
code review complete, looks great! testing now...

mattkrick:
this is great. i do love it when PRs can add functionality AND remove more code than they add.
merged!

PR #5237, opened 2021-07-29
Filter Parabol domain from watchlist

nickoferrall:
While it might be helpful to record Parabol sessions for debugging, I think it's probably best if we don't record them to avoid snooping. Plus, as we're power users, this will free up some extra sessions.

### DIFF
--- a/packages/client/components/AnalyticsPage.tsx
+++ b/packages/client/components/AnalyticsPage.tsx
@@ -78,11 +78,12 @@ const AnalyticsPage = () => {
     const expiredErrorProne =
       errorProneAt && new Date(parseInt(errorProneAt)) < new Date(Date.now() - ms('14d'))
     const email = window.localStorage.getItem(LocalStorageKey.EMAIL)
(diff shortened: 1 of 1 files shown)


mattkrick:
this would only record if we hit an error or we turn on `isWatched`, right?
IMO, since we're power users & use the app more than most, we probably bump into bugs more frequently so it might actually be a good thing to log us! in the past when i hit a bug i kicked myself for not having devtools open when the bug occurred, this would eliminate that tension so I'm AOK with leaving us in, but i'm curious for your thoughts!

nickoferrall:
Makes sense! I was thinking that info might be shared that people would want to keep private, e.g. a 1-1 with a manager where someone shares personal information or meetings where we discuss whether we want to hire someone. 

But keeping the session recordings will certainly help with debugging and I'm happy to stick with the current, more transparent approach and trust that no one would be snooping anyway

mattkrick:
we're all superusers, if we wanted to we could all snoop on each other in the DB or GraphQL level anyways :sweat_smile:

PR #5240, opened 2021-07-30
Accordion v2

### LINKED ISSUE #4902: Add accordion to check-in left nav phase change

mattkrick:
@nickoferrall I think this fixes the animations without using a maxheight. let me know if it works for ya.
this is just a rough draft, feel free to merge in any other fixes. hit me up if there's any breaks that require breaking encapsulation & we'll work those out in sync

fix #4902
https://user-images.githubusercontent.com/5514175/127711378-d2df3bd2-a905-49fc-9c71-501108608fea.mp4

cleans up #5086

Burndown:
- [ ] support all meeting types
- [ ] fix TS errors


packages/client/components/ActionMeetingSidebar.tsx
```diff
@@ -57,30 +59,31 @@ const ActionMeetingSidebar = (props: Props) => {
...
-                isFacilitatorPhase={phaseType === facilitatorPhaseType}
-                isUnsyncedFacilitatorPhase={
-                  isUnsyncedFacilitatorPhase && phaseType === facilitatorPhaseType
-                }
-                isUnsyncedFacilitatorStage={isUnsyncedFacilitatorStage}
-                key={phaseType}
-                phaseCount={phaseCount}
-                phaseType={phaseType}
-              >
+              <Fragment key={phaseType}>
```
mattkrick, line 62: The header is now a peer of the list items.
This is good because the DOM tree is a little more shallow, but also it means we can set minHeight on the list item instead of having to set it on the parent, which means fewer props to pass down

packages/client/components/ActionSidebarAgendaItemsSection.tsx
```diff
@@ -1,12 +1,16 @@
...
 import {createFragmentContainer} from 'react-relay'
-
-import {ActionSidebarAgendaItemsSection_meeting} from '../__generated__/ActionSidebarAgendaItemsSection_meeting.graphql'
 import useGotoStageId from '../hooks/useGotoStageId'
 import AgendaListAndInput from '../modules/teamDashboard/components/AgendaListAndInput/AgendaListAndInput'
+import {ActionSidebarAgendaItemsSection_meeting} from '../__generated__/ActionSidebarAgendaItemsSection_meeting.graphql'
 import MeetingSidebarPhaseItemChild from './MeetingSidebarPhaseItemChild'
 
+const StyledRoot = styled(MeetingSidebarPhaseItemChild)({
+  minHeight: 58
```
mattkrick, line 11: height is undefined for the agenda item list so it will just take up all remaining space

packages/client/components/MeetingNavList.tsx
```diff
@@ -7,7 +7,8 @@ const MeetingNavList = styled('ul')({
   listStyle: 'none',
   margin: 0,
   minHeight: 0, // very important! allows children to collapse for overflow
-  padding: 0
+  padding: 0,
+  overflow: 'auto'
```
mattkrick, line 11: the sidebar respects the parabol logo footer

packages/client/components/NewMeetingSidebarPhaseListItem.tsx
```diff
@@ -84,6 +74,7 @@ const NavListItemLink = styled('div')<LinkProps>(
     minHeight: 40,
     textDecoration: 'none',
     userSelect: 'none',
+    fontWeight: 600,
```
mattkrick, line 77: got rid of NavListItem & put the styles on the Link

packages/client/hooks/useAnimatedPhaseListChildren.ts
```diff
@@ -0,0 +1,25 @@
+import {useEffect, useRef, useState} from 'react'
+
+const useAnimatedPhaseListChildren = (isActive: boolean, itemCount: number) => {
+  const [height, setHeight] = useState(isActive ? 'fit-content' : 0)
```
mattkrick, line 4: made this fit-content so the initial measurement doesn't measure the entire height, just what it needs

packages/client/components/MeetingNavList.tsx
nickoferrall: Nice, I think this is the key change! I was trying to get the nav list to have a fixed height and make the individual phases scrollable but making the entire list scrollable works nicely. Thanks a lot for your help!

nickoferrall:
@mattkrick this PR is ready for a review (can't add you as a reviewer as you're the author)

mattkrick:
discuss & icebreaker are broken in the retro meeting. clicking the discuss phase does not show the children.
Screenshot from 2021-08-11 12-36-24

can you write out a checklist of manual tests to perform for the next review?

nickoferrall:
@mattkrick ah did you pull the latest changes in this branch? I've tested both new and existing retro meetings in different browsers and it's working for me: 

retro-sidebar

Checklist:

- [ ] In a check-in meeting, clicking any of the phase titles in the sidebar, e.g. Solo Updates, causes the children to animate in like an accordion 
- [ ] In a check-in meeting, if there's not enough space to show all of the members in the check-in or updates phase, the sidebar navList has a vertical scrollbar 
- [ ] In a check-in meeting, if there's not enough space to show the agenda items children, the "Add Agenda Topic" input is still visible
- [ ] In a retro, clicking the Icebreaker row in the sidebar causes the children to animate in like an accordion 
- [ ] In a retro, the Discuss phase children are not visible until the vote stage is complete
- [ ] In a retro, if there's not enough space to show all of the members in the check-in phase, the sidebar navList has a vertical scrollbar 
- [ ] In a poker meeting, clicking the Icebreaker row in the sidebar causes the children to animate in like an accordion 
- [ ] In a poker meeting, if there's not enough space to show all of the members in the check-in phase, the sidebar navList has a vertical scrollbar 
- [ ] The sidebar will animate in & out for non-facilitators when they click on a different phase title or when the facilitator changes the phase

nickoferrall:
@mattkrick it looks like there's an issue in `master` where it's not possible to view the check-in children when there are many discussion, agendaitems or tasks in a meeting. I don’t think that there have been any changes to the sidebar in master lately so perhaps it’s flown under the radar for a while. I created an issue [here](https://github.com/ParabolInc/parabol/issues/5262) - this accordion PR fixes that issue for me

mattkrick:
whoaaaaaa, good find! definitely worth tackling that at the same time

mattkrick:
wonderful! would you mind moving it through to the next column & grabbing a reviewer?

nickoferrall:
@BartoszJarocki would you mind reviewing this one? You can find the testing steps in [this comment](https://github.com/ParabolInc/parabol/pull/5240#issuecomment-897138296)

BartoszJarocki:
@nickoferrall I'm happy to do it. will be done by EOD.

packages/client/components/ActionMeeting.tsx
```diff
@@ -72,6 +72,7 @@ const ActionMeeting = (props: Props) => {
         />
       </ResponsiveDashSidebar>
       <MeetingArea>
+        <Suspense fallback={''}>
```
BartoszJarocki, line 75: Just out of curiosity, is suspense actually working with Relay? 😅  if yes, do we want a nicer fallback like progress?

packages/client/components/ActionMeeting.tsx
```diff
@@ -72,6 +72,7 @@ const ActionMeeting = (props: Props) => {
...
         <Phase
           meeting={meeting}
           toggleSidebar={toggleSidebar}
```
BartoszJarocki, line 78: I noticed eslint/prettier errors about formatting here, it'd be great to fix that before merging 🙏

packages/client/hooks/useAnimatedPhaseListChildren.ts
```diff
@@ -0,0 +1,25 @@
...
+  }, [itemCount])
+
+  const wasActiveRef = useRef(isActive)
+  useEffect(() => {
+    if (isActive === wasActiveRef.current) return
+    if (isActive) {
+      ref.current!.style.overflow = 'hidden'
+      setTimeout(() => {
+        ref.current!.style.overflow = ''
+      }, 300)
```
BartoszJarocki, line 18: I have two minor suggestions
1. It'd be nice to clear the timeout in effect cleanup just to avoid some edge cases crashing, especially considering we're assuming the ref is always there 
2. Do we want to extract `300` as some kind of const to use in similar scenarios? Naming this would also be helpful in understanding why do we need the timeout in the first place 🙏

BartoszJarocki (commented):
Great work! I've tested on Chrome/Firefox/Safari/Edge and it's working properly in all cases. 
I've left a few minor comments to consider, other than that it's good to go! 👌

packages/client/components/ActionMeeting.tsx
nickoferrall: Yep, the phase components are being lazily loaded which meant that the height calculation was initially incorrect. We're using the empty string fallback option in quite a few components - I think a skeleton UI fallback could work well but I'd prefer to get this PR merged and refactor that in a different issue

packages/client/hooks/useAnimatedPhaseListChildren.ts
nickoferrall: Thanks for the review! Makes sense - I've just pushed an update 👍

nickoferrall:
Moving to the Maintainer Review column and tagging in @mattkrick

mattkrick:
:tada: :tada: :tada: 

fantastic work! thanks for sticking with this one, i know it went a lot longer than we expected, but I can already see that the things you learned in here are paying off in the spotlight PR

nickoferrall:
Thanks Matt! For sure - really appreciate your coaching on this one

Issue #5263, opened 2021-08-11
Grouping kanban columns are not centered on Safari


nickoferrall:
<img width="1433" alt="safari-kanban-center" src="https://user-images.githubusercontent.com/39854876/129115563-de553fc0-f92a-4e71-8b44-d0c56b30e6b5.png">

On Safari, the grouping kanban columns that are used in the Group and Vote stage of a Retro meeting are left aligned whereas they're centered on Chrome.

mattkrick:
this is a great find!

> For developers, Safari is crap and outdated

https://news.ycombinator.com/item?id=27968394

I'm _real_ close to considering that we just don't worry about safari. it has become the IE of our decade

nickoferrall:
It's strange that Apple doesn't seem to prioritise it. The Reddit quotes in [that article](https://blog.perrysun.com/2021/07/15/for-developers-safari-is-crap-and-outdated/) made me laugh 😄

PR #5264, opened 2021-08-11
Center align kanban columns on safari

nickoferrall:
PR to resolve issue #5263 

This PR center aligns the kanban columns on Safari to be consistent with Chrome.

@enriquesanchez mind reviewing this one? It's a tiny change so if you're happy it, I think we can merge it without a second code review


enriquesanchez (approved):
Works as expected 👍 🎉

Issue #5251, opened 2021-08-05
Jira integration - 'Fix it for me' doesn't work


BartoszJarocki:
## Issue - Bug

When using the 'fix it for me' functionality the `submitScore` isn't called. It results in not updating the estimate once the field is successfully added.

### Acceptance Criteria (optional)
- [ ] error message is cleared correctly
- [ ] submit score is called, estimate is update

**Estimated effort:** 8 hours

mattkrick:
bumping to p2, feels bad that we have a "fix it for me" button that doesn't fix it.
if we wanted to make it a p1 i'd be all for it.

BartoszJarocki:
@mattkrick I'm happy to jump in and fix it. I think it looks really bad so I'd love to have that working.

mattkrick:
sounds good! should be a pretty quick fix, regression tests will be rough. this is probably a great area to intro unit tests in the future

PR #5265, opened 2021-08-12
Jira integration - 'Fix it for me' doesn't work

BartoszJarocki:
## Cause 
This PR fixes the behavior of adding missing Jira fields and updating the estimation afterward. The reason, why it's not working currently is this line
```
const submitScore = () => {
   if (submitting || !isStale || !isLocallyValidatedRef.current) return
```
It works in all cases with the exception of handling missing Jira field error. In that case, `isStale` is `false` and as a result, nothing happens when `submitScore` is called.

## Fix
In general, as the existing functionality is pretty complex and I didn't want to break it, I've made the following changes:
1. `submitScore` function has now another parameter, `forceRetry?: boolean` which is only used after we attempt to update the estimate after fixing the missing Jira field. That way, I haven't touched anything related to current functionality.
2. Instead of using `submitScore` directly as `onClick` handler, I created 2 helper functions
```
  const submitCardScore = () => {
    lastCardScoreRef.current = cardScore
    submitScore(cardScore)
  }

  const submitLastCardScore = () => {
    submitScore(lastCardScoreRef.current, true)
  }
```
`submitCardScore` is used in the exact same way as `submitScore` was used previously. `submitLastCardScore` is used after we fix the missing Jira field, and it's used with `forceRetry = true`. 

`cardScore` is reset after an unsuccessful call of `PokerSetFinalScore` so I'm added a reference `lastCardScoreRef` which basically helps me to remember the value we previously tried to set. It's updated whenever someone clicks `Update`. 

## Other
I've also found one more thing, see the screenshot below:
Screenshot 2021-08-12 at 15 45 05
It seems like we have to respect the type of updated field, otherwise, Jira will complain. Commit that fixes it: https://github.com/ParabolInc/parabol/commit/f1e067be09b4c38f0350d033aa60c1e20198e69c 

## Testing
The most important thing is to check if nothing that already exists in the sprint poker related to estimation didn't break. 
- [ ] Update the estimate in fields like a summary
- [ ] Update the estimate in a field that isn't configured in Jira (eg. Story points); Jira fields are configured per `screen` so for example, in my own project I can see all the fields available for the project issues here: https://bjarocki.atlassian.net/secure/admin/ConfigureFieldScreen.jspa?id=10000 (replace the `bjarocki` with your own login data)

**Important** 
`Fix it for me` functionality does not always work. Jira configuration might be really complex and not always we're able to fix it automatically for the user. In this case, the `Fix it for me` dialog should just show an error. 

Fixes: https://github.com/ParabolInc/parabol/issues/5251


BartoszJarocki:
@tianrunhe I'd appreciate the review! 🙏

tianrunhe (commented):
So I tried this locally. I tried to update the "Original estimate" field (which was not configured in screen in Jira), but got error and could not proceed. See screenshot [here](https://www.loom.com/share/e6a0b7b0a1694ca2866c3e8985869fa2)

BartoszJarocki:
```
Important
Fix it for me functionality does not always work. Jira configuration might be really complex and not always we're able to fix it automatically for the user. In this case, the Fix it for me dialog should just show an error.
```

@tianrunhe yeah, it can happen. Sometimes, it's not adding a field is not as easy as adding it to a screen so we're aware we won't be able to help automatically in all the cases. @mattkrick you mentioned a while ago this could be caused by something else in Jira config and I don't remember what was that, do you remember it? 🙏

mattkrick:
LGTM!

Issue #5181, opened 2021-07-19
Spotlight: Animate target reflection into place


mattkrick:
Predecessors: #5106

Now that a skeleton of the spotlight modal exists, let's animate the target reflection into place so the user sees that clicking the search icon triggers a search based on that particular reflection.

AC
- [ ] when the search icon is clicked on a card, the reflection moves from it's place in the column into the top center of the new modal

Suggestions:
- This is a perfect place to use FLIP animations & a pattern is already in place with the expandedReflectionGroups component.
- When the modal is open, the top card is set to opacity 0
- the location of the card in the column is taken (probably via getBoundingClientRect, but offsetTop/left may be necessary depending on the parent element)
- the location of the target card is set via a transform to the bounding box of the card in the column
- set a transition on the card's transform so it takes 300ms to move into place
- experiment with fading in the entire modal after the card moves into place. we don't want the whole transition to take 500ms, but we want it to feel like the card is the thing that is creating the modal

Estimate: 8 hours

nickoferrall:
Hey @mattkrick, for this issue, are you thinking that we'll use the [useExpandedReflections](https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/client/hooks/useExpandedReflections.ts) hook for the spotlight modal or we'll just borrow ideas from `expandedReflectionGroups`? Just want to check before diving into the fairly complex logic in that hook

mattkrick:
just borrow ideas. if we finish the feature & realize we can abstract it we can consider it then, but let's keep it easy for now

nickoferrall:
Rough plan: 

- create a custom hook that uses a flip animation to move the reflection into the Spotlight modal
- it'll take the selected reflection ref and will adjust its style position 
- I'll have a look at the `useFlip` hook and see if I can either use it or borrow logic from it
- the modal will fade in halfway into the reflection animation 
- the reflection will have a larger zIndex than the modal so that the reflection is on top of the modal

Other notes about the Spotlight:

- the selected reflection at the top of the Spotlight modal is just one reflection rather than a reflection group
- when the viewer clicks the reflection to open the spotlight, that reflection will behave like it's in the dragging state to non-viewers, i.e. users won't be able to drag that reflection. This is because it would be confusing if a user drags the reflection to a different group just before or after the viewer moves it in the spotlight. When the spotlight modal is open, the user is in the process of moving the reflection so it's logical that it would be in the dragging state
[...]

PR #5271, opened 2021-08-13
Animate reflection into Spotlight

nickoferrall:
PR to resolve issue #5181 

This PR builds on #5214. 

@enriquesanchez would love to see what you think!


packages/client/components/SpotlightModal.tsx
```diff
@@ -0,0 +1,208 @@
...
+              placeholder='Or search for keywords...'
+              type='text'
+            />
+          </SearchItem>
+        </SelectedReflectionSection>
+        <SimilarReflectionGroups>
+          {reflectionGroupsCount === 0 ? <SpotlightEmptyState /> : null}
+        </SimilarReflectionGroups>
+      </ModalContainer>
+      <ReflectionWrapper ref={flipRef}>
```
nickoferrall, line 171: The modal needs to fade in after the reflection animation starts doing its thing so I'm absolutely positioning it outside of the modal

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -14,26 +14,52 @@ import useMutationProps from '../../hooks/useMutationProps'
...
+
+const SearchButton = styled(CardButton)<{showSearch: boolean}>(({showSearch}) => ({
+  bottom: 2,
+  color: PALETTE.SLATE_700,
+  cursor: 'pointer',
+  opacity: 1,
+  position: 'absolute',
+  right: 2,
+  visibility: showSearch ? 'visible' : 'hidden',
+  zIndex: ZIndex.TOOLTIP,
```
nickoferrall, line 52: It's not strictly a tooltip but hovering over the button triggers the tooltip. I went with the tooltip z-index as I didn't want to add another z-index to our collection

enriquesanchez (changes requested):
This looks sooooo good @nickoferrall 🙌

I was wondering, would it be possible to animate the original reflection? I see that currently the reflection gets cloned and the duplicate is then animated, and for a second you can see both exist at the same time in the board.

Would this be possible?

nickoferrall:
Thanks! 

For sure - I've just pushed an update that fades out the selected reflection. Also, when the Spotlight closes, the reflection fades back in. We could remove this but I kinda like it as it indicates where the Spotlight reflection was previously located. Would be interested to get your thoughts!

enriquesanchez:
@nickoferrall Oh yeah, much better 🙌

Something about the fade out of the modal and the immediate fade in of the reflection feels a bit off to me. As of the UI flashed. Do you notice that too? I wonder if a slight delay on the fade in of the reflection could help?

nickoferrall:
@enriquesanchez yeah I like that it indicates which reflection was in the spotlight but I can see how it might look like it's flashing. 

I played around with delaying the fade in but it had a similar feel so I've removed it. Now, when the spotlight opens, the selected reflection fades out but when the spotlight closes, the user sees the kanban without any transitions. What do you think?

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -179,8 +215,39 @@ const ReflectionCard = (props: Props) => {
...
+      openSpotlight(reflectionId, reflectionRef)
+      setSelectedForSpotlight(true)
+    }
+  }
+
+  useLayoutEffect(() => {
+    // the same reflection card can exist in the kanban and spotlight
+    // when the spotlight closes, make the kanban reflection visible again
+    if (!inSpotlight && selectedForSpotlight) {
+      setSelectedForSpotlight(false)
```
nickoferrall, line 232: Initially I implemented this by editing the ref styles directly but I think using state here is easier to follow

packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx
```diff
@@ -67,9 +68,10 @@ const DraggableReflectionCard = (props: Props) => {
   const {
     isClipped,
     reflection,
-    staticIdx,
+    staticIdx = 0,
```
nickoferrall, line 71: Dragging the reflection when it's in the Spotlight doesn't work yet - it clones the reflection. I think that's out of scope here and plan on tackling that in the issue where we need to drag the Spotlight reflection to the reflection groups. 

To resolve that, I plan on digging into `useDraggableReflectionCard` in more detail and seeing how `staticIdx` and `staticReflectionCount` are used

enriquesanchez:
>Now, when the spotlight opens, the selected reflection fades out but when the spotlight closes, the user sees the kanban without any transitions. What do you think?

@nickoferrall Ah, I still feel the flash without the fade in, so I think it's due to the scrim fading away so fast then.  I'd leave the fade in as it was then. Apologies for the trouble 🙈.

nickoferrall:
@enriquesanchez ahh OK I see what you mean. That's a great catch - I've just pushed a little change that should make the modal & scrim fade away a little slower.

For the fading in reflection when the Spotlight closes, I'll leave it how it is for now (i.e. it does not fade in). I can always re-add it if Matt feels it'd be helpful when it's his turn to review

enriquesanchez (approved):
I like the update. I can't notice the flash anymore 👍

mattkrick:
fix #5181

packages/client/components/SpotlightModal.tsx
```diff
@@ -0,0 +1,209 @@
...
+  meeting: SpotlightModal_meeting
+  flipRef: (instance: HTMLDivElement) => void
+}
+
+const SpotlightModal = (props: Props) => {
+  const {closeSpotlight, meeting, flipRef} = props
+  const {spotlightReflection} = meeting
+  const isDesktop = useBreakpoint(Breakpoint.NEW_MEETING_SELECTOR)
+  const handleKeyDown = (e: KeyboardEvent<HTMLDivElement>) => {
+    if (e.key === 'Escape') closeSpotlight()
```
mattkrick, line 140: ahhh, i'm the reason this is like it is.

the reason `onClose` in the parent doesn't fire is because the input field is currently active & we normally don't want to close modals where that is true: https://github.com/ParabolInc/parabol/blob/3abf58205fdebba284cc699f947807efed8b30aa/packages/client/hooks/usePortal.tsx#L86. A good example is when the expanded reflection stack is open & you're editing a card in it. when you hit escape, you want to finish editing that card, not close the whole stack.

In this case, we _do_ want to close no matter what.
maybe `usePortal` should also accept a `closeIgnoresFocus` boolean? that'd save us from having to introduce another keydown handler. whatcha think?

mattkrick (changes requested):
great work! just a couple things we can simplify & then it's good to merge!

packages/client/components/SpotlightModal.tsx
mattkrick: an alternative would be to put a keydown handler on the input element that would call blur() on escape. that way we could pick & choose which inputs could make it close. :shrug:

nickoferrall: Ahh interesting! 

If the user hasn't typed anything into the input and they hit escape, I think they prob want to close the modal as it's unlikely that you'd accidentally hit escape as the first key when typing.

If the user has started typing, is still focused on the input and they hit escape, it's possible that it was an accident or they want to close the modal. I think adding an onBlur works well in this instance because they'd have to hit escape a second time to close it.

In usePortal, I'm now checking the length of the element's value to determine whether we should blur or close it

packages/client/components/ReflectionEditorWrapper.tsx
```diff
@@ -169,7 +169,11 @@ class ReflectionEditorWrapper extends PureComponent<Props> {
         this.removeModal()
       } else {
         const el = this.props.editorRef.current
-        el && el.blur()
+        // add to callback queue so we can check activeElement and
+        // determine whether modal should close in expandedReflectionStack
+        setTimeout(() => {
+          el && el.blur()
```
nickoferrall, line 175: I noticed that in prod, when editing a reflection in the expanded reflection stack and hitting esc, the modal does actually close. This is because we've got a keydown event handler in [ExpandedReflectionStack](https://github.com/ParabolInc/parabol/blob/6cfe9aac9dd0ae342e1310b3edb45845516d389b/packages/client/components/RetroReflectPhase/ExpandedReflectionStack.tsx#L100) that closes the modal and we're passing and we're preventing the default close behaviour in usePortal [here](https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/client/hooks/useExpandedReflections.ts#L40).

I'm not in love with the solution, but one option is using this setTimeout so that we can see where the activeElement is in `ExpandedReflectionStack`. If it's focused on editing the card, just blur. Otherwise, close the modal

nickoferrall:
@mattkrick I'd love to get your thoughts on the updates I've made to handle esc keydowns before merging. I've switched the `showSpotlight` boolean to false in `ReflectionCard` so we can merge this PR to master if you're happy with it

PR #5273, opened 2021-08-13
Reset retro meeting to group phase without creating all the new phases

### LINKED ISSUE #5249: Bug: server error when voting, unable to advance meeting
## Issue - Bug

This user reported first reported #5248.

When investigating further, the user reported:

> We had several issues but I think the root cause was that we got stuck in the grouping phase
>  * Once we were finished with the Grouping Phase, I advanced the meeting to the Vote Phase. Only one person in the >. 
>  * meeting moved to the Vote Phase with me the rest were in the Grouping phase still.
>  * When the two of us tried to Vote we go a Server Error but if we refreshed our screen the vote would show up
>  * I could not advance to the discussion phase since no one could vote

The meeting ID is `5cVMwQLims`

One note: if attempt to revert the meeting to an incomplete state by removing the `endedAt` and `summarySentAt` fields from the `NewMeeting` record, server error is emitted to the client UI. I wasn't able to track down the actual server error.

Bug triaged:
   - Triaged and fixed or new issue to fix opened with root cause

**Estimated effort:** 4 hours to triage. More if root cause is already identified.

tianrunhe:
Fixes #5249

### Changes
1. Updates existing `resetMeetingToStage` to only works for resetting to group stage for a retro meeting. This is to ensure that we don't need to create dummy phases/stages for the meeting beforehand, hence we no longer rely on the `meetingSettings` which could change during the meeting period. See discussion in #5249 and [this thread](https://parabol.slack.com/archives/C836NA350/p1628885677080400?thread_ts=1628639020.029800&cid=C836NA350).
2. Due to the above changes, rename the mutation to `resetRetroMeetingToGroupStage` across the board, to avoid confusion.

### Tests

- [x] User can reset a retro meeting back to group phase ...
  - [x] from vote phase when voting has not begun
  - [x] from vote phase when voting has begun (verify that all vote counts have been reset)
  - [x] from discuss phase when there's only 1 discussion topic (verify that discussion agenda items have been cleared)
  - [x] from discuss phase when there are multiple discussion topics (verify that discussion agenda items have been cleared)
  - [x] any number of times freely
- [x] User can properly reset a retro meeting with icebreakers to group phase, even when there's a second retro meeting running without icebreaker (i.e., the reproduction step described in #5249)

### DIFF
--- a/packages/server/graphql/mutations/resetRetroMeetingToGroupStage.ts
+++ b/packages/server/graphql/mutations/resetRetroMeetingToGroupStage.ts
@@ -13,7 +13,7 @@ import MeetingRetrospective from '../../database/types/MeetingRetrospective'
 import {getUserId} from '../../utils/authorization'
 import publish from '../../utils/publish'
 import standardError from '../../utils/standardError'
-import ResetMeetingToStagePayload from '../types/ResetMeetingToStagePayload'
+import ResetRetroMeetingToGroupStagePayload from '../types/ResetRetroMeetingToGroupStagePayload'
 import {primePhases} from './helpers/createNewMeetingPhases'
 
 const resetMeetingPhase = (phase: GenericMeetingPhase) => {
@@ -28,8 +28,8 @@ const resetMeetingPhase = (phase: GenericMeetingPhase) => {
   phase.stages = newStages
 }
 
-const resetMeetingToStage = {
-  type: GraphQLNonNull(ResetMeetingToStagePayload),
+const resetRetroMeetingToGroupStage = {
+  type: GraphQLNonNull(ResetRetroMeetingToGroupStagePayload),
   description: `Reset a retro meeting to group stage`,
   args: {
     meetingId: {
@@ -136,9 +136,15 @@ const resetMeetingToStage = {
     const data = {
       meetingId
     }
-    publish(SubscriptionChannel.MEETING, meetingId, 'ResetMeetingToStagePayload', data, subOptions)
+    publish(
+      SubscriptionChannel.MEETING,
+      meetingId,
+      'ResetRetroMeetingToGroupStagePayload',
+      data,
+      subOptions
+    )
     return data
   }
 }
 
-export default resetMeetingToStage
+export default resetRetroMeetingToGroupStage
--- a/packages/server/graphql/mutations/resetRetroMeetingToGroupStage.ts
+++ b/packages/server/graphql/mutations/resetRetroMeetingToGroupStage.ts
@@ -1,14 +1,9 @@
 import {GraphQLID, GraphQLNonNull} from 'graphql'
 import {SubscriptionChannel} from 'parabol-client/types/constEnums'
-import findStageById from 'parabol-client/utils/meetings/findStageById'
-import {DISCUSS, GROUP, VOTE} from '../../../client/utils/constants'
+import {CHECKIN, DISCUSS, GROUP, REFLECT, VOTE} from '../../../client/utils/constants'
 import getRethink from '../../database/rethinkDriver'
 import DiscussPhase from '../../database/types/DiscussPhase'
-import DiscussStage from '../../database/types/DiscussStage'
-import GenericMeetingPhase, {
-  NewMeetingPhaseTypeEnum
-} from '../../database/types/GenericMeetingPhase'
-import GenericMeetingStage from '../../database/types/GenericMeetingStage'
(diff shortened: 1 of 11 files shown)


tiffanyhan (approved):
all tests passed. lgtm! really nice find and fix.

packages/server/graphql/mutations/resetRetroMeetingToGroupStage.ts
```diff
@@ -1,20 +1,36 @@
...
+import {primePhases} from './helpers/createNewMeetingPhases'
 
-const resetMeetingToStage = {
-  type: GraphQLNonNull(ResetMeetingToStagePayload),
-  description: `Reset meeting to a previously completed stage`,
+const resetMeetingPhase = (phase: GenericMeetingPhase) => {
+  const newStages = phase.stages.map(({id}) => {
+    const newStage =
+      phase.phaseType === DISCUSS
+        ? new DiscussStage({sortOrder: 0, reflectionGroupId: ''})
```
mattkrick, line 23: if there are 10 discussion topics, this will create 10 dummy stages. that probably won't create an error today, but that means future devs will have to be aware that there could be multiple dummy stages. let's instead just have 1 dummy stage

packages/server/graphql/mutations/resetRetroMeetingToGroupStage.ts
```diff
@@ -30,9 +46,12 @@ const resetMeetingToStage = {
 
     // AUTH
     const viewerId = getUserId(authToken)
-    const meeting = await dataLoader.get('newMeetings').load(meetingId)
+    const meeting = (await dataLoader.get('newMeetings').load(meetingId)) as MeetingRetrospective
     if (!meeting) return standardError(new Error('Meeting not found'), {userId: viewerId})
-    const {createdBy, facilitatorUserId, phases, teamId, meetingType, meetingCount} = meeting
+    const {createdBy, facilitatorUserId, phases, meetingType} = meeting
+    if (meetingType != 'retrospective') {
```
mattkrick, line 52: `stageId` is no longer necessary since this mutation will only reset to the group stage

packages/server/graphql/mutations/resetRetroMeetingToGroupStage.ts
```diff
@@ -56,43 +75,31 @@ const resetMeetingToStage = {
...
-      meetingCount,
-      meetingType,
-      dataLoader
-    )
-    const discussionIds = [] as string[]
-    let shouldResetStage = false
+    let discussionIds = [] as string[]
     let resetToPhaseIndex = -1
     const newPhases = [] as GenericMeetingPhase[]
     for (const [phaseIndex, phase] of phases.entries()) {
```
mattkrick, line 81: this feels a little too complex.
let's avoid mutating old objects & having a phaseType conditional (resetMeetingPhase) inside a phaseType switch/case.

```ts
    const newPhases = phases.map((phase) => {
      switch (phase.phaseType) {
        case 'checkin':
        case 'reflect':
          return phase
        case 'group':
        case 'vote':
          return new GenericMeetingPhase(phase.phaseType)
        case 'discuss':
          return new DiscussPhase([])
      }
    })

    const discussionIdsToDelete = phases.find((phase) => phase.phaseType === 'discuss')?.stages.map(({discussionId}) => discussionId)
    
```

packages/server/graphql/mutations/resetRetroMeetingToGroupStage.ts
```diff
@@ -0,0 +1,135 @@
...
+    if (!resetToStage) return standardError(new Error('Group stage not found'), {userId: viewerId})
+    if (!resetToStage.isNavigableByFacilitator)
+      return standardError(new Error('Group stage has not started'), {userId: viewerId})
+    if (!resetToStage.isComplete)
+      return standardError(new Error('Group stage has not finished'), {userId: viewerId})
+    if (meeting.endedAt)
+      return standardError(new Error('The meeting has already ended'), {userId: viewerId})
+
+    // RESOLUTION
+    const discussionPhase = phases.find((phase) => phase.phaseType === DISCUSS) as DiscussPhase
```
mattkrick, line 54: @nickoferrall built a great little helper functrion so we no longer have to cast! https://github.com/ParabolInc/parabol/blob/8118227c74f0df6e3da4146bec61dab3ddc02f4e/packages/server/utils/getPhase.ts

mattkrick:
looks good!

Issue #5182, opened 2021-07-19
Spotlight: similarReflectionGroups query


mattkrick:
Predecessors: none

We need a way to suggest which reflections would make good groups.
We'll do this server-side because in the future we may make this logic a little smarter, but for now it'll be dumb

AC
- [x] `similarReflectionGroups` exists on the graphql User object (open to alternative locations e.g. on the RetrospectiveMeeting object. discuss below)
- [ ] `similarReflectionGroups` accepts 2 args: `{reflectionId, searchQuery}` both are nonnull, but searchQuery could be an empty string
- [x] the resolver makes sure the reflectionId is valid, then makes sure that the viewer has permission to view the reflectionId based off the reflectionId.meetingId & the user has a meetingMember that shares the same meetingId
- [x] the return type is `[ReflectionGroup!]!` 
- [x] to start with, the query simply returns all the active reflectionGroups in the meeting (we'll handle better logic in the next issue) 
- [x] when the spotlight modal is opened, it triggers this query

Estimate: 8 hours

nickoferrall:
@mattkrick happy for me to work on this Spotlight issue next?

Rough plan / thoughts:

- I think it makes sense to add `similarReflectionGroups` to the User object. We could add it to the meeting object as it's related and we're currently querying the meeting and not the user object in the group phase. But, as Spotlight `similarReflectionGroups` are specific to only that user's view rather than everyone in the retro, I think the User object makes sense
- I plan on adding the query to the SpotlightModal component so that when it's opened, it triggers the query right away

mattkrick:
> The meetingId doesn't exist in the User source object so I suggest that we add the meetingId as an argument along with reflectionId & searchQuery

Given the `reflectionId`, could you grab the reflection using a dataloader & get `reflection.meetingId`? That way the user can't pass in a reflectionId that doesn't exist in a meetingId. more client-supplied args = more ways for folks to "hack" us (or at least find gaps in our authorization) 

aside from that, looks great!

PR #5277, opened 2021-08-18
Spotlight similarReflectionGroups query

nickoferrall:
PR to resolve issue #5182 

@tiffanyhan mind reviewing this one when you get the chance?


packages/server/graphql/types/User.ts
```diff
@@ -399,6 +400,40 @@ const User = new GraphQLObjectType<any, GQLContext>({
         return source.overLimitCopy
       }
     },
+    similarReflectionGroups: {
+      type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(RetroReflectionGroup))),
+      description:
+        'The reflection groups that are similar to the selected reflection in the Spotlight',
+      args: {
+        reflectionId: {
+          type: GraphQLID,
```
nickoferrall, line 409: When the Spotlight closes, the `spotlightReflectionId` becomes null and the query in the `SpotlightRoot` fires. To make `reflectionId` nonnullable, we could use [useLazyLoadQuery](https://relay.dev/docs/next/api-reference/use-lazy-load-query/) instead of the `QueryRenderer` and pass in the `reflectionId` as the skip variable?

nickoferrall:
Hey @tiffanyhan, just bumping this PR request

tiffanyhan (approved):
lgtm! i'm not sure the best way in relay to solve for the query firing when the modal is closed and `reflectionId=null` while keeping the var as non-null in the graphql definition, so i'll leave that to @mattkrick

nickoferrall:
Thanks for the review, Tiffany! Requesting a final review from @mattkrick when you get the chance

nickoferrall:
It'd prob be best to review https://github.com/ParabolInc/parabol/pull/5271 before this one as this PR builds on top of that

packages/client/components/SpotlightRoot.tsx
```diff
@@ -0,0 +1,61 @@
...
+interface Props {
+  closeSpotlight: () => void
+  meeting: SpotlightRoot_meeting
+  flipRef: (instance: HTMLDivElement) => void
+}
+
+const SpotlightRoot = (props: Props) => {
+  const {closeSpotlight, meeting, flipRef} = props
+  const {spotlightReflection} = meeting
+  const reflectionId = spotlightReflection?.id
```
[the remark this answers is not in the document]
nickoferrall: Makes sense. Returning null means the modal flashes closed rather than fades out so the ref approach looks good to me. Implemented!

packages/server/graphql/types/User.ts
```diff
@@ -404,6 +405,37 @@ const User = new GraphQLObjectType<any, GQLContext>({
...
+      resolve: async ({id: userId}, {reflectionId}, {dataLoader}) => {
+        const retroReflection = await dataLoader.get('retroReflections').load(reflectionId)
+        if (!retroReflection) return []
+        const {meetingId} = retroReflection
+        const meetingMemberId = toTeamMemberId(meetingId, userId)
+        const [viewerMeetingMember, reflectionGroups] = await Promise.all([
+          dataLoader.get('meetingMembers').load(meetingMemberId),
+          dataLoader.get('retroReflectionGroupsByMeetingId').load(meetingId)
+        ])
+        const {meetingId: viewerMeetingId} = viewerMeetingMember
```
mattkrick, line 431: i probably have a bad type on this dataloader so it's not suggesting that it could be null.
`viewerMeetingMember` is going to be null if `meetingMemberId` isn't valid.
That means you need validate viewerMeetingMember before you destructure it. e.g. `if (!viewerMeetingMember return error)`

mattkrick (commented):
1 itty bitty change, but since users can't call that mutation yet i'll go ahead & merge since you'll be editing that logic in the next PR.
Next time you're in the frontend code give the relay hooks a shot!

PR #5278, opened 2021-08-18
Org backup

### LINKED ISSUE #5043: Support postgres in backupOrganization mutation
`backupOrganization` is a great way to clone just our team's data from the prod database so we can use it as mock data in local development.

We need it to backup our postgres data, too.

AC
- [ ] backupOrganization creates a DB dump that is filtered to only contain rows that pertain to the provided orgId. If possible, this dump happens directly without cloning the rows to a separate DB. 
- [ ] the tables it dumps include:  GitHubAuth, OrganizationUserAudit, PgMigrations, PgPostDeployMigrations, Team, TemplateRef, TemplateScaleRef, User
- [ ] The notion doc https://www.notion.so/parabol/How-to-Get-Mock-Data-To-Develop-Locally-fc0b90425a3049a3aa21abeb7ca88f23 is updated to include any extra manual steps

Estimate: 16 hours

tiffanyhan:
Resolve #5043

This PR is only scoped to the dumping of data, since fetch/restore/documentation depends upon this funtionality.

 - [ ] backupOrganization creates a DB dump for the schema, which includes all data definitions including tables, FK constraints, indexes, sequence tables, etc.
 - [ ] backupOrganization creates a DB dump for the data rows that is filtered to only contain rows that pertain to the provided orgId. If possible, this dump happens directly without cloning the rows to a separate DB.
 - [ ] dumps all pg tables

Testing instructions:
  - go to admin graphiql page locally, click on private schema, do "backupOrganization" mutation
  - test with a locally created organization, as this is the current best way to ensure all pg data exists locally
  - the mutation will timeout in graphiql, and will return ```{
  "message": "Unexpected end of JSON input",
  "stack": "SyntaxError: Unexpected end of JSON input\n    at fetcher (webpack-internal:///./packages/client/modules/admin/containers/Graphql/GraphqlContainer.tsx:74:16)"
}```. this is normal, and it's what it's always done even before adding the postgres part.
  - both `schemaDump.tar.gz` and `orgBackupData.sql` should appear in the root of the project dir under `artifacts` folder
  - just look under created "orgBackup" schema in pgAdmin to verify rows have been dumped relevant to the org. this data along with schema will be the content of the dumps.
  - let's skip manually restoring the backups in the test steps, as this borders on testing `pg_dump` command itself. there are a couple edge cases involved in restoring which will be fully detailed in the documentation. 
  
Organization requirements:
  - the locally created organization should be github integrated (GitHubAuth table)
  - the locally created organization should have sprint poker activity with jira integrated objects and pushed estimate fields (AtlassianAuth and TaskEstimate table)
  - organization should have its own custom template (TemplateRef and TemplateScaleRef tables)
  - organization should have discussion threads (Discussion table)
  - for Poll and PollOption, there is no UI yet for them so just create them manually in pgAdmin, making sure to pass FK constraints.

Restore instructions:
```
cd <:path_to_parabol>/artifacts
```
```
# this will wipe all pg data and start from a clean slate
pg_restore --dbname=postgresql://pgparaboladmin:parabol@localhost:5432/parabol-saas schemaDump.tar.gz --clean
```
```
psql postgresql://pgparaboladmin:parabol@localhost:5432/parabol-saas -c 'ALTER SCHEMA public RENAME TO "orgBackup";'
```
```
pg_restore --dbname=postgresql://pgparaboladmin:parabol@localhost:5432/parabol-saas orgBackupData.tar.gz --disable-triggers
```
```
psql postgresql://pgparaboladmin:parabol@localhost:5432/parabol-saas -c 'ALTER SCHEMA "orgBackup" RENAME TO "public";'
```


tiffanyhan:
@BartoszJarocki mind giving this a review when you get the chance?

BartoszJarocki:
@tiffanyhan I should have said I actually started that today! There are few things I want to understand better so it's taking me a bit longer. It'll be done by Monday. 🙏

BartoszJarocki (approved):
Great work, Tiffany! Looks very solid, I haven't found any issues. 

I've tested locally and it's working as described. There are two things worth mentioning:
1. This PR kind of depends on Polls PR, it'll fail without Polls table
2. I had to run post-deploy migrations manually (not sure if they're supposed to run automatically), otherwise, it'll fail with 
  ```
  {
    "errors": [
      {
        "message": "relation \"PgPostDeployMigrations\" does not exist",
        "locations": [
          {
            "line": 33,
            "column": 3
          }
        ],
        "path": [
          "backupOrganization"
        ]
      }
    ],
    "data": null
  }
  ```

After running the mutation
1. `schemaDump.tar.gz` and `orgBackupData.tar.gz` were created in `artifacts` directory 
2. `orgBackup` table was created in postgres with all the tables
CleanShot 2021-08-23 at 06 37 38

I also run it for multiple org ids, not issues. 

Some questions
1. Do we worry about concurrent usage of `backupOrganization`? Mutation drops the `orgBackup` schema so I'm wondering if this could be an issue 
2. If I understand correctly, the purpose of `orgBackup` schema is to keep the data related to the selected organizations so we can dump only related data, and then, `orgBackup` schema could be deleted?

tiffanyhan:
> This PR kind of depends on Polls PR, it'll fail without Polls table

ah i guess i had that PR locally and assumed it was already in master when it's not! i can remove those tables from the PR. we're going to need to constantly update this mutation every time a table is added in postgres, so that will be a thing going forward.

> I had to run post-deploy migrations manually 

ugh that's a good catch. postdeploy migrations are run on prod but not on dev. this is a discrepancy that we should probably fix.

> Do we worry about concurrent usage of backupOrganization? Mutation drops the orgBackup schema so I'm wondering if this could be an issue

that's a really good question. i think concurrent usage of the mutation can result in problems. a possible solution is to use a locking mechanism for the mutation as a whole, perhaps an application level redis lock or a built-in postgres advisory lock. since this is only for internal usage and it's never been a problem before, i don't think we necessary need to worry about it for the scope of this first iteration. curious to hear of @mattkrick thoughts here.

> If I understand correctly, the purpose of orgBackup schema is to keep the data related to the selected organizations so we can dump only related data, and then, orgBackup schema could be deleted?

Yes the unfortunate thing about `pg_dump` is: (1) it's the only reliable way to copy all constraints in the original schema, and (2) there's no built-in way to only get a slice of rows in the db. The workaround for the former is to have separate dumps for the schema and data; the workaround for the latter is to put the slice of rows we want in its own schema, since `pg_dump` *does* support selective dumping of particular schemas.

BartoszJarocki:
Thank you, @tiffanyhan! I appreciate the answers. I learned a lot by reviewing this PR 🙌

mattkrick (changes requested):
Can you please add manual steps to restore so we can test this?
This should only take a couple hours to complete, reach out if it takes longer

tiffanyhan:
> Can you please add manual steps to restore so we can test this?
> This should only take a couple hours to complete, reach out if it takes longer

@mattkrick here are the manual steps to restore, also added to the PR description. i would pull fresh backup files from the mutation, as I just made a small change to ensure we're pulling extensions too. after this, the only thing that's not resolved is the value of id sequences for tables. this can be resolved using this [script](https://wiki.postgresql.org/wiki/Fixing_Sequences) which will be added to notion doc.

Restore instructions:
```
cd <:path_to_parabol>/artifacts
```
```
# this will wipe all pg data and start from a clean slate
pg_restore --dbname=postgresql://pgparaboladmin:parabol@localhost:5432/parabol-saas schemaDump.tar.gz --clean
```
```
psql postgresql://pgparaboladmin:parabol@localhost:5432/parabol-saas -c 'ALTER SCHEMA public RENAME TO "orgBackup";'
```
```
pg_restore --dbname=postgresql://pgparaboladmin:parabol@localhost:5432/parabol-saas orgBackupData.tar.gz --disable-triggers
```
```
psql postgresql://pgparaboladmin:parabol@localhost:5432/parabol-saas -c 'ALTER SCHEMA "orgBackup" RENAME TO "public";'
```

mattkrick:
it works!
changes added:
- pgRestore script
- inline shell script
- cleanup orgBackup after dump

PR #5280, opened 2021-08-19
Run prettier after pg:build

Dschoordsch:
Having all generated files in the wrong format is distracting during
development. Also format files which weren't correctly formatted.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "start": "pm2-runtime start pm2.config.js",
     "heroku-postbuild": "echo 'Skipping build'",
     "build": "node scripts/prod.js",
-    "pg:build": "pgtyped -c ./packages/server/postgres/pgtypedConfig.js",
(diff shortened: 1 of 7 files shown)


mattkrick:
mmm that's probably the right way to do it!
i was focused on trying to get our commit hooks to ignore those files but that was a losing battle!

Now if only pgtyped deterministically sorted their string unions :sweat: 
If you wanted to PR that codebase I wouldn't mind!

Dschoordsch:
I would prefer not to format generated files, I just figured as these are mostly formatted, that it was intentional. I will try to ignore those instead.

tiffanyhan (approved):
thanks for putting some attention on this. i usually just check in the generated files that i was working on, then restore the rest in git. it's a pesky annoyance though.  i'm good with this fix, but would also prefer #5292 if we can get the `.prettierignore` file to work and not format the generated files on save.

mattkrick:
closing as we're just ignoring these from now on :tada:  #5292

PR #5287, opened 2021-08-19
Fetch user by email from PG

Dschoordsch:
Fetch users by email from Postgres. Split query into two versions (1 user, multiple users) to avoid accidentally indexing the promise due to wrong parenthesis since it happened to me multiple times.

Resolves: #5373

### private:

* [x] draftEnterpriseInvoice
* [x] loginSAML
* [x] updateEmail
* [x] updateWatchlist
* [x] dailyPulse
* [x] user

### public
* [x] deleteUser
* [x] emailPasswordReset
* [x] attemptLogin
* [x] inviteToTeam
* [x] loginWithGoogle
* [x] resetPassword
* [x] verifyEmail
* [x] verifiedInvitation

### scripts
* [ ] backfillHubSpot

### DIFF
--- a/packages/server/postgres/queries/getUsersByEmails.ts
+++ b/packages/server/postgres/queries/getUsersByEmails.ts
@@ -1,0 +1,24 @@
+import {getUsersByEmailsQuery} from './generated/getUsersByEmailsQuery'
+import getPg from '../getPg'
+import AuthIdentity from '../../database/types/AuthIdentity'
+import User from '../../database/types/User'
+import {AuthTokenRole} from 'parabol-client/types/constEnums'
+
+export const getUsersByEmails = async (emails: string[]): Promise<User[]> => {
+  const users = await getUsersByEmailsQuery.run({emails}, getPg())
+  return (
+    // TODO #5372 remove conversion to User
+    users?.map((user) => ({
+      ...user,
+      segmentId: user.segmentId ?? undefined,
+      reasonRemoved: user.reasonRemoved ?? undefined,
+      rol: user.rol === 'su' ? AuthTokenRole.SUPER_USER : undefined,
+      identities: (user.identities as unknown) as AuthIdentity[]
+    })) ?? []
+  )
+}
+
+export const getUserByEmail = async (email: string): Promise<User | null> => {
+  const users = await getUsersByEmails([email])
+  return users[0] ?? null
+}
--- a/packages/server/graphql/intranetSchema/mutations/draftEnterpriseInvoice.ts
+++ b/packages/server/graphql/intranetSchema/mutations/draftEnterpriseInvoice.ts
@@ -12,6 +12,7 @@ import hideConversionModal from '../../mutations/helpers/hideConversionModal'
 import setTierForOrgUsers from '../../../utils/setTierForOrgUsers'
 import DraftEnterpriseInvoicePayload from '../types/DraftEnterpriseInvoicePayload'
 import updateTeamByOrgId from '../../../postgres/queries/updateTeamByOrgId'
+import {getUserByEmail} from '../../../postgres/queries/getUsersByEmails'
 
 const getBillingLeaderUser = async (
   email: string | null,
@@ -20,12 +21,7 @@ const getBillingLeaderUser = async (
 ) => {
   const r = await getRethink()
   if (email) {
-    const user = await r
-      .table('User')
-      .getAll(email, {index: 'email'})
-      .nth(0)
-      .default(null)
-      .run()
+    const user = await getUserByEmail(email)
     if (!user) {
       throw new Error('User for email not found')
     }
--- a/packages/server/graphql/intranetSchema/mutations/loginSAML.ts
+++ b/packages/server/graphql/intranetSchema/mutations/loginSAML.ts
@@ -4,6 +4,7 @@ import {GraphQLID, GraphQLNonNull, GraphQLString} from 'graphql'
 import getSSODomainFromEmail from 'parabol-client/utils/getSSODomainFromEmail'
 import querystring from 'querystring'
 import * as samlify from 'samlify'
+import {getUserByEmail} from '../../../postgres/queries/getUsersByEmails'
 import getRethink from '../../../database/rethinkDriver'
 import AuthToken from '../../../database/types/AuthToken'
 import User from '../../../database/types/User'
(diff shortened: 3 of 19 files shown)


mattkrick:
looking good!

- i like the use of the plural `emails` arg so we can use a dataloader in the future.
- also like the null checks in case no email is returned.
- in the future, we might want to include a `usersByEmail` dataloader because after fetching the user, we can cache it by `id` and then returning the `user` object by id won't require an extra DB hit. That's scope creep for this PR, so I like where you ended it. We can add that functionality in a later PR

Dschoordsch:
@nickoferrall I think I reached the limit of what I can test locally for this patch. Could you please review?

tianrunhe:
`backfillHubSpot` is a one-off script I wrote before. Your change is simple enough so I won't feel strongly against not testing it at all. However, in case you want to, and also for knowledge sharing purpose, here's a way to test it:

1. Grab the production [HubSpot API Key](https://github.com/ParabolInc/action-devops/blob/master/environments/production#L60) to put it in your local `.env`
2. Manually edit [your HubSpot profile](https://app.hubspot.com/contacts/3888472/contact/47044751), e.g., change your `Parabol Preferred Name` field
3. Change `.env` variables to connect to production Postgres instance
4. Run `yarn sucrase-node hubspot/backfillHubSpot.ts [Your Parabol User Id]`
5. Go back to HubSpot and check if the field was updated back

As you can see, it's really no different than testing in production 🤷‍♂️😂

nickoferrall:
[Review in progress]

### Queries/mutations to test
#### Private:

- [ ]  draftEnterpriseInvoice
- [ ]  loginSAML
- [ ]  updateEmail
- [ ]  updateWatchlist
- [ ]  dailyPulse
- [ ]  user

#### Public:
- [ ]  deleteUser
- [ ]  emailPasswordReset
- [ ]  attemptLogin
- [ ]  inviteToTeam
- [ ]  loginWithGoogle
- [ ]  resetPassword
- [ ]  verifyEmail
- [ ]  verifiedInvitation
- [ ]  backfillHubSpot

packages/server/postgres/queries/getUsersByEmail.ts
```diff
@@ -0,0 +1,22 @@
+import {getUsersByEmailQuery} from './generated/getUsersByEmailQuery'
+import getPg from '../getPg'
+import catchAndLog from '../utils/catchAndLog'
+import AuthIdentity from '../../database/types/AuthIdentity'
+import User from '../../database/types/User'
+import {AuthTokenRole} from 'parabol-client/types/constEnums'
+
+export const getUsersByEmail = async (emails: string[]): Promise<User[]> => {
```
nickoferrall, line 8: NIT: as we're passing `emails` plural into this function, perhaps `getUsersByEmails` would be more descriptive?

nickoferrall (approved):
Queries and mutations are looking good to me! If you'd like me to test the loginSAML or backfillHubSpot once you've tested it, just let me know

Dschoordsch:
Since backfillHubSpot is only a one-off script and the change looks identical to other places, I will follow @tianrunhe's assessment and skip testing it.

packages/server/postgres/queries/generated/getUserByEmailQuery.ts
```diff
@@ -0,0 +1,74 @@
+/** Types generated for queries found in "packages/server/postgres/queries/src/getUserByEmailQuery.sql" */
```
mattkrick, line 1: Can delete this file? 
Looks like we only use the plural version

packages/server/postgres/queries/getUsersByEmails.ts
```diff
@@ -0,0 +1,22 @@
+import {getUsersByEmailsQuery} from './generated/getUsersByEmailsQuery'
+import getPg from '../getPg'
+import catchAndLog from '../utils/catchAndLog'
+import AuthIdentity from '../../database/types/AuthIdentity'
+import User from '../../database/types/User'
+import {AuthTokenRole} from 'parabol-client/types/constEnums'
+
+export const getUsersByEmails = async (emails: string[]): Promise<User[]> => {
+  return (
+    (await catchAndLog(() => getUsersByEmailsQuery.run({emails}, getPg())))?.map((user) => ({
```
mattkrick, line 10: let's simplify this to:
```ts
  return getUsersByEmailsQuery.run({emails}, getPg())
```

- the `User` type imported only exists because RethinkDB doesn't have a schema.  `IGetUsersByEmailsQueryResult` will be the single source of truth when RethinkDB goes away :tada: 
- no need for `catchAndLog` because the table exists. at worst it will return an array with `rows.length < emails.length`
-  no need for the `.map` if it returns type `IGetUsersByEmailsQueryResult`. For future reference, if the auto-generated types aren't strict enough we can create a new interface like this:  https://github.com/ParabolInc/parabol/blob/ba49e0e3ba4345857c01c9caa3852b0f51c173a9/packages/server/postgres/queries/getTemplateRefsById.ts#L8-L13. that way we can avoid doing any scrubbing at runtime

packages/server/postgres/queries/getUsersByEmails.ts
```diff
@@ -0,0 +1,22 @@
...
+      segmentId: user.segmentId ?? undefined,
+      reasonRemoved: user.reasonRemoved ?? undefined,
+      rol: user.rol === 'su' ? AuthTokenRole.SUPER_USER : undefined,
+      identities: (user.identities as unknown) as AuthIdentity[]
+    })) ?? []
+  )
+}
+
+export const getUserByEmail = async (email: string): Promise<User | null> => {
+  return (await getUsersByEmails([email]))[0] ?? null
```
mattkrick, line 21: nit: 2 lines might be easier to read:
```ts
const users = await getUsersByEmail([email])
return users[0] ?? null
```

mattkrick (changes requested):
Looks great! Just a couple little things. If the fixes are easy feel free to merge after committing

packages/server/postgres/queries/getUsersByEmails.ts
Dschoordsch: This leads to type mismatches in a lot of places where we still fetch User from Rethink or PG depending on arguments. I would like to keep it like this for now and change it after I converted the other usages. I updated #5372 to cover this.

mattkrick: makes sense. thanks for the update

mattkrick:
LGTM :+1:

mattkrick:
@Dschoordsch small nit-- instead of force pushes, could you just do multiple commits or a squash or rebase? we can also always squash before merging if you like.

 github can't find links to old comments or notifications if they've been deleted in a force push

PR #5290, opened 2021-08-19
Do not format generated files

Dschoordsch:
Commit pg:build files in unformatted form.


Dschoordsch:
Now we have a choice, we should either do this or #5280

Dschoordsch:
Missed the sneaky commit hook formatting

PR #5292, opened 2021-08-19
Do not format generated files

Dschoordsch:
Commit pg:build files in unformatted form.
For VSCode users switch from builtin formatter to prettier extension
one. The builtin one formats different from prettier anyways.
Add the extension as recommendation.
There are two .prettierignore files added since VSCode extension looks
in the root directory while git commit hooks check the package.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -26,6 +26,10 @@
 		}
 	],
 	"npm.packageManager": "yarn",
+	"editor.defaultFormatter": "esbenp.prettier-vscode",
+	"[javascript]": {
+		"editor.defaultFormatter": "esbenp.prettier-vscode",
+	},
 	"javascript.format.semicolons": "remove",
 	"javascript.preferences.quoteStyle": "single",
 	"javascript.preferences.useAliasesForRenames": false,
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,2 @@
+**/__generated__/
+**/generated/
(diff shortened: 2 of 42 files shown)


Dschoordsch:
We should do this or #5280

tiffanyhan:
i pulled the remote branch to check it out. it looks like upon saving the generated files, the code still gets formatted, even though `.prettierignore` is present. not sure why `.prettierignore` is getting ignored (no pun intended)

Dschoordsch:
The file should really be in the root directory next to the `.prettierrc` but then it is ignored during pre-commit formatting. Maybe just putting 2 of those there would solve it, I will check.

Dschoordsch:
@tiffanyhan Could you check again if this works for you now?

tiffanyhan:
it worked for me! this is so great, looks like the extra setting instructions to vscode made the difference.

tiffanyhan (approved):

Dschoordsch:
@mattkrick is this ok to merge?

.vscode/settings.json
```diff
@@ -26,6 +26,10 @@
 		}
 	],
 	"npm.packageManager": "yarn",
+	"editor.defaultFormatter": "esbenp.prettier-vscode",
```
mattkrick, line 29: out .eslint has prettier built in (https://github.com/ParabolInc/parabol/blob/1b2eb07d7804280847c616bdc372b7c61c83f0cb/.eslintrc.js#L3)
I wonder it hasn't been working :thinking:.
Happy to make the switch! At least on my machine there's some kind of memory leak with eslint that causes it to spin forever when trying to format. maybe only prettier will fix it!

PR #5303, opened 2021-08-23
Add poll button

### LINKED ISSUE #5145: Polls implementation: [Frontend] Adding a poll
This issue is part of **Polls implementation: EPIC #5147**

The new polls feature allows users to create a multiple choice question in a discussion thread, and lets meeting attendees vote. 

- [Please watch this Loom for a quick overview](https://www.loom.com/share/3ed5d71323494b24bb850ca55d4ebdcc?sharedAppSource=team_library).
- [The Figma source file can be found here](https://www.figma.com/file/D25gVH1t2alUiOdYNs5Owo/Multiple-content-input-types?node-id=518%3A19375).

Feel free to ping me for questions or comments.

# The 'Add a poll' action

Screen Shot 2021-07-15 at 4 47 51 p m

- [ ] A new 'Add a poll' option is added to the 'action bar' below the comment field in a discussion thread.
- [ ] Both actions are evenly spaced and centered in their container.
- [ ] 'Add a poll' has the same styles that its sibling 'Add a task'.
- [ ] 'Add a poll' uses `poll` (outlined) from the Material Design icon library.

BartoszJarocki:
Fixes #5145 
This PR depends on the #5231 and it's related only to the UI. 

Design 👇 
125862228-8a95e3ec-8bb5-4bf0-9da1-0a71195500b7

Result 👇 
localhost_3000_meet_6gRsuaAcLK_agendaitems_1

AC
- [ ] `Add poll` button is displayed right to the `Add task` button
- [ ] Buttons are centered in the container and match the design


BartoszJarocki:
@nickoferrall may I ask you to take a look? 🙏

nickoferrall (approved):
Great work! I'm excited for this polls feature. I've left a few comments - LGTM!

BartoszJarocki:
Thank you @nickoferrall for the review! @mattkrick I'm moving this one to the maintainer review. I do have one question though, how are we dealing with things that aren't implemented yet and should be hidden on prod? This button is not yet functional so we'd need to hide it. Do we have a feature flags mechanism or should I introduce some temp variable to do so? 🙏🏻

packages/client/components/ActionMeetingAgendaItems.tsx
```diff
@@ -61,7 +61,7 @@ const ActionMeetingAgendaItems = (props: Props) => {
   if (!agendaItem) return null
   const {content, teamMember} = agendaItem
   const {picture, preferredName} = teamMember
-  const allowedThreadables: DiscussionThreadables[] = endedAt ? [] : ['comment', 'task']
+  const allowedThreadables: DiscussionThreadables[] = endedAt ? [] : ['comment', 'task', 'poll']
```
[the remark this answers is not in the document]
BartoszJarocki: Yeah, I was wondering about that. I personally would keep it that way as it's a bit more future proof (maybe at some point backend will send these?). A helper function is kind of like using a singleton, although I agree, it'd be nice to have fewer props. Any thoughts on this @mattkrick?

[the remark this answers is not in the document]
mattkrick: it's a hard problem because the prop is a function of domain + local state. (e.g. is the meeting ended? is it a poker meeting? is someone already adding a task?)
We also don't want the discussion thread to depend on any meeting state because in the future a discussion can live outside a meeting, so we need to do those calcs in a parent.

it's used in 12 files, 6 of those files only pass the props through.

- we could put them in context, but IMO that'd do more harm than good because a dev would still have to climb up the tree to see where the context is set. context always feels a little magical as it's just this state store floating out there
- we could store it in the client cache, but then we'd have to write updaters for things like when the meeting ends & the discussion thread becomes read-only
- we could pass in an object with some methods that are fancier than `Array.includes`


Not sure any of those are substantially better, but if you have a strong opinion let me know!

packages/client/components/Icon.tsx
```diff
@@ -1,28 +1,9 @@
...
-  WebkitFontSmoothing: 'antialiased',
-  // Support for Safari and Chrome
-  textRendering: 'optimizeLegibility',
-  // Support for Firefox
-  mozOsxFontSmoothing: 'grayscale',
-  // Support for IE
-  fontFeatureSettings: 'liga'
-})
+const Icon = styled.i`
+  ${iconStyles};
```
mattkrick, line 5: is there a way to keep typesafety if we use this new pattern?

for example: 

```ts
const Icon = style('i')({
 textRendering: 'optimizeLegibility'
})
```
vscode will offer typeahead support for both the key & value & if you type an invalid value it'll be a typescript error.

alternatively, you could always subclass the component instead of the styles:
```ts
const IconOutlined = styled(Icon)({
  fontFamily: 'Material Icons Outlined'
})
```

BartoszJarocki: Valid concern! 
When using `styled.i` we're still getting eslint errors. 
Screenshot 2021-08-31 at 15 08 25
My idea was to prevent usage of unstyled icons and use the composition: https://emotion.sh/docs/composition

Right now I think it's better to keep things as they're and subclass the icons, ie. have a base `Icon` and then `IconFilled` and `IconOutlined`. Will push changes soon.

packages/client/components/ActionMeetingAgendaItems.tsx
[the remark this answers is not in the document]
BartoszJarocki: Let's leave it as is then! None of these options sounds like a better solution IMO.

packages/client/components/Icon.tsx
BartoszJarocki: @mattkrick I've just changed icons implementation to the way you mentioned. `IconOutlined` is simply created by extending & overriding `Icon`. I really overcomplicated it before 😓 . Thank you for the suggestion!

packages/client/components/DiscussionThreadInput.tsx
```diff
@@ -229,6 +233,9 @@ const DiscussionThreadInput = forwardRef((props: Props, ref: any) => {
     CreateTaskMutation(atmosphere, {newTask}, {})
   }
 
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  const addPoll = () => {}
```
mattkrick, line 239: you can always put an inline comment in the body to get around this lint rule
https://github.com/ParabolInc/parabol/blob/6bf32993d5f6d7386f38c2fc5e7b27609740deea/packages/client/Atmosphere.ts#L59-L62

mattkrick (approved):
LGTM! added a 1-liner feature flag to hide the button in prod

Issue #5146, opened 2021-07-15
Polls implementation: Creating a poll object


enriquesanchez:
This issue is part of Polls implementation: EPIC

The new polls feature allows users to create a multiple choice question in a discussion thread, and lets meeting attendees vote.

- Please watch this Loom for a quick overview.
- The Figma source file can be found here.

Feel free to ping me for questions or comments.

# Creating a poll object

![image](https://user-images.githubusercontent.com/3276087/125862815-0bd89e78-3f01-40c3-b355-23dafa926e68.png)

- [ ] When a poll object is being created, the 'Add a task' and 'Add a poll' actions in the action bar are disabled.
- [ ] Similar to when a task is being created, the user and action ("Enrique is adding a poll...") is indicated next to the avatar. 
- [ ] In the creation state, the poll object contains the following fields and options (all set in 14px font size):
  - [ ] 'Ask a question' (field, required)
  - [ ] 'Add choice 1' (field, required)
  - [ ] 'Add choice 2' (field, required)
  - [ ] 'Add another choice' (text button, set in bold and preceded by the icon `add`)
  - [ ] 'Start' (button, initially disabled)
- [ ] User needs to fill in the question, choice 1, and choice 2 fields in order for the poll to be ready to start. 
- [ ] Not completing all required fields, and/or clicking elsewhere before the poll is submitted, cancels the poll creation flow (similar to tasks), and returns keyboard focus to the 'Add a poll' button in the action bar.
- [ ] A poll can only have up to 4 choices. Choices 3 and 4 being optional.
- [ ] Upon creation, keyboard focus should move to the poll object (blue border, drop shadow), and:
  - [ ] User should be able to start typing and fill the "Ask a question" field.
  - [ ] The next focus stop after "Ask a question" is the 'Add choice 1" field. The user can move focus to this field by either pressing <kbd>TAB</kbd>, or by clicking on the field. The blue border to indicate focus should now move to this field.
 
<img width="833" alt="Screen Shot 2021-07-15 at 4 54 14 p m" src="https://user-images.githubusercontent.com/3276087/125865033-27620604-0f16-4cf6-866c-896e123a9919.png">

- [ ] When a choice field is focused, a character counter ("0/30") is displayed on the upper right corner of the field, in `
Slate 600 #82809A`, and set to `10px` font size.

[...]

PR #5304, opened 2021-08-23
[Backend] Creating a poll object

### LINKED ISSUE #5315: Polls implementation: [Backend] Creating a poll object
## Issue - Enhancement
This issue relates to the backend implementation of creating a poll. Related frontend issue #5146 

- [ ] Create a `CreatePoll` mutation
- [ ] Poll is correctly saved to database
- [ ] Poll options are correctly saved to database

**Estimated effort:** 24h

BartoszJarocki:
Fixes #5315. This PR depends on #5303. 

Scope
* Added `CreatePoll` mutation
* Added `getPollsById`, `getPollsByDiscussionId` and `getPollOptionsByPollId` db queries
* Added `pollLoaders` 
* Added GraphQL types related to polls
* Updated `pgtyped` to the newest version to support readonly arrays (no more nasty casts when passing parameters to generated queries!) 

- [ ] Create a `CreatePoll` mutation
- [ ] Poll is correctly saved to database
- [ ] Poll options are correctly saved to database

Example mutation in graphiql (change `teamId` and `discussionId` to the proper values that exists in your local db)
```
mutation {
  createPoll(
    newPoll: {threadSortOrder: 0.4, teamId: "6gEbA4BOk9", discussionId: "6j6SB4HMXK", title: "poll!", options: [{title: "option title1"}, {title: "option title2"}]}
  ) {
    ... on ErrorPayload {
      error {
        title
        message
      }
    }
    ... on CreatePollSuccess {
      poll {
        id
        title
        discussionId
        threadSortOrder
        options {
          id
          title
        }
      }
    }
  }
}
```

Example result
Screenshot 2021-09-01 at 07 25 00


packages/client/mutations/CreatePollMutation.ts
```diff
@@ -0,0 +1,44 @@
...
+  atmosphere,
+  variables,
+  {onError, onCompleted}
+) => {
+  return commitMutation<TCreatePollMutation>(atmosphere, {
+    mutation,
+    variables,
+    optimisticUpdater: (store) => {
+      const {} = variables
+    },
```
BartoszJarocki, line 38: This will be implemented in frontend part #5146

BartoszJarocki:
@tianrunhe may I ask you to take a look, please? 🙏🏻

packages/server/postgres/queries/src/getPollsOptionsByPollIdQuery.sql
```diff
@@ -0,0 +1,6 @@
+/*
+  @name getPollOptionsByPollIdQuery
```
tianrunhe, line 2: nit: make the name consistent with file name

packages/server/postgres/queries/getPollOptionsByPollId.ts
```diff
@@ -0,0 +1,12 @@
+import {
```
tianrunhe, line 1: Nit: rename this file to `getPollOptionsByPollId`**`s`**

packages/server/postgres/queries/src/getPollsByDiscussionIdQuery.sql
```diff
@@ -0,0 +1,6 @@
+/*
+  @name getPollsByDiscussionIdQuery
```
tianrunhe, line 2: Nit: Rename it to `getPollsByDiscussionId`**`s`**`Query`.

packages/server/postgres/queries/getPollsByDiscussionId.ts
```diff
@@ -0,0 +1,7 @@
+import {getPollsByDiscussionIdQuery} from './generated/getPollsByDiscussionIdQuery'
```
tianrunhe, line 1: Rename this to `getPollsByDiscussionId`**`s`** so it's clear that it expects an array of discussionIds

packages/server/postgres/queries/getPollsByDiscussionId.ts
```diff
@@ -0,0 +1,7 @@
...
+import getPg from '../getPg'
+
+const getPollByDiscussionId = async (discussionIds: readonly string[]) =>
```
tianrunhe, line 4: Nit: try to make this consistent with the file name

packages/server/postgres/queries/src/getPollsByIdQuery.sql
```diff
@@ -0,0 +1,6 @@
+/*
+  @name getPollsByIdQuery
```
tianrunhe, line 2: Rename this to `getPollsById`**`s`**`Query`

packages/server/postgres/queries/getPollsById.ts
```diff
@@ -0,0 +1,8 @@
+import {getPollsByIdQuery, IGetPollsByIdQueryResult} from './generated/getPollsByIdQuery'
```
tianrunhe, line 1: Rename this file to `getPollsbyId`**`s`** so it's clear that it expects an array of ids

packages/server/postgres/queries/getMeetingTaskEstimates.ts
```diff
@@ -9,11 +9,8 @@ export interface MeetingTaskEstimatesResult extends IGetMeetingTaskEstimatesQuer
   stageId: string
 }
 
-const getMeetingTaskEstimates = async (
-  taskIds: MaybeReadonly<string[]>,
-  meetingIds: MaybeReadonly<string[]>
-) => {
-  return getLatestTaskEstimatesQuery.run({taskIds, meetingIds} as any, getPg()) as Promise<
+const getMeetingTaskEstimates = async (taskIds: MaybeReadonly<string[]>) => {
```
tianrunhe, line 12: Is this change of removing `meetingIds` from the argument list expected?

packages/server/dataloader/pollsLoaders.ts
```diff
@@ -0,0 +1,46 @@
...
+import getPollByDiscussionId from '../postgres/queries/getPollsByDiscussionId'
+import getPollsById from '../postgres/queries/getPollsById'
+import {IGetPollOptionsByPollIdQueryResult} from '../postgres/queries/generated/getPollsOptionsByPollIdQuery'
+import {IGetPollsByIdQueryResult} from '../postgres/queries/generated/getPollsByIdQuery'
+import {IGetPollsByDiscussionIdQueryResult} from '../postgres/queries/generated/getPollsByDiscussionIdQuery'
+
+export const pollOptions = (parent: RethinkDataLoader) => {
+  return new DataLoader<number, IGetPollOptionsByPollIdQueryResult[], string>(
+    async (pollIds) => {
+      const rows = (await getPollOptionsByPollId(pollIds)) || []
```
tianrunhe, line 13: I think if the query returns nothing for the `pollIds`, it will return an empty array so this feels redundant

packages/server/graphql/types/Poll.ts
```diff
@@ -0,0 +1,70 @@
...
+  fields: () => ({
+    ...(threadableFields() as any),
+    id: {
+      type: GraphQLNonNull(GraphQLID),
+      description: 'Poll id in a format of `poll:idGeneratedByDatabase`',
+      resolve: ({id}) => PollId.create(id)
+    },
+    meetingId: {
+      type: GraphQLNonNull(GraphQLID),
+      description: 'the foreign key for the meeting the task was created in'
```
tianrunhe, line 25: "the **poll** was created in"

packages/server/graphql/types/Poll.ts
```diff
@@ -0,0 +1,70 @@
...
+        return dataLoader.get('teams').load(teamId)
+      }
+    },
+    title: {
+      type: GraphQLNonNull(GraphQLString),
+      description: 'Poll title'
+    },
+    options: {
+      type: GraphQLNonNull(GraphQLList(GraphQLNonNull(PollOption))),
+      description: 'A list of the most recent estimates for the task',
```
tianrunhe, line 44: Clear copy/paste typo 🤣

packages/server/graphql/mutations/createPoll.ts
```diff
@@ -0,0 +1,129 @@
...
+      threadSortOrder,
+      title
+    })
+    if (insertPollResult.length === 0) {
+      return {error: {message: `Couldn't create a poll`}}
+    }
+    const [{id: pollId}] = insertPollResult
+    await insertPollOptions({pollOptions: options.map(({title}) => ({pollId, title}))})
+
+    const data = {pollId}
```
tianrunhe, line 115: `meetingId` is missing here but it's included in `CreatePollPayload`

packages/server/graphql/mutations/createPoll.ts
```diff
@@ -0,0 +1,129 @@
...
+      }
+    }
+    if (options.some((option) => !option.title)) {
+      return {error: {message: 'All options need a title'}}
+    }
+    const discussion = await dataLoader.get('discussions').load(discussionId)
+    if (!discussion) {
+      return {error: {message: 'Invalid discussion thread'}}
+    }
+    const {meetingId} = discussion
```
tianrunhe, line 90: I'm curious why you chose to not include `meetingId` as another input argument?

packages/server/graphql/mutations/createPoll.ts
```diff
@@ -0,0 +1,129 @@
...
+    const subOptions = {mutatorId, operationId}
+    const {teamId, discussionId, threadSortOrder, title, options} = newPoll
+
+    //AUTH
+    if (!isTeamMember(authToken, teamId)) {
+      return {error: {message: 'Not on team'}}
+    }
+
+    // VALIDATION
+    if (!title) {
```
tianrunhe, line 56: Should we also check if the `title` is a duplicate, i.e., can we create two identical polls?

tianrunhe (changes requested):

packages/server/postgres/queries/getMeetingTaskEstimates.ts
BartoszJarocki: Yes, if you take a look at the underlying query it doesn't have `meetingIds` and after upgrading `pgtyped` TS complained
```
 Object literal may only specify known properties, and 'meetingIds' does not exist in type 'IGetLatestTaskEstimatesQueryParams'.ts(2345)
```

packages/server/graphql/types/Poll.ts
BartoszJarocki: 🤦🏻‍♂️

packages/server/graphql/types/Poll.ts
BartoszJarocki: Inheritance by copy/paste has some downsides ;) I should've noticed that during self-review 🙏🏻

packages/server/graphql/mutations/createPoll.ts
BartoszJarocki: oh, good catch! I don't think it should be included there so I'm going to remove `meetingId` from `CreatePollPayload`

packages/server/graphql/mutations/createPoll.ts
BartoszJarocki: Including `meetingId` would require another check. We can only create a poll in a meeting related to this particular discussion. So I'd say, _not_ having `meetingId` in input argument actually makes it harder to provide wrong arguments.

packages/server/graphql/mutations/createPoll.ts
BartoszJarocki: I'm personally don't think it should be an issue but I'm open to check that. @mattkrick @enriquesanchez any thoughts on that?

BartoszJarocki:
@tianrunhe thank you for the review! May I ask you to do another round? 🙏🏻

tianrunhe (approved):
LGTM!

packages/server/graphql/mutations/createPoll.ts
mattkrick: that's a good consideration.
I feel like the `sortOrder` acts as the unique identifier in the UI. 
If they have multiple rounds of polls, removing 1 option each time, they might keep the same name? IMO that's kinda lame, but i could see folks doing it. I'm cool with leaving it out for now just to get it out & see what folks do with it

mattkrick: can delete this conditional or change it to `title.length === 0` because title is guaranteed to be non-null from `CreatePollInput`

maybe we add a min length?

packages/server/graphql/mutations/createPoll.ts
```diff
@@ -0,0 +1,129 @@
...
+import {GQLContext} from '../graphql'
+import CreatePollInput from '../types/CreatePollInput'
+import MeetingMemberId from 'parabol-client/shared/gqlIds/MeetingMemberId'
+import insertPoll from '../../postgres/queries/insertPoll'
+import insertPollOptions from '../../postgres/queries/insertPollOptions'
+import segmentIo from '../../utils/segmentIo'
+
+const MAX_POLL_OPTIONS = 4
+const MIN_POLL_OPTIONS = 2
+const MAX_TITLE_LENGTH = 100
```
mattkrick, line 16: min title length? maybe 2?

packages/server/graphql/mutations/createPoll.ts
```diff
@@ -0,0 +1,129 @@
...
+import {SubscriptionChannel} from 'parabol-client/types/constEnums'
+
+import {GQLContext} from '../graphql'
+import CreatePollInput from '../types/CreatePollInput'
+import MeetingMemberId from 'parabol-client/shared/gqlIds/MeetingMemberId'
+import insertPoll from '../../postgres/queries/insertPoll'
+import insertPollOptions from '../../postgres/queries/insertPollOptions'
+import segmentIo from '../../utils/segmentIo'
+
+const MAX_POLL_OPTIONS = 4
```
mattkrick, line 14: why 4? In other places we use 12 to signify "a few" (1 for each month). maybe that's too much here, but i worry 4 is too little. @enriquesanchez thoughts?

packages/server/graphql/mutations/createPoll.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+    if (options.length > MAX_POLL_OPTIONS) {
+      return {
+        error: {
+          message: `Poll can have at most ${MAX_POLL_OPTIONS} options, provided ${options.length}`
+        }
+      }
+    }
+    if (options.some((option) => !option.title)) {
+      return {error: {message: 'All options need a title'}}
```
mattkrick, line 84: nit: `option.title.length === 0` just to be super explicit (since non-null is guaranteed)

packages/server/graphql/mutations/createPoll.ts
```diff
@@ -0,0 +1,129 @@
...
+const MIN_POLL_OPTIONS = 2
+const MAX_TITLE_LENGTH = 100
+
+type PollOptionsInputVariables = {
+  title: string
+}
+
+type CreatePollInputVariables = {
+  newPoll: {
+    teamId: string
```
mattkrick, line 24: remove `teamId` here. `teamId` lives on the `Discussion` object provided by `discussionId`

packages/server/graphql/mutations/createPoll.ts
```diff
@@ -0,0 +1,129 @@
...
+    {newPoll}: CreatePollInputVariables,
+    {authToken, dataLoader, socketId: mutatorId}: GQLContext
+  ) => {
+    const viewerId = getUserId(authToken)
+    const operationId = dataLoader.share()
+    const subOptions = {mutatorId, operationId}
+    const {teamId, discussionId, threadSortOrder, title, options} = newPoll
+
+    //AUTH
+    if (!isTeamMember(authToken, teamId)) {
```
mattkrick, line 51: can remove, we'll check `Discussion` for this

packages/server/graphql/mutations/createPoll.ts
```diff
@@ -0,0 +1,129 @@
...
+    }
+    if (options.some((option) => !option.title)) {
+      return {error: {message: 'All options need a title'}}
+    }
+    const discussion = await dataLoader.get('discussions').load(discussionId)
+    if (!discussion) {
+      return {error: {message: 'Invalid discussion thread'}}
+    }
+    const {meetingId} = discussion
+    if (!meetingId) {
```
mattkrick, line 91: let's remove this check. in the future, we might make `Discussion.meetingId` nullable.
Instead, let's check `teamId` and make sure they're a team member

packages/server/graphql/mutations/createPoll.ts
```diff
@@ -0,0 +1,129 @@
...
+      return {error: {message: 'Discussion does not take place in a meeting'}}
+    }
+    const meetingMemberId = MeetingMemberId.join(meetingId, viewerId)
+    const viewerMeetingMember = await dataLoader.get('meetingMembers').load(meetingMemberId)
+    if (!viewerMeetingMember) {
+      return {error: {message: 'Not a member of the meeting'}}
+    }
+
+    // RESOLUTION
+    const insertPollResult = await insertPoll({
```
mattkrick, line 101: this is our first multi-table insert in PG :tada: 
we could do this a few ways:
- chained inserted as you have
- chained inserts in a transaction
- CTEs: https://www.postgresql.org/docs/current/queries-with.html#QUERIES-WITH-MODIFYING

which do you prefer? any pros/cons of each?

one thing worth considering: if the client calls `createPoll` 2 times at once (maybe their internet dropped & reconnected, or we goof up the front-end & let them click submit twice, etc) is there a way we can de-dupe the poll creation?

packages/server/graphql/types/CreatePollInput.ts
```diff
@@ -0,0 +1,37 @@
...
+  GraphQLList,
+  GraphQLNonNull,
+  GraphQLString
+} from 'graphql'
+import PollOptionInput from './CreatePollOptionInput'
+
+const CreatePollInput = new GraphQLInputObjectType({
+  name: 'CreatePollInput',
+  fields: () => ({
+    teamId: {
```
mattkrick, line 14: remove teamId

packages/server/graphql/types/CreatePollInput.ts
```diff
@@ -0,0 +1,37 @@
...
+      type: GraphQLNonNull(GraphQLID),
+      description: 'Foreign key for the team this poll belongs to'
+    },
+    discussionId: {
+      type: GraphQLNonNull(GraphQLID),
+      description: 'Foreign key for the discussion this was created in'
+    },
+    threadSortOrder: {
+      type: GraphQLNonNull(GraphQLFloat),
+      description: 'The order of this threadable, relative to threadParentId'
```
mattkrick, line 24: polls don't have a `threadParentId` because they can't be a "reply" to another type, right?

packages/server/graphql/types/CreatePollPayload.ts
```diff
@@ -0,0 +1,21 @@
+import {GraphQLNonNull, GraphQLObjectType} from 'graphql'
+import {GQLContext} from '../graphql'
+import makeMutationPayload from './makeMutationPayload'
+import Poll from './../types/Poll'
+
+export const CreatePollSuccess = new GraphQLObjectType<any, GQLContext>({
+  name: 'CreatePollSuccess',
+  fields: () => ({
+    poll: {
```
mattkrick, line 9: nit: sometimes i like to include the field `pollId` in case a future use case only needs the Id instead of fetching the whole poll

packages/client/shared/gqlIds/PollId.ts
```diff
@@ -0,0 +1,9 @@
+const PollId = {
+  create: (pollId: number) => `poll:${pollId}`,
```
mattkrick, line 2: nit: the other files in the folder have a `join` method. I know where' not actually joining anything here though. not sure which is better :shrug:

packages/server/graphql/types/PollOption.ts
```diff
@@ -0,0 +1,46 @@
...
+    updatedAt: {
+      type: GraphQLNonNull(GraphQLISO8601Type),
+      description: 'The timestamp the item was updated'
+    },
+    pollId: {
+      type: GraphQLNonNull(GraphQLID),
+      description: 'The foreign key of the poll this option belongs to'
+    },
+    poll: {
+      type: GraphQLNonNull(Team),
```
mattkrick, line 29: type is Team?

packages/server/graphql/types/Poll.ts
```diff
@@ -0,0 +1,70 @@
...
+  isTypeOf: ({title}) => !!title,
+  fields: () => ({
+    ...(threadableFields() as any),
+    id: {
+      type: GraphQLNonNull(GraphQLID),
+      description: 'Poll id in a format of `poll:idGeneratedByDatabase`',
+      resolve: ({id}) => PollId.create(id)
+    },
+    meetingId: {
+      type: GraphQLNonNull(GraphQLID),
```
mattkrick, line 24: `meetingId` is nullable in the DB

packages/server/graphql/types/PollOption.ts
```diff
@@ -0,0 +1,46 @@
...
+    createdAt: {
+      type: GraphQLNonNull(GraphQLISO8601Type),
+      description: 'The timestamp the item was created'
+    },
+    updatedAt: {
+      type: GraphQLNonNull(GraphQLISO8601Type),
+      description: 'The timestamp the item was updated'
+    },
+    pollId: {
+      type: GraphQLNonNull(GraphQLID),
```
mattkrick, line 25: this is where stuff gets dicey.
is this the numerical database ID or the GraphQL ID? e.g.  `1` or `PollOption:1`?
we'll have to be very explicit on which one it is in the description & make sure we always keep it consistent.
My advice is:
- any DB query helper should return the DB foreign key
- any dataloader should return the DB foreign key
- any graphql resolver should return the GraphQL ID

this will keep our database ID abstraction from leaking into the graphql schema. 
 that way, on our client cache, we'll still have globally unique IDs so we can do this: https://relay.dev/docs/guided-tour/reusing-cached-data/filling-in-missing-data/

packages/server/graphql/types/PollOption.ts
```diff
@@ -0,0 +1,46 @@
...
+      description: 'The foreign key of the poll this option belongs to'
+    },
+    poll: {
+      type: GraphQLNonNull(Team),
+      description: 'The poll this option belongs to',
+      resolve: ({pollId}, _args, {dataLoader}) => {
+        return dataLoader.get('polls').load(pollId)
+      }
+    },
+    voteUserIds: {
```
mattkrick, line 35: i foresee a `voteUsers` array too, but we can hold off until the front-end gets built

packages/server/graphql/types/Threadable.ts
```diff
@@ -67,8 +68,16 @@ const Threadable = new GraphQLInterfaceType({
   name: 'Threadable',
   description: 'An item that can be put in a thread',
   fields: threadableFields,
-  resolveType: ({status}) => {
-    return status ? Task : Comment
+  resolveType: ({status, options}) => {
```
mattkrick, line 71: might be worthwhile casting to the `database/Threadable.ts` type so we get a little type safety here. 
then again, you could just merge that file into this one

packages/server/postgres/queries/src/getPollOptionsByPollIdsQuery.sql
```diff
@@ -0,0 +1,6 @@
+/*
+  @name getPollOptionsByPollIdsQuery
+  @param pollIds -> (...)
+*/
+SELECT * FROM "PollOption"
```
mattkrick, line 5: I keep wondering if we'll ever query a PollOption without querying a Poll, or vice versa.
In NoSQL, we'd simply have all the options live in the same document as the poll.
In SQL, it makes sense to properly normalize them. 

Is it worthwhile writing 1 query that fetches a poll and all its options? that means we could get everything we need with a single trip to the database.

mattkrick (changes requested):
great work! all the comments are pretty tiny, feel free to discuss

packages/server/dataloader/customLoaderMakers.ts
```diff
@@ -150,10 +150,9 @@ export const latestTaskEstimates = (parent: RethinkDataLoader) => {
 export const meetingTaskEstimates = (parent: RethinkDataLoader) => {
   return new DataLoader<{meetingId: string; taskId: string}, MeetingTaskEstimatesResult[], string>(
     async (keys) => {
-      const meetingIds = keys.map(({meetingId}) => meetingId)
       const taskIds = keys.map(({taskId}) => taskId)
 
-      const rows = await getMeetingTaskEstimates(taskIds, meetingIds)
+      const rows = await getMeetingTaskEstimates(taskIds)
```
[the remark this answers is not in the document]
BartoszJarocki: sure!

packages/server/postgres/queries/getMeetingTaskEstimates.ts
```diff
@@ -9,11 +9,8 @@ export interface MeetingTaskEstimatesResult extends IGetMeetingTaskEstimatesQuer
...
+  return getLatestTaskEstimatesQuery.run({taskIds}, getPg()) as Promise<
```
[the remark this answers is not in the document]
BartoszJarocki: makes sense 😅  fixed!

packages/server/dataloader/RethinkDataLoader.ts
```diff
@@ -17,7 +18,8 @@ const loaderMakers = {
   ...primaryLoaderMakers,
   ...foreignLoaderMakers,
   ...customLoaderMakers,
-  ...atlassianLoaders
+  ...atlassianLoaders,
+  ...pollLoaders
```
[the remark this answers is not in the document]
BartoszJarocki: @mattkrick which one do you prefer? I think it's better to separate loaders by feature as it's easier to find/maintain.
Also, by extending the type you mean this
```
type CustomLoaderMakers = typeof customLoaderMakers & typeof atlassianLoaders & typeof pollLoaders
```

ie. adding `pollLoaders` here? 

And thank you, I missed that!

packages/server/graphql/mutations/createPoll.ts
BartoszJarocki: I used 4 as this is defined in [AC](https://github.com/ParabolInc/parabol/issues/5146). We can change that to whatever we think is best, thought!

packages/server/graphql/mutations/createPoll.ts
BartoszJarocki: Good idea! I'll move all the consts related to polls to `constEnums` and clean the code.

packages/server/graphql/mutations/createPoll.ts
BartoszJarocki: Hmm, not sure I understand. Should we remove this check entirely or read `teamId` from the discussion and then perform the check? If we're going to remove this entirely, how can we ensure that user with a given access token has the permission to add polls to a given discussion? 🤔

packages/server/graphql/types/CreatePollInput.ts
BartoszJarocki: Oh yeah. The comment should not say `relative to threadParentId`. Just to make sure, `threadSortOrder` is still a valid property used to determine the sort order in the discussion, or am I thinking wrong here? 🙏

packages/server/graphql/types/PollOption.ts
BartoszJarocki: Not sure how did I miss that, sorry!🤦

packages/client/shared/gqlIds/PollId.ts
BartoszJarocki: Join it is!

packages/server/graphql/types/Poll.ts
BartoszJarocki: Good catch!

packages/server/graphql/types/CreatePollPayload.ts
BartoszJarocki: Just to make sure, with a format of `poll:idGeneratedByDatabase`, not raw id generated by DB? 😅

BartoszJarocki: This was a dumb question, it's a resolver so can't be anything else but `poll:idGeneratedByDatabase` to keep things consistent.

packages/server/graphql/mutations/createPoll.ts
BartoszJarocki: I guess the answer is reading `teamId` from the discussion, correct me if I'm wrong, please!

packages/server/graphql/types/Threadable.ts
BartoszJarocki: @mattkrick I've decided to leave `database/Threadable.ts` in a separate file as there's more than one place that uses this type and in my opinion, it's cleaner that way. I've added type checks, though. Let me know what do you think!

packages/server/graphql/types/PollOption.ts
BartoszJarocki: This is a very good catch! Added a better comment and a fix to provide a proper id. 🙏

packages/server/graphql/mutations/createPoll.ts
BartoszJarocki: Implemented with CTE, thank you for pointing that out!

packages/server/graphql/types/Threadable.ts
```diff
@@ -67,8 +69,16 @@ const Threadable = new GraphQLInterfaceType({
   name: 'Threadable',
   description: 'An item that can be put in a thread',
   fields: threadableFields,
-  resolveType: ({status}) => {
-    return status ? Task : Comment
+  resolveType: (threadable: ThreadableDB) => {
```
BartoszJarocki, line 72: I'm not sure if `ThreadableDB` is the best name but there's nothing else that came to my mind. If you there's a better name to avoid collision with gql `Threadable` I'm happy to improve~

BartoszJarocki:
@mattkrick I'd appreciate it if you could take a look. I've fixed all the feedback. One thing is still left, related to fetching polls and poll options at once but it's not a big deal in terms of the performance so I'd leave it as-is for now. Thank you!

packages/server/graphql/types/Poll.ts
```diff
@@ -0,0 +1,70 @@
...
+      type: GraphQLNonNull(GraphQLList(GraphQLNonNull(PollOption))),
+      description: 'A list of all the poll options related to this poll',
+      resolve: async ({id: pollId}, _args, {dataLoader}) => {
+        return dataLoader.get('pollOptions').load(pollId)
+      }
+    }
+  })
+})
+
+const {connectionType, edgeType} = connectionDefinitions({
```
mattkrick, line 52: what are your plans with the connection? unless we're going to return a list of 20+ polls to the viewer we can probably remove poll connection & edge

packages/server/graphql/types/Threadable.ts
mattkrick: i'm right there with ya, i've seen some people call it `CThreadable` where `C` is for `class`, but i think the DB suffix is a little more expressive

packages/server/postgres/queries/src/insertPollWithOptionsQuery.sql
```diff
@@ -0,0 +1,20 @@
...
+   INSERT INTO "Poll" (
+       "createdById",
+       "discussionId",
+       "teamId",
+       "meetingId",
+       "threadSortOrder",
+       "title"
+   ) VALUES :poll RETURNING *
+), pollOptionsData AS (
+  SELECT "title", (SELECT id from poll AS "pollId") FROM (VALUES :pollOptions) AS pollOptionsData(title)
```
mattkrick, line 16: this is the line that caused all the headaches i bet!

mattkrick:
LGTM, only question outstanding is if we need the PollConnection or not. Good to merge after you decide!

packages/server/postgres/queries/src/insertPollWithOptionsQuery.sql
BartoszJarocki: Oh yeah. One thing I've learned about writing SQL is that you don't know the solution, start with a super simple query, data and then go towards the end goal. SQL errors are basically useless so it's hard to find out what's wrong with the query.

packages/server/graphql/types/Poll.ts
BartoszJarocki: You're right, we don't need that. My initial thought was that every threadable would have to expose the connection but now I know it's not true. Removing! ✂️

Issue #5268, opened 2021-08-12
Replace RedisLock with a queue


mattkrick:
In our system, some functions cannot get called at the exact same moment by the same user.
A good example is a GraphQL mutation called `acceptTeamInvitation`. We wouldn't want the same user to accept the team invitation twice and trigger 2 new team members!

Normally, the mutation would check the database and only continue if the `TeamMember` is not found. However, what if both calls occur at the  _exact_ same time? They'd both query the database at the same time. They'd both determine that a `TeamMember` for that user doesn't exist, and they'd both continue on to create a new `TeamMember`!

To solve race conditions like this, we implement a locking mechanism using redis (https://github.com/ParabolInc/parabol/blob/4a7ef7030825bcd2133c69875ff1f90530a0ed15/packages/server/utils/RedisLock.ts)
- `Event 1` and `Event 2` both try to achieve a lock on a redis key called `acceptTeamInvitation:userId=123`.
- `Event 1` achieves the lock, completes the mutation, calls `unlock()`, & returns successfully
- `Event 2` fails to achieve a lock, retries after `Event 1` finishes, determines that the TeamMember already exists, and exits early

This system works well if it doesn't matter which event executes first.

However, if events needs to be executed in a deterministic order, this can cause problems.
If 3 events calls `RedisLock.lock()`, then it is possible that the events are executed: `1, 2, 3` or `1, 3, 2`

This is why:
- `Event 1` gets called & triggers a lock with a max duration of 3 seconds
- `Event 2` gets called, finds the lock has 2.9 seconds left & sets a timeout to check again in 2.9 seconds
- `Event 1` finishes & calls unlock
- `Event 3` gets called, finds there is no lock, and runs to completion
- `Event 2`'s 2.9 second timeout finishes, and it runs to completion

Even if `Event 2` repeatedly checked the lock every 100ms, or 10ms, or even 1ms, there is no guarantee that it will run before `Event 3`.

**We need a way to execute tasks in serial in the order that they were queued**

Acceptance Criteria
- [ ] the locking system must scale horizontally across multiple node instances & a cluster of redis instances
- [ ] items are executed in serial in the the same order that they are queued
- [ ] A list of options with their pros/cons is written out
- [ ] a proof of concept is built in code
[...]

PR #5311, opened 2021-08-24
Implement RedisLockQueue

igorlesnenko:
#5268

In this PR I add `RedisLockQueue` which is a lock that put tasks in a queue and execute them in serial in the order that they were queued.

**It works as follows:**

- Instead of setting a single key, we use a LIST to push a request UID into it.
- Each request periodically checks if it is the first in the LIST. And if so, it is executed.
- After successful execution, the request calls `unlock` function which removes its UID from the LIST.
- Now, another request is able to find its UID in the LIST and execute

**Handling errors:**

It is theoretically possible that some of the events are failed and stuck in a queue. To handle this situations we use the following approach:
- When event starts, we append a timestamp to its UID. Each process checks if timestamp of the first event in a queue is too old or the event didn't event start. If so, it removes that event from the queue.
- If an event waits too long for its turn, it will fail with maxtime error and remove itself from the queue

### DIFF
--- a/packages/server/utils/RedisLockQueue.ts
+++ b/packages/server/utils/RedisLockQueue.ts
@@ -5,15 +5,18 @@ import getRedis from './getRedis'
 export default class RedisLockQueue {
   uid = generateUID()
   queueKey: string
-  waitTime: number
+  waitTimeMs: number
   ttl: number
   queued = false
   redis = getRedis()
 
+  // Check lock every 100 ms
+  private static readonly checkAfterMs = 100
+
   constructor(key: string, ttl: number) {
     this.queueKey = `lock:queue:${key}`
     this.ttl = ttl
-    this.waitTime = 0
+    this.waitTimeMs = 0
   }
 
   unlock = async () => {
@@ -56,21 +59,19 @@ export default class RedisLockQueue {
     return true
   }
 
-  lock = async (maxWait: number) => {
-    // Check lock every 100 ms
-    const checkAfterMs = 100
+  lock = async (maxWaitMs: number) => {
     // "Infinity" loop
     for (let i = 0; i < 1000; i++) {
       const hasLock = await this.checkLock()
       if (!hasLock) {
         return
       } else {
-        await sleep(checkAfterMs)
-        this.waitTime += checkAfterMs
-        if (this.waitTime > maxWait) {
+        await sleep(RedisLockQueue.checkAfterMs)
+        this.waitTimeMs += RedisLockQueue.checkAfterMs
+        if (this.waitTimeMs > maxWaitMs) {
           // Remove event from the queue
           await this.unlock()
-          throw new Error(`Could not achieve lock for ${this.queueKey} after ${maxWait}ms`)
+          throw new Error(`Could not achieve lock for ${this.queueKey} after ${maxWaitMs}ms`)
         }
       }
     }
--- a/packages/server/utils/RedisLockQueue.ts
+++ b/packages/server/utils/RedisLockQueue.ts
@@ -33,10 +33,8 @@ export default class RedisLockQueue {
         .exec()
       this.queued = true
       const resultingListLength = res[0][1]
-      if (resultingListLength === 1) {
-        // If the resulting LIST length is 1, we're already first on the list, return no lock
-        return false
-      }
+      // If the resulting LIST length is 1, we're already first on the list, return no lock
+      return resultingListLength !== 1
     }
     // Checking if we are the first on the list
     const head = await this.redis.lindex(this.queueKey, 0)
--- a/packages/server/utils/RedisLockQueue.ts
+++ b/packages/server/utils/RedisLockQueue.ts
@@ -32,7 +32,7 @@ export default class RedisLockQueue {
   }
 
   private markTaskAsRunning = async () => {
-    // Append timestamp to a key and replace the head
+    // Append a timestamp to uid and replace the head
     this.uidWithTimestamp = `${this.uid}::${Date.now()}`
     return this.redis.lset(this.queueKey, 0, this.uidWithTimestamp)
   }
--- a/packages/server/utils/RedisLockQueue.ts
+++ b/packages/server/utils/RedisLockQueue.ts
@@ -23,23 +23,21 @@ export default class RedisLockQueue {
     this.waitTimeMs = 0
   }
 
+  private removeItem = async (item) => {
+    return this.redis.lrem(this.queueKey, 0, item)
+  }
+
   unlock = async () => {
     if (this.timeoutId) {
       clearTimeout(this.timeoutId)
     }
-    // Remove all occurrences of the event from the list
-    return this.redis.lrem(this.queueKey, 0, this.uidWithTimestamp ?? this.uid)
+    return this.removeItem(this.uidWithTimestamp ?? this.uid)
   }
 
   private markTaskAsRunning = async () => {
     // Append a timestamp to uid and replace the head
     this.uidWithTimestamp = `${this.uid}::${Date.now()}`
     return this.redis.lset(this.queueKey, 0, this.uidWithTimestamp)
-  }
-
-  // Removes the head
-  private flushHead = async () => {
-    return this.redis.lpop(this.queueKey)
   }
 
   private setLockAndValidateQueue = async () => {
@@ -52,14 +50,14 @@ export default class RedisLockQueue {
     }
     // Checking if we are the first on the list
     const head = await this.redis.lindex(this.queueKey, 0)
-    if (head == null) {
+    if (head === null) {
       // If no head element exists, the queue was removed for some reason
       // Now we cannot guarantee the correct order of the event
       throw new Error(`Could not achieve lock for ${this.queueKey}. Queue does not exists`)
     }
 
-    const headUid = head.split(RedisLockQueue.timestampDelimiter)[0]
-    if (headUid === this.uid) {
+    const [, headTimestamp] = head.split(RedisLockQueue.timestampDelimiter)
+    if (head === this.uid) {
       return false
     } else {
       // Check if head is expired or not started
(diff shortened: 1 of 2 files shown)


packages/server/utils/RedisLockQueue.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  lock = async (maxWait: number) => {
+    // Check lock every 100 ms
+    const checkAfterMs = 100
+    // "Infinity" loop
+    for (let i = 0; i < 1000; i++) {
+      const hasLock = await this.checkLock()
+      if (!hasLock) {
+        return
```
mattkrick, line 68: The `ttl` means no operation should take more than 3 seconds.
In other words, if `unlock` doesn't get called in 3 seconds, it should get get called automatically

consider:

```ts
if (!hasLock) {
  this.timeoutId = setTimeout(() => this.unlock(), this.ttl)
  return
}
```

then in `unlock` we can call `clearTimeout(this.timeoutId)`

this protects us from developers accidentally not calling `unlock()`.
however, it still doesn't protect us from restarts.

what _does_ protect us from restarts is detecting when the `head` hasn't changed for > ttl :wink: 
i'll leave that as a hint & see what you come up with

mattkrick (commented):
great stuff!
really well thought out on the ttl, I presented a possible solution, let me know what you think

packages/server/utils/RedisLockQueue.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  checkLock = async () => {
+    // Add an event to the queue only once
+    if (!this.queued) {
+      // Use transaction to add an event to the queue and set initial LIST TTL
+      const res = await this.redis
+        .multi()
+        .rpush(this.queueKey, this.uid)
+        // TODO: Use NX option to set TTL only on LIST creation (Redis 7.0 required)
```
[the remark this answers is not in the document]
igorlesnenko: It is from here https://redis.io/commands/expire

> The EXPIRE command supports a set of options since Redis 7.0

Redis 7.0 should come this year

packages/server/utils/RedisLockQueue.ts
```diff
@@ -0,0 +1,80 @@
...
+        // If the resulting LIST length is 1, we're already first on the list, return no lock
+        return false
+      }
+    }
+    // Checking if we are the first on the list
+    const head = await this.redis.lindex(this.queueKey, 0)
+    if (head == null) {
+      // If no head element exists, the queue was removed for some reason
+      // Now we cannot guarantee the correct order of the event
+      throw new Error(`Could not achieve lock for ${this.queueKey}. Queue does not exists`)
```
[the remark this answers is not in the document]
igorlesnenko: @mattkrick We can do this, however, in this case events might be placed back in a queue in a random order.

Imagine we have 3 or more events in a queue and the queue got removed.
The first event that finds there is no head, becomes a head. The others are just got mixed together and with new events.

Is it fine?

packages/server/utils/RedisLockQueue.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+    if (head === this.uid) {
+      // We are the first on the list, no lock
+      // Refresh TTL on the LIST before running the event
+      const newTTL = await this.redis.pexpire(this.queueKey, this.ttl)
```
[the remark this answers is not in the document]
igorlesnenko: @mattkrick If we simply return `false`, it is possible that two events will run in parallel: 

* Queue got removed.
* We return `false` allowing the Event to run.
* At the same time another Event is added to a queue, sees that `LIST length === 1` and runs in parallel.

As an alternative, we can add the Event back to a queue:
```js
this.queued = false
return this.checkLock()
````
However, in this case, it is possible that before we re-add, some another Events are already added to the queue. So the order of events will be broken.

So throwing an error seems to be the only way to guarantee the correct order of events every time.

In general, I think that the whole edge case is very rare:

> someone could call .unlock() shortly after calling .lock()

This should not cause the edge case — we already added our Event to a queue before checking the `pexpire`, so at least one event will be in a queue, so the another Event’s unlock is unable remove the whole queue.

What do you think?

File: packages/server/utils/RedisLockQueue.ts (hunk 1 above)
Comment (igorlesnenko) [reply packages/server/utils/RedisLockQueue.ts:66]: @mattkrick currently implemented solution handles both automatic `unlock` and has a protection from restarts.

* As we refresh `LIST TTL` right before starting any Event, each event would have enough time to execute.
* If process crashes or developer forgets to unlock, Redis will remove the queue by `TTL` which removes the lock and allows new events to run.
* In case of restart, events are stuck in the queue and not running, so they are not refreshing the `TTL`. Redis will remove this queue along with all stuck events.

However, it has two issues: 
1. Currently, we also refresh `TTL` on adding a new event to a queue. That means, `TTL` might unexpectedly grow. For example, if someone keep adding new events fast, it prevents the stuck queue to be removed.  (It would be possible to avoid this with Redis 7.0)
2. When unlocking the queue automatically, we remove the whole queue instead of a particular stuck events.

I'm not sure if **2** is an issue, because if we need to unlock by `TTL`, it most likely means that server restart occurred, which should be a rare case.

However, considering these issues and your comments, there are several alternative solutions to TTL:

## First solution

Add `setTimeout with unlock()` to current TTL LIST approach.

**Pros**
* This will make a single stuck event to unlock the queue after TTL without delay

**Cons**
* When invalidating a queue after restart, it may still have some delays.
* We still remove the whole queue


## Second solution

As suggested, let's detect when the head hasn't changed for > ttl.

* For this purpose, for each queue we would have a corresponding separate `KEY` with `TTL` on it.
* We refresh this `KEY` `TTL` before running the event
* If another event finds that the `KEY` is missing (was removed by TTL), it drops the whole queue.
* We do not have a `TTL` on `LIST`, as it is redundant in this case
* We do not use `setTimeout with unlock()`, as it is redundant in this case

**Pros**

* No delays when doing unlock and invalidating a queue after restart

**Cons**

* Don't we just duplicate redis functionality?
* We still remove the whole queue

## Third solution

Modify the second solution and `POP` the first event instead of removing the whole queue.

It is possible. However, there is no guarantee that other events in a queue are not stuck too (e.g. server restarted). If we `POPs` a single event, we don't know whether a node process for the next event exists. We must wait for the `TTL` again.
In this case, for example, if we have 10 events stuck in a queue, we would wait for the unlock `(10 * TTL)ms` which might be too long.

**Pros**

* This will make a single stuck event to unlock the queue after TTL without delay
* Do not removes the whole queue on unlock

**Cons**

* Invalidating a restarted queue with many events in it might take a lot of time.

## Fourth solution

Have a health check as was suggested in https://github.com/ParabolInc/parabol/issues/5268#issuecomment-904721874

> 2) Have a health check for each request

> Along with adding request to a queue, for each request we set a simple key process:UID with TTL on it.
> Each request, when checks a queue, refreshes its TTL.
> If process:UID key does not exists for the UID in a queue, we know that the corresponding process crashed.
> So each event always checks the process:UID of the first item in the queue and removes the event if necessary.

**Pros**

* Seems to solve all the previously mentioned problems

**Cons**

* Adds additional complexity

Personally, I'm not sure if **2** is a big issue, therefore keeping solution as is or adding `setTimeout with unlock()` feels to be enough.

Did I understand you correctly? What do you think?

Comment (igorlesnenko) [reply packages/server/utils/RedisLockQueue.ts:91]: @mattkrick there is one issue with this approach:

There is no guarantee that `head === uidToFlush` will be checked after `100ms`. 

We have multiple tasks in a queue, and therefore multiple processes running in background. That means the check will happen at random time. It will also happen when a new event arrived.

So it is possible that head will be instantly removed before it even started.

Comment (igorlesnenko) [reply packages/server/utils/RedisLockQueue.ts:91]: @mattkrick yeah, I think this might work.

Just we need to make sure to select `waitTimeMs` (currently 100ms) to be not less than time required to start a task.

File: packages/server/utils/RedisLockQueue.ts (hunk 4 above)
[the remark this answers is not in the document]
Comment (igorlesnenko) [reply packages/server/utils/RedisLockQueue.ts]: Not relevant anymore, as we decided to not set TTL on list

File: packages/server/utils/RedisLockQueue.ts (hunk 3 above)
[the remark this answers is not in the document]
Comment (mattkrick) [reply packages/server/utils/RedisLockQueue.ts:58]: now that a queue should never disappear without manual intervention (or some case we haven't considered), we can handle this without an error:
```ts
this.queued = false
return true
```

packages/server/utils/RedisLockQueue.ts
```diff
@@ -0,0 +1,107 @@
...
+    // Add an event to the queue only once
+    if (!this.queued) {
+      const resultingListLength = await this.redis.rpush(this.queueKey, this.uid)
+      this.queued = true
+      // If the resulting LIST length is 1, we're already first on the list, return no lock
+      return resultingListLength !== 1
+    }
+    // Checking if we are the first on the list
+    const head = await this.redis.lindex(this.queueKey, 0)
+    if (head == null) {
```
mattkrick, line 55: redis should guarantee that this is exactly `null`, so let's use `===`

packages/server/utils/RedisLockQueue.ts
```diff
@@ -0,0 +1,107 @@
...
+      // If no head element exists, the queue was removed for some reason
+      // Now we cannot guarantee the correct order of the event
+      throw new Error(`Could not achieve lock for ${this.queueKey}. Queue does not exists`)
+    }
+
+    const headUid = head.split(RedisLockQueue.timestampDelimiter)[0]
+    if (headUid === this.uid) {
```
mattkrick, line 62: let's do `head === this.uid` because it tells us that no timestamp has been added yet

packages/server/utils/RedisLockQueue.ts
```diff
@@ -0,0 +1,107 @@
...
+      return false
+    } else {
+      // Check if head is expired or not started
+      const uidToFlush = this.uidToFlush
+      this.uidToFlush = ''
+      if (head === uidToFlush) {
+        // we already gave the other process 100ms to pick up the task and it didn't. it's stale.
+        await this.flushHead()
+      } else {
+        const delimiterIdx = head.lastIndexOf(RedisLockQueue.timestampDelimiter)
```
mattkrick, line 72: let's just do `const [headUid, headTimestamp] = head.split(RedisLockQueue.timestampDelimier)` so we don't have to repeat it down below

mattkrick:
just a couple housekeeping things. other than that lgtm!

igorlesnenko:
@mattkrick Found a bug — it is possible that 2 parallel processes will try to flush the same head.
However, because we use `LPOP`, they will remove two elements instead one head element. I replaced `LPOP` with `LREM`.

mattkrick:
great find! LGTM

Issue #5317, opened 2021-08-25
Error: Team member is undefined. teamMemberId is undefined. phaseType is reflect. stageId is 6kJUXBiwM2....


BartoszJarocki:
https://sentry.io/organizations/parabol/issues/2588743712/?referrer=github_plugin

This issue happened almost 800k times in last 7 days. It's not fatal but we should investigate ASAP.

```
Error: Team member is undefined. teamMemberId is undefined. phaseType is reflect. stageId is 6kJUXBiwM2. meetingId is 6kJUXyK8Bq. localStageId is 6kKuOxK6Br. stage is {"id":"6kJUXBiwM2","isComplete":true,"isNavigable":true,"isNavigableByFacilitator":true}.
  at None (./packages/client/components/MeetingSidebarTeamMemberStageItems.tsx:74:15)
  at Array.map (<anonymous>)
  at c (./packages/client/components/MeetingSidebarTeamMemberStageItems.tsx:63:31)
...
(12 additional frame(s) were not displayed)
```

PR #5320, opened 2021-08-25
Fix undefined teamMember Sentry errors

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/5317

`MeetingSidebarTeamMemberStageItems` was being rendered in phases where we don't show the teamMember avatars in the sidebar. The teamMember could be undefined in those phases which triggered the Sentry error


mattkrick:
ack! sorry for the toe stepping, i failed to assign this to myself & put it in progress

Issue #4867, opened 2021-04-07
Read User from Pg


tiffanyhan:
## Issue - Enhancement

Let's take another step moving from rethinkdb to pg

### Acceptance Criteria (optional)
- [x] #5373
- [x] #5372
- [x] #5407

- **Estimated effort:** 20hrs([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

PR #5329, opened 2021-08-27
Read autopauseUsers user from PG

Dschoordsch:
First step to replace User reads from RethinkDB with reads from Postgres.

This is a separate PR as it's independent and easy to test.

Relates-To: #4867

### DIFF
--- a/packages/server/postgres/queries/getUserIdsToPause.ts
+++ b/packages/server/postgres/queries/getUserIdsToPause.ts
@@ -1,0 +1,13 @@
+import {getUserIdsToPauseQuery} from './generated/getUserIdsToPauseQuery'
+import getPg from '../getPg'
+import catchAndLog from '../utils/catchAndLog'
+
+const getUserIdsToPause = async (activeThreshold: Date): Promise<string[]> => {
+  return (
+    (await catchAndLog(async () =>
+      (await getUserIdsToPauseQuery.run({activeThreshold}, getPg())).map((user) => user.id)
+    )) ?? []
+  )
+}
+
+export default getUserIdsToPause
--- a/packages/server/postgres/queries/src/getUserIdsToPauseQuery.sql
+++ b/packages/server/postgres/queries/src/getUserIdsToPauseQuery.sql
@@ -1,0 +1,5 @@
+/*
+  @name getUserIdsToPauseQuery
+*/
+SELECT id FROM "User"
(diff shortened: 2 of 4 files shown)


packages/server/postgres/queries/getUserIdsToPause.ts
```diff
@@ -0,0 +1,13 @@
+import {getUserIdsToPauseQuery} from './generated/getUserIdsToPauseQuery'
+import getPg from '../getPg'
+import catchAndLog from '../utils/catchAndLog'
+
+const getUserIdsToPause = async (activeThreshold: Date): Promise<string[]> => {
+  return (
```
tianrunhe, line 6: `returns await` is actually unnecessary. I recently learnt this from [Matt's comment](https://github.com/ParabolInc/parabol/pull/5201#discussion_r695311678). There's also an [ESLint rule](https://eslint.org/docs/rules/no-return-await).

tianrunhe (approved):
LGTM except one minor thing.

packages/server/postgres/queries/getUserIdsToPause.ts
Dschoordsch: Sure, saw this used in other places and it's useful for stack traces, but that's also mentioned in the lint rule, so we know the trade off.

Dschoordsch: Wanted to change it, it's actually `(await ...) ?? []`, so I cannot remove it.

Dschoordsch:
@mattkrick could you please review this?

mattkrick:
definitely! could you be sure to drag the issue (https://github.com/ParabolInc/parabol/issues/4867) to the `Maintainer Review` column when ready? I live & die by that kanban :sweat_smile:

packages/server/postgres/queries/src/getUserIdsToPauseQuery.sql
```diff
@@ -0,0 +1,5 @@
+/*
+  @name getUserIdsToPauseQuery
+*/
+SELECT id FROM "User"
+WHERE inactive = false AND ("lastSeenAt" IS NULL OR "lastSeenAt" <= :activeThreshold);
```
mattkrick, line 5: ah! this reminds me, we should probably make `lastSeenAt` non-null.
Out of scope for this PR, but will backlog it.

PR #5330, opened 2021-08-27
Add server tests

Dschoordsch:
Add some tests running graphql queries against the dev server. These
don't prime the database and thus are only meant to be run manually
locally at the moment.

Added missing LocalCache updates to updateWatchlist and updateEmail.

Relates-To: #5407


Dschoordsch:
@BartoszJarocki I would like to get this in as a starting point, so could you please review?

BartoszJarocki:
@Dschoordsch sure, will do this today 👍

packages/server/graphql/intranetSchema/mutations/updateEmail.ts
```diff
@@ -60,6 +64,7 @@ const updateEmail = {
         .run(),
       updateUser(updates, userId)
     ])
+    await dataLoader.get('users').clear(userId)
```
BartoszJarocki, line 67: What's the reason for that change? 🤔

BartoszJarocki (commented):
@Dschoordsch Great work, I've left just one question.

I think having integration tests for API is valuable but given the circumstances, I have some concerns we need to discuss. How are we're going to enforce maintaining these tests? Or adding new ones? The common scenario I saw previously was that tests that aren't run on CI and are meant to run only locally very quickly become outdated. It'd be great to have a check on Github that'd run these tests when PR is submitted but then, it's way more complex to set up as it'd require having deployment preview environments for each PR. 

Anyway, that's a nice starting point and I'm looking forward to a further discussion!

packages/server/graphql/intranetSchema/mutations/updateEmail.ts
Dschoordsch: Mutations updateWatchlist and deleteUser now clear the user from dataloader to return correct result on next query.

Dschoordsch:
I agree, I would like to have these run automatically.
In their current state we could just run those on circleci by spawning a new instance for each run including databases. I don't think test running time is a limiting factor currently.

BartoszJarocki:
@Dschoordsch let's move to maintainer review and continue the discussion!

packages/server/graphql/mutations/deleteUser.ts
```diff
@@ -80,10 +80,14 @@ export default {
...
-    setTimeout(() => {
-      db.write('User', userIdToDelete, update)
-      updateUser(update, userIdToDelete)
-    }, 30000)
+    setTimeout(
+      () =>
+        Promise.all([
+          db.write('User', userIdToDelete, update),
+          updateUser(update, userIdToDelete)
+        ]).then(() => dataLoader.get('users').clear(userIdToDelete)),
```
mattkrick, line 88: curious, why do we need to clear the dataloader here? it doesn't look like we prime the `users` dataloader anywhere above

packages/server/__tests__/common.ts
```diff
@@ -0,0 +1,148 @@
+require('../../../scripts/webpack/utils/dotenv')
+import axios from 'axios'
```
mattkrick, line 2: are you opposed to using `node-fetch` over axios? would love to continue using a single package for all HTTP reqs so we don't have to worry about different APIs & keeping both up to date wrt security stuff.

mattkrick (commented):
you, sir, have brought sanity to our world.
thanks for paving the way for these endpoint tests!
just 1 question wrt possibly dangerous dataloader reuse, aside from that :ok_hand:

packages/server/graphql/intranetSchema/mutations/updateEmail.ts
Dschoordsch: I checked again. Actually this just worked around the problem. `User` is stored in `LocalCache` and this call would clear it from that. So the real issue was a missing `db.write('User', userId, updates)` call. All calls to `updateUser` need to be paired with that and it was missing in `updateEmail` and `updateWatchlist`.
It would be best to combine those two calls into one, but putting it into `updateUser` seems like not the right place since it would mix db writes with the cache. I think long term it would be better to do reads and writes through `db` and call PG from there just like we do with the reads.

Dschoordsch: I think for transitioning `User` away from RethinkDB it's best to extend `customRedisQueries` to writes as this can then be used to write to PG instead of RethinkDB, just like I plan to do with reads: https://github.com/ParabolInc/parabol/blob/aac85bf13878746968236ee32c134fb9a4d6f0be/packages/server/dataloader/customRedisQueries.ts#L61
This would then make the direct `updateUser` call superfluous and avoid this type of error.

packages/server/graphql/mutations/deleteUser.ts
Dschoordsch: You're right, with the answer to the question above this does not make sense.

packages/server/graphql/intranetSchema/mutations/updateWatchlist.ts
```diff
@@ -68,6 +69,8 @@ const updateWatchlist = {
         .run(),
       updateUser(update, userIds)
     ])
+    const dl = dataLoader.get('users')
+    await Promise.all(userIds.map((userId) => dl.clear(userId)))
```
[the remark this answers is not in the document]
Dschoordsch: s.a.

mattkrick:
fix #5407

PR #5351, opened 2021-09-01
Manage team view

### LINKED ISSUE #5002: Adds a Manage Team side sheet to Team view
Solving #4494 

This updates the avatar group for team members and membership management on the team dashboard.

**Note**: we can probably use the updated avatar group component as long as it’s upgraded to handle new details and behaviors.

See [background and design notes](https://www.notion.so/parabol/Avatar-group-improvements-a841c58603e5474782c02a92db3a0ee7) in Notion 🔒 

#### Team view top bar
- [See Figma](https://www.figma.com/file/aFpTjk6wGGfHeYKcUXfdqn/Team-member-management?node-id=1%3A1490)
- [ ] The right side of the team dashboard top bar has Manage Team, Invite, and Agenda components
  - The Manage Team control shows member avatars and toggles the Manage Team side sheet
  - The Invite control toggles the Invite dialog
  - The Agenda control toggles the Agenda side sheet
- [ ] The avatars overlap. The avatar is 24px plus 2px border that matches the background to create the visual separation. This means that the total box is 24+2+2 = 28. It is not a 24 box with 2 + 2 pixel of border on the inside, this would make the actual avatar visual footprint be 20 which is not desired per design. The overlap is 6 pixels (4px of avatar + 2px border)
- [ ] Offline folks are now at 35% opacity. Remove the circular badge as this is an old design pattern. We need a solid background behind each avatar that matches the team dash (`SLATE_200`) so that they overlapping avatars don’t show opacity over each other
- [ ] There’s an overflow badge when applicable
- [ ] Responsive layout and overflow ([See Figma](https://www.figma.com/file/aFpTjk6wGGfHeYKcUXfdqn/Team-member-management?node-id=1%3A1960)):
  - [ ] At smaller breakpoints the manage team, invite, and avatar controls can wrap to a second row
  - [ ] We can limit the width of the avatar group to cause overflow based on how much space is available in the viewport
  - [ ] At any breakpoint, set a max width so that 10 avatars show or 9 avatars + overflow badge
- [ ] Hovering an avatar [shows the name](https://www.figma.com/file/aFpTjk6wGGfHeYKcUXfdqn/Team-member-management?node-id=2%3A194); in the future we could show [the last time they connected](https://www.figma.com/file/aFpTjk6wGGfHeYKcUXfdqn/Team-member-management?node-id=2%3A133) on a second line
- [ ] Tapping an avatar opens a Manage Team side sheet and scrolls to that user if they are scrolled out of view
- [ ] Tapping the overflow badge opens the menu and scrolls to the first user in the overflow state
- [ ] [Layout, type, color etc.](https://www.figma.com/file/aFpTjk6wGGfHeYKcUXfdqn/Team-member-management?node-id=1%3A1514) — the top bar is 72px with 40px tall components and 16px vertical gutter above and below. Icons use the 24px MD variants. The links and labels are 12px font size, 600 weight. Labels use `Slate_700`. Blue components are `Sky_500`

#### Manage team side sheet
- [See Figma](https://www.figma.com/file/aFpTjk6wGGfHeYKcUXfdqn/Team-member-management?node-id=1%3A2010)
- [ ] Each item shows a team member with a 24px avatar and their name
- [ ] The team lead has a secondary meta line ‘Team Lead’ (option: use a badge)
- [ ] If the user can take action on that item there’s a `more_vert` menu toggle
- [ ] Team leads sees a menu for all other members with the options: 1) ‘Promote NAME to Team Lead’ 2) ‘Remove NAME from Team’
- [ ] A team member sees a menu toggle on their user menu item with the option to ‘Leave Team’
- [ ] Only 1 side sheet can be open at a time, either the Agenda or the Manage Team sheet
- [ ] When a side sheet is open the toggles are still visible. They can be used to close and or swap out the content of the side sheet
- [ ] The side sheet can be closed with an X control in the upper right.
- [ ] Mobile uses [a modal side sheet](https://www.figma.com/file/aFpTjk6wGGfHeYKcUXfdqn/Team-member-management?node-id=1%3A2259) that appears over the content with a scrim. At the `SIDEBAR_LEFT` breakpoint, it uses a standard side sheet that pushes in on the right side of the content. For more see [the MD docs for side sheets](https://material.io/components/sheets-side#usage)
- [ ] Note: when a side sheet pushes into the layout (i.e. not mobile, where it overlaps with a scrim) it should push the control bars and floating action buttons, too. The Add Meeting button should be in the bottom right of the main view when any side sheet is open, not over the side sheet ([see example](https://www.figma.com/file/aFpTjk6wGGfHeYKcUXfdqn/Team-member-management?node-id=1%3A2010))

nickoferrall:
Fixes #5002


packages/server/graphql/types/TeamMember.ts
```diff
@@ -39,7 +39,11 @@ const TeamMember = new GraphQLObjectType<any, GQLContext>({
       type: GraphQLBoolean,
       description: 'true if the user is a part of the team, false if they no longer are'
     },
-    isLead: {type: GraphQLBoolean, description: 'Is user a team lead?'},
+    isLead: {
```
nickoferrall, line 42: We have `isLead` on both the `TeamMember` and the `Team`. Perhaps we should change the name on `Team` to `isViewerLead` as it's always referring to the viewer?

packages/server/database/types/TeamMember.ts
```diff
@@ -42,6 +45,7 @@ export default class TeamMember {
     this.teamId = teamId
     this.email = email
     this.hideAgenda = hideAgenda || false
+    this.hideManageTeam = hideManageTeam || true
```
nickoferrall, line 48: I'm hiding the manage team view by default as I assume the agenda list should take precedent? Manage team view is more of a settings page so I imagine users will use the agenda list more

nickoferrall: Also, as we default to true if hideManageTeam is undefined, I assume we don't need a migration here?

packages/client/components/DashboardAvatars/DashboardAvatars.tsx
```diff
@@ -1,94 +1,112 @@
...
+interface Props {
+  team: DashboardAvatars_team
+}
+
 const DashboardAvatars = (props: Props) => {
   const {team} = props
-  const {id: teamId, isLead: isViewerLead, teamMembers} = team
-  const wrapperRef = useRef<HTMLDivElement>(null)
-  const avatarsRef = useRef<HTMLDivElement>(null)
-  const maxAvatars = useAvatarsOverflow(wrapperRef, avatarsRef)
```
nickoferrall, line 68: Now that we have a limit of 10 avatars, I don't think we need to dynamically count the max number of avatars with `useAvatarsOverflow`. On mobile, we drop from 10 to 6 avatars. We could use `useAvatarsOverflow` to see if there's space for 1 or 2 more avatars on mobile, i.e. show 7/8 avatars, but I think the cost of the extra code outweighs the benefit of 1/2 more avatars on mobile

packages/client/modules/teamDashboard/components/TeamDashMain/TeamDrawer.tsx
```diff
@@ -0,0 +1,111 @@
...
+          manageTeamMemberId
+        }
+      }
+    `,
+    props.viewer
+  )
+  const {dashSearch, team, teamMember} = data
+  const hideAgenda = teamMember?.hideAgenda
+  const hideManageTeam = teamMember?.hideManageTeam
+  const manageTeamMemberId = teamMember?.manageTeamMemberId
```
nickoferrall, line 67: `team` and `teamMember` should never be null as we're grabbing the `team` and `teamMember` that belong to the viewer. I could destructure it, use a bang, and save a couple of lines but I did see a bug recently where a user deleted their account, signed up with the same email, and then didn't belong to any teams initially. So, I'd prefer to play it safe here

nickoferrall:
@ackernaut would you mind reviewing this PR please?

ackernaut:
Taking a card to review

ackernaut:
Review in progress…

### Acceptance criteria
#### Team view top bar
- [See Figma](https://www.figma.com/file/aFpTjk6wGGfHeYKcUXfdqn/Team-member-management?node-id=1%3A1490)
- [x] The right side of the team dashboard top bar has Manage Team, Invite, and Agenda components
  - The Manage Team control shows member avatars and toggles the Manage Team side sheet
  - The Invite control toggles the Invite dialog
  - The Agenda control toggles the Agenda side sheet
- [x] The avatars overlap. The avatar is 24px plus 2px border that matches the background to create the visual separation. This means that the total box is 24+2+2 = 28. It is not a 24 box with 2 + 2 pixel of border on the inside, this would make the actual avatar visual footprint be 20 which is not desired per design. The overlap is 6 pixels (4px of avatar + 2px border)
- [x] Offline folks are now at 35% opacity. Remove the circular badge as this is an old design pattern. We need a solid background behind each avatar that matches the team dash (`SLATE_200`) so that they overlapping avatars don’t show opacity over each other
- [ ] There’s an overflow badge when applicable
- [ ] Responsive layout and overflow ([See Figma](https://www.figma.com/file/aFpTjk6wGGfHeYKcUXfdqn/Team-member-management?node-id=1%3A1960)):
  - [ ] At smaller breakpoints the manage team, invite, and avatar controls can wrap to a second row
  - [ ] We can limit the width of the avatar group to cause overflow based on how much space is available in the viewport
  - [ ] At any breakpoint, set a max width so that 10 avatars show or 9 avatars + overflow badge
- [x] Hovering an avatar [shows the name](https://www.figma.com/file/aFpTjk6wGGfHeYKcUXfdqn/Team-member-management?node-id=2%3A194); in the future we could show [the last time they connected](https://www.figma.com/file/aFpTjk6wGGfHeYKcUXfdqn/Team-member-management?node-id=2%3A133) on a second line
- [ ] Tapping an avatar opens a Manage Team side sheet and scrolls to that user if they are scrolled out of view
- [ ] Tapping the overflow badge opens the menu and scrolls to the first user in the overflow state
- [ ] [Layout, type, color etc.](https://www.figma.com/file/aFpTjk6wGGfHeYKcUXfdqn/Team-member-management?node-id=1%3A1514) — the top bar is 72px with 40px tall components and 16px vertical gutter above and below. Icons use the 24px MD variants. The links and labels are 12px font size, 600 weight. Labels use `Slate_700`. Blue components are `Sky_500`

#### Manage team side sheet
- [See Figma](https://www.figma.com/file/aFpTjk6wGGfHeYKcUXfdqn/Team-member-management?node-id=1%3A2010)
- [x] Each item shows a team member with a 24px avatar and their name
- [x] The team lead has a secondary meta line ‘Team Lead’ (option: use a badge)
- [x] If the user can take action on that item there’s a `more_vert` menu toggle
- [x] Team leads sees a menu for all other members with the options: 1) ‘Promote NAME to Team Lead’ 2) ‘Remove NAME from Team’
- [x] A team member sees a menu toggle on their user menu item with the option to ‘Leave Team’
- [x] Only 1 side sheet can be open at a time, either the Agenda or the Manage Team sheet
- [x] When a side sheet is open the toggles are still visible. They can be used to close and or swap out the content of the side sheet
- [x] The side sheet can be closed with an X control in the upper right.
- [x] Mobile uses [a modal side sheet](https://www.figma.com/file/aFpTjk6wGGfHeYKcUXfdqn/Team-member-management?node-id=1%3A2259) that appears over the content with a scrim. At the `SIDEBAR_LEFT` breakpoint, it uses a standard side sheet that pushes in on the right side of the content. For more see [the MD docs for side sheets](https://material.io/components/sheets-side#usage)
- [x] Note: when a side sheet pushes into the layout (i.e. not mobile, where it overlaps with a scrim) it should push the control bars and floating action buttons, too. The Add Meeting button should be in the bottom right of the main view when any side sheet is open, not over the side sheet ([see example](https://www.figma.com/file/aFpTjk6wGGfHeYKcUXfdqn/Team-member-management?node-id=1%3A2010))

#### Detail burndown
- [ ] Tapping the `Manage Team` label under the avatars should also toggle the side sheet
- [ ] The team name and controls need a font-size of 20px and 12px to match the design concept and line up with the right-side controls (see last checklist item under Team view top bar above)
- [ ] The right-side controls break too soon? (Try contracting the window horizontally)
- [ ] The icon controls have icons that need to be centered vertically 24px over 16px to line up with 24px avatar visual footprint
- [ ] Check 1 TS error
- [ ] The X control that closes the side sheet uses the 18px MD icon variant, not 24px
- [ ] The side sheet label is 12/18 (font-size/line-height|box-height) and lines up with the X close icon
- [ ] The Team Lead label uses Slate_800 in the team member list
- [ ] The more_vert icon uses Slate_600 in the team member list

#### Todo
- I haven’t tested many avatars and overflow
- I haven’t tested tapping and it scrolling to the user

nickoferrall:
Thank you for the Saturday review @ackernaut! It was kind of you to squeeze this in before the public holiday.

I've pushed an update with your suggestions. One thing to note:

> The icon controls have icons that need to be centered vertically 24px over 16px to line up with 24px avatar visual footprint

The avatars are 24px but with the 2px transparent border, their height is 28px. The figma sketch shows 24px avatars & icons with 16px labels. To keep the overall height at 40px, I've made the height of the avatars, icons, and team name 28px (while keeping the fontSizes 24px) and the height of the labels 12px.

Alternatively, I could set a 16px line height on the labels and a 24px height on the avatars even though they're 28px with the border, but the label position is then a little close to the avatars imo.

ackernaut:
@nickoferrall I see the issue, thanks for explaining. I was hanging the 2px border around the avatars outside of the 24px block, into the 16px block that holds the label, which looks good to my eyes. But when that 2px border causes the box to be 28px…I’ll think about it. I’d hate to make the layout for the avatars any more complicated than it is.

ackernaut:
An update on the review:
- [x] I agree 12px is too short for those labels. I pushed a quick PR to this branch that uses 28 over 16 for the horizontal alignment
- [x] The overflow works
- [x] Tapping scrolls to the user in a long list

Note: when I was tapping the avatars with a larger team of fake users it felt unexpected for it to close when the side sheet was open. I expected it to scroll around to whatever user I tapped. For your consideration, maybe the Manage Team side sheet only closes when tapping X or tapping the Manage Team label. Tapping avatars either opens the side sheet, and then tries to scroll if necessary, or if the side sheet is open, tries to scroll? It may be weird when nothing happens but I think that may be a better alternative. It’s nuanced. I’m open to your opinion.

I’m gonna read through the code quickly. But if you want a thorough code review you may want to ask a full-stack dev.

packages/server/database/types/TeamMember.ts
ackernaut: Yes, I think hiding manage team view by default is a good choice

packages/client/mutations/ToggleManageTeamMutation.ts
```diff
@@ -0,0 +1,64 @@
...
+        error {
+          message
+        }
+      }
+      ...ToggleManageTeamMutation_teamMember @relay(mask: false)
+    }
+  }
+`
+
+const ToggleManageTeamMutation: StandardMutation<TToggleManageTeamMutation> = (
```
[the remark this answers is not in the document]
ackernaut: One day we may use the right side sheet for lots of things, but I agree here. This is just a short-term solution for the team membership management until we revisit the Team UX.

packages/client/components/DashboardAvatars/DashboardAvatars.tsx
ackernaut: That works for me in this case

ackernaut (changes requested):
Great work here @nickoferrall thanks for running through the list of small details.

If you would, review my small PR and merge if that works for you. Then I’d request a separate code review of a full-stack developer.

nickoferrall:
Thanks @ackernaut! 

Requesting a review from @mattkrick. I'd be great to get your thoughts on Terry's comment above: when the manage side sheet is open and you click on a team member avatar, should the side sheet close (current behaviour) or scroll to that user if they're not in view?

mattkrick:
i like it closing for 3 reasons:
- there's a scrim on mobile & clicking a scrim should always close the drawer
- the mobile behavior is the same as the desktop behavior
- every other component that servers as a menu toggle operates as a toggle

 1 nit on the sorting:
- :heart: the alphanumeric sort in the drawer
- For the avatars, I'd like to sort by: Viewer >> Online >> Alphanumeric. I should always be the first avatar. Not only because we humans are myopic a-holes, but because it guarantees at least 1 avatar has full opacity & seeing that full opacity feels good for the Zoe Zigglers of the world. After that, I wanna see who else is online so I get a sense of togetherness. Then, finally show the riff raff offliners.

mattkrick (changes requested):
fantastic work! pretty minor stuff in comments

packages/server/graphql/mutations/toggleTeamDrawer.ts
```diff
@@ -0,0 +1,65 @@
...
+import {getUserId, isTeamMember} from '../../utils/authorization'
+import ToggleTeamDrawerPayload from '../types/ToggleTeamDrawerPayload'
+import {GQLContext} from '../graphql'
+import standardError from '../../utils/standardError'
+import TeamDrawerEnum from '../types/TeamDrawerEnum'
+import {DrawerTypes} from 'parabol-client/types/constEnums'
+
+const toggleTeamDrawer = {
+  type: GraphQLNonNull(ToggleTeamDrawerPayload),
+  description: `Show/hide the drawer in the team dashboard`,
```
nickoferrall, line 12: In this issue & PR, we've used the terms sidebar, drawer, and side sheet. I'd like to be consistent with the language so I've used drawer as it seems (according to [this](https://stackoverflow.com/questions/61074330/what-is-the-difference-between-matsidenav-and-the-matdrawer#:~:text=The%20difference%20between%20the%20two,the%20user%20to%20filter%20files) and [this](https://material.io/components/navigation-drawer)) that sidebars are global, e.g. for navigation, and drawers are more component-specific. Curious if you have any opinions on this!

packages/client/components/StartMeetingFAB.tsx
```diff
@@ -40,43 +41,54 @@ const MeetingLabel = styled('div')<{isExpanded: boolean}>(({isExpanded}) => ({
   width: isExpanded ? ElementWidth.NEW_MEETING_FAB : 0
 }))
 
-interface Props { }
+interface Props {
+  isAbsolute?: boolean
```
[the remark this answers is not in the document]
nickoferrall: Nice, really appreciate the explanation! That makes a lot of sense. I like the `className` solution as we're only using the absolute position on one occasion. If it happens three times, I'd say we should switch to the declarative solution

nickoferrall:
Thanks @mattkrick! I've pushed an update with your feedback. I haven't been able to merge with master though as I run into this error mentioned here: https://github.com/ParabolInc/parabol/pull/5396#issuecomment-926762610

Once that's resolved, I'll be able to merge with master, fix the conflicts and typecheck issues, and request a follow-up review

packages/server/graphql/mutations/toggleTeamDrawer.ts
mattkrick: drawer is definitely what we should be calling it, especially because you can pull it out like a drawer.

we called the leftnav a sidebar because it used to always be on the side & you couldn't close it like a drawer, but now it definitely is one, but renaming a bunch of components doesn't sound like fun

packages/client/components/DashboardAvatars/DashboardAvatars.tsx
```diff
@@ -1,94 +1,142 @@
...
+  const {id: teamId, teamMembers} = team
+  const isDesktop = useBreakpoint(Breakpoint.SIDEBAR_LEFT)
+  const atmosphere = useAtmosphere()
+  const {viewerId} = atmosphere
+  const maxAvatars = isDesktop ? 10 : 6
   const overflowCount = teamMembers.length > maxAvatars ? teamMembers.length - maxAvatars + 1 : 0
-  const visibleAvatars = overflowCount === 0 ? teamMembers : teamMembers.slice(0, maxAvatars - 1)
+  const sortedAvatars = useMemo(() => {
+    const visibleAvatars = overflowCount === 0 ? teamMembers : teamMembers.slice(0, maxAvatars - 1)
```
mattkrick, line 85: you'll want to slice after you do the sort.
since slice is mutative, i'd do something like 
```ts
teamMembers.slice().sort((a,b) => doStuff(a,b)).slice(0, maxAvatars -1)
```

packages/server/graphql/types/ToggleTeamDrawerPayload.ts
```diff
@@ -0,0 +1,24 @@
+import {GraphQLBoolean, GraphQLNonNull, GraphQLObjectType} from 'graphql'
+import {GQLContext} from '../graphql'
+import makeMutationPayload from './makeMutationPayload'
+
+export const ToggleTeamDrawerSuccess = new GraphQLObjectType<any, GQLContext>({
+  name: 'ToggleTeamDrawerSuccess',
+  fields: () => ({
```
mattkrick, line 7: if you return a `team` object then relay will be able to merge the new team object with the old 1 & get the right values so you don't have to write an updater for that mutation on the client

packages/server/graphql/types/TeamMember.ts
```diff
@@ -39,17 +39,26 @@ const TeamMember = new GraphQLObjectType<any, GQLContext>({
...
       description: 'true if the user prefers to not vote during a poker meeting',
       resolve: ({isSpectatingPoker}) => !!isSpectatingPoker
     },
     hideAgenda: {
       type: new GraphQLNonNull(GraphQLBoolean),
-      description: 'hide the agenda list on the dashboard',
+      description: 'hide the agenda list drawer on the dashboard',
       resolve: ({hideAgenda}) => !!hideAgenda
     },
+    hideManageTeam: {
```
mattkrick, line 57: instead of `hideAgenda` and `hideManageTeam` could we return `openDrawer: {type: DrawerTypeEnum}`? it should be null if nothing should be open. that way there's no ambiguity on what to show if both are false or both are true

packages/server/graphql/types/TeamDrawerEnum.ts
```diff
@@ -0,0 +1,12 @@
+import {GraphQLEnumType} from 'graphql'
+
+const TeamDrawer = new GraphQLEnumType({
```
mattkrick, line 3: this is a great single source of truth!
it might also be nice to export a string union here so you can use it on the server: `export type TeamDrawer = 'agenda' | 'manageTeam'

packages/server/graphql/mutations/toggleTeamDrawer.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    // RESOLUTION
+    const userId = getUserId(authToken)
+    const viewerTeamMemberId = `${userId}::${teamId}`
+    return r
+      .table('TeamMember')
+      .get(viewerTeamMemberId)
+      .update(
+        (teamMember) => ({
+          hideManageTeam: r.branch(
```
mattkrick, line 44: since the drawer can only show 1 thing at a time, let's just store a single enum in the DB instead of 2 booleans

packages/client/types/constEnums.ts
```diff
@@ -69,6 +66,11 @@ export const enum DragAttribute {
   DROPZONE = 'data-dropzone'
 }
 
+export const enum DrawerTypes {
```
mattkrick, line 69: since we've already defined this as `TeamDrawerEnum` on the server, we should use the server definition as the single source of truth.

on the client we can access it like  `import {TeamDrawerEnum} from '../__generated__/ToggleTeamDrawerMutation.graphql`

for context, i think it generally makes sense to opt for string unions over enums if the values of the string unions are human readable (eg they are names like "agenda" and aren't color hex codes)

packages/client/mutations/ToggleTeamDrawerMutation.ts
```diff
@@ -0,0 +1,73 @@
...
+
+const ToggleTeamDrawerMutation: StandardMutation<TToggleTeamDrawerMutation> = (
+  atmosphere,
+  variables,
+  {onError, onCompleted}
+) => {
+  return commitMutation<TToggleTeamDrawerMutation>(atmosphere, {
+    mutation,
+    variables,
+    updater: (store) => {
```
mattkrick, line 36: if the mutation returns a team object you can completely remove the updater. yay less code!

mattkrick (changes requested):
looking good! just a few more changes

packages/server/graphql/types/Organization.ts
```diff
@@ -141,10 +142,9 @@ const Organization: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<a
       type: new GraphQLNonNull(OrganizationUserConnection),
       resolve: async ({id: orgId}, _args, {dataLoader}) => {
         const organizationUsers = await dataLoader.get('organizationUsersByOrgId').load(orgId)
-        organizationUsers.sort((
-          a: OrganizationUser,
-          b: OrganizationUser
-        ) => (a.orgId > b.orgId ? 1 : -1))
+        organizationUsers.sort((a: OrganizationUser, b: OrganizationUser) =>
```
nickoferrall, line 145: Ahh sorry, it looks like there's a bunch of linting changes after merging with master. My prettier settings haven't changed, and I don't see any eslint/prettier node module upgrades. Perhaps I hit save all while merging, and that triggered some linting changes 🤷‍♂️

packages/client/components/DashboardAvatars/DashboardAvatars.tsx
nickoferrall: I think that splice is mutative, but [slice](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice) is non-mutative - it returns a new array? I think your suggestion of slicing after the sort definitely makes sense though, as we want to sort the entire array, not just the first 10 avatars. I've pushed an update with: `teamMembers.sort((a,b) => doStuff(a,b)).slice(0, maxAvatars -1)`

nickoferrall:
Thanks, Matt - I've pushed those updates! I'm now saving the `openDrawer` enum to the `teamMember`

mattkrick:
just 1 change wrt an idempotent mutation. feel free to merge after that's fixed!

nickoferrall:
> just 1 change wrt an idempotent mutation

Sorry, what change are you referring to?

mattkrick (changes requested):
only necessary change is making the mutation take a nullable openDrawer type

mattkrick:
> > just 1 change wrt an idempotent mutation
> 
> Sorry, what change are you referring to?

oops! review submitted

packages/server/graphql/types/Organization.ts
nickoferrall: Ahh yes, it looks like `"printWidth": 100` wasn't applied in [that PR](https://github.com/ParabolInc/parabol/pull/5403/files). @tiffanyhan do you know whether you have a vscode plugin that might have changed the `printWidth` in that PR?

packages/client/modules/teamDashboard/components/TeamDashMain/TeamDrawer.tsx
```diff
@@ -64,44 +64,38 @@ const TeamDrawer = (props: Props) => {
...
-      ? DrawerTypes.AGENDA
-      : hideManageTeam === false
-      ? DrawerTypes.MANAGE_TEAM
-      : null
 
   // as drawer is closing, teamDrawerType is null. use ref to show prev content
-  if (teamDrawerType && drawerTypeRef.current !== teamDrawerType) {
-    drawerTypeRef.current = teamDrawerType
+  if (openDrawer && drawerTypeRef.current !== openDrawer) {
+    drawerTypeRef.current = openDrawer
```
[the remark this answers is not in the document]
nickoferrall: True, the reason I opted for this solution is if we just use `openDrawer`, as the drawer is closing, the content will be empty / incorrect as `openDrawer` will be `null` for the 300ms it takes to close. By using the ref, when the drawer is closing and `openDrawer` is `null`, it'll still show the content in the drawer.

packages/server/graphql/mutations/toggleTeamDrawer.ts
```diff
@@ -22,7 +21,7 @@ const toggleTeamDrawer = {
   },
   resolve: async (
     _source,
-    {teamId, teamDrawerType}: {teamId: string; teamDrawerType: DrawerTypes},
+    {teamId, teamDrawerType}: {teamId: string; teamDrawerType: TeamDrawer},
```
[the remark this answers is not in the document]
nickoferrall: Makes sense! I've updated `teamDrawerType` to be nullable which means `CloseDrawer.tsx` doesn't need to know what type of drawer it's in 🎉   

However, I don't think any changes need to be made to the rethinkdb query or optimistic updater? If the client sends `teamDrawerType` as null, `openDrawer` will become null. If they pass in `teamDrawerType` as `agenda`, and `agenda` is the current `openDrawer`, it'll close.

packages/server/graphql/types/Organization.ts
tiffanyhan: ah my fault, i was on a fresh install of vscode and didn't have prettier installed. installed now, adding linting to github actions sounds nice too

Issue #5326, opened 2021-08-26
Error: No payload received


BartoszJarocki:
https://sentry.io/organizations/parabol/issues/2399285612/?referrer=github_plugin

It's not fatal but it's an indication that a user can't do much in our app, ie. I was getting this during the login, signup when API was down. 

```
Error: No payload received
  at Object.fetch (./node_modules/@mattkrick/graphql-trebuchet-client/lib/GQLHTTPClient.js:17:23)
```

mattkrick:
still a problem in 6.25.0

AC
- [ ] catch this error & log it with the name of the query, variables, & the auth token

Estimate: 4 hours

mattkrick:
we hit this in our check-in meeting! https://sentry.io/organizations/parabol/issues/2399285612/?environment=client&project=107196&query=is%3Aunresolved

bumping to p1

tiffanyhan:
the error originates from `node_modules/@mattkrick/graphql-trebuchet-client/lib/GQLHTTPClient.js`, where the error is passed to `sink.error` method. When `QueryRenderer` does its thing, the error is thrown and then caught by `ErrorBoundary` where it's sent to sentry.

i think to capture more info, such as query and variables, we should `console.info` right before the error is thrown and then caught in `node_modules/@mattkrick/graphql-trebuchet-client/lib/GQLHTTPClient.js`. i don't think it's preferable to make the change in lib code, but it doesn't seem like there's a way to "catch" the error in app code and then log verbose details on the query and variables.

@mattkrick what do you think?

tiffanyhan:
As an aside, when looking through user feedback on sentry: https://sentry.io/organizations/parabol/issues/2399285612/feedback/?project=107196&referrer=github_plugin , it looks like the error occurs randomly. i.e. users reporting it happened when trying to load timeline, adding a card, logging in, adding a comment, etc.

i'm thinking logging the specific query and variables might not help us much, since the error doesn't occur deterministically from one or even a few actions. it seems like looking at server stats and metrics with times correlating to the error events might be a more promising route.

cc. @mattkrick @adaniels-parabol any thoughts?

tiffanyhan:
[...]

PR #5353, opened 2021-09-01
log verbose info when client receives no payload

tiffanyhan:
This PR sends more info to sentry so we can better debug and fix https://github.com/ParabolInc/parabol/issues/5326

**AC:**
I tested this in dev by replacing `fetch` function in `node_modules/@mattkrick/graphql-trebuchet-client/lib/GQLHTTPClient.js` with:
```
    async fetch(payload, sink) {
        console.log('fetching in GQL http client')
        try {
            const res = await this.fetchData({type: 'start', payload})
            // if (!sink)
            //     return
            // if (res?.payload) {
            //     sink.next(res.payload)
            //     sink.complete()
            // }
            // else {
            // console.info('payload:', payload)
            throw new Error('No payload received')
            // }
        }
        catch (e) {
            if (!sink)
                return
            sink.error(e)
        }
    }
```
thereby forcing the error to be thrown always. May need to nuke `dev/dll/yarn.lock.md5` to clear the library cache.

 - [ ] query, variables, & the auth token are logged when the error occurs


packages/client/Atmosphere.ts
```diff
@@ -286,7 +286,17 @@ export default class Atmosphere extends Environment {
 
   handleFetch: FetchFunction = (request, variables, _, uploadables) => {
     return Observable.create((sink) => {
-      this.handleFetchPromise(request, variables, uploadables, sink)
+      const verboseSink = Object.assign({}, sink, {
+        error: (e, isUncaughThrownError) => {
+          if (e.message === 'No payload received') {
+            console.info('request:', request)
+            console.info('variables:', variables)
+            console.info('auth token:', this.authToken)
```
BartoszJarocki, line 294: @tiffanyhan are these going to be logged in Sentry? To make sure we catch that we can also use: https://docs.sentry.io/platforms/javascript/enriching-events/breadcrumbs/ 
Also, I think it'd be better to avoid console logging auth token. Sentry will tell us who's affected by this issue anyway.  Let me know what do you think!

tiffanyhan: yeah i assume they will be logged bc `Error: No payload received` is logged as a breadcrumb in sentry when it's logged in the lib code.

i can definitely use the explicit API which will keep our server logs cleaner. The only problem is, now i can't test locally. i set the `SENTRY_DSN` to dev value locally, but when i check [sentry](https://sentry.io/organizations/parabol/issues/?project=107915&query=&statsPeriod=14d) i can see server errors being sent to sentry but not client errors. and i can see the correct dsn env value set on the window object... 🤔

i had the same thought with the authToken since we already see userId in sentry, but was just going by the AC. happy to remove the authToken part.

BartoszJarocki (approved):
Looks good to me!

packages/client/Atmosphere.ts
BartoszJarocki: hmm! the AC part of logging auth token is a good question, @mattkrick do we really want to do it? 
Regarding local testing, I think it _should_ work. One way or another, this error probably will be hard to reproduce locally 😞

mattkrick:
The big A-HA moment for me here was `'No payload received'`. We should be sending a payload from the server that contains errors & no data.

That told me the server wasn't sending the "timeout" error.
I pushed a fix so the server sends a timeout error. Also fixed the client so it converts that into an Error instance.

Now that the client is getting the correct error message, I don't think we need to log the error from the client because the server is already doing that every time a timeout is hit.
Next step is to offer an action for the user to take--- maybe click a button to retry the query? See #5406

PR #5356, opened 2021-09-02
Upgrade to TypeScript v4.4.2

### LINKED ISSUE #5355: Upgrade to TypeScript v4.4.2
Upgrade TypeScript to fix type errors in #5351 and to get access to new features such as better catch error types and improved control flow analysis (see [here](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables)).

See [this conversation](https://parabol.slack.com/archives/C836NA350/p1630502503048500) in Slack for further context.

nickoferrall:
Fixes #5355 

Most changes are due to the catch error update: error is now `unknown` by default instead of `any`: https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables


packages/client/utils/getDraftCoords.ts
```diff
@@ -18,7 +18,7 @@ const getDraftCoords = () => {
   while (target && target !== document) {
     // make sure the selection is inside draft, this isn't always guaranteed
     if ((target as any).className === 'notranslate public-DraftEditor-content') {
-      return getVisibleSelectionRect(window) as FakeClientRect
+      return getVisibleSelectionRect(window) as ClientRect
```
nickoferrall, line 12: It looks like `FakeClientRect` has the same properties as `ClientRect` except the `height` and `width` values are `readonly` in `ClientRect`. Where `getDraftCoords` is being used, the return value is often being used as type `ClientRect` so I think it's best to cast it here. For example, `cachedCoords` is being used [here](https://github.com/ParabolInc/parabol/blob/df50596b028bd866e4800b36cf36f0b034db5521/packages/client/components/TaskEditor/withEmojis.tsx#L95) as the `originCoords` prop which is of type `ClientRect`.

packages/client/components/VideoControls.tsx
```diff
@@ -45,7 +45,10 @@ const VideoControls = (props: Props) => {
   })
 
   const addMedia = async () => {
-    const descriptors = ['camera', 'microphone'] as PermissionName[]
+    const descriptors = ['camera', 'microphone'] as Extract<
+      'camera' | 'microphone',
+      PermissionName
+    >[]
```
nickoferrall, line 51: This previously returned the error:
```
Conversion of type '("camera" | "microphone")[]' to type 'PermissionName[]' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
```
I think `Extract` is cleaner than casting it to `unknown` and then to `PermissionName`

nickoferrall:
Señor @BartoszJarocki, I see that you liked the issue 😄  Would you mind reviewing this PR when you get the chance?

packages/client/components/VideoControls.tsx
BartoszJarocki: This is funny. It's actually a bug in TS. 
https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/main/baselines/dom.generated.d.ts#L17613 - these permissions are simply missing that's why casting fails. If they were present in the type definition the simplest solution would be to use `const descriptors = ['camera', 'microphone'] as const` - https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-inference 

TS has some issues with these types for some reason. That kind of issues happen from time to time: https://github.com/microsoft/TypeScript/issues/33923

packages/client/hooks/usePortal.tsx
```diff
@@ -33,7 +33,7 @@ const getParent = (parentId: string | undefined) => {
 const usePortal = (options: UsePortalOptions = {}) => {
   const portalRef = useRef<HTMLDivElement>()
   const originRef = useRef<HTMLElement>()
-  const timeoutRef = useRef<number | null>(null)
+  const timeoutRef = useRef<number | NodeJS.Timeout | null>(null)
```
BartoszJarocki, line 36: I'm not sure what's the right thing to do here because I'm pretty sure this code only runs in the browser (or am I wrong?). If that's the case then the TS is wrong here and the only way to let TS know we want the browser version of set timeout is to use `window.setTimeout` directly which is pretty dumb. Anyway, I'm sharing my thoughts here 😅

packages/server/graphql/handleGraphQLTrebuchetRequest.ts
```diff
@@ -57,16 +57,18 @@ const handleGraphQLTrebuchetRequest = async (
...
-          tags: {
-            authToken: JSON.stringify(authToken),
-            docId: docId || '',
-            query: query || '',
-            variables: JSON.stringify(variables)
-          }
-        })
-        return {data: null, errors: [{message: 'The request took too long'}]}
+      if (e instanceof Error) {
+        if (e.message === 'TIMEOUT') {
```
BartoszJarocki, line 61: To avoid nested if statements I'd change the first conditional to `e instanceof Error && e.message === 'TIMEOUT'`

packages/server/graphql/intranetSchema/mutations/draftEnterpriseInvoice.ts
```diff
@@ -120,7 +120,10 @@ export default {
     try {
       user = await getBillingLeaderUser(email, orgId, dataLoader)
     } catch (e) {
-      return {error: {message: e.message}}
+      if (e instanceof Error) {
+        return {error: {message: e.message}}
+      }
+      return {error: {message: 'Unable to get billing leader user'}}
```
BartoszJarocki, line 126: Not a big deal and probably depends on personal preferences but this could be simplified to 
```
      return {
        error: {message: e instanceof Error ? e.message : 'Unable to get billing leader user'}
      }
```
or 
```
      const message = e instanceof Error ? e.message : 'Unable to get billing leader user'
      return {
        error: {message}
      }
```

I don't have any strong preferences here so feel free to do whatever you think is best 👍🏻

packages/server/graphql/intranetSchema/mutations/loginSAML.ts
```diff
@@ -58,7 +58,10 @@ const loginSAML = {
     try {
       loginResponse = await serviceProvider.parseLoginResponse(idp, 'post', {body})
     } catch (e) {
-      return {error: {message: e.message}}
+      if (e instanceof Error) {
+        return {error: {message: e.message}}
+      }
+      return {error: {message: `parseLoginResponse failed: ${JSON.stringify(e)}`}}
```
BartoszJarocki, line 64: Same as above

BartoszJarocki (commented):
Nice work! I've left a couple of comments to consider.

packages/client/hooks/usePortal.tsx
nickoferrall: `timeoutRef` was causing a typecheck issue but it looks like upgrading to v4.4.2 has fixed this so I've reverted `usePortal` back to the way it was

packages/client/components/VideoControls.tsx
nickoferrall: Nice find! I've pushed an update adding a comment with a link to an issue and I'm now casting the name rather than using `Extract` (I think it's a tiny bit cleaner). I was tempted to create a PR to fix the issue but it looks like camera & microphone are considered [at risk](https://w3c.github.io/permissions/#dom-permissionname-camera)

nickoferrall:
Thanks for the review @BartoszJarocki - I've pushed an update! Please let me know if you're happy with the changes and, if so, I'll move it along to maintainer review

BartoszJarocki (approved):
Nice work!

packages/server/graphql/mutations/pokerSetFinalScore.ts
```diff
@@ -137,7 +138,7 @@ const pokerSetFinalScore = {
             dimensionField?.fieldType === 'string' ? finalScore : Number(finalScore)
           await manager.updateStoryPoints(cloudId, issueKey, updatedStoryPoints, fieldId)
         } catch (e) {
-          return {error: {message: e.message}}
+          standardError(e)
```
mattkrick, line 141: missing return statement?
for whatever reason, `Error` instances can get swallowed & unreported (maybe their stack trace is too long? :shrug: ), so I always just return a POJO.
We should return `{error: {message}}` here

packages/server/graphql/mutations/setTaskEstimate.ts
```diff
@@ -124,7 +124,9 @@ const setTaskEstimate = {
           const updatedStoryPoints = fieldType === 'string' ? value : Number(value)
           await manager.updateStoryPoints(cloudId, issueKey, updatedStoryPoints, fieldId)
         } catch (e) {
-          return {error: {message: e.message}}
+          return e instanceof Error
+            ? e
```
mattkrick, line 128: let's return a POJO here instead of an Error instance

packages/server/graphql/mutations/upgradeToPro.ts
```diff
@@ -52,7 +52,9 @@ export default {
     try {
       await upgradeToPro(orgId, stripeToken, email)
     } catch (e) {
-      return standardError(e.param ? new Error(e.param) : e, {userId: viewerId, tags: e})
+      const param = (e as any)?.param
```
mattkrick, line 55: let's put this in a conditional like
```ts
if ('param' in e) {
  const error = new Error(e.param)
  return ...
} else {
 ...
}
```

packages/server/utils/sendToSentry.ts
```diff
@@ -11,10 +11,9 @@ export interface SentryOptions {
 }
 
 // Even though this is a promise we'll never need to await it, so we'll never need to worry about catching an error
-const sendToSentry = async (error: Error, options: SentryOptions = {}) => {
-  // if (!PROD) {
+const sendToSentry = async (err: Error | unknown, options: SentryOptions = {}) => {
```
mattkrick, line 14: let's remove `unknown` from the input union
Some objects don't serialize & if that happens we'll be left with an empty error message

getting into the weeds, `JSON.stringify` simply calls an objects `toJSON` method (or the prototypical parent)
That means I can break it real easy (and many libs do, for good reasons)

e.g.
```ts
JSON.stringify({error: 'very important', toJSON: () => 'eat me'})
```

packages/server/utils/standardError.ts
```diff
@@ -1,6 +1,7 @@
 import sendToSentry, {SentryOptions} from './sendToSentry'
 
-const standardError = (error: Error, options: SentryOptions = {}) => {
+const standardError = (err: Error | unknown, options: SentryOptions = {}) => {
```
mattkrick, line 3: remove `unknown` union member

mattkrick (changes requested):
thanks so much for tackling this so quickly.
couple minor things, good to merge after the little fixes!

packages/server/utils/sendToSentry.ts
nickoferrall: Ahh that's good to know! I was trying to retain the error and save code by avoiding:
```
catch(e){
    const message = e instanceof Error ? e.message : 'Custom error message'
    sendToSentry(new Error(message))
```
in each catch that has a sendToSentry call but it looks like that's the best bet

packages/server/graphql/mutations/upgradeToPro.ts
nickoferrall: Ah I tried to implement this but TypeScript complains that `Object is of type 'unknown'` on `if ('param' in e)`. If the two `any` casts in the catch are too ugly, an alternate option could be disabling an eslint rule that prevent us from doing `catch(e as any){`?

packages/server/graphql/mutations/pokerSetFinalScore.ts
nickoferrall: Interesting, that's good to know! It looks like [standardError](https://github.com/ParabolInc/parabol/blob/3ea2c761bdaa507c0319b6f0da83435b43ecfe51/packages/server/utils/standardError.ts) returns a POJO (when the return statement isn't missing 😅 ) and sends the error to Sentry. Still, I've pushed a change returning `{error: {message}}` as we weren't previously sending the error to Sentry

nickoferrall:
Thanks for the review @mattkrick! I've pushed an update which removes the unknown union type from sendToSentry. This resulted in quite a few changes so I'd prefer to hold off on merging until you've had the chance to scan through the updates

mattkrick:
LGTM!

Issue #5147, opened 2021-07-15
Polls implementation: EPIC


enriquesanchez:
This is an implementation issue for #4990.

![image](https://user-images.githubusercontent.com/3276087/125862136-2e02a9ff-a5b0-4629-a2d9-251ce2b99d9d.png)

The new polls feature allows users to create a multiple choice question in a discussion thread, and lets meeting attendees vote. 

- [Please watch this Loom for a quick overview](https://www.loom.com/share/3ed5d71323494b24bb850ca55d4ebdcc?sharedAppSource=team_library).
- [The Figma source file can be found here](https://www.figma.com/file/D25gVH1t2alUiOdYNs5Owo/Multiple-content-input-types?node-id=518%3A19375).

Feel free to ping me for questions or comments.

# Sub-issues
- [x] #5203
- [x] #5145 
- [x] #5315 
- [x] #5146
- [x] #5441 
- [x] #5148
- [x] #5442
- [x] #5149
- [x] #5443 
- [x] #5150
- [x] #5444 
- [x] #5151
- [x] #5152
- [x] #5153
- [x] #5478

# Additional, related tasks
- [x] #5440  

# Disucssions
 - #5411

BartoszJarocki:
Open questions
1. How to delete a poll that is currently being created?

enriquesanchez:
>How to delete a poll that is currently being created?

@BartoszJarocki I think we should mimic the same behavior we currently have for tasks.

While it's being created:
- If I just pressed 'Add a task' and I haven't typed anythig, pressing <kbd>ESC</kbd> removes the task object.
- If I have already entered some text, then I have to go to the `...` menu and remove it from there ('Set as archived'). In the case of polls, we could say instead 'Delete poll'.

BartoszJarocki:
@enriquesanchez thank you! That makes sense. The reason I'm asking is, there's no '...' on the designs when you're creating a poll so I assume, it should be there all the time, am I right? Also, the difference between poll and task is poll is much harder to remove when losing focus (it has many inputs, for example, we can't remove poll if someone uses a tab to move focus to the 'Start poll' button so I think there are too many edge cases to handle it that way). To summarize, can we do the following?
1. Show `...` button all the time so user can remove the poll anytime
2. If there's nothing added in the poll, hitting ESC will remove the poll
3. We don't remove the poll when a poll is losing focus (it has too many focusable elements to handle that without any hiccups)

[...]

PR #5361, opened 2021-09-03
[Frontend] Creating a poll object

BartoszJarocki:


BartoszJarocki:
@Dschoordsch do you mind taking a look? 🙏

packages/client/components/DiscussionThreadList.tsx
```diff
@@ -123,6 +123,9 @@ export default createFragmentContainer(DiscussionThreadList, {
   threadables: graphql`
     fragment DiscussionThreadList_threadables on Threadable @relay(plural: true) {
       ...ThreadedItem_threadable
+      ... on Poll {
+        updatedAt
+      }
```
BartoszJarocki, line 128: This is needed here so it triggers `useScrollThreadList` when new poll option is added

packages/client/components/DiscussionThread.tsx
```diff
@@ -99,6 +101,7 @@ const DiscussionThread = (props: Props) => {
   const getMaxSortOrder = () => {
     return Math.max(0, ...threadables.map((threadable) => threadable.threadSortOrder || 0))
   }
+  const isCreatingPoll = threadables.some(isLocalPoll)
```
BartoszJarocki, line 104: We can't use `isViewerTyping` as we do with tasks as polls have too many inputs and other focusable elements so it causes a lot of edge cases.

packages/client/components/Poll/Poll.tsx
```diff
@@ -0,0 +1,191 @@
...
+}))
+
+interface Props {
+  children: React.ReactNode
+  poll: Poll_poll$key
+  discussion: Poll_discussion$key
+  dataCy: string
+}
+
+const Poll = React.forwardRef((props: Props, ref: Ref<HTMLDivElement>) => {
```
BartoszJarocki, line 55: Polls are created using compound components pattern, so all the implementation of each part is implemented as separate components and use poll context under the hood. If you see any potential problems with that solution, please let me know!

packages/client/mutations/CreatePollMutation.ts
```diff
@@ -20,21 +31,60 @@ const mutation = graphql`
...
   atmosphere,
   variables,
-  {onError, onCompleted}
+  {localPoll, onError, onCompleted}
 ) => {
   return commitMutation<TCreatePollMutation>(atmosphere, {
     mutation,
     variables,
-    optimisticUpdater: (store) => {
-      const {} = variables
```
BartoszJarocki, line 37: As poll is already in the relay store might not need an optimistic updater. The only problem with that solution is, after clicking 'Start poll' the poll is still in the 'creating' state. We might want to extend the gql `Poll` type on the client and add a property `isLocalPoll` or something like that and update it to `false` in optimistic updater. This would make the `Poll` look like an already created poll just after clicking the `Start poll` button. Let me know what do you think! 💡

packages/client/mutations/handlers/safePutNodeInConn.ts
```diff
@@ -1,11 +1,11 @@
-import {ConnectionHandler, RecordProxy, RecordSourceSelectorProxy} from 'relay-runtime'
+import {ConnectionHandler, RecordProxy, RecordSourceProxy} from 'relay-runtime'
 import getNodeById from '../../utils/relay/getNodeById'
 import {insertEdgeAfter} from '../../utils/relay/insertEdge'
 
 const safePutNodeInConn = (
   conn: RecordProxy | null | undefined,
   node: RecordProxy,
-  store: RecordSourceSelectorProxy,
+  store: RecordSourceProxy,
```
BartoszJarocki, line 8: This was needed so I can reuse this in local store updates, it doesn't break anything.

packages/client/styles/helpers/cardRootStyles.ts
```diff
@@ -8,7 +8,7 @@ const cardRootStyles = {
   boxShadow: cardShadow,
   minWidth: 256,
   maxWidth: 296,
-  position: 'relative' as 'relative',
+  position: 'relative' as const,
```
BartoszJarocki, line 11: Just typescript thingy, it was complaining, probably an old file 👴

packages/client/types/constEnums.ts
```diff
@@ -423,5 +423,15 @@ export const enum Polls {
...
+  MAX_OPTION_TITLE_LENGTH = 30
+}
+
+export const enum PollsAriaLabels {
+  POLL_TITLE_EDITOR = 'Poll Editor',
+  POLL_OPTION_EDITOR = 'Poll Option Editor',
+  POLL_SUBMIT_VOTE = 'Submit poll vote',
+  POLL_START = 'Start a poll',
+  POLL_ADD_OPTION = 'Add a poll option'
 }
```
BartoszJarocki, line 437: I grouped all the aria labels related to polls in a separate enum, it's easier to find

packages/server/graphql/mutations/createPoll.ts
```diff
@@ -116,7 +116,7 @@ const createPoll = {
         teamId
       }
     })
-    publish(SubscriptionChannel.MEETING, meetingId, 'AddPollSuccess', data, subOptions)
+    publish(SubscriptionChannel.MEETING, meetingId, 'CreatePollSuccess', data, subOptions)
```
BartoszJarocki, line 119: Meeting subscription was crashing because of that, I probably missed that when implementing backend 🕵️‍♂️

packages/server/graphql/types/Poll.ts
```diff
@@ -12,6 +13,13 @@ const Poll = new GraphQLObjectType<any, GQLContext>({
   isTypeOf: ({title}) => !!title,
   fields: () => ({
     ...(threadableFields() as any),
+    createdByUser: {
+      type: GraphQLNonNull(User),
+      description: 'The user that created the item',
+      resolve: ({createdById}, _args, {dataLoader}: GQLContext) => {
+        return dataLoader.get('users').load(createdById)
+      }
+    },
```
BartoszJarocki, line 22: I added it here as a non-nullable field, Polls can't be created without a user. Threadable fields are providing nullable type

packages/server/graphql/types/Threadable.ts
```diff
@@ -74,7 +74,7 @@ const Threadable = new GraphQLInterfaceType({
       return Task
     }
 
-    if ('options' in threadable) {
+    if ('title' in threadable) {
```
BartoszJarocki, line 77: If we change the DB queries to include the poll options in a single DB hit, then it'd work. Otherwise, there are no `options` when fetching the `poll`

packages/client/clientSchema.graphql
```diff
@@ -104,3 +104,8 @@ extend type Discussion {
   isAnonymousComment: Boolean!
   replyingToCommentId: String!
 }
+
+extend type PollOption {
+  placeholder: String
+  shouldAutoFocus: Boolean
```
BartoszJarocki, line 110: This looks pretty bad as it's mixing the implementation detail with a data layer. The reason I've added is, when the user clicks `Add another option`, the new option has to be automatically focused. On the other hand, when the user initially creates a poll, the poll title has to be focused so we need a way to figure out if the option is the one initially added, or added as a result of a user action. If you have some ideas on how to improve that, I'd love to do that!

packages/client/components/Poll/local/newPoll.ts
```diff
@@ -0,0 +1,101 @@
...
+import Atmosphere from '../../../Atmosphere'
+import getDiscussionThreadConn from '../../../mutations/connections/getDiscussionThreadConn'
+import safePutNodeInConn from '../../../mutations/handlers/safePutNodeInConn'
+
+const LocalPollOptionsNumber = 2
+
+export const isLocalPoll = (threadable: {id: string}) =>
+  threadable.id.startsWith('poll') && threadable.id.endsWith('tmp')
+
+export const createLocalPoll = (
```
Dschoordsch, line 12: In my mind tasks and polls are quite similar so it seems odd to me why polls are created locally and only shared when final, but tasks are created immediately. Maybe you can explain why?

Dschoordsch (changes requested):
When I tested there where a couple of ACs from the ticket missing:
* [ ] Not completing all required fields, and/or clicking elsewhere before the poll is submitted, cancels the poll creation flow (similar to tasks), and returns keyboard focus to the 'Add a poll' button in the action bar.
* [ ] A poll can only have up to 4 choices. Choices 3 and 4 being optional.
* [ ] After 'Choice 2', the next focus stop is the 'Add another choice button'. This is an optional action and pressing it adds a 'Choice 3' field almost identical to the previous 2. Only difference is the placeholder text reads "Adds choice 3 (optional)...".
* [ ] Once Choice 1 and 2 have been filled in, the 'Start' button becomes active.

Additional I found this irritating
* [ ] Fields with less than 3 characters are not allowed, but there is no UI indication for it. Is this even required for the answers? Example questions: What's your favourite number? What's your favourite letter? How many testers should we hire?

I'm not sure if that should be handled here already, but:
* [ ] There is no "User X added a poll" notification

packages/client/components/DiscussionThreadList.tsx
BartoszJarocki: It has to be here due to the fact that Relay masks the data. The hook `useScrollThreadList` will not see any changes in `threadables` unless this component specifies this data dependency.

packages/client/components/Poll/Poll.tsx
BartoszJarocki: I'm open to changing it. It makes sense to me to design the components that way here. I described that here: https://github.com/ParabolInc/parabol/issues/5145. On the other hand, the 2 downsides you mentioned are valid so if we decide it's too confusing it's not a problem to change it.

BartoszJarocki:
@Dschoordsch thank you for the review! Here are some clarifications related to AC

>  Not completing all required fields, and/or clicking elsewhere before the poll is submitted, cancels the poll creation flow (similar to tasks), and returns keyboard focus to the 'Add a poll' button in the action bar.

https://github.com/ParabolInc/parabol/issues/5147#issuecomment-929218886 
TLDR: too many focusable elements and edge cases to do it that way. We'll have the `...` button with the `Delete` option all the time.

> A poll can only have up to 4 choices. Choices 3 and 4 being optional. 

This works that way unless I missed something? There's one thing left to align after the discussion here: https://github.com/ParabolInc/parabol/issues/5147#issuecomment-931463794 ie. if option 3 or 4 gets added, don't require them to be filled

>  After 'Choice 2', the next focus stop is the 'Add another choice button'. This is an optional action and pressing it adds a 'Choice 3' field almost identical to the previous 2. The only difference is the placeholder text reads "Adds choice 3 (optional)...". 

Correct, adding `(optional)` is missing and was added as a result of the already linked discussion

> Once Choice 1 and 2 have been filled in, the 'Start' button becomes active.

Is there anything missing here except for adding empty option 3 or 4? 

>  Fields with less than 3 characters are not allowed, but there is no UI indication for it. Is this even required for the answers? Example questions: What's your favourite number? What's your favourite letter? How many testers should we hire?

This is good question! @enriquesanchez what do you think about that? I personally think the min length of the answer should be `1` not `2`

> There is no "User X added a poll" notification 
Hmm, where is this defined, and how does it look like?

packages/server/graphql/types/Poll.ts
BartoszJarocki: Good point. I can remove it. The question is, this PR was closed and not merged. @tiffanyhan are we still doing it or was it dropped for some reason? With polls, this is also a matter of changing the DB schema as currently, we have 
```
      "createdById" VARCHAR(100) NOT NULL REFERENCES "User"("id"),
```
so it'd have to be changed to a nullable field.

packages/client/components/Poll/PollContext.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import {Poll_poll} from '~/__generated__/Poll_poll.graphql'
+
+export type PollState = 'creating' | 'created'
+
+interface PollContextType {
+  pollState: PollState
+  canCreatePoll: boolean
+  poll: Poll_poll
+  updatePoll: (optionId: string, title: string) => void
```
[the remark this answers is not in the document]
BartoszJarocki: Good catch!

BartoszJarocki:
@Dschoordsch I've answered all the questions, fixed a few things related to missing AC. I've also changed the min option title len to 1 so it's not annoying anymore, I hope @enriquesanchez agrees on that 😅 
I'd appreciate it if you could do another round, thank you!

PR #5365, opened 2021-09-06
Make sure that users can always see agenda items in check-in

### LINKED ISSUE #5343: Ensure users can always see the agenda list items
When there are many team members in the solo updates section, the user can't see the agenda items. 

Increase the minWidth of the agenda list in the Check-in meeting sidebar

nickoferrall:
Fixes #5343 

Previously, if the height of the meeting members exceeded the available space in the sidebar, the user couldn't see the agenda items:

old-view

This PR ensures that 5 agenda items are always visible (if there are at least 5 agenda items):
new-view

### To Test:

- create a check-in meeting
- add a few other users to the meeting
- reduce the height of the page using the dev tools so that the Icebreaker / Solo Updates section takes up most of the sidebar height
- note that regardless of the number of meeting members in the Icebreaker / Solo Updates phase, you can still see 5 agenda items (if there are at least 5)

Note: there are some typecheck errors in files that this PR doesn't touch. These should be fixed by #5356

### DIFF
--- a/packages/client/components/ActionSidebarAgendaItemsSection.tsx
+++ b/packages/client/components/ActionSidebarAgendaItemsSection.tsx
@@ -1,9 +1,16 @@
+import styled from '@emotion/styled-base'
 import graphql from 'babel-plugin-relay/macro'
 import React from 'react'
 import {createFragmentContainer} from 'react-relay'
 import useGotoStageId from '../hooks/useGotoStageId'
 import AgendaListAndInput from '../modules/teamDashboard/components/AgendaListAndInput/AgendaListAndInput'
 import {ActionSidebarAgendaItemsSection_meeting} from '../__generated__/ActionSidebarAgendaItemsSection_meeting.graphql'
+import MeetingSidebarPhaseItemChild from './MeetingSidebarPhaseItemChild'
+
+const StyledRoot = styled(MeetingSidebarPhaseItemChild)({
+  overflow: 'visible',
+  minHeight: 'fit-content'
+})
 
 interface Props {
   gotoStageId: ReturnType<typeof useGotoStageId>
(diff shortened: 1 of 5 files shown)


nickoferrall:
Hey @enriquesanchez, would you mind reviewing this PR when you get the chance?

enriquesanchez (approved):
Thanks for the patience while I got my environment sorted 🙏
The update LGTM 👍 

Question: why show only 5? Why not all if they can now scroll the list?

nickoferrall:
Thanks for the review @enriquesanchez!

I was thinking that just showing 5 items keeps the focus on the the Icebreaker / Updates and prevents the sidebar from getting too long. But maybe a user could see 5 items and think the rest of their items are missing and not realise that they can scroll down.. I've pushed an update showing 100% of the items

nickoferrall:
@mattkrick  this PR is ready for review when you have the chance. I noticed that in prod, when moving phases, the first child doesn't always scroll perfectly into view if there are many children. It does some of the time but not always.

I've spent some time digging into it and then created a separate issue: https://github.com/ParabolInc/parabol/issues/5377

packages/client/modules/teamDashboard/components/AgendaList/AgendaList.tsx
```diff
@@ -16,16 +16,14 @@ import AgendaItem from '../AgendaItem/AgendaItem'
...
 import {AgendaList_agendaItems} from '~/__generated__/AgendaList_agendaItems.graphql'
 
-const AgendaListRoot = styled('div')({
+const AgendaListRoot = styled('div')<{isMeeting: boolean}>(({isMeeting}) => ({
   display: 'flex',
   flexDirection: 'column',
-  // react-beautiful-dnd supports scrolling on 1 parent
-  // this is where we need it, in order to scroll a long list
-  overflow: 'auto',
+  overflow: isMeeting ? undefined : 'auto',
```
nickoferrall, line 22: I noticed that if there are many agenda items, there could be two vertical scrollbars. Now that the sidebar is scrollable in the meeting, we can remove this overflow

packages/client/components/ActionSidebarAgendaItemsSection.tsx
```diff
@@ -1,4 +1,4 @@
-import styled from '@emotion/styled'
+import styled from '@emotion/styled-base'
```
mattkrick, line 1: why the -base?

mattkrick:
works well! just curious about pulling from styled-base, maybe an auto-import error?
good to merge either way!

packages/client/components/ActionSidebarAgendaItemsSection.tsx
nickoferrall: Huh, that must be the auto-import being strange. I've updated it!

PR #5375, opened 2021-09-07
Use 28 + 16 blocks for team header

ackernaut:
I think the best option for the header alignment is 28 over 16. Sorry for the back and forth. I also added a hover state for those icon controls.


nickoferrall (approved):
I like it, 28 + 16 pixels looks like the winning combination to me!

Issue #5183, opened 2021-07-19
Spotlight: add null case business logic to similarReflections


mattkrick:
Predecessors: #5182

Now that `similarReflections` just returns all reflection groups in the meeting, we need to filter that down if the searchQuery is an empty string.

AC
- [ ] if the searchQuery is an empty string, then we use hierarchical agglomeration to filter & sort the reflectionGroups in the meeting (see https://github.com/ParabolInc/parabol/blob/db45e3b5a4c27e917dcc267d33175bae8a04970b/packages/server/graphql/mutations/helpers/handleCompletedStage.ts#L40 as an example)
- [ ] The heuristic above should attempt to return 10 reflections & if the agglomeration returns less (and there are more than 10 other reflection groups in the meeting), then increase the grouping threshold iteratively until it works
- [ ] on the front end, the groups show up as expected in the results space
- [ ] on another browser, if 2 of those groups are merged in real time, that change is reflected in the search results
- [ ] on another browser, if 1 of those reflections is ungrouped, then that change is reflected in the search results

Estimate: 8 hours

nickoferrall:
I'd be keen to work on this one if there's no objections! Will finish the tasks I'm currently working on and aim to start tomorrow

nickoferrall:
Hey @mattkrick, just want to check whether you're thinking that we'll use the `groupReflections` function here or just borrow some of the hierarchical agglomeration logic?

If we are using the `groupReflections` function, it looks like it sorts reflection groups into an order where groups with similar titles are closer to each other (useful for when going from reflect to group stage). 

However, in the Spotlight, I imagine that we want to sort groups by the ones that are similar to the selected reflection title instead?

mattkrick:
yeah, 2 approaches:
- use the grouping heuristic, find the group that the target reflection is in & keep increasing the coefficient until a satisfactory number of cards are returned
- come up with a new heuristic like a k-nearest-neighbors

the first one should be faster & since this is just a first pass that should be AOK.

nickoferrall:
I think that the `reflectionGroup` function isn't working as expected: 
- I get exactly the same output if I change the `groupingThreshold` from 0.25 to 0.99 or 0.001 or 1.5. 
- the smartTitle and title in the groups are always the content of the first reflection
[...]

Issue #4952, opened 2021-05-11
Retro: Grouping: Spotlight


mattkrick:
From SPA, we designed a pretty amazing spotlight feature:
https://www.notion.so/parabol/Big-Teams-Aviva-Matt-Aaron-7a051021f5bd42779aef38f9440d5609

We decided that this could replace breakout rooms & a few other smaller features. 
Next step is to design it. Here's the pen & ink:
<img width="535" alt="Screen_Shot_2021-03-12_at_12 20 05_PM" src="https://user-images.githubusercontent.com/5514175/117871955-df660680-b252-11eb-8ad9-ee9bca6074e3.png">

AC
- [ ] During the group phase, on each reflection card there is a "search" icon in the card footer.
- [ ] On desktop, the search icon is hidden until hovered
- [ ] Hovering the icon shows a tooltip that says "Find similar" (or whatever copy you like)
- [ ] Clicking the icon opens up the "Spotlight Modal", which is a large modal that takes up the entire meeting area (excluding left nav & right drawer)
- [ ] At the top of the spotlight modal is the card that the user clicked (it will animated into place, just like expanded groups do today)
- [ ] below the card is an autoFocused search input field with the search icon on the left
- [ ] below the input is the "SpotlightSuggestionGrid" that shows the reflection cards that are most similar to the top card. if input is empty, we still show some suggestions. 
- [ ] If no suggestions are available, then the grid says "No reflections match that search query" (copy tbd)
- [ ] If the reflection card has already been grouped, then it's group title is in the card header
- [ ] On each reflection card, the text that matches the search query is highlighted (e.g. if the search query is "cat" then the word "cat" is highlighted if it's visible on the card). Same goes for the reflection group title.
- [ ] If the reflection card is very long, then card is auto-scrolled to the first instance of the highlighted text
- [ ] The top card can be dragged & dropped onto one of the cards in the grid. Once it's dropped, the modal closes. It the drop is cancelled (it's not dropped on a specific card) then the top card animates back up to the top  
- [ ] if the top card is already is a group of 2+ cards, then it is moved out of its current group into the new one
- [ ] the bottom cards can be dragged & dropped onto the top card. if they are in a group of 2+ cards, then they are moved from their current group into the group of the top card. dropping on the top card does _not_ automatically close the modal
- [ ] Bottom cards cannot be dropped on each other (we have to draw the line somewhere. let's force interaction with that top card)

[...]

PR #5376, opened 2021-09-07
Spotlight null query case

nickoferrall:
WIP for issue https://github.com/ParabolInc/parabol/issues/5183

### DIFF
--- a/packages/client/components/ReflectionCard/ReflectionCard.tsx
+++ b/packages/client/components/ReflectionCard/ReflectionCard.tsx
@@ -236,6 +236,7 @@ const ReflectionCard = (props: Props) => {
     !inSpotlight &&
     !isComplete &&
     showSpotlight &&
+    !!openSpotlight &&
     (isHovering || !isDesktop)
   return (
     <ReflectionCardRoot
--- a/packages/client/components/ReflectionCard/ReflectionCard.tsx
+++ b/packages/client/components/ReflectionCard/ReflectionCard.tsx
@@ -230,7 +230,7 @@ const ReflectionCard = (props: Props) => {
     }
   }
 
-  const showSpotlight = false
+  const showSpotlight = !__PRODUCTION__
   const showSearch =
     phaseType === 'group' &&
     !inSpotlight &&
--- a/packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx
+++ b/packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx
@@ -77,9 +77,10 @@ const DraggableReflectionCard = (props: Props) => {
     swipeColumn,
     dataCy
   } = props
-  const {id: meetingId, teamId, localStage} = meeting
+  const {id: meetingId, teamId, localStage, spotlightReflection} = meeting
   const {isComplete, phaseType} = localStage
   const {isDropping, isEditing} = reflection
+  const isSpotlightOpen = !!(spotlightReflection?.id && openSpotlight)
   const [drag] = useState(makeDragState)
   const staticReflectionCount = staticReflections?.length || 0
   const {onMouseDown} = useDraggableReflectionCard(
@@ -89,6 +90,7 @@ const DraggableReflectionCard = (props: Props) => {
     meetingId,
     teamId,
     staticReflectionCount,
+    isSpotlightOpen,
     swipeColumn
   )
   const isDragPhase = phaseType === 'group' && !isComplete
@@ -151,6 +153,9 @@ export default createFragmentContainer(DraggableReflectionCard, {
           isComplete
           phaseType
         }
+      }
+      spotlightReflection {
+        id
       }
     }
   `
--- a/packages/client/components/SpotlightModal.tsx
+++ b/packages/client/components/SpotlightModal.tsx
@@ -146,25 +146,25 @@ const SpotlightModal = (props: Props) => {
             title
           }
           meeting(meetingId: $meetingId) {
-            id
-            teamId
-            localPhase {
-              phaseType
-            }
-            localStage {
-              isComplete
-              phaseType
-            }
-            phases {
-              phaseType
-              stages {
-                isComplete
-                phaseType
-              }
-            }
             ... on RetrospectiveMeeting {
               ...DraggableReflectionCard_meeting
               ...SpotlightGroups_meeting
+              id
+              teamId
+              localPhase {
+                phaseType
+              }
+              localStage {
+                isComplete
+                phaseType
+              }
+              phases {
+                phaseType
+                stages {
+                  isComplete
+                  phaseType
+                }
+              }
               spotlightReflection {
                 ...DraggableReflectionCard_reflection
               }
@@ -176,20 +176,18 @@ const SpotlightModal = (props: Props) => {
     queryRef,
     {UNSTABLE_renderPolicy: 'full'}
   )
-  console.log('🚀 ~ SpotlightModal ~ data', data)
+
   const {viewer} = data
   const {meeting, similarReflectionGroups} = viewer
-  const {spotlightReflection} = meeting
-  console.log('🚀  ~ spotlightReflection', spotlightReflection)
-  // const similarReflectionGroups = viewer?.similarReflectionGroups
+  const spotlightReflection = meeting?.spotlightReflection
   const isLoading = similarReflectionGroups === undefined
-  console.log('🚀 ~ SpotlightModal ~ isLoading', isLoading)
   const isDesktop = useBreakpoint(Breakpoint.NEW_MEETING_SELECTOR)
   const onKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
     if (e.key === 'Escape' && !e.currentTarget.value) {
       closeSpotlight()
     }
   }
+  if (!meeting) return null
   return (
     <>
       <ModalContainer isDesktop={isDesktop}>
@@ -218,7 +216,7 @@ const SpotlightModal = (props: Props) => {
           {isLoading ? (
             <LoadingComponent height={24} width={24} showAfter={0} spinnerSize={24} />
           ) : (
-            <SpotlightGroups meeting={meeting} viewer={viewer!} />
+            <SpotlightGroups meeting={meeting} viewer={viewer} />
           )}
         </SimilarReflectionGroups>
       </ModalContainer>
@@ -237,39 +235,3 @@ const SpotlightModal = (props: Props) => {
 }
 
 export default SpotlightModal
-// export default createFragmentContainer(SpotlightModal, {
-//   meeting: graphql`
-//     fragment SpotlightModal_meeting on RetrospectiveMeeting {
-//       ...DraggableReflectionCard_meeting
-//       ...SpotlightGroups_meeting
-//       id
-//       teamId
-//       localPhase {
-//         phaseType
-//       }
-//       localStage {
-//         isComplete
-//         phaseType
-//       }
-//       phases {
-//         phaseType
-//         stages {
-//           isComplete
-//           phaseType
-//         }
-//       }
-//       spotlightReflection {
-//         ...DraggableReflectionCard_reflection
-//       }
-//     }
-//   `,
-//   viewer: graphql`
-//     fragment SpotlightModal_viewer on User {
-//       ...SpotlightGroups_viewer
-//       similarReflectionGroups(reflectionId: $reflectionId, searchQuery: $searchQuery) {
-//         id
-//         title
-//       }
-//     }
-//   `
-// })
--- a/packages/client/components/SpotlightModal.tsx
+++ b/packages/client/components/SpotlightModal.tsx
@@ -1,21 +1,21 @@
 import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
 import React from 'react'
-import {createFragmentContainer} from 'react-relay'
-import {SpotlightModal_meeting} from '~/__generated__/SpotlightModal_meeting.graphql'
-import {SpotlightModal_viewer} from '~/__generated__/SpotlightModal_viewer.graphql'
+import {PreloadedQuery, usePreloadedQuery} from 'react-relay'
 import useBreakpoint from '../hooks/useBreakpoint'
 import {DECELERATE, fadeUp} from '../styles/animation'
 import {Elevation} from '../styles/elevation'
 import {PALETTE} from '../styles/paletteV3'
 import {ICON_SIZE} from '../styles/typographyV2'
 import {Breakpoint, ElementHeight, ElementWidth, ZIndex} from '../types/constEnums'
+import {SpotlightModalQuery} from '../__generated__/SpotlightModalQuery.graphql'
 import Icon from './Icon'
+import LoadingComponent from './LoadingComponent/LoadingComponent'
 import MenuItemComponentAvatar from './MenuItemComponentAvatar'
 import MenuItemLabel from './MenuItemLabel'
 import PlainButton from './PlainButton/PlainButton'
 import DraggableReflectionCard from './ReflectionGroup/DraggableReflectionCard'
-import SpotlightEmptyState from './SpotlightEmptyState'
+import SpotlightGroups from './SpotlightGroups'
 
 const SELECTED_HEIGHT_PERC = 33.3
 const ModalContainer = styled('div')<{isDesktop: boolean}>(({isDesktop}) => ({
@@ -48,6 +48,7 @@ const SelectedReflectionSection = styled('div')({
 
 const SimilarReflectionGroups = styled('div')({
   display: 'flex',
+  alignItems: 'center',
   justifyContent: 'center',
   height: `${SELECTED_HEIGHT_PERC * 2}%`,
   padding: 16
@@ -129,21 +130,66 @@ const CloseIcon = styled(Icon)({
 
 interface Props {
   closeSpotlight: () => void
-  meeting: SpotlightModal_meeting
   flipRef: (instance: HTMLDivElement) => void
-  viewer: SpotlightModal_viewer
+  queryRef: PreloadedQuery<SpotlightModalQuery>
 }
 
 const SpotlightModal = (props: Props) => {
-  const {closeSpotlight, meeting, flipRef} = props
+  const {closeSpotlight, flipRef, queryRef} = props
+  const data = usePreloadedQuery<SpotlightModalQuery>(
+    graphql`
+      query SpotlightModalQuery($reflectionId: ID!, $searchQuery: String!, $meetingId: ID!) {
+        viewer {
+          ...SpotlightGroups_viewer
+          similarReflectionGroups(reflectionId: $reflectionId, searchQuery: $searchQuery) {
+            id
+            title
+          }
+          meeting(meetingId: $meetingId) {
+            id
+            teamId
+            localPhase {
+              phaseType
+            }
+            localStage {
+              isComplete
+              phaseType
+            }
+            phases {
+              phaseType
+              stages {
+                isComplete
+                phaseType
+              }
+            }
+            ... on RetrospectiveMeeting {
+              ...DraggableReflectionCard_meeting
+              ...SpotlightGroups_meeting
+              spotlightReflection {
+                ...DraggableReflectionCard_reflection
+              }
+            }
+          }
+        }
+      }
+    `,
+    queryRef,
+    {UNSTABLE_renderPolicy: 'full'}
+  )
+  console.log('🚀 ~ SpotlightModal ~ data', data)
+  const {viewer} = data
+  const {meeting, similarReflectionGroups} = viewer
   const {spotlightReflection} = meeting
+  console.log('🚀  ~ spotlightReflection', spotlightReflection)
+  // const similarReflectionGroups = viewer?.similarReflectionGroups
+  const isLoading = similarReflectionGroups === undefined
+  console.log('🚀 ~ SpotlightModal ~ isLoading', isLoading)
   const isDesktop = useBreakpoint(Breakpoint.NEW_MEETING_SELECTOR)
   const onKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
     if (e.key === 'Escape' && !e.currentTarget.value) {
       closeSpotlight()
     }
   }
-  const reflectionGroupsCount = 0
   return (
     <>
       <ModalContainer isDesktop={isDesktop}>
@@ -169,7 +215,11 @@ const SpotlightModal = (props: Props) => {
           </SearchItem>
         </SelectedReflectionSection>
         <SimilarReflectionGroups>
-          {reflectionGroupsCount === 0 ? <SpotlightEmptyState /> : null}
+          {isLoading ? (
+            <LoadingComponent height={24} width={24} showAfter={0} spinnerSize={24} />
+          ) : (
+            <SpotlightGroups meeting={meeting} viewer={viewer!} />
+          )}
         </SimilarReflectionGroups>
       </ModalContainer>
       <ReflectionWrapper ref={flipRef}>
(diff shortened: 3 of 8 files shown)


packages/client/types/constEnums.ts
```diff
@@ -391,9 +391,9 @@ export const enum ZIndex {
   SIDEBAR = 16,
   SIDE_SHEET = 8,
   SNACKBAR = 24 /* snackbar is kind of dialog */,
-  SPOTLIGHT_MODAL = 25, // above the tooltip
   REFLECTION_IN_FLIGHT = 8,
-  REFLECTION_IN_FLIGHT_LOCAL = 26, // keep it above the dialog
+  REFLECTION_IN_FLIGHT_LOCAL = 28, // keep it above the dialog
+  REFLECTION_IN_FLIGHT_SPOTLIGHT = 26,
```
nickoferrall, line 396: Remote reflections that are in the Spotlight (but not those that aren't in the Spotlight) need to appear above the Spotlight modal but below local reflections. The Spotlight modal uses usePortal which uses the dialog zIndex of 24. usePortal is also used by expandedReflectionsStack and a bunch of other modals which, unlike the Spotlight, need to appear above the remote reflections.

I think the cleanest approach is adding a `REFLECTION_IN_FLIGHT_SPOTLIGHT` zIndex and then passing an `isInSpotlight` prop into `useDraggableReflectionCard`

packages/client/mutations/EndDraggingReflectionMutation.ts
```diff
@@ -87,6 +87,42 @@ const handleAddReflectionGroupToGroups = (store, reflectionGroup) => {
...
+    searchQuery: ''
+  })
+  const wasInSpotlightGroups = similarReflectionGroups?.find(
+    (group) => group.getValue('id') === oldReflectionGroupId
+  )
+  const isInSpotlightGroups = similarReflectionGroups?.find(
+    (group) => group.getValue('id') === reflectionGroupId
+  )
+  // check if ungrouping is happening in the spotlight
+  if (!isInSpotlightGroups && wasInSpotlightGroups) {
```
nickoferrall, line 116: If the following scenario happens:
- the Spotlight opens and has 10 reflection groups
- a user on another browser adds a reflection called "dogs" into one of the 10 Spotlight groups
- that user then removes the reflection "dogs" from that group

Would you expect the "dogs" reflection to remain in the Spotlight so that there are 11 groups or that it returns to the kanban? I think it should probably stay (which is how the current implementation works) because while it could have been a accidentally grouped, it's probably more likely that it's a relevant reflection group

packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx
```diff
@@ -89,15 +92,17 @@ const DraggableReflectionCard = (props: Props) => {
...
   )
+
   const isDragPhase = phaseType === 'group' && !isComplete
   const canDrag = isDraggable && isDragPhase && !isEditing && !isDropping
   // slow state updates can mean we miss an onMouseDown event, so use isDragPhase instead of canDrag
   const handleDrag = isDragPhase ? onMouseDown : undefined
   return (
     <DragWrapper
-      ref={(c) => (drag.ref = c)}
+      ref={(c) => (drag.ref = isBehindSpotlight ? null : c)}
```
nickoferrall, line 105: This fixed a tricky bug. When dragging a reflection group in the Spotlight, `useDraggableReflectionCards` would use the ref that's in the kanban instead of the Spotlight. This made dragging, grouping, and ungrouping very buggy. 

Remote reflections still begin from the kanban position which I think is preferable as it indicates that that's where the teammate is dragging the card from

packages/server/graphql/types/User.ts
```diff
@@ -421,19 +423,38 @@ const User = new GraphQLObjectType<any, GQLContext>({
...
+            .run()
         ])
         const {meetingId: viewerMeetingId} = viewerMeetingMember
         if (viewerMeetingId !== meetingId) {
-          standardError(new Error('Not on team'), {userId})
-          return []
+          return standardError(new Error('Not on team'), {userId})
         }
-        return reflectionGroups
+        const {groups} = groupReflections(reflections, AUTO_GROUPING_THRESHOLD)
```
nickoferrall, line 445: Hey @tianrunhe, it looks like you've played with the `groupReflection` function before. I'd love to get your help here.

The AC says:
> The heuristic above should attempt to return 10 reflections & if the agglomeration returns less (and there are more than 10 other reflection groups in the meeting), then increase the grouping threshold iteratively until it works

However, if I change the grouping threshold from 0.25 to 0.95 or 0.05, I get no difference in results. The same number of reflections and reflection groups are returned. See [the issue here](https://github.com/ParabolInc/parabol/issues/5183) for more context and screenshots. Perhaps I'm missing something?

mattkrick: what text are you using for the reflections?
the text has to be long enough & meaningful enough to actually work. it's gonna use a string similarity algo to detect how similar things are. if the reflection texts are too small, the similarity scores will be 0. might be worth using data from real retros to play with

packages/client/mutations/EndDraggingReflectionMutation.ts
```diff
@@ -87,6 +87,42 @@ const handleAddReflectionGroupToGroups = (store, reflectionGroup) => {
   addNodeToArray(reflectionGroup, meeting, 'reflectionGroups', 'sortOrder')
 }
 
+const handleAddReflectionToSpotlight = (
```
nickoferrall, line 90: Hey @mattkrick, I've got two architectural questions I'd like your advice on:

1. Should the remote reflection position begin where the remote user is dragging the reflection, i.e. the kanban, or in the Spotlight? 

Currently, it begins in the kanban (see gif). This tells the user it's being dragged by a teammate but maybe the position feels a bit random.

2. Should the results be groupable/ungroupable?
Enrique mentioned [here](https://github.com/ParabolInc/parabol/issues/4952#issuecomment-862513558) that we should prevent users from grouping/ungrouping the results.
This causes a problem in the following scenario:
- I have the Spotlight open
- another user ungroups a group on their browser that's in my Spotlight results
- that creates a new reflection group id which does not exist in the similarReflectionGroups query
- in my Spotlight, it looks like the card disappears

In this PR, if a results group is ungrouped, I'm adding the newly created reflection group id to the similarReflectionGroups query. This way, the card doesn't disappear and instead remains on the Spotlight.

Alternatively, if the remote reflection begins in the Spotlight, perhaps it could be dragged to the kanban and then disappear from the Spotlight. To do this, I'd try using the ref of the Spotlight card [here](https://github.com/ParabolInc/parabol/blob/0ef8108f6fc8cfce1c59d2fcd40d65a362acee4b/packages/client/components/ReflectionGroup/RemoteReflection.tsx#L63) instead of the `remoteDrag` position.

remote-reflections

Note: I'll create a separate PR with a better Spotlight groups UI

packages/server/package.json
```diff
@@ -131,9 +131,9 @@
     "relay-runtime": "^11.0.2",
     "rethinkdb-ts": "2.4.5",
     "rethinkdbdash": "^2.3.31",
-    "samlify": "^2.6.1",
+    "samlify": "2.7.7",
     "secure-compare": "^3.0.1",
-    "sharp": "^0.26.2",
+    "sharp": "^0.28.0",
```
nickoferrall, line 136: Needed for the m1 chip. I'll create a separate PR with these changes once this is merged: https://github.com/ParabolInc/parabol/pull/5396

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -230,12 +224,13 @@ const ReflectionCard = (props: Props) => {
...
 
-  const showSpotlight = false
+  const showSpotlight = !__PRODUCTION__
   const showSearch =
     phaseType === 'group' &&
-    !inSpotlight &&
+    !isSpotlightOpen &&
     !isComplete &&
     showSpotlight &&
+    !isDemoRoute() &&
```
nickoferrall, line 233: The demo only has a small number of reflections so the Spotlight isn't necessary although it might still be nice to show it off. If you think it's worth including, I suggest that I create a new issue for the demo and tackle that as the last issue in the Spotlight

packages/client/mutations/EndDraggingReflectionMutation.ts
nickoferrall: I think the answers to both 1 & 2 are yes. I've pushed an update following the remote reflection plan I outlined here. In the Spotlight UI groups PR, I'll work on adding new reflection groups to the Spotlight column with the least groups

nickoferrall:
Hey @tianrunhe, would you mind reviewing this PR, please? Specifically, it'd be great to get your eyes on the `similarReflectionsQuery` logic in the `User` query and to test that the AC in the issue has been completed

tianrunhe (changes requested):
- [x] if the searchQuery is an empty string, then we use hierarchical agglomeration to filter & sort the reflectionGroups in the meeting (see https://github.com/ParabolInc/parabol/blob/db45e3b5a4c27e917dcc267d33175bae8a04970b/packages/server/graphql/mutations/helpers/handleCompletedStage.ts#L40 as an example)
- [x] The heuristic above should attempt to return 10 reflections & if the agglomeration returns less (and there are more than 10 other reflection groups in the meeting), then increase the grouping threshold iteratively until it works
- [x] on the front end, the groups show up as expected in the results space
- [x] on another browser, if 2 of those groups are merged in real time, that change is reflected in the search results
- [x] on another browser, if 1 of those reflections is ungrouped, then that change is reflected in the search results

packages/server/graphql/types/User.ts
```diff
@@ -421,19 +423,57 @@ const User = new GraphQLObjectType<any, GQLContext>({
...
-          dataLoader.get('retroReflectionGroupsByMeetingId').load(meetingId)
+          r
+            .table('RetroReflection')
+            .getAll(meetingId, {index: 'meetingId'})
+            .filter({isActive: true})
+            .orderBy('createdAt')
+            .run()
         ])
         const {meetingId: viewerMeetingId} = viewerMeetingMember
         if (viewerMeetingId !== meetingId) {
```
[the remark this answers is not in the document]
nickoferrall: The `meetingId` should always be the same but the `viewerMeetingMember` uses the userId that we get from the source. If the viewer passes a `reflectionId` to the query that lives in a meeting that they're not a part of, the viewer's meeting id should be different.

Now that I'm looking at it, I don't like using `toTeamMemberId` for meeting members so I've added a new helper function

packages/server/graphql/types/User.ts
```diff
@@ -421,19 +423,57 @@ const User = new GraphQLObjectType<any, GQLContext>({
...
+        while (currentThreshold) {
+          const {groupedReflections, nextThresh} = groupReflections(reflections, {
+            groupingThreshold: currentThreshold,
+            maxGroupSize: reflectionsCount,
+            maxReductionPercent: 1
+          })
+          const spotlightGroup = groupedReflections.find(
+            (group) => group.reflectionId === reflectionId
+          )
+          if (!spotlightGroup) break
```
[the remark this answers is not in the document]
nickoferrall: It should always be non-null but TypeScript doesn't know that, unfortunately. I could either use a bang after the `spotlightGroup` or break here

packages/server/graphql/types/User.ts
```diff
@@ -421,19 +423,57 @@ const User = new GraphQLObjectType<any, GQLContext>({
...
+            }
+            if (
+              group.reflectionGroupId === spotlightGroup.reflectionGroupId &&
+              group.oldReflectionGroupId !== spotlightGroup.oldReflectionGroupId
+            ) {
+              similarGroupIds.add(group.oldReflectionGroupId)
+            }
+            currentThreshold = nextThresh
+          }
+        }
```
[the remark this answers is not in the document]
nickoferrall: In the while loop, we're looking through all of the `groupedReflections` rather than the `reflectionGroups`. `spotlightGroup` is the grouped reflection in the source (top bit) of the Spotlight. If we break the loop when we reach 10 `groupedReflections`, there may be less than 10 `reflectionGroups` as several reflections could be in one group. I think we need to keep track of the unique old `reflectionGroupIds` in the loop so that we can guarantee 10 groups.

I agree it's confusing - I've pushed an update changing group to `groupedReflection` to try to add more clarity. Let me know what ya think!

nickoferrall:
Hey @tianrunhe, thanks for the review! I've pushed an update and have responded to the comments. I know you're OOO for a couple of days so if you don't get the chance to review, please let me know and I'll request a different reviewer to keep things moving

tianrunhe:
> Hey @tianrunhe, thanks for the review! I've pushed an update and have responded to the comments. I know you're OOO for a couple of days so if you don't get the chance to review, please let me know and I'll request a different reviewer to keep things moving 

Yeah sorry man, I don't wanna hold you back! I think it's time for you to make some new friends 🤣. I'll definitely check it again when I'm back next Wednesday!

packages/client/mutations/EndDraggingReflectionMutation.ts
mattkrick: > Should the remote reflection position begin where the remote user is dragging the reflection, i.e. the kanban, or in the Spotlight?

The trick here is a single source of truth, ie only 1 card per reflection.
The logic I'd use is something like this:

- when the viewer has spotlight open & a remote user starts to drag a card, make sure the remote user is not dragging the target reflection. if they are & the other user wins the arbitration, close the spotlight. see https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/client/mutations/StartDraggingReflectionMutation.ts#L71-L82
- when the remote user is dragging a card, check to see if that card exists inside the spotlight. 
- if it does, animate the card in front of the spotlight modal
- it it does not, check the `remoteDrag.targetId` 
- if the targetId is null or the targetId does not exist in the spotlight, animate the remote reflection behind the spotlight modal
- if the targetId exists in the spotlight, animate the remote reflection relative to the reflection.

In plain english, let's assume you have a spotlight modal open. if a remote user is dragging a card `A` & hovering over `B`, then you should see that I'm hovering over `B` if it's in your spotlight results. that way when the remote user drops it, it doesn't seem like magic just happened.  

> Should the results be groupable/ungroupable?

- yes, you can drag a result card onto the target card OR you can drag the target card onto a result card.
- like you say, if a remote user drags a result card `A` onto a card that doesn't exist in spotlight `B`, then the spotlight should remove group `A` from the results & add group `B` in its place. visually, it'll look like the card just grew a stack behind it.
- if a remote user drags a card that doesn't exist in spotlight `B` onto a card that _does_ exist in spotlight results `A`, then the user should see that card hovering & then eventually dropped onto `A` (the logic is the same as your question above)

packages/server/graphql/types/User.ts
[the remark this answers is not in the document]
tianrunhe: > If the viewer passes a reflectionId to the query that lives in a meeting that they're not a part of, the viewer's meeting id should be different.

How do we test this case in the code?

We deconstruct the `meetingId` from the reflection in line 430 and then use that to construct a `meetingMemberId` in line 431. If the user is **not** in that meeting, then line 434 should return a null object, which will cause line 442 blow up, yes? On the other hand, if the use is indeed in that meeting, the returned `viewerMeetingMember` object will always have the same `meetingId` as the one we deconstructed from line 430, so the check in line 443 is not helpful.
Tell me if I go completely nut and become blinded in this part of code...

packages/server/graphql/types/User.ts
```diff
@@ -442,37 +443,38 @@ const User = new GraphQLObjectType<any, GQLContext>({
         if (viewerMeetingId !== meetingId) {
           return standardError(new Error('Not on team'), {userId})
         }
-        let currentThreshold: number | null = AUTO_GROUPING_THRESHOLD
-        const similarGroupIds = new Set<string>()
         const reflectionsCount = reflections.length
-        const maxGroupSize = Math.min(reflectionsCount, 10)
-        while (currentThreshold) {
+        const maxReductionPercent = 1
+        const maxResultGroupSize = 10
```
tianrunhe, line 448: Sorry I didn't make it clear... I was proposing to put those in constants like `AUTO_GROUPING_THRESHOLD`

packages/client/components/ReflectionGroup/RemoteReflection.tsx
```diff
@@ -103,16 +106,41 @@ const getHeaderTransform = (ref: RefObject<HTMLDivElement>, topPadding = 18) =>
...
-  const {left, top, minTop} = getCoords(remoteDrag as any)
-  return {nextStyle: {transform: `translate(${left}px,${top}px)`}, minTop}
+  const {left, top, minTop} = getCoords(remoteDrag as any, showAboveSpotlight)
+  const zIndex = showAboveSpotlight ? ZIndex.REFLECTION_IN_FLIGHT_SPOTLIGHT : style.zIndex
+  return {nextStyle: {transform: `translate(${left}px,${top}px)`, zIndex}, minTop}
+}
+
+const getTarget = (meetingId: string, targetId: string, atmosphere: Atmosphere) => {
+  let isTargetInSpotlight = false
+  commitLocalUpdate(atmosphere, (store) => {
```
nickoferrall, line 120: We're not updating the store here - just grabbing a value. I prefer using the Relay store instead of adding another Root query as it saves a trip to the server, and getting the required `reflectionId` and `searchQuery` variables for a Root query can be ugly. 

In [ErrorBoundary](https://github.com/ParabolInc/parabol/blob/89f7d0cceae2c30d7e8757767418ddade4ae98c1/packages/client/components/ErrorBoundary.tsx), I see we use `atmosphere.getStore()`, cast the `store` to `any`, and can then access the records, but I think this is slightly easier to read

nickoferrall: Try useQueryLoader: https://relay.dev/docs/guided-tour/reusing-cached-data/fetch-policies/

packages/client/components/SpotlightModal.tsx
```diff
@@ -168,9 +208,11 @@ const SpotlightModal = (props: Props) => {
             />
           </SearchItem>
         </SelectedReflectionSection>
-        <SimilarReflectionGroups>
-          {reflectionGroupsCount === 0 ? <SpotlightEmptyState /> : null}
-        </SimilarReflectionGroups>
+        <SimilarGroups>
+          <Suspense fallback={''}>
```
nickoferrall, line 212: Suspense is required here as the query in `SpotlightGroup` can be loading. If we remove this, remotely group & ungroup results in the Spotlight, notice that the modal flashes

packages/client/hooks/useDraggableReflectionCard.tsx
```diff
@@ -33,19 +33,37 @@ const useRemoteDrag = (
...
   const setPortal = useContext(PortalContext)
-  const {remoteDrag, isDropping} = reflection
+  const {remoteDrag, isDropping, reflectionGroupId} = reflection
   const setRemoteCard = (isClose: boolean, timeRemaining: number, lastTop?: number) => {
     if (!drag.ref || timeRemaining <= 0) return
     const beforeFrame = Date.now()
     const bbox = drag.ref.getBoundingClientRect()
     if (bbox.top !== lastTop) {
+      const targetId = remoteDrag?.targetId
+      const isTargetInSpotlight = !!document.querySelector(
```
nickoferrall, line 43: @mattkrick I looked into using a ref instead of the `querySelector`, but I think we need to have the logic within `setRemoteCard` as it's recursively updating with every animation frame. In `DraggableReflectionCard.tsx`, we could use the `targetId` to grab the target element and store it in `drag`,  but it won't update as quickly as we need

packages/server/graphql/types/User.ts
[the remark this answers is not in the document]
nickoferrall: True @tianrunhe! I've updated the check so that the error is thrown if `viewerMeetingMember` is null. I tested this by changing `reflectionId` to an id that belongs in a meeting that the viewer is not a part of. I'll move this to Maintainer review unless there's you have any other feedback?

nickoferrall:
FYI: Two issues that I've noticed: 1. After closing the Spotlight, the reflections on the kanban can be stacked incorrectly. 2. The remote reflection position is sometimes slightly off - it looks like it occasionally uses the kanban reflection position instead of the Spotlight. This PR has a fair amount of bloat already, so I'd prefer to press on and tackle these in a separate issue

packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx
```diff
@@ -77,11 +103,15 @@ const DraggableReflectionCard = (props: Props) => {
...
+  const {id: meetingId, teamId, localStage, spotlightReflection} = meeting
   const {isComplete, phaseType} = localStage
   const {isDropping, isEditing} = reflection
-  const [drag] = useState(makeDragState)
+  const isSpotlightOpen = !!spotlightReflection?.id
+  const isInSpotlight = isSpotlightOpen && !openSpotlight
+  const isBehindSpotlight = isSpotlightOpen && !isInSpotlight
   const staticReflectionCount = staticReflections?.length || 0
+  const [drag] = useState(makeDragState)
+  updateSpotlightDrag(drag, isBehindSpotlight, isInSpotlight, spotlightReflection?.id)
```
nickoferrall, line 114: I tried passing in a few args to `makeDragState` and updating the drag object with Spotlight logic that way, but it would update too slowly and throw off the UI

packages/client/mutations/handlers/handleUpdateSpotlightResults.ts
```diff
@@ -0,0 +1,57 @@
...
+    })
+  }
+  // ungrouping created a new group id which needs to be added to Spotlight
+  else if (reflectionsCount === 1) {
+    const sortOrders = similarReflectionGroups.map((group) => ({
+      sortOrder: group.getValue('sortOrder') as number
+    }))
+    const nextSortOrder = getNextSortOrder(sortOrders)
+    reflectionGroup.setValue(nextSortOrder, 'sortOrder')
+    addNodeToArray(reflectionGroup, viewer, 'similarReflectionGroups', 'sortOrder', {
```
nickoferrall, line 48: In the groups to source PR, I noticed that sometimes this can cause an infinite loop when ungrouping. If I removed `handleUpdateSpotlightResults` from the optimistic updater, the infinite loop would stop happening. Not sure if there's anything wrong with the way I'm using the `addNodeToArray` function? If not, there's probably other logic in later PRs causing this issue. It's not happening for me in this branch.

packages/client/hooks/useDraggableReflectionCard.tsx
mattkrick: thanks for digging in!
inside hooks it should be AOK, if there's a better way to do it inside the a functional component let's try that. trying to access the DOM while computing what the vdom should look like might get dicey

possible ideas:

- `useState(() => document.querySelector(...))`
- `useRef` + `useEffect(() => { if (blah blah) document.querySelector}, [])`

packages/client/components/ReflectionCard/ReflectionCard.tsx
mattkrick: i think it's worth showing off, but we can save that for a later time

packages/client/shared/gqlIds/MeetingMemberId.ts
```diff
@@ -1,7 +1,6 @@
-
 const MeetingMemberId = {
-  join: (meetingId: string, userId: string) => `${userId}::${meetingId}`,
-  split: (id: string) => {
+  join: ({meetingId, userId}: {meetingId: string; userId: string}) => `${userId}::${meetingId}`,
```
mattkrick, line 2: are you opposed to keeping the same API as before with 2 params? Typescript will still tell us which is which & then `MeetingMemberId.join` has the same signature as all the other gqlIds

packages/server/graphql/types/User.ts
```diff
@@ -442,19 +442,55 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
...
+          for (const groupedReflection of groupedReflectionsRes) {
+            const {reflectionGroupId, oldReflectionGroupId} = groupedReflection
+            if (
+              reflectionGroupId === spotlightGroup.reflectionGroupId &&
+              oldReflectionGroupId !== spotlightGroup.oldReflectionGroupId
+            ) {
+              spotlightResultGroupIds.add(oldReflectionGroupId)
+            }
+            currentThresh = nextThresh
+            if (spotlightResultGroupIds.size >= spotlightResultGroupSize) break
```
mattkrick, line 486: don't we want to break before the maximum is reached?
- save the results as R1
- compute next results R2
- if size R2 > max, return R1
- else R1 = R2

mattkrick (changes requested):
nice work! couple housekeeping things, i think only having 1 droppable attribute should greatly clean up a lot of the complexity here.

packages/client/shared/gqlIds/MeetingMemberId.ts
nickoferrall: I kinda like passing in an object when two args are of the same type as it prevents us from mixing up the order. But I'm happy to keep things consistent and revert that change 👌

PR #5380, opened 2021-09-08
Fix undefined phaseType bug

### LINKED ISSUE #5307: TypeError: Cannot read property 'phaseType' of undefined
134 events affecting 36 users, first seen in last release 6.24.0. seems related to older #4480

https://sentry.io/organizations/parabol/issues/2588956277/?referrer=github_plugin

```
TypeError: Cannot read property 'phaseType' of undefined
  at c (./packages/client/components/MeetingSidebarTeamMemberStageItems.tsx:57:58)
...
(13 additional frame(s) were not displayed)
```

nickoferrall:
Fixes #5307 

We [previously checked](https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/client/components/ActionSidebarPhaseListItemChildren.tsx#L27) whether the `localPhase` type equaled the `phaseType` before rendering the component.

We now need to always render the `MeetingSidebarTeamMemberStageItems` component during the check-in and updates phases so that the sidebar transition can work.

Checking whether `localPhase` exists in `MeetingSidebarTeamMemberStageItems` does the trick as the `useAnimatedPhaseListChildren` will return a height of 0 if it doesn't exist.

While the `sidebarPhase` should always exist, I've removed the bang operator and added additional checks, just to be extra cautious. 

I'm not sure why the `localPhase` can be null, however. Our client schema shows `localPhase` as non-nullable.

This is a p1 bug and the ship is scheduled today so I'll go straight to maintainer reviewer.

### DIFF
--- a/packages/client/components/MeetingSidebarTeamMemberStageItems.tsx
+++ b/packages/client/components/MeetingSidebarTeamMemberStageItems.tsx
@@ -36,23 +36,25 @@ interface Props {
 const MeetingSidebarTeamMemberStageItems = (props: Props) => {
   const {gotoStageId, handleMenuClick, meeting, phaseType} = props
   const {facilitatorStageId, facilitatorUserId, localPhase, localStage, phases} = meeting
-  const sidebarPhase = phases.find((phase) => phase.phaseType === phaseType)!
+  const sidebarPhase = phases.find((phase) => phase.phaseType === phaseType)
   const localStageId = (localStage && localStage.id) || ''
   const gotoStage = (teamMemberId) => () => {
-    const teamMemberStage = sidebarPhase.stages.find((stage) => stage.teamMemberId === teamMemberId)
+    const teamMemberStage =
+      sidebarPhase && sidebarPhase.stages.find((stage) => stage.teamMemberId === teamMemberId)
     const teamMemberStageId = (teamMemberStage && teamMemberStage.id) || ''
     gotoStageId(teamMemberStageId).catch()
     handleMenuClick()
   }
   const atmosphere = useAtmosphere()
   const {viewerId} = atmosphere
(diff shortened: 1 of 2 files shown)


packages/client/components/MeetingSidebarTeamMemberStageItems.tsx
```diff
@@ -36,23 +36,25 @@ interface Props {
 const MeetingSidebarTeamMemberStageItems = (props: Props) => {
   const {gotoStageId, handleMenuClick, meeting, phaseType} = props
   const {facilitatorStageId, facilitatorUserId, localPhase, localStage, phases} = meeting
-  const sidebarPhase = phases.find((phase) => phase.phaseType === phaseType)!
+  const sidebarPhase = phases.find((phase) => phase.phaseType === phaseType)
   const localStageId = (localStage && localStage.id) || ''
   const gotoStage = (teamMemberId) => () => {
-    const teamMemberStage = sidebarPhase.stages.find((stage) => stage.teamMemberId === teamMemberId)
+    const teamMemberStage =
+      sidebarPhase && sidebarPhase.stages.find((stage) => stage.teamMemberId === teamMemberId)
```
mattkrick, line 43: can always write `sidebarPhase?.stages` to clean it up a bit. not a show stopper

mattkrick (approved):
wasn't sure how to test, but the code looks good. i'll trust ya that it works!

nickoferrall:
Ah I should have left testing steps in the description. I tested it by assigning `localPhase` to `undefined` and then did the same with `sidebarPhase`. Previously, this caused a crash but now it doesn't; the child height is 0 instead

PR #5395, opened 2021-09-13
Add more logging info to Stripe Quantity Mismatch issue

tianrunhe:
Since we couldn't root cause this issue I'm just adding more logging to it so we could get more info. We only have like ~300 events in the last 3/4 months so the excessive amount of logging shouldn't be a big issue.  
  
Let me know if there are anything more to add to help with debugging.


packages/server/billing/helpers/processInvoiceItemHook.ts
```diff
@@ -103,17 +103,23 @@ const processInvoiceItemHook = async (stripeSubscriptionId: string) => {
...
       .filter({
         inactive: false,
         removedAt: null
       })
-      .count()
       .run()
+    const orgUserCount = orgUsers.length
     if (orgUserCount !== nextQuantity) {
-      sendToSentry(new Error('Stripe Quantity Mismatch'), {
+      const errorStr = `Stripe Quantity Mismatch:
```
mattkrick, line 116: we get 8192 chars (https://develop.sentry.dev/sdk/event-payloads/message/)
this will be truncated.
could you make a pg table & push there instead?

packages/server/postgres/queries/insertStripeQuantityMismatchLogging.ts
```diff
@@ -0,0 +1,35 @@
...
+      eventTime,
+      eventType,
+      stripePreviousQuantity,
+      stripeNextQuantity,
+      orgUsers: orgUsers.map(({id, inactive, joinedAt, removedAt, userId, role, tier}) => {
+        return {
+          id,
+          inactive,
+          joinedAt: joinedAt.getTime(),
+          removedAt: removedAt?.getTime() ?? -1,
```
mattkrick, line 25: instead of `getTime()` could we use `toJSON`? that'll give us an ISO8601 date string which will still be sortable, but also human readable.

the fallback value can just be an empty string

packages/server/postgres/migrations/1632333793178_addStripeQuantityMismatchLoggingTable.ts
```diff
@@ -0,0 +1,28 @@
+import {Client} from 'pg'
+import getPgConfig from '../getPgConfig'
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+  CREATE TABLE IF NOT EXISTS "StripeQuantityMismatchLogging" (
```
mattkrick, line 8: love it!

mattkrick (changes requested):
looks great!
only change is to human-readable iso8601 dates

packages/server/postgres/queries/insertStripeQuantityMismatchLogging.ts
```diff
@@ -0,0 +1,35 @@
...
+      eventTime,
+      eventType,
+      stripePreviousQuantity,
+      stripeNextQuantity,
+      orgUsers: orgUsers.map(({id, inactive, joinedAt, removedAt, userId, role, tier}) => {
+        return {
+          id,
+          inactive,
+          joinedAt: JSON.stringify(joinedAt),
+          removedAt: removedAt ? JSON.stringify(removedAt) : "",
```
mattkrick, line 25: nope, JSON.stringify will turn the Date object to a string by calling `.toJSON` and THEN turn that string into JSON, which means it'll give you an extra set of quotes

tianrunhe: Oh! I did not know these nuances and understand you wrongly before! Thanks for pointing it out!

mattkrick: javascript is a horrible, strange language

Issue #5390, opened 2021-09-10
Spotlight: groups UI


nickoferrall:
Predecessors: #5183 

Now that groups are showing up in the Spotlight, make them look like the Figma designs outlined in https://github.com/ParabolInc/parabol/issues/5105

### AC

- [ ] On desktop, there are at least three columns of similar reflection groups
- [ ] On mobile, there is only one column
- [ ] The spacing between the reflection groups follows the designs outlined in Figma
- [ ] Grouping or ungrouping a reflection group does not cause other untouched reflection groups to move
- [ ] if a reflection is removed from the search results by another user grouping it somewhere else, it's vacancy in the search results is animated away in smooth way (this requires absolutely positioning the cards in a masonry grid)

PR #5402, opened 2021-09-15
Spotlight groups UI

nickoferrall:
Fixes https://github.com/ParabolInc/parabol/issues/5390

### Test:

- [ ] Open the Spotlight modal and see that the modal hugs the reflection groups
- [ ] Open the Spotlight modal with only a few reflection groups. See that the size of the Spotlight modal is smaller
- [ ] Open the Spotlight modal with many large (4 lines) reflection cards. See that the size of the Spotlight modal is larger
- [ ] Open on mobile and see that it takes up most of the screen
- [ ] Remotely add reflection groups to the Spotlight or ungroup existing groups to increase the size of the results section. See that the modal's height doesn't increase, and an overflow scroll appears instead.
- [ ] After clicking the icon to open the Spotlight, the selected reflection card animates into the Spotlight and is visible throughout the animation


nickoferrall:
Hey @enriquesanchez, would you mind reviewing this PR?

Two points in particular that I'd like your thoughts on:

1. In Slack, we mentioned preventing the result groups from being groupable. I realised that at the very least, remote reflections (where another user is dragging the reflection) should probably be groupable. If I have a group on the Spotlight with reflections "cat" and "kittens" and another user ungroups them, the Spotlight should reflect that change. We could allow only remote reflections to be groupable & ungroupable and prevent the user from grouping & ungrouping results but I think it's simpler to keep the behaviour consistent.

2. To me, it can feel like there's quite a lot of white space in the Spotlight, particularly if there are less than 10 result groups. However, if the height adjusts based on the number of result groups, when grouping/ungrouping a group, the total height may change causing all of the groups to change position (see gif below). Instead, I think it's best for the height to remain fixed and to design for the expected view of 10 reflection groups on desktop. 

groups

packages/client/components/SpotlightRoot.tsx
```diff
@@ -20,11 +20,15 @@ const SpotlightRoot = (props: Props) => {
...
-    meetingId
-  })
+  const queryRef = useQueryLoaderNow<SpotlightModalQuery>(
+    spotlightModalQuery,
+    {
+      reflectionId: reflectionIdRef.current,
+      searchQuery,
+      meetingId
+    },
+    'network-only'
```
nickoferrall, line 30: The inputs to the query are the `spotlightReflectionId` and the `searchQuery`. While these might not change, the results can change significantly if the user or another user groups/ungroups reflections. As the Spotlight is unlikely to be opened frequently, I think it's best if the policy is `network-only` here so that we don't rely on a potentially outdated cache

enriquesanchez:
I'm having trouble compiling this branch on my dev environment, but I think I can still offer some suggestions:

>We could allow only remote reflections to be groupable & ungroupable and prevent the user from grouping & ungrouping results but I think it's simpler to keep the behaviour consistent.

I agree consistency is preferred here. Nothing more frustating for users that not knowing why things don't work as expected. Let's chat when you get back, so we can hash out any edge cases like this one.

>Instead, I think it's best for the height to remain fixed and to design for the expected view of 10 reflection groups on desktop.

I agree. We can make the spotlight modal narrower instead. [In the initial mockups](https://www.figma.com/file/u5bdWUn9hxkdvS2YOYDnXU/Spotlight?node-id=199%3A4598) we had it cover over only the meeting area. Maybe that could help with this.

nickoferrall:
Thanks for the review! Just letting you know that I'm working on feedback in https://github.com/ParabolInc/parabol/pull/5376 as that's the leading branch. Once that's resolved, I'll move onto this one

nickoferrall:
Hey @enriquesanchez, I've pushed a few updates and recorded a Loom to share the changes with you: 
https://www.loom.com/share/f1b99e76fb7b4a24881372beb6e8516f

I'm now calculating the ideal size of the modal as it's opened to keep it nice and compact. I'd be happy to chat on Zoom on Monday if you prefer or in this issue. Have a good weekend!

packages/client/utils/constants.ts
```diff
@@ -144,6 +144,9 @@ export const RETROSPECTIVE = 'retrospective'
 /* Retro DnD types */
 export const DISCUSSION_TOPIC = 'DISCUSSION_TOPIC'
 
-/* Spotlight Groups */
+/* Spotlight */
 export const MAX_REDUCTION_PERCENTAGE = 1
 export const MAX_RESULT_GROUP_SIZE = 10
+export const MAX_SPOTLIGHT_COLUMNS = 3
+export const SPOTLIGHT_TOP_SECTION_HEIGHT = 236
+export const SPOTLIGHT_GROUPS_PADDING = 60
```
nickoferrall, line 152: If a variable is used in two files, I've added it to the constants. If it's just used in one file, I think it's OK to define it locally

packages/client/hooks/useGetRefHeight.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  const getHeight = () => {
+    const {current: el} = ref
+    const height = el?.clientHeight
+    if (!height) return
+    setHeight(height)
+  }
+
+  useLayoutEffect(getHeight, [ref])
+  useResizeObserver(getHeight, observerRef)
```
nickoferrall, line 19: We want to know the height of the group columns but we don't know what it is until it's opened. `useLayoutEffect` makes an initial calculation and `useResizeObserver` then corrects it if it's slightly off

enriquesanchez:
Thanks for recording the Loom @nickoferrall 🙌 

I have to say this is looking great and I agree with you. The compact modal feels much better, and it's smart to only expand it as needed 👍

I was wondering if its possible to have no padding at the bottom edge of the modal. When you have to scroll the abrupt cut off to white can feel weird and disorienting. 

Screen Shot 2021-10-04 at 1 20 11 p m

nickoferrall:
Thanks, @enriquesanchez! I think we'll need some bottom padding so that when there's not enough space, the bottom reflections aren't touching the modal edge, but I definitely agree that we can reduce it. 

I've pushed an update reducing the bottom padding. Here's a screenshot with many large reflections:

large-init-view

And when a remote reflection is added to the Spotlight: 

with-remote-reflection

enriquesanchez:
@nickoferrall Much better, thank you! 🙌

nickoferrall:
Thank you, @enriquesanchez! Moving this to Maintainer Review

nickoferrall:
Note for @mattkrick: please review https://github.com/ParabolInc/parabol/pull/5376 before this one as this PR uses that one as its base

mattkrick:
CR complete! next big thing is to refactor the modal so the DOM nodes look like they would if we weren't animating it. Then, use a clone to add in some nice animations.

packages/client/components/SpotlightGroups.tsx
```diff
@@ -1,67 +1,109 @@
...
-            phaseRef={phaseRef}
-            reflectionGroup={reflectionGroup}
-          />
-        )
-      })}
-    </Container>
+    <SimilarGroups>
+      <Scrollbar height={scrollHeight} ref={resultsRef}>
+        {groupMatrix.map((row) => (
+          <Column key={`row-${row[0].id}`}>
```
nickoferrall, line 94: `row[0].id` can be the same as the `group.id` key below. Differentiate with `row-` prefix

packages/client/hooks/useAnimatedSpotlightSource.ts
```diff
@@ -0,0 +1,67 @@
...
+import clientTempId from '~/utils/relay/clientTempId'
+import {BezierCurve} from '~/types/constEnums'
+import {Times} from 'parabol-client/types/constEnums'
+import {Elevation} from '~/styles/elevation'
+import cloneReflection from '~/utils/retroGroup/cloneReflection'
+import {PortalStatus} from '~/hooks/usePortal'
+import {MutableRefObject, useLayoutEffect, useRef} from 'react'
+import StartDraggingReflectionMutation from '~/mutations/StartDraggingReflectionMutation'
+
+const useAnimatedSpotlightSource = (
```
nickoferrall, line 11: Any preference on present vs past tense?

packages/client/hooks/useAnimatedSpotlightSource.ts
```diff
@@ -0,0 +1,67 @@
...
+    const sourceCloneBbox = sourceClone.getBoundingClientRect()
+    const {style: sourceStyle} = source
+    sourceStyle.opacity = '0' // hide source while animating sourceClone into modal
+    const clone = cloneReflection(sourceClone, reflectionId)
+    const {style: cloneStyle} = clone
+    const {left: startLeft, top: startTop} = sourceCloneBbox
+    const {left: endLeft, top: endTop} = sourceBbox
+    const roundedEndTop = Math.round(endTop) // fractional top pixel throws off transform position
+    cloneStyle.left = `${startLeft}px`
+    cloneStyle.top = `${startTop}px`
```
nickoferrall, line 35: In the Loom, you mentioned that we don't want to set the initial left & top. However, `cloneReflection` creates an element without defining its left or top; without doing this here, the clone will be at the top left of the screen.

packages/client/hooks/useGroupMatrix.ts
```diff
@@ -0,0 +1,97 @@
...
+    while (getNextWidth(colCount + 1) < width && colCount + 1 <= MAX_SPOTLIGHT_COLUMNS) {
+      colCount++
+    }
+    return colCount
+  }
+
+  const initMatrix = (): Group[][] => {
+    const columnsCount = getColumnsCount()
+    if (columnsCount === null) return []
+    const matrix = Array.from(new Array(columnsCount), () => []) as Group[][]
```
nickoferrall, line 33: Original suggestion was with `const matrix = Array(columnsCount).fill([]) as Group[][]`. I learnt something new as it turns out `Array.fill()` fills the array with a reference to the same array. `Array.from` with a callback has the desired behaviour: https://stackoverflow.com/questions/41121982/strange-behavior-of-an-array-filled-by-array-prototype-fill

packages/client/components/SpotlightModal.tsx
```diff
@@ -1,56 +1,57 @@
...
-  background: '#FFFF',
+import ResultsRoot from './ResultsRoot'
+import {MAX_SPOTLIGHT_COLUMNS, SPOTLIGHT_TOP_SECTION_HEIGHT} from '~/utils/constants'
+import {GroupingKanban_meeting} from '~/__generated__/GroupingKanban_meeting.graphql'
+
+const desktopBreakpoint = makeMinWidthMediaQuery(Breakpoint.SIDEBAR_LEFT)
+const MODAL_PADDING = 72
+
+const Modal = styled('div')<{isLoadingResults: boolean}>(({isLoadingResults}) => ({
+  animation: isLoadingResults
```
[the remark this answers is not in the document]
nickoferrall: We want the source to animate in over 300ms and then the modal to fade in behind it over 300ms. However, we need to know the destination position of the source before we animate it into the modal. This position can change as the height of the modal changes depending on the results. 

To do this, I'm:
- rendering the modal with opacity 0
- using the `portalStatus` to know when the modal has `Entered`
- once it has entered, we know the source destination position and can animate it into the modal
- once the source has animated into the modal, change the modal opacity to 1 and fade it in using a transition 

I've added a comment here to better explain the logic.

packages/client/components/GroupingKanban.tsx
```diff
@@ -38,36 +38,44 @@ export type SwipeColumn = (offset: number) => void
...
-  const {closePortal, openPortal, modalPortal} = useModal({
+
+  const {closePortal, openPortal, modalPortal, portalStatus} = useModal({
     onClose: closeSpotlight,
     id: 'spotlight'
   })
+  const {sourceRef, sourceCloneRef} = useAnimatedSpotlightSource(
+    portalStatus,
+    spotlightReflection?.id,
+    dragIdRef
```
nickoferrall, line 62: `StartDraggingReflectionMutation` needs to happen after the cloning as the reflection height will be altered. So, I've moved the mutation from `useSpotlightSimulateDrag` to `useAnimatedSpotlightSource`. The downside is we need to share `dragIdRef` across both hooks.

packages/client/hooks/useAnimatedSpotlightSource.ts
mattkrick: no strong preference

packages/client/hooks/useAnimatedSpotlightSource.ts
mattkrick: i think i messed up.
setting an initial left, top is AOK, we just want to be sure we don't update the left, top

packages/client/components/GroupingKanban.tsx
mattkrick: the flow is:
- reflection is 88px tall in the kanban
- if you trigger start dragging, then the height will reduce if it's the only one in the group (see useCollapsePlaceholder)
-  so, we need to clone first before calling start drag

mattkrick: i think this is good! it's the smallest nit of wondering if we could combine all the "opening" logic into 1 chunk, but we can think about refactoring later

nickoferrall: One possible solution which might mean we can clone the reflection after executing `StartDraggingReflectionMutation` is adjusting the `isClipped` logic - see [this comment](https://github.com/ParabolInc/parabol/pull/5610/files#r742380685).

But, as you mention in your comment, I'd prefer to merge this PR and refactor later.

PR #5403, opened 2021-09-15
fixed ~200 noImplicitAny errors, mostly on graphql type files

### LINKED ISSUE #5314: Add noImplicitAny to server tsconfig (Part 1 of 8)

- [ ] fix ~200 errors
- [ ] we'll merge it blindly
- [ ] create a copy of this issue named Part (2 of 8) so we can re-prioritize for the next person. We'll keep doing this until it's done

AC

Estimate: 4 hours

tiffanyhan:
Resolves #5314

### DIFF
--- a/packages/server/graphql/types/Organization.ts
+++ b/packages/server/graphql/types/Organization.ts
@@ -7,6 +7,8 @@ import {
   GraphQLObjectType,
   GraphQLString
 } from 'graphql'
+import OrganizationUser from '../../database/types/OrganizationUser'
+import {IGetTeamsByIdsQueryResult} from '../../postgres/queries/generated/getTeamsByIdsQuery'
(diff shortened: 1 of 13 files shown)


packages/server/graphql/types/Organization.ts
```diff
@@ -20,7 +22,7 @@ import Team from './Team'
 import TierEnum from './TierEnum'
 import User from './User'
 
-const Organization = new GraphQLObjectType<any, GQLContext>({
+const Organization: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLContext>({
```
mattkrick, line 25: interesting that TS has to infer `any` because of the circular reference in the fields!

mattkrick:
LGTM! not sure if still in draft or ready to merge, but feel free to merge when ready

PR #5408, opened 2021-09-16
Convert remaining User reads to PG

Dschoordsch:
Added new IUser type which is derived from the generated Postgres type and replaced most usages of the Rethink User type with it. The old type is left to define the Rethink interface and for creation of new Users.

Added custom redis cache query to read user from PG instead of RethinkDB.

Fixes: #5372


packages/server/graphql/types/User.ts
```diff
@@ -119,15 +120,13 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
...
-        return r
-          .table('User')
-          .filter((row: any) => row('email').match(`${domain}$`))
-          .orderBy('createdAt')
-          .nth(0)('id')
-          .eq(userId)
-          .run()
+        const pg = getPg()
+        const patientZeroId = await pg.query(
+          'SELECT id FROM "User" WHERE split_part(email, \'@\', 2) = $1 ORDER BY "createdAt" LIMIT 1',
```
Dschoordsch, line 126: `split_part(email, \'@\', 2) as "domain"` is used quite often, so we should consider storing it instead https://www.postgresql.org/docs/12/ddl-generated-columns.html

Dschoordsch:
@nickoferrall could you please review?

Dschoordsch:
@nickoferrall I forgot to include the list of queries/mutations to test. Would you mind to give it a go?

packages/server/graphql/intranetSchema/queries/helpers/authCountByDomain.ts
```diff
@@ -1,30 +1,33 @@
...
-    .merge((user) => ({
-      domain: user('email')
-        .split('@')(1)
-        .default(null)
-    }))
-    .group('domain')
-    .count()
-    .ungroup()
-    .map((row) => ({domain: row('group'), total: row('reduction')}))
-    .orderBy(r.desc('total'))
```
Dschoordsch, line 26: That `orderBy` got lost

PR #5412, opened 2021-09-17
Spotlight: source to result

### LINKED ISSUE #5185: Spotlight: Support drag & drop for source to result
Predecessor: #5182

Now that the results are populated, we should allow the viewer to drag the source reflection at the top onto a group.
See https://github.com/ParabolInc/parabol/issues/5105 for more info

AC
- [ ] can drag the source reflection onto any group in the results space
- [ ] after dropping on a group, it animates being dropped
- [ ] after the animation, the modal animates closed

Estimate: 4 hours

nickoferrall:
Fixes #5185 

One of the changes in this PR is refactoring`spotlightReflection` to `spotlightGroup`. This is required so that results can be dragged onto the source.

### To test 

- [ ] Can drag the source group onto a result group
- [ ] When this happens, the modal closes
- [ ] If the viewer opens the Spotlight by clicking a reflection group that contains several reflections, the source is the reflection at the top of the stack (other reflections in the group are hidden)


packages/client/components/SpotlightModal.tsx
```diff
@@ -183,8 +186,28 @@ const SpotlightModal = (props: Props) => {
...
+      spotlightReflectionIds.current = [firstReflectionId]
+    }
+    // TODO: uncomment for groups -> source issue
+    // else if (firstReflectionId && secondReflectionId && ids?.includes(secondReflectionId)) {
+    //   spotlightReflectionIds.current = [...ids, firstReflectionId]
+    // }
+    else if (!firstReflectionId || !ids?.includes(firstReflectionId)) {
+      setTimeout(() => {
+        closeSpotlight()
+      }, Times.REFLECTION_DROP_DURATION)
```
nickoferrall, line 208: Instead of `setTimeout`, we could get the reflection that was in the Spotlight, find that reflection in the meeting reflections, wait until `isDropping` is false, and then close the Spotlight. I like to avoid `setTimeout`s wherever possible but I think this is prob easier to understand and more performant.

packages/client/components/GroupingKanban.tsx
```diff
@@ -92,14 +92,14 @@ const GroupingKanban = (props: Props) => {
...
+  const openSpotlight = (reflectionGroupId: string, reflectionRef: RefObject<HTMLDivElement>) => {
+    spotlightGroupRef.current = reflectionRef.current
     openPortal()
     commitLocalUpdate(atmosphere, (store) => {
       const meeting = store.get(meetingId)
-      const reflection = store.get(reflectionId)
-      if (!reflection || !meeting) return
-      meeting.setLinkedRecord(reflection, 'spotlightReflection')
+      const reflectionGroup = store.get(reflectionGroupId)
+      if (!meeting || !reflectionGroup) return
```
BartoszJarocki, line 101: is this a normal case? If not it'd be nice to add a warning here so if that happens we can track what was wrong 💡

packages/client/components/ReflectionGroup/ReflectionGroup.tsx
```diff
@@ -66,24 +66,37 @@ interface Props {
...
+    swipeColumn,
+    dataCy,
+    spotlightReflectionIds
+  } = props
   const groupRef = useRef<HTMLDivElement>(null)
   const {localPhase, localStage} = meeting
   const {phaseType} = localPhase
   const {isComplete} = localStage
   const {reflections, id: reflectionGroupId, titleIsUserDefined} = reflectionGroup
+  const isSpotlightSource = spotlightReflectionIds?.length
```
BartoszJarocki, line 87: The name suggests it should be a boolean so maybe it's better to `!!spotlightReflectionIds?.length`

packages/client/components/SpotlightModal.tsx
```diff
@@ -183,8 +186,28 @@ const SpotlightModal = (props: Props) => {
...
+    const {current: ids} = spotlightReflectionIds
+    if (!ids && firstReflectionId) {
+      spotlightReflectionIds.current = [firstReflectionId]
+    }
+    // TODO: uncomment for groups -> source issue
+    // else if (firstReflectionId && secondReflectionId && ids?.includes(secondReflectionId)) {
+    //   spotlightReflectionIds.current = [...ids, firstReflectionId]
+    // }
+    else if (!firstReflectionId || !ids?.includes(firstReflectionId)) {
+      setTimeout(() => {
```
BartoszJarocki, line 206: It we end up using `setTimeout` it'd be nice to call `clearTimeout` in `useEffect` cleanup

BartoszJarocki (commented):
A few minor comments to consider, looks good otherwise! 👌

nickoferrall:
It looks like this is ready for Maintainer Review, but it's stuck behind the Groups UI PR, which isn't ready for Maintainer Review yet. I'll move this PR to Stuck

packages/client/components/GroupingKanban.tsx
mattkrick: these things should always exist, but using relay's imperative API it can't guarantee `store.get` is going to return something.

You could write something like this

```ts
      const meeting = store.get<GroupingKanban_meeting>(meetingId)!
```

because the meeting object is at least guaranteed to have the fields that this component has requested

nickoferrall:
[Groups UI](https://github.com/ParabolInc/parabol/pull/5402) is now in Maintainer Review, so I'll move this one out of stuck and into Maintainer Review too

packages/client/components/GroupingKanban.tsx
nickoferrall: Personally, I feel more secure using the return statement instead of the bang, even if we know it should exist. The return statement adds an extra line & it could be confusing to devs wondering why it's there (although the bang could also be confusing), but if there's a strange bug in Relay, it could prevent a crash. 

I've pushed an update adding the meeting type. Happy to add the bang instead if anyone feels tension!

packages/client/components/ReflectionGroup/ReflectionGroup.tsx
```diff
@@ -171,6 +192,7 @@ const ReflectionGroup = (props: Props) => {
         {...(isBehindSpotlight ? null : {[DragAttribute.DROPPABLE]: reflectionGroupId})}
         ref={groupRef}
         staticReflectionCount={staticReflections.length}
+        isSpotlightSource={isSpotlightSrcGroup && !isBehindSpotlight}
```
nickoferrall, line 195: We use the `staticReflectionCount` to determine the `bottomPadding` on [L35](https://github.com/ParabolInc/parabol/blob/a20a10d19b23c739a823a07488586e9aede6f0aa/packages/client/components/ReflectionGroup/ReflectionGroup.tsx#L35). The Spotlight clone doesn't know what the `staticReflectionCount` is, resulting in the clone & source having different padding (only when `spotlightGroup` is a stack of reflections). A few options:
1. Set the `staticReflectionCount` to 1 if it's the source group. The source doesn't need extra padding.
2. Pass an `isSpotlightSource` prop & use a ternary in the `paddingBottom`. 
3. Calculate the `staticReflectionCount` in `useAnimatedSpotlightSource` and add the `paddingBottom` the clone. This is tricky as we need the `visibleReflections` (and therefore `hiddenIds`) to calculate the `staticReflection`.

I've opted for option 2 as it's simpler than 3 and more explicit than 1.

packages/client/hooks/useAnimatedSpotlightSource.ts
```diff
@@ -36,6 +36,7 @@ const useAnimatedSpotlightSource = (
     cloneStyle.borderRadius = `4px`
     cloneStyle.boxShadow = `${Elevation.CARD_SHADOW}`
     cloneStyle.overflow = `hidden`
+    cloneStyle.paddingTop = `6px` // reflectionGroup wrapper adds 6px paddingTop
```
mattkrick, line 39: instead of a comment use a shared var `ElementWidth.REFLECTION_CARD_PADDING`

packages/client/mutations/handlers/handleUpdateSpotlightResults.ts
```diff
@@ -29,13 +29,13 @@ const handleUpdateSpotlightResults = (
     oldReflections?.length === 1 && oldReflections[0].getValue('id') === reflectionId
   const reflectionGroupId = reflectionGroup.getValue('id')
   const groupsIds = similarReflectionGroups.map((group) => group.getValue('id'))
-  const isInSpotlight = groupsIds.includes(reflectionGroupId)
-  const wasInSpotlight = groupsIds.includes(oldReflectionGroupId)
+  const isInSpotlight = [...groupsIds, spotlightGroupId].includes(reflectionGroupId)
```
mattkrick, line 32: instead of creating the same array twice just concat spotlightGroupId into groupIds above

packages/server/database/migrations/20211025134235-estimateStageTaskIdBackfill.ts
```diff
@@ -1,153 +0,0 @@
-import JiraIssueId from 'parabol-client/shared/gqlIds/JiraIssueId'
```
mattkrick, line 1: i don't think you meant to delete a migration. possible bad merge?

mattkrick (changes requested):
looks like a merge might've gotten wacky as there's some extra files here & a deleted migration

packages/server/database/migrations/20211025134235-estimateStageTaskIdBackfill.ts
nickoferrall: That's right! I pushed the funky merge just as you started reviewing. I should have moved it to Reviewer Review before pushing. All fixed now!

Issue #5187, opened 2021-07-19
Spotlight: Support drag & drop for result to source


mattkrick:
Predecessor: #5182

Now that the results are populated, we should allow the viewer to drag results onto the source.
See https://github.com/ParabolInc/parabol/issues/5105 for more info

AC
- [ ] can drag a card in the results area to the source reflection
- [ ] after dropping on a the source reflection, the card animates onto the top but the modal doesn't close
- [ ] if another participant is dragging a result, that result is shown as being dragged.

@enriquesanchez what happens if the thing they're dragging comes from a group? do you drag the top card of that group? 
similarly, what if the source reflection is a group? do you drag the whole group or just the top card?

Estimate: 4 hours

enriquesanchez:
@mattkrick I think we should follow existing behaviors. Currently if I drag a card from a group, it removes it from that group and puts it back in the column (or creates a new group if I dropped it on top of a different card). It's also not possible to drag a whole group on to a card, or on another group for that matter. So I think we should keep those constrains here in Spotlight as well. So with that logic in mind:

>what happens if the thing they're dragging comes from a group?
- [ ] If I'm dragging a source card on to a result card, it removes the source card from the group it is currently in, and creates a new group with the card it is dropped on.
- [ ] If instead I drag a reflection from the results and drop it on the source reflection, and this source reflection is already in a group, then the card gets added to the existing group.

>what if the source reflection is a group? do you drag the whole group or just the top card?
Assuming you meant that the source reflection _is in a group_, then the above applies here as well. But if what you meant is that _the source is a group_, then this one I didn't think of before, but I don't think a whole group should be the source for Spotlight, only single cards. This is already constrained by the fact that the spotlight trigger icon-button only appears in cards.

It is a valid use case though, wanting to find a reflection that matches a group. I think it'll be a great v2 addition once we have everything else polished.

How does all this sound to you?

mattkrick:
i agree we should keep the logic the same, dragging just the top one from the top feels a little clunky, but clunky + similar is better than new probably (and definitely faster to build).

i like drawing a hard line at making the source reflection a single reflection instead of a group. feels like the path of least resistance. there will be some edge cases though. for example, let's say you & i are in a meeting:
-  i have spotlight open with card "Cats"
- you drag card "Cats" onto "Dogs"
- What shows up on my screen? Do i still see Cats at the top? Do I see "Dogs" in my search results? How do I know that you did anything to my "Cats" card?

PR #5414, opened 2021-09-17
Spotlight: result to source

nickoferrall:
Fix for issue #5187 

### To test:

- [ ] Can drag a result group onto the source group
- [ ] Cannot drag a result group onto another result group
- [ ] Opening & closing an expanded reflection stack portal in the Spotlight results doesn't close the Spotlight
- [ ] If the source has several reflections in the group, ungroup a reflection by dropping it onto the reflection grid (not a reflection group). See that the source card returns to the Spotlight results rather than its kanban position.


tiffanyhan (approved):
lgtm, lookin very slick!

mattkrick:
code looks good, will move to stuck & merge after the parent gets merged

PR #5396, opened 2021-09-13
Remove postdeploy migrations

mattkrick:
Fix #5374

TEST
- [ ] from a clean repo clone, go through the readme steps & start the app. it works!
- [ ] from a clean repo clone, copy a slice of the prod rethinkdb database & a slice of the pg database (or the whole thing if you can't take a slice of PG) & start the app. it works!

PR #5423, opened 2021-09-22
Push Estimate to GitHub Label

### LINKED ISSUE #4718: Sprint Poker: GitHub: Estimate: Push to tag
GH doesn't offer a story points field.
Instead, they have labels.
a common pattern seen in large OSS projects is to have a label like `Story points: 5`.
That means we need a way for them to create a label template (e.g. `${dimensionName}: ${points}`)
It's also possible that folks use a weighted shortest job first, where the effort is 5 & user value is 10 so the value would be 2. Should we push to a tag like `Value: 2`?
AC
- [ ] in the push-to-github dropdown, above "as comment" is "as label" that has an edit button
- [ ] clicking the edit button opens up a modal that defaults to `{{dimensionName}}: {{score}}`
- [ ] below the input is a example of what the label will look like, where score is the first value from the scale (e.g. `Story points: 5`)
- [ ] submitting the score will add the label to the issue in GH. if the label doesn't exist, we create it (if we can, if not, a friendly error message says the facilitator is not the owner of the repo)

Estimate: 24 hours

mattkrick:
fix #4718

TEST
- [x] start a poker meeting & integrate github & jira
- [x] add a couple github issues & a couple jira issues
- [x] create a new github issue
- [x] when you add an issue, the left nav updates instantly & correctly
- [x] going to the estimate phase, the issue header card looks correct
- [x] by default, a github issue wants to push the score to a comment
- [x] pushing the score to comment adds a comment to the issue on github
- [x] can select "Add as label" and add a label to the GitHub issue
- [x] can change the label template & push as a different label
- [x] if the label doesn't exist, we create the label in github
- [x] the label color matches the color of the card

BURNDOWN
- [x] nest-graphql-endpoint supports custom headers (GitHub's createLabel API requires beta access that is turned on with a header)
- [x] nest-graphql-endpoint requires a custom github schema that includes the beta features
- [x] refactor templateScaleRefs to dataloader
- [x] valid label length limits
- [x] found out i never implemented create github task, so adding that to this PR


packages/client/hooks/useForm.tsx
```diff
@@ -91,7 +91,7 @@ const useForm = (fieldInputDict: FieldInputDict, deps: any[] = []) => {
             }
           }
           return obj
-        }, {} as FieldState),
+        }, {} as FieldState<T>),
```
mattkrick, line 94: this little guy means the return value has the same keys as the input value. 
i hate the extra PR bloat, but i hit a runtime error so decided to add a little extra type safety.

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/SummaryPokerStories.tsx
```diff
@@ -70,7 +70,7 @@ const SummaryPokerStories = (props: Props) => {
                 } else if (integration?.__typename === '_xGitHubIssue') {
                   title = integration.title
                 }
-                const urlPath = `/meet/${meetingId}/estimate/${usedTaskIds.size}`
+                const urlPath = `/meet/${meetingId}/estimate/${idx + 1}`
```
mattkrick, line 73: found a bug that affected the summary of poker meetings that had 2+ dimensions

packages/client/mutations/CreateTaskMutation.ts
```diff
@@ -50,6 +52,9 @@ graphql`
         ... on _xGitHubIssue {
           number
           title
+          repository {
```
mattkrick, line 55: I'd love to use the `@appendNode` directive, but that requires a connectionId that is specific to the viewer. since our subscriptions share the same fragment, i'm not sure how best to do that declaratively

packages/client/mutations/UpdateJiraDimensionFieldMutation.ts
```diff
@@ -99,19 +100,22 @@ const UpdateJiraDimensionFieldMutation: SimpleMutation<TUpdateJiraDimensionField
...
-          stage.getValue('serviceTaskId') as string
-        )
-        if (stage.getValue('dimensionName') === dimensionName && stageCloudId === cloudId) {
+        const dimensionRef = stage.getLinkedRecord('dimensionRef')
+        const dimensionRefName = dimensionRef.getValue('name')
+        if (dimensionRefName !== dimensionName) return
+        const task = stage.getLinkedRecord('task')
+        const _integration = task.getLinkedRecord('integration')
+        if (_integration.getValue('__typename') !== 'JiraIssue') return
+        const integration = _integration as DiscriminateProxy<typeof _integration, 'JiraIssue'>
```
mattkrick, line 111: this just tells typescript that the integration is a proxy of  a JiraIssue (vs. GitHub or anything else).
My PR to relay is pending, so this is how we have to do it for now

packages/client/mutations/handlers/handleGitHubCreateIssue.ts
```diff
@@ -0,0 +1,44 @@
+import {ConnectionHandler, RecordProxy, RecordSourceSelectorProxy} from 'relay-runtime'
+import SearchQueryId from '../../shared/gqlIds/SearchQueryId'
+import toTeamMemberId from '../../utils/relay/toTeamMemberId'
+import {CreateTaskMutationResponse} from '../../__generated__/CreateTaskMutation.graphql'
+import getGitHubIssuesConn from '../connections/getGitHubIssuesConn'
+
+const handleGitHubCreateIssue = (
```
mattkrick, line 7: this function is so gross & i want a better way to do this.
the problem is I don't know a better way that also works with subscriptions.

packages/client/shared/gqlIds/TeamMemberId.ts
```diff
@@ -1,8 +1,8 @@
 
 const TeamMemberId = {
-  join: (teamId: string, userId: string) => `${userId}:${teamId}`,
+  join: (teamId: string, userId: string) => `${userId}::${teamId}`,
   split: (id: string) => {
-    const [userId, teamId] = id.split(':')
+    const [userId, teamId] = id.split('::')
```
mattkrick, line 5: this helper isn't used anywhere (yet) probably because it was buggy.
we'll use this in the future to replace `toTeamMemberId` and `fromTeamMemberId`

packages/client/utils/relay/handlerProvider.ts
```diff
@@ -1,17 +1,15 @@
...
   switch (handle) {
-    case 'connection':
-      return ConnectionHandler
     case 'localTime':
       return LocalTimeHandler
     case 'localPoker':
       return LocalPokerHandler
     default:
-      throw new Error(`Unknown handle ${handle}`)
+      return DefaultHandlerProvider(handle)
```
mattkrick, line 12: relay is adding more & more custom types so this should future proof us better

packages/server/graphql/mutations/endSprintPoker.ts
```diff
@@ -66,7 +66,7 @@ export default {
       return standardError(new Error('Cannot find facilitator stage'), {userId: viewerId})
     }
     const storyCount = new Set(
-      estimateStages.filter(({isComplete}) => isComplete).map(({serviceTaskId}) => serviceTaskId)
+      estimateStages.filter(({isComplete}) => isComplete).map(({taskId}) => taskId)
```
mattkrick, line 69: honestly not sure how this isn't a bug in prod (maybe it is?)

mattkrick:
@nickoferrall GitHub suggested you as my reviewer so here we go!
Sorry about the size, I made a couple typings more strict & that had a chain reaction across a dozen or so files.

nickoferrall:
Hey @mattkrick, here's [the comment](https://github.com/ParabolInc/parabol/pull/5396#issuecomment-926762610) on the postdeploy migration that's preventing me from reviewing this PR. Once that's resolved, I'll review this asap

packages/server/graphql/mutations/helpers/pushEstimateToGitHub.ts
```diff
@@ -0,0 +1,226 @@
...
+      const {dimensions} = templateRef
+      const dimensionRef = dimensions.find((dimension) => dimension.name === dimensionName)
+      if (!dimensionRef) return new Error('Dimension not found')
+      const {scaleRefId} = dimensionRef
+      const scaleRef = await dataLoader.get('templateScaleRefs').load(scaleRefId)
+      const {values} = scaleRef
+      const matchingValue = values.find((scaleValue) => scaleValue.label === value)
+      if (matchingValue) {
+        // remove # prefix
+        color = matchingValue.color.slice(1)
```
nickoferrall, line 160: The label being the same colour as the card is absolute 🔥 . After removing the score from the label, I noticed that the colour doesn't update with the selected card. That seems like the right move as the label is no longer tied to the card and the user may have selected a specific colour for that label name

packages/server/graphql/mutations/helpers/pushEstimateToGitHub.ts
```diff
@@ -0,0 +1,226 @@
...
+    if (meeting) {
+      const {templateRefId} = meeting
+      const templateRef = await dataLoader.get('templateRefs').load(templateRefId)
+      const {dimensions} = templateRef
+      const dimensionRef = dimensions.find((dimension) => dimension.name === dimensionName)
+      if (!dimensionRef) return new Error('Dimension not found')
+      const {scaleRefId} = dimensionRef
+      const scaleRef = await dataLoader.get('templateScaleRefs').load(scaleRefId)
+      const {values} = scaleRef
+      const matchingValue = values.find((scaleValue) => scaleValue.label === value)
```
nickoferrall, line 157: A small bug I ran into:
- create a label called `Points` and click Update
- edit the label by adding an empty space at the end
- update the Story Points final score and click Update
- see the error message: `Name has already been taken`

It looks like GitHub trims the labels so we'll need to do the same. Currently, if the Parabol & GitHub labels are identical except for whitespace, `matchingValue` is undefined so we try to create a new label, GitHub trims it, and then complains that it already exists

packages/client/components/EditGitHubLabelTemplateModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const INPUT_NAME = 'labelTemplate'
+  console.log({defaultValue})
+  const {fields, onChange, setValue} = useForm({
+    [INPUT_NAME]: {
+      getDefault: () => defaultValue
+    }
+  })
+  const labelTemplateField = fields[INPUT_NAME]
+  const {value} = labelTemplateField
+  const onSave = () => {
```
nickoferrall, line 65: It might just be me, but after saving the new GitHub label, I went to GitHub to see the new label and was surprised that it wasn't already there. Then I returned to Parabol and noticed the `Update` button. 

I think the `Update` button works well when updating the final story points as it's right next to the input, and we couldn't submit the value on each keystroke. Still, I wonder whether saving the new label or selecting `As a label`, `As a comment`, or `Do Not Update` could submit immediately without needing to clicking `Update`?

Not a show stopper by any means, just curious to get your thoughts

packages/client/components/GitHubFieldMenu.tsx
```diff
@@ -49,37 +108,76 @@ const GitHubFieldMenu = (props: Props) => {
...
+      UpdateGitHubDimensionFieldMutation(atmosphere, {
+        dimensionName,
+        labelTemplate,
+        nameWithOwner,
+        meetingId
+      })
+    }
     closePortal()
   }
+  const openEditModal = (e: React.MouseEvent) => {
```
nickoferrall, line 131: Curious whether you prefer to be explicit with React event types vs destructuring them when importing, as we do with `useState`, or whether you don't mind either way?

nickoferrall (approved):
Fantastic stuff! I've been through the Testing steps, and everything is working well. It's helpful to see your thinking in the issue and your comments in the code. There's one minor bug I've mentioned in the comments, and then it looks good to go!

BartoszJarocki (approved):
Great work! I've left a couple of nits to think about 👍

packages/server/graphql/mutations/helpers/pushEstimateToGitHub.ts
mattkrick: yeah, i figure we provide the reasonable defaults but after that they're free to plot their own demise

packages/server/graphql/mutations/helpers/pushEstimateToGitHub.ts
mattkrick: top notch find :fire:

packages/client/components/EditGitHubLabelTemplateModal.tsx
mattkrick: that's a real good idea you've got. basically selecting a menu item should also trigger a submit if there is a pendingScore?
if we do it for GH we should do it for Jira, too, & i think that works OK.

So the possible gotcha with the current pattern is: if I change the field, I have to click "Update" again
The gotcha with the future pattern is: the menu item select behavior is conditional (push vs. no push) based on the value of the pending score (non-null vs null). If i click the wrong menu item, there will be a consequence.

So, what's the probability that someone will click a menu item & NOT hit update? probably pretty darn low. We should implement your strategy :+1:

packages/client/mutations/connections/getGitHubIssuesConn.ts
```diff
@@ -0,0 +1,22 @@
+import {ConnectionHandler, ReadOnlyRecordProxy} from 'relay-runtime'
+
+const getGitHubIssuesConn = (
+  githubTeamMemberIntegration: ReadOnlyRecordProxy | null | undefined,
+  query: string | undefined
+) => {
+  if (githubTeamMemberIntegration) {
```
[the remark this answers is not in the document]
mattkrick: the student has become the teacher! great find

packages/client/mutations/UpdateJiraDimensionFieldMutation.ts
```diff
@@ -99,19 +100,22 @@ const UpdateJiraDimensionFieldMutation: SimpleMutation<TUpdateJiraDimensionField
...
-      const stages = estimatePhase.getLinkedRecords('stages')
+      const stages = estimatePhase.getLinkedRecords<JiraFieldMenu_stage[]>('stages')
       stages.forEach((stage) => {
-        const {cloudId: stageCloudId} = JiraIssueId.split(
-          stage.getValue('serviceTaskId') as string
-        )
-        if (stage.getValue('dimensionName') === dimensionName && stageCloudId === cloudId) {
+        const dimensionRef = stage.getLinkedRecord('dimensionRef')
+        const dimensionRefName = dimensionRef.getValue('name')
+        if (dimensionRefName !== dimensionName) return
```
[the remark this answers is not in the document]
mattkrick: oh this'll happen a LOT! it'll happen when:
- the stage refers to a task that is linked to github (or any other integration) 
- the stage refers to a task linked to a jira issue that lives in a different project or cloud

i could probably write this as a `.filter().forEach()`, but i opted for 1 list iteration over expressiveness.

PR #5429, opened 2021-09-23
Fix: creating a Jira issue

BartoszJarocki:
The endpoint we use to convert markdown to adf doesn't work anymore. Calling `manager.convertMarkdownToADF(markdown)` returns 
```
Error: No message available
      at AtlassianServerManager.post (/Users/bartoszjarocki/Work/parabol/web-parabol/dev/gqlExecutor.js:78329:16)
      at async Promise.all (index 1)
      at async createJiraTask (/Users/bartoszjarocki/Work/parabol/web-parabol/dev/gqlExecutor.js:12503:39)
      at async Object.resolve [as MutationcreateJiraTaskIntegrationResolver] (/Users/bartoszjarocki/Work/parabol/web-parabol/dev/gqlExecutor.js:8454:17),
```
 We don't check errors in the response and as a result, we're trying to create a task with the following payload
```
{
  summary: "I'm a task",
  description: Error: No message available
      at AtlassianServerManager.post (/Users/bartoszjarocki/Work/parabol/web-parabol/dev/gqlExecutor.js:78329:16)
      at async Promise.all (index 1)
      at async createJiraTask (/Users/bartoszjarocki/Work/parabol/web-parabol/dev/gqlExecutor.js:12503:39)
      at async Object.resolve [as MutationcreateJiraTaskIntegrationResolver] (/Users/bartoszjarocki/Work/parabol/web-parabol/dev/gqlExecutor.js:8454:17),
  assignee: { id: '60ede678a211b100694a405b' },
  issuetype: { id: '10002' }
}
```
And the Jira API complains. 

This PR is a little proof of concept that this is the root cause of the reported customer issue 
Screenshot 2021-09-23 at 07 46 07.

I replaced `manager.convertMarkdownToADF(markdown)` with a package https://github.com/b-yond-infinite-network/md-to-adf and it works. The only problem is, the package isn't actively supported so I'm not sure if it's safe to use. I haven't found anything better so any ideas are more than welcome!

### DIFF
--- a/packages/server/graphql/mutations/helpers/convertContentStateToADF.ts
+++ b/packages/server/graphql/mutations/helpers/convertContentStateToADF.ts
@@ -1,0 +1,247 @@
+import {ContentState, ContentBlock} from 'draft-js'
+import {Constants, getEntityRanges} from 'draft-js-utils'
+
+const {INLINE_STYLE, BLOCK_TYPE, ENTITY_TYPE} = Constants
+
+interface Mark {
+  type: string
+  attrs?: Record<string, any>
+}
+
+interface Text {
+  type: 'text'
+  text: string
+  marks?: Mark[]
+}
+
+interface Node {
+  type: string
+  attrs?: Record<string, any>
+  content: (Node | Text)[]
+}
+
+interface Doc {
+  version: 1
+  type: 'doc'
+  content: Node[]
+}
+
+class ContentStateToADFConverter {
+  private contentState: ContentState
+
+  private renderText = (text: string, marks?: Mark[]): Text => ({
+    type: 'text',
+    text,
+    ...(marks?.length ? {marks} : {})
+  })
+
+  private renderContent = (block: ContentBlock): Node[] => {
+    const text = block.getText()
+    const charMetaList = block.getCharacterList()
+    const entityPieces = getEntityRanges(text, charMetaList)
+    const content = entityPieces.flatMap(([entityKey, stylePieces]) => {
+      const contentPieces: Text[] = stylePieces
+        .map(([text, style]) => {
+          if (!text) {
+            return null
+          }
+
+          if (style.has(INLINE_STYLE.CODE)) {
+            return {
+              type: 'text',
+              text,
+              marks: [
+                {
+                  type: 'code'
+                }
+              ]
+            }
+          }
+          const marks: Mark[] = []
+          if (style.has(INLINE_STYLE.BOLD)) {
+            marks.push({
+              type: 'strong'
+            })
+          }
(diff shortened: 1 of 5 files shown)


packages/server/graphql/mutations/helpers/createJiraTask.ts
```diff
@@ -28,7 +26,7 @@ const createJiraTask = async (
   const bestType = issuetypes.find((type) => type.name === 'Task') || issuetypes[0]
   const payload = {
     summary,
-    description,
+    description: convertMdToAdf(markdown),
```
Dschoordsch, line 29: If we skip the conversion to markdown, but instead work with [`contentState.getBlockMap()`](https://draftjs.org/docs/api-reference-content-state#getblockmap) directly, it should be quite simple to convert these few [`ContentBlock` types](https://draftjs.org/docs/api-reference-content-block) to [ADF](https://developer.atlassian.com/cloud/jira/platform/apis/document/structure/) directly.

Dschoordsch (commented):
The b-yond people are using the library themselves [ga-sync-issues-to-jira](https://github.com/b-yond-infinite-network/ga-sync-issues-to-jira), so probably support it, but it also depends on this [adf-builder-javascript](https://github.com/LucasRitter/adf-builder-javascript) which is even older. I propose doing the conversion manually.

mattkrick:
waking up & finding a solution to this is the best feeling in the world!
testing now...

BartoszJarocki:
@mattkrick I think we still need to add support for nested lists. @Dschoordsch did a great job implementing our own conversion, and this seems to be the last piece left. 🙏

packages/server/graphql/mutations/helpers/convertContentStateToADF.ts
```diff
@@ -0,0 +1,247 @@
...
+            ...(data.title ? {title: data.title} : {})
+          }
+        }
+        return contentPieces.map((piece) => ({
+          ...piece,
+          marks: [mark, ...(piece.marks ?? [])]
+        }))
+      }
+      /*
+      TODO: In order to show images or other embedded types, we need to upload those separately to Jira and then refer to them by id
```
mattkrick, line 101: AOK to skip, our draftjs implementation doesn't support images, so we couldn't push em even if we wanted to

mattkrick (approved):

mattkrick:
Test result here: https://parabol-2.atlassian.net/browse/DEV3-129

Conversation on atlassian here: https://community.atlassian.com/t5/Jira-questions/convert-jira-ADF-formatted-description-in-HTML-format/qaq-p/1750152

PR #5432, opened 2021-09-23
feat: enable use of smtp endpoints for sending mail

radicand:
This code enables use of a SMTP gateway for sending email, avoiding the need for a SaaS email product where not necessary.


jordanh:
@radicand thank you so much for this! This has been on our minds for a bit. We'll have a look

packages/server/email/MailManagerSMTP.ts
```diff
@@ -0,0 +1,48 @@
...
+    secure: false,
+    pool: true,
+    maxConnections: 10,
+    rateLimit: 50,
+    tls:
+      process.env.MAIL_SMTP_USE_TLS === '1'
+        ? {
+            rejectUnauthorized: false,
+            ciphers:
+              process.env.MAIL_SMTP_CIPHERS ?? 'HIGH:MEDIUM:!aNULL:!eNULL:@STRENGTH:!DH:!kEDH'
```
mattkrick, line 24: could you remove the fallback value here? if the env var is missing we want to know!

mattkrick (changes requested):
thanks for the PR @radicand!
so glad the codebase was easy enough to dig into.
would love to hear more about how you're using parabol!

packages/server/email/MailManagerSMTP.ts
radicand: Good catch, it's removed now.

PR #5464, opened 2021-09-29
Removed SFU and Cypress packages

### LINKED ISSUE #5261: Remove sfu and cypress pacakges from the monorepo
Now that we have ironbank stuff to contend with, it helps to keep package dependencies to a minimum.
since we're moving to playwright & not running any cypress tests, that should be safe to remove 
https://github.com/ParabolInc/parabol/issues/4431

since we're not longer focused on sync video, we should remove that too.

I suggest we keep those 2 in a long-lived git branch so when we decide to energize those areas we can pull from them easily

AC
- [ ] a new branch is created that is called sfu-cypress-checkpoint & that branch is pushed to github
- [ ] sfu and cypress are removed from the repo (except for the branch above)

Estimate: 4 hours

BartoszJarocki:
Fixes #5261


BartoszJarocki:
There's a bunch of `Video` related stuff in the frontend, @mattkrick do we want to remove it all?

BartoszJarocki:
@tiffanyhan do you mind taking a look? 🙏

tiffanyhan:
looks good so far!

in `.circleci/config.yml`, we still have a "Cypress Tests" job which looks like it will fail in a complete CI run since it's referencing files in the cypress package. it doesn't actually run any tests, only migrations. i think we can just get rid of references there.

i think we can also get rid of `scripts/webpack/test.cypress.config.js`, `packages/server/database/cloneTestToCypress.js`, and references to cypress in `.gitignore`. and `README.md`.

tiffanyhan (changes requested):
looks good so far!

in `.circleci/config.yml`, we still have a "Cypress Tests" job which looks like it will fail in a complete CI run since it's referencing files in the cypress package. it doesn't actually run any tests, only migrations. i think we can just get rid of references there.

i think we can also get rid of `scripts/webpack/test.cypress.config.js`, `packages/server/database/cloneTestToCypress.js`, and references to cypress in `.gitignore`. and `README.md`.

BartoszJarocki:
@tiffanyhan thank you, great finds! I've updated the PR with the requested changes and also removed few more cypress references from the code. I'd really appreciate it if you could take a look again, thank you!

tiffanyhan (approved):
lgtm! tested the client demo works well locally

mattkrick:
LGTM!
we still got a handful of `data-cy` things on HTML Elements, but they don't hurt anything & we might reuse them when we rebuild in playwright

Issue #5184, opened 2021-07-19
Spotlight: support searchQuery business logic for similarReflections


mattkrick:
Predecessors: #5183

Now that similarReflectionGroups works if the searchQuery is empty, let's make it work for a search query.

AC
- [ ] if the searchQuery is not an empty string, then that search query searches all the plaintextContent of all the reflections in the meeting
- [ ] a list of unique reflectionGroups from all the filtered reflections are returned

Estimate: 4 hours

PR #5469, opened 2021-09-29
Spotlight searchquery case

tianrunhe:
Implements #5184

Tests:
- [ ] Typing in the search bar triggers results in `SpotlightGroups` changed
- [ ] Results are filtered based on the search string
- [ ] Empty search string returns the null case for `SpotlightGroups` results
- [ ] Search also works when there are formed groups in `SpotlightGroups`
- [ ] Escape key unblur the search bar when typing, another hit closes the modal

@nickoferrall I implemented the structure Matt suggested and we discussed in this [thread](https://parabol.slack.com/archives/C836NA350/p1634163237466600) and I also borrowed some of your code from the SpotlightGroup UI to handle the `'network-only'` fetch policy. I would imagine this could cause some merge pain for your work, so apologize for that in advance!


nickoferrall:
Was the ungrouping bug introduced in this PR or did it exist in the Null Case branch? I _think_ the bug you've noticed is one I've fixed in the [Groups UI PR here](https://github.com/ParabolInc/parabol/pull/5402/files#r710244856). If you open the Spotlight, group/ungroup results, close the Spotlight, click on the same reflection to open the Spotlight, Relay uses the cached query as the variables are the same even though the results could be different. The fix in Groups UI uses the `network-only` fetch policy to make sure the results are always up to date.

packages/server/graphql/types/User.ts
```diff
@@ -459,6 +459,21 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
         if (!viewerMeetingMember) {
           return standardError(new Error('Not on team'), {userId})
         }
+
+        if (searchQuery.trim() !== '') {
+          const matchedReflections = reflections.filter(({plaintextContent}) =>
+            plaintextContent.toLowerCase().match(searchQuery.toLowerCase())
```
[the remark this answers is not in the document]
tianrunhe: I was trying to match the search bar behavior on team & task dashboard but I agree we could just start simple for now.

packages/client/hooks/useQueryLoaderNow.tsx
```diff
@@ -22,7 +23,7 @@ const useQueryLoaderNow = <TQuery extends OperationType>(
   // refetch when reconnected to server
   useEffect(() => {
     const refresh = () => {
-      loadQuery(varRef.current, {fetchPolicy: 'network-only'})
+      loadQuery(variables || {}, {fetchPolicy: fetchPolicy || 'store-or-network'})
```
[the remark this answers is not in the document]
tianrunhe: I merged in [Nick's code from the Spotlight Group UI](https://github.com/ParabolInc/parabol/pull/5402/files#diff-e8932a312141d7e2661d9f6c20dfc5b6246cf0bbd021481562117ce5f51dbcac) for this one and it's used to fix [this issue](https://github.com/ParabolInc/parabol/pull/5469#issuecomment-931477067).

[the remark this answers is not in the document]
mattkrick: look again, the link to Nick's code changes the hook above

packages/client/hooks/useQueryLoaderNow.tsx
```diff
@@ -1,23 +1,48 @@
...
 ) => {
   const [queryRef, loadQuery] = useQueryLoader<TQuery>(preloadableRequest)
   const varRef = useRef(variables)
   const atmosphere = useAtmosphere()
+  const environment = useRelayEnvironment()
   if (!areEqual(variables, varRef.current)) {
     varRef.current = variables
   }
+
+  const refreshQuery = useCallback(() => {
```
mattkrick, line 27: can you give some more detail here? if i set preventSuspense to false what bad things happen? any steps to reproduce the broken behavior?

packages/client/hooks/useSpotlightSimulatedDrag.tsx
```diff
@@ -54,6 +54,10 @@ const useSpotlightSimulatedDrag = (
       const meetingProxy = store.get(meetingId)
       meetingProxy?.setValue(null, 'spotlightGroup')
       meetingProxy?.setValue(null, 'spotlightReflectionId')
+      meetingProxy?.setValue(null, 'spotlightSearchQuery')
+      const reflection = store.get(reflectionId)
+      // set isDropping to true so that the source is added back to its original position in kanban
+      reflection?.setValue(true, 'isDropping')
```
mattkrick, line 60: i'm pretty sure this line was removed @nickoferrall can you confirm if we can remove this?

packages/server/graphql/types/User.ts
```diff
@@ -461,6 +467,23 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
         if (!viewerMeetingMember) {
           return standardError(new Error('Not on team'), {userId})
         }
+
+        if (searchQuery !== '') {
+          const matchedReflections = reflections.filter(({plaintextContent}) =>
+            plaintextContent.toLowerCase().includes(searchQuery.toLowerCase())
```
mattkrick, line 473: searchQuery is already lowercase

yarn.lock
```diff
@@ -3834,7 +3841,12 @@
     "@types/prop-types" "*"
     csstype "^2.2.0"
 
-"@types/relay-runtime@*", "@types/relay-runtime@^11.0.4":
+"@types/relay-runtime@*":
+  version "11.0.3"
```
mattkrick, line 3845: looks like a regression in the lockfile. can remove from this PR?

mattkrick (changes requested):

packages/client/hooks/useQueryLoaderNow.tsx
tianrunhe: I merged in @nickoferrall 's code from [here](https://github.com/ParabolInc/parabol/pull/5650#discussion_r747732466)

packages/client/hooks/useSpotlightSimulatedDrag.tsx
tianrunhe: Yeah I think it's removed. I overlooked it in conflict resolution.

packages/client/hooks/useQueryLoaderNow.tsx
```diff
@@ -1,23 +1,48 @@
...
   preloadableRequest: GraphQLTaggedNode | PreloadableConcreteRequest<TQuery>,
   variables: VariablesOf<TQuery> = {},
-  fetchPolicy?: PreloadFetchPolicy
+  fetchPolicy?: PreloadFetchPolicy,
+  preventSuspense?: boolean
 ) => {
   const [queryRef, loadQuery] = useQueryLoader<TQuery>(preloadableRequest)
   const varRef = useRef(variables)
   const atmosphere = useAtmosphere()
+  const environment = useRelayEnvironment()
```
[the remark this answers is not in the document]
tianrunhe: Yeah, why not? @nickoferrall 🤣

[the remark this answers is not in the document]
nickoferrall: Yeah, we should be able to use atmosphere. I quickly pushed this update at the end of my relay turn as a work in progress to fix the flickering bug

packages/client/hooks/useQueryLoaderNow.tsx
nickoferrall: This prevented the flickering bug shown in the gif in this comment: https://github.com/ParabolInc/parabol/pull/5469#pullrequestreview-782585609

To test:

- set preventSuspense to false
- type in the search query
- notice that after each keystroke, the modal flashes

This is because the Spotlight Results suspend as the query executes. I followed [this pattern](https://relay.dev/docs/next/guided-tour/refetching/refreshing-queries/#if-you-need-to-avoid-suspense) from the Relay docs to prevent this from happening.

packages/client/hooks/useSpotlightSimulatedDrag.tsx
nickoferrall: Yep, should be removed

packages/client/hooks/useQueryLoaderNow.tsx
mattkrick: ohhhhhh! i get it. it's like a workaround for `useTransition`. neat!

mattkrick (changes requested):

packages/client/components/SpotlightSearchBar.tsx
```diff
@@ -0,0 +1,117 @@
...
+          onKeyDown={onKeyDown}
+          autoFocus
+          autoComplete='off'
+          name='search'
+          placeholder='Or search for keywords...'
+          type='text'
+          spellCheck={true}
+          onChange={onChange}
+          ref={inputRef}
+          value={spotlightSearchQuery ?? ""}
```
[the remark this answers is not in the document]
tianrunhe: Because of the `husky` issue (Igor fixes it in #5702), the pre-commit hooks did not run. I manually reformatted it

packages/server/graphql/types/User.ts
```diff
@@ -497,10 +511,7 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
             }
           }
         }
-        return r
-          .table('RetroReflectionGroup')
-          .getAll(r.args(Array.from(currentResultGroupIds)), {index: 'id'})
-          .run()
+        return dataLoader.get('retroReflectionGroups').loadMany(Array.from(currentResultGroupIds))
```
mattkrick, line 514: loadMany could return an error. i made a helper function to aid with this.  see https://github.com/ParabolInc/parabol/blob/9c27f5716901081c321d69ea6f56fb4756a83c1d/packages/server/graphql/isValid.ts

packages/server/graphql/types/User.ts
```diff
@@ -448,19 +454,27 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
...
+          const matchedReflections = reflections.filter(({plaintextContent}) =>
+            plaintextContent.toLowerCase().includes(searchQuery)
+          )
+          const relatedReflections = matchedReflections.filter(
+            ({reflectionGroupId: groupId}: Reflection) => groupId != reflectionGroupId
+          )
+          const relatedGroupIds = [
+            ...new Set(relatedReflections.map(({reflectionGroupId}) => reflectionGroupId))
+          ].slice(0, MAX_RESULT_GROUP_SIZE)
+          return dataLoader.get('retroReflectionGroups').loadMany(relatedGroupIds)
```
mattkrick, line 475: filter out errors here too

mattkrick (changes requested):
1 very minor thing.
feel free to merge after that!

PR #5475, opened 2021-09-30
add hardDeleteUser mutation

### LINKED ISSUE #5035: hardDeleteUser mutation created
## Issue - Enhancement

Today we have the `deleteUser` mutation which marks the user as deleted in our DB but does not remove the record from DB. We need a mutation to "hard" delete a user.

With this mutation exist, we can also better fulfill user privacy requirement in case there are requests for removing the data from a user.

A private mutation exist, which takes `userId` or `email` and a reason text. The mutation will:
- [x] Hard delete the user record from `User` Table
- [x] Hard delete the user record from `TeamMember` table
- [x] Hard delete the user record from `MeetingMember` table
- [x] Hard delete the records in `AtlassianAuth` table with reference to the `userId`
- [x] Hard delete the records in `Notification` table with reference to the `userId`
- [x] Hard delete the user records from `OrganizationUser`
- [x] Hard delete the records in `SuggestedAction` with reference to the `userId`
- [x] Hard delete the records in `Task` with reference to the `userId`
- [x] Hard delete the records in `TimelineEvent` with reference to the `userId`
- [x] Hard delete the records in `AgendaItem` with reference to the `teamMemberId` that takes the form of userId:teamId for every team the user was on
- [x] hard delete Comment where createdBy == userId
- [x] if a meeting has a facilitatorUserId == userId AND another user is still on the team, then change facilitatorUserId to the other team member's userId
- [x] delete PushInvitation where userId === userId
- [x] delete RetroReflection where creatorId === userId
- [x] delete SlackNotification where userId === userId
- [x] delete TeamInvitation where invitedBy === userId

Note: this doesn't solve for Meeting stages, but that's OK for now.

tiffanyhan:
Resolves #5035

UI tests:
Create a new user and invite them to your team, then
- [ ] create a check in meeting, retro meeting, and sprint poker meeting. run thru each meeting and with the new user, create tasks, comment, estimate issues, etc. don't end any of the meetings yet.
- [ ] go to admin graphql and do `hardDeleteUser` mutation while passing in the new user email. should be able to keep going thru each open meeting without crashing, even though you might have to refresh to get some updates.
- [ ] for check-in meetings: user's icebreaker and check in stages should be deleted, agenda items are removed.
- [ ] for retro meetings: user's icebreaker stages should be deleted, reflections and votes remain since they're not PII in the UI.
- [ ] for sprint poker meetings: user's icebreaker stages should be deleted, estimation scores and tasks scoped by users will be removed.
- [ ] in general, tasks created by user will be deleted. tasks created by another user but assigned to the deleted user will be assigned to another user or unassigned. comments will be anonymous.
- [ ] close each meeting. you should be able to "See summary" and "See the discussion" on the timeline without the UI crashing.
- [ ] repeat the above steps, but this time end all meetings before deleting the user. everything should work as outlined above.
- [ ] should be able to sign up for a brand new account again with the previously used email for deleted user


tiffanyhan:
@nickoferrall since Bruce is out this week, do you mind giving this one a review please 🙏?

nickoferrall:
Sure thing 👍  I started reviewing it but ran into a permission error when running the backupOrg mutation. I've added a comment in Slack [here](https://parabol.slack.com/archives/C836NA350/p1633393474310100?thread_ts=1632857249.260300&cid=C836NA350) and will get back to reviewing once that's resolved

packages/server/graphql/intranetSchema/mutations/hardDeleteUser.ts
```diff
@@ -0,0 +1,313 @@
...
+        .table('MeetingMember')
+        .getAll(userIdToDelete, {index: 'userId'})
+        .getField('meetingId')
+        .coerceTo('array')
+        .distinct()
+        .run()
+    ])
+    const orgIds = orgUsers.map((orgUser) => orgUser.orgId)
+
+    const teamIds: string[] = []
```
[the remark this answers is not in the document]
tiffanyhan: great call out. i was previously doing something other side effect in the for loop and didn't want to loop twice. even i kept that, it's probably much more readable to do it declaratively.

packages/server/graphql/intranetSchema/mutations/hardDeleteUser.ts
```diff
@@ -0,0 +1,313 @@
...
+      return {error: {message: 'Provide userId XOR email'}}
+    }
+    if (!userId && !email) {
+      return {error: {message: 'Provide a userId or email'}}
+    }
+    const r = await getRethink()
+    const pg = getPg()
+
+    const user = userId
+      ? (await getUserById(userId))?.[0]
```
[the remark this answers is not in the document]
tiffanyhan: yikes thanks for catching this. the api for the user getter changed a bit after pulling from master and merging, and i subsequently only was testing the mutation with the email getter 🤦🏻‍♀️

packages/server/graphql/intranetSchema/mutations/hardDeleteUser.ts
```diff
@@ -0,0 +1,313 @@
...
+        }))
+        .filter(r.row.hasFields('otherTeamMember'))
+        .pluck('id', 'otherTeamMember')
+        .run(),
+      pg.query(`SELECT "id" FROM "Discussion" WHERE "teamId" = ANY ($1);`, [teamIds])
+    ])
+    const discussionIds = discussions.rows.map(({id}) => id)
+
+    // soft delete first for side effects, then completely remove user dependent stages
+    await softDeleteUser(user, dataLoader, reason, true)
```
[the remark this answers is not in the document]
tiffanyhan: great suggestion, i like that so much more

packages/server/graphql/intranetSchema/mutations/hardDeleteUser.ts
```diff
@@ -0,0 +1,313 @@
...
+
+    for (const teamMemberId of teamMemberIds) {
+      const teamId = teamMemberId.split('::')[1]
+      const filterFn = (stage: CheckInStage | UpdatesStage | EstimateStage) =>
+        (stage as CheckInStage | UpdatesStage).teamMemberId === teamMemberId ||
+        (stage as EstimateStage).creatorUserId === userIdToDelete
+      await removeStagesFromMeetings(filterFn, teamId, dataLoader, true)
+      await removeUserFromMeetingStages(userIdToDelete, teamId, dataLoader, true)
+    }
+    for (const agendaItemId of currentAgendaItemIds) {
```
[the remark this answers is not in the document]
tiffanyhan: cool that makes sense. i usually try to err on the side of not changing existing functions, since they might be written that way for a specific reason i don't know about. moving the logic into `softDeleteUser` definitely simplifies things!

packages/server/graphql/intranetSchema/mutations/hardDeleteUser.ts
```diff
@@ -0,0 +1,313 @@
...
+        undefined,
+        {agendaItemId},
+        {authToken, dataLoader, socketId: undefined}
+      )
+    }
+    for (const agendaItemId of pastAgendaItemIds) {
+      // id is of format 'teamId::randomId'
+      const [teamId] = agendaItemId.split('::')
+      const filterFn = (stage: AgendaItemsStage) => stage.agendaItemId === agendaItemId
+      await removeStagesFromMeetings(filterFn, teamId, dataLoader, true)
```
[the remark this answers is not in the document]
tiffanyhan: i was trying to publish the active agenda items that were deleted to other team members, which isn't necessary for inactive agenda items. however, after testing further, it doesn't look like not publishing has any negative effects so i think we can just do without it!

packages/server/graphql/mutations/removeAgendaItem.ts
```diff
@@ -17,36 +11,5 @@ export default {
...
-    if (!agendaItem) {
-      return standardError(new Error('Agenda item not found'), {userId: viewerId})
-    }
-    const filterFn = (stage: AgendaItemsStage) => stage.agendaItemId === agendaItemId
-    await removeStagesFromMeetings(filterFn, teamId, dataLoader)
-    const data = {agendaItem, meetingId: agendaItem.meetingId}
-    publish(SubscriptionChannel.TEAM, teamId, 'RemoveAgendaItemPayload', data, subOptions)
-    return data
-  }
+  resolve: removeAgendaItemResolver
```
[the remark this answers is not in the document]
tiffanyhan: refactored so this should no longer be necessary

packages/server/graphql/intranetSchema/mutations/hardDeleteUser.ts
```diff
@@ -0,0 +1,313 @@
...
+      createdByTeamInvitations: r
+        .table('TeamInvitation')
+        .getAll(r.args(teamIds), {index: 'teamId'})
+        .filter((row) => row('acceptedBy').eq(r(userIdToDelete)))
+        .update({acceptedBy: ''}),
+      comment: r
+        .table('Comment')
+        .getAll(r.args(discussionIds), {index: 'discussionId'})
+        .filter((row) => row('createdBy').eq(r(userIdToDelete)))
+        .update({createdBy: ''}), // becomes anonymous in UI
```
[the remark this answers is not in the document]
tiffanyhan: great idea

tiffanyhan:
re-requesting review, thanks for the awesome suggestions @mattkrick

one thing i noticed is that after the deleted user gets kicked out of a meeting, we call `SEND TO SENTRY Error: No mutator id in setAppLocation.ts` server side. any suggestions on how to mitigate it?

packages/server/graphql/mutations/deleteUser.ts
```diff
@@ -1,16 +1,27 @@
...
-import getDeletedEmail from '../../utils/getDeletedEmail'
+import softDeleteUserResolver from './helpers/softDeleteUser'
 import {getUserById} from '../../postgres/queries/getUsersByIds'
 import {getUserByEmail} from '../../postgres/queries/getUsersByEmails'
+import db from '../../db'
+import getDeletedEmail from '../../utils/getDeletedEmail'
+import updateUser from '../../postgres/queries/updateUser'
+import {USER_REASON_REMOVED_LIMIT} from '../../postgres/constants'
+
+const delay = (ms) => new Promise((resolve) => setTimeout(resolve, ms))
```
mattkrick, line 13: let's use https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/client/utils/sleep.ts

packages/server/graphql/intranetSchema/mutations/hardDeleteUser.ts
```diff
@@ -0,0 +1,266 @@
...
+  args: {
+    userId: {
+      type: GraphQLID,
+      description: 'a userId'
+    },
+    email: {
+      type: GraphQLID,
+      description: 'the user email'
+    },
+    reasonText: {
```
mattkrick, line 28: i think we can get rid of reasonText since the User object will get deleted unless i missed a place where it is stored?

packages/server/graphql/mutations/deleteUser.ts
```diff
@@ -43,43 +54,24 @@ export default {
...
-      isRemoved: true,
-      email: getDeletedEmail(userId),
-      reasonRemoved: validReason,
-      updatedAt: new Date()
-    }
-    setTimeout(() => {
-      db.write('User', userIdToDelete, update)
-      updateUser(update, userIdToDelete)
-    }, 30000)
+    await delay(30000)
```
mattkrick, line 72: before the long sleep was fine because nothing in our system was really impacted-- we already returned from the resolver & we were just waiting 30 seconds so integrations could still grab the email of the to-be-deleted user.

now that we have 1 function waiting on sleep, we should do this more elegantly.

what we need to do is share the following JSON object with hubspot (https://github.com/ParabolInc/parabol/blob/439d988af10717e2de9a9443c29185f2f5e6b81f/packages/server/utils/updateHubspot.ts#L108-L118)

to do that:
- pass in the referenced string as a query with the userId as the variables to graphql. it should resolve that json object
- next, we pass that object into the properties field here: https://github.com/ParabolInc/parabol/blob/9dbdda9f39b99b0db9b4bf77cbc92557187f0a4c/packages/server/graphql/mutations/helpers/softDeleteUser.ts#L46
- finally, we write a custom conditional for that event to use the data in the `properties` field instead of reaching back out to our endpoint: https://github.com/ParabolInc/parabol/blob/439d988af10717e2de9a9443c29185f2f5e6b81f/packages/server/utils/updateHubspot.ts#L384
the end result is removing the `delay(30000)` 🎉

packages/server/graphql/mutations/helpers/removeStagesFromMeetings.ts
```diff
@@ -14,12 +14,15 @@ const removeStagesFromMeetings = async (
   const now = new Date()
   const r = await getRethink()
   const activeMeetings = await dataLoader.get('activeMeetingsByTeamId').load(teamId)
+  const completedMeetings = await dataLoader.get('completedMeetingsByTeamId').load(teamId)
```
mattkrick, line 17: let's run these in parallel using promise.all

packages/server/graphql/mutations/helpers/removeStagesFromMeetings.ts
```diff
@@ -14,12 +14,15 @@ const removeStagesFromMeetings = async (
...
+  const meetings = activeMeetings.concat(completedMeetings)
+
   await Promise.all(
-    activeMeetings.map((meeting) => {
+    meetings.map((meeting) => {
       const {id: meetingId, phases} = meeting
       phases.forEach((phase) => {
         // do this inside the loop since it's mutative
-        const {facilitatorStageId} = meeting
+        let {facilitatorStageId} = meeting
```
mattkrick, line 25: let's get rid of this line & instead on L31, L35 we just pass in `meeting.facilitatorStageId`. 1 less variable to worry about 🎉

packages/server/graphql/mutations/helpers/removeStagesFromMeetings.ts
```diff
@@ -35,6 +41,7 @@ const removeStagesFromMeetings = async (
               nextStage.startAt = now
               nextStage.viewCount = nextStage.viewCount ? nextStage.viewCount + 1 : 1
               nextStage.isNavigable = true
+              facilitatorStageId = nextStage.id
```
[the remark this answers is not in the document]
mattkrick: Oh! makes sense, let's just use `meeting.facilitatorStageId` everywhere

packages/server/graphql/mutations/helpers/removeTeamMember.ts
```diff
@@ -136,11 +138,20 @@ const removeTeamMember = async (
 
   const archivedTasks = await archiveTasksForDB(integratedTasksToArchive)
   const archivedTaskIds = archivedTasks.map(({id}) => id)
+  const agendaItemIds = await r
+    .table('AgendaItem')
+    .getAll(teamId, {index: 'teamId'})
+    .filter((row) => row('teamMemberId').eq(teamMemberId))
+    .getField('id')
+    .coerceTo('array')
+    .distinct()
```
mattkrick, line 147: no need for `coerceTo` or `distinct` since the coerceTo is implied & they're UIDs

packages/server/graphql/mutations/helpers/removeUserFromMeetingStages.ts
```diff
@@ -14,8 +15,11 @@ const removeUserFromMeetingStages = async (
   const now = new Date()
   const r = await getRethink()
   const activeMeetings = await dataLoader.get('activeMeetingsByTeamId').load(teamId)
+  const completedMeetings = await dataLoader.get('completedMeetingsByTeamId').load(teamId)
```
mattkrick, line 18: let's call these in parallel

packages/server/graphql/mutations/helpers/removeUserFromMeetingStages.ts
```diff
@@ -24,11 +28,19 @@ const removeUserFromMeetingStages = async (
...
-          if (!readyToAdvance) continue
           const userIdIdx = readyToAdvance.indexOf(userId)
-          if (userIdIdx === -1) continue
-          readyToAdvance.splice(userIdIdx, 1)
-          isChanged = true
+          if (userIdIdx !== -1) {
+            readyToAdvance.splice(userIdIdx, 1)
+            isChanged = true
+          }
+          const scores = (stage as EstimateStage).scores
```
mattkrick, line 36: let's rewrite this to:

```ts
if (stage.phaseType === 'ESTIMATE') {
```

that way, we aren't doing any casting & we know that `scores` exists.

packages/server/graphql/mutations/helpers/softDeleteUser.ts
```diff
@@ -0,0 +1,51 @@
...
+  )
+  const teamMemberIds = await r
+    .table('TeamMember')
+    .getAll(userIdToDelete, {index: 'userId'})
+    .getField('id')
+    .coerceTo('array')
+    .run()
+  const teamIds = teamMemberIds.map((id) => TeamMemberId.split(id).teamId)
+
+  await Promise.all([
```
mattkrick, line 36: right here is a good place to do the `Account Removed` query on our own graphql server so we can pass that value into the segmentIo.track properties field.

mattkrick:
Loom: https://www.loom.com/share/ea18990041394edf985131fa6b46944e

mattkrick (changes requested):

packages/server/graphql/intranetSchema/mutations/hardDeleteUser.ts
tiffanyhan: the `reasonText` is passed to `softDeleteUser` resolver which passes it to `segment.track` method

packages/server/utils/updateHubspot.ts
```diff
@@ -390,6 +367,12 @@ const updateHubspot = async (event: string, user: IUser, properties: BulkRecord)
     // wait for hubspot to associate the contact with the company, fn must run in 5 seconds
     await sleep(5000)
     await updateHubspotCompany(email, company)
+  } else if (event === 'Account Removed') {
+    const {parabolPayload} = properties
+    if (!parabolPayload) return
+    const {user} = (parabolPayload as unknown) as {[key: string]: any}
```
tiffanyhan, line 373: i know this is a bit hacky but wasn't sure if `BulkRecord` type should be changed, since this is a special case we're doing

tiffanyhan:
@mattkrick changes have been made, thanks so much for the loom videos. would be great if you could take another look at this 🙏🏽

packages/server/graphql/mutations/deleteUser.ts
```diff
@@ -43,43 +52,20 @@ export default {
...
-      reasonRemoved: validReason,
-      updatedAt: new Date()
-    }
-    setTimeout(() => {
-      db.write('User', userIdToDelete, update)
-      updateUser(update, userIdToDelete)
-    }, 30000)
+    const {id: userIdToDelete} = user
+
+    await softDeleteUserResolver(userIdToDelete, dataLoader, validReason)
```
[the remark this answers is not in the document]
mattkrick: make this fn return `getDeletedEmail(userIdToDelete)` and then pass that into the function below as well as up above in the hard delete mutation resolver

that way getDeletedEmail is only called once & the 2 functions are not tightly coupled to use the same function

packages/server/graphql/mutations/helpers/softDeleteUser.ts
```diff
@@ -0,0 +1,71 @@
...
+  const teamMemberIds = await r
+    .table('TeamMember')
+    .getAll(userIdToDelete, {index: 'userId'})
+    .getField('id')
+    .coerceTo('array')
+    .run()
+  const teamIds = teamMemberIds.map((id) => TeamMemberId.split(id).teamId)
+
+  const [parabolPayload] = await Promise.all([
+    parabolFetch(
```
mattkrick, line 38: reusing the hubspot helper function isn't great. it doesn't reuse the dataloader or auth token & the error message isn't correct.

let's just call executeGraphQL like https://github.com/ParabolInc/parabol/blob/c2112010eae3cc48a47654ff467515401d722976/packages/server/email/ServerEnvironment.ts#L37-L42

packages/server/utils/updateHubspot.ts
mattkrick: let's just define the shape of `parabolPayload` above with all the fields we expect instead of `any`

packages/server/graphql/intranetSchema/mutations/hardDeleteUser.ts
```diff
@@ -0,0 +1,266 @@
...
+        .delete(),
+      organizationUser: r
+        .table('OrganizationUser')
+        .getAll(userIdToDelete, {index: 'userId'})
+        .delete(),
+      suggestedAction: r
+        .table('SuggestedAction')
+        .getAll(userIdToDelete, {index: 'userId'})
+        .delete(),
+      assignedTasks: r
```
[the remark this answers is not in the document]
mattkrick: oh! let's fix that so it deletes all tasks on a 1-person team. we can do that here
https://github.com/ParabolInc/parabol/blob/0ab0a69f8226311a0e8b2b30599cf7781738b4b5/packages/server/graphql/mutations/helpers/removeTeamMember.ts#L44

packages/server/graphql/intranetSchema/mutations/hardDeleteUser.ts
```diff
@@ -0,0 +1,266 @@
...
+        }))
+        .filter(r.row.hasFields('otherTeamMember'))
+        .pluck('id', 'otherTeamMember')
+        .run(),
+      pg.query(`SELECT "id" FROM "Discussion" WHERE "teamId" = ANY ($1);`, [teamIds])
+    ])
+    const teamDiscussionIds = discussions.rows.map(({id}) => id)
+
+    // soft delete first for side effects
+    await softDeleteUser(userIdToDelete, dataLoader, reason)
```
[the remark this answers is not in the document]
mattkrick: see other comment on how to fix

mattkrick (changes requested):

packages/server/graphql/mutations/helpers/removeTeamMember.ts
```diff
@@ -36,56 +41,85 @@ const removeTeamMember = async (
...
+        .table('Team')
+        .get(teamId)
+        .update(updates)
+        .run(),
+      updateTeamByTeamId(updates, teamId),
+      // delete all tasks belonging to a 1-person team
+      r
+        .table('Task')
+        .getAll(teamId, {index: 'teamId'})
+        .delete()
```
tiffanyhan, line 60: moved deleting tasks on 1-person team here

PR #5479, opened 2021-10-04
fix: prevent closing spotlight when closing expanded reflection stack

BartoszJarocki:


BartoszJarocki:
@nickoferrall I'd appreciate if you could take a look, thanks!

nickoferrall:
Nice work, LGTM!

packages/client/hooks/useModal.ts
```diff
@@ -6,17 +6,19 @@ import usePortal, {UsePortalOptions} from './usePortal'
 interface Options extends UsePortalOptions {
   background?: string
   loadingWidth?: number
+  noClosePortal?: boolean
```
mattkrick, line 9: what's the difference between noClosePortal & noClose?

BartoszJarocki:
@mattkrick I've fixed conflicts and updated the solution according to your suggestion, could you please take a look? 🙏

mattkrick:
LGTM! this should definitely help us out in future portal works

PR #5481, opened 2021-10-05
Add GraphQL server tests to circle ci

Dschoordsch:
This PR automatically runs the server `jest` tests as part of the CI run. See inline comments for details.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,6 +27,7 @@ jobs:
       S3_DB_BACKUPS_BUCKET: 'db-backups.parabol.co'
       SERVER_ID: '0'
     working_directory: ~/action
+    resource_class: medium+
     steps:
       - add_ssh_keys:
           fingerprints:
@@ -60,15 +61,28 @@ jobs:
           command: |
             git clone $DEVOPS_REPO $DEVOPS_WORKDIR
       - run:
-          name: GraphQL server tests
+          name: Start dev server
           command: |
             cp $DEVOPS_WORKDIR/environments/cypress .env
             echo SOCKET_PORT=3001 >> .env
-            yarn dev &
-            DEV_PID=$!
-            curl -4 --retry 30 --retry-connrefused --retry-delay 10 http://localhost:3000/graphql
-            yarn workspace parabol-server test --forceExit
-            kill $DEV_PID
+            yarn dev || true
+          background: true
+      - run:
+          name: Wait for dev server to be healthy
+          command: curl -4 --retry 30 --retry-connrefused --retry-delay 10 http://localhost:3000/graphql
+      - run:
+          name: Run server tests
+          command: yarn workspace parabol-server test --runInBand --reporters=default --reporters=jest-junit
+          environment:
+            JEST_JUNIT_OUTPUT_DIR: ./reports/junit
+      - store_test_results:
+          path: ./packages/server/reports/junit/
+      - store_artifacts:
+          path: ./packages/server/reports/junit
+      - run:
+          name: Kill server
+          command: pkill node
+          when: always
       - run:
           name: Build Assets
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,7 +11,7 @@ jobs:
           POSTGRES_DB: 'web'
           POSTGRES_HOST: 'localhost'
           POSTGRES_PORT: '5432'
-      - image: redis:3.2.8
+      - image: redis:6.2.6-alpine
         environment:
           TERM: xterm
     environment:
@@ -59,6 +59,16 @@ jobs:
           name: DevOps checkout
           command: |
             git clone $DEVOPS_REPO $DEVOPS_WORKDIR
+      - run:
+          name: GraphQL server tests
+          command: |
+            cp $DEVOPS_WORKDIR/environments/cypress .env
+            echo SOCKET_PORT=3001 >> .env
+            yarn dev &
+            DEV_PID=$!
+            curl -4 --retry 30 --retry-connrefused --retry-delay 10 http://localhost:3000/graphql
+            yarn workspace parabol-server test --forceExit
+            kill $DEV_PID
       - run:
           name: Build Assets
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,10 +61,6 @@ jobs:
           name: DevOps checkout
           command: |
             git clone $DEVOPS_REPO $DEVOPS_WORKDIR
-
-            # TODO: remove this after merging the linked upstream PR
-            cd "$DEVOPS_WORKDIR"
-            git checkout feature/create-test-env
       - run:
           name: Start dev server
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,32 +6,33 @@ jobs:
       - image: rethinkdb:2.4.0
       - image: circleci/postgres:12.3
         environment:
-          POSTGRES_PASSWORD: 'EcLtsBMRtra33jqUCQWtSZwt4zfGBzdUvJE74fzSxX'
-          POSTGRES_USER: 'parabol'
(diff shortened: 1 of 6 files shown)


mattkrick:
@Dschoordsch just doing some PR housekeeping, since #5330 was merged is this safe to close?

Dschoordsch:
@mattkrick thanks for checking. I'm still working on it from time to time. Since it needs running on CircleCI this is the easiest way for me to test.

blimmer:
I did some digging into this today and I think I understand a bit more about what's going on.

Previously, you were running `yarn dev` in the background with `&`. `pm2` seems to proxy the logs through to the console even when it's running in the background (I was able to repro that locally as well). This makes it a bit tough to parse the output of `jest` and the other commands you're trying to run. So, to work around that, I tried using the `background: true` [config property](https://circleci.com/docs/2.0/configuration-reference/#background-commands). This leaves the task running and still allows you to see the logs, which is kinda handy.

With that step split out into a `background` command, I was able to more easily see that the `jest` command is never exiting:

Screen_Shot_2022-01-25_at_5_43_43_PM

I would expect to see a summary, even if `jest` exited with failures, like I see locally:

Screen Shot 2022-01-25 at 5 45 05 PM

This is why you were never seeing the server shut down before since `jest` was hanging.

Either way, it appears that the server is continuously crashing in CI, which is also causing your `jest` tests to fail. From the logs:

Screen_Shot_2022-01-25_at_5_46_42_PM

The jest tests appear to be receiving some Error message (instead of JSON):

Screen Shot 2022-01-25 at 5 48 35 PM

So, I think the next step is to try to get at the application logs to understand why it's crashing in a loop. I think that'll fix up the rest of the problems. I'm out of time for this week, but I'd be happy to have a look next week if you don't have the bandwidth. CircleCI's [SSH debugging feature](https://circleci.com/docs/2.0/ssh-access-jobs/) might be useful here.

.circleci/config.yml
```diff
@@ -27,6 +27,7 @@ jobs:
       S3_DB_BACKUPS_BUCKET: 'db-backups.parabol.co'
       SERVER_ID: '0'
     working_directory: ~/action
+    resource_class: medium+
```
blimmer, line 30: CPU and memory usage with the default `medium` executor was hitting 100%:

Screen Shot 2022-02-01 at 11 44 22 AM

This will eventually cause flaky/failed builds. So, to avoid this, I bumped up to the [medium+ resource class](https://circleci.com/docs/2.0/configuration-reference/#docker-executor). With this resource runner, we have a bit more breathing room:

Screen Shot 2022-02-01 at 11 46 00 AM

Pricing wise, this will increase cost from 10 credits/min to 15 credits/min ([docs](https://circleci.com/product/features/resource-classes/#docker)).

With your current free plan with CircleCI, you get 30,000 credits per month. Reviewing [past usage](https://app.circleci.com/settings/plan/github/ParabolInc/usage) it looks like you haven't used more than 14k in the last few months, so I believe this is a safe trade-off to make.

Alternatively, we can stick with `medium` for the time being and bump if we start running into transient failures.

.circleci/config.yml
```diff
@@ -59,6 +60,29 @@ jobs:
...
+      - run:
+          name: Start dev server
+          command: |
+            cp $DEVOPS_WORKDIR/environments/cypress .env
+            echo SOCKET_PORT=3001 >> .env
+            yarn dev || true
+          background: true
+      - run:
+          name: Wait for dev server to be healthy
+          command: curl -4 --retry 30 --retry-connrefused --retry-delay 10 http://localhost:3000/graphql
```
blimmer, line 72: This `background: true` [config property](https://circleci.com/docs/2.0/configuration-reference/#background-commands) allows the server task to keep running while continuing to the next steps. The nice part about this is that the `jest` output isn't co-mingled with the PM2 output, which I found to be more usable.

.circleci/config.yml
```diff
@@ -59,6 +60,29 @@ jobs:
...
+      - run:
+          name: Run server tests
+          command: yarn workspace parabol-server test --runInBand --reporters=default --reporters=jest-junit
+          environment:
+            JEST_JUNIT_OUTPUT_DIR: ./reports/junit
+      - store_test_results:
+          path: ./packages/server/reports/junit/
+      - store_artifacts:
+          path: ./packages/server/reports/junit
```
blimmer, line 81: This configuration adds `jest` reporting to the CircleCI UI:

Screen Shot 2022-02-01 at 11 52 01 AM

This provides a few distinct features:

1. Instead of digging through log output, failing tests are easy to find via the UI.
2. CircleCI will monitor tests that fail via their [test insights](https://circleci.com/docs/2.0/insights-tests/) feature. This helps identify flaky tests and more.

Dschoordsch: +2 Nice!

Dschoordsch (commented):
Thank you so much for getting this working!

.circleci/config.yml
mattkrick: oh this is awesome!

BartoszJarocki (approved):
awesome work! one question, can we add tests as a required check in github? it'd prevent us from pushing code which breaks the existing tests and in general, keep them up to date

Dschoordsch:
> awesome work! one question, can we add tests as a required check in github? it'd prevent us from pushing code which breaks the existing tests and in general, keep them up to date

Once merged, this will be blocking, it was ran already for this PR: https://app.circleci.com/pipelines/github/ParabolInc/parabol/4258/workflows/821fd194-aeae-4c77-a3ae-a9c4a169f627/jobs/13063

Dschoordsch:
@adaniels-parabol could you do maintainer review?

.circleci/config.yml
```diff
@@ -59,6 +60,29 @@ jobs:
           name: DevOps checkout
           command: |
             git clone $DEVOPS_REPO $DEVOPS_WORKDIR
+      - run:
+          name: Start dev server
+          command: |
+            cp $DEVOPS_WORKDIR/environments/cypress .env
```
tianrunhe, line 66: We might need `$DEVOPS_WORKDIR/environments/local` here instead. The cypress configuration has not been updated for a while

blimmer: Good call on using something different. I just browsed the `local` config, and I think it might provide too many "real" connections for a testing environment. I'll open a PR over there to create a test env.

packages/server/__tests__/jiraServerIntegration.test.ts
```diff
@@ -51,7 +51,7 @@ test('Add integration provider', async () => {
         oAuth1ProviderMetadataInput: {
           serverBaseUrl,
           consumerKey,
-          privateKey
+          consumerSecret
```
blimmer, line 54: cc @Dschoordsch - this was broken upstream

packages/server/email/getMailManager.ts
```diff
@@ -9,7 +9,8 @@ let mailManager: MailManager
 const managers = {
   mailgun: MailManagerMailgun,
   google: MailManagerGoogle,
-  smtp: MailManagerSMTP
+  smtp: MailManagerSMTP,
+  debug: MailManagerDebug
```
blimmer, line 13: Previously, tests were using the real MailGun instance. Now the `test` env uses the debug mailer.

packages/server/__tests__/jiraServerIntegration.test.ts
Dschoordsch: Yeah, that's me wo broke it

blimmer: No worries! Just making sure this seems correct to you.

adaniels-parabol (approved):

blimmer:
@BartoszJarocki or @Dschoordsch , if either of you have admin on this repo, you can now make the `build-and-deploy` step a "required check" for PRs against your default branch.

Issue #5272, opened 2021-08-13
Remote reflection view for Spotlight reflection


nickoferrall:
### AC:

- when startDraggingReflection has been received via subscription, set isUpdatingDrag flag to true and add a 300ms setTimeout that will turn isUpdatingDrag to false
- When updateDragLocation is received for the given startDraggingReflection event, cancel that setTimeout because the client is now receiving drag coordinates
- when isUpdatingDrag is set to false, that triggers the RemoteReflection's indeterminate mode
- When a RemoteReflection is in indeterminate mode, instead of just sitting there, it moves around in a small circle.
- Bonus points: circles are lame. Instead of making the reflection move around in a circle, make it move around in a fun shape like https://mathworld.wolfram.com/Triquetra.html

See Matt's comment in https://github.com/ParabolInc/parabol/issues/5181#issuecomment-897756589 for context

Dschoordsch:
@nickoferrall it looks like the first part is also missing?
> when the viewer clicks the reflection to open the spotlight, that reflection will behave like it's in the dragging state to non-viewers, i.e. users won't be able to drag that reflection. This is because it would be confusing if a user drags the reflection to a different group just before or after the viewer moves it in the spotlight. When the spotlight modal is open, the user is in the process of moving the reflection so it's logical that it would be in the dragging state
after 30 secs / 1 min, the spotlight should close automatically so its not in the dragging state indefinitely

I don't see any events on the remote side when opening a card.

nickoferrall:
> it looks like the first part is also missing?

What first part are you referring to here? Currently, when the viewer opens the Spotlight, the UI doesn't change for the other users. In this issue, we'll fix that by adding an animation for the remote users when the viewer opens the Spotlight.

> I don't see any events on the remote side when opening a card.

Do you mean subscription events? The Spotlight logic doesn't use any subscriptions but dragging a card does. You can tell that the Spotlight is open because the retro meeting contains a `spotlightGroup`.

PR #5500, opened 2021-10-08
Show remote spotlight activity

Dschoordsch:
When spotlight is opened, `startDragging` is triggered with `isSpotlight=true` to the remote end. Remote then starts an animation without collapsing the place where the card sits. To not time out on the remote end, regular updates are sent.

### Testing preparation
* set `showSpotlight = true` in [ReflectionCard.tsx:233](https://github.com/ParabolInc/parabol/blob/feat/5272/show_remote_spotlight_activity/packages/client/components/ReflectionCard/ReflectionCard.tsx#L233)
* with >= 2 participants join a retrospective group session, one is local, the other the remote end
### Testing steps
* [ ] move cards around, group them, ungroup them and observe if it's still reflected correctly on remote end
* [ ] open spotlight on single card and on group, check that there is an animation shown on the remote end
* [ ] keep spotlight open for more than 5s and check that animation continues on the remote end
* [ ] close spotlight and check that animation stops on the remote end
* [ ] close the browser with spotlight open and check the animation stops at the remote end

Fixes: #5272

### DIFF
--- a/packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx
+++ b/packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx
@@ -107,6 +107,7 @@ const DraggableReflectionCard = (props: Props) => {
     : []
   const isReflectionGroupIdInSpotlight = reflectionGroupIdsInSpotlight.includes(reflectionGroupId)
   const {onMouseDown} = useDraggableReflectionCard(
+    meeting,
     reflection,
     drag,
     staticIdx,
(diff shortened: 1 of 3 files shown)


Dschoordsch:
> Collision logic looks good to me! While testing, I noticed that while the source reflection is animating in a circle, a user can drag & drop a reflection onto the animating source. We can prevent that from happening by checking whether the reflection group contains the source reflection and, if it does, prevent it from being droppable. See:
> 
> https://github.com/ParabolInc/parabol/blob/4e6b5fd8ab5aa3ba5dd1d984dea748af12ab857d/packages/client/components/ReflectionGroup/ReflectionGroup.tsx#L171

That is out of scope for this issue, I create #5556.

Issue #5200, opened 2021-07-22
When a task changes teams, make sure there's no conflicting integrationHash


mattkrick:
We want to guarantee a single source of truth for integrated tasks, or at least get close.
This isn't possible because multiple teams could refer to the same jira issue.

So, what if team A & team B both import jira issue C onto their respective teams to estimate story points on?
I think this should be permitted.
What if a user moves that issue from team b onto team a & now there's a dupe? what if it's archived there? or private?

i think we keep it simple.

AC
- [ ] Allow to change teams for integration tasks
- [ ] if issue B is moved to to team A & they refer to the same jira issue, then issue A gets deleted
- [ ] after A is deleted, then we continue to move B over
 
that will solve the case for dupes, archived, and private

Estimate: 4 hours

mattkrick:
nice find! i forgot switching teams was already locked down for integrated tasks. that makes me less worried about this issue.
i'll leave it up to you, we can either de-prioritize this issue & tackle it later or if you've already made progress on it you can enable switching teams for integrated tasks in the UI.

in the future we'll show both the team & the user assignment on the task card, we just haven't designed that yet.

Dschoordsch:
I only spent little time figuring out how this situation could be triggered. I think we should de-scope this and tackle it later.

jordanh:
**Story Points: 8**
  [See the discussion](https://action.parabol.co/meet/7By0K1Cj6w/estimate/2) in “Sprint `#89` Scoping”

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

Dschoordsch:
If a task B is private and moved from team B to team A, it will be private in team A regardless of whether or not the task A referring to the same integration task C was private, i.e. the visibility is reduced afterwards.

Dschoordsch:
There is a different issue when moving an integration task to another team: What do we do when the owner does not have set up the integration for the target team?
I will add a dialog: "You did not set up GitHub for Team B yet, would you like to do so now? Yes - No" which starts normal flow to add the integration.

Issue #5342, opened 2021-08-31
Integrations: create objects for other user


jordanh:
With our latest integration architecture, we lost the ability for a user who hasn't configured a team integration to use the credentials of another user to create an object. Let's get this functionality back!

AC
- [ ] A non-integrated user can use the credentials of another to create an integrated object
- [ ] Integrated tasks show correctly as long as one team member has the integration set up

EE
4 hours

jordanh:
**Story Points: 8**
  [See the discussion](https://action.parabol.co/meet/7By0K1Cj6w/estimate/5) in “Sprint `#89` Scoping”

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

PR #5513, opened 2021-10-13
Allow moving integration tasks between teams

Dschoordsch:
If a task with the same hash exists in the destination, it is deleted first.

A conflicting integration task in the target is deleted first and the delete notifications are sent out, then the original task is moved.
In order to avoid having duplicated functionality the `teamId` was removed from `updateTask` mutation, so `changeTaskTeam` is the only way left to move a task.

#### Setup
* [x] create 2 teams Team A and Team B
* [x] join both teams with users Alice and Bob
* [x] Alice sets up Jira integration for Team A and Team B
* [x] Bob sets up Jira integration for Team B only
* [x] Alice starts a Sprint Poker in Team A and adds 4 tasks in the Scope stage via Jira integration (JIRA-1 - JIRA-4)
* [x] Bob does the same in Team B with the same 4 Jira tasks
* [x] Alice goes to their Tasks, Archived and un-archive the Jira tasks and distributes them in all 4 columns
* [x] Bob goes to their Tasks, Archived and un-archives JIRA-1 - JIRA-3 and leaves them in the default column
* [x] Open the sessions for Alice and Bob in separate browsers
#### Verification
* [x] Bob opens team B dashboard
* [x] Alice opens Tasks and changes the team of JIRA-1 task from Team A to Team B
* [x] While Alice is doing this, Bob observes the old Team B JIRA-1 disappearing and the new Team A JIRA-1 appearing
* [x] Repeat for JIRA-2 till JIRA-4, the tasks must appear in the right column for Bob
* [x] The archived JIRA-4 of Bob is gone

#### Setup
* [x] Clarissa creates 2 teams: Team Integration and Team No-Integration
* [x] Clarissa adds GitHub and Jira integrations to Team Integration
* [x] Clarissa creates a new GitHub and a new Jira Task
#### Verification
* [x] in Task view Clarissa selects Team No-Integration for the GitHub Task from the team drop down
  * [x] a dialog is presented asking Clarissa to setup GitHub integration
* [x] Clarissa declines
  * [x] task is still in Team No-Integration
* [x] Clarissa repeats the process but now presses 'Add it now'
  * [x] GitHub oauth is started
  * [x] when successfull task is moved to Team Integration
* [x] Repeat verification steps with Jira

Fixes: #5200


nickoferrall:
Hey @Dschoordsch, would you mind clarifying a few things in the testing steps please:
1. Is the second client using the same user account as the first client or are they two different users?
2. By `integration tasks`, are you referring to regular Parabol tasks added in Sprint Poker, or tasks that are integrated with Jira or GitHub?
3. Does `target dashboard` refer to the `/tasks` page?
4. Should I create two Sprint Poker meetings, one for each team, and add the identical task in both meetings?

If you wouldn't mind adding a bit more detail to the testing steps, I'll review the PR asap!

Dschoordsch:
@nickoferrall Awesome find! In my test Bob had the integration set up for both Team A and Team B, then it works. Apparently we're using the integration of the assignee and team combination.
I would argue fixing this is out of scope for this ticket but should be part of AC for #5342

Dschoordsch:
@nickoferrall I mad additional changes to catch the use case when a user tries to move a task to a team without having the integration set up for it yet. I also added separate testing flow for it. Could you please review again?

packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterTeamAssigneeAddGitHubDialog.tsx
```diff
@@ -0,0 +1,117 @@
...
+    })
+  }
+
+  return (
+    <StyledDialogContainer>
+      <DialogTitle>GitHub integration for {teamName}</DialogTitle>
+      <DialogContent>
+        <div>
+          <StyledTip>
+            You do not have GitHub configured for {teamName}, do you want to add it now?
```
nickoferrall, line 87: Small copy suggestion: `You don't have GitHub configured for {teamName}. Do you want to add it now?`

packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterTeamAssigneeAddJiraDialog.tsx
```diff
@@ -0,0 +1,117 @@
...
+    })
+  }
+
+  return (
+    <StyledDialogContainer>
+      <DialogTitle>Jira integration for {teamName}</DialogTitle>
+      <DialogContent>
+        <div>
+          <StyledTip>
+            You do not have Jira configured for {teamName}, do you want to add it now?
```
nickoferrall, line 87: Same copy suggestion as above

packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterTeamAssigneeAddGitHubDialog.tsx
```diff
@@ -0,0 +1,117 @@
...
+})
+
+const ButtonGroup = styled('div')({
+  marginTop: '24px',
+  display: 'flex',
+  justifyContent: 'flex-end'
+})
+
+const StyledTip = styled('p')({
+  fontSize: 13,
```
nickoferrall, line 33: I find the text here a little bit hard to read. Perhaps fontSize 14 with lineHeight 20?

nickoferrall (approved):
That's all looking good to me! Feel free to review my comments and then proceed to Maintainer Review

packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterTeamAssigneeAddGitHubDialog.tsx
Dschoordsch: Yes, looks much better. It came from `AddMissingJiraFieldModal` so we might want to increase size there as well

Dschoordsch:
@enriquesanchez I added a dialog which appears when a user tries to move a task with integration to a team where they didn't set an integration up yet, is this dialog ok:
Screenshot from 2021-11-05 11-09-40
?

Dschoordsch:
@mattkrick could you please review?

enriquesanchez (changes requested):
Hi @Dschoordsch 

What happens if the user clicks cancel? Would the task still be moved and disconnected from the integration? Or would it abort the action altogether?

Design wise looks good! My only suggestion is to use the main Sky blue color for the `Add it now` button instead of the red/orange gradient. We only use that color gradient in buttons in very specific actions (like creating a new meeting), so I think it's best to stick to the main blue for this specific case.

mattkrick:
agreed the blue does feel better, although our prior art is to use the primary CTA for these modals. we'll probably need to revisit all the other modals like this & change the color there, too (removing a team member, removing an org user, asking for super permission from jira to fix their integration, promoting to team lead, etc....)

enriquesanchez:
>agreed the blue does feel better, although our prior art is to use the primary CTA for these modals

Thanks @mattkrick 🙌

@Dschoordsch I think then it's best to keep it as is right now, and further down the line I can look at updating these colors altogether.

enriquesanchez (approved):

Dschoordsch:
> What happens if the user clicks cancel? Would the task still be moved and disconnected from the integration? Or would it abort the action altogether?

The action is aborted.

mattkrick (changes requested):

mattkrick:
Loom: https://www.loom.com/share/22327acd2f8643718f31cad2e572ec52

packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterTeamAssigneeAddGitHubDialog.tsx
```diff
@@ -0,0 +1,117 @@
...
+const StyledPrimaryButton = styled(PrimaryButton)({
+  marginLeft: 16
+})
+
+const Label = styled('div')({
+  fontSize: 14,
+  fontWeight: 600
+})
+
+const TaskFooterTeamAssigneeAddGitHubDialog = (props: Props) => {
```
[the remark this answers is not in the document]
Dschoordsch: I can imagine that people might reconsider adding the integration to the target team, but for me it's more important that they are informed that the new integration is added to the team. To me moving over the integration to the new team is not obvious, if I think about it, then it is logical, but when I'm moving a task I'm not thinking about what the tool might do in the background.

packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterTeamAssigneeAddJiraDialog.tsx
```diff
@@ -0,0 +1,117 @@
...
+  marginLeft: 16
+})
+
+const Label = styled('div')({
+  fontSize: 14,
+  fontWeight: 600
+})
+
+const TaskFooterTeamAssigneeAddJiraDialog = (props: Props) => {
+  const {onClose, onIntegrationAdded, teamId, teamName} = props
```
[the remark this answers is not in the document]
Dschoordsch: I will leave it like this, otherwise it will just be a mess in TaskFooterTeamAssigneeMenu with all the mutation props. If I would refactor this, then into 3 components: 1 with common dialog parts and 2 components just filling in Jira/GitHub specifics.

packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterTeamAssigneeMenu.tsx
```diff
@@ -32,17 +60,57 @@ const TaskFooterTeamAssigneeMenu = (props: Props) => {
...
+  const handleClose = () => {
+    closePortal()
+    menuProps.closePortal()
+  }
+
+  const handleTaskUpdate = (newTeam) => async () => {
     if (!submitting && teamId !== newTeam.id) {
+      setNewTeam(newTeam)
+      if (isGitHubTask || isJiraTask) {
+        const result = await atmosphere.fetchQuery<
```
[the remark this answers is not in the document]
Dschoordsch: I spend quite some time trying to make it work with `loadQuery`, but I wasn't successful. `loadQuery` is really only meant to be used in conjunction with `usePreloadedQuery`. It does not return any data if it doesn't do a network request.
From what I understand from [relay query-retention]](https://relay.dev/docs/guided-tour/reusing-cached-data/presence-of-data/#query-retention) the hooks would only retain data until they are going out of scope which would cause `dispose` to be called on the query. `dispose` should also be called on `loadQuery` thus the data retention would be rather limited.
I can wrap `fetchQuery` in a hook to also properly do the `environment.retain` and `dispose` calls, but using any of the other relay hooks requires much more state in the component* which I always find more difficult to read and debug.

*I would need to use the query in a hook (`useRefetchableFragment`) and use an effect on the `newTeam` state and the result of the hook, make sure it matches and then open the dialog. Not too bad, but I would like to avoid it. Also the queryReference would need to come from somewhere.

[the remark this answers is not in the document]
mattkrick: you're not kidding! yikes, relay makes it _really_ hard to do that outside of react.

Interestingly enough, it looks like `fetchQuery` is doing what we want it to do. it keeps the query in the store & it _could_ be disposed of, but relay doesn't because our `gcReleaseBufferSize` is so large. Looks like you're right, fetchQuery is probably the best way to do things like this moving forward.

packages/server/graphql/mutations/changeTaskTeam.ts
```diff
@@ -55,6 +56,7 @@ export default {
     }
 
     // RESOLUTION
+    // filter mentions of old team members from task content
```
mattkrick, line 59: it's possible that someone calls this mutation outside of our UI.
If that happens, we need this mutation to guarantee that an integration exists on the new `teamId`

I propose the following:
- remove the OAuth2 flow from the client, if they click cancel, we abort, just like you have written. if they click OK, then we call this mutation
- this mutation looks up the AtlassianAuth (or GitHubAuth) for the viewerId/teamId combination
- if it exists, the teamId param is valid
- if it doesn't exist, then it looks up the Auth for the viewerId/oldTeamId
- if that doesn't exist, return error: No valid integration auth
- If that does exist, then copy that row & insert it with the updated teamId

eliminating the extra oauth2 flow means the process goes faster & more importantly we can verify that the new teamId supports that task integration

Dschoordsch: I got a bit lost in edge cases. When the viewer has no integration in the target team, there are these options:
1) if user has integration in source team, then we will use that, but still ask
2) if accessUser is someone else and they have integration for target team, then we will use that without asking
3) else we need to ask the user to integrate with complete oauth flow

For now I ignore 2) and 3) and just assume it's 1) as that's the most common case in the UI.
Better would be of course to detect 2) and 3) as well and handle it appropriately in the UI, but we do not get all information necessary at the moment via GraphQL and I also don't want to repeat the same logic if there is a nicer solution.

Let's just roll with 1) and fix 2) and 3) another day, user value is probably also minimal. Use case 2) I can imagine is a bit annoying, i.e. team lead has 2 parabol teams and adds integration. They're also the only ones with access to Jira. Every time some other team member would move a task, they would get shown the dialog. It would still move the task, but it might get annoying. But if that hypothetical user exists, they can open a ticket.

packages/server/graphql/mutations/changeTaskTeam.ts
```diff
@@ -72,9 +133,25 @@ export default {
     const updates = {
       content: rawContent === nextRawContent ? undefined : JSON.stringify(nextRawContent),
       updatedAt: now,
-      teamId
+      teamId,
+      integration: task.integration && {
```
mattkrick, line 137: the `&&` will turn into a `false`.
currently, the `integration` field doesn't exist on a Task if it is not integrated. 
to keep this same structure, we would want this to be `undefined` if there is no integration or if the integration is unchanged (rethinkdb driver filters out keys with undefined values). Gotta love NoSQL 😞 

For your consideration, I wrapped all the integration stuff up in a single conditional. This might be useful as we move forward with jira server & gitlab integrations. Then you can just write `integration: updatedIntegration`: 

```ts
    const {integration} = task
    let updatedIntegration: typeof integration
    if (integration) {
      const {accessUserId, service} = integration
      const authKeys = [
        {teamId: task.teamId, userId: viewerId},
        {teamId, userId: viewerId},
        {teamId, userId: accessUserId}
      ]
      const errToNullMap = <T>(val: T | Error | null) => (val instanceof Error ? null : val)
      const [sourceTeamAuth, targetTeamAuth, accessUsersTargetTeamAuth] =
        service === 'jira'
          ? (await dataLoader.get('freshAtlassianAuth').loadMany(authKeys)).map(errToNullMap)
          : service === 'github'
          ? (await dataLoader.get('githubAuth').loadMany(authKeys)).map(errToNullMap)
          : []
      if (!targetTeamAuth && !sourceTeamAuth && !accessUsersTargetTeamAuth) {
        return standardError(new Error('No valid integration found'), {
          userId: viewerId
        })
      }
      const data = {teamId, userId: viewerId}
      // use viewer's auth or fallback to the existing accessUserId
      const nextAccessUserId = sourceTeamAuth || targetTeamAuth ? viewerId : accessUserId
      if (nextAccessUserId !== accessUserId) {
        updatedIntegration = {...integration, accessUserId: nextAccessUserId}
      }
      if (sourceTeamAuth && !targetTeamAuth) {
        if (service === 'jira') {
          await upsertAtlassianAuth({
            ...sourceTeamAuth,
            teamId
          })
          publish(SubscriptionChannel.TEAM, teamId, 'AddAtlassianAuthPayload', data, subOptions)
        } else if (service === 'github') {
          await upsertGitHubAuth({
            ...sourceTeamAuth,
            teamId
          })
          publish(SubscriptionChannel.TEAM, teamId, 'AddGitHubAuthPayload', data, subOptions)
        }
      }
    }
```

packages/server/graphql/mutations/changeTaskTeam.ts
```diff
@@ -54,7 +58,64 @@ export default {
       })
     }
 
+    // The task might have been pushed by someone else for viewer (`userId !== accessUserId`).
+    // In that case we still try to use the viewer's target team authentication, but fall back to the
+    // accessUser's in case it is present for the target team.
+    const [sourceTeamAuth, targetTeamAuth, accessUsersTargetTeamAuth] =
+      task.integration?.service === 'jira'
+        ? await Promise.all([
+            dataLoader.get('freshAtlassianAuth').load({teamId: task.teamId, userId: viewerId}),
```
mattkrick, line 67: these might be good candidates for `loadMany`, although those have the possibility of returning an Error instance.
I wrote an example of what that might look like, let me know what you think

packages/server/graphql/mutations/changeTaskTeam.ts
mattkrick: fantastic work exploring all the edge cases.
i think your assumption are reasonable & i feel very comfortable with the solution you've built!

mattkrick (changes requested):

packages/server/graphql/mutations/changeTaskTeam.ts
Dschoordsch: While I like the `updatedIntegration` cleanup, the current `&&` would still work, since it would return `undefined` when `task.integration === undefined`. `&&` doesn't convert the result to boolean.

packages/server/graphql/mutations/changeTaskTeam.ts
Dschoordsch: I don't really like the `loadMany` signature and it's just a wrapper for `Promise.all(load(...))`. Question is whether we want to silently ignore errors and convert them to null or let them bubble up. I tend to default to just let these bubble up since we don't expect any errors here. But I can reduce repetition.

docs/integrations.md
```diff
@@ -0,0 +1,26 @@
...
+
+- [ ] **reading** a task uses **any team member's auth**
+      in the preferred order of:
+  - `task.integration?.accessUserId`
+  - fallback to viewer
+  - fallback to the team lead (reason being is a team lead is less likely to leave a team. maybe this is too optimistic?)
+  - fallback to any other member
+- [x] **pushing a task** requires **viewer auth or assignee's auth**, but in both cases a comment will be added if viewer !== assignee
+- [x] adding tasks in **scoping** requires **viewer's auth**
+- [ ] adding task **estimates** uses **viewer's auth or assignee's auth** (Most likely just added in scoping by the assignee, but even when added to the board before it's little risk as estimating is a team activity)
```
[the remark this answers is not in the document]
Dschoordsch: true

packages/server/graphql/mutations/changeTaskTeam.ts
mattkrick: ah you're right! that's what i get for thinking at night

packages/server/graphql/mutations/changeTaskTeam.ts
mattkrick: agreed, the signature used to be so much nicer! https://github.com/graphql/dataloader/pull/216
we're already silently ignoring any errors inside the batchFn, i wish we could tell the dataloader that we don't need any errors that come out of it.

mapping & wrapping in Promise.all is probably the cleanest way as long as we can guarantee that none of them will throw.

Issue #5243, opened 2021-08-02
Cannot archive/delete meeting


thaywood83:
[A user wrote the following feedback 🔒](https://app.hubspot.com/contacts/3888472/contact/8034101) :

>  I have a rogue meeting that I can't get rid of in the "Meetings" tab. It's called "Week of 2021-06-21 Refinements". It says it's complete but I can't make it go away. Any help would be appreciated! Thanks!

> Oddly enough, this meeting doesn’t even appear in my Timeline 🤔 Only in the Meetings tab. And when you click it, it just redirects right back to where I was – on the Meetings page.

PR #5520, opened 2021-10-14
Fix ending meeting without open phases

Dschoordsch:
With the right sequence of events it is possible to end up with a
complete (no open phases) meeting which did not end yet. This should not
throw.

Relates-To: #5243


mattkrick:
🤦‍♂️ this makes perfect sense, not sure how i didn't see it sooner. Even typescript tried to warn me but i casted over it 🤣
Nice find!

Issue #5523, opened 2021-10-14
After switching to User reads from PG, incorrect objects fetched for some users


jordanh:
## Issue - Bug

Some users are reporting that after logging in or viewing meeting history that they are occasionally seeing User objects for a user who is not theirs (!)

The incorrect object will clear after logging out and logging in.

### Acceptance Criteria (optional)
Users can:
   - always retrieve correct User objects

**Estimated effort:** 2 hour to triage. More if root cause is already identified.

PR #5524, opened 2021-10-14
Map users to corresponding id

Dschoordsch:
In custom redis query return the users in the same order as they were
requested.

Fixes: #5523


Dschoordsch:
@tianrunhe could you please review again?

PR #5538, opened 2021-10-19
Fix/users type from dataloader

Dschoordsch:
Properly handle trying to read non-existing users from dataloader.    
Either ignore the user or throw an error, depending on the situation.


Dschoordsch:
You are right about this, the recent fix in dataloader resulted in stricter typing. Returning undefined for a user read could have happened with rethinkdb before as well. Most of these places should not be reachable for consistent DBs.
 As far as I can see, all these situations would have resulted in "cannot read X of undefined" errors before.

PR #5546, opened 2021-10-20
Rate limit handling

### LINKED ISSUE #5324: Error: Issue does not exist or you do not have permission to see it.
https://sentry.io/organizations/parabol/issues/2529575750/?referrer=github_plugin

Affected 103 users, happened 4.7k times in the last 30 days. Not fatal, but worth looking into.

```
Error: Issue does not exist or you do not have permission to see it.
  File "./packages/client/utils/AtlassianManager.ts", line 317, in p.get
    return new Error(json.errorMessages[0])
  ?, in runMicrotasks
  File "./packages/client/utils/AtlassianManager.ts", line 553, in async p.getIssue
    const issueRes = await this.get
```

tiffanyhan:
Resolves #5324

AC:

- [ ] Create a new sprint poker meeting. In the Scope phase, add a couple of Jira issues.
- [ ] Go to the estimate phase and proceed to one of the Jira issues. Everything should look as normal, as we are either returning the issue directly from jira and then cacheing it, or else displaying a cached redis version and then lazily fetching from jira with no changes to the issue.
- [ ] in `.env`, add `TEST_JIRA_FRESH_ISSUE_UPDATES_CACHE='true'`, then restart your server. if you're on the same url, wait for the page to reload. issues should look as normal as the cache was cleared and they pull from jira.
- [ ] refresh the page. issues should be returned from cache, then quickly lazily updated and you will see their titles change.
- [ ] now add `TEST_JIRA_ISSUE_RATE_LIMITED='true'` to `.env`, and restart your server. the page will reload on the same url. issues should again look the same as the cache was cleared and they pull from jira.
- [ ] refresh the page. issues are returned from cache, and this time they aren't updated as the request to jira was simulated to be rate limited.
- [ ] now add `TEST_JIRA_RETRY_SUCCESS='true'` to `.env`, and restart your server. at first issues should appear as usual.
- [ ] refresh the page. the stale issue will display for a couple seconds, and then will be updated (more slowly than before). what happened was we simulated that the first lazy request to jira got rate limited and then the delayed retry succeeded.


packages/server/utils/atlassian/jiraIssues.ts
```diff
@@ -0,0 +1,111 @@
...
+  issueKey: string,
+  onSuccess: (
+    issue: JiraIssueBean<{description: string; summary: string}, {description: string}>
+  ) => void /* cb invoked when we do lazy async request */,
+  onError: (e: Error) => void /* error invoked when we do lazy async request */,
+  extraFieldIds: string[] = []
+) => {
+  const key = `jira:${cloudId}:${issueKey}:${JSON.stringify(extraFieldIds)}`
+  const redis = getRedis()
+  let cachedIssue = await redis.get(key)
```
mattkrick, line 29: fetching the issue from atlassian should not have to wait for the result to come back from redis.

for your consideration:

```ts

interface CacheHit {
  resolve: undefined
  cachedIssue: string
}

interface CacheMiss {
  resolve: (value: unknown) => void
  cachedIssue: null
}

interface CacheInit {
  resolve: undefined
  cachedIssue: undefined
}

type ReturnValue = CacheInit | CacheMiss | CacheHit

export const getIssue = async (
  manager: AtlassianManager,
  cloudId: string,
  issueKey: string,
  pushUpdateToClient: (issue: JiraIssueRaw) => void /* cb invoked when we do lazy async request */,
  extraFieldIds: string[] = []
) => {
  const key = `jira:${cloudId}:${issueKey}:${JSON.stringify(extraFieldIds)}`
  const redis = getRedis()
  const returnValue = {resolve: undefined, cachedIssue: undefined} as ReturnValue
  const cachedIssuePromise = redis.get(key)
  const issueResPromise = manager.getIssue(cloudId, issueKey, extraFieldIds).then(async (res) => {
    await issueResPromise
    if (res instanceof RateLimitError) {
      // do rate limit handling things
      const {retryAt} = res
      const delay = Math.max(0, retryAt.getTime() - Date.now())
      if (delay > ms('10s')) return res
      await sleep(delay)
      return getIssue(manager, cloudId, issueKey, pushUpdateToClient, extraFieldIds)
    } else if (res instanceof Error) {
      // return an error
      return res
    }

    if (returnValue.cachedIssue === undefined) {
      // redis is guaranteed to have returned, this is a noop
    } else if (returnValue.cachedIssue === null) {
      // the cache was empty when atlassian returned
      const issueFromJira = stringify(res)
      await redis.set(key, issueFromJira, 'PX', ISSUE_TTL_MS)
      returnValue.resolve(res)
    } else {
      // there was a cached value in redis
      const issueFromJira = stringify(res)
      if (issueFromJira !== returnValue.cachedIssue) {
        // the value from jira is fresher than what we have in redis
        // update redis & push an update via pubsub
        redis.set(key, issueFromJira, 'PX', ISSUE_TTL_MS)
        pushUpdateToClient(res)
      }
      // if they're equal, do nothing
    }
  })

  const cachedIssue = await cachedIssuePromise
  if (cachedIssue) {
    returnValue.cachedIssue = cachedIssue
    return cachedIssue
  }
  return new Promise((resolve) => {
    // resolve only exists if there was no cached value
    returnValue.resolve = resolve
  })
}
```

packages/server/utils/atlassian/jiraIssues.ts
```diff
@@ -0,0 +1,98 @@
...
+  onError: any /* error invoked when we do lazy async request */,
+  extraFieldIds: string[] = []
+) => {
+  const key = `jira:${cloudId}:${issueKey}:${JSON.stringify(extraFieldIds)}`
+  const redis = getRedis()
+  let cachedIssue = await redis.get(key)
+  cachedIssue = cachedIssue ? JSON.parse(cachedIssue) : cachedIssue
+
+  if (!cachedIssue) {
+    const issueRes = await manager.getIssue(cloudId, issueKey, extraFieldIds)
```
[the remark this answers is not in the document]
mattkrick: the strategy we want to employ is STORE_AND_NETWORK, so we'll watch to fetch both at the exact same time

packages/server/utils/atlassian/jiraIssues.ts
```diff
@@ -0,0 +1,98 @@
...
+    if (issueRes instanceof Error) {
+      return issueRes
+    }
+    await redis.set(key, JSON.stringify(issueRes), 'PX', ISSUE_TTL_MS)
+    const json = await redis.get(key)
+    return JSON.parse(json!)
```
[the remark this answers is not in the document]
mattkrick: no need to await a set & a get & reparse what we already have.

packages/server/utils/atlassian/jiraIssues.ts
```diff
@@ -0,0 +1,111 @@
...
+  cachedIssue: any,
+  manager: AtlassianManager,
+  cloudId: string,
+  issueKey: string,
+  onSuccess: any,
+  onError: any,
+  extraFieldIds: string[] = [],
+  retryCount = 0
+) => {
+  let freshIssue = await manager.getIssue(cloudId, issueKey, extraFieldIds)
```
mattkrick, line 69: let's not repeat this logic is 2 different places

packages/server/utils/atlassian/jiraIssues.ts
```diff
@@ -0,0 +1,111 @@
...
+      onError(freshIssue)
+    }
+  } else {
+    if (process.env.TEST_JIRA_FRESH_ISSUE_UPDATES_CACHE) {
+      freshIssue.fields.summary = 'issue was edited and lazily fetched'
+    }
+    const issuesAreEqual = jsonEqual(cachedIssue, freshIssue)
+    if (!issuesAreEqual) {
+      const redis = getRedis()
+      const key = `jira:${cloudId}:${issueKey}:${JSON.stringify(extraFieldIds)}`
```
mattkrick, line 106: let's try to generate this key once (see example in comment above)
let's also try not to waterfall the hits to redis

packages/server/dataloader/atlassianLoaders.ts
```diff
@@ -115,48 +119,63 @@ export const jiraIssue = (parent: RethinkDataLoader) => {
             .map((estimate) => estimate.jiraFieldId)
             .filter(isNotNull)
 
-          const issueRes = await manager.getIssue(cloudId, issueKey, estimateFieldIds)
-          if (issueRes instanceof Error) {
-            sendToSentry(issueRes, {userId, tags: {cloudId, issueKey, teamId}})
-            return null
+          const gotIssueCb = async (issueRes: any) => {
```
mattkrick, line 122: see rewrite in comment above, maybe we can move that whole function into this dataloader?

mattkrick:
Loom: https://www.loom.com/share/847c1bda394148d793af5ab6955dabb0

mattkrick (changes requested):

tiffanyhan:
@mattkrick thanks for the review. would love to hear your thoughts on how i refactored this. the main difference is that neither redis nor jira is awaited. we don't assume that redis will always return first, since it could be unresponsive (even though this is unlikely). whichever promise resolves first, is the one we return to the client. this way, if redis is unresponsive we still return a res from jira, instead of timing out and displaying a blank screen.

i also moved the request to jira into its own helper fn so we can reuse the obj which keeps our originally created promise (instead of creating a new one every time we retry). this way, we can return to the client the original resolved promise in any subsequent retries.

i also added back in max retry logic. i don't trust jira to always send us `Retry-After` header, and in that case we won't be infinitely making retry requests.

i'm okay with keeping this core logic in `jiraIssues` file instead of merging it with dataloader, since it seems like there's already so much going on in dataloader. lmk if you have a strong feeling on this

packages/server/utils/atlassian/jiraIssues.ts
```diff
@@ -0,0 +1,121 @@
...
+  pushUpdateToClient: (
+    issue: JiraIssueRaw
+  ) => void /* cb invoked when we get a fresher jira issue */,
+  extraFieldIds: string[],
+  redisKey: string,
+  requests: StoreAndNetworkRequests, // resolve originally created promise in outer fn
+  retryNo = 0
+) => {
+  // get issue from external network (jira api)
+  manager.getIssue(cloudId, issueKey, extraFieldIds).then(async (res) => {
```
mattkrick, line 40: let's refactor this to `await manager.getIssue`

packages/server/utils/atlassian/jiraIssues.ts
```diff
@@ -0,0 +1,121 @@
...
+    if (res instanceof RateLimitError) {
+      const {retryAt} = res
+      const delay = Math.max(0, retryAt.getTime() - Date.now())
+      if (delay > ms('10s') || retryNo > MAX_RETRIES) return
```
mattkrick, line 44: if the delay is > 10 seconds, when does `requests.firstToResolve` ever get called? i think this would cause a mem leak because the promise would never resolve?

packages/server/utils/atlassian/jiraIssues.ts
```diff
@@ -0,0 +1,121 @@
...
+      await sleep(delay)
+      _getIssueFromJira(
```
mattkrick, line 46: let's return this value so if another function wanted to await this value it could

packages/server/utils/atlassian/jiraIssues.ts
```diff
@@ -0,0 +1,121 @@
...
+        requests,
+        retryNo + 1
+      )
+      return
+    } else if (res instanceof Error) {
+      /*
+       * does nothing if redis already returned a res, else resolve with error
+       * don't wait for redis if it's unresponsive and still hasn't returned
+       */
+      requests.firstToResolve!(res)
```
mattkrick, line 62: can remove non-null coercion after the primary function guarantees that firstToResolve always exists

packages/server/utils/atlassian/jiraIssues.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    const redis = getRedis()
+    // got a successful response from jira
+    if (!requests.cachedIssue) {
+      // redis is unresponsive or returned null
+      requests.firstToResolve!(res)
+      redis.set(redisKey, stringify(res), 'PX', ISSUE_TTL_MS) // set cache behind
+    } else {
+      // there was a cached value in redis, firstToResolve already called
+      const issueFromJiraStr = stringify(res)
```
mattkrick, line 74: can move to L66 since this is used in both branches of the conditional

packages/server/utils/atlassian/jiraIssues.ts
```diff
@@ -0,0 +1,121 @@
...
+      if (issueFromJiraStr !== requests.cachedIssue) {
+        // update redis & push an update via pubsub
+        pushUpdateToClient(res)
+        redis.set(redisKey, issueFromJiraStr, 'PX', ISSUE_TTL_MS) // set cache behind
```
mattkrick, line 78: can move below the conditional since it is used in both branches

packages/server/utils/atlassian/jiraIssues.ts
```diff
@@ -0,0 +1,121 @@
...
+  pushUpdateToClient: (
+    issue: JiraIssueRaw
+  ) => void /* cb invoked when we get a fresher issue from jira */,
+  extraFieldIds: string[] = []
+) => {
+  const requests = {
+    firstToResolve: undefined,
+    cachedIssue: undefined
+  } as StoreAndNetworkRequests
+  const returnValuePromise = new Promise((resolve) => {
```
mattkrick, line 98: let's get rid of this & just
```ts
  return new Promise<Unpromise<ReturnType<typeof manager['getIssue']>>>((resolve) => {
  const requests = {
    firstToResolve: resolve,
    cachedIssue: undefined
  } 
  ...
})
```

that way requests.firstToResolve is guaranteed to exist & we don't need non-null coercion everywhere

mattkrick: the big generic is to make sure that this returns the same type as manager.getIssue

packages/server/dataloader/atlassianLoaders.ts
```diff
@@ -116,48 +120,62 @@ export const jiraIssue = (parent: RootDataLoader) => {
...
-            })
-          )
-
-          return {
-            ...fields,
-            descriptionHTML: updatedDescription,
-            teamId,
-            userId
+          if (issueRes instanceof Error) {
+            logError(issueRes)
```
mattkrick, line 174: let's inline this function since it's only used once

packages/client/subscriptions/TeamSubscription.ts
```diff
@@ -80,6 +80,13 @@ const subscription = graphql`
       ...UpdateTeamNameMutation_team @relay(mask: false)
       ...UpdateUserProfileMutation_team @relay(mask: false)
       ...UpgradeToProMutation_team @relay(mask: false)
+
+      ... on JiraIssue {
+        id
```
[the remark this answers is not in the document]
mattkrick: thinking about it more-- yep, since 1 user is requesting this, the update should be pushed to that same user only

mattkrick (changes requested):
the only major thing is not resolving the function when the retry is greater than 10 seconds.

packages/server/utils/atlassian/jiraIssues.ts
tiffanyhan: i like this much better, thanks for the suggestion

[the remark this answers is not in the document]
tiffanyhan: resolved via the below suggestion

packages/client/utils/AtlassianManager.ts
```diff
@@ -307,6 +325,15 @@ export default abstract class AtlassianManager {
       return res
     }
     const json = (await res.json()) as AtlassianError | JiraNoAccessError | JiraGetError | T
+
+    if (res.status === 429) {
```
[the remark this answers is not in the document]
tiffanyhan: ah good point, this was an attempt to resolve your prior suggestion:

> message should probably be the error that jira gives us instead of error because they might give something more informative so we can use that to debug when that error is hit

afaik atlassian doesn't give specific details on the shape. i'll revert back to the hard coded message for now, but did you have a different idea about where the error msg would come from? @mattkrick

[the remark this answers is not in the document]
tiffanyhan: ah good point, this was an attempt to resolve your prior suggestion:

> message should probably be the error that jira gives us instead of error because they might give something more informative so we can use that to debug when that error is hit

afaik atlassian doesn't give specific details on the shape. i'll revert back to the hard coded message for now, but did you have a different idea about where the error msg would come from? @mattkrick

packages/client/utils/AtlassianManager.ts
```diff
@@ -541,9 +568,7 @@ export default abstract class AtlassianManager {
   async getIssue(cloudId: string, issueKey: string, extraFieldIds: string[] = []) {
     const baseFields = ['summary', 'description']
     const reqFields = [...baseFields, ...extraFieldIds].join(',')
-    const issueRes = await this.get<
-      JiraIssueBean<{description: string; summary: string}, {description: string}>
-    >(
+    const issueRes = await this.get<JiraIssueRaw>(
       `https://api.atlassian.com/ex/jira/${cloudId}/rest/api/3/issue/${issueKey}?fields=${reqFields}&expand=renderedFields`
     )
     if (issueRes instanceof Error) return issueRes
```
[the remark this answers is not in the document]
tiffanyhan: i'm not sure why this would be necessary. a `RateLimitError` obj should pass the instanceof Error check with true. Am i missing something?

packages/server/utils/atlassian/jiraIssues.ts
tiffanyhan: you are totally correct. i overlooked this, thank you for catching it

packages/client/utils/AtlassianManager.ts
```diff
@@ -187,7 +192,10 @@ interface JiraAddCommentResponse {
   jsdPublic: true
 }
 
-export type JiraGetIssueRes = JiraIssueBean<JiraGQLFields>
+export type JiraGetIssueRes = JiraIssueBean<
```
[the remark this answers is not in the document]
tiffanyhan: oof this was a mistake, reverted back

packages/client/subscriptions/TeamSubscription.ts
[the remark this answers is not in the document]
tiffanyhan: @mattkrick doesn't this get called when we are in the estimate phase of a poker meeting too, which means all team members in the meeting should be able to see the updated integrated issue they are scoping?

packages/server/utils/atlassian/jiraIssues.ts
tiffanyhan: ah actually i don't think i can pull out the redis set since in the second conditional, we only want to reset the key if it's different. it's not harmful, but seems unnecessary

packages/client/subscriptions/TeamSubscription.ts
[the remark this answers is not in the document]
tiffanyhan: @mattkrick i just tested pushing the update on `NotificationSubscription` instead of `TeamSubscription` channel. when doing it this way, the integrated user will see the updated issue when pushed async but the other teammate won't. this is bc we would only push to the userId who owns the atlassian auth. this behavior seems un-inuitive to me, so you can confirm if this is correct or not?

Screen Shot 2021-12-03 at 4 19 31 PM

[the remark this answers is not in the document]
tiffanyhan: ok i just pushed a commit changing it to publish on NotificationSubscription. If there is another teammate, they won't get the update if you refresh and get a fresher issue. but they'll get the fresher issue from the cache if they refresh their page. i think this is probably okay behavior.

mattkrick:
TEST:
- [ ] mock out atlassian response with a repeated rate-limited response
- [ ] same as above, but with only 1 rate-limited response
- [ ] BAU

Issue #5549, opened 2021-10-20
Initial Mattermost webhook-based integration implemented


jordanh:
## Issue - Enhancement

This issue is too large, I apologize. Design is at: [#4412](https://github.com/ParabolInc/parabol/issues/4412#issuecomment-948003051)

### Acceptance Criteria (optional)
From `team/:teamid/settings/integrations`, users can:
   - Press `Connect` to reveal an integration panel
   - Add a webhook configuration
   - If a configuration is added and active, the panel stays visble
   - Press the 3-dots menu and remove the integration configuration (setting isActive to false)
  
Receive notifications via the webhook:
  - Start meeting
     - check-in
     - retrospective
     - sprint poker
  - Stop meeting
     - check-in
     - retrospective
     - sprint poker
 - Start retrospective stage timer
 - Stop retrospective stage timer

Further, on the stage timer:
   - Stage timer panel indicates when Mattermost is configured
   - When Mattermost is configured, Slack Notification setup does not appear
   - When neither are configured, only Slack Notification setup appears
   - When both are configured, both appear and the panel doesn't look like garbage

**Estimated effort:** 21 hours ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

PR #5550, opened 2021-10-20
Add mattermost

### LINKED ISSUE #4412: Mattermost integration designed
## Issue - Enhancement

[Mattermost](https://mattermost.com/) is a collaboration platform similar to Slack popular in the regulated market. We're contractually obligated to build a Mattermost integration for the DoD by 2021 Jan 01.

The first step to developing this integration will be to design it.

Related: #3538

Users can:
   - CRUD Mattermost integration on their team
   - Notifications designed, mirroring today's Slack integration (at a minimum)
   - Set of implementation issues created

- **Estimated effort:** 13 points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

jordanh:
fix #4412
Implements design [#4412](https://github.com/ParabolInc/parabol/issues/4412#issuecomment-948003051) and #5549

### DIFF
--- a/packages/server/graphql/mutations/helpers/notifications/makeMattermostAttachments.ts
+++ b/packages/server/graphql/mutations/helpers/notifications/makeMattermostAttachments.ts
@@ -1,7 +1,6 @@
 import {PALETTE} from '~/styles/paletteV3'
+import {ExternalLinks} from '~/types/constEnums'
 import authorIcon from '../../../../../../static/images/brand/mark-color@3x.png'
-import appOrigin from '../../../../appOrigin'
-import makeAppURL from '../../../../../client/utils/makeAppURL'
 
 export interface ButtonSpec {
   label: string
@@ -28,20 +27,15 @@ interface FieldAttachment extends FieldAttachmentHeader {
   fields: Array<Field>
 }
 
-const isUrlAbsolute = (url) => url.indexOf('://') > 0 || url.indexOf('//') === 0
-
 export const makeFieldsAttachment = (
   fields: Array<Field>,
   header?: FieldAttachmentHeader
 ): FieldAttachment => {
-  const absoluteAuthorIconUri = isUrlAbsolute(authorIcon)
-    ? authorIcon
-    : makeAppURL(appOrigin, authorIcon)
   const defaultHeader = {
     fallback: 'Parabol has a message for you',
-    author_icon: absoluteAuthorIconUri,
+    author_icon: authorIcon,
     author_name: 'Parabol',
-    author_link: 'https://parabol.co/',
+    author_link: ExternalLinks.LANDING_PAGE,
     color: PALETTE.GRAPE_500
   }
   return {
--- a/packages/client/types/constEnums.ts
+++ b/packages/client/types/constEnums.ts
@@ -117,7 +117,8 @@ export const enum ExternalLinks {
   INTEGRATIONS_GITHUB = 'https://www.parabol.co/integrations/github',
   RESOURCES = 'https://www.parabol.co/resources',
   SUPPORT = 'https://www.parabol.co/support',
-  TEAM = 'https://www.parabol.co/team/'
+  TEAM = 'https://www.parabol.co/team/',
+  LANDING_PAGE = 'https://www.parabol.co/'
 }
 
 export const enum Filter {
--- a/packages/server/graphql/mutations/helpers/notifications/notifySlack.ts
+++ b/packages/server/graphql/mutations/helpers/notifications/notifySlack.ts
@@ -1,25 +1,25 @@
 import formatTime from 'parabol-client/utils/date/formatTime'
 import formatWeekday from 'parabol-client/utils/date/formatWeekday'
+import makeAppURL from 'parabol-client/utils/makeAppURL'
 import findStageById from 'parabol-client/utils/meetings/findStageById'
 import {phaseLabelLookup} from 'parabol-client/utils/meetings/lookups'
+import appOrigin from '../../../../appOrigin'
 import getRethink from '../../../../database/rethinkDriver'
+import MeetingAction from '../../../../database/types/MeetingAction'
+import MeetingPoker from '../../../../database/types/MeetingPoker'
+import MeetingRetrospective from '../../../../database/types/MeetingRetrospective'
 import SlackAuth from '../../../../database/types/SlackAuth'
 import SlackNotification, {
   SlackNotificationEvent
 } from '../../../../database/types/SlackNotification'
 import {toEpochSeconds} from '../../../../utils/epochTime'
-import makeAppURL from 'parabol-client/utils/makeAppURL'
 import segmentIo from '../../../../utils/segmentIo'
 import sendToSentry from '../../../../utils/sendToSentry'
 import SlackServerManager from '../../../../utils/SlackServerManager'
 import errorFilter from '../../../errorFilter'
 import {DataLoaderWorker} from '../../../graphql'
-import appOrigin from '../../../../appOrigin'
-import MeetingRetrospective from '../../../../database/types/MeetingRetrospective'
-import MeetingAction from '../../../../database/types/MeetingAction'
-import MeetingPoker from '../../../../database/types/MeetingPoker'
-import {makeSection, makeSections, makeButtons} from './makeSlackBlocks'
 import getSummaryText from './getSummaryText'
+import {makeButtons, makeSection, makeSections} from './makeSlackBlocks'
 
 const getSlackDetails = async (
   event: SlackNotificationEvent,
@@ -191,10 +191,7 @@ export const notifySlackTimeLimitStart = async (
   const {name: teamName} = team
   const stageRes = findStageById(phases, facilitatorStageId)
   const {stage} = stageRes!
-  const maybeMeetingShortLink = makeAppURL(
-    process.env.INVITATION_SHORTLINK || appOrigin,
-    `${meetingId}`
-  )
+  const maybeMeetingShortLink = makeAppURL(process.env.INVITATION_SHORTLINK!, `${meetingId}`)
   const meetingUrl = makeAppURL(appOrigin, `meet/${meetingId}`)
   const {phaseType} = stage
   const phaseLabel = phaseLabelLookup[phaseType]
--- a/packages/server/jiraImagesHandler.ts
+++ b/packages/server/jiraImagesHandler.ts
@@ -1,11 +1,11 @@
 import {promises as fsp} from 'fs'
 import mime from 'mime-types'
+import path from 'path'
 import {HttpRequest, HttpResponse} from 'uWebSockets.js'
+import jiraPlaceholder from '../../static/images/illustrations/imageNotFound.png'
 import sleep from '../client/utils/sleep'
 import uWSAsyncHandler from './graphql/uWSAsyncHandler'
 import getRedis from './utils/getRedis'
-
-const jiraPlaceholderPath = 'static/images/illustrations/imageNotFound.png'
 
 const getImageFromCache = async (fileName: string, tryAgain: boolean) => {
   const redis = getRedis()
@@ -22,10 +22,15 @@ const getImageFromCache = async (fileName: string, tryAgain: boolean) => {
 let jiraPlaceholderBuffer: Buffer | undefined
 const servePlaceholderImage = async (res: HttpResponse) => {
   if (!jiraPlaceholderBuffer) {
-    jiraPlaceholderBuffer = await fsp.readFile(jiraPlaceholderPath)
+    jiraPlaceholderBuffer = await fsp.readFile(
+      path.join(__dirname, jiraPlaceholder.slice(__webpack_public_path__.length))
+    )
   }
   res.cork(() => {
-    res.writeStatus('200').writeHeader('Content-Type', 'image/png').end(jiraPlaceholderBuffer)
+    res
+      .writeStatus('200')
+      .writeHeader('Content-Type', 'image/png')
+      .end(jiraPlaceholderBuffer)
   })
 }
 
@@ -49,7 +54,10 @@ const jiraImagesHandler = uWSAsyncHandler(async (res: HttpResponse, req: HttpReq
   }
 
   res.cork(() => {
-    res.writeStatus('200').writeHeader('Content-Type', mimeType).end(imageBuffer)
+    res
+      .writeStatus('200')
+      .writeHeader('Content-Type', mimeType)
+      .end(imageBuffer)
   })
 })
 
--- a/packages/server/server.ts
+++ b/packages/server/server.ts
@@ -1,5 +1,5 @@
 import tracer from 'dd-trace'
-
+import {r} from 'rethinkdb-ts'
 import uws, {SHARED_COMPRESSOR} from 'uWebSockets.js'
 import stripeWebhookHandler from './billing/stripeWebhookHandler'
 import createSSR from './createSSR'
@@ -9,10 +9,11 @@ import webhookGraphQLHandler from './graphql/webhookGraphQLHandler'
 import ICSHandler from './ICSHandler'
 import './initSentry'
 import githubWebhookHandler from './integrations/githubWebhookHandler'
+import jiraImagesHandler from './jiraImagesHandler'
 import listenHandler from './listenHandler'
+import PROD from './PROD'
 import PWAHandler from './PWAHandler'
 import selfHostedHandler from './selfHostedHandler'
-import jiraImagesHandler from './jiraImagesHandler'
 import handleClose from './socketHandlers/handleClose'
 import handleMessage from './socketHandlers/handleMessage'
 import handleOpen from './socketHandlers/handleOpen'
@@ -21,8 +22,6 @@ import SSEConnectionHandler from './sse/SSEConnectionHandler'
 import SSEPingHandler from './sse/SSEPingHandler'
 import staticFileHandler from './staticFileHandler'
 import SAMLHandler from './utils/SAMLHandler'
-import PROD from './PROD'
-import {r} from 'rethinkdb-ts'
 
 tracer.init({
   enabled: process.env.DD_TRACE_ENABLED === 'true',
--- a/packages/server/types/githubTypes.ts
+++ b/packages/server/types/githubTypes.ts
@@ -8496,7 +8496,9 @@ export type Mutation = {
   /** Pin an issue to a repository */
   pinIssue?: Maybe<PinIssuePayload>
   /** Regenerates the identity provider recovery codes for an enterprise */
-  regenerateEnterpriseIdentityProviderRecoveryCodes?: Maybe<RegenerateEnterpriseIdentityProviderRecoveryCodesPayload>
+  regenerateEnterpriseIdentityProviderRecoveryCodes?: Maybe<
+    RegenerateEnterpriseIdentityProviderRecoveryCodesPayload
+  >
   /** Regenerates a verifiable domain's verification token. */
   regenerateVerifiableDomainToken?: Maybe<RegenerateVerifiableDomainTokenPayload>
   /** Reject all pending deployments under one or more environments */
@@ -8576,35 +8578,61 @@ export type Mutation = {
   /** Updates the role of an enterprise administrator. */
   updateEnterpriseAdministratorRole?: Maybe<UpdateEnterpriseAdministratorRolePayload>
   /** Sets whether private repository forks are enabled for an enterprise. */
-  updateEnterpriseAllowPrivateRepositoryForkingSetting?: Maybe<UpdateEnterpriseAllowPrivateRepositoryForkingSettingPayload>
+  updateEnterpriseAllowPrivateRepositoryForkingSetting?: Maybe<
+    UpdateEnterpriseAllowPrivateRepositoryForkingSettingPayload
+  >
   /** Sets the base repository permission for organizations in an enterprise. */
-  updateEnterpriseDefaultRepositoryPermissionSetting?: Maybe<UpdateEnterpriseDefaultRepositoryPermissionSettingPayload>
+  updateEnterpriseDefaultRepositoryPermissionSetting?: Maybe<
+    UpdateEnterpriseDefaultRepositoryPermissionSettingPayload
+  >
   /** Sets whether organization members with admin permissions on a repository can change repository visibility. */
-  updateEnterpriseMembersCanChangeRepositoryVisibilitySetting?: Maybe<UpdateEnterpriseMembersCanChangeRepositoryVisibilitySettingPayload>
+  updateEnterpriseMembersCanChangeRepositoryVisibilitySetting?: Maybe<
+    UpdateEnterpriseMembersCanChangeRepositoryVisibilitySettingPayload
+  >
   /** Sets the members can create repositories setting for an enterprise. */
-  updateEnterpriseMembersCanCreateRepositoriesSetting?: Maybe<UpdateEnterpriseMembersCanCreateRepositoriesSettingPayload>
+  updateEnterpriseMembersCanCreateRepositoriesSetting?: Maybe<
+    UpdateEnterpriseMembersCanCreateRepositoriesSettingPayload
+  >
   /** Sets the members can delete issues setting for an enterprise. */
-  updateEnterpriseMembersCanDeleteIssuesSetting?: Maybe<UpdateEnterpriseMembersCanDeleteIssuesSettingPayload>
+  updateEnterpriseMembersCanDeleteIssuesSetting?: Maybe<
+    UpdateEnterpriseMembersCanDeleteIssuesSettingPayload
+  >
   /** Sets the members can delete repositories setting for an enterprise. */
-  updateEnterpriseMembersCanDeleteRepositoriesSetting?: Maybe<UpdateEnterpriseMembersCanDeleteRepositoriesSettingPayload>
+  updateEnterpriseMembersCanDeleteRepositoriesSetting?: Maybe<
+    UpdateEnterpriseMembersCanDeleteRepositoriesSettingPayload
+  >
   /** Sets whether members can invite collaborators are enabled for an enterprise. */
-  updateEnterpriseMembersCanInviteCollaboratorsSetting?: Maybe<UpdateEnterpriseMembersCanInviteCollaboratorsSettingPayload>
+  updateEnterpriseMembersCanInviteCollaboratorsSetting?: Maybe<
+    UpdateEnterpriseMembersCanInviteCollaboratorsSettingPayload
+  >
   /** Sets whether or not an organization admin can make purchases. */
-  updateEnterpriseMembersCanMakePurchasesSetting?: Maybe<UpdateEnterpriseMembersCanMakePurchasesSettingPayload>
+  updateEnterpriseMembersCanMakePurchasesSetting?: Maybe<
+    UpdateEnterpriseMembersCanMakePurchasesSettingPayload
+  >
   /** Sets the members can update protected branches setting for an enterprise. */
-  updateEnterpriseMembersCanUpdateProtectedBranchesSetting?: Maybe<UpdateEnterpriseMembersCanUpdateProtectedBranchesSettingPayload>
+  updateEnterpriseMembersCanUpdateProtectedBranchesSetting?: Maybe<
+    UpdateEnterpriseMembersCanUpdateProtectedBranchesSettingPayload
+  >
   /** Sets the members can view dependency insights for an enterprise. */
-  updateEnterpriseMembersCanViewDependencyInsightsSetting?: Maybe<UpdateEnterpriseMembersCanViewDependencyInsightsSettingPayload>
+  updateEnterpriseMembersCanViewDependencyInsightsSetting?: Maybe<
+    UpdateEnterpriseMembersCanViewDependencyInsightsSettingPayload
+  >
   /** Sets whether organization projects are enabled for an enterprise. */
-  updateEnterpriseOrganizationProjectsSetting?: Maybe<UpdateEnterpriseOrganizationProjectsSettingPayload>
+  updateEnterpriseOrganizationProjectsSetting?: Maybe<
+    UpdateEnterpriseOrganizationProjectsSettingPayload
+  >
   /** Updates an enterprise's profile. */
   updateEnterpriseProfile?: Maybe<UpdateEnterpriseProfilePayload>
   /** Sets whether repository projects are enabled for a enterprise. */
-  updateEnterpriseRepositoryProjectsSetting?: Maybe<UpdateEnterpriseRepositoryProjectsSettingPayload>
+  updateEnterpriseRepositoryProjectsSetting?: Maybe<
+    UpdateEnterpriseRepositoryProjectsSettingPayload
+  >
   /** Sets whether team discussions are enabled for an enterprise. */
   updateEnterpriseTeamDiscussionsSetting?: Maybe<UpdateEnterpriseTeamDiscussionsSettingPayload>
   /** Sets whether two factor authentication is required for all users in an enterprise. */
-  updateEnterpriseTwoFactorAuthenticationRequiredSetting?: Maybe<UpdateEnterpriseTwoFactorAuthenticationRequiredSettingPayload>
+  updateEnterpriseTwoFactorAuthenticationRequiredSetting?: Maybe<
+    UpdateEnterpriseTwoFactorAuthenticationRequiredSettingPayload
+  >
   /** Updates an environment. */
   updateEnvironment?: Maybe<UpdateEnvironmentPayload>
   /** Sets whether an IP allow list is enabled on an owner. */
@@ -8612,7 +8640,9 @@ export type Mutation = {
   /** Updates an IP allow list entry. */
   updateIpAllowListEntry?: Maybe<UpdateIpAllowListEntryPayload>
   /** Sets whether IP allow list configuration for installed GitHub Apps is enabled on an owner. */
-  updateIpAllowListForInstalledAppsEnabledSetting?: Maybe<UpdateIpAllowListForInstalledAppsEnabledSettingPayload>
+  updateIpAllowListForInstalledAppsEnabledSetting?: Maybe<
+    UpdateIpAllowListForInstalledAppsEnabledSettingPayload
+  >
   /** Updates an Issue. */
   updateIssue?: Maybe<UpdateIssuePayload>
   /** Updates an IssueComment object. */
--- a/packages/server/types/modules.d.ts
+++ b/packages/server/types/modules.d.ts
@@ -11,6 +11,7 @@ declare module '*.graphql' {
 
 declare const __PRODUCTION__: string
 declare const __SOCKET_PORT__: string
+declare const __webpack_public_path__: string
 interface Window {
   __ACTION__: any
 }
--- a/scripts/webpack/dev.servers.config.js
+++ b/scripts/webpack/dev.servers.config.js
@@ -1,3 +1,4 @@
+require('./utils/dotenv')
 const path = require('path')
 const nodeExternals = require('webpack-node-externals')
 const transformRules = require('./utils/transformRules')
@@ -30,7 +31,8 @@ module.exports = {
   output: {
     filename: '[name].js',
     path: path.join(PROJECT_ROOT, 'dev'),
-    libraryTarget: 'commonjs'
+    libraryTarget: 'commonjs',
+    publicPath: `http://localhost:${process.env.PORT}/static/`
   },
   resolve: {
     alias: {
@@ -71,10 +73,7 @@ module.exports = {
         test: /\.(png|jpg|jpeg|gif|svg)$/,
         use: [
           {
-            loader: 'file-loader',
-            options: {
-              publicPath: (url) => `static/${url}`
-            }
+            loader: 'file-loader'
           }
         ]
       }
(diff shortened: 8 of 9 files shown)


jordanh:
Proposed tests:

From `team/:teamid/settings/integrations`, users can:
   - [ ] Press `Connect` to reveal an integration panel
   - [ ] Add a webhook configuration
   - [ ] If a configuration is added and active, the panel stays visble
   - [ ] Press the 3-dots menu and remove the integration configuration (setting isActive to false)
  
Receive notifications via the webhook:
  - Start meeting
     - [ ] check-in
     - [ ] retrospective
     - [ ] sprint poker
  - Stop meeting
     - [ ] check-in
     - [ ] retrospective
     - [ ] sprint poker
 - [ ] Start retrospective stage timer
 - [ ] Stop retrospective stage timer

Further, on the stage timer:
   - [ ] Stage timer panel indicates when Mattermost is configured
   - [ ] When Mattermost is configured, Slack Notification setup does not appear
   - [ ] When neither are configured, only Slack Notification setup appears
   - [ ] When both are configured, both appear and the panel doesn't look like garbage

Slack integration regression testing:
   - without Mattermost webhook configured
        - [ ] can add slack integration
        - [ ] can start/stop meeting, see notification
        - [ ] can start retrospective, set stage timer and see notification
        - [ ] can see slack notification when stage timer expires
     - _with_ Mattermost webhook configured
        - [ ] can add slack integration
        - [ ] can start/stop meeting, see notification
        - [ ] can start retrospective, set stage timer and see notification
        - [ ] can see slack notification when stage timer expires

packages/server/graphql/intranetSchema/mutations/runScheduledJobs.ts
```diff
@@ -35,28 +30,52 @@ const processMeetingStageTimeLimits = async (job: ScheduledJobMeetingStageTimeLi
...
+  const meetingUrl = makeAppURL(appOrigin, `meet/${meetingId}`)
+
+  let sendViaSlack = false
+  let sendViaMattermost = mattermostAuth?.isActive
+
+  if (slackAuth?.botAccessToken && slackNotification?.channelId) {
+    sendViaSlack = true
+    const manager = new SlackServerManager(slackAuth.botAccessToken)
+    const slackText = `Time’s up! Advance your meeting to the next phase: ${meetingUrl}`
+    const res = await manager.postMessage(slackNotification.channelId, slackText)
```
mattkrick, line 64: this is a hot take & feel free to fight me

I think both SlackManager & MattermostManager should implement an abstract class called NotificationManager.

NotificationManager has abstract methods like `timeLimitEnd(meetingId, teamId, dataLoader)`
then, we can do things like this:

```ts
sendViaMattermost && mattermostManager.timeLimitEnd()
sendViaSlack && slackManager.timeLimitEnd()
```

mattkrick: if we wanted to get really fancy, then we could have `abstract class NotificationManager` have statics like 

```
static DefaultTimeLimitEndMessage = (meetingUrl) => `Time up for ${meetingUrl}`
```

packages/server/graphql/intranetSchema/mutations/runScheduledJobs.ts
```diff
@@ -65,19 +84,13 @@ const processMeetingStageTimeLimits = async (job: ScheduledJobMeetingStageTimeLi
...
-  // get the meeting
-  // get the facilitator
-  // see if the facilitator has turned on slack notifications for the meeting
-  // if so, send the facilitator a slack notification
-  // if not, send the facilitator an in-app notification
 }
 
 const jobProcessors = {
   MEETING_STAGE_TIME_LIMIT_END: processMeetingStageTimeLimits
 }
```
mattkrick, line 91: if we cast this `as const` then we don't need `job as any` down below. this'll keep our types nice & strong

packages/server/graphql/intranetSchema/mutations/runScheduledJobs.ts
```diff
@@ -103,7 +116,7 @@ const runScheduledJobs = {
     //   description: 'filter jobs by their type'
     // }
   },
-  resolve: async (_source, {seconds}, {authToken}) => {
+  resolve: async (_source, {seconds}, {authToken, dataLoader}) => {
```
mattkrick, line 119: last arg is `:GQLContext` types

packages/server/graphql/mutations/helpers/notifications/makeMattermostAttachments.ts
```diff
@@ -0,0 +1,64 @@
...
+// Mattermost does not allow for Button actions to map to external URLs,
+// so we hack a little button using a Markdown table
+export const makeHackedFieldButtonValue = (spec: ButtonSpec): string => `
+| [${spec.label}](${spec.link}) |
+|:-----------------------------:|
+||`
+
+export const makeHackedButtonPairFields = (
+  buttonSpecPair: FixedLengthArray<ButtonSpec, 2>
+): FixedLengthArray<Field, 2> =>
```
mattkrick, line 60: since we're casting the return as `[Field, Field]` we don't need an explicit return value (in general, it's nice to not have 2 sources of truth for the return value)

for your consideration:
```ts
export const makeHackedButtonPairFields = (
  button1: ButtonSpec,
  button2: ButtonSpec
) =>
  [button1, button2].map((i) => ({
    short: true,
    value: makeHackedFieldButtonValue(i)
  })) as [button1: Field, button2: Field]
```

packages/server/graphql/mutations/helpers/notifications/notifyMattermost.ts
```diff
@@ -0,0 +1,300 @@
...
+import formatWeekday from 'parabol-client/utils/date/formatWeekday'
+import formatTime from 'parabol-client/utils/date/formatTime'
+import {toEpochSeconds} from '../../../../utils/epochTime'
+
+const getWebhookForTeamId = async (dataLoader: DataLoaderWorker, teamId: string) => {
+  const mattermost = await dataLoader.get('mattermostAuthByTeamId').load(teamId)
+  return mattermost?.webhookUrl
+}
+
+const notifyMattermost = async (
```
mattkrick, line 28: curious for your comments on the `NotificationsManager` implementation in the comment above

packages/server/graphql/mutations/removeMattermostAuth.ts
```diff
@@ -0,0 +1,41 @@
...
+    }
+  },
+  resolve: async (_source, {teamId}, {authToken, socketId: mutatorId, dataLoader}) => {
+    const operationId = dataLoader.share()
+    const subOptions = {mutatorId, operationId}
+    const viewerId = getUserId(authToken)
+
+    // AUTH
+    if (!isTeamMember(authToken, teamId)) {
+      return standardError(new Error('Team not found'), {userId: viewerId})
```
mattkrick, line 27: allowing any team member to remove any other team member's integration feels a little bad.

perhaps we have a 1:1 connection between team member & integration (just like with do with github & atlassian & slack)

then, when it comes time to send a message, we can pick the "best" token to use (ie the viewer's token, or any one else's, if a viewer token doesn't exist).

packages/server/graphql/intranetSchema/mutations/runScheduledJobs.ts
mattkrick: Finally, we might have a `ServiceNotifier` class that does the following:
- loops through all notification integrations & creates NotificationManagers for each valid one.
- This also implements the `NotificationManager` interface.
- End result, you call `notificationManager.timeLimitEnd()` and that calls slack & mattermost's implementations of that same method.

packages/server/graphql/types/AddMattermostAuthPayload.ts
```diff
@@ -0,0 +1,32 @@
+import {GraphQLObjectType} from 'graphql'
+import MattermostIntegration from './MattermostIntegration'
+import makeMutationPayload from './makeMutationPayload'
+import User from './User'
+import {GQLContext} from '../graphql'
+
+export const AddMattermostAuthSuccess = new GraphQLObjectType<any, GQLContext>({
+  name: 'AddMattermostAuthSuccess',
+  fields: () => ({
+    MattermostIntegration: {
```
mattkrick, line 10: hot take: let's keep this all lowercase. since we have `integration.mattermost`, don't wanna mix up camel & pascal

packages/server/graphql/types/AddMattermostAuthPayload.ts
```diff
@@ -0,0 +1,32 @@
...
+      type: MattermostIntegration,
```
mattkrick, line 11: the success payload should try to have all responses be nonnull.
the error payload is the other payload type. that's why 
`AddMattermostAuthPayload = 'AddMattermostAuthSuccess | MutationError'`

packages/server/graphql/types/MattermostIntegration.ts
```diff
@@ -0,0 +1,33 @@
...
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'The team that the token is linked to'
+    },
+    updatedAt: {
+      type: new GraphQLNonNull(GraphQLISO8601Type),
+      description: 'The timestamp the token was updated at'
+    },
+    userId: {
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'The id of the user that integrated Slack'
```
mattkrick, line 28: mattermost, not slack

packages/server/graphql/types/RemoveMattermostAuthPayload.ts
```diff
@@ -0,0 +1,25 @@
+import {GraphQLID, GraphQLObjectType} from 'graphql'
+import StandardMutationError from './StandardMutationError'
+import User from './User'
+import {GQLContext} from '../graphql'
+
+const RemoveMattermostAuthPayload = new GraphQLObjectType<any, GQLContext>({
```
mattkrick, line 6: let's use the new Success | Error pattern that is used here:  https://github.com/ParabolInc/parabol/blob/f49e74c0ca6a56a6ae13b385f094d434fd0ed6df/packages/server/graphql/types/AddMattermostAuthPayload.ts

packages/server/postgres/migrations/1634351004998_addMattermostAuth.ts
```diff
@@ -0,0 +1,26 @@
...
+  await client.connect()
+  await client.query(`
+  CREATE TABLE IF NOT EXISTS "MattermostAuth" (
+    "createdAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    "updatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    "isActive" BOOLEAN DEFAULT TRUE NOT NULL,
+    "webhookUrl" VARCHAR(2083) NOT NULL,
+    "userId" VARCHAR(100) NOT NULL,
+    "teamId" VARCHAR(100) NOT NULL,
+    PRIMARY KEY ("teamId")
```
mattkrick, line 15: I think we should make this teamId, userId. 
me having the ability to remove your integration feels bad.

packages/server/utils/MattermostServerManager.ts
```diff
@@ -0,0 +1,70 @@
...
+  webhookUrl: string
+  abstract fetch
+  headers: any
+
+  constructor(webhookUrl: string) {
+    this.webhookUrl = webhookUrl
+  }
+
+  private async post(payload: any): Promise<MattermostApiResponse> {
+    const res = await fetch(this.webhookUrl, {
```
mattkrick, line 25: this is my bad, we need to refactor the slack manager interface.
The best example of an interface we have today is AtlassianManager.
2 things to note:
- It has a `fetchWithTimeout` so if the request takes 2 minutes it won't bork our server (like Google did :cry: )
- we return an `Error` instance based on the gateway or payload error provided (ie the `ok` prop). This great because then whenever you call `post` you can look at the return value, check it's an error, & if so,  handle it & then you have a great strongly typed payload after that. Example usage here: https://github.com/ParabolInc/parabol/blob/31cc0e270ad29dd9b5a32fd55899eb6362426b09/packages/server/dataloader/atlassianLoaders.ts#L114-L118
- For your consideration: https://github.com/ParabolInc/parabol/blob/295c8fb078878d42a28c4d84bca4834b4d0d29a8/packages/client/utils/AtlassianManager.ts

packages/server/utils/MattermostServerManager.ts
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  async postMessage(textOrAttachmentsArray: string | unknown[], notificationText?: string) {
+    const prop =
+      typeof textOrAttachmentsArray === 'string'
+        ? 'text'
+        : Array.isArray(textOrAttachmentsArray)
+        ? 'attachments'
+        : null
+    if (!prop) throw new Error('Invalid matterMost message')
```
mattkrick, line 47: nit: all lowercase mattermost? 

for github, the rule I use is it's either `GitHub` or `github`, never `gitHub`.

mattkrick:
Loom video: https://www.loom.com/share/b77ab0fe44734e98b2a757396a238ec1

Be warned-- it's 1:45:00 :sweat_smile: if the comments make sense, no need to watch.

mattkrick (changes requested):

mattkrick:
fix #4412

(used to link the PR to the issue)

packages/client/modules/teamDashboard/components/ProviderRow/MattermostPanel.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const handleBlur = (e: React.FocusEvent<HTMLInputElement>) => setDirtyField(e.target.name)
+
+  return (
+    <MattermostPanelStyles>
+      <ConnectionGroup>
+        <Heading>Connection</Heading>
+      </ConnectionGroup>
+      <form onSubmit={onSubmit}>
+        <Row>
+          <Label onMouseEnter={openTooltip} onMouseLeave={closeTooltip} ref={originRef}>
```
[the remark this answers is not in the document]
jordanh: TIL!

packages/client/modules/teamDashboard/components/ProviderRow/MattermostPanel.tsx
```diff
@@ -0,0 +1,152 @@
...
+            Mattermost Webhook
+          </Label>
+          {tooltipPortal('Configure in Mattermost: Main Menu > Integrations > Incoming Webhook')}
+          <BasicInput
+            value={fields.webhookUrl.value}
+            error=''
+            onChange={onChange}
+            onBlur={handleBlur}
```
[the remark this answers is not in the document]
jordanh: Good call. Love this user empathy!

packages/client/modules/teamDashboard/components/ProviderRow/MattermostPanel.tsx
```diff
@@ -0,0 +1,152 @@
...
+          <StyledButton size='medium'>Update</StyledButton>
+        </Row>
+        {fieldError && <StyledError>{fieldError}</StyledError>}
+        {!fieldError && mutationError && <StyledError>{mutationError.message}</StyledError>}
+      </form>
+    </MattermostPanelStyles>
+  )
+}
+
+export default createFragmentContainer(MattermostConfigConnection, {
```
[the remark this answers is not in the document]
jordanh: I wasn't aware! That's what I get for cut-n-paste :)

Do you think it'd be helpful to global search and replace `createFragmentContainer` and add a comment above with `// createFragmentContainer deprecated` or stg?

packages/server/graphql/intranetSchema/mutations/runScheduledJobs.ts
jordanh: YAAAAS! I want to do this so badly. I decided to stop short in this PR because I have a similar proposal for this pattern that includes rolling up default oauth2 methods into the abstract class. I'd like to refactor the existing Slack & Mattermost integrations to take advantage.

Since I've already started this work in another branch, I don't want to pull all that into this PR.

packages/server/graphql/intranetSchema/mutations/runScheduledJobs.ts
jordanh: I couldn't get that to work. I needed to recast `job` itself, but do it based on what was in `jobProcessors`. I decided to do this instead:

```typescript
processor(job as Parameters<typeof processor>[0], {dataLoader}).catch(console.log)
```

packages/server/utils/MattermostServerManager.ts
jordanh: This is real, real nice. This `fetchWithTimeout` (and associating typing) definitely belongs on an abstract class. I'll lightly refactor what I have now and then pull this same pattern into the abstract integration manager class I'm working on

packages/server/graphql/intranetSchema/mutations/runScheduledJobs.ts
```diff
@@ -35,28 +30,52 @@ const processMeetingStageTimeLimits = async (job: ScheduledJobMeetingStageTimeLi
...
+  const {teamId, facilitatorUserId} = meeting
+  const {slackNotification, slackAuth} = await getSlackNotificationAndAuth(
+    teamId,
+    facilitatorUserId
+  )
+  const mattermostAuth = await dataLoader.get('mattermostAuthByTeamId').load(teamId)
+  const meetingUrl = makeAppURL(appOrigin, `meet/${meetingId}`)
+
+  let sendViaSlack = false
+  let sendViaMattermost = mattermostAuth?.isActive
```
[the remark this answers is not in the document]
jordanh: I missed this one the first time around, sorry!

packages/client/modules/teamDashboard/components/ProviderRow/MattermostProviderRow.tsx
```diff
@@ -0,0 +1,157 @@
...
+                menuProps={menuProps}
+                mutationProps={mutationProps}
+                teamId={teamId}
+              />
+            )}
+          </ListAndMenu>
+        ) : (
+          <ProviderActions>
+            <StyledButton
+              onClick={() => setConnectClicked(!isConnectClicked)}
```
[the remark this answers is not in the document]
jordanh: Slack stays open, I'm going to leave Mattermost at parity for now. I/we may refactor in the future as we enhance this integration

jordanh:
@mattkrick I had some time today to address most of your comments—including re-scoping the integration authorization to the `userId`:`teamId` pair.

I ran it through its paces, and I think it's ok. When you have a sec, give it another look and if you think it's ready, merge it back

jordanh:
N.B. this has been rebased against `master`

packages/client/modules/teamDashboard/components/ProviderRow/MattermostPanel.tsx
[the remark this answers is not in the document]
mattkrick: that definitely moves us from tribal knowledge to explicit! i like it, i'm a little afraid of executing the find & replace correctly, i suppose we just regex capture the whole line & then add something above it? i've never done that tbh, only codeshifts

packages/client/modules/teamDashboard/components/ProviderRow/MattermostProviderRow.tsx
```diff
@@ -0,0 +1,159 @@
...
+  flexDirection: 'column',
+  padding: 0
+})
+
+const MattermostProviderRow = (props: Props) => {
+  const {viewerRef, teamId} = props
+  const viewer = useFragment(
+    graphql`
+      fragment MattermostProviderRow_viewer on User {
+        ...MattermostProviderRowViewer @relay(mask: false)
```
mattkrick, line 90: here we've got a User fragment nested in a User fragment. We can simplify by removing the nesting.
 
let's inline the fragment below so we get

```gql
fragment MattermostProviderRow_viewer on User {
        ...MattermostPanel_viewer
        teamMember(teamId: $teamId) {
          integrations {
            mattermost {
              isActive
            }
          }
        }
      }
```

then in the Remove mutation, we just reference `MattermostProviderRow_viewer` instead

packages/server/graphql/intranetSchema/mutations/runScheduledJobs.ts
```diff
@@ -35,49 +30,62 @@ const processMeetingStageTimeLimits = async (job: ScheduledJobMeetingStageTimeLi
...
-  // if so, send the facilitator a slack notification
-  // if not, send the facilitator an in-app notification
+  // detect integrated services
+  // if slack, send slack
+  // if mattermost, send mattermost
+  // if no integrated notification services, send an in-app notification
+  const {meetingId} = job
+  const meeting = await dataLoader.get('newMeetings').load(meetingId)
+  const {teamId, facilitatorUserId} = meeting
+  const [{slackNotification, slackAuth}, {mattermostAuth}] = await Promise.all([
```
mattkrick, line 50: I think you want `mattermostAuth` not `{mattermostAuth}`

packages/server/graphql/intranetSchema/mutations/runScheduledJobs.ts
```diff
@@ -35,49 +30,62 @@ const processMeetingStageTimeLimits = async (job: ScheduledJobMeetingStageTimeLi
...
+    getSlackNotificationAndAuth(teamId, facilitatorUserId),
+    dataLoader.get('mattermostBestAuthByUserIdTeamId').load({userId: facilitatorUserId, teamId})
+  ])
+  const meetingUrl = makeAppURL(appOrigin, `meet/${meetingId}`)
+
+  const sendViaMattermost = mattermostAuth?.isActive
```
mattkrick, line 56: `isActive` is guaranteed to be true, so the following will always be true (i think the object destructure up above ruined the strong types. noImplicitAny will save us all soon 😄 )

```
  const sendViaMattermost = !!mattermostAuth
```

packages/server/graphql/mutations/helpers/notifications/makeMattermostAttachments.ts
```diff
@@ -0,0 +1,58 @@
...
+  fields: Array<Field>
+}
+
+export const makeFieldsAttachment = (
+  fields: Array<Field>,
+  header?: FieldAttachmentHeader
+): FieldAttachment => {
+  const defaultHeader = {
+    fallback: 'Parabol has a message for you',
+    author_icon: makeAppURL(appOrigin, `static/images/brand/mark-color%403x.png`),
```
mattkrick, line 38: this will grab the icon from our webserver instead of our CDN.
it'd be worthwhile making it grab the icon from our CDN to reduce traffic to our poor little webserver, especially because mattermost doesn't close the icon. they just fetch it from us every time it loads.

to do that, at the top of this file, simply import the image:
```
import authorIcon from '../../../../../../static/images/brand/mark-color@3x.png'
...
author_icon: authorIcon
```
webpack takes care of the rest 🎉  🪄

packages/server/graphql/mutations/helpers/notifications/notifyMattermost.ts
```diff
@@ -0,0 +1,309 @@
...
+}
+
+export const startMattermostMeeting = async (
+  meetingId: string,
+  teamId: string,
+  dataLoader: DataLoaderWorker
+) => {
+  const meeting = await dataLoader.get('newMeetings').load(meetingId)
+  const webhookUrl = await getWebhookForUserIdTeamId(meeting.facilitatorUserId, teamId, dataLoader)
+  if (!webhookUrl) return okResponse
```
mattkrick, line 101: since we're not using the `okResponse`, i'd probably just `return` to keep it simple

packages/server/graphql/mutations/helpers/notifications/notifyMattermost.ts
```diff
@@ -0,0 +1,309 @@
...
+          value: meetingName
+        },
+        {
+          short: false,
+          value: constraint
+        },
+        {
+          short: false,
+          title: 'Link',
+          value: `[https:/prbl.in/${meetingId}](${meetingUrl})`
```
mattkrick, line 265: `prbl.in` only works for our cloud offering, not PPMI or staging servers.
might i suggest `process.env.INVITATION_SHORTLINK`

packages/server/graphql/types/MattermostIntegration.ts
```diff
@@ -0,0 +1,33 @@
...
+      type: new GraphQLNonNull(GraphQLURLType),
+      description: 'the Mattermost server to integrate against'
+    },
+    teamId: {
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'The team that the token is linked to'
+    },
+    updatedAt: {
+      type: new GraphQLNonNull(GraphQLISO8601Type),
+      description: 'The timestamp the token was updated at'
```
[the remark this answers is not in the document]
mattkrick: agreed! let's add `createdAt`

packages/server/graphql/types/RemoveMattermostAuthPayload.ts
```diff
@@ -0,0 +1,28 @@
...
+import makeMutationPayload from './makeMutationPayload'
+
+export const RemoveMattermostAuthSuccess = new GraphQLObjectType<any, GQLContext>({
+  name: 'RemoveMattermostAuthSuccess',
+  fields: () => ({
+    teamId: {
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'The team with updated mattermost auth'
+    },
+    user: {
```
mattkrick, line 13: nit-- i like to offer foreign keys as well as the objects they get on mutation payloads, but that's probably prematurely optimizing for when we make our API public.

e.g. add `team, userId` to the fields

packages/server/postgres/queries/getMattermostAuthByUserIdTeamId.ts
```diff
@@ -0,0 +1,18 @@
+import getPg from '../getPg'
+import {
+  getMattermostAuthByUserIdTeamIdQuery,
+  IGetMattermostAuthByUserIdTeamIdQueryResult
+} from './generated/getMattermostAuthByUserIdTeamIdQuery'
+
+// this table has a composite primary key (userId, teamId),
+// which cannot use the index with a WHERE IN or JOIN on VALUES
+// so if we want to query multiple userIds/teamIds, just call this multiple times
```
mattkrick, line 9: fun fact, i tested this again, it's not true! postgres can use an index on a composite primary key if there are enough records for it to do so. my first test only had a few. i tested again with ~1000 records & it used the index. The more you know 🌠

packages/server/postgres/queries/getMattermostBestAuthByUserIdTeamId.ts
```diff
@@ -0,0 +1,15 @@
...
+  IGetMattermostBestAuthByUserIdTeamIdQueryResult
+} from './generated/getMattermostBestAuthByUserIdTeamIdQuery'
+
+export interface GetMattermostBestAuthByUserIdTeamIdResult
+  extends IGetMattermostBestAuthByUserIdTeamIdQueryResult {
+  isActive: true
+}
+const getMattermostBestAuthByUserIdTeamId = async (userId: string, teamId: string) => {
+  const [res] = await getMattermostBestAuthByUserIdTeamIdQuery.run({userId, teamId}, getPg())
+  return res as GetMattermostBestAuthByUserIdTeamIdResult
```
mattkrick, line 13:
```ts
return res as GetMattermostBestAuthByUserIdTeamIdResult | undefined
```

packages/server/postgres/queries/src/getMattermostBestAuthByUserIdTeamIdQuery.sql
```diff
@@ -0,0 +1,8 @@
+/*
+  @name getMattermostBestAuthByUserIdTeamIdQuery
+*/
+SELECT ("userId" = :userId)::int AS "userEquality", *
```
mattkrick, line 4: i like this a lot

packages/server/graphql/mutations/helpers/notifications/notifyMattermost.ts
```diff
@@ -0,0 +1,309 @@
...
+}
+
+export const notifyWebhookConfigUpdated = async (
+  webhookUrl: string,
+  userId: string,
+  teamId: string,
+  dataLoader: DataLoaderWorker
+) => {
+  const {name: teamName} = await dataLoader.get('teams').load(teamId)
+  const message = `Integration webhook configuration updated for team \`${teamName}\``
```
mattkrick, line 76: VERY nitty- i don't particularly like the team name in a code block. 
to avoid magical parsing, we can just inject a 0-width char in every word.
e.g. "My Team" becomes `M\u200by T\u200beam`

mattkrick (changes requested):

mattkrick:
Loom for part 2: https://www.loom.com/share/b810e920a7aa470a83303b3e1dff2da4

packages/server/utils/MattermostServerManager.ts
```diff
@@ -0,0 +1,94 @@
...
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+        Accept: 'application/json'
+      },
+      body: JSON.stringify(payload)
+    })
+    if (res instanceof Error) return res
+    // Mattermost returns text if all was ok, JSON error otherwise
+    if (!res.ok) {
```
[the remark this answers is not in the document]
mattkrick: sauce: https://developers.mattermost.com/integrate/incoming-webhooks/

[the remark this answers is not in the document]
jordanh: If you omit the `text` field from the JSON or you malform the JSON, you'll get a JSON response. It looks like:

```
HTTP/1.1 400 Bad Request
Content-Type: application/json
X-Request-Id: ci3gcjo557f3djxsa3nby6ftaa
X-Version-Id: 6.0.0.6.0.0.063d36a3f13988730f77b911e39d71eb.false
Date: Tue, 09 Nov 2021 03:59:31 GMT
Content-Length: 172

{"id":"model.incoming_hook.parse_data.app_error","message":"Unable to parse incoming data.","detailed_error":"","request_id":"ci3gcjo557f3djxsa3nby6ftaa","status_code":400}%
```

I'll make a change to parse the status code, for clarity if for nothing else

packages/client/modules/teamDashboard/components/ProviderRow/MattermostProviderRow.tsx
jordanh: This bugged me too!

I copied the pattern from here: https://github.com/ParabolInc/parabol/blob/2f90cb3592e9fedd8f0badcce9db90a6150dae5f/packages/client/modules/teamDashboard/components/ProviderRow/SlackProviderRow.tsx#L155

I figured, you had your reasons 😉 

I've gone ahead and made this a single fragment (but left Slack alone, for now)

packages/client/modules/teamDashboard/components/ProviderRow/MattermostPanel.tsx
```diff
@@ -0,0 +1,152 @@
...
+          </Label>
+          {tooltipPortal('Configure in Mattermost: Main Menu > Integrations > Incoming Webhook')}
+          <BasicInput
+            value={fields.webhookUrl.value}
+            error=''
+            onChange={onChange}
+            name='webhookUrl'
+            placeholder='https://my.mattermost.com:8065/hooks/abc123'
+          />
+          <StyledButton size='medium'>Update</StyledButton>
```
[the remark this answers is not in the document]
jordanh: Good call. Will do.

packages/server/graphql/intranetSchema/mutations/runScheduledJobs.ts
jordanh: Arrrrrrrrgh. Great catch.

packages/server/graphql/types/RemoveMattermostAuthPayload.ts
jordanh: I'm going to leave this alone for now, because this mutation will go away on the generalized interface I've cooked up

packages/server/postgres/queries/getMattermostAuthByUserIdTeamId.ts
jordanh: I'll strike the comment in both places!

jordanh: (granted, the better thing to do would be to refactor this to fetch multiple...but I am going to leave this alone this will change in a subsequent change set)

packages/server/graphql/mutations/helpers/notifications/notifyMattermost.ts
jordanh: Great review comment. Ugh, passing user input to a 3rd-party API should have raised sanitization warning bells for me. I'll just strike it for now. It's not a big add, obvi

jordanh:
Note: forced commit rebased to `master`

jordanh:
@mattkrick thank you for the 2nd review, merge if you think this is ready

packages/server/graphql/mutations/helpers/notifications/notifyMattermost.ts
```diff
@@ -0,0 +1,298 @@
...
+  webhookUrl: string,
+  userId: string,
+  teamId: string,
+  textOrAttachmentsArray: string | unknown[],
+  notificationText?: string
+) => {
+  const manager = new MattermostServerManager(webhookUrl)
+  const result = await manager
+    .postMessage(textOrAttachmentsArray, notificationText)
+    .catch((err) => new Error(`Mattermost error: ${err.message}`))
```
mattkrick, line 43: the only error that throws on this is:

```ts
    if (!prop) throw new Error('Invalid mattermost message')
```

which isn't a mattermost error, it's an error we make due to a bad `textOrAttachmentsArray`.

let's have `postMessage` always return an error & never throw & then we can just catch that in the line below

packages/server/graphql/mutations/helpers/notifications/notifyMattermost.ts
mattkrick: still a bug!

packages/server/postgres/queries/getMattermostAuthByUserIdTeamId.ts
```diff
@@ -0,0 +1,16 @@
...
+} from './generated/getMattermostAuthByUserIdTeamIdQuery'
+
+// if we want to query multiple userIds/teamIds, just call this multiple times
+export interface GetMattermostAuthByUserIdTeamIdResult
+  extends IGetMattermostAuthByUserIdTeamIdQueryResult {
+  isActive: true
+}
+const getMattermostAuthByUserIdTeamId = async (userId: string, teamId: string) => {
+  const [res] = await getMattermostAuthByUserIdTeamIdQuery.run({userId, teamId}, getPg())
+  return res as GetMattermostAuthByUserIdTeamIdResult
```
mattkrick, line 14: `return res as GetMattermostAuthByUserIdTeamIdResult | undefined`

packages/server/utils/MattermostServerManager.ts
```diff
@@ -0,0 +1,95 @@
...
+      clearTimeout(timeout)
+      return res
+    } catch (e) {
+      clearTimeout(timeout)
+      return new Error('Mattermost is not responding')
+    }
+  }
+
+  private async post(payload: any) {
+    const res = await this.fetchWithTimeout(this.webhookUrl, {
```
mattkrick, line 44: let's add a comment pointing to the mattermost webhook docs: 

```ts
    // https://developers.mattermost.com/integrate/incoming-webhooks/
```

packages/server/utils/MattermostServerManager.ts
```diff
@@ -0,0 +1,95 @@
...
+        'Content-Type': 'application/json',
+        Accept: 'application/json'
+      },
+      body: JSON.stringify(payload)
+    })
+    if (res instanceof Error) return res
+    if (res.status !== 200) {
+      if (res.status === 400) {
+        // Bad request, Matter supplies a JSON error message
+        const {message: error} = await res.json()
```
mattkrick, line 56: how do we reproduce a 400 error with a json payload? 
i want to make sure that a 400 error has a `json` body payload & not a `text` payload otherwise this method will throw & we'll need to add a catch

packages/server/utils/MattermostServerManager.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  async postMessage(textOrAttachmentsArray: string | unknown[], notificationText?: string) {
+    const prop =
+      typeof textOrAttachmentsArray === 'string'
+        ? 'text'
+        : Array.isArray(textOrAttachmentsArray)
+        ? 'attachments'
+        : null
+    if (!prop) throw new Error('Invalid mattermost message')
```
mattkrick, line 72: let's return the error here too since this.post will also return an Error instance

packages/server/graphql/mutations/helpers/notifications/makeMattermostAttachments.ts
mattkrick: still a bug!

mattkrick (changes requested):
couple new things, couple old

mattkrick:
loom part 3: https://www.loom.com/share/90a29bc7c634423b86174455cbe5e029

jordanh:
@mattkrick Loom on Mattermost error messages. Oh, the joy!

https://www.loom.com/share/cf1408c5d759470ba4297405104cb0a2

packages/server/graphql/mutations/helpers/notifications/notifyMattermost.ts
jordanh: Fixed!

packages/server/utils/MattermostServerManager.ts
jordanh: ICYMI: https://www.loom.com/share/cf1408c5d759470ba4297405104cb0a2

jordanh:
Ok @mattkrick, done with my latest (last?) round of changes. Please have a look, and, take care to look at the Webpack configs especially. On my end I tested:

- yarn start (with CDN)
- yarn start (without CDN)
- yarn dev (with CDN)
- yarn dev (without CDN)

And all looked ok. I also made sure to test the jiraImagesHandler placeholder image. You'll see there I decided to take the easy way out.

I'm hopeful this can be merged so I can rebase these changes over to the `add-gitlab-auth` branch and get the team working on it. You'd have my thanks if you find anything small to make the change.

TY!

mattkrick (approved):

PR #5552, opened 2021-10-20
Better logging for timeout in publishInternalGQL

### LINKED ISSUE #5189: Gracefully handle slow responding GQL Executor

tianrunhe:
Resolves #5189 .

The Sentry error reports that we still have the TIMEOUT message. That's because in `publishInternalGQL` we just send to Sentry directly when we have a timeout & rejected promise from [PubSubPromise](https://github.com/ParabolInc/parabol/blob/18db1e79655749880c5de3de45ac8ed7e482f696/packages/server/utils/PubSubPromise.ts#L45).

The commit simply add the same set of logging in `publishInternalGQL` so we can aggregate these events in Sentry.

@tiffanyhan Can you take a look? Should be a simple & small one.

### DIFF
--- a/packages/server/graphql/handleGraphQLTrebuchetRequest.ts
+++ b/packages/server/graphql/handleGraphQLTrebuchetRequest.ts
@@ -59,6 +59,7 @@ const handleGraphQLTrebuchetRequest = async (
     } catch (e) {
       if (e instanceof Error && e.message === 'TIMEOUT') {
         sendToSentry(new Error('GQL executor took too long to respond'), {
+          userId: getUserId(authToken),
           tags: {
             authToken: JSON.stringify(authToken),
             docId: docId || '',


tianrunhe:
Oops wrong issue! There are two issues in this sprint whose title starts with "gracefully..." LOL

PR #5561, opened 2021-10-22
Allow to push integration tasks with assignee's and own integration

Dschoordsch:
When pushing a task that is not assigned to the viewer, we try to reuse the assignee's integration if they have one and fall back to the viewer's. In this case we also add a comment in the new Jira/GitHub issue stating that this was pushed by someone else.

This PR does not address other adjustments about reusing integrations mentioned in comments in #5342.

### Setup
* create 3 users Alice, Bob, Clarissa in Team A
* only Alice sets up GitHub and Jira integrations for Team A
* Alice creates 4 tasks: Alice-GitHub-A, -B and Alice-Jira-A, -B
* Bob creates 3 tasks, Bob-GitHub-A, Bob-Jira-A and Bob-none
### Verification
* [ ] Alice can push Alice-GitHub-A and Alice-Jira-A to GitHub and Jira
* [ ] Alice can push Bob-GitHub-A and Bob-Jira-A
  * [ ] the menu shows 'Push using your credentials'
  * [ ] the tickets contain a comment 'Created by Alice for Bob' and a link to Team A
* [ ] Bob can push Alice-GitHub-B and Alice-Jira-B
  * [ ] the menu shows 'Push as Alice'
  * [ ] the tickets contain a comment 'Created by Bob for Alice' and a link to Team A
* [ ] Bob tries to push Bob-none
  * [ ] menu shows add Jira and add GitHub
* [ ] Clarissa tries to push Bob-none
  * [ ] menu shows add Jira and add GitHub
  * [ ] menu shows message 'Neither Bob nor you have integrations set up'
* [ ] Alice, Bob and Clarissa can see correct  *-GitHub-* and *-Jira-* tickets correctly

Resolves: #5342


packages/client/components/TaskFooterIntegrateMenu.tsx
```diff
@@ -93,6 +136,16 @@ export default createFragmentContainer(TaskFooterIntegrateMenu, {
             ...TaskFooterIntegrateMenuViewerGitHubIntegration @relay(mask: false)
           }
         }
+      }
+      membership: teamMember(userId: null, teamId: $teamId) {
```
Dschoordsch, line 140: will do `useFragment` refactoring separately

Dschoordsch:
@BartoszJarocki could you please take a look?

Dschoordsch:
@enriquesanchez I modified the push to integration dialog in the following cases
* Viewer has integration set up, assignee does not
Screenshot from 2021-11-01 16-04-24
* Assignee has integration set up, viewer does not
Screenshot from 2021-11-01 16-05-48
* Neither viewer nor assignee have integrations set up
Screenshot from 2021-11-01 16-06-39

Would be nice if you could give me feedback on those.

enriquesanchez (changes requested):
Thanks for the ping, @Dschoordsch 🙌

These look good! Do you think we could use the following style for the clarifying message?

image

We use something very similar in other popups, like the team dropdown. I'm using `#82809A` and regular style for the text (better contrast and readability). The separator line helps with clarity by keeping the message separated from the rest.

And another comment is on the message `Neither you nor Georg has any integrations for this team`. I believe we should use `have` instead of `had` because it's plural.

Dschoordsch:
@enriquesanchez  Thanks for the fast feedback! I will change the style.

For the have/has, to me have sounds nicer as well, but I googled some time and always found that the subject closes to the verb determines it's case, for example https://editorsmanual.com/articles/neither-singular-or-plural/
If you can think of a better phrase to avoid this altogether, I'm all ears.

enriquesanchez:
>For the have/has, to me have sounds nicer as well, but I googled some time and always found that the subject closes to the verb determines it's case

@Dschoordsch Oh, I did not know this! Thanks for clarifying. I agree it sounds nicer with `have`, but if this is the way... then... it's the way 😃

Dschoordsch:
@enriquesanchez now looks like this:
Screenshot from 2021-11-02 10-31-46
Screenshot from 2021-11-02 10-39-11
Screenshot from 2021-11-02 10-47-43

enriquesanchez (approved):
Looks great! 👍

packages/server/utils/makeCreateGitHubTaskComment.ts
```diff
@@ -0,0 +1,15 @@
...
+const makeCreateGitHubTaskComment = (
+  creator: string,
+  assignee: string,
+  teamName: string,
+  teamDashboardUrl: string
+) =>
+  `Created by ${creator} for ${assignee}
+  See the dashboard of [${teamName}](${teamDashboardUrl})
+
+  *Powered by [Parabol](${ExternalLinks.INTEGRATIONS_JIRA})*`
```
BartoszJarocki, line 13: should be `INTEGRATIONS_GITHUB` 🕵️‍♂️

packages/server/utils/makeCreateJiraTaskComment.ts
```diff
@@ -0,0 +1,71 @@
...
+          ]
+        },
+        {
+          type: 'text',
+          text: 'Parabol',
+          marks: [
+            {
+              type: 'link',
+              attrs: {
+                href: ExternalLinks.GETTING_STARTED_CHECK_INS
```
BartoszJarocki, line 58: Shouldn't this be `INTEGRATIONS_JIRA`?

packages/server/utils/makeCreateJiraTaskComment.ts
```diff
@@ -0,0 +1,71 @@
+import {Doc} from './convertContentStateToADF'
+import {ExternalLinks} from '../../client/types/constEnums'
+
+const makeCreateJiraTaskComment = (
+  creator: string,
+  assignee: string,
+  teamName: string,
+  teamDashboardUrl: string
+): Doc => ({
```
BartoszJarocki, line 9: nit: We try to avoid specifying return type as it's easier to maintain in a long term. TS will infer the type when the return type has a type specified, for example
```
const makeCreateJiraTaskComment = (
  creator: string,
  assignee: string,
  teamName: string,
  teamDashboardUrl: string
) =>
  ({
    version: 1 as const,
    type: 'doc' as const,
    content: [
      {
        type: 'paragraph',
        content: [
          {
            type: 'text',
            text: `Created by ${creator} for ${assignee}`
          },
          {
            type: 'hardBreak'
          },
          {
            type: 'text',
            text: 'See the dashboard of '
          },
          {
            type: 'text',
            text: teamName,
            marks: [
              {
                type: 'link',
                attrs: {
                  href: teamDashboardUrl
                }
              }
            ]
          },
          {
            type: 'hardBreak'
          },
          {
            type: 'text',
            text: 'Powered by ',
            marks: [
              {
                type: 'em'
              }
            ]
          },
          {
            type: 'text',
            text: 'Parabol',
            marks: [
              {
                type: 'link',
                attrs: {
                  href: ExternalLinks.GETTING_STARTED_CHECK_INS
                }
              },
              {
                type: 'em'
              }
            ]
          }
        ]
      }
    ]
  } as Doc)
```

packages/server/utils/makeScoreJiraComment.ts
```diff
@@ -1,11 +1,12 @@
 import {ExternalLinks} from '../../client/types/constEnums'
+import {Doc} from './convertContentStateToADF'
 
 const makeScoreJiraComment = (
   dimensionName: string,
   finalScore: string,
   meetingName: string,
   discussionURL: string
-) => ({
+): Doc => ({
```
BartoszJarocki, line 9: Same as comment as for `makeCreateJiraTaskComment`

BartoszJarocki (approved):
Great job, all the scenarios work as expected. Only a couple of minor comments, after that let's move it to maintainer review.

One question, there's a small difference in the way we handle Jira vs Github integration. ie. Jira has this condition: 

```
if (!userId) {
      // we should probably remove this constraint in the future
      return {error: {message: 'Task must have assignee before it can be integrated with Jira'}}
    }
```
any reason to keep it?

packages/server/utils/makeCreateJiraTaskComment.ts
Dschoordsch: > Great job, all the scenarios work as expected. Only a couple of minor comments, after that let's move it to maintainer review.
> 
> One question, there's a small difference in the way we handle Jira vs Github integration. ie. Jira has this condition:
> 
> ```
> if (!userId) {
>       // we should probably remove this constraint in the future
>       return {error: {message: 'Task must have assignee before it can be integrated with Jira'}}
>     }
> ```
> 
> any reason to keep it?

Probably not. It was added [here](https://github.com/ParabolInc/parabol/pull/4312/files#diff-13ccf9270fcf560f88b8b298fced28727d4aa6e20731d7537f7d18d39ad8d896) without any explanation in PR or commit message. @mattkrick can this be removed?

Dschoordsch:
> Great job, all the scenarios work as expected. Only a couple of minor comments, after that let's move it to maintainer review.
> 
> One question, there's a small difference in the way we handle Jira vs Github integration. ie. Jira has this condition:
> 
> ```
> if (!userId) {
>       // we should probably remove this constraint in the future
>       return {error: {message: 'Task must have assignee before it can be integrated with Jira'}}
>     }
> ```
> 
> any reason to keep it?

Probably not. It was added [here](https://github.com/ParabolInc/parabol/pull/4312/files#diff-13ccf9270fcf560f88b8b298fced28727d4aa6e20731d7537f7d18d39ad8d896) without any explanation in PR or commit message. @mattkrick can this be removed?

packages/server/utils/makeCreateGitHubTaskComment.ts
Dschoordsch: Good catch, I especially added this constant :facepalm:

packages/server/utils/makeCreateJiraTaskComment.ts
BartoszJarocki: oh, nice! I didn't know that. 👌

packages/client/components/TaskFooterIntegrateMenu.tsx
```diff
@@ -93,6 +137,16 @@ export default createFragmentContainer(TaskFooterIntegrateMenu, {
...
+        integrations {
```
mattkrick, line 142: let's break this into its own team member fragment so we can repeat it in both places

e.g. 

```gql
fragment TaskFooterIntegrateMenuTeamMember on TeamMember {
  integrations {
atlassian {
            ...TaskFooterIntegrateMenuViewerAtlassianIntegration @relay(mask: false)
          }
          github {
            ...TaskFooterIntegrateMenuViewerGitHubIntegration @relay(mask: false)
          }
  }
}
```

packages/client/components/TaskFooterIntegrateMenu.tsx
```diff
@@ -23,38 +22,83 @@ const makePlaceholder = (hasGitHub: boolean, hasAtlassian: boolean) => {
...
+        hasAtlassian,
+        hasGitHub
+      }
+    : null
+}
+
 const TaskFooterIntegrateMenu = (props: Props) => {
   const {menuProps, mutationProps, task, viewer} = props
-  const {id: viewerId, teamMember} = viewer
+  const {id: viewerId, membership, teamMember} = viewer
```
mattkrick, line 41: nit-- let's rename teamMember to assigneeTeamMember or something similar & membership to viewerTeamMember

packages/server/graphql/mutations/createGitHubTaskIntegration.ts
```diff
@@ -68,44 +71,77 @@ export default {
...
+    const auth = assigneeAuth ?? viewerAuth
+    if (!auth) {
       return standardError(
-        new Error(`Assignment failed! The assignee does not have access to GitHub`),
+        new Error(`Assignment failed! Neither you nor the assignee has access to GitHub`),
         {userId: viewerId}
       )
     }
+    const teamMembers = await dataLoader.get('teamMembersByTeamId').load(teamId)
+    const viewer = teamMembers.find(({userId}) => userId === viewerId)
```
mattkrick, line 88: nit-- we can call this viewerTeamMember since viewer usually refers to a User object

packages/server/graphql/mutations/createGitHubTaskIntegration.ts
```diff
@@ -68,44 +71,77 @@ export default {
...
+    const assignee = userId && teamMembers.find((user) => user.userId === userId)
+
+    if (!viewer || (userId && !assignee)) {
```
mattkrick, line 91: we already know the viewer is on the team because we checked that in L56

mattkrick: task.userId is guaranteed to be an active team member because that is validated in the create/update task mutations. i think we can remove this validation check

packages/server/graphql/mutations/createGitHubTaskIntegration.ts
```diff
@@ -68,44 +71,77 @@ export default {
...
+          userId: userId ?? '',
+          viewerId,
+          teamId,
+          taskId
+        }
+      })
+    }
+
+    // RESOLUTION
+    const accessUserId = assigneeAuth && userId ? userId : viewerId
```
mattkrick, line 104: the person calling the mutation is the person who likely has the most access to jira (i know this logic doesn't apply to github, but if they change permissions we'll be ready. plus, it's just nice to have both integrations using the same logic chain)

. for example, maybe the assignee can't create tasks in jira?

let's favor using the viewer's auth first:

```ts
const accessUserId = viewerAuth ? viewerId : userId
```

packages/server/graphql/mutations/createGitHubTaskIntegration.ts
```diff
@@ -68,44 +71,77 @@ export default {
...
+    ])
+    const auth = assigneeAuth ?? viewerAuth
+    if (!auth) {
       return standardError(
-        new Error(`Assignment failed! The assignee does not have access to GitHub`),
+        new Error(`Assignment failed! Neither you nor the assignee has access to GitHub`),
         {userId: viewerId}
       )
     }
+    const teamMembers = await dataLoader.get('teamMembersByTeamId').load(teamId)
```
mattkrick, line 87: can move to the promise.all up above

packages/server/graphql/mutations/createJiraTaskIntegration.ts
```diff
@@ -69,15 +72,55 @@ export default {
       )
     }
 
-    const viewerAuth = await dataLoader.get('freshAtlassianAuth').load({teamId, userId: viewerId})
-    if (!viewerAuth) {
-      return standardError(new Error('The assignee does not have access to Jira'), {
+    const [viewerAuth, assigneeAuth, team] = await Promise.all([
+      dataLoader.get('freshAtlassianAuth').load({teamId, userId: viewerId}),
+      dataLoader.get('freshAtlassianAuth').load({teamId, userId}),
```
mattkrick, line 77: let's remove 
```
    if (!userId) {
      // we should probably remove this constraint in the future
      return {error: {message: 'Task must have assignee before it can be integrated with Jira'}}
    }
```

and then we can do the same checks here that you do for the github version:
- check userId exists before calling the dataloader
- check userId exists before createdBySomeoneElseComment

packages/server/graphql/mutations/createJiraTaskIntegration.ts
```diff
@@ -69,15 +72,55 @@ export default {
...
+      dataLoader.get('teams').load(teamId)
+    ])
+    const auth = assigneeAuth ?? viewerAuth
+    if (!auth) {
+      return standardError(new Error('Neither you nor the assignee has access to Jira'), {
         userId: viewerId
       })
     }
+    const teamMembers = await dataLoader.get('teamMembersByTeamId').load(teamId)
```
mattkrick, line 86: can move to promise.all above

packages/server/graphql/mutations/createJiraTaskIntegration.ts
```diff
@@ -69,15 +72,55 @@ export default {
...
+    const viewer = teamMembers.find(({userId}) => userId === viewerId)
+    const assignee = teamMembers.find((user) => user.userId === userId)
+
+    if (!viewer || !assignee) {
```
mattkrick, line 90: just like the gh file, these are guaranteed to be true when their values are set, no need to re-validate the already validated data

packages/server/graphql/mutations/createJiraTaskIntegration.ts
```diff
@@ -69,15 +72,55 @@ export default {
...
+          viewerId,
+          teamId,
+          taskId
+        }
+      })
+    }
 
     // RESOLUTION
-    const res = await createJiraTask(rawContentStr, cloudId, projectKey, viewerAuth)
+    const accessUserId = assigneeAuth ? userId : viewerId
```
mattkrick, line 103: let's favor viewerId instead of assigneeId

packages/server/utils/makeCreateGitHubTaskComment.ts
```diff
@@ -0,0 +1,15 @@
+import {ExternalLinks} from '../../client/types/constEnums'
+
+// wrap anything that looks like an issue in ticks so it doesn't turn into a link to that issue
+const makeCreateGitHubTaskComment = (
+  creator: string,
+  assignee: string,
+  teamName: string,
+  teamDashboardUrl: string
+) =>
+  `Created by ${creator} for ${assignee}
```
mattkrick, line 10: let's sanitize the creator & assignee name in case they have a # in their name followed by a number:

```
.replace(/#(\d+)/g, '#​\u200b$1')
```

mattkrick (changes requested):
mostly little stuff,
for the permissions, let's try to use the viewer's credentials before falling back to the assignee's credentials. it's always if people can be themselves 🙂

mattkrick:
loom: https://www.loom.com/share/570760e8be494f0aa4dbbbdaec800c05

packages/server/graphql/mutations/createGitHubTaskIntegration.ts
Dschoordsch: I used assignee first because then the assignee matches in GitHub/Jira. Can do it the other way round.

packages/server/graphql/mutations/createGitHubTaskIntegration.ts
```diff
@@ -68,44 +71,62 @@ export default {
...
     }
 
-    // RESOLUTION
     const {content: rawContentStr, meetingId} = task
-    const viewerAuth = await dataLoader.get('githubAuth').load({teamId, userId: viewerId})
-
-    if (!viewerAuth) {
+    const [viewerAuth, assigneeAuth, team, teamMembers] = await Promise.all([
+      dataLoader.get('githubAuth').load({teamId, userId: viewerId}),
+      userId && dataLoader.get('githubAuth').load({teamId, userId}),
```
mattkrick, line 77: if you use a ternary here you'll get strong types returned from Promise.all, otherwise everything is `unknown`

```ts
userId ? dataLoader(...) : null
```

packages/server/graphql/mutations/createGitHubTaskIntegration.ts
```diff
@@ -68,44 +71,62 @@ export default {
...
+      dataLoader.get('teams').load(teamId),
+      dataLoader.get('teamMembersByTeamId').load(teamId)
+    ])
+    const auth = viewerAuth ?? assigneeAuth
+    if (!auth) {
```
mattkrick, line 82: we have a tight coupling between userId & assigneeAuth & we should tell typescript about that so we don't have to do a nonnullish coercion down below on L122.

```ts
const accessUserId = viewerAuth ? viewerId : assigneeAuth ? userId : null
if (!accessUserId) {
```

then `accessUserId` is guaranteed non-null, which is what we'd expect

packages/server/graphql/mutations/createJiraTaskIntegration.ts
```diff
@@ -69,15 +67,41 @@ export default {
       )
     }
 
-    const viewerAuth = await dataLoader.get('freshAtlassianAuth').load({teamId, userId: viewerId})
-    if (!viewerAuth) {
-      return standardError(new Error('The assignee does not have access to Jira'), {
+    const [viewerAuth, assigneeAuth, team, teamMembers] = await Promise.all([
+      dataLoader.get('freshAtlassianAuth').load({teamId, userId: viewerId}),
+      userId && dataLoader.get('freshAtlassianAuth').load({teamId, userId}),
```
mattkrick, line 72: ternary for strong types, a short circuit could result in the value of `false` which isn't a `PromiseLike` type


```ts
// from lib.d.ts
    then<TResult1 = T, TResult2 = never>(onfulfilled?: ((value: T) => TResult1 | PromiseLike<TResult1>) | undefined | null, onrejected?: ((reason: any) => TResult2 | PromiseLike<TResult2>) | undefined | null): PromiseLike<TResult1 | TResult2>;
```

packages/server/graphql/mutations/createJiraTaskIntegration.ts
```diff
@@ -69,15 +67,41 @@ export default {
...
+      dataLoader.get('teams').load(teamId),
+      dataLoader.get('teamMembersByTeamId').load(teamId)
+    ])
+    const auth = viewerAuth ?? assigneeAuth
+    if (!auth) {
```
mattkrick, line 77: same pattern here so accessUserId is nonnull

mattkrick (changes requested):
almost perfect! just a couple minor things & then merge away

packages/server/graphql/mutations/createGitHubTaskIntegration.ts
Dschoordsch: That's a good tip!

PR #5577, opened 2021-10-26
Extend GQL Timeout for GraphiQL Queries

mattkrick:
Base: feat/scale-gql-executor
Based off this branch to avoid a merge conflict.

### Why
If the request comes from GraphiQL, the timeout is now 1 minute.
This makes the UI nice for things like DB backups & moving teams to different orgs.

TEST
- [ ] call `backupOrganization` on the graphiql private schema. it should return a nice looking response instead of a bad timeout message


packages/server/utils/PubSubPromise.ts
```diff
@@ -45,10 +47,12 @@ export default class PubSubPromise<Request extends BaseRequest, Response> {
     return new Promise<Response>((resolve, reject) => {
       const nextJob = numToBase64(this.jobCounter++)
       const jobId = `${SERVER_ID}:${nextJob}`
+      const {isAdHoc} = request
+      const timeout = isAdHoc ? ms('1m') : MAX_TIMEOUT
```
Dschoordsch, line 51: nit: you could use a constant `ADHOC_MAX_TIMEOUT` here

Dschoordsch (approved):

PR #5590, opened 2021-10-29
Added updateOAuthRefreshTokens mutation

### LINKED ISSUE #5510: Add cronjob to update atlassian refresh token
The atlassian refresh token has a 30 day lifespan.
after that, the user will likely see an error & have to re-integrate.
to avoid this, we should update the refresh tokens every 14 days.
we'll do this by having an internal graphql mutation that finds all AtlassianAuth rows that are older than 14 days & refresh them.
we'll call this mutation every 14 days using action-chronos.

AC
- [ ] on the intranet schema there exists a mutation called `updateOAuthRefreshTokens`
- [ ] in that mutation, it queries all AtlassianAuth rows where updatedAt is older than 14 days
- [ ] for each returned row, it calls the `freshAtlassianAuth` dataloader
- [ ] there is a cron job that calls the mutation every 14 days. That lives here https://github.com/ParabolInc/action-chronos/blob/master/src/jobs/index.js
- [ ] AtlassianServerManager.refreshToken is nonnullable
Estimate: 8 hours

BartoszJarocki:
Fixes #5510. 
Cron job PR: https://github.com/ParabolInc/action-chronos/pull/6 

Tests:
1. Integrate Atlassian for a few users
2. Change the `updatedAt` to be more than 14 days ago
3. Open http://localhost:3000/admin/graphql 
4. Call the mutation below
```
mutation {
  updateOAuthRefreshTokens
}
```
5. Check if the integer output equals the number of rows which supposed to be updated

AC:
- [x] `updateOAuthRefreshTokens` exists in internal schema
- [ ] `updateOAuthRefreshTokens` when called updates all the AtlassianAuth rows older than 14 days
- [x] `updateOAuthRefreshTokens` can be called only by admins
- [ ] `updateOAuthRefreshTokens` is called 2 times per month by cron job
- [ ] Atlassian integration still works correctly


packages/server/dataloader/foreignLoaderMakers.ts
```diff
@@ -43,28 +43,7 @@ export const agendaItemsByMeetingId = new LoaderMakerForeign(
...
-  'atlassianAuths',
-  'teamId',
-  async (teamIds) => {
-    const r = await getRethink()
-    return r
-      .table('AtlassianAuth')
-      .getAll(r.args(teamIds), {index: 'teamId'})
-      .run()
-  }
-)
```
BartoszJarocki, line 67: I've searched for usages and I found none so I removed them

packages/server/graphql/types/AtlassianIntegration.ts
```diff
@@ -8,7 +8,6 @@ import {
   GraphQLString
 } from 'graphql'
 import ms from 'ms'
-import AtlassianAuth from '../../database/types/AtlassianAuth'
```
BartoszJarocki, line 11: I changed the type we use from the old Rethink model to the type used by PG. In PG type both access and refresh token is not nullable so it simplifies the code.

packages/server/graphql/types/AtlassianIntegration.ts
```diff
@@ -108,12 +108,7 @@ const AtlassianIntegration = new GraphQLObjectType<any, GQLContext>({
           standardError(err, {tags: {teamId, userId}, userId: viewerId})
           return connectionFromTasks([], 0, err)
         }
-        if (!accessToken) {
-          const err = new Error('Not integrated with Jira')
-          standardError(err, {tags: {teamId, userId}, userId: viewerId})
-          return connectionFromTasks([], 0, err)
-        }
-        const manager = new AtlassianServerManager(accessToken)
```
BartoszJarocki, line 116: Even though I've tested all the scenarios I thought of and it all works I still do not feel confident about that change. Technically, this field is not nullable, am I missing some scenario in which this condition would be needed?

packages/server/graphql/types/TeamMemberIntegrations.ts
```diff
@@ -19,10 +20,7 @@ const TeamMemberIntegrations = new GraphQLObjectType<any, GQLContext>({
       description: 'All things associated with an atlassian integration for a team member',
       resolve: async ({teamId, userId}, _args, {authToken, dataLoader}) => {
         if (!isTeamMember(authToken, teamId)) return null
-        const atlassianIntegration = await dataLoader
-          .get('freshAtlassianAuth')
-          .load({teamId, userId})
-        return atlassianIntegration
+        return dataLoader.get('freshAtlassianAuth').load({teamId, userId})
```
BartoszJarocki, line 23: I've just inlined the variable here

packages/server/dataloader/atlassianLoaders.ts
```diff
@@ -35,15 +35,10 @@ export const freshAtlassianAuth = (parent: RethinkDataLoader) => {
...
         keys.map(async ({userId, teamId}) => {
           const atlassianAuth = await getAtlassianAuthByUserIdTeamId(userId, teamId)
-
-          if (!atlassianAuth?.refreshToken) {
-            // Not always an error! For suggested integrations, this won't exist.
-            // sendToSentry(new Error('No atlassian access token exists for team member'), {
-            //   userId,
-            //   tags: {teamId}
-            // })
+          if (!atlassianAuth) {
```
BartoszJarocki, line 38: The result returned by `getAtlassianAuthByUserIdTeamId` is `Promise<null | AtlassianAuth>` so we definitely need a null check here. On the other hand, refresh token is not nullable anymore so I don't see any reason to keep the refresh token check. I'm happy to improve if I'm missing something though!

packages/server/dataloader/atlassianLoaders.ts
```diff
@@ -63,8 +58,7 @@ export const freshAtlassianAuth = (parent: RethinkDataLoader) => {
           return atlassianAuth
         })
       )
-      const res = results.map((result) => (result.status === 'fulfilled' ? result.value : null))
-      return res
+      return results.map((result) => (result.status === 'fulfilled' ? result.value : null))
```
BartoszJarocki, line 61: Inlined the thing

packages/server/graphql/types/AtlassianIntegration.ts
```diff
@@ -40,7 +40,7 @@ const AtlassianIntegration = new GraphQLObjectType<any, GQLContext>({
       description:
         'The access token to atlassian, useful for 1 hour. null if no access token available or the viewer is not the user',
       type: GraphQLID,
-      resolve: async ({accessToken, userId}, _args, {authToken}) => {
+      resolve: async ({accessToken, userId}: AtlassianAuth, _args, {authToken}) => {
```
BartoszJarocki, line 43: Added types in a few places

BartoszJarocki:
@tianrunhe could you please take a look? 🙏

packages/server/graphql/intranetSchema/intranetSchema.ts
```diff
@@ -92,7 +93,8 @@ const mutation = new GraphQLObjectType<any, GQLContext>({
       stripeUpdateInvoiceItem,
       stripeInvoiceFinalized,
       updateEmail,
-      updateWatchlist
+      updateWatchlist,
+      updateOAuthRefreshTokens
```
tianrunhe, line 97: Maybe a better name is `updateAtlassianOAuthRefreshTokens` in case we need to update GitHub in the future?

tianrunhe (changes requested):
Great work! The only issue I've encountered during testing is error handling. When [this line](https://github.com/ParabolInc/parabol/blob/feat/5510/updateOAuthRefreshTokens/packages/server/dataloader/atlassianLoaders.ts#L47) throws an error, there's no catch. Below are the errors I've been able to catch on my local:

```
tokenJson {
  error: 'invalid_grant',
  error_description: 'Unknown or invalid refresh token.'
}
tokenJson {
  error: 'invalid_grant',
  error_description: 'The client associated with this refresh token (WWY65WgtWwCypZVkmcRxqfiUKAZ0Tioj) is different than the one sent in the request (l9HwJj6ubcL1zxKbhwa3KiwBwGz9SD90).'
}
tokenJson {
  error: 'invalid_grant',
  error_description: 'The client associated with this refresh token (WWY65WgtWwCypZVkmcRxqfiUKAZ0Tioj) is different than the one sent in the request (l9HwJj6ubcL1zxKbhwa3KiwBwGz9SD90).'
}
tokenJson {
  error: 'invalid_grant',
  error_description: 'Unknown or invalid refresh token.'
}
tokenJson {
  error: 'invalid_grant',
  error_description: 'The client associated with this refresh token (WWY65WgtWwCypZVkmcRxqfiUKAZ0Tioj) is different than the one sent in the request (l9HwJj6ubcL1zxKbhwa3KiwBwGz9SD90).'
}
tokenJson {
  error: 'invalid_grant',
  error_description: 'The client associated with this refresh token (WWY65WgtWwCypZVkmcRxqfiUKAZ0Tioj) is different than the one sent in the request (l9HwJj6ubcL1zxKbhwa3KiwBwGz9SD90).'
}
tokenJson {
  error: 'invalid_grant',
  error_description: 'Unknown or invalid refresh token.'
}
tokenJson {
  error: 'invalid_grant',
  error_description: 'The client associated with this refresh token (WWY65WgtWwCypZVkmcRxqfiUKAZ0Tioj) is different than the one sent in the request (l9HwJj6ubcL1zxKbhwa3KiwBwGz9SD90).'
}
tokenJson {
  error: 'invalid_grant',
  error_description: 'Unknown or invalid refresh token.'
}
tokenJson {
  error: 'invalid_grant',
  error_description: 'The client associated with this refresh token (WWY65WgtWwCypZVkmcRxqfiUKAZ0Tioj) is different than the one sent in the request (l9HwJj6ubcL1zxKbhwa3KiwBwGz9SD90).'
}
```

When these events happen, the mutation still returns number of rows updated. However those rows were not updated at all.

packages/server/graphql/intranetSchema/intranetSchema.ts
BartoszJarocki: oh that's a nice catch. TIL: result of `loadMany` has to be filtered [`errorFilter`](https://github.com/ParabolInc/parabol/blob/8dbff9b4c6d19467c96b323be63f61e2926dc0a5/packages/server/graphql/errorFilter.ts)

BartoszJarocki:
@tianrunhe I've updated the error handling, now it shows only successful results. I've also fixed the cron expression, thank you very much! I'd appreciate it if you could take a look one more time 🙏

tianrunhe:
> @tianrunhe I've updated the error handling, now it shows only successful results. I've also fixed the cron expression, thank you very much! I'd appreciate it if you could take a look one more time 🙏

Unfortunately the issue still exists. I think it's because when there are errors, in [this line](https://github.com/ParabolInc/parabol/blob/feat/5510/updateOAuthRefreshTokens/packages/server/dataloader/atlassianLoaders.ts#L87) we will get sth like:  

```  
[  
 {  
   status: 'rejected',  
   reason: Error: Atlassian: invalid\_grant  
       at Function.fetchToken (/Users/brucetian/Projects/parabol/dev/gqlExecutor.js:65193:13)  
       at processTicksAndRejections (node:internal/process/task\_queues:96:5)  
       at async /Users/brucetian/Projects/parabol/dev/gqlExecutor.js:58258:29  
 },  
...  
 {  
   status: 'rejected',  
   reason: Error: Atlassian: invalid\_grant  
       at Function.fetchToken (/Users/brucetian/Projects/parabol/dev/gqlExecutor.js:65193:13)  
       at processTicksAndRejections (node:internal/process/task\_queues:96:5)  
       at async /Users/brucetian/Projects/parabol/dev/gqlExecutor.js:58258:29  
 }  
]
```

So the returned value for `dataLoader.get('freshAtlassianAuth').loadMany(atlassianAuthsToUpdate)` would be `[null, null, ...]`. Applying the error filter to that array has no effect.

BartoszJarocki:
Ugh, I somehow missed it's returning null when promise rejects. In this case, I see no reason to filter by error, filtering nulls should do the job. Thank you for catching this!

tianrunhe (approved):

packages/server/graphql/intranetSchema/intranetSchema.ts
mattkrick: let's keep it as is for now, we can decide whether we want to put github/gitlab logic in it later or break em out into their own

packages/client/types/constEnums.ts
```diff
@@ -331,6 +331,7 @@ export const enum TierLabel {
 
 export const enum Threshold {
   MAX_NUMBER_OF_TASKS_TO_LOAD = 1000,
+  OAUTH_REFRESH = 1209600000, // 14 days
```
mattkrick, line 334: if the cronjob runs every ~14 days & we have a var here for every 14 days does that mean the 2 are tightly coupled? 
not sure

packages/server/graphql/intranetSchema/mutations/updateOAuthRefreshTokens.ts
```diff
@@ -0,0 +1,27 @@
...
+const updateOAuthRefreshTokens = {
+  type: GraphQLInt,
+  description: `Updates Atlassian OAuth tokens that haven't been updated for more than 14 days`,
+  args: {},
+  resolve: async (_source, {}, {authToken, dataLoader}: GQLContext) => {
+    //AUTH
+    requireSU(authToken)
+
+    // RESOLUTION
+    const olderThan14DaysThreshold = new Date(Date.now() - Threshold.OAUTH_REFRESH)
```
mattkrick, line 17: let's not put `14Days` in here because OAUTH_REFRESH could change.
maybe just `olderThanThresh`?

packages/server/graphql/intranetSchema/mutations/updateOAuthRefreshTokens.ts
```diff
@@ -0,0 +1,27 @@
...
+    const atlassianAuthsToUpdate = await getAtlassianAuthsToUpdate(olderThan14DaysThreshold)
+    const updatedAtlassianAuths = (
+      await dataLoader.get('freshAtlassianAuth').loadMany(atlassianAuthsToUpdate)
+    ).filter(isNotNull)
```
mattkrick, line 21: let's also apply `errorFilter`
the only way we'd get an `instanceof Error` is if our batchFn throws an error that we don't know is an error.
e.g. if we added a JSON.parse() somewhere in the batchFn & it threw

packages/server/postgres/queries/getAtlassianAuthsToUpdate.ts
```diff
@@ -0,0 +1,8 @@
+import getPg from '../getPg'
+import {getAtlassianAuthsToUpdateQuery} from './generated/getAtlassianAuthsToUpdateQuery'
+
+const getAtlassianAuthsToUpdate = async (updatedAtThreshold: Date) => {
+  return getAtlassianAuthsToUpdateQuery.run({updatedAtThreshold}, getPg())
```
mattkrick, line 5: nit: return type as `AtlassianAuth` from https://github.com/ParabolInc/parabol/blob/38e415ebf3af01c39af15fe01a6adf8b35fc75c7/packages/server/postgres/queries/getAtlassianAuthByUserIdTeamId.ts

packages/server/utils/AtlassianServerManager.ts
```diff
@@ -66,9 +66,9 @@ class AtlassianServerManager extends AtlassianManager {
     return new AtlassianServerManager(accessToken, refreshToken)
   }
 
-  refreshToken?: string
+  refreshToken: string
 
-  constructor(accessToken: string, refreshToken?: string) {
+  constructor(accessToken: string, refreshToken: string) {
```
mattkrick, line 71: let's completely remove `refreshToken` as a field on the Manager. 
big picture, the spirit of a refreshToken is something that is kept super secure, seldom used, and never passed around. 

Changes:
- `manager.fetchToken` should return `{accessToken, refreshToken, error, scope}`

then, if you need a new instance like in `addAtlassianAuth` it just looks like this:
```ts
 const {accessToken, refreshToken} = await AtlassianServerManager.init(code)
 const manager = new AtlassianServerManager(accessToken)
```

doing this means the manager never needs a refresh token in the constructor 🎉

packages/server/graphql/types/AtlassianIntegration.ts
mattkrick: looks great! i think the chance of a missing accessToken was from the beforetimes when we didn't guarantee its freshness. now we do in all cases

mattkrick (changes requested):
you highlighted a really bad pattern that i created with sometimes sticking the refreshToken on the manager.

the biggest change is to simply have `fetchToken` return an oauth2 response & then the manager instance will never have to know about a refresh token

packages/client/types/constEnums.ts
mattkrick: let's remove this here & make it a mutation variable.
then, then cronjob can tell us the threshold like `mutation { oauthRefresh(before: 2020-05-01) }`
that way we can call this ad-hoc 🎉

packages/server/graphql/intranetSchema/mutations/updateOAuthRefreshTokens.ts
```diff
@@ -0,0 +1,27 @@
...
+import {requireSU} from '../../../utils/authorization'
+import {GQLContext} from '../../graphql'
+import getAtlassianAuthsToUpdate from '../../../postgres/queries/getAtlassianAuthsToUpdate'
+import {Threshold} from '~/types/constEnums'
+import {isNotNull} from '../../../utils/predicates'
+
+const updateOAuthRefreshTokens = {
+  type: GraphQLInt,
+  description: `Updates Atlassian OAuth tokens that haven't been updated for more than 14 days`,
+  args: {},
```
mattkrick, line 11: add
```ts
before : {
  type: GraphQLISO8601
}
```

see `dailyPulse for the pattern

packages/server/utils/AtlassianServerManager.ts
BartoszJarocki: This is a very good idea, the only question remaining is do we want to move error handling outside of the `fetchToken` method? You mentioned `manager.fetchToken should return {accessToken, refreshToken, error, scope}` and it'd mean we'd need to remove this check from `fetchToken` 
```
    if (error) {
      throw new Error(`Atlassian: ${error}`)
    }
``` 
and check for error whenever we use it manually. I think it's better that way as we can handle it without try/catch, just wanted to make sure if that's what you meant

packages/server/graphql/intranetSchema/mutations/updateOAuthRefreshTokens.ts
BartoszJarocki: Good to know! Added 👍

packages/client/types/constEnums.ts
BartoszJarocki: That's a very good idea, done!

packages/server/graphql/intranetSchema/mutations/updateOAuthRefreshTokens.ts
BartoszJarocki: No longer needed, we have an input param

packages/server/postgres/queries/getAtlassianAuthsToUpdate.ts
BartoszJarocki: Done!

packages/server/utils/AtlassianServerManager.ts
BartoszJarocki: I've removed throwing and I check the error manually, IMO that's cleaner but let me know if you think it's better to do it differently, we can discuss 😁

packages/server/graphql/intranetSchema/mutations/updateOAuthRefreshTokens.ts
BartoszJarocki: Done!

packages/server/dataloader/atlassianLoaders.ts
```diff
@@ -35,22 +35,24 @@ export const freshAtlassianAuth = (parent: RethinkDataLoader) => {
...
           if (!decodedToken || decodedToken.exp < inAMinute) {
-            const manager = await AtlassianServerManager.refresh(refreshToken)
-            const {accessToken, refreshToken: newRefreshToken} = manager
+            const {
+              accessToken,
+              refreshToken: newRefreshToken,
+              error
+            } = await AtlassianServerManager.refresh(refreshToken)
+            if (error) {
+              sendToSentry(new Error(error))
```
BartoszJarocki, line 53: This is questionable, do we want to know about the fact it failed? 🤔

BartoszJarocki:
@mattkrick thanks for the great suggestions, I implemented them all. Could you please do another round? Thank you!

packages/server/utils/AtlassianServerManager.ts
mattkrick: since we never put that in a try/catch, we didn't do any error checking anyways. now we can & we can do it without the try/catch block so it should be cleaner

packages/server/dataloader/atlassianLoaders.ts
mattkrick: i don't think we've ever seen an error come from this, so sure!

mattkrick: eh, sure! if we wanted to do it right we'd make the return type a union of a success | error since the return object can't have both an error & an access_token, but 🤷 

https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/

mattkrick (approved):

PR #5594, opened 2021-11-01
[RFC] Add GitLab auth

jordanh:


packages/client/hooks/useMenu.ts
```diff
@@ -30,7 +30,15 @@ const useMenu = <T extends HTMLElement = HTMLButtonElement>(
   if (originCoords) {
     ;(originRef as any).current = {getBoundingClientRect: () => originCoords} as RectElement
   }
-  const {portal, closePortal, togglePortal, portalStatus, setPortalStatus, openPortal} = usePortal({
+  const {
```
jordanh, line 33: Changes here are only to return `terminatePortal` so it may be used

packages/server/graphql/types/GitHubIntegration.ts
```diff
@@ -63,7 +63,7 @@ const GitHubIntegration = new GraphQLObjectType<any, GQLContext>({
     },
     scope: {
       type: GraphQLNonNull(GraphQLString),
-      description: 'The comma-separated list of scopes requested from GitHub'
+      description: 'The comma-separated list of oauthScopes requested from GitHub'
```
jordanh, line 66: TODO: chance comment

packages/server/integrations/MakeIntegrationServerManager.ts
```diff
@@ -0,0 +1,34 @@
...
+  IntegrationProvider,
+  IntegrationProviderTypesEnum
+} from '../types/IntegrationProviderAndTokenT'
+import IntegrationProviderId, {
+  IntegrationProviderIdT
+} from 'parabol-client/shared/gqlIds/IntegrationProviderId'
+import IntegrationServerManager from './IntegrationServerManager'
+import GitLabServerManager from './GitLabServerManager'
+
+// TODO fix me
```
jordanh, line 12: TODO: fix me w.r.t. Mattermost

packages/server/postgres/migrations/1634351004998_addMattermostAuth.ts
```diff
@@ -12,7 +12,7 @@ export async function up() {
     "webhookUrl" VARCHAR(2083) NOT NULL,
     "userId" VARCHAR(100) NOT NULL,
     "teamId" VARCHAR(100) NOT NULL,
-    PRIMARY KEY ("userId", "teamId")
+    PRIMARY KEY ("teamId")
```
jordanh, line 15: TODO: revert this change

scripts/toolboxSrc/getClientActionVars.ts
```diff
@@ -0,0 +1,23 @@
...
+// By building them in the toolbox, they are accessible to both the
+// webpack JS and application TS domains.
+//
+// `yarn dev` will load these via scripts/webpack/dev.client.config.js
+// `yarn start` will load these via packages/server/createSSR.ts
+
+const getClientActionVars = (): ClientActionVarsT => ({
+  atlassian: process.env.ATLASSIAN_CLIENT_ID,
+  github: process.env.GITHUB_CLIENT_ID,
+  gitlab: process.env.GITLAB_CLIENT_ID,
```
jordanh, line 14: TODO: remove gitlab

jordanh (commented):
Self-review pass one: will make some small fixups

jordanh:
👀  **Watch the demo & overview 👇 **

[image](https://www.loom.com/share/22037d51a1404fa5aab5d6ba10c98778)

Specific requests from reviewers:
   - Are these tables `IntegrationProvider` and `IntegrationToken` architected well for our needs?
   - Are the introduced GraphQL types/query patterns roughly correct? What might need to change now to make this safe to try?
   - Are the introduced mutations roughly correct? What might need to change now to make this safe to try?
   - Is the postDeploy utility function to populate global providers safe to try?
   - What would you change about the GitLab auth implementation? (Or, is it safe enough to try?)
   - What might you change about the Mattermost auth implementation? (Or, is it safe enough to try?)
   - Is the `IntegrationServerManager` pattern in the right direction? What might need to change now?
   - Are the minor refactoring (e.g. separating some dataLoaders into their own files) beneficial enough to include (or should these changes be reverted)?
   - Anything else you notice?

BartoszJarocki:
Great work @jordanh! 

Regarding the database schema, I'd avoid having separate columns for things that are related only to a particular provider type. For example, when the provider type is different than OAuth we'll have all OAuth columns empty. It'd be better to keep provider data in a separate JSONB column. This way we can deserialize data into a proper type when querying and no matter what provider type we add, we won't need to change the DB schema. And we avoid having provider-specific columns in a generic table. Then, the table would look like

```
    CREATE TABLE IF NOT EXISTS "IntegrationProvider" (
      "id" INT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
      "createdAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
      "updatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
      "type" "IntegrationProviderTypesEnum" NOT NULL,
      "tokenType" "IntegrationProviderTokenTypeEnum" NOT NULL,
      "scope" "IntegrationProviderScopesEnum" NOT NULL,
      "orgId" VARCHAR(100),
      "teamId" VARCHAR(100),
      "isActive" BOOLEAN DEFAULT TRUE NOT NULL,
      "name" VARCHAR(250) NOT NULL,
      "metadata" JSONB NOT NULL DEFAULT '{}', // here we'd store all the oauth/whatever data
       ...
    );
```

Depending on what differences are in `IntegrationToken` table between providers I'd do the same there (if depending on provider type authentication is or could be different), so instead of columns, just use JSONB.

packages/server/integrations/IntegrationServerManager.ts
```diff
@@ -0,0 +1,102 @@
...
+      grant_type: 'authorization_code',
+      redirect_uri: redirectUri
+    }
+
+    const headers = {
+      Accept: 'application/json',
+      'Content-Type': 'application/json'
+    }
+
+    const uri = `${this.provider.serverBaseUri}/oauth/token?${stringify(queryParams)}`
```
tianrunhe, line 53: Does most of the services have OAuth token request URI like this?

scripts/toolboxSrc/postDeploy.ts
```diff
@@ -62,12 +68,34 @@ const storePersistedQueries = async () => {
...
+  dotenvExpand(myEnv)
+
   try {
     const r = await getRethink()
     await flushSocketConnections()
     await storePersistedQueries()
     await r.getPoolMaster().drain()
+
+    const pg = getPg()
+    await upsertGlobalIntegrationProvidersFromEnv()
```
tianrunhe, line 97: Would there be delay between inserting the rows and app starting?

packages/server/graphql/mutations/addIntegrationProvider.ts
```diff
@@ -0,0 +1,96 @@
...
+            ...maybeToken
+          }
+          await insertIntegrationProviderWithToken(dbProvider, dbToken)
+        } else {
+          await insertIntegrationProvider(dbProvider)
+        }
+    }
+
+    const data = {userId: viewerId, teamId}
+    publish(SubscriptionChannel.TEAM, teamId, 'AddIntegrationProvider', data, subOptions)
```
tianrunhe, line 91: Seems like the subscription level should be depend on the scope.

tianrunhe (commented):
Great work! I second Bartosz's suggestion and made some minor suggestions

packages/server/postgres/migrations/1635354572171_addIntegrationProviderAndToken.ts
```diff
@@ -0,0 +1,98 @@
...
+      CONSTRAINT global_provider_must_be_oauth2 CHECK (
+        "scopeGlobal" IS NULL OR ("scopeGlobal" = TRUE AND "tokenType" = 'OAUTH2')
+      )
+    );
+    CREATE INDEX IF NOT EXISTS "idx_IntegrationProvider_typeAndScope"
+      ON "IntegrationProvider"("type", "scope");
+    CREATE INDEX IF NOT EXISTS "idx_IntegrationProvider_orgId" ON "IntegrationProvider"("orgId");
+    CREATE INDEX IF NOT EXISTS "idx_IntegrationProvider_teamId" ON "IntegrationProvider"("teamId");
+    CREATE TABLE IF NOT EXISTS "IntegrationToken" (
+      "teamId" VARCHAR(100) NOT NULL,
```
Dschoordsch, line 61: Some thoughts on whether or not we should have the IntegrationToken per User, Team, Provider (1, current solution) or just User, Provider and then a separate relation TeamMemberIntegration (2).

1) **Pro**
  * easy, it's what we have now
  * no joins
**Con**
  * updates to the token need to be propagated to all rows
2) **Pro**
  * it would simplify refreshing a token and similar operations on the token itself, because it would not require to update multiple rows.
  * integration specific settings could be stored on TeamMemberIntegration and thus could also be stored without having a token
 **Con**
  * additional join for retrieving the token
  * the key User,IntegrationProvider might not work. While I think it is unlikely that someone wants to have access to Jira with 2 different accounts in 1 team, the likelyhood increases when it's spread over multiple teams. This is probably still not a use case, only wanted to mention it.

packages/server/integrations/IntegrationServerManager.ts
jordanh: This will almost certainly need to be parameterized

scripts/toolboxSrc/postDeploy.ts
jordanh: The pool is drained, so these row(s) will be updated before the app starts

packages/server/graphql/mutations/addIntegrationProvider.ts
jordanh: Indeed. If/when we update the Org view to manage integration providers, this logic will need to be added here

packages/server/graphql/types/IntegrationProvider.ts
```diff
@@ -0,0 +1,113 @@
...
+const IntegrationProvider = new GraphQLObjectType<any, GQLContext>({
+  name: 'IntegrationProvider',
+  description: 'An authentication provider configuration',
+  fields: () => ({
+    id: {
+      type: new GraphQLNonNull(GraphQLID),
+      description: "The provider's unique identifier",
+      resolve: ({id}) => IntegrationProviderId.join(id)
+    },
+    type: {
```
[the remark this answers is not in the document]
jordanh: It's certainly possible! For Mattermost we could support Oauth2, PATs, and Webhooks if we wanted to as Mattermost supports each of these modes

packages/server/postgres/migrations/1635354572171_addIntegrationProviderAndToken.ts
```diff
@@ -0,0 +1,98 @@
...
+      "teamId" VARCHAR(100),
+      "isActive" BOOLEAN DEFAULT TRUE NOT NULL,
+      "name" VARCHAR(250) NOT NULL,
+      "serverBaseUri" VARCHAR(2083) NOT NULL,
+      "oauthScopes" VARCHAR(100)[],
+      "oauthClientId" VARCHAR(2600),
+      "oauthClientSecret" VARCHAR(2600),
+      "createdAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+      "updatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+      UNIQUE("scopeGlobal", "type"),
```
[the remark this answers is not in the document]
jordanh: :)

packages/server/postgres/migrations/1635354572171_addIntegrationProviderAndToken.ts
```diff
@@ -0,0 +1,98 @@
...
+    CREATE TABLE IF NOT EXISTS "IntegrationToken" (
+      "teamId" VARCHAR(100) NOT NULL,
+      "userId" VARCHAR(100) NOT NULL,
+      "providerId" INT NOT NULL,
+      "isActive" BOOLEAN DEFAULT TRUE NOT NULL,
+      "accessToken" VARCHAR(2600),
+      "expiresAt" TIMESTAMP WITH TIME ZONE,
+      "oauthRefreshToken" VARCHAR(2600),
+      "oauthScopes" VARCHAR(100)[],
+      "attributes" JSONB,
```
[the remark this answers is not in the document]
jordanh: Yeah, I agree with you and @BartoszJarocki, let's move these `oauth*` vars to the `attributes` column.

If an implementer wanted to, we could use `attributes` to store things like your last github search, although, personally I think these should likely belong on their own table. I mostly intended these abstracts to only handle auth. The fact we conflate auth and state is a bit hacky IMO.

packages/server/postgres/queries/getGitHubAuthByUserIdTeamId.ts
```diff
@@ -4,9 +4,7 @@ import {
   IGetGitHubAuthByUserIdTeamIdQueryResult
 } from './generated/getGitHubAuthByUserIdTeamIdQuery'
 
-// this table has a composite primary key (userId, teamId),
-// which cannot use the index with a WHERE IN or JOIN on VALUES
-// so if we want to query multiple userIds/teamIds, just call this multiple times
+// if we want to query multiple userIds/teamIds, just call this multiple times
```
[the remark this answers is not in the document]
jordanh: @mattkrick tells me this is no longer true

BartoszJarocki:
Closing in favor of https://github.com/ParabolInc/parabol/pull/5805

Issue #5556, opened 2021-10-21
Make RetroReflection opened in spotlight by another member non-droppable


Dschoordsch:
Requires #5272 

While testing, I noticed that while the source reflection is animating in a circle, a user can drag & drop a reflection onto the animating source. We can prevent that from happening by checking whether the reflection group contains the source reflection and, if it does, prevent it from being droppable. See: https://github.com/ParabolInc/parabol/blob/4e6b5fd8ab5aa3ba5dd1d984dea748af12ab857d/packages/client/components/ReflectionGroup/ReflectionGroup.tsx#L171

_Originally posted by @nickoferrall in https://github.com/ParabolInc/parabol/pull/5500#pullrequestreview-785092289_

PR #5600, opened 2021-11-03
Animate remote reflection when it enters or leaves Spotlight

### LINKED ISSUE #5489: Spotlight: remote reflection animation
When a remote reflection enters or leaves the Spotlight, it should fade in or out of the Spotlight. See Matt's [comment](https://github.com/ParabolInc/parabol/pull/5376#discussion_r721825888) here:
> let's animate a change in z-index so things don't just appear & disappear.
if something used to be above & now it's below, we want the card to fade below the spotlight & appear in full opacity under the spotlight.
to do that, we'll probably want to transition the opacity down to 25%, then change the z-index, then quickly bump the opacity back up to 100%

Expected behaviour: 

- User A opens the Spotlight.
- User B drags a reflection that doesn’t exist in the Spotlight and hovers it above a reflection that does exist in the Spotlight. As it’s hovering, User A should now see the remote reflection hovering above the Spotlight.
- Animate the remote reflection’s opacity so that it fades into the Spotlight.
- If the remote reflection hovers above a group in the Spotlight, but it’s not dropped on a Spotlight group and returns to the kanban, User A should see the remote reflection fade into the Spotlight and then fade out of it.

### AC

- [ ] If a remote reflection that doesn't exist in the Spotlight is hovering above a Spotlight group, the viewer can see the reflection hovering above the Spotlight group.
- [ ] There's a smooth transition as the remote reflection fades into the Spotlight.
- [ ] If the remote reflection that transitioned into the Spotlight moves out of the Spotlight (i.e. it's `targetId` is no longer a Spotlight result group), it smoothly fades out of the Spotlight.

igorlesnenko:
fix #5489

**How to test:**

- Create a retro meeting
- Add various reflections
- Go to **Group** phase
- Join the meeting from another browser window as User B
- As User A, open a Spotlight on any reflection
- As User B, drag a reflection that doesn’t exist in the Spotlight and hover it above a reflection that does exist in the Spotlight. As it’s hovering, User A should now see the remote reflection hovering above the Spotlight with a fade-in animation
- As User B, hover a reflection above a reflection that does **not** exist in the Spotlight. User A should now see the remote reflection disappearing from the Spotlight with a fade-out animation.
- As User B, drop the reflection not in a group so its returns to the Kanban. As user A, see it disappears with a fade-out animation.


packages/client/components/ReflectionGroup/RemoteReflection.tsx
```diff
@@ -49,7 +50,8 @@ const RemoteReflectionModal = styled('div')<{
   transform,
   animation:
     isSpotlight && !isDropping ? `${circleAnimation(transform)} 3s ease infinite;` : undefined,
-  zIndex: ZIndex.REFLECTION_IN_FLIGHT
+  zIndex: ZIndex.REFLECTION_IN_FLIGHT,
+  ...styleProps
```
igorlesnenko, line 54: Spread instead of using `style` prop to support emotion `keyframes`

igorlesnenko:
@nickoferrall Can you review, please?

packages/client/utils/retroGroup/updateClonePosition.ts
```diff
@@ -25,22 +52,49 @@ const getTransition = (isClipped: boolean, timeRemaining: number) => {
...
+  targetId?: string | null,
+  isClose?: boolean,
+  lastStyle?: React.CSSProperties
 ) => {
   const spotlightEl = document.getElementById('spotlight')
-  const showAboveSpotlight = spotlightEl ? spotlightEl.contains(targetEl) : false
+  const isTargetInSpotlight = targetId
+    ? spotlightEl && document.querySelector(`div[${DragAttribute.DROPPABLE}='${targetId}']`)
+    : false
+  const isBelongsToSpotlight = spotlightEl && spotlightEl.contains(element)
```
nickoferrall, line 67: NIT: `inInSpotlight` might be easier to read here

packages/client/utils/retroGroup/updateClonePosition.ts
```diff
@@ -25,22 +52,49 @@ const getTransition = (isClipped: boolean, timeRemaining: number) => {
...
   const clippedTop = Math.min(Math.max(minTop, top), maxTop - bbox.height)
   const isClipped = clippedTop !== top
+
+  const fadeInAnimation = `${reflectionSpotlightFadeIn.toString()} 0.5s linear 0s forwards`
+  const fadeOutAnimation = `${reflectionSpotlightFadeOut.toString()} 0.5s linear 0s forwards`
+  let animation: string | undefined = undefined
+
+  if (spotlightEl && !isBelongsToSpotlight && lastStyle) {
+    if (isTargetInSpotlight && !isClose) {
+      animation = fadeInAnimation
```
nickoferrall, line 80: Perhaps we could create two `const` variables, `isFadingIn` and `isFadingOut` instead of a `let` so it's easier to follow. For example, `const isFadingIn = isTargetInSpotlight && !isClose etc` and then on line 97 we could do `animation: isFadingIn ? fadeInAnimation : isFadingOut ? fadeOutAnimation : undefined`

packages/client/hooks/useDraggableReflectionCard.tsx
```diff
@@ -37,17 +37,38 @@ const useRemotelyDraggedCard = (
...
+        drag.ref,
+        bbox,
+        windowDims.clientHeight,
+        timeRemaining,
+        targetId,
+        isClose,
+        lastStyle
+      )
+
+      setLastStyle(style)
```
nickoferrall, line 58: I'm seeing some janky behaviour that might be related to this state update. `requestAnimationFrame` is being executed before every repaint whereas `setState` is asynchronous and might happen in a later frame. 

Is there a way that this could be accomplished without the state update? I'd be happy to look into this with you on Monday if you like!

remote-anim

It also looks like the reflectionCard jumps back to its position on the kanban at the end of the gif below rather than transitioning into its position.

packages/client/utils/retroGroup/updateClonePosition.ts
```diff
@@ -25,22 +52,49 @@ const getTransition = (isClipped: boolean, timeRemaining: number) => {
...
+    }
+
+    const isCurrentlyBehindSpotlight =
+      lastStyle.zIndex === ZIndex.REFLECTION_IN_FLIGHT || lastStyle.zIndex == null
```
nickoferrall, line 84: `null` isn't a valid value for the `zIndex` - it's best to use `undefined` here

nickoferrall (changes requested):
Great work getting stuck into this tricky issue. The animation is looking good - just some janky behaviour that needs fixing before moving to Maintainer Review

packages/client/hooks/useDraggableReflectionCard.tsx
igorlesnenko: @nickoferrall Removing the state seems to have no effect. I think I fixed both issues. Can you check if it works for you?

https://user-images.githubusercontent.com/466991/141094912-d02b8515-2170-4689-b476-4dcffe456f32.mov

packages/client/components/ReflectionGroup/RemoteReflection.tsx
```diff
@@ -189,16 +200,18 @@ const RemoteReflection = (props: Props) => {
...
   return (
     <>
       <RemoteReflectionModal
         ref={ref}
-        style={nextStyle}
+        style={remoteReflectionModalStyle}
         isDropping={isDropping}
         isSpotlight={isSpotlight}
         transform={transform}
+        animation={animation as string | undefined}
```
igorlesnenko, line 214: As a separate prop, as it uses `keyframes` feature of emotion

packages/client/utils/retroGroup/updateClonePosition.ts
```diff
@@ -25,22 +52,47 @@ const getTransition = (isClipped: boolean, timeRemaining: number) => {
...
+
+  const fadeInAnimation = `${reflectionSpotlightFadeIn.toString()} 0.5s linear 0s forwards`
+  const fadeOutAnimation = `${reflectionSpotlightFadeOut.toString()} 0.5s linear 0s forwards`
+
+  const isCurrentlyBehindSpotlight =
+    lastStyle &&
+    (lastStyle.zIndex === ZIndex.REFLECTION_IN_FLIGHT || lastStyle.zIndex === undefined)
+  const isFadingIn = spotlightEl && !isInSpotlight && lastStyle && isTargetInSpotlight && !isClose
+  const isFadingOut =
+    spotlightEl &&
```
nickoferrall, line 82: `spotlightEl && !isInSpotlight && lastStyle` are being used in `isFadingIn` and `isFadingOut` so perhaps we can create a variable to make this easier to read. Maybe `const wasBehindSpotlight = spotlightEl && !isInSpotlight`?

packages/client/utils/retroGroup/updateClonePosition.ts
```diff
@@ -25,22 +52,47 @@ const getTransition = (isClipped: boolean, timeRemaining: number) => {
...
+  const minTop = getMinTop(top, element)
   const clippedTop = Math.min(Math.max(minTop, top), maxTop - bbox.height)
   const isClipped = clippedTop !== top
+
+  const fadeInAnimation = `${reflectionSpotlightFadeIn.toString()} 0.5s linear 0s forwards`
+  const fadeOutAnimation = `${reflectionSpotlightFadeOut.toString()} 0.5s linear 0s forwards`
+
+  const isCurrentlyBehindSpotlight =
+    lastStyle &&
+    (lastStyle.zIndex === ZIndex.REFLECTION_IN_FLIGHT || lastStyle.zIndex === undefined)
```
nickoferrall, line 79: It looks like we only need the `zIndex` from `lastStyle`. Rather than storing `lastStyle` in local state in the parent component, could we just store `lastZIndex` in local state or a ref?

packages/client/hooks/useDraggableReflectionCard.tsx
nickoferrall: Beautiful! It's working well for me 👍

packages/client/utils/retroGroup/updateClonePosition.ts
```diff
@@ -25,22 +52,47 @@ const getTransition = (isClipped: boolean, timeRemaining: number) => {
...
+  const isFadingIn = spotlightEl && !isInSpotlight && lastStyle && isTargetInSpotlight && !isClose
+  const isFadingOut =
+    spotlightEl &&
+    !isInSpotlight &&
+    lastStyle &&
+    !isCurrentlyBehindSpotlight &&
+    (!showAboveSpotlight || isClose)
```
nickoferrall, line 86: Could you add a comment explaining why each variable is required? That should help the next dev feel more confident touching the code

nickoferrall (commented):
The animation is looking great! I've left a few comments - if you make some changes feel free to request another review from me or move it to Maintainer Review if you prefer

packages/client/utils/retroGroup/updateClonePosition.ts
igorlesnenko: @nickoferrall As far as I see, we are unable to use local state here as it is just a plain function calculating value, not a component.

igorlesnenko:
@nickoferrall Make some changes, can you check?

packages/client/utils/retroGroup/updateClonePosition.ts
```diff
@@ -25,22 +52,56 @@ const getTransition = (isClipped: boolean, timeRemaining: number) => {
...
   const {top, left} = bbox
-  const minTop = getMinTop(top, targetEl)
+  const minTop = getMinTop(top, element)
   const clippedTop = Math.min(Math.max(minTop, top), maxTop - bbox.height)
   const isClipped = clippedTop !== top
+
+  const fadeInAnimation = `${reflectionSpotlightFadeIn.toString()} 0.5s linear 0s forwards`
+  const fadeOutAnimation = `${reflectionSpotlightFadeOut.toString()} 0.5s linear 0s forwards`
+
+  const notInSpotlightSuggestions = spotlightEl && !isInSpotlight
```
nickoferrall, line 77: It looks like `isInSpotlight` already checks whether `spotlightEl` exists so we can remove this line and just use `isInSpotlight`

packages/client/utils/retroGroup/updateClonePosition.ts
```diff
@@ -25,22 +52,56 @@ const getTransition = (isClipped: boolean, timeRemaining: number) => {
...
+  const isCurrentlyBehindSpotlight = lastZIndex !== ZIndex.REFLECTION_IN_FLIGHT_SPOTLIGHT
+  const isFadingIn =
+    // Spotlight is open and the element is not a part of the spotlight suggestions
```
nickoferrall, line 80: The original issue just used the terminology "source" for the reflection group at the top of the Spotlight and "results" for all of the similar Spotlight groups suggested at the bottom. I've been guilty of introducing new terminology like "Spotlight groups", but I think we should stick with just "source" and "results" wherever possible. 

So, I'd prefer to avoid using "Spotlight suggestions" if possible. It looks like we could just say "not a part of the spotlight" here.

nickoferrall (approved):
This is looking good to me! On to Maintainer Review

packages/client/utils/retroGroup/updateClonePosition.ts
igorlesnenko: @nickoferrall There is a small difference — in the second case, we check that the element is not in the spotlight **but** spotlight is open. It could be that spotlight is closed and therefore the element is not in the spotlight.
Changed the code and comments a bit. Can you check if it is more clear now?

packages/client/utils/retroGroup/updateClonePosition.ts
```diff
@@ -25,22 +52,56 @@ const getTransition = (isClipped: boolean, timeRemaining: number) => {
...
-  timeRemaining: number
+  timeRemaining: number,
+  targetId?: string | null,
+  isClose?: boolean,
+  lastZIndex?: number
 ) => {
   const spotlightEl = document.getElementById('spotlight')
-  const showAboveSpotlight = spotlightEl ? spotlightEl.contains(targetEl) : false
+  const isTargetInSpotlight = targetId
+    ? !!(spotlightEl && document.querySelector(`div[${DragAttribute.DROPPABLE}='${targetId}']`))
```
nickoferrall, line 65: Before moving to Maintainer Review, I think it'd be best to merge with master and check how things are working. I suspect `isTargetInSpotlight` might be inaccurate as Spotlight results won't always have the droppable attribute [see here](https://github.com/ParabolInc/parabol/blob/7168dcb35410e60c57f8b68719e0490b69eeb35c/packages/client/components/ReflectionGroup/ReflectionGroup.tsx#L208).

Instead, we could follow the pattern in `DraggableReflectionCard` where we query the local Relay store to get the Spotlight result ids. We could copy the lines [here](https://github.com/ParabolInc/parabol/blob/2360f8cb7609ba483b2473c9e7635c204cdf7a85/packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx#L89-L110) and add them to the `useDraggableReflectionCard` hook on line 40. Then, we can pass the `resultIds` to `getDroppingStyles` and see if `resultsIds.includes(targetId)`.

This would mean we're passing 4 extra arguments into `getDroppingStyles` just to get the Spotlight animation. Therefore, it might be cleaner to create a new function called `getSpotlightAnimation` and do all of the calculations there. The function would return `animation`, and we'd pass that to `RemoteReflection` as a prop. This would mean that we wouldn't need to destucture `animation` from `nextStyle` in the `RemoteReflection`.

packages/client/utils/retroGroup/updateClonePosition.ts
nickoferrall: That makes sense - thanks!

packages/client/utils/retroGroup/updateClonePosition.ts
igorlesnenko: @nickoferrall Thanks! You are right. After rebase, nothing is working as expected. I made changes according to your comment. 

Additionally, one new issue appeared after rebase: the remote reflection is now appearing in wrong coordinates — it targets the reflection on the background, not in spotlight. I'm going to look into this soon, but maybe you can point me to right direction here too

igorlesnenko: I think I found the issue, we are still using
```
 const targetEl = targetId
    ? (document.querySelector(`div[${DragAttribute.DROPPABLE}='${targetId}']`) as HTMLElement)
    : null
    ```
    
To determine the remote reflection coordinates

Comment (nickoferrall) [reply packages/client/utils/retroGroup/updateClonePosition.ts]: Ah how annoying, sorry to hear that. I've checked out your branch to see I can replicate the issue, but there were some merge conflicts. After fixing them, the remote reflection appears at the top left of the screen, i.e. coordinate 0,0. 

In [this PR](https://github.com/ParabolInc/parabol/pull/5610), I pushed [a fix](https://github.com/ParabolInc/parabol/blob/23a510aff8bc9d0a396113c92738f0ffe2263c9f/packages/client/components/ReflectionGroup/ReflectionGroup.tsx#L206) to prevent remote reflections from using the position on the kanban. 

Perhaps we could Zoom and go through this bit together?

File: packages/client/components/ReflectionGroup/ReflectionGroup.tsx [hunk 13]
```diff
@@ -205,7 +205,7 @@ const ReflectionGroup = (props: Props) => {
         />
       )}
       <Group
-        {...(isDroppable ? {[DragAttribute.DROPPABLE]: reflectionGroupId} : null)}
+        {...(isBehindSpotlight ? null : {[DragAttribute.DROPPABLE]: reflectionGroupId})}
```
igorlesnenko, line 208: Temporary fix. Will be solved by another PR. See discussion here https://github.com/ParabolInc/parabol/pull/5600#discussion_r751129213

packages/client/utils/retroGroup/updateClonePosition.ts
igorlesnenko: @nickoferrall pushed a fix, can you check, please?

packages/client/utils/retroGroup/updateClonePosition.ts
```diff
@@ -24,18 +51,61 @@ const getTransition = (isClipped: boolean, timeRemaining: number) => {
...
+  isClose: boolean,
+  lastZIndex: number | undefined
+) => {
+  const spotlightEl = document.getElementById('spotlight')
+  const isTargetInSpotlight = targetId && groupIdsInSpotlight.includes(targetId)
+  const isInSpotlight = spotlightEl && spotlightEl.contains(element)
+  const notInSpotlight = spotlightEl && !spotlightEl.contains(element)
+  const showAboveSpotlight = isInSpotlight || isTargetInSpotlight
+
+  const fadeInAnimation = `${reflectionSpotlightFadeIn.toString()} 0.5s linear 0s forwards`
```
mattkrick, line 67: let's use `BezierCurve.DECELERATE` instead of linear. the different is _real_ small, but basically it means 80% of the transition is completed in the first 50% of the time.

packages/client/components/ReflectionGroup/ReflectionGroup.tsx
```diff
@@ -101,13 +101,13 @@ const ReflectionGroup = (props: Props) => {
...
   const isInSpotlight = !openSpotlight
   const isBehindSpotlight = isSpotlightOpen && !isInSpotlight
-  const isDroppable = useMemo(() => {
-    const isSourceGroup = spotlightGroup?.id === reflectionGroupId
-    const isDraggingSource = !!spotlightGroup?.reflections.find(
-      ({isViewerDragging}) => isViewerDragging
-    )
-    return isSpotlightOpen ? isDraggingSource || isSourceGroup : true // prevent grouping results into results
-  }, [spotlightGroup])
+  // const isDroppable = useMemo(() => {
```
mattkrick, line 104: can remove?

packages/client/utils/retroGroup/updateClonePosition.ts
```diff
@@ -24,18 +51,61 @@ const getTransition = (isClipped: boolean, timeRemaining: number) => {
...
 
+export const getSpotlightAnimation = (
+  element: HTMLDivElement,
+  targetId: string | null | undefined,
+  groupIdsInSpotlight: string[],
+  isClose: boolean,
+  lastZIndex: number | undefined
+) => {
+  const spotlightEl = document.getElementById('spotlight')
+  const isTargetInSpotlight = targetId && groupIdsInSpotlight.includes(targetId)
```
mattkrick, line 62: let's concat the `sourceReflectionGroupId` into this array, that way if a remote drag hovers the card at the top it shows above

packages/client/utils/retroGroup/updateClonePosition.ts
```diff
@@ -24,18 +51,61 @@ const getTransition = (isClipped: boolean, timeRemaining: number) => {
...
+  const isInSpotlight = spotlightEl && spotlightEl.contains(element)
```
mattkrick, line 63: `contains` is a DOM method so it's a little expensive (at least on mobile devices).

can we write something like
```ts
const isInSpotlight = spotlightEl?.contains(element) ?? false
const notInSpotlight = spotlightEl ? !isInSpotlight : false
```

mattkrick (changes requested):
THIS LOOKS SO GOOD! I seriously didn't think animating a z-index like this could look so good.

packages/client/components/ReflectionGroup/ReflectionGroup.tsx
igorlesnenko: Removed, as now fixed in master

.husky/post-commit
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+yarn run post-commit
```
igorlesnenko, line 4: added here fixes from #5702 , otherwise commit does not work

packages/client/utils/retroGroup/updateClonePosition.ts
igorlesnenko: Added, but it does not seem to be working. Rotating spotlight card on remote browser does not have `targetId`. Checking what is possible to do.

igorlesnenko: @nickoferrall can you please advise? 

Recently, we removed droppable attribute from the spotlight source reflection with circular animation (#5556)
Now, because of this, hovering above the source reflection cause remote reflection to `fadeOut`, as `targetId` is not submitted.

As I see we pick `targetId` from `DragAttribute.DROPPABLE` here
https://github.com/ParabolInc/parabol/blob/7ade67ab0cf2af96878785a5c30752e634dd8def/packages/client/utils/retroGroup/getTargetGroupId.ts#L18

Maybe we need to set an additional always existing attribute like `data-target-id`?

nickoferrall: Ah so I actually disagree that we should show remote drags above the source. We want to show remote drags above the Spotlight when they're hovering over a result group so that if they're grouped, the remote reflection fades into the Spotlight rather than suddenly appearing.

However, remote reflections can't be dropped onto the circular animating source, so I don't think there's any advantage to showing them above the source.

igorlesnenko: @nickoferrall thank you
@mattkrick what do you think? Should I remove this concatenation?

mattkrick: sure, safe to leave this one be

igorlesnenko: @mattkrick Do you mean keep my change as is?

nickoferrall: Yeah, I think he means leave it how it is, i.e. don't make any changes

Issue #5259, opened 2021-08-11
Dashboard's top bar is missing focus styles


enriquesanchez:
## Issue - Bug

<img width="1050" alt="Screen Shot 2021-08-11 at 11 07 54 a m" src="https://user-images.githubusercontent.com/3276087/129064453-0eb4c298-7167-4aa1-a9ed-4144c9739090.png">

All of the elements in the app's dashboard are interactive yet not one of them has a focus style to indicate keyboard focus.

A few of them, like the `logo`, the `Search` submit button, and the `user` menu are no focusable at all. They can only be interacted with if the user is using a mouse.

Users who's primary way of navigating is the keyboard have no way of knowing their position when interacting with the dashboard.

Related WCAG success criteria:
- [2.4.7 Focus Visible](https://www.w3.org/WAI/WCAG21/quickref/#focus-visible)
- [2.1.1 Keyboard](https://www.w3.org/WAI/WCAG21/quickref/#keyboard)

**How serious is the bug?** P3
- **OS:** All
- **Browser:** All

### Acceptance Criteria 
Users can:
   - Operate with a keyboard all of the interactive items in the app's dashboard. 
   - Know what their focus position is at all times in the app's dashboard

### Suggestions

I strongly suggest we stick to semantic HTML for our interactive components. The `a`, `button` elements already provide the necessary interactive and a11y features that are standard and expected.

### Estimated effort

1 hour to triage. More if root cause is already identified.

ackernaut:
We probably need hover and or focus states designed for this? We may need to add some design AC.

enriquesanchez:
@acressall and I have been looking into this and came up with a focus style that has enough contrast against the top bar's dark purple background, and also stays consistent with any focus styles we already use elsewhere in the app.

We decided to use a border of `Sky 400 #61B1EB` as it's the shade that has enough ratio:

![imagen](https://user-images.githubusercontent.com/3276087/134384161-0421108e-2173-498f-9a75-2cab6db768c6.png)

Note: image above shows focus styles applied to _all_ elements of the top bar for reference. Normally only one item would receive focus at a time. 

Next steps:

- Implement these updates

PR #5607, opened 2021-11-03
adding focus states to topbar

acressall:
Reviewing top bar focus states with @enriquesanchez based on #5259


enriquesanchez:
Looks sooooo good! 🙌

https://user-images.githubusercontent.com/3276087/140164425-8ab6ee0d-288b-4e20-aca4-fbb0f4a8d9f9.mp4

acressall:
@ackernaut can you make sure I didn't break any best practices? In order to add focus states to images, I added wrapper buttons to them

ackernaut:
Taking a card!

packages/client/components/DashTopBar.tsx
```diff
@@ -35,9 +35,15 @@ const Wrapper = styled('header')({
 })
 
 const LeftNavToggle = styled(PlainButton)({
+  borderRadius: 100,
   fontSize: ICON_SIZE.MD24,
   lineHeight: '16px',
-  paddingLeft: 16
+  paddingLeft: 16,
```
ackernaut, line 41: Here `paddingLeft` gets overridden by `padding` that comes after it, so it should be removed.

MD icons in the top bar have a `24px` footprint. Here you are adding some padding so that you can have a box-shadow as a circle. This is fine as long as the 24px footprint of the icon is centered vertically and spaced correctly horizontally. For the left nav icon, it needs to be centered in a 56x56 area. The left and right gutters need to be 16 to line up with the navigation items below. I believe the margin can be set to 12 instead of 8 to line it up. (margin 12 plus padding 4 gives us the 16 on both sides)

packages/client/components/DashTopBar.tsx
```diff
@@ -35,9 +35,15 @@ const Wrapper = styled('header')({
...
+  margin: 8,
+  padding: 4,
+  ':focus': {
+    boxShadow: '0 0 0 2px #61B1EB'
```
ackernaut, line 45: Instead of the hex for the color, use a template literal with the `SKY_400` with the constant from paletteV3 — this needs to be replaced in 3 other places below

Example
```
import {PALETTE} from '../styles/paletteV3'
...
boxShadow: `0 0 0 2px ${PALETTE.SKY_400}`
```

packages/client/components/DashTopBar.tsx
```diff
@@ -48,11 +54,21 @@ const LeftNavHeader = styled('div')({
...
-  margin: '8px 16px'
+  margin: '8px 0',
+  padding: '8px 8px 4px 8px',
+  ':focus': {
+    boxShadow: '0 0 0 2px #61B1EB',
+    outline: 'none'
+  }
 })
 
+const Img = styled('img')({})
```
ackernaut, line 70: You can remove this and use a simple `<img` below instead of a styled component `<Img

packages/client/components/DashTopBar.tsx
```diff
@@ -87,7 +103,9 @@ const DashTopBar = (props: Props) => {
         <LeftNavToggle onClick={toggle} aria-label='Toggle dashboard menu'>
           <Icon>{'menu'}</Icon>
         </LeftNavToggle>
-        <Img onClick={gotoHome} crossOrigin='' src={parabolLogo} alt='Parabol logo' />
+        <LogoWrapper onClick={gotoHome}>
+          <Img crossOrigin='' src={parabolLogo} alt='Parabol logo' />
```
ackernaut, line 107: Use `<img …`

packages/client/components/TopBarAvatar.tsx
```diff
@@ -9,10 +9,21 @@ import {TopBarAvatar_viewer} from '~/__generated__/TopBarAvatar_viewer.graphql'
...
+  marginLeft: 8,
+  padding: 4,
+  ':focus': {
+    boxShadow: '0 0 0 2px #61B1EB',
+    cursor: 'pointer',
+    outline: 'none'
+  }
 })
 
+const SpacedAvatar = styled(Avatar)({})
```
ackernaut, line 25: If you don’t need to style `Avatar` you can remove this and use `<Avatar />` instead of `<SpacedAvatar />` below.

packages/client/components/TopBarAvatar.tsx
```diff
@@ -29,15 +40,16 @@ const TopBarAvatar = (props: Props) => {
   )
   return (
     <>
+    <AvatarWrapper onClick={togglePortal}>
       <SpacedAvatar
```
ackernaut, line 44: Use <Avatar />

ackernaut (changes requested):
One thing to note on the interaction behaviors.

Tabbing through looks good, each element gets a circle.

When tapping on the elements the behavior varies.

Tap the avatar or icons to the left of it, they focus briefly, then lose focus once the menu opens. Do we want this flash?

When tapping the menu icon or the Parabol logo, the focus remains. Do we want it to remain?

When tapping the Parabol icon on the Meetings view, nothing happens, because we’re already at that view.

Otherwise, there are few styled components without styles to clean up, use the palette constant in a few places, and bump the horizontal alignment of the menu icon.

enriquesanchez:
Thanks for the review @ackernaut 🙌

We fixed the code changes you requested.

> When tapping on the elements the behavior varies.
>Tap the avatar or icons to the left of it, they focus briefly, then lose focus once the menu opens. Do we want this flash?

Good catch. We'll fix this on this PR as well.

>When tapping the menu icon or the Parabol logo, the focus remains. Do we want it to remain?
I think the `LeftNavToggle` should have the following behavior:

1. When I open the left nav, focus should move to the first item on the list (Meetings).
2. When I close it, focus should remain on the menu icon.

We will address this on a separate PR to keep things clean.

>When tapping the Parabol icon on the Meetings view, nothing happens, because we’re already at that view.
Agreed. In this case focus should reset. I think what's happening right now is that because this is a single-page app, the page is not reloading to the browser is keeping the focus in the logo. 

We'll also address this on a separate PR.

For the `Help`, `Notifications`, and `Avatar` menus, we are also going to create an additional PR to handle their behavior. In those cases, focus should move down into the first item of the dropdown when they are opened, which is not happening right now.

ackernaut:
@enriquesanchez cc @acressall 

Thanks for making those changes. Those were relatively minor, but I think seeing the details will help you catch them next time.

Thanks for addressing the UI behavior concerns. So you’ll get 1 more change in regarding the flashing behavior? I’ll take a look at that change and meanwhile we can ask @mattkrick (hello 👋 ) if this type of PR requires maintainer review

Please open a separate implementation issue including the AC and EE that will describe how you’ll handle the behaviors around toggles and moving focus to the first menu items (like you mentioned above). I’m in favor of deferring that work for progress here

enriquesanchez:
>So you’ll get 1 more change in regarding the flashing behavior? 

@ackernaut Just pushed an update:

https://user-images.githubusercontent.com/3276087/144454227-56111b94-3751-4e72-94bf-63e63a542b78.mp4

You’ll notice that the focus style is gone when clicking on it, but when I close the dropdown the focus style appears again. I don’t think we can avoid that, since I’m clicking on the button and the focus naturally moves back to the button.

I was looking at the WordPress block editor, for reference, since extensive a11y testing and work goes into it. It looks like it has the same behavior:

https://user-images.githubusercontent.com/3276087/144454449-6c8eaaf4-e147-4939-9346-6e91c4ad2568.mp4

You'll notice the flash of focus style as well when the button is active for a brief second is there. I'm fine with removing on our end.

enriquesanchez:
Related: #5736

enriquesanchez:
@ackernaut Pinging in case this slipped through the cracks. The requested changes have been updated.

ackernaut:
@enriquesanchez I did miss the changes here. Thanks for pinging. Taking a card.

ackernaut:
This looks good 🙌 . I’m going to merge master in, see if there are any problems, then I’ll merge this PR

ackernaut (approved):

PR #5514, opened 2021-10-13
Prevent remote reflection from loosing its position when opening spotlight

BartoszJarocki:
Fixes #5415 

@nickoferrall seems like the root cause was pretty trivial. I went through the flow of dragging the remote reflection and it was only a matter of a wrong check if the spotlight is opened. Do you mind taking a look? 🙏 
CleanShot 2021-10-13 at 16 16 29
l

PR #5610, opened 2021-11-03
Remote Spotlight animation polish

### LINKED ISSUE #5556: Make RetroReflection opened in spotlight by another member non-droppable

nickoferrall:
Fixes #5556 and #5621

### To test

- [ ] Unable to drop a reflection onto the animating remote Spotlight reflection.
- [ ] When the remote Spotlight is animating, if the other kanban reflections are dragged or grouped, the animating remote reflection adjusts its position.
- [ ] Clicking on the animating remote Spotlight reflection doesn't do anything (previously added a header with opacity 0, which moved the reflections in the kanban).
- [ ] Alice in Client A opens the Spotlight and sees "Dogs" as a reflection in the results. Bob in Client B opens his own Spotlight by clicking on "Dogs". Alice sees the "Dogs" card animate above the Spotlight.
- [ ] Alice opens the Spotlight. The reflection "Cats" is not in the Spotlight results; it's on the kanban behind the Spotlight. Bob opens the Spotlight by clicking on "Cats". "Cats" animates behind the Spotlight.
- [ ] Unable to click open Spotlight icon on animating remote reflection card.
- [ ] If the viewer can see the animating remote Spotlight reflection and then refreshes the page, they can still see the animating remote Spotlight reflection

### Additional tests from issue #5621: 

- [ ] It's not possible to drag a reflection that is underneath the animating remote Spotlight reflection
- [ ] Clicking a card that's underneath the animating remote reflection does not open the expanded reflection stack
- [ ] If there are two reflections in a group, a large one underneath a small one, and the viewer clicks the group to open the Spotlight, the larger reflection is clipped, i.e. it remains 44px in height.
- [ ] If Alice opens the Spotlight, drags the source, doesn't drop it on a reflection group, and returns the source to its position in the Spotlight modal, Bob sees the remote animation restart once the source is back in its original position. 

### Tests from Matt's review

- [ ] Search icon is not permanently stuck (on mobile)
- [ ] When spotlight hits its timeout, it closes
- [ ] Remote drag of a card in similarReflections starts in the expected position (see [comment](https://github.com/ParabolInc/parabol/pull/5610#issuecomment-969194831) for caveat)

Currently, if Alice opens the Spotlight and then Bob opens the Spotlight, the position of the animating remote reflection on Bob's screen will move as the Spotlight opens. This is fixed in: https://github.com/ParabolInc/parabol/pull/5514


packages/client/components/ReflectionGroup/ReflectionGroup.tsx
```diff
@@ -239,7 +246,7 @@ const ReflectionGroup = (props: Props) => {
                   dataCy={`${dataCy}-card-${staticIdx}`}
                   key={reflection.id}
                   staticIdx={staticIdx}
-                  isClipped={staticIdx !== 0}
+                  isClipped={staticIdx > 0}
```
nickoferrall, line 249: It looks like `isClipped` sets the maxHeight of a reflection card to `44px` [here](https://github.com/ParabolInc/parabol/blob/750457f997fe1f3c233450f13bdb40c4fff0e460/packages/client/components/ReflectionEditorWrapper.tsx#L64). We want this to happen when there's a stack of reflections, e.g. if we group a small reflection on top of a large one, we want the large one to shrink/be clipped. 

I'm not sure when we would want to clip a card that has a `staticIdx` of -1? By only clipping cards that have an idx > 0, we prevent the space behind the animating remote reflection from shrinking to 44px. 

If we revert this change and Alice opens a very large reflection in the Spotlight, Bob will see reflections move into the empty space underneath the animating remote reflection.

nickoferrall:
Hey @Dschoordsch, could you review this PR please?

Dschoordsch:
Testing for
> When the remote Spotlight is animating, there is an empty space under the remote reflection. Other reflections on the kanban don't move into the space underneath it.

failed
Peek 2021-11-04 13-13

The placeholder underneath the animation needs to continue being rearranged automatically.

packages/client/components/ReflectionGroup/RemoteReflection.tsx
```diff
@@ -148,15 +152,39 @@ interface Props {
...
+    {reflectionGroupId: spotlightGroup?.id || '', searchQuery: ''},
+    {fetchPolicy: 'store-only'}
+  )
+  const isInViewerSpotlightResults = useMemo(() => {
+    const {viewer} = spotlightSearchResults
+    const {similarReflectionGroups} = viewer
+    if (!similarReflectionGroups) return false
+    return !!similarReflectionGroups.find((group) => group.id === reflectionGroupId)
+  }, [spotlightSearchResults])
+
```
Dschoordsch, line 187: As I understand this, you're querying the spotlight search to find out whether or not the RemoteReflection is in the Viewer's spotlight, because positioning/zIndex work slightly different there?

I think if we need this information on a reflection to be able to display it correctly, then we should add it to reflection directly once they are included in the spotlight window. Having the same search query here again seems error prone.

Dschoordsch (changes requested):

packages/client/components/ReflectionGroup/RemoteReflection.tsx
nickoferrall: That's right - if Alice and Bob both open the Spotlight, we need to know whether Alice's remote Spotlight reflection exists in Bob's results. If it does, the zIndex needs to be higher than the modal zIndex, otherwise it should be less.

> we should add it to reflection directly once they are included in the spotlight window

I'm not sure I follow - are you suggesting that we adds the `groupIds` to Relay client's `reflection` object when we first make the query in `ResultsRoot`?

packages/client/mutations/EndDraggingReflectionMutation.ts
```diff
@@ -48,6 +48,7 @@ graphql`
       id
       dragUserId
       dragUserName
+      isSpotlight
```
nickoferrall, line 51: Add this so `remoteDrag?.isSpotlight` is no longer true when the Spotlight closes

packages/client/hooks/useDraggableReflectionCard.tsx
```diff
@@ -337,6 +354,9 @@ const useCollapsePlaceholder = (
       // the card is the only one in the group, shrink the group!
       style.height = scrollHeight + 'px'
       style.transition = `height ${Times.REFLECTION_DROP_DURATION}ms`
+      const {remoteDrag} = reflection
+      // do not collapse if remote opened spotlight
+      if (remoteDrag?.isSpotlight) return
```
nickoferrall, line 359: Moved this here so that the `style.height` is the `scrollHeight`. Otherwise, when the Spotlight closes and the animating remote reflection is dropping into the kanban, the reflection below think the height is 0 and briefly move into its position.

packages/client/hooks/useDraggableReflectionCard.tsx
```diff
@@ -59,8 +65,24 @@ const useRemotelyDraggedCard = (
         const newTimeRemaining = timeRemaining - (Date.now() - beforeFrame)
         setRemoteCard(isClose, newTimeRemaining, bbox.top)
       })
+    } else if (isSpotlight) {
+      // move animating remote Spotlight when other reflections move
+      spotlightAnimRef.current = requestAnimationFrame(() => {
+        const newTimeRemaining = timeRemaining - (Date.now() - beforeFrame)
+        setRemoteCard(isClose, newTimeRemaining, bbox.top, isSpotlight)
```
nickoferrall, line 72: Update the remote Spotlight reflection position so that when other cards in the kanban move, the animating Spotlight moves rather than hovers over another card

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -245,6 +247,7 @@ const ReflectionCard = (props: Props) => {
     !isSpotlightOpen &&
     !isComplete &&
     !isDemoRoute() &&
+    !isRemoteSpotlight &&
```
nickoferrall, line 250: Prevent clicking Open Spotlight on the animating remote Spotlight reflection

packages/client/components/ReflectionGroup/RemoteReflection.tsx
Dschoordsch: I don't like having to somehow pass `searchQuery` to `RemoteReflection`. This doesn't seem like the right thing to do.
I would either convert this into a `preloadedQuery` and pass the queryRef from `ResultsRoot` or simply pass `isInSimilarReflectionGroups` property down from `ResultsRoot` through `useRemoteLyDraggedCard` to `RemoteReflection`. Either way I don't think `searchQuery` belongs here, but @mattkrick might have a better idea here.

Dschoordsch (changes requested):
I found an interesting bug #5625 while testing, but that exists also on master.

I would like you to change the `visibility` and for the `spotlightSearchResults` we wait for Matt to give his opinion on it.

packages/client/components/ReflectionCard/ReflectionCard.tsx
nickoferrall: Nice, I really like that suggestion! Pushed & moved to Maintainer Review

packages/client/components/ReflectionGroup/ReflectionGroup.tsx
```diff
@@ -58,9 +58,9 @@ const ReflectionWrapper = styled('div')<{
     bottom: 0,
     left: 0,
     outline: 0,
-    opacity: isHidden ? 0 : undefined,
     transform: `translateY(${translateY}px) scaleX(${scaleX})`,
     transition: isHidden ? undefined : `transform ${Times.REFLECTION_DROP_DURATION}ms`,
+    visibility: isHidden ? 'hidden' : undefined,
```
nickoferrall, line 63: Updated following: https://github.com/ParabolInc/parabol/pull/5610#discussion_r744679525

Dschoordsch (approved):

mattkrick:
Breaking cases:

### Search icon is permanently stuck:
- [ ] have a column full of reflections
- [ ] drag the top reflection, see that the search icon is stuck there
- [ ] see loom around 6 or 7 minutes in

### Reflection doesn't go home in kanban column
- [ ] have a column full of reflections
- [ ] have 2 browsers open
- [ ] drag the top one & then release in browser A
- [ ] in browser B notice the top reflection goes all the way to the top of the viewport

mattkrick:
Loom: https://www.loom.com/share/ac14b15310f447b5bd4f2ef076f317f0

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -50,7 +50,7 @@ const SpotlightButton = styled(CardButton)<{showSpotlight: boolean}>(({showSpotl
   opacity: 1,
   position: 'absolute',
   right: 2,
-  visibility: showSpotlight ? 'visible' : 'hidden',
+  visibility: showSpotlight ? 'inherit' : 'hidden',
```
nickoferrall, line 53: Fixes the permanently stuck search icon on mobile

packages/client/components/ReflectionGroup/ReflectionGroup.tsx
```diff
@@ -92,22 +93,19 @@ const ReflectionGroup = (props: Props) => {
...
-    return isSpotlightOpen ? isDraggingSource || isSourceGroup : true // prevent grouping results into results
-  }, [spotlightGroup])
+  const spotlightGroupId = spotlightGroup?.id
+  const visibleReflections = useMemo(
+    () => reflections.filter(({id}) => !reflectionIdsToHide?.includes(id)),
+    [reflections, reflectionIdsToHide]
+  )
+  const isSpotlightSrcGroup = spotlightGroupId === reflectionGroupId
+  const isBehindSpotlight = !!(spotlightGroupId && openSpotlight)
+  const [hideSpotlightHeader, disableDrop] = useSpotlightReflectionGroup(
```
nickoferrall, line 103: I felt that using a hook to tidy away some of the logic would be better than creating a separate `SpotlightReflectionGroup` file because:
1. The kanban `reflectionGroup` would still need to know if the Spotlight is open for the `disableDrop` logic as well as the `isHiddenSpotlightSource` logic
2. There would be a lot of repeated code
3. With the hook, there's only 12 Spotlight related lines, some of which would be needed anyway due to point 1

I don't love that `hideSpotlightHeader` & `disableDrop` are largely unrelated but living in the same hook. I considered separating them but they both require `isRemoteSpotlightSrc` and keeping them in one hook reduces the length of the logic here.

nickoferrall:
Hey @mattkrick, I've pushed an update with fixes for 3/4 those breaking changes. `Reflection doesn't go home in kanban column` isn't a regression in this PR and I noticed that it still existed in v6.25.0, which was before any Spotlight logic was pushed, so I've created an issue for that instead: https://github.com/ParabolInc/parabol/issues/5657

I have noticed that if user A open the Spotlight, user B sees the animating remote reflection, but if user B refreshes the page, the animation stops. The animating remote reflection relies on the `isSpotlight` field in `remoteDrag`. I wonder whether reverting back to triggering the `UpdateDragReflectionMutation` after a certain interval would be a good approach for fixing this?

packages/client/hooks/useDraggableReflectionCard.tsx
nickoferrall: To test:
- open the Spotlight in client A and see that the remote reflection is animating in a circle in client B
- drag a reflection on the kanban that's positioned above the animating Spotlight 
- see that the animating remote reflection adjusts as the kanban reflections change position. Previously, the animating Spotlight didn't change position, meaning it would hover above other reflections on the kanban

In the Loom, you saw this bug, but then couldn't reproduce it later on. This was because you were on master initially and then switched to this branch, so it should be all fixed now.

packages/client/hooks/useAnimatedSpotlightSource.ts
```diff
@@ -43,11 +43,15 @@ const useAnimatedSpotlightSource = (
     }, 0)
     dragIdRef.current = clientTempId()
     // execute mutation after cloning as the mutation will cause reflection height to change
-    StartDraggingReflectionMutation(atmosphere, {
-      reflectionId,
-      dragId: dragIdRef.current,
-      isSpotlight: true
-    })
+    const dragInterval = setInterval(() => {
+      // execute every second so that the remote animation continues when subscribers refresh page
```
nickoferrall, line 47: To test: 
- remove interval
- client A opens Spotlight so client B sees remote animation 
- client B refreshes the page & the remote animation stops

packages/client/components/ReflectionGroup/RemoteReflection.tsx
nickoferrall: In the [Loom](https://www.loom.com/share/ac14b15310f447b5bd4f2ef076f317f0) ~34 mins, Matt mentioned a couple of possible approaches here but said that he'd probably keep it the way it is. It looks like the comment was typed but not sent at the end of the review. I'll resolve this conversation for now.

packages/client/components/ReflectionGroup/ReflectionGroup.tsx
```diff
@@ -92,22 +93,19 @@ const ReflectionGroup = (props: Props) => {
...
-    return isSpotlightOpen ? isDraggingSource || isSourceGroup : true // prevent grouping results into results
-  }, [spotlightGroup])
+  const spotlightGroupId = spotlightGroup?.id
+  const visibleReflections = useMemo(
+    () => reflections.filter(({id}) => !reflectionIdsToHide?.includes(id)),
+    [reflections, reflectionIdsToHide]
+  )
+  const isSpotlightSrcGroup = spotlightGroupId === reflectionGroupId
+  const isBehindSpotlight = !!(spotlightGroupId && openSpotlight)
+  const [isRemoteSpotlightSrc, disableDrop] = useSpotlightReflectionGroup(
```
nickoferrall, line 103: I think a hook is best here instead of creating a separate `SpotlightReflectionGroup` file because:

1. The kanban reflectionGroup would still need to know if the Spotlight is open for the `disableDrop` logic as well as the `isHiddenSpotlightSource` logic
2. There would be a lot of repeated code
3. Now that I've implemented the hook, there's only 12 Spotlight related lines, some of which would be needed even if we created a separate file due to point 1

I don't love that `isRemoteSpotlightSrc` & `disableDrop` are largely unrelated but live in the same hook. I considered separating them but `disableDrop` uses `isRemoteSpotlightSrc` and keeping them in one hook reduces the length of the logic here.

mattkrick:
LGTM!

PR #5617, opened 2021-11-04
mitigate safari popups on SSO

### LINKED ISSUE #5576: Behold the metaverse
A large enterprise is evaluating adopting Parabol. SSO was included in their evaluation.

We set up their SSO. Users authenticate against their SSO correctly, but the SSO popup does not close after they authenticate, and redirects to Parabol itself. This is a crummy user experience and they cannot complete their evaluation.​

- AC
   - Users contacted and determined if it is our SSO or their SSO configuration at fault ([see ticket](https://app.hubspot.com/contacts/3888472/ticket/603043895/) 🔒 )

- EE 2 hours

mattkrick:
Fix #5576
A user reported that SSO was broken in Safari.
This is because we open up the SSO window in a popup.
Safari only allows trusted events (e.g. a user's click) to open popups
The trusted event must open the popup synchronously, i.e. before any `await` command

If the user manually enables popups for that session by going to the address bar & clicking "allow popup" then the popup was opened by the browser instead of by our window, which means the popup's `window.opener === null`

To get around this, we quickly determine if we'll need to open a popup for auth.
If we do, we open an popup immediately.
Then, we can later go back to that popup & change the URL. 

TEST
- [ ] add a valid SAML record to your DB
- [ ] try to login as a person who is a part of that SAML domain the popup opens
- [ ] try to login as the same person with a username & password, it still works
- [ ] login as a non-SSO user with username & password, it works
- [ ] Perform the edge case seen in the code comments. The popup window quickly opens & closes, the UX is acceptable for it being such an edge case

 
PS you can ignore the connectionNames in the server files, that's for other debugging that we need to do & I failed to stash before I started working on this 😝 )


tianrunhe:
I tested it on my local. Two issues:

1.  Instead of a pop-up windows, Safari just open a new tab
2.  I guess because of 1, after I was authenticated with Okta, the new tab won't close and I'm not able to login

Loom [here](https://www.loom.com/share/42975012591e4c1798fcebf9858421a8)

tianrunhe (changes requested):
I tested it on my local. Two issues:

Instead of a pop-up windows, Safari just open a new tab
I guess because of 1, after I was authenticated with Okta, the new tab won't close and I'm not able to login

Loom [here](https://www.loom.com/share/42975012591e4c1798fcebf9858421a8)

mattkrick:
@tianrunhe ahhh, i can't reproduce!
i took a shot in the dark, could you check again & see if it works now?
if not, could you do a `console.log(getOAuthPopupFeatures({width: 385, height: 550, top: 64}))` and tell me what you get?

tianrunhe:
Unfortunately it still doesn't work. Here's the log:  
  
`popup,width=385,height=550,left=1953,top=875`  
  
I'm using Safari Version 15.1 (17612.2.9.1.20)

mattkrick:
@tianrunhe can you check your settings? https://support.apple.com/guide/safari/use-tabs-for-webpages-sfri11576/mac

Automatically open pages in tabs
In the Safari app  on your Mac, choose Safari > Preferences, then click Tabs.

Click the “Open pages in tabs instead of windows” pop-up menu, then choose an option:

Never: Links designed to open in new windows open in new windows.

Automatically: Links designed to open in new windows open in new tabs.

Always: All links designed to open in new windows open in new tabs, including links designed to open in specially formatted separate windows.

tianrunhe:
Safari development is hard, isn't it...? Alright I changed the setting to "Never" but I still have the issue:

https://www.loom.com/share/4c7cacf75d1748498366a645ed8ec75a

mattkrick:
wanted to make sure this got in to get some incremental benefit. we can open a new issue if the bug presents itself again for a user

Issue #5625, opened 2021-11-08
Exception when opening spotlight with 3 cards in 2 columns


Dschoordsch:
## How to reproduce
Start a Retrospective meeting and add exactly **3 cards** to the 'What's working?' column. Resize the window so it has a **two column layout**. Open spotlight.

![Screenshot from 2021-11-08 15-38-18](https://user-images.githubusercontent.com/7331043/140761985-f7a59265-3e94-4e9f-ba8a-cfe6b782f724.png)

## Expected
Spotlight opens

## Actual
Exception is thrown
```
TypeError
Cannot read properties of undefined (reading 'id')
Call Stack
 eval
  packages/client/components/SpotlightGroups.tsx:100:24
 SpotlightGroups
  packages/client/components/SpotlightGroups.tsx:99:18
 renderWithHooks
  %5Bname%5D/./node_modules/react-dom/cjs/react-dom.development.js:14985:18
 updateFunctionComponent
  %5Bname%5D/./node_modules/react-dom/cjs/react-dom.development.js:17356:20
 beginWork
  %5Bname%5D/./node_modules/react-dom/cjs/react-dom.development.js:19063:16
 HTMLUnknownElement.callCallback
  %5Bname%5D/./node_modules/react-dom/cjs/react-dom.development.js:3945:14
 Object.invokeGuardedCallbackDev
  %5Bname%5D/./node_modules/react-dom/cjs/react-dom.development.js:3994:16
 invokeGuardedCallback
  %5Bname%5D/./node_modules/react-dom/cjs/react-dom.development.js:4056:31
 beginWork$1
  %5Bname%5D/./node_modules/react-dom/cjs/react-dom.development.js:23959:7
```
This does not happen for 4 or 5 cards or when aligned in a single column

### Acceptance Criteria
* [ ] spotlight can be opened in this situation without a crash

**Estimated effort:** 1 hour to triage. More if root cause is already identified.

Issue #5622, opened 2021-11-05
Spotlight: fix source slowly returning to kanban


nickoferrall:
When the Spotlight closes, it takes around 500ms for the source reflection to reappear on the kanban.

### AC:

- [ ] When the Spotlight closes, the source reflection shows up on the kanban immediately

PR #5626, opened 2021-11-08
Fix source returning to kanban & 2 column error

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/5625 and https://github.com/ParabolInc/parabol/issues/5622

### To test

- [ ] When closing the Spotlight, the source animation is immediately returned to the kanban, i.e. there's no 500ms wait for it to return
- [ ] Can open the Spotlight if there's only 3 reflections in the kanban
- [ ] After dropping a result reflection on the source, the reflection isn't briefly added back to the results group
- [ ] After grouping a result reflection into the source and clicking on the source group to open the expanded reflection stack, closing the expanded reflection stack modal does not close the Spotlight modal

### DIFF
--- a/packages/client/components/SpotlightModal.tsx
+++ b/packages/client/components/SpotlightModal.tsx
@@ -192,6 +192,7 @@ const SpotlightModal = (props: Props) => {
               reflectionGroup={spotlightGroup}
               meeting={meeting}
               reflectionIdsToHide={reflectionIdsToHideRef.current}
+              expandedReflectionGroupPortalParentId='spotlight'
             />
           )}
         </Source>
--- a/packages/client/hooks/useGroupMatrix.ts
+++ b/packages/client/hooks/useGroupMatrix.ts
@@ -19,9 +19,10 @@ const useGroupMatrix = (
     const width = el?.clientWidth
     if (!width) return null
     let colCount = 1
+    const currentMaxColumns = Math.min(resultsGroups.length, MAX_SPOTLIGHT_COLUMNS)
     const getNextWidth = (count: number) =>
       ElementWidth.MEETING_CARD * count + ElementWidth.MEETING_CARD_MARGIN * (count - 1)
-    while (getNextWidth(colCount + 1) < width && colCount + 1 <= MAX_SPOTLIGHT_COLUMNS) {
+    while (getNextWidth(colCount + 1) < width && colCount + 1 <= currentMaxColumns) {
       colCount++
     }
     return colCount
--- a/packages/client/hooks/useSpotlightSimulatedDrag.tsx
+++ b/packages/client/hooks/useSpotlightSimulatedDrag.tsx
@@ -54,9 +54,6 @@ const useSpotlightSimulatedDrag = (
       const meetingProxy = store.get(meetingId)
       meetingProxy?.setValue(null, 'spotlightGroup')
       meetingProxy?.setValue(null, 'spotlightReflectionId')
-      const reflection = store.get(reflectionId)
-      // set isDropping to true so that the source is added back to its original position in kanban
-      reflection?.setValue(true, 'isDropping')
     })
     dragIdRef.current = undefined
     reflectionIdRef.current = undefined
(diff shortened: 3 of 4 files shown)


packages/client/components/SpotlightModal.tsx
```diff
@@ -192,6 +192,7 @@ const SpotlightModal = (props: Props) => {
               reflectionGroup={spotlightGroup}
               meeting={meeting}
               reflectionIdsToHide={reflectionIdsToHideRef.current}
+              expandedReflectionGroupPortalParentId='spotlight'
```
nickoferrall, line 195: Not sure if this was previously added and then removed while merging branches, but this prevents the Spotlight modal from closing when the expanded reflection stack closes

packages/client/hooks/useGroupMatrix.ts
```diff
@@ -19,9 +19,10 @@ const useGroupMatrix = (
     const width = el?.clientWidth
     if (!width) return null
     let colCount = 1
+    const currentMaxColumns = Math.min(resultsGroups.length, MAX_SPOTLIGHT_COLUMNS)
```
nickoferrall, line 22: Fixes https://github.com/ParabolInc/parabol/issues/5625. If there's only enough reflections for 1 or 2 columns, we don't want to return 3 columns

packages/client/hooks/useSpotlightSimulatedDrag.tsx
```diff
@@ -54,9 +54,6 @@ const useSpotlightSimulatedDrag = (
       const meetingProxy = store.get(meetingId)
       meetingProxy?.setValue(null, 'spotlightGroup')
       meetingProxy?.setValue(null, 'spotlightReflectionId')
-      const reflection = store.get(reflectionId)
```
nickoferrall, line 57: Looks like this is redundant code now. Removing this ensures that the source reflection returns to the kanban immediately once the Spotlight modal closes

packages/client/mutations/handlers/handleUpdateSpotlightResults.ts
```diff
@@ -30,11 +30,11 @@ const handleUpdateSpotlightResults = (
   const reflectionGroupId = reflectionGroup.getValue('id')
   const groupsIds = similarReflectionGroups
     .map((group) => group.getValue('id'))
-    .concat(reflectionGroupId)
+    .concat(spotlightGroupId)
   const isInSpotlight = groupsIds.includes(reflectionGroupId)
   const wasInSpotlight = groupsIds.includes(oldReflectionGroupId)
   // added to another group. Remove old reflection group
-  if (isOldGroupEmpty) {
+  if (isOldGroupEmpty && wasInSpotlight) {
```
nickoferrall, line 37: `wasInSpotlight` isn't strictly necessary but it prevents the function from firing if groups in the kanban are ungrouped

packages/client/mutations/handlers/handleUpdateSpotlightResults.ts
```diff
@@ -43,7 +43,9 @@ const handleUpdateSpotlightResults = (
     })
   }
   // ungrouping created a new group or was added to a group in the kanban
-  if ((reflectionsCount === 1 && wasInSpotlight) || (isOldGroupEmpty && !isInSpotlight)) {
```
nickoferrall, line 46: To test, add `isOldGroupEmpty` back and see the following:
- open Spotlight modal by clicking on a reflection called "kittens"
- add result "cats" to source "kittens" so the source group has 2 reflections
- ungroup "cats" without dropping it on another reflection group
- see that "cats" is not added to the results

nickoferrall:
Hey @BartoszJarocki, could you review this one please?

BartoszJarocki:
@nickoferrall sure, will do today 👍

packages/client/components/SpotlightModal.tsx
BartoszJarocki: yeah I've added that, we probably lost it along the way

BartoszJarocki (approved):
Nice work! All the scenarios work. I haven't found any new regressions. 👌

PR #5628, opened 2021-11-08
Updating all Atlassian auths with the same account id when refreshing the auth token

### LINKED ISSUE #5601: Update all atlassian refresh tokens for a given user
@Dschoordsch pointed this out during our 1:1 & we should fix it this sprint.

The latest jira changes mean a userId can only have 1 valid refresh token for a given app. that means whenever we update the refresh token for 1 team member, we need to also update it for all other team members.

AC
- [ ] when the refresh token is updated for a team member, it updates the refresh token across all teams
- [ ] if a parabol user has 2 separate jira integrations, the refresh token doesn't apply across both (ie you can't just apply the update across all integrations for the given userId.  it'll have to be for the jira user id.

Estimate: 4 hours

BartoszJarocki:
Fixes #5601 

To test, create 2 Parabol teams and integrate the same Jira account in each one.
- [x] check if after refreshing the token in one of them, the second one is updated to the same value as a first one

Question
Do we want to add an index to `userId` column in the `AtlassianAuth` table?


packages/server/dataloader/atlassianLoaders.ts
```diff
@@ -2,16 +2,15 @@ import DataLoader from 'dataloader'
 import {decode} from 'jsonwebtoken'
 import JiraIssueId from 'parabol-client/shared/gqlIds/JiraIssueId'
 import {JiraGetIssueRes, JiraProject} from 'parabol-client/utils/AtlassianManager'
-import getAtlassianAuthByUserIdTeamId, {
-  AtlassianAuth
-} from '../postgres/queries/getAtlassianAuthByUserIdTeamId'
+import {AtlassianAuth} from '../postgres/queries/getAtlassianAuthByUserIdTeamId'
```
BartoszJarocki, line 5: This is a bit concerning. As we're adding more and more queries, we end up having a lot of redundant generated types. How do we decide which one to use as a database layer type? For example, I've added `getAtlassianAuthsByUserId` which is basically the same type as generated by `getAtlassianAuthByUserIdTeamId`. I'm afraid this could go out of control and it'd be nice to agree on how we handle that.

packages/server/dataloader/atlassianLoaders.ts
```diff
@@ -34,37 +33,53 @@ export const freshAtlassianAuth = (parent: RethinkDataLoader) => {
     async (keys) => {
       const results = await Promise.allSettled(
         keys.map(async ({userId, teamId}) => {
-          const atlassianAuth = await getAtlassianAuthByUserIdTeamId(userId, teamId)
-
-          if (!atlassianAuth?.refreshToken) {
+          const userAtlassianAuths = await getAtlassianAuthsByUserId(userId)
+          const atlassianAuthToRefresh = userAtlassianAuths.find(
+            (atlassianAuth) => atlassianAuth.teamId === teamId
+          )
```
BartoszJarocki, line 39: I've changed this call to fetch all the user Atlassian auths to avoid another DB hit when getting Atlassian auths to update with the same Jira account id

packages/server/dataloader/atlassianLoaders.ts
```diff
@@ -34,37 +33,53 @@ export const freshAtlassianAuth = (parent: RethinkDataLoader) => {
...
 
-            await upsertAtlassianAuth(atlassianAuth)
+            await upsertAtlassianAuths([
+              {
+                ...atlassianAuthToRefresh,
+                accessToken,
+                refreshToken: updatedRefreshToken
+              },
+              ...updatedSameJiraAccountAtlassianAuths
+            ])
```
BartoszJarocki, line 76: Added upsert to handle multiple items. Not sure if we want to remove `upsertAtlassianAuth` that takes only one item and simply use this one everywhere or leave both. Either way, I'm happy to improve if needed!

BartoszJarocki:
@tiffanyhan may I ask you to take a look? Thank you!

packages/server/postgres/queries/src/getAtlassianAuthsByUserId.sql
```diff
@@ -0,0 +1,5 @@
+/*
+  @name getAtlassianAuthsByUserIdQuery
+*/
+SELECT * from "AtlassianAuth"
+WHERE "userId" = :userId AND "isActive" = TRUE;
```
tiffanyhan, line 5: > Do we want to add an index to userId column in the AtlassianAuth table?

Since we already have a composite index on `("userId", "teamId")`, this shouldn't be necessary. The way that postgres indexes are implemented under the hood, querying for `userId` will be able to use the composite index efficiently. If we were to query using only `teamId`, that's a different story. more info: https://stackoverflow.com/questions/11352056/postgresql-composite-primary-key

packages/server/dataloader/atlassianLoaders.ts
tiffanyhan: i think this would be great since both queries are so similar. can you refactor to remove the one that returns only a single query? it would be a good general pattern imo to write queries to support multiple values instead of only single.

packages/server/dataloader/atlassianLoaders.ts
tiffanyhan: i definitely share your worries as well. for this PR, i think it would be good to refactor to remove the single `upsertAtlassianAuth` to avoid repetition. to really address the concern, i think we should start a discussion in checkin meeting/slack or open a separate issue.  curious to hear matt's thoughts here.

packages/server/dataloader/atlassianLoaders.ts
```diff
@@ -34,37 +33,53 @@ export const freshAtlassianAuth = (parent: RethinkDataLoader) => {
...
+                refreshToken: updatedRefreshToken
+              }))
 
-            await upsertAtlassianAuth(atlassianAuth)
+            await upsertAtlassianAuths([
+              {
+                ...atlassianAuthToRefresh,
+                accessToken,
+                refreshToken: updatedRefreshToken
+              },
```
tiffanyhan, line 74: this update logic seems to be the same as lines 63-67. what do you think of merging all atlassianAuths to update into one list and then mapping them all once to updatedAuths together? 

i think we should also remember to do set `updatedAt` field to now.

tiffanyhan (changes requested):
can we refactor everything to use `upsertAtlassianAuths` and remove `upsertAtlassianAuth`?

can we also update all atlassianAuths `updatedAt = Date.now()`, and possibly refactor repeated update logic in `atlassianLoaders`?

have not done any manual tests, will do after changes are made.

packages/server/postgres/queries/src/getAtlassianAuthsByUserId.sql
BartoszJarocki: Thank you for the explanation! We came to the same conclusion with Matt on our 1:1 and it's really good to know it works that way. 👍

packages/server/dataloader/atlassianLoaders.ts
BartoszJarocki: @tiffanyhan it's set by the sql query itself

```
VALUES :auths ON CONFLICT ("userId", "teamId") DO
UPDATE
SET (
    "isActive",
    "updatedAt",
    "accessToken",
    "refreshToken",
    "cloudIds",
    "scope",
    "accountId",
    "teamId",
    "userId"
  ) = (
    TRUE,
    CURRENT_TIMESTAMP,
    EXCLUDED."accessToken",
    EXCLUDED."refreshToken",
    EXCLUDED."cloudIds",
    EXCLUDED."scope",
    EXCLUDED."accountId",
    EXCLUDED."teamId",
    EXCLUDED."userId"
  );
```

So I think we should be good, but please lmk if you see some advantages in setting that in the code 💡

tiffanyhan: oh i missed that, no this looks good!

packages/server/dataloader/atlassianLoaders.ts
BartoszJarocki: Done!

BartoszJarocki:
@tiffanyhan I've pushed all the requested changes. I've also fixed one more edge case (ie. when the user refreshes the token). Would be great to do another round, thank you!

tiffanyhan (approved):
lgtm! tested adding second integration to same jira account on different team, refreshing token on one of the teams, and triggering a refresh by simulating an expired token.

packages/server/dataloader/atlassianLoaders.ts
Dschoordsch: I would put the canonical version (i.e. get...ByUserId) into postgres/types and declare this as the return type. That's what I did for User anyway.

mattkrick: as far as type systems go, typescript is structural (vs nominal) so any object that has the same shape as the canonical version will still fulfill the contract.

so, for example, if `byUserId` exports `AtlassianAuth` then the return value of `getAtlassianAuthByUserIdTeamId` will work in all the functions that accept `AtlassianAuth`.
if we wanted to be fancy, the return type of `getAtlassianAuthByUserIdTeamId` could `extend AtlassianAuth` which would be a way to point to the canonical version without a runtime penalty or code comment

mattkrick (approved):
i've got nothing to add, this looks great!

Issue #5623, opened 2021-11-05
Spotlight: fix collapsed reflections in kanban


nickoferrall:
After opening the Spotlight, grouping and ungrouping reflections, and then closing the Spotlight, the viewer will sometimes see that the reflections in the kanban are stacked on top of each other. 

I suspect that the height of the ref of either the source and/or result reflections is set to 0 in `useDraggableReflectionCard`, and its height isn't restored once the Spotlight closes. This results in the following stacking effect:

<img width="317" alt="Screenshot 2021-11-05 at 18 03 06" src="https://user-images.githubusercontent.com/39854876/140578206-1ee42936-477a-4b0f-8aa7-042e038237b8.png">

### AC:

- [ ] After closing the Spotlight, the kanban reflections are not stacked on top of each other

Issue #5627, opened 2021-11-08
Spotlight: remotely ungrouping result group bug


nickoferrall:
A reflection doesn't return to the correct position in the Spotlight modal if it's remotely ungrouped in the results section.

To reproduce:
- Alice opens the Spotlight in client A. One of the result groups contains 2 reflections
- Bob, in client B, ungroups the result group that contains 2 reflections
- Alice sees that the reflection returns to the incorrect position in the result groups

### AC

- [ ] If a results reflection group is remotely ungrouped, it returns to the expected position in the Spotlight

nickoferrall:
This is caused by the`isFinishingRemoteDragging` check [here](https://github.com/ParabolInc/parabol/blob/2360f8cb7609ba483b2473c9e7635c204cdf7a85/packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx#L143) which was added to fix https://github.com/ParabolInc/parabol/issues/5415

We need to find a way of preventing remote reflections from losing their position when the Spotlight is opened while also making sure that result groups remain visible when they're remotely ungrouped.

nickoferrall:
Hey @BartoszJarocki, if you're looking for a new issue and you'd like to return to the painful world of the `DraggableReflectionCard` ref logic, this could be a good one 😅  No worries if not!

BartoszJarocki:
@nickoferrall challenge accepted ;) Seems like https://github.com/ParabolInc/parabol/issues/5415 won't let me forget about it too soon 😂

nickoferrall:
Hey @BartoszJarocki, I found a solution for this that's not related to the `isFinishingRemoteDragging` logic, but the `isDroppable` logic in `reflectionGroup` instead! I'll push an update, most likely on Monday, and tag this issue in the PR.

nickoferrall:
Spoke too soon - this is still related to the `isFinishingRemoteDragging` logic 😅

BartoszJarocki:
No problem, I'll most probably have a time on Monday to start the investigation 😅 Please let me know if anything changes! 🙏

nickoferrall:
Will do!

BartoszJarocki:
@nickoferrall After updating from the newest master the behavior has changed. Now the remote ungroup can be seen in the realtime but the card returns to the wrong position. Should I continue working on this or do we have a separate issue for this behavior? 
[...]

PR #5642, opened 2021-11-10
Fix collapsed reflections 

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/5623

When a reflection group is ungrouped, a new `reflectionGroupId` is created for the ungrouped reflection. When this happens, the new reflection group is added to the Spotlight [here](https://github.com/ParabolInc/parabol/blob/378944330900be841e7cc903974c0afc9a1c17f8/packages/client/mutations/handlers/handleUpdateSpotlightResults.ts#L47). However, momentarily, before the new reflection group is added to the Spotlight, the `reflectionGroupId` doesn't exist in the `resultGroupIdsInSpotlight` meaning `!isReflectionGroupIdInSpotlight` and therefore [isPriorityCard](https://github.com/ParabolInc/parabol/blob/45cae6676c2bf6051cd8042962e5bee243c04419/packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx#L151) are briefly true when they shouldn't be.

This means that the kanban reflection will be a priority card when it shouldn't be, `drag.ref` will be assigned to `c`, and the height will collapse [here](https://github.com/ParabolInc/parabol/blob/3af17a3c22980852d20e9c53e4171b8d94360341/packages/client/hooks/useDraggableReflectionCard.tsx#L341).

While the `reflectionGroupId` doesn't exist in the Spotlight, the `reflectionId` does. Checking if `isReflectionIdInSpotlight` is true fixes the issue.

### To test

- Alice opens the Spotlight in Client A. The Spotlight results include a reflection group that contains two reflections
- Bob, in Client B, ungroups the group in the Spotlight that contains two reflections
- Alice closes the Spotlight and sees that the kanban does not include any collapsed reflections (see screenshot in issue for an example).

Note: after remotely ungrouping, the group doesn't return to the correct position in the results. We have an open issue for this here: https://github.com/ParabolInc/parabol/issues/5627

### DIFF
--- a/packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx
+++ b/packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx
@@ -109,10 +109,16 @@ const DraggableReflectionCard = (props: Props) => {
   const {viewer} = spotlightSearchResults
   const {similarReflectionGroups} = viewer
   const isReflectionIdInSpotlight = useMemo(() => {
-    const resultReflectionIdsInSpotlight =
-      similarReflectionGroups?.flatMap(({reflections}) => reflections.map(({id}) => id)) || []
-    return [...resultReflectionIdsInSpotlight, spotlightReflectionId].includes(reflectionId)
-  }, [similarReflectionGroups, reflectionId])
+    return (
+      reflectionId === spotlightReflectionId ||
+      !!(
+        reflectionId &&
+        similarReflectionGroups?.find(({reflections}) =>
+          reflections.find(({id}) => id === reflectionId)
+        )
+      )
+    )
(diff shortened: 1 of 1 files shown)


nickoferrall:
Hey @BartoszJarocki, I pestered you with my last CR request, but as this is touching the `drag.ref` logic in `DraggableReflectionCard`, this PR might be relevant for your work. Mind giving it a review?

Dschoordsch (changes requested):
I tested it and it worked.

nickoferrall:
Thanks for the proactive review @Dschoordsch! Change pushed - moving to Maintainer Review

Dschoordsch (approved):

mattkrick:
LGTM

PR #5650, opened 2021-11-11
Spotlight search proposed changes

nickoferrall:
Proposed changes to the search query following some merge conflict fun.

### DIFF
--- a/packages/client/hooks/useGroupMatrix.ts
+++ b/packages/client/hooks/useGroupMatrix.ts
@@ -1,11 +1,11 @@
-import {SpotlightGroups_viewer} from './../__generated__/SpotlightGroups_viewer.graphql'
+import {SpotlightGroupsQueryResponse} from './../__generated__/SpotlightGroupsQuery.graphql'
 import {RefObject, useEffect, useState} from 'react'
 import {ElementWidth} from '../types/constEnums'
 import useResizeObserver from './useResizeObserver'
 import {MAX_SPOTLIGHT_COLUMNS} from '~/utils/constants'
 import useInitialRender from './useInitialRender'
 
-type Group = SpotlightGroups_viewer['similarReflectionGroups'][0]
+type Group = SpotlightGroupsQueryResponse['viewer']['similarReflectionGroups'][0]
 
 const useGroupMatrix = (
   resultsGroups: readonly Group[],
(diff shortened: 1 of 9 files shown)


packages/client/hooks/useGroupMatrix.ts
```diff
@@ -1,46 +1,46 @@
...
+    while (getNextWidth(colCount + 1) < width && colCount + 1 <= currentMaxColumns) {
       colCount++
     }
     return colCount
   }
 
-  const initMatrix = (): Group[][] => {
+  const initMatrix = () => {
     const columnsCount = getColumnsCount()
-    if (columnsCount === null) return []
```
nickoferrall, line 32: When we update the search query, `queryRef` is briefly undefined which means we don't render the results. This means that the `resultsRef` is undefined so the groupMatrix is set to `[]`. To prevent this, I've moved `initMatrix` to the `useEffect` and only update state if `columnsCount` isn't null.

packages/client/hooks/useGroupMatrix.ts
```diff
@@ -56,40 +56,38 @@ const useGroupMatrix = (
...
-        .filter(({length}) => length)
+    const newGroups = resultsGroups.filter(({id}) => !matrixGroupsIds.includes(id))
+    const resultsGroupsIds = resultsGroups.map(({id}) => id)
+    const removedGroupsIds = matrixGroupsIds.filter((id) => !resultsGroupsIds.includes(id))
+    if (!removedGroupsIds.length && !newGroups.length) return
+    // grouping results or updating the search query can remove groups
+    const filteredMatrix = groupMatrix
+      .map((row) => row.filter(({id}) => !removedGroupsIds.includes(id)))
+      .filter(({length}) => length) // remove column if empty
+    if (!newGroups.length) {
```
nickoferrall, line 67: Previously, if a results group was grouped/ungrouped it would be added or removed from the matix. With the search query, many groups could be added or removed, so I've updated the logic.

nickoferrall (commented):
Hey @tianrunhe, I've made these changes in a separate PR so you can choose whether or not to accept them. 

I think one issue that'll need to be resolved before the search query can be merged is the modal flashes after each keystroke. We discussed this before and I think it's become more noticeable after the component order was changed. I'll continue to look for a solution now, but feel free to merge this PR into yours if you're happy with it.

packages/client/hooks/useQueryLoaderNow.tsx
```diff
@@ -1,23 +1,48 @@
...
 ) => {
   const [queryRef, loadQuery] = useQueryLoader<TQuery>(preloadableRequest)
   const varRef = useRef(variables)
   const atmosphere = useAtmosphere()
+  const environment = useRelayEnvironment()
   if (!areEqual(variables, varRef.current)) {
     varRef.current = variables
   }
+
+  const refreshQuery = useCallback(() => {
```
nickoferrall, line 27: This logic prevents the results from suspending on each keystroke, which causes the modal to flash. It could probably be cleaned up a little - I don't like the name `refreshQuery` for example when we have `refresh` lower down in the file

nickoferrall:
The last two commits fixed the modal flashing on each keystroke 🎉 

One final change could be preventing the height of the modal from changing when the results change. One way I was thinking about implementing that was passing the search query into [useResultsHeight](https://github.com/ParabolInc/parabol/blob/541b938b1fd92fec20a00b38b0a6c11b050fd279/packages/client/hooks/useResultsHeight.ts) and preventing the height from updating if `spotlightSearchQuery !== ''`, but it's up to you. 

I'll pass the baton back to you now!

tianrunhe (approved):

PR #5652, opened 2021-11-12
make atlassian oauth2 response spec compliant

mattkrick:
@BartoszJarocki since we've got a handful of new integration work to do i figured we should probably set a good example pattern for others to follow. this is a little more verbose, but now it's clear that a success payload can't have an error

spec grabbed from https://www.oauth.com/oauth2-servers/access-tokens/access-token-response/


BartoszJarocki (approved):
Great work!

PR #5656, opened 2021-11-15
recently used emojis are managed using local storage

### LINKED ISSUE #3639: emoji mart fixups

sinansonmez:
Fixes #3639

- [x] autofocus on open
- [x] persist frequently used in the local storage

I also added number of usage for each emoji, so in the future I can improve and sort the emojis by usage and show the sorted version to user. 

I couldn't complete the last task from issue (i.e.: consider allowing folks to tab through options in the menu (currently pressing tab closes the menu altogether)), if I understood correctly, this would require some changes in the library itself (discussed in this [PR](https://github.com/missive/emoji-mart/pull/431) for library and maintainer doesn't approve the changes)

Note: As I am new to coding and recently completed my bootcamp, please guide me for any changes to improve this PR


packages/client/utils/reactjiPickerUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+  rocket: 6,
+  fire: 5,
+  white_check_mark: 4,
+  confused: 3,
+  cry: 2,
+  x: 1
+}
+
+const getRecent = () => {
+  const localStorageRecent = localStorage.getItem('recent')
```
ackernaut, line 14: I imagine ‘recent’ could be more specific, like ‘recentReactjiPicker’

ackernaut: Or ‘frequentlyUsedReactji’

sinansonmez: I agree. As an improvement,I can create a constant variable, as I am using it in two different places.

packages/client/utils/reactjiPickerUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+  return localStorageRecent === null ? DEFAULT : JSON.parse(localStorageRecent)
+}
+
+const handleRecent = (emoji) => {
```
Dschoordsch, line 18: Please add a type to the argument since our goal is to have noImplicitAny rule turned on.

packages/client/utils/reactjiPickerUtils.ts
```diff
@@ -0,0 +1,28 @@
+const DEFAULT = {
```
Dschoordsch, line 1: Better name would be `DEFAULT_ORDER`

packages/client/utils/reactjiPickerUtils.ts
Dschoordsch: Naming can be improved, like `updateUsage`

packages/client/utils/reactjiPickerUtils.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const handleRecent = (emoji) => {
+  const frequentlyUsed = getRecent()
+  const {id} = emoji
+  frequentlyUsed[id] || (frequentlyUsed[id] = 0)
+  frequentlyUsed[id] += 1
+  localStorage.setItem('recent', JSON.stringify(frequentlyUsed))
+}
+
+const getRecentArray = () => Object.keys(getRecent())
```
Dschoordsch, line 26: This will not order the array by usage. Also it's not returning all keys if only some were set. It would be best to order all keys with usage by their order and then followed by the remaining keys from `DEFAULT_ORDER`. Since you're using this as learning exercise I leave that for you ;-)

Naming could also be improved, like `getRecentOrder`.

packages/client/components/ReactjiPicker.tsx
```diff
@@ -6,6 +6,7 @@ import {PALETTE} from '~/styles/paletteV3'
 import {MenuProps} from '../hooks/useMenu'
 import Menu from './Menu'
 import appleEmojis from 'emoji-mart/data/apple.json'
+import {getRecentArray, handleRecent} from '../utils/reactjiPickerUtils'
```
Dschoordsch, line 9: nit (nit-pick, feel free to ignore): `...Utils` is quite generic. Since this is for handling recents logic only, a better name would be `utils/reactjiPickerRecents`

Dschoordsch (changes requested):
Thank you for taking this on!

sinansonmez:
Thanks for the comments @Dschoordsch (very useful also for my future PRs). I updated this PR according to your comments. Please let me know if any further change is needed. I also added a screen record below to show the functionality after improvement

https://user-images.githubusercontent.com/37421564/142174116-7ad91e03-120e-455f-b9fd-b51fc6522813.mov

packages/client/utils/reactjiPickerRecent.ts
```diff
@@ -0,0 +1,40 @@
...
+  frequentlyUsed[emojiId] || (frequentlyUsed[emojiId] = 0)
+  frequentlyUsed[emojiId] += 1
+  localStorage.setItem(RECENT_REACTJI_PICKER, JSON.stringify(frequentlyUsed))
+}
+
+const getRecentArray = () => {
+  const frequentlyUsed = getRecent()
+
+  const sortedFrequentlyUsed: [string, number][] = []
+  for (const emoji in frequentlyUsed) sortedFrequentlyUsed.push([emoji, frequentlyUsed[emoji]])
```
Dschoordsch, line 31: Since we want to have all emojis in the result, we need to iterate over the keys of `DEFAULT_ORDER` in order to not miss any.
For style it would be nicer to use `map` instead, so together this could be
```
const sortedFrequentlyUsed = Object.keys(DEFAULT_ORDER).map(
  (emoji) => [emoji, frequentlyUsed[emoji] ?? 0]
)
```
If there is no value stored, we set the count to 0.

packages/client/utils/reactjiPickerRecent.ts
```diff
@@ -0,0 +1,40 @@
...
+  sortedFrequentlyUsed.sort((a, b) => a[1] - b[1]).reverse()
```
Dschoordsch, line 32: Instead of calling `reverse()` at the end, you can simply invert the sorting function `(a, b) => b[1] - a[1]`

packages/client/utils/reactjiPickerRecent.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const sortedFrequentlyUsedKeys: string[] = []
+  for (const emoji of sortedFrequentlyUsed) sortedFrequentlyUsedKeys.push(emoji[0])
```
Dschoordsch, line 35: nit: here again calling `map` to transform the array would be nicer

Dschoordsch (changes requested):

packages/client/utils/reactjiPickerRecent.ts
sinansonmez: I agree for the map for better styling. However, instead of `DEFAULT_ORDER`, I need to iterate over `frequentlyUsed` keys. Because in `getRecent()` I am using `DEFAULT_ORDER` if there is no value in local storage. Then I am adding new emojis on top of `DEFAULT_ORDER` in local storage
```
const getRecent = () => {
   const localStorageRecent = localStorage.getItem(RECENT_REACTJI_PICKER)
   return localStorageRecent === null ? DEFAULT_ORDER : JSON.parse(localStorageRecent)
 }
```

packages/client/utils/reactjiPickerRecent.ts
sinansonmez: Thanks for the hint. I was skeptical to `reverse()` when I was using (thinking myself: there should be better way 😄 )

packages/client/utils/reactjiPickerRecent.ts
sinansonmez: Thanks. I replaced `for loop` with `map` and deleted redundant `sortedFrequentlyUsedKeys` variable and return directly `map`

packages/client/utils/reactjiPickerRecent.ts
Dschoordsch: You are right, I missed that in `updateUsage` you initialize the local storage with the full map. I thought we only store the values which are > 0.

packages/client/utils/reactjiPickerRecent.ts
```diff
@@ -0,0 +1,39 @@
...
+const getRecentArray = () => {
+  const frequentlyUsed = getRecent()
+
+  const sortedFrequentlyUsed = Object.keys(frequentlyUsed).map((emoji) => [
+    emoji,
+    frequentlyUsed[emoji] ?? 0
+  ])
+  sortedFrequentlyUsed.sort((a, b) => b[1] - a[1])
+
+  return Object.keys(sortedFrequentlyUsed).map((emoji) => sortedFrequentlyUsed[emoji][0])
```
Dschoordsch, line 36: This can just be `sortedFrequentlyUsed.map(([emoji]) => emoji).slice(0, MAX_RECENT_EMOJIS)`. We slice the array so over time the recent list doesn't grow too much. I think starting with `MAX_RECENT_EMOJIS = 9` is reasonable.

Dschoordsch (changes requested):
I tested it and it works fine. But we should limit the max size. If you've done the two changes I requested we can move this forward to maintainer review.
In that stage our maintainers have a final look before merging.

packages/client/utils/reactjiPickerRecent.ts
```diff
@@ -0,0 +1,39 @@
...
+  frequentlyUsed[emojiId] += 1
+  localStorage.setItem(RECENT_REACTJI_PICKER, JSON.stringify(frequentlyUsed))
+}
+
+const getRecentArray = () => {
+  const frequentlyUsed = getRecent()
+
+  const sortedFrequentlyUsed = Object.keys(frequentlyUsed).map((emoji) => [
+    emoji,
+    frequentlyUsed[emoji] ?? 0
```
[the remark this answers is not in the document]
sinansonmez: Yeah, you are right. I also added `as [string, number][]` to explicitly define types

packages/client/utils/reactjiPickerRecent.ts
sinansonmez: I thought `MAX_RECENT_EMOJIS` can be 18 to limit frequently used to 2 rows max but 9 is also reasonable to have just 1 row. After number of iteration, the code is very clean now thanks to you 🚀

Dschoordsch (approved):
I noticed a quirk: if you choose an emoji twice for the same card, it is counted as double usage while in actuality the second time it's chosen it will be removed from the card.
No need to do anything about it though.

Dschoordsch:
@mattkrick can you do the maintainer review?

packages/client/utils/reactjiPickerRecent.ts
```diff
@@ -0,0 +1,35 @@
...
+  cry: 0,
+  x: 0
+}
+
+const RECENT_REACTJI_PICKER = 'recentReactjiPicker'
+const MAX_RECENT_EMOJIS = 9
+
+const getRecent = () => {
+  const localStorageRecent = localStorage.getItem(RECENT_REACTJI_PICKER)
+  return localStorageRecent === null ? DEFAULT_ORDER : JSON.parse(localStorageRecent)
```
mattkrick, line 18: `localStorageRecent` may not be JSON if the value is editing outside of our app (or if we introduce a bug)
if we don't wrap it in try/catch then the whole app could crash.

In order to guard against edge cases, I'd probably use the pre-made solution. emoji-mart already has this functionality built in (see https://github.com/missive/emoji-mart#storage)

I'd propose just adding the following to the parent component & getting rid of the recent array

```ts
  useEffect(() => {
    const existingRecent = localStorage.getItem('emoji-mart.frequently')
    if (existingRecent) return
    localStorage.setItem(
      'emoji-mart.frequently',
      JSON.stringify({
        heart: 0,
        tada: 0,
        smile: 0,
        rocket: 0,
        fire: 0,
        white_check_mark: 0,
        confused: 0,
        cry: 0,
        x: 0
      })
    )
  }, [])
```

mattkrick (changes requested):

packages/client/utils/reactjiPickerRecent.ts
sinansonmez: Thanks for the feedback. Initially, I didn't use their local storage because I couldn't find a proper way to override their default emoji selection. However using `useEffect` in the parent component makes total sense.

sinansonmez:
PR is updated according to comments

mattkrick:
LGTM!

PR #5667, opened 2021-11-17
Fix kanban flash when remotely ungrouping

### LINKED ISSUE #5645: Spotlight: remotely ungrouping can cause kanban flash
Notice the kanban columns height changes after ungrouping:

remote-ungroup-flash

### AC

- [ ] Remotely ungrouping a result in the Spotlight does not cause the kanban to flash

nickoferrall:
Fix #5645 

### To reproduce

- Remove the Suspense change in this PR
- Open Client A and create a reflection group that contains two reflections
- In Client B, ungroup the reflection group
- In Client A, you will often see that the kanban briefly flashes (see gif in issue). This also happens when ungrouping reflections in the Spotlight

@tianrunhe would you mind reviewing this one, please?


tianrunhe (approved):
easy peasy!

mattkrick:
this worries me because AFAIK there is no child component of ReflectionGroup that lazily loads.
Do we know why this happens? My concern is that wrapping the component in a suspense just masks the problem (ie the suspense force renders nothing for 1 animation frame, which basically means instead of it rendering the wrong thing, it renders nothing). Before merging I want to make sure that we're not having the DOM paint something incorrectly & then just hiding the problem.

To debug, you'll need to catch it in the act:
- console.log the column height whenever the column re-renders
- next, set up a debugger whenever the column height equals that erroneous height
- now the DOM is frozen in the erroneous state, inspect some elements & see what exists that shouldn't. my guess is we're hiding the card asynchronously when we should be setting that css value synchronously & not flushing until it's correct

nickoferrall:
@mattkrick the column height remained the same in each console log, but I was able to get the debugger working by adding it to the fallback option in the Suspense.

With the debugger on, I didn't find an element that existed that shouldn't, but I have noticed that on Chrome, Firefox, and Safari, this bug only happens when the dev tools are open. I suspect that this is because the browser can be less performant when the dev tools are open so when the new reflection group is created, perhaps the update doesn't happen as efficiently and ReflectionGroup is still loading.

I've removed as much code as possible to see if there could be an async update happening anywhere else that could be causing the problem, but it seems that `handleAddReflectionGroupToGroups`, which is triggered by the updater in `EndDraggingReflectionMutation` is causing the mischief.

mattkrick:
i can't seem to reproduce this on master.
i noticed in your screenshot that ungrouping caused 2 errors to occur
is it possible that the thing that caused those errors (relay erroneously suspended) caused the flicker? 
could you retest on master & see if you can still reproduce?

nickoferrall:
You're right! Perhaps refactoring `ReflectionGroup` to hooks fixed the suspense issue. I'll close this issue & PR now 👍

PR #5674, opened 2021-11-17
fix relay error on drag end

mattkrick:
In dev, relay would print some errors saying the same object in the same payload had different properties.

specifically, `reflection.reflectionGroupId` was different from `reflectionGroup.reflection.reflectionGroupId` in the endDragging payload.

the root cause was a dataloader with stale data.
@nickoferrall you might be interested in this


nickoferrall (approved):
Nice, LGTM!

Issue #5333, opened 2021-08-30
Add Datadog APM for GraphQL


mattkrick:
fixes #5189

We've had dev sin the past try to add datadog APM & it's made our whole system very slow.
If we try it again, we should be really targeted with it.

AC
- [ ] add the datadog graphql apm for graphql ONLY (no redis, no nothing else, turn off all defaults)
- [ ] make sure it works graphql-jit. if an out-of-the-box solution doesn't work, then we should check out the sdk & just make the same calls.

Estimate: 16 hours

Issue #5468, opened 2021-09-29
Use graphql-jit for subscriptions


mattkrick:
the latest version of graphql-jit supports subscriptions :tada: https://github.com/zalando-incubator/graphql-jit/issues/74

that means we don't have to have different caching strategies for subscribe & execute calls. Should relieve a few CPU cycles on our server

AC
- [ ] subscriptions are compiled & cached, just like executions

Estimate: 4 hours

mattkrick:
not a pre-req, but depending on how hard the implementation of #5333 is, it might be worthwhile to do this first.

pros of doing this first:
- subs are handled the exact same

cons: 
- additional change in how things work

Dschoordsch:
At the moment graphql-jit does not expose `createSourceEventStream` (https://github.com/zalando-incubator/graphql-jit/blob/f522b86bb2d776f5b201734f7be9c500838b3961/src/execution.ts#L1729). It's not too much work to expose it and create an upstream PR for it.
However the use case for `createSourceEventStream` is quite different than for `query` and `subscribe`. In the latter we need to run all field resolvers, thus it is beneficial to compile it beforehand. For the source event stream we're only using the result of the fields `subscribe` function while running the resolvers in a different process. As this is the use case source event streams are made for, it is counterproductive to compile all resolvers.
A separate implementation next to `compileQuery` which just does `compileSourceEventStream` could yield a small benefit as parsing the document (not relevant for us as it's cached) and preparing the execution context could be saved.

Dschoordsch:
I've did some performance measurements and it appears there is no significant performance increase by using `createSourceEventStream` from graphql-jit.

mattkrick:
this is a really fun problem to think about because I've often wondered if we could remove graphql from the web/socket server completely.

what if each websocket subscribed to its own redis pubsub channel & the graphql executor handled source & response streams & then sent the final result to the websocket server?

not something we need to change anytime soon, but it's a fun problem to consider

Dschoordsch:
Since I didn't want my work go to waste I added some benchmark on graphql-jit side. Although I couldn't see any performance improvement when running our app, just looking at `createSourceEventStream` I gained between 50% and 100%. I created a PR to discuss whether or not the owners are interested in this change.

Dschoordsch:
@mattkrick That sounds like a fun problem indeed.
What we have now:
[...]

PR #5681, opened 2021-11-18
Add datadog tracing to compiled graphql queries

Dschoordsch:
Adding the tracing functions is a 2 step process. Step 1 wraps all field
resolvers in the schema itself. Step 2 wraps the CompiledQuery to
initialize the tracing.
The tracing funcitonality itself is heavily inspired by dd-trace-js
graphql plugin.

This changeset will only enable tracing for graphql-jit compiled queries. It can be added to subscriptions after #5468.

Resolves: #5333


packages/gql-executor/tracer.ts
```diff
@@ -1,14 +1,17 @@
...
-  hooks: {
-    execute: (span, args) => {
-      span.setTag('viewerId', args?.contextValue?.authToken?.sub ?? "null")
+if (process.env.DD_TRACE_ENABLED === 'true') {
+  tracer.init({
+    plugins: false,
+    debug: false
+  })
+  tracer.use('graphql', {
+    enabled: true,
```
Dschoordsch, line 9: Since the resolvers in the schema are patched for graphql-jit, this plugin might be broken.

tianrunhe (changes requested):
Tested locally, works well. Great work! Two suggestions:
1. Comparing the output between this version vs. vanilla DataDog GraphQL plugin, it looks like this version is missing the query doc on the resource column. I think with the query content there is super helpful when we investigate performance issue later:
Screen Shot 2021-11-22 at 3 09 18 PM
2. Similarly, is it possible to include the `variables` in the query doc, or maybe tags?

nickoferrall:
Great PR, @Dschoordsch! I'm trying to get a better understanding of how it's working. Did the old implementation slow down the app because we were tracing all queries, including cached queries, in `executeGraphQL`? Whereas now that we're wrapping the schema, it will only trace queries that aren't cached and require a network request?

Dschoordsch:
The previous version required us to use plain `graphql` while now we can use the compiled queries from `graphql-jit`. This brings some performance boost as the field resolvers don't need to be collected etc. This is what we normally use anyways but didn't work with dd-trace.
I think we also ran with the default config which enables `signature` for the tracing, which is an expensive operation and it was done again for each query.
Tracing is still done for all queries and it's not super cheap, maybe there is still some room for improvement, or we decrease the sample factor, but the 2 biggest issues should be solved.

nickoferrall:
Makes sense, thanks for the explanation!

Dschoordsch:
@tianrunhe I addressed the resource name and added the `viewerId` hook back. I'm not sure if that's what you meant with adding tags?

tianrunhe (approved):
This is good enough! All I want is that when we have this in production and say we see a query/mutation takes much longer than expected to execute, we could dive into:

1.  What that query/mutation is about? I think now by having the query name, we could look that up in the code. Would be easier to have the entire query there as the vanilla GraphQL plugin, but I think what we have here is good enough and additionally it saves a lot of traffic sent to DataDog.
2.  Does the query perform badly in general or is it just slow in certain scenario (e.g., a meeting with hundreds of users)? That's my initial request on query variables but I agree with your points on avoiding sending sensitive data to DataDog. Having the viewerId there will definitely help us!

GREAT work! Can't wait to see this in production!

packages/server/graphql/traceGraphQL.ts
```diff
@@ -0,0 +1,357 @@
...
+
+function wrapFieldResolve(tracer: Tracer, config: Config, field?: GraphQLField<any, any>) {
+  if (!field || !field.resolve) return
+  field.resolve = wrappedResolve(tracer, config, field.resolve)
+}
+
+function wrapFieldType(tracer: Tracer, config: Config, field?: GraphQLField<any, any>) {
+  if (!field || !field.type) return
+
+  let unwrappedType = field.type
```
mattkrick, line 211: can use graphql built-in

```ts
  const namedType = getNamedType(field.type)
  wrapFields(tracer, config, namedType)
```

packages/server/graphql/traceGraphQL.ts
```diff
@@ -0,0 +1,357 @@
...
+}
+
+function withCollapse(responsePathAsArray: typeof pathToArray) {
+  return function (path: Path) {
+    return responsePathAsArray(path).map((segment) => (typeof segment === 'number' ? '*' : segment))
+  }
+}
+
+function isOperationDefinitionNode(node: DefinitionNode): node is OperationDefinitionNode {
+  return ['query', 'mutation', 'subscription'].includes((node as OperationDefinitionNode).operation)
```
mattkrick, line 344:
```ts
  return node.kind === 'OperationDefinition'
```

packages/server/graphql/traceGraphQL.ts
```diff
@@ -0,0 +1,357 @@
...
+  return function compileTracedQuery(
+    schema: GraphQLSchema,
+    document: DocumentNode,
+    operationName?: string,
+    partialOptions?: Partial<CompilerOptions>
+  ) {
+    wrapSchema(tracer, config, schema)
+    const query = compileQuery(schema, document, operationName, partialOptions)
+    if (isCompiledQuery(query)) {
+      const operation = getOperation(document, operationName)
```
mattkrick, line 72: having getOperation return `OperationDefinitionNode | undefined` is correct. 
however, in this specific case, i'd do a nonnull coercion inside this conditional because if the query compiled, then we are guaranteed that there is one matching definition & then we can get rid of 1 of the optional chains


```ts
      const operation = getOperation(document, operationName)!
```

mattkrick (changes requested):
I wish I had more feedback to give, but all the changes are pretty minor. It's really solid work! I'll do some testing with it tomorrow morning

mattkrick:
Testing complete!
Results here: https://app.datadoghq.com/apm/traces?query=%40_top_level%3A1%20service%3Aparabol-local-dev%20env%3Alocal&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&historicalData=false&messageDisplay=inline&sort=desc&streamTraces=true&start=1638375774866&end=1638376674866&paused=false

Extra thoughts:
- [ ] Let's add `DD_TRACE_ENABLED='false'` to the `.env.example`, to signify that the var is used in the app
- [ ] We should either add docs or add an entry to our `dev.yml` docker-compose setup to make this easier to setup locally. As a starting point: `docker run -d --name dd-agent -v /var/run/docker.sock:/var/run/docker.sock:ro -v /proc/:/host/proc/:ro -v /sys/fs/cgroup/:/host/sys/fs/cgroup:ro --env-file .env -p 8126:8126 gcr.io/datadoghq/agent:7`. I then just call `yarn build && yarn start` to run in prod so it kicks in
- [ ] I noticed the webserver is using the tracer from dd-trace instead of `./tracer.ts`, which means it's ignoring the env var.  To remove confusion, should we delete `tracer.ts` & just call `init` directly on each server? https://github.com/ParabolInc/parabol/blob/4406819325be2c2f9c6362504bccc044dec053b8/packages/server/server.ts

For your consideration:
```ts
  tracer.init({
    enabled: process.env.DD_TRACE_ENABLED === 'true',
   // might be worthwhile knowing which service it came from
    service: `GQLExecutor ${process.env.SERVER_ID}`,
    plugins: false
  })
```

- [ ] What do you think about

packages/server/graphql/traceGraphQL.ts
```diff
@@ -0,0 +1,357 @@
...
+  _datadog_graphql: {
+    span: Span
+    fields: {[name: string]: Field}
+    source?: string
+  }
+}
+
+type PatchedGraphQLObjectType = GraphQLOutputType & PatchedMarker
+
+function wrapFields(tracer: Tracer, config: Config, type?: PatchedGraphQLObjectType | null) {
```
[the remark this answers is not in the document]
Dschoordsch: good to know

packages/server/graphql/traceGraphQL.ts
```diff
@@ -0,0 +1,357 @@
...
+
+  const responsePathAsArray = config.collapse ? withCollapse(pathToArray) : pathToArray
+
+  async function resolveWithTrace(source, args, context, info) {
+    if (!context._datadog_graphql) return resolve(source, args, context, info)
+
+    const path = responsePathAsArray(info && info.path)
+
+    if (config.depth !== undefined && config.depth >= 0) {
+      const depth = path.filter((item) => typeof item === 'string').length
```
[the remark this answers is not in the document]
Dschoordsch: Right, that's from dd-trace. I think depth only increases for fields of GraphQLObjectType because those can have separate resolve functions. Having lists of lists doesn't increase the depth of those resolvers.

Dschoordsch:
I tested locally with the dev server as dd agent, this way we don't add unwanted infrastructure to datadog.
DD_TRACE_ENABLED is a standard variable from datadog and evaluated if enabled is not passed to init. It's already present in .env.example.

packages/server/graphql/traceGraphQL.ts
```diff
@@ -0,0 +1,361 @@
...
+  if (process.env.DD_TRACE_ENABLED !== 'true') {
+    return compileQuery
+  }
+  return function compileTracedQuery(
+    schema: GraphQLSchema,
+    document: DocumentNode,
+    operationName?: string,
+    partialOptions?: Partial<CompilerOptions>
+  ) {
+    wrapSchema(tracer, config, schema)
```
mattkrick, line 71: I like how you call `wrapSchema` here instead of in another file like `rootSchema`. Feels more modular.

Ideally we could avoid traversing the schema for every new compiled query by flagging the entire schema as "wrapped".

Additionally, if we could guarantee that each type and type field is visited exactly once, then we could remove the recursive calls and the `PatchedMarker` logic.

For your consideration:

```ts
let isSchemaWrapped = false
function wrapSchema(tracer: Tracer, config: Config, schema: GraphQLSchema) {
  if (isSchemaWrapped) return
  isSchemaWrapped = true
  const typeMap = schema.getTypeMap()
  Object.values(typeMap).forEach((namedType) => {
    // ignore introspection and scalar types
    if (namedType.name.startsWith('__') || !isObjectType(namedType)) return
    const fields = namedType.getFields()
    Object.values(fields).forEach((field) => {
      if (field.resolve) {
        field.resolve = wrappedResolve(tracer, config, field.resolve)
      }
    })
  })
}
```

packages/gql-executor/gqlExecutor.ts
```diff
@@ -1,10 +1,16 @@
...
 import Redis from 'ioredis'
 import {ServerChannel} from 'parabol-client/types/constEnums'
 import GQLExecutorId from '../client/shared/gqlIds/GQLExecutorId'
 import executeGraphQL, {GQLRequest} from '../server/graphql/executeGraphQL'
 import '../server/initSentry'
 import RedisStream from './RedisStream'
-import './tracer'
+
+tracer.init({
+  enabled: process.env.DD_TRACE_ENABLED === 'true',
```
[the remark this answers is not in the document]
Dschoordsch: Yes, that's what I've written in my comment. I added it anyways because it's more explicit.

Dschoordsch:
Please see also https://github.com/ParabolInc/action-devops/pull/76

PR #5682, opened 2021-11-18
Language tooltip is updated on New Meeting CheckIn Greeting

### LINKED ISSUE #4868: Icebreaker phase: Language tip
Parabol's product has a language tip in each icebreaker round. It displays a greeting in a randomised language. You can hover over the word and it reveals the translation and the language. 

This enhancement covers two issues:

**1. Hover time**
I realised that people have to hover for a _really long time_ to reveal the translation of the ord. I wondered in the past, why is it that a question mark shows up when I hover over the word but nothing happens? It was underlined so I thought it was a hyperlink and clicked it, again nothing. So I figured it was just my browser. 

To enhance this feature I propose that we:

**– Make the hover time shorter** – currently it's around 2.5s – perhaps we make it 0.5-1.0s?

**– Make the word clickable** – If I see an underlined thing, I want to click it b/c it looks like a URL. If users click it bring up the same 'hover over' text.

**2. Additional languages**
It would be great to add some more languages to Parabol on the basis that “If you talk to a man in a language he understands, that goes to his head. If you talk to him in his own language, that goes to his heart". 

I've included a list of potential candidates here: https://docs.google.com/spreadsheets/d/1OEsfcgiOuIvzjD5zMWdMI-364QM3S2x08rf5yHki9rc/edit?usp=sharing

sinansonmez:
Closes #4868

Language tooltip is updated with standard tooltip component. Please see Before and After screen recordings below. 
This change also eliminated long hover time (around 2.5s) mentioned in the issue. 

**BEFORE**

https://user-images.githubusercontent.com/37421564/142458132-7e8fe1bb-20a5-4ae1-bedc-a338a6050f6f.mov

**AFTER**

https://user-images.githubusercontent.com/37421564/142458230-7e6aab15-c00f-4f02-8bcd-aa3e2d6194e6.mov

### DIFF
--- a/packages/client/hooks/useTooltip.ts
+++ b/packages/client/hooks/useTooltip.ts
@@ -14,7 +14,7 @@ const useTooltip = <T extends HTMLElement = HTMLElement>(
   preferredMenuPosition: MenuPosition,
   options: Options = {}
 ) => {
-  const delay = options.delay || Duration.TOOLTIP_DELAY
+  const delay = options.delay ?? Duration.TOOLTIP_DELAY
   const isDisabled = !!options.disabled
   const disabledRef = useRef(isDisabled)
   useEffect(() => {
--- a/packages/client/modules/meeting/components/NewMeetingCheckInGreeting.tsx
+++ b/packages/client/modules/meeting/components/NewMeetingCheckInGreeting.tsx
@@ -27,9 +27,12 @@ const NewMeetingCheckInGreeting = (props: Props) => {
   const {teamMember, checkInGreeting} = props
   const {content, language} = checkInGreeting
   const {preferredName} = teamMember
-  const {tooltipPortal, openTooltip, closeTooltip, originRef} = useTooltip(
-    MenuPosition.UPPER_CENTER
-  )
+  const {
+    tooltipPortal,
+    openTooltip,
+    closeTooltip,
+    originRef
+  } = useTooltip(MenuPosition.UPPER_CENTER, {delay: 0})
   return (
     <GreetingBlock>
       <GreetingSpan ref={originRef} onMouseEnter={openTooltip} onMouseLeave={closeTooltip}>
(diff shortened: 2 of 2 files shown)


Dschoordsch (changes requested):
Thank you for taking this!

This works as advertised, but #4868 also states the greeting should be clickable. While not a big difference I think it's easy enough to achieve. I think you need to modify `useTooltip` though to have another method which opens the tooltip immediately without a timeout.

sinansonmez:
Thanks for the feedback @Dschoordsch

I didnt implement the clicking part because in the old case it was taking 2 seconds to open the tooltip. Due to this behaviour, the user feels the need to click on the word. 

In the new case, since the tooltip   is opened in a fraction of second on hover, I believe there is no need for clicking. 

If you agree, I can check the tooltip to make sure that tooltip is opened immediately on hover without any delay. In that case there may be no need to implement clicking. Please let me know what you think.

Dschoordsch:
Yes, let's try decreasing the delay even further. Not sure if it isn't irritating when it pops up immediately.

sinansonmez:
I updated and reduced the delay time from default 400 ms to 0. Please let me know what you think?

Dschoordsch:
I checked, it doesn't work with `delay: 0` unless you fix the following line
https://github.com/ParabolInc/parabol/blob/9f19d8e66d9d1589ab7d78b91fe5762f3634ca00/packages/client/hooks/useTooltip.ts#L17
to
```
  const delay = options.delay ?? Duration.TOOLTIP_DELAY
```
After that it looks good and doesn't need `onClick`.

sinansonmez:
You are right `0` is treated as falsy value with `||` operator as explained [here](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing) (which is sth I didn't know, thanks for the hint). Therefore, the PR is updated accordingly with `??`.

Dschoordsch (approved):
Thank you! @mattkrick can you do the maintainer review, please?

mattkrick:
LGTM! great work

Issue #5553, opened 2021-10-21
New members don't show up with name in team overview


Dschoordsch:
When a new member is invited and joins, their avatar shows up but without a name, see
![Missing name for new members](https://user-images.githubusercontent.com/7331043/138253001-d48addd3-6e58-4142-91b1-0cbdca8e580a.gif)
Issue is gone after refresh.

### Acceptance criteria:
* newly added members show up with name

**Estimated effort:** 3 hours

RiyaNegi:
Hey, this is still an issue. Mind if I pick this up? 😄

PR #5683, opened 2021-11-18
Fix empty name on tooltip when onboarding new member

RiyaNegi:
Fixed [Issue 5553](https://github.com/ParabolInc/parabol/issues/5553)

Had to change the PropType interface for `DashboardAvatar_teamMember` since the graphQL data structure could not be changed.
@jordanh

### DIFF
--- a/packages/client/components/DashboardAvatars/DashboardAvatar.tsx
+++ b/packages/client/components/DashboardAvatars/DashboardAvatar.tsx
@@ -37,12 +37,12 @@ const StyledAvatar = styled(Avatar)<{isConnected: boolean; picture: string}>(
 
 const DashboardAvatar = (props: Props) => {
   const {teamMember} = props
-  const {id: teamMemberId, picture, teamId} = teamMember
+  const {id: teamMemberId, picture, teamId, preferredName} = teamMember
   const {user} = teamMember
   if (!user) {
     throw new Error(`User Avatar unavailable. ${JSON.stringify(teamMember)}`)
   }
-  const {isConnected, preferredName} = user
+  const {isConnected} = user
   const atmosphere = useAtmosphere()
   const {submitting, onError, onCompleted, submitMutation} = useMutationProps()
   const {tooltipPortal, openTooltip, closeTooltip, originRef} = useTooltip<HTMLDivElement>(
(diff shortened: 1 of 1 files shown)


packages/client/components/DashboardAvatars/DashboardAvatar.tsx
```diff
@@ -37,12 +37,12 @@ const StyledAvatar = styled(Avatar)<{isConnected: boolean; picture: string}>(
...
 const DashboardAvatar = (props: Props) => {
   const {teamMember} = props
-  const {id: teamMemberId, picture, teamId} = teamMember
+  const {id: teamMemberId, picture, teamId, preferredName} = teamMember
   const {user} = teamMember
   if (!user) {
     throw new Error(`User Avatar unavailable. ${JSON.stringify(teamMember)}`)
   }
-  const {isConnected, preferredName} = user
+  const {isConnected} = user
```
Dschoordsch, line 45: Did you check why preferredName on User was empty?

Dschoordsch (approved):
Thank you for picking this up!

I verified and you fixed the issue. I'm curious why the previous solution did not work though.
If you want to look into it, let me know, otherwise I will check.

packages/client/components/DashboardAvatars/DashboardAvatar.tsx
RiyaNegi: I'm not well versed with GraphQL, but I noticed `packages/server/graphql/types/User.ts` does have `preferredName` field mentioned. Very curious as to why it did not return. 🤔

Dschoordsch: No worries. I checked and it's because we don't query this field in the notification for new team members
https://github.com/ParabolInc/parabol/blob/6fcd60a5924577cde641b8b19b9a619ab0b12b79/packages/client/mutations/AcceptTeamInvitationMutation.ts#L31

Dschoordsch (approved):
@mattkrick could you do the maintainer review?

mattkrick:
great bug find!
The implicit dependency between mutation & query is a sharp edge in relay.
This PR fixes the problem at hand, but we could face a similar problem in the future unless we explicitly couple the mutation payload to that query.

To address the root cause, could you augment that mutation payload to fetch everything that query needs?
e.g.

```gql
teamMember {
  ...DashboardAvatar_teamMember
}
```

that way whenever the query changes we won't have to worry about the mutation not fetching the right fields 👍

PR #5685, opened 2021-11-18
Feat/spotlight wiggle

mattkrick:
Needed a break from code reviews.
Now when the spotlight window is open for a remote user, you see their spotlight card move around in a fun pattern.

TEST
- [ ] Browser A opens the spotlight
- [ ] Browser B sees that card move around in constant time in a trinity knot pattern (kinda like a loosely defined circle)

I switched the animation from `ease` to `linear` because how long the spotlight stays open is indeterminate. With `ease` it was slowing down & coming to a halt every 3 seconds which kinda gave the impression that the duration was not indeterminate. https://material.io/components/progress-indicators#linear-progress-indicators

### DIFF
--- a/packages/client/components/ReflectionGroup/RemoteReflection.tsx
+++ b/packages/client/components/ReflectionGroup/RemoteReflection.tsx
@@ -199,7 +199,6 @@ const RemoteReflection = (props: Props) => {
   const {nextStyle, transform, minTop} = getStyle(remoteDrag, isDropping, isSpotlight, style)
 
   const {headerTransform, arrow} = getHeaderTransform(ref, minTop)
-  console.log({transform})
   return (
     <>
       <RemoteReflectionModal
--- a/packages/client/components/ReflectionGroup/RemoteReflection.tsx
+++ b/packages/client/components/ReflectionGroup/RemoteReflection.tsx
@@ -17,20 +17,16 @@ import getBBox from '../RetroReflectPhase/getBBox'
 import UserDraggingHeader, {RemoteReflectionArrow} from '../UserDraggingHeader'
 import useSpotlightResults from '~/hooks/useSpotlightResults'
 
-const circleAnimation = (transform?: string) => keyframes`
-  0%{
-    transform:translate(5px)
-              rotate(0deg)
-              translate(-5px)
-              rotate(0deg)
-              ${transform ?? ''}
-  }
-  100%{
-    transform:translate(5px)
-              rotate(360deg)
-              translate(-5px)
-              rotate(-360deg)
-              ${transform ?? ''}
+// original from https://codepen.io/Maelig/pen/KrjAgq
+// scale(0.75)
+// translate(-39.5px, 1px)
+// transform determined by hand by delta(kanban el, floating el)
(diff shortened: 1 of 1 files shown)


packages/client/components/ReflectionGroup/RemoteReflection.tsx
```diff
@@ -17,20 +17,16 @@ import getBBox from '../RetroReflectPhase/getBBox'
...
-              rotate(360deg)
-              translate(-5px)
-              rotate(-360deg)
-              ${transform ?? ''}
+// original from https://codepen.io/Maelig/pen/KrjAgq
+// scale(0.75)
+// translate(-39.5px, 1px)
+// transform determined by hand by delta(kanban el, floating el)
+// round all decimals, convert to relative
+const triquetaPath = `path('m0 0c0 10-8 18-18 18s-18-8-18-18c3-2 6-2 9-2c10 0 18 8 18 18c0 6-3 12-9 16c-6-3-9-9-9-16c0-10 8-18 18-18c3 0 6 0 9 2z')`
```
nickoferrall, line 25: The animation moves down and to the left of the reflection's original position in the kanban. I think that works well, but it can hide some of the reflection below. 

Perhaps the reflection's original kanban position could be the centre of the animation rather than at the top right, i.e. the animation moves up and to the right initially before moving down and to the left. That way, it hides less off the card below. 

wiggling

packages/client/components/ReflectionGroup/RemoteReflection.tsx
```diff
@@ -49,8 +45,10 @@ const RemoteReflectionModal = styled('div')<{
     isDropping ? Times.REFLECTION_REMOTE_DROP_DURATION : Times.REFLECTION_DROP_DURATION
   }ms ${BezierCurve.DECELERATE}`,
   transform,
-  animation:
-    isSpotlight && !isDropping ? `${circleAnimation(transform)} 3s ease infinite;` : undefined,
+  transformOrigin: '0 0',
+  offsetPath: triquetaPath,
+  offsetRotate: '0deg',
+  animation: isSpotlight && !isDropping ? `${spotlightWiggle} 4s linear infinite;` : undefined,
```
nickoferrall, line 51: Now that we're not using `transform` here, it looks like we can remove the `transform` on line 47, remove `isSpotlight` from `getStyle`, and remove the first if statement in `getStyle` as well as the comment above the `getStyle` function

nickoferrall (changes requested):
Wonderful wiggling 👌

packages/client/components/ReflectionGroup/RemoteReflection.tsx
Dschoordsch: Not using `transform` is the reason the animation is offset, isn't it?

Dschoordsch: Yes, the animation is using a path, but the transform property of the previous animation was added for an offset. I think it was dynamic or passed from the parent component or something similar, that's why I added it there. Sadly I didn't leave a comment.

mattkrick:
low priority, closing for now

PR #5687, opened 2021-11-19
Ironbank security patches for v6.36.0

adaniels-parabol:
Update contains security patches reported by Ironbank.


Dschoordsch:
Wanted to merge master to check CI, but accidentally merged my local master branch :-/ Had to force push to fix it.

packages/server/utils/predicates.ts
```diff
@@ -1 +1,2 @@
-export const isNotNull = <T>(item: T | null): item is T => item !== null
+export const isNotNull = <T>(item: T | null | undefined): item is T =>
```
mattkrick, line 1: we should probably rename this to `nullish` but that can wait for another time. LGTM

PR #5688, opened 2021-11-19
Show all packages errors on `yarn typecheck`

Dschoordsch:
If we run this command we want to see all errors, so don't stop lerna
run when the first package fails.


adaniels-parabol (approved):
I agree, good fix. Should allow developers see all the tests that failed

PR #5689, opened 2021-11-19
Fix missing `isSpotlight` in `useSpotlightReflectionGroup`

Dschoordsch:
The property was missing from the `ReflectionGroup`s fragment. Also
moved the hook in the `ReflectionGroup` subdirectory as it's tightly
coupled to `ReflectionGroup`


nickoferrall (approved):
LGTM!

PR #5697, opened 2021-11-22
Fix no implicit any errors (Part 3 of 8)

### LINKED ISSUE #5620: Add noImplicitAny to server tsconfig (Part 3 of 8)

- [ ] fix ~200 errors
- [ ] we'll merge it blindly
- [ ] create a copy of this issue named Part (3 of 8) so we can re-prioritize for the next person. We'll keep doing this until it's done

AC

Estimate: 4 hours

nickoferrall:
Fix #5620 

Following [this comment](https://github.com/ParabolInc/parabol/issues/5314#issuecomment-923128109), I'll move this PR straight to Maintainer Review for a light CR. 

I've created an issue for part 4 here: https://github.com/ParabolInc/parabol/issues/5696


packages/server/graphql/types/PokerTemplate.ts
```diff
@@ -17,9 +17,9 @@ const PokerTemplate = new GraphQLObjectType<any, GQLContext>({
     dimensions: {
       type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(TemplateDimension))),
       description: 'The dimensions that are part of this template',
-      resolve: async ({id: templateId}, _args, {dataLoader}) => {
+      resolve: async ({id: templateId}, _args: unknown, {dataLoader}) => {
         const dimensions = await dataLoader.get('templateDimensionsByTemplateId').load(templateId)
-        const activeDimensions = dimensions.filter(({removedAt}) => !removedAt)
+        const activeDimensions = dimensions.filter(({removedAt}: {removedAt: Date}) => !removedAt)
```
mattkrick, line 22: this one seems odd, should already be `Date | undefined`. were you getting a noImplicityAny here too?

packages/server/graphql/types/RetrospectiveMeetingSettings.ts
```diff
@@ -69,7 +73,7 @@ const RetrospectiveMeetingSettings = new GraphQLObjectType<any, GQLContext>({
         const {orgId} = team
         const templates = await dataLoader.get('meetingTemplatesByOrgId').load(orgId)
         const organizationTemplates = templates.filter(
-          (template) =>
+          (template: MeetingTemplate) =>
```
mattkrick, line 76: this is another odd one! it should already be a MeetingTemplate without the cast, right?

packages/server/graphql/types/SetPhaseFocusPayload.ts
```diff
@@ -18,9 +19,9 @@ const SetPhaseFocusPayload = new GraphQLObjectType<any, GQLContext>({
     },
     reflectPhase: {
       type: new GraphQLNonNull(ReflectPhase),
-      resolve: async ({meetingId}, _args, {dataLoader}) => {
+      resolve: async ({meetingId}, _args: unknown, {dataLoader}) => {
         const meeting = await dataLoader.get('newMeetings').load(meetingId)
```
mattkrick, line 23: nit: 

```ts
const meeting = (await dataLoader
          .get('newMeetings')
          .load(meetingId)) as MeetingRetrospective
```

will make the error down below go away (not sure why type unions give this error in typescript anyways).

packages/server/graphql/types/Threadable.ts
```diff
@@ -35,21 +35,22 @@ export const threadableFields = () => ({
   createdByUser: {
     type: require('./User').default,
     description: 'The user that created the item',
-    resolve: ({createdBy}, _args, {dataLoader}: GQLContext) => {
+    resolve: ({createdBy}: {createdBy: Date}, _args: unknown, {dataLoader}: GQLContext) => {
```
mattkrick, line 38: this is a string

packages/server/graphql/types/UpdateUserProfilePayload.ts
```diff
@@ -19,9 +19,9 @@ const UpdateUserProfilePayload = new GraphQLObjectType<any, GQLContext>({
     teamMembers: {
       type: new GraphQLList(new GraphQLNonNull(TeamMember)),
       description: 'The updated team member',
-      resolve: ({teamIds, userId}, _args, {dataLoader}) => {
+      resolve: ({teamIds, userId}, _args: unknown, {dataLoader}) => {
```
mattkrick, line 22: nit: always nice to declare the type as early as possible, that way if we do something else with teamIds then we don't have to do multiple casts down below

mattkrick (changes requested):
just 1 bug, the rest are nits. thanks so much for this!

packages/server/graphql/types/PokerTemplate.ts
nickoferrall: I was! It knows the type but still shouts at me 🤷‍♂️  I've changed `Date` to `Date | undefined`

removed-at

packages/server/graphql/types/RetrospectiveMeetingSettings.ts
nickoferrall: Same as above - it knows the type but still complains!

PR #5698, opened 2021-11-22
Remove wrtc signal server

Dschoordsch:


BartoszJarocki (approved):
Looks good! I tested and everything works fine. As we're removing all webrtc stuff we can also remove one more package from server: `@mattkrick/fast-rtc-swarm` as it's not used anymore 🕵️‍♂️

mattkrick:
this is fantastic. painful to see all that work go away, but fantastic nonetheless.

PR #5699, opened 2021-11-22
Spotlight feature flag

### LINKED ISSUE #5691: Put spotlight behind a feature flag
​

implementor's choice: 

- turn it on via cheatcode (e.g. type "spotlight" in the grouping phase & it turns on)

- add a featureflag on the User object & then we can turn on for a per-user/per-domain basis

​

Estimate: 4 hours

nickoferrall:
Fix #5691 


packages/server/graphql/mutations/addFeatureFlag.ts
```diff
@@ -34,14 +35,26 @@ export default {
...
+      flag
+    }: {emails: string[] | null; domain: string | null; flag: UserFeatureFlagEnum},
     {authToken, dataLoader}: GQLContext
   ) {
     const operationId = dataLoader.share()
     const subOptions = {operationId}
 
     // AUTH
-    requireSU(authToken)
+    if (!isSuperUser(authToken)) {
```
nickoferrall, line 49: We want users to grant themselves access to the Spotlight feature by typing in the cheat code. The other flags, poker and jira, aren't currently being used so we don't need to require super user access here. 

As we add more feature flags, I think it's unlikely that we would want to prevent users from adding themselves to a feature. If they're so excited by a feature that they read the codebase and add themself, that's probably fine. If we create a feature that only paying users can access, we can then refactor this logic or create a private schema mutation.

Dschoordsch (changes requested):

packages/client/mutations/AddFeatureFlagMutation.ts
```diff
@@ -0,0 +1,37 @@
+import {commitMutation} from 'react-relay'
+import graphql from 'babel-plugin-relay/macro'
+import {StandardMutation} from '../types/relayMutations'
+import {AddFeatureFlagMutation as TAddFeatureFlagMutation} from '../__generated__/AddFeatureFlagMutation.graphql'
+
+graphql`
+  fragment AddFeatureFlagMutation_notification on AddFeatureFlagPayload {
+    user {
```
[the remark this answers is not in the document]
nickoferrall: It looks like we do need to include the id so that Relay can find the object and update the values ([see here](https://relay.dev/docs/guided-tour/updating-data/graphql-mutations/#updating-data-once-a-request-is-complete)). 

However, [here](https://github.com/ParabolInc/parabol/blob/cf55ed4d0877b1756ffc7213062dde5056ed662b/packages/client/mutations/UpdatePokerScopeMutation.ts#L53) we don't query `id` within `user`, although perhaps that's OK as it's nested within `scores` which contains `userId`.

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -95,10 +95,11 @@ const ReflectionCard = (props: Props) => {
     reactjis,
     reflectionGroupId
   } = reflection
-  const phaseType = meeting ? meeting.localPhase.phaseType : null
-  const isComplete = meeting?.localStage?.isComplete
-  const phases = meeting ? meeting.phases : null
-  const spotlightGroupId = meeting?.spotlightGroup?.id
+  const {localPhase, localStage, spotlightGroup, viewerMeetingMember, phases} = meeting!
```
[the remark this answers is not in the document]
nickoferrall: We shouldn't be able to get to the reflection card if the meeting is null. I've pushed an update changing the `meeting` type to `ReflectionCard_meeting` and I've removed the non-null assertion after `meeting`

packages/client/mutations/AddFeatureFlagMutation.ts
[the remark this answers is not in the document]
Dschoordsch: That's what I thought, thanks for checking.

packages/server/graphql/mutations/addFeatureFlag.ts
Dschoordsch: I agree, I don't see any risk in users enabling beta features for themselves.

Dschoordsch (approved):
Works as advertised

mattkrick (changes requested):

packages/server/graphql/mutations/addFeatureFlag.ts
nickoferrall: I like it! Super users could still add the feature flag for other emails & domains in my previous implementation, but not including `email` when adding the flag for the viewer is much cleaner. Updated!

packages/server/graphql/mutations/addFeatureFlag.ts
```diff
@@ -31,14 +34,27 @@ export default {
...
     const operationId = dataLoader.share()
     const subOptions = {operationId}
 
     // AUTH
-    requireSU(authToken)
+    const viewerId = getUserId(authToken)
+    const viewer = await getUserById(viewerId)
+    const isAddingFlagToViewer = !emails?.length && !domain
+    if (!isAddingFlagToViewer) {
+      requireSU(authToken)
```
mattkrick, line 52: this is going to throw an error instead of returning an object with an error property.
any thrown errors have their message replaced with the string "Server Error" before being sent to the client.
so, for example, if we have a bug in our frontend code that adds a value to the domain prop, the user will just get a cryptic message

packages/server/graphql/mutations/addFeatureFlag.ts
```diff
@@ -31,14 +34,27 @@ export default {
...
+    }: {emails: string[] | null; domain: string | null; flag: UserFeatureFlagEnum},
     {authToken, dataLoader}: GQLContext
   ) {
     const operationId = dataLoader.share()
     const subOptions = {operationId}
 
     // AUTH
-    requireSU(authToken)
+    const viewerId = getUserId(authToken)
+    const viewer = await getUserById(viewerId)
```
mattkrick, line 49: superusers will never need to query the user.
come to think of it, same with regular users. the viewerId comes from a server-signed token, so the viewer is guarnateed to exist unless we hard deleted them & they're still using their old token.
even if that happened, the DB mutation below wouldn't fail, it'd just update 0 records

mattkrick (changes requested):

packages/server/graphql/mutations/addFeatureFlag.ts
nickoferrall: I see! So is the reason that we throw an error from `requireSU` elsewhere in the codebase because we want to keep the error message ambiguous for potential attackers? Whereas it's a bad idea to use it here because I'm conditionally calling `requireSU` based on the mutation's arguments? So if the frontend code accidentally adds a domain prop, we want to send more helpful feedback because it's more likely to be a bug from a dev rather than an attack.

mattkrick: it's really just an old pattern from back when we didn't have an `error` field as part of the response payload. doing it on internal calls (e.g. a stripe webhook calls our graphql server) is fine because it'll still get logged to sentry, but it'll never make it down to the client

PR #5700, opened 2021-11-22
Fix sticky tooltip in agenda item pins

nickoferrall:
Sometimes the tooltip for the agenda item pins can be a little sticky. `onMouseOver` & `onMouseOut` fire more frequently as they execute when the cursor leaves the child element as well as the current element. Matt mentioned the issue and solution here: https://github.com/ParabolInc/parabol/pull/5550#discussion_r739439423

Hey @tianrunhe, would you mind reviewing this one, please?


tianrunhe (approved):
LGTM!

PR #5704, opened 2021-11-23
Prevent ungrouping Spotlight results

### LINKED ISSUE #5703: Spotlight: prevent viewer from ungrouping results
In the Spotlight, the viewer cannot group results into other results as we want them to only focus on grouping the source card. However, it's currently possible for the viewer to ungroup a result group. To be consistent, we should prevent the viewer from ungrouping results too.

Fixing this issue would prevent the viewer from accidentally ungrouping a result and then being unable to re-group the reflections, as it's not possible to group results.

### AC

- [ ] The viewer is unable to ungroup a result reflection group

nickoferrall:
Fix #5703

Hey @igorlesnenko, would you mind reviewing this PR, please?


igorlesnenko (approved):
Looks good to me

PR #5706, opened 2021-11-23
update password requirements for local accounts

adaniels-parabol:
We've had the question a couple times regarding our password complexity in security questionnaires. We are currently only requiring any 6 characters. This PR increases the requirements to: Minimum eight characters, at least one uppercase letter, one lowercase letter, one number and one special character.

### DIFF
--- a/packages/client/components/EmailPasswordAuthForm.tsx
+++ b/packages/client/components/EmailPasswordAuthForm.tsx
@@ -80,8 +80,8 @@ const validateEmail = (email: string) => {
 const validatePassword = (password: string) => {
   return new Legitity(password)
     .required('Please enter a password')
-    .min(6, '6 character minimum')
     .max(1000, `That's a book, not a password`)
+    .matches(new RegExp("^(?=.*\\d)(?=.*[A-Z])(?=.*[a-z])(?=.*[^\\w\\d\\s:])([^\\s]){8,}$"), 'Minimum eight characters, at least one uppercase letter, one lowercase letter, one number and one special character')
 }
 
 const EmailPasswordAuthForm = forwardRef((props: Props, ref: any) => {
--- a/packages/client/components/EmailPasswordAuthForm.tsx
+++ b/packages/client/components/EmailPasswordAuthForm.tsx
@@ -80,8 +80,9 @@ const validateEmail = (email: string) => {
 const validatePassword = (password: string) => {
   return new Legitity(password)
     .required('Please enter a password')
+    .min(8, '8 character minimum')
     .max(1000, `That's a book, not a password`)
-    .matches(new RegExp("^(?=.*\\d)(?=.*[A-Z])(?=.*[a-z])(?=.*[^\\w\\d\\s:])([^\\s]){8,}$"), 'Minimum eight characters, at least one uppercase letter, one lowercase letter, one number and one special character')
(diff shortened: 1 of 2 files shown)


packages/client/components/EmailPasswordAuthForm.tsx
```diff
@@ -80,8 +80,8 @@ const validateEmail = (email: string) => {
 const validatePassword = (password: string) => {
   return new Legitity(password)
     .required('Please enter a password')
-    .min(6, '6 character minimum')
     .max(1000, `That's a book, not a password`)
+    .matches(new RegExp("^(?=.*\\d)(?=.*[A-Z])(?=.*[a-z])(?=.*[^\\w\\d\\s:])([^\\s]){8,}$"), 'Minimum eight characters, at least one uppercase letter, one lowercase letter, one number and one special character')
```
Dschoordsch, line 84: I think it's nicer to split off the minimal length. This simplifies the regex and gives a better user experience. I would keep the rest together though, so the user can see all these requirements at once. The length should also be checked after the special characters: `required(...).matches(...).min(...).max(...)`

I'm also a bit concerned as it creates additional friction for signups. Especially with these more complicated requirements it's much nicer if the user can see them before trying to enter something invalid. Also what is considered special characters is handled differently from site to site.
I think this deserves its own ticket to improve the UI.

adaniels-parabol: I changed this to break out the length using .min set to 8, but made everything else a grouped regex, so you get the error of what else is required all at once, instead of doing multiple messages based on each match
image
image

adaniels-parabol: Additionally, I used any characters no present in the set for the match a-z, A-Z, 0-9 so they have to include something special. I didn't want to require a certain special. What are your thoughts @Dschoordsch ?

Dschoordsch: Good enough I think. I still think putting the `match` first would be better, otherwise you come up with 8 characters only to discover you need to include a number and special character as well.

Dschoordsch (approved):
I would still switch `.matches` and `.min`

mattkrick:
this is gonna add a huuuuuuuge friction & means a bunch of extra error messages for users.
it also means if their "goto password" isn't sufficient they'll have to think up a new one that they'll forget, which means a lot more "forgot password" clicks. Overall, that means free/pro users get a worse experience & our current goal is to prioritize them over enterprise users.

For security questionnaires could we instead say we force SAML for enterprise users? my concern is no enterprise is actually using passwords anyways so we're making it harder to kick the tires and we're not getting enough in return for it.

login attempts for a particular user are limited to 10 per IP address. That means if an attacker KNEW that your password was exactly 6 lowercase characters & didn't have any numbers it would take 26 ^6 / 10 = 30891578 ip addresses. that's a hell of a botnet!

Gun to my head, if we absolutely have to increase password strength, we need to make the UX as friendly as possible:
- show requirements with a ✅  or 🚫  next to each one before the user starts typing
- as the user types, update the ✅  or 🚫

adaniels-parabol:
There is no urgent need to merge this, we can always close this PR as tension is near zero.

Currently, in security questionnaires (when asked) we respond "password requirements for local users (e.g. username/password) require 6 characters. SAML (e.g. SSO) and Google oAuth inherit security password paraments from the authentication provider. If specific password requirements are needed it is recommended to utilize SAML to enforce any password requirements per your organizations security policy."

mattkrick:
i think that security questionnaire response is great! if we lose an enterprise contract because someone doesn't want to use SAML we can re-eval.
closing for now

Issue #5415, opened 2021-09-17
Dropping remote reflection loses position when opening the Spotlight


nickoferrall:
To reproduce:

- Go to the grouping stage in a retro
- Open a second browser and drag and drop a reflection
- On the first browser, click the open spotlight icon before the remote reflection has completely dropped
- Note that it either loses its position or moves into the Spotlight

### AC

- [ ] The dropping remote reflection position doesn't change when opening the Spotlight

nickoferrall:
@BartoszJarocki oh well that's good news! Which branch are you using? I just tested it on my most recent branch, and I'm still seeing the issue. I'll test it on the one you're using

BartoszJarocki:
@nickoferrall you're right, sorry! yesterday wasn't my best day! 🤕

jordanh:
**Story Points: 13**
  [See the discussion](https://action.parabol.co/meet/7By0K1Cj6w/estimate/7) in “Sprint `#89` Scoping”

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

PR #5708, opened 2021-11-23
Fix remotely ungrouped results position in Spotlight

### LINKED ISSUE #5627: Spotlight: remotely ungrouping result group bug

nickoferrall:
Fix #5627 

### To test

- [ ] If Alice remotely ungroups a result group, Bob sees the ungrouped reflection move into the expected position in the Spotlight modal
- [ ] If a remote reflection is dropping and the viewer opens the Spotlight, the dropping remote reflection does not move into the Spotlight (see gif example here https://github.com/ParabolInc/parabol/issues/5415#issuecomment-937912945)

### DIFF
--- a/packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx
+++ b/packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx
@@ -56,6 +56,7 @@ interface Props {
   staticReflections: DraggableReflectionCard_staticReflections | null
   swipeColumn?: SwipeColumn
   dataCy?: string
+  isSpotlightEntering?: boolean
 }
 
 export interface TargetBBox {
@@ -76,7 +77,8 @@ const DraggableReflectionCard = (props: Props) => {
     openSpotlight,
     isDraggable,
     swipeColumn,
-    dataCy
+    dataCy,
+    isSpotlightEntering
   } = props
   const {id: meetingId, teamId, localStage, spotlightGroup, spotlightReflectionId} = meeting
   const {isComplete, phaseType} = localStage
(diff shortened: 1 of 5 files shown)


packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx
```diff
@@ -112,28 +114,17 @@ const DraggableReflectionCard = (props: Props) => {
...
-        }
-        // if the spotlight is closed, this card is the single source of truth
+        // If the spotlight is closed, this card is the single source of truth
         // Else, if it's a remote drag that is not in the spotlight
         // Else, if this is the instance in the source or search results
+        // And Spotlight modal isn't entering. This throws off dropping remote card position
         const isPriorityCard =
-          !isSpotlightOpen || (!isReflectionIdInSpotlight && remoteDrag) || isInSpotlight
+          (!isSpotlightOpen || (!isReflectionIdInSpotlight && remoteDrag) || isInSpotlight) &&
+          !isSpotlightEntering
```
nickoferrall, line 127: I don't love passing an extra prop down through a few components, but `isSpotlightEntering` is descriptive, it means we can remove the `useEffect` logic above, and it fixes the two tricky bugs mentioned in the PR description

nickoferrall:
Hey @BartoszJarocki, in the unlikely event you have time left in your workday, would you mind reviewing this PR? No worries if not - I'll tag someone else in tomorrow

BartoszJarocki:
I'm doing it now!

BartoszJarocki (approved):
@nickoferrall works perfectly! Even though I agree, the props drilling is pretty bad, it's verbose what it's doing so I'd be all for simplifying the logic. Thank finding the solution so quickly, I owe you! 😁

PR #5714, opened 2021-11-24
Spotlight tracking

### LINKED ISSUE #5493: Spotlight tracking
It'd be interesting to track how users are using the Spotlight as it would give us a clearer understanding of how we can improve it.

### AC: 

With Segment, track how often users are doing the following:

- [ ] Opening the Spotlight 
- [ ] Dragging the source to the results
- [ ] Dragging the results to the source
- [ ] Trying to drag a result to another result

nickoferrall:
Fix #5493 

To test, copy the `SEGMENT` .env vars from the devops repo and add them to your `.env`.


packages/client/mutations/SendClientSegmentEventMutation.ts
```diff
@@ -9,7 +9,7 @@ const mutation = graphql`
 `
 
 interface Options {
-  eventId: number
+  eventId?: number
```
nickoferrall, line 12: `eventId` is being used in [useSegmentTrack](https://github.com/ParabolInc/parabol/blob/feadf80b9ab8561b495c5e9a43cddd3e8e07a918/packages/client/hooks/useSegmentTrack.ts) to check whether the hook is firing too frequently. It's not a requirement for Segment so I think it can be optional

packages/client/hooks/useAnimatedSpotlightSource.ts
```diff
@@ -52,6 +54,10 @@ const useAnimatedSpotlightSource = (
     dragIdRef.current = clientTempId()
     // execute mutation after cloning as the mutation will cause reflection height to change
     startDrag(reflectionId, dragIdRef.current)
+    SendClientSegmentEventMutation(atmosphere, 'Opened Spotlight', {
```
nickoferrall, line 57: We could use the `segmentIo` utility function in the `startDraggingReflection` mutation on the server, which is the typical way that we're tracking events in segment. However, as `startDraggingReflection` is firing every second so that the remote animating reflection continues to work even when subscribers refresh the page, I think we should use the client mutation instead as it will fire only once.

nickoferrall: When a Spotlight event happens, we're sending the following to Segment:
- `viewerId` so we know which users are using the Spotlight
- `reflectionId` so we know if users open the Spotlight with different reflections, if multiple team members open the same reflection, and if the Spotlight is usually opened with the same type of reflection
- `meetingId` so we know in which types of meetings users open the Spotlight. I'd imagine it happens in meetings where there are many reflections, but we'll see

Here, we're not sending `meetingId`. We could pass `meetingId` into the hook or move the mutation to `GroupingKanban` where we close the portal, but as we're tracking the `reflectionId`, we could always track the meeting by looking at the `Closed Spotlight` event, which does send the `meetingId`.

packages/client/hooks/useDraggableReflectionCard.tsx
```diff
@@ -229,6 +228,16 @@ const useDragAndDrop = (
         ? 'REFLECTION_GRID'
         : null
     handleDrop(atmosphere, reflectionId, drag, targetType, targetGroupId)
+    if (spotlightGroup?.id) {
+      const event = isReflectionInSpotlightResults
+        ? `Spotlight result to ${targetType === 'REFLECTION_GROUP' ? 'source' : 'result'}`
+        : `Spotlight source to ${targetType === 'REFLECTION_GROUP' ? 'result' : 'grid'}`
```
nickoferrall, line 234: The source can be dropped on a result or the grid. It's dropped on the grid when the source is dragged but not dropped on a result, or when there are 2+ reflections in the source group and they're ungrouped. 

Results can only be dropped on the source. If the viewer tries to drop them on another result, it will return to its original position. It's worth tracking this as we might want to rethink preventing results from being grouped onto results if users are frequently doing this.

nickoferrall:
Hey @igorlesnenko, could you review this PR, please? https://github.com/ParabolInc/parabol/pull/5704 is its predecessor so it's best to review that one first

igorlesnenko:
@nickoferrall when I drag Spotlight source to result, I see two spotlight closed events in segment logs. Is it expected behaviour?

nickoferrall:
Ah nice catch! This was happening because `onCloseSpotlight` was fired when the source was dropped on a result, but the `onClose` function of the modal also executes `closeSpotlight`. I'm now passing `closePortal` to `SpotlightModal` so it only fires once.

igorlesnenko (approved):
Looks good to me

packages/client/mutations/SendClientSegmentEventMutation.ts
mattkrick: oh yeah! i'll go ahead & remove that, that issue was resolved

packages/client/hooks/useSpotlightSimulatedDrag.tsx
```diff
@@ -48,6 +50,11 @@ const useSpotlightSimulatedDrag = (
       dropTargetId: null,
       dragId: dragIdRef.current
     })
+    SendClientSegmentEventMutation(atmosphere, 'Closed Spotlight', {
```
mattkrick, line 53: curious, what can we learn from watching the close events? if we're tracking how long the spotlight is open, should we attach an eventId so we can match the open to the close?

packages/client/hooks/useAnimatedSpotlightSource.ts
mattkrick: why do this on the client vs server?

for example, if we did this on the server (inside the `similarReflections` query) we could capture this info for folks who have adblockers, too.

might also be worthwhile capturing how many reflections there are in the meeting

mattkrick (commented):
just a couple questions, if no changes are desired feel free to merge

packages/client/hooks/useSpotlightSimulatedDrag.tsx
nickoferrall: Exactly, my thinking was that we could track how long the Spotlight has been open for. I thought that we could identify how long it has been opened for by using the `viewerId` and `reflectionId`, but adding an `eventId` might make life easier. I've pushed an update using the `dragId` for the `eventId`.

packages/client/hooks/useAnimatedSpotlightSource.ts
nickoferrall: I initially implemented this on the server, but as we're firing `startDraggingReflectionMutation` every second so that the remote Spotlight animation keeps wiggling after a refresh, I figured it would be best to do this once from the client. 

I've added the `reflectionsCount` when closing the Spotlight.

packages/client/hooks/useSpotlightSimulatedDrag.tsx
mattkrick: -1 those event IDs are going to get reset on refresh
either we need a globally unique event id or save when `open spotlight` gets called & then when this gets called you can determine the open duration. i'd probably opt for the latter to make it easy for marketing folks to see the duration

nickoferrall: > save when open spotlight gets called

Ah did you see the 'Opened Spotlight' event [here](https://github.com/ParabolInc/parabol/blob/d34564218947580b3430066469e822dc710a23af/packages/client/hooks/useAnimatedSpotlightSource.ts#L57)?

My original thinking was the marketing folks could use the `viewerId`, `reflectionId`, and `createdAt` (provided by Segment) fields in the 'Opened Spotlight' event and compare it with the subsequent 'Closed Spotlight' event with the same `viewerId` & `reflectionId`. I'm not sure the `eventId` is required so I've removed that in the latest update.

Also, previously if the user opened the Spotlight and refreshed the page, there would be an 'Opened Spotlight' event but not a 'Closed Spotlight' event. That could muddy the data so I've added a `window.onbeforeunload` function that closes the portal causing the 'Closed Spotlight' event to happen.

nickoferrall:
Hey @Dschoordsch, Matt mentioned that the PRs that were in Maintainer Review could be merged if a second person reviews it. Would you mind reviewing this one please, and merging if you're happy with it?

packages/client/components/GroupingKanban.tsx
```diff
@@ -174,6 +177,7 @@ export default createFragmentContainer(GroupingKanban, {
           id
         }
       }
+      spotlightSearchQuery
```
Dschoordsch, line 180: +1 this seems like an unrelated fix

packages/client/components/ReflectionGroup/DraggableReflectionCard.tsx
```diff
@@ -190,6 +190,7 @@ export default createFragmentContainer(DraggableReflectionCard, {
         id
       }
       spotlightReflectionId
+      spotlightSearchQuery
```
Dschoordsch, line 193: +1 Looks like an unrelated fix?

Dschoordsch (approved):

Dschoordsch:
Tested also that the events are passed to `segmentIo.track(...)`.

packages/client/components/GroupingKanban.tsx
nickoferrall: We need `spotlightSearchQuery` here and in `DraggableReflectionCard` so that we have access to the query string in `useDraggableReflectionCard` and the search bar. Thanks for the review - now that we've had 2 reviewers, I'll merge the PR

Issue #5545, opened 2021-10-19
Spotlight UI: matching keyword should be highlighted


tianrunhe:
## Issue - Enhancement

Predecessor: #5469 

In search query case, the returned matched reflection groups should highlight the matching keyword, like:

![Screen Shot 2021-10-19 at 3 07 30 PM](https://user-images.githubusercontent.com/1879975/137997484-ed7c432a-8d5f-4da2-97dc-c116d856cb9e.png)

See [Figma](https://www.figma.com/file/u5bdWUn9hxkdvS2YOYDnXU/Spotlight?node-id=56%3A0) for more detail

### Acceptance Criteria (optional)

- [ ] When a user searches a keyword in Spotlight, the returned reflection cards should highlight the matching keyword.
- [ ] If the matched reflection card is deep inside a reflection group, that card should pop up on the top at the stack. 
	> If the user searches "kitten" and there's a group with two reflections, one with "cat" at the top of the group stack and "kitten" underneath, the "cat" group shows up even though we can't see the word "kitten".

nickoferrall:
We want to highlight a reflection that matches the query even if it’s not at the top of the stack. An alternate approach to popping-up the reflection to the top of the stack (see the 2nd checkbox in the AC) could be changing the opacity of the reflections that are on top of the reflection that matches the search query. That way, the user is still aware that the reflection is nested within the group, but they can also see that it matches the query.

igorlesnenko:
For now implemented the approach with card popping-up to the top of the stack. See video here: #5716

I also tried to I tried to experiment with changing opacity:

**Opacity: 0.2**

![image](https://user-images.githubusercontent.com/466991/142882603-4ec82b7f-6039-453c-bc5f-ddd4be6cc000.png)


**Opacity: 0.09**
![image](https://user-images.githubusercontent.com/466991/142882464-8c3822c5-90f1-46d0-ad89-8d8d51f2e34c.png)


**Opacity: 0.05**

![image](https://user-images.githubusercontent.com/466991/142882152-cc8d8521-85ed-41c3-8639-0babe0421d6d.png)

PR #5716, opened 2021-11-26
Spotlight — highlight search keyword

igorlesnenko:
#5545

https://user-images.githubusercontent.com/466991/143566195-bcefd471-f8ab-4490-b90f-1844ea13f4b4.mov

**How to test**

- Open spotlight
- Type anything in a search field
- See matched keyword highlighted
- If the matched reflection card is deep inside a reflection group, that card should pop up on the top at the stack


igorlesnenko:
@tianrunhe can you review please?

packages/client/components/TaskEditor/decorators.ts
```diff
@@ -12,11 +13,34 @@ const findEntity = (entityType) => (contentBlock, callback, contentState) => {
...
+    return
+  }
+
+  const text = contentBlock.getText()
+  const textLower = text.toLowerCase()
+  const searchQueryLower = searchQuery.toLowerCase()
+
+  let start = 0
+  let found = true
+  while (found) {
```
nickoferrall, line 27: This is probably just personal preference, but I find `find` methods to be more readable than `while` loops. Would something like this work?
```
textLower.split('').findIndex((letter) => searchQueryLower.includes(letter))
```

packages/client/hooks/useEditorState.ts
```diff
@@ -3,10 +3,11 @@ import {convertFromRaw, EditorState} from 'draft-js'
 import {useEffect, useRef, useState} from 'react'
 import makeEditorState from '../utils/draftjs/makeEditorState'
 import mergeServerContent from '../utils/mergeServerContent'
+import remountDecorators from '../utils/draftjs/remountDecorators'
 
-const useEditorState = (content?: string | null | undefined) => {
+const useEditorState = (content?: string | null | undefined, searchQuery?: string | null) => {
```
nickoferrall, line 8: Ideally, I think `useEditorState` shouldn't need to know about the `spotlightSearchQuery`. I'm not too familiar with DraftJS, and I've always found it to be a pain when dealing with it, but I think we can set state with the decorator. See here: https://draftjs.org/docs/advanced-topics-decorators/#setting-new-decorators

packages/client/components/ReflectionGroup/ReflectionGroup.tsx
```diff
@@ -135,6 +136,8 @@ const ReflectionGroup = (props: Props) => {
             dragUserId
             isSpotlight
           }
+          content
```
nickoferrall, line 139: Is `content` being used in this file?

packages/client/components/TaskEditor/decorators.ts
igorlesnenko: @nickoferrall Here, I need to find all occurrences `startindex` and `endIndex` without overlapping and call a callback function for each found.

Not sure I got the idea with `split` and `findIndex`. 
With `split` I see kind of following solution:

```js
textLower.split(searchQueryLower).map(function (token) {
    return this.pos += token.length + searchQueryLower.length
}, { pos: -searchQueryLower.length }).slice(0, -1).forEach(i => { 
  callback(i, i + searchQueryLower.length)
})
```

What do you think?

packages/client/components/ReflectionGroup/useSpotlightVisibleReflections.ts
```diff
@@ -0,0 +1,28 @@
...
+
+    if (spotlightSearchQuery != null && visibleReflections.length > 1) {
+      const spotlightSearchQueryLower = spotlightSearchQuery.toLowerCase()
+      const matchIndex = visibleReflections.findIndex((reflection) => {
+        const textLower = reflection.plaintextContent.toLowerCase()
+        return textLower.indexOf(spotlightSearchQueryLower) !== -1
+      })
+
+      if (matchIndex > 0) {
+        visibleReflections.unshift(visibleReflections.splice(matchIndex, 1)[0])
```
nickoferrall, line 20: NIT: I'd find it easier to read if the splice was moved out of the unshift function. Something like: 
```
const matchingReflection = visibleReflections.splice(matchIndex, 1)[0]
visibleReflections.unshift(matchingReflection)
```

packages/client/components/TaskEditor/decorators.ts
nickoferrall: Ahh OK, thanks for clarifying! If we can avoid making `findText` a thunk, I would be very happy as I find them quite tricky. That might just be me so feel free to ignore this suggestion and stick with your original version, but could this work?

```
  if (searchQuery) {
    const findMatchingText = (
      contentBlock: ContentBlock,
      callback: (startIdx: number, endIdx: number) => void
    ) => {
      const textLower = contentBlock.getText().toLowerCase()
      const searchQueryLower = searchQuery.toLowerCase()
      const matchingIdx = textLower.indexOf(searchQueryLower)
      if (matchingIdx !== -1) {
        callback(matchingIdx, matchingIdx + searchQueryLower.length)
      }
    }

    compositeDecorator.push({
      strategy: findMatchingText,
      component: SearchHighlight
    })
  }
```

nickoferrall (changes requested):

packages/client/components/TaskEditor/decorators.ts
igorlesnenko: @nickoferrall yeah, this removes thunk, but iteration is still needed. Made it as you suggested.

nickoferrall: @igorlesnenko why do we need the while loop as well as `indexOf`? `indexOf` is already iterating through the array?

nickoferrall: Ahh OK cool. Thanks, that makes sense. LGTM!

mattkrick:
🤯  wow! great PR @igorlesnenko and great review @nickoferrall
this looks amazing!

PR #5718, opened 2021-11-27
Updated VSCode settings 

BartoszJarocki:


.vscode/settings.json
```diff
@@ -38,6 +27,8 @@
   "typescript.preferences.useAliasesForRenames": false,
   "typescript.format.insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": false,
   "javascript.format.insertSpaceAfterOpeningAndBeforeClosingNonemptyBraces": false,
+  "editor.tabSize": 2,
+  "editor.defaultFormatter": "dbaeumer.vscode-eslint",
```
Dschoordsch, line 31: This conflicts with the pre-commit hook causing the files to be formatted differently, pure nightmare.

BartoszJarocki: Good find. Do you have any ideas about what could be the cause? We use prettier through eslint plugin so using a separate prettier extension to format the code is kind of a hack. I see pre-commit hook also simply uses eslint. I wonder why it would format files differently...

Dschoordsch: I think the issue was, that the configuration files were not read correctly. It might have been mainly about the ignored files.

BartoszJarocki: Does it work correctly now? I want to make sure that if we move forward with these changes they won't break anything in our workflow.

Dschoordsch: Seems to work. It does not format on save though. I don't like this combination, because formatting errors are indistinguishable from real errors. Is there a setting to improve this?

BartoszJarocki: It should work just fine. If it doesn't it means there's something in your VSCode config that prevents eslint from running on save. Regarding liking or not, this PR was supposed to help with providing the same workspace settings for developers and keep things consistent between what happens in VSCode and what happens when running commit hooks. I guess it's a pointless effort though. I did some reading and having eslint working correctly in VSCode seems to be a pretty random process. I'll keep these changes local as they work better for me and let everybody fight his own battle with the settings. Thank you for trying it, though!

Dschoordsch: I didn't want to discourage you from proceeding with this PR. You removed `"autoFix": true` so I assumed this was intentional and thought you might have a better alternative in mind?

BartoszJarocki: No worries, you haven't discouraged me. I checked on the fresh machine and VSCode and it didn't work as well. Once I find out how to make that work, I'll submit another one. Technically, `"autoFix": true" should not be needed as it's still using `"editor.formatOnSave": true" which should run prettier via eslint. 

If you want to try more things, you can see if adding 
```
  "editor.codeActionsOnSave": {
    "source.fixAll.eslint": true,
  },
```

does make any difference, thank you!

PR #5729, opened 2021-12-01
Returning refreshed token from fresh atlassian auth loader

### LINKED ISSUE #5690: Atlassian Integration - Error: Unauthorized
Affected 81 users in a couple of days, happened more than 1.5k times
We should double check if our token refresh mechanism works correctly

https://sentry.io/organizations/parabol/issues/2563520365/?referrer=github_plugin

```
Error: Unauthorized
  File "./packages/client/utils/AtlassianManager.ts", line 314, in p.get
    return new Error(json.message)
  ?, in runMicrotasks
  File "./packages/client/utils/AtlassianManager.ts", line 548, in async p.getIssue
    const issueRes = await this.get
```

BartoszJarocki:
Fixes #5690

### DIFF
--- a/packages/server/dataloader/atlassianLoaders.ts
+++ b/packages/server/dataloader/atlassianLoaders.ts
@@ -63,6 +63,12 @@ export const freshAtlassianAuth = (parent: RootDataLoader) => {
                 refreshToken: updatedRefreshToken
               }))
             await upsertAtlassianAuths(updatedSameJiraAccountAtlassianAuths)
+
+            return {
+              ...atlassianAuthToRefresh,
+              accessToken,
+              refreshToken: updatedRefreshToken
+            }
           }
 
           return atlassianAuthToRefresh


packages/server/dataloader/atlassianLoaders.ts
```diff
@@ -63,6 +63,12 @@ export const freshAtlassianAuth = (parent: RootDataLoader) => {
                 refreshToken: updatedRefreshToken
               }))
             await upsertAtlassianAuths(updatedSameJiraAccountAtlassianAuths)
+
+            return {
+              ...atlassianAuthToRefresh,
+              accessToken,
```
tianrunhe, line 69: nit: rename this to `updatedAccessToken` to make it clear it's the new one, and also matches the naming pattern with `updatedRefreshToken`

tianrunhe (approved):

Issue #5384, opened 2021-09-08
Audit PG User Table mutations


mattkrick:
See https://github.com/ParabolInc/parabol/blob/7a0c2031dc2ce8b82ce31aeaa6e5c8d4ab2e7fe4/packages/server/billing/helpers/adjustUserCount.ts#L58-L65

Those 2 mutations aren't the same! 
We should check em all before we move to pg completely for the User table

AC
- [ ] audit all user mutations & maybe change them to be more explicit when needed

Estimate: 16 hours

Dschoordsch:
I reviewed
* addFeatureFlag
* addNewFeature
* addTeamIdToTMS
* adjustUserCount
* bootstrapNewUser
* connectSocket
* deleteUser
* dismissNewFeature
* emailPasswordReset
* flagOverLimit
* helper/updateUserProfile
* loginSAML
* loginWithGoogle
* payLater
* removeTeamMember
* resetPassword
* safeArchiveTeam
* signUpWithPassword
* updateEmail
* updateWatchlist
* verifyEmail

Most of these don't write to Redis anymore since #5723. Therfore I compared the current PG writes with the pre-PR Redis writes.

Findings:
* dismissNewFeature does not update the updatedAt field anymore
* addFeatureFlag, payLater and removeTeamMember never updated the updatedAt field

Dschoordsch:
@mattkrick I don't see the difference between PG and Rethink in the example in the ticket description, both update only 2 fields
https://github.com/ParabolInc/parabol/blob/7a0c2031dc2ce8b82ce31aeaa6e5c8d4ab2e7fe4/packages/server/billing/helpers/adjustUserCount.ts#L53-L56

I propose adding `updatedAt` to the 4 mutations mentioned above and then consider this resolved.

mattkrick:
**Story Points: 21**
  [See the discussion](https://action.parabol.co/meet/8Kr615RAwE/estimate/10) in Sprint Poker #​​9

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

PR #5723, opened 2021-11-29
Hotfix - Can create new teams

mattkrick:
Fix #5721

Caching the User objects in redis is hurting us again, so this PR rips out all User object caching.
Instead of using a permanent cache, this PR uses a dataloader if possible, or goes out to PG if a dataloader isn't easy to access.

This also fixes that pesky husky bug that just started appearing on my machine this morning.

TEST

- [ ] can create a new team & refresh the page & still be on that team page
- [ ] can archive a team
- [ ] can create a new user

PR #5719, opened 2021-11-29
Add calculated domain field to User

Dschoordsch:
## Mutations to test:
* [x] `dailyPulse`
* [x] `addFeatureFlag`*
* [x] `updateWatchlist`*

*tests added

Resolves: #5502

PR #5733, opened 2021-12-01
Add updatedAt trigger to User table

Dschoordsch:
Relates-To: #5723 
Resolves: #5384


packages/server/postgres/migrations/1638368905980_userUpdatedAtTrigger.ts
```diff
@@ -0,0 +1,23 @@
+import {ColumnDefinitions, MigrationBuilder} from 'node-pg-migrate'
+
+export const shorthands: ColumnDefinitions | undefined = undefined
+
+export async function up(pgm: MigrationBuilder): Promise<void> {
+  pgm.sql(`
```
tianrunhe, line 6: We should use:

```javascript
  const client = new Client(getPgConfig())
  await client.connect()
  await client.query(`
    ...
  `)
  await client.end()
```

tianrunhe (changes requested):
Where do those generated changes come from? You don't seem to edit any SQL files?

Dschoordsch:
The `domain` changes were artefacts from #5719

tianrunhe (approved):

PR #5735, opened 2021-12-01
Add spotlightSearchQuery to similarReflectionGroups queries

### LINKED ISSUE #5734: Spotlight: add spotlightSearchQuery to similarReflectionGroups queries
Now that the Spotlight search query is in master, we need to add `spotlightSearchQuery` to some of the `similarReflectionGroup` queries that currently have a `TODO`.

nickoferrall:
Fix #5734


nickoferrall:
Hey @tianrunhe, could you review this one, please?

igorlesnenko (approved):
Look good to me

tianrunhe (approved):

packages/client/hooks/useSpotlightResults.ts
```diff
@@ -8,8 +8,8 @@ import graphql from 'babel-plugin-relay/macro'
 type ResultGroups = useSpotlightResultsLocalQueryResponse['viewer']['similarReflectionGroups']
 
 const useSpotlightResults = (
-  spotlightGroupId: string | undefined,
-  searchQuery: string | undefined,
+  spotlightGroupId?: string,
+  searchQuery?: string | null,
```
mattkrick, line 12: Adding `searchQuery` in this PR meant that all the parent components had to add that to their fragments.
Since `spotlightGroupId, searchQuery` both come from the relay store we could replace them with a `meetingRef` which means the parent components don't have to fetch data on behalf of this hook.

for example

```ts
const useSpotlightResults = (meetingRef) => {
  const meeting = readInlineData<useSpotlightResults_meeting$key>(
    graphql`
      fragment useSpotlightResults_meeting on RetrospectiveMeeting @inline {
        spotlightGroup {
          id
        }
        spotlightSearchQuery
      }
    `,
    meetingRef
  )
  const {spotlightGroup, spotlightSearchQuery = ''} = meeting
  const spotlightGroupId = spotlightGroup?.id ?? ''
```

PS I think we could also remove the `getReflections` param since this query is `store-only`, but maybe I'm missing something?

nickoferrall: Ah nice, I like that idea! Update pushed 👍

PR #5737, opened 2021-12-01
noUncheckedIndexedAccess fixes

### LINKED ISSUE #5710: Add noUncheckedIndexedAccess to tsconfig.base (Part 2 of 3)
see #5313 for context

AC
- [ ] add `noUncheckedIndexedAccess` to tsconfig.base
- [ ] fix ~100 typescript errors
- [ ] remove `noUncheckedIndexedAcces` from tsconfig.base
- [ ] submit PR

Estimate: 4 hours

nickoferrall:
Fix #5710


packages/client/components/EmailPasswordAuthForm.tsx
```diff
@@ -117,7 +117,9 @@ const EmailPasswordAuthForm = forwardRef((props: Props, ref: any) => {
 
   const handleBlur = async (e: React.FocusEvent<HTMLInputElement>) => {
     const {name} = e.target
-    setDirtyField(name)
+    if (name === 'email' || name === 'password') {
```
nickoferrall, line 120: I added extra type safety to `useForm` so that `setDirtyField` and `setValue` must use values that were passed into `useForm`. One downside is  `handleBlur` is less reusable. We could use `name!` but that's error-prone, so I think checking whether it's a valid name is the best approach.

nickoferrall:
I think the `noUncheckedIndexedAccess` will be a big help, but one area where it's a little frustrating is when:
```
if (arr.length > 0){
    arr[0] // TypeScript says it could be undefined
}
```

The reason is we could theoretically do:
```
let a = []
a[1000] = 1 
a.length // 1001
```
However, it seems unlikely that the latter would become an issue whereas using non-null assertion operators unnecessarily will make the codebase more confusing and error-prone. 

To find some middle ground, I thought about adding a util function but discovered the following:
```
  if (arr.length && arr[0] !== undefined) {
    arr[0] // OK
  }

  const hasLength = (arr: unknown[]) => {
    return arr.length > 0 && arr[0] !== undefined
  }
  if (hasLength(arr)) {
    arr[0] // number | undefined
  }
```

It looks like the new and improved [control flow analysis](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-4.html#control-flow-analysis-of-aliased-conditions-and-discriminants) in TS 4.4 might not be working with `noUncheckedIndexedAccess`. 
I haven't seen an issue in their repo relating to this so I'll create one if you also think this is a bug.

nickoferrall:
Hey @Dschoordsch, could you give this one a review, please?

packages/client/modules/userDashboard/containers/UserColumns/UserColumnsContainer.tsx
```diff
@@ -85,7 +85,7 @@ const UserColumnsContainer = (props: Props) => {
       <TaskColumns
         area={areaForTaskCard}
         tasks={filteredTasks}
-        myTeamMemberId={toTeamMemberId(filteredTeams[0].id, userIds ? userIds[0] : viewer.id)}
+        myTeamMemberId={toTeamMemberId(filteredTeams[0]!.id, userIds ? userIds[0]! : viewer.id)}
```
Dschoordsch, line 88: Would be nicer if you would replace the `if(filteredTeams.length)` with
```
const firstTeam = filteredTeams[0]
if (firstTeam) {
```
and then in this line with
```
        myTeamMemberId={toTeamMemberId(firstTeam.id, userIds?.[0] ?? viewer.id)}
```

packages/server/graphql/executeGraphQL.ts
```diff
@@ -48,7 +48,7 @@ const executeGraphQL = async (req: GQLRequest) => {
   } = req
   // never re-use a dataloader since the things it cached may be old
   const dataLoader = getDataLoader(dataLoaderId)
-  dataLoader.share()
+  dataLoader?.share()
```
Dschoordsch, line 51: You should fix the type here instead:
https://github.com/ParabolInc/parabol/blob/3e4e15427df9b9b03d3de2a79ab0bd70e4663281/packages/server/graphql/DataLoaderCache.ts#L46-L49

packages/server/graphql/executeGraphQL.ts
```diff
@@ -81,7 +81,7 @@ const executeGraphQL = async (req: GQLRequest) => {
     console.log((firstError as Error).stack)
     console.trace({error: JSON.stringify(response)})
   }
-  dataLoader.dispose()
+  dataLoader?.dispose()
```
Dschoordsch, line 84: s.a.

packages/server/graphql/mutations/helpers/sendMeetingJoinToSegment.ts
```diff
@@ -8,7 +8,7 @@ const sendMeetingJoinToSegment = async (userId: string, newMeeting: Meeting) =>
     event: 'Meeting Joined',
     properties: {
       facilitatorUserId,
-      hasIcebreaker: phases[0].phaseType === 'checkin',
+      hasIcebreaker: phases[0]!.phaseType === 'checkin',
```
Dschoordsch, line 11: Why not stick with `?` here?

packages/server/graphql/mutations/helpers/sendMeetingStartToSegment.ts
```diff
@@ -8,7 +8,7 @@ const sendMeetingStartToSegment = async (newMeeting: Meeting, template?: Meeting
     userId: facilitatorUserId,
     event: 'Meeting Started',
     properties: {
-      hasIcebreaker: phases[0].phaseType === 'checkin',
+      hasIcebreaker: phases[0]!.phaseType === 'checkin',
```
Dschoordsch, line 11: Why not stick with `?` here?

packages/server/graphql/mutations/helpers/sendPokerMeetingRevoteToSegment.ts
```diff
@@ -14,7 +14,7 @@ const sendPokerMeetingRevoteToSegment = async (
       userId,
       event: 'Poker Meeting Team Revoted',
       properties: {
-        hasIcebreaker: phases[0].phaseType === 'checkin',
+        hasIcebreaker: phases[0]!.phaseType === 'checkin',
```
Dschoordsch, line 17: Why not stick with `?` here?

Dschoordsch (changes requested):

packages/client/hooks/useMakeStageSummaries.ts
```diff
@@ -50,12 +50,12 @@ const useMakeStageSummaries = (phaseRef: any, localStageId: string) => {
...
     const summaries = [] as StageSummary[]
     for (let i = 0; i < stages.length; i++) {
-      const stage = stages[i]
+      const stage = stages[i]!
       const {taskId, task} = stage
       const batch = [stage]
       for (let j = i + 1; j < stages.length; j++) {
         const nextStage = stages[j]
-        if (nextStage.taskId !== taskId) break
+        if (nextStage?.taskId !== taskId) break
```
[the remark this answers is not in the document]
nickoferrall: If we add an `!` here, the line below will complain as `nextStage` could be undefined

packages/client/components/EditorLinkChanger/EditorLinkChanger.tsx
```diff
@@ -125,16 +125,16 @@ const EditorLinkChanger = (props: Props) => {
           <TextBlock>
             <InputLabel>{'Text'}</InputLabel>
             <InputBlock>
-              <BasicInput {...fields.text} onChange={onChange} autoFocus name='text' />
+              <BasicInput {...fields.text!} onChange={onChange} autoFocus name='text' />
```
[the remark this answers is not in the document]
nickoferrall: Nice! I was hesitant to refactor as I figure we should probably use `useForm` instead of `withForm`, but this is a strong improvement from the non-nullable operators I was using previously

packages/server/graphql/mutations/helpers/sendMeetingStartToSegment.ts
Dschoordsch: Why not `?`?

Dschoordsch (approved):
Just one tiny comment, you probably just missed, can move to maintainer review afterwards.

packages/server/graphql/mutations/helpers/sendMeetingStartToSegment.ts
nickoferrall: Thanks, @Dschoordsch, I missed that one. Update pushed! I'll move it to Maintainer Review

packages/client/hooks/useForm.tsx
```diff
@@ -44,7 +46,7 @@ type FormAction = SetError | SetDirty | SetValue
 const reducer = (state: FieldState<any>, action: FormAction) => {
   switch (action.type) {
     case 'setDirty':
-      if (state[action.name].dirty) return state
+      if (state[action.name]?.dirty) return state
```
mattkrick, line 49: +2 great work fixing setDirty & friends! 


-1 you can remove all of the conditional access things in this file if you do the following:

```ts
interface SetError<T extends string> {
  type: 'setError'
  name: T
  error: string | undefined
}

interface SetDirty<T extends string> {
  type: 'setDirty'
  name: T
}

interface SetValue<T extends string> {
  type: 'setValue'
  name: T
  value: string
}

type FormAction<T extends string> = SetError<T> | SetDirty<T> | SetValue<T>

...
  const [state, dispatch] = useReducer<Reducer<FieldState<T>, FormAction<FieldStateKey<T>>>>(
```

packages/client/utils/relay/fromTeamMemberId.ts
```diff
@@ -1,6 +1,6 @@
 const fromTeamMemberId = (teamMemberId: string) => {
   const [userId, teamId] = teamMemberId.split('::')
-  return {userId, teamId}
+  return {userId, teamId} as {userId: string; teamId: string}
```
mattkrick, line 3: +2 nice work avoiding the scope creep of refactoring to `TeamMemberId.split`!

mattkrick (changes requested):
-1 Just 1 little change to make, otherwise good to merge!

packages/client/hooks/useForm.tsx
mattkrick: ping on this issue @nickoferrall

nickoferrall: Thanks, @mattkrick. I was away on Thursday & Friday, so just getting to this now. Update pushed, and I'll now merge!

PR #5743, opened 2021-12-02
Customise release snackbar

### LINKED ISSUE #5742: Allow for specific URL in new release snack bar
## Issue - Enhancement

For most releases, the changelog is the best spot to see what's new. 

For larger updates, we write a[ blog post announcing the feature](https://www.parabol.co/blog/tag/new-features), which includes more detail on what's changed. In those cases, it would be nice to be able to include a link to the blog post in the 'new release' snack bar notification, so users get the best info about what's changed and why it matters. 

Per @adaniels-parabol, to do this, we would need to update this line of code to be changed to a ENV var that would get updated each release: 
https://github.com/ParabolInc/parabol/blob/master/packages/client/hooks/useServiceWorkerUpdater.ts#L25

When new releases are shipped, users can see a link to an announcement blog post for new features, rather than a changelog link. 

nickoferrall:
Fix #5742 

I don't love adding a bunch of env variables, but I think it's the right move because:

1. If we don't include the `launchVersion` env var, we could forget to update it and accidentally launch twice
2. If we decide to change the message or misspell something, we can update the env var without needing to hotfix the app

@adaniels-parabol would you mind reviewing this one, please?


nickoferrall:
Hey @adaniels-parabol, just a friendly reminder. Are you happy to review this one?

adaniels-parabol (approved):
Looks good too me!

mattkrick (changes requested):
this will only work in dev, you'd have to update the vars in createSSR, too.

Overall, I think we need to change the strategy here a bit. Here's why: 
- we'll have to update the env vars manually on staging & production on every release. if we do set custom messaging, we'll have to be sure to _unset_ those vars before the next release
- this message would pop up every time we re-deployed the app. That means if we change nginx, or restart rethinkdb, everyone see the popup again. We should be able to redeploy without negatively affecting the UX. We need a way to flag the message as having already been seen.
- We have this feature already built (see NewFeatureBroadcast) but it doesn't appear as a snackbar. To make sure our code stays maintainable, we should probably refactor the UI of that feature to appear in a snackbar instead

Next steps: 
- let's get the business requirements from growth & see what fields we need
- compare those business reqs to the NewFeatureBroadcast that we already have built in the app
- Decide to refactor the broadcast to a snackbar, or move towards another solution (e.g. a raw HTML like what the sales team does for the modal)

nickoferrall:
That sounds good to me! That way the Growth team can also use the private mutation, add the copy they want, and see how it looks like on Staging.

I've pushed an update with that change. The Growth team would ideally like to share the update during the week beginning 17th Jan, so if you get time to re-review before the ship that week, that would be wonderful!

packages/client/hooks/useNewFeatureSnackbar.ts
```diff
@@ -0,0 +1,45 @@
...
+    if (newFeature) {
+      const {actionButtonCopy, snackbarMessage, url} = newFeature
+      const snack = {
+        key: 'newFeature',
+        autoDismiss: 0,
+        message: snackbarMessage,
+        onDismiss: () => DismissNewFeatureMutation(atmosphere, {}, {onCompleted, onError}),
+        action: {
+          label: actionButtonCopy,
+          callback: () => window.open(url, '_blank', 'noopener')?.focus()
```
mattkrick, line 37: +1 no need for the `.focus()` but it doesn't hurt

mattkrick (approved):
2 nits, good to merge!
loom here, but it's not really worthwhile: https://www.loom.com/share/d2ff14d10c4a4b6891f4948239bc125a

PR #5747, opened 2021-12-06
Spotlight height doesn't change when searching

### LINKED ISSUE #5739: Spotlight: modal height changes when searching
See here:
modal-height-change

### AC

- [ ] When the user types in the search query, the height of the modal doesn't change

nickoferrall:
Fix #5739 

This PR also renames `SpotlightGroups` to `SpotlightResults` to improve naming consistency.

### DIFF
--- a/packages/client/components/SpotlightResults.tsx
+++ b/packages/client/components/SpotlightResults.tsx
@@ -6,10 +6,10 @@ import ReflectionGroup from './ReflectionGroup/ReflectionGroup'
 import {ElementHeight, ElementWidth} from '~/types/constEnums'
 import useGroupMatrix from '../hooks/useGroupMatrix'
 import useResultsHeight from '~/hooks/useResultsHeight'
-import SpotlightGroupsEmptyState from './SpotlightGroupsEmptyState'
-import {SpotlightGroupsQuery} from '~/__generated__/SpotlightGroupsQuery.graphql'
+import SpotlightResultsEmptyState from './SpotlightResultsEmptyState'
+import {SpotlightResultsQuery} from '~/__generated__/SpotlightResultsQuery.graphql'
 
-const SimilarGroups = styled('div')({
+const ResultsWrapper = styled('div')({
   padding: '40px 0px 24px',
   height: '100%',
   width: '100%',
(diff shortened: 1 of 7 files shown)


packages/client/hooks/useResultsHeight.ts
```diff
@@ -1,21 +1,23 @@
...
+    if (newHeight && height !== newHeight) {
       setHeight(newHeight)
     }
   }, [height])
 
   useResizeObserver(() => {
-    // when resized, set height to 100% so that useLayoutEffect can calc the new height
-    setHeight('100%')
+    // when resized & has ref, set height to 100% so that useLayoutEffect can calc the new height
+    if (resultsRef.current) {
```
nickoferrall, line 18: If there's no `resultsRef`, e.g. when we're showing the empty state, the height was being set to `100%` but not being recalculated in `useLayoutEffect` as `newHeight` was `undefined`

packages/client/components/SpotlightResults.tsx
```diff
@@ -88,33 +87,34 @@ const SpotlightGroups = (props: Props) => {
...
-                isSpotlightEntering={isSpotlightEntering}
-              />
-            ))}
-          </Column>
-        ))}
-      </Scrollbar>
-    </SimilarGroups>
+    <ResultsWrapper>
+      {!similarReflectionGroups.length ? (
+        <SpotlightResultsEmptyState height={scrollHeight} />
```
nickoferrall, line 97: EmptyState needs to inherit the padding in `ResultsWrapper`. There's a few different ways we could do this, and I don't love the ternary operator within the JSX, but I prefer this DRY approach vs copying padding to empty state component, two result statements wrapped in the same `ResultsWrapper`, or adding it to the parent component that doesn't contain any styles.

nickoferrall:
Hey @igorlesnenko, could you review this one, please?

igorlesnenko (changes requested):
I still see some height change on my side:

https://user-images.githubusercontent.com/466991/145042147-6447d2da-ea04-4f32-b65d-cc89d56db10e.mov

nickoferrall:
Thanks, @igorlesnenko! I've pushed an update. It's ready when you are

igorlesnenko (approved):
Looks good to me

Issue #5548, opened 2021-10-20
Deployment Reliability: Disable refetch queries on new version bump


mattkrick:
Today when a disconnect occurs, the frontend will refetch all queries to make sure the data is still fresh.
This is the right thing to do if they were disconnected for a handful of seconds, e.g. a mobile device going through a tunnel.
This is the wrong thing to do when we push a new version out, because the time disconnected is likely < 1 second.

AC
- [ ] Determine if we should disable refetches based on a version bump or a disconnected duration. Pick a side & fight me
- [ ] Implement the refetch logic to first check the disconnect duration (or version bump). Only trigger a refetch if necessary. If it is necessary, consider adding a long jitter

Estimate: 8 hours

Dschoordsch:
6.36.0 release deployment resulted in 5min outage, we should prioritize this issue.

igorlesnenko:
@mattkrick I would prefer the version bump approach.

If we go with measuring disconnect time, I'm not sure if it is possible to select that `REFETCH_AFTER` variable correctly:

- If `REFETCH_AFTER = 1 second`, there is no guarantee that the next deploy will not take `1.1 seconds`
- If `REFETCH_AFTER` would be big enough, we might miss a case of `mobile device going through a tunnel`

Checking the version would guarantee that we always handle only a particular redeploy-disconnect case.

Another question is: do we care about theoretical desynchronisation that might occur in this short period of time (<1 second)? 

As soon as we figure out that we don't want to refetch, we can show some snackbar with the text "You’re offline, refresh"?
Or alternatively, we can show it if redeploy-disconnect took some big time.

What do you think?

igorlesnenko:
@mattkrick I created a PR with a version bump approach for now #5749: Left some comments in self-review there.
What do you think?

mattkrick:
> Another question is: do we care about theoretical desynchronization that might occur in this short period of time (<1 second)?

Nope! With the socket heartbeat we can catch anything over 15s, which is probably OK.

> As soon as we figure out that we don't want to refetch, we can show some snackbar with the text "You’re offline, refresh"?
Or alternatively, we can show it if redeploy-disconnect took some big time.

I think the magic all revolves around this line: https://github.com/ParabolInc/parabol/blob/2b26b54a11ecab8569149274978c48dae3123de1/packages/client/hooks/useTrebuchetEvents.ts#L58

Maybe we do something as simple as
```ts
if (oldVersion == newVersion) {
[...]

PR #5749, opened 2021-12-06
Disable refetch queries on new version bump

igorlesnenko:
#5548

See discussion starting from here: https://github.com/ParabolInc/parabol/issues/5548#issuecomment-984867643

How to test:

**1.**

- `yarn build && yarn start`
- Open any page
- Shut down the server
- See 'You are offline' snackbar
- Run `yarn build && yarn start` again
- See queries are refetched after reconnect, the offline message is gone (you probably would see the page goes blank for a second)

**2.**

- `yarn build && yarn start`
- Open any page
- Shut down the server
- See 'You are offline' snackbar
- Edit the version in `package.json` in the root folder
- Run `yarn build && yarn start`
- See offline message is gone
- See that there was no refetch


packages/client/hooks/useTrebuchetEvents.ts
```diff
@@ -55,16 +55,34 @@ const useTrebuchetEvents = () => {
 
     const onReconnected = () => {
       setConnectedStatus(true)
-      atmosphere.retries.forEach((retry) => retry())
```
igorlesnenko, line 58: Looks like there is no way to know the new version on reconnect as it is not possible to read handshake headers. 
The only way to manage reconnect logic based on a version is to retrieve the version from the backend. Backend is already sending the new version automatically on receiving a connection. So I tried to move this logic in `onData` below.

packages/client/hooks/useTrebuchetEvents.ts
```diff
@@ -55,16 +55,34 @@ const useTrebuchetEvents = () => {
...
+            }
+          }
+        })
+        atmosphere.eventEmitter.emit('removeSnackbar', ({key}) => key === 'offline')
+
+        if ('serviceWorker' in navigator) {
+          const registration = await navigator.serviceWorker.getRegistration()
+          registration?.update().catch()
+        }
+      } else if (obj.version && recentDisconnectsRef.current.length > 0) {
```
igorlesnenko, line 83: Only refetch if reconnect. Avoid refetching on initial version load.

packages/client/hooks/useTrebuchetEvents.ts
```diff
@@ -55,16 +55,34 @@ const useTrebuchetEvents = () => {
...
       const obj = typeof payload === 'string' ? JSON.parse(payload) : payload
-      if (obj.version && obj.version !== __APP_VERSION__ && 'serviceWorker' in navigator) {
-        const registration = await navigator.serviceWorker.getRegistration()
-        registration?.update().catch()
+
+      if (obj.version && obj.version !== __APP_VERSION__) {
+        atmosphere.retries.clear()
+        atmosphere.eventEmitter.emit('addSnackbar', {
+          autoDismiss: 0,
+          message: 'You’re offline, Please refresh to continue',
```
igorlesnenko, line 68: We may not want to show this snackbar if:
- We are sure that redeploy would always cause an extra small amount of downtime
- We are sure (or ignore) that in this extra small amount of downtime no data would become stale

Dschoordsch (changes requested):
It doesn't work for me. When I updated the version, it tried reconnecting multiple times but when it finally reconnected, I got the "You're offline, Please refresh to continue" message. Pressing the "Refresh" button reloads the page, but the offline message reappears.

packages/client/hooks/useTrebuchetEvents.ts
Dschoordsch: -1 I'm not sure if I get it correctly. When we're updating, after we reconnected to the server, we're now showing the you're offline message? What's the reasoning behind it?

igorlesnenko:
@Dschoordsch yeah, that is because of versions mismatch. In normal circumstances, if the backend returns a new version and you refresh a frontend, it would also have a new version.
However, I fixed this anyway — now I run that logic only on reconnect and not on browser refresh (initial page load)

packages/client/hooks/useTrebuchetEvents.ts
igorlesnenko: @Dschoordsch As far as I understand, we want to not refetch the queries, if reconnect occurred because of a version bump. 
As data might be outdated in this case, my proposal is to show the refresh button.
This will be shown only on socket reconnect and in case of version change. So the frontend version would be different from the backend. 
I'm actually not sure if It is really needed, we can remove this snackbar.

Dschoordsch: What usually happens on update is that the frontend is reloaded and then it tries to connect to the backend, which is unavailable for some time. So I'm not even sure if that's the right condition, as in that case once both reconnect, the version should be identical, shouldn't it?

igorlesnenko: @Dschoordsch Updated test cases to make it cleaner.
Removed the snackbar as discussed here: https://github.com/ParabolInc/parabol/issues/5548#issuecomment-991183359

How it currently works in master:

- When socket reconnects, frontend is not reloaded, we only refetch GraphQL queries
- When the backend receives a connection it sends an app version to the frontend.
- If this version is different we update service workers.

In my PR, If reconnect occurred, I refetch GraphQL queries only if the version didn't change.

packages/client/hooks/useTrebuchetEvents.ts
```diff
@@ -55,16 +55,27 @@ const useTrebuchetEvents = () => {
...
       // hacky but that way we don't have to double parse huge json payloads. SSE graphql payloads are pre-parsed
       const obj = typeof payload === 'string' ? JSON.parse(payload) : payload
-      if (obj.version && obj.version !== __APP_VERSION__ && 'serviceWorker' in navigator) {
-        const registration = await navigator.serviceWorker.getRegistration()
-        registration?.update().catch()
+
+      // obj.version is sent once on connection, if we have disconnects logged, it's reconnect
+      const isReconnect = obj.version && recentDisconnectsRef.current.length > 0
+
+      if (obj.version && obj.version !== __APP_VERSION__) {
```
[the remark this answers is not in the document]
igorlesnenko: @Dschoordsch thanks, make sense! Fixed.

packages/client/hooks/useTrebuchetEvents.ts
```diff
@@ -13,6 +13,7 @@ const useTrebuchetEvents = () => {
   const {history} = useRouter()
   const firewallMessageSentRef = useRef(false)
   const recentDisconnectsRef = useRef([] as number[])
+  const serverVersion = useRef(__APP_VERSION__)
```
Dschoordsch, line 16: +1 I know I named it `serverVersion`, but in context it should be `serverVersionRef` to be consistent.

packages/client/hooks/useTrebuchetEvents.ts
```diff
@@ -55,16 +56,28 @@ const useTrebuchetEvents = () => {
...
+
+      if (obj.version && obj.version !== serverVersion.current) {
+        serverVersion.current = obj.version
+        if ('serviceWorker' in navigator) {
+          const registration = await navigator.serviceWorker.getRegistration()
+          registration?.update().catch()
+        }
+
+        if (isReconnect) {
+          atmosphere.retries.clear()
```
Dschoordsch, line 77: -1 It didn't occur to me before, but this is wrong. Retries are statically added by hooks
https://github.com/ParabolInc/parabol/blob/a181ff6e9102be689d12641d3bef3a6009097ecd/packages/client/hooks/useQueryLoaderNow.tsx#L27-L30
and just re-executed on each reconnect. That means if we clear the queue here, these will never be re-executed unless the hook is re-evaluated. We should just remove this clear part.

Dschoordsch (changes requested):

packages/client/hooks/useTrebuchetEvents.ts
igorlesnenko: Fixed

packages/client/hooks/useTrebuchetEvents.ts
igorlesnenko: @Dschoordsch fixed

Dschoordsch (approved):
Tested:
* reconnect
* version upgrade
* reconnect after version upgrade
* reopen the site

packages/client/hooks/useTrebuchetEvents.ts
```diff
@@ -55,16 +56,24 @@ const useTrebuchetEvents = () => {
...
 
     const onData = async (payload: string | object) => {
       // hacky but that way we don't have to double parse huge json payloads. SSE graphql payloads are pre-parsed
       const obj = typeof payload === 'string' ? JSON.parse(payload) : payload
-      if (obj.version && obj.version !== __APP_VERSION__ && 'serviceWorker' in navigator) {
-        const registration = await navigator.serviceWorker.getRegistration()
-        registration?.update().catch()
+
+      // obj.version is sent once on connection, if we have disconnects logged, it's reconnect
+      const isReconnect = obj.version && recentDisconnectsRef.current.length > 0
```
mattkrick, line 67: +1 nit-- since checking for `obj.version` is in both branches of the conditional, we can clean it up a little like

```ts
if (obj.version) {
  // this is the initial payload
  if (obj.version !== serverVersionRef.current) {
    // handle upgrade
  } else if (recentDisconnectsRef.current.length > 0) {
    // handle reconnect
  }
}

mattkrick (approved):
just 1 nit, check it out & merge when ready

PR #5755, opened 2021-12-08
Fix integration search

### LINKED ISSUE #5612: Search in integration dropdown does not work
## Issue - Bug

See Loom: https://www.loom.com/share/85c7a11754324cc882d4fb1f5a677257

Users can:

- [ ] Search properly in task integration dropdown menu

nickoferrall:
Fix #5612 

Initially, when we open the dropdown, it returns the projects/repos that we've previously integrated with a task. We only see all of the available projects & repos when we search. From a UX perspective, I think this is a little confusing, but the performance advantages probably mean it's the right move.

After adding or removing an integration, we have to refresh the page before the change is reflected in the dropdown search. I've created an issue for this: #5756


packages/client/components/TaskFooterIntegrateMenuList.tsx
```diff
@@ -70,13 +70,13 @@ const Label = styled('div')({
 const getValue = (
   item: NonNullable<TaskFooterIntegrateMenuList_suggestedIntegrations['items']>[0]
 ) => {
-  const jiraItemName = item?.remoteProject?.name ?? ''
+  const jiraItemName = item?.projectKey ?? ''
   const githubName = item?.nameWithOwner ?? ''
-  const name = jiraItemName ?? githubName
+  const name = jiraItemName || githubName
```
nickoferrall, line 75: The nullish coalescing operator `??` only returns the right-hand side of the operand when the left-hand side is `null` or `undefined`. So, when `jiraItemName` was `""`, we were still returning the `jiraItemName`. With the logical or `||` operator, we return the `githubName` when `jiraItemName` is falsy, e.g. when it's an empty string.

packages/client/components/TaskFooterIntegrateMenuList.tsx
```diff
@@ -70,13 +70,13 @@ const Label = styled('div')({
 const getValue = (
   item: NonNullable<TaskFooterIntegrateMenuList_suggestedIntegrations['items']>[0]
 ) => {
-  const jiraItemName = item?.remoteProject?.name ?? ''
+  const jiraItemName = item?.projectKey ?? ''
```
nickoferrall, line 73: `remoteProject` is undefined because `teamId` is undefined [here](https://github.com/ParabolInc/parabol/blob/8409e5ee0ab86c9effcb616dd82cdf30053c81ee/packages/server/graphql/types/SuggestedIntegrationJira.ts#L33). I'm not sure why the source doesn't have the `teamId`, as `SuggestedIntegration` belongs to the `TeamMember` which must have the `teamId`. 

As I'm away tomorrow & Friday, I thought it would be best to share this PR now rather than wait until I get to the bottom of this, which would probably be on Monday. 

We can still filter by `projectKey` as this is what we're displaying in the dropdown.

packages/server/graphql/queries/helpers/suggestedIntegrationHelpers.ts
```diff
@@ -79,7 +79,7 @@ export const getTeamIntegrationsByUserId = async (
   return (
     res
       // jira integrations are making it through that don't have a projectKey
-      .filter((res) => res.service !== 'jira' || res.projectKey)
+      .filter((res) => permLookup[res.service] && (res.service !== 'jira' || res.projectKey))
```
nickoferrall, line 82: This fixes the following issue:

- Remove this change
- Add the Jira & GitHub integrations
- Integrate a task with Jira and then remove the Jira integration
- Open the dropdown menu on a task card
- See that it shows the recently used Jira project, even though we're no longer integrated with Jira

I felt that it's best to filter here in the helper file rather than in `suggestedIntegration` as everything in `suggestedIntegration` uses `teamIntegrationsByTeamId`, which is what is returned from this function. By filtering at the source, we don't need to filter elsewhere.

packages/client/components/TaskFooterIntegrateMenu.tsx
```diff
@@ -91,18 +89,16 @@ const TaskFooterIntegrateMenu = (props: Props) => {
       isViewerIntegrated.hasGitHub,
       isViewerIntegrated.hasAtlassian
     )
-    const label = 'Push with your credentials.'
```
nickoferrall, line 94: I was a bit confused by this - is it informing users that they can push to Jira or GitHub with their user credentials instead of team credentials? Or is it letting them know that they can push a task straight to Jira/GitHub? Either way, I felt that removing the full stop was a tiny bit more friendly

nickoferrall:
Hey @tianrunhe, would you mind reviewing this one, please?

nickoferrall:
@tianrunhe just bumping this one in case you didn't see the review request

nickoferrall:
It looks like Bruce is OOO now so I'll request a new reviewer. @BartoszJarocki could you take a look at this one, please?

packages/server/graphql/queries/helpers/suggestedIntegrationHelpers.ts
```diff
@@ -35,34 +36,33 @@ export const useOnlyUserIntegrations = (
...
   // the user is integrated against every team integration, use those
   return userIntegrationsForTeam.length === integrationSet.size
     ? userIntegrationsForTeam
     : undefined
 }
 
-export const getTeamIntegrationsByUserId = async (
-  teamId: string
-): Promise<IntegrationByUserId[]> => {
+export const getTeamIntegrationsByTeamId = async (
```
BartoszJarocki, line 47: +2 thanks for the naming fixes, it's all easier to read now!

BartoszJarocki (approved):
nicely done, works like a charm!

packages/server/graphql/queries/suggestedIntegrations.ts
```diff
@@ -33,33 +38,30 @@ export default {
         return standardError(new Error('Not on same team as user'), {userId: viewerId})
       }
     }
-    const teamIntegrationsByUserId = await getTeamIntegrationsByUserId(teamId)
+    const permLookup: PermLookup = await getPermsByTaskService(dataLoader, teamId, userId)
```
mattkrick, line 41: -1 let's remove the `PermLookup` here as the type is already inferred automatically

packages/server/graphql/queries/helpers/suggestedIntegrationHelpers.ts
```diff
@@ -35,34 +36,33 @@ export const useOnlyUserIntegrations = (
...
+  teamId: string,
+  permLookup: PermLookup
```
mattkrick, line 49: -1 since we're tightly coupling `getPermsByTaskService` to this call, we can just write 
`permLookup: ReturnType<typeof getPermsByTaskService>` and then we have a single source of truth instead of also maintaining a separate interface.

mattkrick (changes requested):
-1 just 2 little fixes & good to merge.
meta: feels weird to score this a -1 because the changes are so inconsequential, but i wanna make sure they land so we don't have to maintain an extra little interface. maybe that's a +1? not sure

nickoferrall:
Thanks, fixes pushed @mattkrick. I'll merge it to master

PR #5782, opened 2021-12-15
Fix isSpectating toggle & Reveal Votes count

### LINKED ISSUE #5773: Poker "I don't vote" button is buggy
Clicking on the "I don't vote" or "Let me vote!" button seems to fire twice sometimes. The Poker deck can quickly flash away and then reappear or vice versa. 

dont-vote

### AC: 

- [ ] Clicking the "I don't vote" or "Let me vote!" button fires once and the deck of card either disappears or appears

### LINKED ISSUE #5775: If the facilitator isn't voting, don't include them in Reveal Votes count
In Poker, the Reveal Votes progress bar includes the facilitator even when they select that they're not voting.

reveal-votes

### AC

- [ ] If the facilitator has selected that they're not voting, don't include them in the Reveal Votes count

nickoferrall:
Fix #5775 & fix #5773

### DIFF
--- a/packages/client/components/EstimateDimensionColumn.tsx
+++ b/packages/client/components/EstimateDimensionColumn.tsx
@@ -75,7 +75,7 @@ const EstimateDimensionColumn = (props: Props) => {
   const setSpectating = (isSpectating: boolean) => () => {
     if (submitting) return
     submitMutation()
-    SetPokerSpectateMutation(atmosphere, {meetingId, isSpectating, stageId}, {onError, onCompleted})
+    SetPokerSpectateMutation(atmosphere, {meetingId, isSpectating}, {onError, onCompleted})
   }
   const showVoting = isVoting || isClosing
   return (
--- a/packages/client/components/EstimateDimensionColumn.tsx
+++ b/packages/client/components/EstimateDimensionColumn.tsx
@@ -3,6 +3,7 @@ import graphql from 'babel-plugin-relay/macro'
 import React from 'react'
 import {createFragmentContainer} from 'react-relay'
 import useMutationProps from '~/hooks/useMutationProps'
+import VoteForPokerStoryMutation from '~/mutations/VoteForPokerStoryMutation'
 import useAtmosphere from '../hooks/useAtmosphere'
 import useIsInitializing from '../hooks/useIsInitializing'
 import useIsPokerVotingClosing from '../hooks/useIsPokerVotingClosing'
@@ -61,7 +62,7 @@ const EstimateDimensionColumn = (props: Props) => {
   const {endedAt, facilitatorUserId, id: meetingId, viewerMeetingMember} = meeting
   const isSpectating = viewerMeetingMember?.isSpectating
   const isFacilitator = viewerId === facilitatorUserId
-  const {id: stageId, dimensionRef} = stage
+  const {id: stageId, dimensionRef, scores} = stage
   const {name} = dimensionRef
   const {isVoting} = stage
   const {onError, onCompleted, submitMutation, error, submitting} = useMutationProps()
@@ -76,6 +77,10 @@ const EstimateDimensionColumn = (props: Props) => {
     if (submitting) return
     submitMutation()
     SetPokerSpectateMutation(atmosphere, {meetingId, isSpectating}, {onError, onCompleted})
+    const viewerHasVoted = !!scores.find(({userId}) => userId === viewerId)
+    if (viewerHasVoted) {
+      VoteForPokerStoryMutation(atmosphere, {meetingId, stageId}, {onError, onCompleted})
+    }
   }
   const showVoting = isVoting || isClosing
   return (
@@ -128,6 +133,9 @@ export default createFragmentContainer(EstimateDimensionColumn, {
       isVoting
       dimensionRef {
         name
+      }
+      scores {
+        userId
       }
     }
   `,
--- a/packages/client/components/EstimateDimensionColumn.tsx
+++ b/packages/client/components/EstimateDimensionColumn.tsx
@@ -3,7 +3,6 @@ import graphql from 'babel-plugin-relay/macro'
 import React from 'react'
 import {createFragmentContainer} from 'react-relay'
 import useMutationProps from '~/hooks/useMutationProps'
-import VoteForPokerStoryMutation from '~/mutations/VoteForPokerStoryMutation'
 import useAtmosphere from '../hooks/useAtmosphere'
 import useIsInitializing from '../hooks/useIsInitializing'
 import useIsPokerVotingClosing from '../hooks/useIsPokerVotingClosing'
@@ -62,7 +61,7 @@ const EstimateDimensionColumn = (props: Props) => {
   const {endedAt, facilitatorUserId, id: meetingId, viewerMeetingMember} = meeting
   const isSpectating = viewerMeetingMember?.isSpectating
   const isFacilitator = viewerId === facilitatorUserId
-  const {id: stageId, dimensionRef, scores} = stage
+  const {id: stageId, dimensionRef} = stage
   const {name} = dimensionRef
   const {isVoting} = stage
   const {onError, onCompleted, submitMutation, error, submitting} = useMutationProps()
@@ -76,11 +75,7 @@ const EstimateDimensionColumn = (props: Props) => {
   const setSpectating = (isSpectating: boolean) => () => {
     if (submitting) return
     submitMutation()
-    SetPokerSpectateMutation(atmosphere, {meetingId, isSpectating}, {onError, onCompleted})
-    const viewerHasVoted = !!scores.find(({userId}) => userId === viewerId)
-    if (viewerHasVoted) {
-      VoteForPokerStoryMutation(atmosphere, {meetingId, stageId}, {onError, onCompleted})
-    }
+    SetPokerSpectateMutation(atmosphere, {meetingId, isSpectating, stageId}, {onError, onCompleted})
   }
   const showVoting = isVoting || isClosing
   return (
@@ -133,9 +128,6 @@ export default createFragmentContainer(EstimateDimensionColumn, {
       isVoting
       dimensionRef {
         name
-      }
-      scores {
-        userId
       }
     }
   `,
--- a/packages/client/components/PokerActiveVoting.tsx
+++ b/packages/client/components/PokerActiveVoting.tsx
@@ -1,6 +1,6 @@
 import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
-import React from 'react'
+import React, {useMemo} from 'react'
 import {createFragmentContainer} from 'react-relay'
 import useMutationProps from '~/hooks/useMutationProps'
 import {PALETTE} from '~/styles/paletteV3'
@@ -100,8 +100,11 @@ const PokerActiveVoting = (props: Props) => {
   const {id: stageId, scores} = stage
   const hasVotes = scores.length > 0
   const isFacilitator = viewerId === facilitatorUserId
-  const viewerHasVoted = Boolean(scores.find(({userId}) => userId === viewerId))
-  const checkedInCount = meetingMembers.length
+  const viewerHasVoted = useMemo(() => !!scores.find(({userId}) => userId === viewerId), [scores])
+  const checkedInCount = useMemo(
+    () => meetingMembers.filter(({isSpectating}) => !isSpectating).length,
+    [meetingMembers]
+  )
   const votePercent = scores.length / checkedInCount
   const allVotesIn = scores.length === checkedInCount
   // Show the facilitator a tooltip if nobody has voted yet
@@ -176,6 +179,7 @@ export default createFragmentContainer(PokerActiveVoting, {
       id
       meetingMembers {
         id
+        isSpectating
       }
     }
   `
--- a/packages/client/components/PokerActiveVoting.tsx
+++ b/packages/client/components/PokerActiveVoting.tsx
@@ -100,7 +100,7 @@ const PokerActiveVoting = (props: Props) => {
   const {id: stageId, scores} = stage
   const hasVotes = scores.length > 0
   const isFacilitator = viewerId === facilitatorUserId
-  const viewerHasVoted = useMemo(() => !!scores.find(({userId}) => userId === viewerId), [scores])
+  const viewerHasVoted = scores.find(({userId}) => userId === viewerId)
   const checkedInCount = useMemo(
     () => meetingMembers.filter(({isSpectating}) => !isSpectating).length,
     [meetingMembers]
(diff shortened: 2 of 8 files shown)


packages/client/components/EstimateDimensionColumn.tsx
```diff
@@ -76,6 +77,10 @@ const EstimateDimensionColumn = (props: Props) => {
     if (submitting) return
     submitMutation()
     SetPokerSpectateMutation(atmosphere, {meetingId, isSpectating}, {onError, onCompleted})
+    const viewerHasVoted = !!scores.find(({userId}) => userId === viewerId)
+    if (viewerHasVoted) {
+      VoteForPokerStoryMutation(atmosphere, {meetingId, stageId}, {onError, onCompleted})
```
nickoferrall, line 82: If a user clicks "I don't vote", we shouldn't include them in the "Reveal Votes" progress count. There's a few ways of doing this:
1. Filter spectators in `scores` in `EstimateStage` [here](https://github.com/ParabolInc/parabol/blob/8685720f380ccd0081aa413c7d64fa8fb0cb91fe/packages/server/graphql/types/EstimateStage.ts#L177). This would mean grabbing all users every time we query `scores`, and we would need to run the query after every `SetPokerSpectateMutation`.
2. Filter on the client. This would add an extra burden to the client, and we could forget to filter elsewhere
3. My approach is removing the viewer's vote if they click "I don't vote", which means we don't need to think about filtering spectator scores. However, we do need to query `scores` on the client. Also, if the viewer votes for 5, clicks "I don't vote", clicks "Let me vote!", they won't see their vote for number 5. I'd say this is arguably the expected UX and keeps things clean logic wise.

nickoferrall:

packages/client/components/EstimateDimensionColumn.tsx
Dschoordsch: -1 I think the idea is right, but I think it should be part of `setPokerSpectate` mutation on server side and the optimistic updater.

Dschoordsch (changes requested):

packages/server/graphql/mutations/setPokerSpectate.ts
```diff
@@ -47,6 +50,9 @@ const setPokerSpectate = {
         .get(teamMemberId)
         .update({isSpectatingPoker: isSpectating, updatedAt: now})
     }).run()
+
+    // mutate the dataLoader cache
+    meetingMember.isSpectating = isSpectating
```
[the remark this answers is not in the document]
seabornlee: Use a fresh DataLoader or call dataLoader.clear() can fix this bug. But when I check other mutations such as `renameMeeting`, it doesn't need to neither update nor clear the cache, it also works well.
Do you know the root cause? @Dschoordsch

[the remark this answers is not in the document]
seabornlee: I got other two questions. Begging for your help.
1. I try to remove `did` from `subOptions` in the mutation, but it still get the same `dataLoader` in subscriptions.
2. When I remove `publish(....)` from mutation, everything works fine. Is it redundant?
Thx!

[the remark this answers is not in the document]
Dschoordsch: @seabornlee the `operationId` in `subOptions` is passed on as `dataloaderId` here https://github.com/ParabolInc/parabol/blob/f9c00c48cb8600bff2f8e3040a77580f006be45a/packages/server/graphql/ResponseStream.ts#L23 so you would need to remove that to test it. But we pass it intentionally to reduce the DB roundtrips for all subscriptions. Publish is not redundant since it notifies all meeting participants, who would otherwise have stale data.
We always have to update or clear the dataloader if we modify an entry in DB after we've loaded it through the dataloader and we pass the dataloader to the subscription. Skimming through the mutations it seems most of the time we're not publishing the same entities we've loaded before, that's why it's not an issue most of the time.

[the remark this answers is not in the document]
nickoferrall: I see, so the sequence of events with the previous implementation was:
- call `SetSpectateMutation` which creates a new dataLoader and cache
- in the mutation, grab the `meetingMember` with the dataLoader
- update the `meetingMember` in the db
- subscription then uses the stale dataLoader when reading the `meetingMember`
- call `SetSpectateMutation` again, which creates another dataLoader and cache using the previously updated `meetingMember`
- now the subscription will read the `meetingMember` that was updated in the previous mutation execution

That makes sense, but it feels error-prone. I wonder whether we could clear the dataLoader automatically after executing a mutation?

packages/client/mutations/SetPokerSpectateMutation.ts
```diff
@@ -33,14 +32,6 @@ const SetPokerSpectateMutation: StandardMutation<TSetPokerSpectateMutation> = (
   return commitMutation<TSetPokerSpectateMutation>(atmosphere, {
     mutation,
     variables,
-    optimisticUpdater: (store) => {
```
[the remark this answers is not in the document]
nickoferrall: The response is super fast with just the updater, even with a slow 3G connection and minimal CPU, so the optimistic updater feels unnecessary here

packages/client/components/EstimateDimensionColumn.tsx
nickoferrall: I was umming and arring about client vs server side here. It's more verbose server-side vs client, but I think you're right. If the next dev uses `SetPokerSpectateMutation ` and forgets to use `VoteForPokerStoryMutation`, they'll run into the same bug.

packages/server/graphql/mutations/setPokerSpectate.ts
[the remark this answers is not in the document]
Dschoordsch: That's the trade of we're making for performance. It's specifically designed so that subscriptions reuse the data and don't query the db again.
It seems like it doesn't occur that often, but we could add a verification mode, where instead of reusing the dataloader it would create a new one and compare it. This we could activate for dev and staging to find these errors earlier.

packages/client/components/PokerActiveVoting.tsx
```diff
@@ -100,8 +100,11 @@ const PokerActiveVoting = (props: Props) => {
   const {id: stageId, scores} = stage
   const hasVotes = scores.length > 0
   const isFacilitator = viewerId === facilitatorUserId
-  const viewerHasVoted = Boolean(scores.find(({userId}) => userId === viewerId))
-  const checkedInCount = meetingMembers.length
+  const viewerHasVoted = useMemo(() => !!scores.find(({userId}) => userId === viewerId), [scores])
+  const checkedInCount = useMemo(
+    () => meetingMembers.filter(({isSpectating}) => !isSpectating).length,
```
Dschoordsch, line 105: I think it is better to do this on server side as well, and publish the scores on updating `isSpectating`, see my test results

Dschoordsch (changes requested):
I did some testing and found some odd behaviour:
* start poker with 2 members Frank the facilitator and Viktor the non-voter and 2 issues
* Viktor votes 8 on the first issue
* Viktor goes to the second issue and presses 'I don't vote'
* Frank now sees 'Reveal votes' although he didn't vote yet. Revealing will show Viktors 8 vote and Franks 🚫

packages/server/graphql/mutations/setPokerSpectate.ts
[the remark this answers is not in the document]
nickoferrall: Makes sense, and thanks for creating the issue. I look forward to getting Matt's thoughts on this. The same stale dataLoader bug did crop up in a recent PR here: https://github.com/ParabolInc/parabol/pull/5674

packages/client/components/PokerActiveVoting.tsx
nickoferrall: Nice catch with the test! That was happening because I was removing the viewer's vote in the stage where they clicked "I don't vote". However, it looks like "I don't vote" applies to all stages, so I'm now removing their vote in all stages. I've moved this logic to the server. 

I think the `checkedInCount` needs to be calculated on the client as we want to know the total number of meetingMembers that are not spectating. Update pushed!

packages/server/graphql/mutations/setPokerSpectate.ts
```diff
@@ -47,7 +69,19 @@ const setPokerSpectate = {
...
+    stages.forEach(async (stage) => {
+      const {id: stageId, scores} = stage
+      const viewerVotedInStage = scores.find(({userId}) => userId === viewerId)
+      if (viewerVotedInStage) {
+        const newStageScores = scores.filter(({userId}) => userId !== viewerId)
+        stage.scores = newStageScores
+        await removeVoteForUserId(viewerId, stageId, meetingId)
+      }
+    })
+    const data = {meetingId, userId: viewerId, stages, teamId, phaseType: 'ESTIMATE'}
```
Dschoordsch, line 84: +1 I think we could optimize this slightly to match your description of this field, i.e. maintain a second array `updatedStages` and only send this here instead of all stages.

Dschoordsch (approved):
Tested, works.

packages/client/components/PokerActiveVoting.tsx
```diff
@@ -100,8 +100,11 @@ const PokerActiveVoting = (props: Props) => {
   const {id: stageId, scores} = stage
   const hasVotes = scores.length > 0
   const isFacilitator = viewerId === facilitatorUserId
-  const viewerHasVoted = Boolean(scores.find(({userId}) => userId === viewerId))
-  const checkedInCount = meetingMembers.length
+  const viewerHasVoted = useMemo(() => !!scores.find(({userId}) => userId === viewerId), [scores])
```
mattkrick, line 103: AFAIK sometimes using `useMemo` is more expensive than not using it.
that said, it doesn't matter much either way, so typically if the arrays are < 100 I just don't memoize things.
but IIRC, you had a blog post that was talking about this? would love to learn more

packages/client/mutations/SetPokerSpectateMutation.ts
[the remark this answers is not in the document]
mattkrick: +1 it's typically the roundtrip time that we're concerned with, not the server processing time. by removing an optimistic updater, the `isSpectating` value is incorrect for the entire round trip, which could be a couple seconds! up to you

packages/server/graphql/mutations/setPokerSpectate.ts
```diff
@@ -32,13 +36,32 @@ const setPokerSpectate = {
...
-
+    const [meetingMember, meeting] = await Promise.all([
+      dataLoader.get('meetingMembers').load(meetingMemberId) as Promise<PokerMeetingMember>,
+      dataLoader.get('newMeetings').load(meetingId)
+    ])
+    if (!meeting) {
+      return {error: {message: 'Meeting not found'}}
+    }
+    const {endedAt, phases, meetingType, teamId} = meeting
+    if (!isTeamMember(authToken, teamId)) {
```
mattkrick, line 47: +1 not strictly necessary since we have the `meetingMember` check below

packages/server/graphql/types/SetPokerSpectatePayload.ts
```diff
@@ -21,6 +25,24 @@ export const SetPokerSpectateSuccess = new GraphQLObjectType<any, GQLContext>({
         const meetingMember = await dataLoader.get('meetingMembers').load(meetingMemberId)
         return meetingMember
       }
+    },
+    stages: {
```
mattkrick, line 29: +1 I'd call this `updatedStages` (because I didn't read the description & just assumed this was all stages)

packages/server/graphql/types/SetPokerSpectatePayload.ts
```diff
@@ -21,6 +25,24 @@ export const SetPokerSpectateSuccess = new GraphQLObjectType<any, GQLContext>({
...
+      type: new GraphQLList(new GraphQLNonNull(EstimateStage)),
```
mattkrick, line 30: -1 `type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(EstimateStage)))`
always nice to guarantee that you'll get an empty array on the client so you don't have to check nullability before iterating over it.


+1 the type here is fixed to `EstimateStage` which means we don't need the `phaseType` in the resolver source

packages/server/graphql/types/SetPokerSpectatePayload.ts
```diff
@@ -21,6 +25,24 @@ export const SetPokerSpectateSuccess = new GraphQLObjectType<any, GQLContext>({
...
+        phaseType,
+        teamId,
+        meetingId
+      }: {
+        dirtyStages: EstimateStageDB[]
+        phaseType: NewMeetingPhaseTypeEnum
+        teamId: string
+        meetingId: string
+      }) => {
+        return dirtyStages.map((stage) => augmentDBStage(stage, meetingId, phaseType, teamId))
```
mattkrick, line 44: can hardcode in `ESTIMATE` in lieu of phaseType here

packages/client/components/PokerCardDeck.tsx
```diff
@@ -15,12 +15,14 @@ import {BezierCurve, PokerCards} from '../types/constEnums'
...
 
-const Deck = styled('div')<{left: number}>(({left}) => ({
+const Deck = styled('div')<{left: number; isSpectating: boolean}>(({left, isSpectating}) => ({
   bottom: 0,
   display: 'flex',
   left,
+  opacity: isSpectating ? 0 : 1,
   position: 'fixed',
   transition: `200ms ${BezierCurve.DECELERATE}`,
+  visibility: isSpectating ? 'hidden' : 'visible',
```
mattkrick, line 25: +2 for the opacity + visibility! the transition is REALLY nice

packages/client/components/PokerCardDeck.tsx
```diff
@@ -182,11 +184,11 @@ const PokerCardDeck = (props: Props) => {
     }
   }, [maxSlide, isCollapsed])
   // const transform = maxSlide > 0 && !isCollapsed ? `translateX(${swipe.translateX}px)` : undefined
-  if (isSpectating) return null
   return (
     <Deck
       ref={deckRef}
       left={left}
+      isSpectating={!!isSpectating}
```
mattkrick, line 191: +1 I'd set this to a boolean on L52. the earlier we can establish strict types, the better!

mattkrick (changes requested):
adding non-null to the return type of the dirty stages is the only negative & tbh that is probably a nit in itself. nice work!
loom doesn't have much, but here it is for posterity: https://www.loom.com/share/3b8ab199dfda49b98b38abc029994c84

packages/client/components/PokerActiveVoting.tsx
nickoferrall: True! Adding `useMemo` adds a function call which makes the performance worse when arrays are small, plus it adds extra complexity. Removed!

PR #5783, opened 2021-12-15
Adds skip link to dashboard

enriquesanchez:
This PR adds a skip link to the app's dashboard views. It mainly consists of two changes:

1. Creates an `a` tag that is visually hidden unless it receives keyboard focus
2. Replaces a `div` with a `main` tag for the main content area. This is so the skip link can point to the main area.

In short, this makes the first interactive item in the Web page a link to the beginning of the main content.

**Related WCAG criteria:**
[2.4.1 Bypass Blocks](https://www.w3.org/WAI/WCAG21/quickref/?versions=2.1#qr-navigation-mechanisms-skip)
_A mechanism is available to bypass blocks of content that are repeated on multiple Web pages._

This is a common a11y feature and there are many examples and best practices. For reference, check out [gov.uk](https://www.gov.uk/) and [whitehouse.gov](https://www.whitehouse.gov/).

Screen Shot 2021-12-15 at 12 23 45-fullpage

It's important to note that the skip link will only be visible when focused using a keyboard or a screen reader. Users who navigate with touch or mouse will not see the skip link. 

**How to test**
Open the app on a web browser and press <kbd>TAB</kbd>.

### DIFF
--- a/packages/client/components/Dashboard.tsx
+++ b/packages/client/components/Dashboard.tsx
@@ -14,6 +14,7 @@ import MobileDashSidebar from './Dashboard/MobileDashSidebar'
 import DashTopBar from './DashTopBar'
 import MobileDashTopBar from './MobileDashTopBar'
 import SwipeableDashSidebar from './SwipeableDashSidebar'
+import {PALETTE} from '~/styles/paletteV3'
 
 const MeetingsDash = lazy(() =>
   import(/* webpackChunkName: 'MeetingsDash' */ '../components/MeetingsDash')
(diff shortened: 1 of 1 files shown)


ackernaut (approved):
This looks real good, and on a first pass! 🙇 

When using a screen reader this will help folks bypass the nav etc. first hitting the main section (i.e. the meeting cards or whatever is in that view)?

Noticeably, when using the screen and keyboard it doesn’t do much ‘skipping’ because we fix and overflow the main view, but I agree with the overall pattern!

It works particularly well when tabbing to the next items in the top bar (from the other PR), I like that! 🙌 

I’m merged master to see how that works. Once it clears tests I’ll merge. 2 PRs in one day!

enriquesanchez:
Thanks for the review, @ackernaut 🙌

>When using a screen reader this will help folks bypass the nav etc. first hitting the main section (i.e. the meeting cards or whatever is in that view)?

That's right!

PR #5804, opened 2021-12-17
breakout REDIS_URL and REDIS_PORT into different env vars

adaniels-parabol:
getaddrinfo at GetAddrInfoReqWrap.onlookup in node dns.js (core node:dns code) will try to resolve DNS on failure with the connection string used for the Redis connection (e.g. REDIS_URL). If the port number is part of the connection string, node:dns tries to resolve the hostname as IP:PORT (e.g. 10.59.244.137:6379) vs just the IP (10.59.244.137). As IP:PORT (e.g. 10.59.244.137:6379) is not a valid DNS entry so DNS resolution of hostname to IP fails.

getaddrinfo uses OS DNS cache/resolver, thus when a refresh occurs the hostname DNS lookup includes the port number causing a failure. This issue occurs on PPMIs when a redis container may scale or restart. Currently, healthchecks detect this issue and resolve the problem with a container-replace which re-establishes network connectivity to redis and resolves the error until a new DNS record for the redis host is requested. This is a permanent fix and doesn't risk any type of outage during container replace.

I've validated this fix on a PPMI and was able to identify the incorrectly formatted hostname after analyzing a tcpdump of packets with the incorrect dns format e.g. 10.59.244.137:6379.


packages/gql-executor/RedisStream.ts
```diff
@@ -9,7 +9,7 @@ export default class RedisStream<T> implements AsyncIterableIterator<T> {
   private stream: string
   private consumerGroup: string
   // xreadgroup blocks until a response is received, so this needs its own connection
-  private redis = new Redis(REDIS_URL)
+  private redis = new Redis(Number(REDIS_PORT), REDIS_URL)
```
Dschoordsch, line 12: -1 this doesn't fix the problem. ioredis parses the arguments
https://github.com/luin/ioredis/blob/98ebeec1acd2067e52169f53e57e989ebbf671ff/lib/redis/index.ts#L199-L200
so both versions (master and PR) result in the same redis options
```
{
  port: 6379,
  host: 'localhost',
...
}
```
Unless there is some library reading the environment variable directly?

Dschoordsch (changes requested):
Nice find, but seems to be deeper in ioredis

packages/gql-executor/RedisStream.ts
adaniels-parabol: Interesting @Dschoordsch I did verify in tcp dump that the dns resolver isn't adding the port number in the resolution request any longer. So, this appears to have resolved the problem. I'll continue to monitor and see if the issue arises again, but thus far the change appears to have resolved the issue.

.env.example
```diff
@@ -47,9 +47,11 @@ POSTGRES_DB=parabol-saas
 POSTGRES_HOST=localhost
 POSTGRES_PORT=5432
 PROTO='http'
-REDIS_URL='redis://localhost:6379'
+REDIS_URL='redis://localhost'
+REDIS_PORT='6379'
 RETHINKDB_URL='rethinkdb://localhost:28015/actionDevelopment'
-#REDIS_URL='redis://redis:6379'
+#REDIS_URL='redis://redis'
+#REDIS_PORT='6379'
```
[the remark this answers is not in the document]
adaniels-parabol: Sure

packages/gql-executor/RedisStream.ts
Dschoordsch: Then it must be fixed in the meantime and we're running different versions locally and on PPMI. I'm fairly certain it has no effect on my system because I printed the redis options with both versions. Since parsing is done by node Url module, maybe the node versions are different? I'm running v16.9.1

Dschoordsch: @adaniels-parabol  Could this be related? https://github.com/nodejs/node/issues/13708

adaniels-parabol: I don't think so. We are also running 16.9.1 on PPMI. I've checked again, still no further errors after implementing above code. Is it safe to move forward with this PR?

Dschoordsch: Yes, it's safe to move forward with it, but unless we had a REDIS_URL without protocol, like `example.com:1234` I don't see this changing anything. I would rather like finding the real cause. Can you try logging the options directly in the redis constructor?

mattkrick:
closing for now as it looks like ioredis already does the parsing for us & the problem is somewhere else. if i'm wrong feel free to reopen!

Issue #5567, opened 2021-10-24
Design: supporting self-hosted services


jordanh:
## Overview

Extending Parabol to support self-hosted integrations in addition to the multi-tenant SaaS integrations we already support takes careful consideration. There are some useful patterns from other tools we can look at. There are also some interesting questions raised by envisioning the user's journey when using our app. Let's start by illustrating this journey.

### User journey

Imagine an agile Team Leader working at a security-sensitive organization. Many of their team's tools are self hosted: They use Mattermost (see: #4412) for short messaging and GitLab (see: #5564) for backlog management and CI/CD. Our Team Leader wants to use Parabol to help facilitate their agile meetings and workflow, and wants to begin by running a Sprint Poker meeting with their team. When they run a test Sprint Poker meeting, they see GitLab is supported, they want to add the GitLab integration so that they can import their GitLab Issues for scoping.

### Present state

When an integration is added by a user today, they may navigate to the Team→Settings→Integrations area _or_ they may add the integration via an in-line affordance—such as during the Sprint Poker Scope phase. Adding an integration, such as GitHub, will trigger an oauth authentication to fetch a token that can be used on behalf of the user/team to access the service provided by the integration. In all cases, Parabol will raise a popup browser window to begin the authenticate with the integration provider's service.

However, how will our journey change to support selecting a self-hosted oauth provider?

### More questions

Where will users go to add additional oauth providers (i.e. self-hosted services)? Which users can add these additional providers?

Will we support multiple self-hosted servers per team for the same integration (i.e. two Jira Server instances on the same team)? How will users select which oauth provider to use?

How will users choose between a canonical public SaaS service (i.e. Jira Cloud) and a self-hosted service (i.e. Jira Server)?

Not all self-hosted servers have oauth enabled by default, most (all?) also provide the ability to manually generate a personal acces token ([GitLab's personal access token ](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html) ("PAT"); Jira Server and Mattermost also have the same affordances), should we support adding services via an access token? If we do, should we support adding services via oauth _and_ access tokens?

## Options

| option | nickname | oauth config location | token types |
| :-----: | :----- | :-----: | :-----: |
| 1 | `the full monty` | per team | both |
| 2 | `PAT overrides central oauth` | per org | both |
| 3 | `access token only` | n/a | personal access token |
| 4 | `centralized oauth` | per org | oauth |

[...]

PR #5805, opened 2021-12-17
Added IntegrationProvider, IntegrationToken tables, GitLab auth, migrated Mattermost to the new data format

BartoszJarocki:
This PR adds two new tables: `IntegrationProvider` and `IntegrationToken`, migrates existing Mattermost integration to a new type, and adds a `GitLab` auth integration. See more information [here](https://github.com/ParabolInc/parabol/pull/5805/files#diff-028fb6de06467163f1e86397b524a255daa3545a56d75d9ff9507cf935540e92R31).

Base PR for reference: https://github.com/ParabolInc/parabol/pull/5594
Base RFC: https://github.com/ParabolInc/parabol/issues/5567 

The main goal of this PR was to remove provider-specific columns from `IntegrationProvider` and `IntegrationToken` tables, fix the TS errors, and do some code cleanup.

Leftovers
- remove `authorizeOAuth2` helper method, as we continue to improve our integrations the differences between GitLab/GitHub/Atlassian OAuth2 grow
- Figure out a better table name for `IntegrationToken`, maybe `UserIntegration`? 

@jordanh's Loom about new tables: https://www.loom.com/share/22037d51a1404fa5aab5d6ba10c98778

Self-review, my first Loom so I'm sorry for being chaotic! https://www.loom.com/share/9435223bbeeb47da977343d65d6c95bc


jordanh:
>  Align client mutations and data types to the new structure (will we ever get GraphQL input union types? 😢)

I felt this too, so hard, when I was coding this set of features

packages/server/integrations/IntegrationServerManager.ts
```diff
@@ -1,102 +1,88 @@
...
+  refresh(refreshToken: string): Promise<OAuth2IntegrationTokenMetadata | Error>
+}
+
+export const isOAuth2AuthorizationManager = (
+  authorizationManager: AuthorizationManager
+): authorizationManager is OAuth2AuthorizationManager =>
+  authorizationManager.provider.tokenType === 'oauth2'
 
+/**
+ * Are webhooks ever require authorization?
```
jordanh, line 33: No, webhooks never require additional authorization

packages/server/integrations/gitlab/GitLabAuthorizationManager.ts
```diff
@@ -0,0 +1,69 @@
+import {
```
jordanh, line 1: I was hoping to create a more generic authorization method that could live as a mixin or within a super as I noticed our oauth2 authorization code was _very_ repetitive

packages/server/postgres/migrations/1636229261909_mattermostToIntegrationsTables.ts
```diff
@@ -121,13 +86,14 @@ export async function down() {
     CREATE INDEX IF NOT EXISTS "idx_MattermostAuth_teamId" ON "MattermostAuth"("teamId");
   `)
 
-  if (mattermostTokensWithProvider.length > 0) {
-    const mattermostAuthsToInsert = mattermostTokensWithProvider.map((row) =>
+  if (mattermostTokensWithProvider.rows.length > 0) {
```
jordanh, line 89: Nice catch

packages/server/postgres/migrations/1636229261909_mattermostToIntegrationsTables.ts
```diff
@@ -121,13 +86,14 @@ export async function down() {
...
+    const mattermostAuthsToInsert = mattermostTokensWithProvider.rows.map((row) =>
```
jordanh, line 90: Ditto that!

packages/server/integrations/gitlab/GitLabAuthorizationManager.ts
BartoszJarocki: This is still in the works

packages/server/postgres/migrations/1635354572171_addIntegrationProviderAndToken.ts
```diff
@@ -58,17 +55,13 @@ export async function up() {
     CREATE INDEX IF NOT EXISTS "idx_IntegrationProvider_orgId" ON "IntegrationProvider"("orgId");
     CREATE INDEX IF NOT EXISTS "idx_IntegrationProvider_teamId" ON "IntegrationProvider"("teamId");
     CREATE TABLE IF NOT EXISTS "IntegrationToken" (
```
BartoszJarocki, line 57: Should we rename it to UserIntegration? For example with webhook integration provider there will be no token but we'll still have an entry in this table, so `IntegrationToken` could be misleading

packages/server/postgres/migrations/1635354572171_addIntegrationProviderAndToken.ts
```diff
@@ -29,6 +29,8 @@ export async function up() {
     END IF;
     CREATE TABLE IF NOT EXISTS "IntegrationProvider" (
       "id" INT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+      "createdAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+      "updatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
       "type" "IntegrationProviderTypesEnum" NOT NULL,
```
BartoszJarocki, line 34: Should we call it a `service`? I think it makes more sense. Type feels more like `oauth2` or `webhook`

packages/server/postgres/migrations/1635354572171_addIntegrationProviderAndToken.ts
```diff
@@ -29,6 +29,8 @@ export async function up() {
...
       "tokenType" "IntegrationProviderTokenTypeEnum" NOT NULL,
```
BartoszJarocki, line 35: This should probably be just a `type` as not all providers are token based

packages/server/postgres/migrations/1636229261909_mattermostToIntegrationsTables.ts
```diff
@@ -122,39 +114,28 @@ export async function down() {
...
+            mattermostTokenWithProvider.createdAt,
+            mattermostTokenWithProvider.updatedAt,
+            mattermostTokenWithProvider.isActive,
+            (mattermostTokenWithProvider.provider.providerMetadata as any).webhookUrl,
+            mattermostTokenWithProvider.userId,
+            mattermostTokenWithProvider.teamId
+          ]
+        )
     )
     await Promise.all(mattermostAuthsToInsert)
```
BartoszJarocki, line 134: Add removing from the `IntegrationToken` table

packages/server/integrations/helpers/authorizeOAuth2.ts
```diff
@@ -0,0 +1,45 @@
...
+    Accept: 'application/json',
+    'Content-Type': 'application/json',
+    ...additonalHeaders
+  }
+
+  let uri = `${authUrl}`
+  if (queryParams) {
+    uri = uri.concat(`?${stringify(queryParams)}`)
+  }
+  const oauth2Response = await fetch(uri, {
```
BartoszJarocki, line 32: Wrap in try/catch

packages/server/integrations/helpers/authorizeOAuth2.ts
```diff
@@ -0,0 +1,45 @@
...
+type OAuth2Response = OAuth2Success | OAuth2Error
+
+type AuthorizeOAuth2Params = {
+  authUrl: string
+  queryParams?: Record<string, string>
+  body?: Record<string, string>
+  additonalHeaders?: Record<string, string>
+}
+
+export const authorizeOAuth2 = async ({
```
BartoszJarocki, line 16: Remove me

packages/server/postgres/migrations/1636229261909_mattermostToIntegrationsTables.ts
BartoszJarocki: done

packages/server/postgres/migrations/1635354572171_addIntegrationProviderAndToken.ts
BartoszJarocki: Provider feels like the right name

BartoszJarocki:
@Dschoordsch I'd really appreciate it if you could take a look, thank you!

packages/server/postgres/types/IntegrationTokenWithProvider.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export interface IntegrationTokenWithProvider extends IntegrationToken {
+  provider: IntegrationProvider
+}
+
+/**
+ * @param flatIntegrationTokenWithProvider raw flattened integration token with provider fromn database
+ * @returns {IntegrationTokenWithProvider}
+ */
+export const nestIntegrationProviderOnIntegrationToken = (
```
mattkrick, line 13: -1 keep em separated & it's so much cleaner

packages/server/types/modules.d.ts
```diff
@@ -12,8 +12,18 @@ declare module '*.graphql' {
 declare const __PRODUCTION__: string
 declare const __SOCKET_PORT__: string
 declare const __webpack_public_path__: string
+
 interface Window {
-  __ACTION__: any
+  __ACTION__: {
```
mattkrick, line 17: +1 i think this belongs in the client?

packages/server/utils/AtlassianServerManager.ts
```diff
@@ -1,8 +1,8 @@
 import fetch from 'node-fetch'
 import AtlassianManager from 'parabol-client/utils/AtlassianManager'
 import makeAppURL from 'parabol-client/utils/makeAppURL'
+import {authorizeOAuth2} from '../integrations/helpers/authorizeOAuth2'
 import appOrigin from '../appOrigin'
-import {OAuth2Error, OAuth2Success} from '../types/custom'
 
 interface AuthQueryParams {
   grant_type: 'authorization_code'
```
mattkrick, line 8: this is defined elsewhere, feels bad having the same defs floating around under different names

packages/server/utils/MattermostServerManager.ts
```diff
@@ -91,6 +92,29 @@ class MattermostServerManager extends MattermostManager {
   constructor(webhookUrl: string) {
     super(webhookUrl)
   }
+
+  static async getBestWebhook(userId: string, teamId: string, dataLoader: DataLoaderWorker) {
```
mattkrick, line 96: -1 opt to remove & put this logic in the datalaoder. dataloaders should use managers, but try to avoid managers using dataloaders

mattkrick (changes requested):
I'll take a crack at these comments over winter break, i know there are a lot!

packages/server/postgres/migrations/1635354572171_addIntegrationProviderAndToken.ts
```diff
@@ -0,0 +1,91 @@
...
+          WHEN "scope" = 'global' THEN TRUE
+          ELSE NULL
+        END
+      ) STORED,
+      "orgId" VARCHAR(100),
+      "teamId" VARCHAR(100),
+      "isActive" BOOLEAN DEFAULT TRUE NOT NULL,
+      "name" VARCHAR(250) NOT NULL,
+      "providerMetadata" JSONB NOT NULL DEFAULT '{}',
+      UNIQUE("scopeGlobal", "type"),
```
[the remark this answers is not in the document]
BartoszJarocki: Seems like you looked on outdated a bit PR, previously `type` represented provider name. I changed `type` to `provider` as we discussed, so considering this, this constraint means `we can have only one global provider with a given name

packages/server/graphql/types/AddIntegrationProviderInput.ts
```diff
@@ -0,0 +1,91 @@
...
+    },
+    serverBaseUrl: {
+      type: new GraphQLNonNull(GraphQLURLType),
+      description: 'The base URL used to access the provider'
+    },
+    clientId: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'The client id to give to the provider'
+    },
+    clientSecret: {
```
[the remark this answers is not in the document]
BartoszJarocki: A little explainer, this is an input type which in certain cases (like adding a global provider, which I'm not sure we even want) can be called only by super users.  `clientSecret` can't be fetched by the client. The field does not exist in the [graphql return type.](https://github.com/ParabolInc/parabol/pull/5805/files#diff-cee2a0044da3c70d7ca73e34603a4642f90a083602bc95d7a585ad290bfa7c82R52)

packages/server/graphql/types/AddIntegrationProviderInput.ts
```diff
@@ -0,0 +1,91 @@
...
+  fields: () => ({
+    orgId: {
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'The org that the access token is attached to'
+    },
+    teamId: {
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'The team that the token is linked to'
+    },
+    type: {
```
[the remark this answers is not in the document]
BartoszJarocki: whoops, I already changed the name to the `provider` which I think makes a little more sense. No strong feelings though!

packages/server/postgres/types/IntegrationProvider.ts
```diff
@@ -0,0 +1,185 @@
...
+ */
+export interface IntegrationProvider extends Omit<_IntegrationProvider, 'providerMetadata'> {
+  providerMetadata: IntegrationProviderMetadata
+}
+
+/**
+ * Metadata structure of the OAuth2 integration provider.
+ */
+export interface OAuth2IntegrationProviderMetadata {
+  scopes: string[]
```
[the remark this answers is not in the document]
BartoszJarocki: Good point about scope being possibly different depending on the use case. 
Any idea where we want to store it? Scope in all cases is used to obtain the token (with given permissions), so we either have to hardcode it, like we do in `AtlassianManager` 
```
export default abstract class AtlassianManager {
  abstract fetch: typeof fetch
  static SCOPE: JiraPermissionScope[] = [
    'read:jira-user',
    'read:jira-work',
    'write:jira-work',
    'offline_access'
  ]
```
or store it here, maybe with additional scopes defined in different metadata fields 🤔

packages/client/modules/teamDashboard/components/ProviderRow/GitLabConfigMenu.tsx
```diff
@@ -0,0 +1,37 @@
...
+  const {onError, onCompleted, submitMutation, submitting} = mutationProps
+  const atmosphere = useAtmosphere()
+
+  const removeGitLabAuth = () => {
+    if (submitting) return
+    submitMutation()
+    RemoveIntegrationTokenMutation(atmosphere, {providerId, teamId}, {onCompleted, onError})
+    // Our parent component does not unmount, and it often re-renders before the CSS menu transition
+    // can complete. We nuke the portal here to ensure the menu is closed.
+    terminatePortal()
```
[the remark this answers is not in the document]
mattkrick: we're removing the token here, but in mattermost we're removing the provider. i think this is right, but perhaps allowing anyone to remove the provider is too much

packages/server/graphql/types/GitLabIntegration.ts
```diff
@@ -0,0 +1,92 @@
...
+
+const OAuth2TokenMetadata = new GraphQLObjectType<any, GQLContext>({
+  name: 'OAuth2TokenMetadata',
+  description: 'OAuth2 token metadata for an Integration Provider',
+  fields: () => ({
+    accessToken: {
+      type: GraphQLString,
+      description: 'The access token'
+    },
+    refreshToken: {
```
mattkrick, line 23: refresh token should never be exposed, it's kept nice & secret in our DB

packages/server/graphql/types/GitLabIntegration.ts
```diff
@@ -0,0 +1,92 @@
...
+import GraphQLISO8601Type from './GraphQLISO8601Type'
+import IntegrationProvider from './IntegrationProvider'
+
+const OAuth2TokenMetadata = new GraphQLObjectType<any, GQLContext>({
+  name: 'OAuth2TokenMetadata',
+  description: 'OAuth2 token metadata for an Integration Provider',
+  fields: () => ({
+    accessToken: {
+      type: GraphQLString,
+      description: 'The access token'
```
mattkrick, line 21: viewer only

packages/server/postgres/migrations/1635354572171_addIntegrationProviderAndToken.ts
mattkrick: you've got a point, i just think of a webhook URL as a token & then I can sleep at night 😅

packages/server/postgres/migrations/1635354572171_addIntegrationProviderAndToken.ts
```diff
@@ -0,0 +1,91 @@
...
+      CONSTRAINT global_provider_must_be_oauth2 CHECK (
+        "scopeGlobal" IS NULL OR ("scopeGlobal" = TRUE AND "tokenType" = 'oauth2')
+      )
+    );
+    CREATE INDEX IF NOT EXISTS "idx_IntegrationProvider_typeAndScope"
```
mattkrick: i couldn't find a use for this, given it is coming from a user we'd have their token, which means we'd have the providerId, so we shouldn't need this

packages/server/graphql/types/AddIntegrationProviderInput.ts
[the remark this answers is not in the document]
mattkrick: oh you're absolutely right!

Dschoordsch (commented):
I'm not sure how this change relates to #5829 so I just have a few comments for now.

packages/client/shared/gqlIds/GitLabIntegrationId.ts
```diff
@@ -0,0 +1,9 @@
+const GitLabIntegrationId = {
+  join: (teamId: string, userId: string) => `gli:${teamId}:${userId}`,
```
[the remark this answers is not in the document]
jordanh: I agree with you sir. It would be nicer not to be so restrictive

packages/client/modules/teamDashboard/components/ProviderRow/GitLabProviderRow.tsx
```diff
@@ -0,0 +1,223 @@
...
+          integrations {
+            gitlab {
+              availableProviders {
+                id
+                scope
+                tokenType
+                name
+                updatedAt
+                providerMetadata {
+                  ... on OAuth2ProviderMetadata {
```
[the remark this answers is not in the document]
mattkrick: You're asking _really_ good questions!
Enrique & I talked about this over break. I logged the takeaway conversation here: https://github.com/ParabolInc/parabol/issues/5567#issuecomment-999081074

tl;dr 1 team member can have 0 or 1 `type`. i can add `oauth2` and you could add `pat`, but then I couldn't work with you. i _cannot_ add both oauth2 and pat. i _cannot_ add a self-hosted and cloud offering for any single team.

packages/server/postgres/types/IntegrationProvider.ts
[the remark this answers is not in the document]
mattkrick: i'm cool hardcoding that stuff into the app, that way we can be pretty expressive in what scope is used for what functionality & maintaining them should be easier than having that logic decoupled in a DB

packages/client/mutations/AddIntegrationTokenMutation.ts
```diff
@@ -0,0 +1,53 @@
...
+    }
+  }
+`
+
+const mutation = graphql`
+  mutation AddIntegrationTokenMutation(
+    $providerId: ID!
+    $oauthCodeOrPat: ID!
+    $teamId: ID!
+    $redirectUri: URL!
```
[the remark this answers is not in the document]
mattkrick: ah! you're absolutely right, i honestly don't know what i was thinking here...

packages/server/graphql/mutations/addIntegrationToken.ts
```diff
@@ -0,0 +1,156 @@
...
+    // RESOLUTION
+    await upsertIntegrationToken({
+      providerId: providerDbId,
+      teamId,
+      userId: viewerId,
+      tokenMetadata: {...tokenMetadata}
+    })
+
+    const data = {userId: viewerId, teamId}
+    publish(SubscriptionChannel.TEAM, teamId, 'AddIntegrationToken', data, subOptions)
```
mattkrick, line 151: NOTIFICATION channel, not team. team doesn't get to know about 1 persons token

packages/server/postgres/migrations/1635354572171_addIntegrationProviderAndToken.ts
```diff
@@ -0,0 +1,91 @@
...
+        CASE
+          WHEN "scope" = 'global' THEN TRUE
+          ELSE NULL
+        END
+      ) STORED,
+      "orgId" VARCHAR(100),
+      "teamId" VARCHAR(100),
+      "isActive" BOOLEAN DEFAULT TRUE NOT NULL,
+      "name" VARCHAR(250) NOT NULL,
+      "providerMetadata" JSONB NOT NULL DEFAULT '{}',
```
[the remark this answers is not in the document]
mattkrick: i think that was chosen to differentiate between tokenMetadata & providerMetadata (it gets confusing when you're in the thick of it, especially because the types weren't originally very strict).

IMO i think we should just use columns, then we get types for free!

packages/server/graphql/mutations/addIntegrationToken.ts
```diff
@@ -0,0 +1,156 @@
...
+) => {
+  const {providerMetadata} = provider
+  if (isOAuth2ProviderMetadata(providerMetadata)) {
+    return createOAuth2TokenMetadata(provider, oauthCodeOrPat, redirectUri, info, context)
+  }
+
+  return null
+}
+
+const addIntegrationToken = {
```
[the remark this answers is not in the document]
mattkrick: i agree `token` referring to an entire DB row feels bad.
technically, it isn't related to the User, but the TeamMember.
maybe IntegrationTeamMemberAuth? IntegrationAuth? IntegrationAuthStrategy?

packages/client/shared/gqlIds/GitLabIntegrationId.ts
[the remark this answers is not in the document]
mattkrick: updated to `${service}:....`

BartoszJarocki:
closing in favor of #5829

PR #5815, opened 2021-12-20
monkeypatch uWS res.cork

mattkrick:
this guarantees that we no longer attempt to access responses that have already been disposed of.
it'll still emit a console log, but we can turn those off later & it'd be AOK.

instead of turning them off, i'm thinking we track both `done` and `aborted` booleans & only log if it is done but not aborted. that way we'll still get log statements when the handler isn't doing it right

### DIFF
--- a/packages/server/pipeStreamOverResponse.ts
+++ b/packages/server/pipeStreamOverResponse.ts
@@ -8,11 +8,11 @@ const pipeStreamOverResponse = (
 ) => {
   res.onAborted(() => {
     readStream.destroy()
-    res.done = true
+    res.aborted = true
   })
   readStream
     .on('data', (chunk: Buffer) => {
-      if (res.done) {
+      if (res.aborted) {
         readStream.destroy()
         return
       }
@@ -40,7 +40,7 @@ const pipeStreamOverResponse = (
     })
 
     .on('error', () => {
-      if (!res.done) {
+      if (!res.aborted) {
         res.writeStatus('500').end()
       }
       readStream.destroy()
--- a/packages/server/safetyPatchRes.ts
+++ b/packages/server/safetyPatchRes.ts
@@ -37,6 +37,15 @@ const safetyPatchRes = (res: HttpResponse) => {
     return res._close()
   }
 
+  res._cork = res.cork
+  res.cork = (cb: () => void) => {
+    if (res.done) {
+      console.log(`uWS DEBUG: Called cork after done`)
+      return
+    }
+    res._cork(cb)
+  }
+
   res._tryEnd = res.tryEnd
   res.tryEnd = (fullBodyOrChunk: RecognizedString, totalSize: number) => {
     if (res.done) {
--- a/packages/server/safetyPatchRes.ts
+++ b/packages/server/safetyPatchRes.ts
@@ -5,7 +5,7 @@ const safetyPatchRes = (res: HttpResponse) => {
     throw new Error('already patched')
   }
   res.onAborted(() => {
-    res.done = true
+    res.aborted = true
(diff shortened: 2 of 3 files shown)


packages/server/safetyPatchRes.ts
```diff
@@ -37,6 +37,15 @@ const safetyPatchRes = (res: HttpResponse) => {
...
 
+  res._cork = res.cork
+  res.cork = (cb: () => void) => {
+    if (res.done) {
+      console.log(`uWS DEBUG: Called cork after done`)
+      return
+    }
+    res._cork(cb)
+  }
+
```
Dschoordsch, line 48: -1 all methods from `HttpResponse` interface (including getters) potentially throw this error
https://github.com/uNetworking/uWebSockets.js/blob/master/src/HttpResponseWrapper.h#L26-L34
Of the missing ones we at lease use `upgrade` and `getRemoteAddressAsText`.

Dschoordsch (changes requested):

packages/server/safetyPatchRes.ts
mattkrick: great catch! i forgot we call those after an `await` will add.
while i'm in here, i'm gonna break done between aborted & done so we get more meaningful errors

packages/server/pipeStreamOverResponse.ts
```diff
@@ -8,11 +8,11 @@ const pipeStreamOverResponse = (
 ) => {
   res.onAborted(() => {
     readStream.destroy()
-    res.done = true
+    res.aborted = true
```
mattkrick, line 11: rename wasn't necessary, but i did it so `aborted` means the same thing throughout the app

packages/server/safetyPatchRes.ts
```diff
@@ -5,7 +5,7 @@ const safetyPatchRes = (res: HttpResponse) => {
     throw new Error('already patched')
   }
   res.onAborted(() => {
-    res.done = true
+    res.aborted = true
```
mattkrick, line 8: we now have a separate flag for when the client aborts the connection (socket hangup, et al).
this is useful because now if we see a warning in the console we know it wasn't from a client abort.

we use this safetyPatchRes file so we don't have to worry about client aborts. otherwise we'd have to check for an abort after every single async call (see handleUpgrade.ts) 

However, we shouldn't be calling a terminating event after res.done is true. That's a bug in our own business logic

packages/server/safetyPatchRes.ts
```diff
@@ -20,58 +20,85 @@ const safetyPatchRes = (res: HttpResponse) => {
...
     if (res.done) {
-      console.error(`uWS DEBUG: Called writeStatus after done ${status}`)
-      return res
+      console.error(`uWS: Called writeStatus after done ${status}`)
     }
+    if (res.done || res.aborted) return res
     return res._writeStatus(status)
   }
+
+  res._upgrade = res.upgrade
```
mattkrick, line 85: added the remaining used methods. thanks!

packages/server/socketHandlers/handleUpgrade.ts
```diff
@@ -21,7 +20,7 @@ const handleUpgrade: WebSocketBehavior['upgrade'] = async (res, req, context) =>
     return
   }
   res.onAborted(() => {
-    aborted.done = true
+    res.aborted = true
```
mattkrick, line 23: keeping aborted on the res like in the other files

mattkrick:
@Dschoordsch would you mind giving this another look?

Dschoordsch (approved):

PR #5829, opened 2022-01-04
Add GitLab

mattkrick:
This supersedes #5805 

Burndown:
- [x] shared provider query
- [x] remove updateIntegrationProvider ifo upsert...
- [x] flatten tokenMetadata, providerMetadata so it only appears to/from database
- [x] move subscription channel to NOTIFICATION
- [x] rename IntegrationToken to IntegrationAuth (maybe?, OOS for this PR)
- [x] rename type to authStrategy (OOS for this PR)
- [ ] create 1 dataloader per service so the return type is strict (OOS for this PR)
- [ ] archiving a team should move the integration to another team so others can keep using it (or remove it, since you must be on the team that created it in order to remove it)
- [ ] AtlassianServerManager uses the clientId, clientSecret from the DB, not the env (OOS)


mattkrick (commented):
@BartoszJarocki would you mind taking a look at this?

For testing, could you set up a gitlab app (if gitlab only allows 1 redirect url, then set up 4 apps: local/dev/staging/prod). Also, please give the mattermost integration a go & make sure i didn't break anything 😛 

could you also look into setting up a gitlab enterprise for developers, if they offer it? i'd like to test adding gitlab enterprise (although the front-end for that is out of scope, but we'll need to do that after we merge this PR)

BartoszJarocki:
Matt, I’ve run the PR, migrations are running smoothly, no problems there.

I’ve found a few issues related to integrations, though. 

1. If you try to add Mattermost (or anything that uses `addIntegrationProvider` mutation) you’ll get an error: `Provided 0 metadata types, expected 1`, it’s because there’s OR instead of AND used in the condition below
```
   if (!oAuth2ProviderMetadataInput || !webhookProviderMetadataInput) {
      return {error: {message: 'Provided 0 metadata types, expected 1'}}
   }
```

2. After fixing the above, to add Mattermost I had to add the `webhookProviderMetadataInput` object to the `AddIntegrationProviderMutation` input
```
      UpdateIntegrationProviderMutation(
        atmosphere,
        {
          provider: {
            id: activeProvider.id,
            scope: 'team',
            webhookProviderMetadataInput: {
              webhookUrl
            }
          }
        },
        {onError, onCompleted}
      )
```
on the frontend part, otherwise, the mutation will fail.

3. After fixing the above, the mutation runs successfully, and as a result, the DB entry is added to the `IntrgrationProvider` table. The UI doesn’t show active Mattermost integration though. From what I see, we’re trying to fetch the Mattermost data from `auth` which doesn’t exist (we’re not adding anything to `TeamMemberIntegrationAuth` related to the Mattermost), so frontend does not see anything there. I'm not sure what's the expected behavior here, whether you want to keep some empty entries in the `TeamMemberIntegrationAuth` table (then, `addIntegrationProvider` mutation will also have to add `TeamMemberIntegrationAuth` entry for webhooks, or frontend will have to call another mutation in sequence to do so) or change the way we fetch the Mattermost integration data so I'll leave it to you. 


Regarding GitLab enterprise, did you mean `self-hosted`? And sure, I'll prepare some apps for us to use.

packages/server/graphql/mutations/addIntegrationProvider.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (!isTeamMember(authToken, teamId)) {
+      return {error: {message: 'Must be on the team that created the provider'}}
+    }
+
+    // VALIDATION
+    const {oAuth2ProviderMetadataInput, webhookProviderMetadataInput, ...rest} = input
+    if (oAuth2ProviderMetadataInput && webhookProviderMetadataInput) {
+      return {error: {message: 'Provided 2 metadata types, expected 1'}}
+    }
+    if (!oAuth2ProviderMetadataInput || !webhookProviderMetadataInput) {
```
BartoszJarocki: -2, Replace `||` with `&&`, otherwise, it'll always fail

packages/client/modules/teamDashboard/components/ProviderRow/MattermostPanel.tsx
```diff
@@ -104,18 +113,43 @@ const MattermostPanel = (props: Props) => {
...
+      )
+    } else {
+      AddIntegrationProviderMutation(
+        atmosphere,
+        {
+          input: {
+            scope: 'team',
+            service: 'mattermost',
+            teamId,
+            authStrategy: 'webhook'
```
BartoszJarocki, line 147: -2, we're missing the following here
```
   webhookProviderMetadataInput: {
      webhookUrl
   }
```

packages/client/modules/teamDashboard/components/ProviderRow/MattermostPanel.tsx
```diff
@@ -68,8 +69,12 @@ const MattermostPanel = (props: Props) => {
...
           integrations {
             mattermost {
-              isActive
-              webhookUrl
+              auth {
+                provider {
+                  id
+                  webhookUrl
+                }
+              }
```
BartoszJarocki, line 77: -2, there's no `auth` for webhook providers added currently, this will always be empty currently. as a result, UI won't show active integration

BartoszJarocki (commented):
Added a few comments so it's easier to find issues I've mentioned

packages/server/graphql/mutations/addIntegrationProvider.ts
BartoszJarocki: That was fast ;)

mattkrick:
@BartoszJarocki realized i didn't include my last self-review, just affects a couple files, but 1 is the mattermost migration so you might have to `yarn pg:migrate down && yarn pg:migrate up` again

Dschoordsch (changes requested):

packages/server/postgres/migrations/1639354572171_addIntegrationProviderAndToken.ts
```diff
@@ -0,0 +1,103 @@
...
+        'gitlab',
+        'mattermost'
+      );
+    END IF;
+    IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'IntegrationProviderAuthStrategyEnum') THEN
+      CREATE TYPE "IntegrationProviderAuthStrategyEnum" AS ENUM (
+        'pat',
+        'oauth2',
+        'webhook'
+      );
```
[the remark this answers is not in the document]
mattkrick: it's true, removing individual values from enums is tricky, but not impossible. However, since migrations are immutable (or at least we treat them like that once they hit production), that means the up-migration will ALWAYS get called eventually, so leaving the enum value in the down-migration will be OK.

That said, if we did have to remove a value in an up-migration (e.g. we stop supporting `pat`) you can do it in 3 lines:

```sql
ALTER TYPE "IntegrationProviderAuthStrategyEnum" RENAME TO "IntegrationProviderAuthStrategyEnumOLD";
CREATE TYPE "IntegrationProviderAuthStrategyEnum" AS ENUM('oauth2', 'webhook');
ALTER TABLE "IntegrationProvider" ALTER COLUMN "authStrategy" TYPE "IntegrationProviderAuthStrategyEnum" USING "authStrategy"::text::"IntegrationProviderAuthStrategyEnum";
```

The good parts of keeping an enum is `yarn pg:build` will create a typescript string union for us, which is really great. It also only takes up 4 bytes, which is nice, but not critical. Definitely not the final word though. Open for debate!

mattkrick:
@BartoszJarocki could you give this another pass? I fixed a glaring security bug that I had (since our IDs are not globally unique it'd be possible to guess the providerId of a webhook provider & create your own auth record to it) & now when you add a webhook provider it follows up by adding a teamMemberIntegrationAuth record for that user

packages/server/postgres/migrations/1639354572171_addIntegrationProviderAndToken.ts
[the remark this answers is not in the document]
Dschoordsch: I should have elaborated more. I did exactly what you described to modify an enum. In our case we would also need to drop and recreate the `global_provider_must_be_oauth2` otherwise we cannot simply alter the column type. Simple enough now with only 1 constraint, but going forward we want to add more which makes modifying these types more error prone because we have to find all constraints and copy and paste them.
Losing the type from pgtyped is a real bummer though.
So it's basically about which part we make more difficult: migrations or queries. Queries we write much more often but migrations are more critical. On production we run down migrations only in emergencies and then it's far easier to miss something while double checking. Bad typescript types are caught during build.
So while I really like the convenience of the generated type, I'm not sure we're strict enough with testing migrations to justify the risk. I don't think we test down migrations thorough enough and merge order could also break them.
I leave it up to you since I don't know how often we needed to roll back in the past.

BartoszJarocki:
@mattkrick I've tested and everything works correctly now 🎉  I'll do another round of CR, but so far everything looks good!

packages/server/postgres/migrations/1639354572171_addIntegrationProviderAndToken.ts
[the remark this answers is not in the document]
Dschoordsch: In the spirit of recent learnings, just stick with what you have now.

packages/server/graphql/types/GitLabIntegration.ts
```diff
@@ -0,0 +1,51 @@
...
+        const orgTeams = (await dataLoader.get('teamsByOrgIds').loadMany(orgIds))
+          .flat()
+          .filter(isValid)
+        const orgTeamIds = orgTeams.map(({teamId}) => teamId)
+        return dataLoader
+          .get('sharedIntegrationProviders')
+          .load({service: 'gitlab', orgTeamIds, teamIds})
+      }
+    }
+    // The GitLab schema get injected here as 'api'
```
BartoszJarocki, line 48: is this valid? I don't see anything related

packages/server/postgres/migrations/1640229261909_mattermostToIntegrationsTables.ts
```diff
@@ -0,0 +1,73 @@
...
+  })
+  await Promise.all(mattermostAuthsToInsert)
+  await client.end()
+  await r.getPoolMaster().drain()
+}
+
+export async function down() {
+  // This down function was broken
+  // To save time, I removed dropping MattermostAuth table from the up migration
+  // We can drop that table in a later migration
```
BartoszJarocki, line 72: Do we want to create a separate task for that?

BartoszJarocki (approved):
Just a couple of minor comments, let's 🚢 it!

packages/server/graphql/types/GitLabIntegration.ts
mattkrick: rootSchema is where the magic happens.
Can always check `./schema.graphql` for the truth!

packages/server/postgres/migrations/1640229261909_mattermostToIntegrationsTables.ts
mattkrick: i don't have tension around a random table floating around in PG (we've got a couple already 😋 ) but if you do don't let me stop ya!

mattkrick:
gonna merge this just so i can unblock Georg's work on Jira

packages/server/postgres/migrations/1639354572171_addIntegrationProviderAndToken.ts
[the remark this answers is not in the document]
mattkrick: ahhh the other constraints (and other columns in other tables) is definitely a pain! 
perhaps we just leave enum values there & never remove them if they get too painful? or rename them to something like `pat_DO_NOT_USE_OR_YOU_WILL_BE_FIRED`.  

to this day we still haven't used a down migration in production 😅 . that's not to say we've had perfect up-migrations, but we typically hotfix by adding an another up-migration.

PR #5840, opened 2022-01-05
enforce server side lock

### LINKED ISSUE #5267: Check Team.isPaid before starting a meeting
We have something that people want so badly they're willing to get around our silly modals :tada: 
@gcrickman 

AC
- [ ] in every startMeeting server mutation, check the `isPaid` boolean and `lockMessageHTML`
- [ ] if `lockMessageHTML` is present, then they're naughty hackers & the error message should say something cheeky
- [ ] if lockMessageHTML is not present & isPaid is false, then they should get a boring error message

Estimate: 4 hours

tiffanyhan:
Resolve #5267

**AC:**
- [ ] if `team.isPaid` is false and `team.lockMessageHTML` is null and you try to start any meeting, you get the error msg "Team is not paid"
- [ ] if `team.isPaid` is false and `team.lockMessageHTML` is non-null and you try to start any meeting, you get the error msg "Wow, you're determined to use Parabol! That's awesome! Do you want to keep sneaking over the gate, or walk through the door with our Sales team?"

### DIFF
--- a/packages/server/graphql/mutations/startCheckIn.ts
+++ b/packages/server/graphql/mutations/startCheckIn.ts
@@ -5,8 +5,9 @@ import ActionMeetingMember from '../../database/types/ActionMeetingMember'
 import {MeetingTypeEnum} from '../../database/types/Meeting'
 import MeetingAction from '../../database/types/MeetingAction'
 import generateUID from '../../generateUID'
+import getTeamsByIds from '../../postgres/queries/getTeamsByIds'
 import updateTeamByTeamId from '../../postgres/queries/updateTeamByTeamId'
-import {getUserId, isTeamMember} from '../../utils/authorization'
+import {getUserId, isLocked, isPaid, isTeamMember} from '../../utils/authorization'
 import publish from '../../utils/publish'
 import standardError from '../../utils/standardError'
 import {GQLContext} from '../graphql'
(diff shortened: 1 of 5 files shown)


tiffanyhan:
@nickoferrall mind giving this one a review?

packages/server/graphql/mutations/startCheckIn.ts
```diff
@@ -38,6 +39,14 @@ export default {
     if (!isTeamMember(authToken, teamId)) {
       return standardError(new Error('Team not found'), {userId: viewerId})
     }
+    const teams = await getTeamsByIds([teamId])
+    const team = teams[0]!
+    if (!isPaid(team)) {
+      const errMsg = isLocked(team)
+        ? "Wow, you're determined to use Parabol! That's awesome! Do you want to keep sneaking over the gate, or walk through the door with our Sales team?"
+        : 'Team is not paid'
```
nickoferrall, line 47: +1 I think we could provide a little more info here. Perhaps instead of `Team is not paid` we could say something like `Unable to start a meeting as your team has an overdue payment`?

packages/server/graphql/mutations/startCheckIn.ts
```diff
@@ -38,6 +39,14 @@ export default {
...
+      return standardError(new Error(errMsg), {userId: viewerId})
```
nickoferrall, line 48: It's not in the AC for the issue so feel free to ignore, but I wonder whether we should track when this happens? We could send an event to Segment with the `isLocked` status so the Growth team can find a list of mischievous users.

nickoferrall (approved):
LGTM!

packages/server/graphql/mutations/startCheckIn.ts
tiffanyhan: great idea, updated it to: 'Sorry! We are unable to start your meeting because your team has an overdue payment'

packages/server/graphql/mutations/startCheckIn.ts
tiffanyhan: great idea, updated it to: 'Sorry! We are unable to start your meeting because your team has an overdue payment'

tiffanyhan:
@nickoferrall i think the PR for this is probably good to merge as-is, but can I pass this issue to you? The only thing left to do is make changes if any are found to be needed in maintainer review step.

Issue #5340, opened 2021-08-31
Create primary/foreign key loaders pattern for postgres


mattkrick:
Now that we're using pg, we need to create some good patterns for dataloaders.

AC
- [ ] there's a new file called pgLoaders that works like customLoaders except it doesn't take this.dataLoaderoptions, it's just the `batchFn`
- [ ] there's a pgPrimaryLoader that takes in a pgLoader
- [ ] there's a pgForeignLoaders file that works like fkLoader that
- [ ] a pgLoader can be a foreign key loader, too. it can optionally take in primary key loader which will clear & prime the optional primary key loader 

Estimate: 4 hours

jordanh:
@mattkrick If this is still a thing, does the above AC need to change at all given recent changes?

Dschoordsch:
@mattkrick what's the goal of this task? Do we want to be able to have loader makers like
```
export const users = pgPrimaryLoaderMaker(getUsersByIdsQuery)
```
and
```
export const usersByEmails = pgForeignLoaderMaker('users', 'emails', getUsersByEmailsQuery)
```
?

PR #5847, opened 2022-01-07
Add primary and foreign key loaders

Dschoordsch:
Primary and foreign key loaders intended for postgres. The loaders have no direct dependency to postgres though, since we want to take advantage of the generated types.

The primaryKeyLoader return `T | undefined`
If no item exists with the key, undefined is returned. This was already
done for some loaders (e.g. Users).
If the id is an input of the resolver, the call site must check the
return, otherwise we assume database integrety and assert the value.
ForeignKeyDataLoader is not affected since it returns arrays for each
key, which results in just an empty array.

Resolves: #5340


Dschoordsch:
@BartoszJarocki could you review this change, please?

packages/client/types/generics.ts
```diff
@@ -64,6 +64,7 @@ export type UnshiftToTuple<V, T extends any[]> = Parameters<(a: V, ...args: T) =
 export type Unpromise<T> = T extends Promise<infer U> ? U : T
 export type Unref<T> = T extends MutableRefObject<infer U> ? U : T
 export type RefCallbackInstance<T extends HTMLElement = HTMLElement> = T | null
+export type Unarray<T> = T extends ArrayLike<infer U> ? U : T
```
BartoszJarocki, line 67: +2 wow, I didn't know it's possible!

packages/server/dataloader/foreignKeyLoaderMaker.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * Used to register loaders for types by foreign key.
+ *
+ * This is intended for types which already have a {@link primaryKeyLoaderMaker} defined.
+ * When an item is loaded via this loader, the primary loader will be primed with the result as well.
+ * It reflects a one to many relationship, i.e. for each key passed, an array will be returned.
+ */
+export function foreignKeyLoaderMaker<
+  LoaderName extends LoaderKeys,
```
BartoszJarocki, line 24: +2, it's great to see more verbose generic keys, way easier to read

packages/server/graphql/mutations/helpers/notifications/notifySlack.ts
```diff
@@ -112,7 +112,7 @@ export const startSlackMeeting = async (
   const title = 'Meeting started :wave: '
   const blocks = [
     makeSection(title),
-    makeSections([`*Team:*\n${team.name}`, `*Meeting:*\n${meeting.name}`]),
+    makeSections([`*Team:*\n${team?.name}`, `*Meeting:*\n${meeting.name}`]),
```
BartoszJarocki, line 115: -1 If a team is `undefined` this will result in showing *Team*: undefined. Not a big deal, but IMO it's better to not send the notification if something went wrong than to show `undefined` to the users. I'd add a check to make sure the team is defined

BartoszJarocki (changes requested):
Great work! I tested and everything works. I just left one little comment about possible `undefined` in a string, otherwise, it's good to go!

BartoszJarocki (approved):

packages/server/graphql/mutations/acceptTeamInvitation.ts
```diff
@@ -74,7 +74,7 @@ export default {
       const {meetingId, teamId} = invitation
       const meeting = meetingId ? await dataLoader.get('newMeetings').load(meetingId) : null
       const activeMeetingId = meeting && !meeting.endedAt ? meetingId : null
-      const team = await dataLoader.get('teams').load(teamId)
+      const team = (await dataLoader.get('teams').load(teamId))!
```
mattkrick, line 77: +1 this is correct, but i sure wish there were another way we could do this.
could we override the `load` method coming from DataLoader with a generic so writing `load<true>(teamId)` returns a `Promise<Omit<V, null>>`?

true is just a dummy, we could use any value, even a string.

packages/server/utils/predicates.ts
```diff
@@ -1,2 +1,5 @@
 export const isNotNull = <T>(item: T | null | undefined): item is T =>
   item !== null && item !== undefined
+
+export const isNeitherNullNorError = <T>(item: T | Error | null | undefined): item is T =>
```
mattkrick, line 4: -1 no preference on which helper we choose, but this functionality already exists here: https://github.com/ParabolInc/parabol/blob/9c27f5716901081c321d69ea6f56fb4756a83c1d/packages/server/graphql/isValid.ts

would like to only have 1 instance of this in the codebase

mattkrick (changes requested):
fantastic pattern!
2 small things, nothing breaking, just want to make sure we don't duplicate helper methods.
if it interests you, it'd be fun to see if we could re-declare the DataLoader type & give `load` a generic so we have a way to coerce to non-null without the `!`, but perhaps that isn't cleaner after all? up to you!

mattkrick:
you know, adding an extra method is a lot cleaner than my dumb idea. great work!
Just a couple merge conflicts, ready to merge after that

mattkrick (approved):

Dschoordsch:
> you know, adding an extra method is a lot cleaner

It's mostly much easier, which is always a good reason

Issue #5823, opened 2021-12-21
add eslint rule: disable `!=`


mattkrick:
we like things strict. let's make `!=` an error to force `!==`

AC
- [ ] eslint rule added & all offenders fixed

Estimate: 2 hours

tiffanyhan:
@tianrunhe can you take a look at this one please?

tiffanyhan:
@igorlesnenko as this is my last day, may i pass this issue to you, or you can reassign it if your plate is full? 🙏

PR #5848, opened 2022-01-07
add strict eq eslint rule

tiffanyhan:
Resolve: #5823

**AC**
- [ ] strict eq enforced by linter


tiffanyhan:
@tianrunhe can you take a look at this one please?

tianrunhe (changes requested):
Looks like it only applies to `/server`, not `/client`? e.g., [here](https://github.com/ParabolInc/parabol/blob/strict-eq/packages/client/utils/AtlassianManager.ts#L382) it's still just `==`.

tiffanyhan:
> Looks like it only applies to `/server`, not `/client`? e.g., [here](https://github.com/ParabolInc/parabol/blob/strict-eq/packages/client/utils/AtlassianManager.ts#L382) it's still just `==`.

ah thanks for catching! looks like we have a number of eslint problems that are causing the command to exit early. i'll triage

packages/client/.eslintrc.js
```diff
@@ -1,9 +1,9 @@
 module.exports = {
   plugins: ['@typescript-eslint', 'prettier', 'emotion', 'react', 'react-hooks'],
   extends: ['plugin:react/recommended', '../../.eslintrc.js', 'plugin:prettier/recommended'],
-  plugins: ['@typescript-eslint', 'prettier'],
+  // plugins: ['@typescript-eslint', 'prettier'], // revert a4963401
   parserOptions: {
```
tiffanyhan, line 5: look like https://github.com/ParabolInc/parabol/commit/a4963401 broke the client lint command. i get lots of `Definition for rule was not found` errors. reverting to the earlier plugins fixes the issue, but it means we're no longer supporting eslint v7+.

packages/client/.eslintignore
```diff
@@ -0,0 +1,3 @@
+__generated__/*
+types/graphql.ts
+serviceWorker/*
```
tiffanyhan, line 3: `.eslintignore` in root of parabol project dir was not working, so i broke it out into separate client and server files and put them in the root of their respective packages.

packages/server/.eslintignore
```diff
@@ -0,0 +1,5 @@
+database
+postgres/queries/generated
+segmentFns/*
+types/githubTypes.ts
+postgres/migrationTemplate.ts
```
tiffanyhan, line 5: `.eslintignore` in root of parabol project dir was not working, so i broke it out into separate client and server files and put them in the root of their respective packages.

packages/client/tsconfig.eslint.json
```diff
@@ -0,0 +1,14 @@
...
+    "validation/*",
+    "utils/*",
+    "types/*",
+    "styles/*",
+    "stories/*",
+    "components/*",
+    "mutations/*",
+    "hooks/*"
+  ]
+}
```
tiffanyhan, line 14: was getting tons of `The file does not match your project config` errors. this file fixes it.

packages/client/components/StandardHubUserMenu.tsx
```diff
@@ -30,7 +30,7 @@ const TallMenu = styled(Menu)({
   maxHeight: 256
 })
 
-interface Props extends RouteComponentProps<{}> {
+interface Props extends RouteComponentProps<{[x: string]: string | undefined}> {
```
tiffanyhan, line 33: went ahead and fixed "Don't use `{}` as a type" errors

packages/client/hooks/useAutoCheckIn.ts
```diff
@@ -37,7 +37,7 @@ const useAutoCheckIn = (meetingRef: any) => {
       JoinMeetingMutation(
         atmosphere,
         {meetingId},
-        {onCompleted: subscribeToMeeting, onError: () => { }}
+        {onCompleted: subscribeToMeeting, onError: () => undefined}
```
tiffanyhan, line 40: went ahead and fixed no-empty-function errors

packages/client/.eslintrc.js
tiffanyhan: if this change is accepted, we should create a separate issue for supporting eslint v7+ in client package?

.eslintrc.js
```diff
@@ -7,7 +7,8 @@ module.exports = {
     sourceType: 'module'
   },
   rules: {
-    '@typescript-eslint/ban-ts-ignore': 'off',
+    eqeqeq: 'error',
+    '@typescript-eslint/ban-ts-comment': 'off',
```
tiffanyhan, line 11: ["ban-ts-comment" is the correct rule name](https://github.com/typescript-eslint/typescript-eslint/blob/main/CHANGELOG.md#2180-2020-01-27) for later versions of `typescript-eslint`

tiffanyhan:
Good news: we can run `yarn lint` successfully now, without the command exiting early bc of problems with client linting.

Bad news: there are still a few (2-3 instances) of loose equality that's not being detected by eslint. E.g. [movePokerTemplateScaleValue.ts](https://github.com/ParabolInc/parabol/blob/strict-eq/packages/server/graphql/mutations/movePokerTemplateScaleValue.ts#L54), [Bug.ts](https://github.com/ParabolInc/parabol/blob/strict-eq/packages/client/utils/screenBugs/Bug.ts#L77), [PhaseItemEditor](https://github.com/ParabolInc/parabol/blob/strict-eq/packages/client/components/RetroReflectPhase/PhaseItemEditor.tsx#L89). I'm pretty certain these files should be in the dependency graph and detected by eslint based on [client/tsconfig.esilnt.json](https://github.com/ParabolInc/parabol/blob/strict-eq/packages/client/tsconfig.eslint.json#L0-L1). I don't know why the files aren't being picked up. I tested adding other errors in the files that eslint should complain about, with the same result. So it's not detecting the file at all.

Since these files not being detected is a general eslinting problem and not specific to adding the strict eq rule, I propose we should create a separate issue for it so it needn't block this PR.

packages/client/.eslintrc.js
```diff
@@ -1,9 +1,9 @@
 module.exports = {
   plugins: ['@typescript-eslint', 'prettier', 'emotion', 'react', 'react-hooks'],
   extends: ['plugin:react/recommended', '../../.eslintrc.js', 'plugin:prettier/recommended'],
-  plugins: ['@typescript-eslint', 'prettier'],
+  // plugins: ['@typescript-eslint', 'prettier'], // revert a4963401
```
tiffanyhan, line 4: look like a496340 broke the client lint command. i get lots of Definition for rule was not found errors. reverting to the earlier plugins fixes the issue, but i suppose it means we’re no longer supporting eslint v7+.

if this change is accepted, we should create a separate issue for supporting eslint v7+ in client package?

tiffanyhan:
Good news: we can run `yarn lint` successfully now, without the command exiting early bc of problems with client linting.

Bad news: there are still a few (2-3 instances) of loose equality that’s not being detected by eslint. E.g. [movePokerTemplateScaleValue.ts](https://github.com/ParabolInc/parabol/blob/strict-eq/packages/server/graphql/mutations/movePokerTemplateScaleValue.ts#L54), [Bug.ts](https://github.com/ParabolInc/parabol/blob/strict-eq/packages/client/utils/screenBugs/Bug.ts#L77), [PhaseItemEditor](https://github.com/ParabolInc/parabol/blob/strict-eq/packages/client/components/RetroReflectPhase/PhaseItemEditor.tsx#L89). I’m pretty certain these files should be in the dependency graph and detected by eslint based on [client/tsconfig.esilnt.json](https://github.com/ParabolInc/parabol/blob/strict-eq/packages/client/tsconfig.eslint.json#L0-L1). I don’t know why the files aren’t being picked up. I tested adding other errors in the files that eslint should complain about, with the same result. So it’s not detecting the file at all.

Since these files not being detected is a general eslinting problem and not specific to adding the strict eq rule, I propose we should create a separate issue for it so it needn’t block this PR.

mattkrick:
This highlights an error I found in how we call eslint-- https://github.com/ParabolInc/parabol/issues/5952
The ignore file also doesn't work because we need to pass in a flag on where to find it e.g. `--ignore-file ../../.eslintignore` via CLI. 

All that said, this PR moves us ahead a little bit, so we'll merge this & then we can tackle how we ignore files after we fix https://github.com/ParabolInc/parabol/issues/5952

igorlesnenko:
Yeah, looks like #5952 is also the reason why eslint is not picking some files. 
It does not pick files nested deeper than the second level: https://stackoverflow.com/questions/54165756/eslint-glob-is-not-considering-all-directories-recursively

igorlesnenko (approved):

mattkrick:
progress! i'll see if i can carve off the server to fix the pattern & we can make some more progress there

PR #5854, opened 2022-01-10
Make User.lastSeenAt non null

### LINKED ISSUE #5360: Make User.lastSeenAt non-null in postgres
We don't do sneaky things like created Users before they sign up.
So, that means a User should always have a `lastSeenAt` date.
This goes towards our postgres migration goal.

AC
- [ ] User.lastSeenAt is non-null with a default of NOW() in postgres
- [ ] getUserIdsToPauseQuery.sql does not have a null check for lastSeenAt

Estimate: 2 hours

tianrunhe:
Fixes #5360.

@Dschoordsch can you take a look? Thanks!

### DIFF
--- a/packages/server/__tests__/autopauseUsers.test.ts
+++ b/packages/server/__tests__/autopauseUsers.test.ts
@@ -37,7 +37,7 @@ test('Autopause users who where never active', async () => {
     data: {
       user: {
         id: userId,
-        inactive: true
+        inactive: false // user will have lastSeenAt populated upon sign up so won't be inactive
       }
     }
   })
--- a/packages/server/__tests__/autopauseUsers.test.ts
+++ b/packages/server/__tests__/autopauseUsers.test.ts
@@ -1,6 +1,6 @@
 import {sendIntranet, signUp} from './common'
 
-// As of 01/11/2021, this test is not testing the AutoPauseUsers mutation
+// As of 01/11/2022, this test is not testing the AutoPauseUsers mutation
 // because there is no easy way to set the lastSeenAt field to a date in the past
 test('Autopause users who where never active', async () => {
   const {userId} = await signUp()
--- a/packages/server/__tests__/autopauseUsers.test.ts
+++ b/packages/server/__tests__/autopauseUsers.test.ts
@@ -1,5 +1,7 @@
 import {sendIntranet, signUp} from './common'
 
+// As of 01/11/2021, this test is not testing the AutoPauseUsers mutation
+// because there is no easy way to set the lastSeenAt field to a date in the past
 test('Autopause users who where never active', async () => {
   const {userId} = await signUp()
   const paused = await sendIntranet({
--- a/packages/server/__tests__/autopauseUsers.test.ts
+++ b/packages/server/__tests__/autopauseUsers.test.ts
@@ -1,6 +1,6 @@
 import {sendIntranet, signUp} from './common'
 
-// As of 01/11/2022, this test is not testing the AutoPauseUsers mutation
+// As of 2022-01-11, this test is not testing the AutoPauseUsers mutation
 // because there is no easy way to set the lastSeenAt field to a date in the past
 test('Autopause users who where never active', async () => {
   const {userId} = await signUp()
--- a/packages/server/__tests__/loginWithPassword.test.ts
+++ b/packages/server/__tests__/loginWithPassword.test.ts
@@ -61,4 +61,5 @@ test('Login after signup', async () => {
   })
   const lastSeenAt = Date.parse(login.data.loginWithPassword.user.lastSeenAt)
   expect(lastSeenAt).toBeGreaterThan(Date.now() - 1000 * 60) // lastSeenAt should be more recent than 1 minute ago
+  expect(lastSeenAt).toBeLessThan(Date.now())
 })
--- a/packages/server/__tests__/loginWithPassword.test.ts
+++ b/packages/server/__tests__/loginWithPassword.test.ts
@@ -22,6 +22,7 @@ test('Login after signup', async () => {
             picture
             preferredName
             createdAt
+            lastSeenAt
           }
         }
         acceptTeamInvitation(invitationToken: $invitationToken) {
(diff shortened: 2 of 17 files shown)


packages/server/__tests__/loginWithPassword.test.ts
```diff
@@ -58,4 +59,6 @@ test('Login after signup', async () => {
       }
     }
   })
+  const lastSeenAt = Date.parse(login.data.loginWithPassword.user.lastSeenAt)
+  expect(lastSeenAt).toBeGreaterThan(Date.now() - 1000 * 60) // lastSeenAt should be more recent than 1 minute ago
```
Dschoordsch, line 63: +1 Would be good to also check
```
expect(lastSeenAt).toBeLessThan(Date.now())
```

packages/server/__tests__/autopauseUsers.test.ts
```diff
@@ -37,7 +37,7 @@ test('Autopause users who where never active', async () => {
     data: {
       user: {
         id: userId,
-        inactive: true
+        inactive: false // user will have lastSeenAt populated upon sign up so won't be inactive
```
Dschoordsch, line 40: +1 This invalidates the test since it's not verifying that autopauseUsers actually pauses some users. That's ok for now, but could you add a comment to the test?

Dschoordsch (changes requested):
With `lastSeenAt` being not null, we can adapt the GraphQL schema
https://github.com/ParabolInc/parabol/blob/a9ca12459364f1d24bee3cfd8e6a4bbc3cc314df/src/server/graphql/types/User.ts#L250
and also fix all the places where we checked or asserted `lastSeenAt`, like these
https://github.com/ParabolInc/parabol/blob/0f149354f886da275cf61c6bb3086c9eb765fe80/packages/client/hooks/useMeetingMemberAvatars.ts#L38
https://github.com/ParabolInc/parabol/blob/35a83dd162e6c37bcc3ba3fa2d5ff630bce06463/packages/client/modules/meeting/components/MeetingAvatarGroup/NewMeetingAvatarGroup.tsx#L98
https://github.com/ParabolInc/parabol/blob/09d0089d0adb1bea2bbe63511dae3d9dfacfc2a5/packages/server/dataloader/hydrateRedisDoc.ts#L12
https://github.com/ParabolInc/parabol/blob/dd4e0189a715685ee5f49fd77887aca0806f5089/packages/server/graphql/intranetSchema/mutations/connectSocket.ts#L51

I didn't check all places, so there are probably more.

packages/server/graphql/intranetSchema/mutations/connectSocket.ts
```diff
@@ -53,12 +53,11 @@ export default {
       adjustUserCount(userId, orgIds, InvoiceItemType.UNPAUSE_USER).catch(console.log)
       // TODO: re-identify
     }
-    const datesAreOnSameDay = now.toDateString() === lastSeenAt?.toDateString()
+    const datesAreOnSameDay = now.toDateString() === lastSeenAt.toDateString()
     if (!datesAreOnSameDay) {
       await updateUser(
         {
-          inactive: false,
-          lastSeenAt: now
```
Dschoordsch, line 61: -1 This is the only place were we update `lastSeenAt`

packages/server/dataloader/hydrateRedisDoc.ts
```diff
@@ -8,9 +8,6 @@ const hydrators = {
     return publicTemplates
   },
   User: (user: IUser) => {
```
Dschoordsch, line 10: -1 I think the function is wrongly typed and we still want to ensure the fields are converted to `Date`

Dschoordsch (changes requested):

packages/server/graphql/intranetSchema/mutations/connectSocket.ts
tianrunhe: oops nice catch!

packages/server/dataloader/hydrateRedisDoc.ts
tianrunhe: Actually I don't quite understand this function at all, can you elaborate a bit more? 🙏

Dschoordsch: I don't know the exact context the function is used, only what's in this file. In line 19 we can see it's used to convert parsed JSON object into the proper type. Since JSON does not have a date type, that's why the conversion was here.
The argument type of this User conversion function is incorrect since the input does not have the correct type for the date fields yet.

tianrunhe: Ohhh I see... it's the type of argument `IUser` gets me confused. Thanks!

Dschoordsch (approved):

tianrunhe:
@Dschoordsch do you have some sort of IOT devices where whenever there's a GitHub change, it alarms loudly to alert you??? I didn't even request a re-review yet! Thanks for that lightspeed🚀 response!

Dschoordsch:
@tianrunhe I just observe the notification channel on slack.

packages/server/__tests__/loginWithPassword.test.ts
```diff
@@ -58,4 +59,7 @@ test('Login after signup', async () => {
       }
     }
   })
+  const lastSeenAt = Date.parse(login.data.loginWithPassword.user.lastSeenAt)
```
mattkrick, line 62: +2 i've never seen the `parse` method before, i always used new Date().getTime(). TIL

packages/server/dataloader/hydrateRedisDoc.ts
```diff
@@ -1,15 +1,13 @@
...
-
 const hydrators = {
   publicTemplates: (publicTemplates: any[]) => {
     publicTemplates.forEach((template) => {
       template.createdAt = new Date(template.createdAt)
     })
     return publicTemplates
   },
-  User: (user: IUser) => {
+  User: (user: any) => {
```
mattkrick, line 8: +1 i think this hydrator is safe to remove now that we're always fetching users from PG, but happy to treat that as out of scope

mattkrick:
LGTM. Apologies for not getting to the other PRs first, I only get 15 minute breaks during this retreat so I'm picking off the small PRs 😅 

nice work!

Issue #5858, opened 2022-01-11
Highlight hovered cards in check-in meeting updates


Dschoordsch:
As a user in check-in meetings I intuitively point on a card and then proceed "This one is in review" without giving any context. If the remote participants could see which card I'm hovering above, it would be easier for them to follow.

![Peek 2022-01-11 22-39](https://user-images.githubusercontent.com/7331043/149026145-70dcefcd-d51d-43b2-ab42-c4ffc98ef129.gif)

### Acceptance Criteria (optional)
In check-in meetings in solo-updates phase
* [ ] check with a designer if this is ok
* [ ] when in right phase an task is hovered, mutation `setTaskHighlight` is triggered with `meetingId`, `taskId` and `isHighlighted`
* [ ] `SetTaskHighlightPayload` is added to `MeetingSubscription`
* [ ] `Task` is extended on client side with `isHighlighted` and `OutcomeCard` highlighted accordingly
* [ ] Only send the highlight when the highlighter is the author
* [ ] Clear all highlights when moving to next stage

ackernaut:
@Dschoordsch do you mind adding an initial estimation? Otherwise waiting for a poker session is fine, thanks

jordanh:
See also: #3430 and #2191

I would _love_ to make this a BP, so I will :)

jordanh:
Removing BP now that I see there is a PR :)

Dschoordsch:
@jordanh I took a look at my PR and I think it should be solved differently. The highlight belongs on the stage, not the task itself, so there is enough meat left to make it a BP. I would like to implement it, but it's not my top priority, so I could ditch it.

elielytz:
@Dschoordsch is this functionality ready to be tested meanwhile?

Dschoordsch:
@elielytz Matt had requested major refactoring, only after that testing makes sense.

elielytz:
okay great

PR #5859, opened 2022-01-11
Highlight checkin tasks when hovered in current solo update

Dschoordsch:
When a task is hovered in current solo update, it should be highlighted
for remote participants, so everyone sees what the current solo updater
is talking about.

## Testing
2 users: Hans the highlighter and Olga the observer
### both users in check-in
* join a Check-In meeting with two participants Hans and Olga
* Hans needs a couple of tasks
* advance the meeting to the solo update of Hans
* [ ] Hans now hovers over different tasks and Olga observes that the task is highlighted
* [ ] Hans drags the card and while it is dragged, it's still highlighted until Hans drops the card
* [ ] advance the meeting and go back with Olga to the solo updates of Hans, the task must not be highlighted anymore

### Hans in check-in, Olga in Team dashboard
* Hans joins a check-in meeting and advances to his solo updates
* Olga opens the Team dashboard and sees the same tasks
* [ ] Hans hovers over different tasks of his and Olga does not see any highlights

### Hans in check-in, Olga in retrospective 
* Olga advances a retrospective till the discussion phase and creates a new task for Hans
* Hans advances a check-in till his solo update
* [ ] Hans hovers over the new task and Olga must not see any highlights

### Olga joins after Hans highlighted a task
* Olga is at team dash
* Hans hovers a card in his check-in solo update and stays hovered
* [ ] Olga enters the meeting and sees that the task is hovered

Resolves: #5858


jordanh:
Oh wow. It is done already! Incredible!

Dschoordsch:
@enriquesanchez could you please take a look if the task highlight is ok? I just made it up.
Solo update highlight

@nickoferrall could you please review?

enriquesanchez:
Thanks for the ping @Dschoordsch! 

> could you please take a look if the task highlight is ok? I just made it up.

🤔 The orange drop shadow feels a bit too much... do you think we could try something like this instead?

Captura de Pantalla 2022-02-11 a la(s) 11 34 18 a m

I think we can keep the raised drop shaodow in its current dark color, and add a 1.5px border using `Sky Blue 500` from our color palette. 

What do you think?

enriquesanchez (changes requested):
Requested changes above ☝️

enriquesanchez:
> What do you think?

Ah, this is much better @Dschoordsch 🙌
Did you feel that `Sky 500` was too strong? Or why did you decide to go for `300`?

enriquesanchez:
Sounds reasonable. Thank you @Dschoordsch 👍

enriquesanchez (approved):

packages/client/modules/teamDashboard/components/TaskColumn/TaskColumnInner.tsx
```diff
@@ -8,12 +8,21 @@ import DraggableTask from '../../../../containers/TaskCard/DraggableTask'
 interface Props {
   area: AreaEnum
   tasks: TaskColumnInner_tasks
+  isMyMeetingSection?: boolean
```
nickoferrall, line 11: +1 `isViewerMeetingSection` might be more consistent with the codebase

packages/client/mutations/SetTaskHighlightMutation.ts
```diff
@@ -0,0 +1,98 @@
...
+  const meeting = store.get<ClientNewMeeting>(meetingId)
+  if (!meeting) {
+    return
+  }
+
+  const highlightedTaskId = meeting.getValue('highlightedTaskId')
+  const taskId = payload.getValue('taskId')!
+  const isHighlighted = payload.getValue('isHighlighted')
+
+  // If we missed the un-highlighting of the previous task, then clear it here.
```
nickoferrall, line 67: I wasn't able to reproduce this. Are there repro steps I can take or does it just happen occasionally?

packages/client/mutations/SetPhaseFocusMutation.ts
```diff
@@ -18,11 +18,11 @@ const mutation = graphql`
     }
   }
 `
-interface LocalHanders {
+interface LocalHandlers {
```
nickoferrall, line 21: +2 nice catch

packages/server/graphql/types/SetTaskHighlightPayload.ts
```diff
@@ -0,0 +1,34 @@
...
+      type: NewMeeting,
+      resolve: resolveNewMeeting
+    },
+    meetingId: {
+      type: GraphQLID
+    },
+    taskId: {
+      type: GraphQLID
+    },
+    task: {
```
nickoferrall, line 24: +1 it looks like we don't need to return `task` or `meeting` in the payload

nickoferrall (approved):
This looks amazing! Really excited to see it in prod

CR time: 1.5 hours

packages/client/mutations/SetTaskHighlightMutation.ts
Dschoordsch: This is just a precaution, it could happen when losing the connection for example.

packages/client/mutations/SetPhaseFocusMutation.ts
Dschoordsch: thanks

packages/client/modules/outcomeCard/containers/OutcomeCard/OutcomeCardContainer.tsx
```diff
@@ -42,6 +45,16 @@ const OutcomeCardContainer = memo((props: Props) => {
   const [editorState, setEditorState] = useEditorState(content)
   const {useTaskChild, isTaskFocused} = useTaskChildFocus(taskId)
 
+  React.useEffect(() => {
+    if (!isMyMeetingSection || !meetingId) return
+
+    SetTaskHighlightMutation(atmosphere, {
+      taskId,
+      meetingId,
+      isHighlighted: isTaskHovered
```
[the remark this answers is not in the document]
Dschoordsch: Good catch!

mattkrick:
TEST
- [ ] Olga enters meeting, then leaves to team dash
- [ ] Hans hovers a card & stays hovered
- [ ] Olga re-enters the meeting & sees that the task is hovered

packages/client/mutations/SetTaskHighlightMutation.ts
```diff
@@ -0,0 +1,98 @@
...
+  if (!task) {
+    return
+  }
+  task.setValue(isHighlighted, 'isHighlighted')
+  if (isHighlighted) {
+    meeting.setValue(taskId, 'highlightedTaskId')
+  }
+}
+
+const SetTaskHighlightMutation = (
```
mattkrick, line 85: +1 this will give you the correct handlers, vars & atmosphere

```ts
SetTaskHighlightMutation: SimpleMutation<TSetTaskHighlightMutation>
```

packages/client/subscriptions/MeetingSubscription.ts
```diff
@@ -79,6 +81,7 @@ const updateHandlers = {
   SetStageTimerPayload: setStageTimerMeetingUpdater,
   ResetRetroMeetingToGroupStagePayload: resetRetroMeetingToGroupStageUpdater,
   StartDraggingReflectionPayload: startDraggingReflectionMeetingUpdater,
+  SetTaskHighlightSuccess: setTaskHighlightUpdater,
```
mattkrick, line 84: can remove when isHighlighted is on the Task object

mattkrick (changes requested):
-1 big picture is the `hovering` state is determined by the server, so let's move it to the server & then handle whether or not to highlight by the component

packages/server/graphql/mutations/setTaskHighlight.ts
```diff
@@ -0,0 +1,59 @@
...
+    const operationId = dataLoader.share()
+    const subOptions = {operationId, mutatorId}
+
+    // AUTH
+    const viewerId = getUserId(authToken)
+    const task = await dataLoader.get('tasks').load(taskId)
+    if (!task) {
+      return standardError(new Error('Task not found'), {userId: viewerId})
+    }
+    if (task.userId !== viewerId) {
```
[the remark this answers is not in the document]
Dschoordsch: Yes, it's solely for solo updates. Only in that situation there is one clear presenter and a bunch of observers.

packages/server/graphql/mutations/setTaskHighlight.ts
```diff
@@ -0,0 +1,59 @@
...
+      return standardError(new Error('Not your turn'), {userId: viewerId})
+    }
+    const meeting = await dataLoader.get('newMeetings').load(meetingId)
```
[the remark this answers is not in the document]
Dschoordsch: No, wouldn't happen since you can only highlight your own tasks.

packages/client/subscriptions/MeetingSubscription.ts
Dschoordsch: Thanks for pointing this out and the other places which could be simplified by moving the state to the server.
Speaking for myself: When you request these bigger changes in how a problem is approached a lot of those smaller things follow automatically and are nice to figure out on my own. I think finding the balance between reducing review rounds and giving room to learn and grow is hard, but I think you could try to leave a bit more room for individual learning here.
I'm saying this because I looked at the big picture comment, reworked it accordingly and then a lot of those comments were obsolete. I wouldn't have gotten the same learning experience had I followed these comments one by one.

Dschoordsch:
@enriquesanchez A side effect of my recent changes is, that the highlight is also happening for the presenter. I think that's good because it is more consistent. Any objections?
Solo update highlight both highlighted

mattkrick (approved):
LGTM! I liked seeing this land in redis, that way folks who join late still see what everyone else sees 🎉

enriquesanchez:
>A side effect of my recent changes is, that the highlight is also happening for the presenter. I think that's good because it is more consistent. Any objections?

@Dschoordsch No objections. I agree it's more consistent and I actually like it 👍

packages/client/subscriptions/MeetingSubscription.ts
mattkrick: great feedback! it's always a struggle to know how many parts of the same issue to flag. will limit to just 1 or 2 places in the future

Issue #5438, opened 2021-09-27
Error dispatch(@mattkrick/graphql-trebuchet-client/lib/GQLTrebuchetClient)


Dschoordsch:
Affected 18 users in the last 11 days. Looks similar to #5381, looks like from 6.28.0 we changed the error handling of this.

https://sentry.io/organizations/parabol/issues/2654611644/?referrer=github_plugin

```
Error: Server error
  at dispatch (./node_modules/@mattkrick/graphql-trebuchet-client/lib/GQLTrebuchetClient.js:44:28)
  at call (./node_modules/@mattkrick/graphql-trebuchet-client/lib/GQLTrebuchetClient.js:9:18)
  at emit (./node_modules/eventemitter3/index.js:202:33)
  at WebSocket.<anonymous> (./node_modules/@mattkrick/trebuchet-client/lib/SocketTrebuchet.js:113:26)
```

mattkrick:
this makes sense, i assume these are errors where no `data` field is received from the gql executor due to a server timeout.
related https://github.com/ParabolInc/parabol/issues/5406

mattkrick:
AC
- [ ] the server should ALWAYS send an `{errors, data}` response in the event of a timeout.

igorlesnenko:
As far as I see `Server error` is not a timeout error

It is returned here:

[https://github.com/ParabolInc/parabol/blob/9e1b280bc2246d90426fa892b9be8d38e38ed881/packages/server/graphql/handleGraphQLTrebuchetRequest.ts#L55](https://github.com/ParabolInc/parabol/blob/9e1b280bc2246d90426fa892b9be8d38e38ed881/packages/server/graphql/handleGraphQLTrebuchetRequest.ts#L55)

While timeout is handled separately here:

[https://github.com/ParabolInc/parabol/blob/9e1b280bc2246d90426fa892b9be8d38e38ed881/packages/server/graphql/handleGraphQLTrebuchetRequest.ts#L60](https://github.com/ParabolInc/parabol/blob/9e1b280bc2246d90426fa892b9be8d38e38ed881/packages/server/graphql/handleGraphQLTrebuchetRequest.ts#L60)

For timeout, the message would be `The request took too long`

In `sanitizeGraphQLErrors` we hide the real error message and replace it with `Server error`

So I checked the exact same error timings in Datadog, and it is most likely the following error:

```
SEND TO SENTRY [Error: Cannot return null for non-nullable field Task.plaintextContent.] undefined
[...]

PR #5869, opened 2022-01-13
Add migration to fill missing plaintextContent for tasks

igorlesnenko:
#5438

How to test

1. If needed, remove some `plaintextContent`

```
r.db('actionDevelopment').table('Task').filter({
    "meetingId": "put_your_own_value"
}).replace(r.row.without('plaintextContent'))
```

2. Run migration `yarn db:migrate`

3. Check that all tasks have `plaintextContent`

```
r.db('actionDevelopment').table('Task').filter(function (row) {
    return row.hasFields({ plaintextContent: true }).not()
})
```

### DIFF
--- a/packages/server/database/migrations/20220112175838-addMissingPlainTextContentToTask.ts
+++ b/packages/server/database/migrations/20220112175838-addMissingPlainTextContentToTask.ts
@@ -1,0 +1,46 @@
+import {R} from 'rethinkdb-ts'
+import extractTextFromDraftString from 'parabol-client/utils/draftjs/extractTextFromDraftString'
+
+export const up = async function (r: R) {
+  // We have about 2500 such tasks
+  const missedTasks = await r
+    .table('Task')
+    .filter((row) => row.hasFields('plaintextContent').not())
+    .run()
+
+  console.log('Missed tasks found:', missedTasks.length)
+
+  if (missedTasks.length) {
+    const updates = [] as {
+      taskId: string
+      plaintextContent: string
+    }[]
+    for (const task of missedTasks) {
+      updates.push({
+        taskId: task.id,
+        plaintextContent: extractTextFromDraftString(task.content)
+      })
+    }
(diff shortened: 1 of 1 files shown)


packages/server/database/migrations/20220112175838-addMissingPlainTextContentToTask.ts
```diff
@@ -0,0 +1,46 @@
+import {R} from 'rethinkdb-ts'
+import extractTextFromDraftString from 'parabol-client/utils/draftjs/extractTextFromDraftString'
+
+export const up = async function (r: R) {
+  // We have about 2500 such tasks
+  const missedTasks = await r
+    .table('Task')
+    .filter((row) => row.hasFields('plaintextContent').not())
```
mattkrick, line 8: +1 i'm a little concerned about the 2500 rows, i know we've crashed when doing batches of <10,000 but I can't remember exactly what the size was. As long as we deploy during an offpeak time I'm willing to roll the dice.

For future reference, skip/limit and using `,pluck('id', 'content')` are great at keeping things well under memory constraints.

tiffanyhan (approved):
tested locally and it works, lgtm!

packages/server/database/migrations/20220112175838-addMissingPlainTextContentToTask.ts
igorlesnenko: @mattkrick I didn't split into batches as I saw a similar migration that uses batch size = 10000: https://github.com/ParabolInc/parabol/blob/5e04ddd020f21692c9c6547dede7fc73064fb54f/packages/server/database/migrations/20201008134420-addPlainTextContentToTask.ts#L36

So I thought that 2500 would be fine too.

I can split on smaller batches if needed.

jordanh:
LGTM! Great find and fix, sir

mattkrick:
great stuff!

PR #5873, opened 2022-01-15
fix #5554 support fast poker scoping

### LINKED ISSUE #5554: Adding Jira tasks in fast succession In Sprint Poker Scope stage is unreliable
### How to reproduce
Run a Sprint Poker and import new tasks from Jira during Scope stage in fast succession

### Expected
All tasks are added to the Estimate stage

### Actual
Some tasks seem randomly be deselected

Adding Jira tasks in fast succession

**Estimated effort:** 4 hours

mattkrick:
Fix #5554 

To test, see repro steps in #5554

First root cause:
- [ ] add 2 items to the poker scope
- [ ] quickly remove both items (call these events A,B)
- [ ] Even though A got called first, B would resolve sooner because it had 0 stages. A stage takes a long time to resolve because it includes a task, which includes a task estimate, which requires going out to jira or github
- [ ] Because B resolves sooner, it gets sent to the client before A
- [ ] When A makes it to the client, the client is no longer consistent to what is in the DB & a refresh is required

To fix:
- [ ] When ordering is important, assign the mutation to a group called a `casualOrderingGroup`. (To learn more, see http://websrv.cs.fsu.edu/~xyuan/cop5611/lecture5.html)
- [ ] For a given group, when the client receives a message from the server, it makes sure to dispatch events in the same order that they were sent 

Second root cause:
- [ ] the redisLock was being applied after fetching the meeting object from the dataloader. By the time the lock was achieved, that meeting object was stale
- [ ] For example, event A comes in fetches the meeting
- [ ] Event B comes in, fetches the same meeting
- [ ] Event A applies a lock, updates the meeting, and unlocks
- [ ] Event B waits for A to finish, but it's already using the stale meeting object

To fix:
- [x] lock the meeting early

I recorded 3 videos on debugging, but they're like 3 hours total, so I don't recommend them https://www.loom.com/team-videos/Code%20Tutorials


BartoszJarocki:
For reference, this is a [fix on the graphql-trebuchet-client ](https://github.com/mattkrick/graphql-trebuchet-client/commit/eac7bab6008d9e0e93f457adfb19bdb17be66089)

tianrunhe:
Now with ordering guaranteed, we can turn off the `isSubmitting` check in `JiraScopingResultItem` so user can fast check & uncheck a single issue.

packages/server/graphql/mutations/updatePokerScope.ts
```diff
@@ -58,21 +61,19 @@ const updatePokerScope = {
     }
 
     const {endedAt, teamId, phases, meetingType, templateRefId, facilitatorStageId} = meeting
-    if (endedAt) {
-      return {error: {message: `Meeting already ended`}}
-    }
     if (!isTeamMember(authToken, teamId)) {
+      // bad actors could be naughty & just lock meetings that they don't own. Limit bad actors to team members
```
tianrunhe, line 65: +2 careful thinking there!

tianrunhe (approved):

packages/client/components/JiraScopingSearchResultItem.tsx
```diff
@@ -52,11 +52,11 @@ const JiraScopingSearchResultItem = (props: Props) => {
   const {id: serviceTaskId, issueKey, summary, url} = issue
   const isSelected = usedServiceTaskIds.has(serviceTaskId)
   const atmosphere = useAtmosphere()
-  const {onCompleted, onError, submitMutation, submitting} = useMutationProps()
+  const {onCompleted, onError, submitMutation} = useMutationProps()
   const disabled = !isSelected && usedServiceTaskIds.size >= Threshold.MAX_POKER_STORIES
   const isTemp = isTempId(serviceTaskId)
   const onClick = () => {
-    if (submitting || disabled || isTemp) return
+    if (disabled || isTemp) return
```
BartoszJarocki, line 59: Should we do the same for `JiraScopingSelectAllIssues`? The only potential problem I see, this mutation can be really slow so it potentially keeps updating the UI after you already finished clicking. On the other hand, not allowing users to click at all while mutation is in progress is really bad UX. 

I'd personally do the same, and if performance problems starts to annoy us, create a separate task to fix that

BartoszJarocki (approved):
Great work! 👏🏻

packages/client/components/JiraScopingSearchResultItem.tsx
mattkrick: i'll make the change

BartoszJarocki:
@mattkrick this one is good to go!

Issue #5809, opened 2021-12-20
 Jira Server: integration provider


Dschoordsch:
Jira Server uses OAuth1.0a for authentication. [[docs](https://developer.atlassian.com/server/jira/platform/oauth/)] This requires each message to be signed. https://github.com/ddo/oauth-1.0a seems to be the best bet for that. There's no recent activity, but it's small enough that we could take over bug-fixing in case issues arise.

## Integration provider fields
* Jira Home URL
* Private key
* Consumer key

## Acceptance criteria
* [ ] modify `addIntegrationProvider` (based on #5805) to allow adding the required values
  * [ ] add `jiraServer` to `IntegrationProviderTypeEnum`
  * [ ] add `oauth1` to `IntegrationProviderTokenTypeEnum`
  * [ ] add `OAuth1ProviderMetadata` with `consumerKey`, `serverBaseUrl` and `privateKey`, add it to `ProviderMetadata` union
  * [ ] modify `createIntegrationProviderInsertParams` to work with OAuth1

jordanh:
I'm sure you could guess this, but what I'd love to ask you to do would be to complete the AC for this as you review #5805, when @BartoszJarocki says it's ready

Dschoordsch:
We have two options here for the mutation:

## 1) Provide private key and consumer key as input
### Pros
* easier to fit into existing mutation
* easier to implement
### Cons
* client of the mutation (sales) needs to generate those keys

## 2) Generate private key and consumer key and return as output
### Pros
* we can keep the private key private and only return the public key
* easier for the client of the mutation (sales)
* we can present all data the customer needs to enter at once (including redirect URL etc.)
### Cons
* the order to add the integration provider would be inverted, the provider would be added to parabol first before it is added in Jira Server. The provider would need to be marked as pending/inactive and activated later once the customer added the integration in Jira Server

While 2) would probably be a better customer experience, I would still go with 1) for this. Our sales can still generate these data and present the customer all at once.

Dschoordsch:
Stuck on #5829 
While the PR is ready to review, it targets feat/gitlab branch and thus needs it to be ready before it can be merged. I don't want to start adding reviewers in case there are major changes to feat/gitlab => putting issue into stuck

elielytz:
Is this tested locally? @Dschoordsch

PR #5874, opened 2022-01-17
Add JiraServer integration provider fields

Dschoordsch:
Add required fields to the database and `addIntegrationProvider`
mutation so the provider can be added.

## Testing
* generate public/private key pair and a consumer key
  ```sh
  openssl genrsa -out jira_privatekey.pem 1024
  openssl req -newkey rsa:1024 -x509 -key jira_privatekey.pem -out jira_publickey.cer -days 365
  openssl pkcs8 -topk8 -nocrypt -in jira_privatekey.pem -out jira_privatekey.pcks8
  openssl x509 -pubkey -noout -in jira_publickey.cer >jira_publickey.pem
  openssl rand -base64 32 >jira_consumerkey
  ```
* run the mutation
  ```graphql
  mutation {
    addIntegrationProvider(input: {
      teamId: " paste valid team id "
      service: jiraServer
      authStrategy: oauth1
      scope:org
      oAuth1ProviderMetadataInput: {
        serverBaseUrl: "https://jira.parabol.co/"
        consumerKey: " paste content of jira_consumerkey "
        privateKey: """ paste content of jira_privatekey.pem """
      }
    }) {
      __typename
      ... on AddIntegrationProviderSuccess {
        provider {
          id
          isActive
          teamId
          ... on IntegrationProviderOAuth1 {
            serverBaseUrl
          }
        }
      }
    }
  }
  ```
* verify values in DB

Resolves: #5809


Dschoordsch:
@igorlesnenko could you please review this PR?

igorlesnenko (approved):
Looks good to me

packages/server/graphql/mutations/addIntegrationProvider.ts
```diff
@@ -35,22 +36,43 @@ const addIntegrationProvider = {
     const subOptions = {mutatorId, operationId}
 
     // AUTH
-    if (!isTeamMember(authToken, teamId)) {
+    if (!isTeamMember(authToken, teamId) && !isSuperUser(authToken)) {
```
mattkrick, line 39: -1 this can't be a superUser because we grab the viewerId from L33 and provide it in the data payload. that means the data payload will return the superuser to all members of the team & we may not want that team to have that info on the superuser

packages/server/graphql/types/IntegrationProviderMetadataInputOAuth1.ts
```diff
@@ -0,0 +1,28 @@
...
+export const IntegrationProviderMetadataInputOAuth1 = new GraphQLInputObjectType({
+  name: 'IntegrationProviderMetadataInputOAuth1',
+  description: 'OAuth1 provider metadata',
+  fields: () => ({
+    serverBaseUrl: {
+      type: new GraphQLNonNull(GraphQLURLType),
+      description: 'The base URL used to access the provider'
+    },
+    consumerKey: {
+      type: new GraphQLNonNull(GraphQLString),
```
mattkrick, line 18: +1 if we know the consumerKey or privateKey doesn't have any spaces in it, we can use a GraphQLID type.

packages/server/postgres/migrations/1642094555266_addJiraServerIntegrationProvider.ts
```diff
@@ -0,0 +1,61 @@
...
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+  DO $$
+  BEGIN
+    ALTER TYPE "IntegrationProviderServiceEnum" ADD VALUE 'jiraServer';
+    ALTER TYPE "IntegrationProviderAuthStrategyEnum" ADD VALUE 'oauth1';
+    ALTER TABLE "IntegrationProvider"
+      ADD COLUMN IF NOT EXISTS "consumerKey" VARCHAR(255),
+      ADD COLUMN IF NOT EXISTS "privateKey" TEXT;
```
mattkrick, line 14: +1 nit: I _think_ we could get away with a VARCHAR(1024) or VARCHAR(2048) because we know it's going to be an RSA-1024 (+ a little extra for base64 encoding maybe?). if it won't always be RSA1024 then TEXT is probably the right solution

packages/server/postgres/queries/getIntegrationProvidersByIds.ts
```diff
@@ -14,6 +14,13 @@ interface IntegrationProviderWebhook extends IntegrationProviderActive {
   webhookUrl: string
 }
 
+interface IntegrationProviderOAuth1 extends Omit<IntegrationProviderActive, 'scope'> {
+  authStrategy: 'oauth1'
+  scope: Omit<IntegrationProviderScopeEnum, 'global'>
```
mattkrick, line 19: +1 i suppose there could be an integration that uses oauth1 that is global? i think confluence cloud might, but we can cross that bridge when it comes. safe for now!

packages/server/postgres/queries/src/upsertIntegrationProviderQuery.sql
```diff
@@ -29,6 +33,8 @@ ON CONFLICT ("teamId", "service", "authStrategy") DO UPDATE SET
   "scope" = EXCLUDED."scope",
   "clientId" = EXCLUDED."clientId",
   "clientSecret" = EXCLUDED."clientSecret",
+  "consumerKey" = EXCLUDED."consumerKey",
```
mattkrick, line 36: +1 
Names are hard! I leave the following to your discretion:

Oauth2 calls these `{clientId, clientSecret}`
Oauth1 calls these `{consumerKey, consumerSecret}` https://oauth.net/core/1.0a/
Jira Server calls these `{consumerKey, privateKey}`

As is, we're using Jira's terminology for all oauth1 stuff, which is probably OK since I hope it's the only time we'll ever use Oauth1.

However, if we wanted to reduce the number of columns in this table, we could share Oauth1 & Oauth2 fields. We'd have to change clientSecret to type `TEXT` (or a larger varchar), but that'd be it.

mattkrick (changes requested):
Just 1 minor. Great stuff!

packages/server/postgres/migrations/1642094555266_addJiraServerIntegrationProvider.ts
Dschoordsch: There is no benefit of using `VARCHAR(1024)` over `TEXT` except that the DB will catch malicious inserts maybe. According to Postgres recommendation I shouldn't make up arbitrary limits.
See Tip at https://www.postgresql.org/docs/9.1/datatype-character.html

packages/server/graphql/mutations/addIntegrationProvider.ts
Dschoordsch: The use case for this is us adding the integration provider on behalf of the organization. Given that the viewerId is only used for the subscription which has a big TODO in line 80 I would argue that the AddIntegrationProviderPayload needs rework. I think it should only have `provider` and `teamId`/`team` fields. It should also be sent to all teams in the org for org scope. The reason is that the `IntegrationProvider` itself is not bound to a `User` or `TeamMember`, but should be available to all TeamMembers in `sharedProviders` field.

packages/server/graphql/types/IntegrationProviderMetadataInputOAuth1.ts
Dschoordsch: `privateKey` has line breaks

packages/server/postgres/queries/src/upsertIntegrationProviderQuery.sql
Dschoordsch: I feel the names are confusing enough and I don't want to set a precedence for reusing columns for different data (although it's the same here, the different naming conventions ruin it). I will rename `privateKey` to `consumerSecret` though to keep the OAuth1 spirit.

packages/server/postgres/migrations/1642094555266_addJiraServerIntegrationProvider.ts
mattkrick: well defended! i like it

packages/server/graphql/types/IntegrationProviderMetadataInputOAuth1.ts
mattkrick: whoa they got rid of that (last time i checked was before the spec existed 😅 ) thanks for correcting my old intel.

packages/server/graphql/mutations/addIntegrationProvider.ts
mattkrick: +1 oh! that use case makes sense, i didn't consider that.
removing the user field in the payload definitely works.
alternatively, we could get the userId from the authToken.sub and that way each subscriber would only get _their_ TeamMember object

mattkrick:
great stuff, i learned a couple things, too!

PR #5876, opened 2022-01-17
Fix alignment of Add task button in discussions

enriquesanchez:
This super tiny PR fixes the center alignment of the `Add task` button in `Discussions`. It removes a  `4px` left margin on the icon.

**Before**
Screen Shot 2021-08-13 at 12 43 41 p m

**After**
Captura de Pantalla 2022-01-17 a la(s) 1 21 25 p m

### DIFF
--- a/packages/client/components/AddTaskButton.tsx
+++ b/packages/client/components/AddTaskButton.tsx
@@ -22,7 +22,7 @@ const AddTaskIcon = styled(Icon)({
   fontSize: 20,
   width: 20,
   height: 20,
-  margin: '0 4px'
+  margin: '0 4px 0 0'
 })
 
 const AddTaskLabel = styled('div')({


packages/client/components/AddTaskButton.tsx
```diff
@@ -22,7 +22,7 @@ const AddTaskIcon = styled(Icon)({
   fontSize: 20,
   width: 20,
   height: 20,
-  margin: '0 4px'
+  margin: '0 4px 0 0'
```
Dschoordsch, line 25: +1 can you fix it on the `AddPollButton` https://github.com/ParabolInc/parabol/blob/d82610163cfb75ed13f23d6b1bdb75748832e176/packages/client/components/AddPollButton.tsx#L25 as well? This way we don't forget it.

Dschoordsch (approved):

BartoszJarocki (approved):
good catch! let's fix what @Dschoordsch mentioned and merge 🚢

packages/client/components/AddTaskButton.tsx
enriquesanchez: Will do, thank you! 🙌

Issue #5633, opened 2021-11-08
Only show GitHub repos that have issues enabled


nickoferrall:
If it's not possible to add an issue to a GitHub repo, we shouldn't show that repo in our task integration.

While testing the GitHub integration, I selected a repo to integrate the task with and saw an error message. I assumed there was a bug until I noticed that the repo I selected had issues disabled.

A user might go through the same experience if they want to test the integration before using it in a "real" repo.

It looks like we can use the `has_issues` field [here](https://docs.github.com/en/rest/reference/repos) to check if issues are enabled. 

### AC

- [ ] In the GitHub task integration, don't show repos that have issues disabled

mattkrick:
closing in favor of #5094 to be fixed in the pr #5883

Issue #5094, opened 2021-07-01
Remove SuggestedIntegration in favor of GitHubRepo and JiraProject


mattkrick:
The SuggestedIntegration graphql type is basically the same thing as a github repo or jira project.
let's remove it & replace it with a interface that resolves to those things to make it a little easier to understand.

AC
- [ ] SuggestedIntegration and all subtypes are removed
- [ ] a new interface type is created that resolves to a _xGitHubRepo or JiraRemoteProject, perhaps `IntegratedRepo` or `RepoIntegration`?
- [ ] projectKey is no longer required to keep on the Task.integration for Jira tasks (today it is used to determine which project to recommend)
- [ ] list of suggestions are checked to be valid (ie WAMP is not available in the dropdown for jira because the project no longer exists)

Estimate: 8 hours

mattkrick:
> We want to remove the SuggestedIntegration type and replace it with a interface, but it looks like SuggestedIntegration is already an interface (see here)?

Yep, let's replace this with `TaskIntegration`.
TaskIntegration is an interface of _xGitHubRepo & JiraRemoteProject (and more to come later)
To determine if it is a _xGitHubRepo or JiraRemoteProject, today we use `isTypeOf`& that should probably be good enough.

The solution can be 1 or the other:
- Either we add `isTypeOf` to the _xGitHubRepo here https://github.com/ParabolInc/parabol/blob/33c3478b8699f0a61e092b533c36e4e6ac03c704/packages/server/graphql/rootSchema.ts#L32
- Or we could always fallback resolveType on the TaskInterface type
- e.g. `  isTypeOf: (value) => value.service === 'jira',`

> projectKey is no longer required to keep on the Task.integration for Jira tasks

see https://github.com/ParabolInc/parabol/blob/fa276fb44fff283b51e242221a1129610fd43045/packages/client/shared/gqlIds/JiraProjectKeyId.ts if you need projectKey, but the work should be done already

> What does WAMP mean in this context?

WAMP is just the name of an archived project on jira. Suggestions probably shouldn't come from most commonly integrated tasks. If we can get rid of this https://github.com/ParabolInc/parabol/blob/8729ec5e362606172b51eba7af1814bac71b3cc9/packages/server/graphql/queries/helpers/suggestedIntegrationHelpers.ts#L46, then SUPER!

nickoferrall:
[...]

PR #5883, opened 2022-01-18
Replace suggested integration with new interface

nickoferrall:
Fix issues #5094 and #5633 

Hey @igorlesnenko, would you mind reviewing this PR, please?

When you open the footer menu on a task card, you'll see a dropdown menu with Jira projects and GitHub repos (if they're both integrated). This PR refactors these suggested integrations to a new interface called `repoIntegration` that returns either a `RemoteJiraProject` or a `_xGitHubRepository`.


packages/client/mutations/CreateJiraTaskIntegrationMutation.ts
```diff
@@ -53,52 +52,6 @@ const CreateJiraTaskIntegrationMutation: StandardMutation<TCreateJiraTaskIntegra
...
-        const cloudId = integration.getValue('cloudId')
-        if (!projectName || !cloudId) return
-        const nextItem = {
-          cloudId,
-          projectKey,
-          projectName,
-          service: 'jira'
-        } as const
-        const id = makeSuggestedIntegrationId(nextItem)
-        // the fallback is likely never used
```
nickoferrall, line 90: The task successfully integrates with Jira and updates without the updater. I'm not sure how to test this fallback option or whether I should replace `suggestedIntegrationJira` with `jiraRemoteProject` here

packages/server/dataloader/atlassianLoaders.ts
```diff
@@ -109,7 +109,7 @@ export const jiraRemoteProject = (
             sendToSentry(projectRes, {userId, tags: {teamId, projectKey}})
             return null
           }
-          return projectRes
+          return {...projectRes, cloudId, userId, teamId}
```
nickoferrall, line 112: `cloudId`, `userId`, and `teamId` are all required fields on `remoteJiraProject` so I think we should return them from the dataLoader. They're needed in `JiraIssue` [here](https://github.com/ParabolInc/parabol/blob/6f13ec0e76efac0112db7123974c3f370a85593b/packages/server/graphql/types/JiraIssue.ts#L67)

packages/server/graphql/queries/helpers/fetchAllIntegrations.ts
```diff
@@ -10,15 +13,16 @@ const fetchAllIntegrations = async (
...
 ) => {
-  const results = await Promise.allSettled([
+  const results = (await Promise.allSettled([
     fetchAtlassianProjects(dataLoader, teamId, userId),
     fetchGitHubRepos(teamId, userId, dataLoader, context, info)
-  ])
-  const allIntegrations = results.flatMap((result) =>
-    result.status === 'fulfilled' ? result.value : []
-  )
+  ])) as PromiseSettledResult<Integration[]>[]
```
nickoferrall, line 19: If I remove this casting, results is type:
```
[PromiseSettledResult<JiraProject[]>, PromiseSettledResult<{
    id: string;
    service: string;
    nameWithOwner: string;
}[]>]
```
TypeScript complains on line 21 that `Type 'JiraProject[] | { id: string; service: string; nameWithOwner: string; }[]' is not assignable to type 'JiraProject | readonly JiraProject[]'.` I spent some time digging into a better solution without much luck, so I've opted for this for now.

packages/server/graphql/queries/helpers/fetchGitHubRepos.ts
```diff
@@ -62,11 +68,13 @@ const fetchGitHubRepos = async (
   const orgs = organizations.nodes || []
   const personalRepos = repositories.nodes || []
   const uniqueRepos = getUniqueRepos(orgs as any, personalRepos as any)
-  return uniqueRepos.map((repo) => ({
-    id: repo.nameWithOwner,
-    service: 'github',
-    nameWithOwner: repo.nameWithOwner
-  }))
+  return uniqueRepos
+    .filter(({hasIssuesEnabled}) => hasIssuesEnabled)
```
nickoferrall, line 72: Fixes https://github.com/ParabolInc/parabol/issues/5633

packages/server/graphql/queries/helpers/repoIntegrationHelpers.ts
```diff
@@ -0,0 +1,77 @@
...
+    }))
+    .run()
+
+  const threeMonthsAgo = new Date(Date.now() - ms('90d'))
+  return res
+    .filter(
+      (res) =>
+        permLookup[res.service] &&
+        (res.service !== 'jira' || res.projectKey) && // jira integrations are making it through that don't have a projectKey
+        res.lastUsedAt > threeMonthsAgo
```
nickoferrall, line 57: If a task was integrated with a project or repo within the past 3 months, it's probably more relevant than other projects & repos and should show up first in the dropdown menu

packages/client/components/NewJiraIssueInput.tsx
```diff
@@ -245,11 +245,11 @@ export default createFragmentContainer(NewJiraIssueInput, {
...
         ... on TeamMember {
-          suggestedIntegrations {
-            ...NewJiraIssueMenu_suggestedIntegrations
+          repoIntegrations {
+            ...NewJiraIssueMenu_repoIntegrations
             items {
-              ... on SuggestedIntegrationJira {
-                projectKey
+              ... on JiraRemoteProject {
+                key
```
nickoferrall, line 252: What's the reasoning behind using `key` on `JiraRemoteProject` and `projectKey` on `JiraIssue`? Perhaps `key` could be changed to `projectKey` to be more consistent?

igorlesnenko (approved):
Looks good to me, except I do not see GitHub repositories in the dropdown. I created a separate issue for this, as I don't see it in master too: #5889

packages/server/graphql/queries/repoIntegrations.ts
```diff
@@ -0,0 +1,51 @@
...
+      const {tms} = user!
+      const onTeam = authToken.tms.find((teamId) => tms!.includes(teamId))
+      if (!onTeam) {
+        return standardError(new Error('Not on same team as user'), {userId: viewerId})
+      }
+    }
+    const permLookup = await getPermsByTaskService(dataLoader, teamId, userId)
+    const [userIntegrationIds, allIntegrations] = await Promise.all([
+      getUserIntegrationIds(userId, teamId, permLookup),
+      fetchAllIntegrations(dataLoader, teamId, userId, context, info)
```
mattkrick, line 33: +1 the primary use case is someone clicking a repo that they have clicked before. For example:
- parabolinc/parabol
- parabolinc/action-devops

If we fetch every single repo from every single integration, we're going to make the user wait a few extra seconds & not provide much value for the primary use case. 

No need to fix now, I wanna get this PR merged,  we'll address this in #6007

packages/server/graphql/queries/helpers/fetchAllIntegrations.ts
mattkrick: +1 we can just use `Promise.all` here because we're guaranteed neither will throw

packages/server/graphql/queries/helpers/fetchAllIntegrations.ts
```diff
@@ -10,15 +13,16 @@ const fetchAllIntegrations = async (
...
+
+  const allIntegrations = results.flatMap((result) => {
+    return result.status === 'fulfilled' ? result.value : []
+  })
 
-  const getValue = (item) => (item.nameWithOwner || item.projectName).toLowerCase()
+  const getValue = (item) => (item.nameWithOwner || item.name)?.toLowerCase()
```
mattkrick, line 25: -1 remove here & let's do the sorting in the calling function (see comment above)

packages/client/utils/AtlassianManager.ts
```diff
@@ -42,6 +42,9 @@ export interface JiraProject {
   }
   simplified: boolean
   style: string
+  cloudId: string
```
mattkrick, line 45: -1 these values aren't coming from the Jira API, so we can't list them here. let's handle adding those in the calling function (maybe fetchAtlassianProjects?) we can augment the return value in the dataloader

packages/client/mutations/CreateJiraTaskIntegrationMutation.ts
mattkrick: agreed, no updater needed for this anymore!

packages/client/components/NewJiraIssueInput.tsx
mattkrick: good find!
reason being is the jira API return a project like `{key: 'DEV3}`, so i tried to match our project to the jira API.
I'd equate to to a foreign key.  for example, `Task.teamId` === `Team.id` (just replace the word id with key)

mattkrick (changes requested):
-1 Looking good! Just a couple finds

packages/server/dataloader/atlassianLoaders.ts
Dschoordsch: I think we shouldn't return those here because it's the input to this loader.
```
export interface JiraRemoteProjectKey {
  userId: string
  teamId: string
  cloudId: string
  projectKey: string
}
```
There shouldn't be a need to return the input unchanged. It's easy enough to add these on the call side.

packages/client/utils/makeRepoIntegrationId.ts
```diff
@@ -0,0 +1,29 @@
...
+type JiraItem = {
+  cloudId: string
+  lastUsedAt?: string
+  nameWithOwner?: null
+  projectKey: string
+  service: 'jira'
+  userId: string
+}
+
+const makeRepoIntegrationId = (item: GitHubItem | JiraItem) => {
```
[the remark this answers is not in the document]
nickoferrall: Makes sense 👍  For the GH case, I just returned the `nameWithOwner` as we already have that value

packages/server/dataloader/atlassianLoaders.ts
nickoferrall: Yeah, it feels strange returning the inputs untouched, but I also don't like calling `jiraRemoteProject` and not getting the non-nullable fields `cloudId`, `teamId`, or `userId`. I don't feel strongly either way so I've removed the inputs from the loader and added them to JiraIssue.project.

packages/server/graphql/queries/helpers/fetchAtlassianProjects.ts
```diff
@@ -1,46 +1,40 @@
...
+  const token =
+    accessToken || (await dataLoader.get('freshAtlassianAuth').load({teamId, userId}))?.accessToken
+  if (!token) return []
+  const manager = new AtlassianServerManager(token)
+  const jiraCloudIds = cloudIds || (await getCloudIds(manager))
+  if (!jiraCloudIds) return []
+  const projects = await manager.getAllProjects(jiraCloudIds)
+  return projects.map((project) => ({
+    ...project,
+    id: IntegrationRepoId.join({...project, projectKey: project.key, service: 'jira'}),
```
mattkrick, line 33: +1 passing in `cloudId` instead of a rest param is a little more expressive because we don't have to dive into the child function to know what fields it actually needs.

+1 since the service is already known, you can use `JiraProjectId.join` instead.

packages/server/graphql/queries/helpers/fetchAllRepoIntegrations.ts
```diff
@@ -0,0 +1,23 @@
...
+  context: GQLContext,
+  info: GraphQLResolveInfo
+) => {
+  const {dataLoader} = context
+  const [jiraProjects, githubRepos] = await Promise.all([
+    fetchAtlassianProjects(teamId, userId, context),
+    fetchGitHubRepos(teamId, userId, dataLoader, context, info)
+  ])
+  const getValue = (item) => (item.nameWithOwner || item.name).toLowerCase()
+  return [...jiraProjects, ...githubRepos].sort((a, b) => {
```
mattkrick, line 18: -1 we're doing half a sort here, not sorting in `allAvailableRepoIntegrations`, and sorting in `repoIntegrations`.
I propose we do all the sorting here

packages/server/graphql/queries/helpers/fetchAtlassianProjects.ts
mattkrick: +1 what are your thoughts on turning this whole function into a dataloader?
all it would need is `{teamId, userId}`. it could call the `freshAtlassianAuth` dataloader to get the auth, and then `atlassianCloudNameLookup` dataloader to get the cloudIds, and then call `getAllProjects`. that could eliminate a roundtrip to atlassian for the cloudIds in some cases & we could get rid of `getCloudIds`

packages/server/graphql/types/AtlassianIntegration.ts
```diff
@@ -157,17 +158,9 @@ const AtlassianIntegration = new GraphQLObjectType<any, GQLContext>({
       resolve: async (
         {accessToken, cloudIds, teamId, userId}: AtlassianAuth,
         _args: unknown,
-        {authToken}: GQLContext
+        context
       ) => {
-        const viewerId = getUserId(authToken)
-        if (viewerId !== userId) return []
```
mattkrick, line 163: -1 is this safe to remove? if you're a consultant in multiple jira clouds & i'm on your team in parabol, I could call

```gql
viewer {
  teamMember(teamId: 'mattAndNick', userId: 'nickO') {
    integrations {
      atlassian {
        projects {...}
      }
    }
  }
}
```

and get access to all your projects using your access token

mattkrick (changes requested):
almost there! just a little security thing & a small refactor

mattkrick:
Review time: 3 hours

packages/server/graphql/queries/helpers/fetchAllRepoIntegrations.ts
nickoferrall: That makes sense! I've pushed an update with that change

packages/server/graphql/types/AtlassianIntegration.ts
nickoferrall: Doh sorry, that wasn't deliberate! While refactoring and removing the code below, I accidentally removed those lines too

packages/server/graphql/queries/helpers/fetchAtlassianProjects.ts
nickoferrall: I like it! I've refactored this to a dataloader. We previously discussed with Georg that it's best not to return the inputs unchanged in the outputs of a dataloader. So, I'm just returning `manager.getAllProject` in the dataloader without the `userId` or `teamId`. I think that's OK, but it does concern me that [JiraRemoteProject](https://github.com/ParabolInc/parabol/blob/0c0a286dd28266dc1eec710029af6780574b1a3b/packages/server/graphql/types/JiraRemoteProject.ts#L29) requires a `userId` and `teamId` so we have to always remember modify the returned values of the dataloader.

packages/server/graphql/queries/helpers/fetchAllRepoIntegrations.ts
```diff
@@ -0,0 +1,66 @@
...
+    prevRepoIntegrationsIds.add(integrationId)
+  })
+  // always have lastUsedAt be a Date (to make the sort easier below)
+  const repos = allRepoIntegrations.map((repo) => {
+    if (prevRepoIntegrationsIds.has(repo.id)) {
+      const prevRepoIntegration = prevRepoIntegrations.find((repoIntegration) => {
+        const integrationId = IntegrationRepoId.join(repoIntegration)
+        return repo.id === integrationId
+      })
+      return {
```
mattkrick, line 44: -1 isn't this function supposed to filter out integrations that exist in prevRepoIntegrations?

packages/server/graphql/queries/helpers/fetchAtlassianProjects.ts
mattkrick: Up to you! As a challenge question: What's bad about returning an input value in the output? Every `tasksByTeamId` accepts a teamId & returns a task with the teamId on it. 

- teamId, userId are required inputs whether or not they're included in the output
- Every call to the dataloader eventually returns a JiraRemoteProject which requires teamId, userId
- we already do it with cloudId & `manager.getAllProjects`

it's less code, no chance of forgetting to add a mapper.

nickoferrall: I'd say that returning an input value unchanged as the output is arguably confusing as we can just include it in the place where we're calling the loader. But, I'll go with my original instinct of including it in the dataLoader because I think we could run into bugs where `RemoteJiraProject` doesn't include `userId` or `teamId`, and, as you say, it's less code.

packages/server/graphql/queries/helpers/fetchAllRepoIntegrations.ts
nickoferrall: No, this function is fetching the `prevRepoIntegrations` because the `lastUsedAt` field doesn't exist on the Jira projects or GH repos. So, I'm getting all of the ids of previous integrations for that user & team, and then, when mapping through the `repoIntegrations`, I check whether the `repoIntegration` exists in `prevRepoIntegrationsIds`. If it does, I use the `lastUsedAt` date, if not, I create a new one with `new Date(0)`.

I have a comment [here](https://github.com/ParabolInc/parabol/blob/be868d328092806995a27002d1ab696c1d776181/packages/server/graphql/queries/helpers/fetchAllRepoIntegrations.ts#L17) that tries to explain this, but I'd be happy to add more info if it's unclear.

mattkrick: ohhhh! i think i was stuck in the old way of thinking. this makes sense!
we can remove that extra find by storing the lastUsedAt in a lookup table instead of using the Set

For your consideration:


```ts
  const fetchedRepoIntegrations = [...jiraProjects, ...githubRepos]
  const repoIntegrationsLastUsedAt = {} as {[repoIntegrationId: string]: Date}
  prevRepoIntegrations.forEach((integration) => {
    const integrationId = IntegrationRepoId.join(integration)
    if (!integrationId) return
    repoIntegrationsLastUsedAt[integrationId] = integration.lastUsedAt
  })
  return fetchedRepoIntegrations
    .map((repo) => ({
      ...repo,
      // always have lastUsedAt be a Date (to make the sort easier below)
      lastUsedAt: repoIntegrationsLastUsedAt[repo.id] ?? new Date(0)
    }))
    .sort((a, b) =>
      a.lastUsedAt > b.lastUsedAt
        ? -1
        : a.service < b.service
        ? -1
        : a.id.toLowerCase() < b.id.toLowerCase()
        ? -1
        : 1
    )
```

mattkrick (approved):
+1 very nice!
1 comment on simplifying the code a bit, feel free to merge when ready!

PR #5884, opened 2022-01-18
Added caching Jira project avatars

### LINKED ISSUE #5833: JIra - getProjectAvatar refactored to cached images in redis
Datadog APM tells us that getting project avatars is a [pretty expensive thing to resolve.](https://app.datadoghq.com/apm/resource/gqlexecutor_01/graphql/f5c522ac77a9e943?query=env%3Aproduction%20service%3Agqlexecutor_01%20operation_name%3Agraphql%20resource_name%3A%22query%20JiraScopingSearchFilterMenuRootQuery%22&env=production&event=AgAAAX4mB-LpRtIq0gAAAAAAAAAYAAAAAEFYNG1CLWZrQUFCMHdvZTBJUGQzWWVzSAAAACQAAAAAMDE3ZTI2MDgtODBhOS00OWQ0LThkYjUtMGY2ZjcxNjdkMTgx&index=apm-search&spanID=570770132399154101&timeHint=1641315558121&topGraphs=latency%3Alatency%2Cerrors%3Aversion_count%2Chits%3Aversion_count&traceID=570770132399154101&start=1641313883530&end=1641317483530&paused=false) 

We've already fixed this for jira images in issues. We just cache those images in redis & then send down a URL instead of sending a base64 encoded image. We should do that here, too.

AC
- [ ] AtlassianManager.getProjectAvatar uses the same image strategy as images in JiraIssues

Effort: 8 hours

BartoszJarocki:
Fixes #5833 

To be able to use the existing code related to caching Jira images I had to make one change. Previously, the image mime type was calculated from filename (extension). The problem with this solution was that the file not necessarily has the extension in the filename. So far, it worked fine as Jira always returned the files with an extension. This isn't the case with avatars. 

Projects avatars are returned as an array of file URLs, without the extension.
```
{
  '48x48': 'https://api.atlassian.com/ex/jira/b8999f88-721c-49c7-aa0e-d60eeffc1eb6/rest/api/3/universal_avatar/view/type/project/avatar/10411',
  '24x24': 'https://api.atlassian.com/ex/jira/b8999f88-721c-49c7-aa0e-d60eeffc1eb6/rest/api/3/universal_avatar/view/type/project/avatar/10411?size=small',
  '16x16': 'https://api.atlassian.com/ex/jira/b8999f88-721c-49c7-aa0e-d60eeffc1eb6/rest/api/3/universal_avatar/view/type/project/avatar/10411?size=xsmall',
  '32x32': 'https://api.atlassian.com/ex/jira/b8999f88-721c-49c7-aa0e-d60eeffc1eb6/rest/api/3/universal_avatar/view/type/project/avatar/10411?size=medium'
}
```

Currently, the mime-type for a given file is taken from the Jira API response and it's stored in redis. I also prefixed all the redis keys for Jira images with `jira-image`. 

Tests
1. Check if project avatars are displayed correctly
Screenshot 2022-01-18 at 16 56 12

2. Check if images in Jira issues descriptions are displayed correctly
Screenshot 2022-01-18 at 16 56 04

### DIFF
--- a/packages/server/jiraImagesHandler.ts
+++ b/packages/server/jiraImagesHandler.ts
@@ -1,5 +1,4 @@
 import {promises as fsp} from 'fs'
-import mime from 'mime-types'
 import path from 'path'
 import {HttpRequest, HttpResponse} from 'uWebSockets.js'
 import jiraPlaceholder from '../../static/images/illustrations/imageNotFound.png'
@@ -8,13 +7,14 @@ import uWSAsyncHandler from './graphql/uWSAsyncHandler'
 import getRedis from './utils/getRedis'
 
 const getImageFromCache = async (fileName: string, tryAgain: boolean) => {
+  const imageKey = `jira-image:${fileName}`
   const redis = getRedis()
-  const imageBuffer = await redis.getBuffer(fileName)
+  const imageBuffer = await redis.getBuffer(imageKey)
   if (imageBuffer === null || imageBuffer.length === 0) return null
   if (imageBuffer.length > 1) return imageBuffer
   if (tryAgain) {
     await sleep(500)
-    return getImageFromCache(fileName, false)
+    return getImageFromCache(imageKey, false)
   }
   return null
 }
@@ -38,7 +38,8 @@ const jiraImagesHandler = uWSAsyncHandler(async (res: HttpResponse, req: HttpReq
     return
   }
 
-  const mimeType = mime.lookup(fileName)
+  const redis = getRedis()
+  const mimeType = await redis.get(`jira-image:mime-type:${fileName}`)
   if (!mimeType || !mimeType.startsWith('image/')) {
     await servePlaceholderImage(res)
     return
--- a/packages/server/utils/atlassian/jiraImages.ts
+++ b/packages/server/utils/atlassian/jiraImages.ts
@@ -1,7 +1,6 @@
 import cheerio from 'cheerio'
 import base64url from 'base64url'
 import crypto from 'crypto'
-import path from 'path'
 import ms from 'ms'
 import getRedis from '../getRedis'
 import AtlassianManager from 'parabol-client/utils/AtlassianManager'
@@ -32,13 +31,10 @@ export const updateJiraImageUrls = (cloudId: string, descriptionHTML: string) =>
       if (!imageUrl) return
 
       const absoluteImageUrl = `${projectBaseUrl}${imageUrl}`
-      const extname = path.extname(absoluteImageUrl)
-      const imageUrlWithoutExt = absoluteImageUrl.slice(0, absoluteImageUrl.length - extname.length)
-      const hashedImageUrl = createImageUrlHash(imageUrlWithoutExt)
-      const hashedImageWithExt = `${hashedImageUrl}${extname}`
-      imageUrlToHash[absoluteImageUrl] = hashedImageWithExt
+      const hashedImageUrl = createImageUrlHash(absoluteImageUrl)
+      imageUrlToHash[absoluteImageUrl] = hashedImageUrl
 
-      $(img).attr('src', createParabolImageUrl(hashedImageWithExt))
+      $(img).attr('src', createParabolImageUrl(hashedImageUrl))
     })
 
   return {updatedDescription: $.html(), imageUrlToHash}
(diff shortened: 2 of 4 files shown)


packages/server/jiraImagesHandler.ts
```diff
@@ -8,13 +7,14 @@ import uWSAsyncHandler from './graphql/uWSAsyncHandler'
 import getRedis from './utils/getRedis'
 
 const getImageFromCache = async (fileName: string, tryAgain: boolean) => {
```
Dschoordsch, line 9: +1 should we rename `fileName` to `imageUrlHash` to be consistent?

packages/server/jiraImagesHandler.ts
```diff
@@ -38,7 +38,8 @@ const jiraImagesHandler = uWSAsyncHandler(async (res: HttpResponse, req: HttpReq
     return
   }
 
-  const mimeType = mime.lookup(fileName)
+  const redis = getRedis()
+  const mimeType = await redis.get(`jira-image:mime-type:${fileName}`)
```
Dschoordsch, line 42: -1 we should get the buffer and mime type at once with `redis.multi` and not having 2 roundtrips to redis. Let's use the same
```
{
      imageBuffer: Buffer,
      contentType: string
    }
```
return type in `getImageFromCache` as we use in `getImage` in `AtlassianManager`

packages/server/utils/atlassian/jiraImages.ts
```diff
@@ -48,40 +44,42 @@ export const downloadAndCacheImages = async (
...
+  const imageResponse = await manager.getImage(imageUrl)
+  if (!imageResponse || !imageResponse.contentType) {
+    await redis.del(imageKey)
     return
   }
-  return redis.setBuffer(hashedImageUrl, imageBuffer, 'PX', IMAGE_TTL_MS)
+
+  return Promise.all([
+    redis.setBuffer(imageKey, imageResponse.imageBuffer, 'PX', IMAGE_TTL_MS),
+    redis.set(imageMimeTypeKey, imageResponse.contentType, 'PX', IMAGE_TTL_MS)
```
Dschoordsch, line 75: +1 why not use `redis.multi` here?

Dschoordsch (changes requested):

packages/server/utils/atlassian/jiraImages.ts
BartoszJarocki: `mset` does not support setting expiration time, that's why I used it that way. If there's any better way, I'm happy to improve

Dschoordsch: Wouldn't
```
return redis.multi()
    .setBuffer(imageKey, imageResponse.imageBuffer, 'PX', IMAGE_TTL_MS)
    .set(imageMimeTypeKey, imageResponse.contentType, 'PX', IMAGE_TTL_MS)
    .exec()
```
work just fine?

BartoszJarocki: you're right. not sure why, but I mistook `multi` with the `mset`, [which can't handle setting expiration times.](https://stackoverflow.com/questions/63722979/does-ioredis-mset-accepts-options-to-set-expiry-like-set-operation-does). I fixed it in both places, thank you!

packages/server/jiraImagesHandler.ts
BartoszJarocki: done

packages/server/jiraImagesHandler.ts
BartoszJarocki: done

Dschoordsch (approved):
Tested, works

igorlesnenko (approved):

packages/server/jiraImagesHandler.ts
```diff
@@ -1,20 +1,25 @@
...
+  const [[imageErr, imageBuffer], [contentTypeErr, contentType]] = await redis
+    .multi()
+    .getBuffer(`jira-image:${imgUrlHash}`)
+    .get(`jira-image:mime-type:${imgUrlHash}`)
+    .exec()
+
+  if (imageErr || contentTypeErr) return null
   if (imageBuffer === null || imageBuffer.length === 0) return null
-  if (imageBuffer.length > 1) return imageBuffer
+  if (imageBuffer.length > 1) return {imageBuffer, contentType}
```
mattkrick, line 19: can we trust that contentType will always be non-null?

packages/server/jiraImagesHandler.ts
```diff
@@ -1,20 +1,25 @@
...
 import sleep from '../client/utils/sleep'
 import uWSAsyncHandler from './graphql/uWSAsyncHandler'
 import getRedis from './utils/getRedis'
 
-const getImageFromCache = async (fileName: string, tryAgain: boolean) => {
+const getImageFromCache = async (imgUrlHash: string, tryAgain: boolean) => {
   const redis = getRedis()
-  const imageBuffer = await redis.getBuffer(fileName)
+  const [[imageErr, imageBuffer], [contentTypeErr, contentType]] = await redis
+    .multi()
```
mattkrick, line 12: -1 since we always fetch these in tandem & they both share the same expiration, we can save some lookup time & memory by using a hash data type in redis.
Not strictly necessary, but it could clean things up a bit & set a a good example for how to use hashes moving forward https://redis.io/commands#hash

mattkrick (changes requested):
Looks great!
Only change would be to use a hashmap vs. 2 separate keys in redis.
Hashes in redis are great for representing objects because the search time is faster & it expresses that the 2 values are related.

The only downside is that hash keys can't have their own expirations, which is why we don't use them more frequently in our app.

That said, I've been looking for a good PR to set the example in using hashes in redis & I think this is it! However, if time is pressing, we can keep it as is & do it another time

packages/server/jiraImagesHandler.ts
BartoszJarocki: Technically, web servers have to provide a proper content type. Here, I'm just returning whatever was cached in its original shape. Before sending back the response, we're checking both `imageBuffer` and `contentType` so we should be good!

```
  const cachedImage = await getImageFromCache(imgUrlHash, true)
  if (!cachedImage?.imageBuffer || !cachedImage?.contentType) {
    await servePlaceholderImage(res)
    return
  }
```

packages/server/jiraImagesHandler.ts
mattkrick: > storing data

yep! this looks great. a pipeline with `pexpire` is the right call

> read the data

you could use `hgetallBuffer` and then call `imageBuffer.contentType.toString()` if you don't want to use a pipeline. it's probably slightly more efficient (i don't think redis internally caches recent reads like that), but tbh i don't think the performance is a big deal either way.

The command exists in redis, but it's not in the typescript file! it's a secret 🤭

packages/server/jiraImagesHandler.ts
mattkrick: ah good call!

packages/server/jiraImagesHandler.ts
BartoszJarocki: thanks, @mattkrick. i've updated the PR so I'd appreciate it if you could take a look again 🙏🏻

mattkrick:
LGTM!
Review time: 1.5 hours

PR #5893, opened 2022-01-19
Remove 4px  left margin from Add Poll button

enriquesanchez:
Another tiny PR that removes a `4px` left margin from the icon of the `Add a poll` button. See #5876 for additional context


Dschoordsch (approved):
Works as advertised

PR #5830, opened 2022-01-04
No implicit any fixes

nickoferrall:
Fix #5757 

579 errors -> 358 errors

Hey @BartoszJarocki, could you review this PR, please?

PR #5904, opened 2022-01-20
Reverted connectionContext type check

nickoferrall:
In https://github.com/ParabolInc/parabol/pull/5830, I checked whether `connectionContext` was of type `ConnectionContext`, which is the type that `handleGraphQLTrebuchetRequest` is expecting. 

This broke the login & signup functionality as `connectionContext` can be of type `StatelessContext`. I've reverted the change. I haven't updated `handleGraphQLTrebuchetRequest` to expect `ConnectionContext | StatelessContext` yet as this will require many subsequent changes to the gql logic, which can be done in a separate PR.

I'll merge this PR right away as master is broken and this should fix it.


Dschoordsch:
Looks good.

PR #5906, opened 2022-01-20
send CC error message to client

### LINKED ISSUE #5905: Payment: Error from stripe doesn't propagate to the client
AC
- [ ] When updating a CC results in a `do_not_honor` the message should be sent to the client

Estimate: 4 hours

mattkrick:
Fix #5905

Updating credit cards did not show an error message

TEST
- [x] promote a team to Pro (use CC 4242424242424242 in dev to upgrade with any exp and cvc)
- [x] click to update the credit card
- [x] enter the CC `4000000000000002`
- [x] set the exp to 08/99 & get an error message
- [ ] set the exp to 08/25 and submit. see that the card was declined

### DIFF
--- a/packages/client/modules/userDashboard/components/CreditCardModal/CreditCardForm.tsx
+++ b/packages/client/modules/userDashboard/components/CreditCardModal/CreditCardForm.tsx
@@ -111,6 +111,8 @@ const CreditCardForm = (props: Props) => {
   const handleError = (param: string, fallback = 'Invalid details') => {
     const inputField = paramToInputLookup[param]
     if (inputField) {
+      // set submitting to false and clear general error
+      onCompleted()
       fields[inputField].setError(StripeError[inputField])
     } else {
       onError(new Error(fallback))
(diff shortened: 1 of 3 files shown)


tianrunhe (changes requested):
When you get the "invalid expiration date" error and you try to re-submit, the error went away so it looks like it was accepted but the modal did not close. If you click again, the error showed up again:

https://www.loom.com/share/b09f10e5bff44b2db5b8c84652e00fef

mattkrick:
oh that's good! will push a fix

mattkrick:
TEST
- [ ] submitting a bad request when there is already an error state shows the new error (submit `4000000000000002` twice)

mattkrick:
yikes, the problem was actually pretty large! handleSubmit was setting a new state & then kept using the stale state 🙈 
Would you mind poking this one again & making sure it works? @tianrunhe

tianrunhe (approved):

PR #5907, opened 2022-01-20
ensure window.opener is same-origin

mattkrick:
One of our customers is using google apps which is opening parabol with a `window.opener` which is a huge vulnerability to them trusting that we won't send any phishing attacks their way. There's no risk to us, but it does cause a bug with logging in because we assume that if something opened up the login page in a popup, we should close that popup.

This fix addresses that problem & makes sure that the thing that opened the login page is from the same origin.

AC
- [ ] login with SSO from parabol. it works. you'll need a SAML row in the DB & an okta account
- [ ] login with SSO from a different origin, it works. To achieve a different origin, just make a dumb HTML page & save it & open it & click the button (see below)
- [x] Above tested in Safari (we're using an optional catch binding, which wasn't supported before 2019)

```html
<html>
  <body>
    <button onclick="NewTab()">Parabol</button>

    <script>
      function NewTab() {
        window.open(
          "http://localhost:3000/saml-redirect?token=badToken"
        );
      }
    </script>
  </body>
</html>
```

### DIFF
--- a/packages/client/components/SAMLRedirect.tsx
+++ b/packages/client/components/SAMLRedirect.tsx
@@ -1,10 +1,10 @@
 import React, {useEffect, useState} from 'react'
-import DialogTitle from './DialogTitle'
-import DialogContent from './DialogContent'
-import StyledError from './StyledError'
-import InviteDialog from './InviteDialog'
 import useAtmosphere from '../hooks/useAtmosphere'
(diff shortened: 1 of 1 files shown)


nickoferrall (approved):
Tested & it works. LGTM!

PR #5923, opened 2022-01-21
Fix/demo bug - mattermost.auth undefined

mattkrick:
Demo was borked in here: https://github.com/ParabolInc/parabol/blob/eb295331feba7e2534c48012d252c2a52b802905/packages/client/components/StageTimerModalEndTimeSlackToggle.tsx#L74

Now it's not.

In the future, we can show folks that they can schedule notification in slack/mattermost.
but not today.


mattkrick:
i based my fix off master, i'll close this & cherry-pick the fix to staging

PR #5924, opened 2022-01-21
better impersonation

### LINKED ISSUE #4947: session-only impersonation
Impersonation could use a better UX.

AC
- [x] can visit `/admin/impersonate/<EMAIL>` to impersonate an email (no userId needed)
- [x] impersonating a user only lasts for the browser session. once the browser is closed, you're logged out. (technically, when the browser is opened, it checks to see if the token is impersonated. if so, it removes it & you have to sign-in again.)
- [x] when impersonating user X, it doesn't tell other team members that user X is online
 
@drewhousman1010 @gcrickman @ackernaut for your consideration

Estimate: 4 hours

mattkrick:
fix #4947

20% passion project, i've been wanting this for over a year now.

TEST
- [ ] create an account that you want to impersonate (`a@a.co`) & join the team of a superuser (rol = 'su')
- [ ] login as a superuser & 
- [ ] goto /admin/impersonate?email=a@a.co & you are them! (always works with ?e=a@a.co if you don't like to type)
- [ ] in another browser, login as the superuser. notice that the impersonated person is not online! (in development, dont' forget to run yarn postDeploy if you restarted the server)
- [ ] as the impersonated user, exit the browser tab. open a new one & see you are logged out

Loom here: https://www.loom.com/share/122527fe02174e63ad6f4c6e81657ff5

### DIFF
--- a/packages/server/socketHandlers/handleConnect.ts
+++ b/packages/server/socketHandlers/handleConnect.ts
@@ -37,6 +37,11 @@ mutation ConnectSocket {
 
 const handleConnect = async (connectionContext: ConnectionContext) => {
   const {authToken, ip, id: socketId} = connectionContext
+  const {rol} = authToken
+  if (rol === 'impersonate') {
+    connectionContext.ready()
+    return null
+  }
   const result = await publishInternalGQL({authToken, ip, query, socketId})
   if (!result) return null
   const {data} = result
--- a/packages/client/Atmosphere.ts
+++ b/packages/client/Atmosphere.ts
@@ -326,8 +326,9 @@ export default class Atmosphere extends Environment {
     }
 
     if (!this.authObj) return
-    const {exp, sub: viewerId} = this.authObj
-    if (exp < Date.now() / 1000) {
+    const {exp, sub: viewerId, rol, iat} = this.authObj
+    // impersonation token must be < 10 seconds old (ie log them out automatically)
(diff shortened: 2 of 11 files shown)


nickoferrall:
Hey @mattkrick, could you make that Loom video public, please? I get the message `Due to the privacy settings for this video, it cannot be played here at this time.`

nickoferrall:
Hey @mattkrick, just bumping this ^

mattkrick:
It should be available to all Parabol folks now. Felt weird making it public since the feature is a little secret

packages/server/socketHandlers/handleConnect.ts
```diff
@@ -37,6 +37,11 @@ mutation ConnectSocket {
 
 const handleConnect = async (connectionContext: ConnectionContext) => {
   const {authToken, ip, id: socketId} = connectionContext
+  const {rol} = authToken
+  if (rol === 'impersonate') {
+    connectionContext.ready()
+    return null
```
nickoferrall, line 43: Is the reason we don't want to run `publishInternalGQL` because we don't need subscriptions when impersonating?

nickoferrall (approved):
Tested & it works as expected. Nice work!

packages/server/socketHandlers/handleConnect.ts
mattkrick: @nickoferrall that would call the `connectSocket` mutation, which would make the user appear online.
we don't want that 🤫

PR #5938, opened 2022-01-24
Add generic createTaskIntegration mutation

### LINKED ISSUE #5819: Make createJiraTaskIntegration/createGitHubTaskIntegration generic
Both mutations are virtually identical and are just different in the IntegrationManage used. Since we also need a mutation like this for Jira Server and GitLab it is sensible to make it generic first

## Acceptance criteria
* [ ] replace `createJiraTaskIntegration` and `createGitHubTaskIntegration` with one `createTaskIntegration` mutation
  * [ ] arguments `integration` of type `IntegrationProviderTypeEnum`, `taskId` and `projectId`
* [ ] all integration specific actions are performed in their respective IntegrationManager

igorlesnenko:
Resolves #5819

**How to test:**
- Go to the team board
- Add jira integration
- Create a task on the team board
- Push a task to Jira
- See it pushed successfully
- Push someone else (another user) task to Jira, see a comment added on Jira
- See it pushed successfully
- Do the same for GitHub
- Check that pushing tasks to integrations works in the demo


packages/server/integrations/BaseTaskIntegrationManager.ts
```diff
@@ -0,0 +1,54 @@
...
+export default abstract class BaseTaskIntegrationManager {
+  readonly context: GQLContext
+  readonly info: GraphQLResolveInfo
+  readonly teamId: string
+  readonly viewerId: string
+  readonly userId: string | null
+  readonly task: Task
+  readonly team: Team
+  readonly rawContentStr: string
+  readonly createdBySomeoneElse: boolean
```
Dschoordsch, line 20: -2 Before this change the logic "If the task is pushed by someone else than the author, then add a comment" is now spread over 3 files, while before it was all in the mutation itself. I think this makes it much harder in the long run to maintain. Also the state which is stored in the base class versus the arguments in subclasses seem arbitrary and it seems conceptually wrong to have an instance depending on the `GQLContext` and `GraphQLResolveInfo`.

I think we would have a cleaner solution when we decouple this a bit and instantiate the `TaskIntegrationManager` with auth and userId

```
class TaskIntegrationManagerFactory {
  authLoaderKeys: Record<IntegrationProviderServiceEnumType, Loaders>
  getManager(service: IntegrationProviderServiceEnumType, userId, auth): TaskIntegrationManager

}

interface TaskIntegrationManager {
  createTask(rawContentStr: string, projectId: string): {
    integrationHash: string,
    integration: {
      accessUserId: string,
      service: IntegrationProviderServiceEnumType
    }
  }
  addCreatedBySomeoneElseComment(projectId, viewerName, assigneeName, teamName, teamDashboardUrl)
}
```

Dschoordsch (changes requested):
Let's try to get the manager right. It should have way less internal state and be independent of GraphQL context.

packages/client/mutations/CreateTaskIntegrationMutation.ts
```diff
@@ -0,0 +1,184 @@
...
+        id,
+        ...nextItem
+      })
+    const nextSuggestedIntegrations = [latestIntegration, ...items].slice(0, hasMore ? 3 : 1)
+    suggestedIntegrations.setLinkedRecords(nextSuggestedIntegrations, 'items')
+    suggestedIntegrations.setValue(true, 'hasMore')
+  }
+}
+
+const jiraTaskIntegrationOptimisticUpdater = (store, variables) => {
```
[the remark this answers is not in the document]
igorlesnenko: yeah, copy pasted as is for now

packages/server/integrations/BaseTaskIntegrationManager.ts
igorlesnenko: @Dschoordsch What if we don't use the base class state at all? I made some changes based on your comments. Can you please check?

packages/client/mutations/CreateTaskIntegrationMutation.ts
```diff
@@ -0,0 +1,184 @@
...
+        ... on _xGitHubIssue {
+          bodyHTML
+          title
+          number
+          repository {
+            nameWithOwner
+          }
+        }
+        ...TaskIntegrationLinkIntegrationGitHub
+        ...TaskIntegrationLinkIntegrationJira
```
Dschoordsch, line 37: -1 without `@relay(mask: false)` I get the error
```
Error: Fragment TaskIntegrationLinkIntegrationJira was used, but not defined
    at githubRequest (/home/man/Work/parabol/parabol/dev/gqlExecutor.js:72097:28)
    at async Promise.all (index 0)
    at async resolve (/home/man/Work/parabol/parabol/dev/gqlExecutor.js:49001:62)
    at async Object.TaskintegrationResolver (/home/man/Work/parabol/parabol/dev/gqlExecutor.js:33111:22) undefined
```

Looking just at the [docs](https://relay.dev/docs/api-reference/graphql-and-directives/#relaymask-boolean) it's not apparent why though. Also the documentation mentions `@inline` should be preferred instead, but I don't see the overlap from just the docs.
I'm just mentioning this as a learning opportunity, changing it to `@relay(mask: false)` should be good enough for this PR and @mattkrick will intervene otherwise.

packages/server/integrations/TaskIntegrationManagerFactory.ts
```diff
@@ -0,0 +1,54 @@
...
+    context?: GQLContext
+    info?: GraphQLResolveInfo
+  }): Promise<CreateTaskResponse>
+
+  getCreatedBySomeoneElseComment(
+    viewerName: string,
+    assigneeName: string,
+    teamName: string,
+    teamDashboardUrl: string
+  ): Doc | string
```
Dschoordsch, line 41: +1 Good enough for now. In the future when making more mutations generic we probably want to change this to `addCreatedBySomeoneElseComment`, but that's easier once we have the `addComment` functionality here. Let's move in small steps.

Dschoordsch (changes requested):
Only small nits for now, otherwise it looks like a good base to use for the other integration migrations which we want to make generic in the future.

packages/server/integrations/TaskIntegrationManagerFactory.ts
igorlesnenko: Yeah, the current functions just accepting a comment as a param, I just did not refactor them for this task

packages/client/mutations/CreateTaskIntegrationMutation.ts
igorlesnenko: @Dschoordsch I don't see this error. How can I reproduce it?

mattkrick: this is very strange because the stack trace is on the server & the `@relay` gets removed before sending the query to the server. There must be more here 🤔

Dschoordsch: @igorlesnenko I had this error with only GitHub integration added and then pushing to GitHub. You're right, since it happened on the server, it looks like a relay bug.

packages/server/graphql/mutations/createTaskIntegration.ts
```diff
@@ -61,92 +67,91 @@ export default {
...
-      })
+
+    const integrationManager = TaskIntegrationManagerFactory.getManager(integrationProviderService)
+
+    if (!integrationManager) {
+      return standardError(new Error('Integration provider is not supported'))
     }
 
-    const {content: rawContentStr, meetingId} = task
+    const authDataLoader = dataLoader.get(integrationManager.authLoaderKey)
```
Dschoordsch, line 77: @igorlesnenko 
> as I see for Jira server we load auth like
> ```
> return dataLoader
>  .get('teamMemberIntegrationAuths')
>  .load({service: 'jiraServer', teamId, userId})
> ```
> While for jira there is no server param
> ```
> parent.get('freshAtlassianAuth').load({teamId, userId})
> ```
> Maybe I should revert here to not return authLoaderKey but the whole auth loader?
> https://github.com/ParabolInc/parabol/pull/5938#discussion_r793422311

Very good observation, I think what could be a good solution is moving the `dataLoader.get` part into the `TaskIntegrationManagerFactory` so we get

```ts
const viewerIntegrationManager = TaskIntegrationManagerFactory.initManager(dataLoader, integrationProviderService, {teamId, userId})
const assigneeIntegrationManager = !viewerIntegrationManager && TaskIntegrationManagerFactory.initManager(dataLoader, integrationProviderService, {teamId, userId})
```

packages/client/mutations/CreateTaskIntegrationMutation.ts
Dschoordsch: That's a weird one, the fragment is present when the error is thrown
```gql
mutation CreateTaskIntegrationMutation(                                                                                                                          
  $integrationProviderService: IntegrationProviderServiceEnum!                                                         
  $projectId: ID!
  $taskId: ID!                                                                                                         
) {                                                                                                                        
  createTaskIntegration(integrationProviderService: $integrationProviderService, projectId: $projectId, taskId: $taskId) {
    error {
      message
    }
    task {
      ...IntegratedTaskContent_task
      integration {
        __typename
        ... on JiraIssue {
          cloudId
          cloudName
          url
          issueKey
          summary
          descriptionHTML
          projectKey
          project {
            name
            id
          }
        }
        ... on _xGitHubIssue {
          bodyHTML
          title
          number
          repository {
            nameWithOwner
            id
          }
        }
        ...TaskIntegrationLinkIntegrationGitHub
        ...TaskIntegrationLinkIntegrationJira
        id
      }
      updatedAt
      teamId
      userId
      id
    }
  }
}
  
fragment IntegratedTaskContent_task on Task {
  integration {
    __typename
    ... on JiraIssue {
      descriptionHTML
      summary
    }
    ... on _xGitHubIssue {
      bodyHTML
      title
    }
    id
  }
}
    
fragment TaskIntegrationLinkIntegrationGitHub on _xGitHubIssue {
  number
  repository {
    nameWithOwner
    id
  }
}

fragment TaskIntegrationLinkIntegrationJira on JiraIssue {
  issueKey
  projectKey
  cloudName
}
```
so it seems something in the parsing is wrong. With `@relay(mask: false)` the fragments are inlined and the error disappears. There are still fragments present in that case, so I don't know what exactly the issue is.

packages/server/graphql/mutations/createTaskIntegration.ts
Dschoordsch: As discussed, in place this would look like
```ts
    const [viewerIntegrationManager, assigneeIntegrationManager, team, teamMembers] = await Promise.all([
      TaskIntegrationManagerFactory.initManager(dataLoader, integrationProviderService, {teamId, userId})
      userId ? TaskIntegrationManagerFactory.initManager(dataLoader, integrationProviderService, {teamId, userId}) : null,
      dataLoader.get('teams').load(teamId),
      dataLoader.get('teamMembersByTeamId').load(teamId)
    ])
    const integrationManager = viewerIntegrationManager ?? assigneeIntegrationManager
    const accessUserId = viewerIntegrationManager ? viewerId : assigneeIntegrationManager ? userId : null

    if (!integrationManager) {
      return standardError(new Error('No auth exists for a given task!'), {userId: viewerId})
    }
```

packages/server/integrations/TaskIntegrationManagerFactory.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export default class TaskIntegrationManagerFactory {
+  public static getManager(
```
Dschoordsch, line 44: As discussed, we should change this method to
```ts
public static async initManager(
  dataLoader: DataLoaderWorker,
  service: IntegrationProviderServiceEnumType,
  {teamId, userId}: {teamId: string, userId: string}
): TaskIntegrationManager | null {
  if (service === 'jira') {
    const auth = await dataLoader.get('freshAtlassianAuth').load({teamId, userId})
    return auth && new JiraTaskIntegrationManager(auth)
  }
  ...
}

File: packages/server/integrations/JiraTaskIntegrationManager.ts [hunk 7]
```diff
@@ -0,0 +1,67 @@
...
+import JiraProjectId from '~/shared/gqlIds/JiraProjectId'
+import createJiraTask from '../graphql/mutations/helpers/createJiraTask'
+import JiraIssueId from '~/shared/gqlIds/JiraIssueId'
+import {CreateTaskResponse, TaskIntegrationManager} from './TaskIntegrationManagerFactory'
+import {AtlassianAuth} from '../postgres/queries/getAtlassianAuthByUserIdTeamId'
+import {Doc} from '../utils/convertContentStateToADF'
+
+export default class JiraTaskIntegrationManager implements TaskIntegrationManager {
+  public title = 'Jira'
+  public authLoaderKey = 'freshAtlassianAuth' as const
```
Dschoordsch, line 11: As discussed, instaed of storing the `authLoaderKey` here the `AtlassianAuth` should be passed in the constructor
```
private auth: AtlassianAuth
constructor(auth: AtlassianAuth) {
  this.auth = auth
}
```
and removed from the methods.

packages/client/mutations/CreateTaskIntegrationMutation.ts
igorlesnenko: Added @relay(mask: false

packages/server/integrations/JiraTaskIntegrationManager.ts
igorlesnenko: Made changes as discussed

packages/server/integrations/GitHubTaskIntegrationManager.ts
```diff
@@ -0,0 +1,74 @@
...
+    viewerName: string,
+    assigneeName: string,
+    teamName: string,
+    teamDashboardUrl: string
+  ): string {
+    return makeCreateGitHubTaskComment(viewerName, assigneeName, teamName, teamDashboardUrl)
+  }
+
+  async createTask({
+    accessUserId,
```
Dschoordsch, line 28: -1 we should remove this argument and use `this.auth.userId` instead

packages/server/integrations/GitHubTaskIntegrationManager.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    if (res.error) {
+      return {
+        error: res.error
+      }
+    }
+
+    const {issueNumber} = res
+
+    return {
```
Dschoordsch, line 62: +1 We can remove one level of nesting here
```ts
{
  integrationHash: GitHubIssueId.join(projectId, issueNumber),
  integration: {
    accessUserId,
    service: 'github',
    issueNumber,
    nameWithOwner: projectId
  }
}
```

Dschoordsch (changes requested):
Just some small things left.

Dschoordsch (approved):
Tested, pushing tasks to GitHub or Jira works

packages/client/mutations/CreateTaskIntegrationMutation.ts
```diff
@@ -0,0 +1,184 @@
...
+        ... on _xGitHubIssue {
+          bodyHTML
+          title
+          number
+          repository {
+            nameWithOwner
+          }
+        }
+        ...TaskIntegrationLinkIntegrationGitHub
+        ...TaskIntegrationLinkIntegrationJira @relay(mask: false)
```
mattkrick, line 37: Found the bug! It's in the code that I wrote for nest-graphql-endpoint.

That package must be removing this inline fragment but still sending the fragment body.
I'll push a fix for it 

-1 this is stuck on me pushing a fix for nest-graphql-endpoint.

mattkrick: fix pushed directly to the PR (it's just a change to package.json).

If you're curious, the code to fix it is here:  https://github.com/ParabolInc/nest-graphql-endpoint/commit/bc4b6fb482f9cb5180deae525d2a4f27e035e9a9#diff-56b72bf7eed61753d4b74cd03f03f5a6d152685f1ca5f055db96ac1c58fb2de6

loom: https://www.loom.com/share/e9c3e95cae6248e6bd9c139e945fa84b

packages/server/integrations/TaskIntegrationManagerFactory.ts
```diff
@@ -0,0 +1,55 @@
...
+      }
+    }
+
+export interface TaskIntegrationManager {
+  title: string
+
+  createTask(params: {
+    rawContentStr: string
+    projectId: string
+    createdBySomeoneElseComment?: Doc | string
```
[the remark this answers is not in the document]
igorlesnenko: Later, we will remove this param and replace it with `addCreatedBySomeoneElseComment` function. I would keep it as is for now

packages/server/graphql/types/SuggestedIntegrationJira.ts
```diff
@@ -23,6 +24,12 @@ const SuggestedIntegrationJira = new GraphQLObjectType<any, GQLContext>({
       description:
         'The name of the project, prefixed with the cloud name if more than 1 cloudId exists'
     },
+    projectId: {
```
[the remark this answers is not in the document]
igorlesnenko: If ok, I'll keep it as is for now then. If #5883 will be merged first, we will see the conflicts here and I'll resolve

packages/client/mutations/CreateTaskIntegrationMutation.ts
```diff
@@ -0,0 +1,184 @@
...
+  atmosphere,
+  variables,
+  {onCompleted, onError}
+) => {
+  const {integrationProviderService} = variables
+
+  return commitMutation<TCreateTaskIntegrationMutation>(atmosphere, {
+    mutation,
+    variables,
+    updater: (store) => {
```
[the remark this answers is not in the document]
igorlesnenko: I'll keep in mind and remove it after #5883 merged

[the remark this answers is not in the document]
igorlesnenko: Merged #5883 into my branch and changed the PR base

packages/server/graphql/types/SuggestedIntegrationJira.ts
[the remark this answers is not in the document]
igorlesnenko: Removed. Merged #5883 into my branch and changed the PR base

mattkrick:
LGTM!

Review time: 1 hour

Issue #5777, opened 2021-12-14
Dashboard filter controls enhanced to incl. search


jordanh:
## Enhancement

Today, when a task card is pushed to an integration we have a wonderful control that looks like this:

![image](https://user-images.githubusercontent.com/103000/146066679-9131e4d4-a9ec-4242-817a-acbfa1dba5d2.png)

When team's have many, many projects they can narrow the list by typing in a filter string.

Other drop downs, like these filter controls on https://action.parabol.co/me/tasks look like this:

![image](https://user-images.githubusercontent.com/103000/146066849-6a11ed7a-6577-41ed-b699-26b7d51d15e3.png)

They could really benefit from the same filter-string pattern.

### Acceptance Criteria

- Filter controls on the following routes updated to use a control with a filter string
   - https://action.parabol.co/me/tasks
   - https://action.parabol.co/team/:teamId

### Estimated effort

- 13 hours

elielytz:
## Test Results

### Task card integration filter check:
Status: Passed

- Case sensitive: Checked
- Filter using single/double letters : Chaecked
- Feedback if not searching results: available

https://user-images.githubusercontent.com/45159239/151502592-d651515a-5cad-4a59-b15d-4c15618c6d28.mp4

### Team member filter dropdown
Status: It seems is not in the staging environment yet, I am testing it on local environment

PR #5945, opened 2022-01-25
Add string filtering to team / team member filter dropdowns

jmtaber129:
fixes https://github.com/ParabolInc/parabol/issues/5777.

Adds search functionality to the team / team member filter dropdowns.  This is done by creating a couple of new things:
* A `SearchMenuItem` component - a styled menu label that accepts text input
* A `useSearchFilter` hook - A hook that takes in an array of items and a function for getting a string value of those items, and returns:
  * An `onQueryChange` handler - Takes a query string.  Pass this as `SearchMenuItem`'s `onChange` handler, or any other text input.
  * `filteredItems` - An array of items that match the query string
  * `query` - The last query passed to `onQueryChange`.
  
These patterns are based on the [filtering logic](https://github.com/ParabolInc/parabol/blob/2d134896d2da154b34fc80a5025022a63cd35a4b/packages/client/components/GitHubScopingSearchFilterMenu.tsx#L145-L154)+[components](https://github.com/ParabolInc/parabol/blob/2d134896d2da154b34fc80a5025022a63cd35a4b/packages/client/components/GitHubScopingSearchFilterMenu.tsx#L172-L177) for integrations mentioned in the original ticket.

This also refactors several existing search filters to use this component+hook:
* TaskFooterIntegrateMenuList.tsx
* GitHubScopingSearchFilterMenu.tsx
* JiraScopingSearchFilterMenu.tsx
* NewGitHubIssueMenu.tsx
* NewJiraIssueMenu.tsx

(Each of these refactors have been tested with local GH + Atlassian integrations)

Screen Shot 2022-01-26 at 2 40 51 PM

**Testing**
New search filtering:
* /me/tasks
  * Open the team dropdown (defaults to "All teams"), search for a team
  * Open the team member filter dropdown (defaults to current user), and search for a team member
  * Add a task to the board, click the task card's team, and search for a different team
* /team/:teamId
  * Open the team member filter dropdown ("all team members"), and search for a team member.  Confirm that team members are filtered and "All team members" is not shown when there's a search query.
  * Add a task card, click the task card's team member, and search for a different team member

Regression testing (for the refactors on existing search filtering):
* /me/tasks
  * Click the task card's "push to integration" button and search for GH and Atlassian integrations
* Sprint poker scoping
  * Filter by GH integration in scoping view
  * Filter by Jira integration in scoping view (search only shows up when more than 10 integrations are available)
  * For both Jira and GH, click new issue, and search for a different integration

### DIFF
--- a/packages/client/components/UserDashTeamMemberMenu.tsx
+++ b/packages/client/components/UserDashTeamMemberMenu.tsx
@@ -73,7 +73,7 @@ const UserDashTeamMemberMenu = (props: Props) => {
     query,
     filteredItems: matchedFilteredTeamMembers,
     onQueryChange
-  } = useSearchFilter(filteredTeamMembers, (item) => item.preferredName.toLowerCase())
+  } = useSearchFilter(filteredTeamMembers, (item) => item.preferredName)
 
   return (
     <Menu
--- a/packages/client/components/UserDashTeamMemberMenu.tsx
+++ b/packages/client/components/UserDashTeamMemberMenu.tsx
@@ -13,11 +13,22 @@ import useRouter from '~/hooks/useRouter'
 import constructUserTaskFilterQueryParamURL from '~/utils/constructUserTaskFilterQueryParamURL'
 import useAtmosphere from '~/hooks/useAtmosphere'
 import {UserTaskViewFilterLabels} from '~/types/constEnums'
+import styled from '@emotion/styled'
+import {PALETTE} from '~/styles/paletteV3'
+import MenuItemLabel from './MenuItemLabel'
 
 interface Props {
   menuProps: MenuProps
   viewer: UserDashTeamMemberMenu_viewer | null
 }
+
+const NoResults = styled(MenuItemLabel)({
+  color: PALETTE.SLATE_600,
+  justifyContent: 'center',
+  paddingLeft: 8,
+  paddingRight: 8,
+  fontStyle: 'italic'
+})
 
 const UserDashTeamMemberMenu = (props: Props) => {
   const {history} = useRouter()
@@ -72,6 +83,9 @@ const UserDashTeamMemberMenu = (props: Props) => {
     >
       <DropdownMenuLabel>{'Filter by team member:'}</DropdownMenuLabel>
       <SearchMenuItem placeholder='Search team members' onChange={onQueryChange} value={query} />
+      {query && matchedFilteredTeamMembers.length === 0 &&
+        <NoResults key='no-results'>No team members found!</NoResults>
+      }
       {query === '' && showAllTeamMembers && (
         <MenuItem
           key={'teamMemberFilterNULL'}
--- a/packages/client/components/UserDashTeamMemberMenu.tsx
+++ b/packages/client/components/UserDashTeamMemberMenu.tsx
@@ -73,7 +73,7 @@ const UserDashTeamMemberMenu = (props: Props) => {
     query,
     filteredItems: matchedFilteredTeamMembers,
     onQueryChange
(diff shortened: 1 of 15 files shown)


packages/client/components/UserDashTeamMenu.tsx
```diff
@@ -30,33 +32,47 @@ const UserDashTeamMenu = (props: Props) => {
...
+      defaultActiveIdx:
+        filteredTeams.findIndex((team) => teamIds?.includes(team.id)) + (showAllTeams ? 2 : 1)
     }
   }, [userIds, teamIds])
+
+  const {
+    query,
+    filteredItems: matchedFilteredTeams,
+    onQueryChange
+  } = useSearchFilter(filteredTeams!, (team) => team.name)
```
Dschoordsch, line 49: -1 please remove `!`

jmtaber129:
Added a "No results" placeholder for when a search query is present, but there are no results.

Fixed those bugs - the `getValue` methods were returning non-lower-case strings, so it was matching a lower-case query to mixed case strings.  Fixed this by calling `toLowerCase()` on the result of `getValue` in the `useSearchFilter` hook

packages/client/components/UserDashTeamMemberMenu.tsx
```diff
@@ -5,18 +5,31 @@ import Menu from './Menu'
...
+import styled from '@emotion/styled'
+import {PALETTE} from '~/styles/paletteV3'
+import MenuItemLabel from './MenuItemLabel'
 
 interface Props {
   menuProps: MenuProps
   viewer: UserDashTeamMemberMenu_viewer | null
 }
 
+const NoResults = styled(MenuItemLabel)({
```
Dschoordsch, line 25: +1 You could put this component in a separate file as well and reuse it.

Dschoordsch (approved):
Works now. I have just one comment you don't need to address. I will move this forward to maintainer review. Good work!

mattkrick:
This'll be my top PR to review tomorrow

mattkrick:
Code looks super clean! Great abstraction using the hook & refactoring all the similar components.

Issue 1:
- Search box appears when there is only 1 person on the team. Maybe hide the search box if the total results are < 6. Business logic is up to you, share your thoughts. 
Screen Shot 2022-01-28 at 10 27 24 AM


Issue 2:
- The active value appears in the search results. Was this intended? With the existing logic, if "Matt" is the current assignee, then "Matt" is removed from the results (1 fewer thing to scroll). No strong opinion on this, but want to get your thoughts

Issue 3:
- Clicking a team member when scrollTop > 0 doesn't work (at least i _think_ that's what's going on?) repro here:

https://user-images.githubusercontent.com/5514175/151603971-be92c350-5668-4b27-b8b0-163b38b0ff24.mov

mattkrick (changes requested):
-1 little usability bug with clicking in the mid or bottom of the scroll container. I agree with Georg about the NoResults refactor. Aside from that, looks _clean_!

jmtaber129:
* Re: issue 1 - Changed to only show the search bar when there are more than 5 items, since this seems to be around when a scrollbox is necessary.
* Nice catch on issue 2 - This just needed to pass in the `keepParentFocus` flag to the `Menu` component to prevent the search box from keeping/stealing focus (if the search box was scrolled out of view, the search box would regain focus when any part of the menu is clicked, including when attempting to select an item).  Fixed.
* Refactored out the `NoResults` component into an `EmptyDropdownMenuItemLabel` component.

mattkrick:
Retested & works great! 👍

Issue #5900, opened 2022-01-20
Release Test v6.44.0


adaniels-parabol:
- The basic checklist is required
- Adding tests from merged PRs is optional

## The basics
Run through this list at least once at [staging](https://action-staging.parabol.co):
- [ ] Smoke tested the [Demo](https://action-staging.parabol.co/retrospective-demo), unauthenticated
- [ ] Created an account
- [ ] Verified invite via mass link works (Team > Invite Button > Invite Link, visit URL in an incognito window)
- [ ] Verified invite via email works
- [ ] Added Slack, verified meeting notifications (use `#t_product_actiontime`)
- [ ] Added Mattermost, verified meeting notifications (check the test webhook in [Parabol Mattermost](https://mattermost.parabol.co/product/integrations/incoming_webhooks))
- [ ] Added GitHub, verified issue created (use `ParabolInc/publictestrepo`)
- [ ] Added Jira, verified issue created (use `parabol-2`)
- [ ] Smoke tested the Retro meeting with 2 players
- [ ] Smoke tested the Sprint Poker meeting with 2 players
- [ ] Smoke tested the Action meeting with 2 players
- [ ] Smoke tested cards on the dashboard
- [ ] Created a 2nd team
- [ ] Created a 2nd organization
- [ ] Upgraded to Pro (Credit card number: `4242 4242 4242 4242`, expiration date: any month in the future, CVC: `123`)
- [ ] Smoke tested the app on a mobile device (e.g. navigate between views, smoke test a Retro meeting, etc.)

### Added

- Add GitLab (#5829)
- Customise release snackbar #5743
- Add migration to fill missing plaintextContent for tasks #5869
- add undefined to wrappedQuery vars #5898

### Changed

- enforce server side lock (#5840)
- SECURITY: bump react-refresh-webpack-plugin #5896
- Security/bump graphql jit #5897
-

### Fixed

- Fix Original Estimate field can not be updated in Sprint Poker (#5865)
- Fix isSpectating toggle & Reveal Votes count #5782
- No implicit any fixes #5830
- Fix alignment of Add task button in discussions #5876
- Fix all noUncheckedIndexedAccess for parabol-client #5837

elielytz:
## Testing Summary
Most of the test cases passed except the following have **_passed with some errors_** or _**failed**_.

### Demo test
Passed with some errors: Most of the features are working except,
When the End Timer option is clicked in the timer button, **_"You found a bug"_** notification is popped.
[...]

PR #5961, opened 2022-01-25
feat(integration-tests): add integration tests using Playwright

blimmer:
This pull request demos what it might look like to use Playwright for integration testing the Parabol app.

## Background

[Playwright](https://playwright.dev/) is a next-gen integration testing framework built by Microsoft. Their [homepage](https://playwright.dev/) describes their features in more detail (TLDR; it's easy to write resilient tests and automatically run them against lots of browsers, both desktop, and mobile).

## Scope

This PR creates a prototype so folks can get a feel for how using Playwright for integration tests might work. I wanted to open up a draft pull request early on to get feedback. See the "TODO" section for other work that, in my opinion, we should do before merging the prototype.

## Demo

I recorded a demo [on Loom](https://www.loom.com/share/6249d7f95b424bf4a58168784bed1e4b) that describes a few things, including:

- Basic Playwright features
- Shows a failing test & how the "trace" is recorded
- Talks about the smoke test and "Reflect" page tests implemented
- Demos how to use the inspector to codegen new tests

## Desired Feedback

I'd love to hear what people think about this strategy, specifically:

- Does Playwright seem like a tool we'd be comfortable using?
- What were the problems with the existing Cypress setup? Does this prototype seem like it might work around those problems?
- If we want to move forward with this plan, what's the scope of the initial tests we write?

TYIA for taking the time to review!

## MVP TODO

- [x] Run integration tests in CircleCI. See also https://github.com/ParabolInc/parabol/pull/5481, which will be similar. @Dschoordsch mentioned that there were some issues starting the dev server in CI on that PR, so we'll need to get that fixed before this works. I took an initial look and I've got a few ideas on how we might fix the problem.
- [x] Add documentation to the README

## Resources

- When thinking about other tests that could be automated, use the [Release Process checklist](https://github.com/ParabolInc/parabol/issues/5900) as inspiration.


packages/integration-tests/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+1. Start the dev server according to the main [README instructions](/README.md#installation)
+2. In this directory, run `yarn test`.
+
+## TODO
+
+- [ ] Add eslint, especially for [no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/main/packages/eslint-plugin/docs/rules/no-floating-promises.md) rule
+- [ ] Run these tests in CircleCI (see also [https://github.com/ParabolInc/parabol/pull/5481](https://github.com/ParabolInc/parabol/pull/5481), which is related).
+- [ ] Junit reporting for CircleCI
+- [ ] Add authenticated tests
```
blimmer, line 16: This might not be MVP. I think we get a lot of value by adding tests for the un-authenticated demo before tackling authenticated tests. Feedback is welcome!

packages/integration-tests/tests/config.ts
```diff
@@ -0,0 +1,25 @@
...
+type TestingMode = 'production' | 'local'
+
+export class Config {
+  readonly testingMode: TestingMode
+  readonly rootUrlPath: string
+
+  constructor(env = process.env) {
+    this.testingMode = env.INTEGRATION_TESTING_MODE === 'production' ? 'production' : 'local'
+    this.rootUrlPath =
+      this.testingMode === 'production' ? 'https://action.parabol.co' : 'http://localhost:3000'
```
blimmer, line 12: We might not use the production-testing mode of this straight away, but I had discussed this with @mattkrick and @jordanh during our intro calls, so I thought I'd demo it as part of the prototype. My thought is that it's easier to use this `Config` object from the get-go so we have flexibility in the future.

packages/integration-tests/tests/retrospective-demo/reflect.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    await page.press(startTextbox, 'Enter')
+
+    await expect(
+      page.locator('[data-cy="reflection-stack-Start"] :text("Start doing this")')
+    ).toBeVisible()
+  })
+
+  // test('allows the user to enter feedback in the stop column')
+  // test('allows the user to enter feedback in the continue column')
+  // test('allows the user to delete previously entered feedback')
```
blimmer, line 21: I tried to balance getting something out there for quick review and demoing a solid bit of Playwright's functionality. I'm happy to add more tests to this prototype if that would be helpful.

packages/integration-tests/tests/retrospective-demo/smoke.test.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    // When they enter a response
+    const startTextbox = '[data-cy=reflection-column-Start] [role=textbox]'
+    await page.click(startTextbox)
+    await page.type(startTextbox, 'Start doing this')
+    await page.press(startTextbox, 'Enter')
+
+    // When they click the "Next" button twice
+    await page.click('[data-cy="next-phase"]:not([disabled=true])')
+    await page.click('[data-cy="next-phase"]:not([disabled=true])')
```
blimmer, line 34: I'm leveraging some of the leftover `data-cy` HTML attributes as part of this PR. However, it's increasingly popular to use integration tests as a way of improving the accessibility (a11y) of your frontends.

Frameworks like `@testing-library`, encourage finding elements by text, role, or HTML attribute type instead of these data attributes, which screen-readers ignore. For example, this "clickable div" is harder for a screen reader to understand vs. using the `button` HTML attribute (or `role=button`). It's also challenging to identify in this test, which is why we probably added the data attribute.

See [`@testing-library`'s docs](https://testing-library.com/docs/queries/about) and [this blog post](https://developers.livechat.com/updates/livechat-automated-a11y-testing) for more examples.

This is more of a long-term consideration, but another added benefit we might get from adding back an integration/through-the-glass testing setup.

packages/integration-tests/tests/config.ts
blimmer: Oh, another thing I just thought about is that this could be useful for smoke testing for customers who have on-prem deployments (cc @adaniels-parabol)

adaniels-parabol: That would be great @blimmer I currently do PPMI deployments in gitlab. It would be great to have a gitlab pipeline I could just put a run var in at the time of the deployment and test. Triggered via a webhook to the pipeline <3

packages/integration-tests/tests/retrospective-demo/smoke.test.ts
tianrunhe: I'm sure @enriquesanchez will love this!

tianrunhe:
This is soooo cool! Thanks for this and the demo! I'm bought in!

What do you think about testing multiple users joining meeting from different browsers? Is it simple for Playwright to tackle that?

blimmer:
@tianrunhe while I'm not sure about having the same test join from multiple browsers, I'm pretty sure we would be able to test multiple "browser windows" via the [`Page`](https://playwright.dev/docs/pages#multiple-pages) object.

This would allow us to verify that when someone does something on `page1`, something happens on `page2`.

packages/integration-tests/tests/config.ts
```diff
@@ -0,0 +1,25 @@
+import {Page} from '@playwright/test'
+
+type TestingMode = 'production' | 'local'
+
+export class Config {
+  readonly testingMode: TestingMode
+  readonly rootUrlPath: string
+
+  constructor(env = process.env) {
+    this.testingMode = env.INTEGRATION_TESTING_MODE === 'production' ? 'production' : 'local'
```
blimmer, line 10: @mattkrick mentioned that there's an `env` variable `host` - we should use that here.

mattkrick: perfect!

blimmer: updates made with [`160bed1` (#5961)](https://github.com/ParabolInc/parabol/pull/5961/commits/160bed146a1c2e275b052b98f60af4aa2592c17c)

packages/integration-tests/tests/retrospective-demo/smoke.test.ts
mattkrick: i'm 100% in favor of this. i propose we get rid of all the data-cy fields & if we can't get to a field via attribute then our a11y needs improvement!

blimmer: Great! I've got another branch (https://github.com/ParabolInc/parabol/compare/feature/continue-testing-demo?expand=1) that I've got some additional tests. I'll start following this pattern moving forward.

PR #5973, opened 2022-01-27
GitLab Oauth2 Poker Integration

### LINKED ISSUE #5404: Sprint Poker: GitLab: Scope: Oauth2
The next integration we'd like to add to sprint poker is GitLab :tada:.

To do that, we need the following:
- A UI
- A DB Table to store GitLab credentials
- A mutation to add credentials to the DB

### Acceptance Criteria
- [ ] When the user creates a new poker meeting, a GitLab tab is visible with the same functionality as the GitHub tab. It should look pretty similar to https://github.com/ParabolInc/parabol/blob/091037bf777be5d7e5d138e111b474e56e948f27/packages/client/components/ScopePhaseAreaAddGitHub.tsx. Note: we use Relay v11 now, so it's encouraged to use hooks e.g. `useFragment` instead of HOCs like `createFragmentContainer`. Pro tip: refactor the GitHub component to use hooks, copy the file, do a search & replace to change `Hub` to `Lab`.

Screenshot from 2021-09-15 13-13-23

#### Frontend
- [ ] Clicking "Import stories from GitLab" opens an oauth2 flow
- [ ] Completing the oauth2 flow should create a new record in the `GitLabAuth` table and the component showing should just say "Integrated!" 

#### Backend
- [ ] A postgres migration exists that creates a new table called `GitLabAuth`. The columns are about the same as the columns in `GitHubAuth`. To create a migration, call `yarn pg:migrate create gitlabAuth`. To explore postgres use pgadmin at localhost:5050
- [ ] A mutation to support adding gitlab auth to the database is created. Use `yarn newMutation addGitLabAuth` and it will create all the necessary files 

Estimate: 16 hours

nickoferrall:
Fix #5404 

#### To Test

- copy the GitLab env vars from the devops repo under `environments/local` and paste them into your `env` file in the `parabol` repo. If you don't have access to the devops repo, send a message in Slack requesting access
- if you're a super user, go to /admin/graphql and execute the `addFeatureFlag` mutation, adding gitlab. You can add yourself as a super user in postgres `localhost:5050`
- you may also need access to Parabol's GitLab app. I've invited Georg with have owner privileges and sent an email to elias@parabol.co (if that's the right email?) with guest privileges
- return to the `parabol` repo and run `yarn postdeploy`. This will run the `postDeploy` script which will call `primeIntegrations.ts` which will add the GitLab Parabol app credentials to your local db
- create a Sprint Poker meeting, go to the `Scope` phase, click on the `GitLab` tab, and click `Import issues from GitLab`. This should open the GitLab oauth2 page and, once you've clicked `Authorize`, you should see `Integrated!` within the GitLab card


packages/client/components/ScopePhaseArea.tsx
```diff
@@ -61,43 +63,49 @@ const TabContents = styled('div')({
...
+const baseTabs = [
   {icon: <GitHubSVG />, label: 'GitHub'},
   {icon: <JiraSVG />, label: 'Jira'},
   {icon: <ParabolLogoSVG />, label: 'Parabol'}
 ]
 
 const ScopePhaseArea = (props: Props) => {
   const {meeting} = props
+  const {viewerMeetingMember} = meeting
+  if (!viewerMeetingMember) return null
```
nickoferrall, line 75: Why is `viewerMeetingMember` nullable?

packages/client/modules/teamDashboard/components/ProviderRow/GitLabProviderRow.tsx
```diff
@@ -157,34 +157,32 @@ const GitLabProviderRow = (props: Props) => {
...
+            onClick={() => openOAuth(cloudProviderId, clientId, serverBaseUrl)}
+            palette='warm'
+            waiting={submitting}
+            onMouseOver={cloudOpenTooltip}
+            onMouseOut={cloudCloseTooltip}
+            ref={cloudRef}
+          >
+            {isDesktop ? 'Connect' : <Icon>add</Icon>}
+          </StyledPrimaryButton>
+          {/* {cloudTooltipPortal('Connect to GitLab Cloud')} */}
```
nickoferrall, line 170: I think we should hide the tooltip until GitLab Self-hosted is ready. I was wondering why GitLab had a tooltip but all the other integrations didn't

nickoferrall:

packages/client/components/ScopePhaseArea.tsx
Dschoordsch: -1 React really doesn't like it when you return before all hooks 
https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

packages/server/graphql/types/GitLabIntegration.ts
```diff
@@ -19,7 +19,7 @@ const GitLabIntegration = new GraphQLObjectType<any, GQLContext>({
     },
     cloudProvider: {
       description:
-        'The cloud provider the team member may choose to integrate with. Nullable based on env vars',
+        'The cloud provider the team member may choose to integrate with. Nullable based on env vars, e.g. a self-hosted instance',
```
Dschoordsch, line 22: -1 This description adds more confusion for me, do you mean "Could be null for self hosted parabol instances" or "Could be set to null if only self hosted GitLab servers are supported"?

packages/client/components/ScopePhaseAreaAddJira.tsx
```diff
@@ -50,14 +50,14 @@ const ScopePhaseAreaAddJira = (props: Props) => {
...
+  const authJira = () => {
     AtlassianClientManager.openOAuth(atmosphere, teamId, mutationProps)
   }
   return (
     <AddJiraArea>
-      <AddJiraButton onClick={importStories} size={'medium'}>
+      <AddJiraButton onClick={authJira} size={'medium'}>
         <JiraSVG />
-        Import stories from Jira
+        Import issues from Jira
```
Dschoordsch, line 60: +1 traditionally the team votes on stories, bugs are not estimated because by their very nature these are uncertain

packages/client/modules/teamDashboard/components/ProviderRow/GitLabProviderRow.tsx
Dschoordsch: +1 I agree, could you add a TODO in the comment?

Dschoordsch (changes requested):
Only small changes. Did not test yet.

packages/server/graphql/types/GitLabIntegration.ts
nickoferrall: Yeah, I meant it could be null for self-hosted Parabol instances, but I'm happy to remove this if it adds more confusion

packages/client/components/ScopePhaseArea.tsx
nickoferrall: Darn, I managed to miss those hooks below. Updated 👍

packages/client/components/ScopePhaseAreaAddJira.tsx
nickoferrall: Once we integrate with Jira, we refer to everything as issues instead of stories, e.g. "Search issues", "New Issue". I think we should be consistent with one or the other. I'd be happy to update everything to "stories" if that's more accurate in Jira land

Screenshot 2022-01-27 at 18 09 59

packages/client/components/ScopePhaseArea.tsx
```diff
@@ -71,33 +73,39 @@ const ScopePhaseArea = (props: Props) => {
   const {meeting} = props
   const [activeIdx, setActiveIdx] = useState(1)
   const isDesktop = useBreakpoint(Breakpoint.SIDEBAR_LEFT)
+  const {viewerMeetingMember} = meeting
+  if (!viewerMeetingMember) return null
+  const {user} = viewerMeetingMember
+  const {featureFlags} = user
+  const {gitlab: allowGitlab} = featureFlags
```
Dschoordsch, line 80: +1 I don't know how modular the integrations should be, ideally we would check if any `cloudProvider` or `sharedProvider` is available before we show the tab, otherwise it will just be blank.

Dschoordsch (approved):
Tested, works, but shows blank tab if no `cloudProvider` was added. Not technically part of AC, but might be better for self-hosted instances once we remove the feature flag.

packages/client/components/ScopePhaseArea.tsx
nickoferrall: Makes sense, I've pushed an update checking whether the `cloudProvider` & `sharedProviders` exist. I also noticed a bug in the `sharedProviders` logic and made a change.

Also, I understand that the `cloudProvider` has the GitLab Parabol app credentials which it gets from the `postDeploy` script using our env vars. But how will external users integrate their GitLab apps with Parabol? Is the `cloudProvider` currently just a placeholder that we'll update before releasing the integration?

packages/server/graphql/types/GitLabIntegration.ts
```diff
@@ -33,13 +33,13 @@ const GitLabIntegration = new GraphQLObjectType<any, GQLContext>({
       type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(IntegrationProviderOAuth2))),
       resolve: async ({userId}, _args, {dataLoader}) => {
         const teamMembers = await dataLoader.get('teamMembersByUserId').load(userId)
+        const organizationUsers = await dataLoader.get('organizationUsersByUserId').load(userId)
+        const orgIds = organizationUsers.map(({orgId}) => orgId)
```
Dschoordsch, line 37: -1 As stated in line 9 and line 32 the integration is per team member, thus we don't want all organizations, but only the one where the current team is in.

Dschoordsch (changes requested):

packages/server/graphql/types/GitLabIntegration.ts
nickoferrall: Could you clarify the difference between the `cloudProvider` and `sharedProviders` please? Am I right in saying that the `cloudProvider` is the default GitLab application (or Mattermost etc) that everyone can access, whereas the `sharedProviders` are the applications that a team or org has integrated with and only they can access? I'm a little unclear on this (see [comment here](https://github.com/ParabolInc/parabol/pull/5973#discussion_r794453361)). 

On line 35, we're getting all `teamMembers`, including ones that don't belong to this org or team. If we only want the providers integrated with this team or its org, I could do something like this:

```
        const team = await dataLoader.get('teams').load(teamId)
        if (!team) return standardError(new Error('Team not found'), {userId: viewerId})
        const {orgId} = team
        const orgTeams = await dataLoader.get('teamsByOrgIds').load(orgId)
        const orgTeamIds = orgTeams.map(({id}) => id)
        return dataLoader
          .get('sharedIntegrationProviders')
          .load({service: 'gitlab', orgTeamIds, teamIds: [teamId]})
```

Dschoordsch: You're right about `cloudProvider` and `sharedProviders` and your solution looks correct as well.

nickoferrall: OK thanks, update pushed 👍

Dschoordsch (approved):
Tested, works. Also mattermost still works.

packages/client/components/ScopePhaseArea.tsx
```diff
@@ -71,33 +73,43 @@ const ScopePhaseArea = (props: Props) => {
   const {meeting} = props
   const [activeIdx, setActiveIdx] = useState(1)
   const isDesktop = useBreakpoint(Breakpoint.SIDEBAR_LEFT)
+  const {viewerMeetingMember} = meeting
+  if (!viewerMeetingMember) return null
+  const {user} = viewerMeetingMember
+  const {featureFlags} = user
+  const gitlabIntegration = viewerMeetingMember.teamMember.integrations?.gitlab
```
mattkrick, line 80: +1 grab teamMember on L78
+1 integrations is nonnull

packages/client/components/ScopePhaseArea.tsx
```diff
@@ -71,33 +73,43 @@ const ScopePhaseArea = (props: Props) => {
...
+  const hasGitLabProvider = !!(
```
mattkrick, line 81: +1 _very_ nitty: `hasGitLabProvider` makes me think the viewer has already integrated (they "have" a gitlab integration).

Perhaps we say `isGitLabProviderAvailable`?

packages/client/components/ScopePhaseAreaGitLab.tsx
```diff
@@ -0,0 +1,62 @@
...
+import ScopePhaseAreaAddGitLab from './ScopePhaseAreaAddGitLab'
+
+interface Props {
+  isActive: boolean
+  gotoParabol: () => void
+  meetingRef: ScopePhaseAreaGitLab_meeting$key
+}
+
+graphql`
+  fragment ScopePhaseAreaGitLab_teamMember on TeamMember {
```
mattkrick, line 14: -1 this fragment should be added to the mutation response here: https://github.com/ParabolInc/parabol/blob/918b002b3315f0c5cfba9e5b5e50c80ed17ccdf9/packages/client/mutations/AddTeamMemberIntegrationAuthMutation.ts#L9

packages/client/components/ScopePhaseAreaGitLab.tsx
```diff
@@ -0,0 +1,62 @@
...
+    integrations {
+      gitlab {
+        auth {
+          provider {
+            scope
```
mattkrick, line 19: -1 scope isn't ever used in the component. I think we're requesting this in order to see if there is a provider. If that's the case, you can grab the `id` or `__typename` field to express that (since those fields are automatically requested & immutable)

packages/client/components/ScopePhaseAreaGitLab.tsx
```diff
@@ -0,0 +1,62 @@
...
+        }
+      }
+    `,
+    meetingRef
+  )
+  const {viewerMeetingMember} = meeting
+  if (!viewerMeetingMember || !isActive) return null
+  const {teamMember} = viewerMeetingMember
+  const {integrations} = teamMember
+  const hasAuth = !!integrations?.gitlab?.auth
```
mattkrick, line 45: +1 integrations is non-null

packages/client/mutations/AddTeamMemberIntegrationAuthMutation.ts
```diff
@@ -4,7 +4,7 @@ import {StandardMutation} from '../types/relayMutations'
 import {AddTeamMemberIntegrationAuthMutation as TAddTeamMemberIntegrationAuthMutation} from '../__generated__/AddTeamMemberIntegrationAuthMutation.graphql'
 
 graphql`
-  fragment AddTeamMemberIntegrationAuthMutation_part on AddTeamMemberIntegrationAuthSuccess {
+  fragment AddTeamMemberIntegrationAuthMutation_teamMember on AddTeamMemberIntegrationAuthSuccess {
```
mattkrick, line 7: I need to do a better job of expressing the naming conventions for these fragments at the top of mutation files.

For the fragments at the top of mutation files, the suffix is the name of the subscription (if any).
Since this mutation doesn't publish anything to a subscription channel, we can inline this gql fragment in the `mutation` fragment below. See minute ~16 of the loom video. Let me know if you have any questions or ideas on how to better communicate this with team members since this naming convention isn't enforced by a compiler like the other ones are.

packages/server/graphql/types/GitLabIntegration.ts
```diff
@@ -11,7 +12,11 @@ const GitLabIntegration = new GraphQLObjectType<any, GQLContext>({
     auth: {
       description: 'The OAuth2 Authorization for this team member',
       type: TeamMemberIntegrationAuthOAuth2,
-      resolve: async ({teamId, userId}, _args, {dataLoader}) => {
+      resolve: async (
+        {teamId, userId}: {teamId: string; userId: string},
+        _args: unknown,
+        {dataLoader}: GQLContext
```
mattkrick, line 18: +1 GQLContext isn't necessary here since it's passed through from L8

packages/server/graphql/types/GitLabIntegration.ts
```diff
@@ -31,18 +36,23 @@ const GitLabIntegration = new GraphQLObjectType<any, GQLContext>({
...
+          standardError(new Error('Team not found'), {userId: viewerId})
+          return []
+        }
+        const {orgId} = team
+        const orgTeams = await dataLoader.get('teamsByOrgIds').load(orgId)
+        const orgTeamIds = orgTeams.map(({id}) => id)
         return dataLoader
           .get('sharedIntegrationProviders')
-          .load({service: 'gitlab', orgTeamIds, teamIds})
+          .load({service: 'gitlab', orgTeamIds, teamIds: [teamId]})
```
mattkrick, line 55: +2 fixing my terrible business logic. This makes perfect sense!

packages/server/graphql/types/GitLabIntegration.ts
```diff
@@ -31,18 +36,23 @@ const GitLabIntegration = new GraphQLObjectType<any, GQLContext>({
...
-          .flat()
-          .filter(isValid)
-        const orgTeamIds = orgTeams.map(({teamId}) => teamId)
+      resolve: async (
+        {teamId}: {teamId: string},
+        _args: unknown,
+        {authToken, dataLoader}: GQLContext
+      ) => {
+        const viewerId = getUserId(authToken)
+        const team = await dataLoader.get('teams').load(teamId)
```
mattkrick, line 45: +1 you can use Georg's new API `loadNonNull` instead of `load` here because this GitLabIntegration object lives on a TeamMember object & you can't have a teamMember without a team, so we know team is guaranteed non-null

mattkrick (changes requested):
Really solid, works like a charm! Loom here (it ends abruptly because I trimmed the end of the video where i exposed development env vars) https://www.loom.com/share/facca2ad2d9948c4ab1cc10016e460c1

mattkrick:
looks great!

PR #5920, opened 2022-01-21
New retro templates

nickoferrall:
Fix #5890, fix #5891, fix #5892 

This PR adds new retro templates and prevents the template menu from scrolling up when you click on a menu item.

PR #5977, opened 2022-01-28
fix retro template scrolling

mattkrick:
fix for #5920

TEST
- [ ] have a list of 20+ retro templates. make the middle one active
- [ ] refresh & open the public template list. the active one is immediately visible without a scroll animation
- [ ] when you click any template items that are partially visible, they animate fully into view
- [ ] when you click any template items that are fully visible, no animation is applied

loom here: https://www.loom.com/share/69ecc2234fe64a3fbef064a064ffc457


nickoferrall (approved):
This looks great! Appreciate you sharing the journey through Loom

PR #5979, opened 2022-01-28
Added needed data dependencies to add agenda item mutation result

### LINKED ISSUE #5400: Check-in meeting navigation can't rejoin or click left nav item sometimes
We experience this ourselves during our check-in meetings.
Here's the meeting object that is borked: https://github.com/ParabolInc/action-devops/issues/73

## How to reproduce
Start a check-in meeting with 2 participants, the facilitator Fritz and Paula the participant. Add two agenda items, when on the first one Paula starts to write a comment and Fritz moves the meeting forward to the next item.
Paula stays on the first item and can finish their comment

## Expected behaviour
Paula can rejoin the meeting by either
1) pressing the "rejoin" button
2) selecting the active agenda item on the left

## Actual behaviour
Paula cannot move to another agenda item, neither by selecting it directly, nor by pressing "rejoin". Only refreshing the page works.

AC
- [ ] can navigate to items that i should be able to

Estimate: 4 hours

BartoszJarocki:
Fixes #5400.

This was a fun one. I went through the whole process of team subscriptions, updating the relay store, checking how `isInterruptingChickenPhase` works to find out that `AddAgendaItemMutation` does not fetch data needed to check if stage `isNavigable`. As a result, `navigateMeetingTeamUpdater` was updating the relay store with data that wasn't enough to handle navigation in some cases.

### DIFF
--- a/packages/client/mutations/AddAgendaItemMutation.ts
+++ b/packages/client/mutations/AddAgendaItemMutation.ts
@@ -26,7 +26,6 @@ graphql`
         ...ActionSidebarAgendaItemsSectionAgendaItemPhase @relay(mask: false)
         stages {
           ...ActionMeetingAgendaItemsStage @relay(mask: false)
-          # For useGotoStageId
           isNavigableByFacilitator
         }
       }
--- a/packages/client/mutations/AddAgendaItemMutation.ts
+++ b/packages/client/mutations/AddAgendaItemMutation.ts
@@ -21,14 +21,12 @@ graphql`
       }
     }
     meeting {
+      ...useGotoStageId_meeting
       phases {
         ...ActionSidebarAgendaItemsSectionAgendaItemPhase @relay(mask: false)
         stages {
           ...ActionMeetingAgendaItemsStage @relay(mask: false)
           # For useGotoStageId
-          id
-          isComplete
-          isNavigable
           isNavigableByFacilitator
         }
       }
--- a/packages/client/mutations/AddAgendaItemMutation.ts
+++ b/packages/client/mutations/AddAgendaItemMutation.ts
@@ -25,7 +25,10 @@ graphql`
         ...ActionSidebarAgendaItemsSectionAgendaItemPhase @relay(mask: false)
         stages {
           ...ActionMeetingAgendaItemsStage @relay(mask: false)
-          # For useMeetingGotoStageId
+          # For useGotoStageId
+          id
+          isComplete
+          isNavigable
           isNavigableByFacilitator
         }
       }
--- a/yarn.lock
+++ b/yarn.lock
@@ -2852,10 +2852,10 @@
     shortid "^2.2.12"
     tslib "^1.9.3"
 
-"@mattkrick/graphql-trebuchet-client@^2.2.0":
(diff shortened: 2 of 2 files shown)


BartoszJarocki:
@tianrunhe may I ask you to take a look?

packages/client/mutations/AddAgendaItemMutation.ts
```diff
@@ -25,7 +25,10 @@ graphql`
         ...ActionSidebarAgendaItemsSectionAgendaItemPhase @relay(mask: false)
         stages {
           ...ActionMeetingAgendaItemsStage @relay(mask: false)
-          # For useMeetingGotoStageId
+          # For useGotoStageId
+          id
+          isComplete
+          isNavigable
```
Dschoordsch, line 31: -1 It is better to make the dependency explicit so the relay compiler can tell us when something is not matching up
```
    meeting {
      ...useGotoStageId_meeting
      phases {
        ...ActionSidebarAgendaItemsSectionAgendaItemPhase @relay(mask: false)
        stages {
          ...ActionMeetingAgendaItemsStage @relay(mask: false)
          isNavigableByFacilitator
        }
      }
    }
  ```

**Review by Dschoordsch — CHANGES_REQUESTED** [review]
It's always the thing you check last ;-)

File: packages/client/mutations/AddAgendaItemMutation.ts (hunk 1 above)
Comment (BartoszJarocki) [reply packages/client/mutations/AddAgendaItemMutation.ts]: @Dschoordsch good point, fixed

**Comment by BartoszJarocki** [comment]
@Dschoordsch do you mind giving it another look?

File: packages/client/mutations/AddAgendaItemMutation.ts [hunk 2]
```diff
@@ -21,11 +21,12 @@ graphql`
...
     }
     meeting {
+      ...useGotoStageId_meeting
       phases {
         ...ActionSidebarAgendaItemsSectionAgendaItemPhase @relay(mask: false)
         stages {
           ...ActionMeetingAgendaItemsStage @relay(mask: false)
-          # For useMeetingGotoStageId
+          # For useGotoStageId
           isNavigableByFacilitator
```
Dschoordsch, line 30: -1 This is part of the `useGotoStageId_meeting` fragment and can be removed here.

Dschoordsch (changes requested):

packages/client/mutations/AddAgendaItemMutation.ts
BartoszJarocki: this was here before my changes so removing that could break _something_ unless we add `@relay(mask: false)`.

Dschoordsch: Fair point, then please update the comment accordingly.

BartoszJarocki: as `useMeetingGotoStageId` doesn't exist I guess it's best to just remove it

Dschoordsch (approved):
Tested, works
* rejoin by clicking on current agenda stage
* rejoin by clicking rejoin button

PR #5990, opened 2022-02-01
feat(dd-rum): identify user on Datadog session

blimmer:
This PR adds logic to identify users with Datadog, so we can easily see their IDs in the Datadog UI. Here's an example of me running this code locally:

Screen_Shot_2022-02-01_at_2_10_03_PM

### DIFF
--- a/packages/client/Atmosphere.ts
+++ b/packages/client/Atmosphere.ts
@@ -104,7 +104,8 @@ export default class Atmosphere extends Environment {
   upgradeTransportPromise: Promise<void> | null = null
   // it's only null before login, so it's just a little white lie
   viewerId: string = null!
-  userId: string | null = null // DEPRECATED
+  /** @deprecated */
+  userId: string | null = null
   constructor() {
     super({
       store,
--- a/packages/client/components/AnalyticsPage.tsx
+++ b/packages/client/components/AnalyticsPage.tsx
@@ -44,7 +44,9 @@ if (dsn) {
   })
 }
 
-if (__PRODUCTION__ && datadogClientToken && datadogApplicationId && datadogService) {
+const datadogEnabled =
+  __PRODUCTION__ && datadogClientToken && datadogApplicationId && datadogService
+if (datadogEnabled) {
   datadogRum.init({
     applicationId: `${datadogApplicationId}`,
     clientToken: `${datadogClientToken}`,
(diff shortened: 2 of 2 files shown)


packages/client/Atmosphere.ts
```diff
@@ -104,7 +104,8 @@ export default class Atmosphere extends Environment {
   upgradeTransportPromise: Promise<void> | null = null
   // it's only null before login, so it's just a little white lie
   viewerId: string = null!
-  userId: string | null = null // DEPRECATED
+  /** @deprecated */
```
blimmer, line 107: By using the `jsdoc`-style syntax, many IDEs will mark uses of this property with a strikethrough:

Screen Shot 2022-02-01 at 2 26 39 PM

This is a nice visual indicator that makes me think twice before using it.

packages/client/components/AnalyticsPage.tsx
```diff
@@ -57,17 +59,21 @@ if (__PRODUCTION__ && datadogClientToken && datadogApplicationId && datadogServi
...
 }
+
 // page titles are changed in child components via useDocumentTitle, which fires after this
 // we must guarantee that this runs after useDocumentTitle
 // we can't move this into useDocumentTitle since the pathname may change without chaging the title
 const TIME_TO_RENDER_TREE = 100
 
 const AnalyticsPage = () => {
-  const key = window.__ACTION__.segment
-  if (!key) return null // development use
```
blimmer, line 67: Since this "Page" is doing more than just dealing with Segment, I'm disabling on the `__PRODUCTION__` flag instead of the presence of the key.

packages/client/components/AnalyticsPage.tsx
```diff
@@ -57,17 +59,21 @@ if (__PRODUCTION__ && datadogClientToken && datadogApplicationId && datadogServi
...
+  if (!__PRODUCTION__) {
+    return null
+  }
+
   /* eslint-disable */
   const {href, pathname} = location
   const pathnameRef = useRef(pathname)
+  const segmentKey = window.__ACTION__.segment
```
blimmer, line 76: I renamed `key` to `segmentKey` for clarity.

packages/client/components/AnalyticsPage.tsx
```diff
@@ -157,6 +163,23 @@ const AnalyticsPage = () => {
...
+
+  useEffect(() => {
+    if (!datadogEnabled) {
+      return
+    }
+
+    const {viewerId} = atmosphere
+    if (viewerId) {
+      datadogRum.setUser({
+        id: atmosphere.viewerId,
```
blimmer, line 175: Ideally, we'd pass the `email` here, too. That's data I can get back from the `AnalyticsPageQuery`, defined above. But that's requiring a bit more refactoring, so I'm going to break it into a separate PR for clarity. I think that change will also potentially fix a few other bugs that I'll describe on the next PR.

tianrunhe (approved):
LGTM! This would be super helpful! Thanks for doing this!

PR #5991, opened 2022-02-01
feat(analytics): create a canonical location to identify users with third-parties

blimmer:
This PR takes #5990 and expands upon it. The main goal here was to reduce the number of places we identify users with third-party integrations, like Segment, LogRocket, Datadog, etc.

The inspiration to work on this came out of what I learned implementing #5990.


packages/client/components/AnalyticsPage.tsx
```diff
@@ -57,17 +64,37 @@ if (__PRODUCTION__ && datadogClientToken && datadogApplicationId && datadogServi
...
+    setViewerInfo(analyticsQueryRes?.viewer)
+  }
+  useEffect(() => {
+    fetchViewerInfo()
+  }, [atmosphere.viewerId])
+  useEffect(() => {
+    setupLogRocket(viewerInfo)
+    identifyUserWithDatadog(viewerInfo)
+    identifyUserWithSegment(viewerInfo)
+    identifyUserWithSentry(viewerInfo)
```
blimmer, line 87: My main goal with this change was to reduce the number of places we were identifying users with third-parties. Now, whenever the `viewerInfo` changes (when it's set or removed with a log out), we have this single effect that calls into all the relevant `identify` methods.

Previously, this method contained a duplicated call to the server endpoint. This also cleaned up the `ErrorBoundary` component, by moving all user identification to this consolidated place. Now, when adding another third-party plugin that needs to identify the user, there's a single place to look.

packages/client/components/AnalyticsPage.tsx
```diff
@@ -157,7 +144,62 @@ const AnalyticsPage = () => {
...
+function identifyUserWithSegment(viewerInfo?: ViewerInfo) {
+  const segmentKey = window.__ACTION__.segment
+  if (!segmentKey) return
+  if (!viewerInfo) return
+
+  const {id, email} = viewerInfo
+  safeIdentify(id, email)
+}
+
+function setupLogRocket(viewerInfo?: ViewerInfo) {
```
blimmer, line 160: Currently, there are four distinct use cases where we init, identify and report to LogRocket. In all of these examples, "error prone" means "encountered an error that bubbled to the Error Boundary within the last 14 days".

1. Logged out and error prone
2. Logged in and error prone
3. Logged in and watched
4. Any user who encounters an error that bubbles the ErrorBoundary (this is also where the Error Prone flag is set).

This logic could be simplified if we simply initialized for all users, identifying the user based on the `viewerInfo` properties set. However, I'm assuming the reason we don't do this today is because LogRocket charges per-session. 

To make this all simpler, I created a simple `LogRocketManager` class. This allows me to call `setUser` on the class whenever it changes. Then, whenever the LogRocket SDK is initialized, it'll use whatever user is current upon initialization.

packages/client/components/AnalyticsPage.tsx
```diff
@@ -157,7 +144,62 @@ const AnalyticsPage = () => {
...
+      id,
+      email,
+      isWatched
+    })
+  } else {
+    datadogRum.removeUser()
+  }
+}
+
+function identifyUserWithSentry(viewerInfo?: ViewerInfo) {
```
blimmer, line 192: I moved this out of `ErrorBoundary` and into this component so everything's in one place.

packages/client/components/ErrorBoundary.tsx
```diff
@@ -25,21 +23,11 @@ class ErrorBoundary extends Component<Props, State> {
   }
 
   componentDidCatch(error: Error, errorInfo: ErrorInfo) {
-    const {atmosphere} = this.props
```
blimmer, line 28: By creating the `LogRocketManager` class and moving the Sentry user identification into `AnalyticsPage`, we no longer need to reach into the Atmosphere class to fetch this information when an error occurs.

packages/client/components/ErrorBoundary.tsx
```diff
@@ -25,21 +23,11 @@ class ErrorBoundary extends Component<Props, State> {
...
-    const {viewerId} = atmosphere
-    const store = atmosphere.getStore()
-    const email = (store as any)?._recordSource?._records?.[viewerId]?.email ?? ''
```
blimmer, line 31: I don't think this was working, in general, before:

Screen Shot 2022-02-01 at 2 03 27 PM

because `_records` is a `Map`, you have to use `.get` as opposed to `[]` syntax.

packages/client/utils/LogRocketManager.ts
```diff
@@ -0,0 +1,70 @@
+import LogRocket from 'logrocket'
+
+interface User {
+  id: string
+  email: string
+}
+
+class LogRocketManager {
```
blimmer, line 8: As I mentioned above, this class allows deferring initialization to reduce the number of sessions (and, thus, the cost of LogRocket).

Now, in `AnalyticsPage`, we can call `setUser/unsetUser` whenever the user changes. Then, we can identify the proper user anytime the SDK is initialized.

blimmer (commented):
I know there's a good deal of code change here. It would be pretty straightforward for me to break this PR up by third-party (e.g., one PR for LogRocket, one for Sentry, etc) if that would make testing easier.

Overall, though, I think this PR provides a solid improvement to the current state and sets us up to easily identify users with future third-party apps.

packages/client/components/AnalyticsPage.tsx
nickoferrall: That's correct! LogRocket charges per session so we created the errorProne property to reduce costs. Nice work simplifying this

packages/client/components/AnalyticsPage.tsx
```diff
@@ -164,22 +145,61 @@ const AnalyticsPage = () => {
...
-    if (!datadogEnabled) {
-      return
-    }
+  return null
+}
+
+function identifyUserWithSegment(viewerInfo?: ViewerInfo) {
+  const segmentKey = window.__ACTION__.segment
+  if (!segmentKey) return
+  if (!viewerInfo) return
```
nickoferrall, line 154: +1 very nitty but `if (!segmentKey || !viewerInfo) return` would save us a line

nickoferrall (approved):
LGTM! Great work simplifying this and making it more performant

mattkrick:
TEST
- [ ] guarantee that segment page & identify events (seen in devtools XHR as p & i, respectively) are sent is the same order as they were previously. Below are a few example scenarios
- [ ] be logged out, goto /, then log in, make sure segment events are in the same order with the same payload as before
- [ ]  be logged in, goto a meeting, check for the same
- [ ] accept an invite while logged out, check for the same
- [ ] accept an invite while logged in, check for the same

packages/client/utils/LogRocketManager.ts
blimmer: Sounds good - should I remove logrocket as part of this PR? Or should it be a follow-up?

packages/client/Atmosphere.ts
```diff
@@ -437,4 +446,11 @@ export default class Atmosphere extends Environment {
     this.viewerId = null!
     this.userId = null // DEPRECATED
   }
+
+  private resetAuth() {
+    this.authToken = null
+    this.authObj = null
+    window.localStorage.removeItem(LocalStorageKey.APP_TOKEN_KEY)
+    window.localStorage.removeItem(LocalStorageKey.EMAIL)
```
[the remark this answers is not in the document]
blimmer: Thanks for this context. What do you think about caching the entire `ViewerInfo` result in LocalStorage? Then, on load, we could default the `viewerInfo` state to the values in LocalStorage. I think that would satisfy the "identify with Segment ASAP" requirement. 

Additionally, this would identify with all the other third parties faster as well.

Related to your comment below (https://github.com/ParabolInc/parabol/pull/5991#discussion_r813272371), if we're not concerned about the extra traffic, we could use the values from LocalStorage and also trigger a refresh against the endpoint on load. Then, if there are updates, we'd re-identify with the updated info early-on in the app lifecycle. This behavior would be similar to Apollo's `cache-and-network` [fetchPolicy](https://www.apollographql.com/docs/react/data/queries/#supported-fetch-policies).

packages/client/components/AnalyticsPage.tsx
```diff
@@ -66,11 +71,27 @@ if (datadogEnabled) {
...
-  if (!__PRODUCTION__) {
-    return null
+  const atmosphere = useAtmosphere()
+  const [viewerInfo, setViewerInfo] = useState<ViewerInfo | undefined>(undefined)
+  const fetchViewerInfo = async () => {
+    const analyticsQueryRes = await atmosphere.fetchQuery<AnalyticsPageQuery>(query)
+    setViewerInfo(analyticsQueryRes?.viewer)
   }
+  useEffect(() => {
+    fetchViewerInfo()
```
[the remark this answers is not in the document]
blimmer: Gotcha, yep I see that as well. I can add a `if` statement to not call the endpoint if the user is not logged in.

packages/client/Atmosphere.ts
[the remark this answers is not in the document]
mattkrick: we've gotta keep this out of the relay cache entirely since the only new field is email (the userId is persisted in the JWT) and we want to access the email before a fetch & doing that within relay means persisting the object in the local storage, rehydrating a relay store & validating it

mattkrick:
It looks like we're still a couple cycles away from merging & I think we've probably spent too much time on it to be worthwhile. The risk of messing up the analytics for marketing is high & the value of refactoring is too low when compared to other work currently on the sprint. I propose we keep this branch on remote & we can prioritize it during a future sprint planning session. I created a new issue for it here: #6149

PR #5992, opened 2022-02-04
fix old migrations, pg:build

mattkrick:
@Dschoordsch couple DB down migration & generated file fixes for your PR

### DIFF
--- a/packages/server/postgres/migrations/1642094555266_addJiraServerIntegrationProvider.ts
+++ b/packages/server/postgres/migrations/1642094555266_addJiraServerIntegrationProvider.ts
@@ -24,8 +24,8 @@ export async function down() {
   DO $$
   BEGIN
     ALTER TABLE "IntegrationProvider"
-      DROP COLUMN "consumerKey",
-      DROP COLUMN "consumerSecret",
+      DROP COLUMN IF EXISTS "consumerKey",
+      DROP COLUMN IF EXISTS "consumerSecret",
       DROP CONSTRAINT global_provider_must_be_oauth2;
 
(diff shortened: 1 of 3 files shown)


packages/server/postgres/queries/generated/getIntegrationProvidersByIdsQuery.ts
```diff
@@ -27,6 +27,7 @@ export interface IGetIntegrationProvidersByIdsQueryResult {
   clientSecret: string | null;
   serverBaseUrl: string | null;
   webhookUrl: string | null;
+  privateKey: string | null;
```
mattkrick, line 30: @Dschoordsch this was deprecated in favor of consumerSecret, right?
should we drop the column in the newest migration?

packages/server/postgres/migrations/1642094555266_addJiraServerIntegrationProvider.ts
```diff
@@ -45,7 +45,7 @@ export async function down() {
 
     ALTER TABLE "IntegrationProvider"
       ALTER COLUMN "service" TYPE "IntegrationProviderServiceEnum" USING "service"::text::"IntegrationProviderServiceEnum",
-      ALTER COLUMN "authStrategy" TYPE "IntegrationProviderAuthStrategyEnum" USING "service"::text::"IntegrationProviderAuthStrategyEnum",
+      ALTER COLUMN "authStrategy" TYPE "IntegrationProviderAuthStrategyEnum" USING "authStrategy"::text::"IntegrationProviderAuthStrategyEnum",
```
Dschoordsch, line 48: +2 good find!

Dschoordsch (approved):
I rebased on master

PR #5996, opened 2022-02-04
Track fatal bugs per 1,000 sessions

### LINKED ISSUE #5995: Track fatal bugs per 1,000 sessions
By keeping track of the number of fatal bugs per 1,000 sessions, we can better understand whether our app is becoming more or less buggy over time.

### AC

- [ ] Send an event to Segment when a fatal bug happens

nickoferrall:
Fix #5995 

Once we have an idea of the expected number of fatal bugs per 1,000 sessions, we could send a notification to Slack when there's a sudden sharp increase.

I created a chart in amplitude to track this: https://analytics.amplitude.com/parabol/chart/8m2qx3s

### To test

- add the segment env vars
- add a fatal bug, e.g. add an undefined variable to the `TimelineEventCard` component, and then visit the Timeline page at `/me`
- search for your `userId` in amplitude and see that the `Fatal Error` event has been tracked

### DIFF
--- a/packages/client/components/ErrorBoundary.tsx
+++ b/packages/client/components/ErrorBoundary.tsx
@@ -2,6 +2,7 @@ import * as Sentry from '@sentry/browser'
 import LogRocket from 'logrocket'
 import React, {Component, ErrorInfo, ReactNode} from 'react'
 import withAtmosphere, {WithAtmosphereProps} from '~/decorators/withAtmosphere/withAtmosphere'
+import SendClientSegmentEventMutation from '~/mutations/SendClientSegmentEventMutation'
 import {LocalStorageKey} from '~/types/constEnums'
 import safeInitLogRocket from '../utils/safeInitLogRocket'
 import ErrorComponent from './ErrorComponent/ErrorComponent'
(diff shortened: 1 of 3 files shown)


packages/client/components/ErrorBoundary.tsx
```diff
@@ -15,20 +17,31 @@ interface State {
...
 class ErrorBoundary extends Component<Props, State> {
   state: State = {
     error: undefined,
     errorInfo: undefined,
-    eventId: undefined
+    eventId: undefined,
+    isOldBrowserErr: false
+  }
+
+  componentDidUpdate() {
```
nickoferrall, line 31: I didn't refactor this component to hooks because there isn't a hooks equivalent to `componentDidCatch`: https://reactjs.org/docs/hooks-faq.html#do-hooks-cover-all-use-cases-for-classes

nickoferrall:
Hey @tianrunhe, could you review this one, please?

packages/client/components/ErrorBoundary.tsx
```diff
@@ -15,20 +17,31 @@ interface State {
...
+    const {error, isOldBrowserErr} = this.state
+    if (!error || isOldBrowserErr) return
+    const {atmosphere} = this.props
+    const {viewerId} = atmosphere
+    SendClientSegmentEventMutation(atmosphere, 'Fatal Error', {viewerId})
```
tianrunhe, line 36: I don't get why you couldn't do this in `componentDidCatch`?

nickoferrall: `componentDidCatch` re-renders four times and would send four events to Segment

tianrunhe:
@nickoferrall  What's the best/easiest way to test this locally?

nickoferrall:
@tianrunhe I've added testing steps to the PR description 👍

tianrunhe:
Hmm... I couldn't see the events published to Amplitude: https://www.loom.com/share/21b5ee7d9ea14517b312a386b3be378a

nickoferrall:
Thanks for the Loom! It looks like you have the `SEGMENT_FN_KEY` set-up, but the `SEGMENT_WRITE_KEY` is commented out. If you uncomment that one, do you see the event in Amplitude?

I've deleted the comment with the Loom as you could see some dev env vars. Might be worth checking with Matt on Slack whether we should change some of those env vars, just in case?

tianrunhe:
Approving as adding `SEGMENT_WRITE_KEY` works. Please see my question on Slack [here](https://parabol.slack.com/archives/C836NA350/p1644343578729209?thread_ts=1643302326.411000&cid=C836NA350).

tianrunhe (approved):

mattkrick:
LGTM!

Review time: 0 hours

Issue #5902, opened 2022-01-20
TypeError: Cannot read properties of undefined (reading 'some')


igorlesnenko:
Happened 6 times for 2 users in 3 days.

https://sentry.io/organizations/parabol/issues/2936475353/?referrer=github_plugin

```
TypeError: Cannot read properties of undefined (reading 'some')
  at None (./packages/client/components/GroupingKanban.tsx:86:37)
  at Array.forEach (<anonymous>)
  at a (./packages/client/components/GroupingKanban.tsx:82:22)
  at c (./packages/client/components/GroupingKanban.tsx:79:42)
...
(14 additional frame(s) were not displayed)
```

Dschoordsch:
Issue seems to be old since reflections being null was already addressed in a different location in #4354

Dschoordsch:
#6009 might be related, although I could only reproduce this locally on master. It might be that relay has different debug settings in dev and prod, I have to check.
Otherwise I could not reproduce this issue. Looking at the traces 4 of 6 people switched phases back to grouping before it happened. Some had opened spotlight. Nothing strikes me as odd in the logs.

PR #6010, opened 2022-02-07
Workaround for null reflections in GroupingKanban

Dschoordsch:
There is nothing obvious causing this issue and I couldn't reproduce it. Handle the error for now so the client is not affected but still track the error in Sentry so we can eventually find the root cause.

Resolves: #5902

### DIFF
--- a/packages/client/components/GroupingKanban.tsx
+++ b/packages/client/components/GroupingKanban.tsx
@@ -1,4 +1,5 @@
 import styled from '@emotion/styled'
+import {captureException} from '@sentry/minimal'
 import graphql from 'babel-plugin-relay/macro'
 import React, {RefObject, useEffect, useMemo, useRef, useState} from 'react'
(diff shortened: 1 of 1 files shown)


Dschoordsch:
@nickoferrall could you please take a look?

nickoferrall (approved):
LGTM 👍

mattkrick:
man, this bug has stumped me for literally years... 
We also catch it here: https://github.com/ParabolInc/parabol/blob/213890d97e38bdb32f7fac045373f3b2ee212965/packages/client/components/RetroDiscussPhase.tsx

This PR is safe & definitely an improvement.
I think the next step is to figure out if this is caused by an optimistic updater function. If so, we know where to look. If not, we need to see if this affects the person calling the mutation, or if it happens to subscribers who receive the payload, too.

mattkrick:
Next issue here: https://github.com/ParabolInc/parabol/issues/6055

PR #6012, opened 2022-02-08
fix gql executor with bad queries in dev

mattkrick:
Fixes https://github.com/ParabolInc/action-devops/issues/80

Repro is found in issue above


Dschoordsch (approved):
Works, I also checked the restart issue and it appears only on dev, so we're fine.

PR #6014, opened 2022-02-08
chore: Add orgId to IntegrationProvider

Dschoordsch:
Integration providers are added with different scopes. Currently it is
cumbersome to query for org scope since only the `teamId` is part of the
integration. This migration would add the `orgId` for org scope and add
constraints that the correct id is set depending on scope.

## Summary by CodeRabbit

- **New Features**
  - Added support for organisation identifiers in integration provider schemas, allowing for broader data handling.
  - Introduced optional organisation identifier in the input types for integration provider creation.
  - Enhanced subscription capabilities by including updates specific to integration provider success.
  - Expanded component functionality to accommodate team-specific identifiers in the provider details.

- **Bug Fixes**
  - Improved error handling for cases with invalid organisational data during integration processes.

- **Improvements**
  - Streamlined authorisation checks to ensure only appropriate users can manage integration providers based on scope.
  - Updated GraphQL schema to reflect nullable team identifiers, enhancing the flexibility of integration provider configurations.
  - Removed obsolete fields to simplify data structure in GraphQL responses.


packages/server/postgres/migrations/1643386252698_orgIdForIntegrationProvider.ts
```diff
@@ -0,0 +1,58 @@
...
+        "teamId" = NULL
+      WHERE "scope" = 'org';
+    UPDATE "IntegrationProvider"
+      SET
+        "teamId" = NULL
+      WHERE "scope" = 'global';
+    ALTER TABLE "IntegrationProvider"
+      ADD CONSTRAINT "scope_org_has_only_orgId" CHECK
+        ("scope" <> 'org' OR ("orgId" IS NOT NULL AND "teamId" IS NULL)),
+      ADD CONSTRAINT "scope_team_has_only_reamId" CHECK
```
BartoszJarocki, line 25: `teamId` instead of `reamId`

packages/server/postgres/migrations/1643386252698_orgIdForIntegrationProvider.ts
```diff
@@ -0,0 +1,58 @@
...
+
+export async function down() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+  DO $$
+  BEGIN
+    ALTER TABLE "IntegrationProvider"
+      DROP CONSTRAINT "scope_org_has_only_orgId",
+      DROP CONSTRAINT "scope_team_has_only_reamId",
```
BartoszJarocki, line 42: `teamId` instead of `reamId`

BartoszJarocki (approved):
Just two typos otherwise it looks really good! I like how easy it is now to understand the relation between the team, org, and global scopes, nicely done 👌🏻

Dschoordsch:
@BartoszJarocki could you take a look again?

Dschoordsch:
@BartoszJarocki thanks for finding this bug! I was missing a unique constraint. I tested with adding and updating a mattermost webhook and adding and removing Jira server auth.

mattkrick:
Teams can change orgs 🙂 

Today, if a team changes an org:
- The provider moves with the team to the new org
- teamMemberIntegrationAuths keep using the same provider (perhaps this is wrong, but i had to get that massive gitlab PR merged)
- tasks that are integrated with that provider still load (because they use the teamMemberIntegrationAuths which uses the old provider)

IMO, anytime the ER model goes full circle with two 1:Ms on the same entity, things get messy 
image

That said, if you prefer updating `moveTeamToOrg`, we can do that!
Also if there's a good place to store this reasoning in the repo let me know. It'd be good to have it stored there to explain why it's built the way it is.

Dschoordsch:
@mattkrick
* I don't think the ER circle is an issue because both half-circles are mutually exclusive by constraints
* Moving the organization wide integration with the team is a bit counter-intuitive for me. It really depends on the UI/UX if that is sensible. If we keep all the current UI and just add a "Make available to whole organization" checkbox while integrating, then it seems reasonable. If on the other hand we have a separate integrations menu under organization, then it would be more sensible to associate it with the org directly.

mattkrick:
Makes sense, the UI is gonna be real important with regards to how intuitive sharing feels.
I imagine integrations will be owned by a team & shared with the rest of their org, just like your first option.

The closest design we have is how we do templates today:
Screen Shot 2022-03-23 at 4 36 07 PM

mattkrick:
@Dschoordsch doing some PR housekeeping. would you like to keep this one open?

mattkrick:
pinging @Dschoordsch on next steps for this. shall we close this but leave the branch alive so we can pick it back up when we prioritize?

Dschoordsch:
There is no good way of attaching an org scoped IntegrationProvider to any specific team from the Organization Settings. I think this indicates that this is necessary after all. I think you're right about a team moving orgs and I agree the currently team scoped integration providers (MS Teams and Mattermost webhook providers) should move with the team.
The usually org scoped providers Jira Server and soon GitLab however are setup by some admin role for the organization and thus should stay with the org.
I think in this case the architecture should follow the UI/UX requirements and allow attaching an IntegrationProvider to an org directly without an intermediate team.

Issue #5993, opened 2022-02-04
Jira Server: push task menu


Dschoordsch:
Requires:
* #5094
* #5817 
## Acceptance criteria
* [ ] Implement getProjects` in [`JiraServerRestManager`](https://github.com/ParabolInc/parabol/pull/5881/files#diff-3fb777e72fa40efad485860e18c797bbfe957b627dd37fe3b578921b6fb93e85)
* [ ] add [`RepoIntegration`](https://github.com/ParabolInc/parabol/pull/5883/files#diff-f672217766433fb8b2155e38cab685b0ccebac51b133486739d1f042ea25648d) for Jira Server
* [ ] add necessary UI changes so the projects show up in the integration menu

Dschoordsch:
**Include in next sprint?: Y**
  [See the discussion](https://action.parabol.co/meet/aGjmYpJBIs/estimate/36) in Planning for Sprint #​​96

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

elielytz:
@Dschoordsch how can i test this?

PR #6026, opened 2022-02-09
Jira server push task to integration menu

Dschoordsch:
## Testing
* add Jira Server authentication
* in Task view open push to integration menu, it shows "Parabol" Jira Server repo

Resolves:
* #5993

Needs:
* #5094


packages/server/graphql/types/JiraServerRemoteProject.ts
```diff
@@ -0,0 +1,58 @@
...
+      description: 'The parabol userId this issue was fetched for'
+    },
+    name: {
+      type: new GraphQLNonNull(GraphQLString)
+    },
+    avatar: {
+      type: new GraphQLNonNull(GraphQLString),
+      resolve: async ({avatarUrls, teamId, userId}, _args: unknown, {dataLoader}) => {
+        const url = avatarUrls['48x48']
+        const auth = await dataLoader.get('freshAtlassianAuth').load({teamId, userId})
```
igorlesnenko, line 41: Looks like we are using Jira auth instead of JiraServer auth here. Is this correct?

igorlesnenko (changes requested):

packages/server/graphql/types/JiraServerRemoteProject.ts
Dschoordsch: You're right. It was just copy&paste since it's unused atm. I fixed it.

Dschoordsch:
@igorlesnenko could you take another look?

igorlesnenko (approved):
Looks good to me!

packages/client/components/TaskFooterIntegrateMenuList.tsx
```diff
@@ -103,6 +104,19 @@ const TaskFooterIntegrateMenuList = (props: Props) => {
             />
           )
         }
+        if (__typename === 'JiraServerRemoteProject') {
+          const onClick = () => {
+            // TODO
```
Dschoordsch, line 109: `createTaskIntegration` #6059 would be called here

packages/client/shared/gqlIds/IntegrationRepoId.ts
```diff
@@ -11,16 +12,22 @@ type JiraRepoIntegration = {
...
+  join: (integration: GitHubRepoIntegration | JiraRepoIntegration | RepoIntegration) => {
     const {service} = integration
     switch (service) {
       case 'github':
         return integration.nameWithOwner
       case 'jira':
         return JiraProjectId.join(integration.cloudId, integration.projectKey)
       default:
-        return ''
+        return `${integration.service}:${integration.providerId}:${integration.id}`
```
Dschoordsch, line 30: I'm uncertain about this bit. I like that with the service prefixed, we always can split it without having to pass the `service` separately, but it's redundant since it's already stored in the provider.

packages/server/graphql/types/JiraServerIntegration.ts
```diff
@@ -37,6 +43,14 @@ const JiraServerIntegration = new GraphQLObjectType<{teamId: string; userId: str
           .load({service: 'jiraServer', orgTeamIds, teamIds: [teamId]})
         return providers
       }
+    },
+    projects: {
+      type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(JiraServerRemoteProject))),
+      description:
+        'A list of projects accessible by this team member. empty if viewer is not the user',
+      resolve: async ({teamId, userId}, _args: unknown, {dataLoader}) => {
+        return fetchJiraServerProjects(teamId, userId, dataLoader)
```
mattkrick, line 52: -1 `fetchJiraServerProjects` looks like a dataloader, especially considering how expensive it might be. 
perhaps refactor into a `jiraServerProjects` dataloader?

packages/server/graphql/queries/helpers/fetchJiraServerProjects.ts
```diff
@@ -0,0 +1,36 @@
+import {DataLoaderWorker} from './../../graphql'
+import JiraServerRestManager from '../../../integrations/jiraServer/JiraServerRestManager'
+
+const fetchJiraServerProjects = async (
```
mattkrick, line 4: -1 let's turn this whole thing into a dataloader!

packages/server/graphql/queries/helpers/fetchAllRepoIntegrations.ts
```diff
@@ -11,13 +12,15 @@ const fetchAllRepoIntegrations = async (
...
-  const [prevRepoIntegrations, jiraProjects, githubRepos] = await Promise.all([
+  const [prevRepoIntegrations, jiraProjects, githubRepos, jiraServerProjects] = await Promise.all([
     getPrevRepoIntegrations(userId, teamId, permLookup),
     dataLoader.get('allJiraProjects').load({teamId, userId}),
-    fetchGitHubRepos(teamId, userId, dataLoader, context, info)
+    fetchGitHubRepos(teamId, userId, dataLoader, context, info),
+    fetchJiraServerProjects(teamId, userId, dataLoader)
   ])
-  const fetchedRepoIntegrations = [...jiraProjects, ...githubRepos]
+  const fetchedRepoIntegrations = [...(jiraProjects ?? []), ...githubRepos, ...jiraServerProjects]
```
mattkrick, line 23: +1 jiraProjects is guaranteed to be an array now so no need for the nullish coalescing

packages/server/graphql/rootSchema.ts
```diff
@@ -73,6 +73,9 @@ const withNestedSchema = mergeSchemas({
   typeDefs: `
     type _xGitHubIssue implements TaskIntegration
     type _xGitHubRepository implements RepoIntegration
+    extend type _xGitHubRepository {
```
mattkrick, line 76: -2 this would be SO COOL, but unfortunately nest-graphql-endpoint isn't smart enough to remove `service` from the request to github, so it requests that field from github & github immediately returns an error (you can reproduce it by integrating github on a team, creating a new team & then trying to push that task to github).

mattkrick (changes requested):
-2 this looks awesome! creating a separate graphql object for jira server projects was the right call. Everything else looking good minus a couple exceptions

packages/server/graphql/rootSchema.ts
Dschoordsch: I could not reproduce it like you described. I got an unrelated error which was fixed after merging in master. Right now  we're not querying the nested schema directly, so the resolver in line 89 is called and everything is fine. I prepared a fix  https://github.com/ParabolInc/nest-graphql-endpoint/pull/8 so I can keep this field.

packages/client/components/RepoIntegrationJiraServerMenuItem.tsx
```diff
@@ -0,0 +1,42 @@
...
+          </RepoIntegrationMenuItemAvatar>
+          <TypeAheadLabel query={query} label={name} />
+        </MenuItemLabel>
+      }
+      onClick={onClick}
+    />
+  )
+})
+
+export default createFragmentContainer(RepoIntegrationJiraServerMenuItem, {
```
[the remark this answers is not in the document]
Dschoordsch: I always forget to change this when copy&pasting existing components

packages/server/graphql/types/JiraServerRemoteProject.ts
```diff
@@ -0,0 +1,66 @@
...
+        if (!auth) return defaultJiraProjectAvatar
+        const provider = await dataLoader.get('integrationProviders').loadNonNull(auth.providerId)
+        const manager = new JiraServerRestManager(
+          provider.serverBaseUrl,
+          provider.consumerKey,
+          provider.consumerSecret,
+          auth.accessToken,
+          auth.accessTokenSecret
+        )
+        const avatar = await manager.getProjectAvatar(url)
```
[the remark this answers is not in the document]
Dschoordsch: This is unused at the moment, so let's convert this separately

packages/server/integrations/jiraServer/JiraServerRestManager.ts
```diff
@@ -30,6 +49,13 @@ export default class JiraServerOAuth1Manager {
     }
   }
 
+  readError(json: any) {
+    if (json.id === 'https://docs.atlassian.com/jira/REST/schema/error-collection#') {
```
mattkrick, line 53: 😳 i have no idea how to test this, but if you're confident in it i like it!

packages/server/integrations/jiraServer/JiraServerRestManager.ts
```diff
@@ -46,7 +72,34 @@ export default class JiraServerOAuth1Manager {
...
+        `Fetching projects failed with status ${response.status}, ${this.readError(json)}`
+      )
+    }
+
+    return json
+  }
+
+  async getProjects(): Promise<JiraServerRestProject[] | Error> {
+    const response = await this.request('GET', '/rest/api/latest/project')
+    const projects = this.parseJsonResponse(response)
```
mattkrick, line 92: +1 doesn't change anything, but may be nice to `await` this or call it `projectsPromise`

packages/server/integrations/jiraServer/JiraServerRestManager.ts
```diff
@@ -46,7 +72,34 @@ export default class JiraServerOAuth1Manager {
     return response
   }
 
-  async getProjects() {
-    return this.request('GET', '/rest/api/latest/project')
+  async parseJsonResponse(response: Response) {
```
mattkrick, line 75: +1 I really like this!
May be convenient to make this a generic so it doesn't return an implicit `any` & we don't need to explicitly declare a return type on L90.

e.g.

```ts
  const projects = this.parseJsonResponse<JiraServerRestProject[]>(response) // Error | JiraServerRestProject[]
```

mattkrick (approved):
+1 LGTM!

PR #6028, opened 2022-02-09
Bug fix: dragging issue in sidebar estimate section may cause duplicates

### LINKED ISSUE #5997: Poker: Dragging Scoped items in Estimate agenda unexpected behavior
## Issue - Bug

See this video: https://www.loom.com/share/8fd2cc36a9eb4cec946d6eb94f8a1b8d

Repro steps:

1. Scope at a few issues (repro'd with ~10)
2. Drag an issue from the middle of the list upward in the list, note duplication
3. Drag issue down, note duplicate issue _may_ go away

- **How serious is the bug?** P2

Users can:
   - Reprioritize issues without duplication

tianrunhe:
Fixes #5997 .

### Tests:

*   [ ] Start a poker meeting, select a couple of issues and drag issues on the sidebar. One should **not** see duplicates created
*   [ ] When starting a poker meeting, choose an template with more than 1 dimensions, e.g., "Weighted Shortest Job First". Performs the above mentioned actions and validate the behavior. Also, the ordering of the dimensions should be preserved no matter how you drag the issue.

### Implementation details: 

Each dimension is a stage itself. The bug happened because in dragging we only updated one of the stages for a dimension. The fix is to update **all** stages with the same issue. Also, to make the ordering of stages deterministic, we first compare the `sortOrder`, if those are same (when there are more than 1 dimensions), we then compare `dimensionRefIdx`.

@nickoferrall Can you please take a look? Thanks!


nickoferrall (changes requested):
Great fix! While testing, I noticed a bug when dropping estimate issues in their original position. It doesn't look like it's related to this bug though: https://github.com/ParabolInc/parabol/issues/6029

nickoferrall (approved):
LGTM!

packages/server/graphql/mutations/dragEstimatingTask.ts
```diff
@@ -48,16 +45,22 @@ export default {
...
-      return standardError(new Error('Meeting stage not found'), {userId: viewerId})
+    const draggedStages = stages.filter((stage) => stageIds.includes(stage.id))
+    if (!draggedStages.length) {
+      return standardError(new Error('No meeting stages were found'), {userId: viewerId})
     }
 
     // RESOLUTION
     // MUTATIVE
-    draggedStage.sortOrder = sortOrder
+    draggedStages.forEach((draggedStage) => (draggedStage.sortOrder = sortOrder))
```
mattkrick, line 55: can use same logic as in the client. the noise will be different, but the order will be the same

mattkrick (changes requested):

packages/client/components/PokerSidebarEstimateSection.tsx
```diff
@@ -78,19 +78,20 @@ const PokerSidebarEstimateSection = (props: Props) => {
...
+      sortOrder = destinationTopic.sortOrderFirstDimension - SORT_STEP + dndNoise()
     } else if (destination.index === stageSummaries!.length - 1) {
-      sortOrder = destinationTopic.sortOrder + SORT_STEP + dndNoise()
+      sortOrder = destinationTopic.sortOrderLastDimension + SORT_STEP + dndNoise()
     } else {
       const offset = source.index > destination.index ? -1 : 1
       sortOrder =
-        (stageSummaries[destination.index + offset]!.sortOrder + destinationTopic.sortOrder) / 2 +
+        (stageSummaries[destination.index + offset]!.sortOrderLastDimension + destinationTopic.sortOrderFirstDimension) / 2 +
         dndNoise()
```
tianrunhe, line 88: I need to do this because the old way would break when number of dimensions > 2.

packages/server/graphql/mutations/dragEstimatingTask.ts
```diff
@@ -48,14 +45,19 @@ export default {
...
+    const draggedStages = stages.filter((stage) => stage.taskId === taskId)
+    if (!draggedStages.length) {
+      return standardError(new Error('No meeting stages were found'), {userId: viewerId})
     }
+    const stageIds = draggedStages.map((stage) => stage.id)
 
     // RESOLUTION
     // MUTATIVE
-    draggedStage.sortOrder = sortOrder
+    const noise = Math.random() / 1e10
```
tianrunhe, line 56: We need a small positive number as increment to preserve the dimension ordering. The `dndNoise()` produces both negative and positive numbers.

packages/server/graphql/types/DragEstimatingTaskPayload.ts
```diff
@@ -17,12 +19,22 @@ export const DragEstimatingTaskSuccess = new GraphQLObjectType<any, GQLContext>(
...
-      resolve: resolveStage('ESTIMATE')
+    stages: {
+      type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(EstimateStage))),
+      resolve: async ({meetingId, stageIds}, _args: unknown, {dataLoader}) => {
+        const meeting = await dataLoader.get('newMeetings').load(meetingId)
+        const {phases, teamId} = meeting
+        const phase = phases.find((phase: GenericMeetingPhase) => phase.phaseType === 'ESTIMATE')!
+        const {stages} = phase
+        const dbStages = stages.filter((stage: GenericMeetingStage) => stageIds.includes(stage.id))
+        dbStages.forEach((dbStage: GenericMeetingStage) =>
```
mattkrick, line 33: -1 does this work? augmentDBStage is a pure function & you're running `forEach` instead of `map`

packages/server/graphql/mutations/dragEstimatingTask.ts
mattkrick: -1 this is an interesting constraint!
dndNoise is small enough to let someone drop 1 item between 2 items up to ~32 times before it got too small, which was probably fine.
with a larger noise coefficient + multiple items safely reorder `Math.floor(10 / dimensionCount)` times. That might not be enough for real world usage. I'll check my math & try to break it in testing. We may just have to update every succeeding stage sortOrder. I don't love the O(log(n)) updates, but better than running into that really weird bug again

tianrunhe: I knew this will get your attention! What I found out before is that sometimes the number generated from `dndNoise` with `Math.random() / 1e15` is too small it's considered equivalently as zero. A series of output from my experimental tests:

Screen Shot 2022-02-18 at 11 43 10 AM

mattkrick: interesting! that's a good enough reason to move away from a float sort.
what do you think about moving to an integer-based sort? 
discussion here: https://github.com/ParabolInc/parabol/issues/6086

mattkrick:
Dragging haphazardly i got this, but can't reliably reproduce yet... the fact that react-swipeable-views is involved makes me think that we need to tell it to ignore the sidebar somewhere...

Screen Shot 2022-02-28 at 5 25 00 PM

mattkrick:
TEST
- [ ] drag an item & drop it on the facilitator (or anywhere outside the list). it goes back home without an error

packages/client/utils/constants.ts
```diff
@@ -46,6 +46,7 @@ export const AGENDA_ITEM = 'agendaItem'
 
 /* Sorting */
 export const SORT_STEP = 1
+export const ESTIMATE_TASK_SORT_ORDER = 2 ** 43
```
mattkrick, line 49: +1 i think you've created a really great pattern here by hiding the sortOrder complexity from the client! since we'll probably use this in the future perhaps we name is something like `INTEGER_SORT_STEP` & add a comment telling future people to use this one instead of the sort_step above?

packages/server/graphql/mutations/dragEstimatingTask.ts
```diff
@@ -48,14 +51,36 @@ export default {
...
+    const stageIds = draggedStages.map((stage) => stage.id)
 
     // RESOLUTION
     // MUTATIVE
-    draggedStage.sortOrder = sortOrder
+    const numberOfDimensions = Math.floor(stages.length / numberOfTasks)
+    const oldPositionIndex = taskIds.indexOf(taskId) / numberOfDimensions
+    let sortOrder
+    if (newPositionIndex === 0) {
+      sortOrder = Math.floor(stages[0].sortOrder / 2)
```
mattkrick, line 71: -1 integers in JS & postgres are signed, which means if we subtract ESTIMATE_TASK_SORT_ORDER  vs. halving it from the previous value we'll get double the number of drags!

packages/server/graphql/mutations/dragEstimatingTask.ts
```diff
@@ -48,14 +51,36 @@ export default {
...
+    } else if (newPositionIndex === numberOfTasks - 1) {
+      sortOrder = stages[stages.length - 1].sortOrder + ESTIMATE_TASK_SORT_ORDER
+    } else {
+      const offSet = oldPositionIndex > newPositionIndex ? 0 : 1
+      const prevStage = stages[(newPositionIndex + offSet) * numberOfDimensions - 1]
+      const nextStage = stages[(newPositionIndex + offSet) * numberOfDimensions]
+      sortOrder = Math.floor((prevStage.sortOrder + nextStage.sortOrder) / 2)
```
mattkrick, line 78: -1 this is the only bummer with the client giving us an index vs. it giving us an integer sort order-- we need to find the previous & next stages! Here it's easy because stages live on the Meeting object. If this were an agenda item (or comment, task, reflection, template dimension, template scale, etc.) where each item is its own row in the DB, we'd have to fetch ALL the items to determine where to put it. for those cases, it'd be more efficient to send a single query to the DB that updates every item where the index is greater than this one.

should we leave this as is & worry about setting a correct pattern later or create the new pattern now? up to you!

mattkrick (changes requested):

packages/server/graphql/mutations/dragEstimatingTask.ts
```diff
@@ -7,24 +7,30 @@ import publish from '../../utils/publish'
...
       type: new GraphQLNonNull(GraphQLID)
     },
-    stageId: {
+    taskId: {
       type: new GraphQLNonNull(GraphQLID)
     },
-    sortOrder: {
-      type: new GraphQLNonNull(GraphQLFloat)
+    newPositionIndex: {
+      description: 'The index of the tasks will be moved to',
```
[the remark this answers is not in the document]
tianrunhe: I feel mentioning `index / dimensionCount` is even more confusing an unnecessary? From client side, they see 3 tasks on the siderbar, they wanna move the 3rd task to the middle, so they put `newPositionIndex = 1` (0-based indexing). They don't need to think about number of dimensions, do they?

packages/server/graphql/mutations/dragEstimatingTask.ts
tianrunhe: I think we only need to worry about using this integer approach because we need to deal with dimensions, i.e., when we need to move N items and put them in between of two elements and also preserve the ordering. For single dimension items like agenda items, when you do `-SORT_STEP`, `+SORT_STEP` or `(a+b)/2` it's guaranteed the newly generated order is in right position. So I feel it'd be a while before we have another use case that needs this approach so I choose to not optimize earlier.

packages/client/utils/constants.ts
tianrunhe: For the same reasoning as I mentioned [below](https://github.com/ParabolInc/parabol/pull/6028#discussion_r817258215), let's leave it till we have another use case of this approach. WDYT?

tianrunhe:
> Dragging haphazardly i got this, but can't reliably reproduce yet... the fact that react-swipeable-views is involved makes me think that we need to tell it to ignore the sidebar somewhere...
> 
> Screen Shot 2022-02-28 at 5 25 00 PM

This is fixed in [PR #6122](https://github.com/ParabolInc/parabol/pull/6122)

packages/server/graphql/mutations/dragEstimatingTask.ts
mattkrick: i'm not sure i follow.
could you explain how you could change the sort order of an agenda item without fetching all the agenda items in the meeting?

PR #6030, opened 2022-02-10
Yarn audit at pre-commit to block commit if moderate or above vulnerabilities found

adaniels-parabol:
Implements yarn audit during the pre-commit stage. This will block commits if moderate or above vulnerabilities found. This will aid in keeping moderate and above (high, critical) out of our code base. This also has the benefit of reducing security findings during Ironbank vulnerability scan as yarn audit is one security feed used for vulnerability findings.

The goal is to have developers fix findings as they go. Developers can always override the pre-commit hook should an emergency push be needed (e.g. in the event of a hot fix) where a vulnerability is above the moderate level.

Failure results in an output of the vulnerability with a link to the fix/documentation. Success results in a continued commit.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -7,7 +7,7 @@ vulnerabilities=$?
 if [ $vulnerabilities -eq 0 ]; then
   echo "No vulnerabilities found."
 else
-  echo "Vulnerabilities moderate or above found. Resolve issues before commiting code. See console output for more details."
+  echo "Vulnerabilities moderate or above found. Resolve issues before committing code. See console output for more details."
   exit 1
 fi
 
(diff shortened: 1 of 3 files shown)


adaniels-parabol:
@mattkrick I feel like I could merge this one in, but it does impact committing code if moderate or above vulnerabilities are found. Thus I wanted to get your LGTM.

mattkrick:
IMO I'd run this as a github workflow so it'd throw up a red flag before merging to master vs. on every local commit.
Maybe put it here? https://github.com/ParabolInc/parabol/blob/7be22aa9856e85e5c284fa06c5bad88253e31822/.github/workflows/typecheck.yml


thinking aloud, if a new vuln pops up, that means every dev is gonna have to fix that vuln on their own before they can commit, and we'll get a bunch of commits that also happen to include fixing a vuln 😅

Dschoordsch:
Closing in favour of #6489

PR #5974, opened 2022-01-27
Bump samlify, remove unused server deps, remove gql2ts (20%)

mattkrick:
**This is low priority.**
I had an extra hour this morning & wanted to do some cleaning up. 

What it does:
- Bumps samlify, which fixes a vuln in xmldom
- Removes unused dependencies from server/package.json. This helps with the privately managed instances. Fewer deps = fewer vulns 
- Removes gql2ts. We haven't used this in a couple years, but we were still building gql types every time we started up the dev server. Faster dev server startups!
- Refactors s3.js into the S3Manager. s3.js wasn't in typescript, 80% unused, and started an s3 instance when the server booted up.

TEST
- [ ] can update your profile picture on staging (to test that the S3Manager still works)

PR #6035, opened 2022-02-11
Remove graphql.ts

Dschoordsch:
graphql.ts is no longer generated since #5974. Remove references to it.


tianrunhe (changes requested):
I got this error when running `yarn dev` and no pages can be opened...?  

```
ERROR in ./packages/server/utils/predicates.ts 3:25
Module parse failed: Unexpected token (3:25)
File was processed with these loaders:
 * ./node_modules/@pmmmwh/react-refresh-webpack-plugin/loader/index.js
You may need an additional loader to handle the result of these loaders.
| __webpack_require__.$Refresh$.runtime = require('/Users/brucetian/Projects/parabol/node_modules/react-refresh/runtime.js');
| 
> export const isNotNull = <T>(item: T | null | undefined): item is T =>
|   item !== null && item !== undefined
| 
webpack 5.64.1 compiled with 1 error in 5515 ms
```

Dschoordsch:
@tianrunhe I fixed the issue. It was caused by including a `.ts` file from server into client. Fixed by moving it to client.

tianrunhe (approved):
LGTM!

mattkrick:
🎉

Issue #6037, opened 2022-02-11
cloudProvider can be undefined


nickoferrall:
Affected 80 users in the past two days. 

https://sentry.io/organizations/parabol/issues/3003222297/?referrer=github_plugin

```
TypeError: Cannot read properties of undefined (reading 'cloudProvider')
  at fn (./packages/client/components/ScopePhaseArea.tsx:82:57)
  at _optionalChain (./packages/client/components/ScopePhaseArea.tsx:1:437)
  at c (./packages/client/components/ScopePhaseArea.tsx:82:5)
...
(10 additional frame(s) were not displayed)
```

PR #6038, opened 2022-02-11
Make gitlab integration nullable

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6037

`TeamMemberIntegrations.gitlab` should be non-nullable because `sharedProviders` is non-nullable. However, [this Sentry issue](https://sentry.io/organizations/parabol/issues/3003222297/?project=107196&referrer=github_plugin) argues differently. 

I was able to reproduce the bug in prod by going to an old Poker meeting. After refreshing the page, I was unable to reproduce it again.

In this PR, I've made gitlab nullable and I'm capturing the issue in Senty if `sharedProviders` is null.

### DIFF
--- a/packages/server/graphql/types/GitLabIntegration.ts
+++ b/packages/server/graphql/types/GitLabIntegration.ts
@@ -1,5 +1,4 @@
 import {GraphQLList, GraphQLNonNull, GraphQLObjectType} from 'graphql'
-import sendToSentry from '../../utils/sendToSentry'
 import {GQLContext} from '../graphql'
 import IntegrationProviderOAuth2 from './IntegrationProviderOAuth2'
 import TeamMemberIntegrationAuthOAuth2 from './TeamMemberIntegrationAuthOAuth2'
@@ -35,25 +34,14 @@ const GitLabIntegration = new GraphQLObjectType<any, GQLContext>({
     sharedProviders: {
       description: 'The non-global providers shared with the team or organization',
       type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(IntegrationProviderOAuth2))),
-      resolve: async (
-        {teamId, userId}: {teamId: string; userId: string},
-        _args: unknown,
-        {dataLoader}
-      ) => {
+      resolve: async ({teamId}: {teamId: string}, _args: unknown, {dataLoader}) => {
         const team = await dataLoader.get('teams').loadNonNull(teamId)
         const {orgId} = team
         const orgTeams = await dataLoader.get('teamsByOrgIds').load(orgId)
         const orgTeamIds = orgTeams.map(({id}) => id)
-        const sharedProviders = await dataLoader
+        return dataLoader
           .get('sharedIntegrationProviders')
           .load({service: 'gitlab', orgTeamIds, teamIds: [teamId]})
-        if (!sharedProviders) {
-          sendToSentry(new Error('Shared providers is undefined'), {
-            userId,
-            tags: {teamId, orgTeamIds: JSON.stringify(orgTeamIds), userId}
-          })
-        }
-        return sharedProviders
       }
     }
     // The GitLab schema get injected here as 'api'
--- a/packages/server/graphql/types/GitLabIntegration.ts
+++ b/packages/server/graphql/types/GitLabIntegration.ts
@@ -53,6 +53,7 @@ const GitLabIntegration = new GraphQLObjectType<any, GQLContext>({
             tags: {teamId, orgTeamIds: JSON.stringify(orgTeamIds), userId}
           })
         }
+        return sharedProviders
       }
     }
     // The GitLab schema get injected here as 'api'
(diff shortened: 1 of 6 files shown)


nickoferrall:
@tianrunhe could you review this PR, please?

nickoferrall:
Hey @tianrunhe, just pinging you to keep this top of mind. I know I only tagged you on Friday, but I'd love to squeeze this into this week's Ship if possible!

packages/server/graphql/types/GitLabIntegration.ts
```diff
@@ -34,14 +35,25 @@ const GitLabIntegration = new GraphQLObjectType<any, GQLContext>({
...
+      ) => {
         const team = await dataLoader.get('teams').loadNonNull(teamId)
         const {orgId} = team
         const orgTeams = await dataLoader.get('teamsByOrgIds').load(orgId)
         const orgTeamIds = orgTeams.map(({id}) => id)
-        return dataLoader
+        const sharedProviders = await dataLoader
           .get('sharedIntegrationProviders')
           .load({service: 'gitlab', orgTeamIds, teamIds: [teamId]})
+        if (!sharedProviders) {
```
tianrunhe, line 50: `sharedProviders` should always be non-null here, no? I thought if there are no results matching the query, it will return an empty array.

tianrunhe (changes requested):

packages/server/graphql/types/GitLabIntegration.ts
nickoferrall: It should be but that bug in Sentry suggests that `gitlab` can be null which suggests that `sharedProviders` must be null too

tianrunhe: hmmm... I bet if we will ever see this new error in Sentry

tianrunhe (approved):
Approving for not showing error page on the front end for our user but I suspect if this is helpful to root cause the issue. This is a weird one. It seems to only happen when assessing an old poker meeting opened before 6.46.0 so maybe caching or migration issue?

nickoferrall:
Copying [this comment](https://parabol.slack.com/archives/C836NA350/p1644939269568349?thread_ts=1644864683.325929&cid=C836NA350) from Slack here for more context.

The Sentry error tells us that gitlab can be undefined. gitlab is non-nullable because sharedProviders is non-nullable. [This](https://app.datadoghq.com/rum/replay/sessions/20ecc7a1-f839-4a87-9c8f-4ca4a25a2596?from=1644807973572) & [this](https://app.datadoghq.com/rum/replay/sessions/2ec13502-d501-4bab-a5c6-23807634ac13?from=1644822902301) datadog session recording show that the Scope phase can crash because gitlab can be undefined, but if the user refreshes the page, it works.

As Bruce suggested above, I think there’s a caching issue. There was a bug in `SharedProviders` which was fixed in the [GitLab oauth PR](https://github.com/ParabolInc/parabol/pull/5973/files) that was shipped last week. `SharedProviders` has always had a non-nullable type, but until that PR, it would return null. We were just not calling it on the frontend.

Now we are calling it on the frontend, but I suspect its initially using the old server code until the user refreshes. 

So, I’ve made gitlab nullable, safeguarded the frontend, and I’m sending an error to Sentry if SharedProviders  is null. If there are Sentry errors, there’s an issue that’s not related to caching the old server code. If there are no Sentry errors, I suggest reverting the code and making gitlab non-null. 

I'd be happy to create an issue to make gitlab non-null if you agree with my thinking, @mattkrick.

mattkrick:
> The Sentry error tells us that gitlab can be undefined. gitlab is non-nullable because sharedProviders is non-nullable.

Non-Nullable fields can exist in a nullable object.  For example:

```gql
user(name: "matt") {
  bestFriend {
    name
  }
}
```

i may not have a best friend, but if i do, they definitely have a name.


> Root cause

The spirit of this PR is good, but if there's a contract stating that gitlab is nonnull, then we should focus effort on why that contract isn't being honored instead of guarding against it.

Since the server guarantees that gitlab is nonnull, i suspect the problem must be an optimisticUpdater (hence our developer FFR 😉 )

A quick search shows this might be the culprit: https://github.com/ParabolInc/parabol/blob/0f149354f886da275cf61c6bb3086c9eb765fe80/packages/client/mutations/JoinMeetingMutation.ts#L57-L61

Assuming this is the problem, can we reproduce? Is that optimsticUpdater more trouble than it's worth & we just get rid of it? Are there other optimisticUpdaters that could be borking it?

nickoferrall:
Thanks, @mattkrick! 

You're right, the optimistic updater was causing the bug. We could add temporary GitLab & Mattermost fields, but removing the optimistic updater seems like the right call. It's difficult to notice any difference when joining a meeting without the optimistic updater, and keeping it might cause more bugs down the line. 

I've checked for similar optimistic updaters that could cause the bug and it looks like this is the only one.

mattkrick:
simple is best! LGTM

PR #6042, opened 2022-02-12
fix unresponsive github poker meeting

### LINKED ISSUE #6015: Poker meeting crashing
## Issue - Bug

One of our [poker meeting](https://action.parabol.co/meet/aGjmYpJBIs) isn't loading. 
Investigate what's the root cause and figure out the next steps.

related slack thread: https://parabol.slack.com/archives/C836NA350/p1644328590076379

- [ ] root cause identified 

### LINKED ISSUE #6023: Reduce fetching estimate labels from GitHub
One of the identified causes of #6015 is that we are requesting too much from GitHub at once & they are slow to respond (not rate limiting us, which i think is interesting!)

In that poker meeting, there are about 40 requests going to github requesting all the labels attached to each issue. Those requests aren't necessary! We should only be fetching estimate updates a client is fetching estimates. In other words, all that logic should live in the `Task.estimates` resolver

AC

- [ ] For GitHub, move all estimate upserting logic that currently lives in the `Task.integration` into the `Task.estimates` resolver.
- [ ] Move the DB query outside of the estimates.forEach clause so the query only happens once instead of once per estimate (cache it, don't call it eagerly). There may be other low hanging fruit here. Call that query as little possible. Great find @igorlesnenko! https://github.com/ParabolInc/parabol/blob/d681cf1e283f4bc497d456d96bcdedc393e1a49a/packages/server/graphql/types/Task.ts#L152-L156

Estimate: 4 hours

### LINKED ISSUE #6024: Improve batches & deduping inside nest-graphql-endpoint
While debugging #6015 one of the identified root causes was nest-graphql-endpoint is including multiple identical selection sets that each have different aliases. This should not be! if 2 fields have identical selection sets, then they should share the same alias (or selection name, if no alias is needed).

AC
- [ ] Look at the JSON response from GitHub. For each entry, if the key is an alias of another key, then the values are different (no duplicate values)

Estimate: 16 hours

mattkrick:
Fix #6024
Fix #6015 
Fix #6023

Our big poker meeting shined the light on some big bugs!

- If 2 fragments have a field with the same name but a different type, then 1 or both must be aliased. `_xGitHubIssue.url` is a `_xGitHubURI` whereas `JiraIssue.url` is a `string`. Since we use graphql-jit, our server technically doesn't perform this validation so we could keep getting away with it, but while debugging I like to remove graphql-jit to get better stack traces so I figured it was a good time to fix it. The same goes for a nullable vs. non-null type, as see in  `createdByUser` (since that can be null for a comment but non-null for a poll or task)
- The above bug showed me that nest-graphql-endpoint couldn't handle client-aliased fields. The workaround is to edit the default resolver for external fields that requires the use of an alias. That added a lot of complexity, so instead I aliased the JiraIssue field (since we control that resolver). However, the day may come when we'll have to tackle this, so I fixed it for the GitHubIssue.url field as a proof of concept. All that's left is to write something that creates a defaultResolver for GitHub fields. Note: other schema stitching packages face this same problem! https://github.com/ardatan/graphql-tools/issues/967
- As @igorlesnenko discovered, the `taskIdsByTeamAndGitHubRepo` query is CRAZY expensive for the parabol product team in prod. I clocked it in at 1500ms! This was getting called once per estimate per task, or a total of 42 times per meeting (in the buggy meeting). I turned this into a dataloader so it only gets called once per meeting. Then, i moved it to the `task.estimates` field so now it doesn't even run during a meeting 🎉 
- nest-graphql-endpoint had 2-3 bugs in it & some medium-hanging fruit for HUGE performance gains. It now recursively  merges fields instead of just at the root level (like https://www.graphql-tools.com/docs/batch-execution). GitHub's rate limit is calculated by multiplying all the `first, last` args it traverses in the request (https://docs.github.com/en/graphql/overview/resource-limitations). For every `first,last` we merge, we reduce the cost by that __factor__, which is huge!

### DIFF
--- a/packages/client/components/PokerEstimateHeaderCardJira.tsx
+++ b/packages/client/components/PokerEstimateHeaderCardJira.tsx
@@ -83,7 +83,7 @@ const PokerEstimateHeaderCardJira = (props: Props) => {
     setIsExpanded((isExpanded) => !isExpanded)
   }
   const isDesktop = useBreakpoint(Breakpoint.SIDEBAR_LEFT)
-  const {issueKey, summary, descriptionHTML, url} = issue
+  const {issueKey, summary, descriptionHTML, jiraUrl} = issue
   return (
     <HeaderCardWrapper isDesktop={isDesktop}>
       <HeaderCard>
@@ -100,7 +100,7 @@ const PokerEstimateHeaderCardJira = (props: Props) => {
           dangerouslySetInnerHTML={{__html: descriptionHTML}}
         />
         <StyledLink
-          href={url}
+          href={jiraUrl}
           rel='noopener noreferrer'
           target='_blank'
           title={`Jira Issue #${issueKey}`}
@@ -119,7 +119,7 @@ export default createFragmentContainer(PokerEstimateHeaderCardJira, {
       issueKey
       summary
       descriptionHTML
-      url
+      jiraUrl: url
     }
   `
 })
--- a/packages/client/components/ThreadedCommentBase.tsx
+++ b/packages/client/components/ThreadedCommentBase.tsx
@@ -62,9 +62,16 @@ const ThreadedCommentBase = (props: Props) => {
   } = props
   const isReply = !!props.isReply
   const {id: discussionId, meetingId, replyingToCommentId, teamId} = discussion
-  const {id: commentId, content, createdByUser, isActive, reactjis, threadParentId} = comment
+  const {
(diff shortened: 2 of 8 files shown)


mattkrick:
yep, pulling estimates is hopefully coming in the future, maybe showing folks the estimate on the Task card itself or pulling existing estimates during the estimate phase 🤷

Issue #5994, opened 2022-02-04
Jira Server: show integrated task


Dschoordsch:
Requires:
* #5810 
* #5993 

## Acceptance criteria
* [ ] call `createTaskIntegration` from the `TaskIntegrationMenu` for Jira Server
* [ ] add Jira Server to [`integration`](https://github.com/ParabolInc/parabol/blob/6893c722d92f4a3b0daed839c0b0e9aa977f2f4d/packages/server/graphql/types/Task.ts#L78) field on `Task`
* [ ] adapt UI to show integrated task

Dschoordsch:
**Include in next sprint?: Y**
  [See the discussion](https://action.parabol.co/meet/aGjmYpJBIs/estimate/37) in Planning for Sprint #​​96

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

mattkrick:
closing in favor of #6059

PR #6059, opened 2022-02-15
feat(jira-server): push a task to Jira Server integration

### LINKED ISSUE #5810: Jira Server: `createTaskIntegration`
Needs #5819
Relates to #5094

## New acceptance criteria
Let's reduce this in scope a bit and create a separate issue for UI and retrieving the `Task` content from Jira Server.
* [ ] add `createIssue` and `addComment` to [`JiraServerRestManager`](https://github.com/ParabolInc/parabol/pull/5881/files#diff-3fb777e72fa40efad485860e18c797bbfe957b627dd37fe3b578921b6fb93e85)
* [ ] create new `JiraServerTaskIntegrationManager` and add to `TaskIntegrationManagerFactory`
* [ ] create `JiraServerIssueId` and store data on `Task`
  ```
  integrationHash: JiraServerIssueId.join(providerId, projectId, issueKey)
  integration: {
    accessUserId
    service
    providerId
  }
  ```

igorlesnenko:
Resolves #5810 #5994 #6238

**How to test:**

- Add Jira Server integration
- Create a task in parabol
- Push task to Jira Server using push to integration menu
- See the task was created on Jira Server
- Add another member to a parabol team
- Login as a new member
- Create a task
- Login as a previous member
- Run the above query for the other member task
- See the task was created on Jira Server and a comment added
- Edit task in Jira Server
- Open parabol board, see the task is loaded correctly

image


packages/server/integrations/JiraServerTaskIntegrationManager.ts
```diff
@@ -0,0 +1,96 @@
...
+    rawContentStr,
+    integrationRepoId
+  }: {
+    rawContentStr: string
+    integrationRepoId: string
+  }): Promise<CreateTaskResponse> {
+    const manager = this.getManager()
+
+    const {title: summary, contentState} = splitDraftContent(rawContentStr)
+    // TODO: implement stateToJiraServerFormat
```
Dschoordsch, line 44: +1 could you find any documentation which format they use?

Dschoordsch (commented):
Did not test yet, but looks good!
You could already integrate the call side which would make testing it easier: https://github.com/ParabolInc/parabol/pull/6026/files#r814033866

packages/server/integrations/JiraServerTaskIntegrationManager.ts
igorlesnenko: @Dschoordsch created an additional issue about this #6130

igorlesnenko:
@Dschoordsch I integrated the call in the next PR: #6092. It is based on this PR, so I guess it is possible to use it to test

packages/server/integrations/jiraServer/JiraServerRestManager.ts
```diff
@@ -20,6 +20,26 @@ interface JiraServerProject {
   archived: boolean
 }
 
+interface JiraServerCreateMeta {
+  projects: Array<{
+    id: string
+    issuetypes: Array<{
```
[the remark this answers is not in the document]
igorlesnenko: @Dschoordsch no particular reason, fixed

packages/server/graphql/mutations/createTaskIntegration.ts
```diff
@@ -113,7 +113,7 @@ export default {
 
     const teamDashboardUrl = makeAppURL(appOrigin, `team/${teamId}`)
     const createdBySomeoneElseComment =
-      userId && viewerId !== userId
+      userId && viewerId !== userId && taskIntegrationManager.getCreatedBySomeoneElseComment
```
Dschoordsch, line 116: +1 you could simplify this to 
```
      userId && viewerId !== userId ? taskIntegrationManager.getCreatedBySomeoneElseComment?.(
            viewerName,
            assigneeName,
            team.name,
            teamDashboardUrl
          )
      : undefined
```
(notice the `?.`)

packages/server/graphql/mutations/createTaskIntegration.ts
```diff
@@ -134,6 +134,16 @@ export default {
       return {error: {message: createTaskResponse.message}}
     }
 
+    if (userId && viewerId !== userId && taskIntegrationManager.addCreatedBySomeoneElseComment) {
+      await taskIntegrationManager.addCreatedBySomeoneElseComment(
```
Dschoordsch, line 138: +1 same here, you could use `?.` for optional function call

Dschoordsch (changes requested):

packages/client/types/constEnums.ts
```diff
@@ -117,6 +117,7 @@ export const enum ExternalLinks {
   INTEGRATIONS_GITHUB = 'https://www.parabol.co/integrations/github',
   INTEGRATIONS_MATTERMOST = 'https://www.parabol.co/integrations/mattermost',
   INTEGRATIONS_GITLAB = 'https://www.parabol.co/integrations/gitlab',
+  INTEGRATIONS_JIRASERVER = 'https://www.parabol.co/integrations/jiraserver',
```
Dschoordsch, line 120: I created #6168 to make sure we create this page eventually

packages/server/graphql/queries/helpers/fetchJiraServerProjects.ts
```diff
@@ -0,0 +1,36 @@
+import {DataLoaderWorker} from './../../graphql'
```
Dschoordsch, line 1: -1 I think this file is a merge artefact, the file should be deleted as it's now a dataloader.

packages/server/integrations/jiraServer/JiraServerRestManager.ts
```diff
@@ -54,45 +90,104 @@ export default class JiraServerRestManager {
...
-  async request(method: string, path: string) {
+  async parseJsonResponse<T>(response: Response): Promise<T | Error> {
+    const contentType = response.headers.get('content-type') || ''
+    if (!contentType.includes('application/json')) {
+      return new Error('Received non-JSON Jira Server Response')
+    }
+    const json = await response.json()
+
+    if (response.status !== 201 && response.status !== 200) {
+      console.error(`Jira server error ${response.status}`, json)
```
Dschoordsch, line 101: +1 do we need this additional `console.error`?

Dschoordsch (changes requested):
`integrationRepoId` issue still persists. Feel free to move to maintainer review once you've addressed the comments

packages/server/integrations/jiraServer/JiraServerRestManager.ts
igorlesnenko: Removed

packages/client/components/TaskIntegrationLink.tsx
```diff
@@ -46,6 +46,7 @@ const TaskIntegrationLink = (props: Props) => {
     )
   } else if (integration.__typename === 'JiraServerIssue') {
     const {url, issueKey, projectKey} = integration
+    console.log('integration', integration)
```
Dschoordsch, line 49: -1 please remove

packages/server/graphql/mutations/createTaskIntegration.ts
```diff
@@ -42,6 +42,7 @@ export default {
     context: GQLContext,
     info: GraphQLResolveInfo
   ) => {
+    console.log('integrationRepoId', integrationRepoId)
```
Dschoordsch, line 45: -1 please remove

packages/server/integrations/JiraServerTaskIntegrationManager.ts
```diff
@@ -44,15 +45,21 @@ export default class JiraServerTaskIntegrationManager implements TaskIntegration
     // TODO: implement stateToJiraServerFormat
     const description = contentState.getPlainText()
 
-    const res = await manager.createIssue(integrationRepoId, summary, description)
+    const {repoId, providerId} = IntegrationRepoId.split(integrationRepoId)
+
+    if (parseInt(providerId ?? '', 10) !== this.provider.id || !repoId) {
```
Dschoordsch, line 50: -1 if `providerId` is supposed to be an int, `IntegrationRepoId.split` should already return an int

Dschoordsch (changes requested):

Dschoordsch (approved):

packages/server/integrations/JiraServerTaskIntegrationManager.ts
```diff
@@ -0,0 +1,108 @@
...
+    const manager = this.getManager()
+
+    const {title: summary, contentState} = splitDraftContent(rawContentStr)
+    // TODO: implement stateToJiraServerFormat
+    const description = contentState.getPlainText()
+
+    const {repoId, providerId} = IntegrationRepoId.split(integrationRepoId)
+
+    if (providerId !== this.provider.id || !repoId) {
+      throw new Error('Incorrect IntegrationRepoId')
```
mattkrick, line 51: +1 we don't need to write unit tests or achieve 100% branch coverage, but if something is unreachable, then we probably don't need to write code for it.
Here, the repoId should be nonnull & we can trust the providerId from the split

packages/server/integrations/JiraServerTaskIntegrationManager.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  private static makeCreateJiraServerTaskComment(
+    creator: string,
+    assignee: string,
+    teamName: string,
+    teamDashboardUrl: string
+  ) {
+    const sanitizedCreator = creator.replace(/#(\d+)/g, '#​\u200b$1')
+    const sanitizedAssignee = assignee.replace(/#(\d+)/g, '#​\u200b$1')
```
mattkrick, line 83: -1 these sanitizations are used for GitHub issues because github searches for the # and turns it into a link.
I couldn't get Jira Server to do anything like that, so I think these are safe to remove. If not, please add a comment to the code showing a use case why they're needed

packages/server/integrations/jiraServer/JiraServerRestManager.ts
```diff
@@ -54,45 +90,103 @@ export default class JiraServerRestManager {
...
+    return this.request<JiraServerIssue>(
+      'GET',
+      `/rest/api/latest/issue/${issueId}?expand=renderedFields`
+    )
+  }
+
+  async createIssue(projectId: string, summary: string, description: string) {
+    const meta = await this.getCreateMeta()
+    if (meta instanceof Error) {
+      throw meta
```
mattkrick, line 145: -1 your calling function in the TaskIntegrationManager is expecting the return value to possibly be an Error, but it isn't using a try/catch. all these errors should be returned & not thrown

mattkrick (changes requested):
-1 few little things, but nothing major. looking good!

packages/client/components/TaskWatermark.tsx
```diff
@@ -5,7 +5,8 @@ import JiraSVG from './JiraSVG'
 
 const iconLookup = {
   _xGitHubIssue: GitHubSVG,
-  JiraIssue: JiraSVG
+  JiraIssue: JiraSVG,
+  JiraServerIssue: JiraSVG
```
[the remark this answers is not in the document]
igorlesnenko: I fixed the logo in next PR: https://github.com/ParabolInc/parabol/pull/6157/files#diff-d419600284cc40027d3e86f70437fc58426edbd40fe307d9d6e3dafc8a29a186R42

packages/server/integrations/JiraServerTaskIntegrationManager.ts
```diff
@@ -0,0 +1,108 @@
+import JiraServerIssueId from '~/shared/gqlIds/JiraServerIssueId'
+import {CreateTaskResponse, TaskIntegrationManager} from './TaskIntegrationManagerFactory'
+import JiraServerRestManager from './jiraServer/JiraServerRestManager'
+import {IGetTeamMemberIntegrationAuthQueryResult} from '../postgres/queries/generated/getTeamMemberIntegrationAuthQuery'
+import {TIntegrationProvider} from '../postgres/queries/getIntegrationProvidersByIds'
+import splitDraftContent from '~/utils/draftjs/splitDraftContent'
+import {ExternalLinks} from '~/types/constEnums'
+import IntegrationRepoId from '~/shared/gqlIds/IntegrationRepoId'
+
+export default class JiraServerTaskIntegrationManager implements TaskIntegrationManager {
```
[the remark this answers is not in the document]
igorlesnenko: @mattkrick As I see, class in TypeScript can't extend 2 classes. 

I also think `JiraServerTaskIntegrationManager` and `JiraServerRestManager`  have different purposes.

`JiraServerTaskIntegrationManager` exposes a unified API that is defined by abstract class or interface.
All Integration managers extend same abstract class and expose same API, for example `taskIntegrationManager.createTask`

This way avoid a lot of `if` constructions here https://github.com/ParabolInc/parabol/blob/5c97d33947be89855eb9bd6c92641eee0d2ee1b9/packages/server/graphql/mutations/createTaskIntegration.ts#L125-L131

However, internally, `createTask` relies on different `createTask` APIs — for GitHub it is GraphQL mutation, for JiraServer it is `JiraServerRestManager` etc.
If I simply extend, then there will be no unified shape of manager API, unified shape of input and output.

What do you think?

igorlesnenko:
Made some fixes and little refactoring, including #6238

packages/client/components/TaskWatermark.tsx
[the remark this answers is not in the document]
igorlesnenko: I feel like for tabs in poker meetings we use an older logo. I can replace it, but not sure which logo to use. 

Here I just used the same logo, as expected there will be rarely both Jira and JiraServer integrations added. 
However, in poker meeting tabs,  2 identical logos looks weird to me.

Not sure, if there is a separate server logo exists. Maybe add some `s` badge to the logo or something?

cc @Dschoordsch

packages/server/database/types/TaskIntegrationJiraServer.ts
```diff
@@ -0,0 +1,16 @@
...
+  providerId: number
+}
+
+export default class TaskIntegrationJiraServer extends BaseTaskIntegration {
+  providerId: number
+  service!: 'jiraServer'
+  constructor(input: Input) {
+    const {accessUserId, providerId} = input
+    super({accessUserId, service: 'jiraServer'})
+    this.providerId = providerId
```
mattkrick, line 14: -1 we'll need an issueKey (and maybe projectKey?) in here, too

packages/server/graphql/mutations/createTaskIntegration.ts
```diff
@@ -134,11 +124,33 @@ export default {
...
     }
 
+    const {issueId, ...updateTaskInput} = createTaskResponse
+
+    if (
+      userId &&
+      viewerId !== userId &&
+      'addCreatedBySomeoneElseComment' in taskIntegrationManager
+    ) {
+      const addCommentResponse = await taskIntegrationManager.addCreatedBySomeoneElseComment(
```
mattkrick, line 134: +1 interesting! so instead of tagging this to the bottom of the task we're adding it as a comment?

packages/client/components/TaskWatermark.tsx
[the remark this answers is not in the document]
mattkrick: This is the logo on our instance of Jira Server, whatcha think? 
jira-software-logo-jira-software-logo-11562914188wp8r59nt10

packages/server/graphql/mutations/createTaskIntegration.ts
igorlesnenko: @mattkrick Yes, we already did that for GitHub and Jira. Probably, because adding it to the text will show it in the card on a Parabol board as well?

packages/client/components/TaskWatermark.tsx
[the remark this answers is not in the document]
igorlesnenko: @mattkrick I see this square logo both on Jira Server and Jira in the header. However, I think we can use it for Jira Server to have different logos.

packages/server/database/types/TaskIntegrationJiraServer.ts
igorlesnenko: @mattkrick now storing `issueId` and `repositoryId` here too

packages/server/integrations/JiraServerTaskIntegrationManager.ts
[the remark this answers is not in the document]
mattkrick: ahh, you're making a lot of sense. i'm all for making things non null wherever we can!
What if we invert the abstraction so we still only have 1 class that communicates with the service?

For example:

```ts
// Use an interface, just like your original, but no optional args on createTask!
interface TaskIntegrationManager {
  title: string

  createTask(params: {
    rawContentStr: string
    integrationRepoId: string
  }): Promise<CreateTaskResponse>
}

// The managers that talk to the service implement the TaskIntegrationManager
export default class JiraServerRestManager implements TaskIntegrationManager {
  constructor(auth, provider) {}
  createTask(str, repoId){}
}

// we grab the context, info in the constructor so methods like createTask don't need extra stuff
export default class GitHubServerManager implements TaskIntegrationManager {
  constructor(auth, provider, context, info) {}
  createTask(str, repoId){}
}
```

that way, our server managers can have all the wacky methods they want, but they'll still need to implement a few basic ones if they want to call themselves a TaskIntegrationManager

whatcha think?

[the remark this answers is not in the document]
igorlesnenko: @mattkrick Yes, I think we may merge two classes (`JiraServerRestManager` and `JiraServerTaskIntegrationManager`) into one. 

But in this case, the resulting `class` would expose different methods that sound similar. For example: `createIssue` which is just a REST API call wrapper, and `createTask` — above mentioned unified method that wraps REST API method and might include additional actions.

Is it OK?

packages/client/components/TaskWatermark.tsx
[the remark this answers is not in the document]
Dschoordsch: That's the logo for Jira Software, I think we should use it instead of the 3 arrows on the task card. Better version is available in the [atlassian logo library](https://atlassian.design/resources/logo-library). For the integration banner I had used the old logo they still use in their Jira Server documentation, could be the same Jira Software logo though.

packages/server/integrations/JiraServerTaskIntegrationManager.ts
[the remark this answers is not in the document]
mattkrick: i think that's great!
if you wanna get fancy, the taskIntegrationFactory could return this Manager as a TaskIntegrationManager interface so those other methods will be hidden 😉

packages/client/components/TaskWatermark.tsx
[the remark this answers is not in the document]
igorlesnenko: Updated the logo  
image

packages/server/integrations/JiraServerTaskIntegrationManager.ts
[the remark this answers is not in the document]
igorlesnenko: @mattkrick made some refactoring, can you take a look?

packages/server/graphql/mutations/createTaskIntegration.ts
```diff
@@ -112,33 +125,40 @@ export default {
...
     if (createTaskResponse instanceof Error) {
       return {error: {message: createTaskResponse.message}}
     }
 
+    const {issueId, ...updateTaskInput} = createTaskResponse
+
+    if (
+      userId &&
+      viewerId !== userId &&
+      'addCreatedBySomeoneElseComment' in taskIntegrationManager
```
mattkrick, line 142: i imagine this'll become non-null in the future, but happy to leave as is for now so we can get this big chunk merged master

mattkrick:
wonderful stuff! the extra focus applied here to set the patterns should pay off when we repeat this pattern for the next integrations

packages/server/graphql/mutations/createTaskIntegration.ts
igorlesnenko: @mattkrick it is already non-null in the interface, so removing this check in my next PR

Issue #6018, opened 2022-02-08
Add first-class unit testing support (backend)


blimmer:
## Issue - Enhancement

Today, the backend test suite only contains [a few tests](https://github.com/ParabolInc/parabol/tree/d681cf1e283f4bc497d456d96bcdedc393e1a49a/packages/server/__tests__). This makes it challenging to write tests for new features and against bugs as they're fixed.

There are currently a few missing infrastructural components that make adding additional tests more challenging:

1. Tests that rely on the database have side-effects. The database should be truncated after each test runs.
2. Rate limiter can cause test suites to fail.

### Database Truncation

For Postgres I've seen a pattern that uses the excellent [`pg-structure`](https://www.pg-structure.com/) NPM package to build a database function for truncating all tables. Then `beforeEach()` test, you call the function within the test database to ensure tests do not leak state between each other.

I'm not super familiar with RethinkDB but, because there's active work to migrate over to Postgres, perhaps we could hard-code some reset logic for the data stored in Rethink?

### Rate Limiter

It looks like the Rate Limiter is currently using an in-memory store since the app isn't horizontally scaled. For the short-term, I think we should be able to find a way to reset the store `beforeEach()` test, similar to the database.

Speaking more long-term, it might be nice to introduce a `Context` object that wraps up "tools" used throughout the server. For instance:

```ts
interface Context {
  rateLimiter: RateLimiter
  mailer: Mailgun
  logger: Pino
  stripe: Stripe
}
```

Then, in `test` you can inject test-versions of the services in the `Context` to reduce the need to mock the third-party services.

### Acceptance Criteria
Developers can:
  - Be sure that the data stores are reset between each test that talks with the database
  - Not be rate limited when calling the same endpoint multiple times

**Estimated effort:** 13 points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes)). This is just an estimate. I mostly wanted to open this up for discussion for now. I think the outcome of this ticket would be a few smaller tickets that would be easier to scope and point.

mattkrick:
Notes from chatting with Ben:
- This is really good & probably the pattern that we want to move towards
- Before we do this, we'll want to complete #4778. Getting that extra type safety will provide the biggest bang for the buck & then moving things like DB drivers into the context can come next
[...]

PR #6063, opened 2022-02-16
feat: add StubRateLimiter

blimmer:
This PR introduces a new `StubRateLimiter` class that we can use in tests.

Previously, when running the integration-tests or backend unit tests that talk to the database,  we'd get transient failures due to rate limiting. To work around this, the rate limiter type is now configurable via an environment variable, which will default to the `StubRateLimiter` in test.

Additionally, as you look to horizontally scale the application, you could use the framework I built with this PR to introduce a rate limiter with a different persistent backend (e.g. Redis).

This Loom shows off these changes: https://www.loom.com/share/311c313cb99246619a735debb51ef817


.env.example
```diff
@@ -52,6 +52,8 @@ POSTGRES_DB=parabol-saas
 POSTGRES_HOST=localhost
 POSTGRES_PORT=5432
 PROTO='http'
+# RATE_LIMITER: in-memory | stub
+RATE_LIMITER='stub'
```
blimmer, line 56: See https://github.com/ParabolInc/action-devops/pull/83 for the updates to the environment files.

packages/server/utils/__tests__/rateLimiters/InMemoryRateLimiter.test.ts
```diff
@@ -0,0 +1,182 @@
...
+  describe('gc', () => {
+    it('cleans up users who have not called the service in the last hour', () => {
+      // Given an InMemoryRateLimiter with scheduled GC enabled
+      const inMemoryRateLimiter = subject({scheduleGc: true})
+      const userId = '123'
+      const fieldName = 'FieldName'
+
+      // When a user requested a field
+      inMemoryRateLimiter.log(userId, fieldName, false)
+      // When two hours elapse (triggering GC twice)
```
blimmer, line 170: I like given/when/then style testing: https://martinfowler.com/bliki/GivenWhenThen.html. I think it helps make tests, especially more complicated tests like this one, more understandable.

packages/server/graphql/getRateLimiter.ts
```diff
@@ -1,11 +1,41 @@
...
+ * This function returns the appropriate rate limited based on the RATE_LIMITER
+ * environment variable, falling back to the InMemoryRateLimiter (the previous,
+ * uncustomizable default).
+ *
+ * Exposed for testing. Use the default export outside of test.
+ *
+ * @private
+ */
+export function _getRateLimiter(): RateLimiter {
+  const rateLimiterEnvVar = process.env.RATE_LIMITER
```
blimmer, line 26: Plug for my friend Stephen's `ts-app-env` project: https://github.com/stephenh/ts-app-env/. This makes dealing with environment variables really easy!

packages/server/graphql/__tests__/getRateLimiter.test.ts
```diff
@@ -0,0 +1,57 @@
...
+    // Then it is an in-memory rate limiter
+    expect(_getRateLimiter()).toBeInstanceOf(InMemoryRateLimiter)
+  })
+
+  it('uses the in-memory rate limiter when specified in the env variable', () => {
+    // Given that the RATE_LIMITER env variable is set to 'in-memory'
+    process.env.RATE_LIMITER = 'in-memory'
+
+    // When getting the rate limiter
+    // Then it is an stub rate limiter
```
tianrunhe, line 34: +1 typo: Then it is an **in-memory** rate limiter

packages/server/graphql/getRateLimiter.ts
```diff
@@ -1,11 +1,41 @@
-import RateLimiter from './RateLimiter'
+import {InMemoryRateLimiter, StubRateLimiter, RateLimiter} from '../utils/rateLimiters'
+
+const RATE_LIMITERS = Object.freeze({
+  'in-memory': InMemoryRateLimiter,
+  stub: StubRateLimiter
```
tianrunhe, line 5: +1 better to use consistent key format: `'stub'`

tianrunhe (approved):
LGTM! Just two small nits

packages/server/utils/__tests__/rateLimiters/InMemoryRateLimiter.test.ts
Dschoordsch: +1 I like structuring the tests in these 3 sections (I know this as arrange, act, assert / AAA), but I'm not fond of all those verbose comments repeating the test description or the code itself. I would prefer some middle ground only stating the non-obvious, otherwise I find these distract me from the actual code. If it's mostly about the structure, I think
```ts
      // Given
      const inMemoryRateLimiter = subject({scheduleGc: true})
      const userId = '123'
      const fieldName = 'FieldName'

      // When two hours elapse between requests, triggering GC twice
      inMemoryRateLimiter.log(userId, fieldName, false)
      jest.advanceTimersByTime(ms('1h'))
      jest.advanceTimersByTime(ms('1h'))
      const secondResult = inMemoryRateLimiter.log(userId, fieldName, false)

      // Then
      // We're relying on the behavior that you receive 1 lastHour the first time the user calls without
      // extended logging for this assertion. If the user was not GC'ed we've get lastMinute: 1, lastHour: undefined
      expect(secondResult).toEqual({lastMinute: 1, lastHour: 1})
```
would be verbose enough.

packages/server/graphql/getRateLimiter.ts
blimmer: `prettier` actually won't let me do this.

Screen Shot 2022-02-22 at 08 42 00

The reason `in-memory` must be quoted is because it contains a dash.

packages/server/utils/__tests__/rateLimiters/InMemoryRateLimiter.test.ts
blimmer: Sounds good - I removed some verbosity.

packages/server/utils/rateLimiters/StubRateLimiter.ts
```diff
@@ -0,0 +1,18 @@
+import {RateLimiter, RateLimitStats} from './RateLimiter'
+
+type StubRateLimiterMode = 'always-allow'
+
+export class StubRateLimiter implements RateLimiter {
```
mattkrick, line 5: -1 overall, I'd like to avoid stubs if at all possible because whether a stub passes or fails doesn't tell us anything about the code that's in production. 

It looks like the spirit of this PR is to circumvent the rate limiter while testing.

Could we instead add a `if (NODE_ENV === 'testing')` conditional here & achieve the same result?
https://github.com/ParabolInc/parabol/blob/33c3478b8699f0a61e092b533c36e4e6ac03c704/packages/server/graphql/rateLimit.ts#L31-L35

mattkrick (changes requested):

packages/server/utils/rateLimiters/StubRateLimiter.ts
blimmer: The downside to using the global `if (NODE_ENV === 'testing')` behavior is that we'd globally disable rate limits for all unit tests. There are use-cases where we'd want to test that the rate limiter is working correctly in a test.

My thought is that we'd default to the stub rate limiter in the test environment. Then, after introducing a server context (see https://github.com/ParabolInc/parabol/issues/6018), it would be pretty straightforward to pass a real rate limiter in a specific testing context where you want to test rate limiting.

> I'd like to avoid stubs if at all possible because whether a stub passes or fails doesn't tell us anything about the code that's in production

I think the same logic applies to disabling with an environment variable check. Any way you shuck it; we're short-circuiting production logic to be able to run unit tests without hitting the rate limit.

That said, I'm happy to go in whatever direction you prefer. @mattkrick let me know how you'd like to proceed.

blimmer: From our sync conversation, I'm going to go the environment variable route for now. I won't squash the existing logic in case we choose to come back to this in the future.

PR #6065, opened 2022-02-16
6032 making the scope issue picker wider and taller to account for la…

acressall:
- Updating the width to use a percentage, so it's wider on larger screen sizes
- adding a max-width so the checkboxes and tabs on the left don't get too far from the actions on the right
- updating the height to a percentage so more issues can be viewed


nickoferrall:
Excellent work, @acressall! This looks great.

Now that the scope card is wider, the loading screen might need to be wider too. It looks like [MockScopingList](https://github.com/ParabolInc/parabol/blob/5da8f4d0eddf52d5ad77d5885bb950bdf409e4b5/packages/client/modules/meeting/components/MockScopingList.tsx) is the component to dig into. 

At the moment, the issue text can take up the full width but the loading screen only takes up a small portion of it. If we increase the width of the loading component to 100%, it looks a lil aggressive IMO, so I wonder whether 75% would be best? The shine in the loading component might need editing as there used to be one shine on the screen, but if we increase the width, there are two.

loading-scope

acressall:
Thanks @nickoferrall! I've updated MockScopingTask to:

- increase the width of the task
- increase the background size
- extend the endpoint of the animation

nickoferrall (approved):
LGTM! I'll move this along to Maintainer Review

mattkrick:
+1 safe to merge, only thing i noticed is that now the mock loading list looks a little empty. Should we grow the list to more than 6 items @ackernaut 

Screen Shot 2022-02-17 at 2 52 00 PM

acressall:
@ackernaut we have it hardcoded to 6 items, but if we assume that a mobile user won't scroll while loading we can increase to 12. WDYT?

acressall:
I made a couple small changes @ackernaut, @mattkrick! 

- increased the padding between mock items, so they resemble the height of the real tasks and also take up more height
- randomized the width of each list item so it feels more like a real task
- increased the amount of items to display to 8 (12 felt really crowded on a laptop)

Here's a screen recording of how it looks:
https://user-images.githubusercontent.com/18582225/155246507-aed888e8-46d9-4534-b766-36ced02f4e96.mov

mattkrick:
> Any way to know the total while loading?

nope


the randomly updating lengths feels a little weird, i always try to make sure movement has meaning for the user. IMO i'd just go with the full width bars since that's what other loading lists across our app use

acressall:
Thanks for the feedback @mattkrick! Reverted to 80% width

mattkrick:
LGTM!

mattkrick:
Review time: 1 hour

Issue #5917, opened 2022-01-21
Add a new meeting GQL types


BartoszJarocki:
Depending on how the new meeting data structure will look like, prepare all the backend types needed.

AC
- [ ] add a new meeting gql type, and add this type to a gql NewMeeting type
- [ ] add a new meeting type to a database TMeetingTypeEnum type
- [ ] add a new meeting type to a gql MeetingTypeEnum type
- [ ] add new meeting phases types to NewMeetingPhaseTypeEnum
- [ ] add new meeting type settings

BartoszJarocki:
**Include in next sprint?: Y**
  [See the discussion](https://action.parabol.co/meet/aGjmYpJBIs/estimate/30) in Planning for Sprint #​​96

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

BartoszJarocki:
**Include in next sprint?: Y**
  [See the discussion](https://action.parabol.co/meet/aOko8PmZr2/estimate/15) in Planning for Sprint #​​97

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

Issue #5916, opened 2022-01-21
Design proposed for standup database schema


BartoszJarocki:
Create a standup database schema while keeping in mind that, standup is only one type of the possible use of the new meeting type. `team prompt` might be a more appropriate name for internal use. 

This issue involves designing and architecting data structures for a new meeting type.
1 . how the data structure should look like? 
2. how do we store responses? 
3. how can we design the schema with recurrence in mind, does it change anything? 
4. how do we support multiple days/weeks/whatever? 
5. what are the phases of the meeting? 
6. what are the stages?

Before writing any code, it is required to outline the solution and discuss it with the maintainer. 

AC:
- [ ] data structure proposed and discussed with maintainer
- [ ] data structure implemented via migrations

BartoszJarocki:
WIP

There are at least two ways to accomplish the goal of having team prompt activity with recurrence in mind.

## `NewMeeting` table changes

### First way - each interval instance is a new meeting phase
For example, we have a team prompt meeting, interval set to `daily`. In this case, each day will be a new meeting phase.

Changes needed

1. Extend the `NewMeeting` table [rethinkdb]
   - add new meeting type - `teamPrompt`
   - add new fields related to recurrence, `isRecurring`, `interval` [daily, weekly, monthly, yearly?], `startDate` (with tz), `endDate` (with tz)
   - depending on the further specification we can have either always a single stage or create a stage for each meeting participant

Example JSON structure of a meeting

```json
{
  "id": "...",
  "createdBy": "...",
  "createdAt": "...",
  "endedAt": "...",
  "meetingNumber": 1,
  "meetingType": "teamPrompt", // new type
  "name": "Async Standup #1",
  "isRecurring": true, // new field
  "interval": "daily", // new field
  "startDate": "...", // new field
  "endDate": "...", // new field
  "summarySentAt": {
    "$reql_type$": "TIME",
    "epoch_time": 1644211686.138,
    "timezone": "+00:00"
  },
  "teamId": "...",
  "templateId": "...",
  "updatedAt": {
    "$reql_type$": "TIME",
    "epoch_time": 1644211683.29,
    "timezone": "+00:00"
  },
  "phases": [
    {
      "id": "...",
      "phaseType": "responses", // new type
      "stages": [
        {
          "id": "...",
          "phaseType": "responses",
          "startAt": {
            "$reql_type$": "TIME",
            "epoch_time": 1644211647.127,
            "timezone": "+00:00"
          },
[...]

Issue #5965, opened 2022-01-26
Designed response grid for async stand-ups


enriquesanchez:
### Acceptance Criteria
Design a concept that answers the following questions:

- How does the grid layout? Masonry? How responsive are the columns?
- How to do the empty state?
- Can there be a simpler empty state?
- Can we defer the empty state that shows team members?
- How to handle when the thread is open?
- What does this look like on mobile?

**Estimated effort:** 
4 hrs.

jordanh:
**Include in next sprint?: Y**
  [See the discussion](https://action.parabol.co/meet/aGjmYpJBIs/estimate/17) in Planning for Sprint #​​96

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

enriquesanchez:
![imagen](https://user-images.githubusercontent.com/3276087/154165893-c4de2bf5-467f-4f8c-a028-aedd7a3549f1.png)


[Here are the Figma mockups](https://www.figma.com/file/bhiEOV4HK9Kv91HFsJmevV/Stand-up-Concepts?node-id=1220%3A74323).

### The Grid:
- The grid is responsive.
- The grid follows a masonry style layout, where responses stack and occupy any space that is available.
- The grid will have a maximum of 3 columns, and a minimum of 1. This will depend on the viewport's size.

### Response threads:
- Work on response threads is being tracked on #5966
- When a response thread is open, and there's is enough space in the viewport to fit the grid and the thread, the thread acts as a sidebar/drawer that pushes the grid to the left.
- When there isn't enough space, the thread acts as a sidebar/drawer that sits on top of the grid.

ackernaut:
Thanks for pushing forward here!

- Do we start with a masonry grid, or something easier for the MVP? If we use a masonry grid, do we need to truncate cards? If we go with a simple wrapping flex group or CSS grid, do we need to truncate to avoid vertical gaps?
[...]

PR #6068, opened 2022-02-16
Initial team prompt db schema

BartoszJarocki:
Implements #5916 

Scope

- [ ] Add new type teamPrompt to `type MeetingTypeEnum = 'poker' | 'retrospective' | 'action'`
- [ ] Add new table in PG TeamPromptResponse with fields id, createdAt, updatedAt, creatorId, meetingId, content, reactjis
- [ ] Change the Discussion table add new enum field teamPromptResponse to discussionTopicType enum

Important: `recurrence` is still being discussed and it's out of scope in this phase 

GQL types will be implemented in https://github.com/ParabolInc/parabol/issues/5917


packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
```diff
@@ -0,0 +1,72 @@
...
+    CREATE TYPE "MeetingTypeEnum" AS ENUM (
+      'action',
+      'retrospective',
+      'poker'
+    );
+    ALTER TABLE "Team"
+      ALTER COLUMN "lastMeetingType" DROP DEFAULT,
+      ALTER COLUMN "lastMeetingType" TYPE "MeetingTypeEnum" USING "lastMeetingType"::text::"MeetingTypeEnum",
+      ALTER COLUMN "lastMeetingType" SET DEFAULT 'retrospective';
+    DROP TYPE "MeetingTypeEnum_delete";
```
BartoszJarocki, line 65: reverting changes to enums is such a pain. I haven't found anything better but if there's a way, I'm happy to improve

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
```diff
@@ -0,0 +1,72 @@
...
+
+      CREATE TABLE IF NOT EXISTS "TeamPromptResponse" (
+        "id" VARCHAR(50) PRIMARY KEY,
+        "createdAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+        "updatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+        "meetingId" VARCHAR(100) NOT NULL,
+        "userId" VARCHAR(100) NOT NULL,
+        "content" TEXT NOT NULL,
+        "plainTextContent" TEXT NOT NULL,
+        "reactjis" "Reactji"[] NOT NULL DEFAULT array[]::"Reactji"[],
```
BartoszJarocki, line 26: I think this is enough and we don't need a separate table for reactjis. if there's a use case that'd benefit from that, I'm happy to add it

packages/server/database/types/Meeting.ts
```diff
@@ -1,7 +1,7 @@
 import generateUID from '../../generateUID'
 import GenericMeetingPhase from './GenericMeetingPhase'
 
-export type MeetingTypeEnum = 'poker' | 'retrospective' | 'action'
+export type MeetingTypeEnum = 'poker' | 'retrospective' | 'action' | 'teamPrompt'
```
BartoszJarocki, line 4: should we remove this type and use PG-generated enum to keep one source of truth?

BartoszJarocki:
@tianrunhe would you mind taking a look? thank you!

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
```diff
@@ -0,0 +1,72 @@
...
+import getPgConfig from '../getPgConfig'
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+    DO $$
+    BEGIN
+      IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'Reactji') THEN
+        CREATE TYPE "Reactji" AS (id text, userId text);
```
tianrunhe, line 11: -1 should we also put foreign key constraint on the `userId` here?

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
tianrunhe, line 26: +1 suggestion: add one more column `sortOrder` as v1

tianrunhe (changes requested):
I like we move fast with pieces we can build today, while waiting for discussion on other parts! Just a few small suggestions.

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
```diff
@@ -0,0 +1,72 @@
+import {Client} from 'pg'
+import getPgConfig from '../getPgConfig'
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+    DO $$
+    BEGIN
+      IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'Reactji') THEN
```
mattkrick, line 10: +2 nice use of composite types!

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
```diff
@@ -0,0 +1,72 @@
...
+      ALTER TYPE "MeetingTypeEnum" ADD VALUE 'teamPrompt';
+
+      CREATE TABLE IF NOT EXISTS "TeamPromptResponse" (
+        "id" VARCHAR(50) PRIMARY KEY,
+        "createdAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+        "updatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+        "meetingId" VARCHAR(100) NOT NULL,
+        "userId" VARCHAR(100) NOT NULL,
+        "content" TEXT NOT NULL,
+        "plainTextContent" TEXT NOT NULL,
```
mattkrick, line 25: +1 everywhere else we call this `plaintextContent` (lowercase T). all depends if plaintext is a single word or 2 words 😅

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
```diff
@@ -0,0 +1,72 @@
...
+      IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'Reactji') THEN
+        CREATE TYPE "Reactji" AS (id text, userId text);
+      END IF;
+
+      ALTER TYPE "DiscussionTopicTypeEnum" ADD VALUE 'teamPromptResponse';
+
+      ALTER TYPE "MeetingTypeEnum" ADD VALUE 'teamPrompt';
+
+      CREATE TABLE IF NOT EXISTS "TeamPromptResponse" (
+        "id" VARCHAR(50) PRIMARY KEY,
```
mattkrick, line 19: +1 curious why no `"id" INT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY`
Was it no good for the polls table?

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
mattkrick: for downmigrations i just wouldn't do it. up migrations are guaranteed to run eventually, so if an enum has an extra value for a couple minutes it won't hurt anything.

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
```diff
@@ -0,0 +1,72 @@
...
+    );
+    ALTER TABLE "Team"
+      ALTER COLUMN "lastMeetingType" DROP DEFAULT,
+      ALTER COLUMN "lastMeetingType" TYPE "MeetingTypeEnum" USING "lastMeetingType"::text::"MeetingTypeEnum",
+      ALTER COLUMN "lastMeetingType" SET DEFAULT 'retrospective';
+    DROP TYPE "MeetingTypeEnum_delete";
+
+    DROP TABLE IF EXISTS "TeamPromptResponse";
+
+    DROP TYPE "Reactji";
```
mattkrick, line 69: +1 IF EXISTS. sometimes things get wonky, so i just add that check for everything

mattkrick (changes requested):

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
BartoszJarocki: @mattkrick sure! 

@tianrunhe I think it isn't possible to add foreign key constraints in composite types, I googled the internet and couldn't find anything 🤔  

```
The syntax is comparable to CREATE TABLE, except that only field names and types can be specified; no constraints (such as NOT NULL) can presently be included. Note that the AS keyword is essential; without it, the system will think a different kind of CREATE TYPE command is meant, and you will get odd syntax errors.
```
https://www.postgresql.org/docs/13/rowtypes.html#ROWTYPES-DECLARING

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
BartoszJarocki: yeah, I was actually wondering what's the correct way 😅 will fix!

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
BartoszJarocki: I focused too much on enums and composite types and actually didn't think too much about this, very much my bad, it should be exactly like in polls!

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
BartoszJarocki: good catch!

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
BartoszJarocki: i'm happy to add it but what do you think is the use case for it?

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
BartoszJarocki: as it's working and I already spend time on this I'll leave it, next time I'll skip it 👍🏻

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
tianrunhe: That's totally fine! I wasn't sure if there's a way to do it either 😅

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
tianrunhe: Could be used in response grid sorting. See [here](https://github.com/ParabolInc/parabol/issues/5965#issuecomment-1042116030). Although I haven't thought about this carefully to see if `updatedAt` can be used to achieve sorting.

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
```diff
@@ -0,0 +1,72 @@
...
+      ALTER TYPE "MeetingTypeEnum" ADD VALUE 'teamPrompt';
+
+      CREATE TABLE IF NOT EXISTS "TeamPromptResponse" (
+        "id" INT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+        "createdAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+        "updatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+        "meetingId" VARCHAR(100) NOT NULL,
+        "userId" VARCHAR(100) NOT NULL,
+        "content" TEXT NOT NULL,
+        "plaintextContent" TEXT NOT NULL,
```
mattkrick, line 25: +1 don't hesitate to tell me no, but if `content` is stored as JSON, then `plaintextContent` could be GENERATED and we won't need to do that at the application layer

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
```diff
@@ -0,0 +1,72 @@
...
+
+      ALTER TYPE "MeetingTypeEnum" ADD VALUE 'teamPrompt';
+
+      CREATE TABLE IF NOT EXISTS "TeamPromptResponse" (
+        "id" INT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+        "createdAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+        "updatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+        "meetingId" VARCHAR(100) NOT NULL,
+        "userId" VARCHAR(100) NOT NULL,
+        "content" TEXT NOT NULL,
```
BartoszJarocki, line 24: This should be JSON

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
BartoszJarocki: @tianrunhe thanks! considering recent discussions around `sortOrder` I guess it should be an int. feel free to correct me, thought!

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
BartoszJarocki: I haven't found any easy way to do it so I'll create a task for it and let's continue with doing that conversion in the application layer for now.

BartoszJarocki:
@mattkrick may I ask you to check it again?

packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts
mattkrick: > any chance you know a simple way of converting json to text in pg?

i think we talked about this in the 1:1. it'd have to be a function that traversed the json recursively picking out the text property on text nodes & aggregating them. definitely worth holding off for now

> I was thinking about it and I have one more question, is there any benefit of storing the same data in a different format in DB? what's the use case for that field, in general?

rich text is hard to query against, read for debugging, & no good for searching if we even wanted to implement something like elastic search

mattkrick:
LGTM!

Review time: 2 hours

PR #6087, opened 2022-02-19
enable application security for datadog monitoring

adaniels-parabol:
This PR enables application security monitoring within DataDog.


Dschoordsch:
Looks good, but right now the service `Web ...` does not send any APM data und thus is not available for app-sec. Even when I change it locally to send some and it appears as a service in DD, I still cannot choose it for app-sec.

PR #6090, opened 2022-02-22
Rethrow exceptions in traceGraphQL

Dschoordsch:
Errors thrown in resolvers were swallowed silently when tracing is
enabled.


BartoszJarocki (approved):
good find!

mattkrick:
LGTM!

PR #6092, opened 2022-02-22
Jira Server: show integrated task 

igorlesnenko:
Resolves #5994

Needs #6059

image

**How to test:**

- Add Jira Server integration
- Create a task in parabol
- Using push to integration menu, push a task to the selected Jira Server
- See the task was pushed successfully
- Open task in Jira Server, see it was created correctly
- Edit task in Jira Server
- Open parabol board, see the task is loaded correctly

PR #6104, opened 2022-02-22
build(server): fix jest transforms

blimmer:
@igorlesnenko ran into some issues on his PR (#6092) where `jest` was throwing this error in [CI](https://app.circleci.com/pipelines/github/ParabolInc/parabol/4405/workflows/c60f5290-e766-485c-a1eb-e0cf00d53f7d/jobs/13207):

```console
 FAIL  dataloader/__tests__/usersCustomRedisQueries.test.ts
  ● Test suite failed to run

    Cannot find module '~/shared/gqlIds/JiraProjectId' from 'integrations/JiraTaskIntegrationManager.ts'

    Require stack:
      integrations/JiraTaskIntegrationManager.ts
      integrations/TaskIntegrationManagerFactory.ts
      dataloader/jiraServerLoaders.ts
      dataloader/RootDataLoader.ts
      graphql/getDataLoader.ts
      dataloader/__tests__/usersCustomRedisQueries.test.ts

      1 | import makeCreateJiraTaskComment from '../utils/makeCreateJiraTaskComment'
    > 2 | import JiraProjectId from '~/shared/gqlIds/JiraProjectId'
        | ^
      3 | import createJiraTask from '../graphql/mutations/helpers/createJiraTask'
      4 | import JiraIssueId from '~/shared/gqlIds/JiraIssueId'
      5 | import {CreateTaskResponse, TaskIntegrationManager} from './TaskIntegrationManagerFactory'

      at Resolver.resolveModule (../../node_modules/jest-resolve/build/resolver.js:324:11)
      at Object.<anonymous> (integrations/JiraTaskIntegrationManager.ts:2:1)
```

The problem is because `jest`/`ts-jest` doesn't automatically map `tsconfig.json`'s `paths` to its internal module mapper (used for mocking in Jest). There's documentation on this [here](https://kulshekhar.github.io/ts-jest/docs/getting-started/paths-mapping/).

Additionally, once that module mapping was working, we ran into another problem:

```console
 FAIL  dataloader/__tests__/usersCustomRedisQueries.test.ts
  ● Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /Users/blimmer/code/parabol/parabol/packages/server/utils/githubQueries/createIssue.graphql:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){mutation createIssue($input: CreateIssueInput!) {
                                                                                               ^^^^^^^^^^^

    SyntaxError: Unexpected identifier

      11 |   GetRepoInfoQueryVariables
      12 | } from '../../../types/githubTypes'
    > 13 | import createIssueMutation from '../../../utils/githubQueries/createIssue.graphql'
         | ^
      14 | import addComment from '../../../utils/githubQueries/addComment.graphql'
      15 | import getRepoInfo from '../../../utils/githubQueries/getRepoInfo.graphql'
      16 | import {GQLContext} from '../../graphql'

      at Runtime.createScriptFromCode (../../node_modules/jest-runtime/build/index.js:1728:14)
      at Object.<anonymous> (graphql/mutations/helpers/createGitHubTask.ts:13:1)
```

To fix that problem, I added a `graphql` transformer.

See inline comments for additional details.

### DIFF
--- a/packages/server/jest.config.js
+++ b/packages/server/jest.config.js
@@ -1,4 +1,4 @@
-const {jsWithTs: tsjPreset} = require('ts-jest/presets')
+const tsJestPresets = require('ts-jest/presets')
 
 module.exports = {
   preset: 'ts-jest',
@@ -9,8 +9,14 @@ module.exports = {
     }
   },
   transform: {
-    ...tsjPreset.transform
+    '\\.(gql|graphql)$': 'jest-transform-graphql',
+    ...tsJestPresets.jsWithBabel.transform
   },
   modulePaths: ['<rootDir>/packages/'],
+  moduleNameMapper: {
+    'server/(.*)': ['<rootDir>/$1'],
+    'parabol-client/(.*)': ['<rootDir>/../client/$1'],
+    '~/(.*)': ['<rootDir>/../client/$1']
+  },
   testRegex: '/__tests__/.*.test\\.ts?$'
 }
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -63,6 +63,7 @@
     "jest": "^27.0.6",
     "jest-extended": "^1.2.0",
     "jest-junit": "^13.0.0",
+    "jest-transform-graphql": "^2.1.0",
     "jsdom": "^16.6.0",
     "json-loader": "^0.5.7",
     "lint-staged": "^10.1.7",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4764,6 +4764,14 @@ are-we-there-yet@^2.0.0:
     delegates "^1.0.0"
     readable-stream "^3.6.0"
 
+are-we-there-yet@^3.0.0:
+  version "3.0.0"
+  resolved "https://registry.yarnpkg.com/are-we-there-yet/-/are-we-there-yet-3.0.0.tgz#ba20bd6b553e31d62fc8c31bd23d22b95734390d"
+  integrity sha512-0GWpv50YSOcLXaN6/FAKY3vfRbllXWV2xvfA/oKJF8pzFhWXPV+yjhJXDBbjscDYowv7Yw1A3uigpzn5iEGTyw==
+  dependencies:
+    delegates "^1.0.0"
+    readable-stream "^3.6.0"
+
 are-we-there-yet@~1.1.2:
   version "1.1.7"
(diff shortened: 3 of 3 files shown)


packages/server/jest.config.js
```diff
@@ -9,8 +9,14 @@ module.exports = {
     }
   },
   transform: {
-    ...tsjPreset.transform
+    '\\.(gql|graphql)$': 'jest-transform-graphql',
+    ...tsJestPresets.jsWithBabel.transform
```
blimmer, line 13: In our current `tsconfig.js`, `allowJs` is not set so we're using the default `false`. Instead, `js` is transpiled with `babel` by `webpack`. Therefore, I believe this config more accurate represents our build config.

packages/server/jest.config.js
```diff
@@ -9,8 +9,14 @@ module.exports = {
...
   },
   modulePaths: ['<rootDir>/packages/'],
+  moduleNameMapper: {
+    'server/(.*)': ['<rootDir>/$1'],
+    'parabol-client/(.*)': ['<rootDir>/../client/$1'],
+    '~/(.*)': ['<rootDir>/../client/$1']
+  },
```
blimmer, line 20: ts-jest does provide [a helper method](https://kulshekhar.github.io/ts-jest/docs/getting-started/paths-mapping/), but because `client` is one directory down it doesn't quite work for this use case. It was easy enough to just write this out.

igorlesnenko (approved):
It works!

PR #6107, opened 2022-02-22
Add migration to update prompt colors for Three Little Pigs template

### LINKED ISSUE #4703: 🐷🐷🐷 template colors
## Issue - Bug

A user 🔒 writes:

image
> We were thinking that:
> House of Bricks should be green
> House of Straw should be yellow

I agree these colors should change. At least, how can Straw not be yellow??

EE: trivial
AC: 
  - new, more appropriate prompt label colors chosen

tianrunhe:
Fixes #4703 

Test steps:

1.  Pull down the branch and run `yarn dev`
2.  Migration should run: \[migrate-rethinkdb\]  ↑  up  ↑  20220222141104 updateThreeLittlePigsPromptColors
3.  In the template picker for retro meeting, select Three Little Pigs from Public templates, see the correct color should be present.

### DIFF
--- a/packages/server/database/migrations/20220222141104-updateThreeLittlePigsPromptColors.ts
+++ b/packages/server/database/migrations/20220222141104-updateThreeLittlePigsPromptColors.ts
@@ -1,0 +1,44 @@
+import {R} from 'rethinkdb-ts'
+import {PALETTE} from '../../../client/styles/paletteV3'
+
+export const up = async function (r: R) {
+  try {
+    await r({
+      houseOfBricks: r
+        .table('ReflectPrompt')
+        .get('promptHouseOfBricks')
+        .update({groupColor: PALETTE.JADE_400}),
+      houseOfSticks: r
+        .table('ReflectPrompt')
+        .get('promptHouseOfSticks')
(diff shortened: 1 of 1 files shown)


tianrunhe:
@ackernaut wanna take a look? I took this chance to update the color using the new palette. Thank you!

ackernaut:
Hey Bruce, thanks for pushing here. I like that you are using the palette in the migration.

I was confused at the color choices so I looked back at the original issue. The user had recommended trying green for house of bricks (green meaning the good ‘go’ connotation). For a second I expected the colors to match the illustration, haha. Then house of bricks would be red and counterintuitive without the illustration. This color connection I describe (or lack thereof) should only be possible when looking at the template, which only affects a few. In the meeting, which affects everybody participating, we use color and words only. So green makes more sense in that context. Straw seems to be more serious than sticks so I would think it would be red and sticks would be yellow, like the stoplight metaphor. Putting straw at yellow follows the visual metaphor. Putting brick at green does not. Still kinda weird altogether.

This looks safe to try to me! Will you have somebody else look at the migration?

ackernaut (approved):

PR #6118, opened 2022-02-23
fix template picker sizing issue on mobile

acressall:
Changes to both retro and poker template pickers:
- Reducing the height of the template list so the action button can appear below
- Increasing the z-index of the action button so it does not sit behind the list (it was unclickable before)
- adding a max-width to the template list containers so they don't overflow

TODO: 
- make the active tab bar display in the correct position on load

### DIFF
--- a/packages/client/modules/meeting/components/PokerTemplateList.tsx
+++ b/packages/client/modules/meeting/components/PokerTemplateList.tsx
@@ -63,8 +63,8 @@ const TabIcon = styled(Icon)({
   marginRight: 4
 })
 
-const containerStyle = {height: '100%'}
-const innerStyle = {width: '100%', height: '100%'}
+const containerStyle = {width: '100%', height: '85%'}
+const innerStyle = {width: '100%', height: '100%', maxWidth: 'calc(100vw - 48px)'}
 interface Props {
   activeIdx: number
   setActiveIdx: (idx: number) => void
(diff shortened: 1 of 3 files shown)


acressall:
@nickoferrall would you mind taking a look at this? I tried for a while to fix the issue of the active bar displaying too far to the left, but am not sure how to fix it. I _think_ it's happening because the icon for the active element hasn't loaded yet because it only happens when you initially open the dialog

packages/client/modules/meeting/components/PokerTemplateList.tsx
```diff
@@ -63,8 +63,8 @@ const TabIcon = styled(Icon)({
   marginRight: 4
 })
 
-const containerStyle = {height: '100%'}
-const innerStyle = {width: '100%', height: '100%'}
+const containerStyle = {height: '85%'}
+const innerStyle = {width: '100%', height: '100%', maxWidth: 'calc(100vw - 48px)'}
```
nickoferrall, line 67: -1 the `maxWidth` change looks really good, but I think there's an underlying issue with the box model as child elements can be wider than their parent:

child-width

The problem with using `calc(100vw - 48px)` here is other child elements that we add may encounter the same issue, so we could end up with lots of calcs in different places. Ideally, we can arrange the components so that the child width is 100% without it being wider than the parent.

nickoferrall (changes requested):
Looking good! If the active tab bar & box model are still misbehaving after playing around with it a bit more, feel free to break the issue up into small issues

Dschoordsch:
@nickoferrall would this be simple enough for you to take this over from @acressall and implement the proper fix? This has been stale for some time.

nickoferrall:
@Dschoordsch I'd be happy to work on it but I'd prefer to wait until it's included in a Sprint before taking it on as I think there's higher priority issues. Perhaps we can close this PR for now and @acressall or I can reopen it if/when we work on it?

PR #6121, opened 2022-02-23
Fix TypeError: jiraProjects is not iterable

Dschoordsch:
This happened when jira integration is not added an TaskFooterMenu is
opened.

### DIFF
--- a/packages/server/dataloader/atlassianLoaders.ts
+++ b/packages/server/dataloader/atlassianLoaders.ts
@@ -98,19 +98,19 @@ export const allJiraProjects = (
   parent: RootDataLoader
 ): DataLoader<
   TeamUserKey,
-  (JiraProject & {cloudId: string; teamId: string; userId: string})[] | null,
+  (JiraProject & {cloudId: string; teamId: string; userId: string})[],
   string
 > => {
   return new DataLoader<
     TeamUserKey,
-    (JiraProject & {cloudId: string; teamId: string; userId: string})[] | null,
+    (JiraProject & {cloudId: string; teamId: string; userId: string})[],
     string
   >(
     async (keys) => {
       const results = await Promise.allSettled(
         keys.map(async ({userId, teamId}) => {
           const auth = await parent.get('freshAtlassianAuth').load({teamId, userId})
-          if (!auth) return null
+          if (!auth) return []
           const cloudNameLookup = await parent
             .get('atlassianCloudNameLookup')
             .load({teamId, userId})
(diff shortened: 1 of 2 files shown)


mattkrick:
+1 well that's odd, typescript should fail when trying to spread `null`.
before merging we should figure out why typescript didn't pick this up, otherwise it'll just keep happening.

mattkrick:
Here's the minimal repro I came up with. Am I missing anything obvious?


```ts
declare class DataLoader<K, V> {
  load(key: K): V
}

const fruitLoader = new DataLoader<void, number | null>()
const container = {fruitLoader}

const maybeFruit = fruitLoader.load()
const correctlyErrors = [...maybeFruit]

const maybeFruit2 = container.fruitLoader.load()
const whyNoError = [...maybeFruit2]  
```
https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEkoGdnwCJQC5QDIHspQYAeAaQBp4A1APngG8AoeeCQ4ACgGsQBPAFzwyASiHUmAXyZMw+AHbIs8AGYwArgEssBIiBjwAvPHkgA7hmx4O+kgDd8m4FXnqAtgCN98AD4n1EBC0nCKyCkqICjiapgbGDGpaOjYw0mGKym5QfF4AYhraRqoFyXowAHTsRCHpEXIwcGBYEHwAog34MGjGANrl-Vk5IPlJALoychnwg3klAExFk9Gx5YnausSVHDWTEWYAFnwAcvjtMJ1FfQPZs0lzo6xAA

mattkrick:
Found the problem, proposing to close this in favor of #6129

nickoferrall:
I'd like to merge this PR so that we can push to prod with this fixed. #6129 solves the root cause but it's a more substantial PR that'll need reviewing and won't make it into the ship. 

Any objections to me merging this PR as a temporary step?

Dschoordsch:
@nickoferrall let me test #6129 instead or if you deem it to risky, then lets just apply the changes in this file https://github.com/ParabolInc/parabol/pull/6129/files#r814422264

nickoferrall:
Nice, that's working for me. I think it's a good idea to use that file from https://github.com/ParabolInc/parabol/pull/6129 and limit changes.

The dropdown menu is currently broken on master when Jira isn't integrated. The change in this PR comes from Matt's #6129, meaning this change has been approved by Matt, Georg and me, so I think it's safe to merge this PR.

nickoferrall (approved):

PR #6122, opened 2022-02-23
Fix a bug where dragging item to original position cause an error

### LINKED ISSUE #6029: Bug: dropping agenda item/estimated tasks to original position causes error
In Sprint Poker, when dropping an estimate issue in its original position, there's an error in `PokerSidebarEstimateSection`: `Cannot read properties of null (reading 'index')`

poker-sidebar-bug

I can't reproduce this in production, but I can in master.

### AC

- [ ] Users can drag and drop estimate issues in their original position in the sidebar

tianrunhe:
Fixes #6029.

The issue is that when an item is dropped to its original position or somewhere invalid, the `destination` will be `null`. See this [discussion](https://github.com/atlassian/react-beautiful-dnd/issues/558#issuecomment-399205587).

Tests:

- [ ] In a poker meeting, user can drag & drop estimated tasks freely
- [ ] In a retro meeting, user can drag & drop discussion topics freely
- [ ] In team dashboard, user can drag & drop agenda items freely

### DIFF
--- a/packages/client/modules/teamDashboard/components/AgendaList/AgendaList.tsx
+++ b/packages/client/modules/teamDashboard/components/AgendaList/AgendaList.tsx
@@ -1,6 +1,6 @@
 import graphql from 'babel-plugin-relay/macro'
 import React, {useMemo} from 'react'
-import {DragDropContext, Draggable, Droppable} from 'react-beautiful-dnd'
+import {DragDropContext, Draggable, Droppable, DropResult} from 'react-beautiful-dnd'
 import {createFragmentContainer} from 'react-relay'
 import {AgendaList_meeting} from '~/__generated__/AgendaList_meeting.graphql'
 // import SexyScrollbar from 'universal/components/Dashboard/SexyScrollbar'
@@ -48,19 +48,17 @@ const AgendaList = (props: Props) => {
       : agendaItems.filter(({content}) => content)
   }, [dashSearch, agendaItems])
 
-  const onDragEnd = useEventCallback((result) => {
+  const onDragEnd = useEventCallback((result: DropResult) => {
     const {source, destination} = result
-    if (
-      !destination || !source ||
-      destination.droppableId !== AGENDA_ITEM ||
-      source.droppableId !== AGENDA_ITEM ||
-      destination.index === source.index
-    ) {
-      return
-    }
+    if (!destination) return
     const destinationItem = agendaItems[destination.index]
     const sourceItem = agendaItems[source.index]
-    if (!destinationItem || !sourceItem) {
+    if (
+      destination.droppableId !== AGENDA_ITEM ||
+      source.droppableId !== AGENDA_ITEM ||
+      destination.index === source.index ||
+      !destinationItem || !sourceItem
+    ) {
       return
     }
 
--- a/packages/client/modules/teamDashboard/components/AgendaList/AgendaList.tsx
+++ b/packages/client/modules/teamDashboard/components/AgendaList/AgendaList.tsx
@@ -51,7 +51,7 @@ const AgendaList = (props: Props) => {
   const onDragEnd = useEventCallback((result) => {
     const {source, destination} = result
     if (
-      !destination ||
+      !destination || !source ||
       destination.droppableId !== AGENDA_ITEM ||
       source.droppableId !== AGENDA_ITEM ||
       destination.index === source.index
--- a/packages/client/modules/teamDashboard/components/AgendaList/AgendaList.tsx
+++ b/packages/client/modules/teamDashboard/components/AgendaList/AgendaList.tsx
@@ -50,16 +50,17 @@ const AgendaList = (props: Props) => {
 
   const onDragEnd = useEventCallback((result) => {
     const {source, destination} = result
-    const destinationItem = agendaItems[destination.index]
-    const sourceItem = agendaItems[source.index]
     if (
       !destination ||
       destination.droppableId !== AGENDA_ITEM ||
       source.droppableId !== AGENDA_ITEM ||
-      destination.index === source.index ||
-      !destinationItem ||
-      !sourceItem
+      destination.index === source.index
     ) {
+      return
+    }
+    const destinationItem = agendaItems[destination.index]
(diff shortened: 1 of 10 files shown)


tianrunhe:
@nickoferrall in case you missed it. Thanks buddy!

packages/client/modules/teamDashboard/components/AgendaList/AgendaList.tsx
```diff
@@ -50,18 +50,19 @@ const AgendaList = (props: Props) => {
 
   const onDragEnd = useEventCallback((result) => {
```
nickoferrall, line 51: +1 I think TypeScript can save us here. We can get the `DropResult` type from `react-beautiful-dnd` and do `result: DropResult`. That tells us that `destination` can be undefined while `source` won't be. 

The bug happened on line 53: `agendaItems[destination.index]` as `destination` could be undefined. This new solution works because `agendaItems[destination.index]` is now happening after `!destination` on line 54. 

An alternative solution could be:

```
    const {source, destination} = result
    if (!destination) return
    const destinationItem = agendaItems[destination.index]
    const sourceItem = agendaItems[source.index]
    if (
      destination.droppableId !== AGENDA_ITEM ||
      source.droppableId !== AGENDA_ITEM ||
      destination.index === source.index ||
      !destinationItem ||
      !sourceItem
    ) {
      return
    }
```

packages/client/modules/teamDashboard/components/AgendaList/AgendaList.tsx
```diff
@@ -50,18 +50,19 @@ const AgendaList = (props: Props) => {
...
     const {source, destination} = result
-    const destinationItem = agendaItems[destination.index]
-    const sourceItem = agendaItems[source.index]
     if (
-      !destination ||
+      !destination || !source ||
```
nickoferrall, line 54: +1  TS says`source` can't be undefined so we can remove this

nickoferrall (changes requested):
Tested - can confirm the bug has been squished!

packages/client/modules/teamDashboard/components/AgendaList/AgendaList.tsx
```diff
@@ -50,18 +50,19 @@ const AgendaList = (props: Props) => {
...
       destination.droppableId !== AGENDA_ITEM ||
       source.droppableId !== AGENDA_ITEM ||
-      destination.index === source.index ||
-      !destinationItem ||
-      !sourceItem
+      destination.index === source.index
```
JimmyLv, line 57: Since we have repeated similar logic three times, should we be able to consider refactoring: to extract a common method e.g. `isInvaildDroppable(result, type)`, pseudo-code below:

```js
const isInvaildDroppable = ({ source, destination }, type) =>  {
     if (!destination) {
       return false
     }
      return destination.droppableId !== type ||
      source.droppableId !== type ||
      destination.index === source.index
}
```

packages/client/modules/teamDashboard/components/AgendaList/AgendaList.tsx
tianrunhe: I like this a lot!

packages/client/modules/teamDashboard/components/AgendaList/AgendaList.tsx
tianrunhe: I thought about this idea earlier but then you will have more undefined check after this so it doesn't necessarily reduce lots of codes.

Here, you need to check undefined for `destination` so you basically need to move that out of the `isInvaildDroppable` method. Essentially that method is just a wrapper around a if conditional check so 🤷‍♂️.

Screen Shot 2022-02-25 at 4 18 15 PM
:

packages/client/utils/constants.ts
```diff
@@ -40,9 +40,14 @@ export const meetingColumnArray = [DONE, ACTIVE, STUCK, FUTURE] as TaskStatusEnu
...
-export const TASK = 'task'
-export const AGENDA_ITEM = 'agendaItem'
+/* Drag-n-Drop Items DroppableId */
+export const AGENDA_ITEM = 'AGENDA_ITEM'
+export const DISCUSSION_TOPIC = 'DISCUSSION_TOPIC'
+export const ESTIMATING_TASK = 'ESTIMATING_TASK'
+export const TEMPLATE_DIMENSION = 'TEMPLATE_DIMENSION'
+export const TEMPLATE_PROMPT = 'TEMPLATE_PROMPT'
+export const TEMPLATE_SCALE_VALUE = 'TEMPLATE_SCALE_VALUE'
+export const ACTIVE_TASK = 'ACTIVE_TASK'
```
tianrunhe, line 50: I could have create an enum to encapsulate them together but decided to not further increase the scope of these changes.

packages/client/utils/constants.ts
```diff
@@ -40,9 +40,14 @@ export const meetingColumnArray = [DONE, ACTIVE, STUCK, FUTURE] as TaskStatusEnu
 /* Scoping Task Search Filter */
 export const taskScopingStatusFilters = [ACTIVE, STUCK, FUTURE] as TaskStatusEnum[]
 
-/* Drag-n-Drop Items */
-export const TASK = 'task'
```
tianrunhe, line 44: This `TASK` constant is used in `ClientGraphQLServer.ts`  and `LocalAtmosphere.ts` and thus is replaced by `SubscriptionChannel.TASK`. No instances were used in dnd related places.

packages/client/utils/constants.ts
```diff
@@ -116,11 +121,6 @@ export const GITHUB_ENDPOINT = 'https://api.github.com/graphql'
 /* JavaScript specifics */
 export const MAX_INT = 2147483647
 
-/* Relay Subscription Channels */
-export const NOTIFICATION = 'notification'
-export const ORGANIZATION = 'organization'
```
tianrunhe, line 121: No instances of these two constants were used.  `SubscriptionChannel.NOTIFICATION|ORGANIZATION` are used instead.

nickoferrall (approved):
LGTM!

mattkrick:
LGTM

PR #6129, opened 2022-02-25
Fix that weird bug where types from a dataloader aren't completely accurate

mattkrick:
@Dschoordsch @BartoszJarocki we found it!

Remember in the `noImplicitAny` work where we would sometimes have to cast within the predicate like `phases.find((phase: GenericMeetingPhase) =>...)` even though that shouldn't be necessary?

@Dschoordsch also found a weird bug where the dataloader would return something like `null | foo[]` and it wouldn't catch that it could be null. See #6121

This fixes that & in turn found a few extra bugs.

Root cause:
The CacheWorker had a `get` method that inherited its type signature via an abstraction.
This abstraction was used so we could decouple how we cache dataloaders & the shape of the dataloaders themselves
However, to do this, we had to declare that the type was an extension of _something_ so I made up an interface called `DataLoaderBase`
This works fine to return types, but I _assume_ typescript doesn't traverse to the child for control flow analysis.
The quick solution is to couple the RootDataLoader with the CacheWorker. Since we only have 1 RootDataLoader, this is fine for us.
Should we ever want to externalize the CacheWorker into a package & share it with the world, we'll have to revisit 🤷


packages/server/dataloader/atlassianLoaders.ts
```diff
@@ -122,11 +120,12 @@ export const allJiraProjects = (
...
             id: JiraProjectId.join(project.cloudId, project.key),
             userId,
-            teamId
+            teamId,
+            service: 'jira' as const
           }))
         })
       )
-      return results.map((result) => (result.status === 'fulfilled' ? result.value : null))
+      return results.map((result) => (result.status === 'fulfilled' ? result.value : []))
```
mattkrick, line 128: whenever a dataloader return an array for a single key I like to return an empty array. Feels more like a database that way (return 0 rows vs. return null)

packages/server/graphql/DataLoaderCache.ts
```diff
@@ -22,8 +23,13 @@ export class CacheWorker<T extends DataLoaderBase> {
     this.cache = cache
   }
 
-  get: T['get'] = (dataLoaderName) => {
-    return this.dataLoaderBase.get(dataLoaderName)
+  get = <LoaderName extends Loaders>(dataLoaderName: LoaderName) => {
```
mattkrick, line 26: this was the source of all the hardship

packages/server/graphql/mutations/addTeamMemberIntegrationAuth.ts
```diff
@@ -85,8 +85,8 @@ const addTeamMemberIntegrationAuth = {
       }
     } else if (scope === 'org' && teamId !== integrationProvider.teamId) {
       const [providerTeam, authTeam] = await Promise.all([
-        dataLoader.get('teams').load(integrationProvider.teamId),
-        dataLoader.get('teams').load(teamId)
+        dataLoader.get('teams').loadNonNull(integrationProvider.teamId),
```
mattkrick, line 88: kudos to @Dschoordsch for this nice method. now it's catching all the possible nulls!

packages/server/graphql/mutations/createOAuth1AuthorizeUrl.ts
```diff
@@ -55,6 +55,8 @@ export default {
         url
       }
     }
-    return standardError(new Error('Service not supported'), {tags: {service: provider.service}})
+    return standardError(new Error('Service not supported'), {
```
mattkrick, line 58: technically, this is unreachable, but leaving it in in case we add new types or forget to extend one

packages/server/graphql/resolvers/resolveStage.ts
```diff
@@ -1,21 +1,22 @@
...
+const resolveStage =
+  (phaseType: NewMeetingPhaseTypeEnum) =>
+  async (
+    {meetingId, stageId}: {meetingId: string; stageId: string},
+    _args: unknown,
+    {dataLoader}: GQLContext
+  ) => {
+    const meeting = await dataLoader.get('newMeetings').load(meetingId)
+    const {phases, teamId} = meeting
+    const phase = phases.find((phase: GenericMeetingPhase) => phase.phaseType === phaseType)!
```
Dschoordsch, line 16: +1 `: GenericMeetingPhase` could be removed here, too

Dschoordsch (approved):
Gave it a quick test and it resolves the initial issue. Awesome that you found this!

BartoszJarocki:
@mattkrick amazing, thank you for finding that out!

PR #6148, opened 2022-03-02
Allow admins to call set org user role mutation

BartoszJarocki:
This change will allow admin users to set org billing users using the following mutation
```
mutation {
   setOrgUserRole(orgId: "orgId", userId: "local|botpRdWlzi", role: "BILLING_LEADER") {
      updatedOrgMember {
         user {
           email
         }
       }
   }
}
```

context: https://parabol.slack.com/archives/C4JAUUZ9P/p1646147430342679

### DIFF
--- a/packages/server/graphql/mutations/setOrgUserRole.ts
+++ b/packages/server/graphql/mutations/setOrgUserRole.ts
@@ -3,7 +3,7 @@ import {SubscriptionChannel} from 'parabol-client/types/constEnums'
 import getRethink from '../../database/rethinkDriver'
 import NotificationPromoteToBillingLeader from '../../database/types/NotificationPromoteToBillingLeader'
 import {OrgUserRole} from '../../database/types/OrganizationUser'
-import {getUserId, isUserBillingLeader} from '../../utils/authorization'
+import {getUserId, isSuperUser, isUserBillingLeader} from '../../utils/authorization'
 import publish from '../../utils/publish'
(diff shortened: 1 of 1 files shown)


packages/server/graphql/mutations/setOrgUserRole.ts
```diff
@@ -38,8 +38,13 @@ export default {
 
     // AUTH
     const viewerId = getUserId(authToken)
-    if (!(await isUserBillingLeader(viewerId, orgId, dataLoader, {clearCache: true}))) {
-      return standardError(new Error('Must be the organization leader'), {userId: viewerId})
+    if (
+      !isSuperUser(authToken) ||
+      !(await isUserBillingLeader(viewerId, orgId, dataLoader, {clearCache: true}))
```
tianrunhe, line 43: If I **AM** a super user but **NOT** the billing leader for the org, this will still be true and I can't run the mutation

tianrunhe (changes requested):

packages/server/graphql/mutations/setOrgUserRole.ts
BartoszJarocki: oh. this should not have happened. thank you!

tianrunhe (approved):

BartoszJarocki:
@mattkrick would you mind taking a look? 🙏

mattkrick:
LGTM!

PR #6152, opened 2022-03-02
fix #6043 only fetch new tasks

### LINKED ISSUE #6043: Poker: updatePokerScope should not reach out to the integration
While debugging #6015 I noticed that every time an item is added or removed from the scope, all selected items are fetched from the integration! This is not necessary. I assume the mutation payload includes a fragment that is like `{task { integration {...} } }` and it shouldn't have that. This causes our app to feel slow while selecting items.

AC
- [ ] updating the poker scope for integrated items does not reach out to jira or github

Estimate: 4 hours

mattkrick:
fix #6043

Before, every add/remove to the poker scope would trigger a refetch of all stages, including the task for that stage, which included the integration associated with that task. That means if I had 50 stages & removed one, it'd reach out to jira or github 49 times just to remove the stage!

Now, it only reaches out to integrations for new stages, which makes sense because it needs to grab the title, description, etc. 
For removals, it doesn't reach out to any integrations.

TEST
- [ ] add 3 jira issues, look at the websocket response payload for the 3rd issue added. notice that newStages is an array that only includes 1 item per dimension

### DIFF
--- a/packages/client/hooks/useMakeStageSummaries.ts
+++ b/packages/client/hooks/useMakeStageSummaries.ts
@@ -14,33 +14,39 @@ interface StageSummary {
   finalScores: (string | null)[]
 }
 
+graphql`
+  fragment useMakeStageSummaries_stages on EstimateStage @relay(plural: true) {
+    id
+    finalScore
+    isComplete
+    isNavigable
+    sortOrder
+    taskId
+    task {
+      title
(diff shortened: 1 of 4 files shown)


mattkrick:
Effort: 1 hour

tianrunhe (approved):

PR #6154, opened 2022-03-03
feat(analytics): support querying organization for superusers

mattkrick:
I realized i couldn't get the name of an organization from graphiql, now i can.

TEST
```gql
viewer {
  organization(orgId: "foo") {
    name
  }
}
```
- [ ] the above query works for an organization that you're not a part of as long as you're a super user

### DIFF
--- a/packages/server/graphql/queries/organization.ts
+++ b/packages/server/graphql/queries/organization.ts
@@ -1,8 +1,7 @@
-import {GQLContext} from './../graphql'
 import {GraphQLID, GraphQLNonNull} from 'graphql'
+import {getUserId, isSuperUser} from '../../utils/authorization'
 import Organization from '../types/Organization'
-import {getUserId} from '../../utils/authorization'
-import OrganizationUser from '../../database/types/OrganizationUser'
+import {GQLContext} from './../graphql'
 
 export default {
   type: Organization,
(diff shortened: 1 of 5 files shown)


packages/server/graphql/queries/organization.ts
```diff
@@ -12,26 +11,18 @@ export default {
...
-    if (viewerId === userId) return organization
-
-    const viewerOrganizationUsers = await dataLoader.get('organizationUsersByUserId').load(userId)
-    const viewerOrganizationUser = viewerOrganizationUsers.find(
-      (organizationUser: OrganizationUser) => organizationUser.orgId === orgId
-    )
-    return viewerOrganizationUser ? organization : null
+    const [organization, viewerOrganizationUser] = await Promise.all([
+      dataLoader.get('organizations').load(orgId),
+      dataLoader.get('organizationUsersByUserIdOrgId').load({userId: viewerId, orgId})
```
Dschoordsch, line 23: +1 While this reduces the amount of loading required here, if any of the `OrganizationUser` fields are queried from `User` we will load `organizationUsersByUserId` again. We could at least change `User.organizationUser` to use the same dataloader.
On the other hand it is unlikely to be queried this way, up to you.

Dschoordsch (approved):
Works

packages/server/graphql/queries/organization.ts
mattkrick: you're absolutely right. I noticed `organizationUsersByUserId` is an overfetch in most use cases since it's quickly followed by a `find`. at the risk of making this PR too big i'll go through & clean em up

Issue #6088, opened 2022-02-21
Sprint Poker: GitLab v1: Issues functionality 


nickoferrall:
Predecessor: #6034 

### AC:

- [ ] As the user scrolls the list, more results are loaded
- [ ] Clicking an item adds the issue to the scope
- [ ] Each issue shows the issue number in the subtitle
- [ ] The select all functionality works
- [ ] Issues are sorted deterministically. We could use the `lastActivityAt` field

mattkrick:
From 1:1

- As a first pass, we only allow 2 search criteria: search (ie text in title & description) and project (which is an array of projectIds selectable from the dropdown)
- the problem is the default query-- gitlab doesn't allow us to do a search among issues without specifying a project
- since GitLab provides a GraphQL endpoint, we'd like to use that without writing any custom code on our backend

```gql
query getIssuesByProjectIdAndSearch($search: String, $projectIds: [ID!], $isProjectFiltered: Boolean!, $first: Int, $after: String) {
  # If they have selected at least 1 project filter, then use that filter
  projects(ids: $projectIds) @include(if: $isProjectFiltered) {
    ...nodeSearch_nodes
  }
  # If they haven't selected a project to filter on, filter on the 3 most recently updated projects
  defaultProjects: projects(membership: true, first: 3, sort: "UPDATED_DESC") @skip(if: $isProjectFiltered) {
    ...nodeSearch_nodes
  }
}

fragment nodeSearch_nodes on ProjectConnection {
  nodes {
    issues(
      includeSubepics: true
      state: opened
      search: $search
      sort: UPDATED_DESC
      first: $first
      after: $after
    ) {
      nodes {
        id
        descriptionHtml
        title
      }
    }
  }
}

```

Out of scope:
- Currently the only way to see the project filters is to open the filter dropdown. this probably isn't great. we should see active filters without having to click a dropdown box. Nick says, "maybe we put little chits below the search input field?" CC @enriquesanchez
![Screen Shot 2022-02-23 at 10 51 35 AM](https://user-images.githubusercontent.com/5514175/155387432-18acbe8f-c97c-484e-a490-3756b007c8e8.png)

enriquesanchez:
> Nick says, "maybe we put little chits below the search input field?" 

This is a great idea.

One thing that bothers me here is the lack of a border for the search field. It makes it difficult to correlate the field with the filter button. We could add a simple border around the whole thing and increase the clarity of the component, something like this:

[...]

Issue #6046, opened 2022-02-14
Sprint Poker: GitLab v1: Project filter menu


nickoferrall:
Predecessor: https://github.com/ParabolInc/parabol/issues/6045
Successor: https://github.com/ParabolInc/parabol/issues/6138

### AC:

- [ ] There's a filter button on the right-hand side of the search input, just like GitHub, Jira, and Parabol tasks
- [ ] Clicking the filter button opens a dropdown menu that contains all of the projects that the user has access to
- [ ] It is sorted in a sensible, deterministic fashion. `lastActivityAt` could be a good field to sort by
- [ ] Selecting a project in the menu filters the issues displayed in the card
- [ ] If there’s more than 10 projects, include a search box just like Jira & GitHub
- [ ] As the user types, the projects displayed in the menu are filtered

nickoferrall:
**Story points: 13**
  [See the discussion](https://action.parabol.co/meet/chrwZ1fHoc/estimate/13) in Sprint Poker #​​1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

nickoferrall:
**Include in next sprint?: Y**
  [See the discussion](https://action.parabol.co/meet/chrwZ1fHoc/estimate/14) in Sprint Poker #​​1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

mattkrick:
fixed in #6323

PR #6160, opened 2022-03-03
GitLab Issues Functionality

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6088

Note: in the AC, it says that we should be able to select all issues. This was working when I created the PR, but since then this seems to be borked in prod. I've created an issue to fix this https://github.com/ParabolInc/parabol/issues/6213 and have added it to the next Sprint Poker meeting.

Also, in the Estimate phase, the issue is not currently visible in the card at the top of the page. This will be added here: https://github.com/ParabolInc/parabol/issues/6244


packages/client/components/GitLabScopingSearchResults.tsx
```diff
@@ -94,16 +82,35 @@ const GitLabScopingSearchResults = (props: Props) => {
...
                   }
                   query {
-                    projects(membership: true, first: $count, after: $cursor)
-                      @connection(key: "GitLabScopingSearchResults_projects") {
+                    projects(
+                      membership: true
+                      first: $projectsFirst
+                      after: $projectsAfter
+                      sort: "latest_activity_desc"
+                      ids: null # $projectIds
```
nickoferrall, line 90: Originally, we discussed using the skip & include directives because the projects query would be null if `ids` is an empty array. However, when `ids` is `null`, the query works and we can successfully search issues. 

So, I suggest we don't use skip or include and just have one query that accepts the `ids` argument. If the user hasn't selected any projects, `ids` is null. If they have, then we have an array of project `ids`.

One problem that I've run into is when scrolling down the list, triggering the Intersection Observer, and grabbing the next results, the data returned from the query is `null`. I suspect that the `useLoadNextOnScrollBottom` code may need to be tweaked, but this can be done when implementing the project filter menu here: https://github.com/ParabolInc/parabol/issues/6046. For now, I've used `null` as a placeholder.

packages/client/components/GitLabScopingSearchResults.tsx
```diff
@@ -78,7 +59,14 @@ const GitLabScopingSearchResults = (props: Props) => {
   >(
     graphql`
```
nickoferrall, line 60: The query is a little long and could be moved to its own file, but I've kept it here to be consistent with the GitHub logic, and so that it's easily accessible

packages/client/components/GitLabScopingSearchResults.tsx
```diff
@@ -78,7 +59,14 @@ const GitLabScopingSearchResults = (props: Props) => {
...
       fragment GitLabScopingSearchResults_query on Query
-        @argumentDefinitions(cursor: {type: "String"}, count: {type: "Int", defaultValue: 50})
+        @argumentDefinitions(
+          projectsFirst: {type: "Int", defaultValue: 5}
+          issuesFirst: {type: "Int", defaultValue: 25}
```
nickoferrall, line 64: We can't add a connection to issues as connections can't appear inside plural fields like `projects`. So, I've added the connection to the projects, and I'm getting the first 25 issues in each project.

If the issue that the user is looking for isn't included within that list, they'll be able to search for the title or description to be able to find it.

packages/server/graphql/nestedSchema/nestGitLabEndpoint.ts
```diff
@@ -21,7 +21,8 @@ const defaultExecutor: Executor<{
     controller.abort()
   }, endpointTimeout)
   try {
-    const result = await fetch(`${baseUri}/api/graphql`, {
+    const uri = baseUri || 'https://gitlab.com' // TODO: why is baseUri undefined when adding issues to the Estimate phase?
```
nickoferrall, line 24: When adding an issue to the Estimate phase, the `baseUri` that we get from `context` is null. I've played around with the nestGitLab logic, but I haven't discovered why yet. Possibly an issue with the package?

packages/server/graphql/types/Task.ts
```diff
@@ -181,6 +182,28 @@ const Task: GraphQLObjectType = new GraphQLObjectType<any, GQLContext>({
...
+          const accessToken = gitlabAuth?.accessToken
+          if (!accessToken) return null
+          const {gid} = integration
+          const query = `
+          {
+             issue(id: "${gid}"){
+                   __typename
+                   id
+                   iid
+                   title
```
nickoferrall, line 198: When we query `xGitLabIssue` on the frontend, we use the GitLab schema which says that several fields, such as `author.id`, are non-nullable. However, these are nullable fields for us as we're not returning them from this `Task` query. 

There are over a dozen non-nullable fields for GitLab issues. If we want to return all of them, the query would be very expensive. We'd need to do this in the `UpdatePokerScopeMutation` optimistic updater too. This also applies to our GitHub logic.

I think it's the right approach to only return the fields that we're likely to use, but nullable fields disguised as non-nullable fields certainly fields error prone. I know that we're refactoring the way that task integrations work, so this might be worth considering.

igorlesnenko:
@nickoferrall I'm trying to run the PR, but I see no issues displayed. What I might be doing wrong?

image

nickoferrall:
@igorlesnenko sometimes the auth can get stale. Have you tried reintegrating with GitLab and refreshing the page?

igorlesnenko:
@nickoferrall yeah, parabol1. It looks like it fetches only the first 5 projects, and they do not have any issues. Is it expected behavior?
Changing query  to `projectsFirst: {type: "Int", defaultValue: 100}` fixes an issue for me.

nickoferrall:
@igorlesnenko thanks for letting me know. Using the connection, it should re-run the query with a larger `first` value when there aren't enough issues to fill the list. Perhaps this only happens when there are some but not many issues within the first 5 projects. I'll take a look 👍

packages/client/components/GitLabScopingSearchResults.tsx
```diff
@@ -120,89 +127,71 @@ const GitLabScopingSearchResults = (props: Props) => {
...
+  const estimatePhase = phases.find(({phaseType}) => phaseType === 'ESTIMATE')!
+  const usedServiceTaskIds = useGetUsedServiceTaskIds(estimatePhase)
+  const handleAddIssueClick = () => setIsEditing(true)
 
-  // even though it's a little herky jerky, we need to give the user feedback that a search is pending
-  // TODO fix flicker after viewer is present but edges isn't set
+  // when there are no issues in the initial query, gitlab can return an error with hasMore=false even though there are more projects
+  const noProjectsWithIssuesErrMsg = 'The user aborted a request.'
+  const errorMessage = errors?.[0]?.message
+  if (errorMessage === noProjectsWithIssuesErrMsg && !isLoadingNext) {
```
nickoferrall, line 174: When integrating with Parabol1, if there are no issues in the initial query, GitLab returns an error message. To test:

- in the pagination query, change the default value of `projectsFirst` to 1
- change the sort to `sort: "latest_activity_asc"`

You can play with the [explorer](https://gitlab.com/-/graphql-explorer) to experiment with this too.

I created a new test GitLab project, didn't add any issues, and ran the same query without getting an error message.

I don't like this solution as I'm not sure why the error happens with Parabol1, and the GitLab API could change the error message meaning `loadNext` wouldn't execute, so I'm open to other ideas. I've increased the default `projectsFirst` value from 5 to 10 to reduce the chance that this happens.

nickoferrall:
Hey @igorlesnenko, I've pushed an update so you can see some feedback while the query is executing, and it will run again if there are no issues returned from the initial query

igorlesnenko:
@nickoferrall reconnected GitLab and now it works but takes some time to load the issues list. I guess several queries are needed

igorlesnenko (commented):
Looks good to me. The only thing that does not work for me is select all functionality. 
When I click on it, I get the following result
image
@nickoferrall Does it work for you?

packages/client/components/GitLabScopingSearchResults.tsx
```diff
@@ -78,7 +59,14 @@ const GitLabScopingSearchResults = (props: Props) => {
   >(
     graphql`
       fragment GitLabScopingSearchResults_query on Query
-        @argumentDefinitions(cursor: {type: "String"}, count: {type: "Int", defaultValue: 50})
+        @argumentDefinitions(
+          projectsFirst: {type: "Int", defaultValue: 20}
```
nickoferrall, line 63: @mattkrick in our 1-1, I mentioned that Igor and I ran into an issue where GitLab would return a server error depending on the `projectsFirst` value.

For me, if the `projectsFirst` number is 15 or lower and the sort order is `latest_activity_asc`, I get the server error. In the GQL explorer, the message is `Internal server error`. In our codebase, I can repeat the same result, but the error is `The user aborted a request`. If `projectsFirst` is greater than 15, it always works.

However, if I sort by `latest_activity_asc`, it doesn't matter what the `projectsFirst` value is, it always works. 

It seems unlikely that a borked Parabol project is causing the issue when `projectsFirst` is 15, because it's still included in the successful results when it's 16.

It seems like there's a bug in GitLab's API. Googling hasn't helped yet, but I'll keep trying tomorrow. 

Here's the query FYI. If I query anything other than `__typename`, I get the error:

```
query {
    projects(membership: true, first:15, sort: "latest_activity_asc") {
    nodes {
      createdAt
    }
  }
}  
```

mattkrick (changes requested):

packages/client/utils/parseWebPath.ts
```diff
@@ -0,0 +1,8 @@
+import {upperFirst} from './upperFirst'
+
+export const parseWebPath = (webPath: string) => {
```
nickoferrall, line 3: @mattkrick you mentioned in your CR ([here](https://github.com/ParabolInc/parabol/pull/6160#discussion_r829255475)) that we should just parse, rather than parse, filter, and concatenate. 

However, if we just parse and return `[fullPath, issuesAndIid]`, those values will always need to edited afterwards. Doing this in multiple places feels more error-prone that doing it all here imo

packages/server/graphql/types/Task.ts
```diff
@@ -181,6 +182,31 @@ const Task: GraphQLObjectType = new GraphQLObjectType<any, GQLContext>({
...
+          if (!accessToken) return null
+          const {gid} = integration
+          const query = `
+            query {
+              issue(id: "${gid}"){
+                __typename
+                id
+                iid
+                title
+                ...info
```
nickoferrall, line 199: You mentioned [here](https://github.com/ParabolInc/parabol/pull/6160#discussion_r829211225) that we should use `...info` here as it'll pass the requested fields from the client to GitLab. 

Did you mean that we should include `...info` in addition to the other fields? Otherwise, we won't have access to those fields. 

To test: remove the other fields, like `id` and `title`, add an issue to the Estimate phase, and see that the title doesn't show up in the sidebar

packages/server/integrations/gitlab/GitLabServerManager.ts
```diff
@@ -1,22 +1,20 @@
...
 class GitLabServerManager {
-  readonly provider: IntegrationProviderGitLabOAuth2
   readonly accessToken: string
   readonly serverBaseUrl: string
 
-  constructor(accessToken: string, serverBaseUrl: string) {
+  constructor(accessToken: string) {
     this.accessToken = accessToken
-    this.serverBaseUrl = serverBaseUrl
+    this.serverBaseUrl = 'https://gitlab.com'
```
nickoferrall, line 12: `serverBaseUrl` will always be `gitlab.com` so we don't need to include this as an arg? Instead, I've added it here

packages/server/graphql/types/Task.ts
mattkrick: The `...info` should be enough, could you do a little triage and see if those fields are being included in the info.fieldNodes? perhaps a fragment is missing in the request somewhere? might be worth comparing the fragment for github and the one for gitlab to determine the delta

packages/server/integrations/gitlab/GitLabServerManager.ts
mattkrick: that's not true for self-hosted versions

packages/server/graphql/types/Task.ts
mattkrick: you know, i'm not sure i love that design decision either... perhaps it should return the whole thing instead of just the info object? My goal was to make it hard to include fields in addition to the `...info`, but it may just be a gotcha

mattkrick:
looking good! loved the attention to detail troublshooting the `...info` stuff & refactoring the webPath

PR #6161, opened 2022-03-03
chore(noImplicitAny): fix more errors

### LINKED ISSUE #5831: Add noImplicitAny to server tsconfig (Part 6 of 8)

- [ ] fix ~200 errors
- [ ] we'll merge it blindly
- [ ] create a copy of this issue named Part (3 of 8) so we can re-prioritize for the next person. We'll keep doing this until it's done

AC

Estimate: 4 hours

mattkrick:
fix #5831

notes:
- removed stripe.js in favor of the new StripeManager, quicker than writing types for the legacy file


packages/client/modules/demo/handleCompletedDemoStage.ts
```diff
@@ -61,9 +61,10 @@ const addStageToBotScript = (stageId: string, db: RetroDemoDB, reflectionGroupId
         }))
       )
     }
-    const comment = commentLookup[reflection.id]
-    if (comment) {
-      comments.push(comment)
+    const comment = commentLookup[reflection.id as keyof typeof commentLookup]
+    if (Array.isArray(comment)) {
+      // I think this was a bug before...
```
mattkrick, line 66: noImplicitAny caught a bug here

packages/client/modules/email/components/SummaryEmail/ExportToCSV.tsx
```diff
@@ -1,5 +1,6 @@
 import graphql from 'babel-plugin-relay/macro'
-import Parser from 'json2csv/lib/JSON2CSVParser' // only grab the sync parser
+import type {Parser as JSON2CSVParser} from 'json2csv'
```
mattkrick, line 2: i don't usually import type, but i wanted to be sure we aren't imported the whole package because it's a big one

packages/client/types/constEnums.ts
```diff
@@ -307,14 +307,6 @@ export const enum SubscriptionChannel {
   ORGANIZATION = 'organization'
 }
 
-export const enum SuggestedActionTypeEnum {
```
mattkrick, line 310: we now have a single source of truth for these in the graphql enum

packages/server/database/migrations/20161125141100-addOrgs.js
```diff
@@ -119,21 +118,23 @@ exports.up = async (r) => {
...
-        metadata: customer.metadata,
-        plan: 'parabol-pro-monthly',
-        quantity: Object.keys(orgs[customer.metadata.orgId].orgUserMap).length,
-        trial_period_days: TRIAL_PERIOD_DAYS
-      })
-    })
-  )
+  // turned into a noop to remove dependency on stripe
+  const subscriptions = []
+  // const orgIds = Object.keys(orgs)
```
mattkrick, line 123: deleted the stripe helper, which this depends on, so had to remove any logic using it.
in the future, we shouldn't do stripe things in migrations, but this was 6 years ago

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/MeetingMembersWithoutTasks.tsx
```diff
@@ -18,9 +17,9 @@ const headerStyle = {
   paddingTop: 24
 }
 
-const getHeaderText = (meetingType) => {
+const getHeaderText = (meetingType: 'retrospective' | 'action' | 'poker') => {
```
JimmyLv, line 20: shall we use the same way as `TSuggestedActionTypeEnum`, to use the `MeetingTypeEnum` type extracted from `GraphQLEnumType`?

packages/client/types/constEnums.ts
JimmyLv: yeah, single source of truth! the `keyof typeof` is really helpful when doing TypeScript legacy support.
btw: have we tried any migration tool like [ts-migrate](https://github.com/airbnb/ts-migrate) before? I'm curious about how it will actually work.

packages/client/utils/date/relativeDate.ts
```diff
@@ -52,16 +52,16 @@ const relativeDate = (date: string | Date, opts: Opts = {}) => {
     day: (abs % MONTH) / DAY,
     hour: (abs % DAY) / HOUR,
     minute: (abs % HOUR) / MIN
-  }
+  } as const
 
-  let k
-  let val
+  let k: any
+  let val: any
```
nickoferrall, line 58: +1 would `k` and `val` both be `number | string`?

packages/server/graphql/queries/helpers/makeUpcomingInvoice.ts
```diff
@@ -22,10 +22,10 @@ export default async function makeUpcomingInvoice(
     amountDue: stripeInvoice.amount_due,
     total: stripeInvoice.total,
     endAt: fromEpochSeconds(stripeInvoice.period_end),
-    invoiceDate: fromEpochSeconds(stripeInvoice.date),
+    invoiceDate: fromEpochSeconds(stripeInvoice.date!),
     orgId,
     startAt: fromEpochSeconds(stripeInvoice.period_start),
-    startingBalance: stripeInvoice.startingBalance,
+    startingBalance: stripeInvoice.starting_balance,
```
nickoferrall, line 28: +2 TypeScript FTW 🎉

packages/server/graphql/types/AtlassianIntegration.ts
```diff
@@ -102,11 +107,8 @@ const AtlassianIntegration = new GraphQLObjectType<any, GQLContext>({
             'A list of projects to restrict the search to. format is cloudId:projectKey. If null, will search all'
         }
       },
-      resolve: async (
-        {teamId, userId, accessToken, cloudIds}: AtlassianAuth,
-        {first, queryString, isJQL, projectKeyFilters},
-        {authToken}: GQLContext
-      ) => {
+      resolve: async ({teamId, userId, accessToken, cloudIds}, args: any, {authToken}) => {
```
nickoferrall, line 110: I was wondering why you didn't do `args: IssueArgs` and remove the casting below, but I see it makes the `field` type incompatible. Annoying that you can't define the args type in a `GraphQLObjectType`

nickoferrall (approved):
LGTM!

packages/client/modules/demo/handleCompletedDemoStage.ts
mattkrick: all the values in `commentLookup` are arrays. perhaps that was an error? i didn't want to change that though

packages/client/types/constEnums.ts
mattkrick: the last time i checked it out it was still a little buggy & at best it would just explicitly cast things as `any` as a fallback.
this little exercise probably took the team 30-40 hours to complete, but we uncovered some cool bugs & got some really useful patterns (like keyof typeof) in place so future developers have some good patterns to go off of.

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/MeetingMembersWithoutTasks.tsx
mattkrick: yep! the relay-compiler gives us those enums in the generated files so we don't have to reference the server from the client so I'll just pull from there.

```ts
import {
  MeetingMembersWithoutTasks_meeting,
  MeetingTypeEnum
} from 'parabol-client/__generated__/MeetingMembersWithoutTasks_meeting.graphql'
```

packages/client/modules/email/components/EmailBlock/EmailBlock.tsx
```diff
@@ -26,7 +26,7 @@ const innerStyle = {
   width: '100%'
 } as const
 
-const EmailBlock = (props) => {
+const EmailBlock = (props: any) => {
```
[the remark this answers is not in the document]
mattkrick: this component is HTML4 😨  
that's the only way to write components so they work in email

PR #6167, opened 2022-03-04
More typesafety for GraphQL code

### LINKED ISSUE #4754: Create GraphQL ping
​

AC

- [ ] ping GQL, returns an object. each key is a database (redis, rethink, pg, no db). - - [ ] object must return even if a DB hangs, throws, etc. within 3 seconds

​

Estimate: 8 hours

mattkrick:
Fix #4754 
- [ ] Run the following on the private schema & see the response times for various services
```gql
query {
  ping {
    rethinkdb
    postgres
  	redis
  }
}
```

Today, we don't get a lot of type safety when we create objects with graphql
- We must remember to add types to the `source` 
- args are broken until we upgrade to v16
- We have to remember to include the context
- The return value of a resolver (e.g. `{taskId}`) isn't typed at all

So, i took a shot at what all the cool kids do these days.
- The schema is created using an SDL. You write GraphQL & it just creates the objects for you. All that's left are the resolvers
- codegen turns that SDL into types that you can apply to your resolvers
- the source value of a resolver (the first param, e.g. `{taskId}`) is called a "Model", which can be thought of like a database object vs. an object returned from graphql. that separation is nice. I'm calling them Sources in this PR so the name matches what GraphQL calls them.

Reading material:
- https://www.graphql-code-generator.com/plugins/typescript-resolvers
- https://www.graphql-tools.com/docs/schema-merging
- https://www.graphql-tools.com/docs/resolvers

Would love to get some feedback!


packages/server/graphql/intranetSchema/resolvers/ping.ts
```diff
@@ -0,0 +1,7 @@
+import {QueryResolvers} from '../resolvers/types'
+
+export interface PingPayload {}
```
mattkrick, line 3: not the greatest example, but say we wanted to just return `{taskId}`, we could define that here

packages/server/graphql/intranetSchema/typeDefs/ping.graphql
```diff
@@ -0,0 +1,10 @@
+union PingPayload = ErrorPayload | PingSuccess
```
mattkrick, line 1: is writing a file like this easier than the old way?

scripts/dev.js
```diff
@@ -45,7 +45,10 @@ const dev = async (maybeInit) => {
     await Promise.all([removeArtifacts()])
   }
   // Enable this if you're creating new github schemas
-  // await generate(codegenSchema)
+  try {
+    await generate(codegenSchema)
```
mattkrick, line 49: note: need to figure out how to get eslint to ignore fixing up the autogenerated files before we uncomment this out for good

packages/server/graphql/intranetSchema/sdl/resolvers.ts
```diff
@@ -0,0 +1,26 @@
...
+const importAll = (context: __WebpackModuleApi.RequireContext) => {
+  const collector = {} as Record<string, any>
+  context.keys().forEach((relativePath) => {
+    const {name} = path.parse(relativePath)
+    collector[name] = context(relativePath).default
+  })
+  return collector
+}
+
+const resolverMap: Resolvers = {
```
mattkrick, line 16: @blimmer this is my attempt to emulate what you were doing with that codegen script

codegen.json
```diff
@@ -12,6 +12,17 @@
       "schema": "packages/server/graphql/nestedSchema/GitLab/gitlabSchema.graphql",
       "documents": "packages/server/graphql/nestedSchema/GitLab/queries/*.graphql",
       "plugins": ["typescript", "typescript-operations", "add"]
+    },
+    "packages/server/graphql/intranetSchema/sdl/resolverTypes.ts": {
```
blimmer, line 16: question: how do we get rid of this SDL folder?

blimmer (commented):
I'm submitting these initial ideas for now, going to shift back over to #6183 to complete that work, then come back over here to make additional progress.

codegen.json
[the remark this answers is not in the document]
mattkrick: added to gitignore

packages/server/graphql/intranetSchema/sdl/resolvers.ts
mattkrick: my current thinking is to have 3 folders inside `/graphql`:
- public
  - public/queries
  - public/mutations
  - public/subscriptions
  - public/types
- private
  - private/queries
  - private/mutations
  - private/types
- github
  - queries
  - mutations
- gitlab
  - queries
  - mutations

the above won't touch the current hierarchy so we can gradually move over to this new format.

codegen.json
```diff
@@ -12,6 +12,16 @@
...
       "documents": "packages/server/graphql/nestedSchema/GitLab/queries/*.graphql",
       "plugins": ["typescript", "typescript-operations", "add"]
+    },
+    "packages/server/graphql/intranetSchema/resolvers/types.ts": {
+      "schema": "packages/server/graphql/intranetSchema/intranetSchema.graphql",
+      "plugins": ["typescript", "typescript-resolvers"],
+      "config": {
+        "mapperTypeSuffix": "Model",
+        "mappers": {
+          "PingPayload": "./ping#PingPayload"
```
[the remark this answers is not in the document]
Dschoordsch: This part is a bit annoying, but currently we mostly use `any` for it.

codegen.json
```diff
@@ -12,6 +12,17 @@
       "schema": "packages/server/graphql/nestedSchema/GitLab/gitlabSchema.graphql",
       "documents": "packages/server/graphql/nestedSchema/GitLab/queries/*.graphql",
       "plugins": ["typescript", "typescript-operations", "add"]
+    },
+    "packages/server/graphql/intranetSchema/sdl/resolverTypes.ts": {
+      "schema": "packages/server/graphql/intranetSchema/intranetSchema.graphql",
+      "plugins": ["typescript", "typescript-resolvers", "add"],
+      "config": {
+        "contextType": "../../graphql#InternalContext",
+        "mappers": {
```
Dschoordsch, line 21: +1 We might want to set `"defaultMapper": void` or similar to avoid the case where people don't define it.
We could even define `export type No_model_added_in_codegen_json = {}` and set it as default mapper, just so developers are pointed to it more directly.

packages/server/graphql/intranetSchema/sdl/types/PingableServices.ts
```diff
@@ -0,0 +1,48 @@
...
+
+const pingService = async <TSuccess>(thunk: () => Promise<TSuccess>) => {
+  const start = Date.now()
+  const res = await Promise.race([thunk(), sleep(5000)])
+  const end = Date.now()
+  const duration = end - start
+  return res ? duration : -1
+}
+
+export type PingableServicesSource = Record<string, never>
```
Dschoordsch, line 15: -1 I think if there is no input, it should be `undefined` or `void` instead.

Dschoordsch (changes requested):
Looks good overall. The source mapping in codegen.json is a bit annoying and having definitions spread across more files, but overall it's probably the way to go.

I wish there was something like [typebox](https://github.com/sinclairzx81/typebox) for GraphQL instead, but the GraphQL types don't contain enough information and with codegen we're on a more mainstream path.

packages/server/graphql/intranetSchema/intranetSchema.graphql
```diff
@@ -0,0 +1,5186 @@
+type Query {
```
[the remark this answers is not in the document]
mattkrick: Yep! Every new piece of business logic will have its own `.graphql` typeDef. We'll use a tool that merges all those typeDefs together into a generated schema that we won't have to look at or touch.

I finished the refactor & left all the legacy stuff in a big `_legacy.graphql` for the sake of time.
https://github.com/ParabolInc/parabol/pull/6228/files#diff-77adac95bcf72265d1a92f5b3932ccf2b8a6055e41e1baf2337fd7c5ee8fc830

packages/server/graphql/intranetSchema/sdl/mutations/updateEmail.ts
```diff
@@ -0,0 +1,48 @@
+import getRethink from '../../../../database/rethinkDriver'
```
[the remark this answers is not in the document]
mattkrick: Agreed!
What do you think of this hierarchy?

inside `/graphql`:
- public
  - public/queries
  - public/mutations
  - public/subscriptions
  - public/types
- private
  - private/queries
  - private/mutations
  - private/types
- github
  - queries
  - mutations
- gitlab
  - queries
  - mutations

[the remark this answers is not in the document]
mattkrick: See https://github.com/ParabolInc/parabol/pull/6228/files#diff-0281cf5ab4a0b4d7573df4f26731ffcd6bf36bdbc723eb82d1f0d83dd667e4d5 for a test implementation

packages/server/types/webpackEnv.ts
```diff
@@ -0,0 +1,386 @@
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-nocheck
+// Type definitions for webpack (module API) 1.16
```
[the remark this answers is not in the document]
mattkrick: i couldn't figure out how to import this without using the `/// references` pattern, which is deprecated. It also had some incorrect types, so i figure I'd just copy it over & remove the dep

packages/server/graphql/intranetSchema/sdl/types/PingableServices.ts
mattkrick: if we return undefined then the graphql traversal won't continue, it only continues on objects.

To test: remove `return {}` from ping.ts.

codegen.json
mattkrick: i like the idea of forcing an explicit mapper!
i'll play around with the subsequent PR to test `void` and `never` and see which one works best.

mattkrick:
looks like master fails the build now, but not because of this.
gonna merge to master anyways & we can triage that later...

all issues that weren't fixed here (folder hierarchy, explicit Source types) will be fixed or trialed in the following PR: https://github.com/ParabolInc/parabol/pull/6228

Issue #5931, opened 2022-01-24
Handle a start standup mutation


BartoszJarocki:

mattkrick:
close in favor of #5930

PR #6171, opened 2022-03-04
feat(standups): Added start team prompt mutation

### LINKED ISSUE #5930: Implement a start standup mutation
In similar fashion to `startCheckIn` or `startSprintPoker` mutations, we need `startStandup` mutation.

AC:
- [ ] evaluate if we can have a single, generic `startTeamPrompt` mutation or do we need `startStandup`, discuss with the maintainer
- [ ] implement the solution

BartoszJarocki:
Fixes #5930 and #5931.

Test:
1. open new meeting screen
2. start async standup
3. empty view with a Parabol logo is displayed
4. close empty view by clicking the logo in the center of the screen
5. async standup meeting is displayed in the meetings view

CleanShot 2022-03-18 at 15 46 36


BartoszJarocki:
@tianrunhe do you mind taking a look? thanks!

packages/server/graphql/mutations/helpers/createNewMeetingPhases.ts
```diff
@@ -132,6 +133,20 @@ const createNewMeetingPhases = async (
         case LAST_CALL:
         case 'SCOPE':
           return new GenericMeetingPhase(phaseType, durations)
+        case 'RESPONSES':
+          const teamMembers = await dataLoader.get('teamMembersByTeamId').load(teamId)
+          const teamMemberIds = teamMembers.map(({id}) => id)
```
BartoszJarocki, line 138: If we add a new team member, the meeting will be missing a stage for him. What would be the best place/way to handle that scenario? Same for team member removal.

BartoszJarocki: After some investigation I think it'd be good to add this login into: `JoinMeetingMutation` 🤔

packages/server/database/types/MeetingTeamPrompt.ts
```diff
@@ -0,0 +1,30 @@
+import TeamPromptResponsesPhase from './TeamPromptResponsesPhase'
+import Meeting from './Meeting'
+import GenericMeetingPhase from './GenericMeetingPhase'
+
+type TeamPromptPhase = TeamPromptResponsesPhase | GenericMeetingPhase
```
tianrunhe, line 5: +1 Do we need `GenericMeetingPhase` here? Does the team prompt meeting has any of the generic meeting phases?

packages/server/database/types/MeetingTeamPrompt.ts
```diff
@@ -0,0 +1,30 @@
...
+  constructor(input: Input) {
+    const {id, teamId, meetingCount, name, phases, facilitatorUserId} = input
+    super({
+      id,
+      teamId,
+      meetingCount,
+      phases,
+      facilitatorUserId,
+      meetingType: 'teamPrompt',
+      name: name ?? `Async standup #${meetingCount + 1}`
```
tianrunhe, line 27: +1 Let's stick with **S**tandup to be consistent

packages/server/graphql/mutations/helpers/createNewMeetingPhases.ts
```diff
@@ -132,6 +133,20 @@ const createNewMeetingPhases = async (
...
+          const teamPromptResponsesPhase = new TeamPromptResponsesPhase(teamMemberIds)
+          const {stages: teamPromptStages} = teamPromptResponsesPhase
+          const teamMemberResponseDiscussion = teamPromptStages.map((stage) => ({
+            id: stage.discussionId,
+            teamId,
+            meetingId,
+            discussionTopicId: stage.teamMemberId,
```
tianrunhe, line 145: -1 This will cause different team prompt meetings have the same `discussionTopicId`, yes?

packages/server/graphql/types/TeamPromptResponse.ts
```diff
@@ -0,0 +1,55 @@
+import {GraphQLFloat, GraphQLID, GraphQLNonNull, GraphQLObjectType, GraphQLString} from 'graphql'
+import {GQLContext} from '../graphql'
+import Team from './Team'
+
+const TeamPromptResponse: GraphQLObjectType = new GraphQLObjectType<any, GQLContext>({
+  name: 'TeamPromptResponse',
+  description: 'A response of a single team member in a team prompt',
+  isTypeOf: ({status}) => !!status,
```
tianrunhe, line 8: Where does this `status` come from?

packages/server/graphql/types/TeamPromptResponse.ts
```diff
@@ -0,0 +1,55 @@
...
+  fields: () => ({
+    createdBy: {
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'The userId that created the item'
+    },
+    createdByUser: {
+      type: new GraphQLNonNull(require('./User').default),
+      description: 'The user that created the item',
+      resolve: ({createdBy}, _args: unknown, {dataLoader}: GQLContext) => {
```
tianrunhe, line 17: -2 In DB, the field is named `userId`

packages/server/graphql/types/TeamPromptResponse.ts
```diff
@@ -0,0 +1,55 @@
...
+        return dataLoader.get('users').load(createdBy)
+      }
+    },
+    content: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'the content of the response'
+    },
+    plaintextContent: {
```
tianrunhe, line 25: -2 `plain**T**extContent` is the column name in DB

tianrunhe (changes requested):

packages/server/graphql/types/TeamPromptResponse.ts
BartoszJarocki: hmm, they actually match! see the discussion here: https://github.com/ParabolInc/parabol/pull/6068/files#r809458469

packages/server/graphql/types/TeamPromptResponse.ts
BartoszJarocki: that wasn't really needed, it's removed now

packages/server/graphql/types/TeamPromptResponse.ts
BartoszJarocki: this whole thing wasn't really needed, we already have `user` and `userId` which is enough.

packages/server/database/types/MeetingTeamPrompt.ts
BartoszJarocki: We don't but it's really a side effect of `createNewMeetingPhases` method returning `GenericMeetingPhase[]` not matter what. I don't want to touch this area of code too much as it affects all other meetings so I've added it to silence the TS. Other meetings like retro, do not specify the phase types at all here: https://github.com/ParabolInc/parabol/blob/7c60878ee1740fd9b60a1d62c10b584d04802d11/packages/server/database/types/MeetingRetrospective.ts#L9 so I'm not sure what's the lesser evil ;)

packages/server/graphql/mutations/helpers/createNewMeetingPhases.ts
BartoszJarocki: this may sound dumb but does `discussionTopicId` have to be globally unique? my understanding is it has to be unique within the meeting but I'm open to changing that if I overlooked something!

packages/server/graphql/types/TeamPromptResponse.ts
```diff
@@ -0,0 +1,45 @@
...
+import {GQLContext} from '../graphql'
+import Reactable, {reactableFields} from './Reactable'
+import Team from './Team'
+
+const TeamPromptResponse: GraphQLObjectType = new GraphQLObjectType<any, GQLContext>({
+  name: 'TeamPromptResponse',
+  description: 'A response of a single team member in a team prompt',
+  interfaces: () => [Reactable],
+  fields: () => ({
+    ...reactableFields(),
```
BartoszJarocki, line 11: I've added a reactable fields here but I guess it'd be best to create a task for the implementation of that. DB has the list of reactjis for each response but I'm not sure if the data format matches 100%.

packages/server/graphql/types/GenericMeetingStage.ts
```diff
@@ -2,7 +2,15 @@ import {GraphQLObjectType} from 'graphql'
...
 
-const phaseTypes = ['reflect', 'group', 'vote', 'firstcall', 'lastcall', 'SCOPE'] as const
+const phaseTypes = [
+  'reflect',
+  'group',
+  'vote',
+  'firstcall',
+  'lastcall',
+  'SCOPE',
+  'RESPONSES'
```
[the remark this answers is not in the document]
BartoszJarocki: yeah, I don't really like the `responses` naming either. I also don't like prefixing the stage name with the meeting name, so I'm not sure what to do here yet 🤔

packages/server/database/types/MeetingTeamPrompt.ts
BartoszJarocki: fixed!

packages/server/graphql/mutations/startTeamPrompt.ts
```diff
@@ -0,0 +1,79 @@
...
+import createNewMeetingPhases from './helpers/createNewMeetingPhases'
+import generateUID from '../../generateUID'
+import MeetingTeamPrompt from '../../database/types/MeetingTeamPrompt'
+import {startMattermostMeeting} from './helpers/notifications/notifyMattermost'
+import {startSlackMeeting} from './helpers/notifications/notifySlack'
+import sendMeetingStartToSegment from './helpers/sendMeetingStartToSegment'
+
+const startTeamPrompt = {
+  type: GraphQLNonNull(StartTeamPromptPayload),
+  description: ``,
```
tianrunhe: \-1 description was missing

packages/server/graphql/mutations/startTeamPrompt.ts
```diff
@@ -0,0 +1,79 @@
...
+  args: {
+    teamId: {
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'The team starting the meeting'
```
tianrunhe, line 24: -1 "Id of the team..."

packages/server/graphql/types/TeamPromptResponse.ts
tianrunhe: Weird... On my local machine, it was capitalized T. But I checked prod it's lowercase so we are good.

packages/server/graphql/types/TeamPromptResponse.ts
```diff
@@ -0,0 +1,45 @@
...
+    },
+    content: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'the content of the response'
+    },
+    plaintextContent: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'the plain text content of the response'
+    },
+    sortOrder: {
```
tianrunhe, line 31: -1 we don't have that field in DB so it won't resolve now.

tianrunhe (changes requested):

packages/server/graphql/mutations/startTeamPrompt.ts
BartoszJarocki: it's there: https://github.com/ParabolInc/parabol/pull/6171/files#diff-49ad6d7c7278b17294131a1bceddfb74714031fee6e4433287256dd997708d50R24

packages/server/graphql/types/TeamPromptResponse.ts
BartoszJarocki: seems like an artifact of the old migration

packages/server/graphql/types/TeamPromptResponse.ts
BartoszJarocki: we do, I've added that per your suggestion: https://github.com/ParabolInc/parabol/pull/6068/files#diff-17d65a8dd63bfe43cfadd39f2b9814637352fff193e0cbd8ce3ee5c0d2bd0c53R24 😁

packages/server/graphql/mutations/startTeamPrompt.ts
BartoszJarocki: fixed!

tianrunhe (approved):

packages/server/graphql/types/TeamPromptResponse.ts
mattkrick: +1 safe to keep a float for now, but in the future we'll probably want to create a new type which is equivalent to `Number.MAX_SAFE_INTEGER` so we don't have to validate that it's an integer

mattkrick (changes requested):
great work! only big thing is preventing 2 "startMeeting" mutations that happen at the same time.

packages/client/mutations/StartTeamPromptMutation.ts
```diff
@@ -0,0 +1,58 @@
...
+import {commitMutation} from 'react-relay'
+import {HistoryLocalHandler, StandardMutation} from '../types/relayMutations'
+import {StartTeamPromptMutation as TStartTeamPromptMutation} from '../__generated__/StartTeamPromptMutation.graphql'
+
+graphql`
+  fragment StartTeamPromptMutation_team on StartTeamPromptSuccess {
+    meeting {
+      id
+      name
+      meetingMembers {
```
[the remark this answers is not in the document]
BartoszJarocki: You're very much correct here, we only need 
```
    meeting {
      id
    }
```
as it is used in mutation's `onCompleted` handler to navigate to the meeting route. I removed them all, as a result we have a noticeable performance improvement, mutation completes faster! 🎉

packages/server/database/types/MeetingTeamPrompt.ts
BartoszJarocki: sure. if you remove `GenericMeetingPhase` from this line, in `startTeamPrompt` mutation TS will complain because it can't automatically cast `GenericMeetingPhase[]` to `TeamPromptResponsesPhase[]` when creating a new instance of `MeetingTeamPrompt`. 
Screenshot 2022-03-24 at 15 29 23

packages/server/database/types/TeamPromptMeetingMember.ts
```diff
@@ -0,0 +1,15 @@
...
+interface Input {
+  id?: string
+  updatedAt?: Date
+  teamId: string
+  userId: string
+  meetingId: string
+}
+
+export default class TeamPromptMeetingMember extends MeetingMember {
+  constructor(input: Input) {
```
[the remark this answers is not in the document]
BartoszJarocki: good hint, done

packages/server/graphql/mutations/helpers/handleCompletedStage.ts
```diff
@@ -85,7 +86,7 @@ const handleCompletedRetrospectiveStage = async (
 
 const handleCompletedStage = async (
   stage: GenericMeetingStage,
-  meeting: MeetingRetrospective | MeetingPoker | MeetingAction,
+  meeting: MeetingRetrospective | MeetingPoker | MeetingAction | MeetingTeamPrompt,
```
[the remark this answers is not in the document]
Dschoordsch: Why not just use the base type `Meeting`?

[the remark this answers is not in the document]
BartoszJarocki: I'm not sure how this whole schema thingy works in this case so I played safe 😎 happy to improve if there's a better & simpler way though! @mattkrick any thoughts?

packages/server/graphql/types/StartTeamPromptPayload.ts
```diff
@@ -0,0 +1,27 @@
...
+    meeting: {
+      type: new GraphQLNonNull(TeamPromptMeeting),
+      resolve: resolveNewMeeting
+    },
+    meetingId: {
+      type: new GraphQLNonNull(GraphQLID)
+    },
+    team: {
+      type: new GraphQLNonNull(Team),
+      resolve: resolveTeam
```
BartoszJarocki, line 20: @Dschoordsch I applied your review notes from https://github.com/ParabolInc/parabol/pull/6250 here as well, just fyi

BartoszJarocki:
@mattkrick do you mind taking a look one more time? I _think_ I covered it all except for https://github.com/ParabolInc/parabol/pull/6171#discussion_r833829673. I'd rather create a separate task as this PR is growing and growing, thanks!

packages/server/graphql/mutations/helpers/handleCompletedStage.ts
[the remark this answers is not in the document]
mattkrick: @Dschoordsch discriminating on the super (base `Meeting`) won't yield a more specific Meeting type.

```ts
class Meeting {
  type = ''
}

class RetroMeeting extends Meeting {
  type = 'retro' as const
  fun = true
}

class PokerMeeting extends Meeting {
  type = 'poker' as const
  addictive = true
}

type AnyMeeting = RetroMeeting | PokerMeeting

const foo = null as unknown as AnyMeeting

if (foo.type === 'retro') {
  foo.fun // ts knows this is a RetroMeeting
}
```

mattkrick:
Safe to merge, would love to hear your thoughts on https://github.com/ParabolInc/parabol/pull/6171#discussion_r833832059 before this goes to GA as it'll be way cheaper to make the change now

packages/server/graphql/types/DiscussionTopicTypeEnum.ts
```diff
@@ -7,6 +7,7 @@ const DiscussionTopicTypeEnum = new GraphQLEnumType({
     agendaItem: {},
     reflectionGroup: {},
     task: {},
+    teamPromptResponse: {},
```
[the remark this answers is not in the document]
BartoszJarocki: I personally think it we should have one thread per user (ie. and one response per user) but I might be missing a bigger picture here. So currently the `discussionTopicId` is set to `teamMemberId` and I think it's ok. Are there any pros of using just the `userId` here? 

@ackernaut do you see any potential use cases where we'd need more than one response from a single user in a team prompt meeting? would we also need a discussion thread per every user response? I'd appreciate your thoughts here! 🙏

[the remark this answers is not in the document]
ackernaut: @BartoszJarocki I think in the future, if you could build your own meeting or add your own meeting phase, maybe that would be an option (e.g. number of responses? n (1, 2, 3…) or unlimited?

That being said I don’t think we have a good enough conception of what that is in ‘structured conversations’. Some use cases come to mind. For a proposal, maybe the prompt is ‘clarifying questions’ or ‘reactions’ and each person can ask more than 1 and they get a little stack of responses that represent their questions or reactions. Maybe we could group similar questions and reactions? However, maybe using cards isn’t the way to handle that UX. 

If I had to make a call, I would just keep it simple and go with 1 response per user at this phase. If you think we will be stuck later, is there a way where you can allow for multiple, but we limit it superficially in the UI for the standup UX? We can discuss more in a proper thread or sync if you need.

PR #6181, opened 2022-03-08
feat(zoom): Add Zoom integration provider

Dschoordsch:
Add createVideoMeeting mutation which can create a new zoom meeting and
returns the link. Zoom is added as an intagration provider and primed
during postdeploy. The OAuth2 flow currently is triggered only from a
provider row.

TODO:
- [ ] finish integration provider row (remove menu, icon)
- [ ] prompt to integrate when there's no active zoom integration before calling create video meeting mutation

Resolves:
* #6187


mattkrick:
PR housekeeping: should we keep this open?

PR #6182, opened 2022-03-08
ci: use official postgres image & sync with digital ocean version

blimmer:
I noticed today that CircleCI started badging our builds with this warning about using deprecated images ([link from screenshot](https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034)).

Screen Shot 2022-03-08 at 16 35 00

We just needed to switch over to using the official postgres image vs. the `circle/` prefixed one.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ jobs:
     docker:
       - image: cimg/node:16.14.0
       - image: rethinkdb:2.4.0
-      - image: circleci/postgres:12.3
+      - image: postgres:12.3
         environment:
           # These env values need to be synced up with the `test` environment file
           POSTGRES_PASSWORD: "p9_p455w02d_f02_73575"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ jobs:
     docker:
       - image: cimg/node:16.14.0
       - image: rethinkdb:2.4.0
-      - image: postgres:12.3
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -4,7 +4,7 @@ jobs:
     docker:
       - image: cimg/node:16.14.0
       - image: rethinkdb:2.4.0
-      - image: circleci/postgres:12.3
+      - image: postgres:12.3
```
Dschoordsch, line 7: +1 We should probably just switch to [`cimg/postgres:12.9`](https://circleci.com/developer/images/image/cimg/postgres) instead

Dschoordsch (approved):

.circleci/config.yml
adaniels-parabol: We should use cimg/postgres, we recently moved the node image to cimg/node as well.

adaniels-parabol: Just FYI, we are actually at 12.10 on DO

blimmer: Gotcha, so would it make sense to sync all of these version specifications up to 12.10?

blimmer: Sounds good. I'll update the local docker setups to use `postgres/12.10` and this to use `cimg/postgres:12.9` until they merge this change: https://github.com/CircleCI-Public/cimg-postgres/pull/42

Issue #6180, opened 2022-03-07
Replace Deprecated Scopes in Jira Integration


lora-reames:
https://developer.atlassian.com/cloud/jira/platform/changelog/#CHANGE-517
> ## Deprecation Notice Scopes for Forge and OAuth 2.0 (3LO) apps
> The original OAuth scopes for Forge and OAuth 2.0 (3LO) apps on Confluence Cloud, Jira Cloud platform, and Jira Service Management Cloud are deprecated.
> 
> All Forge and OAuth 2.0 (3LO) **apps must use the new scopes by 23 August 2022 PDT / 24 August 2022 AEST**. After this date, only apps using the new scopes will function as expected, as the original scopes will have been removed.

Parabol currently uses the following [deprecated scopes](https://developer.atlassian.com/cloud/jira/platform/scopes-for-oauth-2-3LO-and-forge-apps/#deprecated-scopes):
- 'read:jira-user'
- 'read:jira-work'
- 'write:jira-work'
- 'manage:jira-project'

lora-reames:
found this while working on https://github.com/ParabolInc/parabol/issues/6153

Dschoordsch:
@igorlesnenko you mentioned you've read the deadline moved? Could you update the description?

igorlesnenko:
Yeah, they paused this change
https://developer.atlassian.com/cloud/jira/platform/changelog/#CHANGE-603

> On 22 February, we announced the [availability of new scopes for Forge and OAuth 2.0 (3LO) apps](https://community.developer.atlassian.com/t/action-required-update-scopes-for-forge-and-oauth-2-0-3lo-apps/53299). Following concerns raised by our developer community, we have now paused the deprecation of classic scopes and removed the requirement to transition your app to new scopes by 23 August 2022.

> There is no action for you to take regarding your app scopes at this stage. Apps using either classic or new scopes can continue to do so. For those that haven’t adopted new scopes yet, please continue using classic scopes – we’ll let you know when you can start transitioning.

jordanh:
Good to know! It is still coming up, but we can safely kick the can down the road just a little while longer...

PR #6184, opened 2022-03-09
feat(jira-integration): register webhooks for jira updates

### LINKED ISSUE #6153: Jira webhook Proposal
## Issue - Enhancement
resolve https://github.com/ParabolInc/parabol/issues/6031 by registering a webhook for issue updates and updating linked issues in parabol
> A [user](https://app.hubspot.com/contacts/3888472/contact/14760601) 🔒 writes:
> We're using the Sprint Poker Meeting tool and link our issues via Jira. One thing we miss is: if a story's description is being updated in Jira, there is no option to reload/update the linked issue in the ongoing Sprint Poker Meeting. Would love to see that option available in the future.

### Acceptance Criteria

- [ ]     When a user has parabol open & they're looking at a Task that is integrated with Jira, when they update that issue on the jira website, the task in parabol gets updated for them & their team without them having to refresh
   
### Proposed Solution
- #### update jira scopes
    add scope `read:issue-details:jira` required for issue webhooks
- #### add fields to atlassianAuth Table (psql) for projects tracked and webhook expiration
- #### add side effect to [addAtlassianAuth.ts](https://github.com/ParabolInc/parabol/blob/8409e5ee0ab86c9effcb616dd82cdf30053c81ee/packages/server/graphql/mutations/addAtlassianAuth.ts) register webhook with currently known jira projects
- #### add side effect to [removeAtlassianAuth.ts](https://github.com/ParabolInc/parabol/blob/33c3478b8699f0a61e092b533c36e4e6ac03c704/packages/server/graphql/mutations/removeAtlassianAuth.ts#L11) to deregister webhook when auth is removed.
- #### schedule task (daily) to query the atlassianAuth table for hooks that are expiring soon and refresh them
- #### add side effects to AtlassianManager.ts [getAllProjects](https://github.com/ParabolInc/parabol/blob/7c60878ee1740fd9b60a1d62c10b584d04802d11/packages/client/utils/AtlassianManager.ts#L480) to add previously unknown projects to the webhook and update table, remove projects that dont exist anymore. if list differs from table update webhook
   - investigate if this is actually necessary, it may be possible to create a webhook which targets an entire cloud id without filtering to the projects 
- #### webhook handler
    1. implement a jiraWebhookHandler to handle update events
    3. discard events for issues which do not match an existing parabol task
    4. events
        1. issue-updated → update a task in parabol via [updateTask](https://github.com/ParabolInc/parabol/blob/7ea866e83a250d097462c0074b9ee0f93155be12/packages/server/graphql/mutations/updateTask.ts#L49) mutation which *should* update it for viewers via subscription and keep task history updated

**Estimated effort:** 8hrs

lora-reames:
**DRAFT**
closes #6153
also trying for #6180

### DIFF
--- a/packages/client/utils/AtlassianManager.ts
+++ b/packages/client/utils/AtlassianManager.ts
@@ -1,6 +1,7 @@
 import AbortController from 'abort-controller'
 import JiraIssueId from '../shared/gqlIds/JiraIssueId'
 import {SprintPokerDefaults} from '../types/constEnums'
+import sleep from './sleep'
 
 export interface JiraUser {
   self: string
@@ -264,12 +265,12 @@ interface JiraPageBean<T> {
 
 interface WebhookDetails {
   jqlFilter: string
-  fieldIdsFilter: string[]
+  fieldIdsFilter?: string[]
   events: jiraWebhookEventType[]
 }
 
 interface Webhook {
-  id: string
+  id: number
   jqlFilter: string
   fieldIdsFilter: string[]
   issuePropertyKeysFilter: string[]
@@ -919,7 +920,7 @@ export default abstract class AtlassianManager {
     )
   }
 
-  async registerWebhook(cloudId: string, url: string, webhooks: WebhookDetails[]) {
+  async registerWebhooks(cloudId: string, url: string, webhooks: WebhookDetails[]) {
     return this.post(`https://api.atlassian.com/ex/jira/${cloudId}/rest/api/3/webhook`, {
       url,
       webhooks
@@ -941,6 +942,58 @@ export default abstract class AtlassianManager {
   async getWebhooks(cloudId: string) {
     return this.get<JiraPageBean<Webhook>>(
       `https://api.atlassian.com/ex/jira/${cloudId}/rest/api/3/webhook`
+    )
+  }
+  async getAllWebhooks(cloudIds: string[]) {
+    return Promise.all(cloudIds.map((cloudId) => this.getWebhooks(cloudId)))
+  }
+
+  async updateWebhook(webhookId: number, cloudId: string, url: string) {
+    // register new hook
+    await this.registerAllProjectsWebhooks([cloudId], url)
+    // delete old hook
+    await this.deleteWebhooks(cloudId, [webhookId])
+  }
+
+  async registerOrUpdateWebhooks(cloudIds: string[], url: string) {
+    const webhooksAndErrors = await this.getAllWebhooks(cloudIds)
+    return Promise.all(
+      webhooksAndErrors.map(async (entry, index) => {
+        if (entry instanceof Error) {
+          return Promise.reject(Error)
+        } else if (entry instanceof RateLimitError) {
+          // retry
+          const waitTime = Date.now() - entry.retryAt.getUTCMilliseconds()
+          if (waitTime > 0) await sleep(waitTime)
+          return this.registerOrUpdateWebhooks([cloudIds[index]!], url)
+        } else if (entry.total > 0) {
+          return this.updateWebhook(entry.values[0]!.id, cloudIds[index]!, url)
+        } else {
+          return this.registerAllProjectsWebhooks([cloudIds[index]!], url)
+        }
+      })
+    )
+  }
+
+  async registerAllProjectsWebhooks(cloudIds: string[], url: string) {
+    const allProjects = await this.getAllProjects(cloudIds)
+
+    const projectsByCloudId = allProjects.reduce((cloudIds, project) => {
+      const index = cloudIds.findIndex(({cloudId}) => project.cloudId === cloudId)
+      cloudIds[index]!.projects.push(project)
+      return cloudIds
+    }, [] as Array<{cloudId: string; projects: JiraProject[]}>)
+
+    return Promise.all(
+      projectsByCloudId.map((cloudIdProjects) => {
+        const {cloudId, projects} = cloudIdProjects
+        const jqlFilter = `project in ${projects.map(({key}) => key).join(',')}`
+        const webhook: WebhookDetails = {
+          jqlFilter,
+          events: ['jira:issue_updated']
+        }
+        this.registerWebhooks(cloudId, url, [webhook])
+      })
     )
   }
 }
--- a/packages/client/utils/AtlassianManager.ts
+++ b/packages/client/utils/AtlassianManager.ts
@@ -277,34 +277,23 @@ export type JiraPermissionScope =
   | 'delete:webhook:jira'
   | 'offline_access'
   | 'read:application-role:jira'
-  | 'read:application-role:jira'
-  | 'read:avatar:jira'
   | 'read:avatar:jira'
   | 'read:field-configuration:jira'
-  | 'read:group:jira'
   | 'read:group:jira'
   | 'read:issue-details:jira'
   | 'read:issue-meta:jira'
   | 'read:issue-security-level:jira'
   | 'read:issue-type-hierarchy:jira'
-  | 'read:issue-type-hierarchy:jira'
-  | 'read:issue-type:jira'
   | 'read:issue-type:jira'
   | 'read:issue:jira'
   | 'read:issue.changelog:jira'
   | 'read:issue.vote:jira'
   | 'read:project-category:jira'
-  | 'read:project-category:jira'
-  | 'read:project-version:jira'
   | 'read:project-version:jira'
   | 'read:project:jira'
-  | 'read:project:jira'
   | 'read:project.component:jira'
-  | 'read:project.component:jira'
-  | 'read:project.property:jira'
   | 'read:project.property:jira'
   | 'read:status:jira'
-  | 'read:user:jira'
   | 'read:user:jira'
   | 'read:webhook:jira'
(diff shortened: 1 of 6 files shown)


packages/client/utils/AtlassianManager.ts
```diff
@@ -831,4 +923,87 @@ export default abstract class AtlassianManager {
...
+        }
+      })
+    )
+  }
+
+  async registerAllProjectsWebhooks(cloudIds: string[]) {
+    if (cloudIds.length === 0) return
+    const allProjects = await this.getAllProjects(cloudIds)
+
+    const initial = cloudIds.map<{cloudId: string; projects: JiraProject[]}>((id) => ({
```
Dschoordsch, line 987: +1 specifying the type is redundant here

packages/client/utils/AtlassianManager.ts
```diff
@@ -831,4 +923,87 @@ export default abstract class AtlassianManager {
...
+      cloudId: id,
+      projects: [] as JiraProject[]
+    }))
+
+    const projectsByCloudId = allProjects.reduce((projectsByCloud, project) => {
```
Dschoordsch, line 992: +1 while technically correct to use `reduce` here, since you're just modifying and returning `initial` it would be easier to read if you would rename `initial` to `projectsByCloudId` and then do `allProjects.forEach` instead.

packages/client/utils/AtlassianManager.ts
```diff
@@ -831,4 +923,87 @@ export default abstract class AtlassianManager {
...
+      const index = projectsByCloud.findIndex(({cloudId}) => project.cloudId === cloudId)
+      projectsByCloud && projectsByCloud[index]!.projects.push(project)
+      return projectsByCloud
+    }, initial)
+    return Promise.all(
+      projectsByCloudId.map((cloudIdProjects) => {
+        const {cloudId, projects} = cloudIdProjects
+        const jqlFilter = `project IN (${projects.map(({key}) => key).join(',')})`
```
Dschoordsch, line 1000: +1 since we only need the project key, we could already just collect those in `projectsByCloudId`

packages/client/utils/AtlassianManager.ts
```diff
@@ -398,10 +489,11 @@ export default abstract class AtlassianManager {
     }
     return new Error(`Unknown Jira error: ${JSON.stringify(error)}`)
   }
-  private readonly delete = async (url) => {
+  private readonly delete = async (url, payload?: any) => {
```
Dschoordsch, line 492: +1 would be nice to add a type to `url: string,` as well

packages/client/utils/AtlassianManager.ts
```diff
@@ -831,4 +923,87 @@ export default abstract class AtlassianManager {
       `https://api.atlassian.com/ex/jira/${cloudId}/rest/api/3/screens/${screenId}/tabs/${tabId}/fields/${fieldId}`
     )
   }
+
+  async registerWebhooks(cloudId: string, webhooks: WebhookDetails[]) {
+    return this.post(`https://api.atlassian.com/ex/jira/${cloudId}/rest/api/3/webhook`, {
+      url: makeAppURL('http://localhost:3000', '/webhooks/jira'),
+      // url: makeAppURL(appOrigin, '/webhooks/jira'),
```
Dschoordsch, line 930: +1 that line should be correct, maybe you're missing some `.env` variables?
```
HOST='localhost'
PORT='3000'
PROTO='http'
```

packages/server/integrations/jiraWebhookHandler.ts
```diff
@@ -0,0 +1,73 @@
...
+  | 'jira:issue_created'
+  | 'jira:issue_updated'
+  | 'jira:issue_deleted'
+  | 'comment:created'
+  | 'comment_updated'
+  | 'comment_deleted'
+  | 'issue_property_set'
+  | 'issue_property_deleted'
+
+const jiraWebhookHandler = uWSAsyncHandler(async (res: HttpResponse, _req: HttpRequest) => {
```
Dschoordsch, line 64: -1 Is there some verification how we can check the request is really coming from Atlassian?

Dschoordsch (changes requested):

packages/client/utils/AtlassianManager.ts
```diff
@@ -831,4 +923,87 @@ export default abstract class AtlassianManager {
...
+    })
+  }
+
+  async getWebhooks(cloudId: string) {
+    return this.get<JiraPageBean<Webhook>>(
+      `https://api.atlassian.com/ex/jira/${cloudId}/rest/api/3/webhook`
+    )
+  }
+  async getAllWebhooks(cloudIds: string[]) {
+    return Promise.all(cloudIds.map((cloudId) => this.getWebhooks(cloudId)))
```
mattkrick, line 953: +1 if the return value is extended with `cloudId` then the parent function wouldn't need to dangerously index into the `cloudIds` array

packages/client/utils/AtlassianManager.ts
```diff
@@ -831,4 +923,87 @@ export default abstract class AtlassianManager {
...
+    await Promise.all(
+      webhooksAndErrors.map(async (entry, index) => {
+        if (entry instanceof Error) {
+          return Promise.reject(Error)
+        } else if (entry instanceof RateLimitError) {
+          // retry
+          const waitTime = Date.now() - entry.retryAt.getUTCMilliseconds()
+          if (waitTime > 0) await sleep(waitTime)
+          return this.registerOrUpdateWebhooks([cloudIds[index]!])
+        } else if (entry.total > 0) {
```
mattkrick, line 974: +1 checking for `entry.values[0]` here means no non-null coercion is needed later

packages/server/integrations/jiraWebhookHandler.ts
```diff
@@ -0,0 +1,73 @@
...
+interface jiraIssueChangeLogItem {
+  toString: string
+  to: string | null
+  fromString: string
+  from: string | null
+  fieldtype: string // more specific? i think this is jira vs custom
+  field: string // name of field altered
+}
+
+type jiraWebhookEventType =
```
mattkrick, line 54: +1 types should all be PascalCase

packages/client/utils/AtlassianManager.ts
```diff
@@ -831,4 +923,87 @@ export default abstract class AtlassianManager {
...
+        } else {
+          return this.registerAllProjectsWebhooks([cloudIds[index]!])
+        }
+      })
+    )
+  }
+
+  async registerAllProjectsWebhooks(cloudIds: string[]) {
+    if (cloudIds.length === 0) return
+    const allProjects = await this.getAllProjects(cloudIds)
```
mattkrick, line 985: -1 why not just watch all projects? by filtering on project, that means the webhook will go stale when they add a new project to a cloud

packages/client/utils/AtlassianManager.ts
```diff
@@ -831,4 +923,87 @@ export default abstract class AtlassianManager {
...
+    await this.deleteWebhooks(cloudId, [webhookId])
+  }
+
+  async registerOrUpdateWebhooks(cloudIds: string[]) {
+    const webhooksAndErrors = await this.getAllWebhooks(cloudIds)
+    await Promise.all(
+      webhooksAndErrors.map(async (entry, index) => {
+        if (entry instanceof Error) {
+          return Promise.reject(Error)
+        } else if (entry instanceof RateLimitError) {
```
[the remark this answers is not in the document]
lora-reames: I actually wanted to ask you about that but didn't get around to it before the BP time was spent. this section was something of an un-finished thought. rate limit retry doesnt seem to be handled centrally now but i was weary of changing the behavior else where without discussing it. i should have reached out about that when i thought about it, my bad.

packages/client/utils/AtlassianManager.ts
lora-reames: agreed

packages/client/utils/AtlassianManager.ts
lora-reames: good point

packages/client/utils/AtlassianManager.ts
lora-reames: probably i set it aside to investigate other details

packages/server/integrations/jiraWebhookHandler.ts
lora-reames: unfinished handler (time spent on jira scope issue) auth very much needed

packages/client/utils/AtlassianManager.ts
lora-reames: Per my understanding (and i thought yours) its not possible to subscribe per cloud id, we actually discussed this [here](https://github.com/ParabolInc/parabol/issues/6153#issuecomment-1057517811) when we discussed the jql filter being required. Attempting to subscribe all was my first choice

lora-reames:
@mattkrick @Dschoordsch i wish id seen the conversation on this draft, I was surprised not to see some of this feedback mentioned in the BP channel.

packages/client/utils/AtlassianManager.ts
lora-reames: agreed

PR #6209, opened 2022-03-15
create env var and logic for FULLSCREEN auth dialog

### LINKED ISSUE #6204: Make authentication popup configurable in ENV vars
A recent customer using Keycloak as their authentication provider for SAML recently asked could we increase the size of the authentication popup as users have to scroll everytime they login via SSO, (see image below)

image
(customer information hidden)

A/C:

- [ ] ENV var called UI_AUTH_FULLSCREEN (true full screen the auth popup, false, normal size)

adaniels-parabol:
Resolves #6204 this was requested by non Parabol SaaS users as they had to scroll. On the public SaaS Parabol UI_AUTH_FULLSCREEN would be set to false

### DIFF
--- a/packages/client/types/modules.d.ts
+++ b/packages/client/types/modules.d.ts
@@ -28,6 +28,7 @@ interface Window {
     AUTH_INTERNAL_ENABLED: boolean
     AUTH_GOOGLE_ENABLED: boolean
     AUTH_SSO_ENABLED: boolean
+    UI_AUTH_FULLSCREEN: boolean
   }
 }
 declare type Json = null | boolean | number | string | Json[] | {[key: string]: Json}
--- a/packages/client/utils/getOAuthPopupFeatures.ts
+++ b/packages/client/utils/getOAuthPopupFeatures.ts
@@ -1,3 +1,4 @@
+
 interface Box {
   width: number
   height: number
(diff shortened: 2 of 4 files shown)


mattkrick:
@Dschoordsch had the great idea of having the Auth screen persist the window dimensions to local storage. that way, on subsequent visits they'll get the window to open at the right size. proposing we move in that direction 

AC
- [ ] the SAMLRedirect component persist its coords to local storage before it calls postMessage
- [ ] before opening a SAML auth popup, local storage is checked for coords & those are used, if available & validated

ragingpastry:
Hello @mattkrick  ! 

I am one of the requestors of this change. While I agree that persisting the window dimensions to local storage would be great, the main driver of this request was to make the initial login for users a little less cumbersome and funny looking. Currently the pop-up window is very tiny and looks quite strange. 

Would it be possible to have the pop-up window resize dynamically based on the size of the page?

Dschoordsch:
@ragingpastry It is not possible to resize the window according to the content because we don't know the content when we create the popup. I think it would be safe to increase the popup size for everyone in general since there are multiple use cases where it requires scrolling and I think that would be safe to try. Right now it looks like the popup is sized the same as the login form, but that also looks quite lost on a big screen.

Right now the popup is 385x550 and I think we could safely increase it to 600x800 on desktop. This size should also be used for the OAuth popups.

Dschoordsch:
Closing, we should handle it properly in #6488

Issue #6158, opened 2022-03-03
Jira Server: adding tasks to poker scope


igorlesnenko:
Extracted from:  #6096

AC:

- [ ] Allow adding tasks to poker scope 
- [ ] Show task description in poker meeting

mattkrick:
closed in #6214

PR #6214, opened 2022-03-16
feat(jira-server): add issue to poker scope

igorlesnenko:
#6158

How to test

- Add Jira Server integration
- Create a poker meeting
- Add a Jira Server issue to the scope
- See it correctly appears in the sidebar
- Click on the issue in the sidebar — see description displayed correctly


packages/client/components/PokerEstimateHeaderCardJira.tsx
```diff
@@ -1,125 +1,36 @@
...
+import React from 'react'
+import {useFragment} from 'react-relay'
+import {PokerEstimateHeaderCardJira_issue$key} from '../__generated__/PokerEstimateHeaderCardJira_issue.graphql'
+import PokerEstimateHeaderCardHtml from './PokerEstimateHeaderCardHtml'
 
 interface Props {
-  issue: PokerEstimateHeaderCardJira_issue
+  issueRef: PokerEstimateHeaderCardJira_issue$key
 }
 const PokerEstimateHeaderCardJira = (props: Props) => {
```
nickoferrall, line 10: -1 it looks like `PokerEstimateHeaderCardJira` and `PokerEstimateHeaderCardJiraServer` are identical other than the `linkTitle`. Perhaps we could use `PokerEstimateHeaderCardHtml` for both components and use the `__typename` of the issue to determine the `linkTitle`?

packages/client/shared/gqlIds/JiraServerIssueId.ts
```diff
@@ -1,8 +1,8 @@
 const JiraServerIssueId = {
-  join: (providerId: number, integrationRepoId: string, issueId: string) => `${providerId}:${integrationRepoId}:${issueId}`,
+  join: (providerId: number, repoId: string, issueId: string) => `${providerId}:${repoId}:${issueId}`,
   split: (id: string) => {
     const parts = id.split(':')
-    return {providerId: parts[0], integrationRepoId: parts[1], issueId: parts[2]}
+    return {providerId: parseInt(parts[0] ?? '', 10), repoId: parts[1] , issueId: parts[2]}
```
nickoferrall, line 5: +1 `repoId` and `issueId` could be undefined, so when `JiraServerIssueId.split()` is being called, we're adding a default string, e.g. `return api.addComment(comment, issueId ?? '')`.

In my opinion, as `GitHubIssueId` and `JiraIssueId` return strings and `providerId` is a string, we should add a default string for `repoId` and `issueId` in this function too.

packages/server/database/types/ImportedTask.ts
```diff
@@ -7,22 +6,23 @@ interface Input {
...
+    const {meetingId, teamId, integration, userId, integrationHash} = input
     super({
       content: convertToTaskContent(`Task imported from ${integration.service} #archived`),
       createdBy: userId,
       meetingId,
       sortOrder: dndNoise(),
       status: 'future',
       teamId,
-      integrationHash: IntegrationHash.join(integration),
+      integrationHash,
```
nickoferrall, line 25: +2 nice catch!

packages/server/graphql/types/JiraServerIssue.ts
```diff
@@ -14,7 +15,11 @@ const JiraServerIssue = new GraphQLObjectType<any, GQLContext>({
   isTypeOf: ({service}) => service === 'jiraServer',
   fields: () => ({
     id: {
-      type: new GraphQLNonNull(GraphQLID)
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'GUID providerId:integrationRepoId:issueId',
+      resolve: ({id, project, providerId}) => {
```
nickoferrall, line 20: +1 could add types here

packages/client/shared/gqlIds/JiraServerIssueId.ts
```diff
@@ -1,8 +1,8 @@
 const JiraServerIssueId = {
-  join: (providerId: number, integrationRepoId: string, issueId: string) => `${providerId}:${integrationRepoId}:${issueId}`,
```
nickoferrall, line 2: +1 why the name change out of interest? JiraServerIssue description refers to it as `integrationRepoId`: https://github.com/ParabolInc/parabol/blob/a0951776dcf99f90578746d2c30f3563b3e4e3d3/packages/server/graphql/types/JiraServerIssue.ts#L19

nickoferrall (changes requested):

packages/client/shared/gqlIds/JiraServerIssueId.ts
igorlesnenko: @nickoferrall I renamed it to avoid confusion as we have `integrationRepoId` which is `service:providerId:repositoryId`. Renamed to `repositoryId` and fixed the description

packages/client/shared/gqlIds/JiraServerIssueId.ts
igorlesnenko: @nickoferrall Changed to using `!`, assuming the input is valid. As was suggested here: https://github.com/ParabolInc/parabol/pull/6059#discussion_r828400763

packages/client/components/PokerEstimateHeaderCardJira.tsx
igorlesnenko: Now using a single `PokerEstimateHeaderCardHtml`

nickoferrall (approved):
LGTM!

packages/client/components/PokerEstimateHeaderCard.tsx
```diff
@@ -44,7 +54,17 @@ export default createFragmentContainer(PokerEstimateHeaderCard, {
         integration {
           ... on JiraIssue {
             __typename
-            ...PokerEstimateHeaderCardJira_issue
+            issueKey
+            summary
+            descriptionHTML
+            jiraUrl: url
+          }
+          ... on JiraServerIssue {
```
mattkrick, line 62: +1 I see this duplication happens twice in this PR. In the future we'll probably want to make an interface that handles both.

e.g.
```ts
JiraCloudIssue implements JiraIssue
JiraServerIssue implements JiraIssue
```

then, we could still pass `issue` into the component

packages/server/integrations/AbstractTaskIntegrationManager.ts
```diff
@@ -0,0 +1,23 @@
...
+import {TaskIntegration} from '../database/types/Task'
+
+export type CreateTaskResponse =
+  | {
+      integrationHash: string
+      integration: TaskIntegration
+    }
+  | Error
+
+export default abstract class AbstractTaskIntegrationManager {
```
mattkrick, line 13: -1 can delete? i think this pattern went away...

packages/server/database/types/ImportedTask.ts
```diff
@@ -7,22 +6,23 @@ interface Input {
...
+  integrationHash: string
   userId: string
 }
 
 export default class ImportedTask extends Task {
   integrationHash!: string
   integration!: TaskIntegration
   constructor(input: Input) {
-    const {meetingId, teamId, integration, userId} = input
+    const {meetingId, teamId, integration, userId, integrationHash} = input
```
mattkrick, line 17: +1 can you give me some more info on why we can't use the `IntegrationHash.join` here?

packages/client/components/PokerEstimateHeaderCardHtml.tsx
```diff
@@ -74,16 +71,19 @@ const StyledLabel = styled('span')({
...
-  issue: PokerEstimateHeaderCardJira_issue
+  summary: string
+  descriptionHTML: string
+  url: string
+  linkTitle: string
+  linkText: string
 }
-const PokerEstimateHeaderCardJira = (props: Props) => {
-  const {issue} = props
+const PokerEstimateHeaderCardHtml = (props: Props) => {
```
mattkrick, line 80: +1 the github card is just HTML, too, so the name could be a little more discriminatory.

perhaps if we rename `JiraIssue` to `JiraCloudIssue` and then make `JiraCloudIssue` and `JiraServerIssue` implement a `JiraIssue` interface?

mattkrick (approved):
1 question about the serviceTaskId vs. integrationHash & I see a new interface in our future, but we can do that in another PR

packages/client/components/PokerEstimateHeaderCard.tsx
igorlesnenko: Yeah, I thought about it. We can do it in some next PRs

packages/client/components/PokerEstimateHeaderCardHtml.tsx
igorlesnenko: @mattkrick Renamed `PokerEstimateHeaderCardHtml` to `PokerEstimateHeaderCardContent`

What if we make a single interface with unified fields for all integrations? Something like `IntegrationIssue`.
For example, a common `summary` field resolves to `title` for GitHub and to `summary` for Jira and Jira Server
What do you think?

packages/server/database/types/ImportedTask.ts
igorlesnenko: @mattkrick probably I had some issue in past which looks like not relevant anymore, reverted the change back

packages/client/components/PokerEstimateHeaderCardHtml.tsx
mattkrick: that's exactly how it used to work! https://github.com/ParabolInc/parabol/blob/f77f81a4935f3e7bd9e29f77ebc798e757a71710/packages/server/graphql/types/JiraIssue.ts#L45-L51

i had to get rid of it because there was no way to tell the client that the 2 fields were the same. for example, say a component requests `title`, then you navigate to another component that requires `summary`. normally it could use the cached value, but since they have different names, it'd have to make a separate network request because it saw that `summary` was a missing field.

Similarly, when creating an issue in an optimistic updater, you had to remember to populate both fields otherwise you could get an error. 

If 5 integrations all call the same field something different, that'd be 5 extra fields we'd have to remember to request, which just seemed like it was ripe for bugs. We could instead replace e.g. `summary` with `title`, but then that would get confusing when making the request to the integration.

I think the _right_ way to do it is by adding that field on the _client_, not the server, via a clientSchema extension. Then, you can have a client resolver that resolves 1 value to the other. If this sounds interesting to you, you can experiment with it, the details on how to do it are here: https://relay.dev/docs/guided-tour/reusing-cached-data/filling-in-missing-data/

mattkrick:
LGTM

packages/client/components/PokerEstimateHeaderCardHtml.tsx
igorlesnenko: @mattkrick I feel like it is more robust and easier to implement if the backend exposes ready-to-use fields. 
As we would need to pick some approach and made some refactoring, I created a separate discussion issue here #6374 and wrote my concerns and proposal with more details. What do you think?

Issue #5910, opened 2022-01-21
Add UML diagram of infrastructure to this repo 


mattkrick:
It'd be nice to have a UML diagram of what our infrastructure looks like in this repo.
That way, if we were to change how data flowed, then the first step would be submitting a PR to the diagram. Kind of like a single source of truth!
PlantUML is one option, but there are others: https://modeling-languages.com/text-uml-tools-complete-list/

AC
- [ ] UML of infra added to repo in PR

Estimate: 8 hours

mattkrick:
@BartoszJarocki found that github adopted mermaid for their markdown: https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/

PR #6235, opened 2022-03-18
Add flow diagram for services

Dschoordsch:
I created a diagram to help me understand how we're using the different
redis channels to connect the server with the GQL executor.

Relates to: #5910


mattkrick:
this is so darn cool!
We should add a link or the image itself to the root README but definitely safe to merge as is!

PR #6237, opened 2022-03-18
chore: Remove backfilling invoices logic from migration

Dschoordsch:
Migration is many years old and the used stripe manager was deleted in
PR #6161.


nickoferrall (approved):

mattkrick:
🙏  Thanks for finding this!

Issue #4631, opened 2021-02-02
Migrate to PG: Teams: Stop writes to RethinkDB (3 of 3)


mattkrick:
Now that #4630 is complete, we can stop writing to RethinkDB.

AC
- [ ] all writes to RethinkDB are stopped

Estimate: 8 hours

shapel:
Hello @mattkrick I would like you to consider my PR related to this issue. 
I know that it was not marked as "help wanted" but it was so low-hanging fruit that I couldn't resist ;)

PR #6239, opened 2022-03-18
refactor: Stop writes to RethinkDB Team table

shapel:
Hello

Bellow you can find changes relate to #4631

I know that it was not marked as "help wanted" but it was so low-hanging fruit that I couldn't resist ;)

### DIFF
--- a/packages/server/billing/helpers/terminateSubscription.ts
+++ b/packages/server/billing/helpers/terminateSubscription.ts
@@ -8,9 +8,6 @@ const terminateSubscription = async (orgId: string) => {
   // flag teams as unpaid
   const [rethinkResult] = await Promise.all([
     r({
-      updateTeam: r.table('Team').getAll(orgId, {index: 'orgId'}).update({
-        isPaid: false
-      }),
       stripeSubscriptionId: r
         .table('Organization')
         .get(orgId)
--- a/packages/server/database/migrations/20220329101759-dropTeams.ts
+++ b/packages/server/database/migrations/20220329101759-dropTeams.ts
@@ -1,49 +1,45 @@
-import getPg from '../../postgres/getPg'
+import getPg, {closePg} from '../../postgres/getPg'
 import {Team} from '../../postgres/queries/getTeamsByIds'
 import checkTeamEq from '../../postgres/utils/checkTeamEq'
 import {dropUndefined} from '../../postgres/utils/dropUndefined'
+import {closeRethink} from '../rethinkDriver'
 
 export const up = async function (r) {
   const pg = await getPg()
 
-  const teamTableExists = await pg.query<{exists: boolean}>(`SELECT EXISTS (SELECT 1 FROM "pg_tables" WHERE tablename = 'Team');`)
+  const teamTableExists = await pg.query<{exists: boolean}>(
+    `SELECT EXISTS (SELECT 1 FROM "pg_tables" WHERE tablename = 'Team');`
+  )
 
   // table does not exist on fresh DBs
   if (teamTableExists.rows[0].exists) {
+    // implicitly open connection to RethinkDB and PostgresSQL
     const errors = await checkTeamEq()
     if (errors.foundErrors > 1) {
       throw new Error(JSON.stringify(errors, undefined, ' '))
     }
   }
-
   await r.tableDrop('Team').run()
-};
+
+  await Promise.all([closePg(), closeRethink()])
+}
 
 export const down = async function (r) {
   if (await r.tableList().contains('Team').run()) return
-  await r
-    .tableCreate('Team')
-    .run()
-  await r
-    .table('Team')
-    .indexCreate('orgId')
-    .run()
-  await r
-    .table('Team')
-    .indexCreate('updatedAt')
-    .run()
-  await r
-    .table('Team')
-    .indexWait()
-    .run()
+  await r.tableCreate('Team').run()
+  await r.table('Team').indexCreate('orgId').run()
+  await r.table('Team').indexCreate('updatedAt').run()
+  await r.table('Team').indexWait().run()
 
   const pg = await getPg()
   const teams = await pg.query<Team>(`SELECT * FROM "Team";`)
 
   const batchSize = 1000
 
-  for (let current = 0; current * batchSize < teams.rowCount; current+= batchSize) {
+  for (let current = 0; current * batchSize < teams.rowCount; current += batchSize) {
     const filteredTeams = teams.rows.slice(current, current + batchSize).map(dropUndefined)
     await r.table('Team').insert(filteredTeams).run()
   }
-};
+
+  await closePg()
+}
--- a/packages/server/database/migrations/20220329101759-dropTeams.ts
+++ b/packages/server/database/migrations/20220329101759-dropTeams.ts
@@ -1,13 +1,12 @@
-import {Client} from 'pg'
-import getPgConfig from '../../postgres/getPgConfig'
+import getPg from '../../postgres/getPg'
 import {Team} from '../../postgres/queries/getTeamsByIds'
 import checkTeamEq from '../../postgres/utils/checkTeamEq'
 import {dropUndefined} from '../../postgres/utils/dropUndefined'
 
 export const up = async function (r) {
-  const client = new Client(getPgConfig())
-  await client.connect()
-  const teamTableExists = await client.query<{exists: boolean}>(`SELECT EXISTS (SELECT 1 FROM "pg_tables" WHERE tablename = 'Team');`)
+  const pg = await getPg()
+
+  const teamTableExists = await pg.query<{exists: boolean}>(`SELECT EXISTS (SELECT 1 FROM "pg_tables" WHERE tablename = 'Team');`)
 
   // table does not exist on fresh DBs
   if (teamTableExists.rows[0].exists) {
@@ -18,7 +17,6 @@ export const up = async function (r) {
   }
 
   await r.tableDrop('Team').run()
-  await client.end()
 };
 
 export const down = async function (r) {
@@ -39,9 +37,8 @@ export const down = async function (r) {
     .indexWait()
     .run()
 
-  const client = new Client(getPgConfig())
-  await client.connect()
-  const teams = await client.query<Team>(`SELECT * FROM "Team";`)
+  const pg = await getPg()
+  const teams = await pg.query<Team>(`SELECT * FROM "Team";`)
 
   const batchSize = 1000
 
@@ -49,5 +46,4 @@ export const down = async function (r) {
     const filteredTeams = teams.rows.slice(current, current + batchSize).map(dropUndefined)
     await r.table('Team').insert(filteredTeams).run()
   }
-  await client.end()
 };
--- a/packages/server/database/migrations/20220329101759-dropTeams.ts
+++ b/packages/server/database/migrations/20220329101759-dropTeams.ts
@@ -18,6 +18,7 @@ export const up = async function (r) {
   }
 
   await r.tableDrop('Team').run()
+  await client.end()
 };
 
 export const down = async function (r) {
@@ -48,4 +49,5 @@ export const down = async function (r) {
     const filteredTeams = teams.rows.slice(current, current + batchSize).map(dropUndefined)
     await r.table('Team').insert(filteredTeams).run()
   }
+  await client.end()
 };
--- a/packages/server/database/migrations/20220329101759-dropTeams.ts
+++ b/packages/server/database/migrations/20220329101759-dropTeams.ts
@@ -1,0 +1,51 @@
+import {Client} from 'pg'
+import getPgConfig from '../../postgres/getPgConfig'
+import {Team} from '../../postgres/queries/getTeamsByIds'
+import checkTeamEq from '../../postgres/utils/checkTeamEq'
+import {dropUndefined} from '../../postgres/utils/dropUndefined'
+
+export const up = async function (r) {
+  const client = new Client(getPgConfig())
+  await client.connect()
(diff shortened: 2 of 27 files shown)


Dschoordsch (changes requested):
Thank you for working on this! Looks good, the next steps missing would be:
1) run `checkTeamEq` against production data to verify we're not loosing any data
2) drop `Team` table from rethink

If you feel up to the task you could
* create a new rethink migration (run `yarn db:create dropTeam` to create it and then yarn db:migrate to run it)
* the migration should run `checkTeamEq` and only if it passes should drop the `Team` table from rethink. You would need to check why there is `maxErrors` parameter, it might be that we prime both tables slightly differently with the built-in teams
* `checkTeamEq` would be deleted in a separate PR

mattkrick:
ah shoot i see some files have gone stale.
we've moved intranetSchema to `private`, and just extracted the resolvers out of them

shapel:
I've rebase my changes related to stop writing to RethinkDB. 

I tried to create a migration to drop the 'Team' table after checking `checkTeamEq` but I ran into the fact that some PostgreSQL migrations depend on this table (like packages/server/postgres/migrations/1614030642692_refs-for-meetings.ts) this led to a crush for fresh app start (due to the order in which migrations are performed). 
So it requires to change PostgreSQL migrations too and I suppose could be done in separate PR.

Dschoordsch:
@shapel I included the migrations in #6308 but there is an issue with the migrations not finishing, probably missing some pg/rethink driver shutdown. I'm not sure when I have time to look into it, so if you want to give it a go, be my guest.

shapel:
👋 @Dschoordsch looks like I fixed the issue with migration and fresh dev start

Dschoordsch (changes requested):
Looks good, but `yarn db:migrate up` still hangs for me. I think the issue is that `checkTeamEq` indirectly initializes PG and Rethink which are then not closed.

shapel:
hmm, I can't reproduce locally on a fresh environment...
But I've changed usage to Postgres connection pool (same as `checkTeamEq`) instead of creating client (due to fact that transaction is not used here anyway).
@Dschoordsch could you check the case with this change?
PS Please attach db dump in case the issue still exists.

shapel:
👋 @Dschoordsch do you have any news about it?

packages/server/database/migrations/20220329101759-dropTeams.ts
```diff
@@ -0,0 +1,49 @@
...
+  // table does not exist on fresh DBs
+  if (teamTableExists.rows[0].exists) {
+    const errors = await checkTeamEq()
+    if (errors.foundErrors > 1) {
+      throw new Error(JSON.stringify(errors, undefined, ' '))
+    }
+  }
+
+  await r.tableDrop('Team').run()
+};
```
Dschoordsch, line 20: It blocks for me when manually running `yarn db:migrate up`. It works when adding and calling
```
await closeRethink()
await pg.end()
```
where `closeRethink` is from this change https://github.com/ParabolInc/parabol/blob/cfdd929342fbba35b6361b6328dd73122a23c94e/packages/server/database/rethinkDriver.ts#L237

Dschoordsch (changes requested):

packages/server/database/migrations/20220329101759-dropTeams.ts
shapel: @Dschoordsch you were right!
The problem was related to the implicitly openшing RethinkDB connection in `checkTeamEq`.
I also added a postgres closing for faster exit from migration process. 
Also we have to keep in mind that the internal methods for getting the postgres & rethink pool may be used in future migrations, so the pool instances need to be reset. I don't mean just closing them, but doing what is necessary to be able to initiated them again.

packages/server/database/rethinkDriver.ts
```diff
@@ -231,7 +231,16 @@ const getRethink = async () => {
   }
   // this is important because pm2 will restart the process & for whatever reason r isn't always healthy
   await r.waitForHealthy()
-  return (r as unknown) as ParabolR
+  return r as unknown as ParabolR
+}
+
+export const closeRethink = async () => {
+  if (promise) {
+    await (await promise)?.drain()
```
Dschoordsch, line 239: +1 Nice that you thought of resetting the local state. We can now get rid of the optional chaining `?` here since you check promise the line above.

tianrunhe:
I just run 
```graphql
mutation {
  checkRethinkPgEquality(maxErrors: 1000, tableName: "Team", writeToFile: true)
}
```
in production and it looks like we have a lot of false-positive inequalities. It's mostly for the `jiraDimensionFields` fields. The ordering of the elements in the array are different between RethinkDB & PG. Moreover, the ordering of keys in the element are different too but essentially they are the same. Look below as an example:

```JSON
{
    "rethinkRow": {
      "jiraDimensionFields": [
        {
          "cloudId": "bb3ae8ea-9d72-4d37-ad1b-24e2945db313",
          "dimensionName": "Scale",
          "fieldId": "customfield_10004",
          "fieldName": "Story Points",
          "fieldType": "number",
          "projectKey": "SP"
        },
        {
          "cloudId": "bb3ae8ea-9d72-4d37-ad1b-24e2945db313",
          "dimensionName": "Story Points",
          "fieldId": "customfield_10004",
          "fieldName": "Story Points",
          "fieldType": "number",
          "projectKey": "SP"
        }
      ]
    },
    "pgRow": {
      "jiraDimensionFields": [
        {
          "dimensionName": "Scale",
          "cloudId": "bb3ae8ea-9d72-4d37-ad1b-24e2945db313",
          "projectKey": "SP",
          "fieldName": "Story Points",
          "fieldType": "number",
          "fieldId": "customfield_10004"
        },
        {
          "dimensionName": "Story Points",
          "cloudId": "bb3ae8ea-9d72-4d37-ad1b-24e2945db313",
          "projectKey": "SP",
          "fieldName": "Story Points",
          "fieldType": "number",
          "fieldId": "customfield_10004"
        }
      ]
    }
}
```

I think we should add a custom equality checker in `checkRethinkPgEquality` for this field and run it against prod, make sure there are no real differences before we proceed. @mattkrick thoughts?

shapel:
👋 @tianrunhe your mention about improving `checkRethinkPgEquality` looks totally reasonable for me. So I'm going to look what could be improved there.

shapel:
@tianrunhe plz review https://github.com/ParabolInc/parabol/pull/6411 

By the way the example you provided processed by [areEqual](https://github.com/ParabolInc/parabol/blob/master/packages/server/postgres/utils/checkEqBase.ts#L2) as equal (so seems like only the order of the elements in the array matters, not the order of the object keys) 🤔

shapel:
👋 @tianrunhe can you run the mutation again? as you did at https://github.com/ParabolInc/parabol/pull/6239#issuecomment-1100171145 
Hopefully we will see other result due to fact that 6.56.0 2022-April-27 contains #6411.

tianrunhe:
> 👋 @tianrunhe can you run the mutation again? as you did at [#6239 (comment)](https://github.com/ParabolInc/parabol/pull/6239#issuecomment-1100171145) Hopefully we will see other result due to fact that 6.56.0 2022-April-27 contains #6411.

Hi @shapel, once again thanks for doing this dirty work! I have a piece of good news & bad news to share 😂
Good news: the differences for `jiraDimensionFields` are gone! 🎉 Great work on resolving that!
However, bad news... now it seems the major differences are in the `updatedAt` field, which tells me that there are some places in the code base where the writes between RethinkDB & PG are different. Here are some examples, notice in some cases the differences are small, but in others they are pretty wild:

```JSON
    "6IljpRp11n": {
      "error": [
        "updatedAt"
      ],
      "rethinkRow": {
        "updatedAt": "2021-09-09T19:03:23.762Z"
      },
      "pgRow": {
        "updatedAt": "2021-09-08T20:38:52.152Z"
      }
    },
    "5O0WSZYo81": {
      "error": [
        "updatedAt"
      ],
      "rethinkRow": {
        "updatedAt": "2021-09-09T20:23:45.695Z"
      },
      "pgRow": {
        "updatedAt": "2021-09-03T18:24:09.103Z"
      }
    },
    "hmz9s97S4": {
      "error": [
        "updatedAt"
      ],
      "rethinkRow": {
        "updatedAt": "2021-09-10T05:50:14.466Z"
      },
      "pgRow": {
        "updatedAt": "2020-10-15T11:47:59.894Z"
      }
    },
    "2uRH7TRlQi": {
      "error": [
        "updatedAt"
      ],
      "rethinkRow": {
        "updatedAt": "2021-09-10T05:50:14.468Z"
      },
      "pgRow": {
        "updatedAt": "2020-04-01T08:56:47.641Z"
      }
    }
```

Can you find out where those places could be?

shapel:
@tianrunhe thank you for sharing!
I see several places where the system does not set `updatedAt` in Postgres compared to RethinkDB.

I will add review comments about them in this PR.

The question is what should we do about it? 
Looks like we need to start setting `updateAt` in Postgres for mention cases.
Do I need to do this in this PR or in the new one?

packages/server/safeMutations/safeArchiveTeam.ts
```diff
@@ -14,18 +13,9 @@ const safeArchiveTeam = async (teamId: string, dataLoader: DataLoaderWorker) =>
...
-    isArchived: true,
-    updatedAt: new Date()
-  }
   const users = await Promise.all(userIds.map((userId) => dataLoader.get('users').load(userId)))
   const [rethinkResult, pgResult] = await Promise.all([
     r({
-      team: r
-        .table('Team')
-        .get(teamId)
-        .update(updates, {returnChanges: true})('changes')(0)('new_val')
```
shapel, line 27: The system does not set `updatedAt` in Postgres compared to RethinkDB.

packages/server/billing/helpers/terminateSubscription.ts
```diff
@@ -8,9 +8,6 @@ const terminateSubscription = async (orgId: string) => {
   // flag teams as unpaid
   const [rethinkResult] = await Promise.all([
     r({
-      updateTeam: r.table('Team').getAll(orgId, {index: 'orgId'}).update({
```
shapel, line 11: The system does not set `updatedAt` in RethinkDB compared to Postgres.

packages/server/safeMutations/safeArchiveTeam.ts
Dschoordsch: I think instead of adding the `updatedAt` times in code here where they could be forgotten, we should add a trigger like we did for "User" table: https://github.com/ParabolInc/parabol/blob/d05cd9b6c7ae3e2395e3c3e9de25d4786fa1184e/packages/server/postgres/migrations/1638368905980_userUpdatedAtTrigger.ts#L8

Dschoordsch: As a workaround for the current case, it would be good if we could:
1) @shapel allow some delta when comparing dates
2) @tianrunhe pull the dates from rethink and update PG
Let me know whether or not that works for you.

shapel: @Dschoordsch thanks for review.
Of course, we could add a trigger to the "Team" table, but I see that this is still not a common pattern across the system, which can be confusing if the DB and the web server might have time synchronization problems. 
It's always better to have one source of truth...
I could add such a trigger to all tables in a separate PR (to make sure we have the same approach throughout the system).

Regarding to possible workaround...
According to the examples provided by @tianrunhe, there are records in the system with dates that differ by about a year. And I guess they can't potentially be updated in the near function.  Therefore I suggest adding the `updateAt` field to the ignore list when comparing records for the time being.

Dschoordsch: Adding "updatedAt" to the ignored fields is fine for me. We should still try to correct our data while we have the chance, but I leave that up to @tianrunhe to decide.

I know the "set_updatedAt" trigger is not a common practice yet, but I'd like to get there as it's easy to forget updating that field, as you've noticed. This would include removing all manual updates to that field as well.

tianrunhe:
Hi @shapel, since your commits were released today I get a chance to rerun the equality checker again. It seems the `isOnboardTeam` property need to be added in `maybeUndefinedFieldsDefaultValues`? Example output:

```JSON
{
  "rethinkRecordsCompared": 49354,
  "foundErrors": 1000,
  "missingPgRows": [],
  "rJYPNVrwx": {
    "error": [
      "isOnboardTeam"
    ],
    "rethinkRow": {},
    "pgRow": {
      "isOnboardTeam": false
    }
  },
  "ryJjbY_7l": {
    "error": [
      "isOnboardTeam"
    ],
    "rethinkRow": {},
    "pgRow": {
      "isOnboardTeam": false
    }
  },
```

shapel:
Hi @tianrunhe thank to sharing this. I'm going to make this little fix tomorrow.

shapel:
The issue is still there, we just prepared to make sure we can merge the major changes. So I hope @tianrunhe gets a chance to run equality again when the latest fix goes into production.

Dschoordsch:
Was automatically close, reopening. @shapel thanks for paying attention!

shapel:
Hi @tianrunhe, do you have any news to share about 👆?

tianrunhe:
> Hi @tianrunhe, do you have any news to share about 👆?

Hey @shapel sorry I was on vacation last 2 weeks. I just run the validator again and it seems there are only 2 different records left. I suspect they are side effects of some manual edits. I'll dig further and report back!

tianrunhe:
@shapel Congratulations!

```JSON
{
  "rethinkRecordsCompared": 125478,
  "foundErrors": 0,
  "missingPgRows": [],
  "rethinkRowCount": 125478,
  "pgRowCount": 125478,
  "areRowCountsEqual": true
}
```

Please fix merge conflicts so we are ready to move on!

Dschoordsch (approved):

mattkrick:
🎉 fantastic work!!

PR #6250, opened 2022-03-22
feat(standups): Added end team prompt mutation #6171

BartoszJarocki:


Dschoordsch:
@BartoszJarocki can you add me as a reviewer next time? I usually go by this view: https://github.com/ParabolInc/parabol/pulls/review-requested/@me

BartoszJarocki:
yes, sorry, I should have done that!

packages/server/graphql/types/EndTeamPromptPayload.ts
```diff
@@ -0,0 +1,30 @@
...
+import makeMutationPayload from './makeMutationPayload'
+import Team from './Team'
+import TeamPromptMeeting from './TeamPromptMeeting'
+
+export const EndTeamPromptSuccess = new GraphQLObjectType<any, GQLContext>({
+  name: 'EndTeamPromptSuccess',
+  fields: () => ({
+    meeting: {
+      type: new GraphQLNonNull(TeamPromptMeeting),
+      resolve: ({meetingId}: {meetingId: string}, _args: unknown, {dataLoader}) => {
```
Dschoordsch, line 12: +1 We could use `resolve: resolveNewMeeting` here https://github.com/ParabolInc/parabol/blob/133ddb94ac90220ec2c8703470b274127f23a250/packages/server/graphql/resolvers.ts#L29

In the future we probably want to use just `resolve: (source) => source` here and do all the loading in `TeamPromptMeeting` but that's out of scope for now.

packages/server/graphql/types/EndTeamPromptPayload.ts
```diff
@@ -0,0 +1,30 @@
...
+        return dataLoader.get('newMeetings').load(meetingId)
+      }
+    },
+    meetingId: {
+      type: new GraphQLNonNull(GraphQLID)
+    },
+    team: {
+      type: new GraphQLNonNull(Team),
+      resolve: ({teamId}, _args: unknown, {dataLoader}) => {
```
Dschoordsch, line 21: +1 use https://github.com/ParabolInc/parabol/blob/133ddb94ac90220ec2c8703470b274127f23a250/packages/server/graphql/resolvers.ts#L94

packages/server/graphql/types/NotificationSubscriptionPayload.ts
```diff
@@ -34,9 +35,10 @@ const types = [
   CreateTaskPayload,
   DeleteTaskPayload,
   DisconnectSocketPayload,
-  EndCheckInSuccess,
   EndNewMeetingPayload,
+  EndCheckInSuccess,
   EndRetrospectiveSuccess,
+  EndTeamPromptSuccess,
```
Dschoordsch, line 41: -1 You're not publishing anything on the `SubscriptionChannel.NOTIFICATION` so there is no need to add the type here.

packages/server/graphql/mutations/endTeamPrompt.ts
```diff
@@ -0,0 +1,74 @@
...
+    {authToken, dataLoader, socketId: mutatorId}: GQLContext
+  ) => {
+    const r = await getRethink()
+    const viewerId = getUserId(authToken)
+    const now = new Date()
+    const operationId = dataLoader.share()
+    const subOptions = {mutatorId, operationId}
+
+    // AUTH
+    const meeting = (await r
```
Dschoordsch, line 32: +1 I think we should use
```
const meeting = dataLoader.get('newMeetings').load(meetingId) as MeetingTeamPrompt | null
```
here. You just need to make sure after you've updated it in rethink to also update `meeting.endedAt = now`

Dschoordsch (changes requested):

Dschoordsch:
Also tested the change and it works

packages/server/graphql/types/NotificationSubscriptionPayload.ts
BartoszJarocki: this was kind of future-proofing but you're right, for now there's no need to do it

packages/server/graphql/mutations/endTeamPrompt.ts
BartoszJarocki: what's the advantage of using a data loader here instead of a raw query?

Dschoordsch: Data will be cached in the dataLoader already when publishing the subscription, so 1 less load from DB.

BartoszJarocki: right! fixed.

BartoszJarocki:
@Dschoordsch do you mind taking a look one more time?

Dschoordsch (approved):
Works

BartoszJarocki:
@mattkrick do you mind taking a look? 🙏

mattkrick:
LGTM

PR #6253, opened 2022-03-22
chore: remove postcommit git command

blimmer:
This PR removes the `postcommit` hook to call `git update-index` because it slows down git operations like `rebase`. According to @mattkrick - this used to be required by JetBrains IDEs, but we don't need it anymore.


mattkrick:
LGTM!

Issue #6203, opened 2022-03-15
Standup UI - Integrate TipTap


BartoszJarocki:
We want to try doing standups with a different rich text editor - https://tiptap.dev/ 

AC
- [ ] add tip tap as a frontend dependency 
- [ ] add some dummy component which integrates tip tap
- [ ] test mobile support

Issue #5950, opened 2022-01-25
Add handling of the add response mutation


BartoszJarocki:

tianrunhe:
**Story Points: 13**
  [See the discussion](https://action.parabol.co/meet/chorM4RmGA/estimate/21) in PS2 Poker 1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

tianrunhe:
**Include?: Y**
  [See the discussion](https://action.parabol.co/meet/chorM4RmGA/estimate/22) in PS2 Poker 1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

Issue #5948, opened 2022-01-25
Implement mutation to handle adding responses


BartoszJarocki:
- [ ] When user starts a team prompt meeting, empty responses were created
- [ ] A mutation exists to update the content of a response. It takes prompt id and stringnified content as arguments

tianrunhe:
**Story Points: 13**
  [See the discussion](https://action.parabol.co/meet/chorM4RmGA/estimate/19) in PS2 Poker 1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

tianrunhe:
**Include?: Y**
  [See the discussion](https://action.parabol.co/meet/chorM4RmGA/estimate/20) in PS2 Poker 1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

PR #6255, opened 2022-03-22
feat(standups): Integrate TipTap

tianrunhe:
Implements #6203

Very basic editor component using TipTap. Mostly followed the [official guide](https://tiptap.dev/installation/react). Markdown is supported from the StarterKit out-of-box. I'm working on #5948 and #5950 which will actually use this component. See demo:

https://www.loom.com/share/6e4808a7ef9340fabba1256826d04543

Also tested on iOS & Android simulators, the editor works fine:

Screen Shot 2022-03-28 at 5 27 24 PM

Screen Shot 2022-03-28 at 5 27 35 PM


BartoszJarocki (approved):

packages/client/components/promptResponse/PromptResponseEditor.tsx
```diff
@@ -0,0 +1,63 @@
...
+    if (isEditing) return
+    setDoc(initialDoc)
+  }, [initialDoc])
+
+  const setEditing = (isEditing: boolean) => {
+    setIsEditing(isEditing)
+    setAutoFocus(false)
+  }
+
+  const onUpdate = ({editor}: EditorEvents['update']) => {
```
mattkrick, line 29: -1 this looks wrong. we shouldn't be serializing the entire document on update.
the only time we should serialize is when we send it to the server.

packages/client/components/promptResponse/PromptResponseEditor.tsx
```diff
@@ -0,0 +1,63 @@
...
+    const nextDoc = editor.getJSON()
+    setEditing(true)
+    setDoc(nextDoc)
+  }
+
+  const onSubmit = async ({editor}: EditorEvents['blur']) => {
+    const nextDoc = editor.getJSON()
+    setEditing(false)
+    setDoc(nextDoc)
```
mattkrick, line 38: -1 why set a doc to the serialized version of itself?

mattkrick (changes requested):
the update functionality looks wrong, but happy to merge this & fix it in the next version if it's already been fixed there

PR #6259, opened 2022-03-22
Feat/graphql shield

### LINKED ISSUE #6248: DX(PrivateSchema): separate authorization from resolvers
Predecessor: #6228

Using the private schema as a test, how should we separate authorization and resolution?

By breaking them apart, we have a clear separation of concerns, which is good for scalability.
It also means we should be able to identify glaring holes in our permissions.

### Suggested Options

- GraphQL Shield: https://graphql-shield.vercel.app/
- Resolver composition: https://www.graphql-tools.com/docs/resolvers-composition
- Authz: https://www.the-guild.dev/blog/graphql-authz (probably my least favorite)

### Considerations

Some mutations may change the underlying authorization. If authorization is checked using a dataloader, then that dataloader would have to be cleared before returning the mutation resolution, otherwise it'll result in stale authorization.
For example, let's say I call the mutation `removeFromTeam` which requires the viewer to be on TeamX. The authorizer uses a dataloader to see if an active TeamMember exists. Then, the resolver inactivates that team member & returns the team member. if that team member gets resolved using the dataloader, it'll still show up as active. 

### Background
This was the challenge problem for backend devs: https://www.notion.so/parabol/Back-end-Technical-Problem-a9d281e0396b41be85865e2b3a8461b4

We also talk about it here: https://github.com/ParabolInc/parabol/issues/4540#issuecomment-853465969

AC
- [ ] Create a 1 or many draft PRs that show what the code would look like using each option.
- [ ] If there are other options to consider, bring them up & weigh the pros/cons

Estimate: 8 hours

mattkrick:
Fix #6248

This introduces graphql-shield, which is a way to separate authorization from graphql resolvers.

TEST
- Our private schema is protected at the resolver level AND at the endpoint level. To test, comment out this line to remove the endpoint protection: https://github.com/ParabolInc/parabol/blob/155d10eb724ebb396a52e88cc3af78f9ac4ae43a/packages/server/graphql/intranetGraphQLHandler.ts#L17
- [ ] call a query on the private schema without superuser credentials, it says you're not authorized

Loom: https://www.loom.com/share/4e5ed97273fe4e95a92749fe3941057a


mattkrick:
i refactored all this junk when i started building this on the public schema, so i'll close this in favor of #6263 since it touches all the same files

PR #6260, opened 2022-03-23
feat(ado): Initial Azure DevOps integration

bobjac:
Initial PR.  Commits include the following:
- UI changes behind feature flags to connect to Azure DevOps in task integrations
- GraphQL schema updates to support AzureDevOps authentication
- Authentication to AzureDevOps using OAuth2 PKCE and required content-type
- schema update and placeholder for rest api to pull azure devops work items

Next steps:
- Modify AzureDevOps rest api to asynchronously call azure devops rest apis for each work item
- Create data loader function for azure devops work items rest api
- Enhance Azure DevOps graphql schema to incorporate combined id's similar to the jira issue

Happy to review all changes with you.


bobjac:
The tsconfig.json conflicts are cosmetic.  I am happy to resolve.  The gitlabTypes were no modified directly by any of our changes.  If you like, I can resolve by taking the values from the master branch in all cases.

packages/server/postgres/migrations/1647457468652_addAzureDevOpsAuthTable.ts
```diff
@@ -0,0 +1,33 @@
+import {Client} from 'pg'
+import getPgConfig from '../getPgConfig'
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+
+  await client.query(`
+    CREATE TABLE IF NOT EXISTS "AzureDevOpsAuth" (
```
Dschoordsch, line 9: -2 This data should go into the "TeamMemberIntegrationAuth" table. We did not finalize where we're storing `instanceIds` but these should probably also go into "TeamMemberIntegrationAuth".

packages/server/graphql/mutations/addAzureDevOpsAuth.ts
```diff
@@ -0,0 +1,111 @@
...
+import {SubscriptionChannel} from 'parabol-client/types/constEnums'
+import {GQLContext, GQLMutation} from '../graphql'
+import {isTeamMember} from '../../utils/authorization'
+import standardError from '../../utils/standardError'
+import segmentIo from '../../utils/segmentIo'
+import upsertAzureDevOpsAuths from '../../postgres/queries/upsertAzureDevOpsAuths'
+import AzureDevOpsServerManager from '../../utils/AzureDevOpsServerManager'
+import getAzureDevOpsAuthsByUserId from '../../postgres/queries/getAzureDevOpsAuthsByUserId'
+
+const addAzureDevOpsAuth: GQLMutation = {
```
Dschoordsch, line 15: -2 this should go into the `addTeamMemberIntegrationAuth` mutation.

packages/server/graphql/mutations/removeAzureDevOpsAuth.ts
```diff
@@ -0,0 +1,43 @@
+import {GraphQLID, GraphQLNonNull} from 'graphql'
+import removeGitHubAuthDB from '../../postgres/queries/removeGitHubAuth'
+import {getUserId, isTeamMember} from '../../utils/authorization'
+import publish from '../../utils/publish'
+import RemoveAzureDevOpsAuthPayload from '../types/RemoveAzureDevOpsAuthPayload'
+import {SubscriptionChannel} from 'parabol-client/types/constEnums'
+import {GQLContext} from '../graphql'
+import standardError from '../../utils/standardError'
+
+const removeAzureDevOpsAuth = {
```
Dschoordsch, line 10: -2 this should go into the `removeTeamMemberIntegrationAuth` table

Dschoordsch (changes requested):
Sorry, but there are some bigger changes needed. I think you may have followed an outdated pattern in the project as we're in the middle of cleaning up our integrations and we have older integrations (Jira and GitHub) not up to date while integrations following the new pattern (Jira Server and Gitlab) are still in progress.

Main part is the storage of integration data which should go into "IntegrationProvider" and "TeamMemberIntegrationAuth" mainly.

packages/server/integrations/helpers/authorizeOAuth2.ts
```diff
@@ -20,9 +43,12 @@ export const authorizeOAuth2 = async <
 }: AuthorizeOAuth2Params) => {
   const headers = {
     Accept: 'application/json',
-    'Content-Type': 'application/json',
     ...additonalHeaders
   }
+  if (!headers['Content-Type']) {
```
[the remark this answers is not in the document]
bobjac: Sounds good.  I was going to go that route originally.  I'll make the change.

packages/server/graphql/types/TeamMemberIntegrations.ts
```diff
@@ -57,6 +58,16 @@ const TeamMemberIntegrations = new GraphQLObjectType<{teamId: string; userId: st
         const auths = await dataLoader.get('slackAuthByUserId').load(userId)
         return auths.find((auth) => auth.teamId === teamId)
       }
+    },
+    azuredevops: {
```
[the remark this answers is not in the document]
bobjac: Sounds good.  Will modify.

packages/server/integrations/helpers/authorizeOAuth2.ts
```diff
@@ -10,6 +10,29 @@ interface AuthorizeOAuth2Params {
...
+    return JSON.stringify(body)
+  }
+
+  if (contentType.toLowerCase().startsWith('application/x-www-form-urlencoded')) {
+    let transformedBody = ''
+    Object.entries(body).forEach((entry) => {
+      if (transformedBody.length > 0) {
+        transformedBody += '&'
+      }
+      transformedBody += `${entry[0]}=${entry[1]}`
```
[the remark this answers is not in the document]
bobjac: Absolutely.  That should have been there already.  I will make this change.

packages/server/graphql/types/TeamMemberIntegrations.ts
```diff
@@ -57,6 +58,16 @@ const TeamMemberIntegrations = new GraphQLObjectType<{teamId: string; userId: st
...
+      type: AzureDevOpsIntegration,
+      description: 'All things associated with a A integration for a team member',
+      resolve: async ({teamId, userId}, _args: unknown, {authToken, dataLoader}) => {
```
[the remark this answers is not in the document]
bobjac: I will make the change

packages/server/graphql/types/UserFeatureFlags.ts
```diff
@@ -19,6 +19,11 @@ const UserFeatureFlags = new GraphQLObjectType<any, GQLContext>({
       type: new GraphQLNonNull(GraphQLBoolean),
       description: 'true if gitlab is allowed',
       resolve: ({gitlab}) => !!gitlab
+    },
+    azuredevops: {
```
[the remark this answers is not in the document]
bobjac: I will make the change

packages/server/utils/AzureDevOpsServerManager.ts
```diff
@@ -0,0 +1,53 @@
...
+    const body = {
+      ...params,
+      client_id: process.env.AZUREDEVOPS_CLIENT_ID!
+    }
+    const additonalHeaders = {
+      'Content-Type': 'application/x-www-form-urlencoded',
+      // eslint-disable-next-line prettier/prettier
+      'Origin': 'http://localhost:8081'
+    }
+    const authUrl = `https://login.microsoftonline.com/${process.env.AZUREDEVOPS_TENANT}/oauth2/v2.0/token`
```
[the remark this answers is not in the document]
bobjac: Correct.  I think I follow.  I'll add the tenants to the IntegrationProvider table

For context:
In the Azure DevOps world, each organization would be a tenant and each tenant has multiple projects.  A user can be a member of multiple organizations.

User----n*---->Tenant-------n*------>Project--------n*------->Work Item.

There are multiple types of work items (user story, bug, task, etc).

I'll store each tenant in the IntegrationProvider table

packages/server/postgres/migrations/1647457468652_addAzureDevOpsAuthTable.ts
bobjac: tagging @wdfox

packages/server/graphql/mutations/addAzureDevOpsAuth.ts
bobjac: tagging @wdfox

packages/server/graphql/mutations/removeAzureDevOpsAuth.ts
bobjac: tagging @wdfox

packages/server/graphql/types/TeamMemberIntegrations.ts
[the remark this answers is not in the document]
bobjac: Changed and pushed.

packages/server/graphql/types/UserFeatureFlags.ts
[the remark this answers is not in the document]
bobjac: Changed and pushed.

packages/server/integrations/helpers/authorizeOAuth2.ts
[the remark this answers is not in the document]
bobjac: I made this change.

packages/server/integrations/helpers/authorizeOAuth2.ts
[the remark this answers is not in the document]
bobjac: I change the mapper to use the URLSearchParams, but I am still returning a string via toString() since this is encapsulated in a transformBody function.

packages/server/utils/AzureDevOpsServerManager.ts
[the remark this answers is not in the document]
bobjac: @Dschoordsch  - For integration provider data in the IntegrationProvider table, how are you currently populating that table in production?  I am assuming that you are working with your customers' admins to provider the needed info (tenant, client_id, etc).   We didn't see any ui component or anything under /admin (besides /admin/graphql directly) for populating that table.  Should we assume that there is going to be an admin that is populating this table with the needed info, or is there a UI component that we are missing?

packages/server/postgres/migrations/1647457468652_addAzureDevOpsAuthTable.ts
bobjac: This has been updated.  Tagging @wdfox

packages/server/graphql/types/TeamMemberIntegrations.ts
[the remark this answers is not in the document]
bobjac: This has been updated

packages/server/utils/AzureDevOpsServerManager.ts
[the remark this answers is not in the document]
bobjac: I made that change.  I have so design options for the hierarchy.  In Azure DevOps we can get all work items for a user, but that would require multiple rest calls:

1. Find all ADO organizations I am a member of
2. Find all projects in each organization
3. Find all work items of a particular type in that project

We also have some nesting in other places.  We could hide the rest calls/constuction of the types interesting to the Parabol client in the RestManager, or we could expose the hierarchy in the GraphQL schema and handle it in resolvers.  My guess is that you will want to have a consistent graphql schema for your calling clients, but wanted to get your thoughts on it.

packages/server/graphql/mutations/removeAzureDevOpsAuth.ts
bobjac: @wdfox made this change

packages/server/graphql/mutations/addAzureDevOpsAuth.ts
bobjac: @wdfox  made this change

packages/server/utils/AzureDevOpsServerManager.ts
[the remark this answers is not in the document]
bobjac: @Dschoordsch - We made the broader changes that you were looking for.  There is still work to do on extending the schema.  Also, the token refresh pattern we implemented was based off the AtlassianLoaders.  In the newer pattern, it looks like the token refresh is being handled behind the JiraRestManager by an external library.  I can move the token refresh to the rest api layer if you would prefer that.

bobjac:
@Dschoordsch - @wdfox  and I made the changes that you requested.  We are adopting the latest pattern.  There is still work to be done, but we were looking for your input in the direction as we are trying to solidify the foundation to make more frequent prs.  One question I have - we used the token refresh pattern from the Atlassian implementation.  In that implementation, the refresh is executed in the data loader.  For Jira server, it is executed behind the RestManager which seems more fitting to me with the PKCE flow.  The JiraServer RestManager seems to be using a library for that as I didn't see an explicit refresh.

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,301 @@
...
+export const azureDevOpsAllWorkItems = (
+  parent: RootDataLoader
+): DataLoader<TeamUserKey, WorkItem[] | undefined, string> => {
+  return new DataLoader<TeamUserKey, WorkItem[] | undefined, string>(
+    async (keys) => {
+      const results = await Promise.allSettled(
+        keys.map(async ({userId, teamId}) => {
+          const returnWorkItems = [] as WorkItem[]
+          const auth = await parent.get('freshAzureDevOpsAuth').load({teamId, userId})
+          if (!auth) return undefined
```
Dschoordsch, line 144: +1 It's easier to handle in the rest of the code if we just return `[]` here. Not having `auth` is handled separately anyways.

packages/server/graphql/mutations/addTeamMemberIntegrationAuth.ts
```diff
@@ -102,6 +103,16 @@ const addTeamMemberIntegrationAuth = {
         const manager = new GitLabOAuth2Manager(clientId, clientSecret, serverBaseUrl)
         tokenMetadata = await manager.authorize(oauthCodeOrPat, redirectUri)
       }
+      if (service === 'azureDevOps') {
+        if (!oauthVerifier) {
+          return {
+            error: {message: 'Missing OAuth2 Verifier required for Azure DevOps authentication'}
+          }
+        }
+        tokenMetadata = (await AzureDevOpsServerManager.init(oauthCodeOrPat, oauthVerifier)) as
```
Dschoordsch, line 112: -1 Checking for `refreshToken` and asserting it exists should be done inside `init`

scripts/webpack/dev.client.config.js
```diff
@@ -132,7 +132,9 @@ module.exports = {
         prblIn: process.env.INVITATION_SHORTLINK,
         AUTH_INTERNAL_ENABLED: process.env.AUTH_INTERNAL_DISABLED !== 'true',
         AUTH_GOOGLE_ENABLED: process.env.AUTH_GOOGLE_DISABLED !== 'true',
-        AUTH_SSO_ENABLED: process.env.AUTH_SSO_DISABLED !== 'true'
+        AUTH_SSO_ENABLED: process.env.AUTH_SSO_DISABLED !== 'true',
+        azureDevOpsTenant: process.env.AZUREDEVOPS_TENANT,
+        azureDevOpsClientId: process.env.AZUREDEVOPS_CLIENT_ID,
```
Dschoordsch, line 137: -2 These should be removed

packages/server/utils/AzureDevOpsServerManager.ts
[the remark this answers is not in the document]
Dschoordsch: > [...] the token refresh pattern we implemented was based off the AtlassianLoaders. In the newer pattern, it looks like the token refresh is being handled behind the JiraRestManager by an external library. I can move the token refresh to the rest api layer if you would prefer that.

Jira Server token effectively does not expire (valid for ~5 years), so there is no refresh logic at all. Atlassian integration has also some client logic to refetch the query once the token expired [useFreshToken](https://github.com/ParabolInc/parabol/blob/90b642f04f65df862ca941df896c7b09c08bf6db/packages/client/modules/teamDashboard/components/ProviderRow/AtlassianProviderRow.tsx#L48+L61). That's quite fragile though.
 
I think a far better solution would be to check the validity of the token right before usage and refresh it if necessary right before the request. This would require some locking as this could be run from multiple processes simultaneously. [`RedisLock`](https://github.com/ParabolInc/parabol/blob/90b642f04f65df862ca941df896c7b09c08bf6db/packages/server/utils/RedisLock.ts) would be the right thing to use for locking in this case.

This can be handled in subsequent PRs.

packages/server/graphql/types/AzureDevOpsIntegration.ts
```diff
@@ -0,0 +1,158 @@
...
+                id: workItem.id.toString(),
+                url: workItem.url
+              }
+            })
+          )
+          return userStories
+        }
+      }
+    },
+    cloudProvider: {
```
Dschoordsch, line 128: -2 I think you would not have a cloudProvider, would you? This is the one global instance which is publicly available. An example would be gitlab.com as the `cloudProvider` vs. the self hosted GitLab instances as `sharedProvider`. As far as I understand all Azure DevOps instances would be `sharedProvider`s since they differ by the tenant. There is no single "azure.microsoft.com" instance where everyone can authenticate against, is there?

packages/server/graphql/types/AzureDevOpsIntegration.ts
```diff
@@ -0,0 +1,158 @@
...
+      description:
+        'A list of work items coming straight from the azure dev ops integration for a specific team member',
+      resolve: async (
+        {teamId, userId}: AzureDevOpsAuth,
+        {},
+        {authToken, dataLoader}: GQLContext
+      ) => {
+        const viewerId = getUserId(authToken)
+        if (viewerId !== userId) {
+          const err = new Error('Cannot access another team members issues')
```
Dschoordsch, line 108: +1 usually we would check for team membership

Dschoordsch (changes requested):
The current version looks close. Only issue is with the `IntegrationProvider` which should be added manually and not come from the environment.
Sorry for missing the question initially, feel free to ping me if I don't answer in time.

packages/server/graphql/types/AzureDevOpsIntegration.ts
wdfox: Agree completely. Thank you for your clear explanation on the difference between `cloudProvider` and `sharedProvider`. Change is made.

packages/server/graphql/types/UserFlagEnum.ts
```diff
@@ -1,14 +1,15 @@
 import {GraphQLEnumType} from 'graphql'
 
-export type UserFeatureFlagEnum = 'spotlight' | 'standups' | 'gitlab'
+export type UserFeatureFlagEnum = 'spotlight' | 'standups' | 'gitlab' | 'azuredevops'
```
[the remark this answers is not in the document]
bobjac: This change has been made

packages/server/graphql/types/AzureDevOpsIntegration.ts
bobjac: This change has been made

packages/server/graphql/mutations/addTeamMemberIntegrationAuth.ts
bobjac: This change has been made

scripts/webpack/dev.client.config.js
bobjac: @wdfox should have removed these

bobjac:
@Dschoordsch / @wdfox - 

Hi @Dschoordsch, 
We are given the option to merge to main.  I wanted to check and make sure you had no final requests for changes before merging.  When you give the go-ahead, do you use squash/merge or merge commit on your team?

packages/server/database/types/TaskIntegrationAzureDevOps.ts
```diff
@@ -0,0 +1,23 @@
+//import AzureDevOpsProjectKeyId from '../../../client/shared/gqlIds/AzureDevOpsProjectKeyId';
```
Dschoordsch, line 1: -1 please remove

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+}
+
+export const freshAzureDevOpsAuth = (
+  parent: RootDataLoader
+): DataLoader<TeamUserKey, IGetTeamMemberIntegrationAuthQueryResult | null, string> => {
+  return new DataLoader<TeamUserKey, IGetTeamMemberIntegrationAuthQueryResult | null, string>(
+    async (keys) => {
+      const results = await Promise.allSettled(
+        keys.map(async ({userId, teamId}) => {
+          console.log(`Inside keys.map userId: ${userId} | teamId: ${teamId}`)
```
Dschoordsch, line 91: -1 please remove

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+          const azureDevOpsAuthToRefresh = (await parent.get('teamMemberIntegrationAuths').load({
+            service: 'azureDevOps',
+            teamId,
+            userId
+          })) as IGetTeamMemberIntegrationAuthQueryResult | null
+          if (azureDevOpsAuthToRefresh === null) {
+            console.log('error line 61')
```
Dschoordsch, line 98: -1 please remove

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+            return null
+          } else {
+            console.log(`azureDevOpsAuthToRefresh: ${JSON.stringify(azureDevOpsAuthToRefresh)}`)
+          }
```
Dschoordsch, line 102: -1 please remove

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+          // const {accessToken: existingAccessToken, refreshToken} = azureDevOpsAuthToRefresh
+          const {accessToken: existingAccessToken, refreshToken, accessTokenSecret, providerId} = azureDevOpsAuthToRefresh
+          if (!refreshToken) {
+            console.log(`null condition hit - refreshToken:${refreshToken}`)
```
Dschoordsch, line 106: -1 please remove

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+            teamId,
+            userId
+          })) as IGetTeamMemberIntegrationAuthQueryResult | null
+          if (azureDevOpsAuthToRefresh === null) {
+            console.log('error line 61')
+            return null
+          } else {
+            console.log(`azureDevOpsAuthToRefresh: ${JSON.stringify(azureDevOpsAuthToRefresh)}`)
+          }
+          // const {accessToken: existingAccessToken, refreshToken} = azureDevOpsAuthToRefresh
```
Dschoordsch, line 103: -1 please remove

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+          const {accessToken: existingAccessToken, refreshToken, accessTokenSecret, providerId} = azureDevOpsAuthToRefresh
+          if (!refreshToken) {
+            console.log(`null condition hit - refreshToken:${refreshToken}`)
+            return null
+          }
+
+          // const {accessToken: existingAccessToken, refreshToken, accessTokenSecret, providerId} = azureDevOpsAuthToRefresh
```
Dschoordsch, line 110: -1 please remove

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+            if (!refreshToken || !accessTokenSecret) {
+              return null
+            }
+
+            const provider = await parent.get('integrationProviders').loadNonNull(providerId)
+
+            if (!provider) {
+              return null
+            }
+
```
Dschoordsch, line 127: -1 `loadNonNull` will throw if it's null

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+          }
+
+          // const {accessToken: existingAccessToken, refreshToken, accessTokenSecret, providerId} = azureDevOpsAuthToRefresh
+          const decodedToken = existingAccessToken && (decode(existingAccessToken) as any)
+          const now = new Date()
+          const inAMinute = Math.floor((now.getTime() + 60000) / 1000)
+          if (!decodedToken || decodedToken.exp < inAMinute) {
+            // console.log(`calling AzureDevOpsServerManager.refresh`)
+            // const oauthRes = await AzureDevOpsServerManager.refresh(refreshToken)
+            // console.log(`oauthRes: ${JSON.stringify(oauthRes)}`)
```
Dschoordsch, line 117: -1 please remove

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+            const manager = new AzureDevOpsServerManager(azureDevOpsAuthToRefresh, provider as IntegrationProviderAzureDevOps)
+
+            const oauthRes = await manager.refresh(refreshToken)
+            if (oauthRes instanceof Error) {
+              //sendToSentry(oautRes)
+              return null
+            }
+            const {accessToken, refreshToken: newRefreshToken} = oauthRes
+            console.log(`accessToken: ${accessToken} | newRefreshToken: ${newRefreshToken}`)
+            const updatedRefreshToken = !!newRefreshToken ? newRefreshToken : refreshToken
```
Dschoordsch, line 137: +1 same as `const updatedRefreshToken = newRefreshToken || refreshToken`

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+
+            const manager = new AzureDevOpsServerManager(azureDevOpsAuthToRefresh, provider as IntegrationProviderAzureDevOps)
+
+            const oauthRes = await manager.refresh(refreshToken)
+            if (oauthRes instanceof Error) {
+              //sendToSentry(oautRes)
+              return null
+            }
+            const {accessToken, refreshToken: newRefreshToken} = oauthRes
+            console.log(`accessToken: ${accessToken} | newRefreshToken: ${newRefreshToken}`)
```
Dschoordsch, line 136: -1 please remove

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+            const updatedRefreshToken = !!newRefreshToken ? newRefreshToken : refreshToken
+            // const updatedSameAccountAzureDevOpsAuths = userAzureDevOpsAuths
+            //   .filter((auth) => auth.accountId === azureDevOpsAuthToRefresh.accountId)
+            //   .map((auth) => ({
+            //     ...auth,
+            //     accessToken,
+            //     refreshToken: updatedRefreshToken
+            //   }))
+            // console.log(updatedSameAccountAzureDevOpsAuths)
+            // await upsertAzureDevOpsAuths(updatedSameAccountAzureDevOpsAuths)
```
Dschoordsch, line 146: -1 please remove

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+            const newAzureDevOpsAuth = {
+              ...azureDevOpsAuthToRefresh,
+              accessToken,
+              refreshToken: updatedRefreshToken
+            }
+            await upsertTeamMemberIntegrationAuth(newAzureDevOpsAuth)
+            console.log(`newAzureDevOpsAuth: ${newAzureDevOpsAuth}`)
```
Dschoordsch, line 153: -1 please remove

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+            return newAzureDevOpsAuth
+          }
+          console.log(`azureDevOpsAuthToRefresh: ${azureDevOpsAuthToRefresh}`)
```
Dschoordsch, line 156: -1 please remove

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+          const manager = new AzureDevOpsServerManager(auth, provider as IntegrationProviderAzureDevOps)
+
+          // const manager = new AzureDevOpsServerManager(accessToken)
+          const restResult = await manager.getMe()
+          const {error, azureDevOpsUser} = restResult
+          if (error !== undefined || azureDevOpsUser === undefined) {
+            console.log(error)
+            return undefined
+          }
+          const user = azureDevOpsUser as unknown as AzureDevOpsUser
```
Dschoordsch, line 257: -1 cast seems unnecessary, `azureDevOpsUser` is already of type `AzureDevOpsUser` here

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+          if (!auth) {
+            return undefined
+          }
+
+          const provider = await parent.get('integrationProviders').loadNonNull(auth.providerId)
+
+          if (!provider) {
+            return undefined
+          }
+
```
Dschoordsch, line 247: -1 checking `!provider` is redundant

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+            return []
+          }
+
+          const manager = new AzureDevOpsServerManager(auth, provider as IntegrationProviderAzureDevOps)
+
+          const userInfo = await parent.get('azureDevUserInfo').load({teamId, userId})
+          if (!userInfo) return []
+          const {id} = userInfo
+
+          // const manager = new AzureDevOpsServerManager(accessToken)
```
Dschoordsch, line 304: -1 please remove

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+
+export const azureDevOpsAllWorkItems = (
+  parent: RootDataLoader
+): DataLoader<TeamUserKey, WorkItem[] | undefined, string> => {
+  return new DataLoader<TeamUserKey, WorkItem[] | undefined, string>(
+    async (keys) => {
+      const results = await Promise.allSettled(
+        keys.map(async ({userId, teamId}) => {
+          const returnWorkItems = [] as WorkItem[]
+          // const auth = await parent.get('freshAzureDevOpsAuth').load({teamId, userId})
```
Dschoordsch, line 177: -1 getting 'freshAzureDevOpsAuth' here was the right call. Same for all other functions here. You can also think about modifying its return type to include the provider since you always need it.

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+            .get('teamMemberIntegrationAuths')
+            .load({service: 'azureDevOps', teamId, userId})
+
+          if (!auth) {
+            return []
+          }
+
+          const provider = await parent.get('integrationProviders').loadNonNull(auth.providerId)
+
+          if (!provider) {
```
Dschoordsch, line 345: -1 `provider` is never null

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+            return []
+          }
+
+          const manager = new AzureDevOpsServerManager(auth, provider as IntegrationProviderAzureDevOps)
+
+          // const manager = new AzureDevOpsServerManager(accessToken)
```
Dschoordsch, line 351: -1 please remove

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+
+          if (!provider) {
+            return null
+          }
+
+          const manager = new AzureDevOpsServerManager(auth, provider as IntegrationProviderAzureDevOps)
+          // const manager = new AzureDevOpsServerManager(accessToken)
+          const workItemIds: number[] = []
+          const workItemNum = parseInt(workItemId)
+          console.log(`workItemNum: ${workItemNum}`)
```
Dschoordsch, line 397: -1 please remove debug statements

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+          if (!isNaN(workItemNum)) {
+            workItemIds.push(workItemNum)
+          }
+          console.log(`workItemIds: ${workItemIds}`)
```
Dschoordsch, line 401: -1 please remove debug statements

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+          const restResult = await manager.getWorkItemData(instanceId, workItemIds)
+          const {error, workItems} = restResult
+          if (error !== undefined
+            || workItems.length !== 1
+            || !workItems[0]) {
+            console.log(error)
+            return null
+          } else {
+            console.log(`no error and workItems length of 1 returning: ${workItems[0]}`)
```
Dschoordsch, line 410: -1 please remove debug statements

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,495 @@
...
+            const azureDevOpsWorkItem: AzureDevOpsWorkItem = {
+              id: returnedWorkItem.id.toString(),
+              title: returnedWorkItem.fields['System.Title'],
+              teamProject: returnedWorkItem.fields['System.TeamProject'],
+              url: returnedWorkItem.url,
+              state: returnedWorkItem.fields['System.State'],
+              type: returnedWorkItem.fields['System.WorkItemType'],
+              service: 'azureDevOps'
+            }
+            console.log(`azureDevOpsWorkItem: ${JSON.stringify(azureDevOpsWorkItem)}`)
```
Dschoordsch, line 421: -1 please remove debug statements

packages/server/graphql/mutations/updatePokerScope.ts
```diff
@@ -1,5 +1,6 @@
 import {GraphQLID, GraphQLList, GraphQLNonNull} from 'graphql'
 import {SubscriptionChannel, Threshold} from 'parabol-client/types/constEnums'
+// import {XGitHubCheckConclusionState} from 'parabol-client/types/graphql'
```
Dschoordsch, line 3: -1 please remove

packages/server/graphql/mutations/updatePokerScope.ts
```diff
@@ -62,6 +63,12 @@ const updatePokerScope = {
     }
 
     const {endedAt, teamId, phases, meetingType, templateRefId, facilitatorStageId} = meeting
+    console.log(`endedAt: ${endedAt}`)
+    console.log(`teamId: ${teamId}`)
+    console.log(`phases: ${phases}`)
+    console.log(`meetingType: ${meetingType}`)
+    console.log(`templateRefId: ${templateRefId}`)
+    console.log(`facilitatorStageId: ${facilitatorStageId}`)
```
Dschoordsch, line 71: -1 please remove

packages/server/graphql/mutations/updatePokerScope.ts
```diff
@@ -83,9 +90,12 @@ const updatePokerScope = {
     // delete stages
     const subtractiveUpdates = updates.filter((update) => {
       const {action, serviceTaskId} = update
-      return action === 'DELETE' && !!stages.find((stage) => stage.serviceTaskId === serviceTaskId)
+      //return action === 'DELETE' && !!stages.find((stage) => stage.serviceTaskId === serviceTaskId)
```
Dschoordsch, line 93: -1 please use this instead of the debug version below

packages/server/graphql/mutations/updatePokerScope.ts
```diff
@@ -117,7 +127,8 @@ const updatePokerScope = {
     const newDiscussions = [] as Writeable<InputDiscussions>
     const additiveUpdates = updates.filter((update) => {
       const {action, serviceTaskId} = update
-      return action === 'ADD' && !stages.find((stage) => stage.serviceTaskId === serviceTaskId)
+      const a = action === 'ADD' && !stages.find((stage) => stage.serviceTaskId === serviceTaskId)
```
Dschoordsch, line 130: +1 please return directly

packages/server/graphql/mutations/updatePokerScope.ts
```diff
@@ -139,7 +150,7 @@ const updatePokerScope = {
       viewerId,
       meetingId
     )
-
+    console.log(`additiveUpdatesWithTaskIds: ${additiveUpdatesWithTaskIds}`)
```
Dschoordsch, line 153: -1 please remove

packages/server/graphql/mutations/updatePokerScope.ts
```diff
@@ -165,8 +176,11 @@ const updatePokerScope = {
       }))
       // MUTATIVE
       newDiscussions.push(...discussions)
+      console.log(`newDiscussions: ${newDiscussions}`)
       stages.push(...newStages)
+      console.log(`stages: ${stages}`)
       newStageIds = newStages.map(({id}) => id)
+      console.log(`newStageIds: ${newStageIds}`)
```
Dschoordsch, line 183: -1 please remove debug statements

packages/server/graphql/mutations/updatePokerScope.ts
```diff
@@ -182,6 +196,7 @@ const updatePokerScope = {
       })
       .run()
     if (newDiscussions.length > 0) {
+      console.log('inserting newDiscussions')
```
Dschoordsch, line 199: -1 please remove

packages/server/utils/AzureDevOpsServerManager.ts
```diff
@@ -0,0 +1,112 @@
...
+import {
+  OAuth2PkceAuthorizationParams,
+  OAuth2PkceRefreshAuthorizationParams
+} from '../integrations/OAuth2Manager'
+import {IntegrationProviderAzureDevOps} from '../postgres/queries/getIntegrationProvidersByIds';
+// import getIntegrationProvidersByIds, {IntegrationProviderAzureDevOps} from '../postgres/queries/getIntegrationProvidersByIds';
+import {IGetTeamMemberIntegrationAuthQueryResult} from '../postgres/queries/generated/getTeamMemberIntegrationAuthQuery';
+import {isError} from 'util';
+
+class AzureDevOpsServerManager extends AzureDevOpsManager {
```
Dschoordsch, line 15: -1 merge `AzureDevOpsManager` logic into this class

packages/server/utils/AzureDevOpsServerManager.ts
```diff
@@ -0,0 +1,112 @@
...
+      code_verifier: codeVerifier,
+      redirect_uri: makeAppURL(appOrigin, 'auth/ado')
+    })
+  }
+  // private readonly auth: IGetTeamMemberIntegrationAuthQueryResult
+  private readonly provider: IntegrationProviderAzureDevOps
+  // static provider: any;
+  // private readonly serverBaseUrl: string
+  // private readonly oauth: OAuth
+  // private readonly token: OAuth.Token
```
Dschoordsch, line 36: -1 please remove commented out code in this file

packages/server/tsconfig.json
```diff
@@ -9,16 +9,33 @@
...
   "files": [
     "types/webpackEnv.ts",
     "types/modules.d.ts",
     "../client/modules/email/components/SummaryEmail/MeetingSummaryEmail/MeetingSummaryEmail.tsx",
     "graphql/queries/tasks.ts",
     "graphql/queries/archivedTasks.ts",
+    "graphql/types/AzureDevOpsWorkItem.ts",
+    "graphql/types/AzureDevOpsRemoteProject.ts",
+    "graphql/types/AzureDevOpsIntegration.ts",
+    "dataloader/azureDevOpsLoaders.ts",
```
Dschoordsch, line 38: -1 adding these files manually should not be necessary since it is included by others already

packages/server/postgres/migrations/1648135135375_addAzureDevOpsIntegrationProvider.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    CREATE TYPE "IntegrationProviderServiceEnum" AS ENUM (
+      'gitlab',
+      'mattermost',
+      'jiraServer'
+    );
+
+    ALTER TABLE "IntegrationProvider"
+      ALTER COLUMN "service" TYPE "IntegrationProviderServiceEnum" USING "service"::text::"IntegrationProviderServiceEnum",
+      ADD CONSTRAINT global_provider_must_be_oauth2 CHECK (
```
Dschoordsch, line 45: -2 You're not modifying the "authStrategy" field so there is no need to modify this constraint

packages/server/integrations/helpers/authorizeOAuth2.ts
```diff
@@ -16,32 +37,38 @@ export const authorizeOAuth2 = async <
...
+    headers['Content-Type'] = 'application/json'
+  }
   const url = new URL(authUrl)
   if (searchParams) {
     Object.entries(searchParams).forEach((entry) => {
       url.searchParams.append(...entry)
     })
   }
-
+  console.log(`about to call AzureDevOps with url: ${url} and headers: ${headers}`)
```
Dschoordsch, line 58: -1 please remove

packages/server/integrations/OAuth2Manager.ts
```diff
@@ -4,11 +4,23 @@ export interface OAuth2AuthorizationParams {
   redirect_uri: string
 }
 
+export interface OAuth2PkceAuthorizationParams extends OAuth2AuthorizationParams {
+  code_verifier: string
+}
+
+//export interface OAuth2PkceRefreshAuthorizationParams extends OAuth2RefreshAuthorizationParams {}
```
Dschoordsch, line 11: -1 please remove

packages/server/integrations/OAuth2Manager.ts
```diff
@@ -4,11 +4,23 @@ export interface OAuth2AuthorizationParams {
...
+
 export interface OAuth2RefreshAuthorizationParams {
   grant_type: 'refresh_token'
   refresh_token: string
 }
 
+export interface OAuth2PkceRefreshAuthorizationParams extends OAuth2RefreshAuthorizationParams {
+  scope: string
+  redirect_uri: string
+  //client_secret: string
```
Dschoordsch, line 21: -1 please remove

packages/server/graphql/types/Task.ts
```diff
@@ -159,6 +162,7 @@ const Task: GraphQLObjectType = new GraphQLObjectType<any, GQLContext>({
         context,
         info
       ) => {
+        console.log(`entered TaskIntegration with integration" ${JSON.stringify(integration)}`)
```
Dschoordsch, line 165: -1 please remove

packages/server/graphql/types/Task.ts
```diff
@@ -29,8 +31,9 @@ import TaskIntegration from './TaskIntegration'
 import TaskStatusEnum from './TaskStatusEnum'
 import Team from './Team'
 import Threadable, {threadableFields} from './Threadable'
-import JiraServerIssueId from '~/shared/gqlIds/JiraServerIssueId'
-import GitLabServerManager from '../../integrations/gitlab/GitLabServerManager'
+// import AzureDevOpsServerManager from '../../../server/utils/AzureDevOpsServerManager'
+// import {IGetTeamMemberIntegrationAuthQueryResult} from '../../../server/postgres/queries/generated/getTeamMemberIntegrationAuthQuery'
+// import {WorkItem} from 'parabol-client/utils/AzureDevOpsManager'
```
Dschoordsch, line 36: -1 please remove

packages/server/graphql/types/Task.ts
```diff
@@ -176,6 +180,17 @@ const Task: GraphQLObjectType = new GraphQLObjectType<any, GQLContext>({
             issueId,
             providerId: integration.providerId
           })
+        } else if (integration.service === 'azureDevOps') {
+          console.log(`integration.service is azureDevOps`)
```
Dschoordsch, line 184: -1 please remove

packages/server/graphql/types/EstimateStage.ts
```diff
@@ -53,10 +53,13 @@ const EstimateStage = new GraphQLObjectType<Source, GQLContext>({
         _args: unknown,
         {dataLoader}
       ) => {
+        console.log('entered EstimateStage')
```
Dschoordsch, line 56: -1 please remove

packages/server/graphql/types/EstimateStage.ts
```diff
@@ -53,10 +53,13 @@ const EstimateStage = new GraphQLObjectType<Source, GQLContext>({
...
         const NULL_FIELD = {name: '', type: 'string'}
         const task = await dataLoader.get('tasks').load(taskId)
         if (!task) return NULL_FIELD
         const {integration} = task
+        console.log(`integration: ${integration}`)
+        console.log(`integration json: ${JSON.stringify(integration)}`)
```
Dschoordsch, line 62: -1 please remove

packages/server/graphql/types/EstimateStage.ts
```diff
@@ -92,6 +95,23 @@ const EstimateStage = new GraphQLObjectType<Source, GQLContext>({
 
           return {name: SprintPokerDefaults.SERVICE_FIELD_COMMENT, type: 'string'}
         }
+        if (service === 'azureDevOps') {
+          const {instanceId, issueKey, projectKey} = integration
+          console.log(`instanceId: ${instanceId}`)
+          console.log(`issueKey: ${issueKey}`)
+          console.log(`projectKey: ${projectKey}`)
```
Dschoordsch, line 102: -1 please remove

packages/server/graphql/types/EstimateStage.ts
```diff
@@ -92,6 +95,23 @@ const EstimateStage = new GraphQLObjectType<Source, GQLContext>({
...
+          const [dimensionName, team] = await Promise.all([
+            getDimensionName(meetingId),
+            dataLoader.get('teams').load(teamId)
+          ])
+          console.log(`dimensionName: ${dimensionName}`)
+          console.log(`team: ${JSON.stringify(team)}`)
```
Dschoordsch, line 108: -1 please remove

Dschoordsch (changes requested):
Basically just cleanup left

packages/server/graphql/types/EstimateStage.ts
bobjac: Removed

packages/server/graphql/types/EstimateStage.ts
bobjac: Removed

packages/server/graphql/types/EstimateStage.ts
bobjac: Removed

packages/server/graphql/types/EstimateStage.ts
bobjac: Removed

packages/client/components/AzureDevOpsFieldDimensionDropdown.tsx
```diff
@@ -0,0 +1,66 @@
+/*import styled from '@emotion/styled'
```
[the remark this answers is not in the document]
bobjac: Removed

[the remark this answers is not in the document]
bobjac: Removed

packages/client/components/AzureDevOpsScopingSearchResultsRoot.tsx
```diff
@@ -0,0 +1,52 @@
...
+  const queryRef = useQueryLoaderNow<AzureDevOpsScopingSearchResultsQuery>(
+    azureDevOpsScopingSearchResultsQuery,
+    {
+      teamId,
+      first: 25,
+      queryString: normalizedQueryString,
+      isWIQL
+    }
+  )
+  console.log(`AzureDevOpsScopingSearchResultsRoot.queryRef: ${queryRef}`)
```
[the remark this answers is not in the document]
bobjac: Removed

packages/client/components/AzureDevOpsFieldMenu.tsx
```diff
@@ -0,0 +1,111 @@
+/*import graphql from 'babel-plugin-relay/macro'
```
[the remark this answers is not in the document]
bobjac: We are going to need this file for our next set of changes when we push back the estimate fields.  I can either delete the file for this PR or check in a default dropdown.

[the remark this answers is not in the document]
bobjac: Removed

packages/client/components/AzureDevOpsScopingSearchBar.tsx
```diff
@@ -0,0 +1,46 @@
...
+  return (
+    <SearchBar>
+      <SearchIcon>search</SearchIcon>
+      <AzureDevOpsScopingSearchInput meeting={meeting} />
+      <AzureDevOpsScopingSearchFilterToggle meeting={meeting} />
+    </SearchBar>
+  )
+}
+
+export default createFragmentContainer(AzureDevOpsScopingSearchBar, {
```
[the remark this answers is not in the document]
bobjac: Modified

bobjac:
@Dschoordsch / @wdfox  - all the comments have been resovlved.

bobjac:
@Dschoordsch  - Please let me know if there are any other issues that you would like addressed.  If not, do you prefer squash or commit merges?  Also, would you prefer to pull the trigger on the merge or should I?

packages/client/mutations/UpdateAzureDevOpsDimensionFieldMutation.ts
```diff
@@ -0,0 +1,37 @@
+/*import graphql from 'babel-plugin-relay/macro'
```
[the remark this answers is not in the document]
bobjac: Initial implementation of mutation added

packages/server/dataloader/azureDevOpsLoaders.ts
bobjac: Modified to just expand on ...azureDevOpsUser in the returned object

packages/server/utils/AzureDevOpsServerManager.ts
bobjac: Changed

packages/client/components/ScopePhaseArea.tsx
```diff
@@ -95,30 +101,22 @@ const ScopePhaseArea = (props: Props) => {
...
-    return idx === -1 ? 1 : idx
-  })
-  const isTabActive = (label: typeof baseTabs[number]['label']) => {
-    return activeIdx === tabs.findIndex((tab) => tab.label === label)
-  }
 
-  const selectIdx = (idx: number) => {
-    setActiveIdx(idx)
-    const service = tabs[idx]?.label ?? 'Jira'
-    window.localStorage.setItem('favoriteService', service)
```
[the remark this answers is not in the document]
bobjac: Fixed

packages/client/utils/AzureDevOpsClientManager.ts
```diff
@@ -0,0 +1,91 @@
+import {CreateAzureDevOpsAuthorizeUrlMutationResponse} from '~/__generated__/CreateAzureDevOpsAuthorizeUrlMutation.graphql'
+import Atmosphere from '../Atmosphere'
+import {MenuMutationProps} from '../hooks/useMutationProps'
+import AddTeamMemberIntegrationAuthMutation from '../mutations/AddTeamMemberIntegrationAuthMutation'
+import CreateAzureDevOpsAuthorizeUrlMutation from '../mutations/CreateAzureDevOpsAuthorizeUrlMutation'
+import getOAuthPopupFeatures from './getOAuthPopupFeatures'
+import makeHref from './makeHref'
+
+class AzureDevOpsClientManager {
+  fetch = window.fetch.bind(window)
```
Dschoordsch, line 10: -1 fetch is unused

packages/server/graphql/mutations/createAzureDevOpsAuthorizeUrl.ts
```diff
@@ -0,0 +1,72 @@
...
+    if (!provider || provider.authStrategy !== 'oauth2') {
+      return standardError(new Error('Integration provider not found'), {
+        userId: viewerId,
+        tags: {providerId}
+      })
+    }
+
+    if (provider.service === 'azureDevOps') {
+      const scope = '499b84ac-1321-427f-aa17-267ca6975798/.default'
+      const url = `https://login.microsoftonline.com/${provider.tenantId}/oauth2/v2.0/authorize?client_id=${provider.clientId}&response_type=code&redirect_uri=${redirect}&response_mode=query&scope=${scope}&state=${providerState}&code_challenge=${code}&code_challenge_method=S256`
```
Dschoordsch, line 60: -2 we can easily just request `tenantId` and `clientId` in the provider row and then get rid of this mutation. For Jira Server this was added because it uses OAuth1 and requires a roundtrip to their servers to get this URL.

packages/server/graphql/types/AzureDevOpsIntegration.ts
```diff
@@ -0,0 +1,214 @@
...
+  isActive: boolean
+  accessToken: string | null
+  refreshToken: string | null
+  scopes: string | null
+  accessTokenSecret: string | null
+  // Note: instanceIds does not belong here, in fact this type as a whole should be removed eventually
+  instanceIds: string[]
+}
+
+interface AzureDevOpsAuth
```
Dschoordsch, line 40: -1 please remove. Appears to be unused, only `{teamId, userId}` part is used

packages/server/graphql/types/AzureDevOpsIntegration.ts
```diff
@@ -0,0 +1,214 @@
...
+type IntegrationProviderServiceEnum = 'azureDevOps' | 'gitlab' | 'jiraServer' | 'mattermost'
+
+type WorkItemArgs = {
+  first: number
+  after?: string
+  queryString: string | null
+  isWIQL: boolean
+}
+
+interface IGetAzureDevOpsAuthByUserIdTeamIdQueryResult {
```
Dschoordsch, line 24: -1 please remove

packages/server/graphql/types/AzureDevOpsRemoteProject.ts
```diff
@@ -0,0 +1,53 @@
...
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'The parabol userId this issue was fetched for'
+    },
+    self: {
+      type: new GraphQLNonNull(GraphQLID)
+    },
+    instanceId: {
+      type: new GraphQLNonNull(GraphQLID),
+      description:
+        'The instance ID that the project lives on. Does not exist on the Azure DevOps object!'
```
Dschoordsch, line 36: -1 

> Does not exist on the Azure DevOps object!

This part is only confusing. Does that mean this field is broken and needs a `resolve` function? Do we need this field then?

packages/server/graphql/types/AzureDevOpsRemoteProject.ts
```diff
@@ -0,0 +1,53 @@
...
+    },
+    key: {
+      type: new GraphQLNonNull(GraphQLString)
+    },
+    name: {
+      type: new GraphQLNonNull(GraphQLString)
+    },
+    simplified: {
```
Dschoordsch, line 44: -1 This is Jira specific where the field values depend on being legacy/simplified issue type.

Please remove all non-ado fields from this

packages/server/graphql/types/CreateAzureDevOpsAuthorizeUrlPayload.ts
```diff
@@ -0,0 +1,18 @@
+import {GraphQLObjectType, GraphQLString} from 'graphql'
+import {GQLContext} from '../graphql'
+import StandardMutationError from './StandardMutationError'
+
+export const CreateAzureDevOpsAuthorizeUrlPayload = new GraphQLObjectType<any, GQLContext>({
```
Dschoordsch, line 5: -1 this needs to be removed when you remove the mutation.

packages/server/graphql/types/TeamMemberIntegrations.ts
```diff
@@ -57,6 +58,17 @@ const TeamMemberIntegrations = new GraphQLObjectType<{teamId: string; userId: st
...
+    azureDevOps: {
+      type: new GraphQLNonNull(AzureDevOpsIntegration),
+      description: 'All things associated with a A integration for a team member',
+      resolve: (source) => source
+      // resolve: async ({teamId, userId}, _args: unknown, {authToken, dataLoader}) => {
+      //   //if (!isTeamMember(authToken, teamId)) return null
+      //   console.log(`authToken: ${authToken}`)
+      //   console.log(`teamId: ${teamId}`)
+      //   return dataLoader.get('freshAzureDevOpsAuth').load({teamId, userId})
+      // }
```
Dschoordsch, line 71: -1 please remove dead code

packages/server/graphql/types/AzureDevOpsRemoteProject.ts
```diff
@@ -0,0 +1,53 @@
...
+import AzureDevOpsProjectId from 'parabol-client/shared/gqlIds/AzureDevOpsProjectId'
+import {GQLContext} from '../graphql'
+import IntegrationProviderServiceEnum from './IntegrationProviderServiceEnum'
+import RepoIntegration, {repoIntegrationFields} from './RepoIntegration'
+
+const AzureDevOpsRemoteProject = new GraphQLObjectType<any, GQLContext>({
+  name: 'AzureDevOpsRemoteProject',
+  description: 'A project fetched from Azure DevOps in real time',
+  interfaces: () => [RepoIntegration],
+  isTypeOf: ({service}) => service === 'azuredevops',
```
Dschoordsch, line 11: -1 It should be `service === 'azureDevOps'`

packages/server/utils/AzureDevOpsServerManager.ts
```diff
@@ -0,0 +1,457 @@
...
+
+export interface AzureDevOpsError {
+  code: number
+  message: string
+}
+
+const MAX_REQUEST_TIME = 5000
+
+class AzureDevOpsServerManager {
+  fetch = fetch as any
```
Dschoordsch, line 149: -1 no need for this property anymore, just use `fetch` from `node-fetch` directly. Also it was never set.

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,425 @@
...
+            const updatedRefreshToken =  newRefreshToken || refreshToken
+            const newAzureDevOpsAuth = {
+              ...azureDevOpsAuthToRefresh,
+              accessToken,
+              refreshToken: updatedRefreshToken
+            }
+            await upsertTeamMemberIntegrationAuth(newAzureDevOpsAuth)
+            return newAzureDevOpsAuth
+          }
+          return azureDevOpsAuthToRefresh
```
Dschoordsch, line 134: +1 You could also add the `provider` as a field to make client side code easier

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,425 @@
...
+
+export const azureDevOpsAllWorkItems = (
+  parent: RootDataLoader
+): DataLoader<TeamUserKey, WorkItem[] | undefined, string> => {
+  return new DataLoader<TeamUserKey, WorkItem[] | undefined, string>(
+    async (keys) => {
+      const results = await Promise.allSettled(
+        keys.map(async ({userId, teamId}) => {
+          const returnWorkItems = [] as WorkItem[]
+          const auth = await parent
```
Dschoordsch, line 154: -1 Why not use `const auth = await parent.get('freshAzureDevOpsAuth').load({teamId, userId})`?

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,425 @@
...
+}
+
+export const azureDevUserInfo = (
+  parent: RootDataLoader
+): DataLoader<TeamUserKey, AzureUserInfo | undefined, string> => {
+  return new DataLoader<TeamUserKey, AzureUserInfo | undefined, string>(
+    async (keys) => {
+      const results = await Promise.allSettled(
+        keys.map(async ({userId, teamId}) => {
+          const auth = await parent
```
Dschoordsch, line 195: -1 Why not use `const auth = await parent.get('freshAzureDevOpsAuth').load({teamId, userId})`?

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,425 @@
...
+}
+
+export const allAzureDevOpsAccessibleOrgs = (
+  parent: RootDataLoader
+): DataLoader<TeamUserKey, Resource[], string> => {
+  return new DataLoader<TeamUserKey, Resource[], string>(
+    async (keys) => {
+      const results = await Promise.allSettled(
+        keys.map(async ({userId, teamId}) => {
+          const auth = await parent
```
Dschoordsch, line 238: -1 Why not use `const auth = await parent.get('freshAzureDevOpsAuth').load({teamId, userId})`?

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -0,0 +1,425 @@
...
+  )
+}
+
+export const allAzureDevOpsProjects = (
+  parent: RootDataLoader
+): DataLoader<TeamUserKey, TeamProjectReference[], string> => {
+  return new DataLoader<TeamUserKey, TeamProjectReference[], string>(
+    async (keys) => {
+      const results = await Promise.allSettled(
+        keys.map(async ({userId, teamId}) => {
```
Dschoordsch, line 281: -1 Why not use `const auth = await parent.get('freshAzureDevOpsAuth').load({teamId, userId})`?

same for functions below

Dschoordsch (changes requested):

packages/client/components/AzureDevOpsScopingSearchResults.tsx
```diff
@@ -0,0 +1,126 @@
...
+  return (
+    <ResultScroller>
+      {edges.map(({node}) => {
+        console.log(`INside map. usedServiceTaskIds: ${JSON.stringify(usedServiceTaskIds)} `)
+        return (
+          <ScopingSearchResultItem
+            key={node.id}
+            service={'azureDevOps'}
+            usedServiceTaskIds={usedServiceTaskIds}
+            serviceTaskId={'dev.azure.com/jacobsrj:2649cbcf-37c1-484d-a290-1409d9d5b1cf:'+ node.id}
```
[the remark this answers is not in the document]
bobjac: This looks like it was added as part of the searching feature.  I am not sure why that is there.  It should be pulled/constructed from the node.url property.  That will contain the instance/project.  I am fixing this now.

packages/client/components/PokerDimensionFinalScorePicker.tsx
```diff
@@ -124,7 +125,9 @@ const PokerDimensionFinalScorePicker = (props: Props) => {
             />
           )}
 
-          {(integrationType === 'JiraIssue' || integrationType === 'JiraServerIssue') && (
+          {(integrationType === 'JiraIssue' ||
+            integrationType === 'JiraServerIssue' ||
+            integrationType === 'AzureDevOpsWorkItem') && (
             <JiraFieldDimensionDropdown
```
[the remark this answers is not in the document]
bobjac: I modified the code in the other branch to create a AzureDevOpsFieldDimensionDropdown.  It has enough AzureDevOps specific implementation to deviate from the JiraFieldDimensionDropDown.

packages/server/utils/AzureDevOpsServerManager.ts
bobjac: The standard fetch will complain about the signal parameter since it is not cast as any.

packages/server/dataloader/azureDevOpsLoaders.ts
bobjac: changed

packages/server/dataloader/azureDevOpsLoaders.ts
bobjac: changed

packages/server/dataloader/azureDevOpsLoaders.ts
bobjac: changed

packages/server/graphql/types/AzureDevOpsRemoteProject.ts
bobjac: That comment was carry over from the JiraRemoteProject.  Removed

packages/server/graphql/types/AzureDevOpsRemoteProject.ts
bobjac: removed

packages/server/graphql/mutations/createAzureDevOpsAuthorizeUrl.ts
bobjac: tagging @wdfox  here.  I believe @wdfox added the mutation here because the needed information to create the url (tenandId, etc) is now stored server-side in the database.  I believe he will still need to pull it from the backend.  Are you suggesting that we pull it from a graphql query instead of a mutation?

packages/server/utils/AzureDevOpsServerManager.ts
bobjac: I removed it and tested with the node-fetch and it was failing

packages/server/graphql/mutations/createAzureDevOpsAuthorizeUrl.ts
wdfox: Fixed this by removing the mutation and using the provider row query per the recommendation

packages/server/utils/AzureDevOpsServerManager.ts
Dschoordsch: -1 I checked and saw there is a type mismatch between node-fetch and the `AbortController` and `RequestInit` coming from DOM. Adding the imports
```
import fetch, {RequestInit} from 'node-fetch'
import AbortController from 'abort-controller'
```
fixes these. Then we can get rid of this member and use `fetch` directly.

Dschoordsch (changes requested):

packages/client/components/ScopePhaseArea.tsx
```diff
@@ -73,12 +75,18 @@ const ScopePhaseArea = (props: Props) => {
...
   const gitlabIntegration = viewerMeetingMember?.teamMember.integrations.gitlab
   const jiraServerIntegration = viewerMeetingMember?.teamMember.integrations.jiraServer
+  const azureDevOpsIntegration = viewerMeetingMember?.teamMember.integrations.azureDevOps
+
   const isGitLabProviderAvailable = !!(
     gitlabIntegration?.cloudProvider?.clientId || gitlabIntegration?.sharedProviders.length
   )
+
+  const allowAzureDevOps =
+    !!azureDevOpsIntegration.sharedProviders.length && featureFlags.azureDevOps
```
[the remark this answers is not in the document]
bobjac: resolved

Dschoordsch (approved):
I tested with #6448 and was able to complete OAuth flow, so let's get this in.

Issue #5828, opened 2022-01-04
Implement Create abstract NotificationManager class


tiffanyhan:
Implement #5748

## Acceptance criteria
* [ ] Refactor Slack and Mattermost notifications to resemble following structure
```
interface NotificationManagerAuths {
  slack: SlackAuth | null,
  mattermost: MattermostAuth | null
}
type NotificationManagerType = keyof NotificationManagerAuths

class IntegratedServicesNotifier {
  managers: {[key in NotificationManagerType]?: NotificationIntegrationManager} = {}

  constructor(auths: NotificationManagerAuths) {
    // init this.managers based on auths
    if (auths.slack) {
      this.managers.slack = new SlackNotificationIntegrationManager(auths.slack)
    }
...
  }

  // returning void because otherwise errors would need to be picked apart per integration and this manager should deal with retries
  // preferences should be limited to enable/disable an integration for a certain action, we don't want to support sending different events to different channels for example
  startMeeting(meeting: NewMeeting, team: Team, preferences: NotificationManagerType[]): Promise<void>
  endMeeting(meeting, team, preferences): Promise<void>
  startTimeLimit(meeting, team, preferences): Promise<void>
  endTimeLimit(meeting, team, preferences): Promise<void>
}

type NotifyResponse = 'success' | {
  error: Error
  recoverable: boolean
}

export interface NotificationIntegrationManager {
  startMeeting(meeting: NewMeeting, team: Team): Promise<NotifyResponse>
  endMeeting(meeting, team): Promise<NotifyResponse>
  startTimeLimit(meeting, team): Promise<NotifyResponse>
  endTimeLimit(meeting, team): Promise<NotifyResponse>
}

// new dataloaders to load all available notification auths
const auths: NotificationManagerAuths = await dataLoader.get('notificationManagerAuths').load({teamId, userId})
const manager = new IntegratedServicesNotifier(auths)

const preferences = await dataLoader.get('notificationPreferences')
manager.startMeeting(meeting, team, preferences.startMeeting
```

jordanh:
Could you move this issue to the Reviewer Review column and tag some folks (like @Dschoordsch or @BartoszJarocki) to look at this? Once this API has been reviewed, I'll prioritize this for implementation

tiffanyhan:
@Dschoordsch mind reviewing the API in the issue description for this issue?

Dschoordsch:
[...]

PR #6262, opened 2022-03-23
feat(notifications): Refactor Slack/Mattermost into NotificationHelper

Dschoordsch:
A common `NotificationHelper` loads the settings and sends the message
via the individual `NotificationIntegrationHelper`s. These then don't
need to load any additional data and are just responsible for sending
out single messages.

Resolves:
* #5828


packages/server/graphql/mutations/helpers/notifications/NotificationIntegrationHelper.ts
```diff
@@ -0,0 +1,21 @@
+import {Team} from '../../../../postgres/queries/getTeamsByIds'
+import Meeting from '../../../../database/types/Meeting'
+
+export type NotifyResponse =
+  | 'success'
+  | {
+      error: Error
+      // true if the error is transient
+      retry?: boolean
```
Dschoordsch, line 9: These errors are unused just like these were before

packages/server/graphql/mutations/helpers/notifications/getSummaryText.ts
```diff
@@ -1,11 +1,12 @@
...
-import MeetingPoker from '../../../../database/types/MeetingPoker'
+import {isMeetingRetrospective} from '../../../../database/types/MeetingRetrospective'
+import {isMeetingAction} from '../../../../database/types/MeetingAction'
+import {isMeetingPoker} from '../../../../database/types/MeetingPoker'
+import Meeting from '../../../../database/types/Meeting'
 
-const getSummaryText = (meeting: MeetingRetrospective | MeetingAction | MeetingPoker) => {
-  if (meeting.meetingType === 'retrospective') {
+const getSummaryText = (meeting: Meeting) => {
+  if (isMeetingRetrospective(meeting)) {
```
Dschoordsch, line 9: Typecheck functions were added to avoid casting

Dschoordsch:
@nickoferrall Could you take a look please?

packages/server/database/types/MeetingAction.ts
```diff
@@ -14,6 +14,10 @@ interface Input {
   facilitatorUserId: string
 }
 
+export function isMeetingAction(meeting: Meeting): meeting is MeetingAction{
```
nickoferrall, line 17: +2 love the helper functions!

packages/server/graphql/mutations/helpers/notifications/NotificationHelper.ts
```diff
@@ -0,0 +1,96 @@
...
+    getSlack(dataLoader, event, teamId),
+    getMattermost(dataLoader, meeting, teamId)
+  ])
+
+  return [...slack, ...mattermost]
+}
+
+export const NotificationHelper = {
+  // returning void because otherwise errors would need to be picked apart per integration and this manager should deal with retries
+  async startMeeting(dataLoader: DataLoaderWorker, meetingId: string, teamId: string) {
```
nickoferrall, line 54: +2 I love how succinct all of the functions are in this file. It's so easy to read

packages/server/graphql/mutations/helpers/notifications/NotificationHelper.ts
```diff
@@ -0,0 +1,96 @@
...
+    .load({event, teamId})
+  return notifications.map(SlackNotificationHelper)
+}
+
+async function getMattermost(dataLoader: DataLoaderWorker, meeting: Meeting, teamId: string) {
+  const {facilitatorUserId} = meeting
+  const provider = await dataLoader
+    .get('bestTeamIntegrationProviders')
+    .load({service: 'mattermost', teamId, userId: facilitatorUserId})
+  return provider ? [MattermostNotificationHelper(provider as IntegrationProviderMattermost)] : []
```
nickoferrall, line 24: Do you know why we return the types for our primaryKeyLoaderMakers but not for our other dataLoaders?

Dschoordsch: We return a union type for this one so we need to cast to the correct one

https://github.com/ParabolInc/parabol/blob/eb295331feba7e2534c48012d252c2a52b802905/packages/server/postgres/queries/getIntegrationProvidersByIds.ts#L44-L47

mattkrick (changes requested):
Looking good! Just one change to make the `Notifier` pattern look similar to the TaskIntegration pattern

packages/server/graphql/mutations/helpers/notifications/SlackNotificationHelper.ts
```diff
@@ -0,0 +1,178 @@
...
+        text: 'See estimates',
+        url: pokerUrl
+      }
+      return makeButtons([estimateButton, summaryButton])
+    default:
+      throw new Error('Invalid meeting type')
+  }
+}
+
+export const SlackNotificationHelper: NotificationIntegrationHelper<SlackNotificationAuth> = (
```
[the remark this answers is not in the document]
Dschoordsch: I'm still not sure if this is the right thing to do. It is about whether or not we put the responsibility for storing the data into the integration. It should be either the integration knows how to store and retrieve the data or some other part of the app. At the moment it is a mix of both. We put the responsibility for storing clearly into the mutations with the new `IntegrationProvider` and `TeamMemberIntegrationAuth` but the responsibility for retrieving the data is then put into the integration somehow. So I think we have 2 models which we could follow going forward:

1) try to take the structure and storage out of the integrations and manage it uniformly for different ones
  pros:
  * we do that to some extend already with unified auth
  * less duplication in storage/loading code
  cons:
  * storage might not be optimised for single integration, i.e. JSON or lots of NULL fields
 
2) integrations manage the storage on their own
  pros:
  * rest of the app doesn't need to care
  * specifically tailored to the integrations data
  cons:
  * duplication in storing and loading code
  * more possibilities for inconsistent behaviour (e.g. Mattermost notifies only 1 channel while slack can notify multiple and has settings per notification type)

Right now we're doing neither of both which spreads the logic into more places than necessary. I was leaning towards 1) with the caveat that I did not change functionality for how Slack/Mattermost data is loaded, but I would like to have that more consistent going forward.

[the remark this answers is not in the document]
Dschoordsch: @mattkrick could you take a look?

packages/server/graphql/mutations/helpers/notifications/getSummaryText.ts
Dschoordsch: Yes, that's right. Before this worked because the meeting was handed down directly with the correct type. If I have an `endMeeting(meetingId)` function, then at some point I would need to check if it matches any of those types. This is the place now where this is done, when we need this distinction.

Dschoordsch:
@mattkrick could you please take a look?

packages/server/graphql/mutations/helpers/notifications/getSummaryText.ts
mattkrick: I'm not sure I follow. Do you have a code example of the benefit?

packages/server/graphql/mutations/helpers/notifications/SlackNotificationHelper.ts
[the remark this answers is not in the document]
mattkrick: Do you have a code example of each? I'm not 100% sure I understand both solutions.

I'm a big fan of single-purpose classes. My intent with the "Manager" class is something that only facilitates communication with an external API. It can read & write, but it only knows about the service. You should be able to call e.g. `manager.addComment()` and not have to worry whether it'll do that via REST (like jira) or GraphQL (like github). 

Since these managers only care about communication with remote services, I'd like to keep postgres out of them. We can pass their return values to postgres on our own or through an ORM or anything else we choose, but ideally that stuff lives outside of the manager.

All that said, putting a dataloader inside a manager is breaking that encapsulation. Maybe it's OK because it's only a read? Maybe instead we pass in a thunk like `endMeeting(meetingId, getEvents)`. where `getEvents = () => dataLoader.get('slackEvents').load({event: 'meetingEnd'})`?

I'm not sure if that helps. If you write down some code examples we can go through the merits of each option

packages/server/graphql/mutations/helpers/notifications/Notifier.ts
```diff
@@ -0,0 +1,45 @@
+import {DataLoaderWorker} from '../../../graphql'
+
+/**
+ * Empty Notifier prototype to allow iterating the methods typesafe
+ */
+export const EmptyNotifier = {
```
mattkrick, line 6: -1 this pattern of abstract noop methods isn't very conventional. My concern is if other devs start copying this pattern, would the codebase get harder to read?

As constructed, `IntegrationNotifier` is essentially a singleton, it's just built without using the `class` syntactical sugar. 

2 options that are a little more expressive:

```ts
class NotifierManager {
  // put each type in as a method
  endMeeting(dataLoader, meetingId, teamId) {
    return Promise.allSettled(notifiers.map((notifier.endMeeting(dataLoader, meetingId, teamId)
  } 
}
export IntegrationManager = new NotifierManager()
```

```ts
interface Notifier {
  endMeeting: (meetingId: string) => Promise<NotifyResponse>
  startMeeting: (meetingId: string) => Promise<NotifyResponse>
}

interface NotifierManager extends Notifier {}
class NotifierManager {
  constructor() {
    // use a helper to do the work
    const methods = ['endMeeting', 'startMeeting']
    methods.forEach((method) => {
      this[method] = async (...args: any[]) => {
        return notifiers.map((notifier) => notifier[method](...args))
      }
    })
  }
}

export default IntegrationNotifier = new NotifierManager()
```

packages/server/graphql/mutations/helpers/notifications/MattermostNotifier.ts
```diff
@@ -0,0 +1,339 @@
...
+    properties: {
+      teamId,
+      notificationEvent: event
+    }
+  })
+
+  return 'success'
+}
+
+const makeEndMeetingButtons = (meeting: Meeting) => {
```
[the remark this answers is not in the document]
Dschoordsch: This will work locally, but if we only pass `meetingId` to the notifier, it needs to load the meeting and will only get `Meeting` as type. At some point we would need to make sure it's of type `AnyMeeting` which would require basically the same check.

packages/server/graphql/mutations/helpers/notifications/SlackNotificationHelper.ts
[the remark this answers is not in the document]
Dschoordsch: Keeping the dataloader and PG calls out of the managers is basically option 1) which I also wanted to achieve with the version where you've left the comment. Because someone needs to still load the settings/authentication, I had encapsulated it in the `NotificationHelper`. This is similar to `TaskIntegrationManagerFactory` but has the downside, that it needs to know how you load each integration settings. This was amplified by the fact that Slack and Mattermost operate quite differently (from the app logic).

We can also partially solve this by loading all Slack settings for all events (like we've done before). Then we'd only have one initialization step instead of different ones for each method call.

packages/server/graphql/mutations/helpers/notifications/Notifier.ts
Dschoordsch: Going with the copy & paste method as there is no way to ensure all methods are listed. That's why I had introduced the empty object placeholder to have the type information available at runtime.

packages/server/graphql/mutations/helpers/notifications/getSummaryText.ts
Dschoordsch: It's easy enough to do it afterwards. Dataloader is still kept out of the implementations, so resolving for now as it's not a pressing issue.

Dschoordsch:
I addressed Matt's concerns so I will go forward merging it after I did some testing.

Issue #6265, opened 2022-03-24
Record session only if `isWatched === true`


Dschoordsch:
Right now sessions are recorded always in Datadog. For LogRocket we had introduced `isWatched` property on user to decide whether a session should be recorded.
This is probably a good idea for Datadog as well for privacy concerns.

Related:
* #6149
* #6135

Issue #6135, opened 2022-02-28
Remove LogRocket


nickoferrall:
Now that we're using DataDog's session replays, we no longer need LogRocket. We've already stopped paying for LogRocket, so let's remove the relevant code.

### AC

- [ ] Remove LogRocket from the codebase

jordanh:
@nickoferrall if you haven't already, you might want to add this to the Planning meeting

nickoferrall:
Sure thing, added!

Dschoordsch:
I'm not going to remove `isWatched` now as this could also be used with DataDog going forward.

mattkrick:
fixed in #6266

PR #6266, opened 2022-03-24
chore: Remove LogRocket

Dschoordsch:
Left `isWatched` logic in place for now although it is unused, but we
probably want to use that for Datadog #6265.

Resolves:
* #6135


nickoferrall (approved):
LGTM - we can also remove it here: https://github.com/ParabolInc/parabol/blob/174f41d91534648548eea47c0eb1d697acc1dc8f/scripts/webpack/dev.client.config.js#L128

mattkrick:
great!

Issue #6213, opened 2022-03-16
Unable to select all issues in Poker


nickoferrall:
When selecting all issues in Poker, only one is added to the Estimate phase and one "Unknown Story" is added. See gif:

![buggy-select-all](https://user-images.githubusercontent.com/39854876/158602047-b74abbfa-e55a-44a7-a6e2-decc75818363.gif)

If I refresh the page, the issues are visible in the Estimate phase, so there might be a problem with the `UpdatePokerScopeMutation` updater.

The same behaviour happens for GitHub issues, Jira issues, and Parabol tasks. 

### AC

- [ ] Clicking "Select All" successfully adds all issues to the Estimate Phase, and their titles are visible in the sidebar

JimmyLv:
I think this is the same issue as https://github.com/ParabolInc/parabol/issues/6384?

nickoferrall:
Yep, that's right. I'll close #6384

Issue #6050, opened 2022-02-14
GitLab v3: Task Integration Spec Written


nickoferrall:
Once the task integration has been rearchitected (see https://github.com/ParabolInc/parabol/issues/5968), create an issue(s) outlining how we'll integrate Parabol tasks with GitLab.

### AC:

- [ ] Implementation issue(s) created for the GitLab task integration

nickoferrall:
Implementation issue (https://github.com/ParabolInc/parabol/issues/6428) created so I'll close this one

Issue #6007, opened 2022-02-05
Speed up repoIntegrations query


mattkrick:
I want to get #5883 merged, but we need to fix the logic soon to keep the app feeling snappy.
This is high priority because #5883 will introduce a much longer wait to open the menu. For users who have 100+ jira projects, the wait could now be in the seconds, which isn't good if we're trying to increase integration adoption.

The current logic runs a map/reduce on tasks for that team member to see which integrations they have used in the past.
Additionally, it fetches every project that the user has access to. That requires multiple roundtrips to jira because we have to paginate the jira projects.

To fix, we have options:
- Cache a list of used repos in redis
- Cache a list of fetched repos in redis (fetched from all the integrations)
- Don't fetch 3rd party integration repos until they start searching (i don't like this option)
 
Feel free to fight me on this, but I propose the following logic:
- [ ] repoIntegrations checks to see if `repoIntegration:${teamId}:${userId}` exists in redis. if it does, return that
- [ ] if it doesn't exist, then it runs what it currently does today & caches the result in redis
- [ ] to invalidate: when a task is pushed to an integration, the mutation checks to see if that integration repo exists in the redis key above. if `redisResult.find((repo) => repo.id === id)?.lastUpdatedAt > 0` don't invalidate. else, invalidate the redis cached value
- [ ] to invalidate in the case that a repo was recently added in github, I propose the following. When they search for a repo by typing a search query, we filter locally (just like today). If they don't click anything within 5 second, we call `repoIntegrations(force: true)`, which blows away the cache & fetches from integrations again.

jordanh:
General feedback while it's fresh:

Gee, it'd love to see whatever we decided and implement expressed as an interface for a developer to fill out for future integrations. Could be real tricky to just "know" that you should worry about caching this stuff for future integrations

nickoferrall:
Now that we're storing the `favoriteService` in https://github.com/ParabolInc/parabol/pull/6331, we could sort the `favoriteService` integrations above other services. `favoriteService` could be a nullable argument passed into the `repoIntegrations` query.

mattkrick:
related: #6723

- i love the interface
- i love the favoriteService (i can't believe i named it that though 😞)
- i love using defer, but that might take awhile to make the server support it
- i'd like to get rid of `allAvailableRepoIntegrations`, we should have just 1 field, but multiple args

Proposed AC
[...]

PR #6267, opened 2022-03-24
feat(sprint-poker): Add GitLab issue

### LINKED ISSUE #6033: Sprint Poker: GitLab v1: Create new issue
Just like GitHub & Jira, a user is able to create a new issue in GitLab. 

### AC

- [ ] Display a FAB
- [ ] Clicking the FAB shows the new issue input
- [ ] If the user enters a title into the input and submits it, a new issue is created in GitLab
- [ ] If there's an error creating the issue, share feedback with the user

nickoferrall:
Fix #6033 

### To test

- Integrate with GitLab (you may need to add the feature flag if you can't see GitLab in the integration provider row)
- Make sure you're part of the GitLab team Parabol1
- Create a Sprint Poker meeting, go to the Scope phase, and see your GitLab issues
- Click the FAB button, select a project, enter a title, and see that the issue has been created


packages/client/mutations/handlers/handleGitLabCreateIssue.ts
```diff
@@ -0,0 +1,53 @@
...
+    project,
+    'GitLabProjectEdge'
+  )
+  const mysteryIssueInEdge = newEdge!
+    .getLinkedRecord('node')
+    ?.getLinkedRecords('issues')![0]
+    ?.getValue('id')
+  console.log('🚀 ', {mysteryIssueInEdge})
+  newEdge.setValue(now, 'cursor')
+  ConnectionHandler.insertEdgeBefore(gitlabProjectsConn, newEdge)
```
nickoferrall, line 50: Hey @mattkrick, in the GitLab [scoping results](https://github.com/ParabolInc/parabol/blob/79ac25607102e774cb2339dbc0725dcfb4fafbf6/packages/client/components/GitLabScopingSearchResults.tsx#L97), we query the projects and then grab the issues within those projects. This differs from the other integrations where we just query the issues.

When creating a new issue, I want to insert a new edge in the connection. However, we can't have an issues connection that is a child of the projects connection. So, I want the new edge to be a project that contains the new issue. 

I can see here that the new project edge does contain the new issue. I can see it in the Relay dev tools too. However, when you add a new issue, you'll see in this [console log](https://github.com/ParabolInc/parabol/blob/79ac25607102e774cb2339dbc0725dcfb4fafbf6/packages/client/components/GitLabScopingSearchResults.tsx#L139) that a new project edge has been added but it does not contain the new issue.

I'm wondering if you might know why that it and if there's a better way of inserting the issue?

packages/client/types/constEnums.ts
```diff
@@ -266,7 +266,7 @@ export const enum Providers {
   GITHUB_NAME = 'GitHub',
   GITHUB_DESC = 'Use GitHub Issues from within Parabol',
   GITHUB_SCOPE = 'admin:org_hook,read:org,repo,user,write:repo_hook',
-  GITLAB_SCOPE = 'read_api',
+  GITLAB_SCOPE = 'api',
```
nickoferrall, line 269: Ideally, I'd prefer a less invasive [scope](https://docs.gitlab.com/ee/user/project/settings/project_access_tokens.html), but `write_repository` or `write_registry` won't let us read the projects. Including one of the read scopes like `read_api,write_repository` is rejected for being a malformed URL even though it follows the same comma separated pattern shown in their [docs](https://docs.gitlab.com/ee/user/profile/personal_access_tokens.html).

packages/client/components/GitLabScopingSearchResults.tsx
```diff
@@ -97,6 +98,7 @@ const GitLabScopingSearchResults = (props: Props) => {
                       edges {
                         node {
                           ... on _xGitLabProject {
+                            fullPath
```
nickoferrall, line 101: `repoIntegration` requires `fullPath` to determine whether it's a GitLab project. See rootSchema

packages/client/components/NewGitLabIssueInput.tsx
```diff
@@ -0,0 +1,265 @@
...
+      fragment NewGitLabIssueInput_viewer on User {
+        id
+        team(teamId: $teamId) {
+          id
+        }
+        teamMember(teamId: $teamId) {
+          id
+          integrations {
+            gitlab {
+              api {
```
nickoferrall, line 121: Unlike the GitHub and Jira integrations, I'm querying GitLab directly. [See here](https://github.com/ParabolInc/parabol/issues/6033#issuecomment-1069039404) for detailed explanation. TL;DR: repo integrations grabs all possible integrations rather than just the GitLab/Jira/GitHub ones that we want.

packages/server/graphql/queries/helpers/fetchAllRepoIntegrations.ts
```diff
@@ -13,13 +14,25 @@ const fetchAllRepoIntegrations = async (
...
+    prevRepoIntegrations,
+    jiraProjects,
+    githubRepos,
+    gitlabProjects,
+    jiraServerProjects
+  ] = await Promise.all([
     getPrevRepoIntegrations(userId, teamId, permLookup),
     dataLoader.get('allJiraProjects').load({teamId, userId}),
     fetchGitHubRepos(teamId, userId, dataLoader, context, info),
+    fetchGitLabProjects(teamId, userId, dataLoader, context, info),
```
nickoferrall, line 27: This file is being used in `allAvailableRepoIntegrations` and `repoIntegrations` in the `TeamMember` object. 

As mentioned [here](https://github.com/ParabolInc/parabol/pull/6267/files#r835236268), we're not using `repoIntegrations` in the Poker GitLab integrations, unlike GitHub or Jira. 

However, initially, I started following the `repoIntegration` pattern and created the `fetchGitLabProjects` file. I think it's best to keep it because we'll need it in the task integration, where we do want to display all `repoIntegrations`. There's an issue for this here: https://github.com/ParabolInc/parabol/issues/6050

I haven't updated `getPrevRepoIntegrations` to account for GitLab because we don't need it in Poker, and it'll be refactored anyway - see issue: https://github.com/ParabolInc/parabol/issues/6007

packages/client/types/constEnums.ts
Dschoordsch: Did you try setting the exact same scopes in the GitLab app? There seems to be an issue with requesting less scope than the app requires

packages/server/graphql/mutations/helpers/createGitLabTask.ts
```diff
@@ -0,0 +1,50 @@
...
+  context: GQLContext,
+  info: GraphQLResolveInfo,
+  dataLoader: DataLoaderWorker
+) => {
+  const {accessToken, providerId} = gitlabAuth
+  if (!accessToken) return {error: new Error('Invalid GitLab auth')}
+  const {title, contentState} = splitDraftContent(rawContent)
+  const body = stateToMarkdown(contentState)
+  const provider = await dataLoader.get('integrationProviders').load(providerId)
+  if (!provider?.serverBaseUrl) return {error: new Error('serverBaseUrl not found')}
```
Dschoordsch, line 23: +1 Our DB data should be sanitized on insert, so `provider.serverBaseUrl!` should be safe. There is already an issue to add better typed dataloaders.

packages/client/mutations/UpdatePokerScopeMutation.ts
```diff
@@ -216,13 +216,18 @@ const UpdatePokerScopeMutation: StandardMutation<TUpdatePokerScopeMutation, Hand
...
+            })
             const gitlabIssue = store.get(gid)
             const iid = gitlabIssue?.getValue('iid')
             const optimisticGitLabIssue = createProxyRecord(store, '_xGitLabIssue', {
               title,
               iid
             })
-            optimisticTask.setLinkedRecord(optimisticGitLabIssue, 'integration')
+            optimisticTaskIntegration.setLinkedRecords([optimisticGitLabIssue], 'issue')
+            optimisticTask.setLinkedRecord(optimisticTaskIntegration, 'integration')
```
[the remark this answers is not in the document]
nickoferrall: Yeah, the select all functionality isn't working in Sprint Poker in prod. I've created an issue for this and voted to add it to our next Sprint: https://github.com/ParabolInc/parabol/issues/6213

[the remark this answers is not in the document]
Dschoordsch: I think that's unrelated and only happens if some GitLab issues were added before. Does not happen on master.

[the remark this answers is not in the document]
nickoferrall: You're right! I was trying to sneak in a new bug on top of the existing bug. I fixed it by updating the `UpdatePokerScopeMutation` optimistic updater

packages/client/types/constEnums.ts
nickoferrall: Yep, I've played around with a bunch of different scopes and adjusted the GitLab app to require the exact same scopes. 

- When using multiple scopes, like `read_api,write_repository`, the auth page in GitLab says: `The requested scope is invalid, unknown, or malformed.'. 
- When I just used `read_api`, I get an error message saying I can't create an issue. 
- When `api` is the `GITLAB_SCOPE` and it's the scope in the GitLab app, it works for me but you [mentioned in Slack](https://parabol.slack.com/archives/C836NA350/p1648479438025419) that you ran into an error. So I've added `api` and `read_api` to the GitLab app which you mentioned now works for you. You can find the app permissions [here](https://gitlab.com/groups/parabol1/-/settings/applications/209182).

Dschoordsch (changes requested):
I was encouraged to offer my high level opinion more often: I think this button is not very important. I didn't even know it existed. Sadly we don't have any data whether it is actually used by anyone, so maybe if you feel like it, you could sneak in an `integration` field
https://github.com/ParabolInc/parabol/blob/ef26d0b6ef6610af354b202d609a000d7c153db0/packages/server/graphql/mutations/createTask.ts#L72

Dschoordsch (commented):
GitHub started 2 pending reviews for me

nickoferrall:
> we don't have any data whether it is actually used by anyone

True, there's an issue in GitLab v1 to track this and a bunch of other data points, so we'll soon find out! https://github.com/ParabolInc/parabol/issues/6303

packages/client/components/ScopePhaseArea.tsx
```diff
@@ -108,6 +109,39 @@ const ScopePhaseArea = (props: Props) => {
     setActiveIdx(2)
   }
 
+  // swipeable views won't ignore null children so conditionally create them: https://github.com/oliviertassinari/react-swipeable-views/issues/271
+  const contents = [
```
Dschoordsch, line 113: -1 I think this is the right approach, but I don't like that the order here needs to manually be matched to the order of `baseTabs`. Can you refactor it that we have something like:
```ts
const contents: Record<typeof baseTabs[number]['label'], JSX.Element> = {
  'GitHub': <ScopePhaseAreaGitHub
      isActive={isTabActive('GitHub')}
      gotoParabol={goToParabol}
      meetingRef={meeting}
    />,
  'Jira': <ScopePhaseAreaJira
      isActive={isTabActive('Jira')}
      gotoParabol={goToParabol}
      meeting={meeting}
    />,
  'Parabol': <ScopePhaseAreaParabolScoping isActive={isTabActive('Parabol')} meeting={meeting} />,
    'Jira Server': <ScopePhaseAreaJiraServer
      isActive={isTabActive('Jira Server')}
      gotoParabol={goToParabol}
      meetingRef={meeting}
    />,
  'GitLab': <ScopePhaseAreaGitLab
        isActive={isTabActive('GitLab')}
        gotoParabol={goToParabol}
        meetingRef={meeting}
      />
  }
```
and then later iterate over `tabs` to insert the correct component? This way we cannot mess up the order

packages/server/graphql/public/typeDefs/gitlab.graphql
```diff
@@ -1 +1,7 @@
 type _xGitLabIssue implements TaskIntegration
+
+type _xGitLabProject implements RepoIntegration
+
+extend type _xGitLabProject {
+  service: IntegrationProviderServiceEnum!
```
Dschoordsch, line 6: -1 The resolver for this field is missing. You can test it by adding service to your request in `NewGitLabIssueInput`. Check https://github.com/ParabolInc/parabol/blob/b7e062ce11f6886042779753ee06d01cbd84c932/packages/server/graphql/public/types/_xGitHubIssue.ts for an example

Dschoordsch (changes requested):
Almost there!

Dschoordsch:
You should also delete `gitlabTypes.ts`.

packages/client/components/ScopePhaseArea.tsx
```diff
@@ -117,6 +117,44 @@ const ScopePhaseArea = (props: Props) => {
...
+      <ScopePhaseAreaJira
+        isActive={isTabActive('Jira')}
+        gotoParabol={goToParabol}
+        meeting={meeting}
+      />
+    ),
+    Parabol: <ScopePhaseAreaParabolScoping isActive={isTabActive('Parabol')} meeting={meeting} />
+  }
+
+  if (allowJiraServer) {
```
Dschoordsch, line 139: -1 no need to add this conditionally if we iterate over `tabs` instead below.

packages/client/components/ScopePhaseArea.tsx
```diff
@@ -117,6 +117,44 @@ const ScopePhaseArea = (props: Props) => {
...
+    contents['Jira Server'] = (
+      <ScopePhaseAreaJiraServer
+        isActive={isTabActive('Jira Server')}
+        gotoParabol={goToParabol}
+        meetingRef={meeting}
+      />
+    )
+  }
+  if (allowGitLab) {
```
Dschoordsch, line 148: -1 no need to add this conditionally

Dschoordsch (changes requested):

packages/client/components/GitLabScopingSearchResults.tsx
mattkrick: this is a pretty big issue, `allProjects` is actually resolving to `projects`. I try to explain why in the loom. It's 90 minutes, the summary starts at 1:25:00.

https://www.loom.com/share/79e4d705572c40bdba449b12bc3250b7

mattkrick: i baked this functionality into nest-graphql-endpoint in v0.6.0, i'll submit a PR later to bump versions so we never git bitten by this again

packages/client/components/NewGitLabIssueInput.tsx
```diff
@@ -0,0 +1,264 @@
...
+  const {onCompleted, onError} = useMutationProps()
+  const [selectedFullPath, setSelectedFullPath] = useState(gitlabProjects[0]?.fullPath || '')
+  const {fields, onChange, validateField, setDirtyField} = useForm({
+    newIssue: {
+      getDefault: () => '',
+      validate: validateIssue
+    }
+  })
+  const {originRef, menuPortal, menuProps, togglePortal, portalStatus} = useMenu(
+    MenuPosition.UPPER_RIGHT
```
mattkrick, line 169: +1 can you make this a dropdown? eg `useMenu(x, {isDropdown: true})` and update the github menu to use this, too.

dropdowns start with the full width, whereas menus grow from 0 width, nothing major, just looks nice for wider menus

packages/client/components/NewGitLabIssueMenu.tsx
```diff
@@ -0,0 +1,49 @@
...
+    gitlabProjects,
+    getValue
+  )
+
+  const onClick = (fullPath: string) => {
+    handleSelectFullPath(fullPath)
+  }
+
+  return (
+    <Menu ariaLabel='Select GitLab project' keepParentFocus {...menuProps}>
```
mattkrick, line 32: -1 this menu is waaaay too wide! the items are showing ellipsis, but the width is as big as can be.
the input field should dictate the width

Screen Shot 2022-04-06 at 3 08 55 PM

packages/client/components/NewGitLabIssueMenu.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+const NewGitLabIssueMenu = (props: Props) => {
+  const {handleSelectFullPath, menuProps, gitlabProjects} = props
+
+  const {query, filteredItems: filteredProjects, onQueryChange} = useSearchFilter(
+    gitlabProjects,
+    getValue
+  )
+
+  const onClick = (fullPath: string) => {
```
mattkrick, line 27: -1 this function doesn't add any value. consider using the pattern in NewGitHubIssueMenu
+1 anything named `onClick` should probably accept a mouse event (or nothing, if the event isn't used)

packages/client/components/ScopePhaseArea.tsx
```diff
@@ -117,6 +115,38 @@ const ScopePhaseArea = (props: Props) => {
     setActiveIdx(2)
   }
 
+  const contents: Record<typeof baseTabs[number]['label'], JSX.Element> = {
```
mattkrick, line 118: +2 i like the refactor!
+1 you could take this one step further by putting the component in the `tabs` array (just make Jira accept a `meetingRef` and ignore passing `gotoParabol` into the parabol tab)

```ts
 const baseTabs = [
    {icon: <GitHubSVG />, label: 'GitHub', allow: true, Component: ScopePhaseAreaGitHub}
]

tabs.map(({label, Component}) => <Component meeting={meeting} isActive={isTabActive(label)}, gotoParabol={gotoParabol}/>
```

packages/client/mutations/handlers/handleGitLabCreateIssue.ts
```diff
@@ -0,0 +1,55 @@
...
+  const project = createProxyRecord(store, '_xGitLabProject', {
+    fullPath
+  })
+  const issueEdge = createProxyRecord(store, '_xGitLabIssueEdge', {})
+  issueEdge.setLinkedRecord(integration, 'node')
+
+  const issueConn = createProxyRecord(store, '_xGitLabIssueConnection', {})
+  issueConn.setLinkedRecords([issueEdge], 'edges')
+
+  const issueArgs = {first: 25, includeSubepics: true, sort: 'UPDATED_DESC', state: 'opened'}
```
mattkrick, line 25: -1 single source of truth required here. if i change these vars in the gql request, this whole thing will break.

packages/client/types/constEnums.ts
mattkrick: to get around that error, try using space separated or comma + space separated values. (`read_api write_repository` or `read_api, write_repository`). OAuth2 spec has some discrepancies.

packages/server/graphql/mutations/helpers/createGitLabTask.ts
```diff
@@ -0,0 +1,49 @@
...
+  gitlabAuth: IGetTeamMemberIntegrationAuthQueryResult,
+  context: GQLContext,
+  info: GraphQLResolveInfo,
+  dataLoader: DataLoaderWorker
+) => {
+  const {accessToken, providerId} = gitlabAuth
+  if (!accessToken) return {error: new Error('Invalid GitLab auth')}
+  const {title, contentState} = splitDraftContent(rawContent)
+  const body = stateToMarkdown(contentState)
+  const provider = await dataLoader.get('integrationProviders').load(providerId)
```
mattkrick, line 22: -1 prefer `loadNonNull` when the return value is guaranteed so we don't need `!` down below

packages/server/integrations/gitlab/GitLabServerManager.ts
```diff
@@ -38,8 +38,8 @@ class GitLabServerManager {
   ): Promise<[boolean, Error | null]> {
     if (!this.accessToken) return [false, new Error('GitLabServerManager has no access token')]
 
-    const gitLabRequest = this.getGitLabRequest(info, context)
-    const [, error] = await gitLabRequest<GetProfileQuery>(getProfile, {})
+    const gitlabRequest = this.getGitLabRequest(info, context)
```
mattkrick, line 41: -1\ This is the perfect time to make this conform with the interface that @igorlesnenko built.

Can you say `class GitLabServerManager implements TaskIntegrationManager`?

It'll require you to add a `createTask` method. That method should then internally call `this.getGitLabRequest`.
That way you can just call `manager.createTask()` in createGitLabTask.

you can stub out the other method(s) that the interface requires

mattkrick (changes requested):
-1 great work! all pretty minor stuff

packages/client/components/NewGitLabIssueInput.tsx
nickoferrall: Ooh `isDropdown` looks good! I've added it to Jira as well as GH & GitLab

packages/client/mutations/handlers/handleGitLabCreateIssue.ts
nickoferrall: Makes sense. I'm now passing the variables into the `GitLabScopingSearchResultsRoot` query here: https://github.com/ParabolInc/parabol/blob/fed8dce61509d31113972d785915730d520224c0/packages/client/components/GitLabScopingSearchResultsRoot.tsx#L43

packages/client/types/constEnums.ts
nickoferrall: Space-separated works but even with `read_api write_repository write_directory`, it's not possible to create an issue as I get a permission error. So, it looks like `api` is actually the right scope for us

packages/client/components/GitLabScopingSearchResults.tsx
```diff
@@ -45,30 +26,15 @@ const GitLabScopingSearchResults = (props: Props) => {
...
-      fragment GitLabScopingSearchResults_query on Query
-        @argumentDefinitions(
-          projectsFirst: {type: "Int", defaultValue: 20}
-          issuesFirst: {type: "Int", defaultValue: 25}
-          projectsAfter: {type: "String"}
-          issuesAfter: {type: "String"}
-          search: {type: "String"}
-          projectIds: {type: "[ID!]", defaultValue: null}
-        )
-        @refetchable(queryName: "GitLabScopingSearchResultsPaginationQuery") {
```
nickoferrall, line 70: In v1, we're now querying a large number of projects and getting their issues. The user can't scroll down to load more issues. If the issue doesn't exist in the initial query, they can use the search query or filter menu. 

In v2, we'll refactor the nested connections so users can scroll down to load more issues: https://github.com/ParabolInc/parabol/issues/6325

For now, I want to update this query in this parent branch so all child branches follow the same pattern.

PR #6270, opened 2022-03-24
fix(sdl): Default resolveType on mutation payloads

mattkrick:
root cause fix #6268

This bug was caused because graphql couldn't figure out if a mutation returned an error or a success.

TEST
- [ ] run the following on the private schema. you'll see the error message

```gql
mutation {
  enableSAMLForDomain(name:"foo", domains:["foo.co"], metadata: "as") {
    __typename
    ... on ErrorPayload {
      error {
        message
      }
    }
  }
}
```

### DIFF
--- a/packages/server/graphql/private/rootSchema.ts
+++ b/packages/server/graphql/private/rootSchema.ts
@@ -5,19 +5,10 @@ import resolveTypesForMutationPayloads from '../resolveTypesForMutationPayloads'
 import publicSchema from '../rootSchema'
 import resolvers from './resolvers'
 
-// const publicSchema = new GraphQLSchema({
-//   query,
-//   mutation,
-//   types: rootTypes
-// })
-
 const typeDefs = loadFilesSync(
   path.join(__PROJECT_ROOT__, 'packages/server/graphql/private/typeDefs/*.graphql')
 )
 
 const privateSchema = resolveTypesForMutationPayloads(makeExecutableSchema({typeDefs, resolvers}))
-// const publicSchema = require('../rootSchema').default
-// console.log({publicSchema})
-// required to get the public types
 const fullSchema = mergeSchemas({schemas: [privateSchema, publicSchema]})
 export default fullSchema
--- a/packages/server/graphql/private/rootSchema.ts
+++ b/packages/server/graphql/private/rootSchema.ts
@@ -1,6 +1,7 @@
 import {loadFilesSync} from '@graphql-tools/load-files'
 import {makeExecutableSchema} from '@graphql-tools/schema'
 import path from 'path'
+import resolveTypesForMutationPayloads from '../resolveTypesForMutationPayloads'
 import resolvers from './resolvers'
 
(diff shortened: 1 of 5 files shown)


packages/server/graphql/resolveTypesForMutationPayloads.ts
```diff
@@ -0,0 +1,28 @@
...
+  This function finds all conventional MutationPayloads and
+  adds that `resolveType` if one is not already defined
+*/
+
+import {GraphQLSchema} from 'graphql'
+import {isUnionType} from 'graphql/type'
+
+const resolveTypesForMutationPayloads = (schema: GraphQLSchema) => {
+  Object.values(schema.getTypeMap()).forEach((type) => {
+    if (!isUnionType(type) || !type.name.endsWith('Payload')) return
```
Dschoordsch, line 16: -1 we should also return if `type.resolveType` is already present. It is defined in `makeMutationPayload` and thus does not need to be redefined.

https://github.com/ParabolInc/parabol/blob/fcfb8a714245224f1339027427b5e928cecceb69/packages/server/graphql/types/makeMutationPayload.ts#L9

Dschoordsch (changes requested):
Does not fix all issues, there is still a problem with "User" fields (see build results)

packages/server/graphql/resolveTypesForMutationPayloads.ts
```diff
@@ -0,0 +1,28 @@
...
+    const concreteTypes = type.getTypes()
+    const errorType = concreteTypes.find((type) => type.name === 'ErrorPayload')
+    const successName = `${type.name.slice(0, -'Payload'.length)}Success`
+    const successType = concreteTypes.find((type) => type.name === successName)
+    // Abort if the MutationPayload is not a default 2-part union of an ErrorPayload | Success
+    if (!errorType || !successType || concreteTypes.length !== 2 || type.resolveType) return
```
mattkrick, line 22: ah, i had it down here, but up there makes more sense

mattkrick:
@Dschoordsch ready for another look!

packages/server/graphql/private/rootSchema.ts
```diff
@@ -1,11 +1,14 @@
 import {loadFilesSync} from '@graphql-tools/load-files'
-import {makeExecutableSchema} from '@graphql-tools/schema'
+import {makeExecutableSchema, mergeSchemas} from '@graphql-tools/schema'
 import path from 'path'
+import resolveTypesForMutationPayloads from '../resolveTypesForMutationPayloads'
+import publicSchema from '../rootSchema'
```
mattkrick, line 5: The require chain looks like this:
- publicSchema
- updateHubspot
- executeGraphQL
- privateSchema/publicSchema

That circular reference means the private schema can't import the public schema.
To get around this, i used CJS require statements that are called inside function (vs. as imports at the top).
Once we move to an SDL for the public schema this'll resolve itself

PR #6282, opened 2022-03-25
feat: Display a message when there are no more items to paginate

### LINKED ISSUE #1450: Archive: that's all folks!
## Issue - Enhancement

When a user scrolls to the end of the archive, I want to see something that tells me there are no more items to paginate to, so I can be sure I'm done

### Acceptance Criteria
1. Create multiple project cards, enough so that when they are archived they will have to be paginated
2. Archive them all
3. Go to the archive, keep paginating until there are no more archive cards to display
4. Show a message at the bottom, _"That's all folks!_ There are no further Tasks in the archive."

- **Estimated effort:** 12 points ([see CONTRIBUTING.md](https://github.com/ParabolInc/action/blob/master/CONTRIBUTING.md#points-and-sizes))

shapel:
Hello
Below you can find my second attempt to fix #1450 :)

Tech aspects:
- there is no way to configure DOM representation in `react-virtualized` for such case
- the CSS pseudo-element is used as an attempt to implement the simplest way to achieve the required functionality without being deep coupling with implementation of external controls

UX aspects:
- the message should not be displayed if there is no scroll at all
- the message should only be displayed if the user has scrolled down and there are no more items to load

### DIFF
--- a/packages/client/modules/teamDashboard/components/TeamArchive/TeamArchive.tsx
+++ b/packages/client/modules/teamDashboard/components/TeamArchive/TeamArchive.tsx
@@ -194,8 +194,10 @@ const TeamArchive = (props: Props) => {
   }
   const isRowLoaded = ({index}) => index < edges.length
   const maybeLoadMore = () => {
-    if (!hasNext || isLoadingNext) return
-    loadNext(columnCount * 10)
+    if (!hasNext || isLoadingNext) return Promise.resolve()
+    return new Promise<void>((resolve, reject) => {
+      loadNext(columnCount * 10, {onComplete: (err) => (err ? reject(err) : resolve())})
+    })
   }
   const [cellCache] = useState(
     () =>
--- a/packages/client/modules/teamDashboard/components/TeamArchive/TeamArchive.tsx
+++ b/packages/client/modules/teamDashboard/components/TeamArchive/TeamArchive.tsx
@@ -246,7 +246,7 @@ const TeamArchive = (props: Props) => {
     oldEdgesRef.current = edges
   }, [edges, oldEdgesRef])
 
-  const rowRenderer = ({columnIndex, parent, rowIndex, key, style}) => {
+  const cellRenderer = ({columnIndex, parent, rowIndex, key, style}) => {
     // TODO render a very inexpensive lo-fi card while scrolling. We should reuse that cheap card for drags, too
     const index = getIndex(columnIndex, rowIndex)
     if (!isRowLoaded({index})) return undefined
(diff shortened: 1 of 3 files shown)


shapel:
@Dschoordsch it would be nice if you could review my new attempt ;)

Dschoordsch (changes requested):
Looks good. Can you add some margin at the bottom?

packages/client/modules/teamDashboard/components/TeamArchive/TeamArchive.tsx
```diff
@@ -278,79 +287,108 @@ const TeamArchive = (props: Props) => {
...
+                    >
+                      {({onRowsRendered, registerChild}) => {
+                        _onRowsRenderedRef.current = onRowsRendered
+                        return (
+                          <div style={{flex: '1 1 auto'}}>
+                            <AutoSizer>
+                              {({height, width}) => {
+                                return (
+                                  <Grid
+                                    className={!hasNext && isScrolledToBottom ? noMoreMessage : ''}
```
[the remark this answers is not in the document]
shapel: I believe this is redundant checking.
Because there are indicators of a different kinds (server and client). `hasNext` comes from the server's response (connection data), which indicates that there is no more element from the server's point of view.
`isLoadingNext` refers to the status of a request on the client (that we are waiting for the answer to the next piece of data).

After the server responds that there are no more items to load, the client cannot make a pagination request for the next items.
So when `hasNext` is false, `isLoadingNext` is always false too.

packages/client/modules/teamDashboard/components/TeamArchive/TeamArchive.tsx
```diff
@@ -278,79 +303,98 @@ const TeamArchive = (props: Props) => {
...
     _onRowsRenderedRef.current({
       startIndex: getIndex(columnStartIndex, rowStartIndex),
       stopIndex: getIndex(columnStopIndex, rowStopIndex)
     })
   }
 
+  const [isScrolledToBottom, setIsScrolledToBottom] = useState(false)
+  const onScroll = ({clientHeight, scrollHeight, scrollTop}: ScrollParams) => {
+    if (!isScrolledToBottom && scrollTop + clientHeight >= scrollHeight) {
+      setIsScrolledToBottom(true)
```
Dschoordsch, line 317: -1 I don't think this is the best solution. It also results in the scrollbar changing once the end is reached. All we want to know is whether or not the `ReactVirtualized__Grid__innerScrollContainer` is higher than the `ReactVirtualized__Grid`. Since we also only want to show it once the full list was loaded, I think it's best to:
* add 1 to the rowCount
* return "That's all folks" component in `cellRenderer`
  ```
    if (!hasNext && rowCount === rowIndex) {
      return columnIndex === 0 && <CellMeasurer >... That's all folks ...
  ```
  with `gridColumn: 1 / span ${columnCount}`

shapel: You are right! I've changed the approach in #6338

shapel:
close it in favour of https://github.com/ParabolInc/parabol/pull/6338

PR #6299, opened 2022-03-28
fix: button hover mismatched styling on avatar input

### LINKED ISSUE #6169: Button hover styling mismatched on the /me/profile page
## Issue - Bug

- **OS:** macOS
- **Browser:** Chrome

https://user-images.githubusercontent.com/4997466/156768067-8cf7c549-0a70-412c-9bd6-ec667e089437.mp4

### Acceptance Criteria
Users can:
   - see the full button shadow effect with the correct padding when hovering (`/me/profile` page)
   - see UI consistent with other buttons on all pages

Root cause: The current button on the profile page does not have enough padding when hovering, so the shadow effect is obscured.

correct button padding when hover

**Estimated effort:** <1 hour

mikennaji:
closes #6169


Dschoordsch (changes requested):

Dschoordsch:
@mikennaji I think you made a mistake with this changeset. When you do a `git revert` it is not undoing anything, but rather committing a new change which does the reverse of the change you're reverting. This PR now has the revert of a merge to master in it and cannot be merged like this.
You can fix this by doing an interactive rebase, rebasing the last change on top of the first change you've done, discarding the merge and revert of it. Just do `git rebase -i 8d11924` and then delete the two lines relating to the merge (instructions are also in the editor that opens).

PR #6300, opened 2022-03-28
feat(impersonate): keep impersonate token until last tab closes

### LINKED ISSUE #6106: Impersonate should work for opened tabs, too
We shouldn't delete the impersonated tokens if parabol is opened in a popup. That makes looking at invoices impossible.

AC
- [ ] In https://github.com/ParabolInc/parabol/blob/12b990356cb64860ab425f178d18cf1073816972/packages/client/Atmosphere.ts#L334 the business logic includes checking to see if the current window has an opener. if it does, then don't log them out

Estimate: 3 hours

mattkrick:
fix #6106

TEST
- [x] impersonate a pro user
- [x] goto their org, click the top invoice
- [x] you can view the invoice as an impersonated user
- [x] refresh the invoice tab, you can still see it
- [x] close the invoice tab & refresh the main tab. you are logged out


packages/client/Atmosphere.ts
```diff
@@ -314,7 +316,35 @@ export default class Atmosphere extends Environment {
     this.setAuthToken(authToken)
   }
 
-  setAuthToken = (authToken: string | null) => {
+  private validateImpersonation = async (iat: number) => {
+    const isAnotherParabolTabOpen = async () => {
+      const askOtherTabs = new Promise((resolve) => {
+        if (!this.tabCheckChannel) {
+          this.tabCheckChannel = new BroadcastChannel('tabCheck')
```
mattkrick, line 323: this is the first time we've used a broadcast channel in this app, so it deserves some explanation.

When we open up a new tab, that tab doesn't have an `opener` so we can't see if we opened it or who owns it.
we could do things like attach metadata to that new tab, but that means whenever we open a new tab we'd have to remember to attach that metadata.

introduce broadcast channels-- a way for tabs to talk to one another.
if an impersonate token is detected, it opens a channel.
that means when the e.g. invoice tab gets opened, it pings all the other parabol tabs, sees that 1 is open, so it doesn't log the user out.

tianrunhe (changes requested):
After closing the invoice tab, the first refresh on the main tab does not automatically log out the user. Second fresh does though. See [video](https://www.loom.com/share/6e590b9d32e445499d44b5fefe40e9eb).

Issue #5383, opened 2021-09-08
payload can be null in ResetRetroMeetingToGroupStageMutation


nickoferrall:
Affected 9 users in the past 16 days since v6.24.1

https://sentry.io/organizations/parabol/issues/2607259830/?referrer=github_plugin

```
TypeError: Cannot read properties of null (reading 'getLinkedRecords')
  at resetRetroMeetingToGroupStageUpdater (./packages/client/mutations/ResetRetroMeetingToGroupStageMutation.ts:52:43)
  at updater (./packages/client/mutations/ResetRetroMeetingToGroupStageMutation.ts:73:7)
  at Set.forEach (<anonymous>)
  at updateIsEditing (./packages/client/components/ReflectionCard/ReflectionCard.tsx:110:5)
  at g (./packages/client/components/ReflectionCard/ReflectionCard.tsx:121:18)
...
(15 additional frame(s) were not displayed)
```

shapel:
@nickoferrall do you have a chance to review this?

PR #6306, opened 2022-03-30
fix: add agenda item response handling

shapel:
Fixes how mutation response handling by adding types and fixing possible TypeError (similar to #5383).


shapel:
👋 @mattkrick do you have a chance to review or assign someone to do this?

mattkrick:
thanks @shapel! 
Do you have some steps I could follow to reproduce the bug?

shapel:
@mattkrick I don't have steps to reproduce. 
I saw that piece of code with potential problem like https://github.com/ParabolInc/parabol/issues/5383 and decided to fix it too. 

PS May be you can to review fix for the issue I mention too?

mattkrick:
LGTM! 🙏

Issue #6196, opened 2022-03-15
Standup UI - Top Bar - Empty component stub


BartoszJarocki:
Design: https://www.figma.com/file/bhiEOV4HK9Kv91HFsJmevV/Stand-up-Concepts?node-id=1138%3A72617 

AC
- [ ] Empty component is added to the top of the Standups view

PR #6307, opened 2022-03-30
feat(standups): Top Bar Component Stub

jmtaber129:
Fixes https://github.com/ParabolInc/parabol/issues/6196

Create the stub for the top bar, and wrap the existing meeting space placeholder

### DIFF
--- a/packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
+++ b/packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
@@ -1,13 +1,20 @@
+import styled from '@emotion/styled'
 import React from 'react'
 import {HeadingBlock, MeetingTopBarStyles} from '../MeetingTopBar'
-import PhaseHeaderTitle from '../PhaseHeaderTitle'
+
+const TeamPromptHeaderTitle = styled('h1')({
+  fontSize: 16,
+  lineHeight: '24px',
+  margin: 0,
+  padding: 0
+})
 
 const TeamPromptTopBar = () => {
   return (
     <MeetingTopBarStyles>
-      <HeadingBlock isMeetingSidebarCollapsed={true}>
+      <HeadingBlock>
         {/* :TODO: (jmtaber129): Add back button */}
-        <PhaseHeaderTitle>Hard-coded standup title</PhaseHeaderTitle>
+        <TeamPromptHeaderTitle>Hard-coded standup title</TeamPromptHeaderTitle>
         {/* :TODO: (jmtaber129): Add avatars, overflow menu, etc. */}
       </HeadingBlock>
     </MeetingTopBarStyles>
--- a/packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
+++ b/packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
@@ -1,0 +1,17 @@
+import React from 'react'
+import {HeadingBlock, MeetingTopBarStyles} from '../MeetingTopBar'
+import PhaseHeaderTitle from '../PhaseHeaderTitle'
+
+const TeamPromptTopBar = () => {
+  return (
+    <MeetingTopBarStyles>
+      <HeadingBlock isMeetingSidebarCollapsed={true}>
+        {/* :TODO: (jmtaber129): Add back button */}
+        <PhaseHeaderTitle>Hard-coded standup title</PhaseHeaderTitle>
+        {/* :TODO: (jmtaber129): Add avatars, overflow menu, etc. */}
+      </HeadingBlock>
+    </MeetingTopBarStyles>
+  )
+}
+
+export default TeamPromptTopBar
--- a/packages/client/components/TeamPromptMeeting.tsx
+++ b/packages/client/components/TeamPromptMeeting.tsx
@@ -1,15 +1,14 @@
-import styled from '@emotion/styled'
-import React from 'react'
+import React, {Suspense} from 'react'
 import {Link} from 'react-router-dom'
 
-const Root = styled('div')({
-  display: 'flex',
-  alignItems: 'center',
-  justifyContent: 'center',
-  height: '100%'
-})
-
 import logoMarkPurple from '../styles/theme/images/brand/mark-color.svg'
+import MeetingArea from './MeetingArea'
+import MeetingContent from './MeetingContent'
+import MeetingHeaderAndPhase from './MeetingHeaderAndPhase'
+import MeetingStyles from './MeetingStyles'
+import MeetingTopBar from './MeetingTopBar'
+import PhaseHeaderTitle from './PhaseHeaderTitle'
+import PhaseWrapper from './PhaseWrapper'
 
 interface Props {
   meeting: any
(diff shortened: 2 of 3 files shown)


packages/client/components/TeamPromptMeeting.tsx
```diff
@@ -1,27 +1,43 @@
...
-      <Link title='My Dashboard' to='/meetings'>
-        <img alt='Parabol' src={logoMarkPurple} />
-      </Link>
-    </Root>
+    <MeetingStyles>
+      <MeetingArea>
+        <Suspense fallback={''}>
+          <MeetingContent>
+            <MeetingHeaderAndPhase hideBottomBar={true}>
+              <MeetingTopBar
```
BartoszJarocki, line 24: I'd avoid using this component here. Standup meeting has totally different styles and I'm not sure we're gaining anything by reusing it here. It's safer to create a `TeamPromptMeetingTopBar` and just make everything match the design. 

Things that are different
1. Back button (non existing in current meeting view)
2. Title (different font size)
3. Avatars group (smaller)
4. Facilitator button (not needed for now, but not present in current meetings)
5. Invite button 
6. More button

that's why I'd just create a new component. as we're here we can try to group components related to this type of meeting in a single directory (so it's easier to see what belongs where). I'd create `components/TeamPrompt/` and put everything related in there.

packages/client/components/TeamPromptMeeting.tsx
```diff
@@ -1,27 +1,43 @@
...
+                avatarGroup={/*:TODO: (jmtaber129): Add avatar group*/<></>}
+                isMeetingSidebarCollapsed={true}
+              >
+                {/* :TODO: (jmtaber129): Add back button */}
+                <PhaseHeaderTitle>Hard-coded standup title</PhaseHeaderTitle>
```
BartoszJarocki, line 29: I'm not sure if it's the final version of a title, but this component has different styles than what we in the designs for team prompt so I'd just create a new one.

BartoszJarocki (changes requested):

packages/client/components/TeamPromptMeeting.tsx
jmtaber129: Could you point out the style differences for me?  I'm looking through the code + the figma docs, and I'm not seeing any real differences.  Though this specific component's implementation doesn't match the styles as implemented in Figma, it looks like this should still match the styles in the context of other components (e.g. the `MeetingTopBar` has styles that add 16px to the left, etc.)

packages/client/components/TeamPromptMeeting.tsx
BartoszJarocki: that’s a valid tension! in general, we use this rule to decide whether we want to abstract/reuse something: https://www.notion.so/parabol/How-to-Do-a-Code-Review-daaf03fae3484cf5a1e18174170dadff#598e7a6900f5430ab69f85d88c2f055f 

regarding this particular case, here is what worries me

1. we’ll be refactoring a component used in production by all other meetings, to make it fit the designs of a new, experimental meeting which is under heavy development/experimentation mode. So basically, I’m worried about creating a component with bad interface which in long term will cost us more than creating a new one, for now. here’s a good read about that: https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction 
2. `MeetingTopBar` already has a lot of stuff this meeting currently won’t need, like `toggleDrawer`, `toggleSideBar`, `isCommentUnread`, `isMeetingSidebarCollapsed` and list goes on. 

that said, here’s what I’d do. I'd ask a different question. is there a better way of creating a top bar, in more modular, composable way? if yes, let’s do it and create tasks to refactor existing meetings to use the new component. 

for example, new top bar needs navigation and overflow menu which are not present in other meetings. maybe it’d be good to create a separate component, called `NavigationBar` which would only have the navigation and overflow menu. then, we could use it that way:
```
<NavigationBar>
	<MeetingTopBar>
		…
	</MeetingTopBar>
</NavigationBar>
```
this is just an idea, there might be a better way to do this. 

TLDR, instead of changing the existing component, I’d create a better one, more composable and create tasks to reuse it in other meetings. what do you think @jmtaber129?

packages/client/components/TeamPromptMeeting.tsx
BartoszJarocki: sure, sorry I should have been more specific. 
in figma, on desktop version the title has size of `16px` while `PhaseHeaderTitle` uses `20px` 

```
const PhaseHeaderTitle = styled('h1')({
  fontSize: 16,
  lineHeight: '24px',
  margin: 0,
  padding: 0,
  [meetingTopBarMediaQuery]: {
    fontSize: 20
  }
})
```

not a big deal, but there's a difference 👨‍🎨 😅

jmtaber129: Ah, it looks like some of the Figma mockups have different font sizes - "Standup MVP, first layout" has font size 20px, while "Initial state, no responses yet" has font size 16px.

I've been under the impression that the intention behind the designs is to make this consistent with existing meeting phase styling, but I don't know for sure.

Lmk what you think, happy to create a new component if that's the way to go 🙂

BartoszJarocki: oh, didn't notice designs have different font size, good catch! 👏 
@enriquesanchez @ackernaut what size is correct? 🙏

packages/client/components/TeamPromptMeeting.tsx
BartoszJarocki: great! would you like to play with it in this task? once we have a we'll create tasks and handle other meetings separately.

jmtaber129: Yea, I'll create this new component as part of this task

jmtaber129: After thinking about this a bit more, I ended up trying out a slightly different approach.  Specifically:
* Export styled components from `MeetingTopBar`
* Use these components to compose a meeting-/phase-specific top bar.  E.g. a `RetroTopBar` for retros that includes the sidebar and `DemoCreateAccountButton`, a `SidebarMeetingTopBar` for any meetings with sidebars that don't need additional customization, etc.

The thinking here is that the core functionality of the current `MeetingTopBar` is pretty basic - render some children (maybe with a sidebar toggle) and render the avatars that were passed in.  However, the scope of the component as an abstraction has increased quite a bit - maybe show a sidebar toggle, show a `DemoCreateAccountButton` for retro demos, show a discussion toggle drawer for sprint poker estimation.  Refactoring other meetings to use this approach would maintain consistent and reusable styles and allow reuse of most existing logic, while also allowing meeting-/phase-level customization.

If this approach were adopted for the other meetings, the refactor would look something like:
* Create a `SidebarHeadingBlock` for any top bars that need sidebar toggles
* Create a `SidebarMeetingTopBar` component that includes the sidebar toggle, and use that in most phases
* Create a `PokerEstimatePhaseTopBar` component that includes the sidebar toggle and discussion drawer toggle, and use that in the poker estimation phase.
* Create a `RetroTopBar` component that includes the sidebar toggle and `DemoCreateAccountButton`, and use that in retro phases.
* Optionally split the existing `MeetingTopBar.tsx` into a few files within a `MeetingTopBar/` directory.

LMK what you think.

(I'll also refactor the `HeadingBlock` to not need the `isMeetingSidebarCollapsed` on the `TeamPromptTopBar` if this sounds good)

packages/client/components/TeamPromptMeeting.tsx
enriquesanchez: @BartoszJarocki @jmtaber129 Thanks for the ping! It should be 16px 👍

BartoszJarocki (approved):
looks good to me, nice work!

packages/client/components/TeamPromptMeeting.tsx
BartoszJarocki: sounds really good! having multiple building blocks and then being able to construct specific components for each meeting type is the way to go. as you're in the middle of all of these, do you mind creating refactoring tasks with proper descriptions for other meetings? great job!

packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
```diff
@@ -0,0 +1,24 @@
...
+  lineHeight: '24px',
+  margin: 0,
+  padding: 0
+})
+
+const TeamPromptTopBar = () => {
+  return (
+    <MeetingTopBarStyles>
+      <HeadingBlock>
+        {/* :TODO: (jmtaber129): Add back button */}
```
mattkrick, line 16: do you use gitlens in vscode? 
it'll show the blame inline so you don't have to add your name if you don't want to. even lets you hover for extra details. no harm either way
Screen Shot 2022-04-06 at 4 58 11 PM

mattkrick (approved):
love these bite sized PRs!

PR #6308, opened 2022-03-30
chore(dx): Stop writes to rethinkDB Team table

Dschoordsch:
Follow up to
* #6239

Added migration to check and then drop the table.


Dschoordsch:
#6239 is more up to date

Issue #5952, opened 2022-01-25
Eslint fix: glob pattern in quotes for server


mattkrick:
Today, we call eslint & use a glob pattern without quotes.
We either need to pass in a directory (without the glob) OR put the glob in quotes: https://stackoverflow.com/questions/37927772/how-to-silence-warnings-about-ignored-files-in-eslint/66434753#66434753

By doing it how we're doing it, we can't ignore files without a warning, which is painful.
Changing this will also catch a BUNCH of brand new errors. We should do this for the server first, then the client if we so choose.

AC
- [ ] `yarn lint` on the server runs eslint on the server directory, not a glob
- [ ] All new errors caught are fixed

Estimate: 4 hours

mattkrick:
closed in favor of #6314

PR #6313, opened 2022-03-30
fix(lint): faster, deterministic linting

### LINKED ISSUE #6273: fix(prettier): decouple prettier from eslint
When writing code, some things like line length are inconsistent & in vscode, something eslint hangs & takes forever.

I think we should probably follow prettier's guidance: https://prettier.io/docs/en/integrating-with-linters.html

AC
- [ ] our linting & styling does what prettier suggests
- [ ] formatting on save results in the same code as the lint-staged precommit hook

Estimate: 4 hours

mattkrick:
Fix #6273

Today, our linting is slow. Different developers have different settings & PRs often have line diffs that aren't meaningful (e.g. reorder order of imports).

### Goals
- [ ] first, try to format on save (using vscode)
- [ ] if that is disabled or vscode isn't used, it formats on commit (as long as `git commit -n` isn't used)
- [ ] allows you to format manually by running `yarn format` in the server or client package
- [ ] as a last resort, a github action will check for lint & the PR won't be mergable until that test passes (out of scope for this PR)

### What this PR does:
- upgrades all our linting software
- decouples prettier & eslint. this makes it faster than before & means eslint doesn't try to do prettier's job.
- fixes the glob patterns in the CLIs (#5952)
- makes prettier share eslint's ignore file fo reduced complexity
- adds a prettier plugin that organizes imports using the same logic as vscode 

### What this PR does NOT do:
- fixes existing errors. there are about 500 files that will be changed. we can do that in a subsequent PR (#6314)
- adds a github action to fail the PR merge if the format is incorrect

## TEST

PRE-COMMIT HOOK
- [x] using vscode, disable the prettier plugin & turn off formatting on save. Open Action.tsx & hit save. nothing should happen
- [x] in the client package, add `Action.tsx` to `.eslintignore`. Then add a random `42` to an empty line
- [x] stage Action.tsx & then run `yarn precommit` in the client package. you should a few import statements changed
- [x] undo the changes to Action.tsx & to your vscode settings

VSCODE FORMAT ON SAVE
- [x] open up Action.tsx & hit save. the result should be the same as what the precommit hook did


.vscode/settings.json
```diff
@@ -26,6 +26,10 @@
     }
   ],
   "npm.packageManager": "yarn",
+  "editor.codeActionsOnSave": {
+    "source.organizeImports": false,
```
mattkrick, line 30: we have a prettier plugin that does this for us, so organizeImports is no longer necessary

packages/client/lint-staged.config.js
```diff
@@ -1,4 +1,5 @@
 module.exports = {
-  'package/client/**/!(*graphql).{js,ts,tsx}': 'eslint --fix',
+  '*.{ts,tsx}': ['eslint --fix', 'prettier --config ../../.prettierrc --ignore-path ./.eslintignore --write'],
```
mattkrick, line 2: removed the prettier rules from eslint & just calls prettier after eslint.

packages/client/package.json
```diff
@@ -13,10 +13,12 @@
     "url": "https://github.com/ParabolInc/parabol/issues"
   },
   "scripts": {
+    "format": "yarn typecheck && yarn lint && yarn prettier",
     "storybook": "start-storybook -p 6006 -c .storybook",
     "build-storybook": "build-sFtorybook",
-    "lint": "eslint --fix ./**/*.{ts,tsx}",
+    "lint": "eslint --fix \"**/*.{ts,tsx}\"",
```
mattkrick, line 19: fixed glob pattern. now we have like 600 errors that we'll want to fix up at a later time

packages/client/package.json
```diff
@@ -37,15 +39,14 @@
     "@types/stripe-v2": "^2.0.1",
     "babel-plugin-relay": "^12.0.0",
     "eslint": "^8.2.0",
-    "eslint-config-prettier": "^6.15.0",
+    "eslint-config-prettier": "^8.5.0",
     "eslint-plugin-emotion": "^10.0.14",
-    "eslint-plugin-prettier": "^4.0.0",
```
mattkrick, line 42: this plugin converted all of prettiers rules to slower eslint rules. it's faster to run a bare minimum of eslint rules & then run prettier after that

tianrunhe (approved):

PR #6317, opened 2022-03-30
feat(scale): give a pubsub channel to each socket server

mattkrick:
before, a graphql executor would send the response payload to a fixed server (because the channel name was static).
now, it sends it to the server that requested it.

I also renamed `serverChannel` to `executorChannel` since that could get ambiguous.

### DIFF
--- a/packages/server/utils/PubSubPromise.ts
+++ b/packages/server/utils/PubSubPromise.ts
@@ -1,6 +1,6 @@
 import Redis from 'ioredis'
 import ms from 'ms'
-import GQLExecutorId from '../../client/shared/gqlIds/GQLExecutorId'
+import GQLExecutorChannelId from '../../client/shared/gqlIds/GQLExecutorChannelId'
 import numToBase64 from './numToBase64'
 import sendToSentry from './sendToSentry'
 
@@ -64,7 +64,7 @@ export default class PubSubPromise<Request extends BaseRequest, Response> {
       const {executorServerId, ...rest} = request
       const message = JSON.stringify({jobId, socketServerId: SERVER_ID, request: rest})
       if (executorServerId) {
-        const executorChannel = GQLExecutorId.join(executorServerId)
+        const executorChannel = GQLExecutorChannelId.join(executorServerId)
         this.publisher.publish(executorChannel, message)
       } else {
         // cap the stream to slightly more than 1000 entries.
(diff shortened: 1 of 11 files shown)


packages/server/utils/PubSubPromise.ts
```diff
@@ -60,10 +61,11 @@ export default class PubSubPromise<Request extends BaseRequest, Response> {
...
       }
       this.jobs[jobId] = {resolve, timeoutId}
-      const {serverChannel, ...rest} = request
-      const message = JSON.stringify({jobId, request: rest})
-      if (serverChannel) {
-        this.publisher.publish(serverChannel, message)
+      const {executorServerId, ...rest} = request
+      const message = JSON.stringify({jobId, socketServerId: SERVER_ID, request: rest})
+      if (executorServerId) {
+        const executorChannel = GQLExecutorId.join(executorServerId)
```
Dschoordsch, line 67: -1 the result of `...Id.join(...)` should be an id, now it's the opposite, so it should technically be `split(...)`

packages/server/utils/getGraphQLExecutor.ts
```diff
@@ -1,14 +1,15 @@
...
 import PubSubPromise from './PubSubPromise'
 
-let pubsub: PubSubPromise<GQLRequest & {serverChannel?: string}, ExecutionResult>
+let pubsub: PubSubPromise<GQLRequest & {executorServerId?: string}, ExecutionResult>
 const getGraphQLExecutor = () => {
   if (!pubsub) {
     pubsub = new PubSubPromise(
       ServerChannel.GQL_EXECUTOR_STREAM,
-      ServerChannel.GQL_EXECUTOR_RESPONSE
+      SocketServerId.join(process.env.SERVER_ID!)
```
Dschoordsch, line 12: -1 Same as above, the result of this is a channel, not an id, so it should be either `...Id.split(...)` or `...Channel.join(...)`

Dschoordsch (changes requested):

packages/server/utils/PubSubPromise.ts
Dschoordsch: My issue is that we're passing around ids and then we use them by calling `GQLExecutorId.join(id)` to get a channel name. I think safest would be to call it `GQLExecutorChannel.join(id)` or if you like the `Id` suffix at least call it `GQLExecutorChannelId`. Naming is hard.

mattkrick: oh! that makes sense

PR #6318, opened 2022-03-30
feat(standups): Top Bar Back Button

jmtaber129:


BartoszJarocki (approved):
lgtm!

packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  const { name: meetingName } = meeting
+
+  return (
+    <MeetingTopBarStyles>
+      <HeadingBlock>
+        <TeamPromptHeader>
+          <BackButton
+            ariaLabel='Back to Meetings'
+            onClick={() => history.push('/meetings')}
```
mattkrick, line 50: -1 instead of an `onClick`, can you make `BackButton` extend a `Link` component?

e.g. 
```ts
import {Link} from 'react-router-dom'
const BackButton = styled(Link)({})
```

Reason:
- the `Link` Component uses an anchor tag
- If you right click an anchor tag you'll see "open link in new tab" et al
- that's really nice for a11y, but also in case i want to check out my team dash without leaving the meeting

i just learned this little trick like 2 months ago when some a11y folks told me i need to do better, so the codebase is probably filled with cases where it's not done correctly. if you find other places, please don't hesitate to fix em!

mattkrick (changes requested):
-1 just some merge conflicts & a little a11y thing. after that feel free to merge

packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
jmtaber129: Nice tip!  The `BackButton` already internally uses a different component, so I wrapped the existing implementation of `BackButton` with a `Link` and changed it to take a `to` prop passed to `Link` instead of `onChange` passed to `IconButton`

jmtaber129:
Fixed a11y thing + merge conflicts, so merging

Issue #6321, opened 2022-03-31
Inline fragments on field alias influencing each other


Dschoordsch:
There seems to be a bug where both fragments are messed up along the way. If you add additional fields to `NewGitLabIssueInput` `allProjects` it will be returned here as well. The request side looks fine and also the fields requested from GitLab.

_Originally posted by @Dschoordsch in https://github.com/ParabolInc/parabol/pull/6267#discussion_r839396103_

Dschoordsch:
Should be solved by #6361

PR #6323, opened 2022-03-31
feat(sprint-poker): Filter GitLab issues by project

nickoferrall:
Fix: https://github.com/ParabolInc/parabol/issues/6046

Note: there is a bug as mentioned in the [comment below](https://github.com/ParabolInc/parabol/pull/6323#issuecomment-1084813582), but the PR is accomplishing the AC of being able to filter issues by projects, so I think it's safe to move this to Reviewer Review.


nickoferrall:
Hey @mattkrick, the problem we discussed yesterday during our 1-1 is still causing trouble. After selecting one of the project filters in [GitLabScopingSearchFilterMenu](https://github.com/ParabolInc/parabol/blob/87f71fdc141054113692ceb1d83b2ebd0edc1878/packages/client/components/GitLabScopingSearchFilterMenu.tsx), the [GitLabScopingSearchResults](https://github.com/ParabolInc/parabol/blob/2cf7ddd8e84cc75aa82ec1ee6ed7fc755fb68009/packages/client/components/GitLabScopingSearchResults.tsx) query successfully filters by the selected project. However, the GitLabScopingSearchFilterMenu query also filters even though no ids have been passed as an arg. 

projects-filter-bug

In the gif above, I'm using a `network-only` fetch policy. If I use a `store-or-network` policy, unselecting the selected project filter does not result in all menu options returning to the menu, i.e. there's still only one project in the menu.

I suspect the GitLab queries are being tangled up when they're stitched together. Georg came across a similar problem when reviewing one of my other PRs: https://github.com/ParabolInc/parabol/issues/6321

packages/client/components/GitLabScopingSearchResults.tsx
```diff
@@ -126,7 +129,6 @@ const GitLabScopingSearchResults = (props: Props) => {
     query
   )
 
-  const lastItem = useLoadNextOnScrollBottom(paginationRes, {}, 20)
```
nickoferrall, line 129: Scrolling to the bottom of the page when the `projects` query contains a `projectsIds` causes an infinite loop. I spent time digging into this, but as we're refactoring this query in https://github.com/ParabolInc/parabol/issues/6325 and as the user can still find the desired issue using the search & filter menu, I think it's safe to remove for now.

packages/client/components/GitLabScopingSearchFilterMenu.tsx
```diff
@@ -0,0 +1,160 @@
...
+            gitlabSearchQuery.setValue(newSelectedProjectsIds, 'selectedProjectsIds')
+          })
+        }
+        return (
+          <MenuItem
+            key={projectId}
+            label={
+              <StyledMenuItemLabel>
+                <StyledCheckBox active={isSelected} />
+                <TypeAheadLabel query={searchQuery} label={fullPath} />
```
nickoferrall, line 149: In GitLab, a project can have subprojects. In the Parabol1 repo, we have several projects that contain several subprojects. So, if we show the `name` of the project in the filter menu, there can be several projects that have the same name. In Parabol1, we have many subprojects called `ppmi`, for example. To differentiate between projects, I think it's best to use `fullPath` instead of `name` in the menu.

mattkrick:
to fix the search thing, just use `search: ""` on the search filter args list to tell relay it's a totally different query.
loom:  https://www.loom.com/share/33c450c0c9394aab8765f6d0ce009636

packages/client/components/GitLabScopingSearchResults.tsx
```diff
@@ -53,7 +56,7 @@ const GitLabScopingSearchResults = (props: Props) => {
     graphql`
       fragment GitLabScopingSearchResults_query on Query
         @argumentDefinitions(
-          projectsFirst: {type: "Int", defaultValue: 20}
+          projectsFirst: {type: "Int", defaultValue: 75}
```
igorlesnenko, line 59: Is it safe to increase to that number?

igorlesnenko (commented):
Looks good to me. Just a couple of things.

* We removed [gitlabTypes.ts](https://github.com/ParabolInc/parabol/pull/6323/files#diff-7476a08a2a1db903c3e56c7cf900db00262edf93155e304a899aa054c3c59159) from Git but here we seem to be bringing it back
* While testing various GitLab issues, I noticed that sometimes I need to reconnect GitLab. Usually, I open a list of issues and see it is empty but no error is displayed. Reconnecting GitLab fixes the issue for me. Feels like at some point GitLab token expires and we do not handle it. Is it a known issue?

nickoferrall:
Great, thanks for the review. I've removed the GitLab types. I've also noticed GitLab needing to be reintegrated occasionally so I've created an issue to dig into this: https://github.com/ParabolInc/parabol/issues/6343

packages/client/components/GitLabScopingSearchResults.tsx
igorlesnenko: Yeah, I just remember we had this issue. If we have refactor issue, I think it is fine

packages/client/components/GitHubScopingSearchFilterMenu.tsx
```diff
@@ -26,6 +26,10 @@ const StyledCheckBox = styled(Checkbox)({
 })
 const StyledMenuItemLabel = styled(MenuItemLabel)({})
 
+const StyledMenu = styled(Menu)({
+  maxWidth: '100%'
```
mattkrick, line 30: +1 i tried commenting this out & it looked the same. any steps to reproduce the problem?

packages/client/components/GitLabScopingSearchResultsRoot.tsx
```diff
@@ -21,17 +21,26 @@ const GitLabScopingSearchResultsRoot = (props: Props) => {
...
       }
     `,
     meetingRef
   )
   const {teamId, gitlabSearchQuery} = meeting
-  const normalizedQueryString = gitlabSearchQuery.queryString.trim() || ''
+  const {queryString, selectedProjectsIds} = gitlabSearchQuery!
+  const normalizedQueryString = queryString.trim()
+  const normalizedSelectedProjectsIds = selectedProjectsIds?.length
+    ? (selectedProjectsIds as string[])
```
mattkrick, line 35: +1 can remove casting, the end result can still be null

packages/client/components/ScopePhaseArea.tsx
```diff
@@ -70,8 +70,7 @@ const ScopePhaseArea = (props: Props) => {
   const {meeting} = props
   const isDesktop = useBreakpoint(Breakpoint.SIDEBAR_LEFT)
   const {viewerMeetingMember} = meeting
-  if (!viewerMeetingMember) return null
-  const {user, teamMember} = viewerMeetingMember
+  const {user, teamMember} = viewerMeetingMember!
```
mattkrick, line 73: -2 what if a user visits a meeting that they didn't join while it was active?
if this is about a lint rule you can just add a comment to ignore that rule for this file

packages/server/graphql/types/GitLabSearchQuery.ts
```diff
@@ -14,6 +14,12 @@ const GitLabSearchQuery = new GraphQLObjectType<any, GQLContext>({
       type: new GraphQLNonNull(GraphQLString),
       description: 'The query string used to search GitLab issue titles and descriptions'
     },
+    selectedProjectsIds: {
+      type: new GraphQLList(new GraphQLNonNull(GraphQLID)),
+      description:
+        'The list of ids of projects that have been selected as a filter. Null if none have been selected',
+      resolve: ({selectedProjectsIds}) => selectedProjectsIds
```
mattkrick, line 21: -1 can resolve this resolve fn, that's what it'll do anyways

mattkrick (changes requested):
-2 just 1 fatal edge case, everything else is pretty minor. great job!

packages/client/components/GitLabScopingSearchResultsRoot.tsx
nickoferrall: I think we have to keep the casting here because if we remove it, TS complains that `normalizedSelectedProjectsIds` can be readonly

packages/client/components/ScopePhaseArea.tsx
nickoferrall: Yep, I was just trying to remove the lint rule. Change reverted 👍

packages/client/components/GitHubScopingSearchFilterMenu.tsx
nickoferrall: If a project fullPath is particularly long, it gets truncated without setting `maxWidth` to 100%. 
Without `maxWidth: '100%'`:
before

With `maxWidth: '100%'`:
after

mattkrick:
LGTM!

Issue #6327, opened 2022-03-31
feat(sdl): fix args implicit any in sdl resolvers


mattkrick:
For a reason I'm not 100% sure of, when `files: {}` is included in the tsconfig, then `args` in sdl resolvers is implicitly of type any instead of what they should be. 😕

Now that eslint configs have been sorted out, we should be able to safely tackle this one. 

AC
- [ ] refactor enableSAML to have optional args. Typescript throws warnings that the args are optional

Estimate: 8 hours

ackernaut:
**Story Points: 5**
  [See the discussion](https://action.parabol.co/meet/chorM4RmGA/estimate/11) in PS2 Poker 1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

ackernaut:
**Include?: Y**
  [See the discussion](https://action.parabol.co/meet/chorM4RmGA/estimate/12) in PS2 Poker 1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

Dschoordsch:
@ackernaut I don't mind you claiming this one, but technically "Refactor GraphQL SDL" ended up in Squad: 1

ackernaut:
@Dschoordsch do you still want to move this to squad 1? do you want to include it in sprint 100?

Dschoordsch:
@ackernaut if you have open capacity feel free to keep it in squad 2

ackernaut:
Okay, I understand it to be a DX related issue, so should be fine

PR #6329, opened 2022-03-31
refactor(lint): Lint server

### LINKED ISSUE #6328: feat(lint): lint server
successor: #6314

prior to doing a huge format on all our files, we should fix the things that are more than just style changes. so when the big change comes along we don't need to spend much time doing a code review

AC
- [ ] run `yarn lint` on the server and fix all the errors

Estimate: 2 hours

mattkrick:
fix #6328

In efforts to keep #6327 as small as possible, I broke out the server linting (just linting, not prettier) into its own PR here.

With the fixed glob pattern, this found a few unused files, which are deleted here.
It also found that a few of the files that belong in the graphql/private dir were in the public dir, so those have been moved.
I also removed the github/gitlabTypes.ts from git since we generate those on the fly now

Note: this doesn't refactor all the su* queries. I'll do that in a subsequent PR to keep this one small-ish.

TEST
- [ ] just look over the files & make sure the server starts up. No features added, just a refactor


packages/server/graphql/private/queries/suProOrgInfo.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const suProOrgInfo: QueryResolvers['suProOrgInfo'] = async (
+  _source,
+  {includeInactive},
+  {authToken}
+) => {
+  const r = await getRethink()
+
+  // AUTH
+  requireSU(authToken)
```
mattkrick, line 14: I didn't change any part of this resolve fn.
this auth check is removed in the next PR #6330 because auth is now handled in `permissions.ts`.

packages/server/graphql/private/typeDefs/flagOverLimit.graphql
```diff
@@ -0,0 +1,25 @@
+extend type Mutation {
```
mattkrick, line 1: there's no real need for me to have broken these out, except I want to establish the pattern so others can see how new mutations should look

nickoferrall (approved):
LGTM! Got things up and running after removing and reinstalling node_modules

PR #6330, opened 2022-03-31
refactor(requireSU): Remove requireSU

mattkrick:
predecessor: #6329

### What's it do?
- moves all superuser queries into the graphql/private schema dir
- refactor createImposterToken since it's the only superuser mutation on the public schema
- removes the legacy requireSU function to avoid confusion

TEST
- [ ] try impersonating someone when you aren't logged in as a superuser. it doesn't work
- [ ] try it again as a superuser, it works


JimmyLv (approved):
This looks great to me! Just curious what the `impersonate` is doing here? Is it a generic pattern?
```js
new AuthToken({sub: userId, tms, rol: 'impersonate'})
```
I saw a related route ``/admin/impersonate/:userid`` and want to confirm that this would help "us" with admin privileges to troubleshoot the specific user?

mattkrick:
good question! 

auto logs us out & keeps us ghosted: https://www.notion.so/parabol/How-to-impersonate-another-user-795c480be5de40138e2fec7ddaa43c31

PR #6331, opened 2022-03-31
feat(integrations): save favorite integration to local storage

mattkrick:
debugging poker scoping issues sucks when you have to keep clicking a tab after refresh.

TEST
- [ ] start a poker meeting, see that the default tab is jira
- [ ] click the github tab & then hit refresh
- [ ] notice that it loads the github tab by default this time 🎉


nickoferrall (approved):
This is really helpful! Perhaps we could use the `favoriteService` in other areas of the app. For example, when the user opens all available integrations in the task footer menu, we could sort favorite service integrations above services? Added as a comment here: https://github.com/ParabolInc/parabol/issues/6007#issuecomment-1087377164

mattkrick:
💯  this is _just the beginning_.

PR #6332, opened 2022-04-01
chore: Track integrations in createTask

Dschoordsch:
At the moment we do not know how many users use the "Add issue" fab in
Poker meetings to create new tasks in their integration. Add the
integration field in the segment message to track it.


nickoferrall (approved):
LGTM!

PR #6333, opened 2022-04-01
feat(standups): upsertTeamPromptResponse

tianrunhe:
Implements #5948

Tests:
- [ ] When user starts an async stand up meeting, empty prompt responses are created. The number of responses should equal to the number of team members
- [ ] In `localhost:3000/admin/graphql`, admin can run `upsertTeamPromptResponse` mutation successfully:
```GraphQL
mutation ($teamPromptResponse: TeamPromptResponseInput!) {
  upsertTeamPromptResponse(teamPromptResponse: $teamPromptResponse) {
    ... on ErrorPayload {
      error {
        message
      }
    }
    ... on UpsertTeamPromptResponseSuccess {
      teamPromptResponse {
        id
        content
        plaintextContent
      }
      meeting {
        ... on TeamPromptMeeting {
          id
          createdAt
          responses {
            plaintextContent
          }
        }
      }
    }
  }
}

{
  "teamPromptResponse": {
    "meetingId": "cw4A1FHEGs",
    "content": "{\"type\":\"doc\",\"content\":[{\"type\":\"paragraph\",\"content\":[{\"type\":\"text\",\"text\":\"What's up! This editor instance exports its \"},{\"type\":\"text\",\"marks\":[{\"type\":\"bold\"}],\"text\":\"content\"},{\"type\":\"text\",\"text\":\" as JSON.\"}]}]}"
  }
}
```


packages/server/dataloader/primaryKeyLoaderMaker.ts
```diff
@@ -20,3 +20,20 @@ export function primaryKeyLoaderMaker<ReturnT extends {id: string}>(
     )
   }
 }
+
+/**
+ * Register a loader for types loaded by their {id: number} field.
+ */
+export function numberedPrimaryKeyLoaderMaker<ReturnT extends {id: number}>(
```
tianrunhe, line 27: I had hard time merging this with the one above...

tianrunhe (commented):
@Dschoordsch Hey Georg, do you know why would I get the [type error](https://github.com/ParabolInc/parabol/runs/5791170746?check_suite_focus=true)? I suspect this has something to do with the new primary key loader with a `number` key but couldn't find how to fix it.
What's your opinion on keeping the `id` for `TeamPromptResponse` integer in PG? Pro is we could auto generate/increase, con is exactly what I'm facing right now.
Thanks for your time!

packages/server/dataloader/primaryKeyLoaderMaker.ts
Dschoordsch: We should make the key type generic in `primaryKeyLoaderMaker` instead:
```ts
export function primaryKeyLoaderMaker<KeyT extends string | number, ReturnT extends {id: KeyT}>(
  batchFn: (ids: readonly KeyT[]) => Promise<ReturnT[]>
) {
  return (parent: RootDataLoader) => {
    return new NullableDataLoader<KeyT, ReturnT, KeyT>(
      async (ids) => {
        const result = await batchFn(ids)
        return normalizeResults(ids, result)
      },
      {...parent.dataLoaderOptions}
    )
  }
}
```
I also prepared already the `normalizeResults` function but didn't had any use yet, now it's there:
```ts
export const normalizeResults = <KeyT extends string | number, T extends {[key: string]: KeyT}>(
  keys: Readonly<KeyT[]>,
  results: T[],
  key: keyof T = 'id'
) => {
  const map = {} as Record<KeyT, T>
  results.forEach((result: T) => {
    result[key]
    map[result[key]] = result
  })
  const mappedResults = [] as T[]
  keys.forEach((key) => {
    mappedResults.push(map[key])
  })
  return mappedResults
}
```

I don't know where exactly the type gets lost in `foreignKeyLoaderMaker` but sprinkling in some more helper types and strategic casting got it working
```ts
type LoaderKeyType<LoaderName extends LoaderKeys> = Loader<LoaderName> extends DataLoader<
  infer KeyT,
  any,
  infer KeyT
>
  ? KeyT : any

/**
 * Used to register loaders for types by foreign key.
 *
 * This is intended for types which already have a {@link primaryKeyLoaderMaker} defined.
 * When an item is loaded via this loader, the primary loader will be primed with the result as well.
 * It reflects a one to many relationship, i.e. for each key passed, an array will be returned.
 */
export function foreignKeyLoaderMaker<
  LoaderName extends LoaderKeys,
  KeyName extends keyof LoaderType<LoaderName>
>(
  primaryLoaderKey: LoaderName,
  foreignKey: KeyName,
  fetchFn: (keys: readonly LoaderType<LoaderName>[KeyName][]) => Promise<LoaderType<LoaderName>[]>
) {
  type T = LoaderType<LoaderName>
  type PrimaryKeyT = LoaderKeyType<LoaderName>
  type KeyValue = T[KeyName]
  return (parent: RootDataLoader) => {
    const primaryLoader = parent.get(primaryLoaderKey) as DataLoader<PrimaryKeyT, T, PrimaryKeyT>
    return new DataLoader<KeyValue, T[], KeyValue>(
      async (ids) => {
        const items = await fetchFn(ids)
        items.forEach((item) => {
          const key = item?.['id']
          if (key) {
            primaryLoader.clear(key).prime(key, item)
          }
        })
        return ids.map((id) => items.filter((item) => item && item[foreignKey] === id))
      },
      {
        ...parent.dataLoaderOptions
      }
    )
  }
}
```
Basically the `parent.get(primaryLoaderKey)` seems to be the culprit. Some type is not restrictive enough to make it work. Not very satisfactory but it should work.

tianrunhe:
@BartoszJarocki Would you take a look at this one? A lot of meat here 😁

packages/server/dataloader/foreignKeyLoaderMaker.ts
```diff
@@ -13,6 +13,14 @@ type LoaderType<LoaderName extends LoaderKeys> = Loader<LoaderName> extends Data
   ? NonNullable<T>
   : any
 
+type LoaderKeyType<LoaderName extends LoaderKeys> = Loader<LoaderName> extends DataLoader<
+  infer KeyT,
+  any,
+  infer KeyT
+>
+  ? KeyT
+  : any
```
BartoszJarocki, line 22: +2 🤯

packages/server/graphql/mutations/helpers/createNewMeetingPhases.ts
```diff
@@ -145,6 +146,14 @@ const createNewMeetingPhases = async (
             discussionTopicId: stage.teamMemberId,
             discussionTopicType: 'teamPromptResponse' as const
           }))
+          const teamMemberPromptResponses = teamPromptStages.map((stage, index) => ({
+            meetingId,
+            userId: stage.teamMemberId.split('::')[0],
```
BartoszJarocki, line 151: -1 it'd be better to use `TeamMemberId.split` here

packages/server/graphql/mutations/updatePromptResponse.ts
```diff
@@ -0,0 +1,78 @@
...
+  } else if (input.content) {
+    const content = input.content
+    const newLine = input.type === 'paragraph' ? '\n' : ''
+    return (
+      content.map((subContent) => extractTextFromTipTapJSONContent(subContent)).join('') + newLine
+    )
+  } else {
+    return input.type === 'paragraph' ? '\n' : ''
+  }
+}
```
BartoszJarocki, line 24: +1 I'd create a separate directory in `utils` named `tiptap` and would put it there

packages/server/graphql/mutations/updatePromptResponse.ts
```diff
@@ -0,0 +1,78 @@
...
+
+const updatePromptResponse = {
```
BartoszJarocki, line 26: +1 what would you say about using the new way of creating mutations here? definitely not a block, I think it could be fun to explore

packages/server/graphql/mutations/updatePromptResponse.ts
```diff
@@ -0,0 +1,78 @@
...
+  type: GraphQLNonNull(UpdatePromptResponsePayload),
+  description: ``,
+  args: {
+    promptResponseId: {
+      type: new GraphQLNonNull(GraphQLID)
+    },
+    content: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'A stringified TipTap content containing prompt response'
+    }
```
BartoszJarocki, line 36: +1 I'd create the `UpdateTeamPromptResponseInput` input type to be a good graphql citizen

packages/server/graphql/mutations/updatePromptResponse.ts
```diff
@@ -0,0 +1,78 @@
...
+      content.map((subContent) => extractTextFromTipTapJSONContent(subContent)).join('') + newLine
+    )
+  } else {
+    return input.type === 'paragraph' ? '\n' : ''
+  }
+}
+
+const updatePromptResponse = {
+  type: GraphQLNonNull(UpdatePromptResponsePayload),
+  description: ``,
```
BartoszJarocki, line 28: -1 missing description

packages/server/graphql/mutations/updatePromptResponse.ts
```diff
@@ -0,0 +1,78 @@
...
+      type: new GraphQLNonNull(GraphQLID)
+    },
+    content: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'A stringified TipTap content containing prompt response'
+    }
+  },
+  resolve: async (
+    _source: unknown,
+    {promptResponseId, content},
```
BartoszJarocki, line 40: -1 types please (a new way of creating mutation would take care of it automatically)

packages/server/graphql/mutations/updatePromptResponse.ts
```diff
@@ -0,0 +1,78 @@
...
+    const promptResponse = await getTeamPromptResponseById(promptResponseId)
+    if (!promptResponse) {
+      return standardError(new Error('PromptResponse not found'), {userId: viewerId})
+    }
+    const {userId} = promptResponse
+    if (userId !== viewerId) {
+      return standardError(new Error("Can't edit other's response"), {userId: viewerId})
+    }
+    const {meetingId} = promptResponse
+
```
BartoszJarocki, line 57: At some point we'll need to check if the meeting is already ended, or maybe if user stage is completed, that's the future tension though

packages/server/graphql/mutations/updatePromptResponse.ts
BartoszJarocki, line 26: +1 nit: I'd name it `updateTeamPromptReponse` just to be consistent

packages/server/graphql/types/UpdatePromptResponsePayload.ts
```diff
@@ -0,0 +1,31 @@
...
+import TeamPromptResponse from './TeamPromptResponse'
+
+export const UpdatePromptResponseSuccess = new GraphQLObjectType<any, GQLContext>({
+  name: 'UpdatePromptResponseSuccess',
+  fields: () => ({
+    error: {
+      type: StandardMutationError
+    },
+    meeting: {
+      type: NewMeeting,
```
BartoszJarocki, line 16: -1 this should not be nullable

packages/server/graphql/types/UpdatePromptResponsePayload.ts
```diff
@@ -0,0 +1,31 @@
...
+import NewMeeting from './NewMeeting'
+import StandardMutationError from './StandardMutationError'
+import TeamPromptResponse from './TeamPromptResponse'
+
+export const UpdatePromptResponseSuccess = new GraphQLObjectType<any, GQLContext>({
+  name: 'UpdatePromptResponseSuccess',
+  fields: () => ({
+    error: {
+      type: StandardMutationError
+    },
```
BartoszJarocki, line 14: -1 `makeMutationPayload` takes care of resolving to error type if the error is present so this isn't needed

packages/server/graphql/types/UpdatePromptResponsePayload.ts
```diff
@@ -0,0 +1,31 @@
...
+    meeting: {
```
BartoszJarocki, line 15: +1 I'd also add a `meetingId` to `UpdatePromptResponseSuccess ` as it most probably will be needed in updaters on frontend side, we might get away without fetching the whole meeting

packages/server/graphql/types/UpdatePromptResponsePayload.ts
```diff
@@ -0,0 +1,31 @@
...
+      type: NewMeeting,
+      resolve: resolveNewMeeting
+    },
+    promptResponse: {
+      type: TeamPromptResponse,
```
BartoszJarocki, line 20: -1 this should not be nullable

packages/server/graphql/types/TeamPromptResponse.ts
```diff
@@ -1,5 +1,6 @@
 import {GraphQLFloat, GraphQLID, GraphQLNonNull, GraphQLObjectType, GraphQLString} from 'graphql'
 import {GQLContext} from '../graphql'
+import {JSONContent} from '@tiptap/react'
```
BartoszJarocki, line 3: +1 I noticed we sometimes use `@tiptap/react`, sometimes `@tiptap/core`, does it matter?

packages/server/postgres/queries/getTeamPromptResponsesByIds.ts
```diff
@@ -0,0 +1,26 @@
+import {getTeamPromptResponsesByIdsQuery, IGetTeamPromptResponsesByIdsQueryResult} from './generated/getTeamPromptResponsesByIdsQuery'
+import {JSONContent} from '@tiptap/core'
+import getPg from '../getPg'
+
+export interface TeamPromptResponse extends  Omit<IGetTeamPromptResponsesByIdsQueryResult, 'content'> {
+  content: JSONContent
+}
+
+export const mapToTeamPromptResponse = (results: IGetTeamPromptResponsesByIdsQueryResult[]): TeamPromptResponse[] => {
```
BartoszJarocki, line 9: +1 do we need to export that?

packages/server/postgres/queries/getTeamPromptResponsesByIds.ts
```diff
@@ -0,0 +1,26 @@
...
+  return results.map((teamPromptResponse: any) => {
+    return {
+      ...teamPromptResponse,
+      content: teamPromptResponse.content as JSONContent
+    } as TeamPromptResponse
+  })
+}
+
+export const getTeamPromptResponsesByIds = async (promptResponseIds: readonly number[]): Promise<TeamPromptResponse[]> => {
```
BartoszJarocki, line 18: +1 I'd use `teamPromptResponseIds`, I know it's a little verbose but it makes me think less

packages/server/postgres/queries/updateTeamPromptResponseContentById.ts
```diff
@@ -0,0 +1,11 @@
+import getPg from '../getPg'
+import {
+  IUpdateTeamPromptResponseContentByIdQueryParams,
+  updateTeamPromptResponseContentByIdQuery
+} from './generated/updateTeamPromptResponseContentByIdQuery'
+
+export const updateTeamPromptResponseContentById = async (
+  params: IUpdateTeamPromptResponseContentByIdQueryParams
+) => {
+  await updateTeamPromptResponseContentByIdQuery.run(params, getPg())
```
BartoszJarocki, line 10: +1 I'd just return instead of await

BartoszJarocki (changes requested):
looks good! a couple of minor things and a bunch of things to consider

packages/server/dataloader/foreignKeyLoaderMaker.ts
tianrunhe: It's Georg's brain!

tianrunhe:
@BartoszJarocki I rewrite the mutation in SDL! It's pretty fun. Can you take another look? 🙏

packages/server/graphql/public/mutations/updateTeamPromptResponse.ts
```diff
@@ -0,0 +1,57 @@
...
+  {updatedTeamPromptResponse},
+  {authToken, dataLoader, socketId: mutatorId}
+) => {
+  const viewerId = getUserId(authToken)
+  const operationId = dataLoader.share()
+  const subOptions = {mutatorId, operationId}
+  const {teamPromptResponseId, content} = updatedTeamPromptResponse
+
+  // AUTH
+  const promptResponse = await getTeamPromptResponseById(teamPromptResponseId)
```
BartoszJarocki, line 22: +1 if we use `dataLoader` here, we might have one less DB call when resolving the response. I see you're already updating the `promptResponse` after the insert so just changing this line with `dataLoader` will do the job

packages/server/graphql/types/UpdatePromptResponsePayload.ts
```diff
@@ -0,0 +1,33 @@
...
+    }
+  })
+})
+
+const UpdatePromptResponsePayload = makeMutationPayload(
+  'UpdatePromptResponsePayload',
+  UpdatePromptResponseSuccess
+)
+
+export default UpdatePromptResponsePayload
```
BartoszJarocki, line 33: -1 we already have this type created using new method [here](https://github.com/ParabolInc/parabol/pull/6333/files#diff-cad85a115f24af217d3957c2fbd0577f226056c7f0b91566ab1e1df42b57ae9bR10)

BartoszJarocki (changes requested):
looks really good! there's one case I've missed that needs attention, other than that 👌

packages/server/graphql/public/types/UpdateTeamPromptResponsePayload.ts
```diff
@@ -0,0 +1,23 @@
...
+const UpdateTeamPromptResponsePayload: UpdateTeamPromptResponsePayloadResolvers = {
+  teamPromptResponse: async (source, _args, {dataLoader}) => {
+    if ('error' in source) return null
+    const {teamPromptResponseId} = source
+    return dataLoader.get('teamPromptResponses').loadNonNull(teamPromptResponseId)
+  },
+  meeting: async (source, _args, {dataLoader}) => {
+    if ('error' in source) return null
+    const {meetingId} = source
+    return dataLoader.get('newMeetings').load(meetingId)
```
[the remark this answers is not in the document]
tianrunhe: Unfortunately the meeting data loader is still in the old RethinkDB loader and does not have the `loadNotNull` method...

packages/server/graphql/types/UpdatePromptResponsePayload.ts
tianrunhe: Oops forgot to delete this one! Nice catch!

packages/server/graphql/mutations/helpers/createNewMeetingPhases.ts
```diff
@@ -145,6 +147,14 @@ const createNewMeetingPhases = async (
...
             discussionTopicType: 'teamPromptResponse' as const
           }))
+          const teamMemberPromptResponses = teamPromptStages.map((stage, index) => ({
+            meetingId,
+            userId: TeamMemberId.split(stage.teamMemberId).userId,
+            sortOrder: index,
+            content: {},
+            plaintextContent: ''
+          }))
+          asyncSideEffects.push(upsertTeamPromptResponses(teamMemberPromptResponses))
```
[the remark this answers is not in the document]
tianrunhe: Very good catch! I'll create a separate mutation

packages/server/graphql/public/mutations/createTeamPromptResponse.ts
```diff
@@ -0,0 +1,55 @@
...
+  const meeting = await dataLoader.get('newMeetings').load(meetingId)
+  const {endedAt, teamId} = meeting
+  if (!isTeamMember(authToken, teamId)) {
+    return standardError(new Error('Team not found'), {userId: viewerId})
+  }
+  if (endedAt) return standardError(new Error('Meeting already ended'), {userId: viewerId})
+
+  // VALIDATION
+  /**
+   * TODO: check if there's already a response for this user in this meeting
```
tianrunhe, line 30: To check this, we need `getTeamPromptResponsesByMeetingId` query, which I have in [subsequent PR](https://github.com/ParabolInc/parabol/pull/6388/files#diff-00749d0652800c424824c0c23172bbae293db5aa559f1fd57d8295a999656210). To prevent this PR becomes a monster, I'll save this for later.

tianrunhe:
@BartoszJarocki I added a new `createTeamPromptResponse` mutation. Please take a look, thanks!

packages/server/graphql/public/mutations/createTeamPromptResponse.ts
tianrunhe: This issue is addressed [here](https://github.com/ParabolInc/parabol/pull/6388/files#diff-432ce08d5f6c923e8425643ee15814afda231e2941dcb4a278e99a87824e02deR33)

packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
```diff
@@ -0,0 +1,81 @@
...
+  {authToken, dataLoader, socketId: mutatorId}
+) => {
+  const viewerId = getUserId(authToken)
+  const operationId = dataLoader.share()
+  const subOptions = {mutatorId, operationId}
+  const {teamPromptResponseId, meetingId, content} = teamPromptResponse
+
+  // AUTH
+  const promptResponse = teamPromptResponseId
+    ? await dataLoader.get('teamPromptResponses').loadNonNull(teamPromptResponseId)
```
BartoszJarocki, line 22: -2 if I see correctly it'll throw an error if there's no data, so we'd need to use normal `load` method here

packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
```diff
@@ -0,0 +1,81 @@
...
+  {teamPromptResponse},
+  {authToken, dataLoader, socketId: mutatorId}
+) => {
+  const viewerId = getUserId(authToken)
+  const operationId = dataLoader.share()
+  const subOptions = {mutatorId, operationId}
+  const {teamPromptResponseId, meetingId, content} = teamPromptResponse
+
+  // AUTH
+  const promptResponse = teamPromptResponseId
```
BartoszJarocki, line 21: -1 this has implicit any now, let's give it a proper type

packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
```diff
@@ -0,0 +1,81 @@
...
+    : {
+        meetingId,
+        userId: viewerId,
+        sortOrder: 0,
+        content: {},
+        plaintextContent: ''
+      }
+  if (!promptResponse) {
+    return standardError(new Error('PromptResponse not found'), {userId: viewerId})
+  }
```
BartoszJarocki, line 32: -1 this will never happen, now if there's no response in db we provide manually created object

BartoszJarocki (changes requested):
@tianrunhe a couple of things, then it's good to go, nice work!

tianrunhe:
> @tianrunhe a couple of things, then it's good to go, nice work!

@BartoszJarocki I just fixed these things. I'd appreciate if you can take another look. Thanks!

packages/server/postgres/queries/src/upsertTeamPromptResponsesQuery.sql
```diff
@@ -0,0 +1,10 @@
+/*
+  @name upsertTeamPromptResponsesQuery
+  @param responses -> ((meetingId, userId, sortOrder, content, plaintextContent)...)
+*/
+INSERT INTO "TeamPromptResponse" ("meetingId", "userId", "sortOrder", "content", "plaintextContent")
+VALUES :responses
+ON CONFLICT ("meetingId", "userId") DO UPDATE SET
```
BartoszJarocki, line 7: -2 it's on me but there's no unique constraint for both `meetingId` and `userId` so this insert fails with 
```
Query 1 ERROR: ERROR:  there is no unique or exclusion constraint matching the ON CONFLICT specification
```

I wish we could simply use here the `id`. 
1. If `inputTeamPromptResponseId` is provided, provide use it
2. if `inputTeamPromptResponseId` is _not_ provided, use `DEFAULT`

I don't think it's possible with `pgtyped` but it'd be the cleanest way IMO. otherwise, we need another migration to create unique constraint between `meetingId` and `userId`

tianrunhe: Oh darn! It's actually my bad. I got that error before and as a temporary work around to test other things, I added the unique constraint manually to the DB and since then forgot I had that in place...

tianrunhe: seems no easy way to do it so I just added a migration

jmtaber129: FWIW I think it's worth enforcing this uniqueness constraint, even if it means adding a new migration, and even if there's a workaround using `id`.  I have tension around not enforcing uniqueness for things that we know to be unique.

BartoszJarocki (approved):
lgtm!

packages/server/postgres/queries/upsertTeamPromptResponses.ts
```diff
@@ -0,0 +1,25 @@
...
+} from './generated/upsertTeamPromptResponsesQuery'
+
+export type InputTeamPromptResponses = DeepNonNullable<
+  IUpsertTeamPromptResponsesQueryParams['responses']
+>
+export type InputTeamPromptResponse = IUpsertTeamPromptResponsesQueryParams['responses'][0]
+
+export const upsertTeamPromptResponse = async (
+  response: InputTeamPromptResponse
+): Promise<number> => {
```
mattkrick, line 15: +1 try to avoid declaring return types, sometimes it's unavoidable, but it forces good habits because the single source of truth is in the `return` statement

mattkrick (changes requested):

packages/server/graphql/mutations/helpers/tiptap/extractTextFromTipTapJSONContent.ts
```diff
@@ -0,0 +1,17 @@
+import {JSONContent} from '@tiptap/core'
+
+const extractTextFromTipTapJSONContent = (input: JSONContent): string => {
+  if (input.text) {
```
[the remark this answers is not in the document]
tianrunhe: The methods are defined on class `Editor` (which I found confusing as it's more about the state of the editor, so I renamed it in our code), not on `JSONContent`. We [call](https://github.com/ParabolInc/parabol/pull/6392/files#diff-9a4772bf006047ee4c0e2f50318333ead4d455b37ea159ef99cbac2ce69774a5R80) `getJSON()` on it to get the JSON representation of the content in the editor, and the stringinified version of that object is passed into this mutation. 
I guess we could also call `getText()` there and pass both of the JSON structure and the simple text to this mutation?

packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
```diff
@@ -0,0 +1,84 @@
...
+
+const upsertTeamPromptResponse: MutationResolvers['upsertTeamPromptResponse'] = async (
+  _source,
+  {teamPromptResponse},
+  {authToken, dataLoader, socketId: mutatorId}
+) => {
+  const viewerId = getUserId(authToken)
+  const operationId = dataLoader.share()
+  const subOptions = {mutatorId, operationId}
+  const {teamPromptResponseId: inputTeamPromptResponseId, meetingId, content} = teamPromptResponse
```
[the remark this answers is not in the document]
tianrunhe: But when we insert a new response, we need `meetingId`. I adjust the validation logic accordingly.

packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
```diff
@@ -0,0 +1,84 @@
...
+  }
+  if (endedAt) return standardError(new Error('Meeting already ended'), {userId: viewerId})
+
+  // RESOLUTION
+  const contentJSON: JSONContent = JSON.parse(content)
+  const plaintextContent = extractTextFromTipTapJSONContent(contentJSON)
+  const teamPromptResponseInput: InputTeamPromptResponse = {
+    meetingId,
+    userId: viewerId,
+    sortOrder: 0, //TODO: set correct sort order, need to fetch all responses for meeting
```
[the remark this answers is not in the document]
tianrunhe: It's currently [defined as non-null](https://github.com/ParabolInc/parabol/blob/master/packages/server/postgres/migrations/1644936872423_addTeamPromptResponseTable.ts#L24). If we think we don't need it, we can remove it later, with a migration etc.

packages/server/postgres/queries/src/getTeamPromptResponsesByIdsQuery.sql
```diff
@@ -0,0 +1,6 @@
+/*
+  @name getTeamPromptResponsesByIdsQuery
+  @param ids -> (...)
+*/
+SELECT "id", "createdAt", "updatedAt", "meetingId", "userId", "sortOrder", "content", "plaintextContent" FROM "TeamPromptResponse"
```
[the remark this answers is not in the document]
tianrunhe: I explicitly dropped `reactjis` from the list of columns due to the discussion on custom PG type [here](https://parabol.slack.com/archives/C039L3XG6BH/p1649943641865019).

packages/server/postgres/queries/getTeamPromptResponsesByIds.ts
```diff
@@ -0,0 +1,26 @@
+import {getTeamPromptResponsesByIdsQuery, IGetTeamPromptResponsesByIdsQueryResult} from './generated/getTeamPromptResponsesByIdsQuery'
+import {JSONContent} from '@tiptap/core'
+import getPg from '../getPg'
+
+export interface TeamPromptResponse extends  Omit<IGetTeamPromptResponsesByIdsQueryResult, 'content'> {
+  content: JSONContent
+}
+
+const mapToTeamPromptResponse = (results: IGetTeamPromptResponsesByIdsQueryResult[]): TeamPromptResponse[] => {
```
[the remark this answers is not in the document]
tianrunhe: can you elaborate more on what do you suggest to do here?

packages/server/graphql/mutations/helpers/tiptap/extractTextFromTipTapJSONContent.ts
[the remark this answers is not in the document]
mattkrick: on the server, rehydrate the JSONContent into an Editor & then call getText on it

packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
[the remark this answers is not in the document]
mattkrick: could you point out the validation that would prevent the above from happening?

packages/server/postgres/queries/src/getTeamPromptResponsesByIdsQuery.sql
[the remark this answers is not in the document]
mattkrick: we need reactjis here to fulfill the contract. we can cast the json to whatever we please in the helper

packages/server/graphql/mutations/helpers/tiptap/extractTextFromTipTapJSONContent.ts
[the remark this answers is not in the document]
BartoszJarocki: rehydration sounds good but seems like `Editor` doesn't really work in server env.

When creating it that way
```
  const editor = new Editor({
    element: null,
    content,
    extensions: createEditorExtensions()
  })
```
we get 
```
  "errors": [
    {
      "message": "document is not defined",
      "locations": [
        {
          "line": 33,
          "column": 3
        }
      ],
      "path": [
        "upsertTeamPromptResponse"
      ]
    }
  ],
  "data": null
}
```

It also doesn't work with JSDOM:
```
  const {document} = new JSDOM().window
  const editor = new Editor({
    element: document.createElement('tmp'),
    content,
    extensions: createEditorExtensions()
  })
```

@mattkrick, @jmtaber129 @JimmyLv any ideas?

packages/server/graphql/public/typeDefs/upsertTeamPromptResponse.graphql
```diff
@@ -0,0 +1,44 @@
...
+  """
+  upsertTeamPromptResponse(
+    """
+    The input object containing the team prompt response id and modified content
+    """
+    teamPromptResponse: TeamPromptResponseInput!
+  ): UpsertTeamPromptResponsePayload!
+}
+
+input TeamPromptResponseInput {
```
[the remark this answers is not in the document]
BartoszJarocki: done

packages/server/postgres/queries/src/getTeamPromptResponsesByIdsQuery.sql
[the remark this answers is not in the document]
BartoszJarocki: as `Reactji` is a composite type we can't use `SELECT *` as it breaks pg-typed
```
Error: Postgres type '_Reactji' is not supported by mapping
    at TypeAllocator.use (/Users/bjarocki/Work/parabol/web-parabol/node_modules/@pgtyped/cli/lib/types.js:125:38)
    at /Users/bjarocki/Work/parabol/web-parabol/node_modules/@pgtyped/cli/lib/generator.js:58:36
    at Array.forEach (<anonymous>)
    at /Users/bjarocki/Work/parabol/web-parabol/node_modules/@pgtyped/cli/lib/generator.js:57:21
    at Generator.next (<anonymous>)
    at fulfilled (/Users/bjarocki/Work/parabol/web-parabol/node_modules/@pgtyped/cli/lib/generator.js:5:58)
```

@mattkrick I suggest to leave it as is here. @jmtaber129 already handled it here very nicely: https://github.com/ParabolInc/parabol/pull/6407/commits/747f1551fbc0fec5d62c2a347a068557386a4ddc#diff-00749d0652800c424824c0c23172bbae293db5aa559f1fd57d8295a999656210R5

packages/server/graphql/public/types/UpsertTeamPromptResponsePayload.ts
```diff
@@ -0,0 +1,15 @@
+import {ErrorPayload, UpsertTeamPromptResponsePayloadResolvers} from '../../private/resolverTypes'
+import {UpsertTeamPromptResponseSuccessSource} from './UpsertTeamPromptResponseSuccess'
+
+export type UpsertTeamPromptResponsePayloadSource =
+  | UpsertTeamPromptResponseSuccessSource
+  | ErrorPayload
+
+const UpsertTeamPromptResponsePayload: UpsertTeamPromptResponsePayloadResolvers = {
+  __resolveType: (source) => {
```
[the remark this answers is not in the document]
BartoszJarocki: oh, my bad. `addRequestors` returns an union of `GraphQLSchema` and gitlab/github stuff and has to be used as the last, my bad! all good now

```
const rootSchema = addRequestors(resolveTypesForMutationPayloads(withNestedResolversSchema))
```

packages/server/graphql/mutations/helpers/tiptap/extractTextFromTipTapJSONContent.ts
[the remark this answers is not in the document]
jmtaber129: @BartoszJarocki  Directly importing `generateText` seems to work for me:
```
import {JSONContent, generateText} from '@tiptap/core'
import StarterKit from '@tiptap/starter-kit'

const extractTextFromTipTapJSONContent = (input: JSONContent): string => {
  return generateText(input, [StarterKit], {})
}

export default extractTextFromTipTapJSONContent
```

Screen Shot 2022-04-27 at 12 56 26 PM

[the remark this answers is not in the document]
BartoszJarocki: oh this is great find! I don't see that in their docs but seems to do exactly what we need: https://github.com/ueberdosis/tiptap/blob/main/packages/core/src/helpers/generateText.ts 

thank you @jmtaber129 🙏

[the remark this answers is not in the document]
BartoszJarocki: @mattkrick done. i also extracted a helper method to create an array of extensions just to make sure we use the same config on client/backend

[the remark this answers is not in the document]
jmtaber129: Yea, I was suspicious of using it at first since it's in the `helpers` folder, but it's used in an example added in the PR - https://github.com/ueberdosis/tiptap/pull/1875/files#diff-ce42067e2376ec11487bcf3eb3cfc63fe82c8c1fe236794954e7a5e01f4cc72aR46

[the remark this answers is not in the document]
BartoszJarocki: yeah, should be safe to use. great find!

PR #6335, opened 2022-04-01
feat(lint): Lint client

mattkrick:
~~predecessor: #6330~~

Squad 2 has all the DX fun 😉 

93 files changed, but half are deletions, so the code review shouldn't be too bad.

I ran `yarn lint` on the client & fixed all the errors using the following logic:
- if the file was no longer used, I deleted it
- if the file was used but not included in the client bundle (e.g. used by the server) it was added to `tsconfig.eslint.json`.
- If the file had `==` I made it `===`
- if the file violated a rule of react-hooks, i disabled eslint for that line
- if the file use `withRouterProps<[]>` i converted it to `useRouter` to avoid the empty object error
- if there was an empty function, i added an inline comment to it to show that the noop was intentional

we should now have a lint-free codebase 🎉


tianrunhe:
👀 Taking a look

Issue #6199, opened 2022-03-15
Standup UI - Top Bar - Options Menu


BartoszJarocki:
Design - https://www.figma.com/file/bhiEOV4HK9Kv91HFsJmevV/Stand-up-Concepts?node-id=1138%3A72617

AC
- [ ] Options menu is located in the top left corner
- [ ] For now it can have an option for ending the meeting

PR #6336, opened 2022-04-01
feat(standups): Show meeting title

jmtaber129:
fixes https://github.com/ParabolInc/parabol/issues/6198

https://www.loom.com/share/f497ec22108442fb87d7c8d6e5dd67b0

Needs rebase once https://github.com/ParabolInc/parabol/pull/6318 is merged

PR #6342, opened 2022-04-04
feat(standups): Options menu - end meeting

jmtaber129:
fixes https://github.com/ParabolInc/parabol/issues/6199

Current behavior is to end the meeting, but stay on the meeting page.  After meeting is ended, "End Meeting" button is still present, but disabled.

(commits https://github.com/ParabolInc/parabol/commit/f0635559c344aacda1bdbcc76763b97095c56f4b and later are the relevant ones; the diff against https://github.com/ParabolInc/parabol/pull/6336's branch is inaccurate due to the merge from `master`)

Needs rebase after https://github.com/ParabolInc/parabol/pull/6336 is merged.

https://www.loom.com/share/b1871e6f2f29435f84e83e570e2db457

### DIFF
--- a/packages/client/components/MeetingTopBar.tsx
+++ b/packages/client/components/MeetingTopBar.tsx
@@ -47,7 +47,7 @@ const PrimaryActionBlock = styled('div')({
   display: 'flex'
 })
 
-const AvatarGroupBlock = styled('div')({
+export const AvatarGroupBlock = styled('div')({
   alignItems: 'center',
   display: 'flex',
   justifyContent: 'center',
--- a/packages/client/components/MeetingTopBar.tsx
+++ b/packages/client/components/MeetingTopBar.tsx
@@ -1,15 +1,15 @@
-import React, {ReactElement, ReactNode} from 'react'
 import styled from '@emotion/styled'
-import DemoCreateAccountButton from './DemoCreateAccountButton'
-import SidebarToggle from './SidebarToggle'
-import isDemoRoute from '../utils/isDemoRoute'
-import hasToken from '../utils/hasToken'
-import {meetingAvatarMediaQueries} from '../styles/meeting'
-import makeMinWidthMediaQuery from '../utils/makeMinWidthMediaQuery'
-import PlainButton from './PlainButton/PlainButton'
+import React, {ReactElement, ReactNode} from 'react'
 import {PALETTE} from '~/styles/paletteV3'
-import Icon from './Icon'
 import {ICON_SIZE} from '~/styles/typographyV2'
+import {meetingAvatarMediaQueries} from '../styles/meeting'
+import hasToken from '../utils/hasToken'
+import isDemoRoute from '../utils/isDemoRoute'
+import makeMinWidthMediaQuery from '../utils/makeMinWidthMediaQuery'
+import DemoCreateAccountButton from './DemoCreateAccountButton'
+import Icon from './Icon'
+import PlainButton from './PlainButton/PlainButton'
+import SidebarToggle from './SidebarToggle'
 
 const localHeaderBreakpoint = makeMinWidthMediaQuery(600)
 
@@ -47,7 +47,7 @@ const PrimaryActionBlock = styled('div')({
   display: 'flex'
 })
 
-export const AvatarGroupBlock = styled('div')({
+export const IconGroupBlock = styled('div')({
   alignItems: 'center',
   display: 'flex',
   justifyContent: 'center',
@@ -167,7 +167,7 @@ const MeetingTopBar = (props: Props) => {
         )}
         <ChildrenBlock>{children}</ChildrenBlock>
       </HeadingBlock>
-      <AvatarGroupBlock>
+      <IconGroupBlock>
         {showButton && (
           <PrimaryActionBlock>
             <DemoCreateAccountButton />
@@ -184,7 +184,7 @@ const MeetingTopBar = (props: Props) => {
             </DiscussionButton>
           </ButtonContainer>
         )}
-      </AvatarGroupBlock>
+      </IconGroupBlock>
     </MeetingTopBarStyles>
   )
 }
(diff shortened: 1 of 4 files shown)


packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
```diff
@@ -34,24 +51,27 @@ const TeamPromptTopBar = (props: Props) => {
...
         <TeamPromptHeader>
-          <BackButton
-            ariaLabel='Back to Meetings'
-            onClick={() => history.push('/meetings')}
-          />
+          <BackButton ariaLabel='Back to Meetings' onClick={() => history.push('/meetings')} />
           <TeamPromptHeaderTitle>{meetingName}</TeamPromptHeaderTitle>
         </TeamPromptHeader>
       </HeadingBlock>
+      <AvatarGroupBlock>
```
jmtaber129, line 70: Putting this button within the `AvatarGroupBlock` to stay consistent with the styling patterns of the existing `MeetingTopBar`

packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
```diff
@@ -0,0 +1,80 @@
...
+          <BackButton ariaLabel='Back to Meetings' onClick={() => history.push('/meetings')} />
+          <TeamPromptHeaderTitle>{meetingName}</TeamPromptHeaderTitle>
+        </TeamPromptHeader>
+      </HeadingBlock>
+      <AvatarGroupBlock>
+        <ButtonContainer>
+          <TeamPromptOptionsMenu meetingRef={meeting} />
+        </ButtonContainer>
+      </AvatarGroupBlock>
+      {/* :TODO: (jmtaber129): Add avatars, overflow menu, etc. */}
```
BartoszJarocki, line 75: +1 I guess that's need a little update, we already have a menu :)

packages/client/components/MeetingTopBar.tsx
```diff
@@ -47,7 +47,7 @@ const PrimaryActionBlock = styled('div')({
   display: 'flex'
 })
 
-const AvatarGroupBlock = styled('div')({
+export const AvatarGroupBlock = styled('div')({
```
BartoszJarocki, line 50: +1 now that we use this section not only for avatars, should we figure out a better name for it? not a big deal but it'd be nice

BartoszJarocki (approved):
just two little things to consider, looks really nice!

packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
jmtaber129: Updated + moved into the avatar block

packages/client/components/MeetingTopBar.tsx
jmtaber129: Definitely agree.  Renamed to `IconGroupBlock` (block of avatar icons, options icon, etc.), but open to suggestions!

BartoszJarocki: i don't have any better suggestions so I'll remain silent, `IconGroupBlock` sounds much better!

packages/client/components/TeamPrompt/TeamPromptOptionsMenu.tsx
```diff
@@ -0,0 +1,96 @@
...
+      <MenuItem
+        key='copy'
+        isDisabled={!!endedAt}
+        label={
+          <OptionMenuItem>
+            <StyledIcon>flag</StyledIcon>
+            <span>{'End this activity'}</span>
+          </OptionMenuItem>
+        }
+        onClick={async () => {
```
mattkrick, line 78: +1 no async needed

packages/client/components/TeamPrompt/TeamPromptOptionsMenu.tsx
```diff
@@ -0,0 +1,96 @@
...
+      }
+    `,
+    meetingRef
+  )
+
+  const {id: meetingId, endedAt} = meeting
+  const atmosphere = useAtmosphere()
+  const {onCompleted, onError} = useMutationProps()
+
+  const renderedMenu = (
```
mattkrick, line 67: +1 it's nice to break menus out into their own components early because they almost inevitably outgrow their parents eventually. By convention I'd usually call the parent component a `Toggle` and then the one inside the portal a `Menu`

mattkrick (changes requested):
+1 just 1 nit & a merge conflict

packages/client/components/TeamPrompt/TeamPromptOptionsMenu.tsx
jmtaber129: Could you elaborate a bit on what you'd expect this to look like in the context of the top bar where this is rendered?  E.g. should these toggle + menu components be controlled from the `TopBar` where `TeamPromptOptionsMenu` is currently being rendered as an uncontrolled component, should the toggle + menu still be in their own component but just with the menu pulled into its own component, etc.

mattkrick: yep, pulling the menu into its own component should suffice.  you already know the react fragment trick for handling menus, so no worries there.

A good example is how we do the dashboard topbar today (this component isn't named Toggle, but the same pattern applies): https://github.com/ParabolInc/parabol/blob/a88782807ca3a4bb1c291ec455b554f137e02476/packages/client/components/TopBarHelp.tsx

no need for the lazy preloads for small menus, but if the menus get large it's always nice to break it off into its own chunk, that way webpack can keep it out of the primary bundle & the page will load a little faster for first time visitors or when folks upgrade

jmtaber129: Gotcha, I think I refactored in the way you're suggesting but lmk if I misunderstood!

Issue #6138, opened 2022-02-28
Sprint Poker: GitLab v1: New Search Input UI


nickoferrall:
Predecessor: https://github.com/ParabolInc/parabol/issues/6046

In Jira and GitLab, the only way to see the project filters is to open the project filter menu:

![image](https://user-images.githubusercontent.com/39854876/156038846-7ba18b5e-5026-4a73-ae9a-12ebf363c5aa.png)

Instead, let's show the current project filters under the search input. While we're here, we can improve the UI by adding a border to the search input and making the filter button blue to draw more attention to it:

![image](https://user-images.githubusercontent.com/39854876/156039138-4a631d7c-1348-4a33-a5b2-f1110da6fcfe.png)

Note: we don't need to add the current filters in the GitHub integration as the search query already includes the repos we're filtering by. See discussion here: https://github.com/ParabolInc/parabol/issues/6088

### AC

- [ ] In all Sprint Poker integrations, the search input has a border and the filter menu has a blue background like the image above
- [ ] In Jira & GitLab integrations, list the current filters below the search input
- [ ] If there are many filters and there's not enough space to list them all, show `... and 9 more` at the end of the line
- [ ] The GitHub integration does not include the "current filters" text as users can see the selected repos in the search query

mattkrick:
https://user-images.githubusercontent.com/5514175/156233871-1dc1a6f1-6cd0-4fc2-947b-986915d84161.mov

nickoferrall:
^ that screen recording is the functionality that I suggest we tackle in this issue: https://github.com/ParabolInc/parabol/issues/6047

nickoferrall:
**Story points: 13**
  [See the discussion](https://action.parabol.co/meet/chrwZ1fHoc/estimate/11) in Sprint Poker #​​1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

nickoferrall:
**Include in next sprint?: Y**
  [See the discussion](https://action.parabol.co/meet/chrwZ1fHoc/estimate/12) in Sprint Poker #​​1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

Issue #6325, opened 2022-03-31
Sprint Poker: GitLab v2: Refactor nested issues query


nickoferrall:
In Poker, we want to show users their GitLab issues. However, the GitLab API offers a `projects` query but not an `issues` one. So, we're currently querying their most recently updated projects and getting the issues in those projects.

One problem with this is when the user scrolls down, we want to render more issues. As we can't have nested connections, we have to increase the number of projects being loaded which isn't what the user necessarily wants. 

See the discussion issue here for more context: https://github.com/ParabolInc/parabol/issues/6272

### AC

- [ ] For the initial fetch, use the projects API with a set number of issues & projects. it can't be increased. if gitlab returns an error, we tell the user to pick a project from the dropdown menu. Scrolling to the bottom does nothing.
- [ ] When the user has at least 1 selected project, we call the project(id) API instead, since that returns a single object. we map over the fitleredProjectIds and trigger a new query for each. Put a @connection on the issues field in this refetchable query
- [ ] When the user scrolls to the bottom, we call loadMore(20) to load 20 more issues on each. This will be a noop if all the issues have already been fetched for a given project since hasNextPage will be false
- [ ] The results from each of the queries is flatmapped together, just like today

jordanh:
Out of curiosity, could this perhaps be broken down into smaller issues?

nickoferrall:
> Out of curiosity, could this perhaps be broken down into smaller issues?

Yes, I have an idea on how I could break this into two PRs. My current approach:

We need to loop through the selected project id filters and execute multiple `project` queries. From the returned project queries, we need to flatMap the issues within those projects. As a lot of data manipulation is required, I think we should add a field called `projectIssues` to `GitLabIntegration` so that the server does this rather than the client each time. On the client, we would also have to loop through an array and call a Relay hook query, which would break the rules of hooks.

`projectIssues` will be a connection data type so that we can add a refetchable argument to the query. This will allow us to increase the `first` argument as the user scolls down the page, meaning that more issues will render as they scroll.

While it would be preferable to query the GitLab API directly from our frontend, as there's no `issues` endpoint, I think this is the best bet. 

One way that I could potentially break this down into smaller PRs is to implement `projectIssues` on the server in one PR. In a second PR, I could add the logic where a user scrolls and new issues render.

Issue #6070, opened 2022-02-16
Designed notifications side sheet


ackernaut:
Currently notifications are in a cramped menu. There are probably opportunities to clarify the meaning and state of notifications. Also, if we encourage async activities, folks may need to see which conversations have had new comments etc.. E.g. in async stand-ups, I might benefit from a notification that somebody commented on my response. Otherwise, I have to have a habit of checking the activity often to see if there are responses. The basic UX is reengaging folks where they left off participating in an activity.

### Questions
- Is a side sheet the best pattern? Why or why not? Does it make sense to put it on the left side of the dashboard where the upper-right user hub is?
- What are general patterns for notifications? Do they load indefinitely? Can they be cleared? If so, what does empty look like?
- How do I know if I’ve seen a notification or not, or taken action somehow?
- How can we style the avatar, meta, messages, and actions so that it makes good use of space but maintains rhythm, quick readability, usability, balance, etc.
- Stretch goal: what types of notifications could it support? Take an action as admin or team member? Reminders to take actions in new phases? Alerts that there are new comments, or changes to something? What should we not support as a notification? (Note: this is not an issue to design the specifics of individual notifications, but feel free to mock some up for the sake of showing different types)
- How does this open on mobile? Tablets? Larger viewports?

### Acceptance criteria
- Designed concepts of a notification side sheet or similar that shows states, workflows, etc.
- Proposed written criteria that can be used in an implementation issue
- Reviewed with a designer and developer

### Estimated effort
- 8 hours

ackernaut:
cc @enriquesanchez does this issue make sense? Anything missing? Does the scope match the effort close enough? Do you think this is a good idea overall, to use a side sheet?

enriquesanchez:
@ackernaut 

These questions are really good 🙌

> Do you think this is a good idea overall, to use a side sheet?

I think a side sheet that pushes the content to the left makes sense when the side sheet's content is related to the content being pushed, the use case being I may want to see both things at the same time. For example comments in a Notion doc.

In the case of notifications, I'd be more inclined to explore a dropdown pattern instead. Similar to what we currently have, but taller.

ackernaut:
>I think a side sheet that pushes the content to the left makes sense when the side sheet's content is related to the content being pushed

[...]

PR #6344, opened 2022-04-05
feat(sprint-poker): New Poker Scope UI

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6138


packages/client/components/GitLabScopingSearchCurrentFilters.tsx
```diff
@@ -0,0 +1,67 @@
...
+const Items = styled('div')({
+  color: PALETTE.SLATE_600,
+  fontSize: 16,
+  fontWeight: 600,
+  fontStyle: 'italic',
+  padding: '0px 24px 0px 4px',
+  width: '100%',
+  whiteSpace: 'nowrap',
+  overflow: 'hidden',
+  textOverflow: 'ellipsis'
```
nickoferrall, line 31: I went with truncating & using an ellipsis if the "Current filters" text is too long. We discussed in the issue showing three filters and showing "and x more" if there's no room to show all filters, but three filters could take up half a line or several lines. This approach is also easy to implement and requires less code.

packages/client/components/JiraScopingSearchCurrentFilters.tsx
```diff
@@ -0,0 +1,83 @@
...
+  )
+  const {jiraSearchQuery} = meeting
+  const {viewerMeetingMember} = meeting
+  const {teamMember} = viewerMeetingMember!
+  const {integrations} = teamMember
+  const {atlassian} = integrations
+  const {projects} = atlassian!
+  const {projectKeyFilters} = jiraSearchQuery
+  const projectFilterNames = projectKeyFilters.map((projectId, idx) => {
+    const project = projects.find(({id}) => id === projectId)!
```
nickoferrall, line 70: This has a worst-case performance of O(n*m). Still, I decided not to memorize it because the component tree structure means that this component usually only re-renders when `projectKeyFilters` changes anyway. Also, in practice, `projectKeyFilters` will be <10 and the total number of projects will normally be <100.

packages/client/components/GitHubScopingSearchBar.tsx
```diff
@@ -32,9 +42,11 @@ const GitHubScopingSearchBar = (props: Props) => {
 
   return (
     <SearchBar>
-      <GitHubScopingSearchHistoryToggle meetingRef={meeting} />
-      <GitHubScopingSearchInput meetingRef={meeting} />
-      <GitHubScopingSearchFilterToggle meetingRef={meeting} />
+      <SearchBarWrapper>
```
nickoferrall, line 45: We're reusing a lot of the same code in the Poker scope search bars so I considered creating reusable components. However, I decided not to because the GitHub integration doesn't use the CurrentFilters component, `ScopingSearchHistoryToggle` doesn't exist in the Parabol or GitLab integration, and the data we're getting in each of the components differs quite a bit.

packages/server/graphql/types/GitLabSearchQuery.ts
```diff
@@ -14,11 +15,10 @@ const GitLabSearchQuery = new GraphQLObjectType<any, GQLContext>({
       type: new GraphQLNonNull(GraphQLString),
       description: 'The query string used to search GitLab issue titles and descriptions'
     },
-    selectedProjectsIds: {
-      type: new GraphQLList(new GraphQLNonNull(GraphQLID)),
+    selectedProjects: {
```
nickoferrall, line 18: In the "Current Filters", we want to display a list of the selected GitLab project `fullPath`s: 

image

There are a few ways that we could get the projects `fullPath`s:

1. Current approach: in the `GitLabSearchQuery`, we can change `selectedProjectIds` to `selectedProjects`, which contains the `id`s and `fullPath`s for each project. This is easy to query on the frontend, but in GitLab v2, when we store persisted queries in the db, we'll need to store an extra field, `fullPath`, in the db. 
2. In [GitLabIntegration](https://github.com/ParabolInc/parabol/blob/f39ccbb0bc4fa3b680668c74985ff6e36f060140/packages/server/graphql/types/GitLabIntegration.ts), we could store a `projects` field. In the resolver, we could query projects, in a similar way to how we query an issue in Task ([see here](https://github.com/ParabolInc/parabol/blob/4712bec016a670388622caae3f8653f3b1f74f69/packages/server/graphql/types/Task.ts#L217)). The downside is we would be sending a query to GitLab to get all projects here, in the projects filter menu, and in the create new issue dropdown menu. 
3. Like the [AtlassianIntegration](https://github.com/ParabolInc/parabol/blob/072cdfcbe5cf1a5eed3a9c72ec1da51756d2b9fe/packages/server/graphql/types/AtlassianIntegration.ts#L163), we could create a GitLab custom dataLoader that grabs all projects. The custom dataLoader would be used here, the projects filter menu, and in the create new issue dropdown menu. As the dataLoader caches the results, we wouldn't need to reach out to GitLab each time, which would make the integration feel more snappy.

I'm thinking the third approach is best, but I want to get a second opinion before implementing it. Perhaps there's a fourth option that's even better 🤷‍♂️

nickoferrall:
Hey @enriquesanchez, this review implements the design you crafted in https://github.com/ParabolInc/parabol/issues/6138. Would you mind reviewing the design please?

packages/server/graphql/types/GitLabSearchQuery.ts
mattkrick: Storing `fullPath` isn't great because that value is mutable, ie GitLab users can rename their repo. We should try to only store immutable keys that we can exchange for fresh data from the integration, even if that means an extra round trip, it's better than stale data! (I used to store our github repo name in the DB & then we changed parabolinc/action to parabolinc/parabol & it was stale forever)

Option 3 wouldn't work because a dataloader only caches per-request & the filter & create new issue dropdown are separate requests. You could cache the projects in redis if you were concerned about latency, but then you'd have to verify that the value in redis has all the fields that you require (e.g. if you cached `{id, title}` in redis and then a 2nd request tries to get `{id, fullPath}`, you'd need a way to tell it to go back out to gitlab.

Option 2 is actually more efficient than you might think. You can use relay's `store-or-network` mode to only trigger a fetch if there is missing data. So as long as your 2nd query is a subset of your first query, there won't be a 2nd network request.

nickoferrall: That all makes sense, thanks! I was thinking that as `fullPath` is unique, it essentially acts as an id, but you're right, users can just rename it 😄 

I've pushed an update that implements option 2 👍

packages/client/components/GitLabScopingSearchResults.tsx
```diff
@@ -42,27 +31,6 @@ const GitLabScopingSearchResults = (props: Props) => {
...
-    graphql`
-      fragment GitLabScopingSearchResults_query on Query
-        @argumentDefinitions(
-          projectsFirst: {type: "Int", defaultValue: 75}
-          issuesFirst: {type: "Int", defaultValue: 25}
-          projectsAfter: {type: "String"}
-          issuesAfter: {type: "String"}
-          projectIds: {type: "[ID!]", defaultValue: null}
-        )
-        @refetchable(queryName: "GitLabScopingSearchResultsPaginationQuery") {
```
nickoferrall, line 65: This is no longer a refetchable query as we're now just getting a large number of issues and letting the user filter them with the search and menu. This will be refactored in this issue: https://github.com/ParabolInc/parabol/issues/6325

packages/client/components/NewGitLabIssueInput.tsx
```diff
@@ -127,16 +126,13 @@ const NewGitLabIssueInput = (props: Props) => {
...
                 }
-                query {
-                  # use alias otherwise projects args must be the same as GitLabScopingSearchResults query
-                  allProjects: projects(
-                    membership: true
-                    first: 100
-                    sort: "latest_activity_desc"
-                  ) {
+                # use alias to tell relay that this query shouldn't be cached with GitLabScopingSearchResults query
+                newIssueQuery: query {
```
nickoferrall, line 130: I ran into the same problem where adding the `selectedProjectsIds` filter to `GitLabScopingSearchResults` resulted in the new issue input query getting filtered when we don't want it to. Changing the `projects` args didn't work, but aliasing the query did.

nickoferrall:
Hey @igorlesnenko, would you mind doing a code review for this PR while @enriquesanchez reviews the design?

enriquesanchez (changes requested):
@nickoferrall This looks good! 🙌

Tested on FF 98.0.1 (macOS). Here are a couple of screenshots:

Captura de Pantalla 2022-04-06 a la(s) 12 07 39 p m

...and testing with a lot of filters:

Captura de Pantalla 2022-04-06 a la(s) 12 07 56 p m

I like that you truncate the text string 👍

The only change I suggest is to have the filter button have no background when there are no active filters, and only change the bg to blue when one or more are active:

Captura de Pantalla 2022-04-06 a la(s) 12 14 45 p m

I know you wanted the filter icon to stand out, so if this causes tension, we can bump it up a little with a gray bg:

Captura de Pantalla 2022-04-06 a la(s) 12 16 42 p m

(I'm using `Slate 200 #FF0FA` in the mockup above ☝️)

What do you think?


Questions: 

1. Does `Current filters` feel obvious to you? Should we say instead `Current projects`? Or something like that?

nickoferrall:
> The only change I suggest is to have the filter button have no background when there are no active filters

My thinking is it's _more_ important that the filter stands out when there are no filters. My concern atm is users may be unaware that the filter button exists. This is particularly concerning with the GitLab integration where, in v1, scrolling down won't load more issues. 

To make sure that the user is immediately aware of the filter button, I'd prefer to stick with the blue background, or a colour that contrasts more than slate 200. What do you think?

> Should we say instead Current projects?

Yeah, we can change it to current projects! My thinking was we want to keep it consistent across the integrations. Jira and GitLab refer to them as projects, but the Parabol integration uses status filters. Subsequent integrations will probably have different terms too. But, as you mentioned in a previous issue, users will probably only use one integration, so why not make the language as integration specific as possible.

enriquesanchez:
> I'd prefer to stick with the blue background, or a colour that contrasts more than slate 200. What do you think?

That's reasonable and super safe to try. I would still like to revisit this in the near future. [We are currently exploring filtering in other parts of the app](https://github.com/ParabolInc/parabol/issues/6070#issuecomment-1089362659) and I would prefer if we standardize some of these components and patterns 👍

> as you mentioned in a previous issue, users will probably only use one integration, so why not make the language as integration specific as possible.

I'm unfamiliar with how much extra work or code this represents. So I defer to you. My concern is to make it obvious that these are their projects they are filtering, but I guess by the name themselves users should be aware of that 🤔

I'm also uncertain if `Current projects` makes sense from an english grammar stand point. Maybe it's `Filtering projects:`?

nickoferrall:
> My concern is to make it obvious that these are their projects they are filtering, but I guess by the name themselves users should be aware of that 🤔 I'm also uncertain if Current projects makes sense from an english grammar stand point. Maybe it's Filtering projects:?

Yeah, you're right, it could be unclear that `Current projects` relates to the filters. I think `Current filters` works quite well tbh. I'll press on with the current solution unless you have any tension. 

@igorlesnenko this PR is ready for your code review when you get the chance 🙏

packages/client/components/GitLabScopingSearchCurrentFilters.tsx
```diff
@@ -0,0 +1,86 @@
...
+    `,
+    meetingRef
+  )
+  const {gitlabSearchQuery} = meeting
+  const {selectedProjectsIds} = gitlabSearchQuery
+  const {viewerMeetingMember} = meeting
+  const {teamMember} = viewerMeetingMember!
+  const {integrations} = teamMember
+  const {gitlab} = integrations
+  const {projects} = gitlab
```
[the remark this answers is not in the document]
nickoferrall: I think you're right - optional chaining is the cleaner approach. Update pushed!

igorlesnenko (approved):

packages/client/components/GitLabScopingSearchCurrentFilters.tsx
```diff
@@ -0,0 +1,86 @@
...
+        }
+      }
+    `,
+    meetingRef
+  )
+  const {gitlabSearchQuery, viewerMeetingMember} = meeting
+  const {selectedProjectsIds} = gitlabSearchQuery
+  const projects = viewerMeetingMember?.teamMember.integrations.gitlab.projects
+
+  const selectedProjectsNames = [] as string[]
```
mattkrick, line 68: -1 little messy here, can simplify to

```ts
  const selectedProjectsNames = projects
    .filter((project) => selectedProjectIds.contains(project.id))
    .map(({fullPath}) => fullPath)
  const currentFilters = selectedProjectsNames.length === 0 ? 'None' : selectedProjectsNames.join(', ')
```

packages/client/components/JiraScopingSearchCurrentFilters.tsx
```diff
@@ -0,0 +1,83 @@
...
+          }
+        }
+      }
+    `,
+    meetingRef
+  )
+  const {jiraSearchQuery, viewerMeetingMember} = meeting
+  const {projectKeyFilters} = jiraSearchQuery
+  const projects = viewerMeetingMember?.teamMember.integrations.atlassian?.projects
+  const projectFilterNames = [] as string[]
```
mattkrick, line 65: -1 see same pattern in above comment

packages/client/components/ParabolScopingSearchCurrentFilters.tsx
```diff
@@ -0,0 +1,65 @@
...
+        parabolSearchQuery {
+          statusFilters
+        }
+      }
+    `,
+    meetingRef
+  )
+  const {parabolSearchQuery} = meeting
+  const {statusFilters} = parabolSearchQuery
+  const formattedStatusFilters = statusFilters?.map((status, idx) =>
```
mattkrick, line 52: -1 the `join` method is your friend 🙂

mattkrick (changes requested):
very minor fix, other than that good to go!

packages/client/components/GitLabScopingSearchCurrentFilters.tsx
nickoferrall: Nice, the `join` method is cleaner. I've updated the three filters to include that.

However, I think we want to loop through `selectedProjectIds` instead of projects so that newly selected project filters are appended to the list. 

For example, if `const projects = ['DEV', 'DESIGN', 'TEST']`, and the text says `Current filters: DESIGN`. If the user then also selects the `DEV` project, I'd expect the text to update to `Current filters: DESIGN, DEV`. 

So, I've kept some of the old `forEach` and `find` logic.

Issue #6200, opened 2022-03-15
Standup UI - Static, non-editable prompt


BartoszJarocki:
Design - https://www.figma.com/file/bhiEOV4HK9Kv91HFsJmevV/Stand-up-Concepts?node-id=1138%3A72617

AC
- [ ] There’s a static prompt in the center at the top of the main view. 
- [ ] It’s not editable yet

BartoszJarocki:
**Story Points: 3**
  [See the discussion](https://action.parabol.co/meet/chorM4RmGA/estimate/3) in PS2 Poker 1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

BartoszJarocki:
**Include?: Y**
  [See the discussion](https://action.parabol.co/meet/chorM4RmGA/estimate/4) in PS2 Poker 1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

PR #6353, opened 2022-04-06
feat(standups): Responses grid with static prompt

### LINKED ISSUE #6201: Standup UI - Responses grid
Design - https://www.figma.com/file/bhiEOV4HK9Kv91HFsJmevV/Stand-up-Concepts?node-id=1138%3A72617
Design issue - https://github.com/ParabolInc/parabol/issues/5965 

Questions
 - How does it scale? is it the same as meetings view?

AC
- [ ] Grid with placeholders components is added

BartoszJarocki:
Fixes #6201 

For now, we'll use the same grid as we use on the meeting dashboard. In next iterations we'll improve and implement #6389.
Earlier discussion for the reference: https://github.com/ParabolInc/parabol/issues/5965

This PR also adds https://github.com/ParabolInc/parabol/issues/6200 as it was a little change.


BartoszJarocki:
@jmtaber129 would you mind taking a look? thanks!

packages/client/components/TeamPromptMeeting.tsx
```diff
@@ -22,23 +86,80 @@ const TeamPromptMeeting = (props: Props) => {
...
+              <ErrorBoundary>
+                <ResponsesGridContainer maybeTabletPlus={maybeTabletPlus}>
+                  <ResponsesGrid>
+                    {transitioningTeamMembers.map((teamMember) => {
+                      const {child, onTransitionEnd, status} = teamMember
+                      const {id, picture, preferredName} = child
+
+                      return (
+                        <TeamMemberResponse
+                          key={id}
```
jmtaber129, line 148: +1 Would it make sense to use `child.key` here?  That would guarantee consistency between the `key` used by `useTransition` and the `key` used by React.

packages/client/components/TeamPromptMeeting.tsx
```diff
@@ -22,23 +86,80 @@ const TeamPromptMeeting = (props: Props) => {
     graphql`
       fragment TeamPromptMeeting_meeting on TeamPromptMeeting {
         ...TeamPromptTopBar_meeting
+        id
+        ...useMeeting_meeting
+        phases {
+          id
+          phaseType
```
jmtaber129, line 93: +1 `id` and `phaseType` don't seem to be used anywhere

jmtaber129 (approved):

packages/client/components/TeamPromptMeeting.tsx
```diff
@@ -1,16 +1,80 @@
...
+})
+
+const ResponseHeader = styled('div')({
+  display: 'flex',
+  flexDirection: 'row',
+  alignItems: 'center',
+  padding: '0 8px'
+})
+
+const TeamMemmberName = styled('h3')({
```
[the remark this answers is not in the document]
BartoszJarocki: 🤦

packages/client/components/TeamPromptMeeting.tsx
```diff
@@ -22,23 +86,80 @@ const TeamPromptMeeting = (props: Props) => {
...
       }
     `,
     meetingRef
   )
 
+  const maybeTabletPlus = useBreakpoint(Breakpoint.FUZZY_TABLET)
+  const teamMembers = useMemo(() => {
+    return meeting.phases
+      .flatMap((phase) => phase.stages)
+      .flatMap((stage) => stage.teamMember)
```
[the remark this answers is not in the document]
BartoszJarocki: yeah, I'm pretty sure it'll need to be adjusted to the card component anyway so I didn't want to guess. if you don't mind, I'll leave it as is

BartoszJarocki:
@jmtaber129 I fixed every comment except for the mapping which I think should be aligned in the PR related to cards. thank you for the review, moving on to the maintainer review 🙏

BartoszJarocki:
@mattkrick I'd be great if you could take a look 🙏 thank you!

packages/client/components/TeamPromptMeeting.tsx
```diff
@@ -22,23 +86,76 @@ const TeamPromptMeeting = (props: Props) => {
     graphql`
       fragment TeamPromptMeeting_meeting on TeamPromptMeeting {
         ...TeamPromptTopBar_meeting
+        ...useMeeting_meeting
+        phases {
+          stages {
```
mattkrick, line 91: -1 fragment on `TeamPromptResponsesPhase` instead. it's a little more expressive way to say "we only care about this 1 phase in particular"

as it's written, it's saying, "look at EVERY phase & for EVERY phase, give me the TeamPromptResponsesStages"

Also, if you fragment on the phase, then the teamMember object is guaranteed non null, which is nice.

packages/client/components/TeamPromptMeeting.tsx
```diff
@@ -22,23 +86,76 @@ const TeamPromptMeeting = (props: Props) => {
...
+            ... on TeamPromptResponseStage {
```
mattkrick, line 92: -1 Could you please rename `TeamPromptResponsesPhase` to `TeamPromptResponsePhase`? (notice  the missing "s"). not a problem in this PR, but i didn't discover it until now

packages/client/components/TeamPromptMeeting.tsx
```diff
@@ -22,23 +86,76 @@ const TeamPromptMeeting = (props: Props) => {
...
+          return null
+        }
+
+        return {
+          ...teamMember,
+          key: teamMember.id
+        }
+      })
+      .filter(isNotNull)
+  }, [meeting.phases])
```
mattkrick, line 122: -1 with the changes above, you can simplify this to:

```ts
 const responsePhase = meeting.phases.find((phase) => phase.__typename === 'TeamPromptResponsePhase')!
 const teamMembers responsePhase.stages.map((stage) => {...stage.teamMember, key: stage.teamMember.id}))
```

mattkrick (changes requested):
some very minor client-side stuff, after that feel free to merge!

packages/client/components/TeamPromptMeeting.tsx
BartoszJarocki: typescript is a little dumb in that case and `responsePhase` can be another type as well.
see the screenshot below
Screenshot 2022-04-19 at 15 35 16

do we already have some generic utility to generate type predicates? we'd need something like this and I just want to make sure I haven't missed anything in the codebase 
https://stackoverflow.com/questions/59050071/narrow-return-type-of-find-from-discriminated-union-array

packages/client/components/TeamPromptMeeting.tsx
BartoszJarocki: no strong opinions here, the origin of the name is the fact that this phase is for all responses, not one. let me know if that makes sense. if you think it's still better to use `TeamPromptResponsePhase` I will fix

packages/client/components/TeamPromptMeeting.tsx
BartoszJarocki: oh cool, will fix

packages/client/components/TeamPromptMeeting.tsx
mattkrick: ah gotcha, that works!

packages/client/components/TeamPromptMeeting.tsx
mattkrick: ah good call, we have `GQLType`, which does the trick. I'll write a little commit to showcase it

packages/client/components/TeamPromptMeeting.tsx
```diff
@@ -102,24 +103,15 @@ const TeamPromptMeeting = (props: Props) => {
...
-        }
 
-        return {
-          ...teamMember,
-          key: teamMember.id
-        }
-      })
-      .filter(isNotNull)
-  }, [meeting.phases])
+  const phase = getPhaseByTypename(phases, 'TeamPromptResponsesPhase')
```
mattkrick, line 109: @BartoszJarocki let me know if this gets you whatcha need. if so, go ahead & merge this PR & we'll start using this more in the future

BartoszJarocki: yup, looks good! thank you!

PR #6355, opened 2022-04-06
Feat/6244/gitlab visible in estimate phase

### LINKED ISSUE #6244: Sprint Poker: GitLab v1: Issue is visible in Estimate phase
After adding a GitLab issue to the Estimate phase, the user should be able to see the issue in a card at the top of the selected Estimate stage. See GitHub example:

gh-estimate

### AC

- [ ] GitLab issue is visible in the Estimate phase

nickoferrall:
Fixes #6244


packages/client/components/PokerEstimateHeaderCard.tsx
```diff
@@ -4,12 +4,58 @@ import {createFragmentContainer} from 'react-relay'
...
+        descriptionHTML: bodyHTML,
+        url,
+        linkTitle: `GitHub Issue #${number}`,
+        linkText: `#${number}`
+      }
+    case '_xGitLabIssue':
+      const {iid, title, descriptionHtml, webUrl} = integration
+      return {
+        cardTitle: title,
+        descriptionHTML: descriptionHtml || '',
```
igorlesnenko, line 50: Is there a reason to use `||` instead of `??` ?

packages/client/components/PokerEstimateHeaderCard.tsx
```diff
@@ -4,12 +4,58 @@ import {createFragmentContainer} from 'react-relay'
...
+  url: string
+  linkTitle: string
+  linkText: string
+}
+
+const getHeaderFields = (integration: Integration | null): IntegrationHeaderFields | null => {
+  if (!integration) return null
+  const {__typename} = integration
+  switch (__typename) {
+    case 'JiraServerIssue':
```
igorlesnenko, line 26: I think we would need to do something in the future to not have a lot of mappings like this. Created a separate  discussion issue: #6374

igorlesnenko (approved):
Looks good to me

packages/client/components/PokerEstimateHeaderCard.tsx
nickoferrall: We could use either one. `||` will use the right-hand side of the operation if the left-hand side is falsy, e.g. `0`, whereas `??` will only do this if it is null or undefined. Here, we're checking whether `descriptionHtml` is a string or null, so `??` is probably more descriptive. I've pushed an update changing this 👍

packages/client/components/PokerEstimateHeaderCard.tsx
```diff
@@ -4,12 +4,58 @@ import {createFragmentContainer} from 'react-relay'
...
 import PokerEstimateHeaderCardContent from './PokerEstimateHeaderCardContent'
 import PokerEstimateHeaderCardError from './PokerEstimateHeaderCardError'
-import PokerEstimateHeaderCardGitHub from './PokerEstimateHeaderCardGitHub'
 import PokerEstimateHeaderCardParabol from './PokerEstimateHeaderCardParabol'
 
 interface Props {
   stage: PokerEstimateHeaderCard_stage
 }
+
+type Integration = NonNullable<NonNullable<PokerEstimateHeaderCard_stage['task']>['integration']>
```
Dschoordsch, line 13: +1 The outer `NunNullable<>` is not necessary since you use the type later with `Integration | null`. For me it also doesn't add to the readability since I rely on hover types to know what's the final type.

packages/client/components/PokerEstimateHeaderCard.tsx
```diff
@@ -23,26 +69,21 @@ const PokerEstimateHeaderCard = (props: Props) => {
...
-  }
-  return null
+  const {cardTitle, descriptionHTML, url, linkTitle, linkText} = headerFields
+  return (
+    <PokerEstimateHeaderCardContent
+      cardTitle={cardTitle}
+      descriptionHTML={descriptionHTML}
+      url={url}
+      linkTitle={linkTitle}
+      linkText={linkText}
```
Dschoordsch, line 84: +1 Especially if you move the `IntegrationHeaderFields` type over to `PokerEstimateHeaderCardContent` and rename it to `PokerEstimateHeaderCardContentProps` you could just spread them here
```
<PokerEstimateHeaderCardContent {...headerFields}/>
```

Dschoordsch (changes requested):
Tested, works

packages/client/components/PokerEstimateHeaderCard.tsx
nickoferrall: That's much cleaner. Update pushed!

Dschoordsch (approved):

PR #6361, opened 2022-04-06
fix(gitlab): add proper client-side alias handling

mattkrick:
This fixes that problem where fullPath was required on the client side.
adding fullPath to the client didn't actually fix the problem, it just masked it.
full loom here: https://www.loom.com/share/79e4d705572c40bdba449b12bc3250b7


mattkrick:
gonna merge this in just to keep the PR review going along...

PR #6363, opened 2022-04-06
feat(gitlab): bump nest-graphql-endpoint

### LINKED ISSUE #6362: fix(nestedEndpoints): Add default resolver that looks for the alias first
The client should be able to alias GraphQL fields without the developer having to modify the server.

This is a VERY sharp edge in development. Frontend developers shouldn't have to know the quirks of nested graphql endpoints to do their job.

AC
- [ ] `resolveToFieldNameOrAlias` should be the default resolver for every single field in github & gitlab schemas. I assume this means iterating through every single resolver & replacing the default resolver with this one.
- [ ] it looks like we can pass in a default one here https://github.com/mattkrick/graphql-tools/blob/61c5221f8feed66b2e4a3de01bc85ee4b4893b09/packages/schema/src/addResolversToSchema.ts#L52. Maybe we use that API, but I fear that we can't do that before `mergeSchemas` already gives us default resolvers. maybe we put that logic in nestGraphQLEndpoint?

Estimate: 3 hours

mattkrick:
fix #6362
The latest version now supports client-side alias without having to touch the server (see deleted file to see what we no longer have to do).

@nickoferrall with this merge we can get rid of all the instances of `resolveToFieldNameOrAlias`


nickoferrall (approved):
🙌

PR #6366, opened 2022-04-07
refactor: Use AvatarList component in DashboardAvatars

### LINKED ISSUE #4493: Update overflow chit to start at +2 (like we do in poker)
​

shapel:
# Description

Fixes #4493

## Demo

## Testing scenarios

- [ ] Scenario A
  - Step 1
  - Step 2...

- [ ] Scenario B
  - Step 1
  - Step 2....

## Final checklist


packages/client/hooks/useOverflowAvatars.ts
```diff
@@ -25,7 +25,7 @@ const useOverflowAvatars = <T extends {id: string}>(
     if (!el) return
     const {clientWidth: totalWidth} = el
     const lappedAvatarWidth = avatarWidth - avatarOverlap
-    const maxAvatars = Math.floor((totalWidth - avatarWidth) / lappedAvatarWidth)
+    const maxAvatars = Math.floor((totalWidth - avatarWidth) / lappedAvatarWidth) + 1
```
shapel, line 28: added + 1 due to fact that by subtracting `avatarWidth` we already assume that we have at least one full avatar

packages/client/components/Avatar/Avatar.tsx
```diff
@@ -1,13 +1,24 @@
...
+  Props,
+  'sansRadius' | 'sansShadow' | 'picture' | 'size' | 'onClick' | 'isConnected'
+>
 
 const ImageBlock = styled('div')<ImageBlockProps>(
-  ({sansRadius, sansShadow, picture, size, onClick}) => ({
-    backgroundImage: `url(${picture}), url(${defaultUserAvatar})`,
+  ({isConnected, sansRadius, sansShadow, picture, size, onClick}) => ({
+    backgroundImage: `${
+      isConnected || isConnected === undefined
```
shapel, line 14: Copy-paste logic from DashboardAvatar related to 'opacity' for not connected users + backward compatibility for handling components not provided `isConnected` prop  (`undefined`).

packages/client/components/AvatarList.tsx
```diff
@@ -29,10 +29,10 @@ const widthToOverlap = {
 // hard coding for now until we have
 // a better solution for handling gutters
 // for a component that may not be there
-const sizeToHeightBump = {
+export const sizeToHeightBump = Object.freeze({
```
shapel, line 32: explicitly use this hack in other place (to be sure don't forget to change in case of refactoring)

packages/client/components/AvatarListUser.tsx
```diff
@@ -83,6 +82,8 @@ const AvatarListUser = (props: Props) => {
         isAnimated={isAnimated}
         borderColor={borderColor}
         width={width}
+        isConnected={!!isConnected}
+        onClick={onClick}
```
shapel, line 86: moved `onClick` to Avatar so it can change pointer style based on this

packages/client/components/OverflowAvatar.tsx
```diff
@@ -55,30 +57,23 @@ interface Props {
...
-    <Wrapper offset={offset} onClick={onClick}>
+    <Wrapper offset={offset}>
       <OverflowCount
         width={width}
         ref={ref}
         status={status}
         onTransitionEnd={onTransitionEnd}
         isAnimated={isAnimated}
         borderColor={borderColor}
+        onClick={onClick}
```
shapel, line 74: moved `onClick` to OverflowCount so it can change pointer style based on this

packages/client/components/OverflowAvatar.tsx
```diff
@@ -55,30 +57,23 @@ interface Props {
...
       >
-        {label}
+        +{label}
```
shapel, line 76: added `+` before overflow count (as DashboardAvatars did)
It seems to me more understandable from a UX point of view but I'm not sure about the current guideline.

packages/client/components/DashboardAvatars/DashboardAvatars.tsx
```diff
@@ -2,56 +2,34 @@ import styled from '@emotion/styled'
...
-import ErrorBoundary from '../ErrorBoundary'
+import AvatarList, {sizeToHeightBump} from '../AvatarList'
 import PlainButton from '../PlainButton/PlainButton'
-import DashboardAvatar from './DashboardAvatar'
 
-const AvatarsList = styled('div')<{isDesktop: boolean}>(({isDesktop}) => ({
+const Wrapper = styled('div')({
   display: 'flex',
   flexDirection: 'column',
-  marginRight: 6,
```
shapel, line 20: `marginRight` & `left` usage looks deprecated (tested on desktop and mobile view)

shapel:
👋 @mattkrick just ping you to be sure that this is on you list 🤞

shapel:
@nickoferrall maybe you can review my PR 🙏

shapel:
👋 @mattkrick @nickoferrall plz pay attention to this PR or suggest someone to review 🤞

nickoferrall:
No problem, @shapel. I can review this one 👍

packages/client/components/AvatarList.tsx
nickoferrall: Why use `Object.freeze` here? What problem does this solve?

packages/client/components/DashboardAvatars/DashboardAvatars.tsx
```diff
@@ -64,93 +42,104 @@ const StyledButton = styled(PlainButton)({
...
+  },
+  marginTop: sizeToHeightBump[ElementWidth.DASHBOARD_AVATAR] * -1 // coupling with AvatarList styles
 })
 
 interface Props {
   team: DashboardAvatars_team
 }
 
-type Avatar = DashboardAvatars_team['teamMembers'][0]
+type Avatar = DashboardAvatars_team['teamMembers'][0]['user']
```
nickoferrall, line 53: -1 this type is now the user rather than the avatar so we should update the name. We also need to update the names in the logic below to reflect this

nickoferrall (changes requested):
It looks like it's not working as expected. When I have 5 team members and I load the page, It only shows two avatars with a +3 icon when it should show all 5. I'm using Chrome on a large monitor. Let me know if you have any difficulty reproducing it. 

dashboard-bug

packages/client/components/AvatarList.tsx
shapel: I used `Object.freeze` as an explicit intent that this object should not be modified in any module that imports it. 
But you're right, we can use a compile-time restriction instead of a run-time limit.

shapel:
In the beginning, I misunderstood AC a bit. So @nickoferrall  you're absolutely right that it didn't work.

Let me describe some of the challenges of this PR.
The tricky part here is that we need to have good balance for the avatars area:
- we want to see as many avatars as possible
- we want the design to stay the same (title "Manage Team" under avatars should be centered with them)

Since the AvatarList component heavily depends on the current width of container, we need to somehow "set" a specific size on it.
To do this I've changed layout a bit. The avatars area will try to grow on big screen sizes.  As for small screen sizes, this is debatable.
Screenshot 2022-04-20 at 15 13 48

We can leave the avatars area to grow on small screen sized as well (its' looks good in my opinion) but this will change  the current design.
Screenshot 2022-04-20 at 15 15 22
Screenshot 2022-04-20 at 15 15 52

Otherwise we may only have a small number of avatars to display (so keep current design). It's still not perfect because it displays only fixed set of avatars. 
Screenshot 2022-04-20 at 15 14 23
Screenshot 2022-04-20 at 15 17 44


But I had a hard time finding a better solution to make the area grow, but not too much. 
I'm more than happy to hear some idea how to handle this differently!

packages/client/components/AvatarList.tsx
```diff
@@ -42,21 +46,30 @@ interface Props {
   emptyEl?: ReactElement
   isAnimated?: boolean
   borderColor?: string
+  // the current container may not be sized so that we can observe some parent container
+  containerRef?: React.RefObject<HTMLDivElement>
```
shapel, line 50: We need to somehow be able to watch some parent container size, because the HTML layout may not guarantee that the current container will be an adequate size. 
I can't find a better solution for this. 🤷🏻‍♂️

nickoferrall:
Thanks for letting me know @shapel and for pushing the update! Would you mind sharing some repro steps so that I can see the bug that this is fixing?

shapel:
@nickoferrall  I'm not sure I fully understood your question.
If we talk about https://github.com/ParabolInc/parabol/issues/4493, then the point is that the system should only show the overflow chit if there are avatars that don't fit on the screen.

without fix:
Screenshot 2022-04-26 at 10 05 22

with fix:
Screenshot 2022-04-26 at 10 02 16

packages/client/modules/teamDashboard/components/TeamTasksHeader/TeamTasksHeader.tsx
```diff
@@ -94,7 +95,8 @@ const Avatars = styled('div')({
   display: 'flex',
   paddingTop: 12,
   [desktopBreakpoint]: {
-    paddingTop: 0
+    paddingTop: 0,
+    flexGrow: 1
```
nickoferrall, line 99: -1 it looks like this fixes the issue I mentioned in my earlier comment where only a few avatars would show up on desktop, but on mobile/tablet, I only see three avatars (including overflow) when there's space for many more

packages/client/components/DashboardAvatars/DashboardAvatars.tsx
```diff
@@ -64,93 +54,107 @@ const StyledButton = styled(PlainButton)({
   WebkitTapHighlightColor: 'transparent',
   '&:hover': {
     cursor: 'pointer'
-  }
+  },
+  marginTop: sizeToHeightBump[ElementWidth.DASHBOARD_AVATAR] * -1 // coupling with AvatarList styles
```
nickoferrall, line 58: +1 if it needs to be a negative margin, I'd prefer to put a `-sizeToHeightBump[ElementWidth.DASHBOARD_AVATAR]` as the multiplication implies a calculation is required. 

I haven't dug into the calculations in this PR review yet, but in general, I'd love to minimise the number of calculations in this PR. They may all be necessary, but there's a lot of calculations for marginTop, minHeight, minWidth. If we can use the box model as much as possible so that we don't need to define the exact heights * margins, we'll save ourselves plenty of pain down the line.

nickoferrall (changes requested):
Thanks for sharing that example. I've just reviewed the code and I noticed that when I switch between teams, the avatars wouldn't update. See gif: 

switching-teams-bug

packages/client/components/AvatarList.tsx
```diff
@@ -9,30 +9,34 @@ import {AvatarList_users} from '../__generated__/AvatarList_users.graphql'
...
+    // setting the min-width so that the current container has some size
+    minWidth
+  })
+)
 
 const widthToOverlap = {
   28: 8,
   46: 16
-}
+} as const
```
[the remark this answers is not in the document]
shapel: According to https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-4.html#const-assertions this construct mark that object literals get readonly properties. 
So we have a compile-time check that we're not changing a property of that object anywhere (especially true for things we're exporting from a module).

packages/client/components/AvatarList.tsx
```diff
@@ -9,30 +9,34 @@ import {AvatarList_users} from '../__generated__/AvatarList_users.graphql'
...
 
 // hard coding for now until we have
 // a better solution for handling gutters
 // for a component that may not be there
-const sizeToHeightBump = {
+export const sizeToHeightBump = {
   28: 4,
   46: 0
-}
+} as const
```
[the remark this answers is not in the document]
shapel: see 👆

packages/client/components/AvatarList.tsx
```diff
@@ -42,21 +46,30 @@ interface Props {
...
+    size,
+    overlap
+  )
   const showAnimated = isAnimated ?? true
   const activeTChildren = transitionChildren.filter(
     (child) => child.status !== TransitionStatus.EXITING
   )
   const minHeight = activeTChildren.length === 0 ? 0 : size + sizeToHeightBump[size]
+  const minWidth =
+    activeTChildren.length === 0 ? 0 : (activeTChildren.length - 1) * offsetSize + size
```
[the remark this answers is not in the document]
shapel: The logic here is that there can only be one full-size avatar in the system (real avatar or overflow chit), all others are located with overlapping each other (size - overlap). 
Thus, `offsetSize` is used here only as a memorized calculation (size - overlap).

I see the name `offsetSize` being a bit confusing here, but I couldn't figure out a better to use for calculating the offset and memorising the non-overlapping size.

packages/client/modules/teamDashboard/components/TeamTasksHeader/TeamTasksHeader.tsx
shapel: You are absolutely right. I mentioned here https://github.com/ParabolInc/parabol/pull/6366#issuecomment-1103866909 that there is a controversial thing and we need to agree on how to deal with it.

packages/client/components/AvatarListUser.tsx
shapel: I've changed the logic to only change the cursor if an onClick handler is specified.

shapel:
@nickoferrall thanks for reviewing and finding the issue with switching between teams!

packages/client/components/DashboardAvatars/DashboardAvatars.tsx
shapel: I've removed calculation usage.

Regarding the number of calculations, I had a hard time finding a better solution.
I'm more than happy to hear some ideas on how to handle this differently!

shapel:
@nickoferrall just a friendly ping to make sure it's on your radar 🤞

mattkrick:
bumping this to reviewer review internally so it can get some 👀

nickoferrall:
Sorry for the slow review cycle @shapel. I'm working on a couple of integrations that are time urgent so I need to prioritise those for now. I've shared your PR with the team so either someone else will review this soon or I'll get back to it as quickly as I can

shapel:
Just ping to make sure 👆 is on your radar.

packages/client/components/Avatar/Avatar.tsx
```diff
@@ -1,19 +1,28 @@
...
+      backgroundImage: `linear-gradient(rgba(255,255,255,.5), rgba(255,255,255,.5)),
+    url(${picture}), url(${defaultUserAvatar})`
+    },
     backgroundPosition: 'center center',
     backgroundRepeat: 'no-repeat',
     backgroundSize: 'cover',
     borderRadius: sansRadius ? 0 : '100%',
     boxShadow: sansShadow ? 'none' : undefined,
-    cursor: onClick ? 'pointer' : 'default',
+    ...(onClick ? {cursor: 'pointer'} : null),
```
nickoferrall, line 25: +1 why set the cursor to null instead of default? I'd prefer to following the existing pattern of `cursor: onClick ? 'pointer' : 'default'`

packages/client/components/DashboardAvatars/DashboardAvatars.tsx
```diff
@@ -64,93 +54,107 @@ const StyledButton = styled(PlainButton)({
...
+        <StyledButton onClick={() => handleClick()}>Manage Team</StyledButton>
+      </Column>
+    </Row>
   )
 }
 
+graphql`
+  fragment DashboardAvatars_teamMember on TeamMember {
+    ...TeamMemberAvatarMenu_teamMember
+    ...LeaveTeamModal_teamMember
```
nickoferrall, line 133: -1 this fragment is repeating a lot of the logic below. Which fields do we need here?

packages/client/components/OverflowAvatar.tsx
```diff
@@ -40,7 +41,8 @@ const OverflowCount = styled('div')<{
     : 'scale(1)',
   transition: `all 300ms ${BezierCurve.DECELERATE}`,
   userSelect: 'none',
-  width
+  width,
+  ...(onClick ? {cursor: 'pointer'} : null)
```
nickoferrall, line 45: +1 same as the comment above. I'd prefer to stick with the existing cursor pattern if possible

packages/client/hooks/useTransition.ts
```diff
@@ -80,16 +80,19 @@ const useTransition = <T extends {key: Key}>(children: T[]) => {
     const mountingKeys = [] as Key[]
     children.forEach((nextChild) => {
       const idxInPrev = prevTChildren.findIndex(({child}) => child.key === nextChild.key)
-      const status = idxInPrev === -1 ? TransitionStatus.MOUNTED : prevTChildren[idxInPrev]!.status
+      const needToUpdateStatus =
+        idxInPrev === -1 || prevTChildren[idxInPrev]!.status === TransitionStatus.EXITING
```
nickoferrall, line 84: Could you provide repro steps so that I can see what bug this fixes? As `useTransition` is being used in many files, it would be great if we can solve this issue without updating `useTransition`

packages/client/components/DashboardAvatars/DashboardAvatars.tsx
```diff
@@ -64,93 +54,107 @@ const StyledButton = styled(PlainButton)({
...
+      atmosphere,
+      {teamId, teamDrawerType: 'manageTeam'},
+      {onError, onCompleted}
+    )
+    if (!userId) return
+    commitLocalUpdate(atmosphere, (store) => {
+      const viewer = store.getRoot().getLinkedRecord('viewer')
+      const teamMember = viewer?.getLinkedRecord('teamMember', {teamId})
+      if (!teamMember) return
+      const memberInFocus = teamMembers.find((teamMember) => teamMember.user.id === userId)
```
nickoferrall, line 104: -1 `teamMember` here refers to the `teamMember` in the `teamMembers` array, so we don't need to split up the `teamMember` and `memberInFocus` checks. Can revert the change, so: `if (!teamMember || !memberInFocus) return`

nickoferrall (changes requested):
Sorry for the slow review. Almost there now!

packages/client/components/Avatar/Avatar.tsx
shapel: I just thought that because `cursor` is an inherited CSS property, we might not want to override the value if the avatar doesn't have a click handler...

shapel: But I agree that it's may be now is not the right time to change cursor pattern, so I'm going to revert this change.

packages/client/components/OverflowAvatar.tsx
shapel: same as https://github.com/ParabolInc/parabol/pull/6366#discussion_r877857585

mattkrick:
looks like this'll need another commit or 2 before @nickoferrall can give another review.
moving this back to self-review

Issue #6303, opened 2022-03-29
Sprint Poker: GitLab v1: Track key events


nickoferrall:
To better understand how users are using the GitLab integration, we should track key events. 

See discussion for further context: https://github.com/ParabolInc/parabol/discussions/6296

For each event, we'll probably want to track the user id, timestamp, and meeting id at a minimum. 

### AC

- [ ] Track the start of the search. We could do this when they first enter a character rather than when they focus on the input as they could click on the input by accident.
- [ ] Track the end of the search. This could be when the user leaves the input. 
- [ ] Track when a user selects an issue in the Scope phase. Track the current search query string and the number of projects selected in the project filter.
- [ ] When a new issue is created.
- [ ] When a project filter menu item is selected. 
- [ ] When the select all button is used. 
- [ ] When a search input has been cleared. See the [discussion thread here](https://github.com/ParabolInc/parabol/discussions/6296#discussioncomment-2462151) for a scenario we might want to account for.
- [ ] Include a `trace_id` in the GitLab integration interface to track sessions.

PR #6367, opened 2022-04-07
feat(sprint-poker): Track GitLab events

nickoferrall:
Fixes https://github.com/ParabolInc/parabol/issues/6303

To test: 

- Copy the `SEGMENT_WRITE_KEY` and `SEGMENT_FN_KEY` from the production env vars in the devops repo. Paste them in your `.env` file
- Go to Amplitude and search for the user with your `viewerId`. Select on the user and see the GitLab events 
- Comment out the Segment env vars once you've finished testing

This PR adds event tracking in GitLab to the following:

- [x] The start of a search
- [x]  When a user selects an issue in the Scope phase
- [x]  When a project filter menu item is selected
- [x]  When the select all button is used
- [x]  When a search input has been cleared

This PR tracks when a project has been created: https://github.com/ParabolInc/parabol/pull/6332/


packages/client/components/GitHubScopingSearchInput.tsx
```diff
@@ -98,6 +99,7 @@ const GitHubScopingSearchInput = (props: Props) => {
   }
   const clearSearch = () => {
     setSearch(atmosphere, meetingId, '')
+    inputRef.current?.focus()
```
nickoferrall, line 102: After clearing a search input, I think we should keep the focus in the input as a user may want to continue searching

packages/client/components/GitHubScopingSelectAllIssues.tsx
```diff
@@ -86,7 +86,12 @@ const GitHubScopingSelectAllIssues = (props: Props) => {
       )
       return issue?.title ?? 'Unknown Story'
     })
-    UpdatePokerScopeMutation(atmosphere, variables, {onError, onCompleted, contents})
+    UpdatePokerScopeMutation(atmosphere, variables, {
+      onError,
+      onCompleted,
+      contents,
+      selectedAll: true
```
nickoferrall, line 93: To track whether a user has selected all issues, we could try to calculate this in the mutation, but it would be tricky. We'd need to query all of the issues from each integration using the same first arg. We'd then need to check which issues had been previously selected and what the `Threshold.MAX_POKER_STORIES` is. Then, if the number of issues added to the mutation is equal to that calculation, we'd know the selectAll button had been selected.

While I don't love passing an extra variable to the mutation, I think it's less error-prone and easier to maintain than the calculation solution mentioned above.

packages/client/utils/getSearchQueryFromMeeting.ts
```diff
@@ -0,0 +1,29 @@
+import {
+  TaskServiceEnum,
+  UpdatePokerScopeMutationResponse
+} from '../__generated__/UpdatePokerScopeMutation.graphql'
+
+const getSearchQueryFromMeeting = (
```
nickoferrall, line 6: The name may be misleading as we just want the query string and sometimes the filters

packages/client/mutations/UpdatePokerScopeMutation.ts
```diff
@@ -260,7 +276,30 @@ const UpdatePokerScopeMutation: StandardMutation<TUpdatePokerScopeMutation, Hand
...
+      const {meeting} = updatePokerScope
+      if (!meeting) return
+      const {viewerId} = atmosphere
+      const {meetingId, updates} = variables
+      const update = updates[0]!
+      const {service, action} = update
+      const searchQuery = getSearchQueryFromMeeting(meeting, service)
+      if (!searchQuery) return
+      const [queryString, searchQueryFilters] = searchQuery
+      SendClientSegmentEventMutation(atmosphere, 'Updated Poker Scope', {
```
igorlesnenko, line 293: Select all functionality seems to be broken for me, and I don't see event in segment logs
image

https://user-images.githubusercontent.com/466991/162723931-ff59e6fb-6158-4245-9209-45a17cd1365d.mov

nickoferrall: Yeah, select all functionality is broken in production. I created an issue to fix it: https://github.com/ParabolInc/parabol/issues/6213

I'll look into why the event isn't firing

packages/server/graphql/types/SegmentEventTrackOptions.ts
```diff
@@ -13,7 +14,12 @@ const SegmentEventTrackOptions = new GraphQLInputObjectType({
     meetingId: {type: GraphQLID},
     reflectionId: {type: GraphQLID},
     viewerId: {type: GraphQLID},
-    reflectionsCount: {type: GraphQLInt}
+    reflectionsCount: {type: GraphQLInt},
+    action: {type: GraphQLString},
+    queryString: {type: GraphQLString},
```
igorlesnenko, line 19: Should we name it `searchQueryString`?

packages/client/utils/getSearchQueryFromMeeting.ts
igorlesnenko: I feel like it is ok, maybe consider inlining this into mutation if only needed for one particular case.

igorlesnenko:
Got some crash when selecting a project.


https://user-images.githubusercontent.com/466991/162727822-89e98357-38ed-4ca6-976b-ce79e23cfcc2.mov



Seeing this in logs, might be related
```
SEND TO SENTRY Error: Query has complexity of 333, which exceeds max complexity of 250
    at /Users/igor/www/parabol/dev/gqlExecutor.js:56094:30
    at async Object.resolve [as TaskintegrationResolver] (/Users/igor/www/parabol/dev/gqlExecutor.js:49853:33) undefined
 ```

**Comment by igorlesnenko** [comment]
I didn't see the following event:

> Track the end of the search. This could be when the user leaves the input.

Did we decide to not implement it?

When creating a new issue, I see the following data in the segment event

```
analytics.track({
  userId: 'local|beuHr0gzra',
  event: 'Task added',
  properties: {
    email: 'igor@parabol.co',
    isReply: false,
    meetingId: 'csz3yDmpoc',
    teamId: 'beuHrkZhVT'
  }
});
```

Should we add more data like query string or is it fine?

File: packages/client/mutations/UpdatePokerScopeMutation.ts (hunk 4 above)
Comment (igorlesnenko) [reply packages/client/mutations/UpdatePokerScopeMutation.ts:293]: @nickoferrall I think the issue is the following
```
Field "selectedAll" is not defined by type SegmentEventTrackOptions.
```

**Comment by nickoferrall** [comment]
> Got crash when selecting a project

Ooh good to know. Does this happen every time you select a project from the filter? Had you selected all projects just beforehand?

> Did we decide to not implement [track end of search]?

I started implementing it but then wondered how much value it would add. We know when the user starts searching, clears the search, and adds an issue to the scope. If we measure the `onBlur` event for the input, it doesn't necessarily mean that the user has finished searching because they may have clicked the clear search icon & continued searching. But I'd be happy to add this back in if you think it'd be helpful.

> Should we add more data like query string [to create task event] or is it fine?

We want to know the search query string when the user adds an issue to the scope because it tells us whether their search was successful. But creating a new issue isn't related to the search query, so I don't think we'd learn much from tracking it.

**Review by igorlesnenko — APPROVED** [review]

**Review by Dschoordsch — CHANGES_REQUESTED** [review]

File: packages/client/mutations/UpdatePokerScopeMutation.ts [hunk 6]
```diff
@@ -260,7 +276,30 @@ const UpdatePokerScopeMutation: StandardMutation<TUpdatePokerScopeMutation, Hand
...
+      const {meeting} = updatePokerScope
+      if (!meeting) return
+      const {viewerId} = atmosphere
+      const {meetingId, updates} = variables
+      const update = updates[0]!
+      const {service, action} = update
+      const searchQuery = getSearchQueryFromMeeting(meeting, service)
+      if (!searchQuery) return
+      const [searchQueryString, searchQueryFilters] = searchQuery
+      SendClientSegmentEventMutation(atmosphere, 'Updated Poker Scope', {
```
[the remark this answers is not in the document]
nickoferrall: The difficulty is that the search query only exists in the client. So, we could add an additional `searchQuery` argument to `updatePokerScope`, but I'm not sure it's worth it just for tracking?

packages/client/components/GitLabScopingSearchFilterMenu.tsx
```diff
@@ -142,6 +144,11 @@ const GitLabScopingSearchFilterMenu = (props: Props) => {
               : [...selectedProjectsIds, projectId]
             gitlabSearchQuery.setValue(newSelectedProjectsIds, 'selectedProjectsIds')
           })
+          SendClientSegmentEventMutation(atmosphere, 'Selected GitLab Project Filter', {
```
Dschoordsch, line 147: -1 I think for future analysis it would be better if we keep these events more generic `'Selected Poker Scope Project Filter'` and adding `service: 'gitlab'` as a property.

packages/client/components/GitLabScopingSearchInput.tsx
```diff
@@ -64,13 +65,28 @@ const GitLabScopingSearchInput = (props: Props) => {
...
+    SendClientSegmentEventMutation(atmosphere, eventTitle, {
+      viewerId,
+      meetingId
+    })
+  }
 
   const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     const {value} = e.target
+    if (!queryString) {
+      trackEvent('Start GitLab search')
```
Dschoordsch, line 81: -1 Rename event to `'Started Poker Scope Search'`

packages/client/components/GitLabScopingSearchInput.tsx
```diff
@@ -64,13 +65,28 @@ const GitLabScopingSearchInput = (props: Props) => {
...
+    }
     setSearch(atmosphere, meetingId, value)
   }
+
   const clearSearch = () => {
+    trackEvent('Cleared GitLab search')
```
Dschoordsch, line 87: -1 `'Cleared Poker Scope Search'`

Dschoordsch (changes requested):

packages/client/components/GitLabScopingSearchInput.tsx
```diff
@@ -64,13 +65,29 @@ const GitLabScopingSearchInput = (props: Props) => {
...
+      viewerId,
+      meetingId,
+      service: 'gitlab'
+    })
+  }
 
   const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     const {value} = e.target
+    if (!queryString) {
+      trackEvent('Started Poker Scope search')
```
Dschoordsch, line 82: +1 Not a big deal, but the random capitalization seems odd to me as it is inconsistent with 'Selected Poker Scope Project Filter'. Seeing the existing events, these also seem to have random capital letters.

Dschoordsch (approved):

packages/client/components/GitLabScopingSearchInput.tsx
nickoferrall: True, I've capitalised `search` to make it more consistent

PR #6337, opened 2022-04-03
feat(DX): Fast dev mode

mattkrick:
predecessor: #6379
fix #6316 

Our development flow has gotten kinda crazy. Before a dev can work, the following must happen
- migrate rethinkdb
- migrate postgres
- flush redis
- build the servers
- run the gql executor
- run the socket server
- build the client vendor DLL
- run the webpack dev server
- watch graphql codegens
- update the graphql schema
- watch relay fragments
- build CLI toolbox

This PR parallelizes this work across 8 processes. 

TEST
- [ ] run `yarn dev`
- [ ] the app is available 3x faster than before
- [ ] in `createTask`, add `console.log('hi')` to the top of the resolve function. see the server restarts automatically. create a task in parabol & see the log statement in the server logs (proving webpack watch + output watchers are working)
- [ ] add `foo: Int!` as a field in `graphql/private/typeDefs/ping.graphql` & save the file. open `graphql/private/schema.graphql` and see the updates you just made (proving the schema updater is working)
- [ ] open `graphql/private/types/PingableServices` and rename the `postgres` field to `foo`. Notice that it doesn't throw an error (proving the codegen watcher is working)
- [ ] In DashSidebar.tsx L69 change "Add a team" to "ZZZZ" and hit save. see the field update in the browser without refreshing (proving HMR is working)
- [ ] in the same file, add an `id` field to the relay fragment add the bottom & hit save. Look at the definition for `DashSidebar_viewer` and see id is now included (proving relay compiler is working)
- [ ] run `git clean -fdx && yarn`, then copy the `.env` back into place, and then `yarn dev`. the app works (you can ignore the errors that happen) 
- [ ]  run `git clean -fdx` and then run `yarn postcheckout`. It builds the graphql schema, relay fragments, and codegens
- [ ] run `yarn build && yarn start` and it works (no regressions in prod.js)

PR #6379, opened 2022-04-10
feat(CI): use prod build for integration tests

mattkrick:
This PR runs the integration tests against a production build vs. a development build.
This is beneficial for a few reasons:
- it's closer to production, so we may catch an extra bug at some point
- the CI already performs a prod build, so CI goes faster because it no longer has to run a dev build
- the dev build in #6337 includes a bunch of watchers which break in CI. We could make it work, but given the above 2 benefits, this seems like the best solution

Also deleted a handful of unused files.

TEST
- [ ] the tests in CI pass


packages/server/__tests__/common.ts
```diff
@@ -30,13 +32,27 @@ export async function sendIntranet(req: {
   return response.json()
 }
 
+const persistQuery = async (query: string) => {
+  const r = await getRethink()
+  const docId = persistFunction(query.trim())
```
Dschoordsch: -2 `await` is missing

blimmer:
See also https://github.com/ParabolInc/parabol/issues/6381, which we should consider as part of this PR

packages/server/database/rethinkDriver.ts
```diff
@@ -231,7 +231,11 @@ const getRethink = async () => {
   }
   // this is important because pm2 will restart the process & for whatever reason r isn't always healthy
   await r.waitForHealthy()
-  return (r as unknown) as ParabolR
+  return r as unknown as ParabolR
+}
+
+export const closeRethink = async () => {
+  await (await promise)?.drain()
```
mattkrick, line 238: little nit, but we'll want to drain the whole pool, not just the connection (r.getPoolMaster()?.drain())

mattkrick:
i don't know what gypsy magic jest is doing, but the connection pool started in the tests is not the same one that is started in globalTeardown. They have the same pid & ppid, so some kind of sandboxing is going on. 

Jest says the following:
```
Note: Any global variables that are defined through globalSetup can only be read in globalTeardown. You cannot retrieve globals defined here in your test suites.
```

Anyways, the "right" way to do this is to close down the DB after the tests in a single file are complete, so that's what i did.
in the future we'll reuse this `afterAll` pattern to clean up certain tables probably, so it's really just introducing the pattern a little early.

packages/server/database/migrations/20180206135107-renameProjectToTask.js
```diff
@@ -1,24 +1,13 @@
...
-    .table('ProjectHistory')
-    .config()
-    .update({name: 'TaskHistory'})
-    .run()
-  await r
-    .table('TaskHistory')
-    .indexDrop('projectIdUpdatedAt')
-    .run()
+  await r.table('Project').config().update({name: 'Task'}).run()
+  await r.table('Task').wait()
```
mattkrick, line 4: moved this line up a few to give it a few more ms to kick in.
this shouldn't be necessary, but we live in a crazy world.

mattkrick:
@Dschoordsch ready for another review

mattkrick:
@Dschoordsch if we move `afterAll` to its own file will we always have to run tests with `--runInBand` and guarantee that `setup.ts` is run last?

Dschoordsch:
> @Dschoordsch if we move `afterAll` to its own file will we always have to run tests with `--runInBand` and guarantee that `setup.ts` is run last?

No, it's part of the setup for each test https://github.com/ParabolInc/parabol/blob/f7d24ed7de01d02ca66936534c194c688d67056d/packages/server/jest.config.js#L22 so it will be run for each file

mattkrick:
THAT IS AWESOME

PR #6382, opened 2022-04-11
fix(integration-tests): support running integration tests against another env

blimmer:
This PR allows us to run the integration test suite against a different HOST (other than `localhost`). See inline comments for details.

### DIFF
--- a/packages/integration-tests/playwright.config.ts
+++ b/packages/integration-tests/playwright.config.ts
@@ -1,4 +1,4 @@
-import {PlaywrightTestConfig, devices} from '@playwright/test'
+import {devices, PlaywrightTestConfig} from '@playwright/test'
 
 /**
  * See https://playwright.dev/docs/test-configuration.
@@ -27,7 +27,7 @@ const config: PlaywrightTestConfig = {
   workers: process.env.CI ? 1 : undefined,
 
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
-  reporter: 'html',
+  reporter: [['list'], ['html']],
 
   /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
   use: {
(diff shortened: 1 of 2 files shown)


packages/integration-tests/playwright.config.ts
```diff
@@ -27,7 +27,7 @@ const config: PlaywrightTestConfig = {
   workers: process.env.CI ? 1 : undefined,
 
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
-  reporter: 'html',
+  reporter: [['list'], ['html']],
```
blimmer, line 30: In GitLab CI, we were just getting the [`dot` reporter](https://playwright.dev/docs/test-reporters#dot-reporter) which made failures difficult to debug.

This change forces the default reporter (`list`) in addition to `html`.

packages/integration-tests/tests/config.ts
```diff
@@ -24,7 +24,7 @@ export class Config {
 
   private rootUrlPathFromEnv({HOST, PORT}: typeof EnvConfig): string {
     const scheme = HOST === 'localhost' ? 'http' : 'https'
-    return `${scheme}://${HOST}${PORT ? `:${PORT}` : ''}`
+    return `${scheme}://${HOST}${HOST === 'localhost' ? `:${PORT}` : ''}`
```
blimmer, line 27: Since we're reading the `.env` file, `PORT` was set to the `.env` value, even when we wanted to test against a different `HOST`. For example `HOST=action-staging.parabol.co yarn test` was trying to hit `https://action-staging.parabol.co:3000`.

Dschoordsch: -1 The right thing to do then would be to unset PORT, wouldn't it? So the check should be `PORT ? `:${PORT}` : ''`

Dschoordsch (changes requested):

packages/integration-tests/tests/config.ts
mattkrick: port is a funny one because it's only exposed on the ingress for localhost (see https://github.com/ParabolInc/parabol/blob/f221880dac24332874939b6c0a3f7328f4b76023/packages/server/appOrigin.ts#L4)

gonna merge this as is because today we're just testing on localhost. if there's a use case that's still broken we can fix it then

PR #6387, opened 2022-04-12
chore(dx): Don't truncate TypeScript types

Dschoordsch:
This setting influences truncation from TSServer in hover and error
messages.


BartoszJarocki (approved):
TIL, useful option 👌

PR #6388, opened 2022-04-12
feat(standups): Wire up UpdateTeamPromptResponseMutation

tianrunhe:
Implements #5950. Based on https://github.com/ParabolInc/parabol/pull/6333

To tests:
1. Start an async startup meeting
2. Go into the meeting and see response cards
3. Update the content on the response card of yours
4. Refresh the page or go to other page & then come back, see the updated content preserved.

### DIFF
--- a/packages/server/postgres/queries/getTeamPromptResponsesByIds.ts
+++ b/packages/server/postgres/queries/getTeamPromptResponsesByIds.ts
@@ -10,17 +10,6 @@ export interface TeamPromptResponse
   content: JSONContent
 }
 
-export const mapToTeamPromptResponse = (
-  results: IGetTeamPromptResponsesByIdsQueryResult[]
-): TeamPromptResponse[] => {
-  return results.map((teamPromptResponse: any) => {
-    return {
-      ...teamPromptResponse,
-      content: teamPromptResponse.content as JSONContent
-    } as TeamPromptResponse
-  })
-}
-
 export const getTeamPromptResponsesByIds = async (
   teamPromptResponseIds: readonly number[]
 ): Promise<TeamPromptResponse[]> => {
@@ -28,5 +17,5 @@ export const getTeamPromptResponsesByIds = async (
     {ids: teamPromptResponseIds},
     getPg()
   )
-  return mapToTeamPromptResponse(teamPromptResponses)
+  return teamPromptResponses as TeamPromptResponse[]
 }
--- a/packages/server/postgres/queries/getTeamPromptResponsesByMeetingIds.ts
+++ b/packages/server/postgres/queries/getTeamPromptResponsesByMeetingIds.ts
@@ -1,12 +1,12 @@
 import getPg from '../getPg'
 import {getTeamPromptResponsesByMeetingIdQuery} from './generated/getTeamPromptResponsesByMeetingIdQuery'
-import {mapToTeamPromptResponse, TeamPromptResponse} from './getTeamPromptResponsesByIds'
+import {TeamPromptResponse} from './getTeamPromptResponsesByIds'
 
 export const getTeamPromptResponsesByMeetingIds = async (
   meetingIds: readonly string[]
 ): Promise<TeamPromptResponse[]> => {
   const responses = await getTeamPromptResponsesByMeetingIdQuery.run({meetingIds}, getPg())
-  return mapToTeamPromptResponse(responses)
+  return responses as TeamPromptResponse[]
 }
 
 export const getTeamPromptResponsesByMeetingId = async (


packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
```diff
@@ -57,11 +66,27 @@ const TeamPromptResponseCard = (props: Props) => {
...
 
-  const {teamMember} = responseStage
+  const {teamMember, meetingId, response} = responseStage
   const {picture, preferredName, userId} = teamMember
 
+  const contentJSON: JSONContent | null = response ? JSON.parse(response.content) : null
+  const plaintextContent = response?.plaintextContent ?? ''
   const isCurrentViewer = userId === viewerId
-  const isEmptyResponse = !isCurrentViewer // :TODO: (jmtaber129): Determine based on actual response, too
+  const isEmptyResponse = !isCurrentViewer && !plaintextContent // :TODO: (jmtaber129): Determine based on actual response, too
```
jmtaber129, line 75: +1 Remove TODO

jmtaber129 (changes requested):
This mostly looks good, but I'm not sure the `TeamPromptResponse` id change is what we want to do here

packages/server/graphql/types/TeamPromptResponse.ts
```diff
@@ -10,6 +11,12 @@ const TeamPromptResponse: GraphQLObjectType = new GraphQLObjectType<any, GQLCont
   interfaces: () => [Reactable],
   fields: () => ({
     ...reactableFields(),
+    id: {
+      type: new GraphQLNonNull(GraphQLID),
+      description:
+        'Team prompt response id in a format of `teamPromptResponse:idGeneratedByDatabase`',
+      resolve: ({id}) => TeamPromptResponseId.join(id)
+    },
```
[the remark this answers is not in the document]
BartoszJarocki: if you see the places where wrong ids are used, then it's definitely a bug. client-facing code has to use prefixed value, backend operations non-prefixed. regarding the problem, yes, in the new PG tables we use `INT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY` which is simply an int, that is only unique in the scope of the given table. Relay, on the other hand [requires globally unique ids ](https://relay.dev/graphql/objectidentification.htm#sel-DAFFCAACEyBgnH), more info [here](https://stackoverflow.com/questions/33399901/in-relay-what-role-do-the-node-interface-and-the-global-id-spec-play/33411416#33411416). check the [gqlIds](https://github.com/ParabolInc/parabol/tree/master/packages/client/shared/gqlIds) directory. there's more, I haven't introduced this pattern in this PR.

packages/server/graphql/types/TeamPromptResponseStage.ts
```diff
@@ -21,11 +23,17 @@ const TeamPromptResponseStage = new GraphQLObjectType<any, GQLContext>({
...
       description: 'The response to the prompt',
-      resolve: () => {
-        // TODO: implement fetching responses
-        return {}
+      resolve: async (
+        {meetingId, teamMemberId}: {meetingId: string; teamMemberId: string},
+        _args: unknown,
+        {}
+      ) => {
+        // TODO: implement getTeamPromptResponsesByMeetingIdAndUserId
```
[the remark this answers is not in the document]
BartoszJarocki: nope, just a leftover. we should probably use a data loader here anyway. i don't want to slow down the PR, will create a separate task to double-check data loaders usage across team prompt gql types.

packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
BartoszJarocki: done

packages/server/graphql/types/TeamPromptResponseStage.ts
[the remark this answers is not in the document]
BartoszJarocki: task: https://github.com/ParabolInc/parabol/issues/6496

packages/server/graphql/types/TeamPromptResponse.ts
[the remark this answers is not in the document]
jmtaber129: Ahh gotcha.  In that case, I think there may be some flaws in this overall pattern that might be worth discussing (e.g. inconsistencies in `TeamPromptResponse` `id`s between what's just come out of the DB vs an `id` returned by this resolver), but that shouldn't block the rest of the PR.

jmtaber129 (approved):

packages/client/mutations/UpsertTeamPromptResponseMutation.ts
```diff
@@ -0,0 +1,96 @@
...
+      )
+      if (!stage) return
+      const response = stage.getLinkedRecord('response')
+      if (!response) {
+        // insertion
+        const newResponse = createProxyRecord(store, 'TeamPromptResponse', {
+          id: clientTempId(viewerId),
+          meetingId,
+          sortOrder: 0,
+          content
```
mattkrick, line 83: -1 in the card component above it's referencing `plaintextContent`. 

either we need to add plaintextContent to this optimisticUpdater or in the component above we should remove `plaintextContent` and instead calculate that value.

packages/client/mutations/UpsertTeamPromptResponseMutation.ts
```diff
@@ -0,0 +1,96 @@
...
+  variables,
+  {onError, onCompleted}
+) => {
+  return commitMutation<TUpsertTeamPromptResponseMutation>(atmosphere, {
+    mutation,
+    variables,
+    optimisticUpdater: (store) => {
+      const {viewerId} = atmosphere
+      const {meetingId, content} = variables
+      if (!meetingId) return
```
mattkrick, line 65: +1 meetingId is guaranteed, can remove.

packages/server/postgres/queries/getTeamPromptResponsesByIds.ts
```diff
@@ -10,7 +10,7 @@ export interface TeamPromptResponse
   content: JSONContent
 }
 
-const mapToTeamPromptResponse = (
+export const mapToTeamPromptResponse = (
```
mattkrick, line 13: -1 don't do in runtime what you can do in build time. casting a response to a type shouldn't require a function

packages/client/mutations/UpsertTeamPromptResponseMutation.ts
BartoszJarocki: fixed

packages/server/postgres/queries/getTeamPromptResponsesByIds.ts
BartoszJarocki: could you please elaborate a bit more? how can we do the casting in the build time? also, this function is heavily updated in https://github.com/ParabolInc/parabol/pull/6407/files#diff-373264581dbe0ee3e09bdbe59799c483616655423e01c4f6b3ecc9338f17a4a0R18-R30 to cast reactjis to proper type so we'd want to keep that in mind

packages/server/postgres/queries/src/getTeamPromptResponsesByMeetingIdQuery.sql
```diff
@@ -0,0 +1,6 @@
+/*
+  @name getTeamPromptResponsesByMeetingIdQuery
+  @param meetingIds -> (...)
+*/
+SELECT "id", "createdAt", "updatedAt", "meetingId", "userId", "sortOrder", "content", "plaintextContent" FROM "TeamPromptResponse"
```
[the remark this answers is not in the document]
BartoszJarocki: here @jmtaber129 casts it to json and then in the runtime to the proper type: https://github.com/ParabolInc/parabol/pull/6407/files#diff-00749d0652800c424824c0c23172bbae293db5aa559f1fd57d8295a999656210R5

BartoszJarocki:
@mattkrick could you please take a look again? I added some comments, thank you!

packages/client/mutations/UpsertTeamPromptResponseMutation.ts
mattkrick: the query requires the field `plaintextContent`, but it's not provided here. which could cause an error if the component depends on it (e.g. add something like `plaintextContent.trim()` to the component & it'll complain that it's undefined.

when creating this new optimistic object, you'll need to do something like:
```ts
{
content: JSON.stringify(editorState.getJSON()),
plaintextContent: editor.state.doc.textContent // or however you get the plaintext
}
```

packages/server/postgres/queries/getTeamPromptResponsesByIds.ts
mattkrick: after typescript compiles this function, it will just be an identity function where the input is the same as the output. 
instead, you just want to cast:

```ts
type Foo extends Omit<IGetTeamPromptResponsesByIdsQueryResult[0], 'content'> {
  content: JSONContent
}
return responses as Foo[]
```

packages/server/postgres/queries/src/getTeamPromptResponsesByMeetingIdQuery.sql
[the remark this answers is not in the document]
mattkrick: sounds good!

packages/client/mutations/UpsertTeamPromptResponseMutation.ts
BartoszJarocki: oh got it. let's have this conversation in this PR - https://github.com/ParabolInc/parabol/pull/6504. I don't think we need an optimistic updater here at all.

Issue #6322, opened 2022-03-31
Sprint Poker: GitLab v1: Can vote on an issue


nickoferrall:
### AC

- [ ] In the Estimate phase of a Poker meeting, a user can successfully vote on a GitLab issue

Dschoordsch:
@nickoferrall can you please add your estimate until we can vote in a poker meeting?

nickoferrall:
@Dschoordsch sure, as in add my estimate as a label to this issue?

nickoferrall:
**Story Points: 8**
  [See the discussion](http://localhost:3000/meet/curomvzm0g/estimate/4) in Sprint Poker #​​32

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

nickoferrall:
**Story Points: 8**
  [See the discussion](http://localhost:3000/meet/curomvzm0g/estimate/4) in Sprint Poker #​​32

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

Issue #6287, opened 2022-03-25
Sprint Poker: GitLab v2: Add "Powered by Parabol" comment after voting


nickoferrall:
After voting on a GitHub issue in Sprint Poker, if the "As comment" label is selected, we add a comment to the issue saying "Powered by Parabol".

We should do the same for GitLab. 

### AC

- [ ] After voting on a GitLab issue, if the "As comment" label is selected, add a comment to the GitLab issue saying "Powered by Parabol" 
- [ ] Like GitHub, the comment has a hyperlink linking back to Parabol

Issue #6397, opened 2022-04-13
Poker Update/Edit Score button flickers


nickoferrall:
After selecting a new dimension from the GitHub label, the Update/Edit Score button flickers:

![flickering-update-btn](https://user-images.githubusercontent.com/39854876/163214211-3258db53-9a0f-4075-af3a-fbfd0f992156.gif)

Jira seems to be working as expected.

### AC

- [ ] The update/edit score button does not flicker after selecting a new dimension from the GitHub label

Dschoordsch:
@nickoferrall Do we republish the vote when changing the field? If that is the case, then it seems ok if the button changes state while we're updating.

PR #6398, opened 2022-04-13
feat(sprint-poker): Can vote on GitLab issue

nickoferrall:
Fixes https://github.com/ParabolInc/parabol/issues/6322 & fixes https://github.com/ParabolInc/parabol/issues/6287

This PR adds:

- a GitLab label to the Estimate phase
- the user can select "As Comment" or "Do not update"
- if they select "As Comment", voting on a GitLab issue in Parabol adds a comment to the GitLab issue
- the GitLab comment includes the "Powered by Parabol" link 

Note: when selecting the "As Comment" or "Do not update" label, the Update/Edit Score button can flicker. This bug exists in GitHub in prod, so I created an issue for it: https://github.com/ParabolInc/parabol/issues/6397


packages/server/graphql/mutations/helpers/pushEstimateToGitLab.ts
```diff
@@ -0,0 +1,72 @@
...
+    dataLoader.get('gitlabDimensionFieldMaps').load({teamId, dimensionName, gid})
+  ])
+  if (!auth) return new Error('User no longer has access to GitLab')
+
+  const labelTemplate = fieldMap?.labelTemplate ?? SprintPokerDefaults.SERVICE_FIELD_COMMENT
+  if (labelTemplate === SprintPokerDefaults.SERVICE_FIELD_NULL) return undefined
+
+  const {accessToken, providerId} = auth
+  const provider = await dataLoader.get('integrationProviders').load(providerId)
+  if (!provider?.serverBaseUrl || !accessToken) return new Error('Invalid integration provider')
```
igorlesnenko, line 43: Should we use `!` instead if it is not a possible scenario?

igorlesnenko (changes requested):
Only small tweaks:

packages/server/graphql/mutations/helpers/pushEstimateToGitLab.ts
nickoferrall: As in `provider!.serverBaseUrl!`?

igorlesnenko: @nickoferrall I mean consider serverBaseUrl non-nullable below and remove this check. Not sure if invalid integration provider is really possible
`const manager = new GitLabServerManager(accessToken, provider.serverBaseUrl!)`

igorlesnenko: Also, above we checked `auth`: `if (!auth) return new Error('User no longer has access to GitLab')`
Could we have `accessToken` empty after that?

igorlesnenko: maybe consider just:

```
if (!provider) return new Error('integration provider not found')
```

packages/server/graphql/mutations/helpers/pushEstimateToGitLab.ts
```diff
@@ -39,14 +39,12 @@ const pushEstimateToGitLab = async (
...
 
   const {accessToken, providerId} = auth
+  if (!accessToken) return new Error('Invalid GitLab auth')
   const provider = await dataLoader.get('integrationProviders').load(providerId)
-  if (!provider?.serverBaseUrl || !accessToken) return new Error('Invalid integration provider')
+  if (!provider?.serverBaseUrl) return new Error('Invalid GitLab provider')
   const manager = new GitLabServerManager(accessToken, provider.serverBaseUrl)
   const gitlabRequest = manager.getGitLabRequest(info, context)
   if (labelTemplate === SprintPokerDefaults.SERVICE_FIELD_COMMENT) {
-    if (!stageId || !meeting) {
```
igorlesnenko: I think we still need meeting check, but just moving this above

```
  const [task, meeting] = await Promise.all([
    dataLoader.get('tasks').load(taskId),
    dataLoader.get('newMeetings').load(meetingId)
  ])
  if (!meeting) {
    return new Error('Meeting does not exist')
  }
  ```

File: packages/server/graphql/mutations/helpers/pushEstimateToGitLab.ts (hunk 1 above)
Comment (igorlesnenko) [reply packages/server/graphql/mutations/helpers/pushEstimateToGitLab.ts]: @nickoferrall I meant something like

```
if (!auth) return new Error('User no longer has access to GitLab')
...
if (!provider) return new Error('integration provider not found')
...
const manager = new GitLabServerManager(accessToken!, provider.serverBaseUrl!)

```

But I think your approach is also fine

Comment (nickoferrall) [reply packages/server/graphql/mutations/helpers/pushEstimateToGitLab.ts]: Makes sense, thanks! Update pushed

**Review by igorlesnenko — APPROVED** [review]

File: packages/server/graphql/mutations/helpers/pushEstimateToGitLab.ts [hunk 3]
```diff
@@ -0,0 +1,70 @@
...
+    const {stages} = estimatePhase
+    const stageIdx = stages.findIndex((stage) => stage.id === stageId)
+    const discussionURL = makeAppURL(appOrigin, `meet/${meetingId}/estimate/${stageIdx + 1}`)
+    const body = makeScoreGitLabComment(
+      dimensionName,
+      value || '<None>',
+      meetingName,
+      discussionURL
+    )
+    const [, commentError] = await gitlabRequest(createNote, {
```
mattkrick, line 59: +1 I know we pushed it off on the last PR. Is this PR the right time to have GitLabServerManager implement TaskIntegrationManager? Having another integration using that abstraction could do some good in how they shape up & remove the need for a large refactoring PR at the end of this milestone

packages/server/graphql/mutations/updateGitLabDimensionField.ts
```diff
@@ -0,0 +1,77 @@
+import {GraphQLID, GraphQLNonNull, GraphQLString} from 'graphql'
+import {SubscriptionChannel} from 'parabol-client/types/constEnums'
+import MeetingPoker from '../../database/types/MeetingPoker'
+import upsertGitLabDimensionFieldMap from '../../postgres/queries/upsertGitLabDimensionFieldMap'
+import {isTeamMember} from '../../utils/authorization'
+import publish from '../../utils/publish'
+import {GQLContext} from '../graphql'
+import UpdateGitLabDimensionFieldPayload from '../types/UpdateGitLabDimensionFieldPayload'
+
+interface Args {
```
mattkrick, line 10: -1 could you write this as a resolver using the new SDL pattern? that way these args will be provided for you

mattkrick (changes requested):
looking good! My only advice would be to consider filling up the GitLabServerManager with all the helper functions you can to set a good pattern moving forward

packages/server/graphql/mutations/helpers/pushEstimateToGitLab.ts
nickoferrall: I've implemented the new pattern including `createNote` in this PR: https://github.com/ParabolInc/parabol/pull/6427

packages/server/graphql/mutations/updateGitLabDimensionField.ts
nickoferrall: Nice, it was cool to see the new pattern! The type safety is so good 🙌  Update pushed!

packages/server/postgres/migrations/1649750150687_GitLabDimensionFieldMap.ts
```diff
@@ -0,0 +1,27 @@
+import {Client} from 'pg'
+import getPgConfig from '../getPgConfig'
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+  CREATE TABLE IF NOT EXISTS "GitLabDimensionFieldMap" (
+    "id" SERIAL,
+    "teamId" VARCHAR(120) NOT NULL,
```
Dschoordsch, line 10: -1 "teamId" is a `VARCHAR(100)`

packages/server/postgres/migrations/1649750150687_GitLabDimensionFieldMap.ts
```diff
@@ -0,0 +1,27 @@
...
+    "dimensionName" VARCHAR(120) NOT NULL,
+    "gid" VARCHAR(140) NOT NULL,
+    "labelTemplate" VARCHAR(100) NOT NULL,
+    PRIMARY KEY ("teamId", "dimensionName", "gid")
```
Dschoordsch, line 14: -1 Why have an `id` field if it's not the primary key?  It does not even have a unique constraint.

Dschoordsch (changes requested):

packages/server/postgres/migrations/1649750150687_GitLabDimensionFieldMap.ts
nickoferrall: I followed the GitHub example here: https://github.com/ParabolInc/parabol/blob/cf55ed4d0877b1756ffc7213062dde5056ed662b/packages/server/postgres/migrations/1631822263196_GitHubDimensionFieldMap.ts

I'm not sure why. Perhaps @mattkrick could help clarify?

packages/server/postgres/migrations/1649750150687_GitLabDimensionFieldMap.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+  CREATE TABLE IF NOT EXISTS "GitLabDimensionFieldMap" (
+    "id" SERIAL,
+    "teamId" VARCHAR(100) NOT NULL,
+    "dimensionName" VARCHAR(120) NOT NULL,
+    "gid" VARCHAR(140) NOT NULL,
```
Dschoordsch, line 12: -2 I don't understand why we store the voting dimension by issue id. I think we should have the `providerId`, `projectId` instead, so it is set for the project and all upcoming voting issues.

Dschoordsch (changes requested):

packages/server/postgres/migrations/1649750150687_GitLabDimensionFieldMap.ts
nickoferrall: For the `[Integration]DimensionFieldMap`s, my understanding is we want to store the `dimensionName`, e.g. Story Points, and `labelTemplate`, e.g. `__comment`, associated with a particular project/repo.

We do this so that the user's preferences are stored so they don't need to keep changing between "As a comment" and "Do not update", for example. 

Previously, I thought it was associated with an issue, so I was storing the `gid`. But, I now see that when I update the "As a comment" label to "As a label", in the GitHub integration, it updates the label for all other issues in that repo. 

So, I agree that I should change `gid` to `projectId` so that we store the preferences for all issues within a project in GitLab.

What I'm unclear on is why we should have `providerId`? Are you thinking we should have this instead of `teamId`? The same issue could exist in multiple teams. By including the `teamId`, we only update the label preference on that team.

packages/server/postgres/migrations/1649750150687_GitLabDimensionFieldMap.ts
mattkrick: every node needs an ID in graphql/relay so i just gave it a one so we didn't have to concat 3 fields together `("teamId", "dimensionName", "nameWithOwner")`. safe to remove if you'd rather concat em

packages/server/postgres/migrations/1649750150687_GitLabDimensionFieldMap.ts
Dschoordsch: I would like to tie all integration settings to their provider. As these are potentially self-hosted a client could switch between two instances where both use the same ids belonging to different projects. Or teams move orgs, there are some scenarios where this could happen. We should keep the `teamId` as well.

nickoferrall: Storing the providerId makes sense for that self-hosted switching edge case. So you're happy for us to store teamId, projectId, dimensionName, and providerId here?

Dschoordsch: As the key/unique constraint to lookup the setting, yes. And then "labelTemplate" to store the actual setting.

packages/server/postgres/migrations/1649750150687_GitLabDimensionFieldMap.ts
Dschoordsch: I'm also fine with keeping it, but it must have a `unique` constraint since you can just insert/update serial fields.

Dschoordsch: I'm also fine with keeping it, but it must have a `unique` constraint since you can just insert/update serial fields.

packages/server/postgres/migrations/1649750150687_GitLabDimensionFieldMap.ts
nickoferrall: Sounds good. Update pushed!

packages/server/graphql/mutations/helpers/pushEstimateToGitLab.ts
```diff
@@ -0,0 +1,77 @@
...
+    userId: accessUserId
+  })
+  if (!auth) return new Error('User no longer has access to GitLab')
+  const {accessToken, providerId} = auth
+  const provider = await dataLoader.get('integrationProviders').load(providerId)
+  if (!provider) return new Error('Integration provider not found')
+  const manager = new GitLabServerManager(accessToken!, provider.serverBaseUrl!)
+  const gitlabRequest = manager.getGitLabRequest(info, context)
+
+  const [data] = await gitlabRequest<GetIssueQuery>(getIssue, {gid})
```
nickoferrall, line 45: In this PR https://github.com/ParabolInc/parabol/pull/6427 this is refactored to use the `TaskIntegrationManager` pattern

packages/server/graphql/types/EstimateStage.ts
```diff
@@ -109,6 +116,36 @@ const EstimateStage = new GraphQLObjectType<Source, GQLContext>({
             type: 'string'
           }
         }
+        if (service === 'gitlab') {
+          const {gid} = integration
+          const gitlabAuth = await dataLoader
+            .get('teamMemberIntegrationAuths')
+            .load({service: 'gitlab', teamId, userId: viewerId})
+          if (!gitlabAuth?.accessToken) return null
```
Dschoordsch, line 124: -2 This is failing for meeting participants without gitlab integration. They will not be able to join a Poker Meeting with GitLab issues at all.
```
GraphQLError: Cannot return null for non-nullable field EstimateStage.(serviceField,serviceField,serviceField,serviceField,serviceField,serviceField,serviceField,serviceField).
```
Reproduce: Ingo has GitLab integration, creates a Poker Meeting, adds GitLab tasks and moves to estimate stage. Paul has no integration and tries to join the meeting.

Dschoordsch (changes requested):

Dschoordsch (approved):

Dschoordsch:
Since Matt had no -2 you are free to merge per our policy

Issue #6097, opened 2022-02-22
Jira Server: Allow searching for issues in Sprint Poker


Dschoordsch:
Needs:
* #6096 
 
## Acceptance criteria
* [ ] Allow custom search
* [ ] pagination of results

Dschoordsch:
#6406 was merged

PR #6406, opened 2022-04-14
feat(jira-server): Allow searching for issues in Sprint Poker

igorlesnenko:
#6097
image

**How to test:**
* connect jira server
* create poker meeting
* go to scope on jira server tab
* Check that search, JQL search, project filters works


packages/client/components/JiraScopingSearchBar.tsx
```diff
@@ -1,55 +1,123 @@
...
     meetingRef
   )
 
+  const {id: meetingId, teamId} = meeting
+  const {jiraSearchQueries} = meeting.viewerMeetingMember?.teamMember.integrations?.atlassian ?? {}
+  const searchQueries =
+    jiraSearchQueries?.map((jiraSearchQuery) => {
+      const {id, queryString, isJQL, projectKeyFilters} = jiraSearchQuery
+
+      const selectQuery = () => {
```
nickoferrall, line 61: -2 I like that `ScopingSearchHistoryToggle` is now a reusable component, but the problem is that this complex `searchQueries` logic is now taking place in the `JiraScopingSearchBar`. This means that:

1. On every keystroke in the search input, we run through this logic which contains a nested `map`, `filter`, `indexOf` and creates a few functions. The number of saved `jiraSearchQueries` will probably be fairly small, but if a user is searching & selecting issues every week for over a year, they could have hundreds of saved queries. 
2. The rest of the search bar doesn't need to know about the `searchQueries`. I'd say that it's easier to read the component & component tree if each component only contains the minimum amount of data it needs.

I'd be in favour of reverting back to `JiraScopingSearchHistoryToggle` and keeping the logic there, but feel free to disagree!

packages/client/components/JiraScopingSearchResults.tsx
```diff
@@ -87,12 +87,17 @@ const JiraScopingSearchResults = (props: Props) => {
...
               key={node.id}
-              issue={node}
+              service={'jira'}
               usedServiceTaskIds={usedServiceTaskIds}
+              serviceTaskId={node.id}
               meetingId={meetingId}
               persistQuery={persistQuery}
+              summary={node.summary}
+              url={node.url}
+              linkText={node.issueKey}
```
nickoferrall, line 99: +1 could destructure these but it's NIT

packages/client/components/ScopingSearchBar.tsx
```diff
@@ -0,0 +1,65 @@
...
+  overflow: 'hidden',
+  textOverflow: 'ellipsis'
+})
+
+interface Props {
+  children: ReactNode
+  currentFilters?: string
+}
+
+const ScopingSearchBar = (props: Props) => {
```
nickoferrall, line 50: +2 lovely stuff making this reusable! 

+1 It looks like you've implemented this in Jira, but GitHub & GitLab are using the old version. If you want to prevent PR bloat and don't want to change them in this PR, perhaps we should create an issue to change them? Otherwise, someone might edit the `ScopingSearchBar`, see that it's updated the UI in Jira & Jira Server, and assume it's done the same for the other integrations.

packages/client/components/ScopingSearchHistoryToggle.tsx
```diff
@@ -0,0 +1,63 @@
...
+  alignItems: 'center',
+  justifyContent: 'center',
+  paddingRight: 8
+})
+
+interface Props {
+  searchQueries?: SearchQueries[]
+}
+
+const ScopingSearchHistoryToggle = (props: Props) => {
```
nickoferrall, line 40: +1 as this is only being used in Jira, perhaps `JiraScopingSearchHistoryToggle` might be more descriptive?

packages/client/components/ScopingSearchInput.tsx
```diff
@@ -0,0 +1,73 @@
...
+}))
+
+interface Props {
+  placeholder: string
+  queryString: string
+  meetingId: string
+  linkedRecordName: string
+}
+
+const ScopingSearchInput = (props: Props) => {
```
nickoferrall, line 41: +1 same here. I think we should either use the `Jira` prefix, or implement this in GitHub & GitLab, or create an issue to do so

packages/server/graphql/types/JiraServerIntegration.ts
```diff
@@ -20,6 +19,16 @@ import IntegrationProviderOAuth1 from './IntegrationProviderOAuth1'
...
 import TeamMemberIntegrationAuthOAuth1 from './TeamMemberIntegrationAuthOAuth1'
+import IntegrationRepoId from "~/shared/gqlIds/IntegrationRepoId";
+
+type IssueArgs = {
+  first: number
+  after?: string
+  queryString: string | null
+  isJQL: boolean
+  projectKeyFilters: string[] | null
+  [argName: string]: any
```
nickoferrall, line 30: -1 why `argName: string`? When is this used?

packages/server/integrations/jiraServer/JiraServerRestManager.ts
```diff
@@ -116,9 +117,7 @@ export default class JiraServerRestManager implements TaskIntegrationManager {
     const json = await response.json()
 
     if (response.status !== 201 && response.status !== 200) {
-      return new Error(
-        `Fetching projects failed with status ${response.status}, ${this.formatError(json)}`
```
nickoferrall, line 120: -1 why remove the description and `response.status`?

packages/server/graphql/types/JiraServerIntegration.ts
```diff
@@ -119,11 +124,14 @@ const JiraServerIntegration = new GraphQLObjectType<{teamId: string; userId: str
...
 
-        const issueRes = await integrationManager.getIssues()
+        const projectKeys = (projectKeyFilters ?? []).map((projectKeyFilter) => IntegrationRepoId.split(projectKeyFilter).projectKey!)
+
+        const issueRes = await integrationManager.getIssues(queryString, isJQL, projectKeys)
 
         if (issueRes instanceof Error) {
-          sendToSentry(issueRes, {userId, tags: {teamId}})
-          return null
+          return connectionFromTasks([], first, {
```
nickoferrall, line 132: Why return an empty connection out of interest instead of null & sending the error to Sentry?

Also, it's not from this PR, but I noticed that we're checking is `provider` exists above, but we're using `loadNonNull` so it's guaranteed that it won't be nullable

nickoferrall (changes requested):

packages/server/integrations/jiraServer/JiraServerRestManager.ts
igorlesnenko: @nickoferrall `Fetching projects failed` is an incorrect message, as it is not about projects, but an abstract error.
We also show the message directly on a frontend for `Jira`. e.g. incorrect JQL syntax. 
```
Error in the JQL Query: Expecting operator before the end of the query. The valid operators are '=', '!=', '<', '>', '<=', '>=', '~', '!~', 'IN', 'NOT IN', 'IS' and 'IS NOT'.
```
So I removed the status here as not required to display

packages/server/graphql/types/JiraServerIntegration.ts
igorlesnenko: @nickoferrall As here we are returning an error that should be displayed in the UI
image

igorlesnenko: Removed provider check

packages/server/graphql/types/JiraServerIntegration.ts
igorlesnenko: I don't see it used anywhere, so removed

packages/client/components/ScopingSearchBar.tsx
igorlesnenko: @nickoferrall created an issue to keep track of it #6435

packages/client/components/ScopingSearchInput.tsx
igorlesnenko: Created an issue https://github.com/ParabolInc/parabol/issues/6435

packages/client/components/ScopingSearchHistoryToggle.tsx
igorlesnenko: It will be used for all integrations and refactored in my next Jira Server search history PR to use a single table for all search history.

packages/client/components/JiraScopingSearchBar.tsx
igorlesnenko: @nickoferrall Re-render happens because we use relay store as a state and request a search query in a parent fragment.
```
jiraSearchQuery {
  projectKeyFilters
  queryString
  isJQL
}
```
I wrapped `ScopingSearchHistoryToggle` and `ScopingSearchBar` into their own `Jira` prefixed components with their own fragments. It solved the issue.
In my next PR I'm going to try to add more refactoring to `ScopingSearchHistoryToggle` as we are going to use a single table for all search queries.

I'm curious, do we really need to use this approach with storing a search query in the relay store?

packages/client/components/JiraServerScopingSearchBar.tsx
```diff
@@ -0,0 +1,71 @@
...
+            }
+          }
+        }
+        ...JiraServerScopingSearchFilterToggle_meeting
+      }
+    `,
+    meetingRef
+  )
+
+  const {id, viewerMeetingMember, jiraServerSearchQuery} = meeting
```
nickoferrall, line 44: +1 convention in the repo is to name it `meetingId`, so `id: meetingId`

nickoferrall (approved):

packages/client/components/JiraScopingSearchBar.tsx
igorlesnenko: @nickoferrall I was thinking about using a react context, feels like it could be less hacky, though I need to check. What do you think?

nickoferrall: I remember @mattkrick saying that React Context is nice and easy to use, but the way it's implemented isn't great for performance. 

I now quite like being able to query global state in Relay in the same way that we query the db, but I do remember finding it confusing when I first discovered the pattern. 

My main concern with putting the search query in React Context is we'd be storing global state in two places which could be confusing.

packages/client/components/JiraScopingSearchHistoryToggle.tsx
```diff
@@ -1,73 +1,88 @@
...
+        PersistJiraSearchQueryMutation(atmosphere, {
+          teamId,
+          input: {queryString, isJQL, projectKeyFilters, isRemove: true}
+        })
       }
-    }
-  `
-})
+
+      const handleRemoveJiraSearchQueryClick = () => {
```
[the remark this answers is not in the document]
igorlesnenko: ah sorry didn't push this commit, removed

packages/client/components/JiraScopingSearchHistoryToggle.tsx
```diff
@@ -1,73 +1,85 @@
...
+        })
+      }
+      const queryStringLabel = isJQL ? queryString : `“${queryString}”`
+      const projectFilters = projectKeyFilters
+        .map((filter) => filter.slice(filter.indexOf(':') + 1))
+        .join(', ')
+
+      return {
+        id,
+        ariaLabel: 'Remove this Jira Search Query',
```
Dschoordsch, line 64: +1 We could make this ariaLabel a bit more generic and not pass it down but hard code it: "Remove this search query"

Dschoordsch, line 64: +1 We could make this ariaLabel a bit more generic and not pass it down but hard code it: "Remove this search query"

packages/client/components/JiraScopingSearchHistoryToggle.tsx
```diff
@@ -1,73 +1,85 @@
...
+        labelFirstLine: queryStringLabel,
+        labelSecondLine: projectFilters && `in ${projectFilters}`,
+        onClick: selectQuery,
+        onDelete: () => {
```
Dschoordsch, line 68: +1 you could make this a `const deleteQuery = ...` to be aligned with the `selectQuery`

Dschoordsch, line 68: +1 you could make this a `const deleteQuery = ...` to be aligned with the `selectQuery`

packages/client/components/ScopingSearchInput.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const {placeholder, queryString, meetingId, linkedRecordName} = props
+  const atmosphere = useAtmosphere()
+  const inputRef = useRef<HTMLInputElement>(null)
+  const isEmpty = !queryString
+
+  const setSearch = (meetingId: string, value: string) => {
+    commitLocalUpdate(atmosphere, (store) => {
+      const meeting = store.get(meetingId)
+      if (!meeting) return
+      const jiraSearchQuery = meeting.getLinkedRecord(linkedRecordName)!
```
Dschoordsch, line 51: -1 please rename to `searchQuery`

Dschoordsch, line 51: -1 please rename to `searchQuery`

packages/client/shared/gqlIds/IntegrationRepoId.ts
```diff
@@ -12,28 +12,43 @@ type JiraRepoIntegration = {
...
-  join: (integration: GitHubRepoIntegration | JiraRepoIntegration | RepoIntegration) => {
+  join: (integration: GitHubRepoIntegration | JiraRepoIntegration | RepoIntegration | JiraServerRepoIntegration) => {
     const {service} = integration
     switch (service) {
       case 'github':
         return integration.nameWithOwner
       case 'jira':
         return JiraProjectId.join(integration.cloudId, integration.projectKey)
+      case 'jiraServer':
+        return `${integration.service}:${integration.providerId}:${integration.id}:${integration.key}`
```
Dschoordsch, line 37: -1 `integration.key` is redundant since we already have the `id`

Dschoordsch, line 37: -1 `integration.key` is redundant since we already have the `id`

packages/server/graphql/types/JiraServerIntegration.ts
```diff
@@ -83,12 +91,8 @@ const JiraServerIntegration = new GraphQLObjectType<{teamId: string; userId: str
...
         }
       },
-      resolve: async (
-        {teamId, userId},
-        // {first, queryString, isJQL, projectKeyFilters},
-        {first},
-        {authToken, dataLoader}: GQLContext
-      ) => {
+      resolve: async ({teamId, userId}, args: any, {authToken, dataLoader}) => {
+        const {first, queryString, isJQL, projectKeyFilters} = args as IssueArgs
```
Dschoordsch, line 95: -1 If you type args, then do it in the function signature

Dschoordsch, line 95: -1 If you type args, then do it in the function signature

Dschoordsch (changes requested):
Only minor stuff

packages/server/graphql/types/JiraServerIntegration.ts
igorlesnenko: @Dschoordsch I took the approach from here 
https://github.com/ParabolInc/parabol/commit/072cdfcbe5cf1a5eed3a9c72ec1da51756d2b9fe#diff-00c4ea32101ae6a99d7e1beea2cfe1d6e9fa38b1a1c11d010558fa2c3db37b0dR111

Attempting to set type in function signature cause an error on `fields: () =>` above: 

image

Dschoordsch (changes requested):
Works for the most part. When entering only 1 character for search I get the message back, that no issues match, while it would be better to show a message like "Search term too short, enter at least 2 characters"
JiraServer search wrong message

packages/client/components/JiraScopingSearchFilterMenu.tsx
```diff
@@ -45,27 +43,32 @@ const FilterLabel = styled(DropdownMenuLabel)({
...
 
-type JiraSearchQuery = NonNullable<
-  NonNullable<JiraScopingSearchFilterMenu_viewer['meeting']>['jiraSearchQuery']
->
-
 const getValue = (item: {name: string}) => item.name
 
 const MAX_PROJECTS = 10
 
 const JiraScopingSearchFilterMenu = (props: Props) => {
```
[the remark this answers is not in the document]
Dschoordsch: I don't know what happened here, but somehow GitHub messed up my review comments badly, this is a leftover from the previous one.

Dschoordsch (approved):

Issue #6261, opened 2022-03-23
Standup UI - Response Reactjis 


BartoszJarocki:
I've added a reactable fields here but I guess it'd be best to create a task for the implementation of that. DB has the list of reactjis for each response but I'm not sure if the data format matches 100%.

_Originally posted by @BartoszJarocki in https://github.com/ParabolInc/parabol/pull/6171#discussion_r832922673_

BartoszJarocki:
**Story Points: 5**
  [See the discussion](https://action.parabol.co/meet/chorM4RmGA/estimate/1) in PS2 Poker 1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

BartoszJarocki:
**Include?: Y**
  [See the discussion](https://action.parabol.co/meet/chorM4RmGA/estimate/2) in PS2 Poker 1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

BartoszJarocki:
This issue's estimation was terribly wrong and would it would benefit from planning and further splitting

PR #6407, opened 2022-04-14
feat(standups): Response Reactjis

jmtaber129:
Fixes https://github.com/ParabolInc/parabol/issues/6261

Needs rebase once https://github.com/ParabolInc/parabol/pull/6388 lands.

https://www.loom.com/share/d72412e5897348d4b1d1f72e7f0088fc


jmtaber129:
@BartoszJarocki I would recommend selecting the diff for only the most recent commits (e.g. https://github.com/ParabolInc/parabol/pull/6407/files/47033c3ab5c31ca6960ab6b414a8edfd12f250c3..8abfd81d2d0ce8d42f866aa19e70a5e693cc96e7) and reviewing that.

IIRC the branch for https://github.com/ParabolInc/parabol/pull/6388 is missing some commits on `master`, so I'll want to hold off on changing the base to that

jmtaber129:
@BartoszJarocki this should be good to review now.

packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
```diff
@@ -60,13 +59,15 @@ const upsertTeamPromptResponse: MutationResolvers['upsertTeamPromptResponse'] =
...
   }
 
-  const teamPromptResponseId = await upsertTeamPromptResponseQuery({
-    meetingId,
-    userId: viewerId,
-    sortOrder: 0, //TODO: placeholder as currently it's defined as non-null. Might decide to remove the column entirely later.
-    content,
-    plaintextContent
-  })
+  const teamPromptResponseId = TeamPromptResponseId.join(
```
BartoszJarocki, line 62: -1 when we decide to move the id splitting/joining id part to DB queries level we should also return the proper id from that query so we don't have to remember to do that manually ie. let's try to keep DB queries as a single source of truth about id and do not use `TeamPromptResponseId` outside of them at all.

packages/client/components/ReflectionCard/ReactjiCount.tsx
```diff
@@ -62,7 +62,8 @@ const ReactjiCount = (props: Props) => {
   const {onToggle, reactji, status, onTransitionEnd} = props
   if (!reactji) return null
   const {count, id, isViewerReactji} = reactji
-  const [, name = ''] = id.split(':')
+  const splitIndex = id.lastIndexOf(':')
+  const name = splitIndex !== -1 ? id.slice(splitIndex + 1) : ''
```
BartoszJarocki, line 66: out of curiosity, what's the reason for this change?

BartoszJarocki (changes requested):
just one little thing related to ID handling, great work!

packages/client/components/ReflectionCard/ReactjiCount.tsx
jmtaber129: Previously, the `id` would be of the form `<generated UUID>:<name>` so the name is always at index 1 after splitting, but with the team prompt response ID changes, the `id` is of the form `teamPromptResponse:<number>:<name>` so it's instead the last element after splitting, or the part of the string after the last index of `:`

packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
jmtaber129: That makes sense.  Moved the interop between the prefixed graphql queries and the numerical DB queries into the SQL queries themselves.

BartoszJarocki: I thought about moving into ts part but what you did looks even better, awesome work!

BartoszJarocki (approved):

packages/client/components/ReflectionCard/ReactjiCount.tsx
mattkrick: +1 can we refactor this to use the gqlId helper pattern? https://github.com/ParabolInc/parabol/blob/2708b9553cc212d6da9be951c6b486e46f57b68d/packages/client/shared/gqlIds/JiraProjectId.ts

That way future devs won't have to know how an ID is composed, they can just call `reactjiId.split(id)`

packages/server/graphql/mutations/addReactjiToReactable.ts
```diff
@@ -69,11 +76,15 @@ const addReactjiToReactable = {
     const subOptions = {mutatorId, operationId}
 
     //AUTH
-    const dbTable = tableLookup[reactableType]
-    const reactable = (await r
-      .table(dbTable)
-      .get(reactableId)
-      .run()) as Reactable
+    let reactable: Reactable
+    const isPgTable = reactableType === 'RESPONSE'
```
mattkrick, line 80: +1 super nitty, but if you do the lookup here (`const pgLoaderName = pgDataloaderLookup[reactableType]`) then you'll save a line & it'll future proof this bit of code for when we introduce more types.

packages/server/graphql/mutations/addReactjiToReactable.ts
```diff
@@ -69,11 +76,15 @@ const addReactjiToReactable = {
...
+    const rethinkDbTable = rethinkTableLookup[reactableType]
+    if (isPgTable) {
+      const loaderName = pgDataloaderLookup[reactableType]
+      reactable = (await dataLoader.get(loaderName).load(reactableId)) as Reactable
```
mattkrick, line 84: +1 `.loadNonNull(reactableId)` and then you can remove the casting

mattkrick: +1 clear the dataloader cache right here on L85 so you don't need to do it at the bottom of the function

packages/server/postgres/queries/generated/getTeamPromptResponsesByIdsQuery.ts
```diff
@@ -26,13 +27,13 @@ export interface IGetTeamPromptResponsesByIdsQueryQuery {
...
+const getTeamPromptResponsesByIdsQueryIR: any = {"name":"getTeamPromptResponsesByIdsQuery","params":[{"name":"ids","codeRefs":{"defined":{"a":53,"b":55,"line":3,"col":9},"used":[{"a":318,"b":320,"line":6,"col":46}]},"transform":{"type":"array_spread"}}],"usedParamSet":{"ids":true},"statement":{"body":"SELECT CONCAT('teamPromptResponse:', \"id\") as id, \"createdAt\", \"updatedAt\", \"meetingId\", \"userId\", \"sortOrder\", \"content\", \"plaintextContent\", to_json(\"reactjis\") as \"reactjis\" FROM \"TeamPromptResponse\"\nWHERE CONCAT('teamPromptResponse:', \"id\") in :ids","loc":{"a":69,"b":320,"line":5,"col":0}}};
 
 /**
  * Query generated from SQL:
  * ```
- * SELECT "id", "createdAt", "updatedAt", "meetingId", "userId", "sortOrder", "content", "plaintextContent" FROM "TeamPromptResponse"
- * WHERE id in :ids
+ * SELECT CONCAT('teamPromptResponse:', "id") as id, "createdAt", "updatedAt", "meetingId", "userId", "sortOrder", "content", "plaintextContent", to_json("reactjis") as "reactjis" FROM "TeamPromptResponse"
+ * WHERE CONCAT('teamPromptResponse:', "id") in :ids
```
mattkrick, line 36: -1 i'd argue this is a leaky abstraction, the database shouldn't need to be aware that we prefix its IDs in order to make them globally unique for the client. Ideally, we should be able to change the prefix that gets sent to the client without affecting anything that touches the database.

The standard practice that I've seen (https://github.com/graphql/graphql-relay-js#object-identification) does the following:
- when a GUID comes in via a mutation argument, it's transformed to a database ID at the top of the resolve function
- all server functions (db helpers, dataloaders, lookups, etc) use the database ID
- in the GraphQL type resolve function, the database ID is transformed into a GUID before going back to the client

In doing so, the scope of the prefix is limited to just the ingress/egress.
happy to debate or change thinking though! just want to make sure we do this the same way with every entity (e.g. https://github.com/ParabolInc/parabol/blob/c431a3b26753a7cb57e8820677df43c0a766a9e5/packages/server/graphql/types/Poll.ts#L26)

mattkrick (changes requested):
only thing to mention is where we do the GUID transformation (in graphql or in the DB). This would be a good convo for the whole product team to have! I'd say bring it up in slack & see what everything thinks, happy to go either way as long as we're all on the same page. You can use my last comment as a point to argue for/against

packages/server/graphql/mutations/addReactjiToReactable.ts
jmtaber129: I disagree with clearing the cache here, because although we've performed the only read before the cache is invalidated, we haven't actually done anything to invalidate the cache.  The current placement of the cache-clearing logic is immediately after the underlying data is mutated.

jmtaber129: I've made this a bit clearer by rearranging the `isPg` and `isRemove` branches such that the cache-clearing happens at the end of the `isPg` branch containing both removals and additions

packages/client/components/ReflectionCard/ReactjiCount.tsx
jmtaber129: Done

packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
```diff
@@ -20,10 +20,9 @@ const upsertTeamPromptResponse: MutationResolvers['upsertTeamPromptResponse'] =
 
   // VALIDATION
   if (inputTeamPromptResponseId) {
-    const responseId = TeamPromptResponseId.split(inputTeamPromptResponseId)
     const teamPromptResponse: TeamPromptResponse = await dataLoader
       .get('teamPromptResponses')
-      .load(responseId)
+      .load(inputTeamPromptResponseId)
```
BartoszJarocki, line 25: +1 I'm a little lost, where do we convert the ID now? wouldn't it be better to keep it as is to keep the PR as small as possible?

packages/server/graphql/types/TeamPromptResponse.ts
```diff
@@ -14,8 +13,7 @@ const TeamPromptResponse: GraphQLObjectType = new GraphQLObjectType<any, GQLCont
     id: {
       type: new GraphQLNonNull(GraphQLID),
       description:
-        'Team prompt response id in a format of `teamPromptResponse:idGeneratedByDatabase`',
-      resolve: ({id}) => TeamPromptResponseId.join(id)
+        'Team prompt response id in a format of `teamPromptResponse:idGeneratedByDatabase`'
```
BartoszJarocki, line 16: +1 same here, wouldn't it be better to not touch this in this PR?

BartoszJarocki (commented):
a couple of clarifying questions, otherwise lgtm!

packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
jmtaber129: The conversion (both joining and splitting) happens within the dataloader when the dataloader is used.  This reduces some complexity around the custom `reactables` loader, which would otherwise need to do some special casing to handle both number IDs and string IDs.  I originally tried doing things that way and leaving these IDs as-is, but IIRC it ended up breaking some existing abstractions and would've been difficult to refactor.

packages/server/postgres/queries/generated/getTeamPromptResponsesByIdsQuery.ts
jmtaber129: In the interest of getting this landed (especially since it's still behind a feature flag), I've reverted the change that makes this a leaky abstraction, back towards the existing GQLID patterns.

I'm going to put some thought into how this should be refactored in the near-/medium-future and start a discussion with product.

jmtaber129:
@Dschoordsch adding you since Matt's out.  I've addressed most of Matt's review, with the one big open thing being around IDs.  Since this ID question has already pushed the scope of this PR quite a bit (and this is all still behind a feature flag), I have a strong preference towards landing this with the ID patterns as-is (mod any minor requested changes) and discuss the ID approach with other engineers per Matt's comment as a follow-up.  Any follow-on refactors around these IDs should be code-only (i.e. no migrations), so I think this is safe to try.

packages/server/postgres/queries/getTeamPromptResponsesByIds.ts
```diff
@@ -1,21 +1,41 @@
...
+export const mapToTeamPromptResponse = (
+  results: IGetTeamPromptResponsesByIdsQueryResult[]
+): TeamPromptResponse[] => {
+  return results.map((teamPromptResponse: any) => {
+    return {
+      ...teamPromptResponse,
+      id: TeamPromptResponseId.join(teamPromptResponse.id),
+      reactjis: teamPromptResponse.reactjis.map(
+        (reactji: {shortname: string; userid: string}) =>
+          new Reactji({id: reactji.shortname, userId: reactji.userid})
```
Dschoordsch, line 27: +1 would be nice to use the same structure `{id, userId}` in Postgres so we could get away with just casting.

Dschoordsch (approved):

Dschoordsch:
I did not test the change.

packages/server/postgres/queries/getTeamPromptResponsesByIds.ts
Dschoordsch: Could you elaborate what you mean by SQL's case-insensitivity? The columns have the exact case as defined.

Issue #5131, opened 2021-07-13
Add tooltip to long agenda item names


nickoferrall:
If the agenda item name is too long to show the entire name, display a tooltip on hover so that users can still read the agenda item.

![tooltip](https://user-images.githubusercontent.com/39854876/125484320-9e2af5f3-67f6-4944-9960-e3bbe0696532.png)

### AC:

- [ ] If the agenda item name requires an ellipsis, show a tooltip when hovering over the text of the name

jordanh:
Would love to tackle this one the same time we do #4600

PR #6410, opened 2022-04-16
feat: editable agenda items

### LINKED ISSUE #4600: Editable agenda items
## Issue - Enhancement

Truly I'm a bit shocked it's taken until now to hear this from a user.

[A user writes ](https://app.hubspot.com/contacts/3888472/contact/10876051)🔒 :
> Discussion Topics are not editable
> It didn’t appear that we could edit a discussion topic when it had been created
> Workaround was to delete and re-write

Users can:
   - have an easy affordance (e.g. does the cursor change to an edit bone?) to edit agenda item names
   - 
Also:
   - internal validation performed
   - implementation issue created

- **Estimated effort:** 6 hours

shapel:
# Description

Fixes/Partially Fixes #4600, #5131 

## Demo

<a href="https://www.loom.com/share/f75a096ff61f47bf971aae19eaf693b2">
    <p>Agenda items - 28 May 2022 - Watch Video</p>
    
  </a>

### Functionality
- [ ] As a team member, I should be able to add/edit/delete/pin/unpin an agenda item on the team page
  - Open a team page.
  - Add new agenda item.
  - Click the ":" button near the agenda item in right corner (will be visible on hover).
  - Try to edit/pin/unpin/delete the agenda item

- [ ] As a team member, I should not be able to edit /  delete agenda item for already completed meeting. But I should be able to pin / unpin agenda items for completed meeting.
  - Complete some Check-in meeting with some agenda items 
  - Open the Timeline page.
  - Click the "See the discussion" link for the meeting.
  - Try to change any agenda item.

- [ ] As a team member, I should be able to see full validation error during agenda item editing.
  - Open a team page.
  - Click the ":" button near some agenda item in right corner (will be visible on hover).
  - Delete all text and press Enter.

- [ ] As a team member, I should to be able to see the longest agenda item.
  - Open a team page.
  - Add an agenda item with long text.
  - A full text tooltip should be displayed on hover for such an item.

### Design 
- [ ] Menu with icons should have the same design (padding/height of elements) throughout the system.
  - Help menu
    -  Click on help icon at top bar.
  - Meeting options menu
    - Open meetings page. 
    - Click the ":" button in the right corner of any meeting to see the menu.
  - Comment options menu
    - Start sprint poker meeting.
    - Add comment during estimation stage of some item.
    - Click the ":" button next to the comment.
  - Agenda item options menu 
    - Open team page. 
    - Click the ":" button near some agenda item in right corner (will be visible on hover).

- [ ] Headings within the agenda item options menu should follow the design guidelines.
    - Open team page. 
    - Click the ":" button near some agenda item in right corner (will be visible on hover).
    - Check that menu item name follow the design guidelines.

## Final checklist

### DIFF
--- a/packages/client/components/MenuItemWithIcon.tsx
+++ b/packages/client/components/MenuItemWithIcon.tsx
@@ -3,29 +3,21 @@ import React from 'react'
 import {PALETTE} from '~/styles/paletteV3'
 import {ICON_SIZE} from '~/styles/typographyV2'
 import Icon from './Icon'
-
-const Wrapper = styled('div')({
-  alignItems: 'center',
-  display: 'flex',
-  paddingLeft: 8,
-  paddingRight: 8,
-  width: '100%'
-})
+import MenuItemLabel from './MenuItemLabel'
 
 const ItemIcon = styled(Icon)({
   color: PALETTE.SLATE_600,
   fontSize: ICON_SIZE.MD24,
-  padding: 8
+  marginRight: 8
 })
 
-const Label = styled('div')({
-  color: PALETTE.SLATE_700,
-  fontSize: 14,
-  lineHeight: '32px'
+const IconWrapper = styled('span')({
+  opacity: 0.7,
+  height: ICON_SIZE.MD24,
+  marginRight: 8
 })
-
 interface Props {
-  icon: string
+  icon: string | React.ReactElement
   label: string
   dataCy: string
 }
@@ -33,10 +25,10 @@ interface Props {
 const MenuItemWithIcon = (props: Props) => {
   const {icon, label, dataCy} = props
   return (
-    <Wrapper data-cy={`${dataCy}`}>
-      <ItemIcon>{icon}</ItemIcon>
-      <Label>{label}</Label>
-    </Wrapper>
+    <MenuItemLabel data-cy={dataCy}>
+      {typeof icon === 'string' ? <ItemIcon>{icon}</ItemIcon> : <IconWrapper>{icon}</IconWrapper>}
+      <span>{label}</span>
+    </MenuItemLabel>
   )
 }
 
--- a/packages/client/modules/teamDashboard/components/AgendaItem/AgendaItem.tsx
+++ b/packages/client/modules/teamDashboard/components/AgendaItem/AgendaItem.tsx
@@ -2,6 +2,16 @@ import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
 import React, {useEffect, useRef, useState} from 'react'
 import {createFragmentContainer} from 'react-relay'
+import EditableText from '~/components/EditableText'
+import Menu from '~/components/Menu'
+import MenuItem from '~/components/MenuItem'
+import MenuItemHR from '~/components/MenuItemHR'
+import MenuItemWithIcon from '~/components/MenuItemWithIcon'
+import useMenu from '~/hooks/useMenu'
+import useMutationProps from '~/hooks/useMutationProps'
+import useTooltip from '~/hooks/useTooltip'
+import Legitity from '~/validation/Legitity'
+import {AGENDA_ITEM_MAX_CHARS, contentValidator} from '~/validation/makeAgendaItemSchema'
 import {AgendaItem_meeting} from '~/__generated__/AgendaItem_meeting.graphql'
 import Avatar from '../../../../components/Avatar/Avatar'
 import IconButton from '../../../../components/IconButton'
@@ -10,34 +20,16 @@ import useAtmosphere from '../../../../hooks/useAtmosphere'
 import {MenuPosition} from '../../../../hooks/useCoords'
 import useGotoStageId from '../../../../hooks/useGotoStageId'
 import useScrollIntoView from '../../../../hooks/useScrollIntoVIew'
-import useTooltip from '../../../../hooks/useTooltip'
 import RemoveAgendaItemMutation from '../../../../mutations/RemoveAgendaItemMutation'
 import UpdateAgendaItemMutation from '../../../../mutations/UpdateAgendaItemMutation'
 import pinIcon from '../../../../styles/theme/images/icons/pin.svg'
 import unpinIcon from '../../../../styles/theme/images/icons/unpin.svg'
-import {ICON_SIZE} from '../../../../styles/typographyV2'
 import findStageAfterId from '../../../../utils/meetings/findStageAfterId'
 import {AgendaItem_agendaItem} from '../../../../__generated__/AgendaItem_agendaItem.graphql'
 
 const AgendaItemStyles = styled('div')({
-  position: 'relative',
-  // show the DeleteIconButton on hover
-  '&:hover > button': {
-    opacity: 1
-  }
+  position: 'relative'
 })
-
-const DeleteIconButton = styled(IconButton)<{disabled?: boolean}>(({disabled}) => ({
-  display: 'block',
-  // we can make the position of the del (x) more centered when there’s a low number of agenda items
-  left: 19,
-  lineHeight: ICON_SIZE.MD18,
-  opacity: 0,
-  position: 'absolute',
-  top: '.6875rem',
-  transition: 'opacity .1s ease-in',
-  visibility: disabled ? 'hidden' : undefined
-}))
 
 const IconBlock = styled('div')({
   display: 'flex',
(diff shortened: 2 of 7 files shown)


packages/client/validation/makeAgendaItemSchema.ts
```diff
@@ -1,10 +1,18 @@
-import {compositeId, id} from './templates'
 import legitify from './legitify'
 import Legitity from './Legitity'
+import {compositeId, id} from './templates'
+
+export const AGENDA_ITEM_MAX_CHARS = 63
+
+export const content = (value: Legitity) =>
```
shapel, line 7: I'm not sure about naming here...

I think it's better to keep the validation logic for some model in one module (high coupling). This way we avoid  inconsistent logic changes associated with this model.

So I keep it there instead of moving to `packages/client/validation/templates.ts` or similar.

nickoferrall:
Thanks for the reminder @shapel! I've been away since Thursday, but I'll do my best to review this PR this week 👍

packages/client/validation/makeAgendaItemSchema.ts
Dschoordsch: +1 `contentValidator` might be a bit better

Dschoordsch (changes requested):
Testing failed: When trying to edit an agenda item as facilitator, the meeting also moves to the corresponding stage. This is not what I want, it should be possible to edit the item without advancing the meeting

shapel:
👋 @ackernaut @enriquesanchez could you plz help me with a UX challenge of implementing ability to edit agenda item.

I would like you to share your thoughts on how to better deal with the fact that there are many UI actions in the system related to the agenda item, for example:
* delete (left corner on hover)
* pin/unpin (right corner on hover)
* rearrange (drag & drop)
* move on to the agenda item (by clicking)
   * plus in feature potentially allow the facilitator to switch to other agenda item without advancing meeting #6418
* edit (need to introduce it somehow)

I suggest to introduce a menu in the right corner on hover that will contain all these actions, like this:
* move to different stages (by clicking)
* rearrange (drag)
* menu (right corner on hover)
   * pin/unpin
   * edit
   * delete
   * switch to item without advancing the meeting #6418

I know this introduces another click for some actions, but showing another icon in the right corner (edit + pin/unpin) on hover doesn't look so good for me. As well as the fact that trying to switch to edit mode on click leads to misunderstandings, as @Dschoordsch mentioned.

enriquesanchez:
Hi @shapel 👋 

Thank you so much for your work here! You make a good observation that having to many actions fall onto one UI element introduces problems.

I agree that an ellipsis menu might be the right call here, especially now that we have quite a few actions to consider.

> menu (right corner on hover)
> - pin/unpin
> - edit
> - delete
> - switch to item without advancing the meeting

The only reservation I have is with the option to switch to item without advancing the meeting. I feel that one needs a little more exploration because I feel there are other solutions to consider.

I think a good next step to try is to only have these 3 actions inside the menu:

> - pin/unpin
> - edit
> - delete

How does that feel to you?

nickoferrall (changes requested):
Thanks for the PR! Two issues that I'm running into:

1. I can't read the whole error message:
err-message
 
2. If the agenda item text is long, I can't see the edit icon

edit-icon

shapel:
Hi @enriquesanchez ✋
What do you think about how to better deal with the fact @nickoferrall mentioned that it's hard to read the error message while editing an agenda item due to the small area size?

I think we could:
1 Show error message with an ellipsis, show a tooltip when hovering over the text of the error.
2 Try using a shorter message like "the field is required" instead of "Please enter an agenda item".
The problem here is what to do in case of any server error (with long text) and it needs to be shown here.
3 Use a dialog to edit the agenda item. But it seems that such a template is not used in the system.

As far as I'm concerned, the 1th option looks like a pretty good solution.

enriquesanchez:
Thanks for the ping @shapel!

Is it possible to have the error message wrap into more lines? Something like:

```
Please enter an
agenda item
```

The reason I suggest this is that, like you mentioned, these error messages can be longer, and there's also the scenario where a user might have their browser text set to a bigger font size. So the element needs to be flexible and adapt.

shapel:
Hi @enriquesanchez 
Can you help me understand the naming convention for the menu item? 

I see several patterns in the system such as:
* Screenshot 2022-05-28 at 14 18 38
* Screenshot 2022-05-28 at 14 18 51
* Screenshot 2022-05-28 at 14 19 15

For the agenda item options menu, I use the same pattern as the comment options menu. Was it the right choice or not?
Screenshot 2022-05-28 at 14 21 20

packages/client/components/MenuItemWithIcon.tsx
```diff
@@ -3,40 +3,32 @@ import React from 'react'
...
 })
 
-const Label = styled('div')({
-  color: PALETTE.SLATE_700,
-  fontSize: 14,
-  lineHeight: '32px'
+const IconWrapper = styled('span')({
+  opacity: 0.7,
+  height: ICON_SIZE.MD24,
+  marginRight: 8
```
nickoferrall, line 17: Why do the opacity & margins in these two styled-components need to be updated? Would you mind giving me an example use case / screenshot? We generally try to use padding wherever possible rather than margin as margin can throw off other components' positioning.

Also, when components like this are being reused throughout the app, I'm also hesitant to update them unless there's no other option as we'll need to check the UI at different breakpoints in each of the components that it's being used.

packages/client/modules/teamDashboard/components/AgendaItem/AgendaItem.tsx
```diff
@@ -186,19 +285,32 @@ const AgendaItem = (props: Props) => {
...
-          palette='midGray'
-        />
       </AgendaItemStyles>
       {tooltipPortal(
-        pinned ? `Unpin "${content}" from every check-in` : `Pin "${content}" to every check-in`
+        endedAt
+          ? pinned
+            ? `Unpin "${content}" from every check-in`
+            : `Pin "${content}" to every check-in`
+          : `Select the action for "${content}"`
```
nickoferrall, line 311: -1 I don't think we need a tooltip for `Select the action` as I feel that the menu icon is clear enough. There's a risk of having too many tooltips and it becoming annoying for the user

packages/client/modules/teamDashboard/components/AgendaItem/AgendaItem.tsx
```diff
@@ -171,12 +267,15 @@ const AgendaItem = (props: Props) => {
...
+                tooltipRef.current = ref
+                menuRef.current = ref
+              }}
+              onClick={endedAt ? handleIconClick : togglePortal}
               onMouseOver={openTooltip}
               onMouseOut={closeTooltip}
-              ref={originRef}
             >
-              {getIcon()}
+              {getIcons()}
```
nickoferrall, line 278: +1 we're only displaying one icon so I think `getIcon` is more explicit

packages/client/modules/teamDashboard/components/AgendaItem/AgendaItem.tsx
```diff
@@ -153,15 +157,107 @@ const AgendaItem = (props: Props) => {
...
+      if (!pinned) return <Avatar hasBadge={false} picture={picture} size={24} />
+      else return <SvgIcon alt='pinnedIcon' src={pinIcon} />
+    }
   }
 
+  const getMenu = () => {
+    return (
+      <Menu ariaLabel={'Select the action for the agenda item'} {...menuProps}>
+        {pinned ? (
+          <MenuItem
```
nickoferrall, line 185: +1 alternatively we could create an array of objects that contain the label, icon, and onClick function for each item. We could then map through the array and use the label as a key. That way, we wouldn't repeat the menuItem logic and we'd reduce the size of this already large file.

packages/client/modules/teamDashboard/components/AgendaItem/AgendaItem.tsx
```diff
@@ -153,15 +157,107 @@ const AgendaItem = (props: Props) => {
...
+    submitMutation()
+    RemoveAgendaItemMutation(atmosphere, {agendaItemId}, {meetingId, onError, onCompleted})
   }
 
-  const getIcon = () => {
-    if (pinned && isHovering) return <SvgIcon alt='unpinIcon' src={unpinIcon} />
-    else if (!pinned && !isHovering) return <Avatar hasBadge={false} picture={picture} size={24} />
-    else return <SvgIcon alt='pinnedIcon' src={pinIcon} />
+  const getIcons = () => {
+    if (endedAt) {
```
nickoferrall, line 166: +2 nice touch only showing the delete logic for active meetings

nickoferrall (changes requested):
This is looking good! Just a few small changes requested

PR #6411, opened 2022-04-18
fix: ignore comparison order for equality

shapel:
unlock: #6239 

`checkRethinkPgEquality` should handle cases where the fields are an array with different element order.

### DIFF
--- a/packages/server/postgres/utils/__tests__/checkEqBase.test.ts
+++ b/packages/server/postgres/utils/__tests__/checkEqBase.test.ts
@@ -17,7 +17,7 @@ describe('areEqual', () => {
     expect(areEqual(a, b)).toBe(true)
     expect(areEqual(b, a)).toBe(true)
   })
-  it('should ignore object keys with `undefined` value', async () => {
+  it("should ignore object keys with `undefined` value if they're missing in another object", async () => {
     const a = {
       issueKey: undefined,
       fieldName: 'Story point estimate',
@@ -33,6 +33,18 @@ describe('areEqual', () => {
 
     expect(areEqual(a, b)).toBe(true)
     expect(areEqual(b, a)).toBe(true)
+  })
+  it("should not ignore object keys with `undefined` value if they're present in another object", () => {
+    const a = {
+      issueKey: undefined,
+      fieldName: 'Story point estimate'
+    }
+    const b = {
+      issueKey: 'issue key'
+    }
+
+    expect(areEqual(a, b)).toBe(false)
+    expect(areEqual(b, a)).toBe(false)
   })
   it('should ignore the order of the elements in the array', () => {
     const a = [{k: 1}, {k: 2}, {k: 3}]
--- a/packages/server/postgres/utils/__tests__/checkEqBase.test.ts
+++ b/packages/server/postgres/utils/__tests__/checkEqBase.test.ts
@@ -1,0 +1,51 @@
+import {areEqual} from '../checkEqBase'
+
+describe('areEqual', () => {
+  it('should ignore the order of keys in objects', () => {
+    const a = {
+      fieldName: 'Story point estimate',
+      fieldType: 'number',
+      fieldId: 'customfield_10016'
+    }
+
+    const b = {
+      fieldId: 'customfield_10016',
+      fieldName: 'Story point estimate',
+      fieldType: 'number'
+    }
+
+    expect(areEqual(a, b)).toBe(true)
+    expect(areEqual(b, a)).toBe(true)
+  })
+  it('should ignore object keys with `undefined` value', async () => {
+    const a = {
+      issueKey: undefined,
+      fieldName: 'Story point estimate',
+      fieldType: 'number',
+      fieldId: 'customfield_10016'
+    }
+
+    const b = {
+      fieldName: 'Story point estimate',
+      fieldType: 'number',
+      fieldId: 'customfield_10016'
+    }
+
+    expect(areEqual(a, b)).toBe(true)
+    expect(areEqual(b, a)).toBe(true)
+  })
+  it('should ignore the order of the elements in the array', () => {
+    const a = [{k: 1}, {k: 2}, {k: 3}]
+    const b = [{k: 3}, {k: 1}, {k: 2}]
+
+    expect(areEqual(a, b)).toBe(true)
+    expect(areEqual(b, a)).toBe(true)
+  })
+  it('should check all elements in array', () => {
+    const a = [{k: 1}, {k: 1}, {k: 2}]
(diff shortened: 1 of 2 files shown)


packages/server/postgres/utils/checkEqBase.ts
```diff
@@ -1,7 +1,18 @@
...
 import {RTable, TableSchema} from '../../database/stricterR'
-import areEqual from 'fbjs/lib/areEqual'
 import getPg from '../getPg'
 
+function areEqualIgnoringOrder(a, b): boolean {
+  if (Array.isArray(a) && Array.isArray(b)) {
+    return (
+      a.length === b.length &&
+      a.every((aa) => b.some((bb) => areDeep(aa, bb))) &&
+      b.every((bb) => a.some((aa) => areDeep(aa, bb)))
```
shapel, line 10: Not sure if we need to tweak the optimization here (O(n^2) -> O(n) by tracking visited items and etc) and complicate the code.  It depends on the amount of data the system has to process.

Dschoordsch:
@tianrunhe could you please review this?

packages/server/postgres/utils/checkEqBase.ts
```diff
@@ -1,7 +1,18 @@
+import areDeep from 'fbjs/lib/areEqual'
 import {RTable, TableSchema} from '../../database/stricterR'
-import areEqual from 'fbjs/lib/areEqual'
 import getPg from '../getPg'
 
+function areEqualIgnoringOrder(a, b): boolean {
+  if (Array.isArray(a) && Array.isArray(b)) {
+    return (
+      a.length === b.length &&
+      a.every((aa) => b.some((bb) => areDeep(aa, bb))) &&
```
tianrunhe, line 9: `areDeep` won't work for our use case: there could be a field in RethinkDB whose value is `undefined` but in PG the key is absent.

I run your code locally and see the mutation still reports difference for entries like this one below:

```javascript
a = [
  {
    dimensionName: 'Story Points',
    cloudId: 'bba9c111-f180-4970-9b7d-7232bf08f4f3',
    projectKey: 'BRUCE',
    issueKey: undefined,
    fieldName: 'Story point estimate',
    fieldType: 'number',
    fieldId: 'customfield_10016'
  }
]
b = [
  {
    cloudId: 'bba9c111-f180-4970-9b7d-7232bf08f4f3',
    dimensionName: 'Story Points',
    fieldId: 'customfield_10016',
    fieldName: 'Story point estimate',
    fieldType: 'number',
    projectKey: 'BRUCE'
  }
]
```

In fact, that's exactly the reason why we do customized comparison in `getPairNeFields`.
I think to make this actually work, you need a recursion.

tianrunhe (changes requested):

packages/server/postgres/utils/checkEqBase.ts
shapel: @tianrunhe thx for review.
I fixed the cases you mentioned. If you know of other cases, feel free to let me know.

packages/server/postgres/utils/checkEqBase.ts
```diff
@@ -1,7 +1,97 @@
...
+      const _: never = aType
+      return a === b
+  }
+}
+
+function areEqualDate(a, b): boolean | undefined {
+  const isADate = a instanceof Date
+  const isBDate = b instanceof Date
+  if (isADate && isBDate) {
+    return +a === +b
```
tianrunhe, line 45: In production, I've seen a few edge cases where timestamps from PG and RethinkDB are 1 or 2 millisecs different. We can run this in production first and see if there are that many cases and then we can then tweak this comparison.

tianrunhe (approved):
Tests passed. Great work! Thanks for doing this!
I have a future tension but safe to merge this version first and run it in production!

tianrunhe:
Passing to [Maintainer Review](Tests passed. Great work! Thanks for doing this! I have a future tension but safe to merge this version first and run it in production!) per our policy

packages/server/postgres/utils/checkEqBase.ts
```diff
@@ -1,7 +1,97 @@
...
+  if (isARegExp && isBRegExp) {
+    return (
+      a.source === b.source &&
+      a.global === b.global &&
+      a.multiline === b.multiline &&
+      a.ignoreCase === b.ignoreCase
+    )
+  } else if (isARegExp || isBRegExp) {
+    return false
+  }
```
Dschoordsch, line 63: -1 I get warnings when running the test. You need to explicitly `return undefined` at the end, same in other functions

packages/server/postgres/utils/checkEqBase.ts
```diff
@@ -1,7 +1,97 @@
...
+
+      return areEqualObject(a, b)
+    default:
+      // ts check that all types are handled
+      const _: never = aType
+      return a === b
+  }
+}
+
+function areEqualDate(a, b): boolean | undefined {
```
Dschoordsch, line 41: -1 TS complains, we should explicitly type here `a: unknown, b: unknown`, same for other functions

packages/server/postgres/utils/checkEqBase.ts
```diff
@@ -1,7 +1,97 @@
...
+}
+
+function areEqualObject(a: object, b: object): boolean {
+  const aKeys = Object.keys(a).filter((key) => a[key] !== undefined)
+  const bKeys = Object.keys(b).filter((key) => b[key] !== undefined)
+
+  if (aKeys.length !== bKeys.length) return false
+  if (!aKeys.every((key) => bKeys.includes(key))) return false
+
+  for (let i = 0; i < aKeys.length; i++) {
```
Dschoordsch, line 87: +1 for TS to know that `aKeys[i]` is not undefined, it would be better to instead iterate over the values directly
```ts
for (const key of aKeys) {
  if (!areEqual(a[key], b[key]))
```

Dschoordsch (changes requested):
Just some typescript fixes

packages/server/postgres/utils/__tests__/checkEqBase.test.ts
```diff
@@ -0,0 +1,51 @@
...
+    const b = {
+      fieldId: 'customfield_10016',
+      fieldName: 'Story point estimate',
+      fieldType: 'number'
+    }
+
+    expect(areEqual(a, b)).toBe(true)
+    expect(areEqual(b, a)).toBe(true)
+  })
+  it('should ignore object keys with `undefined` value', async () => {
```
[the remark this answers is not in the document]
shapel: done

packages/server/postgres/utils/checkEqBase.ts
shapel: done

PR #6412, opened 2022-04-18
fix: disable rate limiting in test

### LINKED ISSUE #6381: Skipping rate limiter in test does not work
## Issue - Bug

- **How serious is the bug?** P3
- **OS:** all
- **Browser:** N/A

The current behavior to disable rate limiting in test does not work as expected. This is causing failures in https://github.com/ParabolInc/parabol/pull/6183.

Originally, I had implemented the RateLimiter as configurable via the `.env` file (see https://github.com/ParabolInc/parabol/pull/6063). However, I received feedback that we should instead use the `NODE_ENV` environment variable to disable to rate limiter instead (see superseding PR https://github.com/ParabolInc/parabol/pull/6146).

The `NODE_ENV` approach causes problems in the current testing behavior (where we use the development server) and the proposed changes in https://github.com/ParabolInc/parabol/pull/6379, which uses the production build for testing.

In both dev and CI, `process.env.NODE_ENV` is built into the code at compile time by Webpack. So, the environment variable is not checked at runtime.

### Problem in Dev

In dev, the environment variable approach causes problems because the `NODE_ENV` is compiled into the code by Webpack. This makes it difficult for users to easily switch between `development` and `test` `NODE_ENV` values (they must recompile). Additionally, the `NODE_ENV` variable does not appear to be passed to the webpack build in dev (e.g., `NODE_ENV=test yarn build` does not replace `process.env.NODE_ENV` with `test` during the build process).

### Problem in CI

In CI, we have a similar issue today and will also have the same issue once https://github.com/ParabolInc/parabol/pull/6379 is complete. The goal of https://github.com/ParabolInc/parabol/pull/6379 is good - using the `production` build for testing. However, the approach of using the `NODE_ENV` environment variable breaks down that strategy. We'd have to choose **either** testing against a more production-like environment __or__ disabling rate limiting. Many checks through that code check if the environment is production, so tests will be less prod-like following the current disable approach from https://github.com/ParabolInc/parabol/pull/6146.

### StubRateLimiter

We could change the rate-limiting disabling logic back to the initially proposed solution from https://github.com/ParabolInc/parabol/pull/6063. This will allow us to selectively disable rate limiting via the `.env` file in CI and Dev.

### Dyanmically Reference `NODE_ENV` vs. Building via Webpack

Alternatively, we could look into dynamically reading `NODE_ENV` from the environment vs. building the variable into the webpack code. However, this would make tests less production-like and would take longer to accomplish.

### Ignore Localhost-y IP Addresses

At least running on my local machine, I see my IP address as `0000:0000:0000:0000:0000:ffff:7f00:0001` (the IPv6 equivalent of `127.0.0.1`). Assuming this is the same in CI, we could ignore that IP address in the rate limiter.

### Future Use-Case

In the future, we'd like to run integration tests as part of the GitLab CI deployment pipeline. I don't think any of these three ideas above will support that workflow, so we might need to brainstorm the future of our rate-limiting to work around this issue in the longer term. 

As we migrate over to a horizontally-scalable application, we'll need to reconsider the current in-memory rate limiting behavior, anyway, as it is "per instance," not global. It might make sense to treat this as two issues - one short-term to get integration tests working and one longer-term for the future of rate limiting in the new infrastructure environment.

blimmer:
This PR allows disabling rate limiting when testing. See #6381 for details on this approach.

I confirmed that this PR resolves the issues with https://github.com/ParabolInc/parabol/pull/6183

Closes #6381


packages/server/graphql/__tests__/getRateLimiter.test.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  afterEach(() => {
+    jest.useRealTimers()
+  })
+
+  afterAll(() => {
+    process.env.CI = originalRateLimiterEnvVariable
+  })
+
+  it('defaults to the in-memory rate limiter when not in CI', () => {
```
blimmer, line 20: This test ensures we're still rate limiting as expected when not on CI.

packages/server/graphql/getRateLimiter.ts
```diff
@@ -1,11 +1,23 @@
...
+/**
+ * This function returns a stub rate limiter when testing, based on the CI
+ * environment variable.
+ *
+ * Exposed for testing. Use the default export outside of test.
+ *
+ * @private
+ */
+export function _getRateLimiter(): RateLimiter {
+  return process.env.CI ? new StubRateLimiter() : new InMemoryRateLimiter()
```
blimmer, line 20: I confirmed in the built files that we're referencing the environment variable, not building in the static value.

Screen Shot 2022-04-18 at 09 33 44

packages/server/utils/rateLimiters/StubRateLimiter.ts
```diff
@@ -0,0 +1,18 @@
+import {RateLimiter, RateLimitStats} from './RateLimiter'
```
blimmer, line 1: I pulled this over from https://github.com/ParabolInc/parabol/pull/6063

PR #6415, opened 2022-04-18
fix(dx): eslint all files

blimmer:
Today, linting fails in `server` and `client`:

```
> pwd
parabol/packages/server
> yarn lint
yarn run v1.22.18
$ eslint --fix "**/*.{ts,tsx}"

/Users/blimmer/code/parabol/parabol/packages/server/graphql/types/UserTiersCount.ts
  0:0  error  Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: graphql/types/UserTiersCount.ts.
The file must be included in at least one of the projects provided

✖ 1 problem (1 error, 0 warnings)

error Command failed with exit code 1.
```

```
> pwd
parabol/packages/client
> yarn lint
yarn run v1.22.18
$ eslint --fix "**/*.{ts,tsx}"

/Users/blimmer/code/parabol/parabol-2/packages/client/components/AtmosphereProvider/PortalProvider.tsx
  6:57  error  Unexpected empty arrow function  @typescript-eslint/no-empty-function

/Users/blimmer/code/parabol/parabol-2/packages/client/components/Dashboard/DashNavTeam.tsx
  0:0  error  Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser.
The file does not match your project config: components/Dashboard/DashNavTeam.tsx.
The file must be included in at least one of the projects provided

/Users/blimmer/code/parabol/parabol-2/packages/client/components/FacilitatorMenu.tsx
  22:77  error  Expected '!==' and instead saw '!='  eqeqeq

/Users/blimmer/code/parabol/parabol-2/packages/client/components/JiraFieldMenu.tsx
  29:28  warning  React Hook "useMemo" is called conditionally. React Hooks must be called in the exact same order in every component render  react-hooks/rules-of-hooks

/Users/blimmer/code/parabol/parabol-2/packages/client/components/MeetingCard.tsx
  172:27  warning  React Hook "useBreakpoint" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return?           react-hooks/rules-of-hooks
  173:60  warning  React Hook "useMenu" is called conditionally. React Hooks must be called in the exact same order in every component render. Did you accidentally call a React Hook after an early return?                 react-hooks/rules-of-hooks
[...]

PR #6414, opened 2022-04-18
build: use mainline workbox for sw.js generation

blimmer:
## Background

I saw that we were getting deprecation warnings when building:

> (node:19119) [DEP_WEBPACK_COMPILATION_CACHE] DeprecationWarning: Compilation.cache was removed in favor of Compilation.getCache()
(Use `node --trace-deprecation ...` to show where the warning was created)

After digging into it via the `--trace-deprecation` flag, I tracked it down to the `wrkbx` plugin:

```
(node:19165) [DEP_WEBPACK_COMPILATION_CACHE] DeprecationWarning: Compilation.cache was removed in favor of Compilation.getCache()
    at /Users/blimmer/code/parabol/parabol/node_modules/wrkbx/build/inject-manifest.js:94:24
    at _next36 (eval at create (/Users/blimmer/code/parabol/parabol/node_modules/tapable/lib/HookCodeFactory.js:19:10), <anonymous>:50:1)
```

Interestingly, it seems like `wrkbx` was an unofficial, temporary fork of `workbox` which added Webpack compilation to `workbox` (see [the original comment](https://github.com/GoogleChrome/workbox/issues/1513#issuecomment-496024997)).

This behavior is now officially supported in `workbox`, so we can now get back to the mainline version. I also [asked the original author](https://github.com/GoogleChrome/workbox/issues/1513#issuecomment-1101578912) to deprecate `wrkbox` so it's more discoverable that people should migrate to the original version.

## sw.js diffs

I diffed the before and after of the `sw.js` file produced to confirm this change. Here's that diff:

```diff
1,2d0
< importScripts("key_CDN_BASE_URL/build/v6.54.0/precache-manifest.c3bb2fee8723020cc02ce220b1caf93d.js");
< 
20c18
<   const urls = self.__precacheManifest.map(({
---
[...]

PR #6416, opened 2022-04-18
build: reduce docker image size

### LINKED ISSUE #4827: Reduce Docker image size by using a multi-stage build
## Issue - Enhancement

The current Docker image produced by `docker/Dockerfile.prod` is quite large. As of today, the resulting image was 3.59gb in size. By reducing the image size, users will be able to scale out more quickly and ultimately reduce their resource requirements to save money on cloud deployments.

I did a little experiment this afternoon to see how much space `devDependencies` were taking up and it looks like removing `devDependencies` alone gets the image down to 2.41gb (an ~33% decrease in size). For this hacky test, I applied this diff to `docker/Dockerfile.prod`. This isn't a good solution long-term but is good for demonstrating some of the opportunities we have by optimizing the Dockerfile from its current state.

```diff
diff --git a/docker/Dockerfile.prod b/docker/Dockerfile.prod
index a76e99856..b8d50a98a 100644
--- a/docker/Dockerfile.prod
+++ b/docker/Dockerfile.prod
@@ -10,8 +10,10 @@ ADD yarn.lock .
 COPY . .

 FROM base as prod
-RUN yarn install
-RUN yarn build
+RUN yarn install --cache-folder /tmp/.yarn-cache && \
+    yarn build && \
+    rm -fr node_modules /tmp/.yarn-cache
+RUN yarn install --cache-folder /tmp/.yarn-cache --prod && rm -fr /tmp/.yarn-cache
 RUN cp docker/entrypoint.prod.sh /bin/entrypoint
 RUN chmod +x /bin/entrypoint
 EXPOSE 80
```

It looks like there are a few `devDependencies` that're actually required at runtime (I was getting some errors when starting up the server), so we'll need to look into those.

After taking a look at the Dockerfile, there are a few opportunities I saw that could reduce the final size of the image.

### Implement a multi-stage build

Today, you do have two stages in your Dockerfile (`base` and `prod`), but the `yarn install` and `yarn build` is happening in your `prod` layer, instead of the `base` layer. To more effectively utilize a multi-stage build, we could hoist the installation of `devDependencies` and the `yarn build` up to the `base` layer. Then, in the `prod` layer, install only prod NPM dependencies and copy the built code from the `base` layer.

Here's a more concrete example of what this would look like:

```dockerfile
FROM node:14.10.0 as base
COPY . .
RUN yarn install --cache-folder /tmp/.yarn-cache && \
	yarn build && \
	rm -fr /tmp/.yarn-cache

FROM node:14.10.0
WORKDIR /parabol/

# Copy all files from the base image. This is all the `js` files transpiled from TS in the `yarn build` step.
# And any other supporting files needed (e.g., `package.json` & `yarn.lock`)
COPY --from=base <YOUR BUILT FILES> ./

# Only install prod dependencies
RUN yarn install --prod --cache-folder /tmp/.yarn-cache && \
	rm -fr /tmp/.yarn-cache

RUN cp docker/entrypoint.prod.sh /bin/entrypoint
RUN chmod +x /bin/entrypoint
EXPOSE 80
ENTRYPOINT [ "entrypoint" ]
CMD ["yarn", "start"]
```

There are more information and examples on this approach [here](https://github.com/goldbergyoni/nodebestpractices/blob/master/sections/docker/multi_stage_builds.md).

### Consider another node image variant for your `prod` stage

Another benefit of running your build in a separate stage is that you may be able to use `alpine`, or another smaller [variant](https://github.com/nodejs/docker-node#image-variants) of the `node` image to further reduce the size of the image.

### Review `.dockerignore`

You'll get the most bang for your buck from the step above, but you can also reduce the final image size by reviewing your `.dockerignore` file. It looks like some files intended for humans (like `SECURITY.md`) and dev (`devTemplate.html`) are included in the final production image. Ignoring the `docs` folder will free up 19mb alone.

The resulting docker image is:

- Significantly smaller than the current size (3.59gb)
- The application still runs properly

- **Estimated effort:** 11 points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

blimmer:
This PR reduces the size of Docker image by only installing production dependencies, not `devDependencies`.

See inline comments for details.

Closes #4827


.dockerignore
```diff
@@ -6,4 +6,7 @@
 /.git
 /.github
 /.vscode
-/packages/client/__generated__
\ No newline at end of file
+/packages/client/__generated__
+/.husky
+/.prettierrc
+/docs
```
blimmer, line 12: These files don't need to be included in the docker image.

docker/Dockerfile.prod
```diff
@@ -1,19 +1,42 @@
-FROM node:16.14.0 as base
+# First, install all dependencies, including devDependencies, to run the build process
+FROM node:16.14.0 as build
```
blimmer, line 2: This pattern is called a [multi-stage build](https://docs.docker.com/develop/develop-images/multistage-build/). At a high level, we have two distinct phases: `build` and `prod`.

The first stage, `build`, is defined here first. We install all dependencies, including `devDependencies` so `webpack` will work properly.

Then in the second stage, `prod`, we copy over the generated files, but only install the production dependencies to reduce the final image size.

This pattern reduced the image size from 2.51 GB to 1.43 GB (a 43% decrease!).

docker/Dockerfile.prod
```diff
@@ -1,19 +1,42 @@
...
 
-ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
+WORKDIR /parabol
 
-RUN mkdir -p /parabol/node_modules && mkdir -p /home/node/.npm-global
+# Only copy dependency-related files here (vs. COPY . .) to avoid breaking the cache and running
+# the slow `yarn install` more than necessary
```
blimmer, line 7: This might seem overly verbose, but having the docker layer caching here is beneficial. By using this pattern, the only time `yarn install` needs to run is:

- if the layer is not cached (e.g., you've never built before)
- if any of the `package.json` files or `yarn.lock` changes

Otherwise, docker knows `yarn install` will produce the exact same resulting layer so it'll use the cache.

docker/Dockerfile.prod
```diff
@@ -1,19 +1,42 @@
...
+
+WORKDIR /parabol
+
+COPY package.json yarn.lock ./
+COPY packages/client/package.json ./packages/client/package.json
+COPY packages/gql-executor/package.json packages/gql-executor/package.json
+COPY packages/integration-tests/package.json packages/integration-tests/package.json
+COPY packages/server/package.json packages/server/package.json
+# Only install production dependencies
+RUN yarn install --prod --frozen-lockfile && \
```
blimmer, line 30: `--prod` is the important bit here.

docker/Dockerfile.prod
```diff
@@ -1,19 +1,42 @@
...
-RUN mkdir -p /parabol/node_modules && mkdir -p /home/node/.npm-global
+# Only copy dependency-related files here (vs. COPY . .) to avoid breaking the cache and running
+# the slow `yarn install` more than necessary
+COPY package.json yarn.lock ./
+COPY packages/client/package.json ./packages/client/package.json
+COPY packages/gql-executor/package.json packages/gql-executor/package.json
+COPY packages/integration-tests/package.json packages/integration-tests/package.json
+COPY packages/server/package.json packages/server/package.json
+RUN yarn install --frozen-lockfile && \
+      yarn cache clean
```
blimmer, line 14: `yarn cache clean` reduces the overall size of the layer.

docker/Dockerfile.prod
```diff
@@ -1,19 +1,42 @@
...
+COPY packages/integration-tests/package.json packages/integration-tests/package.json
+COPY packages/server/package.json packages/server/package.json
+# Only install production dependencies
+RUN yarn install --prod --frozen-lockfile && \
+      yarn cache clean
+
+COPY . .
+COPY --from=build /parabol/build ./build
+COPY --from=build /parabol/dist ./dist
+COPY --from=build /parabol/queryMap.json ./
```
blimmer, line 36: This `--from=build` is what copies the built files from the `build` stage to this stage. No `devDependencies` are copied over!

packages/server/tsconfig.eslint.json
```diff
@@ -8,6 +8,7 @@
     "postgres/migrations/postDeploy/*",
     "postgres/types/*",
     "graphql/queries",
+    "graphql/types/**/*.ts",
```
blimmer, line 11: See #6415 for a more comprehensive fix

packages/server/package.json
```diff
@@ -123,8 +114,8 @@
     "graphql-shield": "^7.5.0",
     "html-minifier-terser": "^5.0.2",
     "immutable": "3.8.2",
-    "intersection-observer": "^0.7.0",
     "ioredis": "^4.11.2",
+    "jsdom": "^16.6.0",
```
blimmer, line 118: This was the only dependency I found that needed to be moved from `devDependencies` to `dependencies`. It is used at runtime to sanitize SVGs: https://github.com/ParabolInc/parabol/blob/master/packages/server/fileStorage/validateAvatarUpload.ts

packages/client/package.json
```diff
@@ -38,13 +38,14 @@
     "@types/relay-runtime": "^12.0.0",
     "@types/segment-analytics": "^0.0.31",
     "@types/stripe-v2": "^2.0.1",
+    "@types/linkify-it": "^3.0.2",
     "babel-plugin-relay": "^12.0.0",
+    "debug": "^4.1.1",
```
blimmer, line 43: It didn't feel like `debug` needed to be included in the production bundle.

packages/client/package.json
```diff
@@ -38,13 +38,14 @@
...
     "eslint": "^8.2.0",
     "eslint-config-prettier": "^8.5.0",
     "eslint-plugin-emotion": "^10.0.14",
     "eslint-plugin-react": "^7.16.0",
     "eslint-plugin-react-hooks": "^1.6.1",
-    "expect": "^24.8.0",
```
blimmer, line 47: `jest` provides `expect`-style syntax as a dependency

packages/client/package.json
```diff
@@ -79,21 +79,18 @@
     "graphql-typed": "^0.6.1",
     "hoist-non-react-statics": "^3.3.0",
     "immutable": "3.8.2",
-    "intersection-observer": "^0.7.0",
```
blimmer, line 82: I don't think we need to polyfill this anymore. It's pretty well-supported: https://caniuse.com/intersectionobserver

packages/client/package.json
```diff
@@ -79,21 +79,18 @@
...
     "json2csv": "5.0.5",
     "jwt-decode": "^2.1.0",
     "linkify-it": "^2.0.3",
     "mousetrap": "^1.6.3",
     "ms": "^2.0.0",
-    "oy-vey": "^0.11.0",
```
blimmer, line 88: I moved this to `server/package.json`, since that's the only place it's used.

packages/client/package.json
```diff
@@ -79,21 +79,18 @@
...
     "prop-types": "^15.5.10",
     "react": "^17.0.2",
     "react-beautiful-dnd": "13.0.0",
     "react-copy-to-clipboard": "^5.0.0",
     "react-day-picker": "^7.1.6",
     "react-dom": "^17.0.2",
     "react-dom-confetti": "^0.0.10",
-    "react-hotkey-hoc": "^0.2.1",
```
blimmer, line 96: I couldn't find any uses of this in code.

packages/server/package.json
```diff
@@ -36,29 +36,22 @@
     "@types/jest": "^27.0.1",
     "@types/json2csv": "^5.0.3",
     "@types/jsonwebtoken": "^8.3.0",
-    "@types/linkify-it": "^3.0.2",
```
blimmer, line 39: Only used in `client` - moved.

packages/server/package.json
```diff
@@ -36,29 +36,22 @@
...
     "@types/mime-types": "^2.1.0",
     "@types/ms": "^0.7.30",
     "@types/node": "16",
     "@types/nodemailer": "^6.4.4",
-    "@types/react-relay": "^11.0.2",
```
blimmer, line 44: `react-relay` isn't used in `server`, just `relay-runtime`.

packages/server/package.json
```diff
@@ -36,29 +36,22 @@
...
     "@types/relay-runtime": "^11.0.4",
     "@types/sharp": "^0.29.4",
     "@types/string-similarity": "^3.0.0",
     "@types/stripe": "^6.32.0",
     "@types/uuid": "^8.3.3",
+    "@pgtyped/cli": "^0.11.0",
```
blimmer, line 48: I moved this to `devDependencies`. It didn't seem like something that needed to be in the prod deps.

packages/server/package.json
```diff
@@ -36,29 +36,22 @@
...
     "babel-plugin-relay": "^12.0.0",
     "clean-webpack-plugin": "^3.0.0",
     "copy-webpack-plugin": "7.0.0",
-    "cross-env": "^5.1.6",
     "css-loader": "5.0.1",
-    "dynamic-serializer": "^0.2.1",
```
blimmer, line 55: I couldn't find uses of either of these packages.

packages/server/package.json
```diff
@@ -36,29 +36,22 @@
...
     "eslint": "^8.2.0",
     "eslint-config-prettier": "^8.5.0",
-    "eslint-plugin-emotion": "^10.0.14",
-    "eslint-plugin-react": "^7.16.0",
-    "eslint-plugin-react-hooks": "^1.6.1",
```
blimmer, line 60: These plugins are only used in `client`, where they're already defined.

packages/server/package.json
```diff
@@ -68,11 +61,10 @@
     "jest-extended": "^1.2.0",
     "jest-junit": "^13.0.0",
     "jest-transform-graphql": "^2.1.0",
-    "jsdom": "^16.6.0",
     "json-loader": "^0.5.7",
     "lint-staged": "^10.1.7",
-    "mocha": "^9.0.3",
```
blimmer, line 74: We're using `jest`.

packages/server/package.json
```diff
@@ -68,11 +61,10 @@
...
     "prettier": "^2.6.1",
+    "prettier-plugin-organize-imports": "^2.3.4",
```
blimmer, line 67: moved to `devDependencies`.

packages/server/package.json
```diff
@@ -136,17 +127,15 @@
...
     "pg": "^8.5.1",
     "pm2": "^5.1.1",
-    "prettier-plugin-organize-imports": "^2.3.4",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
-    "react-relay": "^12.0.0",
     "relay-runtime": "^12.0.0",
     "rethinkdb-ts": "2.4.5",
     "samlify": "^2.8.2",
-    "secure-compare": "^3.0.1",
```
blimmer, line 149: I couldn't find uses of this

mattkrick:
merged https://github.com/ParabolInc/parabol/pull/6414 and master into this, should be ready to merge after tests pass. excited to see this help out the build times!

PR #6427, opened 2022-04-20
feat(sprint-poker): Push task to GitLab

### LINKED ISSUE #6365: Sprint Poker: GitLab v2: Implement TaskIntegrationManager in GitLabServerManager
To be consistent with other integrations, the `GitLabServerManager` should implement the new `TaskIntegrationManager`.

See comment for more details: https://github.com/ParabolInc/parabol/pull/6267#discussion_r844481608

### AC

- [ ] `GitLabServerManager` implements the `TaskIntegrationManager`

### Updated AC:

- [ ] A user can integrate a Parabol task card with a GitLab project

nickoferrall:
Fix #6365 and fix https://github.com/ParabolInc/parabol/issues/6428

This PR refactors the GitLab queries and mutations to use the `TaskIntegrationManagerFactory` pattern that Igor created. Using this class, I've also implemented pushing a Parabol task card to a GitLab project. 

push-task-to-gitlab


packages/server/graphql/queries/helpers/fetchGitLabProjects.ts
```diff
@@ -16,12 +14,11 @@ const fetchGitLabProjects = async (
     .get('teamMemberIntegrationAuths')
     .load({service: 'gitlab', teamId, userId})
   if (!auth?.accessToken) return []
-  const {accessToken, providerId} = auth
+  const {providerId} = auth
   const provider = await dataLoader.get('integrationProviders').load(providerId)
   if (!provider?.serverBaseUrl) return []
```
nickoferrall, line 19: Sometimes we've just checked whether `provider` exists and then written `provider.serverBaseUrl!` below. Personally, I prefer to include `serverBaseUrl` in the if statement as it doesn't increase the number of lines of code and it means we don't need to use a `!`

packages/client/components/TaskFooterIntegrateMenu.tsx
```diff
@@ -131,6 +137,8 @@ const TaskFooterIntegrateMenu = (props: Props) => {
       mutationProps={mutationProps}
       teamId={teamId}
       label={label}
+      gitlabRef={viewerIntegrations.gitlab}
```
nickoferrall, line 140: I don't love that this is so GitLab specific, but we do need the GitLab cloud provider info to be able to complete the auth [here](https://github.com/ParabolInc/parabol/blob/e11747dfe132959d3ff09221debdb2398e2b4c66/packages/client/components/AddToGitLabMenuItem.tsx#L53). 

I considered passing down `integrationsRef` instead, but as we're just using GitLab here, I felt it was best to pass down the smallest amount of data possible.

igorlesnenko: Maybe also consider passing `cloudProvider` which should be generic across different integrations

packages/server/integrations/TaskIntegrationManagerFactory.ts
```diff
@@ -54,6 +61,17 @@ export default class TaskIntegrationManagerFactory {
...
 
+    if (service === 'gitlab') {
+      const auth = await dataLoader
+        .get('teamMemberIntegrationAuths')
+        .load({service: 'gitlab', teamId, userId})
+      if (!auth) return null
+      const {providerId} = auth
+      const provider = await dataLoader.get('integrationProviders').load(providerId)
+      if (!provider?.serverBaseUrl) return null
+      return auth && new GitLabServerManager(auth, context, info, provider.serverBaseUrl)
```
igorlesnenko, line 72: +1 Looks like we already checked for `auth` above

packages/client/components/TaskIntegrationLink.tsx
```diff
@@ -77,6 +78,23 @@ const TaskIntegrationLink = (props: Props) => {
         {children}
       </StyledLink>
     )
+  } else if (integration.__typename === '_xGitLabIssue') {
+    const {webPath, iid, webUrl} = integration
+    const {fullPath} = parseWebPath(webPath)
+    const href =
+      fullPath === 'ParabolInc/ParabolDemo' ? 'https://github.com/ParabolInc/parabol' : webUrl
```
igorlesnenko, line 85: +1  Still `github` here

packages/client/components/TaskFooterIntegrateMenu.tsx
nickoferrall: I like it and agree that it's more generic, but `cloudProvider` can be null so `cloudProviderRef` won't work as the ref in the child fragment

packages/server/graphql/queries/helpers/fetchGitLabProjects.ts
Dschoordsch: You can also do `const provider = await dataLoader.get('integrationProviders').loadNonNull(providerId)` since you know it must exist if the `auth` was valid. I think the `teamMemberIntegrationAuths` loader should return the provider as well, since we always need it.

nickoferrall: `loadNonNull` guarantees that the `provider` is non-null, but the `serverBaseUrl` can still be nullable. As we have to check whether `serverBaseUrl` is null, I think there's no harm is using `if (!provider?.serverBaseUrl) return []`

packages/client/components/TaskFooterIntegrateMenu.tsx
```diff
@@ -18,25 +18,36 @@ interface Props {
...
 }
 
 type Integrations = NonNullable<TaskFooterIntegrateMenu_viewer['viewerTeamMember']>['integrations']
 
 const isIntegrated = (integrations: Integrations) => {
-  const {atlassian, github, jiraServer} = integrations
+  const {atlassian, github, jiraServer, gitlab} = integrations
   const hasAtlassian = atlassian?.isActive ?? false
   const hasGitHub = github?.isActive ?? false
+  const hasGitLab = gitlab?.auth?.isActive ?? false
```
nickoferrall, line 43: Check if `gitlab` exists as it doesn't yet in the demo. It'll be implemented in this issue: https://github.com/ParabolInc/parabol/issues/6443

nickoferrall:
@mattkrick or @Dschoordsch it'd be great to get a Maintainer Review on this one when you get the chance. Once this PR and https://github.com/ParabolInc/parabol/pull/6398 are merged, we can launch v1 of the GitLab integration. 

I mentioned to @lorena-cure that I estimate that v1 will be ready in prod by the next ship on Wednesday.

Dschoordsch (changes requested):
Works fine. Only one small change in line with your own concern about `gitlabRef`

packages/client/components/TaskFooterIntegrateMenuSignup.tsx
```diff
@@ -27,8 +34,16 @@ const Label = styled('div')({
 })
 
 const TaskFooterIntegrateMenuSignup = (props: Props) => {
-  const {menuProps, mutationProps, teamId, label} = props
+  const {menuProps, mutationProps, teamId, label, gitlabRef, featureFlags} = props
   const {submitting} = mutationProps
+  const gitlab = useFragment(
+    graphql`
+      fragment TaskFooterIntegrateMenuSignup_GitLabIntegration on GitLabIntegration {
```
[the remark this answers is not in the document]
nickoferrall: I like it! I've pushed an update swapping `gitlabRef` for `integrationsRef`

Dschoordsch (approved):
Works for the owner of the integration. Doesn't work for non-integrated users:
* Gerrit has GitLab integration and creates a task "Push me"
* Tina is a teammate of Gerrit and tries to push the task "Push me" on behalf of Gerrit
For Jira/Git this will push the task using Gerrit's credentials and add a comment that it was pushed by Tina on his behalf. With this PR it shows a server error
```
  error: '{"data":{"createTaskIntegration":null},"errors":[{"message":"Method not implemented yet.","locations":[{"line":6,"column":3}],"path":["createTaskIntegration"]}]}'
```
I'm fine moving this into a separate issue/PR.

nickoferrall:
Nice catch! I've implemented the `addCreatedBySomeoneElseComment` method for GitLab so that should be working now. 

I'll move this issue into the Stuck column until the base branch is merged

PR #6439, opened 2022-04-25
fix(sprint-poker): Select All issues

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6213


packages/server/graphql/mutations/updatePokerScope.ts
```diff
@@ -166,10 +166,15 @@ const updatePokerScope = {
       // MUTATIVE
       newDiscussions.push(...discussions)
       stages.push(...newStages)
-      newStageIds = newStages.map(({id}) => id)
```
nickoferrall, line 169: We were looping through all the updates in `additiveUpdatesWithTaskIds` and then using `dimensions` to get the `newStages`. However, we would then overwrite `newStageIds` instead of appending them. 

This meant that only one of two (depending on the number of dimensions) `newStageIds` would be returned from the mutation.

packages/client/mutations/UpdatePokerScopeMutation.ts
```diff
@@ -113,17 +113,6 @@ const UpdatePokerScopeMutation: StandardMutation<TUpdatePokerScopeMutation, Hand
...
     variables,
-    updater: (store) => {
-      const payload = store.getRootField('updatePokerScope')
-      const meeting = payload.getLinkedRecord('meeting')
-      const newStages = payload.getLinkedRecords('newStages')
-      if (!meeting || !newStages) return
-      const phases = meeting.getLinkedRecords('phases')
-      const estimatePhase = phases.find((phase) => phase.getType() === 'EstimatePhase')!
-      const stages = estimatePhase.getLinkedRecords('stages')
-      const nextStages = [...stages, ...newStages]
```
nickoferrall, line 124: When selecting all issues, `stages` will include many of the same issues as `newStages` so it will duplicate the issues. 

We could make sure that `nextStages` only contains unique stages, but it looks like the `updater` here isn't very helpful so I'd prefer to remove it. I've tested updating the poker scope with and without the updater with two users by selecting & deselecting all as well as selecting & deselecting individual issues. The behaviour is identical either way.

packages/client/types/constEnums.ts
```diff
@@ -366,6 +366,7 @@ export const enum Threshold {
   MAX_POKER_DIMENSION_NAME = 50,
   MAX_REACTJIS = 12,
   MAX_POKER_STORIES = 50,
+  MAX_GITLAB_POKER_STORIES = 20,
```
nickoferrall, line 369: This results in ugly logic elsewhere, but I get an error if I select/deselect 50 GitLab issues: `Query too large`. This doesn't happen with the GitHub or Jira integration. The reason is the [GitLab query](https://github.com/ParabolInc/parabol/blob/5235acc69ab0094009a15832ad9f5344a11024b1/packages/client/components/GitLabScopingSearchResults.tsx) is more complex as it needs to grab projects and then issues within those projects rather than just querying the issue. 

The complex query will be fixed in this issue: https://github.com/ParabolInc/parabol/issues/6325 I've updated the AC to include removing this.

packages/client/hooks/useUnusedRecords.ts
```diff
@@ -3,19 +3,21 @@ import {Threshold} from '../types/constEnums'
...
-  usedServiceTaskIds: Set<string>
+  usedServiceTaskIds: Set<string>,
+  thresh?: number
 ): [string[], boolean | null] => {
   return useMemo(() => {
     const unusedServiceTaskIds = [] as string[]
     serviceTaskIds.forEach((servieTaskId) => {
       if (!usedServiceTaskIds.has(servieTaskId)) unusedServiceTaskIds.push(servieTaskId)
     })
+    const maxStories = thresh ? thresh : Threshold.MAX_POKER_STORIES
```
BartoszJarocki, line 14: +1 you could make the `tresh` param mandatory and provide `Threshold.MAX_POKER_STORIES` as default, then this line could be removed

packages/client/mutations/UpdatePokerScopeMutation.ts
BartoszJarocki: i tested and it works. out of curiosity, do we where the relay store is updated so it actually shows selected/deselected issues without this updater? i suspect it'd be a meeting subscription. i just want to make sure we don't rely on some accidental side effect

packages/client/types/constEnums.ts
BartoszJarocki: +2 i appreciate the context!

BartoszJarocki (approved):
lgtm!

packages/client/mutations/UpdatePokerScopeMutation.ts
nickoferrall: I'm not sure. It might be because the local data update isn't very complex, but I'd like to get the Maintainer's input on this

packages/server/graphql/types/EstimateStage.ts
```diff
@@ -130,7 +132,13 @@ const EstimateStage = new GraphQLObjectType<Source, GQLContext>({
             info,
             provider.serverBaseUrl!
           )
-          const [issueData] = await manager.getIssue({gid})
+          const [issueData, issueError] = await manager.getIssue({gid})
+          if (issueError) {
```
nickoferrall, line 136: This fixes the error mentioned in https://github.com/ParabolInc/parabol/issues/6511 where a user can't join a meeting. 

If there are many issues selected, GitLab will stop returning `issueData` and return `issueError` saying that the query complexity has exceeded the limit of 250. By returning a null field here, we prevent the app from crashing.

However, if you select all issues in GitLab and refresh the page, GitLab can't retrieve all the issues and shows the dummy error text in the sidebar. I'm working on fixing this in https://github.com/ParabolInc/parabol/pull/6520

packages/server/graphql/types/EstimateStage.ts
```diff
@@ -130,7 +132,13 @@ const EstimateStage = new GraphQLObjectType<Source, GQLContext>({
...
+            console.log(issueError)
```
Dschoordsch, line 137: +1 please remove

packages/client/mutations/UpdatePokerScopeMutation.ts
Dschoordsch: Would love to look into the actual implementation, but basically because the objects in "stages" and "newStages" have the same id they are treated as identical, i.e. merged in the store.

Dschoordsch (approved):
Tested, works for the most part, one small issue when we selected already 50 issues with a different integration and then switch to the GitLab tab it has -35 issues remaining:
Screenshot from 2022-05-11 12-42-22
Since this is an edge case and it works better than before, I'm fine with merging as is and fixing in a follow up

nickoferrall:
Thanks for the review! 

> when we selected already 50 issues with a different integration and then switch to the GitLab tab it has -35 issues remaining

This issue will be fixed in this PR: https://github.com/ParabolInc/parabol/pull/6541

Issue #6440, opened 2022-04-25
viewerMeetingMember can be undefined


nickoferrall:
https://sentry.io/organizations/parabol/issues/3209960940/?referrer=github_plugin

Affected 153 users in 5 days.

To reproduce:

- Alice creates a new team and start a meeting with that team
- Bob is not a member of the team
- Bob clicks on the meeting link, Alice accepts it, Bob see's the error
- Bob can refresh the page and the error goes away

```
TypeError: Cannot destructure property 'user' of 'i' as it is null.
  at c (./packages/client/components/ScopePhaseArea.tsx:73:10)
  at Set.forEach (<anonymous>)
  at setConnectedStatus (./packages/client/hooks/useTrebuchetEvents.ts:20:7)
  at call (./packages/client/hooks/useTrebuchetEvents.ts:94:7)
  at y.promiseToUpgrade (./packages/client/Atmosphere.ts:251:23)
...
(25 additional frame(s) were not displayed)
```

PR #6441, opened 2022-04-25
fix: viewerMeetingMember can be undefined

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6440

In a previous PR, I used the non-null assertion to fix the eslint complaint that we could return null before a hook. Matt mentioned that we could ignore the eslint complaint and return null if `viewerMeetingMember` is undefined. 

Unfortunately, while I made the change in that PR, a subsequent PR built off of that branch must have still included the change. This PR reverts that change. 

I'd like to get this PR merged into master before the ship on Wednesday. As the change is trivial and the urgency is high, I'll request a review from a Maintainer.

https://www.loom.com/share/3fdd2fbb01ef42a9a7691f9c1132a276

### DIFF
--- a/packages/client/components/ScopePhaseArea.tsx
+++ b/packages/client/components/ScopePhaseArea.tsx
@@ -70,16 +70,14 @@ const ScopePhaseArea = (props: Props) => {
   const {meeting} = props
   const isDesktop = useBreakpoint(Breakpoint.SIDEBAR_LEFT)
   const {viewerMeetingMember} = meeting
-  if (!viewerMeetingMember) return null
-  const {user, teamMember} = viewerMeetingMember
-  const {featureFlags} = user
-  const gitlabIntegration = teamMember.integrations.gitlab
-  const jiraServerIntegration = teamMember.integrations.jiraServer
+  const featureFlags = viewerMeetingMember?.user.featureFlags
+  const gitlabIntegration = viewerMeetingMember?.teamMember.integrations.gitlab
+  const jiraServerIntegration = viewerMeetingMember?.teamMember.integrations.jiraServer
   const isGitLabProviderAvailable = !!(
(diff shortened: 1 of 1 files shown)


packages/client/components/ScopePhaseArea.tsx
```diff
@@ -70,7 +70,8 @@ const ScopePhaseArea = (props: Props) => {
   const {meeting} = props
   const isDesktop = useBreakpoint(Breakpoint.SIDEBAR_LEFT)
   const {viewerMeetingMember} = meeting
-  const {user, teamMember} = viewerMeetingMember!
+  if (!viewerMeetingMember) return null
```
Dschoordsch, line 73: -1 While I don't have a big issue with breaking the hook rules here, I think it would be better to address the underlying issue. From your issue description it appears as if we should add [`ScopePhaseArea_meeting`](https://github.com/ParabolInc/parabol/blob/e2117f4ba74a893e00925d77275f3dd6c8e19717/packages/client/components/ScopePhaseArea.tsx#L176) to the [`AcceptTeamInvitationPayload`](https://github.com/ParabolInc/parabol/blob/e2117f4ba74a893e00925d77275f3dd6c8e19717/packages/client/mutations/AcceptTeamInvitationMutation.ts#L39) instead

Dschoordsch (changes requested):
Please check first if we can't fix this properly

packages/client/components/ScopePhaseArea.tsx
nickoferrall: I tried adding the `ScopePhaseArea_meeting` fragment to the payload and I also tried adding `viewerMeetingMember` to the payload, but `viewerMeetingMember` could still be null in `ScopePhaseArea`. 

I've pushed an update that uses conditionals to grab the values from `viewerMeetingMember`, so it no longer returns before the hook.

PR #6447, opened 2022-04-25
feat(enableSAMLforDomain): support updates

### LINKED ISSUE #6327: feat(sdl): fix args implicit any in sdl resolvers

### LINKED ISSUE #6409: GraphQL schema is broken in /admin/grapqhl
## Issue - Bug

Screen Shot 2022-04-15 at 3 51 44 PM
Screen Shot 2022-04-15 at 3 54 56 PM

This causes running GraphQL queries really hard in `/admin/graphql` with no auto-suggestions etc. If you manage to get the queries correct they run fine though.

Users can:
   - [ ] Run GraphQL queries in `admin/graphql` with auto-suggestions

mattkrick:
fix #6327
fix #6409

enableSAMLForDomain had some bugs in it:
- you could register a domain for multiple different clients
- you couldn't just update the domains without also including the metadata, or vice versa

Fixing this also exposed a bug in typescript not employing strictNullChecks, so that's fixed here, too.

TEST
- [ ] add a new SAML record locally & it works (you can copy valid metadata from staging)
- [ ] edit that record's domains by calling it again without metadata (e.g. `{name: 'foo.co', domains: ['example.co']}`)
- [ ] create a new SAML record & try to usurp that domain (e.g. `{name: 'test.co', domains: ['example.co']}`). it should fail because example.co is already owned by another SAML record

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,7 +48,7 @@
     "prepare": "husky install"
   },
   "resolutions": {
-    "typescript": "4.5.5",
+    "typescript": "4.6.3",
     "hoist-non-react-statics": "^3.3.0",
     "@types/react": "16.9.11",
     "@types/react-dom": "16.9.4",
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -50,7 +50,7 @@
     "prettier": "^2.6.1",
     "react-refresh": "^0.9.0",
     "strict-event-emitter-types": "^2.0.0",
-    "typescript": "4.5.5",
+    "typescript": "4.6.3",
     "webpack": "5.64.1",
     "webpack-dev-server": "^4.7.3"
   },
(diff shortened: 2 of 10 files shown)


mattkrick:
@BartoszJarocki would you mind taking a look at this when you get a chance?

BartoszJarocki (approved):
- [x] add a new SAML record locally & it works (you can copy valid metadata from staging)
- [x] edit that record's domains by calling it again without metadata (e.g. `{name: 'foo.co', domains: ['example.co']}`)
- [x] create a new SAML record & try to usurp that domain (e.g. `{name: 'test.co', domains: ['example.co']}`). it should fail because example.co is already owned by another SAML record
fails nicely with
```
{
  "data": {
    "enableSAMLForDomain": {
      "__typename": "ErrorPayload",
      "error": {
        "message": "Domain is already owned by okta"
      }
    }
  }
}
```

everything works as expected 👍

Issue #6349, opened 2022-04-05
Demo: pushing to task error


jordanh:
## Issue - Bug

Repro:
1. Start demo
2. Advance to Discussion phase
3. Hover and press push to task icon
4. Observe error

See:

![2022-04-05 12 21 02](https://user-images.githubusercontent.com/103000/161833877-13537a2e-d89f-4712-93af-4f81f255209b.gif)

### Acceptance Criteria (optional)
Users can:
   - Push to task demo feature works, or
Triage is performed to find the root cause of the bug, timeboxed to ~1 hour. 

**Estimated effort:** 1 hour to triage. More if root cause is already identified.

jordanh:
Elevated to p1 because I hate giving folks poor impressions of product quality. Will set at the top of the backlog

Issue #5297, opened 2021-08-20
Relay: Refactor renderQuery to Suspense + hooks


mattkrick:
Relay v11 lets us leverage the new react suspense patterns.
Instead of the ugly `renderQuery` helper function that we used, we can write something that is typesafe, doesn't have funky loading states, and is just easier to reason about.

We can turn this:

```ts
  return (
    <QueryRenderer<TeamRootQuery>
      environment={atmosphere}
      query={query}
      variables={{teamId}}
      fetchPolicy={'store-or-network' as any}
      render={({props: renderProps}) => {
        const viewer = renderProps ? renderProps.viewer : null
        // @ts-ignore
        return <TeamContainer location={location} match={match} viewer={viewer} teamId={teamId} />
      }}
    />
  )
```

Into this:
```ts
  const queryRef = useQueryLoaderNow<TeamContainerQuery>(teamContainerQuery, {teamId})
  return (
    <Suspense fallback={''}>
      {queryRef && (
        <TeamContainer location={location} match={match} queryRef={queryRef} teamId={teamId} />
      )}
    </Suspense>
  )
```

For an example in code, see https://github.com/ParabolInc/parabol/commit/016f2761305132ec471941bab013aa0d899d0089

AC
- [x] `renderQuery` and `QueryRenderer` are removed in favor of relay hooks

Estimate: 8 hours

jordanh:
Given to @JimmyLv as an intro issue

JimmyLv:
related PR: #5294
another example commit: https://github.com/ParabolInc/parabol/commit/864956a9b34ce1504b40de8354199bc29e4e207e

PR #6460, opened 2022-04-27
fix: Integrate task with Jira & GitHub in the demo

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6349


packages/client/modules/demo/ClientGraphQLServer.ts
```diff
@@ -248,7 +248,7 @@ class ClientGraphQLServer extends (EventEmitter as GQLDemoEmitter) {
         }
       }
     },
-    TaskFooterIntegrateMenuRootQuery: (_teamId: unknown, userId: string) => {
+    TaskFooterIntegrateMenuQuery: (_teamId: unknown, userId: string) => {
```
nickoferrall, line 251: These queries were recently renamed. Trying to integrate a task card with an integration or ending a demo causes a crash

packages/client/components/TaskFooterIntegrateMenu.tsx
```diff
@@ -33,7 +33,7 @@ const isIntegrated = (integrations: Integrations) => {
   const {atlassian, github, jiraServer} = integrations
   const hasAtlassian = atlassian?.isActive ?? false
   const hasGitHub = github?.isActive ?? false
-  const hasJiraServer = jiraServer.auth?.isActive ?? false
+  const hasJiraServer = jiraServer?.auth?.isActive ?? false
```
nickoferrall, line 36: JiraServer hasn't been added to the demo so we need to check if it exists here. I've created an issue to add GitLab to the demo. It'd be great to create one for Jira Server too if there isn't one already

packages/client/modules/demo/ClientGraphQLServer.ts
JimmyLv: Just curious, `ops` here refers to simulated operations? The query result here is actually mocked, so it needs to be consistent with the query statement returned by the real GraphQL API. Is my understanding correct? 😁

JimmyLv (approved):
LGTM!

JimmyLv (changes requested):
Oh, if my [understanding](https://github.com/ParabolInc/parabol/pull/6460/files#r863846573) is correct, I think all of these mocked `RootQuery` need to be renamed, because recently I did the refactoring https://github.com/ParabolInc/parabol/issues/5297 but I didn't realize that I had to take care of these ops in the demo.

```ts
TaskFooterIntegrateMenuRootQuery
NewMeetingSummaryRootQuery
TaskFooterUserAssigneeMenuRootQuery
SuggestMentionableUsersRootQuery
```

nickoferrall:
Sure thing - I've updated two more queries 👍

mattkrick:
tested jira & github, seems to work!

Issue #6372, opened 2022-04-07
Standup UI - Integrate Response Cards with Discussion Threads 


jmtaber129:
Design - https://www.figma.com/file/bhiEOV4HK9Kv91HFsJmevV/Stand-up-Concepts

Blocked on https://github.com/ParabolInc/parabol/issues/6202 and https://github.com/ParabolInc/parabol/issues/5796

AC
- [ ] Show Response Card at top of Discussion Thread
- [ ] Emojis on added to Response Card appear on the card at the top of discussion thread, and vice versa
- [ ] Response Card shows number of replies
- [ ] Response Card shows commentor avatars
- [ ] Tapping response card opens discussion thread

PR #6469, opened 2022-04-27
feat(standups): Integrate response cards with discussion drawer

jmtaber129:
Fixes https://github.com/ParabolInc/parabol/issues/6372

https://www.loom.com/share/154a2c6d7da3438d945fbe9df01d065e


packages/server/dataloader/customLoaderMakers.ts
```diff
@@ -74,6 +74,7 @@ export const commentCountByDiscussionId = (parent: RootDataLoader) => {
         r
           .table('Comment')
           .getAll(r.args(discussionIds as string[]), {index: 'discussionId'})
+          .filter({isActive: true})
```
jmtaber129, line 77: Also leftover from when I was using commentCount as number of replies instead of the number of threadables.  Doesn't impact any existing features (since the meeting summary uses a comment count attached to the meeting object on meeting completion) but is still a latent bug where deleted comments would still be counted for the `Discussion` object

packages/client/components/TeamPrompt/TeamPromptDiscussionDrawer.tsx
```diff
@@ -104,30 +142,51 @@ const TeamPromptDiscussionDrawer = ({meetingRef, isDesktop}: Props) => {
     })
   }
 
-  return localDiscussionId ? (
+  // :TODO: (jmtaber129): Use content from 'stage.stage'.
```
BartoszJarocki, line 145: +1 what's the reason for not using the response from the actual gql query?

packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
```diff
@@ -44,24 +63,60 @@ const TeamPromptResponseCard = (props: Props) => {
...
-  const {teamMember} = responseStage
+  const {teamMember, meeting, discussion} = responseStage
   const {picture, preferredName, userId} = teamMember
 
+  const discussionEdges = discussion.thread.edges
+  const replyCount = discussionEdges.length
+
   const isCurrentViewer = userId === viewerId
-  const isEmptyResponse = !isCurrentViewer // :TODO: (jmtaber129): Determine based on actual response, too
+  const isEmptyResponse = false // :TODO: (jmtaber129): Determine based on actual response, too
```
BartoszJarocki, line 119: -1 even though the upsert mutation isn't yet wired up we can improve here to use the real response or am I missing something?

packages/server/graphql/mutations/joinMeeting.ts
```diff
@@ -141,6 +142,15 @@ const joinMeeting = {
       // only add a new stage for the new users (ie. invited to the team after the meeting was started)
       if (teamMemberResponseStage) return
       const responsesStage = new TeamPromptResponseStage({teamMemberId})
+      await insertDiscussions([
```
BartoszJarocki, line 145: +2 nice catch

BartoszJarocki (commented):
definitely, not a blocker but the drawer opening animation doesn't work when opening it for the firs
CleanShot 2022-05-06 at 08 33 41
t time, see gif below

other than that, a couple of questions about using the real responses, nice work!

packages/client/components/TeamPrompt/TeamPromptDiscussionDrawer.tsx
jmtaber129: The resolver for the stage's `response` is stubbed on the backend for this branch's base (returns an empty object), so felt better to leave this as a stub on the frontend, and either implement it in a follow-up PR, or fix this before merging if https://github.com/ParabolInc/parabol/pull/6388 lands first

packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
jmtaber129: Same issue as in the drawer - this isn't implemented on the backend in this branch's base

jmtaber129:
@BartoszJarocki re: the animation - I'll log this as a separate issue unless you know an easy way to do this, but I agree it shouldn't block this.  I think the fix would involve updating the state such that the component renders off-screen (as a result of `localStageId` being changed to non-null), then animating it onto the screen (as a result of `isRightDrawerOpen` being changed to `true`).

Re: comments around using the GQL results - would you prefer that this be based on https://github.com/ParabolInc/parabol/pull/6388 for the purposes of testing?  I originally based it on `master` since the two aren't directly coupled, but I don't have a strong preference if it helps with review.

packages/client/components/TeamPrompt/TeamPromptDiscussionDrawer.tsx
BartoszJarocki: makes sense, thank you!

BartoszJarocki (approved):

packages/client/components/TeamPrompt/TeamPromptRepliesAvatarList.tsx
```diff
@@ -0,0 +1,103 @@
...
+  width,
+  marginRight: '8px'
+}))
+
+interface Props {
+  edgesRef: TeamPromptRepliesAvatarList_edges$key
+}
+
+const SIZE = 24
+const OVERLAP = 4
```
Dschoordsch, line 23: +1 could be more specific: What size? Which overlap?

Dschoordsch (approved):
Did not do extensive testing.

@MindEvo could you please double check the AvatarList in Sprint Poker meetings voting face and on meeting cards in the overview?

MindEvo:
AvatarLists appear to be working properly however I'm having this issue with the filter button.

https://user-images.githubusercontent.com/35351786/168260641-67eb2772-27ba-4cd2-b370-595847f9ff8b.mp4

Dschoordsch:
> AvatarLists appear to be working properly however I'm having this issue with the filter button.

Thanks for testing! The issue you mentioned is fixed on master and should work after merging master back in, so no concerns there from my side.

packages/client/components/TeamPrompt/TeamPromptDiscussionDrawer.tsx
```diff
@@ -104,30 +146,39 @@ const TeamPromptDiscussionDrawer = ({meetingRef, isDesktop}: Props) => {
     })
   }
 
-  return localDiscussionId ? (
+  const response = stage.stage.response
```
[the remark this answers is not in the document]
jmtaber129: Done

packages/client/components/TeamPrompt/TeamPromptRepliesAvatarList.tsx
jmtaber129: Prefixed each with `AVATAR_`

PR #6476, opened 2022-04-28
feat(restrictDomains): Add table and mutations

### LINKED ISSUE #6095: Allow an organization to restrict signup to SSO / SAML on public SAAS
Relates to:
* #5512
* #4666
* #5663

From a support request:
As an organization (with enterprise tier and SAML) I want to restrict the email domains that allowed to be invited to
1) control license consumption
2) restrict access to the content

Restricting just the invite to certain domains would not solve their problem since it would still allow people to accept the invite with non-company addresses which they could still use for signup after leaving the company.

## Acceptance criteria
* [ ] add a field to Organization to enforceSAML
* [ ] if a new user accepts an invite to Org, they are required to signup with SAML if enforceSAML is set
* [ ] if an existing user tries to accept the invite and he's not a SAML user of that Org, they must create a new account

mattkrick:
~Stuck on: #6447~
fixes part of #6095

This allows users to create a whitelist of emails and/or domains that are allowed into the org.
To add a domain, it must be an enterprise organization & the user must be the billing leader
OR any superuser can make the change.

This PR doesn't include the logic necessary to prevent someone from accepting an invite.
That will be in a future PR

TEST
1. Run the follow code in graphiql multiple times. It's idempotent & the returned result is as expected (change the orgId)
```gql
mutation addApproval {
  addApprovedOrganizationDomains(orgId:"bbITHTgkmc", emailDomains:["parabol.co", "prbl.io"]) {
    ... on AddApprovedOrganizationDomainsSuccess {
      organization {
        approvedDomains
      }
    }
  }
}
```

2. In a new graphiql tab (yeah, i added tabs), run the following code to remove the some domains. The result should be as expected

```gql
mutation removeApproval {
  removeApprovedOrganizationDomains(orgId:"bbITHTgkmc", emailDomains:["parabol.co"]) {
    ... on RemoveApprovedOrganizationDomainsSuccess {
      organization {
        approvedDomains
      }
    }
  }
}
```


mattkrick:
This branch is based on #6447. Once that merges we can give this a code review.

packages/client/modules/admin/containers/Graphql/GraphqlContainer.tsx
```diff
@@ -45,7 +45,7 @@ const GraphqlContainer = () => {
 
   return (
     <GQL>
-      <GraphiQL fetcher={fetcher} ref={graphiql}>
+      <GraphiQL fetcher={fetcher} ref={graphiql} tabs>
```
mattkrick, line 48: graphiql now has tabs, it should be easier to test this

codegen.json
```diff
@@ -28,31 +28,29 @@
       "config": {
         "contextType": "../graphql#GQLContext",
         "mappers": {
+          "AddApprovedOrganizationDomainsSuccess": "./types/AddApprovedOrganizationDomainsSuccess#AddApprovedOrganizationDomainsSuccessSource",
```
mattkrick, line 31: this file is now auto-sorted, so we shouldn't see many more changes like this

packages/server/graphql/public/typeDefs/addApprovedOrganizationDomains.graphql
```diff
@@ -0,0 +1,35 @@
...
+type AddApprovedOrganizationDomainsSuccess {
+  orgId: ID!
+
+  """
+  Organization with the updated approvals
+  """
+  organization: Organization!
+}
+
+extend type Organization {
```
mattkrick, line 29: by extending here, we get type safety in the resolver & don't have to touch the legacy /graphql/types/Organization.ts

packages/server/graphql/public/types/Organization.ts
```diff
@@ -0,0 +1,12 @@
+import getApprovedOrganizationDomainsFromPG from '../../../postgres/queries/getApprovedOrganizationDomainsFromPG'
+import {OrganizationResolvers} from '../resolverTypes'
+
+const Organization: OrganizationResolvers = {
+  approvedDomains: async ({id: orgId}) => {
+    // no need for a dataloader since it's infrequently used
```
mattkrick, line 6: i add a dataloader in the subsequent PR

JimmyLv (approved):
looks great to me! learned the context of how the backend works ✨

PR #6486, opened 2022-05-02
chore(DX): Fix Postgres DB path in dev.yml

Dschoordsch:
The mounted volume did not include the default Postgres storage path,
thus after every db:stop the postgres database was wiped while the
Rethink database was preserved.


mattkrick:
this didn't seem to work for me

repro:
```
git checkout chore/fixPostgresVolume
yarn db:stop
yarn db:start
```

i'll try to dig in more tomorrow.
this is weird because i'm certain the old way used to work, & now i can confirm it doesn't 😕

Dschoordsch:
I'm pretty sure this works. The old way never worked for me. Maybe you've had some environment set to change the path or an older PG version had a different default location?
Repro is
```
git checkout chore/fixPostgresVolume
yarn db:stop
yarn db:start
```
make some changes to DB
```
yarn db:stop
yarn db:start
```
changes are still there

BartoszJarocki:
@Dschoordsch for some reason it works only when I add the same fix to the root `docker-compose.yml`.  I'll dig into this a bit more

Dschoordsch:
@BartoszJarocki Are you running `docker-compose up` manually?

BartoszJarocki:
@Dschoordsch no, I'm using `yarn db:start` and `yarn dev`. I'll dig into this today, maybe it's something OS-specific

BartoszJarocki:
@Dschoordsch ok I know what was the problem. 
I run `docker-compose -v` and it I was on `version 1.29.2, build 5becea4c` even though I had the newest docker installed. 
turns out macOS docker by default does not use docker-compose v2. it has to be enabled manually here
Screenshot 2022-05-18 at 14 44 43

I guess they want everybody to migrate to `docker compose` instead of `docker-compose`. Everything works now.

BartoszJarocki (approved):

BartoszJarocki (approved):
lgtm!

PR #6487, opened 2022-05-02
feat(restrictDomains): Add restriction to acceptTeamInvitation

mattkrick:
You don't need to do all the tests below, but try to log in/signup with different invitations & strategies.

TEST
- [x] add a restriction on email domains as seen in https://github.com/ParabolInc/parabol/pull/6476
- [x] not logged in: visit a mass-invitation link. log in via google & see the error
- [x] not logged in: visit a mass-invitation link. log in via password & see the error
- [x] not logged in: visit a mass-invitation link. log in via password with unverified email & see the error
- [x] not logged in: visit a mass-invitation link. log in via saml & see the error
- [x] not logged in: visit a team-invitation link. log in via google & see the error
- [x] not logged in: visit a team-invitation link. log in via password & see the error
- [x] not logged in: visit a team-invitation link. log in via password with unverified email & see the error
- [x] not logged in: visit a team-invitation link. log in via saml & see the error (in the form, not a snack. good enough 😓)
- [x] logged in: visit a mass-invitation link. see the error
- [x] logged in: visit a team-invitation link. see the error
- [x] logged in: visit a meeting url. see the error
- [x] logged in: get invited & then have your email blacklisted. accept team invite via notification. see the error
- [x] logged in: visit a route on a team you don't belong to that's locked. the other team members don't see a push invitation.
- [x] when typing in a team-invitation to a bad domain, get an error (error doesn't say why, but that'll take more work, sorry)
- [x] ~can't sign up with prbl.in email address with password, it'll ask you to verify the email~ (requires adding prbl.in to SecureDomain before running the last migration. this test isn't necessary as it's a lot of work)
- [x] can login without invitation via saml/password/google
- [x] can sign up without invitation via saml/password/google


packages/client/components/EmailPasswordAuthForm.tsx
```diff
@@ -180,7 +180,11 @@ const EmailPasswordAuthForm = forwardRef((props: Props, ref: any) => {
     atmosphere.setAuthToken(token)
     if (invitationToken) {
       localStorage.removeItem(LocalStorageKey.INVITATION_TOKEN)
-      AcceptTeamInvitationMutation(atmosphere, {invitationToken}, {history, onCompleted, onError})
+      AcceptTeamInvitationMutation(
+        atmosphere,
+        {invitationToken},
+        {history, onCompleted, onError, ignoreApproval: true}
```
mattkrick, line 186: `onError` already propagates the error to the form. passing in `ignoreApproval` prevents a snackbar with the same message from popping up

packages/client/components/EmailPasswordAuthForm.tsx
```diff
@@ -206,14 +210,20 @@ const EmailPasswordAuthForm = forwardRef((props: Props, ref: any) => {
     if (isSignin) {
       LoginWithPasswordMutation(
         atmosphere,
-        {email, password, invitationToken},
+        {email, password, invitationToken: invitationToken || '', isInvitation: !!invitationToken},
```
mattkrick, line 213: i made invitationToken non-null on the server, which means we have to do some stuff to accommodate on the client.

packages/server/email/createEmailVerficationForExistingUser.ts
```diff
@@ -0,0 +1,49 @@
+import base64url from 'base64url'
+import crypto from 'crypto'
+import getRethink from '../database/rethinkDriver'
+import AuthIdentityLocal from '../database/types/AuthIdentityLocal'
+import EmailVerification from '../database/types/EmailVerification'
+import {DataLoaderWorker} from '../graphql/graphql'
+import emailVerificationEmailCreator from './emailVerificationEmailCreator'
+import getMailManager from './getMailManager'
+
+const createEmailVerficationForExistingUser = async (
```
mattkrick, line 10: we already had a helper function to do this with new signups.
this does almost the exact same thing, except it pulls the email & password from the User object instead of from args

packages/server/graphql/mutations/inviteToTeam.ts
```diff
@@ -80,12 +82,24 @@ export default {
...
         return !(user && user.tms && user.tms.includes(teamId))
       })
+
+      // filter out invitees that aren't approved by the org
+      const approvalErrors = await Promise.all(
+        newInvitees.map((email) => {
+          return getIsEmailApprovedByOrg(email, orgId, dataLoader)
+        })
+      )
+      const newAllowedInvitees = newInvitees
```
mattkrick, line 92: if my org is locked down to only inviting people with the company.co domain, then if i invite someone by email that doesn't match that domain, it'll get filtered out

packages/server/graphql/mutations/pushInvitation.ts
```diff
@@ -45,6 +46,16 @@ export default {
       if (!isAuthenticated(authToken)) return standardError(new Error('Not authenticated'))
 
       // VALIDATION
+      const team = await dataLoader.get('teams').load(teamId)
+      if (!team) {
+        return {error: {message: 'Invalid teamId'}}
+      }
+
+      const {orgId} = team
+      const approvalError = await getIsUserIdApprovedByOrg(viewerId, orgId, dataLoader)
```
mattkrick, line 55: if someone goes to the meeting url when they're not on the team & they don't have an org-approved email address, then a push notification is **not** pushed to the other team members asking if they want to invite the stranger to their team.

packages/server/graphql/mutations/signUpWithPassword.ts
```diff
@@ -74,7 +74,7 @@ const signUpWithPassword = {
       if (!nickname || !domain) {
         return {error: {message: 'Invalid email'}}
       }
-      const verificationRequired = await isEmailVerificationRequired(domain)
+      const verificationRequired = await isEmailVerificationRequired(email, dataLoader)
```
mattkrick, line 77: we kinda did domain locking in the past. if an org chose to lock their domain, you couldn't create a parabol account unless you verified your email first. since this PR does that & so much more, i changed the logic to share the same DB table.

packages/server/graphql/public/mutations/acceptTeamInvitation.ts
```diff
@@ -0,0 +1,127 @@
...
+
+const acceptTeamInvitation: MutationResolvers['acceptTeamInvitation'] = async (
+  _source,
+  {invitationToken, notificationId},
+  {authToken, dataLoader, socketId: mutatorId}
+) => {
+  // AUTH WORKAROUND
+  if (!isAuthenticated(authToken) || !invitationToken) {
+    // Workaround for https://github.com/zalando-incubator/graphql-jit/issues/171
+    return {}
```
mattkrick, line 23: found a bug in graphql-jit. this return value is the same thing as if `@include(if: false)` directive worked correctly

packages/server/graphql/public/permissions.ts
```diff
@@ -24,6 +24,9 @@ export type PermissionMap<T> = {
 const permissionMap: PermissionMap<Resolvers> = {
   Mutation: {
     '*': isAuthenticated,
+    // Bug in graphql-jit means it will always get called
+    // https://github.com/zalando-incubator/graphql-jit/issues/171
+    acceptTeamInvitation: and(/* isAuthenticated, */ rateLimit({perMinute: 50, perHour: 100})),
```
mattkrick, line 29: authentication is checked in the resolver until we fix this bug

packages/server/postgres/migrations/1651532750081_domainIndex.ts
```diff
@@ -0,0 +1,18 @@
+import {Client} from 'pg'
+import getPgConfig from '../getPgConfig'
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+  CREATE INDEX IF NOT EXISTS "idx_OrganizationApprovedDomain_domain" ON "OrganizationApprovedDomain"("domain");`)
```
mattkrick, line 8: by indexing on the domain filed, we can get rid of the rethinkdb SecureDomain table

packages/server/postgres/queries/getUsersByDomain.ts
```diff
@@ -1,10 +1,10 @@
-import {getUsersByDomainQuery} from './generated/getUsersByDomainQuery'
 import getPg from '../getPg'
 import IUser from '../types/IUser'
+import {getUsersByDomainQuery} from './generated/getUsersByDomainQuery'
 
 const getUsersByDomain = async (domain: string): Promise<IUser[]> => {
   const users = await getUsersByDomainQuery.run({domain}, getPg())
-  return users as IUser[]
+  return users as unknown as IUser[]
```
mattkrick, line 7: I made IUser more strict

mattkrick:
@nickoferrall let me know what you think next steps should be

nickoferrall:
Makes sense! I imagine that users signing up with their personal email address won't be that unusual (I think it's happened a few times in the Parabol org), so we could give them some feedback in the sign-up flow if they're using an invite link & the domain is `gmail.com`, for example. But I think that's out of scope for this PR.

I mentioned [here](https://github.com/ParabolInc/parabol/pull/6487#pullrequestreview-996826739) that I had difficulty reproducing the error, so if you could clarify the testing steps that would be great.

mattkrick:
Sure, here's the steps to repro:
- Create a new org Foo
- run the following mutation:
```gql
mutation addApproval {
  addApprovedOrganizationDomains(orgId:"foo", emailDomains:["b.co"]) {
    ... on AddApprovedOrganizationDomainsSuccess {
      organization {
        approvedDomains
      }
    }
  }
}
```

- copy the invitation link for a team on org Foo
- in incognito, paste the invitation link
- sign up with a brand new email address that doesn't end in `b.co`, e.g. use a@a1234at.co
- see that your account is created, but you are not on any team. you do get a notification at the bottom of your screen giving you an error (see picture below)
-
Screen Shot 2022-06-06 at 3 07 49 PM

nickoferrall (commented):
LGTM! Thanks, I can see the snackbar message

PR #6490, opened 2022-05-03
fix: ignore updateAt field in checkTableEq for Teams

shapel:
As part of #6239 struggle, we encountered the fact that Postgres & RethinkDB store "Team" records with `updatedAt` that differ by about a year. 

This appears to be because the system does not always update the `updatedAt` field of Postgres or RethinkDB in some cases. This will be fixed in a separate PR.

Therefore, we can skip checking the `updateAt` field for the time being when comparing records.
https://github.com/ParabolInc/parabol/pull/6239#discussion_r863572037

### DIFF
--- a/packages/server/postgres/utils/checkTeamEq.ts
+++ b/packages/server/postgres/utils/checkTeamEq.ts
@@ -1,7 +1,7 @@
-import Team from '../../database/types/Team'
 import getRethink from '../../database/rethinkDriver'
-import {checkTableEq} from './checkEqBase'
+import Team from '../../database/types/Team'
 import getTeamsByIds from '../queries/getTeamsByIds'
+import {checkTableEq} from './checkEqBase'
 
 const alwaysDefinedFields: (keyof Partial<Team>)[] = [
   'name',
@@ -9,8 +9,7 @@ const alwaysDefinedFields: (keyof Partial<Team>)[] = [
   'isArchived',
   'isPaid',
   'tier',
-  'orgId',
-  'updatedAt'
+  'orgId'
 ]
 
 const maybeUndefinedFieldsDefaultValues: {[Property in keyof Partial<Team>]: any} = {
--- a/packages/server/postgres/utils/checkTeamEq.ts
+++ b/packages/server/postgres/utils/checkTeamEq.ts
@@ -3,14 +3,17 @@ import Team from '../../database/types/Team'
 import getTeamsByIds from '../queries/getTeamsByIds'
 import {checkTableEq} from './checkEqBase'
 
-const alwaysDefinedFields: (keyof Partial<Team>)[] = [
+const alwaysDefinedFields: (keyof Team)[] = [
   'name',
   'createdAt',
   'isArchived',
   'isPaid',
   'tier',
-  'orgId'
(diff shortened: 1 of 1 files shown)


shapel:
@tianrunhe can you review this?

packages/server/postgres/utils/checkTeamEq.ts
```diff
@@ -1,16 +1,15 @@
...
+import {checkTableEq} from './checkEqBase'
 
 const alwaysDefinedFields: (keyof Partial<Team>)[] = [
   'name',
   'createdAt',
   'isArchived',
   'isPaid',
   'tier',
-  'orgId',
-  'updatedAt'
```
tianrunhe, line 13: I think it'd be more clear to create another array called `ignoredFields` and use that to filter out fields we don't care.

tianrunhe (changes requested):

packages/server/postgres/utils/checkTeamEq.ts
shapel: I fixed it, plus added somehow missing fields like 'createdBy', 'isOnboardTeam'.

tianrunhe (changes requested):

packages/server/postgres/utils/checkTeamEq.ts
```diff
@@ -1,17 +1,19 @@
...
-const alwaysDefinedFields: (keyof Partial<Team>)[] = [
+const alwaysDefinedFields: (keyof Team)[] = [
   'name',
   'createdAt',
   'isArchived',
   'isPaid',
   'tier',
   'orgId',
-  'updatedAt'
+  'createdBy',
```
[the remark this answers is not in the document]
shapel: Nice caught!
I suppose we should skip this field as well. Because, as far as I understand, the `getPairNeFields` logic only works in cases where the value from RethinkDB can be `null` or `undefined`.

packages/server/postgres/utils/checkTeamEq.ts
```diff
@@ -1,17 +1,18 @@
...
   'name',
   'createdAt',
   'isArchived',
   'isPaid',
   'tier',
   'orgId',
-  'updatedAt'
+  'isOnboardTeam'
 ]
+const ignoredFields: (keyof Team)[] = ['updatedAt', 'createdBy']
```
tianrunhe, line 15: Instead of ignoring it, I think we should put it in the `maybeUndefinedFieldsDefaultValues` field in the following lines. See reference on [checkUserEq](https://github.com/ParabolInc/parabol/blob/09c12a5267f0a04fe7370ac9a63f5b4c879256c9/packages/server/postgres/utils/checkUserEq.ts#L19-L34)

tianrunhe (changes requested):

packages/server/postgres/utils/checkTeamEq.ts
shapel: Yep, you are right!

tianrunhe (approved):
LGTM! Thanks for fixing this!

Dschoordsch (approved):

PR #6493, opened 2022-05-03
refactor: Set `updateAt` field via a Postgres trigger

shapel:
As part of #6239 struggle, we encountered the fact that Postgres & RethinkDB store "Team" records with `updatedAt` that differ by about a year.  This appears to be because the system does not always update the `updatedAt` field of Postgres or RethinkDB in some cases.
As @Dschoordsch suggested at https://github.com/ParabolInc/parabol/pull/6239#discussion_r863542513 this PR adds triggers for all tables that have `updateAt`.


shapel:
👋 @tianrunhe could you review this?

shapel:
Hello @tianrunhe, just pinging to make sure that it's on your radar 🤞

tianrunhe (changes requested):
GREAT work! Thanks so much for doing this!

I only have 1 suggestion: why not combine all of the migrations into one? It'll be faster to execute during deployment.

shapel:
@tianrunhe thanks for review.
I did what you suggested + rebase master

tianrunhe (approved):
LGTM! Thanks again for doing this!

Dschoordsch (approved):
Thanks for taking this on!

Issue #6395, opened 2022-04-13
Standup UI - Activities from other team members get real-time updates


tianrunhe:
AC:
- [ ] updating team response content mutation payload is added to the meeting subscription
- [ ] verify: When Bob joins the team, Alice sees a new response card show up on her screen, with Bob's avatar

Estimate: 5 hours

mattkrick:
since we're using tiptap, we could use this opportunity to introduce collaborative editing so instead of seeing "X is typing..." it looks like a google doc where the cursor just pops up on the document

PR #6504, opened 2022-05-05
feat(standups): Activities from other team members get real-time updates

BartoszJarocki:
Fixes: #6395 

## Important info
indicator 'XYZ is adding a response' will be a separate task
This PR depends on https://github.com/ParabolInc/parabol/pull/6388. 

## Tests

1. Updates
- create a standup meeting, join from 2 browsers
- update the response, click away (currently mutation is fired in `onBlur`), make sure the other user sees the update
- update the response as another user, click away (currently mutation is fired in `onBlur`), and make sure the first one sees the update

2. New team member
- copy the team invite link
- create & join the standup meeting
- open incognito, accept the invite 
- make sure the first user sees the new card for a new team member being added

https://user-images.githubusercontent.com/1017620/167173342-bf914132-f910-4e28-970c-7d8bc6f7cbad.mp4


packages/client/components/promptResponse/PromptResponseEditor.tsx
```diff
@@ -65,6 +65,14 @@ const PromptResponseEditor = (props: Props) => {
     editable: !readOnly
   })
 
+  useEffect(() => {
+    if (!editor) {
+      return
+    }
+
+    editor.commands.setContent(content)
+  }, [editor, content])
```
BartoszJarocki, line 74: this is needed as the `content` specified [here](https://github.com/ParabolInc/parabol/pull/6504/files#diff-b3416557ebc46ac9b7d1fa9ef62f5564cf40767ea9895e098715c22cb32d4549R60) is just the initial content. without that, when the relay store is updated during the subscription event or the mutation, the editor won't update.

packages/client/mutations/UpsertTeamPromptResponseMutation.ts
```diff
@@ -1,32 +1,18 @@
...
-          ... on TeamPromptResponsesPhase {
-            stages {
-              ... on TeamPromptResponseStage {
-                ...TeamPromptResponseCard_stage
-              }
-            }
-          }
-        }
-      }
-    }
```
BartoszJarocki, line 29: we don't actually need that, updater will just insert the response for the stage related to a given user

packages/client/mutations/UpsertTeamPromptResponseMutation.ts
```diff
@@ -59,33 +66,9 @@ const UpsertTeamPromptResponseMutation: StandardMutation<TUpsertTeamPromptRespon
...
-          id: clientTempId(viewerId),
-          meetingId,
-          sortOrder: 0,
-          content
-        })
-        stage.setLinkedRecord(newResponse, 'response')
-      } else {
-        // update
-        response.setValue(content, 'content')
-      }
```
BartoszJarocki, line 88: we don't need an optimistic updater at all. relay store will be updated upon the successful finish of mutation, tip tap always displays the actual content so an optimistic updater doesn't make any difference. if I miss something, I'm happy to improve.

BartoszJarocki:
@jmtaber129 do you mind taking a look? thank you!

packages/client/components/promptResponse/PromptResponseEditor.tsx
jmtaber129: +1 Thoughts on using the dependency list from `useEditor` (see https://github.com/ParabolInc/parabol/pull/6469/commits/1bdb26fd6af9b2eb525473490c5ce3313cd5de24#diff-b3416557ebc46ac9b7d1fa9ef62f5564cf40767ea9895e098715c22cb32d4549R68)?  I'm not sure what the tradeoffs are, but that seems cleaner in the context of hooks.

packages/client/mutations/UpsertTeamPromptResponseMutation.ts
jmtaber129: -1 It seems like the optimistic updater is needed for cases where content changes between null and non-null.  Specifically, there seems to be a bug where the editor state flashes between the old and new state in these cases.  See loom - https://www.loom.com/share/52896ca850d64a57b5ad3bbc20342cee

I think this will also be needed to immediately update the response in an open discussion drawer in https://github.com/ParabolInc/parabol/pull/6469/commits/1bdb26fd6af9b2eb525473490c5ce3313cd5de24#diff-b3416557ebc46ac9b7d1fa9ef62f5564cf40767ea9895e098715c22cb32d4549R68

jmtaber129 (changes requested):
Looking good!  Just one suggestion + one minor bug + merge conflicts

packages/client/components/promptResponse/PromptResponseEditor.tsx
BartoszJarocki: how did I miss that? thanks for pointing that out! `useEffect` should really be the last resort, this is much better.

packages/client/mutations/UpsertTeamPromptResponseMutation.ts
BartoszJarocki: @jmtaber129 oh, that's correct. I've added the `optimisitcUpdater`, could please double-check now? thank you!

jmtaber129 (approved):
LGTM!

packages/client/mutations/UpsertTeamPromptResponseMutation.ts
```diff
@@ -51,41 +39,70 @@ const mutation = graphql`
   }
 `
 
-const UpsertTeamPromptResponseMutation: StandardMutation<TUpsertTeamPromptResponseMutation> = (
-  atmosphere,
-  variables,
-  {onError, onCompleted}
-) => {
+export const upsertTeamPromptResponseUpdater: SharedUpdater<
```
Dschoordsch, line 42: -1 This updater is not needed. Works for me without it.
As a rule of thumb, if you're returning an updated object from a mutation, the relay store will merge it into the existing object. Only case where an updater is required is if you only add part of the modified object (like an element in an array) and you need to update the parent.

Dschoordsch (changes requested):
+1 I have some flicker with Firefox when updating my own response
weird update flicker

packages/client/mutations/UpsertTeamPromptResponseMutation.ts
BartoszJarocki: very much true! I removed the updater, optimistic updater, and replaced it with an optimistic response.

BartoszJarocki:
hm. seems like removing the updater broke the case when initial responses are added. will investigate...

Dschoordsch:
> hm. seems like removing the updater broke the case when initial responses are added. will investigate...

That's weird, I tested that case as well with a new member joining.

BartoszJarocki:
seems like it's not updater but my flickering fix, ugh

BartoszJarocki:
ok. without an updater (not tied to the mutation, but only to the meeting subscription), the following case fails always on my end:
1. Start async standup as Alice
2. Join async standup as Bob
3. Write the answer as Bob - Alice can't see it until refresh
4. Write the answer as Alice - Bob can't see it until refresh

Dschoordsch (approved):
Did not do another test round, I trust you and James there.

PR #6505, opened 2022-05-05
fix(SDL): Resolve subscription payload type

Dschoordsch:
When a subscription union is extended via SDL, it wasn't resolved
properly. Query the schema for non-abstract object types and return
those directly solves it for this use case.


BartoszJarocki (approved):
great find!

mattkrick:
nice find! i totally forgot about that little chunk of logic in there

PR #6513, opened 2022-05-05
chore(deps): bump rethinkdb-ts-migrate

mattkrick:
this bumps some dependencies in rethinkdb-ts-migrate to get rid of the dependabot errors.
we couldn't merge the dependabot fixes because our versions were too old


mattkrick:
merging because it's just a deps bump on the rethinkdb migrator

PR #6517, opened 2022-05-05
chore(deps): bump pm2

mattkrick:
pm2 bupmed cron, which bumped moment, which has a vuln


mattkrick:
merging as it's just a security bump

PR #6518, opened 2022-05-05
chore(deps): remove graphql-typescript-definitions

mattkrick:
this package was unused & had a chain of deps that were vulnerable (including moment)


mattkrick:
safe to merge as soon as tests pass

PR #6519, opened 2022-05-05
chore(dx): remove generated pg files from git

mattkrick:
To make code reviews go faster, i'm removing pg generated files from git.
In production, prod.js will now build these before building the servers.
In dev, pm2 will watch the /src directory and whenever you make a change to a SQL file, the updated file will be generated automatically.


mattkrick:
grrrrr this won't work out.
we've got like 10 generated queries that are used by migrations, so we need the queries to run those migrations, but we need the migrations to build the queries... 

if we want to move forward with this, we'd have to move all those queries from the migrations first

mattkrick:
Having merge conflicts caused by nondeterministically generated files is terrible DX.
I'm gonna timebox this to today & force it to work.

mattkrick:
- [ ] the typecheck test is failing because `yarn pg:build` requires access to pg, so we'll probably have to spin up a pg server to get it to pass.
- [ ] CI is failing for the same reason, pg is on 5432 but it's looking for it on port 25060

.circleci/config.yml
```diff
@@ -126,7 +134,7 @@ jobs:
               export ACTION_VERSION=$(grep '"version":' package.json | cut -d\" -f4)
               git config --global user.name "Parabol CircleCI"
               git config --global user.email "admin+circleci@parabol.co"
-              git add queryMap.json build dist packages/client/__generated__ scripts/toolbox -f
+              git add queryMap.json build dist packages/client/__generated__ scripts/toolbox packages/server/postgres/queries/generated -f
```
mattkrick, line 137: we won't be able to test this until it gets pushed to staging, but i'm pretty sure this should work

.github/workflows/typecheck.yml
```diff
@@ -4,17 +4,43 @@ on: [push]
 
 jobs:
   build:
+    services:
```
mattkrick, line 7: to generate the typed files, pgtyped needs to access the most recent database schema.
to get the most recent database schema, we have to run all the migrations on it.
some of those migration files also touch rethinkdb, so we have to require rethinkdb, too.

.github/workflows/typecheck.yml
```diff
@@ -4,17 +4,43 @@ on: [push]
...
+      rethinkdb:
+        image: rethinkdb:2.4.0
+        ports:
+          - "8080:8080"
+          - "29015:29015"
+          - "28015:28015"
+      postgres:
+        image: postgres:12
```
mattkrick, line 15: very important! postgres:latest breaks pgtyped...

packages/server/database/migrations/20210726140543-estimate-stage-tasks.ts
```diff
@@ -3,11 +3,10 @@ import {R} from 'rethinkdb-ts'
 import dndNoise from '../../../client/utils/dndNoise'
 import convertToTaskContent from '../../../client/utils/draftjs/convertToTaskContent'
 import generateUID from '../../generateUID'
-import getTemplateRefsByIds from '../../postgres/queries/getTemplateRefsByIds'
-import insertTaskEstimate from '../../postgres/queries/insertTaskEstimate'
+import {getTemplateRefsByIds, insertTaskEstimate} from '../../postgres/generatedMigrationHelpers'
```
mattkrick, line 6: all the migration files have a simple change like this.

BartoszJarocki:
as a side note, it is possible to configure github to not show diffs for particular files/directories
https://docs.github.com/en/repositories/working-with-files/managing-files/customizing-how-changed-files-appear-on-github

it's _not_ a perfect solution, as the file will still show in PR but it'll look like this
Screenshot 2022-06-04 at 19 43 00

jmtaber129 (approved):
LGTM, but I'm curious - why is pgtyped generating non-deterministic output?

mattkrick:
i think it's not calling `sort` before writing the enums & interfaces 
e.g. 

```ts
type tier = 'pro' | 'enterprise'
type tier = 'enterprise' | 'pro'
```

it'd be easy enough to PR, but the big win here is sticking a watcher on pgtyped so whenever the SQL changes it regenerates

PR #6520, opened 2022-05-06
Query complexity bug in GitLab

nickoferrall:


mattkrick:
I think a limit of 15 is not going to fly for users, from what I've seen folks love to use a lot of issues.

To tackle this, here's some next steps: 
- print the query that gets sent to gitlab. does it look right? are there parts that are duplicated? the goal is to minimize the number of fields being requested from gitlab
- nest-graphql-endpoint batches queries together before sending them to gitlab. can you come up with a batching heuristic to know when to start a new batch? If you can count the number of fields & the `first, last` arg values for connections you should be able to get a pretty reliable complexity computed locally.

mattkrick:
grrrrr this won't work out.
we've got like 10 generated queries that are used by migrations, so we need the queries to run those migrations, but we need the migrations to build the queries... 

if we want to move forward with this, we'd have to move all those queries from the migrations first

nickoferrall:
Initially, we only need the issue `id`, `iid`, and `title` for the Poker sidebar (which is currently broken at 19 issues when you refresh the page). However, the `NewMeeting` query includes the `descriptionHtml`, `webPath`, and `webUrl` for each issue, even though we only need it when we click on an Estimate Stage. 

`descriptionHtml` alone has a query complexity of 5. By removing the PokerEstimatePhase fragments [here](https://github.com/ParabolInc/parabol/blob/0c4cc3d4e69eb7d19d244a75f33cba02e46bccbb/packages/client/components/PokerEstimatePhase.tsx#L103), I can refresh the page and see 45 issues in the sidebar (50 is still too complex atm). 

My current plan: implement a `PokerEstimatePhaseRoot` component that executes a root query for each individual Estimate phase. This should reduce the number of fields requested in the initial GitLab query in `NewMeeting`.

nickoferrall:
> to make sure i understand, if a meeting has 50 issues, you would issue 50 requests?

No, we have the `NewMeeting` request which grabs 50 issues with fields that we don't need in the sidebar, such as `descriptionHtml`. I want the `NewMeeting` query to only grab the fields needed for the sidebar. 

Then, when the user goes to a specific Estimate stage, the `PokerEstimatePhaseRoot` query executes for that individual stage. It only grabs the fields, such as `descriptionHtml`, for the one issue in that stage. 

On `estimate/1`, there would be one `PokerEstimatePhaseRoot` request. When the user goes to `estimate/2`, a second request would be sent for that issue.

packages/client/components/GitLabScopingSearchResultItem.tsx
```diff
@@ -58,14 +58,13 @@ const GitLabScopingSearchResultItem = (props: Props) => {
         id
         iid
         title
-        webPath
```
nickoferrall, line 61: We can get the `webPath` using `webUrl` so I've removed this

packages/client/components/PokerEstimatePhase.tsx
```diff
@@ -40,13 +38,49 @@ const EstimateAreaWrapper = styled('div')({
...
+              endedAt
+              isCommentUnread
+              isRightDrawerOpen
+              localStage {
+                ...PokerEstimateHeaderCard_stage
+              }
+              # phases {
+              #   ... on EstimatePhase {
+              #     stages {
+              #       ...PokerEstimateHeaderCard_stage
```
nickoferrall, line 68: `PokerEstimateHeaderCard_stage` includes expensive fields such as `descriptionHtml`. Rather than query this for every stage, I just want to query this for the `localStage`, which is what the user currently needs

packages/client/mutations/CreateTaskIntegrationMutation.ts
```diff
@@ -12,8 +12,9 @@ graphql`
     task {
       ...IntegratedTaskContent_task
       integration {
-        __typename
+        # __typename
         ... on JiraIssue {
+          __typename
```
nickoferrall, line 17: I was playing around with removing `__typename` from the gitlab query. Will revert this change

nickoferrall (commented):
Hey @mattkrick, I've implemented the above approach. I can now select 45 GitLab issues, refresh the page, and see all the issues in the sidebar. I've added a [PokerEstimatePhaseRoot](https://github.com/ParabolInc/parabol/blob/87f80445e05f2644c658d40b2d7fcf88e4d26e6d/packages/client/components/PokerEstimatePhaseRoot.tsx) component.

To do:

- [ ] 50 issues doesn't work atm - it has a complexity of 264, just above the 250 limit. Try to remove `__typename` from the query and use the gid or iid field to determine whether it's a GitLab issue.
- [ ] If I proceed from the Scope phase to the Estimate phase, I can click the hyperlink in the `PokerEstimateHeaderCard`. If I refresh the page, I can't and `descriptionHtml` is null.
- [ ] Can't push an estimate to a GitLab issue atm.

mattkrick:
building something that might break if we add an extra couple fields feels bad.
what do you think about changing the batching strategy?
see https://github.com/ParabolInc/nest-graphql-endpoint/pull/10

nickoferrall:
Closing in favour of https://github.com/ParabolInc/parabol/pull/6541

PR #6521, opened 2022-05-06
feat(pg): reduce max pg connections to 30

mattkrick:
Reduce connections to 30 for the application.
Things that connect directly to the DB (e.g. pgadmin, analytics, psql) will still take up a connection, so we'll need to keep those pinned at < 15, but that should be pretty easy to do.

### DIFF
--- a/packages/server/postgres/getPgConfig.ts
+++ b/packages/server/postgres/getPgConfig.ts
@@ -6,7 +6,7 @@ const getPgConfig = () => ({
   database: process.env.POSTGRES_DB,
   host: process.env.POSTGRES_HOST,
   port: Number(process.env.POSTGRES_PORT),
-  max: PROD ? 40 : 5
+  max: PROD ? 30 : 5
 })
 
 export default getPgConfig


mattkrick:
pinging @Dschoordsch to see if this is an appropriate solution for the near term.

packages/server/postgres/getPgConfig.ts
```diff
@@ -6,7 +6,7 @@ const getPgConfig = () => ({
   database: process.env.POSTGRES_DB,
   host: process.env.POSTGRES_HOST,
   port: Number(process.env.POSTGRES_PORT),
-  max: PROD ? 40 : 5
+  max: PROD ? 30 : 5
```
Dschoordsch, line 9: I might miss something, but the total max is 47, so having 30 per process can exhaust the pool with just 2 processes. Given we have 2 executors, 20 seems like a better choice here.

Dschoordsch (changes requested):
Not worse than before

Dschoordsch (approved):

Dschoordsch:
@jchandler-parabol could you please double check that I didn't miss something?

jchandler-parabol:
@Dschoordsch Your logic to further reduce it to 20 makes sense from a ceiling point-of-view. If both processes max out, then you still have 7 for devs during that spike. In the course of creating last week's release, I had to trouble someone to close their pgAdmin instance in order to perform the backup step. So, we're certainly stepping on each other's feet. All that said, I do not yet know enough about the application to validate that 20 (per process) meets the minimal requirements for our current load. I do like the plan to setup pooling and agree that this is a great way to buy some time until we can schedule that story.

PR #6526, opened 2022-05-09
Create migrationToPostgres.md

asherif844:
Documentation on migrating techniques from RethinkDB that is no longer commercially supported to Postgres, an open source database that is the 2nd most popular database in the world.  Documentation also available for deploying Postgres on Azure leveraging Flexible Server that provides enterprise support and scalability.


BartoszJarocki (commented):
@asherif844 thank you for the contribution! Migration to Postgres is indeed on our roadmap. The article about migrating to Postgres using Haskel, [persistent](http://hackage.haskell.org/package/persistent), and [esqueleto](http://hackage.haskell.org/package/esqueleto) was a good read! I'm the tooling is quite different from what we use - Node, TS, and PG typed. Putting that kind of instructions in our repo might be confusing as it's not too relevant to our case.

BartoszJarocki:
I'm closing this as the content isn't too relevant to our case and it might be confusing to put this as official docs.

PR #6539, opened 2022-05-10
fix(metrics): add UTM parameters to URLs in team invitation emails

### LINKED ISSUE #6498: UTM Parameters added to
## Issue - Bug

In our 'you've been invited to a team!' email, we do not have UTM parameters on the links that go to the website, that's these: 
invite-email-annotated

Without these, we can't tell how much this email is driving traffic to the site. 

The UTM parameters should be: 
- Source = `invite email`
- Medium = `email`
- Campaign = `invitations` 

[Documented here. ](https://www.notion.so/parabol/UTM-Parameters-Guidelines-f2e1ada7e80042ff90a304d6511d9e27)

When users click on links in the invite email to go to the website, those links include UTM parameters. 

tianrunhe:
Fixes #6498 

Tests:
1. Invite a user via email
2. Click the links in the invitation email and verify the URL contains `?utm_source=invite+email&utm_medium=email&utm_campaign=invitations`

### DIFF
--- a/packages/client/modules/email/components/TeamInvite.tsx
+++ b/packages/client/modules/email/components/TeamInvite.tsx
@@ -1,5 +1,6 @@
 import {ExternalLinks} from 'parabol-client/types/constEnums'
 import React from 'react'
+import makeAppURL from '../../../utils/makeAppURL'
 import {emailCopyStyle, emailLinkStyle, emailProductTeamSignature} from '../styles'
 import Button from './Button'
 import EmailBlock from './EmailBlock/EmailBlock'
@@ -37,6 +38,12 @@ const meetingCopyLabelLookup = {
   teamPrompt: 'an Async Standup Meeting'
 } as const
 
+const urlParams = {
+  utm_source: 'invite email',
+  utm_medium: 'email',
+  utm_campaign: 'invitations'
+}
+const options = {searchParams: urlParams}
 export interface TeamInviteProps {
   appOrigin: string
   inviteeName: string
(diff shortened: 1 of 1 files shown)


packages/client/modules/email/components/TeamInvite.tsx
```diff
@@ -119,23 +126,31 @@ const TeamInvite = (props: TeamInviteProps) => {
             'Parabol is software for remote teams to run online retrospective and check-in meetings. See the video and links below:'
           }
         </p>
-        <a href={ExternalLinks.GETTING_STARTED_RETROS} style={videoLinkStyle} title='Retro 101'>
+        <a
+          href={makeAppURL('https://www.parabol.co/', 'resources/retrospective-meetings', options)}
```
Dschoordsch, line 130: -1 I would prefer if we continue using the constants, thus if we change the marketing page we don't forget to update it here. You can just put a helper inside this file
```ts
function appendUTM(url: string) {
  const newUrl = new URL(url)
  Object.entries(urlParams).forEach(([name, value]) => {
    newUrl.searchParams.append(name, value)
  })
  return newUrl.toString()
}
```
and call it with 
```ts
            href={appendUTM(ExternalLinks.GETTING_STARTED_RETROS)}
```

Dschoordsch (changes requested):

Dschoordsch:
I took the freedom to make the changes myself so it can be in this ship

tianrunhe:
@Dschoordsch THANK YOU! 🎉

PR #6542, opened 2022-05-10
fix(metrics): fix the bug where user deletion event won't update HubSpot

### LINKED ISSUE #4692: HubSpot property 'is user removed'
## Issue - Bug

The property HubSpot property 'is user removed' does not update from false to true when someone deletes their Parabol account. I think this might also cause deleted users to still receive marketing emails since the workflow to stop this relies on this property.

Source: API - Property changed by use of a HubSpot API endpoint.

**Estimate**: 1 hour

tianrunhe:
Fixes #4692 

Couple of issues:
1. When executing GraphQL query, we should use `ServerAuthToken` rather than using user's token as we need admin access to run the private schema
2. `isRemoved` and company `userCount` & `activeUserCount` will only be accurate after the deletion is done so I moved the HubSpot logic accordingly
3. After deletion, the user email will be augmented. To match with the record in HubSpot (it uses email), we manually set the field to user's email before deletion.

Tests:
1. Copy the [`HUBSPOT_API_KEY`](https://github.com/ParabolInc/action-devops/blob/master/environments/production#L67) to your `.env`
4. Register a test account
5. Delete the account from User->Profile
6. Go to [HubSpot](https://app.hubspot.com/contacts/3888472/contact/76428101), find the test account you created and verify "Is User Removed?" is true.
Screen Shot 2022-05-10 at 3 38 22 PM


nickoferrall (approved):
LGTM! Tested and it does the trick

packages/server/graphql/mutations/deleteUser.ts
```diff
@@ -65,13 +67,37 @@ export default {
...
-      dataLoader,
-      authToken,
-      validReason
-    )
+    const deletedUserEmail = await softDeleteUser(userIdToDelete, dataLoader)
     await markUserSoftDeleted(userIdToDelete, deletedUserEmail, validReason)
+
+    // Update HubSpot after deletion
+    const executeGraphQL = require('../executeGraphQL').default
+    const parabolPayload = await executeGraphQL({
```
Dschoordsch, line 75: -1 you're now missing the event for `hardDeleteUser` mutation which is calling `softDeleteUser` as well.

Dschoordsch (changes requested):

tianrunhe:
Hey @Dschoordsch , I made some fixes but for some reason, after my commit, the app server can no longer start. I couldn't figure out why... Can you see if you can reproduce this issue on your desktop? Thanks!!!

packages/server/graphql/mutations/helpers/sendAccountRemovedToSegment.ts
```diff
@@ -0,0 +1,38 @@
+import ServerAuthToken from '../../../database/types/ServerAuthToken'
+import segmentIo from '../../../utils/segmentIo'
+
+const executeGraphQL = require('../../executeGraphQL').default
```
Dschoordsch, line 4: -2 It's this line which causes the app to not start. It creates an import loop which blocks itself. Move it inside the function (as it was before) and it works.

Dschoordsch (changes requested):

packages/server/graphql/mutations/helpers/sendAccountRemovedToSegment.ts
tianrunhe: Thanks so much for this!!!

Dschoordsch (approved):

Issue #5820, opened 2021-12-21
Jira Server: persist search query


Dschoordsch:
Functionally similar to `persistJiraSearchQuery` and `persistGitHubSearchQuery`.

## Acceptance criteria
* [ ] add generic table to persist integration specific settings (see #5968)
* [ ] add mutation `persistIntegrationSearchQuery` and add logic for Jira Server
* [ ] add logic to UI and `TaskIntegration` so it is used
* [ ] create ticket to migrate `persistJiraSearchQuery` and `persistGitHubSearchQuery` over

jordanh:
**Include in next sprint?: N**
  [See the discussion](https://action.parabol.co/meet/aGjmYpJBIs/estimate/8) in Planning for Sprint #​​96

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

Dschoordsch:
**Story points: 13**
  [See the discussion](https://action.parabol.co/meet/chrwZ1fHoc/estimate/19) in Sprint Poker #​​1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

Dschoordsch:
**Include in next sprint?: Y**
  [See the discussion](https://action.parabol.co/meet/chrwZ1fHoc/estimate/20) in Sprint Poker #​​1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

igorlesnenko:
@Dschoordsch Our search queries have different format for different integrations, and may get even more differences in the future if we want to persist more filters.

For example, for JiraServer we persist the following information:
```
queryString
isJQL
projectKeyFilters
```

For GitHub it is only:

```
queryString
```

if we want a generic mutation `persistIntegrationSearchQuery` we probably need to submit the whole `query` as single JSON field and read an array of JSON on the UI
```
teamMember {
  integrations {
    atlassian {
      searchQueries
    }
  }
```

This way we can probably also implement generic logic on the UI.

Does that sound okay to you?

igorlesnenko:
I also suggest to use a separate table `IntegrationSearchQuery`
```
id 
service
query: jsonb
lastUsedAt
```

As currently, we load all the queries and manually check uniqueness, do sorting, slicing array etc.
This all could be done with a single `insert on conflict update`

Dschoordsch:
[...]

PR #6551, opened 2022-05-12
feat(jira-server): Save and allow to reuse recent search queries in the scope phase of poker meeting for JiraServer

igorlesnenko:
Fixes: #5820

This PR introduces a new PostgreSQL table and new mutaton for storing integrations search queries.

Recent Jira Server search queries are now automatically saved

Also see the next issue with suggeted improvements: #6567

**How to test:**

- Create a poker Meeting with Jira Server integration connected
- On a Jira Server tab:
- Enter any search string and pick any found issue, see the search query was saved in the dropdown menu.
- Select any project in the project filter menu, enter any search string and pick any found issue, see the search query was saved in the dropdown menu.
- Clear the search query and the project filter, select both previous saved queries, see that they were executed correctly
- Remove queries, see they were removed correctly


packages/client/mutations/PersistJiraServerSearchQueryMutation.ts
```diff
@@ -0,0 +1,57 @@
...
+`
+
+const mutation = graphql`
+  mutation PersistJiraServerSearchQueryMutation(
+    $teamId: ID!
+    $service: IntegrationProviderServiceEnum!
+    $providerId: Int
+    $jiraServerSearchQuery: JiraServerSearchQueryInput
+  ) {
+    persistIntegrationSearchQuery(
```
igorlesnenko, line 29: We still need to wrap the generic mutation in an individual one on a frontend to request the specific integration response above. An issue with the suggestion to improve: #6567

packages/server/graphql/types/JiraServerIntegration.ts
```diff
@@ -162,6 +175,53 @@ const JiraServerIntegration = new GraphQLObjectType<{teamId: string; userId: str
...
+      resolve: async ({teamId, userId}, _args: unknown, {dataLoader}) => {
+        const auth = await dataLoader
+          .get('teamMemberIntegrationAuths')
+          .load({service: 'jiraServer', teamId, userId})
+
+        if (!auth) {
+          return null
+        }
+
+        const searchQueries = await getLatestIntegrationSearchQueriesWithProviderId({
```
igorlesnenko, line 202: Two separate SQL queries to be able to request `AND providerId IS null`

packages/server/postgres/queries/src/getLatestIntegrationSearchQueriesWithProviderIdQuery.sql
```diff
@@ -0,0 +1,13 @@
+/*
+  @name getLatestIntegrationSearchQueriesWithProviderIdQuery
+*/
+SELECT * FROM "IntegrationSearchQuery"
+WHERE "teamId" = :teamId
+AND "userId" = :userId
+AND "service" = :service
+AND "providerId" = :providerId
```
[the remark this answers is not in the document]
Dschoordsch: What's the use case where we don't have a providerId?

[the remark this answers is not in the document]
igorlesnenko: @Dschoordsch I think we don't have it for hardcoded integrations like Jira or GitHub

packages/server/postgres/migrations/1652861495447_addIntegrationSearchQuery.ts
```diff
@@ -0,0 +1,44 @@
...
+    CREATE TABLE IF NOT EXISTS "IntegrationSearchQuery" (
+      "id" INT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+      "userId" VARCHAR(100) NOT NULL,
+      "teamId" VARCHAR(100) NOT NULL,
+      "providerId" INT,
+      "service" "IntegrationProviderServiceEnum" NOT NULL,
+      "query" JSONB NOT NULL DEFAULT '{}',
+      "lastUsedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+      "createdAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+      "updatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
```
Dschoordsch, line 17: -1 if we add an "updatedAt" field we should also add a trigger to automatically set it https://github.com/ParabolInc/parabol/pull/6493

packages/server/postgres/queries/generated/getLatestIntegrationSearchQueriesWithProviderIdQuery.ts
```diff
@@ -0,0 +1,52 @@
...
+  service: IntegrationProviderServiceEnum | null | void;
+  providerId: number | null | void;
+}
+
+/** 'GetLatestIntegrationSearchQueriesWithProviderIdQuery' return type */
+export interface IGetLatestIntegrationSearchQueriesWithProviderIdQueryResult {
+  id: number;
+  userId: string;
+  teamId: string;
+  providerId: number | null;
```
Dschoordsch, line 21: -2 The type is correctly generated here, just pass `null` as value if you want to query for `"providerId" is null`

packages/server/postgres/queries/getLatestIntegrationSearchQueriesWithProviderId.ts
```diff
@@ -0,0 +1,17 @@
+import getPg from '../getPg'
+import {getLatestIntegrationSearchQueriesWithProviderIdQuery} from './generated/getLatestIntegrationSearchQueriesWithProviderIdQuery'
+import { IntegrationProviderServiceEnum } from "./generated/getIntegrationProvidersByIdsQuery";
+
+
+interface Params {
+  teamId: string
+  userId: string
+  providerId: number
```
Dschoordsch, line 9: -2 change the type to `providerId: number | null` and it should work.

Dschoordsch (changes requested):
Only looked at the DB part yet. I don't see why we need two separate kinds of queries and indexes. Even with allowing `null` for `"providerId"` that's just another value.

packages/server/postgres/migrations/1652861495447_addIntegrationSearchQuery.ts
```diff
@@ -0,0 +1,44 @@
...
+        ON DELETE CASCADE
+    );
+
+    CREATE UNIQUE INDEX integration_search_query_unique_not_null 
+    ON "IntegrationSearchQuery" ("userId", "teamId", "service", "query", "providerId")
+    WHERE "providerId" IS NOT NULL;
+    
+    CREATE UNIQUE INDEX integration_search_query_unique_null 
+    ON "IntegrationSearchQuery" ("userId", "teamId", "service", "query")
+    WHERE "providerId" IS NULL;
```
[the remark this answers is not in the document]
igorlesnenko: @Dschoordsch for example, for GitHub we have `providerId` which is always `null`.  

In this case, unique index that includes this nullable column will not work as expected. 

In postgres, It is possible to insert multiple identical rows if one column has `null` value. 

> PostgreSQL treats NULL as distinct value, therefore, you can have multiple NULL values in a column with a UNIQUE index.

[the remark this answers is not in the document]
Dschoordsch: Eventually providerId should be defined for all integrations.

Ok, so we need to have 2 indexes, but do we need two sets of queries? I think we don't. Same for upsert.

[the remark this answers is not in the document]
igorlesnenko: @Dschoordsch Changed to a single query. However, under the hood, it still uses 2 slightly different SQL queries.

Get query generated as follows:
```
AND "providerId" = :providerId
```

I can't just submit `null` here, as the query would be invalid. So I need a separate SQL file that explicitly uses `IS NULL`

For upsert queries, `ON CONFLICT` conditions is also a bit different:
```
ON CONFLICT ("userId", "teamId", "query", "service", "providerId") WHERE "providerId" IS NOT NULL DO
```

```
) ON CONFLICT ("userId", "teamId", "query", "service") WHERE "providerId" IS NULL DO
````

**Comment by Dschoordsch** [comment]
Let's get a reviewer review in first, @nickoferrall could you please take a look?

File: packages/client/mutations/PersistJiraServerSearchQueryMutation.ts [hunk 8]
```diff
@@ -0,0 +1,57 @@
...
+`
+
+const PersistJiraServerSearchQueryMutation: SimpleMutation<
+  TPersistJiraServerSearchQueryMutation
+> = (atmosphere, variables) => {
+  return commitMutation<TPersistJiraServerSearchQueryMutation>(atmosphere, {
+    mutation,
+    variables
+    // TODO: Add optimistic updater
+    // optimisticUpdater: (store) => {
```
nickoferrall, line 52: +1 can remove?

packages/server/graphql/public/mutations/persistIntegrationSearchQuery.ts
```diff
@@ -0,0 +1,68 @@
...
+      break
+    // case "jira":
+    //   query = jiraSearchQuery!
+    // break;
+    default:
+      throw new Error('Service not implemented')
+  }
+
+  if (providerId) {
+    const integrationProviders = await getIntegrationProvidersByIds([providerId])
```
nickoferrall, line 35: +1 could save a line and do `const integrationProvider = await getIntegrationProvidersByIds([providerId])[0]`

packages/server/graphql/public/mutations/persistIntegrationSearchQuery.ts
```diff
@@ -0,0 +1,68 @@
...
+  const viewerId = getUserId(authToken)
+  const operationId = dataLoader.share()
+  const subOptions = {mutatorId, operationId}
+
+  //AUTH
+  if (!isTeamMember(authToken, teamId)) {
+    return {error: {message: `Not on team`}}
+  }
+
+  let query
```
nickoferrall, line 22: -1 we should add the type for `query`

packages/server/graphql/types/JiraServerIntegration.ts
```diff
@@ -33,6 +35,17 @@ const JiraServerIntegration = new GraphQLObjectType<{teamId: string; userId: str
...
   fields: () => ({
+    id: {
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'Composite key in jiraServer:providerId format',
+      resolve: async ({teamId, userId}: {teamId: string; userId: string}, _args, {dataLoader}) => {
+        const auth = await dataLoader
+          .get('teamMemberIntegrationAuths')
+          .load({service: 'jiraServer', teamId, userId})
+
+        return `jiraServer:${auth?.providerId}`
```
nickoferrall, line 46: If `auth` is undefined, we'll return `jiraServer:undefined` here. Could we check if `auth` exists, and return null if it doesn't?

nickoferrall: Also where is this `id` currently being used?

nickoferrall (changes requested):
This looks great! I've tested it and everything is working. One tiny issue is when removing a saved query, it currently takes the user to that query. Instead, I think it should just remove the saved query and not alter the current search. 

removing-saved-query

packages/server/graphql/types/JiraServerIntegration.ts
igorlesnenko: @nickoferrall it is needed for relay to correctly update the store. We request `JiraServerintegration.searchQueries` after performing persist search query mutation.

packages/server/graphql/public/mutations/persistIntegrationSearchQuery.ts
igorlesnenko: Added in `upsertIntegrationSearchQuery`

packages/server/graphql/types/JiraServerIntegration.ts
```diff
@@ -33,6 +35,21 @@ const JiraServerIntegration = new GraphQLObjectType<{teamId: string; userId: str
...
   fields: () => ({
+    id: {
+      type: new GraphQLNonNull(GraphQLID),
+      description: 'Composite key in jiraServer:providerId format',
+      resolve: async ({teamId, userId}: {teamId: string; userId: string}, _args, {dataLoader}) => {
+        const auth = await dataLoader
+          .get('teamMemberIntegrationAuths')
+          .load({service: 'jiraServer', teamId, userId})
+
+        if (!auth) {
```
nickoferrall, line 46: +1 I wonder whether the `id` type should be nullable and we return null if there's no `auth`? Otherwise, an empty string might be difficult to debug

packages/client/components/ScopingSearchHistoryMenu.tsx
```diff
@@ -64,9 +64,10 @@ const ScopingSearchHistoryMenu = (props: Props) => {
       )}
       {searchQueries.map(({id, labelFirstLine, labelSecondLine, onClick, onDelete}) => {
         const handleDelete = (event: MouseEvent) => {
-          if (searchQueries.length > 1) {
-            event.stopPropagation() // prevents closing the menu when remove button is clicked
+          if (searchQueries.length === 1) {
+            closePortal()
```
nickoferrall, line 68: +2 nice fix!

nickoferrall (approved):

packages/server/postgres/queries/removeIntegrationSearchQueryToPG.ts
```diff
@@ -0,0 +1,7 @@
+import getPg from '../getPg'
+import {removeIntegrationSearchQueryQuery} from './generated/removeIntegrationSearchQueryQuery'
+
+const removeIntegrationSearchQueryToPG = async (id: number, userId: string, teamId: string) => {
```
Dschoordsch, line 4: -1 That name is inconsistent with otheres, can we just name it `removeIntegrationSearchQuery` please?

packages/server/graphql/types/JiraServerIntegration.ts
```diff
@@ -33,6 +35,21 @@ const JiraServerIntegration = new GraphQLObjectType<{teamId: string; userId: str
...
+          return ''
```
Dschoordsch, line 47: Why did you add the `id` field?

packages/server/graphql/types/JiraServerIntegration.ts
```diff
@@ -162,6 +179,53 @@ const JiraServerIntegration = new GraphQLObjectType<{teamId: string; userId: str
...
       }
+    },
+    providerId: {
+      type: new GraphQLNonNull(GraphQLInt),
+      resolve: async ({teamId, userId}, _args: unknown, {dataLoader}) => {
+        const auth = await dataLoader
+          .get('teamMemberIntegrationAuths')
+          .load({service: 'jiraServer', teamId, userId})
+
+        return auth?.providerId
```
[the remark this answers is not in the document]
Dschoordsch: -1 I agree, should be nullable

Dschoordsch, line 190: -1 The ID should be `ProviderId.join(auth.providerId)`

packages/server/graphql/types/JiraServerIntegration.ts
```diff
@@ -33,6 +35,21 @@ const JiraServerIntegration = new GraphQLObjectType<{teamId: string; userId: str
...
+      resolve: async ({teamId, userId}: {teamId: string; userId: string}, _args, {dataLoader}) => {
+        const auth = await dataLoader
+          .get('teamMemberIntegrationAuths')
+          .load({service: 'jiraServer', teamId, userId})
+
+        if (!auth) {
+          return ''
+        }
+
+        return `jiraServer:${auth.providerId}`
```
Dschoordsch, line 50: -2 The id is not unique, a user can have the same provider for multiple teams, each team has a different auth. Thus you need to use `jiraServer:${auth}.id` here.

The main issue here is that we're mixing the providers (which are per team and should be non-null) with the auth (which is per team member and should be nullable). So what we really should do is:
* make `JiraServerIntegration` nullable (also GitLab, AzureDevOps, MSTeams, ...)
* move `IntegrationProvider` fields to Team object.

No need to do it in this change, but we should address that.

packages/server/graphql/types/JiraServerIntegration.ts
```diff
@@ -162,6 +179,53 @@ const JiraServerIntegration = new GraphQLObjectType<{teamId: string; userId: str
...
+
+        return searchQueries.map((searchQuery) => {
+          const query = searchQuery.query as {
+            queryString: string | null
+            isJQL: boolean
+            projectKeyFilters: string[] | null
+          }
+
+          return {
+            id: searchQuery.id,
```
Dschoordsch, line 221: -1 This will not be unique, so we need to prefix it. There is already `SearchQueryId.join` but that's Meeting specific.

packages/server/graphql/public/types/RemoveIntegrationSearchQuerySuccess.ts
```diff
@@ -0,0 +1,18 @@
...
+      teamId: string
+    }
+  | {error: {message: string}}
+
+const RemoveIntegrationSearchQuerySuccess: {
+  jiraServerIntegration: (
+    source: RemoveIntegrationSearchQuerySuccessSource
+  ) => RemoveIntegrationSearchQuerySuccessSource
+} = {
+  jiraServerIntegration: (source) => {
```
Dschoordsch, line 13: +1 did you test this resolver is actually necessary? Also you would usually import the generated types here
```
const RemoveIntegrationSearchQuerySuccess: RemoveIntegrationSearchQuerySuccessResolvers = {
```
You might need to define the `JiraServerIntegration` source properly though.

Dschoordsch (changes requested):

packages/server/postgres/queries/src/removeIntegrationSearchQueryQuery.sql
```diff
@@ -0,0 +1,5 @@
+/*
+  @name removeIntegrationSearchQueryQuery
+*/
+DELETE FROM "IntegrationSearchQuery"
+WHERE "id" = :id AND "userId" = :userId AND "teamId" = :teamId;
```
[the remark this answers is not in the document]
igorlesnenko: @Dschoordsch I added this as a validation, to not fetch `IntegrationQuery` first and manually check `userId` and `teamId`.

packages/client/mutations/PersistJiraServerSearchQueryMutation.ts
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+`
+
+const mutation = graphql`
+  mutation PersistJiraServerSearchQueryMutation(
+    $teamId: ID!
+    $service: IntegrationProviderServiceEnum!
+    $providerId: Int
+    $jiraServerSearchQuery: JiraServerSearchQueryInput
```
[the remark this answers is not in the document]
igorlesnenko: `oneof` feels cleaner, as we do not need to pass `service`, however I think it is ok as is for now

packages/server/graphql/types/JiraServerIntegration.ts
igorlesnenko: @Dschoordsch it is needed for relay to correctly update the store. We request `JiraServerIntegration.searchQueries` in persist search query mutation. So relay updates `searchQueries` automatically

packages/server/graphql/types/JiraServerIntegration.ts
igorlesnenko: @Dschoordsch it is unique as the `id` comes from postgresql table

packages/server/postgres/queries/src/removeIntegrationSearchQueryQuery.sql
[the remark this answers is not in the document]
igorlesnenko: Added a comment

packages/server/graphql/types/JiraServerIntegration.ts
igorlesnenko: @Dschoordsch pushed a commit that uses `toGlobalId` and `fromGlobalId` from 'graphql-relay' package. I think this way we can avoid creating a new function every time. What do you think?

igorlesnenko: The issue with the proposal #6642

packages/server/graphql/types/JiraServerIntegration.ts
igorlesnenko: @Dschoordsch 
> a user can have the same provider for multiple teams
So in future provider will be added to the whole organisation?
Currently, I see that we have `teamId` column in the `IntegrationProvider` table.
Will we have `IntegrationProviderTeam` table with `providerId | teamId` columns?

igorlesnenko: `auth` object does not have `id`. Made it like this: `return `jiraServer:${teamId}:${auth.providerId}``

Dschoordsch: That's the current state, each "IntegrationProvider" has scope "org" or "team". Right now the organization is derived from the team, but we may store it separately #6014 at some point

packages/server/graphql/public/types/RemoveIntegrationSearchQuerySuccess.ts
igorlesnenko: @Dschoordsch Does not work for me without this resolver, `jiraServerIntegration` returned as `null.
Picking generated type `RemoveIntegrationSearchQuerySuccessResolvers` does not work in this case as I need to just forward plain `source` here, not actually resolve.

Dschoordsch (approved):

Issue #6552, opened 2022-05-12
Remove GitLab feature flag


nickoferrall:
### AC

- [ ] Remove the GitLab feature flag so that it can be used by all

Issue #6540, opened 2022-05-10
Fix GitLab query complexity bug


nickoferrall:
To reproduce:

- Select 50 issues in the GitLab Poker Scope phase
- Refresh the page a few times
- See that the sidebar doesn't show the titles of the issues and the console says that the complexity of the query exceeds the limit of 250

Originally discovered by Matt here: https://github.com/ParabolInc/parabol/issues/6511#issue-1227046122

### AC

- [ ] Can add 50 GitLab issues, refresh the page, and still see all the issues in the sidebar

igorlesnenko:
@nickoferrall is it same issue here #6386 ?

nickoferrall:
@igorlesnenko it is! I didn't see that one. I'll close it now

Issue #6511, opened 2022-05-05
feat(gitlab): Burndown to v1


mattkrick:
1. placeholder should be smaller (Search gitlab? search gitlab projects? gitlab projects?)
<img width="229" alt="Screen Shot 2022-05-05 at 11 32 40 AM" src="https://user-images.githubusercontent.com/5514175/166995253-0866e4c2-a36f-43af-848a-91bbcc718f0b.png">

2. Filters aren't updating
<img width="736" alt="Screen Shot 2022-05-05 at 11 35 50 AM" src="https://user-images.githubusercontent.com/5514175/166997166-740365ff-4645-4b7c-99fc-e13c81cf4ffe.png">

3. clicking select all is broken 
<img width="676" alt="Screen Shot 2022-05-05 at 11 36 29 AM" src="https://user-images.githubusercontent.com/5514175/166997297-ddc5a2be-ff9d-48f1-a25d-6da1c28b756b.png">

4. searching by issue number is broken
<img width="753" alt="Screen Shot 2022-05-05 at 11 37 50 AM" src="https://user-images.githubusercontent.com/5514175/166997854-0db4f99d-36b3-4186-8801-65fe63e92616.png">

5. query complexity was too high & it tells me i'm not on the team
<img width="370" alt="Screen Shot 2022-05-05 at 11 44 14 AM" src="https://user-images.githubusercontent.com/5514175/167001365-4504db16-3de3-4821-9c3b-19d9df4a7ba6.png">

<summary>see
<details>

```
11:40:36 1|GraphQL Executor  | SEND TO SENTRY Error: Query has complexity of 480, which exceeds max complexity of 250
11:40:36 1|GraphQL Executor  |     at /Users/mattkrick/code/parabol/dev/gqlExecutor.js:57348:30
11:40:36 1|GraphQL Executor  |     at processTicksAndRejections (node:internal/process/task_queues:96:5)
11:40:36 1|GraphQL Executor  |     at async Object.resolve [as TaskintegrationResolver] (/Users/mattkrick/code/parabol/dev/gqlExecutor.js:50941:33) undefined
11:40:37 1|GraphQL Executor  | TypeError: Cannot destructure property 'issue' of 'issueData' as it is undefined.
11:40:37 1|GraphQL Executor  |     at Object.resolve [as EstimateStageserviceFieldResolver] (/Users/mattkrick/code/parabol/dev/gqlExecutor.js:39655:13)
[...]

PR #6554, opened 2022-05-12
chore: remove GitLab feature flag

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6552

### To test:

- Open up an incognito window
- Create a new user
- Create a Sprint Poker meeting
- See that you can integrate with GitLab from within the meeting & everything works as expected such as pushing estimates to the issue
- In `/tasks`, see that you can push a task to a GitLab project

This PR should only be merged to master once https://github.com/ParabolInc/parabol/issues/6511 and https://github.com/ParabolInc/parabol/issues/6540 have been closed.

### DIFF
--- a/packages/client/.eslintignore
+++ b/packages/client/.eslintignore
@@ -1,3 +1,2 @@
 __generated__/*
 serviceWorker/*
-AddFeatureFlagMutation.ts
--- a/packages/client/.eslintignore
+++ b/packages/client/.eslintignore
@@ -1,2 +1,3 @@
 __generated__/*
 serviceWorker/*
+AddFeatureFlagMutation.ts
--- a/packages/client/components/AddToGitLabMenuItem.tsx
+++ b/packages/client/components/AddToGitLabMenuItem.tsx
@@ -43,7 +43,6 @@ const AddToGitLabMenuItem = forwardRef((props: Props, ref) => {
   )
   const atmosphere = useAtmosphere()
   const {cloudProvider} = gitlab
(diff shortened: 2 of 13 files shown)


packages/client/.eslintignore
```diff
@@ -1,2 +1,3 @@
 __generated__/*
 serviceWorker/*
+AddFeatureFlagMutation.ts
```
nickoferrall, line 3: AddFeatureFlagMutation was being used for the Spotlight feature. We wanted users to be able to enter a cheat code to add themselves to the feature. 

This is no longer in use but will be a helpful feature in the future. To prevent the `Parsing error: "parserOptions.project" has been set for @typescript-eslint/parser` eslint complaint, I've added it to the eslintignore file here.

Dschoordsch: -1 ignoring the file completely is not good since it will not be linted. We should rather add this explicitly to the project so it's linted correctly. You should add the file here https://github.com/ParabolInc/parabol/blob/066796ca63a54773fab9363e7f5d28eee8d0202c/packages/client/tsconfig.eslint.json instead

Dschoordsch (changes requested):

packages/client/.eslintignore
nickoferrall: Makes sense, thanks. Updated!

Dschoordsch (approved):

Dschoordsch:
Did not test.

Issue #6393, opened 2022-04-12
Standup UI - Last updated time


jmtaber129:
AC:
- [ ] Response card shows when someone last updated their standup response

PR #6557, opened 2022-05-13
feat(standups): Standup UI - last updated time #6393

JimmyLv:
implements #6393

based on #6469 (need to merge firstly)

Google Chrome-Async Standup #2  Jimmy Lv’s Team-002274-20220513


packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
```diff
@@ -123,15 +133,14 @@ const TeamPromptResponseCard = (props: Props) => {
   const {teamMember, meetingId, meeting, discussion, response} = responseStage
   const {picture, preferredName, userId} = teamMember
 
+  const discussionEdges = discussion.thread.edges
+  const replyCount = discussionEdges.length
+
```
JimmyLv, line 138: just moved the position because of a merge conflict. Can be ignored

JimmyLv:
@jmtaber129 Hi James, could you please help to review this PR? 😁

packages/client/components/TeamPrompt/TeamPromptLastUpdatedTime.tsx
```diff
@@ -0,0 +1,26 @@
...
+  updatedAt: string | Date
+}
+
+export default function LastUpdatedTime({updatedAt, createdAt}: Props) {
+  const isEdited = createdAt !== updatedAt
+  const formattedTime = updatedAt ? relativeDate(updatedAt) : relativeDate(createdAt)
+  return (
+    <Timestamp>
+      <span>{formattedTime}</span>
+      {isEdited && <span> · Edited</span>}
```
jmtaber129, line 23: +1 No need for `span` + simplify to one block:
```suggestion
      {formattedTime}{isEdited && ' · Edited'}
```

packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
```diff
@@ -81,11 +82,20 @@ const TeamPromptResponseCard = (props: Props) => {
           picture
           preferredName
         }
+        discussion {
+          thread(first: 1000) @connection(key: "DiscussionThread_thread") {
+            edges {
+              ...TeamPromptRepliesAvatarList_edges
+            }
+          }
+        }
```
jmtaber129, line 91: +1 duplicates the below `discussion`

packages/client/mutations/UpsertTeamPromptResponseMutation.ts
```diff
@@ -13,6 +13,7 @@ graphql`
       userId
       content
       plaintextContent
+      updatedAt
```
jmtaber129, line 16: -1 this needs `createdAt` for new responses

packages/client/mutations/UpsertTeamPromptResponseMutation.ts
```diff
@@ -77,7 +79,8 @@ const UpsertTeamPromptResponseMutation: StandardMutation<
         id: teamPromptResponseId ?? clientTempId(viewerId),
         userId: viewerId,
         content,
-        plaintextContent
+        plaintextContent,
+        updatedAt: now
```
jmtaber129, line 83: -1 this needs `createdAt` for new responses

packages/client/components/TeamPrompt/TeamPromptLastUpdatedTime.tsx
```diff
@@ -0,0 +1,26 @@
...
+})
+
+interface Props {
+  createdAt: string | Date
+  updatedAt: string | Date
+}
+
+export default function LastUpdatedTime({updatedAt, createdAt}: Props) {
+  const isEdited = createdAt !== updatedAt
+  const formattedTime = updatedAt ? relativeDate(updatedAt) : relativeDate(createdAt)
```
jmtaber129, line 19: +1 `updatedAt` should always be non-null, so I think this can be simplified a bit

jmtaber129 (changes requested):
Looking good!

A few high-level comments:
1. -1 After editing, the timestamp flashes to the optimistic update and then immediately back to the old timestamp.  This seems to be because we're not clearing the cache in the upsert mutation (this hasn't been an issue since we're manually setting the content, but that seems like a smell).  Applying this patch should fix this, and prevent similar issues from popping up in the future:
```
diff --git a/packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts b/packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
index 4473e5505..3842bb612 100644
--- a/packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
+++ b/packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
@@ -67,11 +67,10 @@ const upsertTeamPromptResponse: MutationResolvers['upsertTeamPromptResponse'] =
     content,
     plaintextContent
   })
-  const updatedTeamPromptResponse = await dataLoader
+
+  dataLoader
     .get('teamPromptResponses')
-    .load(teamPromptResponseId)
-  updatedTeamPromptResponse.content = contentJSON
-  updatedTeamPromptResponse.plaintextContent = plaintextContent
+    .clear(teamPromptResponseId)

   const data = {
     meetingId,
```
2. -1 Upon creating a new response, the timestamp doesn't update to reflect the `createdAt` time until after refresh (see inline comments for where `createdAt` is missing)
3. -1 A pattern I've seen in other apps like Slack is to use the original message timestamp (`createdAt` in this case) for the primary timestamp, and then surface the edit time through a tooltip over `Edited`.  I think this is the way to go here, but I'd get some design input from @ackernaut 
4. +1 The timestamp introduces some inconsistency with spacing + card size between the header and the actual card for empty cards vs. responded cards (see screenshot) - a card with the timestamp actually more closely matches the [designs](https://www.figma.com/file/bhiEOV4HK9Kv91HFsJmevV/Stand-up-Concepts?node-id=1147%3A73406), but I think we need to enforce a 12px bottom margin on the avatar component to ensure consistency
Screen Shot 2022-05-16 at 1 48 49 PM
5. +1 A tooltip over the relative timestamp showing the actual timestamp might be nice.  @ackernaut what do you think?

ackernaut:
@jmtaber129 
- So simply show ‘Edited’ and when you hover there’s a tooltip to show the timestamp for when it was edited? That sounds straightforward enough to me 👍 
- When the relative timestamp says something like ‘an hour ago’ you can hover that and the timestamp gives you something like ‘Mon May 16 14:34’? I like it 👍

jmtaber129:
@ackernaut yep!  Just to clarify, overall what this would look like is we'd have the format `<createdAt><' * Edited' if edited>`, where hovering over `createdAt` would show a timestamp like `Mon May 16 14:34`, and hovering over `Edited` would show the `updatedAt` timestamp

ackernaut:
@jmtaber129 LGTM!

JimmyLv:
wow, that's so nice, thanks @jmtaber129 for reviewing timely, let me update!

JimmyLv:
@jmtaber129 Hey James, I've fixed all your feedback and updated the tooltip interaction behaviors, can you help review them again? Thanks!

Google Chrome-Async Standup #3  Jimmy Lv’s Team-002439-20220520

JimmyLv:
@jmtaber129 Hey James, I've fixed all your feedback and updated the tooltip interaction behaviors, can you help review them again? Thanks!

Google Chrome-Async Standup #3  Jimmy Lv’s Team-002439-20220520

packages/client/mutations/UpsertTeamPromptResponseMutation.ts
```diff
@@ -77,7 +80,9 @@ const UpsertTeamPromptResponseMutation: StandardMutation<
         id: teamPromptResponseId ?? clientTempId(viewerId),
         userId: viewerId,
         content,
-        plaintextContent
+        plaintextContent,
+        updatedAt: now,
+        createdAt: now
```
jmtaber129, line 85: +1 This causes the createdAt time to flash from `X minutes ago` to `just now` and back.  I think we should only provide this if a response doesn't already exist (i.e. when `teamPromptResponseId` is falsey).  That way, we optimistically provide a `createdAt` of `now` for new responses, and let relay use the value already in the store for updated responses.

```suggestion
        createdAt: !teamPromptResponseId ? now : undefined
```

jmtaber129 (approved):
Looking good!  A few minor issues (two inline comment + the below comment)

+1 `createdAt` still doesn't show up until after refresh, but I found the fix (unrelated to this PR, there's an updater that we need for newly inserted responses - new responses don't get upserted into the local store because the ID doesn't already exist, so filling out an initial response, navigating away from the meeting, and then navigating back to the meeting doesn't show the new response).  Try applying this patch and it should work:
```
diff --git a/packages/client/mutations/UpsertTeamPromptResponseMutation.ts b/packages/client/mutations/UpsertTeamPromptResponseMutation.ts
index f13ea9d9c..2d0b09e33 100644
--- a/packages/client/mutations/UpsertTeamPromptResponseMutation.ts
+++ b/packages/client/mutations/UpsertTeamPromptResponseMutation.ts
@@ -91,6 +91,10 @@ const UpsertTeamPromptResponseMutation: StandardMutation<
     mutation,
     variables,
     optimisticResponse,
+    updater:  (store) => {
+      const payload = store.getRootField('upsertTeamPromptResponse')
+      upsertTeamPromptResponseUpdater(payload as any, {atmosphere, store})
+    },
     onCompleted,
     onError
   })
```

cc @Dschoordsch @BartoszJarocki since this was removed as a result of review of https://github.com/ParabolInc/parabol/pull/6504.  Definitely a weird "gotcha".

packages/client/mutations/UpsertTeamPromptResponseMutation.ts
JimmyLv: Yes, that's the case!

JimmyLv:
@Dschoordsch Hey Georg, could you please help to do the maintainer review for this PR? I'd like to ensure this one can be merged into the next release, thanks! 😁

Dschoordsch:
@JimmyLv please add me to the reviewers once I should do a review, otherwise I might miss it.

packages/client/components/promptResponse/PromptResponseEditor.tsx
```diff
@@ -56,7 +57,11 @@ const PromptResponseEditor = (props: Props) => {
 
   const onSubmit = ({editor: newEditorState}: EditorEvents['blur']) => {
     setEditing(false)
-    handleSubmit && handleSubmit(newEditorState)
+    const newContent = newEditorState.getJSON()
+
+    if (areEqual(content, newContent)) return
```
Dschoordsch, line 62: -1 I think you need to check here for empty and undefined/null to avoid creating an empty post on first blur.

Dschoordsch (changes requested):
It feels weird that the first blur of the input field creates a response, even when it is empty.
Updated at is set when loosing focus

I also would like to have the times updated, but that could happen separately. If you don't address this, make sure to create a ticket for it.

packages/client/components/TeamPrompt/TeamPromptLastUpdatedTime.tsx
```diff
@@ -0,0 +1,52 @@
...
+    tooltipPortal: updatedTimePortal,
+    openTooltip: showUpdatedTime,
+    closeTooltip: closeUpdatedTime,
+    originRef: updatedTimeRef
+  } = useTooltip<HTMLButtonElement>(MenuPosition.UPPER_CENTER)
+
+  const isEdited = createdAt !== updatedAt
+  return (
+    <Timestamp>
+      <Hover onMouseEnter={showCreatedTime} onMouseLeave={closeCreatedTime} ref={createdTimeRef}>
```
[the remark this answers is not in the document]
JimmyLv: I see what you mean, it's because the hover triggers a re-render of the component, so there's a jump from "just now" to "3 minutes ago", but do you expect this time to keep updating automatically? That might require a poll, is that too heavy or do you have other suggestions for it?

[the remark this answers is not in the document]
Dschoordsch: We don't need to poll the server, we would just need to reformat the time. Ideally this would be done outside of React component tree by updating the ref via script directly, i.e. add a setInterval which takes the ref of the text component and updates it without doing a full re-render. @nickoferrall I think you have more React experience, what do you think?

packages/client/components/promptResponse/PromptResponseEditor.tsx
JimmyLv: done! good point, and we can use `newEditorState.isEmpty` to check the empty input content

JimmyLv:
Hey @Dschoordsch , I updated the first empty check to avoid creating an empty post on the first blur. Could you please help to review again, thanks!

https://user-images.githubusercontent.com/4997466/170245563-5e4453f5-2a68-4aae-928f-2cc093b3f1ca.mp4

packages/client/components/TeamPrompt/TeamPromptLastUpdatedTime.tsx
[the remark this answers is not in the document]
nickoferrall: Yep, I'd add a setInterval to a useEffect and update the timestamp that way. Interestingly, in our Discussion Thread, I noticed we're not updating the timestamp until the component rerenders, so if you implement a pattern in standups, it could be useful elsewhere in the app.

Dschoordsch (approved):

PR #6559, opened 2022-05-13
feat(standups): Place viewer's card at the start of the list

jmtaber129:
Test:
- [ ] Viewer's card is always at the start of the card list
- [ ] When a new user joins a standup, their card is added to the start of the list


BartoszJarocki (approved):
@jmtaber129 if you don't mind, I'm going to use `One Review Required` label here. PR is small and safe enough. lgtm! 

edit: more of a question, do we want to add a new user's card to the start of the list (after the viewers card) or to the end of the list? adding it to the end would prevent shifting all other cards and could look better, just an idea

PR #6563, opened 2022-05-17
refactor: gitlab issues query part 1

nickoferrall:
Part 1 of 2 to fix https://github.com/ParabolInc/parabol/issues/6325

Previously, we were querying the GitLab projects and getting the issues within those projects. The problem with this is Relay doesn't support nested connections, so when scrolling down, we could only request more projects, when really we want more issues.

In the issue AC, we said we'd initially use the projects API and then, if a user selects projects filters, we would use the project API, so we can add a connection on issues when the user scrolls down. 

In this PR, I've implemented a field called `projectIssues` (renamed to `projectsIssues` in a later PR), which returns the issues that belong to a project. If no project filters are selected, it returns the first 25 issues from the first 50 projects. 

With this implementation, we can just use the `projectIssues` query rather than two separate queries, simplifying things on the frontend. 

In this PR, we're just returning the issues that belong to a project. In a subsequent PR (https://github.com/ParabolInc/parabol/pull/6604), we'll implement the pagination and ability to add a new issue to the connection.

### To test

- Integrate with GitLab and create a Sprint Poker meeting
- Go to the Scope phase and see GitLab issues rendered in the list
- Select project filters from the dropdown menu, and see that the issues displayed are filtered by project


packages/server/integrations/gitlab/GitLabServerManager.ts
```diff
@@ -142,10 +145,31 @@ class GitLabServerManager implements TaskIntegrationManager {
     return [noteData, noteError] as const
   }
 
-  async getProjects({teamId}: {teamId: string}) {
+  async getProjectIssues({
+    fullPath,
+    first,
+    includeSubepics,
+    searchQuery,
+    sort,
```
nickoferrall, line 153: In the next PR, these are passed in as one object rather than separate variables

igorlesnenko (approved):

Dschoordsch (approved):
Can be fixed in follow up PR

PR #6587, opened 2022-05-19
feat(standups): Sending meeting completed event when standup ends

### LINKED ISSUE #6535: Standups: adds Meeting Completed metric
Pulling the first item from #6385 

### Acceptance criteria
- Sends an event when a stand-up meeting is completed

### Estimated effort
- 2 hours

BartoszJarocki:
# Description

Fixes #6535 
This PR introduces a new pattern on how we could do analytics. There's a loom below describing why I think it's a good idea.

## Demo
https://www.loom.com/share/10621d2a9e7444719f55511f2dd9aa0d 

## Testing scenarios
- [ ] Ending the standups sends the correct event to segment

## Final checklist

- [x] I checked the [code review guidelines](../../docs/codeReview.md)


packages/server/utils/analytics/analytics.ts
```diff
@@ -0,0 +1,102 @@
...
+    })
+  }
+
+  checkInEnd = (completedMeeting: Meeting, meetingMembers: MeetingMember[]) => {
+    meetingMembers.forEach((meetingMember) =>
+      this.meetingEnd(meetingMember.userId, completedMeeting, meetingMembers)
+    )
+  }
+
+  retrospectiveEnd = (
```
BartoszJarocki, line 43: I haven't wired up these methods yet, I just wanted to showcase how it'd look like

packages/server/utils/analytics/analytics.ts
```diff
@@ -0,0 +1,101 @@
...
+    responses: TeamPromptResponse[]
+  ) => {
+    const userIdsResponses: Record<string, string> = responses.reduce(
+      (previous, response) => ({...previous, [response.userId]: response.plaintextContent}),
+      {}
+    )
+    meetingMembers.forEach((meetingMember) => {
+      const plaintextResponseContent = userIdsResponses[meetingMember.userId]
+      this.meetingEnd(meetingMember.userId, completedMeeting, meetingMembers, {
+        responseAdded: plaintextResponseContent && plaintextResponseContent.length > 0
```
tianrunhe, line 33: -1 This will not always resolve to a boolean value, e.g., when `plaintextResponseContent` is empty

packages/server/utils/analytics/analytics.ts
```diff
@@ -0,0 +1,101 @@
...
+    const presentMemberUserIds = meetingMembers.map(({userId}) => userId)
+    const wasFacilitator = userId === facilitatorUserId
+    this.track(userId, 'Meeting Completed', {
+      hasIcebreaker: phases[0]?.phaseType === 'checkin',
+      // include wasFacilitator as a flag to handle 1 per meeting
+      wasFacilitator,
+      userIds: wasFacilitator ? presentMemberUserIds : undefined,
+      meetingType,
+      meetingNumber,
+      teamMembersCount: meetingMembers.length,
```
tianrunhe, line 90: +1 Can you just add a `//TODO: fix me` so we don't forget this

tianrunhe (changes requested):
I love this refactor! Absolutely beautiful! Just one minor issue:

packages/server/utils/analytics/analytics.ts
BartoszJarocki: oh, good catch! fixed

packages/server/utils/analytics/analytics.ts
BartoszJarocki: done

BartoszJarocki:
@tianrunhe thank you! I fixed your notes so it'd if you could double-check it. @Dschoordsch do you mind taking a look?

packages/server/utils/analytics/analytics.ts
```diff
@@ -0,0 +1,101 @@
...
+        createMeetingTemplateAnalyticsParams(template)
+      )
+    )
+  }
+
+  private meetingEnd = (
+    userId: string,
+    completedMeeting: Meeting,
+    meetingMembers: MeetingMember[],
+    meetingSpecificProperties?: any
```
Dschoordsch, line 78: +1 Could be `Record<string,string | number>`, we probably don't want to do any nesting here.

packages/server/utils/analytics/analytics.ts
Dschoordsch: +1 Would be nice if you could wire them up already as it's mostly duplicate code right now.

packages/server/utils/analytics/analytics.ts
```diff
@@ -0,0 +1,101 @@
...
+      meetingType,
+      meetingNumber,
+      teamMembersCount: meetingMembers.length, // TODO: use team members count
+      teamMembersPresentCount: meetingMembers.length,
+      teamId,
+      ...meetingSpecificProperties
+    })
+  }
+
+  private track = (userId: string, event: AnalyticsEvent, properties?: any) =>
```
Dschoordsch, line 97: +1 My 2 cents: not having a generic method for tracking one of a type events will just add more boilerplate. I think for meetings it is fine to do the formatting here, but if I look at inviteToTeam, createTaskIntegration etc. these are easier to handle locally. I'm fine defining the event type here though so there is some place to keep an overview.

Dschoordsch (approved):
I have some reservations about replicating each method as an event in the `Analytics` class, but safe to try.

packages/server/utils/analytics/analytics.ts
BartoszJarocki: yes, I understand the tension. it's a bit of a personal preference, I prefer more readable code even with the cost of more boilerplate. we have a choice of either using `analytics.track(userId, 'Invite', { teamId, invitees: successfulInvitees})` or `analytics.inviteEmailSent(userId, invites)`. The reason I prefer the second is that I think it's a bit better to delegate the formatting of the parameters to the class that's actually responsible for successfully handling that. the less we have to change the resolvers when dealing with their side effects, the better IMO. I'm open to making that public though, it'd still be an improvement over what we do now.

packages/server/utils/analytics/analytics.ts
BartoszJarocki: we also send booleans and enums here so I figured it's ok to keep it that way. in different cases we also send arrays, dates etc

packages/server/utils/analytics/analytics.ts
BartoszJarocki: done

Dschoordsch (approved):

Issue #6592, opened 2022-05-19
Refresh GitLab Auth Tokens


nickoferrall:
In GitLab's May 2022 update, they have deprecated non-expiring tokens and are now expiring tokens after two hours: https://northflank.com/blog/supporting-expiring-oauth-access-tokens-for-gitlab

### AC

- [ ] Refresh GitLab tokens so that users can still access their GitLab integration two hours after integrating

nickoferrall:
Current plan:

- after integrating with GitLab, store the token expiration timestamp in Redis
- create a gitlabAuth dataloader that uses our `teamMemberIntegrationAuths`
- in the dataLoader, get the token's TTL from Redis
- if we're almost at the limit, refresh the token 
- upsert the new access token in the `teamMemberIntegrationAuth` table

Dschoordsch:
> * in the dataLoader, get the token's TTL from Redis
Why would we store the TTL in Redis when the token itself is stored in Postgres? That seems out of place to me.

nickoferrall:
We could store it in Postgres, but I was thinking Redis because:

- Currently, the TTL is GitLab specific, so adding a migration for `teamMemberIntegrationAuths` might be overkill 
- With Redis, you can set the TTL on the key and then just check whether the key exists. In Postgres, we'd need to store the timestamp and compare it with the current timestamp. Not a big deal, but not as clean as Redis

nickoferrall:
@Dschoordsch curious to get your thoughts on the above. I created a PR for this yesterday evening: https://github.com/ParabolInc/parabol/pull/6594

If you feel that the TTL needs to live in Postgres, I'll work on the change today.

PR #6594, opened 2022-05-19
feat(gitlab): refresh gitlab tokens

nickoferrall:
Fixes https://github.com/ParabolInc/parabol/issues/6592

### Testing

- Integrate with GitLab
- In gitlabLoaders.ts, change `if (!isValidAuth) {` to `if (true) {` so that we run the refresh logic every time the dataloader is called
- Comment out the upsert team member integration line: `upsertTeamMemberIntegrationAuth(newGitlabAuth),`
- Now, we're refreshing the token, but we're not upserting & storing the new access token. When refreshing the token, we invalidate the previous accessToken, so we should no longer have access to GitLab.
- Refresh the page and create a new Poker meeting. See that the GitLab issues don't show up in the Scope phase

https://www.loom.com/share/19d6b24adda649eb9175347856b97170


igorlesnenko (changes requested):
I would suggest to store TTL in a `teamMemberIntegrationAuths` where token lives.

I don't see that redis gives us functionality that couldn't be implemented in postgres in this particular case.

If we store it in a PostgreSQL, we don't need any additional query, as we can just fetch datetime and check that in code.

As alternative, we can just use try catch, if gitlab returns an error we refresh the token.

What do you think?

packages/server/dataloader/gitlabLoaders.ts
```diff
@@ -0,0 +1,59 @@
...
+              sendToSentry(new Error('No refresh token in gitlabAuth'), {userId})
+              return null
+            }
+            const provider = await parent.get('integrationProviders').loadNonNull(providerId)
+            const {clientId, clientSecret, serverBaseUrl} = provider
+            const manager = new GitLabOAuth2Manager(clientId!, clientSecret!, serverBaseUrl!)
+            const oauthRes = await manager.refresh(refreshToken)
+            if (oauthRes instanceof Error) return null
+            const {accessToken, refreshToken: newRefreshToken, expires_in} = oauthRes
+            const expiresAtTimestamp = new Date().getTime() + (expires_in - 30) * 1000
```
nickoferrall, line 37: `expires_in is in seconds. Minus 30 to be safe. *1000 to convert it to milliseconds

nickoferrall:
I've pushed an update using the postgres table rather than Redis. Ready for your review @igorlesnenko. 

Fingers crossed it can be squeezed into the ship this week so we can launch GitLab on Wednesday 🤞

packages/server/dataloader/gitlabLoaders.ts
```diff
@@ -0,0 +1,59 @@
...
+            const newGitlabAuth = {
+              ...gitlabAuth,
+              accessToken,
+              refreshToken: newRefreshToken,
+              expiresAt
+            }
+            upsertTeamMemberIntegrationAuth(newGitlabAuth)
+            return newGitlabAuth
+          }
+          return gitlabAuth as IGetTeamMemberIntegrationAuthQueryResult | null
```
Dschoordsch, line 48: -1 why add `| null` when we know it is `!== null` from l22?

packages/server/integrations/helpers/authorizeOAuth2.ts
```diff
@@ -32,7 +32,12 @@ const transformBody = (contentType: string, body?: Record<string, string>): stri
 }
 
 export const authorizeOAuth2 = async <
-  TSuccess = {accessToken: string; refreshToken: string | undefined; scopes: string | undefined}
+  TSuccess = {
+    accessToken: string
+    refreshToken: string | undefined
+    scopes: string | undefined
+    expires_in: number | undefined
```
Dschoordsch, line 39: -1 we're fixing the other fields names to camelCase, so please do it also for `expiresIn`

Dschoordsch (changes requested):
Only small things, feel free to merge after fixing. Did not test.

PR #6599, opened 2022-05-20
fix(Sprint Poker): An exception could occur when modifying the scope in fast succession

### LINKED ISSUE #6334: `Minified exception occurred` error in poker meeting
Some users see the following error during their poker meetings with the Jira integrations:

image

See sentry for additional information:

https://sentry.io/organizations/parabol/issues/?environment=server&environment=client&project=107196&query=minified&statsPeriod=14d

image

- [ ] Triage is performed to find the root cause of the bug
- [ ] The error is fixed and does not prevent user to add Jira issues to the poker scope 

Dschoordsch:
# Description

Fixes #6334 

Minified exception occured on production because all exception messages are stripped out. When adding and removing issues in fast succession, it could happen that the optimistic updater throws an error because the store is in a weird state. This fix just works around this particular case but does not fix the root cause.

## Demo

This is how I was able to reproduce it before the fix:
https://www.loom.com/share/f2990f1feab448bca0d50190e67a16d5
I needed to repeat this a couple of times, it was not easily reproducible.
I tested with Jira as GitHub was probably bottlenecked by the server which resulted in different server side errors.

## Testing scenarios

- [ ] See loom, Sprint Poker with Jira
  - Add and remove tasks in quick succession
  - No error is shown

- [ ] Conclude the sprint poker and then modify the scope
  - Add remove more than 5 items in 10s (individual operations)
  - The server will not print an error like
    ```
11:08:09 1|GraphQL Executor | Error: Could not achieve lock for lock:queue:meeting:dtVNSO7Av6 after 10000ms
    ```

## Final checklist

- [ ] I checked the [code review guidelines](../../docs/codeReview.md)

### DIFF
--- a/packages/client/mutations/UpdatePokerScopeMutation.ts
+++ b/packages/client/mutations/UpdatePokerScopeMutation.ts
@@ -134,12 +134,13 @@ const UpdatePokerScopeMutation: StandardMutation<TUpdatePokerScopeMutation, Hand
         const stagesForTaskId = stages.filter(
           (stage) => stage.getValue('taskId') === firstStageTaskId
         )
-        const prevDimensionRefIds = stagesForTaskId.map((stage) => {
+        stagesForTaskId.forEach((stage) => {
           const dimensionRef = stage.getLinkedRecord('dimensionRef')
-          return dimensionRef?.getValue('id') ?? ''
-        }) as string[]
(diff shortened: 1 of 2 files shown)


BartoszJarocki (approved):
tested all listed scenarios, everything works well 👍

PR #6607, opened 2022-05-20
Jira server pagination

### LINKED ISSUE #6605: Jira Server: pagination of results in Sprint Poker
Implement pagination of results in Sprint poker

The issue extracted from #6097

**Estimated effort:** 8

igorlesnenko:
# Description

Fixes #6605

## Testing scenarios

- Go to the scope phase of a sprint poker meeting
- Open Jira Server tab
- Scroll till the end and see that new issues are loaded


packages/server/graphql/types/JiraServerIssue.ts
```diff
@@ -96,14 +94,10 @@ const {connectionType, edgeType} = connectionDefinitions({
   nodeType: JiraServerIssue,
   edgeFields: () => ({
     cursor: {
-      type: GraphQLISO8601Type
+      type: GraphQLString
     }
   }),
   connectionFields: () => ({
-    pageInfo: {
```
nickoferrall, line 103: -1 why remove `pageInfo`? If the date cursor type isn't working, you can use the `PageInfo` type instead: https://github.com/ParabolInc/parabol/blob/master/packages/server/graphql/types/PageInfo.ts

packages/server/graphql/types/JiraServerIntegration.ts
```diff
@@ -165,7 +175,22 @@ const JiraServerIntegration = new GraphQLObjectType<{teamId: string; userId: str
           }
         })
 
-        return connectionFromTasks(mappedIssues, first)
+        const nodes = mappedIssues.slice(0, first)
+        const edges = mappedIssues.map((node, index) => ({
+          cursor: `${index + afterInt}`,
```
nickoferrall, line 180: -2 this looks like an offset pagination rather than a cursor pagination. Offset paginations don't scale as well and can result in data being skipped or duplicated: https://stackoverflow.com/questions/18314687/how-to-implement-cursors-for-pagination-in-an-api

Instead, is there a unique field associated with each edge that we could use as a cursor?

nickoferrall (changes requested):

packages/server/graphql/types/JiraServerIssue.ts
igorlesnenko: @nickoferrall I removed it here, because we have default one defined here https://github.com/ParabolInc/parabol/blob/8a848a90858a29870572161ee59586f1c2b1fb15/packages/server/graphql/connectionDefinitions.ts#L40

packages/server/graphql/types/JiraServerIntegration.ts
igorlesnenko: @nickoferrall yeah, but looks like Jira Server API supports only this kind of pagination: https://docs.atlassian.com/software/jira/docs/api/REST/7.6.1/ (see Pagination section)

nickoferrall: Ah how annoying. As we're going with an offset pagination, I'm curious why the cursor needs to be a string rather than a number?

igorlesnenko: @nickoferrall relay requires this to be a string https://relay.dev/graphql/connections.htm#sec-Cursor. It throws an error  otherwise

nickoferrall (approved):
LGTM!

packages/server/graphql/types/JiraServerIntegration.ts
```diff
@@ -136,7 +135,18 @@ const JiraServerIntegration = new GraphQLObjectType<{teamId: string; userId: str
           (projectKeyFilter) => IntegrationRepoId.split(projectKeyFilter).projectKey!
         )
 
-        const issueRes = await integrationManager.getIssues(queryString, isJQL, projectKeys)
+        // "first + 1" is a trick to make pagination works easier, see hasNextPage
+        const maxResults = first + 1
+        // Cursor must be a string
```
Dschoordsch, line 140: +1 It's always better to state why something is the case, i.e. why does the cursor need to be a string?

packages/server/graphql/types/JiraServerIntegration.ts
```diff
@@ -136,7 +135,18 @@ const JiraServerIntegration = new GraphQLObjectType<{teamId: string; userId: str
           (projectKeyFilter) => IntegrationRepoId.split(projectKeyFilter).projectKey!
         )
 
-        const issueRes = await integrationManager.getIssues(queryString, isJQL, projectKeys)
+        // "first + 1" is a trick to make pagination works easier, see hasNextPage
```
Dschoordsch, line 138: +1 This is more cryptic than helpful. I think "Request one extra item to see if there are more results" would be easier to understand in a couple of month.

Dschoordsch (approved):

Issue #6561, opened 2022-05-16
Standups: Order response cards


jmtaber129:
We should sort response cards such that they're consistent between page loads and viewers.  Specifically, the order should be:
* Viewer's cards (regardless of whether it's completed)
* Non-empty cards
  * Ordered by response's `createdAt` time
* Empty cards
  * Ordered by when the team member's stage was created
  * For team members who joined the meeting by being a part of the team, the ordering should be consistent within a meeting, but the sort key is unimportant.

Suggestion: do the non-viewer card sorting on the server to ensure consistency.

### Acceptance Criteria
- [ ] Viewer's card is first
- [ ] Non-empty cards appear immediately after viewer's card
- [ ] Empty cards appear after non-empty cards
- [ ] When a response transitions from empty to non-empty, the card moves to between the last non-empty card and the first empty card
- [ ] When a team member joins, their empty card appears at the end of the card list
- [ ] When a response transitions from non-empty to empty (i.e. team member cleared their response), the card moves to the empty cards (exact placement isn't important)

ackernaut:
Overall, this makes sense to me. I do have a question about this detail:

>Non-empty cards Ordered by response's createdAt time

Could we stick the newest cards first in the sort order? This would favor fresh content, putting it at the top. The thinking is if I looked at the view I’d see new responses. That assumption was based on an async flow.

I suppose the layout changes could get busy if a bunch of folks were adding responses at the same time. If they were holding a sync meeting this may be annoying if I’m trying to read other cards. The reason this may be okay is that my card will always be top left, so I can edit without having my card move around.

If there are doubts, I think adding non-empty responses to the grid (oldest first, newest appended) is a safe starting point to have less layout change. The goal is to ship good over perfect.

jmtaber129:
I have a medium preference towards optimizing for less layout changes, at least for non-empty cards, since the impact of things moving around while you're trying to read responses is pretty negative IMO.  E.g. If I'm trying to read through my teammates' responses, and suddenly they all shift while I'm reading because someone just submitted their first response, that would be a pretty significant interrupt, and would make it difficult to get through reading all the responses.

Maybe there's a more targeted enhancement of surfacing unread cards that we could do as a later milestone.  Perhaps a highlight or a little red dot in the corner of cards that were updated since last view?

ackernaut:
Okay, we can start with less layout change. I like deferring ideas around showing what’s new. I think once we and others use it for a while we can revisit.

PR #6618, opened 2022-05-23
feat(standups): Card ordering w/ animated transitions

jmtaber129:
# Description
fixes https://github.com/ParabolInc/parabol/issues/6561
Enforce the ordering defined in the issue.  Though I originally intended to do the ordering on the server-side for stronger consistency, I couldn't figure out a reasonable way to do it in the graphql layer, so it's happening on the client-side.

Also changed the `useTransition` hook to trigger transitions when elements change order (this has primarily been used for mounting and unmounting in other components), and renamed the `useAnimatedMeetingCard` hook to `useAnimatedCard` since that's its level of generalization.

Also added a `sortByISO8601Date` util for stable sorting by string date.

## Demo
https://www.loom.com/share/22b3e13614ed4bc58196f4dfafcc8c20

## Testing scenarios
Sign in with a few accounts: User A and User B on Team X with a few other users, and User C who is _not_ on Team X
- [ ] Create a new standup meeting with User A
- [ ] Join the meeting as User B (should have a card that is not right after the viewer's card)
- [ ] Confirm viewer's card is first for both users A and B
- [ ] Update User B's response
- [ ] Confirm that User B's card animates to right after User A's card on User A's screen
- [ ] Clear out User B's response
- [ ] Confirm that User B's card animates to where it was previously
- [ ] Update User B's response
- [ ] Sign in as User C with an invite link to the meeting
- [ ] Confirm that User C's card appears at the end of the list
- [ ] Update User C's response
- [ ] Confirm that User C's card animates to right after User B's card on User A's screen

## Final checklist

- [X] ~I added the label `One Review Required` if the PR introduces only minor changes, does not contain any architectural changes or does not introduce any new patterns and I think one review is sufficient'~


BartoszJarocki (approved):
great work!

BartoszJarocki:
@Dschoordsch do you mind taking a look?

packages/client/hooks/useTransition.ts
```diff
@@ -76,17 +76,17 @@ const useTransition = <T extends {key: Key}>(children: T[]) => {
...
       const status = idxInPrev === -1 ? TransitionStatus.MOUNTED : prevTChildren[idxInPrev]!.status
       currentTChildren.push({
         status,
         child: nextChild,
         onTransitionEnd: transitionEndFactory(nextChild.key)
       })
-      if (idxInPrev === -1) {
+      if (idxInPrev === -1 || idxInPrev !== idxInNext) {
         touched = true
         mountingKeys.push(nextChild.key)
```
Dschoordsch, line 91: +1 since `mountingKeys` contains now all changed keys, it would be nice to change the name accordingly

packages/client/components/TeamPromptMeeting.tsx
```diff
@@ -99,26 +88,31 @@ const TeamPromptMeeting = (props: Props) => {
...
-      .filter(isNotNull)
-
-    // Find the viewer's card and put it at the beginning.
+    const allStages = phase.stages.filter(isNotNull)
+
+    const nonViewerStages = allStages.filter((stage) => stage.teamMember.userId !== viewerId)
+    const orderedNonEmptyStages = nonViewerStages
+      .filter((stage) => !!stage.response?.plaintextContent)
+      .sort((stageA, stageB) =>
+        sortByISO8601Date(stageA.response!.createdAt, stageB.response!.createdAt)
```
Dschoordsch, line 97: +1 Wouldn't it be more intuitive to order by `updatedAt`?

packages/client/components/TeamPromptMeeting.tsx
```diff
@@ -99,26 +88,31 @@ const TeamPromptMeeting = (props: Props) => {
...
+      )
+    // Empty stages are implicitly ordered by stage creation time on the backend.
+    const orderedEmptyStages = nonViewerStages.filter((stage) => !stage.response?.plaintextContent)
+    let orderedStages = [...orderedNonEmptyStages, ...orderedEmptyStages]
+
+    // Find the viewer's card.
     const viewerCardIndex = allStages.findIndex((stage) => stage.teamMember.userId === viewerId)
```
Dschoordsch, line 104: +1 you could use `const viewerCard = allStages.find(...` to omit the index

Dschoordsch (approved):
Did not test

packages/client/components/TeamPromptMeeting.tsx
jmtaber129: Per the [issue](https://github.com/ParabolInc/parabol/issues/6561), we decided to optimize for fewer layout changes in non-empty cards to avoid the case where a card moves while a user is reading it.  So similar to not pushing a new card to the front of the list, we probably don't want to be moving updated cards to the end of the list (thus shifting all non-empty cards up one space).

We'll probably revisit this ordering if there are tensions.

packages/client/hooks/useTransition.ts
jmtaber129: Changed to `updatedKeys`

packages/client/components/TeamPromptMeeting.tsx
jmtaber129: Nice catch!

PR #6623, opened 2022-05-24
DRAFT: move subscriptions to gqlExecutor

### LINKED ISSUE #6622: Move subscriptions to gqlExecutors
At the moment the websocket server also maintains the subscriptions and their state. When a new date is published from the gqlExecutor it is send via Redis PubSub to the websocket server which then sends a request back to the same gqlExecutor for resolution. This can be streamlined when the gqlExecutor also resolves the subscription immediately and only sends the (already resolved) response to the websocket server.

## Goal
Reduce round trips between gqlExecutor and websocket server.

## Subscription data storage
If the gqlExecutor should resolve the subscription already when new date is published, then it needs to have access to the subscription data `{query, variables, websocket, authToken}`. This is currently maintained in the single websocket server in memory, but should be stored in Redis instead.
Format should be a hash set with name `channel` aka `${topic}:${id}` as key so it is efficient to retrieve with `hmgetall`. The keys can be the `websocketId` so there are no duplicate subscriptions per socket and disconnecting is eaiser

## Publish
Instead of publishing a new date via pubsub to the 

## Subscribing
The websocket server would still execute `createSourceEventStream` and and create an async iterator for the next date. These dates should be fully resolved responses now.
In addition to creating the stream, the subscription data is also written to Redis.

## Unsubscribing
The websocket key is removed from the set.

## Timeouts / server crashes
Should the websocket server crash there is no builtin mechanism to automatically remove the stale subscription data. This is the biggest challenge when implementing this.
There are 2 failure scenarios:
### websocket server crashes
In this case all subscription data is stale and the whole `${topic}:${id}` set should be deleted. We could add a `expire` to this key and update it regularly from the websocket server.
Issue could be if there are new subscriptions for the same map refreshing the ttl while the map still contains a lot of stale subscription data. This case is very likely since the users connected while the server crashes most likely will reconnect once it restarts. This requires some thought.

### websocket connection is aborted
Is this a problem without the websocket crashing? The socket disconnect hooks should be properly processed and the keys removed

**EE**: 13h

Dschoordsch:
# Description

Unsubscribing does not work yet and the source/response streams need
cleanup/merging.

Fixes #6622 

## Demo

## Testing scenarios

- [ ] Scenario A
  - Step 1
  - Step 2...

- [ ] Scenario B
  - Step 1
  - Step 2....

## Final checklist

### DIFF
--- a/packages/server/graphql/ResponseStream.ts
+++ b/packages/server/graphql/ResponseStream.ts
@@ -1,6 +1,6 @@
 import {ExecutionResult} from 'graphql'
-import {getUserId} from '../utils/authorization'
-import getGraphQLExecutor from '../utils/getGraphQLExecutor'
+//import {getUserId} from '../utils/authorization'
+//import getGraphQLExecutor from '../utils/getGraphQLExecutor'
 import sendToSentry from '../utils/sendToSentry'
 import SubscriptionIterator from '../utils/SubscriptionIterator'
 import {SubscribeRequest} from './subscribeGraphQL'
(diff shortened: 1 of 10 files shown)


Dschoordsch:
@Dschoordsch any updates on this?

Dschoordsch:
Not giving this one up yet. Will work on #7630 and once that's in I will benchmark this.

mattkrick:
Just checking in on this one, still good to keep in PR?

mwermuth:
Looked at during backlog grooming.

Issue #6584, opened 2022-05-18
Standups: Edit prompt from meeting


jmtaber129:
Facilitators can edit the prompt after a meeting has started.

Designs - https://www.figma.com/file/bhiEOV4HK9Kv91HFsJmevV/Stand-up-Concepts?node-id=1087%3A71021

Design issue - https://github.com/ParabolInc/parabol/issues/6069

Suggested backend approach (consistent with other meetings):
* New field on `TeamPromptMeeting` in graphql and rethink for the prompt
* `startTeamPrompt` mutation initializes new field on `TeamPromptMeeting` to the current `What are you working on today? Stuck on anything?`.  Note that this will read from the team meeting settings once we support editing the prompt from the "new meeting" view.
* New mutation (e.g. `updateMeetingPrompt`) that takes meeting ID and new prompt and updates the prompt.

AC:
- [ ] Facilitators see a pencil icon to the right of the prompt (per designs)
  - [ ] on-hover styling for the prompt text or pencil is consistent with the styling for editable meeting names in other meetings (text + pencil changes to a lighter color)
- [ ] Clicking the pencil opens an "edit prompt" modal
- [ ] Modal shows editable input field with the current prompt
- [ ] Suggested prompts (can be moved out of scope at dev's discretion):
  - [ ] Modal shows list of static, hard-coded suggested prompts
  - [ ] Clicking a suggested prompt populates the input field, which can still be edited.
- [ ] Clicking "use prompt" updates the prompt
  - [ ] The local client immediately shows the new prompt and closes the modal
  - [ ] After refreshing the page (and reloading the prompt from the server), the new prompt still appears
  - [ ] Other users currently in the meeting see the new prompt as soon as it's updated
- [ ] Non-facilitator members cannot edit the prompt
- [ ] Facilitators cannot edit the prompt of a meeting that has ended

PR #6625, opened 2022-05-24
feat(standups): Editable Prompt within meeting

jmtaber129:
# Description
fixes https://github.com/ParabolInc/parabol/issues/6584
Note that this does _not_ match the styling on mobile.  Due to some styling used as part of the `useModal` hook internals, we'd need to either refactor some styles, or implement a mobile-friendly overlay using lower-level hook APIs.

## Demo
https://www.loom.com/share/7b8aac4235594fc18003a4f26f6c7336

## Testing scenarios
* Create a new meeting as facilitator, and also join as a separate user
* Confirm that there is an initial meeting prompt
* Click the meeting prompt and see a modal open
* Click a suggested prompt and confirm that it populates the text box
* Modify the prompt in the text box
* Click "Use Prompt"
* Confirm that the new prompt appears on the facilitator's view, and the other user's view updated to the new prompt
* As facilitator, end the meeting from the options menu
* Confirm that the prompt no longer appears editable, and the modal does not open when clicking

## Final checklist

- [ ] ~I added the label `One Review Required` if the PR introduces only minor changes, does not contain any architectural changes or does not introduce any new patterns and I think one review is sufficient'~


packages/client/components/TeamPrompt/TeamPromptEditablePrompt.tsx
```diff
@@ -0,0 +1,95 @@
...
+    `,
+    meetingRef
+  )
+
+  const {id: meetingId, meetingPrompt, facilitatorUserId, endedAt} = meeting
+  const {viewerId} = atmosphere
+  const isFacilitator = viewerId === facilitatorUserId
+
+  const handleUpdatePrompt = useEventCallback((newPrompt) => {
+    if (submitting || error) return
```
BartoszJarocki, line 63: -1 if mutation returns an error, it'll be persisted and the user won't be able to try again. with current impl error will be only cleared when the user changes the prompt which isn't ideal in the case where mutation fails because of the network conditions, for example. I'd personally keep validation errors and mutation errors separately and check only validation errors here

BartoszJarocki (changes requested):
It looks really nice! I just found one case that prevents users from trying to change the prompt  when mutation fails. 

+1 is the mobile view, I think we can make fonts smaller and make the dialog content scrollable, see the screenshots
localhost_3000_meet_dS9nIC7pE4_responses
localhost_3000_meet_dS9nIC7pE4_responses(iPhone SE)

packages/client/components/TeamPrompt/TeamPromptEditablePromptModal.tsx
```diff
@@ -0,0 +1,188 @@
...
+          onChange={(e) => {
+            const nextValue = e.target.value || ''
+            setEditablePrompt(nextValue)
+          }}
+          placeholder={'What are you working on today? Stuck on anything?'}
+          value={editablePrompt}
+        />
+        {error && (
+          <ErrorWrapper>
+            <ErrorIcon>info</ErrorIcon>
```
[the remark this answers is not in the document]
jmtaber129: This is consistent with designs: https://www.figma.com/file/bhiEOV4HK9Kv91HFsJmevV/Stand-up-Concepts?node-id=1421%3A88986

BartoszJarocki (approved):

mattkrick (changes requested):

packages/client/components/TeamPrompt/TeamPromptEditablePromptModal.tsx
```diff
@@ -0,0 +1,201 @@
...
+          placeholder={'What are you working on today? Stuck on anything?'}
+        />
+        {displayError && (
+          <ErrorWrapper>
+            <ErrorIcon>info</ErrorIcon>
+            {displayError}
+          </ErrorWrapper>
+        )}
+        <SuggestedPromptsHeader>Suggestions</SuggestedPromptsHeader>
+        {SUGGESTED_PROMPTS.map((prompt, i) => (
```
mattkrick, line 176: +1 optionally filter out the currently selected value if it's easy & doesn't update on keystroke

mattkrick:
-1 Button needs an extra 4 - 8px top padding
Screen Shot 2022-06-09 at 9 06 00 AM

+1 title max width should be the same as the card area max width (subjective, designs don't show this case, but i think it'd look better)
Screen Shot 2022-06-09 at 9 14 55 AM

mattkrick (changes requested):
_super_ minor stuff, nothing blocking a merge, but i figure it'd be easy to fix these up now

PR #6629, opened 2022-05-25
feat(standups): Improved standups options button size

### LINKED ISSUE #6616: Standups: Make meeting options menu button size responsive
+1 the options menu button doesn't match the varying sizes of the avatars.  It looks like the styling for the `Options` component in `TeamPromptOptions` is enforcing a height+width of `32` for the button itself.

_Originally posted by @jmtaber129 in https://github.com/ParabolInc/parabol/pull/6614#pullrequestreview-982099539_

BartoszJarocki:
# Description

Fixes #6616 

Changes:
1. Added support for `active` and `focus` styles so it's easier to navigate using the keyboard
2. Changed the styles to match the back button used in the top bar 
3. Increased icon size

Note: I didn't add any breakpoints for icon size as I wanted to make it work the same way as the existing back button. The size of the icon was increased, IMO it looks good on both mobile and desktop. 

## Demo

https://user-images.githubusercontent.com/1017620/170276983-c6b9e653-13cc-4295-b370-be91255d1de1.mp4

## Testing scenarios

- [ ] Navigate all focusable elements on standups and make sure `active` and `focus` styles work correctly
- [ ] Hover style works correctly
- [ ] Resize the browser to see if the button is centered correctly all the time

## Final checklist

### DIFF
--- a/packages/client/components/TeamPrompt/TeamPromptOptions.tsx
+++ b/packages/client/components/TeamPrompt/TeamPromptOptions.tsx
@@ -4,18 +4,20 @@ import React from 'react'
 import {useFragment} from 'react-relay'
 import {MenuPosition} from '~/hooks/useCoords'
 import useMenu from '~/hooks/useMenu'
+import {PALETTE} from '~/styles/paletteV3'
 import {TeamPromptOptions_meeting$key} from '~/__generated__/TeamPromptOptions_meeting.graphql'
-import {PALETTE} from '../../styles/paletteV3'
-import CardButton from '../CardButton'
+import FlatButton from '../FlatButton'
 import IconLabel from '../IconLabel'
 import TeamPromptOptionsMenu from './TeamPromptOptionsMenu'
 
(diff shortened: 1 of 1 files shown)


BartoszJarocki:
@jmtaber129 as you originally pointed that out, would you mind taking a look? thanks!

jmtaber129 (approved):
+1 The on-hover shape is ovular instead of circular.  Changing back to `CardButton` and keeping `opacity: 1` seems to resolve this, but not sure if there's other tradeoffs with using `FlatButton`.
Screen Shot 2022-05-25 at 5 09 37 PM

BartoszJarocki:
@jmtaber129 thank you for pointing that out, not sure how I missed that 🤦 I changed `OptionsButton` to extend `BaseButton` instead of `FlatButton` (I haven't noticed it was intended to use as a pill-shaped button). The `CardButton` adds more properties and there's no reason to override them here, we only need a few of them, `BaseButton` seems like a better fit here.

PR #6640, opened 2022-05-27
feat(standups): Basic editable prompt

jmtaber129:
# Description
refs https://github.com/ParabolInc/parabol/issues/6584
Full "update prompt" mutation w/ basic UI.  The proper UI (full modal w/ static suggestions) will come in a follow-on PR.

In an effort to get this feature partially landed by the next release, and because this is still behind a feature flag, anything related to a "good" UI is outside the scope of this PR.  As a result, while the non-facilitator view should be unchanged, the facilitator's view of the prompt may seem a bit janky.  Additionally, most (all?) of the changes to `packages/client/components/TeamPromptMeeting.tsx` will be short-lived (only until the UI PR lands).

See https://github.com/ParabolInc/parabol/pull/6625/commits/993d027785be822e39b41a506fd61878e2d73e1f and later commits for context on how the UI will change to match designs.

## Demo
https://www.loom.com/share/55bf3227b9c844fcbe52e4c0c44c1efc

## Testing scenarios
- [ ] Facilitator can update prompt from within meeting
- [ ] Meeting prompt changes on other users' views when the facilitator changes the prompt

## Final checklist

### DIFF
--- a/packages/client/mutations/UpdateMeetingPromptMutation.ts
+++ b/packages/client/mutations/UpdateMeetingPromptMutation.ts
@@ -1,0 +1,50 @@
+import graphql from 'babel-plugin-relay/macro'
+import {commitMutation} from 'react-relay'
+import {StandardMutation} from '../types/relayMutations'
+import {UpdateMeetingPromptMutation as TUpdateMeetingPromptMutation} from '../__generated__/UpdateMeetingPromptMutation.graphql'
+
+graphql`
+  fragment UpdateMeetingPromptMutation_meeting on UpdateMeetingPromptSuccess {
+    meeting {
+      id
+      meetingPrompt
+    }
+  }
+`
+
+const mutation = graphql`
+  mutation UpdateMeetingPromptMutation($meetingId: ID!, $newPrompt: String!) @raw_response_type {
+    updateMeetingPrompt(meetingId: $meetingId, newPrompt: $newPrompt) {
+      ... on ErrorPayload {
+        error {
+          message
+        }
+      }
+      ...UpdateMeetingPromptMutation_meeting @relay(mask: false)
+    }
+  }
+`
+
+const UpdateMeetingPromptMutation: StandardMutation<TUpdateMeetingPromptMutation> = (
+  atmosphere,
+  variables,
+  {onError, onCompleted}
+) => {
+  return commitMutation<TUpdateMeetingPromptMutation>(atmosphere, {
+    mutation,
+    variables,
+    optimisticResponse: {
+      updateMeetingPrompt: {
+        __typename: 'UpdateMeetingPromptSuccess',
+        meeting: {
+          id: variables.meetingId,
+          meetingPrompt: variables.newPrompt
+        }
+      }
+    },
+    onCompleted,
+    onError
+  })
+}
+
+export default UpdateMeetingPromptMutation
--- a/packages/server/graphql/public/mutations/updateMeetingPrompt.ts
+++ b/packages/server/graphql/public/mutations/updateMeetingPrompt.ts
@@ -31,7 +31,7 @@ const updateMeetingPrompt: MutationResolvers['updateMeetingPrompt'] = async (
   }
 
   // VALIDATION
-  if (newPrompt.length < 2) {
+  if (newPrompt.length < 2 || newPrompt.length > 500) {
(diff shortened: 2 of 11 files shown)


packages/server/graphql/public/mutations/updateMeetingPrompt.ts
```diff
@@ -0,0 +1,51 @@
...
+  const meeting = await dataLoader.get('newMeetings').load(meetingId)
+  if (!meeting) {
+    return {error: {message: 'Meeting not found'}}
+  }
+  if (meeting.meetingType !== 'teamPrompt') {
+    return {error: {message: 'Meeting is not a team prompt meeting'}}
+  }
+  const {facilitatorUserId} = meeting
+  if (viewerId !== facilitatorUserId) {
+    return {error: {message: 'Only the facilitator can change the meeting prompt'}}
```
BartoszJarocki, line 27: +1 same as above

BartoszJarocki (approved):
lgtm!

packages/client/mutations/UpdateMeetingPromptMutation.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const UpdateMeetingPromptMutation: StandardMutation<TUpdateMeetingPromptMutation> = (
+  atmosphere,
+  variables,
+  {onError, onCompleted}
+) => {
+  return commitMutation<TUpdateMeetingPromptMutation>(atmosphere, {
+    mutation,
+    variables,
+    optimisticResponse: {
```
mattkrick, line 36: +2 nice!

packages/server/graphql/public/mutations/updateMeetingPrompt.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+  const {facilitatorUserId} = meeting
+  if (viewerId !== facilitatorUserId) {
+    return standardError(new Error('Only the facilitator can change the meeting prompt'), {
+      userId: viewerId
+    })
+  }
+
+  // VALIDATION
+  if (newPrompt.length < 2) {
```
mattkrick, line 34: -1 needs to validate max length. client uses 500, that seems reasonable. (don't want someone to paste in 5MB of junk text)

packages/server/graphql/public/typeDefs/updateMeetingPrompt.graphql
```diff
@@ -0,0 +1,32 @@
...
+"""
+union UpdateMeetingPromptPayload = UpdateMeetingPromptSuccess | ErrorPayload
+
+type UpdateMeetingPromptSuccess {
+  meetingId: ID!
+
+  """
+  the updated meeting
+  """
+  meeting: TeamPromptMeeting
```
mattkrick, line 29: -1 should be `TeamPromptMeeting!`

packages/server/graphql/public/types/UpdateMeetingPromptSuccess.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export type UpdateMeetingPromptSuccessSource = {
+  meetingId: string
+}
+
+const UpdateMeetingPromptSuccess: UpdateMeetingPromptSuccessResolvers = {
+  meeting: async (source, _args, {dataLoader}) => {
+    const {meetingId} = source
+    const meeting = await dataLoader.get('newMeetings').load(meetingId)
+    return (meeting.meetingType !== 'teamPrompt' ? null : meeting) as Maybe<
```
mattkrick, line 11: +1 no need to validate the meetingType here since `meetingId` is provided by the server (we already validated it in the resolve function of the mutation. can just return it as a `TeamPromptMeeting`. returning a nonnull for a success payload is always nice because then the client doesn't have to check to see if `meeting` exists.

my heuristic for these checks is asking myself if we could achieve 100% branch coverage if we wrote an integration test for it. if we can't test the branch that returns `null` then we don't need to build a case for it

mattkrick (changes requested):
-1 nice work! just 1 quick fix to make `meeting` nonnull in the success payload because hey, if it's successful, then it should include a meeting object, right?

BartoszJarocki:
@mattkrick @jmtaber129 as we want to have it in this week's release and I see all the comments were resolved I'll allow myself to merge this 🙏

PR #6658, opened 2022-06-01
feat(standups): Update relative `createdAt` automatically

### LINKED ISSUE #6628: Standups: update displayed `createdAt` and `updatedAt` time continuously
When looking at the standup prompt for a prolonged period, it seems odd if the created time still shows "Just now" and updates on hover to something like "7 minutes ago". The time format should update automatically.

## Acceptance criteria
* [ ] Update `TeamPromptLastUpdatedTime` to automatically reformat the time
  * [ ] setPeriod or setTimeout can be used to update the state as the formatting update does not happen that often
  * [ ] make sure to stop the timer in cleanup of `React.useEffect`

_Originally posted by @Dschoordsch in https://github.com/ParabolInc/parabol/pull/6557#discussion_r881501532_

BartoszJarocki:
# Description

Fixes #6628 

## Demo
No demo

## Testing scenarios

- [ ] Open a standup meeting, add a response wait for more than 1min, and see if `createdAt` is updated automatically (ie. from  `Just now` to `1 minute ago` 

## Final checklist

### DIFF
--- a/packages/client/components/TeamPrompt/TeamPromptLastUpdatedTime.tsx
+++ b/packages/client/components/TeamPrompt/TeamPromptLastUpdatedTime.tsx
@@ -1,5 +1,5 @@
 import styled from '@emotion/styled'
-import React from 'react'
+import React, {useEffect, useState} from 'react'
 import {MenuPosition} from '../../hooks/useCoords'
 import useTooltip from '../../hooks/useTooltip'
 import {PALETTE} from '../../styles/paletteV3'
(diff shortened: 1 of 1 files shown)


packages/client/components/TeamPrompt/TeamPromptLastUpdatedTime.tsx
```diff
@@ -20,7 +20,19 @@ interface Props {
...
 export default function TeamPromptLastUpdatedTime({updatedAt, createdAt}: Props) {
+  const [relativeCreatedAt, setRelativeCreatedAt] = useState(() => relativeDate(createdAt))
+
+  useEffect(() => {
+    const intervalId = setInterval(() => {
+      setRelativeCreatedAt(relativeDate(createdAt))
+    }, UPDATE_INTERVAL_MS)
+
+    return () => clearInterval(intervalId)
+  }, [createdAt])
```
jmtaber129, line 34: +1 Would it make sense to use `useRefreshInterval` here?  That way instead of manually updating the `relativeCreatedAt` on the interval, we just let the full component refresh itself, including the relative timestamp.  That should reduce this to a two-ish line change (just the constant and calling the hook).

jmtaber129 (approved):
See inline comment - I think `useRefreshInterval` may be the way to go here.

packages/client/components/TeamPrompt/TeamPromptLastUpdatedTime.tsx
BartoszJarocki: I'm not sure what's more readable in this case so I'll with the solution you mentioned as it's shorter, thanks!

PR #6667, opened 2022-06-02
fix(sprint-poker): Prevent kicking facilitator off the meeting while modifying the scope

### LINKED ISSUE #6231: Poker removing current stage issue breaks meeting
## How to reproduce
Gina has GitHub integration and Fen facilitates the meeting. Gina adds a task to estimate and Fen advances the meeting to vote on that task. Then Gina goes back to scope phase and removes the task while Fen stays at it.

### Expected behaviour
The meeting is advanced and can be continued normally.

### Actual behaviour
The facilitator is thrown out of the meeting and she cannot rejoin the meeting from the dashboard.
Removing current phase task breaks meeting

BartoszJarocki:
# Description

Fixes #6231 

The bug itself was caused by meeting subscription updating the relay store with no fully populated data. `UpdatePokerScopeMutation` didn't include data needed for `useInitialSafeRoute` and `useUpdatedSafeRoute` hooks to work. `facilitatorStageId` was undefined after the update and as a result, the facilitator was kicked off the meeting. 

## Demo

https://user-images.githubusercontent.com/1017620/171653362-24ce2b37-cc41-4028-b9cd-4bfa97568678.mp4

## Testing scenarios

- [ ] Alice creates the sprint poker and adds some issues to the scope. Bob joins the meeting and removes the issues added by Alice. Alice is still in the meeting and wasn't kicked off the meeting.

Doesn't really matter what integration is used, the scenario described in #6231 was true for all existing integrations. 

## Final checklist

### DIFF
--- a/packages/client/mutations/UpdatePokerScopeMutation.ts
+++ b/packages/client/mutations/UpdatePokerScopeMutation.ts
@@ -56,6 +56,7 @@ graphql`
       }
     }
     meeting {
+      ...PokerMeeting_meeting
       ...useInitialSafeRoute_meeting
       ...useUpdatedSafeRoute_meeting
       gitlabSearchQuery {
--- a/packages/client/mutations/UpdatePokerScopeMutation.ts
+++ b/packages/client/mutations/UpdatePokerScopeMutation.ts
@@ -56,6 +56,8 @@ graphql`
       }
     }
     meeting {
+      ...useInitialSafeRoute_meeting
+      ...useUpdatedSafeRoute_meeting
       gitlabSearchQuery {
         queryString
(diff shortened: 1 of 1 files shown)


BartoszJarocki:
@Dschoordsch do you mind taking a look? thanks!

Dschoordsch (approved):
I added another fragment to solve the case where the facilitator is at the last task. The control bar will show the "End meeting" button. When adding a new task in scoping, the button would not update to "Next" otherwise.

BartoszJarocki:
@Dschoordsch good find! and btw, I removed you because the calendar shows you're off! 🌴

PR #6670, opened 2022-06-02
feat(insights): add feature flag

### LINKED ISSUE #6662: Domain Stats: Backend mutation to flag PQLs
<details>
<summary> Thought process</summary>
Today, a user's company is calculated from their email domain.
That means if we turn on `User.featureFlags.insights`, they'll only see insights for the domain.
That means consultants wouldn't be able to see those insights since they wouldn't have the same domain.

So we've got a problem. Today, 1 user has many orgs, and each org could have a different company. ergo, 1 user has many companies. We can either force 1 user to 1 company, or we can show multiple insights, 1 for each org that they're in.

```gql
viewer {
  organizations {
    company {
      id
    }
  }
}
```

today, 1 company = 1 domain. we need to fix this. 
we kinda handle this in the SAML table. a SAML id is like a company, which then has multiple domains.
what if we break that link out into its own table?
Company table has:
- id
- name
- domains (array?)
- contacts/leads (array?)

in 3NF we'd need 3 tables:
Company table:
- id
- name
CompanyDomain table:
- domain
- companyId
CompanyContact table:
- companyId
- email/userId

not sure what to do about leads, stick them in the OrgUser? the challenge today is that domain & company are synonymous. we need to separate them out like we do in the SAML table.
We get that with the first 2 tables, but the company contacts is tough. this is a trait for a user set by a salesperson. i'd like to build this to support 1 user belonging to multiple companies because it's cheap to build it that way now, but it'll be expensive if we need that later.
so, 1 user has many orgUsers, 1 orgUser has 1 organizations, 1 organization has 1 company (which is currently calculated by activeDomain). So, since 1 org has 1 company, we can put it on the OrgUser document.

then again, hubspot already does a good job of extracting the Primary company from the user.
what if we just use hubspot as the single source of truth? or at least as a good default value.
</details>

v1: Insights for invitees 
- `insights` is added as a feature flag, when true, the insights link appears in the dash
- `allowInsights` is put on the OrganizationUser. that way we don't have to force 1 User to be attached to 1 Company.

v2: Insights for all (waaay out of scope, maybe we never do this)
- Everyone sees the insights link
- in addition to `allowInsights`, if your email address is verified, you can see the insights for the domain attached to your email.

AC

- there is an `insights` featureFlag
- there is a `toggleInsightsContact(value: Boolean!, domain: String!, emails: [String!])` mutation on the private schema. Calling it will set `isInsightsContact` to true for every OrganizationUser that is apart of an Organization where `Organization.activeDomain === domain`. If `value === false`, then that field is unset for everyone in the emails arg. if emails arg is empty, then it turns it false for everyone in every org. if `value === true`, then emails is non-null (because setting it for everyone at once is risky).

Estimate: 16 hours

mattkrick:
# Description

Partially Fixes #6662

This PR adds the insights feature flag & refactors featureFlags to the SDL pattern

## Demo
Screen Shot 2022-06-02 at 2 17 34 PM

## Testing scenarios

- [ ] in GraphiQL, run the following

```gql
mutation {
  addFeatureFlag(domain: "parabol.co", flag:insights) {
    users {
      id
    }
  }
}
```


mattkrick:
closing in favor of #6664

PR #6671, opened 2022-06-02
fix(Retro Templates): Added missing prompts

### LINKED ISSUE #6666: Fix incomplete template prompts
@Dschoordsch noticed that some of the templates are missing prompt text. We should fix this to include the missing prompts.

This currently affects two retrospective templates:

1. Keep Problem Try
2. SaMoLo

Screenshot 2022-06-02 at 13 40 30
Screenshot 2022-06-02 at 13 40 23

The correct prompts for the templates are as follows:

**Keep, Problem, Try 🤔**

Keep
What should we continue doing?

Problem
What’s not going well?

Try
What’s the smallest next step we could take to resolve our problems?

_Source_: #5891

**SaMoLo ⚖️**

Same of
What do we want to keep doing?

More of
What do we want to do more of?

Less of
What do we want to do less of?

_Source_: #5891

Dschoordsch:
# Description

Fixes #6666

Some prompts were missing because of an `id` clash with existing
prompts because the `id` was purely generated from the question itself.
Changed it to be generated including the `templateId`.

## Demo

Screenshot from 2022-06-03 00-11-27
Screenshot from 2022-06-03 00-10-58

## Testing scenarios

New meeting dialog, choose Retro and switch template. "SaMoLo" and "Keep, Problem, Try" have 3 prompts each matching their name.

### DIFF
--- a/packages/server/database/migrations/20220124600507-addMoreRetroTemplates.ts
+++ b/packages/server/database/migrations/20220124600507-addMoreRetroTemplates.ts
@@ -626,6 +626,7 @@ const makePrompt = (promptInfo: PromptInput, idx: number) => {
   const {question, description, promptColor, promptId, templateId, sortOrder} = promptInfo
   const paletteIdx = idx > promptColors.length - 1 ? idx % promptColors.length : idx
   const groupColor = promptColor ? promptColor : promptColors[paletteIdx]
+  // FIXME this creates duplicated ids
   const id = promptId ? promptId : nameToId(question, false)
   return {
     createdAt,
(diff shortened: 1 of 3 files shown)


packages/server/database/migrations/20220124600507-addMoreRetroTemplates.ts
```diff
@@ -626,6 +626,7 @@ const makePrompt = (promptInfo: PromptInput, idx: number) => {
   const {question, description, promptColor, promptId, templateId, sortOrder} = promptInfo
   const paletteIdx = idx > promptColors.length - 1 ? idx % promptColors.length : idx
   const groupColor = promptColor ? promptColor : promptColors[paletteIdx]
+  // FIXME this creates duplicated ids
```
Dschoordsch, line 629: Added this just in case someone wants to copy&paste this code.

nickoferrall (approved):
LGTM

PR #6687, opened 2022-06-07
feat(insights): Add /insights route

mattkrick:
# Description

## Demo

## Testing scenarios

- [ ] Scenario A
  - Step 1
  - Step 2...

- [ ] Scenario B
  - Step 1
  - Step 2....

## Final checklist

### DIFF
--- a/packages/server/graphql/public/types/User.ts
+++ b/packages/server/graphql/public/types/User.ts
@@ -1,3 +1,4 @@
+import {isNotNull} from 'parabol-client/utils/predicates'
 import {isSuperUser} from '../../../utils/authorization'
 import getDomainFromEmail from '../../../utils/getDomainFromEmail'
 import isCompanyDomain from '../../../utils/isCompanyDomain'
@@ -8,6 +9,19 @@ const User: UserResolvers = {
     const domain = getDomainFromEmail(email)
     if (!domain || !isCompanyDomain(domain) || !isSuperUser(authToken)) return null
     return {id: domain}
+  },
+  domains: async ({id: userId}, _args, {dataLoader}) => {
+    const organizationUsers = await dataLoader.get('organizationUsersByUserId').load(userId)
+    const orgIds = organizationUsers
+      .filter(({allowInsights}) => allowInsights)
+      .map(({orgId}) => orgId)
+    const organizations = await Promise.all(
+      orgIds.map((orgId) => dataLoader.get('organizations').load(orgId))
+    )
+    return organizations
+      .map(({activeDomain}) => activeDomain)
+      .filter(isNotNull)
+      .map((id) => ({id}))
   },
   featureFlags: ({featureFlags}) => {
     return Object.fromEntries(featureFlags.map((flag) => [flag as any, true]))
--- a/packages/client/components/Dashboard.tsx
+++ b/packages/client/components/Dashboard.tsx
@@ -6,6 +6,7 @@ import {Route, Switch} from 'react-router'
 import useBreakpoint from '~/hooks/useBreakpoint'
 import useSnackNag from '~/hooks/useSnackNag'
 import useSnacksForNewMeetings from '~/hooks/useSnacksForNewMeetings'
+import {PALETTE} from '~/styles/paletteV3'
 import {Breakpoint} from '~/types/constEnums'
 import useSidebar from '../hooks/useSidebar'
 import {DashboardQuery} from '../__generated__/DashboardQuery.graphql'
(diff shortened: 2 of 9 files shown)


packages/server/graphql/public/types/User.ts
```diff
@@ -9,6 +10,19 @@ const User: UserResolvers = {
     if (!domain || !isCompanyDomain(domain) || !isSuperUser(authToken)) return null
     return {id: domain}
   },
+  domains: async ({id: userId}, _args, {dataLoader}) => {
+    const organizationUsers = await dataLoader.get('organizationUsersByUserId').load(userId)
+    const orgIds = organizationUsers
+      .filter(({allowInsights}) => allowInsights)
```
nickoferrall, line 16: -1 it looks like `allowInsights` doesn't exist on `organizationUser` (see [here](https://github.com/ParabolInc/parabol/blob/231d8d5cf45760a2740e0ee945ce555887f37ca8/packages/server/database/types/OrganizationUser.ts)). Do you want to check the feature flag instead?

packages/client/components/StandardHubUserMenu.tsx
```diff
@@ -30,92 +30,90 @@ const TallMenu = styled(Menu)({
...
-          <MenuItemLabel>
+          <MenuItemLink to={'/me/organizations'}>
             <MenuItemIcon>account_balance</MenuItemIcon>
             {'Organizations'}
-          </MenuItemLabel>
+          </MenuItemLink>
         }
-        onClick={goToOrganizations}
       />
+      {insights && (
```
nickoferrall, line 82: +1 should we check the length of `domains` too? If the feature flag is turned on but there are no active domains, users will see an empty page

packages/server/graphql/public/types/User.ts
```diff
@@ -9,6 +10,19 @@ const User: UserResolvers = {
     if (!domain || !isCompanyDomain(domain) || !isSuperUser(authToken)) return null
     return {id: domain}
   },
+  domains: async ({id: userId}, _args, {dataLoader}) => {
```
nickoferrall, line 13: -1 at the moment, if I create two orgs with the same user, I see the same domain twice. Should we make domains unique? 

domains

packages/client/components/InsightsDomainNudge.tsx
```diff
@@ -0,0 +1,113 @@
...
+        </OverLimitCopy>
+      </OverLimitBlock>
+      <ButtonBlock>
+        <CTA size={'large'} onClick={onClickCTA}>
+          {CTACopy}
+        </CTA>
+      </ButtonBlock>
+      {biggestOrganization &&
+        modalPortal(
+          <CreditCardModal
```
nickoferrall, line 102: +1 after upgrading to pro, I don't get any confetti 🥲 

-1 after upgrading to pro, when the modal closes, the CTA button remains the same until I refresh the page. Some users might think that the upgrade wasn't successful.

packages/server/graphql/private/mutations/toggleAllowInsights.ts
```diff
@@ -0,0 +1,41 @@
+import {r} from 'rethinkdb-ts'
+import {getUsersByEmails} from '../../../postgres/queries/getUsersByEmails'
+import {MutationResolvers} from '../resolverTypes'
+
+const toggleAllowInsights: MutationResolvers['toggleAllowInsights'] = async (
```
nickoferrall, line 5: +1 we could use the [isCompanyDomain](https://github.com/ParabolInc/parabol/blob/4d979ad66a68ad399864f1fcc393cff4bd642b5e/packages/server/utils/isCompanyDomain.ts) function you created to make sure that no one adds `gmail` as a domain. It's a private mutation, so it's very unlikely that one of us would do that, but it would mean that a user could see data for all gmail users

nickoferrall (changes requested):
Great work building this so quickly!

packages/client/components/InsightsDomainNudge.tsx
```diff
@@ -0,0 +1,113 @@
...
+      window.open('mailto:love@parabol.co?subject=Increase Usage Limits')
+    }
+  }
+  const {togglePortal, closePortal, modalPortal} = useModal()
+  if (!showNudge) return null
+  return (
+    <NudgeBlock>
+      <OverLimitBlock>
+        <OverLimitCopy>
+          <b>{domainId}</b> is over the limit of <b>2 Free Teams</b>
```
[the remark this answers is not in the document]
nickoferrall: Maybe I hadn't had enough coffee when I read this - I read the copy as "You're 2 teams over the limit" rather than "over the limit of 2 teams", so I thought we wanted this to be a calculation. I think there's no harm in storing the limit as a variable though.

mattkrick (commented):
yikes that was a rough first pass on my part!

packages/client/components/InsightsDomainNudge.tsx
nickoferrall: > someone with a verified email with that matching domain

Should users with verified email addresses be able to see data for other orgs within the same company? I wonder whether BigCo would be happy with BigCo Europe Org having access to BigCo North America Org's data. 

+2 great stuff fixing the confetti 🎉  

I think the PR is good to go, but the Upgrade to Pro CTA doesn't update after upgrading. I think the button should disappear immediately after so the user knows the upgrade was successful.

mattkrick: > Should users with verified email addresses be able to see data for other orgs within the same company? I wonder whether BigCo would be happy with BigCo Europe Org having access to BigCo North America Org's data.

it's a really good question, i think it's the only way! if i tell you your company is using 100 seats & you only think you're using 10, you're gonna ask me who else is using parabol & i'll show ya all the verified user email addresses.

 > I think the PR is good to go, but the Upgrade to Pro CTA doesn't update after upgrading. I think the button should disappear immediately after so the user knows the upgrade was successful.

it goes away for me! i think this is the bug that only happens for superusers

PR #6700, opened 2022-06-08
fix(Sprint Poker): Remove padding in Discussion Drawer

### LINKED ISSUE #6433: Empty discussion thread width on mobile doesn't take full width
discussion-thread-width

It looks like the width is correct in retros & check-ins; it's just an issue in the Poker meeting.

### AC

- [ ] Empty discussion thread img uses available width

87prashant:
# Description

fix #6433 Remove the padding in the discussion Drawer

:star_struck: My first Open-Source Contribution

## Demo
Before
WhatsApp Image 2022-06-08 at 9 35 03 PM

After
After

## Testing scenarios

- [x] Check the discussion drawer on mobile Chrome/Safari
- [X] Check the discussion drawer on desktop Chrome/Safari

## Final checklist

- [ ] I added the label `One Review Required` if the PR introduces only minor changes, does not contain any architectural changes, or does not introduce any new patterns and I think one review is sufficient'
- [X] PR title is human readable and could be used in the changelog

### DIFF
--- a/packages/client/components/EstimatePhaseDiscussionDrawer.tsx
+++ b/packages/client/components/EstimatePhaseDiscussionDrawer.tsx
@@ -6,13 +6,14 @@ import {desktopSidebarShadow} from '~/styles/elevation'
 import {PALETTE} from '~/styles/paletteV3'
 import {ICON_SIZE} from '~/styles/typographyV2'
 import {EstimatePhaseDiscussionDrawer_meeting} from '~/__generated__/EstimatePhaseDiscussionDrawer_meeting.graphql'
-import {BezierCurve, DiscussionThreadEnum, ZIndex} from '../types/constEnums'
+import {BezierCurve, Breakpoint, DiscussionThreadEnum, ZIndex} from '../types/constEnums'
 import {DiscussionThreadables} from './DiscussionThreadList'
 import DiscussionThreadListEmptyState from './DiscussionThreadListEmptyState'
 import DiscussionThreadRoot from './DiscussionThreadRoot'
 import Icon from './Icon'
 import LabelHeading from './LabelHeading/LabelHeading'
 import PlainButton from './PlainButton/PlainButton'
+import mediaQueryforPoker from './mediaQueryforPoker'
 
 const Drawer = styled('div')<{isDesktop: boolean; isOpen: boolean}>(({isDesktop, isOpen}) => ({
   boxShadow: isDesktop ? desktopSidebarShadow : undefined,
@@ -29,7 +30,12 @@ const Drawer = styled('div')<{isDesktop: boolean; isOpen: boolean}>(({isDesktop,
   transition: `all 200ms ${BezierCurve.DECELERATE}`,
   userSelect: 'none',
   width: isOpen || !isDesktop ? DiscussionThreadEnum.WIDTH : 0,
-  zIndex: ZIndex.SIDEBAR
+  zIndex: ZIndex.SIDEBAR,
+  [mediaQueryforPoker(Breakpoint.POKER_DISCUSSION_DRAWER)]: {
+    width: '100vw',
+    position: 'fixed',
+    right: isOpen ? '-360px' : '-100vw'
+  }
 }))
 
 const ThreadColumn = styled('div')({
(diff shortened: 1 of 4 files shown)


Dschoordsch (changes requested):
@87prashant Thank you for work! Sorry it took so long. I checked the issue and your patch, I think it's going in the right direction. Discussions work differently on Retrospectives and Check-In meetings because they are always visible and cannot be opened/closed, but I think also in Sprint Poker these should be full width, i.e. the gap on the left should be removed when the total width of the screen is below a certain threshold.

Retro:
Screenshot from 2022-06-13 11-53-41
Sprint Poker:
Poker open discussion drawer

To me it seems we should have a breakpoint somewhere near 550px wide where we switch to full width drawer. I don't know if we have nice breakpoints pre-defined. If you've never worked with breakpoints, I see there is one for the meeting control bar
https://github.com/ParabolInc/parabol/blob/a4b8bf1ffca6a47c76024fa58f155d9a6cb33e4f/packages/client/components/MeetingControlBar.tsx#L46

87prashant:
660px seems fine to me, Is that okay?


ezgif com-gif-maker (1)

Dschoordsch:
The design looks good to me, @enriquesanchez what do you think?
Peek 2022-06-16 13-12

enriquesanchez:
Looks good to me as well! 🎉 

Thank you for working on this @87prashant!

enriquesanchez (approved):

packages/client/components/EstimatePhaseDiscussionDrawer.tsx
```diff
@@ -29,7 +30,12 @@ const Drawer = styled('div')<{isDesktop: boolean; isOpen: boolean}>(({isDesktop,
   transition: `all 200ms ${BezierCurve.DECELERATE}`,
   userSelect: 'none',
   width: isOpen || !isDesktop ? DiscussionThreadEnum.WIDTH : 0,
-  zIndex: ZIndex.SIDEBAR
+  zIndex: ZIndex.SIDEBAR,
+  [mediaQueryforPoker(Breakpoint.POKER_DISCUSSION_DRAWER)]: {
+    width: '100vw',
+    position: 'fixed',
+    right: isOpen ? '-360px' : '-100vw'
```
Dschoordsch, line 37: -1 I know it works, but having -360px here (which should have been `DiscussionThreadEnum.WIDTH`) as right offset seems wrong. I think the logic should move here:
https://github.com/ParabolInc/parabol/blob/ab9ee86dcc4abea3815ff71f975ea1f1388de4ca/packages/client/components/PokerEstimatePhase.tsx#L25
instead of counteracting the width set there here.

packages/client/types/constEnums.ts
```diff
@@ -37,6 +37,7 @@ export const enum Breakpoint {
   SIDEBAR_LEFT = 1024,
   NEW_MEETING_GRID = 1112,
   NEW_MEETING_SELECTOR = 500,
+  POKER_DISCUSSION_DRAWER = 660,
```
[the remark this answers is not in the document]
87prashant: Yeah, that makes more sense.

packages/client/components/mediaQueryforPoker.tsx
```diff
@@ -0,0 +1,3 @@
+export default function mediaQueryforPoker(breakpoint: number) {
+  return `@media screen and (max-width: ${breakpoint}px)`
```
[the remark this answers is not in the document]
87prashant: Ok

packages/client/components/EstimatePhaseDiscussionDrawer.tsx
87prashant: @Dschoordsch thanks for guidance but I am not able to figure out to do that.
Can you help me with this?

Dschoordsch (approved):

packages/client/components/EstimatePhaseDiscussionDrawer.tsx
Dschoordsch: I checked again in detail and it seems it's not that straight forward as I thought since the `SwipeableDasSidebar` defines the size and is used in other places as well.

Issue #5921, opened 2022-01-21
Support email domain & oauth2 changes


mattkrick:
User story 1
> An organization changes their domain, e.g. `parabol.co` to `parabol.com` or `parabol-team.co`. All users with the old domain should have their email updated to the new domain.

User story 2
> An organization drops google as an email provider. OAuth2 logins no longer work. They should be able to login using an email + password.

After seeing both of these issues, even though they're related to logging in, I don't think there's 1 solution that would fix them both.

Proposed solution for 1
> private mutation accessible by the sales team:

```gql
mutation {
  changeEmailDomain(oldDomain: "parabol.co", newDomain: "parabol-team.co") {
    users {
      email
    }
  }
}
```

Proposed solution for 2
> Remove Google AuthIdentity from User object
> Add a local AuthIdentity if one doesn't exist with a bogus hashed password
> instruct each user to click "forgot password", which will send them into the reset password flow. We could send this automatically from the mutation for each identity removed.

this would be another private mutation like
```gql
mutation {
  removeAuthIdentity(domain: "parabol.co", identityType: "GOOGLE", addLocal: true) {
    users {
      email
    }
  }
}
```

@gcrickman Would these solutions work?

gcrickman:
Ya, I think that would work!

gcrickman:
This [customer](https://app.hubspot.com/contacts/3888472/company/5353593883) 🔒  also reported a domain change with their Google Suite

Dschoordsch:
@mattkrick @gcrickman 
How often does user story 2 occur? Would it make sense to offer them the possibility to do it themselves? Something like:
* When a user enters a Google AuthIdentity email in forgot password show a popup "You're signed up with google, do you want to switch to username + password?"
* Only when they follow the reset link in their email and they set the new password, we convert it to local account.

Alternatively we could also just modify the email in this case to mention that this would convert their Google ID account to email + password.

jordanh:
Prioritized to backlog. Will increase priority as we hear this more often

jordanh:
@mattkrick could you give me a rough burn down / EE for this? I'd like to weigh this against some other things when we get to focusing on admin feature

ackernaut:
Mentioned today in [a user feedback retro](https://action.parabol.co/meet/aG72wCHz5S/discuss/1)

mattkrick:
> When a user enters a Google AuthIdentity email in forgot password show a popup "You're signed up with google, do you want to switch to username + password?"

[...]

PR #6708, opened 2022-06-09
feat: change email domain

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/5921

To test:

Run the following mutation
```gql
mutation {
  changeEmailDomain(oldDomain: "parabol.co", newDomain: "parabol-team.com"){
    ... on ChangeEmailDomainSuccess {
      users {
        email
      }
    }
  }
}
```
See that the email address or domain is updated in the following tables in RethinkDB:

- [ ] Organization
- [ ] TeamMember
- [ ] SAML
- [ ] Invoice, update the billingLeaderEmails field

In Postgres:

- [ ] OrganizationApprovedDomain
- [ ] User


packages/server/graphql/private/mutations/changeEmailDomain.ts
```diff
@@ -0,0 +1,71 @@
...
+  // VALIDATION
+  if (oldDomain === newDomain) {
+    throw new Error('New domain is the same as the old one')
+  }
+  const normalizedNewDomain = newDomain.toLowerCase()
+  const normalizedOldDomain = oldDomain.toLowerCase()
+
+  // RESOLUTION
+  const [updatedUserIds] = await Promise.all([
+    updateUserEmailDomainsToPG(normalizedOldDomain, normalizedNewDomain),
```
nickoferrall, line 19: After writing the mutation, I noticed we have a `getUsersByDomain` query. I could use `getUsersByDomain` to get the relevant `userIds` and then in `updateUserEmailDomainsToPG` and in the `TeamMember` query, get users with those `userIds` rather than search for matching domains/emails. Do you think it's worth making that change or do you think the current implementation does the job?

packages/server/graphql/private/mutations/changeEmailDomain.ts
```diff
@@ -0,0 +1,71 @@
...
+        (row) => ({
+          billingLeaderEmails: row('billingLeaderEmails').map((email) =>
+            r.branch(
+              email.split('@').nth(1).eq(normalizedOldDomain),
+              email.split('@').nth(0).add(`@${normalizedNewDomain}`),
+              email
+            )
+          )
+        }),
+        {returnChanges: true}
```
mattkrick, line 62: -1 no changes needed

mattkrick (changes requested):
nice! got some good string manipulations in there. just a couple fixes

packages/server/postgres/queries/generated/updateDomainsInOrganizationApprovedDomainQuery.ts
```diff
@@ -0,0 +1,31 @@
...
+}
+
+const updateDomainsInOrganizationApprovedDomainQueryIR: any = {"name":"updateDomainsInOrganizationApprovedDomainQuery","params":[{"name":"oldDomain","transform":{"type":"scalar"},"codeRefs":{"used":[{"a":127,"b":135,"line":5,"col":30},{"a":166,"b":174,"line":6,"col":16}]}},{"name":"newDomain","transform":{"type":"scalar"},"codeRefs":{"used":[{"a":139,"b":147,"line":5,"col":42}]}}],"usedParamSet":{"oldDomain":true,"newDomain":true},"statement":{"body":"UPDATE \"OrganizationApprovedDomain\"\nSET domain = replace(domain, :oldDomain, :newDomain)\nWHERE domain = :oldDomain","loc":{"a":61,"b":174,"line":4,"col":0}}};
+
+/**
+ * Query generated from SQL:
+ * ```
+ * UPDATE "OrganizationApprovedDomain"
+ * SET domain = replace(domain, :oldDomain, :newDomain)
+ * WHERE domain = :oldDomain
```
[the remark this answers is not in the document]
nickoferrall: When would the domain be an email address out of interest? I thought the domain would always be the bit after the `@`, so `parabol.co`. I went with `domain = :oldDomain` originally, to be more precise, but `LIKE` works too, so I've pushed an update!

packages/server/graphql/private/mutations/changeEmailDomain.ts
nickoferrall: Thanks for the explanation! I updated the `WHERE` clause but I can see in pgadmin that it's a `SEQ_SCAN`. Same thing happens when `WHERE domain = :oldDomain`. It looks like the scan is happening because of the concatenation which seems to be necessary for the string manipulation. 

seq-scan

packages/server/postgres/queries/generated/updateDomainsInOrganizationApprovedDomainQuery.ts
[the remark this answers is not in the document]
mattkrick: let's say enterprise.com wants to lock down their domain.
but they want bob@gmail.com to join a meeting.
they don't want to open the floodgates to any gmail domain, just bob in particular

packages/server/graphql/private/mutations/changeEmailDomain.ts
mattkrick: ah, i think i must've confused tables.

`User.domain` is a btree index & will never contain an email, so you can use strict equality (`domain = :oldDomain`)

`OrganizationApproval.domain` is a btree & requires using `LIKE` so it will be a seq scan, which i guess is fine because it'll be a pretty small table.

packages/server/postgres/queries/src/updateUserEmailDomainsQuery.sql
```diff
@@ -0,0 +1,8 @@
+/*
+  @name updateUserEmailDomainsQuery
+*/
+
+UPDATE "User" SET email =
+CONCAT(LEFT(email, POSITION('@' in email)), :newDomain::VARCHAR)
+WHERE domain LIKE :oldDomain
```
mattkrick, line 7: -1 `domain = :oldDomain`

packages/server/graphql/private/mutations/changeEmailDomain.ts
mattkrick: PS, make sure you run that on a table that has a sufficient amount of rows like the staging DB. if there are only 5 rows, it's gonna use a seq scan regardless

Screen Shot 2022-06-10 at 11 05 49 AM

mattkrick (changes requested):

Issue #6709, opened 2022-06-09
Remove auth identity 


nickoferrall:
See User Story 2 in https://github.com/ParabolInc/parabol/issues/5921

> An organization drops google as an email provider. OAuth2 logins no longer work. They should be able to login using an email + password.

### AC

- [ ] Create a private mutation that removes the Google auth identity from the User object
- [ ] Add a local AuthIdentity if one doesn't exist with a bogus hashed password
- [ ] Instruct each user to click "forgot password", which will send them into the reset password flow. We could send this automatically from the mutation for each identity removed.

EE: 5 points

PR #6713, opened 2022-06-09
feat: remove auth identity

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6709

### To test

- [ ] Copy & paste the following mutation

```gql
mutation {
  removeAuthIdentity(domain: "parabol.co", identityType: GOOGLE, addLocal: true){
    ... on RemoveAuthIdentitySuccess {
      users {
        email
        identities {
          isEmailVerified
          type
        }
      }
    }
  }
}
```

- [ ] See that the `GOOGLE` auth identities have been removed from the `User` table
- [ ] You're no longer able to sign in with that Google login
- [ ] You receive a forgot password email and can reset your password
- [ ] You can login with the Parabol email & newly created password

### DIFF
--- a/packages/server/graphql/graphql.ts
+++ b/packages/server/graphql/graphql.ts
@@ -18,6 +18,7 @@ export interface GQLContext {
 export interface InternalContext {
   dataLoader: DataLoaderWorker
   authToken: AuthToken
+  ip: string
   // not present if called ad-hoc
   socketId?: string
 }
--- a/packages/server/graphql/private/mutations/removeAuthIdentity.ts
+++ b/packages/server/graphql/private/mutations/removeAuthIdentity.ts
@@ -1,0 +1,19 @@
+import {getUserId} from '../../../utils/authorization'
+import {MutationResolvers} from '../../public/resolverTypes'
+
+const removeAuthIdentity: MutationResolvers['removeAuthIdentity'] = async (
+  _source,
+  {},
+  {authToken, dataLoader, socketId: mutatorId}
+) => {
+  const viewerId = getUserId(authToken)
+  const now = new Date()
+
+  // VALIDATION
+
+  // RESOLUTION
+  const data = {}
+  return data
+}
+
+export default removeAuthIdentity
--- a/packages/server/graphql/private/mutations/removeAuthIdentity.ts
+++ b/packages/server/graphql/private/mutations/removeAuthIdentity.ts
@@ -21,16 +21,20 @@ const removeAuthIdentity: MutationResolvers['removeAuthIdentity'] = async (
   const dummyPassword = 'dummy'
   const dummyHashedPassword = await bcrypt.hash(dummyPassword, Security.SALT_ROUNDS)
   const updateUsersPromises = users.map((user) => {
-    const {id: userId} = user
-    const identityId = `${userId}:${AuthIdentityTypeEnum.LOCAL}`
-    const newIdentity = new AuthIdentityLocal({
-      hashedPassword: dummyHashedPassword,
-      id: identityId,
-      isEmailVerified: false
-    })
-    const filteredIdentities = user.identities.filter((identity) => identity.type !== identityType)
-    const newIdentities = addLocal ? [...filteredIdentities, newIdentity] : filteredIdentities
-    return updateUser({identities: newIdentities}, userId)
+    const {id: userId, identities} = user
+    const filteredIdentities = identities.filter((identity) => identity.type !== identityType)
+    if (!filteredIdentities.length || addLocal) {
+      const identityId = `${userId}:${AuthIdentityTypeEnum.LOCAL}`
+      const newIdentity = new AuthIdentityLocal({
+        hashedPassword: dummyHashedPassword,
+        id: identityId,
+        isEmailVerified: false
+      })
+      const newIdentities = [...filteredIdentities, newIdentity]
+      return updateUser({identities: newIdentities}, userId)
+    } else {
+      return updateUser({identities: filteredIdentities}, userId)
+    }
   })
   const result = await Promise.all(updateUsersPromises)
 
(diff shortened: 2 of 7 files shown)


packages/server/graphql/graphql.ts
```diff
@@ -18,6 +18,7 @@ export interface GQLContext {
 export interface InternalContext {
   dataLoader: DataLoaderWorker
   authToken: AuthToken
+  ip: string
```
nickoferrall, line 21: I can see that I am able to get my IP from the InternalContext, so I thought I'd add it here. Alternatively, I could make the `ip` arg optional in https://github.com/ParabolInc/parabol/blob/6f207a89341461d03180a8cd157d424e426c7155/packages/server/graphql/private/mutations/removeAuthIdentity.ts#L65

packages/server/graphql/private/mutations/removeAuthIdentity.ts
```diff
@@ -0,0 +1,80 @@
...
+      const newIdentities = [...filteredIdentities, newIdentity]
+      return updateUser({identities: newIdentities}, userId)
+    } else {
+      return updateUser({identities: filteredIdentities}, userId)
+    }
+  })
+  await Promise.all(updateUsersPromises)
+
+  // send forgot password email
+  users.forEach(async ({email}) => {
```
nickoferrall, line 51: If a domain update affects hundreds of users and mailgun is being slow, could this mutation timeout? We want to invalidate all the tokens, and then insert the new password request, and then send the email, so I think we need to `await` each time.

packages/server/graphql/private/mutations/removeAuthIdentity.ts
```diff
@@ -0,0 +1,80 @@
...
+  if (!users.length) {
+    return {error: {message: 'No user emails match that domain'}}
+  }
+
+  // RESOLUTION
+  const updateUsersPromises = users.map(async (user) => {
+    const {id: userId, identities} = user
+    const filteredIdentities = identities.filter((identity) => identity.type !== identityType)
+    if (!filteredIdentities.length || addLocal) {
+      const identityId = `${userId}:${AuthIdentityTypeEnum.LOCAL}`
```
nickoferrall, line 34: The `identityId` will look a lil strange as it includes both the google auth & local strings, e.g. `google-oauth2|e2N8F8q7II:LOCAL`. I think that's OK as long as the `id` is unique as we're using the `type` field rather than the `id` to know which type it is.

packages/server/graphql/graphql.ts
mattkrick: since this is on the private schema, the ip is gonna be a sales person. i'd hate to rate limit them, but i don't think it'll be a problem either way, so this is probably OK

packages/server/graphql/private/mutations/removeAuthIdentity.ts
```diff
@@ -0,0 +1,80 @@
...
+
+const randomBytes = util.promisify(crypto.randomBytes)
+
+const removeAuthIdentity: MutationResolvers['removeAuthIdentity'] = async (
+  _source,
+  {domain, identityType, addLocal},
+  {ip}
+) => {
+  // VALIDATION
+  const users = await getUsersByDomain(domain)
```
mattkrick, line 24: -1 normalize args first

packages/server/graphql/private/typeDefs/removeAuthIdentity.graphql
```diff
@@ -0,0 +1,33 @@
...
+"""
+Return value for removeAuthIdentity, which could be an error
+"""
+union RemoveAuthIdentityPayload = ErrorPayload | RemoveAuthIdentitySuccess
+
+type RemoveAuthIdentitySuccess {
+  """
+  The users who had their auth identity removed
+  """
+  users: [User]!
```
mattkrick, line 32: -1 make non null `[User!]!`

packages/server/graphql/private/mutations/removeAuthIdentity.ts
```diff
@@ -0,0 +1,80 @@
...
+  if (!users.length) {
+    return {error: {message: 'No user emails match that domain'}}
+  }
+
+  // RESOLUTION
+  const updateUsersPromises = users.map(async (user) => {
+    const {id: userId, identities} = user
+    const filteredIdentities = identities.filter((identity) => identity.type !== identityType)
```
mattkrick, line 32: -1 what if they already have a local & google identity & `addLocal` is true?
business logic should be to add local only if local doesn't already exist

packages/server/graphql/private/mutations/removeAuthIdentity.ts
mattkrick: -1 `forEach` isn't equipped to handle async functions.
consider the following 

```ts
[1, 2, 3].forEach(async () => {
    await new Promise((resolve) =>
      setTimeout(() => {
        console.log('resolving')
        resolve()
      }, 1000)
    )
  })
console.log('do i print before or after?')
```
what you want to do is call `Promise.all(users.map(...)`

here's what I'd do...
1. instead of updating the user in `updateUsersPromises`, have that function return an array of users with updated identities.
2. extract out the chunk that was copied from `emailPasswordReset`
3. call `Promise.all(usersWithUpdatedIdentities.map((user) => processEmailPasswordReset(ip, user.identities, user.id...))`

mattkrick (changes requested):

packages/server/graphql/private/mutations/removeAuthIdentity.ts
nickoferrall: If the user only has a Google account and `addLocal` is false, should we leave identities empty or create a new local one? Leaving it empty would mean they can't use forget password or login

mattkrick: i sure how they set `addLocal: true` 😬 
i think that's OK. the only use case for `addLocal: false` is if they move to SSO & don't want people to login via gmail or password. Or if we add a new oauth2 strategy & we just want to remove the 1.

mattkrick (approved):

PR #6722, opened 2022-06-10
fix(jira): Fix add missing jira field new algorithm

### LINKED ISSUE #6675: Jira: not always possible to fix a field for a company-managed project if account has a lot of screens
Following up on the issue here #6589

Here is the another case where the fix field functionality might return a timeout error:

- If Jira account has a lot of projects, it will have a lot of screens.
- In add missing Jira field functionality we only fetching first 100 screens.
- So it is possible that required screen is not fetched.
- Side effect: If we keep iterating a lot of screens without success our API fails with timeout.

It is not possible to filter screens using Jira API or fetch only screens that belong to a specific project. So I suggest to do the following:

### Acceptance Criteria 
- Fetch larger amount of screens, e.g. 2000
- Iterate only most possible screens to avoid timeouts and show meaningful error

**Estimated effort:** 4h

igorlesnenko:
Fixes #6675

Previous PR: #6676

* Looks like Jira API does not allow fetching more than 100 items, set `batchSize` to 100
* Changed `evaluateProbability` function to make it more precise, otherwise it sometimes picks screens that belong to different projects
* Fixed hardcoded value

### DIFF
--- a/packages/server/graphql/mutations/addMissingJiraField.ts
+++ b/packages/server/graphql/mutations/addMissingJiraField.ts
@@ -91,20 +91,21 @@ const addMissingJiraField = {
       `Adding missing Jira field, fieldId: ${fieldId}, cloudId: ${cloudId}, projectKey: ${projectKey}`
     )
 
-    const batchSize = 1000
+    // Jira does not allow fetching more than 100 screens
+    const batchSize = 100
     const screensResponse = await manager.getScreens(cloudId, batchSize)
     const screens: JiraScreen[] = []
     if (screensResponse instanceof Error || screensResponse instanceof RateLimitError) {
       return standardError(screensResponse)
     }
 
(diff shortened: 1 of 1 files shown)


packages/server/graphql/mutations/addMissingJiraField.ts
```diff
@@ -123,8 +124,9 @@ const addMissingJiraField = {
 
     // we're trying to guess what's the probability that given screen is assigned to an issue project
     const evaluateProbability = (screen: JiraScreen) => {
-      if (screen.name.startsWith(projectKey) && screen.name.includes('Default')) return 1
-      if (screen.name.includes(projectKey)) return 0.9
+      const nameProjectPrefix = `${projectKey}:`
```
mattkrick, line 127: i assume the colon cuts down on the number of false positives?

mattkrick, line 127: i assume the colon cuts down on the number of false positives?

PR #6725, opened 2022-06-13
fix: handle duplicates in changeEmailDomain

nickoferrall:
When executing the new `changeEmailDomain` mutation, I got an error saying that the unique email constraint would be violated. This happens in the following scenario:
- `bob@olddomain.com` and `bob@newdomain.com` both exist
- We execute the `changeEmailDomain` mutation to change `olddomain` to `newdomain`
- `bob@olddomain.com` can't be updated as there would be two `bob@newdomain.com`

We can use `ON CONFLICT DO NOTHING` in `INSERT` statements, but `UPDATE`s look trickier. I looked into a `WHERE` condition that would check if the domain equalled the old domain and the name part of the email concatenated with the new domain didn't exist in the email column, but I didn't find a good solution. 

Instead, I think it would be more readable to do this logic in `changeEmailDomain.ts` and then pass the `userIds` into the query.


mattkrick (changes requested):

packages/server/graphql/private/mutations/changeEmailDomain.ts
```diff
@@ -14,9 +15,23 @@ const changeEmailDomain: MutationResolvers['changeEmailDomain'] = async (
...
   }
 
+  const [oldDomainUsers, newDomainUsers] = await Promise.all([
+    getUsersbyDomain(normalizedOldDomain),
+    getUsersbyDomain(normalizedNewDomain)
+  ])
+  const filteredUserIdsByOldDomain = oldDomainUsers
+    .filter((user) => {
+      const emailName = user.email.split('@')[0]
+      const nameExistsInNewDomain = newDomainUsers.find(
```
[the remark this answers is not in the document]
nickoferrall: > this'll fail if the name is included in the tld

As we're checking everything before the `@`, I'm not sure why the name being in the domain would cause it to fail? Still, as we want to return the `usersNotUpdated`, which I think makes sense, I've pushed an update that created a `proposedNewEmails` array.

packages/server/postgres/queries/generated/updateUserEmailDomainsQuery.ts
```diff
@@ -3,8 +3,8 @@ import { PreparedQuery } from '@pgtyped/query';
 
 /** 'UpdateUserEmailDomainsQuery' parameters type */
 export interface IUpdateUserEmailDomainsQueryParams {
+  userIds: readonly (string | null | void)[];
```
mattkrick, line 6: can you delete this file from git? (the whole directory should be ignored...)

packages/server/graphql/private/mutations/changeEmailDomain.ts
```diff
@@ -14,9 +15,33 @@ const changeEmailDomain: MutationResolvers['changeEmailDomain'] = async (
...
+    throw new Error(`No users found with oldDomain: ${oldDomain}`)
+  }
+
+  const makeNewEmail = (email: string) => `${email.split('@')[0]}@${newDomain}`
+  const newDomainUserEmails = newDomainUsers.map((user) => user.email)
+  const proposedNewUserEmails = oldDomainUsers.map((user) => makeNewEmail(user.email))
+  const duplicateUserEmails = proposedNewUserEmails.filter((userEmail) =>
+    newDomainUserEmails.includes(userEmail)
+  )
+  const usersNotUpdatedIds = [] as string[]
```
mattkrick, line 33: +1 L27-40 works, but it's a little hard to read.
Could you take a stab at simplifying it to make it a little more maintainable?

<details>
<summary> Click for Spoiler</summary>

```ts
const newDomainUserEmails = newDomainUsers.map((user) => user.email)
  const userIdsToUpdate = oldDomainUsers.filter(({email}) => !newDomainUserEmails.includes(email)).map(({id}) => id)
  const userIdsNotUpdated = oldDomainUsers.filter(({email}) => newDomainUserEmails.includes(email)).map(({id}) => id)
```
</details>

mattkrick (changes requested):

packages/server/graphql/private/mutations/changeEmailDomain.ts
nickoferrall: Sure thing, updated! What do ya think?

packages/server/graphql/private/mutations/changeEmailDomain.ts
```diff
@@ -10,13 +11,39 @@ const changeEmailDomain: MutationResolvers['changeEmailDomain'] = async (
   // VALIDATION
   const normalizedNewDomain = newDomain.toLowerCase().trim()
   const normalizedOldDomain = oldDomain.toLowerCase().trim()
-  if (oldDomain === newDomain) {
+  if (normalizedOldDomain === normalizedNewDomain) {
     throw new Error('New domain is the same as the old one')
   }
 
+  if (normalizedOldDomain.includes('@') || normalizedNewDomain.includes('@')) {
```
mattkrick, line 18: +2 nice UX!

mattkrick (approved):

PR #6736, opened 2022-06-14
fix: updates caniuse-lite version in yarn lock file

JimmyLv:
# Description

To avoid Dev Server always prompting warnings like this,

WebStorm-parabol  ~ccodingstyleparabol  – …package json-003196-20220614

Why you should do it regularly: https://github.com/browserslist/browserslist#browsers-data-updating

`npx browserslist@latest --update-db` updates `caniuse-lite` version
in your npm, yarn or pnpm lock file. This update will bring data about
new browsers to polyfills tools like Autoprefixer or Babel and reduce
already unnecessary polyfills.

You need to do it regularly for three reasons:

1. To use the latest browser’s versions and statistics in queries like
   `last 2 versions` or `>1%`. For example, if you created your project
   2 years ago and did not update your dependencies, `last 1 version`
   will return 2 year old browsers.
2. Actual browsers data will lead to using less polyfills. It will reduce
   size of JS and CSS files and improve website performance.
3. `caniuse-lite` deduplication: to synchronize version in different tools.

## Final checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,8 @@
     "nanoid": "^3.1.31",
     "@xmldom/xmldom": "^0.7.0",
     "node-gyp": "^8.4.1",
-    "simple-get": "^4.0.1"
+    "simple-get": "^4.0.1",
+    "caniuse-lite": "~1.0.0"
   },
   "devDependencies": {
     "@babel/preset-env": "7.16.0",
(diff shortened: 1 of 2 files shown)


JimmyLv:
Did some research and [here also mentioned using "resolutions"](https://stackoverflow.com/questions/56647770/react-create-app-get-error-about-browserslist-error#comment100063074_56648454), but not sure if [yarn "resolutions"](https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/#toc-why-would-you-want-to-do-this) will automatically force-resolve the latest version every time. maybe it needs the package we're using, to have an update for the caniuse-lite dependency, then to trigger?

I added "resolutions" first, so let's watch it later, the browser "browserslist" [caniuse-db](https://github.com/Fyrd/caniuse#readme) shouldn't been updated so frequent, 🤣

mattkrick:
we shall find out together!

PR #6752, opened 2022-06-14
chore(dx): delete remaining generated pg queries

mattkrick:
# Description

we deleted all generated fragments, but somehow these are still alive.
I'm not sure how these didn't get deleted 🤷


mattkrick:
this was likely a result of a merge conflict, so merging this directly as soon as tests finish

PR #6759, opened 2022-06-15
chore: Refresh HubSpot chat widget on page changes

### LINKED ISSUE #6718: Add HubSpot chatbot
There are two ways we can add chat and I'm not sure which will work best:

1. We create a custom component that uses the [HubSpot Chat API](https://developers.hubspot.com/docs/api/conversation/chat-widget-sdk)
2. We use the embed code (below) and update the settings to have it refresh when the URL changes

Note: On the HubSpot site we can set criteria for which pages we want chat to show up on. BUT because in the app the page never actually refreshes, the chat icon never goes away. To combat this, the chat API has a way to refresh _the bot_ whenever the URL changes (see `widget.refresh` in the docs)

HubSpot Embed Code
`<script type="text/javascript" id="hs-script-loader" async defer src="//js-na1.hs-scripts.com/3888472.js"></script>`

AC:
- The chat widget successfully displays the _correct_ bot (it should only display the bot that is set for that URL)
- The chat widget _does not display_ when it is on a URL that is not included in the chatbot settings

Dschoordsch:
Fixes #6718 

HubSpot matches the URL to decide which chat to show. For SPAs it needs
manual refreshing on page change.

## Demo
https://www.loom.com/share/21d239c305fa4791b4d130dc4bd96333

## Testing scenarios

- Add Hubspot to your local [devTemplate.html](https://github.com/ParabolInc/parabol/blob/4f1c9d807d01ca027f9f300af114a67a2f2992af/devTemplate.html#L7)
  (see #6718)
- Remove non-production return from Analytics page
  https://github.com/ParabolInc/parabol/blob/4f1c9d807d01ca027f9f300af114a67a2f2992af/packages/client/components/AnalyticsPage.tsx#L83-L85
- [ ] load a meeting summary on 'localhost:3000/new-summary' which shows a chat bot as configured on HubSpot
- [ ] switching to another page which changes the URL also makes the chat disappear

## Final checklist

### DIFF
--- a/packages/client/components/AnalyticsPage.tsx
+++ b/packages/client/components/AnalyticsPage.tsx
@@ -25,6 +25,11 @@ const query = graphql`
 declare global {
   interface Window {
     analytics: SegmentAnalytics.AnalyticsJS
+    HubSpotConversations?: {
+      widget?: {
+        refresh?: () => void
+      }
+    }
   }
 }
 
@@ -144,6 +149,11 @@ const AnalyticsPage = () => {
       )
     }, TIME_TO_RENDER_TREE)
   }, [isSegmentLoaded, pathname])
+
+  // We need to refresh the chat widget so it can recheck the URL
+  useEffect(() => {
+    window.HubSpotConversations?.widget?.refresh?.()
+  }, [pathname])
 
   useEffect(() => {
     if (!datadogEnabled) {


igorlesnenko (approved):
works great

PR #6773, opened 2022-06-16
feat: add default tasks to new team

### LINKED ISSUE #6719: Add default tasks to new team
When a user creates a new team, the /team view should have two tasks assigned to that user
1. "Congratulations to your new team, you can now start inviting your team mates[link to open invite modal]"
2. "Start your first meeting in your new team [link to meeting lobby]"

**Events to measure**
- Clicks on the cards
- Cards moved to done
- Cards set to archived

JimmyLv:
# Description

resolved #6719

## Demo

https://user-images.githubusercontent.com/4997466/182640794-9a17f482-f8b6-4352-8617-32c408fa6b8c.mp4

## Testing scenarios

- [x] create new user with the new onboarding team by default
  - create new account
  - check /tasks page
  - see 3 seed tasks, including 1 more `invite` task

- [x] create new team
  - check /team page
  - see 1 seed task, including `invite` task, click to open a new tab, then open the invite modal.

- [x] create new team under new org
  - check /team page
  - see 1 seed task, including `invite` task, click to open a new tab, then open the invite modal.

## Final checklist


mattkrick:
I don't think we should merge this feature, see comments here: https://github.com/ParabolInc/parabol/issues/6719#issuecomment-1165948330

JimmyLv:
Based on the latest discussion at https://github.com/ParabolInc/parabol/issues/6719#issuecomment-1199362998, I implemented an `invite` task and added the related tracking events, for the Amplitude funnel.

packages/client/components/MassInvitationTokenLink.tsx
```diff
@@ -77,13 +78,22 @@ const MassInvitationTokenLink = (props: Props) => {
   const displayToken = isTokenValid ? token : '············'
   const linkLabel = `${window.__ACTION__.prblIn}/${displayToken}`
   const url = getMassInvitationUrl(displayToken)
+
+  const {viewerId} = atmosphere
+  const handleCopy = () => {
+    SendClientSegmentEventMutation(atmosphere, 'Copied Team Invite Link', {
```
tianrunhe, line 84: We should follow [Segment naming convention](https://segment.com/academy/collecting-data/naming-conventions-for-clean-data/) (Objection-action framework), so let's rename the event name to "Team Invitation Link Copied"

packages/client/components/AddTeamMemberModal.tsx
```diff
@@ -175,6 +176,10 @@ const AddTeamMemberModal = (props: Props) => {
 
   const sendInvitations = () => {
     if (invitees.length === 0) return
+    SendClientSegmentEventMutation(atmosphere, 'Clicked Send Invitation Button', {
```
tianrunhe, line 179: We should follow [Segment naming convention](https://segment.com/academy/collecting-data/naming-conventions-for-clean-data/) (Objection-action framework), so let's rename the event name to "Send Invitation Button Clicked"

tianrunhe (changes requested):
Note: I only reviewed metrics related changes

packages/client/components/MassInvitationTokenLink.tsx
JimmyLv: Good to know, thanks @tianrunhe , I just already updated.

tianrunhe (approved):
Approve from metrics side!

igorlesnenko (approved):
This looks good to me. Only one question.
When I click invite your teammates link, it opens same URL in a new tab, which feels weird. Is it difficult to make it to only open a modal?

JimmyLv:
Thanks @igorlesnenko , it's really not technically possible to do a direct trigger the invite modal, as has been discussed before. https://github.com/ParabolInc/parabol/issues/6719#issuecomment-1179906042

It's been mentioned that the "out-app" UX is not good enough, but I can't think of a better way to do it at the moment. https://github.com/ParabolInc/parabol/issues/6719#issuecomment-1205852234

> These cards ideally shouldn’t open in a new tab. The links don’t support direct interactions. Any way to make this a good UX without links?

igorlesnenko:
@JimmyLv I think we have plans to migrate from draft-js everywhere at some point

mattkrick:
Deferring to product manager @jordanh

jordanh:
I'm reviewing this now. I'll have a proposal ~Monday or ~Tuesday for how I suggest we proceed

jordanh:
Ok, thank you for standing by.

When @Matt and I gave feedback on this design idea, we failed you folks in several ways. We asserted that this approach wouldn’t work. Given that we had implemented a similar seed task cards before and exposed it to statistically significant numbers, we have pretty good reason to believe this won’t work. That doesn’t mean we’re correct, however. It just means we’re very likely to be correct.

Seed tasks were [first added here](https://github.com/ParabolInc/parabol/commit/97553a1a108a482404061e664dd54224673a7047) in 2016, and iterated on for 3 years (!) until they were [removed here in 2019](https://github.com/ParabolInc/parabol/commit/0e9d715400ad004dfb84f24da4cdbc7ed4135e13). The complete history can be discovered via:

`git log --follow -- src/server/graphql/mutations/helpers/addSeedTasks.js`

I’d love to show you the actual data for how poorly this performed, but sadly it’s rolled off of Amplitude’s data retention window.

Before merging this, I want us to make sure that we're in a position as a team to understand what the effects are. In other words, I'd like to make certain that @mattkrick and my data are, either either out of date or that the changes to the copy are statistically significant. 

Could squad members answer the following?

1. Is the hypothesis of this experiment that we believe the funnel conversion from creating a new team to sending invitations will increase? And, creating a new team to adding an integration will increase?
2. What are those rate today? Over what time period do we have accurate data?
3. If the rate changes after shipping this change, what do you believe that tells you? What time frame would give us good data here (rule of thumb: roughly n=400 is a 95% confidence interval)
4. What might our next steps be (leave the task card? Improve the new team or team dashboard journey some other way?)

If you come with me on this journey, I'd like to highlight this as an example of what a good experiment could look like here for other squads

Dschoordsch:
See issue

jordanh:
@Dschoordsch @JimmyLv, how would you like to proceed?

Dschoordsch:
@jordanh I checked the numbers and they suggest that most people don't create a new team, thus this is redundant for user activation. I closed the issue.

Issue #6688, opened 2022-06-07
Standups: Improve UX when submitting standup responses


enriquesanchez:
Currently when submitting a standup response, there is no clear action to submit a response. Of the common ways for submitting a text input, none are present:

1. Pressing <kbd>Enter</kbd>: this does not work how we would expect based on the behavior of reflection cards, instead it creates a new text paragraph inside the response.
2. Submit button: there is no button to submit the response, opposite of how submitting a comment in a thread works.

[This has already come up in user feedback sessions](https://www.loom.com/share/d958ddc72d4a4dfa928a06244069e9b2#), where folks are confused by the lack of a clear action, and instead think the `Reply` button is the action for submitting.

The only way for users to "leave" editing mode of a response card is to click elsewhere.

In addition to a lack of clear submit action, the presence of both the `Reply` and reacti buttons can cause additional confusion, particularly because at the moment, there is nothing to reply or react to. 

<img width="300" alt="Captura de Pantalla 2022-06-07 a la(s) 4 34 37 p m" src="https://user-images.githubusercontent.com/3276087/172489745-42d1124f-7dec-4dd3-ba76-195c42d33b55.png">

## Proposal

Let's add a `Submit` button and show the reply and reacti button only once a response has been submitted:

### When it's empty:

<img width="400" alt="Captura de Pantalla 2022-06-07 a la(s) 4 57 45 p m" src="https://user-images.githubusercontent.com/3276087/172489875-f900c0eb-771c-4693-8aed-a76f600368d1.png">

The response card shows the placeholder text and a disabled `Submit` button.

### When a response is entered

<img width="400" alt="Captura de Pantalla 2022-06-07 a la(s) 4 58 34 p m" src="https://user-images.githubusercontent.com/3276087/172489998-da24b069-1897-4947-9d5d-74c94df643cc.png">

The `Submit` button becomes active and the user can click it now.

### When a response has been submitted

<img width="400" alt="Captura de Pantalla 2022-06-07 a la(s) 4 59 09 p m" src="https://user-images.githubusercontent.com/3276087/172490083-5c9ae924-7a66-4f2a-8391-0f0263905905.png">

The `Submit` button goes away, and now we can display the `Reply` and reacti buttons.

### Replies and reactions

[...]

PR #6778, opened 2022-06-16
feat(standups): Response submission UX redesign

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/6688
Update the UX for response submission from onBlur to explicit "submit"/"update", and allow users to undo any in-progress edits with a "cancel" button.

Recommend reviewer sets "ignore whitespace" for the change in `packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx`

## Demo

https://www.loom.com/share/8f09c3adb3ec42089de0b05831cc984f

## Testing scenarios
(check that this matches designs for each step)
- [ ] Start a new meeting
- [ ] Type out a new response
- [ ] Click "Submit"
- [ ] Refresh the page to confirm that the mutation was persisted
- [ ] Click into the response (should have no buttons)
- [ ] Make a minor edit (buttons should appear)
- [ ] Click "cancel" - content should revert to original submission, with no edit timestamp
- [ ] Try to delete the entire response - "update" should be disabled.  Then click cancel.
- [ ] Make a minor edit, and click update - the card should show as edited, and refreshing should show that the mutation persisted.
- [ ] Sign in as a different user - confirm that buttons do not appear on non-viewer cards

## Final checklist

### DIFF
--- a/packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
+++ b/packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
@@ -258,7 +258,7 @@ const TeamPromptResponseCard = (props: Props) => {
               readOnly={!isCurrentViewer}
               placeholder={'Share your response...'}
             />
-            {!!contentJSON && (
+            {!!plaintextContent && (
               <ResponseCardFooter>
                 <StyledReactjis reactjis={reactjis} onToggle={onToggleReactji} />
                 <ReplyButton onClick={() => onSelectDiscussion()}>
--- a/packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
+++ b/packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
@@ -258,19 +258,21 @@ const TeamPromptResponseCard = (props: Props) => {
               readOnly={!isCurrentViewer}
               placeholder={'Share your response...'}
             />
-            <ResponseCardFooter>
-              <StyledReactjis reactjis={reactjis} onToggle={onToggleReactji} />
-              <ReplyButton onClick={() => onSelectDiscussion()}>
-                {replyCount > 0 ? (
-                  <>
-                    <TeamPromptRepliesAvatarList edgesRef={discussionEdges} />
-                    {replyCount} {plural(replyCount, 'Reply', 'Replies')}
-                  </>
-                ) : (
-                  'Reply'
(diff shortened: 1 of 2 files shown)


packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
```diff
@@ -258,19 +258,21 @@ const TeamPromptResponseCard = (props: Props) => {
...
-                  <>
-                    <TeamPromptRepliesAvatarList edgesRef={discussionEdges} />
-                    {replyCount} {plural(replyCount, 'Reply', 'Replies')}
-                  </>
-                ) : (
-                  'Reply'
-                )}
-              </ReplyButton>
-            </ResponseCardFooter>
+            {!!contentJSON && (
```
BartoszJarocki, line 261: +1 to make it easier to read and understand, can we declare a variable for this? `hasResponse`? also, it'd be better to check `plaintextContent` as populated `contentJSON` still can have an empty response

packages/client/components/promptResponse/PromptResponseEditor.tsx
```diff
@@ -80,22 +108,48 @@ const PromptResponseEditor = (props: Props) => {
...
-    </StyledEditor>
+    <>
+      <StyledEditor>
+        <EditorContent editor={editor} />
+      </StyledEditor>
+      {!readOnly && (
+        // The render conditions for these buttons *should* only be true when 'readOnly' is false, but let's be explicit
+        // about it.
+        <SubmissionButtonWrapper>
+          {!!content && isEditing && (
```
BartoszJarocki, line 136: +1 `hasContent` would make this easier to read

BartoszJarocki (approved):
lgtm, just a few nits about potentially making code easier to understand 👍

packages/client/components/promptResponse/PromptResponseEditor.tsx
jmtaber129: @BartoszJarocki could you elaborate a bit on the desired change here?  `content` here is the actual editor content, not a boolean.

jmtaber129: I actually have a weak preference towards leaving this as `!!content` since the use of `!!` is common enough for converting truthy/falsy things to true/false

Issue #6762, opened 2022-06-15
Implement domain stats snackbar


nickoferrall:
As mentioned in https://github.com/ParabolInc/parabol/issues/6681, we want to emit a nagging snackbar to ask relevant users to visit the `/usage` page. 

### AC

- [ ] Store a `hasSeenUsage` boolean field on the User object
- [ ] In `MeetingSelector`, `NewMeeting`, and `DashboardRoot`, add a `useSnackbar` hook that works in a similar way to (or modifies) [useSnackNag](https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/client/hooks/useSnackNag.ts)
- [ ] If the `insights` featureFlag is true, `hasSeenUsage` is false, and is not `/usage` page, the snackbar is emitted
- [ ] When the user visit the `/usage` page, a mutation is executed that updates `hasSeenUsage` to true
- [ ] When the snackbar is emitted or dismissed send an event to Segment
- [ ] If the snackbar is dismissed, it nags the user every 3 minutes until they visit the `/usage` page

**EE**: 5 hours

mattkrick:
>  Store a hasSeenUsage boolean field on the User object

what functionality is dependent on this variable?

nickoferrall:
> what functionality is dependent on this variable?

Whether we should the snackbar. `hasSeenUsage` is false until the user visits `/usage`, when it's updated to true. If `hasSeenUsage` is false, we show the nagging snackbar. If it's true, we don't.

mattkrick:
ahhhhh so if they goto /usage once, they never see the snackbar again?

nickoferrall:
Yep, exactly!

mattkrick:
gotcha! i was thinking about nagging em even more 🤣 

there's a few places we could store `hasSeenUsage`, ranging from most naggy to least naggy

- never store hasSeenUsage, they can only escape the snackbar by clicking dismiss every 3 minutes. this is what we currently use for the useSnackNag.
- store it in the client schema (ie per-session)
- store it in local storage (ie per-device)
- store it on the User object (ie per-user)

my .02 is to not keep it on the user object since that involves the server & a DB migration, but i'll let you & @ackernaut  @avivapinchas decide

nickoferrall:
[...]

PR #6785, opened 2022-06-17
feat(insights): usage snack nag

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6762

This PR shows a nagging snackbar to users who have the `insights` feature flag turned on and whose `tier` is less than their `suggestedTier`, e.g. `tier` is `personal` and `suggestedTier` is `pro`.

### To test

- [ ] Change `NAG_EVERY` to `2s` in `useUsageSnackNag`
- [ ] See that the snackbar pops up throughout the dashboard and meeting selector but not in a meeting
- [ ] If you dismiss the snackbar by clicking on it (not on the CTA), it reappears after `NAG_EVERY` 
- [ ] Sends an event to Segment when the snackbar has been emitted and when the CTA has been clicked on

Note: if you refresh the page or dismiss the snackbar and proceed to the meeting selector, it'll show up right away rather than after 3 minutes. I could add some logic to prevent this, but as [the idea](https://github.com/ParabolInc/parabol/issues/6762#issuecomment-1157976095) is for the snackbar to be essentially persistent, I think that logic would be unnecessary.

### DIFF
--- a/packages/client/hooks/useUsageSnackNag.ts
+++ b/packages/client/hooks/useUsageSnackNag.ts
@@ -1,8 +1,12 @@
+import graphql from 'babel-plugin-relay/macro'
 import ms from 'ms'
 import {useEffect} from 'react'
+import {useLazyLoadQuery} from 'react-relay'
 import {RouterProps, useHistory} from 'react-router'
 import Atmosphere from '../Atmosphere'
 import SendClientSegmentEventMutation from '../mutations/SendClientSegmentEventMutation'
+import {TierEnum} from './../__generated__/NewTeamOrgPicker_organizations.graphql'
+import {useUsageSnackNagQuery} from './../__generated__/useUsageSnackNagQuery.graphql'
 import useAtmosphere from './useAtmosphere'
 
 const NAG_EVERY = ms('3m')
@@ -11,6 +15,13 @@ const getIsNaggingPath = (history: RouterProps['history']) => {
   const {location} = history
   const {pathname} = location
   return !(pathname.includes('/usage') || pathname.includes('/meet/'))
+}
+
+const shouldNag = (tier: TierEnum, suggestedTier: TierEnum | null) => {
+  if (!suggestedTier) return false
+  const suggestPro = suggestedTier === 'pro' && tier === 'personal'
+  const suggestEnterprise = suggestedTier === 'enterprise' && tier !== 'enterprise'
+  return suggestPro || suggestEnterprise
 }
 
 const emitNag = (atmosphere: Atmosphere, history: RouterProps['history']) => {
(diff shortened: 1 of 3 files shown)


packages/client/hooks/useUsageSnackNag.ts
```diff
@@ -0,0 +1,77 @@
...
+    autoDismiss: 0,
+    onDismiss: () => {
+      setTimeout(() => {
+        const isNaggingPath = getIsNaggingPath(history)
+        if (!isNaggingPath) return
+        emitNag(atmosphere, history)
+      }, NAG_EVERY)
+    },
+    action: {
+      label: 'View Usage',
```
nickoferrall, line 40: I wanted the CTA to take the user to the `/usage` page rather than being a `Dismiss` button like the other snack nag because it saves the user a couple of clicks, and the `/usage` page is somewhat hidden.

I considered changing the copy to `Checkout Usage` as `View Usage` doesn't feel super friendly / a bit boring, but users might associate `Checkout` with a shopping cart, so I kept it as `View`.

packages/client/hooks/useUsageSnackNag.ts
```diff
@@ -0,0 +1,77 @@
...
+const useUsageSnackNag = (insights: boolean) => {
+  const atmosphere = useAtmosphere()
+  const history = useHistory()
+  const isNaggingPath = getIsNaggingPath(history)
+  const isUserNaggable = isNaggingPath && insights
+  const {viewer} = useLazyLoadQuery<useUsageSnackNagQuery>(
+    graphql`
+      query useUsageSnackNagQuery($isUserNaggable: Boolean!) {
+        viewer {
+          domains @include(if: $isUserNaggable) {
```
nickoferrall, line 60: I know that the `domains` query is expensive, so rather than run this query for all users in the dashboard, I thought it'd be a good opportunity to use the `include` directive

mattkrick: is it possible to avoid sending a request to the server if the user does not have the feature flag enabled?

mattkrick (changes requested):

packages/client/hooks/useUsageSnackNag.ts
nickoferrall: If the user does not have the `insights` feature flag enabled, `isUserNaggable` will be false, so the `domains` query will be ignored and won't be sent to the server. Are you thinking we should add the `@include` directive to `viewer` instead of `domains`?

As there are no other fields requested in `viewer`, if `domains` is ignored, there is no network request for `useUsageSnackNagQuery`:

contains-query

When `insights` is false:
ignores-query

mattkrick: WHOAAA. I had no idea that relay used the `@include` directive as a client-side filter! i thought it just naively sent it onto the server.
really cool find!

Issue #6793, opened 2022-06-21
Unable to use invitation link if invitee is already in team


nickoferrall:
@acressall shared that a [user](https://app.hubspot.com/contacts/3888472/contact/68344051) 🔒  was unable to use the invite links. 

It looks like they're sharing the invite to team invitation link when they want to invite team members to a meeting. However, it should still be possible to use this.

### To reproduce

- Copy the team invitation link
- Go to a new window and create a new account
- Paste the invitation link into the URL and see that you're successfully added to the team / join a meeting
- Now paste the URL for a second time and see the following screen:

![Screenshot 2022-06-21 at 08 57 13](https://user-images.githubusercontent.com/39854876/174749166-aff7aea9-3e60-49ab-be3f-1e749310d1b3.png)

PR #6795, opened 2022-06-21
fix: reuse invite link

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6793

### To test

- In prod/master, copy the team invitation link
- Go to a new window and create a new account
- Paste the invitation link into the URL and see that you're successfully added to the team 
- Now paste the URL for a second time and see the loading screen (see screenshot in issue)
- Checkout this branch and see that you can use the same invitation link as many times as you wish

### DIFF
--- a/packages/client/mutations/AcceptTeamInvitationMutation.ts
+++ b/packages/client/mutations/AcceptTeamInvitationMutation.ts
@@ -36,6 +36,7 @@ graphql`
     }
   }
 `
+
 graphql`
   fragment AcceptTeamInvitationMutation_notification on AcceptTeamInvitationPayload {
     # this is just for the user that accepted the invitation
--- a/packages/client/mutations/AcceptTeamInvitationMutation.ts
+++ b/packages/client/mutations/AcceptTeamInvitationMutation.ts
@@ -143,6 +143,7 @@ export const handleAcceptTeamInvitationErrors = (
 ) => {
   if (acceptTeamInvitation?.error) {
     const {message} = acceptTeamInvitation.error
+    if (message === InvitationTokenError.ALREADY_ACCEPTED) return true
     atmosphere.eventEmitter.emit('addSnackbar', {
       autoDismiss: 0,
       key: `acceptTeamInvitation:${message}`,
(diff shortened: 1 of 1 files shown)


packages/client/mutations/AcceptTeamInvitationMutation.ts
```diff
@@ -175,10 +177,6 @@ const AcceptTeamInvitationMutation: StandardMutation<
         onCompleted(data, errors)
       }
       const {acceptTeamInvitation} = data
-      const isOK = ignoreApproval
-        ? true
-        : handleAcceptTeamInvitationErrors(atmosphere, acceptTeamInvitation)
-      if (!isOK) return
```
nickoferrall, line 181: The purpose of `ignoreApproval` as explained [here](https://github.com/ParabolInc/parabol/pull/6487/files#r866228637) was:

> onError already propagates the error to the form. passing in ignoreApproval prevents a snackbar with the same message from popping up

However, `ignoreApproval` is only used in `EmailPasswordAuthForm` and if a user has already been accepted into a team, `acceptTeamInvitation` will contain an `ALREADY_ACCEPTED` error message. This means that `isOK` will be false and we'll return here. 

By moving this logic below the error message check on line 185, we should be able to prevent the same error message from popping up twice while fixing this bug.

packages/client/mutations/AcceptTeamInvitationMutation.ts
```diff
@@ -143,6 +144,7 @@ export const handleAcceptTeamInvitationErrors = (
 ) => {
   if (acceptTeamInvitation?.error) {
     const {message} = acceptTeamInvitation.error
+    if (message === InvitationTokenError.ALREADY_ACCEPTED) return true
```
nickoferrall, line 147: I think we shouldn't show the `accepted` snackbar to users after they login, or change the snack message to be more user friendly

mattkrick:
tested the following from #6487 (it used that ignoreApproval flag):

>  not logged in: visit a team-invitation link. log in via saml & see the error 

everything works!

Issue #6791, opened 2022-06-20
Retro template search filter


nickoferrall:
We would like to add a search filter to the retro template settings so that users can quickly find the template they're looking for (see issue https://github.com/ParabolInc/parabol/issues/5972 and discussion https://github.com/ParabolInc/parabol/discussions/6249). 

In this issue, let's implement a basic search filter that filters the retro templates if the template's name doesn't include the search query. I suggest that we implement tags as suggested in https://github.com/ParabolInc/parabol/issues/5972 in a separate issue once a design has been created.

### AC

- [ ] Implement a search bar in the retro template modal 
- [ ] When searching in the search bar, the templates are filtered

Issue #5978, opened 2022-01-28
Fix template modal UI on mobile 


nickoferrall:
On mobile, the meeting template modal UI needs some love & attention:

![template-modal](https://user-images.githubusercontent.com/39854876/151585178-712b800a-eed9-4b45-b809-21d39605d20f.png)

### AC

- [x] Modal contents doesn't flow over the modal
- [x] Template names are hidden behind the action button
- [ ] Tab button underlining is positioned under the button text

PR #6798, opened 2022-06-21
feat: retro template search filter

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6791

This PR implements a search filter so users can quickly find the template they're looking for. The UI does not sparkle on mobile as there are existing template UI bugs as outlined in: https://github.com/ParabolInc/parabol/issues/5978

<details>

template-filter

</details>

### To test

- [ ] Typing in the search bar filters templates in Team, Organization, and Public
- [ ] If you begin typing in one tab, e.g. Public, and switch to the Team tab, the search query is removed

### DIFF
--- a/packages/client/modules/meeting/components/ReflectTemplateListTeam.tsx
+++ b/packages/client/modules/meeting/components/ReflectTemplateListTeam.tsx
@@ -1,10 +1,10 @@
 import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
 import React from 'react'
-import {createFragmentContainer} from 'react-relay'
+import {useFragment} from 'react-relay'
 import useActiveTopTemplate from '../../../hooks/useActiveTopTemplate'
 import {PALETTE} from '../../../styles/paletteV3'
-import {ReflectTemplateListTeam_teamTemplates} from '../../../__generated__/ReflectTemplateListTeam_teamTemplates.graphql'
+import {ReflectTemplateListTeam_settings$key} from '../../../__generated__/ReflectTemplateListTeam_settings.graphql'
 import ReflectTemplateItem from './ReflectTemplateItem'
 
 const TemplateList = styled('ul')({
@@ -38,13 +38,30 @@ interface Props {
   activeTemplateId: string
   showPublicTemplates: () => void
   teamId: string
-  teamTemplates: ReflectTemplateListTeam_teamTemplates
+  settingsRef: ReflectTemplateListTeam_settings$key
 }
 
 const ReflectTemplateListTeam = (props: Props) => {
-  const {isActive, activeTemplateId, showPublicTemplates, teamId, teamTemplates} = props
+  const {isActive, activeTemplateId, showPublicTemplates, teamId, settingsRef} = props
+  const settings = useFragment(
+    graphql`
+      fragment ReflectTemplateListTeam_settings on RetrospectiveMeetingSettings {
+        templateSearchQuery
+        teamTemplates {
+          ...ReflectTemplateItem_template
+          id
+          name
+        }
+      }
+    `,
+    settingsRef
+  )
+  const {teamTemplates, templateSearchQuery} = settings
   const edges = teamTemplates.map((t) => ({node: {id: t.id}})) as readonly {node: {id: string}}[]
   useActiveTopTemplate(edges, activeTemplateId, teamId, isActive, 'retrospective')
+  const filteredTemplates = teamTemplates.filter(({name}) =>
+    name.toLowerCase().includes(templateSearchQuery ?? '')
+  )
   if (teamTemplates.length === 0) {
     return (
       <Message>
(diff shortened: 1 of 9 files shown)


nickoferrall:
Hey @ackernaut, would you mind reviewing this PR and seeing if the design I went with looks good to you?

ackernaut (changes requested):
This is cool @nickoferrall !

- I see a blue outline on the search box at all times. I would style it to change from transparent to blue on focus, as that’s a focus pattern we’re using in other places (tapping on a standup response, tapping on the agenda item input, etc.)
- I noticed that the search box is still available when the team or org list is empty. Should we disable it in that case (visually muted, no input cursor)? I noticed that when the list is not empty but there aren’t any matching results, you dynamically update the help text ‘_No public templates match your search query "ninja"_‘. Should we use that pattern even for an empty list? Or should we hide the search altogether if the list includes 5 templates or fewer?
- I like that the search clears when switching tabs. Should we add a little X icon on the right that can clear it for the current tab?
- I noticed this is not on the Sprint Poker template selector which is okay, there are only 2 public templates. Thinking about this, is this one reason to hide search if there are 5 templates or fewer?

ackernaut:
@nickoferrall 
- Can we inset the search control 8px, use the pill radius (like Poker Scope search) or 4px radius, and then go from Slate 400 to Sky 500?
Screen Shot 2022-06-21 at 10 26 14
Screen Shot 2022-06-21 at 10 27 47
- Hiding the search sounds complex. I think we can leave it as is for now. If the list is empty the results are going to be empty. However, I’m not sure if it’s apparent that the search is scoped to the tab. Maybe we say ‘Search [tab name] templates…’ in the placeholder text? Is that possible?
- Thanks for adding the little X if not too much creep in scope

nickoferrall:
@ackernaut ooh yeah, the poker scope search bar style looks much nicer here. New screenshot below:

<details>

Screenshot 2022-06-21 at 19 28 24

</details>

I implemented the colour change onFocus from Slate 400 to Sky 500 but I actually think it looks better when it remains Slate 400. That's how the GitHub search bar works, so I removed the onFocus logic. Curious what you think - gif below:

<details>

change-border-colour

</details>

> Maybe we say ‘Search [tab name] templates…’ in the placeholder text?

I love this idea! Update pushed which includes the template type in the placeholder. I've also added the close search icon and only display it once there's text to clear.

ackernaut:
@nickoferrall wow this is really coming along!
- I think the divider bar between the icon and the search input text is unnecessary in this case. In Sprint Poker the search icon is sometimes a dropdown button so the bar probably makes it feel like a split control
- I’m okay dropping the blue focus styles in this case, since the control is conventual (outlined text box with a search icon)

packages/client/modules/meeting/components/ReflectTemplateSearchBar.tsx
```diff
@@ -0,0 +1,139 @@
...
+            autoComplete='off'
+            name='search'
+            placeholder={`Search ${normalizedTempType} templates...`}
+            type='text'
+            onChange={onChange}
+            ref={inputRef}
+            value={templateSearchQuery ?? ''}
+          />
+        </InputWrapper>
+        <ClearSearchIcon isEmpty={!templateSearchQuery} onClick={clearSearch}>
```
mattkrick, line 131: -1 the input field should still have focus after clearing

packages/client/modules/meeting/components/ReflectTemplateListTeam.tsx
```diff
@@ -38,13 +38,30 @@ interface Props {
...
+          name
+        }
+      }
+    `,
+    settingsRef
+  )
+  const {teamTemplates, templateSearchQuery} = settings
   const edges = teamTemplates.map((t) => ({node: {id: t.id}})) as readonly {node: {id: string}}[]
   useActiveTopTemplate(edges, activeTemplateId, teamId, isActive, 'retrospective')
+  const filteredTemplates = teamTemplates.filter(({name}) =>
```
mattkrick, line 62: +1 curious, could you reuse the following hook? https://github.com/ParabolInc/parabol/blob/c3a29b0323049581c88e4a3dc5313b94ca8876da/packages/client/hooks/useFilteredItems.ts

packages/client/modules/meeting/components/ReflectTemplateListTeam.tsx
```diff
@@ -53,9 +70,12 @@ const ReflectTemplateListTeam = (props: Props) => {
       </Message>
     )
   }
+  if (filteredTemplates.length === 0) {
+    return <Message>{`No team templates match your search query "${templateSearchQuery}"`}</Message>
+  }
   return (
     <TemplateList>
-      {teamTemplates.map((template) => {
+      {filteredTemplates.map((template) => {
```
mattkrick, line 78: +1 it's nice to see the search query highlighted or made bold in the result. something like this: https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/client/components/TypeAheadLabel.tsx

mattkrick (changes requested):

packages/client/modules/meeting/components/ReflectTemplateListTeam.tsx
nickoferrall: Yeah, I do like seeing the text become bold when it matches the query. As the title is already bold here, I've tried highlighting it instead. I kinda like it, not sure if it feels too busy. What do you think?

template-search

mattkrick: i like it! I was gonna suggest going a shade lighter, but it passes the contrast ratio test, so I'm happy!
Screen Shot 2022-06-24 at 11 50 51 AM

PR #6803, opened 2022-06-22
fix: Added dismiss button to snackbar messages

### LINKED ISSUE #6218: Add dismiss button to snackbar messages
Currently, our snackbar messages look like this:

Screenshot 2022-03-15 at 08 24 49

While the message and CTA are pretty clear, I find that there is no obvious way to dismiss them. We currently support clicking anywhere on the snackbar to remove them, but there are two issues with this approach:

1. Not everyone knows this (I didn't!), since there is no visible affordance that indicates it. 
2. Because focus cannot be placed on the snackbar itself, they can only be dismissed if you are using a mouse or touch input device. Keyboard-only users have not much of a choice.

Snackbars are also intended to [disappear automatically](https://material.io/components/snackbars#behavior):

> Snackbars appear without warning, and don't require user interaction. They automatically disappear from the screen after a minimum of four seconds, and a maximum of ten seconds.

But in our case, it stays in view until clicked on.

I don't mind having them stay in view, it makes sense given the type of messages we are using them for. But I think **we can improve the interaction if we add a close/dismiss button to them**.

This is similar to what GMail is doing:

Captura de Pantalla 2022-03-16 a la(s) 10 45 26 a m

For us, it could look like this:

Captura de Pantalla 2022-03-16 a la(s) 10 45 26 a m

- The puzzle piece emoji (or any emoji we may use in the future) is relocated to the start of the text. This anchors the eyes, but also brings visual balance to the whole component.
- The `Learn more` button maintains its primary CTA status by visually standing out more than anything else.
- The close button is visible, but does not steal any attention from the message nor the primary CTA.

### Acceptance criteria
- Users can dismiss a snackbar via a visible affordance (dimiss button)

### Estimated effort
(Needs input from a developer)

87prashant:
# Description

fixes #6218

## Demo

Screenshot 2022-06-22 140132

Screenshot 2022-06-22 140046

## Testing scenarios

Try to generate different snackbar messages and verify the dismiss button and its functionality.

## Final checklist


87prashant:
https://github.com/ParabolInc/parabol/issues/6218#issuecomment-1163543678
Should the dismiss button be conditional?

Issue #6809, opened 2022-06-22
Replace useAllIntegrationQuery in GitHub & Jira


nickoferrall:
In the GitHub & Jira new issue menus, we use the `useAllIntegration` hook to get the repos and projects we need. However, `useAllIntegration` queries all of our integrations rather than just the one that we need. The more integrations we add, the more expensive this gets.

![useAllIntegrations](https://user-images.githubusercontent.com/39854876/175080013-08696d72-b5c0-4a0c-81c2-801268a2c020.png)

Instead, we can query the service directly and avoid using `useAllIntegrations` as we're doing in the GitLab integration. 

### AC

- [ ] Replace `useAllIntegrations` in GitHub & Jira new issue menus
- [ ] Pushing a task to GitHub or Jira still works in the same way it currently does as we need all integrations

PR #6819, opened 2022-06-23
refactor: new GitHub & Jira issue queries

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6809

Base: #6791 - looks like I didn't switch back to master...

This PR refactors the `NewJiraIssueInput` and `NewGitHubIssueInput` queries away from the `useAllIntegrations` hook. That query would fetch all projects/repos from all integrations, which resulted in it being our 9th slowest query by p99 latency (see issue). This would only become slower as we add more integrations.

Now, GitHub and Jira are only querying GitHub and Jira.

### To test

- [ ] Create a new issue in Jira, open the new issue dropdown menu and see all expected projects
- [ ] Do the same in GitHub and see all expected repos
- [ ] Can successfully add a new issue to GitHub and Jira
- [ ] Can successfully push a task to GitHub and Jira


packages/client/hooks/useGetRepoContributions.ts
```diff
@@ -0,0 +1,90 @@
...
+import {useMemo} from 'react'
+import {useFragment} from 'react-relay'
+import {useGetRepoContributions_teamMember$key} from './../__generated__/useGetRepoContributions_teamMember.graphql'
+
+export type Repo = {
+  readonly id: string
+  readonly nameWithOwner: string
+}
+
+const useGetRepoContributions = (teamMemberRef: useGetRepoContributions_teamMember$key): Repo[] => {
```
nickoferrall, line 12: Previously, in the filter menu, we were getting repos by contributions, but in the new issue menu, we were querying them by organization. By moving this to a hook and querying the same fields, we save an additional query if the user opens the filter menu before the new issue menu, for example.

mattkrick:
@nickoferrall after merging the base branch it found some conflicts. could you resolve?

PR #6821, opened 2022-06-23
feat: allow facilitator to update the standup meeting title

### LINKED ISSUE #6820: Standups: Allow users to change the meeting title
The facilitator should be able to rename the standup meeting title in case it's used as a different activity

AC
- [ ] Facilitator can update the meeting title
- [ ] Normal user can't edit the meeting title

BartoszJarocki:
# Description

Fixes #6820 

## Demo

## Testing scenarios

- [ ] Scenario A
  - Step 1
  - Step 2...

- [ ] Scenario B
  - Step 1
  - Step 2....

## Final checklist

### DIFF
--- a/packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
+++ b/packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
@@ -2,10 +2,15 @@ import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
 import React from 'react'
 import {useFragment} from 'react-relay'
+import useAtmosphere from '~/hooks/useAtmosphere'
+import useMutationProps from '~/hooks/useMutationProps'
 import NewMeetingAvatarGroup from '~/modules/meeting/components/MeetingAvatarGroup/NewMeetingAvatarGroup'
+import RenameMeetingMutation from '~/mutations/RenameMeetingMutation'
+import Legitity from '~/validation/Legitity'
 import {TeamPromptTopBar_meeting$key} from '~/__generated__/TeamPromptTopBar_meeting.graphql'
 import {meetingAvatarMediaQueries} from '../../styles/meeting'
 import BackButton from '../BackButton'
+import EditableText from '../EditableText'
 import {IconGroupBlock, MeetingTopBarStyles} from '../MeetingTopBar'
 import TeamPromptOptions from './TeamPromptOptions'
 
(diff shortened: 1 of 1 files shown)


jmtaber129 (changes requested):
-1 A lot of this seems duplicated from the editable meeting name in `NewMeetingSidebar.tsx`.  We should probably just refactor the editable meeting name out of the sidebar and into its own component, and then reuse that here to ensure consistency in editable meeting names

jmtaber129:
+1 Eh, there's still some code duplication - the code [here](https://github.com/ParabolInc/parabol/pull/6821/commits/b418957d0cf17f2d9c789daacaf5cb36ca9e815f#diff-f3d13923c4f5b000506a131b00622ab307d68fa3983bd9cf31234027338ebf2fR98-R110) is still a copy of [this](https://github.com/ParabolInc/parabol/blob/master/packages/client/components/NewMeetingSidebar.tsx#L114-L126), vs a single `EditableMeetingTitle` that just takes a `meeting` fragment.

The only difference I can see is some minor style differences that I think could be resolved with `styled`.

Not blocking though

jmtaber129 (approved):
I do still think this should be extracted into its own component instead of a hook, but non-blocking, so LGTM

BartoszJarocki:
@jmtaber129 the rule I follow is [here](https://www.notion.so/parabol/How-to-Do-a-Code-Review-daaf03fae3484cf5a1e18174170dadff#598e7a6900f5430ab69f85d88c2f055f). I do think in general copy/paste is cheaper than the wrong abstraction, so unless the pattern is clear, and there's a need for more usage I avoid creating shared interfaces or hiding the implementation inside of the specific component. 

Currently, we have a shared logic in a hook, and we can use it with whatever components we want. If you extract a component, just to avoid one if statement, you either need to provide a way to customize whatever's hidden inside or assume it will always be the same. right now, the pattern isn't clear to me, and having the same if statement in two places in exchange for being able to very easily customize that in any place is ok. 

Anyway, this is always more art than science so it's normal to have different views on that kind of stuff. If that's causing tension on your end, I'd encourage you to start a discussion on that topic. I'll merge that, as it's important to have in this week's release. 

Here are a few articles worth reading about that kind of dilemma:
https://sandimetz.com/blog/2016/1/20/the-wrong-abstraction 
https://overreacted.io/goodbye-clean-code/

PR #6866, opened 2022-07-08
feat: New meeting dialog is now a modal

### LINKED ISSUE #5602: Updated new meeting dialog and route
See #5589 for context
new dialog

## Acceptance criteria
* [ ] existing new meeting page stays functional, new dialog is only accessible when entering route directly
* [ ] existing new meeting page copied and convert it into a dialog
* [ ] dialog has it's own "modal route" which can be accessed by `meetings/new-meeting` or `team/${teamId}/new-meeting` depending on where it is opened
  Inspiration can be gotten from https://github.com/davidmfoley/react-router-modal
* [ ] Bottom row with 'Create', 'Cancel' and existing meeting status is added
* [ ] Title and X buttons are added
* [ ] Existing template select is copied and adapted with new behaviour, **no** design or functional changes
  * [ ] When selecting the template field, the template selection UI should slide in from right to left and take up all the space on the dialog.
  * [ ] Everything should remain the same, with the exception of the addition of a back arrow preceding the title.
  * [ ] Clicking on the back arrow should slide the template selection UI out of view within the modal, and return to the main new meeting view.

## **Not** part of AC
* Carousel and meeting selector
* Pointing 'Add meeting' buttons to new route

Dschoordsch:
# Description

Fixes #5602 

This is the first step in redesigning the dialog. The dialog still uses
the same route as before. If the route is hit directly, the meetings
dashboard is rendered in the background. If it's reached by pressing the
"Add meeting" button, then the current page is put into the location
state to be rendered as background.

## Demo

https://www.loom.com/share/eba28320ed6f4c67806042a440ac1b9f

image

## Testing scenarios

* Create at least 2 teams
  * team 1 with some active meetings
  * team 2 with no active meetings
* [ ] Open the new meeting dialog by the "Add meeting" button from different dashboards
  * [ ] see that the dashboard is still in the background
* [ ] Open the new meeting dialog by hitting the URL directly
  * [ ] meetings dash is in the background
* [ ] test all meeting dialog buttons, including all drop downs
* [ ] check layout on mobile
  * [ ] with active meetings
  * [ ] with no active meetings
  * [ ] with opening the description
* [ ] smoke test some other dialogs
  * [ ] create teams
  * [ ] credit card information
  * [ ] invite members

### DIFF
--- a/packages/client/components/Dashboard.tsx
+++ b/packages/client/components/Dashboard.tsx
@@ -1,8 +1,9 @@
 import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
+import {Location} from 'history'
 import React, {lazy, useRef} from 'react'
 import {PreloadedQuery, usePreloadedQuery} from 'react-relay'
-import {Route, Switch} from 'react-router'
+import {Route, Switch, useLocation} from 'react-router'
 import useBreakpoint from '~/hooks/useBreakpoint'
 import useSnackNag from '~/hooks/useSnackNag'
 import useSnacksForNewMeetings from '~/hooks/useSnacksForNewMeetings'
@@ -37,6 +38,9 @@ const NewTeam = lazy(
     import(
       /* webpackChunkName: 'NewTeamRoot' */ '../modules/newTeam/containers/NewTeamForm/NewTeamRoot'
     )
+)
+const NewMeetingRoot = lazy(
+  () => import(/* webpackChunkName: 'NewMeetingRoot' */ './NewMeetingRoot')
 )
 
 interface Props {
@@ -128,6 +132,10 @@ const Dashboard = (props: Props) => {
   useSnackNag(overLimitCopy)
   useUsageSnackNag(insights)
   useSnacksForNewMeetings(activeMeetings)
+
+  const location = useLocation<{backgroundLocation?: Location}>()
+  const state = location.state
+
   return (
     <DashLayout>
       <SkipLink href='#main'>Skip to content</SkipLink>
@@ -145,9 +153,9 @@ const Dashboard = (props: Props) => {
           </SwipeableDashSidebar>
         )}
         <DashMain id='main' ref={meetingsDashRef}>
-          <Switch>
+          <Switch location={state?.backgroundLocation || location}>
             <Route
-              path='/meetings'
+              path='(/meetings|/new-meeting)'
               render={(routeProps) => (
                 <MeetingsDash {...routeProps} meetingsDashRef={meetingsDashRef} viewer={viewer} />
               )}
@@ -156,6 +164,9 @@ const Dashboard = (props: Props) => {
             <Route path='/team/:teamId' component={TeamRoot} />
             <Route path='/newteam/:defaultOrgId?' component={NewTeam} />
             <Route path='/usage' component={InsightsRoot} />
+          </Switch>
+          <Switch>
+            <Route path='/new-meeting/:teamId?' component={NewMeetingRoot} />
           </Switch>
         </DashMain>
       </DashPanel>
--- a/packages/client/components/NewMeeting.tsx
+++ b/packages/client/components/NewMeeting.tsx
@@ -31,10 +31,11 @@ interface Props {
 
 const MEDIA_QUERY_VERTICAL_CENTERING = '@media screen and (min-height: 840px)'
 
-const IllustrationAndSelector = styled('div')({
+const IllustrationAndSelector = styled('div')<{isDesktop}>(({isDesktop}) => ({
   gridArea: 'picker',
-  width: '100%'
-})
+  width: '100%',
+  paddingBottom: isDesktop ? 0 : 32
+}))
 
 const TeamAndSettings = styled('div')<{isDesktop}>(({isDesktop}) => ({
   alignItems: 'center',
@@ -167,7 +168,7 @@ const NewMeeting = (props: Props) => {
         </CloseButton>
       </Title>
       <NewMeetingInner isDesktop={isDesktop}>
-        <IllustrationAndSelector>
+        <IllustrationAndSelector isDesktop={isDesktop}>
           <NewMeetingIllustration idx={idx} setIdx={setIdx} newMeetingOrder={newMeetingOrder} />
           <NewMeetingMeetingSelector meetingType={meetingType} idx={idx} setIdx={setIdx} />
         </IllustrationAndSelector>
--- a/packages/client/components/NewMeeting.tsx
+++ b/packages/client/components/NewMeeting.tsx
@@ -65,8 +65,11 @@ const Title = styled(DialogTitle)({
 })
 
 const CloseButton = styled(FlatButton)({
-  padding: 8,
-  color: PALETTE.SLATE_600
+  position: 'absolute',
+  top: 8,
+  right: 8,
+  color: PALETTE.SLATE_600,
+  padding: 0
 })
 
 const NewMeetingInner = styled('div')<{isDesktop: boolean}>(
--- a/packages/client/components/NewMeeting.tsx
+++ b/packages/client/components/NewMeeting.tsx
@@ -3,8 +3,8 @@ import graphql from 'babel-plugin-relay/macro'
 import React, {useEffect, useMemo, useRef, useState} from 'react'
 import {PreloadedQuery, usePreloadedQuery} from 'react-relay'
 import {mod} from 'react-swipeable-views-core'
-import WaveSVG from 'static/images/wave.svg'
 import useUsageSnackNag from '~/hooks/useUsageSnackNag'
+import {PALETTE} from '~/styles/paletteV3'
 import {NonEmptyArray} from '~/types/generics'
 import {MeetingTypeEnum, NewMeetingQuery} from '~/__generated__/NewMeetingQuery.graphql'
 import useBreakpoint from '../hooks/useBreakpoint'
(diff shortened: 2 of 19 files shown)


igorlesnenko (approved):
Tweaked markup to keep experience smooth until we implement a new design

packages/client/components/NewMeetingActions.tsx
```diff
@@ -8,61 +8,76 @@ import StartTeamPromptMutation from '~/mutations/StartTeamPromptMutation'
...
-  width: 320,
-  maxWidth: '100%',
-  [newMeetingGridMediaQuery]: {
-    fontSize: 24,
-    height: 64
+const StartButton = styled(FlatPrimaryButton)({
+  fontSize: 20,
+  height: 50,
+  [narrowScreenMediaQuery]: {
+    paddingLeft: '1em',
```
mattkrick, line 55: -1 the em/rem/px debate goes deep, but since we don't really support browser zoom & our base size is 16px, 1em will always equal 16px, so it's nice to just use pixels so you can deal with whole numbers. otherwise if we change this we'll have to write something like `.6875em`, which gets ugly. i know there are some very old legacy components that still use em, we're trying to get rid of those!

packages/client/components/NewMeetingRoot.tsx
```diff
@@ -19,9 +21,30 @@ const NewMeetingRoot = () => {
...
   useSubscription('NewMeetingRoot', TeamSubscription)
   const queryRef = useQueryLoaderNow<NewMeetingQuery>(newMeetingQuery, {teamId})
-  return (
+
+  const location = useLocation<{backgroundLocation?: Location}>()
+  const history = useHistory()
+
+  const onClose = useCallback(() => {
+    const state = location.state
+    history.push(state?.backgroundLocation ?? '/meetings')
```
mattkrick, line 30: -1 as convention, it's nice to use `history.replace` for modals. typically when you hit the browser back button you expect more than to close the modal

packages/client/components/NewMeeting.tsx
```diff
@@ -38,37 +42,32 @@ const TeamAndSettings = styled('div')<{isDesktop}>(({isDesktop}) => ({
...
-      // the wave is 2560x231, so to figure out the offset from the center, we need to find how much scaling there was
-      backgroundPositionY: `calc(50% - ${Math.floor(((innerWidth / 2560) * 231) / 2 - 1)}px), 0`,
-      height: '100%',
-      minHeight: 0,
-      overflow: 'auto'
-    }
-)
+const NewMeetingDialog = styled(DialogContainer)({
+  width: '800px',
+  maxHeight: 'unset'
```
mattkrick, line 55: -1 we never want a modal to be taller than the viewport. in the pic below, the CTA is below the fold, which is no good. instead.

options:
- shrink/hide the image when the instructions are expanded
- make the instructions overflow: auto
- cut some margin between the options & "1 Active Meeting"
- ???

Screenshot from 2022-07-15 16-20-13

mattkrick:
-1 Typically, if a child div is much smaller than its parent, they shouldn't have the same border radius. we'll want a little more visual contrast (see below). In this case, since the illustration is about the same size as a meeting card & the meeting card has a 4px border radius, I'd suggest the illustrations have a 4px border radius (or maybe 8 if you're feeling spicy) & keep the modal with the 12px. 

Screenshot from 2022-07-15 16-26-06

packages/client/components/Dashboard.tsx
```diff
@@ -128,6 +132,10 @@ const Dashboard = (props: Props) => {
   useSnackNag(overLimitCopy)
   useUsageSnackNag(insights)
   useSnacksForNewMeetings(activeMeetings)
+
+  const location = useLocation<{backgroundLocation?: Location}>()
```
mattkrick, line 136: +2 this is some straight up magic here

packages/client/components/NewMeetingActions.tsx
```diff
@@ -8,61 +8,76 @@ import StartTeamPromptMutation from '~/mutations/StartTeamPromptMutation'
...
+    paddingRight: '1em'
   }
 })
 
-const ForwardIcon = styled(Icon)({
-  paddingLeft: 16,
-  [newMeetingGridMediaQuery]: {
-    fontSize: 36 // MD 1.5x
+const CancelButton = styled(FlatButton)({
```
mattkrick, line 60: -1 users can close by clicking the X or the scrim, let's get rid of the cancel to simplify this view a little more

packages/client/components/NewMeeting.tsx
```diff
@@ -38,37 +42,32 @@ const TeamAndSettings = styled('div')<{isDesktop}>(({isDesktop}) => ({
...
+  fontSize: 24,
+  display: 'flex',
+  flexDirection: 'row',
+  alignItems: 'center',
+  justifyContent: 'space-between',
+  paddingBottom: 24
+})
+
+const CloseButton = styled(FlatButton)({
+  padding: 8,
```
mattkrick, line 68: -1 let's move this just a touch more out of the way, get rid of the padding & 

```css
position: absolute;
top: 8px;
right: 8px;
```

packages/client/components/TopBarStartMeetingButton.tsx
```diff
@@ -21,11 +19,12 @@ const MeetingLabel = styled('div')({
 })
 
 const TopBarStartMeetingButton = () => {
+  const location = useLocation()
   const teamId = getTeamIdFromPathname()
   const {history} = useRouter()
 
   const onClick = () => {
-    history.push(`/new-meeting/${teamId}?source=TopBar`)
+    history.push(`/new-meeting/${teamId}?source=TopBar`, {backgroundLocation: location})
```
mattkrick, line 27: -1 history.replace

packages/client/components/StartMeetingFAB.tsx
```diff
@@ -82,7 +84,7 @@ const StartMeetingFAB = (props: Props) => {
     }
   }
   const onClick = () => {
-    history.push(`/new-meeting/${teamId}?source=BottomFAB`)
+    history.push(`/new-meeting/${teamId}?source=BottomFAB`, {backgroundLocation: location})
```
mattkrick, line 87: -1 history.replace

packages/client/components/TopBarStartMeetingButton.tsx
mattkrick: +1 when teamId is unknown, it feels weird that this url flickers `/source=TopBar` & then goes to `/:teamId`. Nothing breaks, it just feels unpolished to have quick redirects.
 
ideally, the flow would be:
- if teamId is unknown, persist source=TopBar somewhere else (context, localSession, global var, react router location, etc.
- when teamId is known & we're going to redirect to /:teamId, attach the source=TopBar to the end

mattkrick (changes requested):
couple minor front-end design things, aside from that looks & works very well!

packages/client/components/NewMeeting.tsx
Dschoordsch: I agree the CTA shouldn't be below the fold, but I think in this case it would be acceptable since:
1) it only moves out of screen once the user interacts with the "How to Run" section
2) it simplifies scrolling for the user since the whole screen is the scrolling area
3) the whole content will be changed soon anyways

packages/client/components/NewMeetingActions.tsx
Dschoordsch: I agree, I wish though this would be caught already when reviewing the design.

packages/client/components/TopBarStartMeetingButton.tsx
igorlesnenko: @Dschoordsch can we use `SendClientSegmentEventMutation` instead of `source=TopBar` as a simple fix?

Dschoordsch: Let's just get rid of `source=TopBar` and `source=BottomFab`. I don't think we need this anymore, as it's not a user choice now, but just depends on screen resolution which button is visible.

igorlesnenko: Removed

packages/client/components/NewMeeting.tsx
Dschoordsch: @mattkrick what do you think about my comment above?

mattkrick: From a UX perspective, we just can't hide a primary CTA. The primary purpose of this modal is to get the user to click that button, so if we hide it, we're defeating that purpose. In this screen, you can see the CTA peek out a little bit, but on a slightly smaller screen, they wouldn't even see the button & there's no hint that they're supposed to scroll.

We could remove the illustration when they click "how to run", remove "how to run" once it's been clicked, condense the margin of the 4 steps, "learn more" and "1 active meeting". all of those would be preferred over hiding the CTA

mattkrick: or we could just remove "how to run" on mobile altogether!

mattkrick:
Only thing remaining is keeping the primary CTA above the fold on mobile. Apologies for being stubborn on this, there's just too much evidence showing what a difference it makes. After that good to merge!

packages/client/components/NewMeeting.tsx
Dschoordsch: Let's remove the "How to Run" if there is not enough space. If we run into this situation with the next iteration of new meeting dialog, we should place some arrow on the bottom or similar to indicate there is scrolling needed (actually I see a scroll bar on your screenshot, but that might be missed).

I think it might be better to make the dialog fullscreen on mobile
* we have less space wasted as the background is not of any value anyways
* the scrollbar would be visible on top of the dialog

igorlesnenko:
Removed how to run on mobile
image

PR #6867, opened 2022-07-08
feat: display Integrations link in the sidebar

### LINKED ISSUE #6556: Move Integrations to the sidebar
As discussed in https://github.com/ParabolInc/parabol/issues/5939, we want to increase the awareness of our integrations. The current flow to find integrations is:

- Click on Team
- Click on `Settings & Integrations` font size 12 hyperlink
- Click on Integrations

I'd love to move Integrations below `Tasks` on our sidebar so that it's more visible to our users and requires one click.

sidebar

In Settings & Integrations, we could just show settings and remove the Team & Integrations tabs. 
settings

### AC

- [ ] Integrations is added to the sidebar
- [ ] Integrations pathname changes from `/team/${teamId}/settings/integrations` to `/team/${teamId}/integrations`
- [ ] `Settings & Integrations` hyperlink changes to `Settings`
- [ ] Remove the Team & Integrations tab

igorlesnenko:
Fixes #6556

https://user-images.githubusercontent.com/466991/178038130-f1021295-61fe-4d5e-9f9c-81eac9badc8a.mov

### DIFF
--- a/packages/client/components/DashNavList/DashNavList.tsx
+++ b/packages/client/components/DashNavList/DashNavList.tsx
@@ -1,12 +1,12 @@
 import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
-import React, {useMemo, Fragment} from 'react'
+import React, {Fragment, useMemo} from 'react'
 import {createFragmentContainer} from 'react-relay'
-import {DashNavList_viewer} from '../../__generated__/DashNavList_viewer.graphql'
-import LeftDashNavItem from '../Dashboard/LeftDashNavItem'
 import {PALETTE} from '~/styles/paletteV3'
 import {Breakpoint} from '~/types/constEnums'
 import makeMinWidthMediaQuery from '~/utils/makeMinWidthMediaQuery'
+import {DashNavList_viewer} from '../../__generated__/DashNavList_viewer.graphql'
+import LeftDashNavItem from '../Dashboard/LeftDashNavItem'
 
 const DashNavListStyles = styled('div')({
   paddingRight: 8,
@@ -44,7 +44,7 @@ interface Props {
 }
 
 const DashNavList = (props: Props) => {
-  const {className, onClick, viewer} = props
+  const {onClick, viewer} = props
   const teams = viewer?.teams
 
   const teamsByOrgKey = useMemo(() => {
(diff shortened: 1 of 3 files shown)


enriquesanchez (changes requested):
Can we indent `Integrations` a bit more to the right? Currently there's not enough indentation so it's hard to tell it's a level "under" the team.

@nickoferrall I had brought this up in the past, but I don't love how it's just "Integrations" under the team tab. Has the squad given any feedback on this? Have y'all considered adding more options here?

Also, is this intended to be an experiment or test?

nickoferrall:
> @nickoferrall I had brought this up in the past, but I don't love how it's just "Integrations" under the team tab. Has the squad given any feedback on this? Have y'all considered adding more options here?

I'm in Squad 3, which is focused on revenue, so I've just been responding to questions on this issue. I think Squad 2 and Enrique, as a designer, should get the final say in how it works.

Still, I'm happy to continue sharing my thoughts. I think more menu item options could be added, like Settings. I also think it'd be great if the menu with Integrations wasn't visible by default. Only after clicking it a dropdown icon would it open up. Similar to how this Stripe example works:

stripe-sidebar

igorlesnenko:
> . I also think it'd be great if the menu with Integrations wasn't visible by default. 

@nickoferrall I think it is currently works like this in my implementation, the sub-menu is not visible until you click on a team link.

@enriquesanchez I think we can consider this as an experiment as well. We assume that the link is not visible enough, so we want to implement a new solution, and test the result.

However, I agree that we should not break the consistency of the current design.

Possible solutions that I currently see:

1. Keep link in sidebar as in this PR, but add more items. For example we can add settings link too. Anything else that could be added here?

2. Maybe  ust keep things as is but show some tooltip hint near the settings and integrations link when the board is just created? 

3. Keep the link as is but change link design to make it more visible. Add icons, split in 2 links: `Settings • Integrations`. 

What do you think? 

Just asked for a squad feedback.

enriquesanchez:
Thanks @igorlesnenko 

>Keep link in sidebar as in this PR, but add more items. For example we can add settings link too. Anything else that could be added here?

I like this option, on the top of my head I can think of adding:

- Members
- Settings
- Integrations

The problem is that none of these options have their own view, they are either a tab (Settings, Integrations) or a drawer (Members).

So it might be a good idea in the future when/if we redesign the admin pages.

For now, I'm on the fence about it. Do let me know what the squad thinks.

igorlesnenko:
@enriquesanchez closing this PR without merging, a new `design` issue created https://github.com/ParabolInc/parabol/issues/6894

PR #6874, opened 2022-07-12
fix(security): package bumps for dependabot fixes

mattkrick:

# Description

This should fix 3 of the CWEs identified by dependabot that it is not able to automatically fix.
Being vague here because security.
Of note-- now that lerna is deprecated, vulns are getting found & not fixed, so we should increase priority to move to non-lerna.

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,15 +66,16 @@
     "@xmldom/xmldom": "^0.7.0",
     "node-gyp": "^8.4.1",
     "simple-get": "^4.0.1",
-    "caniuse-lite": "~1.0.0"
+    "caniuse-lite": "~1.0.0",
+    "git-url-parse": "12.0.0"
   },
   "devDependencies": {
     "@babel/preset-env": "7.16.0",
-    "@graphql-codegen/add": "^3.1.0",
-    "@graphql-codegen/cli": "^2.3.1",
-    "@graphql-codegen/typescript": "^2.4.0",
-    "@graphql-codegen/typescript-operations": "^2.2.0",
-    "@graphql-codegen/typescript-resolvers": "^2.5.2",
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -66,15 +66,16 @@
     "@xmldom/xmldom": "^0.7.0",
     "node-gyp": "^8.4.1",
     "simple-get": "^4.0.1",
-    "caniuse-lite": "~1.0.0"
+    "caniuse-lite": "~1.0.0",
+    "git-url-parse": "12.0.0"
```
mattkrick, line 70: lerna depends on v11, forcing v12 to fix vuln

package.json
```diff
@@ -66,15 +66,16 @@
...
   },
   "devDependencies": {
     "@babel/preset-env": "7.16.0",
-    "@graphql-codegen/add": "^3.1.0",
-    "@graphql-codegen/cli": "^2.3.1",
-    "@graphql-codegen/typescript": "^2.4.0",
-    "@graphql-codegen/typescript-operations": "^2.2.0",
-    "@graphql-codegen/typescript-resolvers": "^2.5.2",
+    "@graphql-codegen/add": "^3.2.0",
+    "@graphql-codegen/cli": "^2.8.0",
```
mattkrick, line 75: 2 vulns from subdeps here

packages/integration-tests/package.json
```diff
@@ -10,7 +10,7 @@
     "test": "playwright test"
   },
   "devDependencies": {
-    "@playwright/test": "^1.19.2",
+    "@playwright/test": "^1.23.2",
```
mattkrick, line 13: 1 subdep vuln here

PR #6879, opened 2022-07-13
feat(standups): Customized start standup notification

### LINKED ISSUE #6849: Standups: Improve the start standup notification
Right now we use a generic notification type that we use for any other meeting. Standup notifications should encourage to add the response. Here's the design: https://github.com/ParabolInc/parabol/issues/6075#issuecomment-1125468744 
Currently, the implementation is the same for every meeting we have so this task might take a little longer as it'll require changes in the notification building implementation.

AC
- [ ] Start standup notification matches the design

BartoszJarocki:
# Description

Fixes #6849 

## Demo
Screenshot 2022-07-14 at 15 44 37
Screenshot 2022-07-14 at 15 43 48

Parabol icon doesn't work as it's a test app configuration

## Testing scenarios

- [ ] Integrate slack and start standup meeting, see if notification matches discussed style
- [ ] Integrate mattermost and start standup meeting, see if notification matches discussed style
- [ ] Integrate ms teams and start standup meeting, see if notification matches discussed style
- [ ] start other meeting types and see if everything works as previously 

I haven't tested MS Teams as I don't have any accounts there

## Final checklist


packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
```diff
@@ -43,6 +44,45 @@ const notifyMSTeams = async (
...
+}
+
+const createTeamPromptMeetingAction = (meetingUrl: string) => {
+  const joinMeetingAction = new AdaptiveCards.OpenUrlAction()
+  joinMeetingAction.title = 'Submit Response'
+  joinMeetingAction.url = meetingUrl
+  joinMeetingAction.id = 'submitResponse'
+
+  return joinMeetingAction
+}
```
jmtaber129, line 74: +1 It seems like a lot of this config is duplicated across Teams, Slack, and Mattermost.  Would it make sense to refactor this into a separate util?  I.e. create an abstraction of "meeting notification" that returns title, URL, button text, etc., and then the individual notifiers could just use that abstracted config instead of doing meeting-specific things for each notification integration.

This would likely reduce shotgun-surgery when creating new meeting types (only change the config, vs changing each notifier) and simplify building new notifier integrations.

jmtaber129 (approved):
One refactor suggestion and a couple minor nits, otherwise LGTM!

packages/server/database/types/MeetingTeamPrompt.ts
```diff
@@ -18,7 +18,7 @@ export function isMeetingTeamPrompt(meeting: Meeting): meeting is MeetingTeamPro
   return meeting.meetingType === 'teamPrompt'
 }
 
-function createTeamPromptDefaultTitle() {
+export function createTeamPromptDefaultTitle() {
```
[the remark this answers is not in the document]
BartoszJarocki: good catch, it's not used anywhere

packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
```diff
@@ -43,6 +44,45 @@ const notifyMSTeams = async (
 }
 export type MSTeamsNotificationAuth = IntegrationProviderMSTeams & {userId: string}
 
+const createTeamPromptMeetingTitle = (meeting: Meeting) => `*${meeting.name}* is open 💬`
+
+const genericMeetingTitle = `Meeting Started 👋`
```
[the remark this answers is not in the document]
BartoszJarocki: right, fixed

packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
BartoszJarocki: agreed! will create a refactor task for that.

BartoszJarocki:
@mattkrick would you mind taking a look? thanks!

packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
```diff
@@ -43,6 +44,45 @@ const notifyMSTeams = async (
 }
 export type MSTeamsNotificationAuth = IntegrationProviderMSTeams & {userId: string}
 
+const createTeamPromptMeetingTitle = (meeting: Meeting) => `*${meeting.name}* is open 💬`
+
+const GENERIC_MEETING_TITLE = `Meeting Started 👋`
+
+const MeetingTypeTitleLookup: Record<MeetingTypeEnum, ((meeting: Meeting) => string) | string> = {
```
mattkrick, line 51: -1 polymorphic values feel bad & clutter up the code down on L104, but I see the problem you're up against-- a lookup table is generally cleaner than a switch/case in javascript.

I'd prefer changing the generic case from a string constant into a thunk, just like you did for the actions! That way you don't have to duck-type the result of the lookup.

```ts
const createGenericTitle = () => `Meeting Started`
{action: createGenericTitle}
```

+1 camelCase. PascalCase is reserved for constructor-like entities (components, GraphQLObjects, client mutations. tbh I'm not sure why client mutations gets that treatment, but that's what facebook does so we all just started doing it like :sheep: )

+1 prefer to just provide `meeting.name`. treat function like the cops, don't give em any more than they need :police_officer:

packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
```diff
@@ -43,6 +44,45 @@ const notifyMSTeams = async (
...
+  action: GENERIC_MEETING_TITLE,
+  poker: GENERIC_MEETING_TITLE,
+  retrospective: GENERIC_MEETING_TITLE,
+  teamPrompt: createTeamPromptMeetingTitle
+}
+
+const createGenericMeetingAction = (meetingUrl: string) => {
+  const joinMeetingAction = new AdaptiveCards.OpenUrlAction()
```
mattkrick, line 59: :scream: did...did a C# dev write this whole package? goodness gracious look at those class extensions & nullary constructors! no action needed, just venting :slightly_smiling_face:

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
```diff
@@ -100,6 +107,61 @@ const makeEndMeetingButtons = (meeting: Meeting) => {
   }
 }
 
+const createLinkSectionContent = (
+  meetingId: string,
+  meetingUrl: string,
+  meetingShortLink?: string
+) => {
+  const shortLink = meetingShortLink || `https:/prbl.in/${meetingId}`
```
mattkrick, line 115: -2 `prbl.in` goes to /invitation-link (please fix in other file, too)

packages/server/graphql/mutations/helpers/notifications/MattermostNotifier.ts
```diff
@@ -87,6 +90,71 @@ const makeEndMeetingButtons = (meeting: Meeting) => {
...
+        title: 'Meeting started 👋',
+        title_link: meetingUrl
+      }
+    )
+  ]
+}
+
+const makeStartMeetingNotificationLookup: Record<
+  MeetingTypeEnum,
+  (team: Team, meeting: Meeting, meetingUrl: string) => FieldAttachment[]
```
mattkrick, line 150: +1

personally, i don't like exporting anything I don't have to.
you could replace `FieldAttachment` with `ReturnType<typeof makeGenericStartMeetingNotification>` and then if  `makeGenericStartMeetingNotification` ever changes its signature you don't have to update anything. you could also stop exporting `FieldAttachment`. up to you

mattkrick (changes requested):
looking good! Just 1 big thing, the rest is all little stuff

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
BartoszJarocki: @mattkrick could you elaborate on this a bit more? previously link was created either using
```
makeSection(`*Link:*\n<${meetingUrl}|https:/prbl.in/${meeting.id}>`),
``` 
or
```
const maybeMeetingShortLink = makeAppURL(process.env.INVITATION_SHORTLINK!, `${meeting.id}`)
makeSection(`*Link:*\n<${meetingUrl}|${maybeMeetingShortLink}>`),
```
so unless I'm missing something obvious, I haven't changed anything in the logic here

BartoszJarocki: Screenshot 2022-07-27 at 18 21 21

Seems like everything after `|` is just a mask for a real link (`meetingUrl`). The screenshot is from the current prod impl.

mattkrick: If that's the case then prod is wrong :slightly_smiling_face: 

`prbl.in == action.parabol.co/invitation-link` (we picked `.in` because it's short for invitation :yum: )
from the screenshot it looks like the link is good, but the link label is bad. If that's the case, we can make it a friendly link label, or use [appOrigin](https://github.com/ParabolInc/parabol/blob/f221880dac24332874939b6c0a3f7328f4b76023/packages/server/appOrigin.ts) to make it the real URL.

or since there's a button that does the same thing as the link, maybe we just remove the url?

any of those work for me! Just feels weird having the link label look like a url, but be invalid

BartoszJarocki: it feels weird for sure, thank you for catching this! as this touches _every_ notification that we send, I don't want to decide on my own. I feel like having a visible link could be useful for easy copy/paste. on the other hand, having a long URL here also doesn't feel good. @enriquesanchez @ackernaut any thoughts?

enriquesanchez: Yep, the intention for having the shortened URL there was to allow for easy copy/paste in case someone needed it to invite others. It's definitely a _nice-to-have_ and not critical, if there are strong tensions against it I don't have a preference.

That said, if we change it we should probably look at the other meeting notifications that also include the shortened link URL.

Captura de Pantalla 2022-07-27 a la(s) 4 09 40 p m

ackernaut: The visible URL is short, but when you hover it, it’s long? Is that true? I would want them both to match. I like the short version. I agree the long version may be too much to print here.

BartoszJarocki: @ackernaut @enriquesanchez, sorry for not being clear enough. I like to short version too. the problem is, it's actually wrong. for example, I just copied the URL from our squad check-in meeting - https:/prbl.in/fiOGvDz7sA. when you try to open that URL, it'll say `invitation invalid` as we're trying to use the invitation URL format with `meetingId`. The question is, how do we want to fix that? It's not specific to this PR, all URLs we send in notifications on production are technically broken. they work if you click on them as we display the short URL only as a mask, the actual link is the long, correct URL. if you copy it by hand, it won't work. 

I also feel like it shouldn't block this issue. This PR doesn't introduce any regressions, and we really need to have it merged before the next release. @mattkrick are you ok with me creating a new issue for the URL problem and moving on with that PR?

ackernaut: I see. If you didn’t make changes to the link part and want to move on, okay by me. I think the short URL should work and the format that’s printed in the view is also what would be copied and used in the browser (or typed out if you were to read it for some reason). We can aim for that AC in another issue, or we can just yank it from the design as it’s confusing and broken otherwise.

packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
BartoszJarocki: yeah, all good points, fixed!

packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
BartoszJarocki: yeah, I was thinking the same 😅

packages/server/graphql/mutations/helpers/notifications/MattermostNotifier.ts
BartoszJarocki: fixed! except for doing `ReturnType<typeof makeGenericStartMeetingNotification>` I did `ReturnType<typeof makeFieldsAttachment>[]` to make `makeStartMeetingNotificationLookup` compatible with both `makeTeamPromptStartMeetingNotification` and `makeGenericStartMeetingNotification`

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
BartoszJarocki: I created a separate issue where we can discuss how to fix it - https://github.com/ParabolInc/parabol/issues/6968

BartoszJarocki:
@mattkrick all fixed, except for links that I created a separate issue for. we're releasing standups as a beta next week so I'd appreciate if you could take a look, thank you!

mattkrick:
LGTM! comments added on the new issue

Issue #6854, opened 2022-07-06
Standups: Activity summary implementation


BartoszJarocki:
Designs - https://github.com/ParabolInc/parabol/issues/6801

AC:
- [ ] Ending the meeting from within the meeting navigates to the meeting summary
- [ ] The meeting summary can be accessed from the timeline

Issue #5927, opened 2022-01-21
Designed timeline summary card for async stand-ups


ackernaut:
When the async standup ends, what do we show on the timeline? This issue should address only the needful for the MVP.

### Acceptance criteria
Designed a concept of a summary card on the timeline that addresses these questions:
- This shows once the meeting is closed?
- Does this link to the closed activity?
- What other meta might we show?
- What might we show in the future? (This can be a high-level, stretch goal)
- What might be the difference between a one-time activity and a recurring activity? (This can be a high-level, stretch goal)

### Estimated effort
- 3 hours

ackernaut:
Here’s what I propose:
- [See the Figma concept](https://www.figma.com/file/bhiEOV4HK9Kv91HFsJmevV/Stand-up-Concepts?node-id=1193%3A74353)
- We use the `group_work` icon. If we have a better idea this is easy enough to change later
- We largely follow the formatting of other summary cards for now
- We emphasize responses on the first line, and thread activity on the second line
- The link takes the user to the activity view. That is the way to follow up with the threads e.g. you had a question or reaction to what I shared etc.
- Important: I propose that all responses and threads remain alive and editable after the activity has ended. This probably has an impact implementation and there’s not consensus on this UX. I think it’s safe to try at any rate.

In the future, not part of this design:
- I think we will use in-app and chat notifications to ping folks when there are new responses and comments, bringing them back to the meeting
- In the future I think the summary card may be a little richer, maybe showing some avatars, maybe showing if there are conversations that are still going, etc.

ackernaut:
Marking this as done, to be implemented in #5800 (AC and EE to be determined) unless there’s a better issue cc @BartoszJarocki

jmtaber129:
> I propose that all responses and threads remain alive and editable after the activity has ended.

@ackernaut should the summary card be updated when responses and comments are added after the meeting ends?  E.g. if 2 responses and 6 comments were added before the meeting is ended, and then a straggler adds their response and 2 comments later, should the summary card show 2 responses + 6 comments or 3 responses + 8 comments?

ackernaut:
@jmtaber129 I think a snapshot of static meta when the meeting ends is a fine place to start. Later we may find some use case for having dynamic meta. We can revisit it when we think about recurring meetings.

Dschoordsch:
[...]

PR #6885, opened 2022-07-14
feat(standups): Summary Page

jmtaber129:
# Description
Fixes https://github.com/ParabolInc/parabol/issues/6854
Note that this does not include email summaries due to some limitations with our server-side rendering of FE components used elsewhere in the app (specifically, the tiptap response editor).

## Demo
https://www.loom.com/share/ab6f284ac73542f88d7fb35c7b247ede

## Testing scenarios
- [ ] Create a new standup meeting
- [ ] Add some responses
- [ ] Add some discussion comments, including tasks
- [ ] End the meeting
- [ ] Meeting summary should be displayed, with the correct stats numbers, and the responses.

## Final checklist

### DIFF
--- a/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/TeamPromptResponseSummary.tsx
+++ b/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/TeamPromptResponseSummary.tsx
@@ -1,0 +1,63 @@
+import styled from '@emotion/styled'
+import graphql from 'babel-plugin-relay/macro'
+import {TeamPromptResponseSummary_meeting$key} from 'parabol-client/__generated__/TeamPromptResponseSummary_meeting.graphql'
+import React from 'react'
+import {useFragment} from 'react-relay'
+import getPhaseByTypename from '~/utils/getPhaseByTypename'
+import {isNotNull} from '~/utils/predicates'
+import sortByISO8601Date from '~/utils/sortByISO8601Date'
+import TeamPromptResponseSummaryCard from './TeamPromptResponseSummaryCard'
+
+const ResponseSummaryGrid = styled('div')({
+  display: 'flex',
+  flexWrap: 'wrap',
+  padding: '24px',
+  justifyContent: 'center'
+})
+
+interface Props {
+  meetingRef: TeamPromptResponseSummary_meeting$key
+}
+
+const TeamPromptResponseSummary = (props: Props) => {
+  const {meetingRef} = props
+  const meeting = useFragment(
+    graphql`
+      fragment TeamPromptResponseSummary_meeting on TeamPromptMeeting {
+        phases {
+          ... on TeamPromptResponsesPhase {
+            __typename
+            stages {
+              id
+              response {
+                plaintextContent
+                createdAt
+              }
+              ...TeamPromptResponseSummaryCard_stage
+            }
+          }
+        }
+      }
+    `,
+    meetingRef
+  )
+  const {phases} = meeting
+  const phase = getPhaseByTypename(phases, 'TeamPromptResponsesPhase')
+  const allStages = phase.stages.filter(isNotNull)
+
+  const orderedNonEmptyStages = allStages
+    .filter((stage) => !!stage.response?.plaintextContent)
+    .sort((stageA, stageB) =>
+      sortByISO8601Date(stageA.response!.createdAt, stageB.response!.createdAt)
+    )
+
+  return (
+    <ResponseSummaryGrid>
+      {orderedNonEmptyStages.map((stage) => {
+        return <TeamPromptResponseSummaryCard key={stage.id} stageRef={stage} />
+      })}
+    </ResponseSummaryGrid>
+  )
+}
+
+export default TeamPromptResponseSummary
--- a/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/TeamPromptResponseSummary.tsx
+++ b/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/TeamPromptResponseSummary.tsx
@@ -1,5 +1,6 @@
 import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
+import useEmailItemGrid from 'parabol-client/hooks/useEmailItemGrid'
 import {TeamPromptResponseSummary_meeting$key} from 'parabol-client/__generated__/TeamPromptResponseSummary_meeting.graphql'
 import React from 'react'
 import {useFragment} from 'react-relay'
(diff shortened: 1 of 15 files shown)


packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/TeamPromptResponseSummary.tsx
```diff
@@ -0,0 +1,63 @@
...
+import {isNotNull} from '~/utils/predicates'
+import sortByISO8601Date from '~/utils/sortByISO8601Date'
+import TeamPromptResponseSummaryCard from './TeamPromptResponseSummaryCard'
+
+const ResponseSummaryGrid = styled('div')({
+  display: 'flex',
+  flexWrap: 'wrap',
+  padding: '24px',
+  justifyContent: 'center'
+})
```
BartoszJarocki, line 16: +1 to make future reuse in emails easier wouldn't it be better to use https://github.com/ParabolInc/parabol/blob/22c7a81694d8d4acb75ea5c57e322d301c828516/packages/client/hooks/useEmailItemGrid.tsx#L10 or is it something you already explored?

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/TeamPromptResponseSummaryCard.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const {picture, preferredName} = teamMember
+  const contentJSON = response ? JSON.parse(response.content) : null
+  return (
+    <ResponseSummaryCard>
+      <ResponseHeader>
+        <Avatar picture={picture} size={48} />
+        <TeamMemberName>{preferredName}</TeamMemberName>
+      </ResponseHeader>
+      <PromptResponseWrapper>
+        <PromptResponseEditor autoFocus={false} content={contentJSON} readOnly={true} />
```
BartoszJarocki, line 67: +1 should we [generate html](https://tiptap.dev/guide/output#option-2-generate-html-from-prosemirror-json) from `ContentJSON` and avoid using `PromptResponseEditor` which isn't SSR friendly?

BartoszJarocki (approved):
lgtm! I left some suggestions about ssr to consider

jmtaber129:
I did some investigation, and it looks like SSR _would_ be possible with `@tiptap/html`'s `generateHTML`, but there's a few hurdles to getting full SSR working as-is:
* Typings on `@tiptap/html` don't seem to match the necessary extensions - this might just be a version mismatch
* Would need to replicate prompt response styling for the summary view
* Would need to do some refactoring to use the `SummaryAvatarHeader` needed for SSR of avatars.

Because of this, I'd prefer to get this landed as-is and tackle SSR for emails as a separate issue

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/TeamPromptResponseSummary.tsx
```diff
@@ -0,0 +1,66 @@
...
+    meetingRef
+  )
+  const {phases} = meeting
+  const phase = getPhaseByTypename(phases, 'TeamPromptResponsesPhase')
+  const allStages = phase.stages.filter(isNotNull)
+
+  const orderedNonEmptyStages = allStages
+    .filter((stage) => !!stage.response?.plaintextContent)
+    .sort((stageA, stageB) =>
+      sortByISO8601Date(stageA.response!.createdAt, stageB.response!.createdAt)
```
mattkrick, line 52: 0 curious, why doesn't `a < b ? -1 : 1` work?

packages/server/graphql/types/TeamPromptMeeting.ts
```diff
@@ -41,6 +43,49 @@ const TeamPromptMeeting = new GraphQLObjectType<any, GQLContext>({
...
+        const meeting = await dataLoader.get('newMeetings').load(meetingId)
+        if (meeting.meetingType !== 'teamPrompt') {
+          return 0
+        }
+        const {phases} = meeting
+        const discussPhase = getPhase(phases, 'RESPONSES')
+        const {stages} = discussPhase
+        const discussionIds = stages.map((stage) => stage.discussionId)
+        const r = await getRethink()
+        return r
```
mattkrick, line 81: -1 these aggregation queries actually end up being _real_ expensive since we fetch them for the timeline event. IIRC we shaved off almost a whole second on the load time for the timeline view by caching these.
Can we cache them on the meeting object like we do for the other meeting types?

mattkrick (changes requested):

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/TeamPromptResponseSummary.tsx
jmtaber129: The result is unstable when `a === b` - you need to return `0` when things are equal to maintain the order

packages/server/graphql/types/TeamPromptMeeting.ts
jmtaber129: @mattkrick These actually only run for the summary itself - there are similar resolvers for these on the existing timeline event.

We _could_ cache these, but the result would be inaccurate if any of the stats change after the meeting is ended (which is quite possible for things like standup discussions), and we made the decision that the timeline should reflect the current state, not the state when the meeting was ended.  See https://github.com/ParabolInc/parabol/issues/5927#issuecomment-1128113425

mattkrick: > These actually only run for the summary itself
Any access to a meeting object will be able to access these, and the timeline event links to the meeting object

> there are similar resolvers for these on the existing timeline event.

could you include a link to where you see this? This field is cached for the other 3 meeting types, e.g.:
https://github.com/ParabolInc/parabol/blob/8409e5ee0ab86c9effcb616dd82cdf30053c81ee/packages/server/graphql/types/ActionMeeting.ts#L52

> we made the decision that the timeline should reflect the current state

we'll need to either undo that decision or create dataloaders for these queries & update the other 3 meeting types so they all have similar behavior. With the number of timeline events that a standup will be creating, I'd opt for the former since each timeline event is going to have to run these aggregation queries & we're gonna run into performance issues again.

jmtaber129: That's fair.

In any case, this issue is somewhat tangential to this PR, since the timeline view currently uses [different resolvers](https://github.com/ParabolInc/parabol/blob/50f7d8268eff5906f6b6a849f6fef1060f49d28d/packages/server/graphql/public/types/TimelineEventTeamPromptComplete.ts#L25-L65), so I'd prefer to address this in a separate issue.

mattkrick: ohhhhh shoot we em in 2 places!
so we've got `taskCount, responseCount, commentCount` that should all live in 1 place together.
I'd say that place should be on the TeamPromptMeeting object, since those are aggregation stats about the meeting & in the future we might want to access those without accessing the timelineevent.

What we don't want is having them exist on 2 objects (TimelineEvent & Meeting) because that removes the single source of truth.

I'm all for moving the caching logic to a new PR since that'll be a little more involved. For this PR could you remove those fields from the timelineEvent object so this PR doesn't introduce a 2nd source of truth?

jmtaber129: Ahh that makes a lot of sense.  Done.

PR #6886, opened 2022-07-14
fix(retro): Give users feedback that they need to vote in Vote phase

### LINKED ISSUE #6631: Give users feedback that they need to vote in Vote phase
In the Reflect phase, if the user doesn't add any reflections and tries to proceed to the Group phase, we give them feedback by preventing them from clicking "Next" in the control bar. "Next" is greyed out and there's a no entry icon on hover:
control-bar

In the Voting phase, if you don't vote on a group, you can't proceed to the Discuss phase, but the control bar doesn't give them any indication that that's the case.

control-bar-two

### AC

- [ ] If the user hasn't voted in the Vote phase, grey out the "Next" button and show a no entry icon on hover

igorlesnenko:
Fixes #6631

https://user-images.githubusercontent.com/466991/179980001-24b8e29e-cdca-4c23-8607-127019bfcee2.mov

_Please let me know if I need to change tooltip text_

**How to test:**

- Create a retro, go to vote phase
- See you can't click next until some votes appeared
- See a tooltip with a hint
- Smoke test another meeting type

### DIFF
--- a/packages/client/components/BottomControlBarReady.tsx
+++ b/packages/client/components/BottomControlBarReady.tsx
@@ -159,17 +159,15 @@ const BottomControlBarReady = (props: Props) => {
     return false
   }
   const disabled = getDisabled()
+  const onMouseOver = () => {
+    if (phaseType === 'vote' && disabled) {
+      openErrorTooltip()
+    }
+  }
+
   return (
     <>
-      <div
-        onMouseOver={() => {
-          if (phaseType === 'vote' && disabled) {
-            openErrorTooltip()
-          }
-        }}
-        onMouseLeave={closeErrorTooltip}
-        ref={errorTooltipRef}
-      >
+      <div onMouseOver={onMouseOver} onMouseLeave={closeErrorTooltip} ref={errorTooltipRef}>
         <BottomNavControl
           dataCy={`next-phase`}
           disabled={disabled}
--- a/packages/client/components/BottomControlBarReady.tsx
+++ b/packages/client/components/BottomControlBarReady.tsx
@@ -44,6 +44,26 @@ const CheckIcon = styled(Icon)<{progress: number; isNext: boolean; isViewerReady
   })
 )
 
+// Disabled button does not fire mouse events, this is a workaround for showing a tooltip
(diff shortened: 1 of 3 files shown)


JimmyLv (approved):
LGTM!

JimmyLv (approved):

igorlesnenko:
@Dschoordsch The issue occurs because `disabled` button does not fire mouse events. Wrapping it in a `div `only partially solves the problem. 

I got around that problem by using `pointerEvents: none`. 

https://user-images.githubusercontent.com/466991/179980001-24b8e29e-cdca-4c23-8607-127019bfcee2.mov

Alternative solution: #6907

packages/client/components/BottomControlBarReady.tsx
```diff
@@ -132,29 +171,48 @@ const BottomControlBarReady = (props: Props) => {
...
+          <BottomControlBarProgress isNext={isNext} progress={progress} />
+          <BottomNavIconLabel label={label} ref={originRef}>
+            <CheckIcon isViewerReady={isViewerReady} isNext={isNext} progress={progress}>
+              {icon}
+            </CheckIcon>
+          </BottomNavIconLabel>
+        </BottomNavControl>
+      </BottomNavContorlWrapper>
       {tooltipPortal(`Tap 'Next' again to Confirm`)}
+      {errorTooltipPortal(`At least one vote required`)}
```
tianrunhe, line 215: I feel this message not quite user friendly but I don't know what would be a better version... tagging @avivapinchas

avivapinchas: How about: 
> Need to vote at least once to advance

I'm not a huge fan of this one either, but not sure what would be better. 

This is perhaps a bigger question, but: why do they need to vote at least once? It seems like an un-needed gate.

I find when copy is hard to write, it's often because the thing we're trying to explain doesn't make sense on some level.

igorlesnenko: Will modify the PR to just allow users to advance to the next phase without voting

igorlesnenko:
Another solution implemented instead #6966

Issue #6884, opened 2022-07-14
Upgrade Stripe


nickoferrall:
We're currently using Stripe version `4.24.0`, but the [latest version](https://github.com/stripe/stripe-node) available is 9.13.0. 

By upgrading, we should get better types, and reduce the chance that bugs or security issues creep in.

### AC

- [ ] Upgrade Stripe to the latest version
- [ ] Fix any issues that are caused by the upgrade
- [ ] Upgrades work
- [ ] Downgrades work
- [ ] Seat changes work
- [ ] Invoices work

PR #6895, opened 2022-07-19
chore: upgrade stripe

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6884

Upgrades Stripe from `4.24.0` to `9.13.0`

### To test

- [ ]  Upgrades work
- [ ]  Downgrades work
- [ ]  Seat changes work
- [ ]  Invoices work

### DIFF
--- a/packages/server/billing/helpers/generateInvoice.ts
+++ b/packages/server/billing/helpers/generateInvoice.ts
@@ -246,11 +246,22 @@ const makeItemDict = (stripeLineItems: Stripe.InvoiceLineItem[]) => {
         nextPeriodCharges.amount = nextPeriodCharges.amount || amount
         nextPeriodCharges.quantity = nextPeriodCharges.quantity || lineItemQuantity
       }
-    } else if (!metadata.type) {
-      unknownLineItems.push(lineItem)
     } else {
-      // at this point, we don't care whether it's an auto pause or manual
-      addToDict(itemDict, lineItem)
+      if (!nextPeriodCharges) {
+        nextPeriodCharges = new NextPeriodCharges({
+          amount,
+          quantity: lineItemQuantity,
+          nextPeriodEnd: fromEpochSeconds(end),
+          unitPrice: lineItem.plan?.amount || undefined,
+          interval: lineItem.plan?.interval || 'month'
+        })
+      }
+      if (!metadata.type) {
+        unknownLineItems.push(lineItem)
+      } else {
+        // at this point, we don't care whether it's an auto pause or manual
+        addToDict(itemDict, lineItem)
+      }
     }
   }
   return {itemDict, nextPeriodCharges, unknownLineItems}
@@ -324,21 +335,22 @@ export default async function generateInvoice(
   ].sort((a, b) => (a.type > b.type ? 1 : -1))
 
   // sanity check
-  const calculatedTotal =
-    invoiceLineItems.reduce((sum, {amount}) => sum + amount, 0) + nextPeriodCharges.amount
-  if (calculatedTotal !== invoice.total) {
-    console.warn(
-      'Calculated invoice does not match stripe invoice',
-      invoiceId,
-      calculatedTotal,
-      invoice.total
-    )
-  }
+  // const calculatedTotal =
+  //   invoiceLineItems.reduce((sum, {amount}) => sum + amount, 0) + nextPeriodCharges.amount
+  // if (calculatedTotal !== invoice.total) {
+  //   console.warn(
+  //     'Calculated invoice does not match stripe invoice',
+  //     invoiceId,
+  //     calculatedTotal,
+  //     invoice.total
+  //   )
+  // }
 
   const [type] = invoiceId.split('_')
   const isUpcoming = type === 'upcoming'
 
   let status: InvoiceStatusEnum = isUpcoming ? 'UPCOMING' : 'PENDING'
+  console.log('🚀 ~ invoice', {invoice, status, isUpcoming})
 
   if (status === 'PENDING') {
     status = invoice.paid ? 'PAID' : 'FAILED'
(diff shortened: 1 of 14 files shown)


packages/server/billing/helpers/generateInvoice.ts
```diff
@@ -227,42 +227,53 @@ const makeItemDict = (stripeLineItems: Stripe.invoices.IInvoiceLineItem[]) => {
...
         nextPeriodCharges.amount = nextPeriodCharges.amount || amount
-        nextPeriodCharges.quantity = nextPeriodCharges.quantity || quantity
+        nextPeriodCharges.quantity = nextPeriodCharges.quantity || lineItemQuantity
       }
-    } else if (!metadata.type) {
-      unknownLineItems.push(lineItem)
     } else {
-      // at this point, we don't care whether it's an auto pause or manual
-      addToDict(itemDict, lineItem)
+      if (!nextPeriodCharges) {
```
nickoferrall, line 250: If `description` is not null on line 234, `nextPeriodCharges` will be undefined even though we guarantee that it's non-nullable on line 222. So, if it's undefined, I'm now creating a new `nextPeriodCharges`

packages/server/billing/helpers/generateInvoice.ts
```diff
@@ -339,7 +350,7 @@ export default async function generateInvoice(
   const isUpcoming = type === 'upcoming'
 
   let status: InvoiceStatusEnum = isUpcoming ? 'UPCOMING' : 'PENDING'
-  if (status === 'PENDING' && invoice.closed === true) {
```
nickoferrall, line 342: The `invoice` object no longer has the `closed` field. I believe our subscriptions always charge the card automatically, although there may be exceptions with Enterprise? If subscriptions are always charged automatically, the invoice is `PENDING`, but not paid, I think it's safe to assume it's a `FAILED` payment.

Stripe has created [invoice statuses](https://stripe.com/docs/invoicing/overview#invoice-statuses), so we could refactor our [InvoiceStatusEnum](https://github.com/ParabolInc/parabol/blob/dd8e82c239be9526a5ee6f443fea4b9aeff51e2d/packages/server/database/types/Invoice.ts#L6) in a separate PR to use this?

packages/server/utils/stripe/StripeManager.ts
```diff
@@ -72,28 +72,20 @@ export default class StripeManager {
...
-  async updateSubscriptionQuantity(
-    stripeSubscriptionId: string,
-    quantity: number,
-    prorationDate?: number
-  ) {
-    return this.stripe.subscriptions.update(stripeSubscriptionId, {
-      quantity,
-      proration_date: prorationDate
+  async getSubscriptionItem(subscriptionId: string) {
+    const allSubscriptionItems = await this.stripe.subscriptionItems.list({
```
nickoferrall, line 76: Stripe now has a [SubscriptionItem](https://stripe.com/docs/api/subscription_items) object and `quantity` no longer exists on the [subscription](https://stripe.com/docs/api/subscriptions/object) object. We only have one `subscriptionItem` per `subscription`, so when we want to get the subscription quantity, we can call `getSubscriptionItem`. 

I named it `get` instead of `retreiveSubscriptionItem` to signal that there's some additional logic going on here, unlike the other API calls that simply reach out to Stripe.

packages/server/utils/stripe/StripeManager.ts
```diff
@@ -108,25 +100,49 @@ export default class StripeManager {
...
+    hookId: string
+  ) {
+    return this.stripe.invoiceItems.update(invoiceItemId, {metadata: {type, userId, hookId}})
   }
 
   async updatePayment(customerId: string, source: string) {
     return this.stripe.customers.update(customerId, {source})
   }
+
+  async updateSubscriptionItemQuantity(
```
nickoferrall, line 138: Moved this so that everything is in alphabetical order and eases my OCD 😅

packages/server/billing/helpers/generateInvoice.ts
mattkrick: IIRC after stripe finalizes an invoice, it waits an hour or 2 until it creates a charge.
During that timeframe would invoices look like they're failed?

packages/server/graphql/mutations/helpers/getCCFromCustomer.ts
```diff
@@ -1,12 +1,19 @@
 import Stripe from 'stripe'
+import {getStripeManager} from '../../../utils/stripe'
+
+export default async function getCCFromCustomer(customer: Stripe.Customer) {
+  if (typeof customer.default_source !== 'string') {
+    throw new Error(`Could not find default_source for ${customer.id}`)
+  }
+  const manager = getStripeManager()
+  const card = await manager.retrieveSource(customer.id, customer.default_source)
```
mattkrick, line 9: -2 now that this is async we'll need to fix the following:
 https://github.com/ParabolInc/parabol/blob/4172e4446fb0b6206fa585a36d037476899e6330/packages/server/graphql/private/mutations/stripeUpdateCreditCard.ts

more importantly-- why didn't typescript tell us about this error?! could you run `yarn typecheck` in the server package & triage why?

mattkrick (changes requested):
really great work! just 1 big thing that typescript should've caught--- i'm more worried about why it didn't catch it than the mistake itself.

I'll do QA tomorrow, my brain is a little fried this evening

packages/server/billing/helpers/generateInvoice.ts
nickoferrall: Interesting! I'm now fetching the `charge` and checking the status there

packages/server/graphql/mutations/helpers/getCCFromCustomer.ts
nickoferrall: How strange! I ran `yarn typecheck` in the server package, but there were no complaints. I played around with the config but didn't get any change in behaviour

mattkrick: issue created for debugging: https://github.com/ParabolInc/parabol/issues/6990

mattkrick:
code looks good! moving to QA...

mattkrick:
Screenshot from 2022-08-02 18-38-39

I forgot how serious our failed payments look :rofl: 

LGTM! great work, this upgrade has been on my mind for, well, literally 5 years

Issue #6897, opened 2022-07-19
Prevent new meeting if any team member belongs to a locked team


nickoferrall:
See issue https://github.com/ParabolInc/parabol/issues/6378 for further context.

### AC

- [ ] Prevent new meeting if any team member belongs to a locked team
- [ ] Let the user know why they're unable to start a new meeting

Issue #6378, opened 2022-04-10
Locking enhanced to be at User level, not Team level


jordanh:
Sales are observing the phenomenon that users are bypassing our org-level over-limit locking in significant numbers. Users are creating new org/team combinations to continue running meetings. Flattering for sure, but this makes forcing a conversation difficult.

This issue exists to explore an alternative locking strategy, in which we can lock out users at the user level _in addition_ to the org level.

### Acceptance Criteria (optional)
An architectural design, written as a GitHub issue that:
   - Allows sales to run a single mutation that would continue to lock an org
   - Enhances our Hubspot integration to set a new Hubspot property indicating that user is locked out
   - Also lock users at the user level to prevent them from starting additional meetings, even if they make additional orgs
      - can they still create an org?
      - do we display the lockout message on any new teams they create?
   - An exploration of alternative ideas that may come to mind

**Estimated effort:** Up to 8 hours to explore and write.

nickoferrall:
Here are a few possible approaches:

### 1. Update `lockTeams` to lock all users within a team by default

The `lockTeams` mutation could include an optional `lockUserIds` arg. If null/undefined, it would lock all users within the team. If the array includes user ids, only those ids are locked. If the array is empty, no users are locked. 

We would add `isLocked` to the User object. Locked users wouldn't be able to create a new team/org, but could create a meeting.

If the `lockTeams` arg `isPaid` is true, it would unlock all users.

Pros:

- We wouldn't need to create a new mutation

Cons:

- It's unclear that `lockTeams` would also lock users so it might need to be renamed to `lockTeamsAndUsers`, which feel spaghetti-y
- It could be expensive to update all of the users in an org
- We wouldn't be able to lock specific users
- Users that have already dodged the rules and created a new team/org wouldn't be affected 

### 2. Create `lockUsers` and `unlockUsers` mutations

Create two new mutations: 
- `lockUsers(teamIds: string[] | null, userIds: string[] | null, lockMeetings: boolean)`. 
- `unlockUsers(teamIds: string[] | null, userIds: string[] | null)`

Add `isLocked` and `isMeetingsLocked` to the User object. 

In `lockUsers`, if `teamIds` are specified, all users within those teams are locked. If `userIds` are specified, it only locks those users. If `lockMeetings` is false, only `isLocked` is updated, which prevents users from creating new teams/orgs. If `lockMeetings` is true, those users can't create new meetings either.

[...]

PR #6900, opened 2022-07-19
feat: increase locking restrictions

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6897

This PR prevents a user from creating a meeting if any of the members of that team belong to a locked team. It largely uses the logic shared [here](https://github.com/ParabolInc/parabol/issues/6378#issuecomment-1185777575), with a few changes. I've added comments where the logic was changed to help with the CR.

### To test

- [ ] Teams that don't contain a user that belongs to a locked team can create a meeting successfully
- [ ] If the team is locked, the user can't start a meeting
- [ ] If a team in the org is locked, the user can't start a meeting
- [ ] If a team member in the team belongs to a team that is locked, the user can't start a meeting
- [ ] The user sees a helpful error message if they're locked & unable to start a meeting

### DIFF
--- a/packages/client/mutations/AcceptTeamInvitationMutation.ts
+++ b/packages/client/mutations/AcceptTeamInvitationMutation.ts
@@ -17,6 +17,8 @@ import handleAddTeamMembers from './handlers/handleAddTeamMembers'
 import handleAddTeams from './handlers/handleAddTeams'
 import handleAuthenticationRedirect from './handlers/handleAuthenticationRedirect'
 
+export const LOCKED_OVERDUE_MSG = `Sorry! You're unable to join this team because one of your teams has an overdue payment`
+
 graphql`
   fragment AcceptTeamInvitationMutation_team on AcceptTeamInvitationPayload {
     teamMember {
@@ -190,10 +192,13 @@ const AcceptTeamInvitationMutation: StandardMutation<
             autoDismiss: 0,
             key: `acceptTeamInvitation:${message}`,
             message,
-            action: {
-              label: 'OK',
-              callback: () => history.push(`/me`)
-            }
+            action:
+              message === LOCKED_OVERDUE_MSG
+                ? {
+                    label: 'Contact Sales',
+                    callback: () => window.open('mailto:love@parabol.co?subject=Overdue Payment')
+                  }
+                : undefined
           })
         }
         if (!ignoreApproval) return
--- a/packages/client/mutations/AcceptTeamInvitationMutation.ts
+++ b/packages/client/mutations/AcceptTeamInvitationMutation.ts
@@ -180,19 +180,26 @@ const AcceptTeamInvitationMutation: StandardMutation<
       const {authToken, team} = acceptTeamInvitation
       const serverError = getGraphQLError(data, errors)
       if (serverError) {
-        if (serverError.message === InvitationTokenError.ALREADY_ACCEPTED) {
+        const message = serverError.message
+        if (message === InvitationTokenError.ALREADY_ACCEPTED) {
           handleAuthenticationRedirect(acceptTeamInvitation, {
             atmosphere,
             history,
             meetingId: locallyRequestedMeetingId
           })
+        } else if (!ignoreApproval) {
+          atmosphere.eventEmitter.emit('addSnackbar', {
+            autoDismiss: 0,
+            key: `acceptTeamInvitation:${message}`,
+            message,
+            action: {
+              label: 'OK',
+              callback: () => history.push(`/me`)
+            }
+          })
         }
-        return
+        if (!ignoreApproval) return
       }
-      const isOK = ignoreApproval
-        ? true
-        : handleAcceptTeamInvitationErrors(atmosphere, acceptTeamInvitation)
-      if (!isOK) return
       atmosphere.setAuthToken(authToken)
       if (!team) return
       const {id: teamId, name: teamName} = team
--- a/packages/client/mutations/AcceptTeamInvitationMutation.ts
+++ b/packages/client/mutations/AcceptTeamInvitationMutation.ts
@@ -144,15 +144,13 @@ export const handleAcceptTeamInvitationErrors = (
 ) => {
   if (acceptTeamInvitation?.error) {
     const {message} = acceptTeamInvitation.error
-    if (message === InvitationTokenError.ALREADY_ACCEPTED) return true
+    if (message === InvitationTokenError.ALREADY_ACCEPTED) return
     atmosphere.eventEmitter.emit('addSnackbar', {
       autoDismiss: 0,
       key: `acceptTeamInvitation:${message}`,
       message
     })
-    return false
   }
-  return true
 }
 
 const AcceptTeamInvitationMutation: StandardMutation<
--- a/packages/client/mutations/AcceptTeamInvitationMutation.ts
+++ b/packages/client/mutations/AcceptTeamInvitationMutation.ts
@@ -1,6 +1,6 @@
 import graphql from 'babel-plugin-relay/macro'
 import {commitMutation} from 'react-relay'
-import {InvitationTokenError} from '~/types/constEnums'
+import {InvitationTokenError, LOCKED_MESSAGE} from '~/types/constEnums'
 import {AcceptTeamInvitationMutation_notification} from '~/__generated__/AcceptTeamInvitationMutation_notification.graphql'
 import Atmosphere from '../Atmosphere'
 import {
@@ -16,8 +16,6 @@ import {AcceptTeamInvitationMutation_team} from '../__generated__/AcceptTeamInvi
 import handleAddTeamMembers from './handlers/handleAddTeamMembers'
 import handleAddTeams from './handlers/handleAddTeams'
 import handleAuthenticationRedirect from './handlers/handleAuthenticationRedirect'
-
-export const LOCKED_OVERDUE_MSG = `Sorry! You're unable to join this team because one of your teams has an overdue payment`
 
 graphql`
   fragment AcceptTeamInvitationMutation_team on AcceptTeamInvitationPayload {
@@ -193,7 +191,7 @@ const AcceptTeamInvitationMutation: StandardMutation<
             key: `acceptTeamInvitation:${message}`,
             message,
             action:
-              message === LOCKED_OVERDUE_MSG
+              message === LOCKED_MESSAGE.TEAM_INVITE
                 ? {
                     label: 'Contact Sales',
                     callback: () => window.open('mailto:love@parabol.co?subject=Overdue Payment')
--- a/packages/server/graphql/mutations/helpers/isStartMeetingLocked.ts
+++ b/packages/server/graphql/mutations/helpers/isStartMeetingLocked.ts
@@ -4,13 +4,15 @@ import {DataLoaderWorker} from '../../graphql'
 const isStartMeetingLocked = async (teamId: string, dataLoader: DataLoaderWorker) => {
   const team = await dataLoader.get('teams').loadNonNull(teamId)
   const {lockMessageHTML, orgId, tier} = team
-  const orgTeams = await dataLoader.get('teamsByOrgIds').load(orgId)
-  const anyOrgTeamIsLocked = orgTeams.some((team) => !team.isPaid)
-  if (tier !== 'personal' && anyOrgTeamIsLocked) {
-    // if this team was manually locked, be mean because they called this by hiding the modal
-    return lockMessageHTML
-      ? 'Wow, you’re determined to use Parabol! That’s awesome! Do you want to keep sneaking over the gate, or walk through the door with our Sales team?'
-      : 'Sorry! We are unable to start your meeting because your organization has an overdue payment'
+  // if this team was manually locked, be mean because they called this by hiding the modal
+  if (lockMessageHTML)
+    return 'Wow, you’re determined to use Parabol! That’s awesome! Do you want to keep sneaking over the gate, or walk through the door with our Sales team?'
+  if (tier !== 'personal') {
+    const orgTeams = await dataLoader.get('teamsByOrgIds').load(orgId)
+    const anyOrgTeamIsLocked = orgTeams.some((team) => !team.isPaid)
+    return anyOrgTeamIsLocked
+      ? 'Sorry! We are unable to start your meeting because your organization has an overdue payment'
+      : null
   }
   // if this team wasn't manually locked, see if any of its members are on locked teams
   const teamMembers = await dataLoader.get('teamMembersByTeamId').load(teamId)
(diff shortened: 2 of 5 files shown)


packages/server/graphql/mutations/helpers/isStartMeetingLocked.ts
```diff
@@ -1,12 +1,30 @@
...
 const isStartMeetingLocked = async (teamId: string, dataLoader: DataLoaderWorker) => {
   const team = await dataLoader.get('teams').loadNonNull(teamId)
-  const {isPaid, lockMessageHTML} = team
-  if (isPaid) return null
-  return lockMessageHTML
-    ? 'Wow, you’re determined to use Parabol! That’s awesome! Do you want to keep sneaking over the gate, or walk through the door with our Sales team?'
-    : 'Sorry! We are unable to start your meeting because your team has an overdue payment'
+  const {lockMessageHTML, orgId, tier} = team
+  const orgTeams = await getTeamsByOrgIds([orgId])
+  const anyOrgTeamIsLocked = orgTeams.some((team) => !team.isPaid)
```
nickoferrall, line 9: I think we should check whether the team is part of an org that hasn't paid. Otherwise, if we just check the team and they create a new team within the same org, they'll be able to use that team as its `tier` will be pro/enterprise. 

Also, why do we lock by team rather than org when the org is the billing entity?

packages/server/graphql/mutations/helpers/isStartMeetingLocked.ts
```diff
@@ -1,12 +1,30 @@
...
+  }
+  // if this team wasn't manually locked, see if any of its members are on locked teams
+  const teamMembers = await dataLoader.get('teamMembersByTeamId').load(teamId)
+  const userIds = teamMembers.map(({userId}) => userId)
+  const allTeamMembers = (await dataLoader.get('teamMembersByUserId').loadMany(userIds))
+    .filter(isValid)
+    .flat()
+  const teamIds = [...new Set(allTeamMembers.map(({teamId}) => teamId))]
+  const allRelatedTeams = (await dataLoader.get('teams').loadMany(teamIds)).filter(isValid)
+  const anyTeamIsLocked = allRelatedTeams.some((team) => team.isPaid === false)
```
nickoferrall, line 24: Rather than checking `team.lockMessageHTML`, I think we should check `team.isPaid` because we can lock a team without including the `lockMessageHTML`.

If a team has paid but a team member belongs to a locked team, the paying team won't be able to start a meeting. I think that's OK - we could remove the team member or unlock the locked team if we think the paying team should be able to start a meeting.

packages/server/graphql/mutations/helpers/isStartMeetingLocked.ts
```diff
@@ -1,12 +1,30 @@
...
+  return anyTeamIsLocked
+    ? 'One or more of your team members is over the usage limit. Please contact sales'
```
nickoferrall, line 26: We could mention the team member(s) & which team(s) are locked in this message. That would be a helpful UX, but it increases the likelihood that they remove that team member / delete the team to get around the rules.

packages/server/graphql/mutations/helpers/isStartMeetingLocked.ts
mattkrick: -1 `isPaid` is denormalized onto the team so if 1 is paid, they're all paid.
100% this should live on the organization, it should be refactored...

all that to say we can avoid that extra DB hit on L8 if tier !== personal

https://github.com/ParabolInc/parabol/blob/4172e4446fb0b6206fa585a36d037476899e6330/packages/server/billing/helpers/terminateSubscription.ts

packages/server/graphql/mutations/helpers/isStartMeetingLocked.ts
```diff
@@ -1,12 +1,30 @@
...
 
 const isStartMeetingLocked = async (teamId: string, dataLoader: DataLoaderWorker) => {
   const team = await dataLoader.get('teams').loadNonNull(teamId)
-  const {isPaid, lockMessageHTML} = team
-  if (isPaid) return null
-  return lockMessageHTML
-    ? 'Wow, you’re determined to use Parabol! That’s awesome! Do you want to keep sneaking over the gate, or walk through the door with our Sales team?'
-    : 'Sorry! We are unable to start your meeting because your team has an overdue payment'
+  const {lockMessageHTML, orgId, tier} = team
+  const orgTeams = await getTeamsByOrgIds([orgId])
```
mattkrick, line 8: -1 why not use a dataloader?

packages/server/graphql/mutations/helpers/isStartMeetingLocked.ts
mattkrick: makes sense!

packages/server/graphql/mutations/helpers/isStartMeetingLocked.ts
mattkrick: heh, i was thinking the same thing! we want them to solve the problem with money, so the less info the better

mattkrick (changes requested):

packages/server/graphql/mutations/helpers/isStartMeetingLocked.ts
nickoferrall: > if 1 is paid, they're all paid

When an org pays, all teams in that org are initially paid. However, when we lock a team, we only update `isPaid` on that team ([see here](https://github.com/ParabolInc/parabol/blob/234bc53534684b29b70e2a4357ba27f42bd1fa20/packages/server/graphql/private/mutations/lockTeams.ts#L16)), so one team could be paid while another isn't. If we don't check whether other teams in the pro/enterprise org are paid, they could keep creating new teams in the org. 

One concern with this approach is if an org has three teams but says they'll never pay, we'd lock the third team so they remain within the two-team free tier. However, with this implementation, they/we would need to delete the third team so that their other two teams can still create meetings. I think that's OK - I'd rather force a conversation with sales than be too lenient.

mattkrick: ah, i gotcha.
@gcrickman when you lock a team with a message is it OK if we lock every team on that org? I forget why we lock teams instead of orgs...

avivapinchas: [This is why: ](https://github.com/ParabolInc/parabol/issues/6378#issuecomment-1190556784)
> A few things can happen:
> - A team lead may choose to pay for their own teams, but then others within the org need to be locked out
> - A company may say 'oh we'll just fit into the free tier' - then we'll lock all but 2 teams, since they can have 2 teams on the free tier

avivapinchas: > we'd need to delete the teams the free-tier org no longer wants to use. Would that work for you?

Can you say more about this? So instead of locking the teams, we'd delete the teams? 

The reason we want to lock is: 
- We want folks trying to get access to get a message about what's happening, so they could request to unlock or trigger a conversation with their own leadership about paying 
- If they do move to paid, they can regain access to their history and data in Parabol

nickoferrall: > The reason we want to lock is:
> - We want folks trying to get access to get a message about what's happening, so they could request to unlock or trigger a conversation with their own leadership about paying
> - If they do move to paid, they can regain access to their history and data in Parabol

This would still be possible. Let's say Cheeky Org has three teams: Team A, Team B, and Team C. They're on the free-tier and don't want to pay, so we lock Team C. Clicking on Team C no longer shows their tasks - it shows the lock modal. Teams A & B can still see their tasks. All three teams can no longer start new meetings and see a helpful error message when they try.

If Cheeky Org does pay, we'll unlock Team C so they have access to their tasks and all teams can create meetings.

If they don't pay, we would archive Team C, so that Teams A & B can start meetings, which is easily done in GraphiQL.

Here's a Loom that demo's how it's currently working: https://www.loom.com/share/8e395fb8b8c94c7fa3ede02cbd12462f

mattkrick: thanks for the reminder @avivapinchas! That makes a bunch of sense.

@nickoferrall I think your logic is sound. Ultimately, we need to move away from the team-based limit & limit on something that compels them to upgrade on their own, but that's another discussion.

avivapinchas: Thanks for that video, super helpful @nickoferrall! 

How I would want this to work is: 
- Teams A & B can still run their meetings
- Members of Team C cannot run meetings *and* cannot create new teams 

The reason for that is: 
- Teams A and B, whether paying or fitting into the free tier, are doing the right thing - we shouldn't put barriers in front of them 
- Team C should visibly see the problem - I don't want to archive them because then whatever attachment they have to their history or tasks or what have you goes away, they have less reason to come back. 

Is it possible to adjust this?

avivapinchas: Actually, I have a clarifying question: If the team members in Team A & B are _not_ members of Team C (there is no overlap in users), can they still run meetings? And: can a team member still leave Team C?

nickoferrall: > If the team members in Team A & B are not members of Team C (there is no overlap in users), can they still run meetings? And: can a team member still leave Team C?

Nope, the current implementation prevents everyone in Cheeky Org from starting meetings. The locked modal would cover the screen preventing users from leaving Team C, but they could leave the org.

The way you would like it to work makes sense to me, but I have one concern:
- catherine@cheeky.org from Team C sees that her team is locked and she can't create a new org/team
- Catherine creates a new user account with email catherine+stingy@cheeky.org
- She creates a new team or org, adds users from Team C, and starts a new meeting
- Alternatively, Catherine asks her colleague in Team A to create a new team/org

We could prevent members of Team C from creating a new team or org by:
- In the [addTeam](https://github.com/ParabolInc/parabol/blob/3939ec06398841ddb42e87c2b3a19683ac112370/packages/server/graphql/mutations/addTeam.ts) and [addOrg](https://github.com/ParabolInc/parabol/blob/3939ec06398841ddb42e87c2b3a19683ac112370/packages/server/graphql/mutations/addOrg.ts) mutations, check if the person creating the team belongs to any locked teams. If they do, show an error message.
- Remove the logic added in this PR to `isStartMeetingLocked`. As is the case in prod currently, members of Team C wouldn't be able to start a meeting, but members of Team A & B could. Members of any other org could too.

For this to be an effective restriction, I think we'd need to prevent users that belong to a locked team from joining new teams or orgs. We could do this by:
- Checking if the user belongs to a locked team in [acceptTeamInvitation](https://github.com/ParabolInc/parabol/blob/218c326c9f6c50f9d1496f2b25041a628bfa0a40/packages/server/safeMutations/acceptTeamInvitation.ts)
- If they do, show the locked modal and don't let them join

The concern here is a consultant could belong to a locked team, have no say in the billing, but need to join another org. They would need to remove themself from Cheeky Org to be able to do that.

----

I see two options:

- The current implementation which prevents teams from creating a new meeting if they have a team member that belongs to a locked org
- Prevent users from creating new teams/orgs and from joining teams if they belong to a locked team

What do you think @avivapinchas @mattkrick?

avivapinchas: I like option 2. 

> The current implementation which prevents teams from creating a new meeting if they have a team member that belongs to a locked team

So would members of Team A&B be able to run meetings, if they had no overlapping team members on Team C?

mattkrick: i've got some tension with sprinkling in too much business logic around this use case, but as long as it doesn't turn into a cat & mouse game & stops here I think we're OK wrt ROI.

I'd argue we should only handle this logic in `accceptTeamInvitation` and NOT in creating new teams/orgs. That'll keep it simpler & teams are pointless if their friends can't join.

I'd add the logic in parallel to this call: https://github.com/ParabolInc/parabol/blob/7279d1fd87288b5629b993fc292ad978baaa9c0d/packages/server/graphql/public/mutations/acceptTeamInvitation.ts#L62

avivapinchas: So what would it loo like for locked users when they go to create new teams/orgs? They could do it, but then when they go to invite, they see the message?

mattkrick: they could create a new team, they could invite new people, but those new people couldn't join. when they tried they would receive a snackbar notification saying something like "You're over limit, freeloader"

avivapinchas: 🤔 I think it's clearer/stronger to show the notice to the person trying to skirt the rules, vs the folks they invited, buuuuttt maybe fun things will happen by doing it this way! Safe to try to move forward with this and see what the impact is.

packages/client/mutations/AcceptTeamInvitationMutation.ts
```diff
@@ -180,19 +178,29 @@ const AcceptTeamInvitationMutation: StandardMutation<
...
-        if (serverError.message === InvitationTokenError.ALREADY_ACCEPTED) {
+        const message = serverError.message
+        if (message === InvitationTokenError.ALREADY_ACCEPTED) {
           handleAuthenticationRedirect(acceptTeamInvitation, {
             atmosphere,
             history,
             meetingId: locallyRequestedMeetingId
           })
+        } else if (!ignoreApproval) {
+          atmosphere.eventEmitter.emit('addSnackbar', {
```
mattkrick, line 189: -1 can we move this logic to `handleAcceptTeamInvitationErrors` since things like loginwithGoogle also call acceptTeamInvitation (see https://github.com/ParabolInc/parabol/blob/218c326c9f6c50f9d1496f2b25041a628bfa0a40/packages/client/mutations/LoginWithGoogleMutation.ts#L28)

as it's written if someone who is logged out got an invite link, pasted it, logged in with google, i don't think they would see the error message, would they?

mattkrick (changes requested):
just 1 client-side thing

packages/client/mutations/AcceptTeamInvitationMutation.ts
nickoferrall: I just double checked and they will still see the locked error message as we return the error from the server and show the message here: https://github.com/ParabolInc/parabol/blob/ee58e1899bb0f04c06921b022e5b4025a94ce4ab/packages/client/mutations/AcceptTeamInvitationMutation.ts#L151

I left a comment here https://github.com/ParabolInc/parabol/pull/6900#discussion_r930911479 mentioning why I'd prefer to separate the logic. If you'd still prefer it to be added in, just let me know!

mattkrick: ahhh gotcha. The only difference I see is that if you're not logged in you don't get the "Contact Sales" CTA, but not a show stopper!

Issue #6429, opened 2022-04-20
Bug: Cannot access task board or timeline when impersonating users


tianrunhe:
## Issue - Bug
Repro steps:
1. Log in as an admin user and impersonate some user
2. Go to their task board or timeline
3. See error

Log from DataDog:
```
SEND TO SENTRY [Error: Variable "$userIds" got invalid value [null]; Expected non-nullable type ID! not to be null at value[0].] undefined
```

### Acceptance Criteria (optional)
  - [ ] Admin user can impersonate user and see their task board

Triage is performed to find the root cause of the bug, timeboxed to ~1 hour. 

**Estimated effort:** 1 hour to triage. More if root cause is already identified.

tianrunhe:
@drewhousman1010 also encountered this issue today. See [Slack thread](https://parabol.slack.com/archives/C4JAUUZ9P/p1651077080259489).

PR #6901, opened 2022-07-19
fix: impersonate user bug

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6429

### To test

- [ ] Impersonate a user and successfully navigate to the Timeline
- [ ] Impersonate a user and successfully navigate to Tasks


mattkrick:
nice find!

Issue #6421, opened 2022-04-19
Emoji menu popups in a wrong place if `:` is the first character typed


igorlesnenko:
## Issue - Bug

- Create a new retro meeting
- Try to add a new reflection but start typing it from the `:` character
- See the emoji menu has appeared at the top left corner of the browser window

https://user-images.githubusercontent.com/466991/164055008-32b3c7cd-157a-4f6c-bd88-be1e73d63238.mov

### Acceptance Criteria (optional)

- Emoji menu always appears in the correct location

**Estimated effort:** 2h

PR #6906, opened 2022-07-20
fix:  Emoji menu popups in a wrong place if `:` is the first character typed 

gitstart:
# Description
Emoji menu popups in a wrong place if : is the first character typed

- `EmogiMenuContainer`  is rendered out of position  because the `draft-js`'s `getVisibleSelectionRect` responsible for calculating and returns the correct page coordinates instead returns undefined when we input `:` at the start of input. 

Partially Fixes  [issue](https://github.com/ParabolInc/parabol/issues/6421)
- Conditionally render `EmojiMenuContainer` if we have the page `coordinates` provided

## Demo

## Testing scenarios

- [ ] Scenario A
  - Step 1
  - Step 2...

- [ ] Scenario B
  - Step 1
  - Step 2....

## Final checklist

### DIFF
--- a/packages/client/components/TaskEditor/withEmojis.tsx
+++ b/packages/client/components/TaskEditor/withEmojis.tsx
@@ -76,16 +76,20 @@ const withEmojis = (ComposedComponent) => {
     renderModal = () => {
       const {query} = this.state
       const {editorState} = this.props
-      this.cachedCoords = getDraftCoords() || this.cachedCoords
+      this.cachedCoords = getDraftCoords()
       return (
-        <EmojiMenuContainer
-          removeModal={this.removeModal}
-          menuItemClickFactory={this.menuItemClickFactory}
-          query={query}
-          menuRef={this.menuRef}
-          editorState={editorState}
-          originCoords={this.cachedCoords}
-        />
+        <>
+          {this.cachedCoords && (
+            <EmojiMenuContainer
+              removeModal={this.removeModal}
+              menuItemClickFactory={this.menuItemClickFactory}
+              query={query}
+              menuRef={this.menuRef}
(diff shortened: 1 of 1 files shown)


packages/client/components/TaskEditor/withEmojis.tsx
```diff
@@ -76,16 +76,20 @@ const withEmojis = (ComposedComponent) => {
...
-        <EmojiMenuContainer
-          removeModal={this.removeModal}
-          menuItemClickFactory={this.menuItemClickFactory}
-          query={query}
-          menuRef={this.menuRef}
-          editorState={editorState}
-          originCoords={this.cachedCoords}
-        />
+        <>
+          {this.cachedCoords && (
```
Dschoordsch, line 82: -2 this just disables the menu for this situation, which is worse than showing it in the wrong place

Dschoordsch (changes requested):

Dschoordsch:
Thanks for taking on this issue, but the current version does just disable the menu. Instead we should calculate the fallback position somehow.

packages/client/components/TaskEditor/withEmojis.tsx
raj360: Well the menu is not entirely disabled, we are only preventing it from rendering with `undefined` coordinates,  its actually not so long when we receive actual `coordinate`  values from `getVisibleSelectionRect` instead of `undefined`.

raj360:
> Thanks for taking on this issue, but the current version does just disable the menu. Instead we should calculate the fallback position somehow.

let look into it and figure out a possible solution

gitstart:
Closing this PR, we could not find a better solution to fix this bug.

Issue #6741, opened 2022-06-14
Design to prompt user to drag cards in group phase


acressall:
[User feedback:](https://parabol.slack.com/archives/C029ZTH0EL9/p1655132122585119)

>  It took two meetings to figure out that he had to drag to group - it wasn't obvious at the start.

Design an animation or prompt to indicate to the user that they can drag and drop cards to group them

### Acceptance Criteria
Users can:
   - See a prompt to start grouping
   - Not see the prompt _after_ they have begun grouping
   - Make the prompt go away
   - do not show this for users who've run >3 retros

**Estimated effort:** 2 points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

acressall:
I have two ideas for helping users see that they can drag and drop in the grouping phase of a retro. 

The animation should only start when the user:
1. is present
2. has not done anything for >=1 minute (too short?)
3. has not participated and/or facilitated more than 2 retrospectives

The animation should only show up once per phase and stop as soon as the user takes _any_ action

acressall:
Option 1: hinting at drag and droppable cards. This animation will repeat once and will occur on the first card that is not in a group


https://user-images.githubusercontent.com/18582225/174649708-21f216d9-d0bd-494a-b376-6669455b8473.mov

acressall:
Option 2: highlighting the tips. This animation prompts the user to view the tips for more information about what to do. It could potentially be used in more than just the group phase


https://user-images.githubusercontent.com/18582225/174649814-4d5df6f5-0d64-45f2-bac6-7cf77e322468.mov

Dschoordsch:
I'd like to use the option 2 style hint for highlighting the next or end meeting depending on context.

acressall:
New issues added to reflect these #6815, #6814

87prashant:
> Option 2: highlighting the tips. This animation prompts the user to view the tips for more information about what to do. It could potentially be used in more than just the group phase

A small observation: 
What I notice is that it is hard to notice the animation, white element against the white background (maybe that's the case only with me).
- We can change/add the color at the time of the animation (though we are changing color already, but it is not significant)

What do others think?

JimmyLv:
I agree that it is hard to notice the animation, but since it is flashing repeatedly, maybe the user will notice after staying for a while? 
> We can change/add the color at the time of the animation
I like the idea of adding color to it though

[...]

PR #6910, opened 2022-07-21
feat(Retro): Prompt user to drag cards in group phase

### LINKED ISSUE #6814: Prompt user to drag cards in group phase
Design: #6741 

Option 1: hinting at drag and droppable cards. 

https://user-images.githubusercontent.com/18582225/175152287-26501b7b-ac93-48de-985a-6baa6c8bfd19.mov

Animation details:
- Delay 30s
- Duration: 900ms (600ms to "lift"; 300ms to "drop")
- Iteration count: 2
- Transition timing: ease-out

AC:
- show the hint after 30s and repeat every 60s until there is at least one reflection group.

87prashant:
# Description

Fixes #6814

## Demo
To show the working the duration of animation changed to 2s  with a 2s delay

https://user-images.githubusercontent.com/106697681/180114304-1e982836-41ae-4c3e-b2f3-ca52cc4ae474.mp4

## Testing scenarios

## Final checklist


packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -254,11 +255,25 @@ const ReflectionCard = (props: Props) => {
     !isComplete &&
     !isDemoRoute() &&
     (isHovering || !isDesktop)
+
+  const reflectionCount = reflectionGroups.reduce(
```
87prashant, line 259: #### Help Wanted
I am currently learning relay and graphql. When I use reflectionCount after adding the data dependency in the fragment container, its value is always 0, though reflections are present in the meeting.
So, I have manually calculated it.  
Any help what I am missing.

87prashant:
@acressall 
Could you confirm is this the intended behavior in the video?

acressall:
@87prashant the animation seems really fast. Is it the video sped up or is that the actual speed? The functionality side of things is good!

87prashant:
Yeah, animation is changed in the video to show functionality.

packages/client/components/ReflectionCard/ReflectionCard.tsx
Dschoordsch: You're not missing anything. `reflectionCount` is not calculated on the server side, but is just statically set when ending the meeting:
* it's read statically here: https://github.com/ParabolInc/parabol/blob/884ac2ef450a5f3f79c5da521c2caca9a3cc7c67/packages/server/graphql/types/RetrospectiveMeeting.ts#L73
* the source for `RetrospectiveMeeting`, i.e. the argument to the resolve function is the object loaded from rethink https://github.com/ParabolInc/parabol/blob/884ac2ef450a5f3f79c5da521c2caca9a3cc7c67/packages/server/database/types/MeetingRetrospective.ts#L33
* it's written in this mutation https://github.com/ParabolInc/parabol/blob/884ac2ef450a5f3f79c5da521c2caca9a3cc7c67/packages/server/graphql/mutations/endRetrospective.ts#L51

It would be possible to move this calculation to the server side, but your current approach is fine

87prashant: Ok, thanks @Dschoordsch!

Dschoordsch:
@JimmyLv could you please review?

packages/client/components/ReflectionCard/ReflectionCardRoot.tsx
```diff
@@ -12,7 +31,8 @@ const ReflectionCardRoot = styled('div')({
   maxWidth: '100%',
   position: 'relative',
   transition: `box-shadow 2000ms ${DECELERATE}`,
-  width: ElementWidth.REFLECTION_CARD
-})
+  width: ElementWidth.REFLECTION_CARD,
+  animation: shouldAnimate && shouldAnimate() ? `${dragPrompt}  60s 30s infinite` : undefined
```
JimmyLv, line 35: -1 same as above

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -254,11 +255,25 @@ const ReflectionCard = (props: Props) => {
...
+    (sum, {reflections}) => sum + reflections.length,
+    0
+  )
+
+  const shouldAnimate = () => {
+    if (phaseType !== 'group') return false
+    if (reflectionCount !== reflectionGroups.length) return false
+    if (reflectionId !== reflectionGroups[0]?.reflections[0]?.id) return false
```
JimmyLv, line 267: -1 If you want to be sure that the value of `reflectionGroups[0]` exists, it might also need another layer of [Optional chaining (?.)](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)?
```js
reflectionGroups?.[0]?.
```

But it's still a good idea to check the TypeScript declaration of `reflectionGroups` and `reflections` directly here.

JimmyLv (changes requested):

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -254,11 +255,25 @@ const ReflectionCard = (props: Props) => {
...
+    return true
+  }
+
   return (
     <ReflectionCardRoot
       data-cy={`${dataCy}-root`}
       onMouseEnter={() => setIsHovering(true)}
       onMouseLeave={() => setIsHovering(false)}
+      shouldAnimate={() => shouldAnimate()}
```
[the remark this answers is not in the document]
87prashant: Ok

87prashant:
I am not able to find the way to stop the animation when dragging the first reflection (i.e. reflection card where animation applies)

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -254,11 +255,30 @@ const ReflectionCard = (props: Props) => {
     !isComplete &&
     !isDemoRoute() &&
     (isHovering || !isDesktop)
+
+  const firstReflection = reflectionGroups?.[0]?.reflections?.[0]
+  //reflectionCount === reflectionGroups.length showing the weird behavior on reload, that's why calculated like this
+  const primaryCondition = reflectionGroups.find((group) => group.reflections.length > 1)
```
JimmyLv, line 261: -1 `primaryCondition` needs a more descriptive name.

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -254,11 +255,30 @@ const ReflectionCard = (props: Props) => {
...
+  const draggingHappening = (() =>
```
JimmyLv, line 262: -1 Given that you already touched the `firstReflection`, here only needs to determine the state of the first card at this point, not need to iterate through the entire array?

JimmyLv (changes requested):
@87prashant Thank you, I think we're almost done with it!
Given that the original condition is "until there is at least one reflection group.", it's the basic thing. So I think the `shouldAnimate` internally should have reversed logic and returns FALSE by default, and only returns TRUE when:

- there is `group` stage
- and there is no reflection group,
- and the state of the first card is not dragging (not need to iterate through the entire array.)

Does this simplify the logic?

packages/client/components/ReflectionCard/ReflectionCard.tsx
JimmyLv: -1 Maybe `Array.prototype.some()` might be more meaningful? with the scenario here to return `boolean` value of `is Dragging`? 

> The some() method tests whether at least one element in the array passes the test implemented by the provided function. - [Array.prototype.some() - JavaScript | MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)

JimmyLv:
Hey @87prashant  I tried to refactor the logic, but haven't verified its functionality yet, would you please see if that's correct?

```js

  
  const isGroupPhase = !isComplete && phaseType === 'group'
  const hasNoGroup = !reflectionGroups.some((group) => group.reflections.length > 1)

  const shouldAnimate = (() => {
    if (isGroupPhase && hasNoGroup) {
      const firstReflection = reflectionGroups?.[0]?.reflections?.[0]
      const isFirst = reflectionId === firstReflection?.id
      const isNotEditing = !firstReflection?.isEditing
      const isNotDragging = !firstReflection.isViewerDragging && !firstReflection.remoteDrag

      return isFirst && isNotEditing && isNotDragging;
    }
    return false
  })()
```

packages/client/components/ReflectionCard/ReflectionCard.tsx
87prashant: > here only needs to determine the state of the first card at this point, not need to iterate through the entire array

packages/client/components/ReflectionCard/ReflectionCard.tsx
87prashant: ok

packages/client/components/ReflectionCard/ReflectionCard.tsx
JimmyLv: > show the hint after 30s and repeat every 60s until there is at least one reflection group.
According to the original requirement description, the `remoteDrag` should be ignored, because the drag is an intermediate state that does not end up forming one `reflectionGroup`.

JimmyLv: So I think the condition of judgment here should be:
```js
const isNotDragging = !reflectionGroups.some(
  (group) =>  group.reflections.some(
    (reflection) => reflection.isViewerDragging)
  )
)
```

87prashant: Ok

87prashant:
> would you please see if that's correct and make sense?

Yeah it is working fine 👍

JimmyLv (changes requested):

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -254,11 +255,31 @@ const ReflectionCard = (props: Props) => {
...
+  const shouldAnimate = (() => {
+    if(isGroupPhase && hasNoGroup) {
+      const firstReflection = reflectionGroups?.[0]?.reflections?.[0]
+      const isFirst = reflectionId === firstReflection?.id
+      const isNotEditing = !firstReflection?.isEditing
+      const isNotDragging = !reflectionGroups.some(
+        (group) => group.reflections.some(
+          (reflection) => reflection.isViewerDragging
+        )
+      )
```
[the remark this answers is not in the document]
87prashant: ok 😁

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -96,14 +152,15 @@ const ReflectionCard = (props: Props) => {
...
+  const {localPhase, localStage, spotlightGroup, phases, spotlightSearchQuery, reflectionGroups} =
+    meeting
   const {phaseType} = localPhase
   const {isComplete} = localStage
   const spotlightGroupId = spotlightGroup?.id
   const isSpotlightSource = reflectionGroupId === spotlightGroupId
   const isSpotlightOpen = !!spotlightGroupId
   const atmosphere = useAtmosphere()
-  const reflectionRef = useRef<HTMLDivElement>(null)
+  const reflectionNewRef = useRef<HTMLDivElement>(null)
```
87prashant, line 163: I am not sure if `reflectionNewRef` refers to the same thing as `reflectionRef`?

JimmyLv: I see that you did a `useFragment` refactoring, are you worried about this will affect the `ref`? but here `reflectionNewRef` should not need to be renamed.

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -60,8 +60,8 @@ const SpotlightButton = styled(CardButton)<{showSpotlight: boolean}>(({showSpotl
 
 interface Props {
   isClipped?: boolean
-  reflection: ReflectionCard_reflection
-  meeting: ReflectionCard_meeting
+  reflectionRef: ReflectionCard_reflection$key
```
87prashant, line 63: I have used `reflectionRef` variable here. So I have to rename the below one

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -86,7 +86,63 @@ const getReadOnly = (
 }
 
 const ReflectionCard = (props: Props) => {
-  const {meeting, reflection, isClipped, openSpotlight, stackCount, showReactji, dataCy} = props
+  const {meetingRef, reflectionRef, isClipped, openSpotlight, stackCount, showReactji, dataCy} =
```
87prashant, line 89: Or I can use an alias for `relectionRef` during the destructuring of `Props`
Can you suggest any appropriate name for the alias

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -86,7 +86,63 @@ const getReadOnly = (
...
+    props
+  const reflection = useFragment(
```
87prashant, line 91: `reflection` is also taken 😁

JimmyLv: aha, if so, it would be better to just rename the return value of `useRef` to `reflectionReference`. 😁

JimmyLv (approved):
Thank you for all your work! LGTM!

mattkrick:
@ackernaut what are your thoughts?

I have a reservation on this design. As we run retros internally, we read all the items first & then we begin dragging. if i'm focused on reading all the reflections & see something jump on my screen, my eye gets pulled to the animation. That's not a great UX, it disrupts the flow of the meeting & continues to do so until i create a group.

I acknowledge that we have a user request where the team didn't realize they were supposed to drag. Frankly, I'm OK with that. I'd rather be snapchat than facebook. If we introduce this feature, we're making it better for first timers & adding a distraction for the rest, however since this is a design consideration, I'll default to @ackernaut

87prashant:
> If we introduce this feature, we're making it better for first timers & adding a distraction for the rest

I am not sure if there is some way we can check/track how many time user run the retro and could do something with that in this situation.

acressall:
@mattkrick this is a good callout! 

WDYT if we could extend the delay on the animation, so it only starts when there has been no action after a few _minutes_?

I've heard from a few users that the facilitator (often scrum master in this case) does most of the interaction with the product, and the rest of the team doesn't do much outside of writing reflections and voting. My hypothesis: if a _participant_ engages more in the meeting, they'll engage more in the product as a whole, and this prompt _might_ help them do that. 

Here's what I propose:

1. We increase the time before this animation starts. Maybe test 2 minutes?
2. We add an event for grouping (not sure if this would be when the user drags or if we can check if they actually made a group)
3. If the amount of grouping increases over the time this is live by 2% or more, we keep it. If it is under that, we remove it (we'll test for a month. Users who run a 2-week cadence will only see it twice, users who are on a  1-mo cadence will see it once). If we want to keep it we can look into limiting it to a team's first retro, if that's possible
4. We take a look at who has been grouping -- who does it the most, what does their engagement look like in and outside of meetings. With the new event we can look at data in Amplitude and also watch some session replays. This way, even if we decide to remove this after the month it is live, it provides learnings about our users. No wasted work. 

WDYT? Does this ease any of your concerns? Would love @ackernaut's thoughts too.

ackernaut:
I took a card, and will drop my thoughts soon

ackernaut:
@acressall @mattkrick thanks for the ping:

I pulled this down and took a look:

- The first card in the first column animates after 30 seconds, then every 60 seconds
- It doesn’t animate if a group has been made, if somebody is editing that first card, or if somebody is dragging any card
- I think the animation itself is a little slow, feels different than when folks pick up cards, which happens fast
  - However, being slow might help the user understand. Emphasis on might…
  - If the user isn’t looking at that card, I suspect they’re going to be confused at what happened; by the time they look the animation is probably on the second iteration
  - I do think this will be weird in async use cases when nobody has reason to group yet. If I show up to an open group phase for any length of time I can see how it would be annoying
  - I’m not sure the animation is clear that you can drag the card; not sure what I would do differently
  - If this stays in the code, there needs to be a comment as to why `shouldAnimate` is in there, it’s not obvious to the functionality of the reflection card but rather a conditional use case. I think another developer probably won’t know why it’s there. There could even be a link to this PR ‘this is why this is here’

Feedback:

- Solving for some, impacting all: I agree that we have to be very careful providing solutions that are highly specific to certain use cases but end up affecting all users
  - Depending on the timing of the animation here, this might not impact many folks. Chances are somebody has grouped something; we could also add a condition to `shouldAnimate` when `meetingNumber` is less than 3.
- I tend to like these principals for first-time use:
  - Minecraft/Snapchat — it’s okay for folks to fumble a little at first, I’d assert folks are gonna fumble with anything we stick in front of them; they eventually learn a simple mechanic and there’s no problem
  - Super Mario — I like how in Super Mario the design of the early levels helps folks to learn the mechanics. The problem is our software doesn’t have levels. One repeats the same game over and over. Therefore we try to detect some first-time use conditions and teach inline. I think this concept is generally in line with that thinking. I don’t want a bunch of dialogs popping up everywhere.
 

  
Experiment?
- I think the conditions make this safe to try.
  - **I recommend we add a condition** to `shouldAnimate` if `meetingNumber` is less than 3
  - The timing (repeat 60, delay first 30) is probably good enough for the first-time use case
  - The fact that if anybody groups, edits the card, or drags makes the animation go away probably reduces the noise for most usage
- I think it may be too much scope creep to add a grouping event for this PR. Not even sure we need to frame this as an experiment
  - I think since this isn’t solving for general participation in grouping, but rather first-time learning, we shouldn’t frame an experiment that way; you might instead compare 1 month to the previous and see if there’s 1 user and the meeting count is less than 3, did those folks group more? (assuming the condition for somebody trying it by themselves, and it being helpful to know that you can drag to group)
- This animation is easy to remove if there’s some external or internal tension

Late to the party

- I only saw 1 data point of user feedback? Has there been more? I briefly looked in issues and discussions but only found that single comment in #6741 — my thought here is that we’ve applied effort with little data, when there are known frictions with lots of data. I’m not sure there’s much of a problem with folks learning how to group
- Root causes: when we have to spell out a lot of specific conditions (should be X time, no grouping yet, maybe only in the first 2-3 meetings) I wonder if we could solve this in more general way. Why isn’t it obvious that you can group?

ackernaut (changes requested):
Can you add 1 more condition to `shouldAnimate`? I think we could scope this to when the meeting number is less than 3.

I believe you can get `meetingNumber` from the `RetrospectiveMeeting` object?

acressall:
Thanks for the detailed feedback @ackernaut! To make sure I understand, are you saying it's safe to try with the `shouldAnimate` parameter on `meetingNumber`?

Another clarifying question:

> I think since this isn’t solving for general participation in grouping, but rather first-time learning, we shouldn’t frame an experiment that way; you might instead compare 1 month to the previous and see if there’s 1 user and the meeting count is less than 3, did those folks group more? (assuming the condition for somebody trying it by themselves, and it being helpful to know that you can drag to group)

Is there a way to do this without a grouping event? Maybe @tianrunhe knows how we can look at the amount of grouping?

tianrunhe:
> Is there a way to do this without a grouping event? Maybe @tianrunhe knows how we can look at the amount of grouping?

It seems we don't have events published on grouping behavior 😂. Feel free to create an issue for me, should be pretty easy to add

ackernaut (approved):

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -314,9 +322,10 @@ const ReflectionCard = (props: Props) => {
 
   const isGroupPhase = !isComplete && phaseType === 'group'
   const hasNoGroup = !reflectionGroups.some((group) => group.reflections.length > 1)
+  const isMeetingNumberLessThanThree = meetingNumber < 3
```
ackernaut, line 325: @87prashant thanks for adding this!

JimmyLv (approved):

87prashant:
https://github.com/ParabolInc/parabol/pull/6910#issuecomment-1192243055

@JimmyLv 
Could you check this as it might cause bug later?

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -254,12 +319,31 @@ const ReflectionCard = (props: Props) => {
...
+
+  const isGroupPhase = !isComplete && phaseType === 'group'
+  const hasNoGroup = !reflectionGroups.some((group) => group.reflections.length > 1)
+  const isMeetingNumberLessThanThree = meetingNumber < 3
+
+  const shouldAnimate = (() => {
+    if (isGroupPhase && hasNoGroup && isMeetingNumberLessThanThree) {
+      const firstReflection = reflectionGroups?.[0]?.reflections?.[0]
+      const isFirst = reflectionId === firstReflection?.id
+      const isNotEditing = !firstReflection?.isEditing
```
Dschoordsch, line 331: -1 If I'm editing any card, then I don't want to be distracted. This should go into the next `.every` condition and then be renamed to `isNotInteracting` or sth. like that.

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -86,7 +86,64 @@ const getReadOnly = (
...
+          stages {
+            id
+            isComplete
+          }
+        }
+        spotlightGroup {
+          id
+        }
+        spotlightSearchQuery
+        reflectionGroups {
```
Dschoordsch, line 134: FYI: This will cause all reflection cards to re-render once something changes in an unrelated part. I thought this could be a problem, but they already re-render on master if there is a change.

Dschoordsch (changes requested):

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -254,12 +319,31 @@ const ReflectionCard = (props: Props) => {
...
     !isDemoRoute() &&
     (isHovering || !isDesktop)
+
+  const isGroupPhase = !isComplete && phaseType === 'group'
+  const hasNoGroup = !reflectionGroups.some((group) => group.reflections.length > 1)
+  const isMeetingNumberLessThanThree = meetingNumber < 3
+
+  const shouldAnimate = (() => {
+    if (isGroupPhase && hasNoGroup && isMeetingNumberLessThanThree) {
+      const firstReflection = reflectionGroups?.[0]?.reflections?.[0]
```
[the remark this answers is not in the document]
87prashant: Yes, I am not able to figure out to resolve that.
I already mentioned it here: https://github.com/ParabolInc/parabol/pull/6910#issuecomment-1192243055

Apart from that, here is one more issue: https://github.com/ParabolInc/parabol/pull/6910#issuecomment-1193069249

jordanh:
As PM, I've also been asked to review this PR. I'll have some feedback ~Monday or ~Tuesday for you all

packages/client/components/ReflectionCard/ReflectionCardRoot.tsx
```diff
@@ -1,9 +1,28 @@
+import {keyframes} from '@emotion/core'
 import styled from '@emotion/styled'
 import {DECELERATE} from '~/styles/animation'
 import {Elevation} from '../../styles/elevation'
 import {Card, ElementWidth} from '../../types/constEnums'
 
-const ReflectionCardRoot = styled('div')({
+const dragPrompt = keyframes`
```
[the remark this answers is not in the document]
87prashant: Ok, I think `dragHintAnimation` would be better as it matches with `showDragHintAnimation`

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -254,12 +319,31 @@ const ReflectionCard = (props: Props) => {
     !isComplete &&
     !isDemoRoute() &&
     (isHovering || !isDesktop)
+
+  const isGroupPhase = !isComplete && phaseType === 'group'
+  const hasNoGroup = !reflectionGroups.some((group) => group.reflections.length > 1)
+  const isMeetingNumberLessThanThree = meetingNumber < 3
```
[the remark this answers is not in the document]
87prashant: 👍

packages/client/components/ReflectionCard/ReflectionCard.tsx
87prashant: Can we do something about it as I need `reflectionGroups` for calculations inside logic?

packages/client/components/ReflectionCard/ReflectionCard.tsx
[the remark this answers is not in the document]
Dschoordsch: The solution would be to also grab the `reflectPrompts`, take the first `id` and search the first reflection to match the `promptId`. You can take a look [here](https://github.com/ParabolInc/parabol/blob/85d92aa5eba8b4f069b2dbf32836e09805ca60f1/packages/client/components/GroupingKanban.tsx#L80) how the reflections are sorted in their columns.

Dschoordsch:
> I am not able to find the way to stop the animation when dragging the first reflection (i.e. reflection card which is animating)

That's an interesting problem. When we drag, we're not actually dragging the React element, but instead we clone the HTML node which means it does not receive any updates, thus the pre-drag state stays. This is done in
https://github.com/ParabolInc/parabol/blob/85d92aa5eba8b4f069b2dbf32836e09805ca60f1/packages/client/hooks/useDraggableReflectionCard.tsx#L334-L336
where we clone in l334 and only start the mutation in l336, thus `isViewerDragging` is still false when we create the clone. Moving the `cloneReflection` call after `StartDraggingReflectionMutation` would solve this.
Right now I can't see a reason why we couldn't just move the clone after the mutation, have to check.

packages/client/components/ReflectionCard/ReflectionCard.tsx
Dschoordsch: Actually this causes 2 renders each time something changes :-/

packages/client/components/ReflectionCard/ReflectionCard.tsx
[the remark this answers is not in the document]
Dschoordsch: I think the best solution would be to use existing information instead and pass down `isFirstColumn` from `GroupingKanban` and then `isFirstGroup` from `GroupingKanbanColumn`. This would also avoid the additional render caused by the additional GraphQL fields.

jordanh:
I had a look into this PR and the history I can find. I'd like to enable our org to try this, but I don't think this is yet safe to try for us. Let me explain why, and what would make it safe.

If there is an objection to proceeding with an idea, safety is about reshaping the idea _and_ the measures around it in order to show the organization the path forward. In the case of this PR, there is risk in not being able to show how this change benefits the user or how it might cause harm (should people find it annoying/more confusing, or whatever)

There are two bits that feel "unsafe" to me: introduction risk and measurement risk

Introduction risk is the inadvertent affect of creating a problem where one didn't exist before. I'd like to know what data we had that led us to wanting to introducing this change.

Measurement risk is not knowing if a change has a positive or negative effect after it is introduced.

Both of these risks may just may be an artifact of not having enough visibility into the end-to-end process that created this change. If you'd please humor a busy product manager, may I ask some questions that will help me determine safety?

1) Could we quantify what the funnel conversion stages are that this hopes to address (is Group->Vote and end-to-end meeting completions for users who have completed fewer than 3 meetings?)
2) What are the measure today (say, over the last 90 days?)
3) If this change is successful, what increases to the metrics from #1 do we hope to achieve? How long would it take for us to get to 95% confidence (roughly 400 users being exposed to the change?)

I may have more questions/feedback after these, but your answers, squad 2, will help us chart a course forward!

87prashant:
> there is no possible conflict, but please add a comment when you re-order the cloning.

Ok

packages/client/components/ReflectionCard/ReflectionCard.tsx
87prashant: I removed the `reflectionGroups` from `ReflectionCard` by passing the required data down from `GroupingKanban` but I am not able to see the difference in terms of render.
Could you please check, if it is what you mean?

acressall:
Alright, thanks for the feedback everyone! 
I'd like to propose we abandon this issue as it is taking far longer than anticipated to decide how to move forward. 
The initial principles of the concept are no longer true:

1. Pushing a quick, low-impact change would give us the information to see if this is a worthwhile lever to pursue further (this is no longer "quick" and the low-impact is not worth the larger effort)
2. We no longer have the time to complete implementation _and_ analyze results by the end of the trimester. This was meant to be low-hanging fruit, but we now have other issues that can take precedence. 

We can move #6814 back to the backlog in case we want to revisit later

Dschoordsch:
Before considering any funnels, let's address the primary action we're trying to encourage: grouping in retros. I spent some time to run the numbers and there is a difference in grouping behaviour [^1]
## meetingNumber <= 3
Screenshot from 2022-08-19 15-35-22

## meetingNumber > 3
Screenshot from 2022-08-19 15-35-04

This data is for the past 30 days. If we can bring the no-groups section from 17% down to 14% it would be success. More interesting question would be why there are no reflections though.

[^1]: https://colab.research.google.com/drive/1r_BUI0Jc7QjltopyOAXe9ty421L3p0vd#scrollTo=cdH_GnjhTINT

ackernaut:
@Dschoordsch this is interesting, so 37% of meetings <=3 have no reflections? Gonna be hard to group those 😆 

I assume folks abandon these early meetings (either leave it running or end it) since they can’t move to the Group phase without a reflection. So I wonder why they abandon? Best case, they ‘get it’ (‘yep, there are the questions’) and come back for another, real meeting with their team. Worse case, they don’t understand, have low curiosity to learn anything more about the app, and churn. I wonder if this is mostly just learning behavior. If they ‘get it’ and don’t feel the need to actually write something then maybe it’s less of a problem than we think. How badly do we need to test these assumptions vs working on known areas of friction with lots of user signal?

I still have a hard time seeing a problem with learning to group if folks are actually using it with their team. I can see the need perhaps for some initial hints when trying it alone in an early meeting number.

Dschoordsch:
> I still have a hard time seeing a problem with learning to group if folks are actually using it with their team. I can see the need perhaps for some initial hints when trying it alone in an early meeting number.

42% of those early meetings are run alone, so it could be helpful for them to figure it out.

jordanh:
Closing this PR

Dschoordsch:
@jordanh Given the data I collected I would still like to keep this and just further limit it to meetings without additional participants.

Dschoordsch (changes requested):
Let's move this forward and add some more conditions when to show this hint:
* meeting number < 3
* no other participants
* number of reflections > 1

jordanh:
@Dschoordsch

ok, thank you for assembling that data and your advocacy. Let's learn on this one.

When merged, could you please take a Parabol task set to ~45 days out that to come back and measure the effect of this change? I think the work you've put in to being focused on what metric you are trying to affect and sharing results will benefit the company!

packages/client/components/GroupingKanbanColumn.tsx
```diff
@@ -183,6 +187,8 @@ const GroupingKanbanColumn = (props: Props) => {
                       phaseRef={phaseRef}
                       reflectionGroupRef={reflectionGroup}
                       swipeColumn={swipeColumn}
+                      firstReflectionId={firstReflectionId}
```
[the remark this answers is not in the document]
87prashant: Done

packages/client/components/GroupingKanban.tsx
```diff
@@ -130,6 +140,8 @@ const GroupingKanban = (props: Props) => {
               reflectionGroups={groupsByPrompt[prompt.id] || []}
               reflectPromptsCount={reflectPromptsCount}
               swipeColumn={swipeColumn}
+              firstReflectionId={firstReflectionId}
```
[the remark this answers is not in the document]
87prashant: 👍

Dschoordsch (changes requested):
Code looks good, there seems to be an issue with just animating the first item
Peek 2022-10-04 20-32

packages/client/components/GroupingKanban.tsx
```diff
@@ -108,6 +119,26 @@ const GroupingKanban = (props: Props) => {
...
   if (!phaseRef.current) return null
+
+  const reflectionCount = reflectionGroups.reduce(
+    (sum, {reflections}) => sum + reflections.length,
+    0
+  )
+  const isGroupPhase = !isComplete && phaseType === 'group'
+  const isRetrospectiveBeginner = meetingNumber < 3 // If the meeting number is low, the user is probably new to retrospectives
+  const hasNoGroup = !reflectionGroups.some((group) => group.reflections.length > 1)
+  let isNotInteracting = false
```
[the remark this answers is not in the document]
87prashant: 👍

Dschoordsch (approved):
Works now! Nice work!

Issue #6310, opened 2022-03-30
Text in reflection cards is not selectable during vote period


tianrunhe:
## Issue - Bug

Scenario:
At voting phase for a user feedback retro meeting, I'd like to select the text from the reflection card. For example, there's a URL in the card so I'd like to copy & paste & check it out.

Expected:
User can select the text from a reflection card.

Actual:
Nothing happens when you click through the text.

### Acceptance Criteria (optional)
Users can:
   - [ ] select text from a reflection card

Triage is performed to find the root cause of the bug, timeboxed to ~1 hour. 

**Estimated effort:** 1 hour to triage. More if root cause is already identified.

PR #6924, opened 2022-07-23
fix: Text in reflection cards is not selectable during vote period

87prashant:
# Description

fixes: #6310

## Final checklist

### DIFF
--- a/packages/client/components/ReflectionCard/ReflectionCard.tsx
+++ b/packages/client/components/ReflectionCard/ReflectionCard.tsx
@@ -4,37 +4,37 @@ import {convertToRaw} from 'draft-js'
 import React, {MouseEvent, useEffect, useRef, useState} from 'react'
 import {commitLocalUpdate, createFragmentContainer} from 'react-relay'
 import AddReactjiToReactableMutation from '~/mutations/AddReactjiToReactableMutation'
+import isDemoRoute from '~/utils/isDemoRoute'
 import {
   NewMeetingPhaseTypeEnum,
   ReflectionCard_meeting
 } from '~/__generated__/ReflectionCard_meeting.graphql'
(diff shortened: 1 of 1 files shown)


87prashant:
Btw, I don't know why the import files showing changed!

Dschoordsch:
> Btw, I don't know why the import files showing changed!

We have a pre-commit hook which also sorts imports. If you make changes to a file which was written before this hook, it will re-order the imports.

@JimmyLv can you please take a look?

packages/client/components/ReflectionCard/ReflectionCard.tsx
```diff
@@ -216,7 +216,7 @@ const ReflectionCard = (props: Props) => {
     phases,
     isSpotlightSource
   )
-  const userSelect = readOnly ? (phaseType === 'discuss' ? 'text' : 'none') : undefined
+  const userSelect = readOnly ? (phaseType === 'discuss' || 'vote' ? 'text' : 'none') : undefined
```
JimmyLv, line 219: -1 I didn't quite get the logic here, when the `phaseType` is `vote`, the result of `'vote' === 'discuss'` is FALSE and then the default value is taken as `'vote'`? This shouldn't be what we want.

JimmyLv (changes requested):
If you want to allow the `vote` phase to be able to select text, you should be making a condition like this?
```js
if (phaseType === 'discuss' || phaseType === 'vote') {
  ...
}
```

packages/client/components/ReflectionCard/ReflectionCard.tsx
87prashant: Thank you @JimmyLv for clarifying the difference between the two:
My code made reflection cards selectable in every phase.

JimmyLv (approved):
LGTM!

PR #6931, opened 2022-07-25
feat(saml): support preferredName, enforce SAML over oauth

### LINKED ISSUE #6668: SAML attribute to define which value is read for preferred name
When using SAML auth we currently take the email address of the user login and trim the @domain from the email, thus making the remaining value the preferred name. A request has been made by an enterprise user to allow the passing of a specific field from their IDP that would allow assignment of a name value to be used that is different from the email address. Below are a few screenshots to further clarify the issue.

image

The issue is some users have firstname.lastname@domain which makes a human readable name. Other users have an id@domain which doesn't result in a human readable name when trimmed and formatted.

AC:

- [ ] allow mapping of customer defined attribute to custom attribute `preferredName` when customer is utilizing SAML auth. If preferredName name is found in claim as attribute then use the value as priority value for name

_If my understanding of how we create the username is incorrect, please further update thread and AC_

### LINKED ISSUE #6930: Enforcing SSO discussion, how should it work?
When enforcing SSO on a domain, should we also keep users from logging in via other methods (local/oauth) and signing up for the Parabol app?

For example: in an enterprise environment SSO is often used as a way to control access to approved apps. So, that users must request access through company resources for proper onboarding e.g. user requests access, they get a screen that says SSO required, they go request access to the app from their company which maybe requires: training to know what's allowed to be stored in the app, a license from a procured pool of licenses, and proper grouping in teams, etc

From a popular app online, this is a screen grab when trying to signup with bobsmith@apple.com
image

As you can see, you can't even signup if SSO is enabled as the "organization uses Single Sign On (SSO) ... Please sign in using your SSO credentials."

This might also drive users who are currently operating for free to purchase a license if their company exco decides SSO is the only approved method for Parabol, thus a conversion.

Currently: enforcing SSO allows users to still login via other methods like Google Oauth, even when using OKTA as SSO provider. (note this company example previously used Google Oauth to sign-in with their company accounts, but not use OKTA and app provisioning for access)

mattkrick:
# Description

fix #6668
fix #6930

Users can add a `preferredName` attribute to their SAML response payload & that will become their new `preferredName` _for new users_. This will not update the name for existing users.

Also, if a user signed up using Google OAuth2 and then their domain upgraded to SSO, we now force them to login with SSO. If they try logging in via Google OAuth2, they get an error telling them to login via SSO.

### DIFF
--- a/packages/server/graphql/private/mutations/loginSAML.ts
+++ b/packages/server/graphql/private/mutations/loginSAML.ts
@@ -51,15 +51,15 @@ const loginSAML: MutationResolvers['loginSAML'] = async (_source, {samlName, que
   const {isInvited} = relayState
   const {extract} = loginResponse
   const {attributes, nameID: name} = extract
-  const caseInsensitiveAtttributes = {} as Record<Lowercase<string>, Lowercase<string>>
+  const caseInsensitiveAtttributes = {} as Record<Lowercase<string>, string | undefined>
   Object.keys(attributes).forEach((key) => {
     const lowercaseKey = key.toLowerCase()
     const value = attributes[key]
-    const lowercaseValue = String(value).toLowerCase()
-    caseInsensitiveAtttributes[lowercaseKey] = lowercaseValue
+    caseInsensitiveAtttributes[lowercaseKey] = String(value)
   })
-  const {email: inputEmail, emailaddress} = caseInsensitiveAtttributes
-  const email = inputEmail || emailaddress
+  const {email: inputEmail, emailaddress, preferredname} = caseInsensitiveAtttributes
+  const preferredName = preferredname || name
+  const email = inputEmail?.toLowerCase() || emailaddress?.toLowerCase()
   if (!email) {
     return {error: {message: 'Email attribute was not included in SAML response'}}
   }
@@ -83,7 +83,7 @@ const loginSAML: MutationResolvers['loginSAML'] = async (_source, {samlName, que
   const newUser = new User({
     id: userId,
     email,
-    preferredName: name,
+    preferredName,
     tier: 'enterprise'
   })
 
--- a/packages/server/graphql/private/mutations/loginSAML.ts
+++ b/packages/server/graphql/private/mutations/loginSAML.ts
@@ -57,8 +57,8 @@ const loginSAML: MutationResolvers['loginSAML'] = async (_source, {samlName, que
     const value = attributes[key]
     caseInsensitiveAtttributes[lowercaseKey] = String(value)
   })
-  const {email: inputEmail, emailaddress, preferredname} = caseInsensitiveAtttributes
-  const preferredName = preferredname || name
+  const {email: inputEmail, emailaddress, displayname} = caseInsensitiveAtttributes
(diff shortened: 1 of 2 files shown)


packages/server/graphql/public/mutations/loginWithGoogle.ts
```diff
@@ -28,8 +29,14 @@ const loginWithGoogle: MutationResolvers['loginWithGoogle'] = async (
...
   }
 
-  const existingUser = await getUserByEmail(email)
+  const [existingUser, samlURL] = await Promise.all([
+    getUserByEmail(email),
+    getSAMLURLFromEmail(email, false)
+  ])
 
+  if (samlURL) {
+    return {error: {message: 'Please login using SSO'}}
```
adaniels-parabol, line 38: Do you think we should be more descriptive and tell them, something like "Your organization administrator requires you to login with SSO. Please click the "Sign in with SSO" but in the form below."

adaniels-parabol: other options would be to change the screen to SSO and enter email or pop SSO window to the url in the SAML table entry for that account along with the message "Your organization administrator requires you to login with SSO. Please click the "Sign in with SSO" but in the form below."

Dschoordsch: I also would also opt for redirecting to SSO sign in with message "Your organization requires you to login with SSO"

packages/server/graphql/private/mutations/loginSAML.ts
```diff
@@ -51,15 +51,15 @@ const loginSAML: MutationResolvers['loginSAML'] = async (_source, {samlName, que
...
   Object.keys(attributes).forEach((key) => {
     const lowercaseKey = key.toLowerCase()
     const value = attributes[key]
-    const lowercaseValue = String(value).toLowerCase()
-    caseInsensitiveAtttributes[lowercaseKey] = lowercaseValue
+    caseInsensitiveAtttributes[lowercaseKey] = String(value)
   })
-  const {email: inputEmail, emailaddress} = caseInsensitiveAtttributes
-  const email = inputEmail || emailaddress
+  const {email: inputEmail, emailaddress, preferredname} = caseInsensitiveAtttributes
```
Dschoordsch, line 60: -1 I didn't find any definition on the fly, but it seems `displayname` is the preferred way to define a preferred name.

Dschoordsch (changes requested):

packages/server/graphql/public/mutations/loginWithGoogle.ts
adaniels-parabol: From Squad 3 meeting, a recommendation "Your organization requires you to login with SSO. Reach out to support@parabol.co for more help!"

If we start getting lots of emails, then that might be signal to better automate this process of mapping a user to the approval POC.

packages/server/graphql/private/mutations/loginSAML.ts
mattkrick: i like it! will update

PR #6934, opened 2022-07-25
fix(pause): decrease max monthly pauses to 1

mattkrick:

# Description

There have been 302 manual account pauses in the last month.
The hypothesis is that folks are gaming the system & pausing until they run a meeting & then unpausing.
We're trying this approach because it's cheap & easy. It'll probably get rid of the worst offenders.
After this, I think we should just focus on providing daily value. If we have a product that folks use more frequently, this problem heals itself.


mattkrick:
closing in favor of Nick's PR to completely remove them

PR #6942, opened 2022-07-26
feat(retro): prompt user to go to next phase or end meeting

### LINKED ISSUE #6815: Prompt user to go to next phase or end meeting
Design from #6741 

https://user-images.githubusercontent.com/18582225/175158956-4dea61a6-048c-4ff1-afa3-a9933e3c4af5.mov

After [length?] of inactivity, the user will see the "next" item in the toolbar do a wiggle animation to hint at the next step. 
This will only appear for the facilitator.
We can test this in the demo. 

**Reflect phase:**
when there are no reflections in progress (other users) and none in focus (facilitator)?

**Group phase:**
when nothing has moved after 30s?

**Vote phase:**
When the remaining votes is at zero for 30s or has not changed for 30s

**Discuss phase:**
this is a tricky one since a lot could be happening sync in a call. Maybe we hint after 5 minutes?

Caveat: if the ready button is "full" before these conditions are met, the animation should start after 5s

JimmyLv:
# Description

Resolved #6815 
After inactivity, the user will see the "next" item in the toolbar do a wiggle animation to hint at the next step.
This will only appear for the facilitator.

## Demo

~~https://www.loom.com/share/73d4c6d84f7d420881373f1f794b076e~~
https://www.loom.com/share/aa33f812fd83490ab690e63c9e2f3080

## Testing scenarios

1. Reflect phase:
when there are no reflections in progress for 30s

2. Group phase:
when nothing has moved after 30s

3. Vote phase:
When the remaining votes is at zero for 30s or has not changed for 30s

4. Discuss phase:
this is a tricky one since a lot could be happening sync in a call, we hint after 5 minutes

Caveat : if the ready button is "full" before these conditions are met, the animation should start after 5s

## Final checklist


packages/client/components/BottomControlBarTips.tsx
```diff
@@ -1,78 +1,83 @@
...
 import Menu from './Menu'
 
 const TallMenu = styled(Menu)({
   maxHeight: 320
 })
 
-const CheckInHelpMenu = lazyPreload(async () =>
-  import(/* webpackChunkName: 'CheckInHelpMenu' */ './MeetingHelp/CheckInHelpMenu')
+const CheckInHelpMenu = lazyPreload(
+  async () => import(/* webpackChunkName: 'CheckInHelpMenu' */ './MeetingHelp/CheckInHelpMenu')
```
JimmyLv, line 24: oh, the linter did it automatically

packages/client/components/BottomControlBarTips.tsx
```diff
@@ -99,12 +104,29 @@ const helps = {
...
 interface Props {
   cancelConfirm: (() => void) | undefined
-  meeting: BottomControlBarTips_meeting
+  meeting: BottomControlBarTips_meeting$key
   status: TransitionStatus
   onTransitionEnd: () => void
 }
 const BottomControlBarTips = (props: Props) => {
-  const {cancelConfirm, meeting, status, onTransitionEnd} = props
+  const {cancelConfirm, meeting: meetingRef, status, onTransitionEnd} = props
```
igorlesnenko, line 112: +2 One more createFragmentContainer refactored 👍

igorlesnenko (commented):
I have few questions

- The issue says that `After inactivity, the user will see the "next" item in the toolbar do a wiggle animation`. Why do we animate `tips` button instead?

- If I open a demo, click next on reflect phase without waiting for it to finish. Then wait on the group phase. Animation will never appear. Is it expected behaviour?

- I noticed some glitchy behaviour. Some times when I start meeting I instantly see button animation on the reflect phase. Sometimes it never happens:
https://www.loom.com/share/90e285da0b4c4782b958b1eb73d9b2b4

JimmyLv:
Hey @igorlesnenko after confirming with @acressall (cc Design Reviewer based on our new process 😁), I already updated to animate the `next` button, could you please help to have a look again? 

new Loom demo: https://www.loom.com/share/aa33f812fd83490ab690e63c9e2f3080

for another two cases, I've fixed them too:

> If I open a demo, click next on reflect phase without waiting for it to finish. Then wait on the group phase. Animation will never appear. Is it expected behaviour?

This is probably due to the previous judgement logic, which is now handled to allow skipping the wait directly, i.e. the grouping phase also triggers the animation.

> when I start meeting I instantly see button animation on the reflect phase.

This is because I didn't judge `hasNoReflection` at first, which means the animation was triggered immediately without any input from anyone

igorlesnenko:
@JimmyLv the button is now animating almost right after creating first reflection, should we wait longer or it is expected behaviour?
https://www.loom.com/share/ae706afeb701430cbbd94a3ca541d00f

JimmyLv:
@igorlesnenko Good find, it does animation quickly when there is only 1 user, I have changed the delay seconds to 30s, i.e.

`when there are no reflections in progress for 30s`

igorlesnenko (approved):
Looks good to me

packages/client/components/BottomControlBarReadyButton.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const atmosphere = useAtmosphere()
+  const {viewerId} = atmosphere
+
+  useEffect(() => {
+    // only enable for facilitatorUser
+    const isFacilitator = viewerId === facilitatorUserId && !endedAt
+    if (!isFacilitator) return
+
+    switch (localPhase.phaseType) {
+      case 'reflect':
```
Dschoordsch, line 117: +1 I think it would be more readable if you would split it up into multiple `useEffect` blocks:
```ts
useEffect(() => {
    // only enable for facilitatorUser
    const isFacilitator = viewerId === facilitatorUserId && !endedAt
    if (!isFacilitator) return
    if (localPhase.phaseType !== 'reflect') return

      const hasNoReflection = reflectionGroups?.every((group) => group.reflections?.length === 0)
      if (hasNoReflection) return

      const reflectPrompts = localPhase.reflectPrompts!
      const beforeDemoStart =
        isDemoRoute() && reflectPrompts.every(({editorIds}) => !Array.isArray(editorIds))
      if (beforeDemoStart) return

      const isNotEditing = reflectPrompts.every(({editorIds}) => {
        return editorIds === undefined || (Array.isArray(editorIds) && editorIds.length === 0)
      })

      setDelaySeconds(30)
      setShouldAnimate(isNotEditing)
    }, [localPhase.phaseType === 'reflect', reflectionGroups, editorIds])
```

packages/client/components/BottomControlBarReadyButton.tsx
```diff
@@ -0,0 +1,183 @@
...
+    votesRemaining
+  } = meeting
+  const [delaySeconds, setDelaySeconds] = useState(2)
+  const [shouldAnimate, setShouldAnimate] = useState(false)
+  const atmosphere = useAtmosphere()
+  const {viewerId} = atmosphere
+
+  useEffect(() => {
+    // only enable for facilitatorUser
+    const isFacilitator = viewerId === facilitatorUserId && !endedAt
```
Dschoordsch, line 113: -1 I think it wasn't clear in the ticket, but we should also add it to the "Ready" button while we're at it.

packages/client/components/BottomControlBarReadyButton.tsx
```diff
@@ -0,0 +1,183 @@
...
+        setShouldAnimate(isNotEditing)
+        break
+      case 'group':
+        const isNotDragging = reflectionGroups?.every((group) =>
+          group.reflections?.every(
+            (reflection) => !reflection.isViewerDragging && !reflection.isDropping
+          )
+        )
+
+        setDelaySeconds(30)
```
Dschoordsch, line 140: -1 This conflicts with #6910. If no groups were created and the user does nothing, they will see the next hint before the drag hint which renders the last one useless for new users.

Dschoordsch (changes requested):

mattkrick:
Clarifying questions:
- how might we measure the success of these animations?
- how might we avoid distracting the viewer with an animation while they are processing the reflections?

packages/client/components/BottomControlBarReadyButton.tsx
```diff
@@ -0,0 +1,183 @@
...
+    return {
+      display: 'flex',
+      alignItems: 'center',
+      justifyContent: 'center',
+      animation: shouldAnimate ? `1s ease-in-out ${delay}s 3 ${shake}` : undefined
+    }
+  }
+)
+
+interface BottomControlBarReadyButton {
```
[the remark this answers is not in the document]
JimmyLv: Thanks for finding it, I found the reason: it's WebStorm "rename" feature that automatically modifies this Prop Interface. need to pay attention to this issue in the future.

packages/client/components/BottomControlBarReadyButton.tsx
```diff
@@ -0,0 +1,183 @@
...
+        break
+      case 'discuss':
+        // this is a tricky one since a lot could be happening sync in a call. Maybe we hint after 5 minutes?
+        setDelaySeconds(5 * 60)
+        setShouldAnimate(true)
+        break
+    }
+
+    // if the ready button is "full" before these conditions are met, the animation should start after 5s
+    // TODO: replace with props from ButtonControlBarReady
```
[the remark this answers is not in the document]
JimmyLv: Removed, I replaced the calculation here with the `progress` props from the ReadyButton in the upper-level component

packages/client/components/BottomControlBarReadyButton.tsx
JimmyLv: Based on the latest comments https://github.com/ParabolInc/parabol/pull/6910#issuecomment-1217101885, if it is decided to stop showing dragging hints for now, then just allow this shake animation hints here.

packages/client/components/BottomControlBarReadyButton.tsx
JimmyLv: Aha, it is mentioned in https://github.com/ParabolInc/parabol/issues/6815: `This will only appear for the facilitator.` However, I have made the code logic here directly inside `return()`, i.e. if is not a Facilitator, just render the normal button directly.

packages/client/components/BottomControlBarReadyButton.tsx
JimmyLv: This is good practice, as #TIL. Just want to confirm, is the `=== condition` inside the deps of `useEffect` a generic pattern? Is there a reference for this? 😁

```ts
[localPhase.phaseType === 'reflect', reflectionGroups, editorIds])
```

Dschoordsch: I don't think this is common practice. You need to match what's used inside the block, so `localPhase.phaseType` would be sufficient, it would just run the block more often than necessary.

Dschoordsch: I took a look and found this
https://github.com/ParabolInc/parabol/blob/7bb4774f8ec3edc9022d0f7d26f7874865a579ae/packages/client/components/GroupingKanban.tsx#L73

JimmyLv:
Hi @Dschoordsch, I did some refactoring Improvement, could you please help to have a look again?

packages/client/components/BottomControlBarReadyButton.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const beforeDemoStart =
+      isDemoRoute() && reflectPrompts.every(({editorIds}) => !Array.isArray(editorIds))
+    if (beforeDemoStart) return
+
+    const isNotEditing = reflectPrompts.every(({editorIds}) => {
+      return editorIds === undefined || (Array.isArray(editorIds) && editorIds.length === 0)
+    })
+
+    setDelaySeconds(isNotEditing ? 30 : undefined)
+  }, [localPhase.phaseType === 'reflect', reflectionGroups, localPhase])
```
Dschoordsch, line 118: -1 dependencies should be
```ts
[localPhase.phaseType === 'reflect', reflectionGroups, localPhase.reflectPrompts]
```

packages/client/components/BottomControlBarReadyButton.tsx
```diff
@@ -0,0 +1,169 @@
...
+    if (localPhase.phaseType !== 'group') return
+
+    const isNotDragging = reflectionGroups?.every((group) =>
+      group.reflections?.every(
+        (reflection) => !reflection.isViewerDragging && !reflection.isDropping
+      )
+    )
+
+    setDelaySeconds(isNotDragging ? 30 : undefined)
+  }, [localPhase.phaseType === 'group', reflectionGroups, localPhase])
```
Dschoordsch, line 130: -1 `localPhase` is not used directly in this effect

packages/client/components/BottomControlBarReadyButton.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+    setDelaySeconds(isNotEditing ? 30 : undefined)
+  }, [localPhase.phaseType === 'reflect', reflectionGroups, localPhase])
+
+  useEffect(() => {
+    if (localPhase.phaseType !== 'group') return
+
+    const isNotDragging = reflectionGroups?.every((group) =>
+      group.reflections?.every(
+        (reflection) => !reflection.isViewerDragging && !reflection.isDropping
```
Dschoordsch, line 125: -1 it's missing `reflection.isEditing` and we probably also only should show the hint if there is at least one group, what do you think?

packages/client/components/BottomControlBarReadyButton.tsx
```diff
@@ -0,0 +1,169 @@
...
+      )
+    )
+
+    setDelaySeconds(isNotDragging ? 30 : undefined)
+  }, [localPhase.phaseType === 'group', reflectionGroups, localPhase])
+
+  useEffect(() => {
+    if (localPhase.phaseType !== 'vote') return
+
+    const teamVotesRemaining = votesRemaining || 0
```
Dschoordsch, line 135: -1 `votesRemaining` should be non-null. Also if you check for `null`/`undefined` you should use `??` to make your intention clear, since `!!0 === false`

packages/client/components/BottomControlBarReadyButton.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const myVotesRemaining = viewerMeetingMember?.votesRemaining || 0
```
Dschoordsch, line 136: -1 `??`

packages/client/components/BottomControlBarReadyButton.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const isNotVoting = teamVotesRemaining === 0 || myVotesRemaining === 0
```
Dschoordsch, line 137: -1 is there a situation where `myVotesRemaining > 0 && teamVotesRemaining === 0`? I think checking for `teamVotesRemaining` would be sufficient.
+1 `noVotesRemaining` would be a better variable name in my opinion.

packages/client/components/BottomControlBarReadyButton.tsx
```diff
@@ -0,0 +1,169 @@
...
+    setDelaySeconds(5 * 60)
+  }, [localPhase.phaseType === 'discuss'])
+
+  useEffect(() => {
+    // if the ready button is "full" before these conditions are met,
+    // the animation should start after 5s
+    if (progress !== 1) return
+
+    setDelaySeconds(5)
+  }, [progress, localPhase.phaseType])
```
Dschoordsch, line 154: -1 why did you add `localPhase.phaseType` here?
+1 you could optimize the dependency array by `[progress === 1]`

packages/client/components/BottomControlBarReadyButton.tsx
```diff
@@ -0,0 +1,169 @@
...
+
+  // only enable for facilitatorUser
+  const isFacilitator = viewerId === facilitatorUserId && !endedAt
+  if (!isFacilitator) {
+    return <>{children}</>
```
Dschoordsch, line 159: +1 you could just `return children` here

Dschoordsch (changes requested):

packages/client/components/BottomControlBarReadyButton.tsx
JimmyLv: there's no situation where `myVotesRemaining > 0 && teamVotesRemaining === 0`? because `teamVotesRemaining` actually includes the count of `myVotesRemaining`, so I'll simplify the conditions here to:
```ts
const noVotesRemaining = teamVotesRemaining === 0
```

Google Chrome-Retro #9  lv460051518+1’s Team-004077-20220823

JimmyLv: The reason I used `myVotesRemaining === 0` before was that it's possible that not all team members were online and participated in the voting for this meeting. So it started prompting when all of the facilitator's own votes had been cast. But I'll address this kind of case in a separate issue.

packages/client/components/BottomControlBarReadyButton.tsx
JimmyLv: Returning `children` directly would result in a JSX element error, so I added a return the type to the FC function:

```ts
const BottomControlBarReadyButton = (props: Props): JSX.Element => {
```

WebStorm-parabol  ~ccodingstyleparabol  – …packagesclientcomponentsBottomControlBarReady tsx-004079-20220823

packages/client/components/BottomControlBarReadyButton.tsx
JimmyLv: Yes, I agree that there needs to be at least one Group before the prompting starts.

packages/client/components/BottomControlBarReadyButton.tsx
```diff
@@ -115,29 +105,30 @@ const BottomControlBarReadyButton = (props: Props) => {
...
+  }, [localPhase.phaseType === 'group', reflectionGroups])
 
   useEffect(() => {
     if (localPhase.phaseType !== 'vote') return
 
-    const teamVotesRemaining = votesRemaining || 0
-    const myVotesRemaining = viewerMeetingMember?.votesRemaining || 0
-    const isNotVoting = teamVotesRemaining === 0 || myVotesRemaining === 0
+    const teamVotesRemaining = votesRemaining ?? 0
+    const noVotesRemaining = teamVotesRemaining === 0
```
Dschoordsch, line 128: +1 this seems a bit redundant, why not just `noVotesRemaining = !votesRemaining`? Or if you want to have it more explicit `noVotesRemaining = votesRemaining ?? 0 === 0`

Dschoordsch (approved):
Did not test again.

PR #6955, opened 2022-07-27
fix(resetPassword): Reset request count daily

### LINKED ISSUE #6712: Password reset problems
I'm turning a discussion we had into an issue and we'll continue to record when users have password reset problems here. There seems to have been an unusual amount of requests coming into support in the last 2 weeks.

Here's the original discussion https://github.com/ParabolInc/parabol/discussions/6301

mattkrick:

# Description

fixes #6712

users who reset their password often can now keep resetting it often

## Testing scenarios

- trigger a password reset 3 times in 1 day. then come back 24+ hours later & trigger another one. it works

### DIFF
--- a/packages/client/components/ForgotPasswordPage.tsx
+++ b/packages/client/components/ForgotPasswordPage.tsx
@@ -2,24 +2,24 @@
  * The password reset page. Allows the user to reset their password via email.
  *
  */
-import React from 'react'
 import styled from '@emotion/styled'
-import EmailInputField from './EmailInputField'
-import PlainButton from './PlainButton/PlainButton'
-import PrimaryButton from './PrimaryButton'
-import {emailRegex} from '../validation/regex'
-import Legitity from '../validation/Legitity'
-import AuthenticationDialog from './AuthenticationDialog'
-import {GotoAuthPage} from './GenericAuthentication'
-import DialogTitle from './DialogTitle'
-import {PALETTE} from '../styles/paletteV3'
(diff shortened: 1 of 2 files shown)


packages/server/graphql/mutations/emailPasswordReset.ts
```diff
@@ -36,6 +36,7 @@ const emailPasswordReset = {
           .table('PasswordResetRequest')
           .getAll(ip, {index: 'ip'})
           .filter({email})
+          .filter((row) => row('time').ge(yesterday))
```
mattkrick, line 39: this is the only real change

packages/client/components/ForgotPasswordPage.tsx
```diff
@@ -131,6 +131,12 @@ const ForgotPasswordPage = (props: Props) => {
     goToPage('signin', params.toString())
   }
 
+  const errorMessage = {
+    [AuthenticationError.USER_NOT_FOUND]: 'We couldn’t find that email. Please try again.',
+    [AuthenticationError.EXCEEDED_RESET_THRESHOLD]:
+      'Too many reset password attempts. Please try again later.'
```
mattkrick, line 137: add a more specific message. this isn't good because we are giving nefarious people a hint, but we'll assume most folks are good.

nickoferrall (approved):
Nice find! Works for me. I'm sure this will bring the Growth team plenty of joy

Issue #6916, opened 2022-07-22
Recurrence MVP: Basic `startRecurrence` mutation + recurrence objects


jmtaber129:
Backend spec: https://github.com/ParabolInc/parabol/issues/6888

Implement the `startRecurrence` mutation, along with any additional recurrence objects (e.g. meeting series) necessary:
* Create the new `MeetingSeries` object in postgres + graphql
* Add the new fields to the `NewMeeting` rethink object
* Create the `startRecurrence` mutation, which takes a meeting ID as an argument

Out of scope: 
* Anything on the frontend
* Restarting recurrence

The intention is to implement the basic foundation needed to execute other recurrence tasks (UI, `stopRecurrence`, `processRecurrence`, etc.) in parallel

### Acceptance Criteria
- [ ] Devs can run the `startRecurrence` mutation for a meeting in the graphql console, which results in the following
  - [ ] A new meeting series with relevant defaults exists
  - [ ] The meeting has a `scheduledEndTime` at a reasonable time in the future
  - [ ] The meeting has a `meetingSeriesId` associated with it in its rethink row
  - [ ] The recurrence rule represents a start time for the next meeting that makes sense.  E.g. the rrule shouldn't represent that the next meeting should start later that same day.
- [ ] Errors are returned if:
  - [ ] The meeting doesn't exist
  - [ ] The meeting already has a `meetingSeriesId` associated with it, and the series is inactive (restarting recurrence is out of scope)

PR #6957, opened 2022-07-27
feat(standups): `startRecurrence` mutation

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/6916

Implement the `startRecurrence` mutation, which takes a meeting that has never recurred, and makes it recurring by creating the `MeetingSeries` and updating the `Meeting` rethink model.

The mutation generation script adds all the UI boilerplate by default, so I went ahead and added a feature flagged recurrence button there.

Changes in this PR:
* `recurrence` feature flag
* `MeetingSeries` DB + GQL model
* `TeamPromptMeeting` updates (will eventually be changed to all `NewMeeting`s)
* Basic "start recurrence" button in the team prompt meeting options behind the feature flag

## Demo
In UI
Screen Shot 2022-07-27 at 7 44 27 PM

In the relay store after clicking the "start recurrence" button:
Screen Shot 2022-07-27 at 7 46 42 PM

The resulting row in the DB:
Screen Shot 2022-07-27 at 7 46 58 PM

## Testing scenarios
- [ ] Start a standup
- [ ] Click the start recurrence button, and check that the relay store is accurate
- [ ] Start a standup
- [ ] Run the `startRecurrence` mutation from the GQL console, and ensure the response is accurate

## Final checklist


packages/server/graphql/private/typeDefs/_legacy.graphql
```diff
@@ -3932,8 +3963,15 @@ type UserFeatureFlags {
...
-  """true if azuredevops is allowed"""
+  """
+  true if azuredevops is allowed
+  """
   azureDevOps: Boolean!
+
+  """
+  true if recurrence is allowed
+  """
+  recurrence: Boolean!
```
jmtaber129, line 3974: This is the only important change

packages/server/graphql/public/types/StartRecurrenceSuccess.ts
```diff
@@ -0,0 +1,16 @@
+import MeetingTeamPrompt from '../../../database/types/MeetingTeamPrompt'
+import {StartRecurrenceSuccessResolvers} from '../resolverTypes'
+
+export type StartRecurrenceSuccessSource = {
+  meetingId: string
+  meetingSeriesId: string
+}
+
+const StartRecurrenceSuccess: StartRecurrenceSuccessResolvers = {
+  meeting: async (source, _args, {dataLoader}) => {
```
BartoszJarocki, line 10: +1 nit, you can destructure source here

packages/server/postgres/migrations/1658834062488_createMeetingSeries.ts
```diff
@@ -0,0 +1,33 @@
...
+  await client.query(`
+    CREATE TABLE IF NOT EXISTS "MeetingSeries" (
+      "id" INT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+      "meetingType" "MeetingTypeEnum" NOT NULL,
+      "title" TEXT NOT NULL,
+      "recurrenceRule" TEXT NOT NULL,
+      "duration" INT NOT NULL,
+      "createdAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+      "updatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+      "cancelledAt" TIMESTAMP WITH TIME ZONE
```
BartoszJarocki, line 16: a little comment just to make sure we're on the same page. `TIMESTAMP WITH TIME ZONE` type in pg is stored as UTC internally but converted to time zone specified in pg config when returning the date. 

from pg docs 

> All timezone-aware dates and times are stored internally in UTC. They are converted to local time in the zone specified by the [TimeZone](https://www.postgresql.org/docs/current/runtime-config-client.html#GUC-TIMEZONE) configuration parameter before being displayed to the client.

imo it's best to have postgres set to UTC, too. this will guarantee we won't make any mistakes when calculating if the meeting should actually be closed, etc. 

to check what timezone is set you can run `SHOW TIMEZONE;`; to change `SET TIMEZONE='UTC` (it's db wise so we have to be very careful to not break anything else, if we need to actually do this)

packages/server/postgres/migrations/1658834062488_createMeetingSeries.ts
```diff
@@ -0,0 +1,33 @@
...
+    );
+
+    DROP TRIGGER IF EXISTS "update_MeetingSeries_updatedAt" ON "MeetingSeries";
```
BartoszJarocki, line 19: +1 do we actually need to drop it here or is it just to be safe?

packages/server/postgres/queries/src/getMeetingSeriesByIdsQuery.sql
```diff
@@ -0,0 +1,6 @@
+/*
+  @name getMeetingSeriesByIdsQuery
+  @param ids -> (...)
+*/
+SELECT "id", "meetingType", "title", "recurrenceRule", "duration", "createdAt", "updatedAt", "cancelledAt" FROM "MeetingSeries"
```
BartoszJarocki, line 5: +1 would `SELECT * FROM "MeetingSeries" be more future-proof?

BartoszJarocki (approved):
great work! I left some suggestions/comments/nits

packages/server/graphql/public/mutations/startRecurrence.ts
```diff
@@ -0,0 +1,76 @@
...
+    duration: 24 * 60 // 24 hours
+  })
+
+  await r
+    .table('NewMeeting')
+    .get(meetingId)
+    .update(
+      {
+        meetingSeriesId: newMeetingSeriesId,
+        scheduledEndTime: new Date(Date.now() + 24 * 60 * 60 * 1000) // 24 hours from now
```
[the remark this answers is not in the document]
jmtaber129: TIL!

packages/server/postgres/migrations/1658834062488_createMeetingSeries.ts
jmtaber129: Just to be safe - similar to a `CREATE TABLE IF NOT EXISTS` but for triggers

packages/server/postgres/migrations/1658834062488_createMeetingSeries.ts
jmtaber129: This probably won't be used when calculating recurrence effects - if this timestamp is used, it will probably only be used in the UI.

packages/server/graphql/public/mutations/startRecurrence.ts
```diff
@@ -0,0 +1,73 @@
...
+  const meeting = await dataLoader.get('newMeetings').load(meetingId)
+  if (!meeting) {
+    return standardError(new Error('Meeting not found'), {userId: viewerId})
+  }
+
+  if (meeting.meetingType !== 'teamPrompt') {
+    return standardError(new Error('Meeting is not a team prompt meeting'), {userId: viewerId})
+  }
+
+  const {teamId, meetingSeriesId} = meeting
```
mattkrick, line 33: -1 check viewer permissions (either they have a MeetingMember, or they are a teamMember, up to you)

packages/server/graphql/public/mutations/startRecurrence.ts
```diff
@@ -0,0 +1,73 @@
...
+  if (meetingSeriesId) {
+    return standardError(new Error('Meeting already has meeting series'), {userId: viewerId})
+  }
+
+  // Next meeting start is tomorrow at 9a UTC
+  const startDate = new Date(
```
mattkrick, line 39: +1 this feels brash, but I'm guessing we're gonna update this mutation by adding a date input or something later? good to ship for now!

packages/server/graphql/public/typeDefs/MeetingSeries.graphql
```diff
@@ -0,0 +1,44 @@
...
+  """
+  meetingType: MeetingTypeEnum!
+
+  """
+  The name of the meeting series used to generate the names of the recurring meetings
+  """
+  title: String!
+
+  """
+  The RFC-2445 recurrence rule that determines when meetings should start
```
mattkrick, line 21: +2 love seeing RFC-2445 called out here

packages/server/graphql/public/typeDefs/_legacy.graphql
```diff
@@ -6521,6 +6521,21 @@ type TeamPromptMeeting implements NewMeeting {
...
+  """
+  meetingSeriesId: ID
+
+  """
+  The meeting series this meeting is associated with if the meeting is recurring
+  """
+  meetingSeries: MeetingSeries
+
+  """
+  The timestamp the meeting is scheduled to end
```
mattkrick, line 6535: +1 add something like "is non-null if meetingSeriesId is non-null" (if that's true!, I think it is?)

packages/server/postgres/migrations/1658834062488_createMeetingSeries.ts
```diff
@@ -0,0 +1,33 @@
...
+import getPgConfig from '../getPgConfig'
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+    CREATE TABLE IF NOT EXISTS "MeetingSeries" (
+      "id" INT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+      "meetingType" "MeetingTypeEnum" NOT NULL,
+      "title" TEXT NOT NULL,
```
mattkrick, line 11: -1 could we make this something reasonable like `VARCHAR(255)`? (same for recurrenceRule)
2 reasons:
- prevent a DOS attack like someone adding a 5MB title that causes our DB to slow down. (I think we do this in the application layer, but it's always nice to prevent it at the DB layer, too!)
- give other devs an estimate about how large this field is

packages/server/graphql/public/types/TeamPromptMeeting.ts
```diff
@@ -0,0 +1,27 @@
...
+      return MeetingSeriesId.join(meetingSeriesId)
+    }
+
+    return null
+  },
+  meetingSeries: async ({meetingSeriesId}, _args, {dataLoader}) => {
+    if (meetingSeriesId) {
+      const series = await dataLoader
+        .get('meetingSeries')
+        .load(MeetingSeriesId.join(meetingSeriesId))
```
mattkrick, line 16: +1 preference to just pass in the database `meetingSeriesId` instead of the GraphQL GUID so we only have to do transforms for the ingress/egress, but if this is how the new dataloaders works i default to that pattern

packages/server/postgres/queries/getMeetingSeriesByIds.ts
```diff
@@ -0,0 +1,24 @@
...
+const getMeetingSeriesByIds = async (meetingSeriesIds: readonly string[]) => {
+  const meetingSeriesResults = await getMeetingSeriesByIdsQuery.run(
+    {ids: meetingSeriesIds.map((id) => MeetingSeriesId.split(id))},
+    getPg()
+  )
+
+  return meetingSeriesResults.map((meetingSeries) => {
+    return {
+      ...meetingSeries,
+      id: MeetingSeriesId.join(meetingSeries.id)
```
mattkrick, line 19: -1 I'll use this comment as a catch-all for the meetingSeriesId:
- There are 4 calls to `MeetingSeriesId.join`. 1 in the mutation, 2 in the Meeting resolver, 1 in the DB query. It feels strange because `TeamPromptMeeting.meetingSeriesId` has it in the resolver, but `MeetingSeries.id` doesn't have it in the resolver (it uses the default resolver). Makes it hard to know where in the code it's the prefixed GUID vs. the DBID (string vs. number). For example, if we added `Team.meetingSeries`, the developer would have to know to map over the `meetingSeries` to give them all GUIDs.

I propose the following:
- Add a `MeetingSeries.id` resolver that inputs a DBID and outputs a GUID
- remove the other calls to `MeetingSeriesId.join` except for `TeamPromptMeeting.meetingSeriesId`

That way everything gets transformed on the way out & we get a nice clean DBID to work with everywhere else. Open to other ideas!

packages/server/graphql/public/mutations/startRecurrence.ts
```diff
@@ -0,0 +1,73 @@
...
+      meetingSeriesId: newMeetingSeriesId,
+      scheduledEndTime: new Date(Date.now() + ms('24h')) // 24 hours from now
+    })
+    .run()
+
+  dataLoader.get('newMeetings').clear(meetingId)
+
+  // RESOLUTION
+
+  const data = {meetingId, meetingSeriesId: MeetingSeriesId.join(newMeetingSeriesId)}
```
mattkrick, line 68: -1 can remove `meetingSeriesId` here since it's not available on the SuccessPayload

mattkrick (changes requested):
really high quality work! just a couple housekeeping items & 1 security thing

packages/server/graphql/public/mutations/startRecurrence.ts
jmtaber129: Nice catch! Checked whether the user is a team member

packages/server/graphql/public/mutations/startRecurrence.ts
jmtaber129: Yep, this will probably change based on some meeting series config in a future iteration. Went ahead and added a TODO

packages/server/postgres/migrations/1658834062488_createMeetingSeries.ts
jmtaber129: TIL!  Fixed

packages/server/postgres/queries/getMeetingSeriesByIds.ts
jmtaber129: Yea, this sounds like the best way forward.  Updated

packages/server/graphql/public/mutations/startRecurrence.ts
jmtaber129: Done (removed here and from the `StartRecurrenceSuccessSource` type)

mattkrick:
_clean!_ nice work

PR #6960, opened 2022-07-28
chore: Added one review required to code review policy

BartoszJarocki:
We've [made a decision](https://parabol.slack.com/archives/C836NA350/p1658847963962769) to make `One Review Required` part of our `Code Review Policy`. This PR removes the part saying it was an experiment. 

Please let me know if you'd like to see any additional clarification on that, thank you!

### DIFF
--- a/docs/codeReview.md
+++ b/docs/codeReview.md
@@ -47,6 +47,7 @@ A PR advances from self-review to reviewer-review and finally maintainer review 
   - If there are only +2 comments, the PR can be merged directly by the maintainer.
 
 ## Metrics Representative
+
 - Metrics Representative ensures any analytics related changes work well with downstream data services
 - This role is currently filled by @tianrunhe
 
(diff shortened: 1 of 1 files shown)


docs/codeReview.md
```diff
@@ -58,14 +59,12 @@ A PR advances from self-review to reviewer-review and finally maintainer review
 - If you need to clarify parts of the code, check if it can be done by adding comments or improve naming of variables/functions/classes
 - When you replied or resolved all comments, move issue back to reviewer- or maintainer-review
 
-## Experiment
+## One Review Required
 
-We want to run an experiment for two months. During the experiment, all PRs will still require a Reviewer Review and a Maintainer Review. However, if the developer creating the PR feels that only one review is required, they can add a label: “One Review Required”. The Reviewer Reviewer then adds a comment saying whether they agree that the PR could be merged once they’ve approved it.
+If the developer creating the PR feels that maintainer review isn't necessary, they can add a label: “One Review Required”. The Reviewer Reviewer then adds a comment saying whether they agree that the PR could be merged once they’ve approved it.
```
Dschoordsch, line 64: +1 While I understand the reasoning behind the comment added by the reviewer, in practice I did not see any such comments added. So either we emphasize this more, or we adopt to the lived practice and remove this requirement.

Dschoordsch (approved):

docs/codeReview.md
nickoferrall, line 64: +1 could say `they should add the label` instead of `they can add a label` as the later implies it's optional

nickoferrall: Perhaps the Reviewer Reviewer only adds a comment if they disagree with the label as they're more likely to agree than disagree?

nickoferrall (approved):
LGTM!

docs/codeReview.md
BartoszJarocki: yes, good point! I'll change that to what @nickoferrall suggested, thank you!

BartoszJarocki: @Dschoordsch @nickoferrall, how about `If the Reviewer thinks it's not safe to have just one review, it's his responsiblity to add a comment saying that Maintainer review is required to merge the PR.`?

BartoszJarocki: good point. will fix!

nickoferrall: Looks good to me! One small edit - we could say `it's their responsibility` instead

BartoszJarocki: oh, that's right! thank you 😅 fixed!

tianrunhe (approved):

JimmyLv (approved):

Issue #6938, opened 2022-07-26
Remove manual pausing


nickoferrall:
Some users are manually pausing their team members after a meeting even though those users are still active. In the prorated invoice, the org receives a discount for the days those team members were "inactive". 

Team members that haven't used Parabol for 30 days are automatically deactivated, so we can remove the option to manually pause members while still being fair to our users.

See Slack discussion [here](https://parabol.slack.com/archives/C03HPHWH55W/p1658760339362439) for further context.

### AC

- [ ] In the org Members page, remove the manual pausing toggles

**EE**: 1 hour

PR #6962, opened 2022-07-28
feat: remove manual pausing

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6938

This PR removes the option to manually pause a user & removes the `inactivateUser` mutation. It doesn't remove `InvoiceItemType.PAUSE_USER` or update the `generateInvoice` logic because users will have already been manually paused and we'll want that to show up in their next invoice.


mattkrick (changes requested):
+1 for posterity, could you also delete `MAX_MONTHLY_PAUSES` from the constEnums?

nickoferrall:
> +1 for posterity, could you also delete MAX_MONTHLY_PAUSES from the constEnums?

Good point, removed 👍 There are no other changes requested so I'll merge this PR

PR #6966, opened 2022-07-28
fix: allow to go to the discuss phase without voting

### LINKED ISSUE #6631: Give users feedback that they need to vote in Vote phase

igorlesnenko:
Fixes #6631

In this PR I make it possible to move to the discussion phase without voting.

Previous discussion here: #6886

How to test:
- Create a new retro meeting
- Advance to the voting phase but do not case any votes
- Click next, see the button works and advances to the discussion phase

### DIFF
--- a/packages/client/mutations/ResetRetroMeetingToGroupStageMutation.ts
+++ b/packages/client/mutations/ResetRetroMeetingToGroupStageMutation.ts
@@ -33,6 +33,7 @@ graphql`
           viewerVoteCount
         }
       }
+      ...MeetingControlBar_meeting
     }
   }
 `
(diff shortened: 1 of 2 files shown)


igorlesnenko:
@mattkrick Could you please take a look? I'm not sure if it is the right way to do this

mattkrick:
LGTM! thanks for fixing this

1 reviewer can be any dev, not just maintainers

Issue #6971, opened 2022-07-29
Add key controls to meeting selector modal


nickoferrall:
I'd like to be able to use my keys to switch between meeting types.

### AC:

- [ ] Can use left & right arrow keys to alternate between meeting types
- [ ] Can press enter to begin a meeting

PR #6972, opened 2022-07-29
feat: meeting modal key shortcuts

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6971

https://www.loom.com/share/772c611a565943b6a2e3032f1fdfd552

### To test

- [ ] Pressing the left & right arrow switches between meeting types
- [ ] Pressing enter starts the selected meeting type
- [ ] Pressing enter doesn't start the meeting if a portal or modal is open


packages/client/components/NewMeetingCarousel.tsx
```diff
@@ -86,24 +87,49 @@ interface Props {
...
 }
 
 const NewMeetingCarousel = (props: Props) => {
-  const {idx, setIdx, meetingOrder} = props
+  const {idx, setIdx, meetingOrder, onStartMeetingClick} = props
 
   // TODO: remove when standups feature flag removed
   const moreThanThreeSlides = meetingOrder.length > 3
 
+  useHotkey('left', () => {
```
nickoferrall, line 99: `swiper` package doesn't recognise left or right key down events, I _think_ because their keyboard functionality lets you swipe using the up & down arrow key. I changed the direction to `horizontal`, but no dice. Went with `useHotKey` to build something quickly

igorlesnenko: In current implementation of carousel selected slide in swiper is not synchronized with the selected slide in our modal. 

It is kind of different concepts:

For example when we drag a view, our selected slide don't change. However, in swiper selected slide changes.  Selected slide for us is a slide that we clicked on and highlighted. Selected slide for swiper is the first slide in the view.

So I think your implementation is good.

And probably we can replace swiper package later with our own implementation.

packages/client/components/NewMeetingCarousel.tsx
```diff
@@ -86,24 +87,49 @@ interface Props {
...
+
+  // keycode is a number but package thinks it's a string
+  const onKeyPress = (_swiper: unknown, keycode: string) => {
+    if (parseInt(keycode, 10) === 13) {
+      // inefficient, but only happens on enter
+      const isModalOpen =
+        document.querySelector(`div[id='templateModal']`) ||
+        document.querySelector(`div[id='portal']`)
+      if (!isModalOpen) {
+        onStartMeetingClick()
```
igorlesnenko, line 116: Not sure I fully understand a comment to this piece. Why on enter `isModalOpen` becomes `false`?

igorlesnenko (approved):
If you go to the last slide by hitting `left` from the beginning, the view is not scrolled to this slide

I think for now it is not critical issue, and could be fixed only if easy. We can create separate issue

https://user-images.githubusercontent.com/466991/182107115-09fa3920-be4c-40f5-baa3-ba7819d12063.mov

packages/client/components/NewMeetingCarousel.tsx
nickoferrall: `isModalOpen` is true when a modal or portal exists in the dom. It's inefficient to search through the dom, but we only do this when the user presses the enter key, so I think it's OK. `isModalOpen` doesn't become `false` when the user presses enter.

I'd be happy to update the comment if that's unclear?

packages/client/components/NewMeetingCarousel.tsx
nickoferrall: Ah great point! I didn't have standups enabled locally, so there were only 3 meeting types.

It looks like there's a bug, probably with the package, that doesn't update the active CSS correctly. For example, when I click "Async Standup", the "Retrospective" still has the `swiper-slide-active` class:

active-idx

I spent some time looking into the package but didn't find a solution, so for now if `idx === 0` and the press left, they'll stay on idx 0. I think you're right that we should probably replace the swiper package with our own implementation.

packages/client/components/NewMeetingCarousel.tsx
```diff
@@ -86,24 +87,54 @@ interface Props {
...
+    }
+  })
+
+  const enterKey = 13
+  // keycode is a number but package thinks it's a string
+  const onKeyPress = (_swiper: unknown, keycode: string) => {
+    if (parseInt(keycode, 10) === enterKey) {
+      // inefficient, but only happens on enter
+      const isModalOpen =
+        document.querySelector(`div[id='templateModal']`) ||
```
mattkrick, line 118: -1 i'd love to avoid resorting to checking the DOM & keep on the logic in the react lifecycle.

there's a few ways to do this, top of mind is to use focus:
- focus the meeting card with the active index (if it's a div you can use tabIndex={0} to make it focusable)
- put the keypress handler on the active card
- stop propagation if necessary

mattkrick (changes requested):

packages/client/components/NewMeetingCarousel.tsx
nickoferrall: Cool, TIL about tabIndex!

PR #6997, opened 2022-08-03
feat(metrics): Add `isActive` and `featureFlags` as new User Properties for Segment identify calls

### LINKED ISSUE #6952: Refactor User properties in Segment
User Properties in Segment are basically dimensions we can slice our analysis on:
Screen Shot 2022-07-26 at 4 27 05 PM

For example, if we build a M1>M2 funnel analysis, we could see the differences by, say, invited vs. non-invited, patient 0 vs. others, enterprise tier vs. personal tier etc.

What do we have for user properties today?

1. `avatar`: I don't know why we needed this here... completely waste. We should remove it?
2. ~`createdAt`: seems useful except in Segment you can't do any date parsing functions so not quite useful either~ (we can use [derived properties](https://help.amplitude.com/hc/en-us/articles/360058731292#date-time-functions)). We can keep this though.
3. `email` & `id`: for identity purpose we should keep them
4. ~`name`: seems useless. I can't think of any use cases where we would do analysis based on users' preferred names. Might just remove it.~ (could be useful when we look up users in Amplitude)
5. `highestTier`: probably the only useful attribute for analysis.

I think what would be quite useful & easy to add are:
1. `isPatient0`
2. `isActive`: this could be a little bit confusing if added as it might change time by time but I think it'd be helpful
3. `isBillingLead`
4. `isTeamLead`

- [ ] `avatar` and `name` are removed from Segment identify call
- [ ] `isPatient0`, `isActive` and `isRemoved` are added to Segment identify call

tianrunhe:
# Description

Partially Fixes #6952. As a first step, I added `isActive` and `featureFlags` because the infomation is available upon socket connection. Any other properties suggested in the issue are expensive to calculate. For example, `isTeamLead`, we could call `identify` on `promoteToTeamLead` but that would only work for **new** team lead. To "backfill" existing team leads, we have to `identify` them on login, but it seems expensive to calculate if the user is a team lead on all the logins for every user? I need to think about the performance impact more hence not including them in this PR.

## Demo

Screen Shot 2022-08-03 at 4 55 37 PM

## Testing scenarios
 - [ ] Login and check Segment to verify an `identify` call was issued, with appropriate `traits` included.

## Final checklist

### DIFF
--- a/packages/server/billing/helpers/adjustUserCount.ts
+++ b/packages/server/billing/helpers/adjustUserCount.ts
@@ -167,6 +167,19 @@ export default async function adjustUserCount(
     }
   }
 
+  if (
+    type === InvoiceItemType.PAUSE_USER ||
+    type === InvoiceItemType.AUTO_PAUSE_USER ||
+    type === InvoiceItemType.REMOVE_USER
+  ) {
+    segmentIo.identify({
+      userId,
+      traits: {
+        isActive: false
+      }
+    })
+  }
+
   const prorationDate = options.prorationDate ? toEpochSeconds(options.prorationDate) : undefined
   const hooks = proOrgs.map((org) => {
     return new InvoiceItemHook({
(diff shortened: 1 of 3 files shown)


tianrunhe:
@nickoferrall GitHub suggests you 😜 Thanks man!

nickoferrall (approved):
LGTM!

PR #6998, opened 2022-08-04
chore(security): upgrade node to 16.16.0

adaniels-parabol:
# Description

Upgrades node to 16.16.0 from 16.14.0 (Ironbank finding security fix)


mattkrick:
I think the @types/node version is for node v18, but that shouldn't really affect us as I don't think we utilize anything solely for v18 & the typecheck passed.

packages/integration-tests/package.json
```diff
@@ -11,7 +11,7 @@
   },
   "devDependencies": {
     "@playwright/test": "^1.23.2",
-    "@types/node": "16",
+    "@types/node": "18.6.3",
```
Dschoordsch, line 14: +1 I assume this should stay at something like `^16...`, also in `packages/server` it's still set to `"16"`

Dschoordsch (approved):

mattkrick:
weird that playwright isn't recognizing node....

Dschoordsch:
> weird that playwright isn't recognizing node....

I reckon this is just a system issue. Since the command is run with `sudo` the search path is different.

PR #6989, opened 2022-08-02
feat: Updated new meeting settings area

igorlesnenko:
Fixes #5603

**Next PR: #7001**

image

image

How to test:

- Open new meeting dialog
- Make sure all new controls works for all meeting types bot on mobile and desktop

PR #7001, opened 2022-08-04
feat: team members preview in new meeting team selector

### LINKED ISSUE #5604: Team member preview in new meeting team selector
See #5589 for details

## Acceptance criteria
- [ ] Below the meeting types is the team selector. It displays a grid of 2x2 avatars of team members (ideally these would be randomly picked from the full group of team members) and the team name.
- [ ] If a team has less than 4 members, [here's how to handle the grid of avatars](https://www.figma.com/file/5ouU5EQOZ0ELcBA0xWZGdQ/Start-a-new-meeting?node-id=685%3A11301).

igorlesnenko:
Fixes #5604

**Prev PR: #6989**

https://user-images.githubusercontent.com/466991/182911071-a49393f5-192c-43e6-978e-7e0a987551b4.mov

How to test:

- Open new meeting dialog
- Switch between teams with different count of members
- See member avatar previews are displayed correctly and randomized


packages/client/components/NewMeetingTeamPicker.tsx
```diff
@@ -32,13 +33,36 @@ const NewMeetingTeamPicker = (props: Props) => {
...
+        ...NewMeetingTeamPickerAvatars_team
+        name
+      }
+    `,
+    selectedTeamRef
+  )
+
+  const teams = useFragment(
+    graphql`
+      fragment NewMeetingTeamPicker_teams on Team @relay(plural: true) {
```
JimmyLv, line 49: TIL, Good to know the [`@relay(plural: Boolean)` GraphQL directive](https://relay.dev/docs/api-reference/graphql-and-directives/#relayplural-boolean)

JimmyLv (approved):
Great work! 👍

packages/client/components/MenuToggleV2Text.tsx
```diff
@@ -37,14 +33,14 @@ const MenuToggleTitle = styled('div')({
 
 interface Props {
   label: string
-  icon?: string
+  icon?: React.ReactElement
   title: string
 }
 const MenuToggleV2Text = forwardRef((props: Props, ref: any) => {
```
Dschoordsch, line 39: +1 Maybe we could find a better name for this component?

packages/client/components/NewMeetingTeamPickerAvatars.tsx
```diff
@@ -0,0 +1,84 @@
...
+const AvatarWrapper = styled('div')<{count: number}>(({count}) => ({
+  width: count > 1 ? 20 : 'auto',
+  height: count > 2 ? 20 : 'auto'
+}))
+
+const StyledAvatar = styled(Avatar)({
+  border: `2px solid ${PALETTE.SLATE_200}`
+})
+
+const getShuffledArr = (arr) => {
```
Dschoordsch, line 29: +1 maybe move `getShuffledArr.ts` to `/client/utils/` and use it on server and client?

packages/client/components/NewMeetingTeamPickerAvatars.tsx
```diff
@@ -0,0 +1,84 @@
...
+          picture
+        }
+      }
+    `,
+    teamRef
+  )
+
+  const {teamMembers} = team
+
+  if (!teamMembers) {
```
Dschoordsch, line 59: -1 It seams `teamMembers` is not nullable and it would violate the hook precondition when you return before `useMemo`.
If you're expecting this as some known exception in rare cases, please add a comment as explanation.

Dschoordsch (changes requested):
Looks good!

packages/client/components/MenuToggleV2Text.tsx
igorlesnenko: Renamed to `startAdornment`, taking inspiration from material-ui library

packages/client/components/NewMeetingTeamPickerAvatars.tsx
igorlesnenko: Removed null check, not needed here

packages/client/components/MenuToggleV2Text.tsx
Dschoordsch: `icon` name was fine, I commented on the `MenuToggleV2Text` name. `startAdornment` is ok, but slightly less intuitive.

packages/client/utils/getShuffledArr.ts
```diff
@@ -0,0 +1,10 @@
+const getShuffledArr = <T>(arr: T[]): T[] => {
```
Dschoordsch, line 1: -1 I didn't want to have this file duplicated, It exists already in `server/utils` and we're fine using `parabol-client/utils` imports in server. So please remove the version in `server/utils`

Dschoordsch (changes requested):

packages/client/components/NewMeetingTeamPickerAvatars.tsx
```diff
@@ -56,12 +53,8 @@ const NewMeetingTeamPickerAvatars = (props: Props) => {
 
   const {teamMembers} = team
 
-  if (!teamMembers) {
-    return null
-  }
-
   const randomAvatars = useMemo(() => {
-    const randomTeamMembers = getShuffledArr(teamMembers)
+    const randomTeamMembers = getShuffledArr(teamMembers as TeamMember[])
```
Dschoordsch, line 57: -1 teamMembers is not resolved as nullable for me? If it is for you, please change to `teamMembers ?? []`

igorlesnenko: @Dschoordsch it is not nullable, but without specifying that type it complains about 
```
  The type 'readonly { readonly id: string; readonly picture: string; }[]' is 'readonly' and cannot be assigned to the mutable type '{ readonly id: string; readonly picture: string; }[]'.
  ```

Comment (Dschoordsch) [reply packages/client/components/NewMeetingTeamPickerAvatars.tsx:57]: Then let's fix the signature of `getShuffledArr` as it does a `slice` at the beginning anyways.

File: packages/client/utils/getShuffledArr.ts (hunk 5 above)
Comment (igorlesnenko) [reply packages/client/utils/getShuffledArr.ts:1]: Ah, I didn't notice we had exactly same function, removed

File: packages/client/components/MenuToggleV2Text.tsx (hunk 2 above)
Comment (igorlesnenko) [reply packages/client/components/MenuToggleV2Text.tsx]: Renamed back to icon, renamed MenuToggleV2Text

File: packages/client/components/NewMeetingTeamPickerAvatars.tsx (hunk 6 above)
Comment (igorlesnenko) [reply packages/client/components/NewMeetingTeamPickerAvatars.tsx]: Fixed

File: packages/client/components/NewMeetingTeamPickerAvatars.tsx [hunk 7]
```diff
@@ -0,0 +1,72 @@
...
+        }
+      }
+    `,
+    teamRef
+  )
+
+  const {teamMembers} = team
+
+  const randomAvatars = useMemo(() => {
+    const randomTeamMembers = getShuffledArr(teamMembers)
```
mattkrick, line 52: +1 @igorlesnenko I'm assuming the goal here is to help the viewer recognize the team without using the team name. What do you think about:
- always showing the team lead, since they're probably the most recognizable user on the team ("Oh yeah, that's Igor's team")
- removing the viewer from the shuffle when the team size > 1 & they're not the team lead (that way, we can provide as much useful info as possible)

up to you, lgtm!

mattkrick (approved):

packages/client/components/NewMeetingTeamPickerAvatars.tsx
Dschoordsch: I like putting the team lead first.
I would only hide the viewer if team size > 4, having only 1 head shown for a team of 2 would be confusing.

Dschoordsch (approved):

packages/client/components/NewMeetingTeamPickerAvatars.tsx
igorlesnenko: This all make sense to me, I'll merge this PR for now and do changes in new issue

PR #7003, opened 2022-08-04
feat(metrics): identify user's tier upon login

tianrunhe:
# Description

Before we only identify user's `highestTier` [upon tier changes](https://github.com/ParabolInc/parabol/blob/master/packages/server/utils/setUserTierForUserIds.ts#L70). This causes [majority of our users in Segment/Amplitude has `none` highestTier](https://analytics.amplitude.com/parabol/chart/mknrler?source=redirect%3A%20chart%20saved). 

## Demo
Screen Shot 2022-08-04 at 3 46 47 PM

## Testing scenarios
- [ ] Login and verify an `identify` call was issued in Segment, with `highestTier` as a property in the User's traits

## Final checklist

### DIFF
--- a/packages/server/graphql/private/mutations/connectSocket.ts
+++ b/packages/server/graphql/private/mutations/connectSocket.ts
@@ -85,7 +85,8 @@ const connectSocket: MutationResolvers['connectSocket'] = async (
     userId,
     traits: {
       isActive: true,
-      featureFlags: user.featureFlags
+      featureFlags: user.featureFlags,
+      highestTier: user.tier
     }
   })
   return user


tianrunhe:
@nickoferrall literally a "1 line code change" 😜

packages/server/graphql/private/mutations/connectSocket.ts
```diff
@@ -85,7 +85,8 @@ const connectSocket: MutationResolvers['connectSocket'] = async (
     userId,
     traits: {
       isActive: true,
-      featureFlags: user.featureFlags
+      featureFlags: user.featureFlags,
+      highestTier: user.tier
```
nickoferrall, line 89: Should we change this to `tier` instead of `highestTier` in Segment?

nickoferrall (approved):
LGTM

packages/server/graphql/private/mutations/connectSocket.ts
tianrunhe: Unfortunately It was [called `highestTier`](https://github.com/ParabolInc/parabol/blob/master/packages/server/utils/setUserTierForUserIds.ts#L70) and populated that way already today. Renaming this would cause two separate properties in Amplitude which is more confusing I think... I don't have much tension here.

PR #7011, opened 2022-08-05
fix: turn on noImplicitAny for server

### LINKED ISSUE #6162: Add noImplicitAny

- [ ] fix ~200 errors
- [ ] we'll merge it blindly
- [ ] create a copy of this issue named Part (8 of 8) so we can re-prioritize for the next person. We'll keep doing this until it's done

AC

Estimate: 4 hours

mattkrick:

# Description

Fixes #6162

No more implicit any in the server!
I tried to avoid changing runtime logic wherever possible
pg migrate uses our tsconfig, which means even though we explicitly ignore typechecking migrations (because the types referenced are probably changed in the migration) we still have to make sure migrations don't have implicitAny


packages/client/modules/outcomeCard/components/OutcomeCardAssignMenu/TaskFooterTeamAssigneeMenu.tsx
```diff
@@ -194,16 +194,15 @@ const TaskFooterTeamAssigneeMenu = (props: Props) => {
           />
         )
       })}
-      {addIntegrationModalPortal(
-        (isGitHubTask || isJiraTask) && (
+      {(isGitHubTask || isJiraTask) &&
```
mattkrick, line 197: changes runtime logic

packages/server/dataloader/foreignKeyLoaderMaker.ts
```diff
@@ -34,18 +26,20 @@ export function foreignKeyLoaderMaker<
...
   primaryLoaderKey: LoaderName,
   foreignKey: KeyName,
-  fetchFn: (keys: readonly LoaderType<LoaderName>[KeyName][]) => Promise<LoaderType<LoaderName>[]>
+  fetchFn: (
+    keys: readonly LoaderType<LoaderName>[KeyName][]
+  ) => Promise<(LoaderType<LoaderName> & {id: string | number})[]>
 ) {
   type T = LoaderType<LoaderName>
-  type PrimaryKeyT = LoaderKeyType<LoaderName>
+  type PrimaryKeyT = string | number
```
mattkrick, line 34: The runtime logic showed that if an item was non-null, it would have an `id` field.
If that's true, then the id field is always gonna be a number or string

packages/server/integrations/jiraServer/JiraServerRestManager.ts
```diff
@@ -230,7 +230,7 @@ export default class JiraServerRestManager implements TaskIntegrationManager {
       return issueTypes
     }
 
-    const bestIssueType = issueTypes.find((type) => type.name === 'Task') || issueTypes.values[0]
+    const bestIssueType = issueTypes.find((type) => type.name === 'Task') || issueTypes[0]
```
mattkrick, line 233: caught a bug!

packages/server/utils/AzureDevOpsServerManager.ts
```diff
@@ -109,18 +109,29 @@ export interface WorkItemBatchResponse {
...
+  fields: WorkItemFields
   id: number
   relations: WorkItemRelations[]
   rev: number
   url: string
 }
 
 export interface ReferenceLinks {
-  links: object
+  html?: {
```
mattkrick, line 132: not sure if correct, but this is what the runtime logic said it looked like

Dschoordsch (approved):

packages/server/postgres/migrations/1625576993065_addAtlassianAuthTable.ts
```diff
@@ -40,7 +39,7 @@ export async function up(): Promise<void> {
 
   const atlassianIntegrations = await r
     .table('AtlassianAuth')
-    .filter((row) => row('accessToken').default(null).ne(null))
+    .filter((row: any) => row('accessToken').default(null).ne(null))
```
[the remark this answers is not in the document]
mattkrick: pure laziness. fixed!

packages/client/components/DropdownMenuToggle.tsx
```diff
@@ -28,7 +28,7 @@ interface InputStyleProps {
 const InputBlock = styled('div')<InputStyleProps>(
   ({disabled, size}) => ({
     ...ui.fieldBaseStyles,
-    ...ui.fieldSizeStyles[size],
+    ...ui.fieldSizeStyles[size as keyof typeof ui.fieldSizeStyles],
```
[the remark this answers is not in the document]
mattkrick: you're right, but it leaves me confused.
why is using a string to access a constant object OK here? shouldn't size be required to be small, medium, large? if string "foo" got passed in, it would yield undefined...

Issue #6858, opened 2022-07-06
Poker: Discussion drawer is not accessible (or troublesome) on mobile devices


87prashant:
## Issue - Bug
Tested on Chrome and Firefox on Android:
#### Chrome
Weird behavior due to URL bar when scrolling:

https://user-images.githubusercontent.com/106697681/177599519-9f05f8c1-218f-4ffb-84e6-1b56df5b27e6.mp4

#### Firefox
Similar to Chrome

https://user-images.githubusercontent.com/106697681/177599448-a76b47ee-721f-4d41-be24-d72d6e19b15a.mp4

Moreover, we could implement similar to this:

https://user-images.githubusercontent.com/106697681/177599493-bcc84f4f-e96a-4931-813d-13860c24836e.mp4

jordanh:
Thank you @87prashant for your report. Going to prioritize this as a p4. If we get other independent reports that this is bugging our users (or we see an uptick Android usage) we will reprioritize higher

87prashant:
`dvh` now supported on chrome android.

PR #7024, opened 2022-08-08
fix: Fixed Discussion drawer is not accessible on mobile devices

gitstart:
# Description

Fixes/Partially Fixes #[6858](https://github.com/ParabolInc/parabol/issues/6858)

## Demo

## Testing scenarios

- [ ] Scenario A
  - Step 1
  - Step 2...

- [ ] Scenario B
  - Step 1
  - Step 2....

## Final checklist


gitstart:
@igorlesnenko please can you review?

igorlesnenko:
@gitstart can you please link the original issue?

gitstart:
@igorlesnenko added link to original issue. thanks for the feedback

packages/client/components/EstimatePhaseDiscussionDrawer.tsx
```diff
@@ -26,6 +26,7 @@ const Drawer = styled('div')<{isDesktop: boolean; isOpen: boolean}>(({isDesktop,
   bottom: 0,
   top: 0,
   right: isDesktop ? 0 : undefined,
+  paddingBottom: isDesktop ? 0 : 56,
```
87prashant, line 29: I think hardcoding would not solve the problem and it also disturbs the design on mobile.

[Here](https://ishwar-rimal.medium.com/fixing-that-address-bar-issue-in-mobile-browsers-once-and-for-all-8c283fc88e56) is a post if it could help.

87prashant:
I don't know how to test it on mobile during development.
@igorlesnenko can you suggest any way to do so?

packages/client/components/EstimatePhaseDiscussionDrawer.tsx
jesjo001: I will look into this but meanwhile to test, you can forward your local development port to a live testing address with ngrok and access the address on mobile.
@87prashant

87prashant: Thanks for helping🙏

jesjo001: After Researching this issue and trying out different suggested fixes of which none of them seems to work, I would highly suggest we go with this hard-coded fix as it seems to be the only thing that works (tested on both Chrome and Firefox on Android, See attached Video). It's also worth noting that there is no official / community agreed on a fix for this issue only workarounds based on different approaches which sometimes would involve modifying the CSS of the body of the webpage which is not visible in this case due to the fact that it will affect multiple pages in different ways. Based on this I highly suggest we go along with this fix as it addresses the issue. 

https://user-images.githubusercontent.com/31279142/184479188-f3821c76-45a6-41f4-93f2-47b51214be49.mp4

Dschoordsch: @jesjo001 did you try setting the height to `100dvh`[^1]?

[^1]: https://dev.to/frehner/css-vh-dvh-lvh-svh-and-vw-units-27k4

Dschoordsch (changes requested):
See my comment about `100dvh`

gitstart:
Hi @Dschoordsch, requested  changes have been implemented you can give this PR another look.

87prashant:
Working fine on Firefox, but is currently not supported on Chrome.
 
Screenshot 2022-09-13 172714

packages/client/components/EstimatePhaseDiscussionDrawer.tsx
```diff
@@ -32,6 +32,7 @@ const Drawer = styled('div')<{isDesktop: boolean; isOpen: boolean}>(({isDesktop,
   zIndex: ZIndex.SIDEBAR,
   [`@media screen and (max-width: ${Breakpoint.POKER_DISCUSSION_FULLSCREEN_DRAWER}px)`]: {
     width: '100vw',
+    height: '100dvh',
```
87prashant, line 35: @Dschoordsch
`dvh` currently does not work on chrome

Dschoordsch: My fault, thanks for checking!

gitstart: @Dschoordsch  | @87prashant  whats the way forward on this.

gitstart:
Hi @Dschoordsch, reverted to the requested changes you can take a look once again.

Dschoordsch (changes requested):
I tested the change. It makes the buttons available, but I think this is just a workaround rather than the correct solution.
In the screen capture you can see that hiding the address bar on the main estimate screen doesn't move the bottom bar. In the discussion drawer however there is big gap when the address bar is visible and hiding it moves it. It should always stay attached to the bottom.
https://user-images.githubusercontent.com/7331043/193242363-c4d46bfa-c001-4e96-8ebc-fe1892984f86.mp4

gitstart:
Hi @Dschoordsch, closing this PR we are not able to complete works to.

Dschoordsch:
@87prashant we could do that. We can combine the solution from @gitstart with `dvh` where supported, similar to
https://github.com/ParabolInc/parabol/blob/0d46c2542ff7e7d78bbf5c8c0104a287858afb69/packages/client/components/TeamPrompt/TeamPromptDiscussionDrawer.tsx#L43-L46

87prashant:
`dvh` now supported on chrome android.

Issue #6917, opened 2022-07-22
Recurrence MVP: `processRecurrence` mutation


jmtaber129:
Backend spec: https://github.com/ParabolInc/parabol/issues/6888

Blocked on https://github.com/ParabolInc/parabol/issues/6916

Implement the idempotent `processRecurrence` mutation - for each meeting series:
* Start any meetings that should be started due to recurrence
* End any meetings that should be ended due to recurrence

### Acceptance Criteria (optional)
- [ ] When `processRecurrence` is executed:
  - [ ] Any meetings that should currently be open are opened
  - [ ] Any meetings with a `scheduledEndTime` before `now` are closed
  - [ ] Inactive meeting series are ignored
- [ ] Running `processRecurrence` twice consecutively (within a minute or two) has no effect (idempotency)

PR #7023, opened 2022-08-08
feat(standups): Recurrence - `stopRecurrence` mutation

BartoszJarocki:
# Description

Fixes #6918 

## Demo

localhost_3000_meet_fEobIylTH2_responses
localhost_3000_meet_fEobIylTH2_responses (1)

## Testing scenarios

- [ ] Create a standup meeting and...
  - Set up recurrence using menu in the top right corner
  - Stop the recurrence using menu in the top right corner
  - Check rethinkdb if given meeting have `scheduledEndTime` set to null
  - Check postgres if related meeting series have `cancelledAt` set to a proper date

## Final checklist

- [x] I checked the [code review guidelines](../docs/codeReview.md)
- [x] I have added [Metrics Representative](../docs/codeReview.md#metrics-representative) as reviewer(s) if my PR invovles metrics/data/analytics related changes
- [x] I have performed a self-review of my code, the same way I'd do it for any other team member
- [x] I have tested all cases I listed in the testing scenarios and I haven't found any issues or regressions
- [x] Whenever I took a non-obvious choice I added a comment explaining why I did it this way
- [x] I added the label `One Review Required` if the PR introduces only minor changes, does not contain any architectural changes or does not introduce any new patterns and I think one review is sufficient'
- [x] PR title is human readable and could be used in changelog

Issue #6888, opened 2022-07-14
Standups: Backend Recurrence MVP Implementation Spec


jmtaber129:
See https://github.com/ParabolInc/parabol/issues/6852 for other discussion and some alternatives considered.

# High-level overview
We store state about recurring meetings/activities in a new postgres table, `MeetingSeries`, which contains metadata about its collection of meetings, and how these meetings recur.  We introduce a new mutation, `processRecurrence`, which uses the recurrence information in `MeetingSeries` to idempotently open and close meetings accordingly.  We then run this new mutation on a 5-minute interval to keep actual meeting state in-sync with what the meeting series' recurrence specifies.

# Spec
### `MeetingSeries` postgres table
* `id` (non null)
* `meetingType` - the type of the meeting.  Should just be `teamPrompt` for now.
* `title` - optional title of recurrence series.  Any newly created meetings should be prefixed with this title.  Should just be `Async Standup` until we enable changing this in the UI.
* `recurrenceRule` - the rule that defines how this meeting series recurs.  This should simply be the result of serialization from the rrule library (see "recurrence format")
* `duration` - the duration (in minutes)
* `cancelledAt` - When recurrence for this series was cancelled.  Null if series is active.  Can be reset to null if reactivated.

### `NewMeeting` rethink table
New field:
* `meetingSeriesId` - The ID of the meeting series if this meeting is associated with one
* `scheduledEndTime` - when the meeting is scheduled to end

## Recurrence format
We should use the [RFC-2445 iCal spec](https://www.kanzaki.com/docs/ical/rrule.html), though we won't use all of its features.  The [rrule library](https://github.com/jakubroztocil/rrule) supports this spec, and supports date generation and parsing+serialization.

For MVP, we should hard-code the recurrence to use the following rule when creating meeting series:
* `freq` - weekly
* `dtstart` - 9am at user's time zone _or_ 12a EST
* `byweekday` - Monday, Tuesday, Wednesday, Thursday, Friday

Note that this does not include enough information to determine duration - let's assume all meetings end 24 hours after they start for the MVP, then we can add duration (including the duration column) as a follow-up.

## `processRecurrence` mutation
This mutation does two things in the MVP:
* Close meetings that are scheduled to close
* Create new meetings that should be started based on a meeting series' recurrence rule

Closing meetings is simple - find any open meetings where `now` is after their `scheduledEndTime`, and close them.

Creating new meetings is a bit more complicated:
[...]

PR #7041, opened 2022-08-10
feat(standups): Recurrence - `processRecurrence` mutation

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/6917

(Half of the diff is tests)

Implement `processRecurrence`, which starts and ends team prompt meetings according to their `scheduledEndTime`s and meeting series.  Can be executed in the private GQL schema as:
```
mutation {
  processRecurrence {
    ... on ProcessRecurrenceSuccess {
      meetingsStarted
      meetingsEnded
    }
  }
}
```

## Demo

See processRecurrence.test.ts

## Testing scenarios

(Reviewers - don't feel obligated to manually do these; these testing scenarios are also included in the unit tests)

- [ ] Start a team prompt meeting
- [ ] Set it to recur
- [ ] Run `processRecurrence`, and confirm that no meetings are started or ended
- [ ] Update the team prompt meeting to have a `scheduledEndTime` in the past via SQL
- [ ] Run `processRecurrence` and confirm that one meeting is ended, and that the team prompt meeting now has a non-null `endedAt`
- [ ] Run `processRecurrence` and confirm that no meetings are started or ended (idempotency).
- [ ] Update the team prompt meeting to have a `createdAt` 24 hours earlier via SQL
- [ ] Run `processRecurrence` and confirm that one meeting is started, and that a new team prompt meeting appears for the team.
- [ ] Run `processRecurrence` and confirm that no meetings are started or ended (idempotency).

## Final checklist

### DIFF
--- a/packages/server/__tests__/processRecurrence.test.ts
+++ b/packages/server/__tests__/processRecurrence.test.ts
@@ -54,12 +54,18 @@ const assertIdempotency = async () => {
   })
 }
 
-test('Should not end meetings that are not scheduled to end', async () => {
+// :TODO: (jmtaber129): Handle cleanup better
+
+beforeEach(async () => {
+  // Process recurrence right before each test to prevent pending effects (i.e. meetings that will
+  // start or end on the next processRecurrence run) from interfering with test results.
   await sendIntranet({
     query: PROCESS_RECURRENCE,
     isPrivate: true
   })
+})
 
+test('Should not end meetings that are not scheduled to end', async () => {
   const r = await getRethink()
   const {userId} = await signUp()
   const {id: teamId} = (await getUserTeams(userId))[0]
@@ -97,11 +103,6 @@ test('Should not end meetings that are not scheduled to end', async () => {
 })
 
 test('Should not end meetings that are scheduled to end in the future', async () => {
-  await sendIntranet({
-    query: PROCESS_RECURRENCE,
-    isPrivate: true
-  })
-
   const r = await getRethink()
   const {userId} = await signUp()
   const {id: teamId} = (await getUserTeams(userId))[0]
@@ -142,11 +143,6 @@ test('Should not end meetings that are scheduled to end in the future', async ()
 })
 
 test('Should end meetings that are scheduled to end in the past', async () => {
-  await sendIntranet({
-    query: PROCESS_RECURRENCE,
-    isPrivate: true
-  })
-
   const r = await getRethink()
   const {userId} = await signUp()
   const {id: teamId} = (await getUserTeams(userId))[0]
@@ -185,11 +181,6 @@ test('Should end meetings that are scheduled to end in the past', async () => {
 })
 
 test('Should end the current meeting and start a new meeting', async () => {
-  await sendIntranet({
-    query: PROCESS_RECURRENCE,
-    isPrivate: true
-  })
-
   const r = await getRethink()
   const {userId} = await signUp()
   const {id: teamId} = (await getUserTeams(userId))[0]
@@ -207,7 +198,7 @@ test('Should end the current meeting and start a new meeting', async () => {
 
   const newMeetingSeriesId = await insertMeetingSeriesQuery({
     meetingType: 'teamPrompt',
-    title: 'Async Standup',
+    title: 'Daily Test Standup',
     recurrenceRule: recurrenceRule.toString(),
     duration: 24 * 60, // 24 hours
     teamId,
@@ -242,6 +233,81 @@ test('Should end the current meeting and start a new meeting', async () => {
       processRecurrence: {
         meetingsStarted: 1,
         meetingsEnded: 1
+      }
+    }
+  })
+
+  await assertIdempotency()
+
+  const actualMeeting = await r.table('NewMeeting').get(meetingId).run()
+  expect(actualMeeting.endedAt).toBeTruthy()
+
+  const lastMeeting = await r
+    .table('NewMeeting')
+    .filter({meetingType: 'teamPrompt', meetingSeriesId: newMeetingSeriesId})
+    .orderBy(r.desc('createdAt'))
+    .nth(0)
+    .run()
+
+  expect(lastMeeting).toMatchObject({
+    name: expect.stringMatching(/Daily Test Standup.*/)
+  })
+})
+
+test('Should only start a new meeting if it would still be active', async () => {
+  const r = await getRethink()
+  const {userId} = await signUp()
+  const {id: teamId} = (await getUserTeams(userId))[0]
+
+  const now = new Date()
+
+  // Create a meeting series that's been going on for a few days, and happens daily at 9a UTC.
+  const startDate = new Date(
+    Date.UTC(now.getUTCFullYear(), now.getUTCMonth(), now.getUTCDate() - 5, 9)
+  )
+  const recurrenceRule = new RRule({
+    freq: RRule.DAILY,
+    dtstart: startDate
+  })
+
+  const newMeetingSeriesId = await insertMeetingSeriesQuery({
+    meetingType: 'teamPrompt',
+    title: 'Async Standup',
+    recurrenceRule: recurrenceRule.toString(),
+    duration: 24 * 60, // 24 hours
+    teamId,
+    facilitatorId: userId
+  })
+
+  const meetingId = generateUID()
+  const meeting = new MeetingTeamPrompt({
+    id: meetingId,
+    teamId,
+    meetingCount: 0,
+    phases: [new TeamPromptResponsesPhase(['foobar'])],
+    facilitatorUserId: userId,
+    meetingPrompt: 'What are you working on today? Stuck on anything?',
+    scheduledEndTime: new Date(Date.now() - ms('73h')),
+    meetingSeriesId: newMeetingSeriesId
+  })
+
+  // The last meeting in the series was created just over 72h ago, so 3 meetings should have started
+  // since then, but only 1 meeting should start as a result of the mutation.
+  meeting.createdAt = new Date(meeting.createdAt.getTime() - ms('73h'))
+  meeting.endedAt = new Date(Date.now() - ms('49h'))
+
+  await r.table('NewMeeting').insert(meeting).run()
+
+  const update = await sendIntranet({
+    query: PROCESS_RECURRENCE,
+    isPrivate: true
+  })
+
+  expect(update).toEqual({
+    data: {
+      processRecurrence: {
+        meetingsStarted: 1,
+        meetingsEnded: 0
       }
     }
   })
--- a/packages/server/__tests__/processRecurrence.test.ts
+++ b/packages/server/__tests__/processRecurrence.test.ts
@@ -1,0 +1,253 @@
+import ms from 'ms'
+import {RRule} from 'rrule'
+import getRethink from '../database/rethinkDriver'
+import MeetingTeamPrompt from '../database/types/MeetingTeamPrompt'
+import TeamPromptResponsesPhase from '../database/types/TeamPromptResponsesPhase'
+import generateUID from '../generateUID'
+import {insertMeetingSeries as insertMeetingSeriesQuery} from '../postgres/queries/insertMeetingSeries'
+import {getUserTeams, sendIntranet, signUp} from './common'
+
+const PROCESS_RECURRENCE = `
+  mutation {
+    processRecurrence {
+      ... on ProcessRecurrenceSuccess {
+        meetingsStarted
+        meetingsEnded
+      }
+      ... on ErrorPayload {
+        error {
+          message
+        }
+      }
+    }
+  }
+`
+
+const assertIdempotency = async () => {
+  // Assert that subsequent mutations do nothing.
+  let res = await sendIntranet({
+    query: PROCESS_RECURRENCE,
+    isPrivate: true
+  })
+
+  expect(res).toEqual({
+    data: {
+      processRecurrence: {
+        meetingsStarted: 0,
+        meetingsEnded: 0
+      }
+    }
+  })
+
+  res = await sendIntranet({
+    query: PROCESS_RECURRENCE,
+    isPrivate: true
+  })
+
+  expect(res).toEqual({
+    data: {
+      processRecurrence: {
+        meetingsStarted: 0,
+        meetingsEnded: 0
+      }
+    }
+  })
+}
+
+test('Should not end meetings that are not scheduled to end', async () => {
+  await sendIntranet({
+    query: PROCESS_RECURRENCE,
+    isPrivate: true
+  })
+
+  const r = await getRethink()
+  const {userId} = await signUp()
+  const {id: teamId} = (await getUserTeams(userId))[0]
+
+  const meetingId = generateUID()
+  const meeting = new MeetingTeamPrompt({
+    id: meetingId,
+    teamId,
+    meetingCount: 0,
+    phases: [new TeamPromptResponsesPhase(['foobar'])],
+    facilitatorUserId: userId,
+    meetingPrompt: 'What are you working on today? Stuck on anything?'
+  })
+
+  await r.table('NewMeeting').insert(meeting).run()
+
+  const update = await sendIntranet({
+    query: PROCESS_RECURRENCE,
+    isPrivate: true
+  })
+
+  expect(update).toEqual({
+    data: {
+      processRecurrence: {
+        meetingsStarted: 0,
+        meetingsEnded: 0
+      }
+    }
+  })
+
+  await assertIdempotency()
+
+  const actualMeeting = await r.table('NewMeeting').get(meetingId).run()
+  expect(actualMeeting.endedAt).toBeFalsy()
+})
+
+test('Should not end meetings that are scheduled to end in the future', async () => {
+  await sendIntranet({
+    query: PROCESS_RECURRENCE,
+    isPrivate: true
+  })
+
+  const r = await getRethink()
+  const {userId} = await signUp()
+  const {id: teamId} = (await getUserTeams(userId))[0]
+
+  const meetingId = generateUID()
+  const meeting = new MeetingTeamPrompt({
+    id: meetingId,
+    teamId,
+    meetingCount: 0,
+    phases: [new TeamPromptResponsesPhase(['foobar'])],
+    facilitatorUserId: userId,
+    meetingPrompt: 'What are you working on today? Stuck on anything?',
+    scheduledEndTime: new Date(Date.now() + ms('5m'))
+  })
+
+  await r.table('NewMeeting').insert(meeting).run()
+
+  const update = await sendIntranet({
+    query: PROCESS_RECURRENCE,
+    isPrivate: true
+  })
+
+  expect(update).toEqual({
+    data: {
+      processRecurrence: {
+        meetingsStarted: 0,
+        meetingsEnded: 0
+      }
+    }
+  })
+
+  await assertIdempotency()
+
+  const actualMeeting = await r.table('NewMeeting').get(meetingId).run()
+  expect(actualMeeting.endedAt).toBeFalsy()
+
+  await r.table('NewMeeting').get(meetingId).delete().run()
+})
+
+test('Should end meetings that are scheduled to end in the past', async () => {
+  await sendIntranet({
+    query: PROCESS_RECURRENCE,
+    isPrivate: true
+  })
+
+  const r = await getRethink()
+  const {userId} = await signUp()
+  const {id: teamId} = (await getUserTeams(userId))[0]
+
+  const meetingId = generateUID()
+  const meeting = new MeetingTeamPrompt({
+    id: meetingId,
+    teamId,
+    meetingCount: 0,
+    phases: [new TeamPromptResponsesPhase(['foobar'])],
+    facilitatorUserId: userId,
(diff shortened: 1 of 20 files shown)


packages/server/postgres/migrations/1659973034251_alterMeetingSeries.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+    DELETE FROM "MeetingSeries"; --Intentionally without 'WHERE' clause.
+
+    ALTER TABLE "MeetingSeries"
+      ADD COLUMN IF NOT EXISTS "teamId" VARCHAR(100) NOT NULL,
+      ADD COLUMN IF NOT EXISTS "facilitatorId" VARCHAR(100) NOT NULL;
```
jmtaber129, line 12: These columns need to be non-null for the meeting series to be usable, and it's non-trivial to set these with data on the rethink meetings.

Since this is a WIP feature, I've decided to simply delete all rows currently in any meeting series table, instead of investing time into making this migration "clean", but I'm very open to ways to improve this.

packages/server/utils/analytics/analytics.ts
```diff
@@ -107,8 +107,16 @@ class Analytics {
...
-  meetingStarted = (userId: string, meeting: Meeting, template?: MeetingTemplate) => {
-    this.track(userId, 'Meeting Started', createMeetingProperties(meeting, undefined, template))
+  meetingStarted = (
+    userId: string,
+    meeting: Meeting,
+    template?: MeetingTemplate,
+    isRecurring?: boolean
+  ) => {
+    this.track(userId, 'Meeting Started', {
+      isRecurring: isRecurring,
```
tianrunhe, line 117: I think `isRecurring` can be a common meeting properties (with default value `False` maybe):
1. I can see we'd also like to have that property for `Meeting Completed` events
2. Suppose in the future we can reuse this recurrence for other meeting types: "create a weekly retro meeting", won't that be cool?

packages/server/graphql/public/mutations/startRecurrence.ts
```diff
@@ -54,7 +54,9 @@ const startRecurrence: MutationResolvers['startRecurrence'] = async (
     meetingType: 'teamPrompt',
     title: 'Async Standup',
     recurrenceRule: recurrenceRule.toString(),
-    duration: 24 * 60 // 24 hours
+    duration: 24 * 60, // 24 hours
+    teamId,
+    facilitatorId: viewerId
   })
```
tianrunhe, line 60: I think we need a separate "Meeting Recurrence Started" event here. Would be interesting to see who always setup the cadence for the team, the team lead? billing lead?

tianrunhe (changes requested):

packages/server/utils/analytics/analytics.ts
jmtaber129: This is true, but this would effectively just be passing `isRecurring` into `createMeetingProperties`, just for `createMeetingProperties` to add it as a property, since we can't reliably determine whether a meeting is recurring just by the `meeting` object

1. I think `isScheduled` may be more representative, and also more future-proof, since it's less about the fact that the meeting's recurring and more that it was ended automatically. WDYT?
2. I hope so!

packages/server/graphql/public/mutations/startRecurrence.ts
jmtaber129: Added!

packages/server/utils/analytics/analytics.ts
tianrunhe: > since we can't reliably determine whether a meeting is recurring just by the meeting object

Can you check if `meetingSeriesId` exists in the Meeting object?

Now I think about it more, I feel like instead of a boolean `isRecurring` or `isScheduled`, we should just inject `meetingSeriesId` as a common meeting property. It acts like a foreign key to the MeetingSeries table, just like `teamId`.

packages/server/utils/analytics/analytics.ts
```diff
@@ -107,8 +108,22 @@ class Analytics {
...
+      ...createMeetingProperties(meeting, undefined, template)
+    })
+  }
+
+  recurrenceStarted = (userId: string, meeting: Meeting) => {
+    // :TODO: (jmtaber129): Add properties related to recurrence settings (duration, frequency,
+    // etc.) after we support configuring those.
+    this.track(userId, 'Meeting Recurrence Started', createMeetingProperties(meeting))
```
tianrunhe, line 126: To my point above (if you agree with me), here instead of the properties for the first meeting, we should log properties for the Meeting Series table. As v1, let's have:
1. `meetingSeriesId` so we can join against meeting events
2. `teamId`
3. `meetingType`

tianrunhe (changes requested):

packages/server/__tests__/processRecurrence.test.ts
```diff
@@ -0,0 +1,383 @@
...
+      ... on ErrorPayload {
+        error {
+          message
+        }
+      }
+    }
+  }
+`
+
+const assertIdempotency = async () => {
```
BartoszJarocki, line 26: +2 huge props for covering such an important part of the functionality in tests

packages/server/graphql/mutations/helpers/createTeamPrompt.ts
```diff
@@ -0,0 +1,34 @@
...
+  )
+  return new MeetingTeamPrompt({
+    id: meetingId,
+    teamId,
+    meetingCount,
+    phases,
+    facilitatorUserId: facilitatorId,
+    meetingPrompt: 'What are you working on today? Stuck on anything?' // :TODO: (jmtaber129): Get this from meeting settings.
+  })
+}
```
BartoszJarocki, line 32: +1 instead of having a bunch of helper methods, I think it'd be nice to have one class `TeamPromptMeetingActions` or something like that, with static methods. it'd allow having a single place for every action related to this meeting.

For example
```
class TeamPromptMeetingActions {
   static start(...) {}
   static startRecurring(rrule) {}
   static end(...) {} 
}
```

Then, we'd just call `TeamPromptMeetingActions.start(...)` 

it doesn't necessarily have to be all static, if there are parts that'd require calling multiple actions, we could instantiate this class and make it easier to call different methods. not a big deal, it works just fine as is, but I think this would be a bit clearer.

packages/server/graphql/private/mutations/processRecurrence.ts
```diff
@@ -0,0 +1,119 @@
...
+      const res = await endTeamPrompt({meeting, now, dataLoader, r, subOptions})
+      if (!('error' in res)) meetingsEnded++
+    })
+  )
+
+  let meetingsStarted = 0
+
+  // For each active meeting series, get the meeting start times (according to rrule) after the most
+  // recent meeting start time and before now.
+  const activeMeetingSeries = await getActiveMeetingSeriesQuery.run(undefined, getPg())
```
BartoszJarocki, line 83: +1 as a good practice, I'd still create a wrapper method for this just in case we'd need to map results and prevent direct use of `getPg` in mutations (just like for other queries)

packages/server/graphql/private/mutations/processRecurrence.ts
```diff
@@ -0,0 +1,119 @@
...
+  for (const meetingSeries of activeMeetingSeries) {
+    const lastMeeting = await r
+      .table('NewMeeting')
+      .filter({meetingType: 'teamPrompt', meetingSeriesId: meetingSeries.id})
```
BartoszJarocki, line 87: +1 do we need to specify `meetingType` here?

packages/server/graphql/private/mutations/processRecurrence.ts
```diff
@@ -0,0 +1,119 @@
...
+  startTime: Date,
+  dataLoader,
+  r,
+  subOptions
+) => {
+  const {teamId, facilitatorId} = meetingSeries
+
+  // AUTH
+  const unpaidError = await isStartMeetingLocked(teamId, dataLoader)
+  if (unpaidError) return standardError(new Error(unpaidError), {userId: facilitatorId})
```
BartoszJarocki, line 29: +1 this will be called for every meeting that's about to start, and it'll hit the DB every time, do we want to check if the team is clocked only once?

packages/server/postgres/migrations/1659973034251_alterMeetingSeries.ts
BartoszJarocki: Recurrence is behind the feature flag, I'd rather delete some rows by hand (if any) than have hard delete in the mutation 💀 (I checked, both staging and prod are empty)

BartoszJarocki (approved):
great work on a hard task. left some comments to consider, otherwise lgtm!

packages/server/graphql/mutations/helpers/createTeamPrompt.ts
jmtaber129: Hmm is this common in other parts of the codebase?  I'm not opposed to making this a little more object-oriented, but AFAIK most of our code is pretty functional, and I don't totally understand the benefits to doing it this way beyond putting things in the same file.

packages/server/graphql/private/mutations/processRecurrence.ts
jmtaber129: Not necessarily, but this seemed like a reasonable guardrail against unsupported meeting types. E.g. if a meeting in the series is somehow a retro, we probably don't want to mess with that.

packages/server/graphql/private/mutations/processRecurrence.ts
jmtaber129: So as it turns out, `isStartMeetingLocked` actually uses the `dataLoader` here, so this should only hit the DB once per teamId.

packages/server/utils/analytics/analytics.ts
jmtaber129: So since this is effectively just a normal meeting event (like meeting started, meeting joined etc.), we'll get `teamId` and `meetingType` as part of `createMeetingProperties`, and I've also just added `meetingSeriesId` to that helper as well.

packages/server/utils/analytics/analytics.ts
jmtaber129: Oops, you're right. Normally, we can't determine whether a meeting is actively recurring based on it's `meetingSeriesId`, but if we're starting a meeting that has a meetingSeriesId, then it's definitely being started as a result of recurrence.

Went ahead and added `meetingSeriesId` to the common meeting properties.

packages/server/postgres/migrations/1659973034251_alterMeetingSeries.ts
jmtaber129: Fair enough! Removed the deletion

packages/server/graphql/mutations/helpers/createTeamPrompt.ts
BartoszJarocki: then don't do this. this is just my personal preference not to put every function in a separate file, especially when they're related. navigating that kind of code requires less context switching and makes it easier to spot things that could be shared/simplified.

tianrunhe (approved):
Signing off from metrics POV. Great work!

mattkrick (changes requested):
lots of comments, but this is a VERY complex topic! let me know if I can help in any way

packages/server/graphql/private/mutations/processRecurrence.ts
BartoszJarocki: I've added a `meetingSeriesId` index here: https://github.com/ParabolInc/parabol/pull/7023/commits/958b34233af1d14ce67e9c47a66064c2861403de

packages/server/graphql/private/mutations/processRecurrence.ts
```diff
@@ -0,0 +1,118 @@
...
+      .run()
+
+    // For meetings that should still be active, start the meeting and set its end time.
+    // Any subscriptions are handled by the shared meeting start code
+    const rrule = RRule.fromString(meetingSeries.recurrenceRule)
+
+    // Only get meetings that should currently be active, i.e. meetings that should have started
+    // within the last 24 hours, started after the last meeting in the series, and started before
+    // 'now'.
+    const fromDate = new Date(
```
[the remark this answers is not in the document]
jmtaber129: Doing these updates as scheduled jobs would make meeting start times more exact, but it would also introduce more moving parts + complexity, so I have a strong preference for keeping all updates contained within this mutation for now, and introducing more precise scheduling if/when customer and product needs require it.  This is also what was agreed upon in the [spec](https://github.com/ParabolInc/parabol/issues/6888)

packages/server/graphql/private/mutations/processRecurrence.ts
```diff
@@ -0,0 +1,118 @@
...
+      Math.max(lastMeeting.createdAt.getTime() + ms('10m'), now.getTime() - ms('24h'))
+    )
+    const newMeetingsStartTimes = rrule.between(fromDate, now)
+    for (const startTime of newMeetingsStartTimes) {
```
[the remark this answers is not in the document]
jmtaber129: That's not a rule we decided on AFAIK.

packages/server/graphql/mutations/helpers/createTeamPrompt.ts
```diff
@@ -0,0 +1,34 @@
+import MeetingTeamPrompt from '../../../database/types/MeetingTeamPrompt'
+import generateUID from '../../../generateUID'
+import {MeetingTypeEnum} from '../../../postgres/types/Meeting'
+import createNewMeetingPhases from './createNewMeetingPhases'
+
+const createTeamPrompt = async (teamId, facilitatorId, r, dataLoader) => {
```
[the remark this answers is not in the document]
jmtaber129: Oops, totally thought I changed `startTeamPrompt` to use this helper 🤦

packages/server/graphql/mutations/helpers/endTeamPrompt.ts
```diff
@@ -0,0 +1,83 @@
...
+  const data = {
+    meetingId,
+    teamId,
+    timelineEventId
+  }
+  publish(SubscriptionChannel.TEAM, teamId, 'EndTeamPromptSuccess', data, subOptions)
+  return data
+}
+
+export default endTeamPrompt
```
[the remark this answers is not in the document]
jmtaber129: The intention here is that this simply has the effect of ending a team prompt meeting (either as a result of someone executing the `endTeamPrompt` mutation, or as an effect of some other mutation, as is the case with `processRecurrence`).  IMO the `publish` line is coupled less with the `endTeamPrompt` resolver and more with the effect of ending a team prompt meeting.  The return signature is a bit more coupled with the actual mutation, but I don't see this as a significant enough smell to put effort into refactoring as part of this PR.

packages/server/graphql/private/mutations/processRecurrence.ts
```diff
@@ -0,0 +1,118 @@
...
+  const operationId = dataLoader.share()
+  const subOptions = {mutatorId, operationId}
+
+  // VALIDATION
+
+  // RESOLUTION
+  // Find any meetings with a scheduledEndTime before now, and close them
+  const teamPromptMeetingsToEnd = (await r
+    .table('NewMeeting')
+    .filter({meetingType: 'teamPrompt'})
```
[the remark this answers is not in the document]
jmtaber129: Somewhat radical suggestion - we land this as is, then come back to fixing this as a very-near-term follow-up (like, within a week).  Reasoning:
* Recurrence is still behind its own feature flag
* This mutation isn't yet being run in prod unless a superuser executes it manually
* 6 seconds isn't _too_ bad for a non-interactive query
* The actual fix will be pretty self-contained (add the index via migration, then change this query to use the migration).
* Getting this landed will enable better end-to-end testing in dev.

LMK what you think - I'm usually pretty aggressive about minimizing scope creep once something's in review, but if you think this is important enough to take care of now, I'll do it as part of this PR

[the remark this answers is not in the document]
jmtaber129: On second thought, I just went ahead and figured out the index

jmtaber129:
@mattkrick I think I addressed everything except the `meetingSeriesId` index - I'll follow up with Bartosz on how to get that sorted with the index he added.  If everything besides that looks good, I'd like to get this landed as-is to avoid merge conflicts, excess WIP, etc., since this mutation doesn't get called in prod yet, and is limited to superusers

packages/server/graphql/private/mutations/processRecurrence.ts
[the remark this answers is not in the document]
mattkrick: Oh! I think I was going off an old draft doc: https://www.notion.so/Recurrence-draft-702693dab9394ca5ab1e66993cd9d91b?d=c168af56d33d48068d8513bb5bdfefc8#740426cb569048eaa1453ed862392be4

If 1 series can have multiple active instances would we show both on the /meetings route? For things like invite links, we just send folks to the most recently opened meeting?

packages/server/graphql/private/mutations/processRecurrence.ts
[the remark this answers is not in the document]
jmtaber129: That's partially my bad - the spec mentions how this is called in the overview but nowhere else. This mutation will be called periodically (once every ~5 minutes initially) by the chronos cron scheduler, so the actual start time should always be within +5 minutes of the scheduled start time regardless.

packages/server/graphql/private/mutations/processRecurrence.ts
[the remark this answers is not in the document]
mattkrick: usually, anything behind a feature flag i'm totally OK with. The exception is where a feature could impact the usability for features not under the feature flag. RethinkDB **_SHOULD_** be able to handle a long-running query while also processing a bunch of updates, they even say they do! https://rethinkdb.com/docs/architecture.

In practice, we've crashed the DB doing this, where queries to single tables hang :sweat_smile:. This is the top reason why I wanna get off of RethinkDB, it's just not stable at longrunning tasks. In the future & in PG, this is totally safe

packages/server/graphql/private/mutations/processRecurrence.ts
```diff
@@ -0,0 +1,124 @@
...
+
+  const res = await Promise.all(
+    teamPromptMeetingsToEnd.map(async (meeting) => {
+      return await safeEndTeamPrompt({meeting, now, dataLoader, r, subOptions})
+    })
+  )
+
+  const meetingsEnded = res.filter((res) => !('error' in res)).length
+
+  let meetingsStarted = 0
```
mattkrick, line 81: +1 would like to see this as `const` like up above

packages/server/database/rethinkDriver.ts
```diff
@@ -91,7 +91,7 @@ export type RethinkSchema = {
   }
   NewMeeting: {
     type: AnyMeeting
-    index: 'facilitatorUserId' | 'teamId' | 'templateId'
+    index: 'facilitatorUserId' | 'teamId' | 'templateId' | 'scheduledEndTime' | 'hasNotEnded'
```
mattkrick, line 94: -1 should be `hasEndedScheduledEndTime` (looks like the `between` method in stricterR isn't strict, so it didn't catch this one)

packages/server/graphql/private/mutations/processRecurrence.ts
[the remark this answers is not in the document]
jmtaber129: _Oof_, that's not good. Does rethink have a timeout-like thing, a la Postgres's `statement_timeout`? If a long-running query like this can DOS the DB, we should probably raise an error anytime the duration of a query crosses some threshold.

packages/server/graphql/private/mutations/processRecurrence.ts
jmtaber129: Changing this is less trivial compared to `meetingsEnded`, since `meetingsStarted` is updated in a nested loop, so getting this as a const would probably require a map-reduce-y thing, which I think would be less readable compared to incrementing a `let` var. WDYT?

packages/server/graphql/private/mutations/processRecurrence.ts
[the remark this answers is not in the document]
mattkrick: gotcha. I'm hesitant to introduce something that has a 5 min latency because everything else in our app (and all other modern apps) is real time. If the start time is early, it's no big deal, the only way they'd know is the timestamp on the slack message. If the facilitator starts the meeting at the time they want everyone to hop in, that might make them look bad as everyone sits around waiting for it to open, but like you say, future tension.

packages/server/graphql/private/mutations/processRecurrence.ts
mattkrick: sure, makes sense

packages/server/graphql/private/mutations/processRecurrence.ts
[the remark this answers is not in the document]
mattkrick: I don't think so... at least i haven't found it! the last couple times it was crashed with an ad-hoc query going via the explorer. It's hard to see which query causes the crash. there's a jobs table, but when the whole thing freezes, so does that table.

Issue #6828, opened 2022-06-24
Poker scope phase: Styling issue in header 


87prashant:
![Screenshot 2022-06-24 231227](https://user-images.githubusercontent.com/106697681/175614915-cca0ad76-e117-48fb-a3ae-3055780b187d.png)

Content of header (i.e Github, Jira, Parabol, Gitlab) is not responsive. I think a slider should be added. 
 
# AC
- [ ] Implement a Slider

jordanh:
Nice find. Good for @nickoferrall or @Dschoordsch to know if they are in this area adding a new integration

jordanh:
This section sill certainly need some whenever we add another integration into the mix

PR #7043, opened 2022-08-11
fix: Poker scope phase: Styling issue in header

gitstart:
# Description

Fixes/Partially Fixes #[6828](https://github.com/ParabolInc/parabol/issues/6828)

## Demo

[Loom video 1](https://www.loom.com/share/a2add52814c5458681cb9b1153bf6eca)
[Loom video 2](https://www.loom.com/share/d8690f4d2a094aa3b313ca250216dc34)
[Loom video 3](https://www.loom.com/share/4ce8142fd42f41dfb03495bffa5d5b72)

## Testing scenarios

- [x] Scenario A
  - Step 1
  - Step 2...

- [x] Scenario B
  - Step 1
  - Step 2....

## Final checklist


87prashant:
Thanks @gitstart  for working on this one
Could you also implement automatically scrolling to the active `tab` on swiping if it is hidden?

gitstart:

Dschoordsch (changes requested):
I think we need to introduce some scrolling in the header to solve this, it still looks off on mobile
Screenshot from 2022-08-19 16-45-22

gitstart:
@Dschoordsch the issue has been fixed. You can review

Dschoordsch (changes requested):
Thanks for your update and the videos. It looks better now. I still see a usability problem on mobile (e.g. iphone 11 resolution): It looks like there are only 3 tabs and there is no indication that the tab bar could be scrolled
No indication of more content
Best way to ensure the user sees that there is more content available would be to ensure that the next element is partially visible.
For the new meeting dialog we use a separate component which has a setting for this case https://github.com/ParabolInc/parabol/blob/3fb122180dfe5d49ff16f67f22fba410d1bc5134/packages/client/components/NewMeetingCarousel.tsx#L138-L159
I'm not sure how well it would work with a variable number of visible elements though, but it gives you some place to start.
If you can solve this issue without changing the component, that would also be fine.

gitstart:
Hi @Dschoordsch, your comments and appended another loom video for reference.

gitstart:
> Thanks for your update and the videos. It looks better now. I still see a usability problem on mobile (e.g. iphone 11 resolution): It looks like there are only 3 tabs and there is no indication that the tab bar could be scrolled No indication of more content [ No indication of more content ](https://user-images.githubusercontent.com/7331043/186629013-c4472b95-bbe9-453e-a5de-db75006a8933.gif) [ ](https://user-images.githubusercontent.com/7331043/186629013-c4472b95-bbe9-453e-a5de-db75006a8933.gif) Best way to ensure the user sees that there is more content available would be to ensure that the next element is partially visible. For the new meeting dialog we use a separate component which has a setting for this case
> 
> https://github.com/ParabolInc/parabol/blob/3fb122180dfe5d49ff16f67f22fba410d1bc5134/packages/client/components/NewMeetingCarousel.tsx#L138-L159
> 
> 
> I'm not sure how well it would work with a variable number of visible elements though, but it gives you some place to start.
> If you can solve this issue without changing the component, that would also be fine.

The suggested implementation wasn't possible, since this component does use `swiper-react`  but enforced the scroll-bar visibility on small sized screens.

packages/client/styles/theme/globalStyles.ts
```diff
@@ -122,23 +122,24 @@ export default `
     font-weight: 600;
   }
 
-  /*::-webkit-scrollbar {
+  .scope-phase-area::-webkit-scrollbar {
```
Dschoordsch, line 125: -1 Please don't modify the global style for local changes

packages/client/components/ScopePhaseArea.tsx
```diff
@@ -143,7 +145,7 @@ const ScopePhaseArea = (props: Props) => {
 
   return (
     <ScopingArea isDesktop={isDesktop}>
-      <StyledTabsBar activeIdx={activeIdx}>
+      <StyledTabsBar className='scope-phase-area' activeIdx={activeIdx}>
```
Dschoordsch, line 148: -1 instead of creating a global style, these settings should be added in the `StyledTabsBar` line 43-48

Dschoordsch (changes requested):

packages/client/components/ScopePhaseArea.tsx
gitstart: Hello @Dschoordsch, suggested changes have been performed you can take another look.

Dschoordsch (changes requested):
Looks good when testing with Chrome. On Firefox the scrollbar is not visible.

gitstart:
> Looks good when testing with Chrome. On Firefox the scrollbar is not visible.

Hi @Dschoordsch, spent some time investigating this issue though to no success yet, only made some clean ups to the code but the functionality still as previous,  issue is `scrollbar only shows up during the scroll effect on Mozilla`. you can further advice if this can stand or closed thanks.

Dschoordsch (approved):
Let's go with it then. It's an edge case anyways.

PR #7049, opened 2022-08-11
feat(standups): Recurrence -  restart recurrence in `startRecurrence` mutation

### LINKED ISSUE #6921: Recurrence MVP: Restart recurrence in `startRecurrence` mutation
Backend spec: https://github.com/ParabolInc/parabol/issues/6888

Blocked on https://github.com/ParabolInc/parabol/issues/6916

Add support for restarting recurrence via the `startRecurrence` mutation.

- [ ]  When `startRecurrence` is executed on a meeting with an inactive meeting series
  - [ ] The series `cancelled_at` is set to null
  - [ ] The `scheduledEndTime` of the current meeting (and any other open meetings in the series) are set to a reasonable time in the future
  - [ ] The next `processRecurrence` run doesn't create or end any meetings in this series

BartoszJarocki:
# Description

Fixes #6921 

## Demo

https://user-images.githubusercontent.com/1017620/184804894-c39bba99-fa44-41b9-9a61-f038c072fc06.mp4

## Testing scenarios

- [ ] Open standups, start recurrence, stop recurrence and then, start recurrence again

## Final checklist


packages/server/graphql/public/mutations/startRecurrence.ts
```diff
@@ -36,7 +37,27 @@ const startRecurrence: MutationResolvers['startRecurrence'] = async (
...
 
   if (meeting.meetingSeriesId) {
-    return standardError(new Error('Meeting already has meeting series'), {userId: viewerId})
+    const meetingSeries = await dataLoader.get('meetingSeries').loadNonNull(meeting.meetingSeriesId)
+    if (!meetingSeries.cancelledAt) {
+      return standardError(new Error('Meeting is already recurring'), {userId: viewerId})
+    }
+
+    // meeting has stopped recurrence associated with it, reenable it
+    await updateMeetingSeries({cancelledAt: null}, meeting.meetingSeriesId)
```
jmtaber129, line 46: -1 We should update the recurrence rule to make sure the next meeting starts at a reasonable time. E.g. if this meeting originally started a few days ago, then the recurrence rule would likely prescribe a meeting in the past 24 hours, so the next `processRecurrence` run would immediately start a new meeting, which would lead to two meetings being open for 24 hours.

I'd suggest the following to handle this case (we'll probably refine these semantics over time):
* If `cancelledAt` is in the past hour or so, or the next start time defined by in the recurrence rule is at least 9 hours (arbitrarily chosen) from now, don't change the recurrence rule
  * For bonus points, also update the `scheduledEndTime` according to duration + startTime
* If `cancelledAt` is more than an hour ago, and the next start time is within the next 9 hours, change the recurrence rule's start date to be 24 hours after that next start time.

This would be in-line with the current `startRecurrence` start date semantics, which choose the following day (UTC) to have the first meeting started.

Longer term, we'll probably build more of a UI for figuring out when a user wants their next meeting to start

packages/server/postgres/queries/src/updateMeetingSeriesByIdQuery.sql
```diff
@@ -7,5 +7,5 @@ UPDATE "MeetingSeries" SET
   "title" = COALESCE(:title, "title"),
   "recurrenceRule" = COALESCE(:recurrenceRule, "recurrenceRule"),
   "duration" = COALESCE(:duration, "duration"),
-  "cancelledAt" = COALESCE(:cancelledAt, "cancelledAt")
+  "cancelledAt" = :cancelledAt
```
jmtaber129, line 10: -1 Wouldn't this reset `cancelledAt` when updating other columns?

jmtaber129 (changes requested):
-1 As-is, in certain cases this will cause a new meeting to be created immediately upon the next `processRecurrence` execution. The exact semantics for when the next meeting starts doesn't need to be solid yet, but we should update the recurrence rule when the next meeting would otherwise start very soon (I'd suggest an arbitrary time period of the next 9 hours)

packages/server/graphql/public/mutations/startRecurrence.ts
BartoszJarocki: @jmtaber129 thank you for catching this! I think I've covered all the cases, do you mind taking a look? thank you!

packages/server/graphql/public/mutations/startRecurrence.ts
```diff
@@ -37,26 +40,76 @@ const startRecurrence: MutationResolvers['startRecurrence'] = async (
...
+        .get(meeting.id)
+        .update({
+          scheduledEndTime: new Date(
+            meeting.createdAt.getTime() + ms(`${MEETING_DURATION_IN_MINUTES}m`)
+          )
+        })
+        .run()
+    )
+    await Promise.all(updates)
+
```
BartoszJarocki, line 102: @tianrunhe do we want to send an event about restarting recurrence?

tianrunhe: Thanks for the mention! Yeah I think we need to send an event here but to start, we could just use the existing `Meeting Recurrence Started` event. That being said, can you also add `Meeting Recurrence Ended` in `stopRecurrence` mutation? My thinking is that with these two, we could dive deep into analysis of restarting behavior if we want to (because we can join them by `meetingSeriesId`) so no need to send a new `Meeting Recurrence Restarted` event. WDYT?

BartoszJarocki: sounds good to me! will do.

BartoszJarocki: only issue I see right now is the fact that we send meeting properties along with the `Meeting Recurrence Started`, like that: 
```
this.track(userId, 'Meeting Recurrence Started', createMeetingProperties(meeting))
```

It's technically possible to have many meetings opened a the same time related to the same meeting series, IMO we should send meeting series params, instead of meeting. For example, what meeting id should we send in the case below? IMO, it'd be nice to have that sorted out, before recurrence goes to prod and I'd suggest to send meeting series params there. 

```

    // update all the active meetings to end at proper time
    const activeMeetings = await r
      .table('NewMeeting')
      .getAll(id, {index: 'meetingSeriesId'})
      .filter({endedAt: null}, {default: true})
      .run()
    const updates = activeMeetings.map((meeting) =>
      r
        .table('NewMeeting')
        .get(meeting.id)
        .update({
          scheduledEndTime: new Date(
            meeting.createdAt.getTime() + ms(`${MEETING_DURATION_IN_MINUTES}m`)
          )
        })
        .run()
    )
    await Promise.all(updates)

    analytics.recurrenceStarted(viewerId, ??)
```

what do you think @tianrunhe @jmtaber129 ?

tianrunhe: Yeah good tension! I think initially James [just re-used the helper function](https://github.com/ParabolInc/parabol/pull/7041#discussion_r947443676) so we can easily get common properties between `Meeting` and `MeetingSeries` but I agree, in this case, we only want **1** `Meeting Series Started` event. In that event I don't really need to know the `meetingId`? As A) there are more than one, B) if I want to know which meetings are affected, I can join by `meetingSeriesId`.
Yeah so it feels like we should have treat `MeetingSeries` separately.

packages/server/utils/analytics/analytics.ts
```diff
@@ -112,10 +119,12 @@ class Analytics {
...
-  recurrenceStarted = (userId: string, meeting: Meeting) => {
-    // :TODO: (jmtaber129): Add properties related to recurrence settings (duration, frequency,
-    // etc.) after we support configuring those.
-    this.track(userId, 'Meeting Recurrence Started', createMeetingProperties(meeting))
+  recurrenceStarted = (userId: string, meetingSeries: MeetingSeriesAnalyticsProperties) => {
+    this.track(userId, 'Meeting Recurrence Started', meetingSeries)
+  }
+
+  recurrenceStopped = (userId: string, meetingSeries: MeetingSeriesAnalyticsProperties) => {
+    this.track(userId, 'Meeting Recurrence Stopped', meetingSeries)
```
BartoszJarocki, line 127: @tianrunhe do you mind taking a look if that looks good? thank you!

packages/server/graphql/public/mutations/startRecurrence.ts
```diff
@@ -37,37 +40,88 @@ const startRecurrence: MutationResolvers['startRecurrence'] = async (
...
+    const isNextScheduledStartDateReasonable =
+      nextScheduledStartDate.getTime() > now.getTime() + ms('9h')
+    if (hasRecentlyStopped || isNextScheduledStartDateReasonable) {
+      // no need to change the recurrence rule, next meeting starts in reasonable time
+      await restartMeetingSeries({cancelledAt: null}, meetingSeriesId)
+    } else {
+      // meeting series was stopped more than ~1h ago and the next scheduled start date is within the next ~9h
+      // to prevent a meeting from being scheduled too soon, we need to change the recurrence rule
+      const newRecurrenceRule = currentRRule.clone()
+      newRecurrenceRule.options.dtstart = new Date(nextMeetingStartDate.getTime() + ms('1d'))
```
jmtaber129, line 73: +1 I think you mean `nextScheduledStartDate` here? If the next meeting start is too soon, we want to just delay that start time by 24h.

packages/server/graphql/public/mutations/startRecurrence.ts
```diff
@@ -37,37 +40,88 @@ const startRecurrence: MutationResolvers['startRecurrence'] = async (
...
   )
   const recurrenceRule = new RRule({
     freq: RRule.WEEKLY,
     byweekday: [RRule.MO, RRule.TU, RRule.WE, RRule.TH, RRule.FR],
-    dtstart: startDate
+    dtstart: nextMeetingStartDate
   })
 
-  const newMeetingSeriesId = await insertMeetingSeriesQuery({
+  if (meeting.meetingSeriesId) {
```
jmtaber129, line 54: +1 This is kind of a long `if` block (nearly half of the entire resolver) - let's pull something like a `restartRecurrence` function out to make this a bit more readable

packages/server/graphql/public/mutations/startRecurrence.ts
```diff
@@ -37,37 +40,88 @@ const startRecurrence: MutationResolvers['startRecurrence'] = async (
...
-  const startDate = new Date(
+  const nextMeetingStartDate = new Date(
     Date.UTC(now.getUTCFullYear(), now.getUTCMonth(), now.getUTCDate() + 1, 9)
   )
   const recurrenceRule = new RRule({
     freq: RRule.WEEKLY,
     byweekday: [RRule.MO, RRule.TU, RRule.WE, RRule.TH, RRule.FR],
-    dtstart: startDate
+    dtstart: nextMeetingStartDate
   })
```
jmtaber129, line 52: +1 Assuming we change the restart logic to not use `nextMeetingStartDate`, this will only be used when initializing a meeting series, so let's move this down below the `if` block

packages/server/graphql/public/mutations/startRecurrence.ts
```diff
@@ -37,37 +40,88 @@ const startRecurrence: MutationResolvers['startRecurrence'] = async (
...
+      .getAll(meetingSeriesId, {index: 'meetingSeriesId'})
+      .filter({endedAt: null}, {default: true})
+      .run()
+    const updates = activeMeetings.map((meeting) =>
+      r
+        .table('NewMeeting')
+        .get(meeting.id)
+        .update({
+          scheduledEndTime: new Date(
+            meeting.createdAt.getTime() + ms(`${MEETING_DURATION_IN_MINUTES}m`)
```
jmtaber129, line 95: +1 This should use the `meetingSeries`' `duration`

packages/server/graphql/public/mutations/startRecurrence.ts
jmtaber129: @tianrunhe @BartoszJarocki IMO, since all recurrence-related mutation are triggered from a specific meeting, the user action here is "**restart recurrence for this meeting**" (even though the effect on the data model is "restart recurrence for this _series_"), so I think it makes sense to continue associating this event with the meeting passed to the mutation (at least for now).

Re: reusing the "Meeting Recurrence Started" event - unless there's an easy way to make a custom event in Amplitude on "first event for a distinct property value", we should track a flag to differentiate initial recurrence starts and recurrence restarts for analysis within Amplitude. E.g. I might want to investigate questions like "what % of recurrence starts are immediately cancelled, broken down by whether it's an initial start or restart".

Later on, we'll probably change our user-facing semantics to be closer to "restart recurrence for meeting _series_", but we can cross that bridge re: analytics when we get to it.

jmtaber129 (approved):
A few minor requests, but nothing that would complicated enough to need another round of review from me (except for maybe the analytics discussion).  Nice work!

tianrunhe (approved):
I think this LGTM as a MVP from analytics perspective. I'd like to avoid feature creep. Niche/Advanced use cases can always be addressed in BQ if one is really interested, or we could refactor later.

packages/server/graphql/public/mutations/startRecurrence.ts
BartoszJarocki: yes, good catch!

packages/server/graphql/public/mutations/startRecurrence.ts
BartoszJarocki: good catch!

packages/server/graphql/public/mutations/startRecurrence.ts
BartoszJarocki: @jmtaber129 I understand the arguments, but IMO it's getting harder and harder to figure out what `meetingId` to pass in certain cases. I'd like to have this PR merged as is and let's create a separate task for analytics if needed.

BartoszJarocki:
@Dschoordsch do you mind taking a look? thanks!

packages/server/graphql/public/mutations/startRecurrence.ts
```diff
@@ -2,14 +2,63 @@ import ms from 'ms'
...
+    .table('NewMeeting')
+    .getAll(meetingSeriesId, {index: 'meetingSeriesId'})
+    .filter({endedAt: null}, {default: true})
+    .run()
+  const updates = activeMeetings.map((meeting) =>
+    r
+      .table('NewMeeting')
+      .get(meeting.id)
+      .update({
+        scheduledEndTime: new Date(meeting.createdAt.getTime() + ms(`${meetingSeries.duration}m`))
```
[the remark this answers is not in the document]
BartoszJarocki: @Dschoordsch good find! I'm not sure if I understand correctly what your solution should look like. you're suggesting using the last meeting and letting it end tomorrow, what if there are other active meetings? what should we do with them? if that'd help, we can meet and discuss it. 

btw, this PR suffered from a huge scope creep already, so I'd like to do something that is acceptable, not perfect for now. recurrence is under the feature flag anyway.

packages/server/postgres/queries/src/restartMeetingSeriesByIdQuery.sql
```diff
@@ -0,0 +1,7 @@
+/*
+  @name restartMeetingSeriesByIdQuery
+*/
+UPDATE "MeetingSeries" SET
+  "recurrenceRule" = COALESCE(:recurrenceRule, "recurrenceRule"),
+  "cancelledAt" = :cancelledAt
```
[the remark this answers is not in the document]
BartoszJarocki: oh damn. fixed!

packages/server/graphql/public/mutations/startRecurrence.ts
```diff
@@ -2,14 +2,63 @@ import ms from 'ms'
...
+  if (!cancelledAt) {
+    return
+  }
+
+  const now = new Date()
+  const currentRRule = RRule.fromString(recurrenceRule)
+  const hasRecentlyStopped = cancelledAt.getTime() > now.getTime() - ms('1h')
+  const nextScheduledStartDate = currentRRule.after(now)
+  const isNextScheduledStartDateReasonable =
+    nextScheduledStartDate.getTime() > now.getTime() + ms('9h')
```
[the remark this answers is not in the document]
BartoszJarocki: the idea for this part of the code was to _not_ modify the recurrence rule when there isn't too much of a time difference between dates set by the previous `rrule`. i agree, it is confusing and if there's a better way, would love to do it!

packages/server/graphql/public/mutations/startRecurrence.ts
```diff
@@ -38,46 +87,51 @@ const startRecurrence: MutationResolvers['startRecurrence'] = async (
...
-    facilitatorId: viewerId
-  })
-
-  const updatedMeeting = (await r
-    .table('NewMeeting')
-    .get(meetingId)
-    .update(
-      {
+    analytics.recurrenceStarted(viewerId, meetingSeries)
+  } else {
```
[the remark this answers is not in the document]
BartoszJarocki: done!

packages/server/graphql/public/mutations/startRecurrence.ts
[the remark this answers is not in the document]
BartoszJarocki: @jmtaber129 @Dschoordsch summary of our call:

When restarting recurrence
1. Schedule the next meeting start date to tomorrow at 9 AM (just like when starting). 
2. Set all active meetings to end tomorrow at 9 AM (the same time when a new meeting will be created). 

It's easy enough to describe in one sentence and should work just fine for the time being. Once we have more complex use cases to handle, we can revisit.

BartoszJarocki:
@Dschoordsch would you mind taking another look? thanks!

packages/server/graphql/public/mutations/startRecurrence.ts
```diff
@@ -2,22 +2,98 @@ import ms from 'ms'
...
+}
+
+const restartExistingMeetingSeries = async (meetingSeries: MeetingSeries) => {
+  const r = await getRethink()
+  const {cancelledAt, recurrenceRule, id: meetingSeriesId} = meetingSeries
+  if (!cancelledAt) {
+    return
+  }
+
+  // to keep things simple, we'll just restart the meeting series at arbitrarty time, 9 AM tomorrow
```
Dschoordsch, line 65: +1 Comment would be better `// to keep things simple, restart the meeting series at the same time a new series would start`

To future proof it we could even use the `currentRRule` to calculate which time it should start tomorrow
```ts
  const nextMeetingStartDate = currentRRule.after(new Date(Date.UTC(now.getUTCFullYear(), now.getUTCMonth(), now.getUTCDate() + 1, 0)))
```

Dschoordsch (approved):

packages/server/graphql/public/mutations/startRecurrence.ts
BartoszJarocki: good idea! fixed comment and added calculating date using `currentRRule`

Issue #7026, opened 2022-08-08
Implement create new issue from Azure Poker meeting


nickoferrall:
In the Azure DevOps Poker integration, the "New User Story" button is displayed when there are no issues available. Clicking it just toggles the `isEditing` boolean, and doesn't let the user add a new issue. 

![azure](https://user-images.githubusercontent.com/39854876/183393999-aefdda38-a3f1-4c41-bb2c-a0b569f7a670.png)
 
When issues are rendered, the "New User Story" button isn't displayed.

Also, in Parabol, the user should only be able to create an issue, not any other type of user story? So I think the button should say: "New issue".

### AC

- [ ] The "New User Story" fab text is changed to "New Issue"
- [ ] The "New Issue" fab is displayed when issues are rendered and when issues aren't rendered
- [ ] By clicking the fab, selecting a project, and entering an issue title, the user can successfully create a new issue in Azure

Issue #7051, opened 2022-08-11
Show desired subtitle in Azure sidebar item


nickoferrall:
After adding an issue to the Estimate phase, the subtitle briefly shows an id rather than the issue number: 

![issue-id](https://user-images.githubusercontent.com/39854876/184170972-48ed3ca7-d4d5-4735-8296-d8f9cdb3dd47.png)

### AC

- Fix optimistic updater so that the Azure Poker sidebar item shows the desired subtitle

nickoferrall:
I think this should still be fixed

Issue #7048, opened 2022-08-11
Change Azure item link to project name


nickoferrall:
In Azure, every item link shows the type rather than the project:

![results-type](https://user-images.githubusercontent.com/39854876/184151610-0f698dd0-ec90-4292-8ac4-29dfff1740b3.png)

Let's change the type to the project so it looks more like the other integrations:

![gh-links](https://user-images.githubusercontent.com/39854876/184152334-2e36d703-4f1b-4374-9345-e338e0150243.png)

### AC

- Change Azure results item link to the project name

Issue #7047, opened 2022-08-11
Azure DevOps timeout error


nickoferrall:
If there are dozens of Azure DevOps results, after a few refreshes, no results are returned and the server returns the following error: `Error: Azure DevOps is down`.

### AC

- Render dozens of Azure work items in the Scope stage, refresh the page a few times, add and remove work items, and see the expected work items
- It looks like the `workItems` query is failing, most probably rate limited by Azure. See if we can make this query more efficient, e.g. making using of the first argument here https://github.com/ParabolInc/parabol/blob/7938dd78615230f2a6501594813a43722004e06c/packages/server/graphql/types/AzureDevOpsIntegration.ts#L109-L111

mattkrick:
I'd suggest caching work items in redis & use a cache-and-network strategy to return the result from our redis cache if it exists & then when the result comes back from azure, we can push that value to redis & send the update to the client if there's a delta.

https://github.com/ParabolInc/parabol/blob/072cdfcbe5cf1a5eed3a9c72ec1da51756d2b9fe/packages/server/utils/atlassian/jiraIssues.ts

PR #7053, opened 2022-08-11
feat: create azure issue

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7026

https://www.loom.com/share/f4a4a7a97e3c4397a5373782cfa230f0

### To test

- Integrate with Azure DevOps. You may need to run `yarn postdeploy` or add the feature flag if it's not an option in your integration table.
- Go to the Scope phase of a Sprint Poker meeting
- Create an issue using the fab
- Can successfully change the project that the issue is sent to 
- The issue shows up at the top of the results
- Can click on the URL in the newly issue and get taken to that issue

Note: 

- The link in the Azure item says "Basic:Issue" rather than the project name. I suggest we change this and have created an issue: https://github.com/ParabolInc/parabol/issues/7048
- If you create many issues and refresh the page a few times, the integration can timeout. I created an issue for this: https://github.com/ParabolInc/parabol/issues/7047
- After ~24 hours, I need to manually refresh the Azure token to see any results: https://github.com/ParabolInc/parabol/issues/7006
- After selecting an issue in the Scope phase or creating a new issue, the subtitle in the newly created item is an id instead of the issue number. Issue here: https://github.com/ParabolInc/parabol/issues/7051


packages/client/components/NewAzureIssueInput.tsx
```diff
@@ -0,0 +1,249 @@
...
+
+interface Props {
+  isEditing: boolean
+  meetingId: string
+  setIsEditing: (isEditing: boolean) => void
+  viewerRef: NewAzureIssueInput_viewer$key
+}
+
+const validateIssue = (issue: string) => {
+  return new Legitity(issue).trim().min(2, `C’mon, you call that an issue?`)
```
igorlesnenko, line 106: +1 The error message feels not obvious :)

packages/client/components/NewAzureIssueInput.tsx
```diff
@@ -0,0 +1,249 @@
...
+            }
+          }
+        }
+      }
+    `,
+    viewerRef
+  )
+  const {id: userId, team, teamMember} = viewer
+  const {id: teamId} = team!
+  const projects = teamMember?.integrations?.azureDevOps.projects || []
```
igorlesnenko, line 136: +1 Using `??` is preferable

packages/client/components/NewAzureIssueInput.tsx
```diff
@@ -0,0 +1,249 @@
...
+    const selectedProject = projects.find((project) => project.name === selectedProjectName)!
+    const serviceProjectHash = AzureDevOpsProjectId.join(
+      selectedProject.instanceId,
+      selectedProject.id
+    )
+    const newTask = {
+      teamId,
+      userId,
+      meetingId,
+      content: convertToTaskContent(`${newIssueTitle} #archived`),
```
igorlesnenko, line 182: Is it expected behaviour? We always creating archived tasks in our board? Do you know the reason behind this?

nickoferrall: By creating an archived task, the `integration` is stored in the `task` and can be used to insert it into the Relay connection:
https://github.com/ParabolInc/parabol/blob/8e96045b57741440e39cd19cd8d38883fcbc629b/packages/client/mutations/handlers/handleGitHubCreateIssue.ts#L11

Also, as we already had this logic built for the task integration, I'm guessing it was easier to use the existing logic in `CreateTaskMutation` than create a new mutation (or update `UpdatePokerScopeMutation`) that adds the integration into the connection.

Does that sound right, @mattkrick?

packages/client/components/NewAzureIssueInput.tsx
nickoferrall: That's a fair point. For me, I prefer using apps that feel playful, and I think informal language helps with that. It's also consistent with other parts of the app - we have 9 uses of the word "C'mon" 😄 

We could allow any length, but it's unlikely that the user meant to create an issue if they haven't hit the two-character requirement, so I think it's a better UX to give them that feedback

packages/client/components/NewAzureIssueInput.tsx
mattkrick: working as intended? yes
correct? no :disappointed: 
the problem was folks would run a poker meeting, import a bunch of tasks & then when they finish the meeting, those tasks would be on their kanban board. they didn't like this, and I can't blame them, they pointed "stories" but in our system, everything is a "Task". Then again, some users like having stories show up on their task board! How do we make both sides happy?
If someone would like to tackle this problem, it'd give me great joy

mattkrick (approved):

packages/client/components/NewAzureIssueInput.tsx
nickoferrall: I see - so the original reason to create a task when creating an integration issue was to point to the story? Why do we need to point to a story when we're querying directly from the API? If we create a GitHub issue from a Poker meeting, we can query that issue directly from GitHub's API without creating a Parabol task?

PR #7060, opened 2022-08-12
fix(usageStats): active teams definition

### LINKED ISSUE #7054: fix: NewMeeting updatedAt updates after meeting ends
AC
- [ ] a meeting's updatedAt should not update after the meeting has ended (or the reason it is should be documented)
- [ ] update Usage stats to use ` max(meeting.createdAt, meeting.endedAt, meeting.summarySentAt)` if it cannot be resolved

Estimate: 4 hours

mattkrick:
# Description

Fixes #7054

An active team was defined by the team having a meeting.updatedAt being less than 30 days old. updatedAt is now replaced with endedAt/createdAt.
Also, updated applied fixes to stop meeting.updatedAt from updating after the meeting has ended.
 
Test:
- [ ] goto /usage on staging, see parabol.co has 66 active teams (or just review the code, I know that's probably a heavy lift)
- [ ] end a meeting, then go back to it as facilitator & move around. the facilitator stage stays the same
- [ ] end a meeting, promote someone else to facilitator, you can't.

### DIFF
--- a/packages/server/graphql/mutations/navigateMeeting.ts
+++ b/packages/server/graphql/mutations/navigateMeeting.ts
@@ -47,7 +47,10 @@ export default {
     const viewerId = getUserId(authToken)
     const meeting = await r.table('NewMeeting').get(meetingId).default(null).run()
     if (!meeting) return standardError(new Error('Meeting not found'), {userId: viewerId})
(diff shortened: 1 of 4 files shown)


nickoferrall (approved):
LGTM!

PR #7061, opened 2022-08-12
fix: carousel

### LINKED ISSUE #7040: Center the active meeting in the new meeting carousel on mobile
Center the active meeting in the new meeting carousel on mobile so the previous & next options are both slightly visible. That way folks can go back & forth just by tapping instead of swiping.

**Mockup:**

image

This could be done by following settings:

```
        centeredSlides={true}
        centeredSlidesBounds={true}
```

However, there is possibly a bug in swiper library and center is calculated incorrectly when `slidesOffsetBefore={24}` set.

### Acceptance Criteria
- active slide is centered on mobile

**Estimated effort:** 8 points

mattkrick:
Fixes #7040

Fixes carousel sliding: https://www.loom.com/share/d9a85b1904be463ea3cb7401e8f4e47d


packages/client/components/NewMeetingCarousel.tsx
```diff
@@ -139,18 +138,17 @@ const NewMeetingCarousel = (props: Props) => {
...
-        spaceBetween={16}
+        slidesOffsetAfter={24}
+        centeredSlides
+        centeredSlidesBounds
+        slideToClickedSlide
+        spaceBetween={24}
         threshold={10}
         keyboard={true}
-        slidesPerView={1.5}
+        slidesPerView={1.99}
```
mattkrick, line 148: there's some magic here. if we `2` slides are visible, then it won't `slideToClickedSlide` for the 2nd slide moving forward or slide 1 when moving backwards :shrug:

igorlesnenko (changes requested):
I see a couple of issues now

1) The card on mobile is not centered correctly

image

This is because library does not consider `slidesOffsetBefore={24}` when calculating center. (Feels like bug in the library)

2) Card on mobile became smaller now, and the description is trimmed

3) Now we show 3 cards instead of 4 on desktop, and the latest one is cut just a little bit

image

This may give the feeling that we have only three types of meetings.

Comparing to previous setup:

image

Is showing 3 cards better?

mattkrick:
all good points!
we could set a min width for card sizes (216px looks like the best bet to keep the subtitles from condensing) 
the off center thing is infuriating, at least it's a _little_ more centered than before? :sweat_smile: 
i get why it works the way it does, but i'm not sure how to get around it. maybe we add in an empty slide on the front & back & disallow folks from navigating to them? that'll add padding and keep things centered.

up to you if you wanna keep the changes or merge without! i think what you've got is good enough to move forward with

igorlesnenko:
@mattkrick what if we do without `centeredSlidesBounds` option. I pushed a new PR: #7079
What do you think?

PR #7063, opened 2022-08-12
feat(dd sourcemaps): add source maps to deploy. fix(deploy): using staging .env, reverted code to

### LINKED ISSUE #6731: Add source map to Datadog RUM
Relates to:
* #6071

Right now error tracking only shows errors in the minified .js files. To evaluate whether or not we will have the same information available, we need to upload the source map
Datadog:
Screenshot from 2022-08-19 10-41-53
Sentry:
Screenshot from 2022-08-19 10-42-00

## Acceptance criteria
* [ ] Deploying to production automatically pushes the correct source map with correct metadata (version/path)

adaniels-parabol:
# Description

- datadog-ci sourcemap upload at build time, however, the cli requires the .js file and .js.map file to have the same root name. As currently built, the name is different between .js and .js.map file. e.g.name]_[contenthash].js.map and [name]_[contenthash].js' This PR changes contenthash to hash to resolve the datadog-ci issue.
- reverts changes that only used staging .env for deployment, thus all deployments were going to staging with build assets since v6.56.0

Note, I chose to add datadog-ci node module at the time it was needed vs having it placed into the build as it's not a runtime dep, only used for uploading sourcemaps.

resolves #6731

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -74,8 +74,14 @@ jobs:
           name: Build Assets
           # Use the staging env to run webpack (NODE_ENV=production)
           command: |
-            cp $DEVOPS_WORKDIR/environments/staging .env
-            yarn build --deploy
+            if [ "${CIRCLE_BRANCH}" == "production" ] || [ "${CIRCLE_BRANCH}" == "staging" ] || [ "${CIRCLE_BRANCH}" == "development" ]; then
+              cp $DEVOPS_WORKDIR/environments/$CIRCLE_BRANCH .env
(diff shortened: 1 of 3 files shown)


mattkrick:
this will be first on my list tomorrow morning. sorry for the delay, my brain is a little too fried atm to make sense of this :sweat_smile:

mattkrick:
looks great! 
I noticed `hash` is now deprecated https://webpack.js.org/configuration/output/#template-strings so I'll use the new name: `fullhash` :shrug:

Issue #6905, opened 2022-07-20
Demo comments are always anonymous


nickoferrall:
When commenting publically in a discussion thread in the demo, it says that the comment was written by an anonymous user. If you toggle to anonymous and comment, it still says the comment was anonymous. 

See Loom: https://www.loom.com/share/46eafca211c84565ae5c6c250b5f3061

### AC

- [ ] When commenting publically in the demo discussion thread, the comment says the name of the person commenting. We won't have their name, so it could be "Demo User".
- [ ] Toggling to Anonymous commenting still shows the comment as anonymous

**Estimated effort:** 1 hour to triage. More if root cause is already identified.

PR #7065, opened 2022-08-15
fix: demo comments are always anonymous

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6905

I've also updated the comment text to mention standups. 

Currently, in prod:
prod-demo

This PR:
new-demo

### To test

- Can add comments to a thread in the demo and see the author (yourself)
- When the bot adds a comment, it says the author's name
- Can add anonymous comments
- Can add comments & anonymous comments in the Retro & Poker discussion thread

### DIFF
--- a/packages/client/components/ThreadedCommentHeader.tsx
+++ b/packages/client/components/ThreadedCommentHeader.tsx
@@ -71,6 +71,9 @@ export default createFragmentContainer(ThreadedCommentHeader, {
   comment: graphql`
     fragment ThreadedCommentHeader_comment on Comment {
       id
+      createdByUser {
+        preferredName
+      }
       createdByUserNullable: createdByUser {
         preferredName
       }
--- a/packages/client/components/ThreadedCommentHeader.tsx
+++ b/packages/client/components/ThreadedCommentHeader.tsx
@@ -32,9 +32,9 @@ interface Props {
 }
 
 const getName = (comment: ThreadedCommentHeader_comment) => {
-  const {isActive, createdByUser, isViewerComment} = comment
+  const {isActive, createdByUserNullable, isViewerComment} = comment
   if (!isActive) return 'Message Deleted'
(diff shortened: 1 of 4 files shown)


packages/client/components/ThreadedCommentHeader.tsx
```diff
@@ -31,10 +31,10 @@ interface Props {
   meetingId: string
 }
 
-const getName = (comment) => {
-  const {isActive, createdByUserNullable, isViewerComment} = comment
```
nickoferrall, line 35: In https://github.com/ParabolInc/parabol/pull/6042, you added the `createdByUserNullable` alias to `createdByUser` because the type can be null for a comment, but non-null for a poll or task. I believe the alias stops the two fragments from being merged as one?

However, in the demo, it looks like aliasing isn't working as `createdByUserNullable` is always `null` here and in https://github.com/ParabolInc/parabol/blob/8e2f30a09de44a299abd217b63a38cf6cad5201e/packages/client/components/ThreadedCommentBase.tsx#L74

Interestingly, by reverting the change here, `createdByUserNullable` is no longer null in ThreadedCommentBase. 

The demo, tasks, and meeting discussion threads are working well with this change, but perhaps you have an idea why the demo might not work with aliases?

mattkrick: ah, you might not see it fail validation because we don't validate persisted queries.
if you change this line to true, you'll probably see the error pop up again: https://github.com/ParabolInc/parabol/blob/fa767a6f719cec03a966d41cc45d1a077a1857fb/packages/server/graphql/executeGraphQL.ts#L57

To fix it in the demo, you could add `createdByUserNullable` to the response & that should fix it

nickoferrall: Makes sense!

PR #7066, opened 2022-08-15
fix: remove double scrollbar on discuss phase

mattkrick:
# Description

We have 2 scrollbar containers.
Note, this isn't visible on macOS or windows, but it's there! it'll be noticeable as you use your scrollwheel on all platforms.
Specifically, when the outer container is scrolled, the cards will sit on top of the "12 Reflections" header.

## Demo

Screenshot from 2022-08-15 11-54-26

## Testing scenarios

[ ] - create a retro with a bunch of reflections in a group. you can scroll them just fine


nickoferrall (approved):
LGTM

PR #7067, opened 2022-08-15
feat(metrics): clean up some properties for Task Created events

### LINKED ISSUE #7056: Improve parameters on `Task Created` event
This is a shadow copy of the issue from https://github.com/ParabolInc/parabol-data-sanctum/issues/347. Creating this one so I have an issue to link to on the PR.

**TLDR & Acceptance Criteria:**

For the `Task Created` event we emit:
1. Remove the misleading `isAsync` property
2. Remove the `service` property from `Task Created` events
3. Remove the `isReply` property from `Task Created` events
4. Add a property `inMeeting` to indicate if the task is created/updated from meeting
5. Add a property `meetingType` to indicate which type of meeting the task is created from
6. When an integrated task is created during the scoping phase of a poker meeting (e.g., a Jira or GitHub issue), emit a `Task Created` **and** a `Task Published` event

tianrunhe:
# Description

Fixes #7056 

## Demo
Screen Shot 2022-08-16 at 2 36 03 PM

Screen Shot 2022-08-16 at 2 35 44 PM

## Testing scenarios

- [ ] Create a task from team dashboard and verify the property `inMeeting` for `Task Created` event
- [ ] Create a task from a meeting and verify the properties `inMeeting` and `meetingType` for `Task Created` event
- [ ] Create an integration task in the scoping phase of a poker meeting, verify `Task Created` and `Task Published` events are emitted and check the properties `inMeeting`, `meetingType` and `service`.

## Final checklist


Dschoordsch (changes requested):
I think the data could be misleading in the following case:
* create a parabol task in a meeting
* publish the task from the dashboard
It now appears as `inMeeting: true` which is true for the task, but not for the publish event itself.

FYI I also noticed that we don't have events for:
* assigning new team to a task
* import task in Sprint Poker

packages/server/utils/analytics/analytics.ts
```diff
@@ -19,6 +18,14 @@ export type OrgTierChangeEventProperties = {
   billingLeaderEmail: string
 }
 
+export type TaskProperties = {
+  taskId: string
+  teamId: string
+  meetingId?: string
+  meetingType?: MeetingTypeEnum
+  inMeeting: boolean
```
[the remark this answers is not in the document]
tianrunhe: No, we agreed they are misleading so we decided to remove them. See A/C in the [issue](https://github.com/ParabolInc/parabol/issues/7056)

tianrunhe:
> I think the data could be misleading in the following case:
> 
> * create a parabol task in a meeting
> * publish the task from the dashboard
>   It now appears as `inMeeting: true` which is true for the task, but not for the publish event itself.
> 
> FYI I also noticed that we don't have events for:
> 
> * assigning new team to a task
> * import task in Sprint Poker

Thanks for the review! Why do you think the use case is misleading? It feels to me the metric is describing what actually happened, i.e., the task was indeed created in a meeting and later it was published outside a meeting?

For the other events I'll backlog and leave them out of the scope of this PR

Dschoordsch:
>Thanks for the review! Why do you think the use case is misleading? It feels to me the metric is describing what actually happened, i.e., the task was indeed created in a meeting and later it was published outside a meeting?

I'm afraid that when someone asks the question later: "How many tasks are pushed to an integration from within a meeting" this will give wrong results. I feel it is misleading that the push event has `inMeeting: true` when it was pushed outside of the meeting. The properties don't belong to this event in my opinion. I would rather like to have the values of the event itself in the event, i.e. if you push from within a meeting it should be true and false otherwise.
I don't know if you could still answer the question: "How many tasks which were created in a meeting were pushed to integration?" without this? Maybe if you have a funnel and hold constant the task id or sth. like that?

Dschoordsch (approved):
I leave it up to you to merge it like this, but I have my concerns.

tianrunhe:
@Dschoordsch I'm sorry I should have read your comments more closely! When I read: 

> It now appears as inMeeting: true which is true for the task, but not for the publish event itself.

My interpretation was that `inMeeting = true` for `Task Created` and `inMeeting = false` for `Task Published` so I thought wasn't that expected? That's why I said:

> the metric is describing what actually happened, i.e., the task was indeed created in a meeting and later it was published outside a meeting?

But now I'm reading my response I find it sounds like I think `inMeeting` should be true for both `Task Created` and `Task Published`. That's not what I meant! I totally agree with you and now I see where the bug is. I should have totally run the scenario locally! My apology!

To fix this issue, I think we need to identify where the mutation fires (e.g., from a meeting page, or a team dashboard?) so maybe I should utilize `SendClientSegmentMutation`. I'll work on this a bit longer.

Once again, sorry for the misinterpretation!

tianrunhe:
@Dschoordsch How does this change look like?

packages/client/components/ScopingSearchInput.tsx
```diff
@@ -38,7 +38,7 @@ interface Props {
   queryString: string
   meetingId: string
   linkedRecordName: string
-  service: TaskServiceEnum
+  service: IntegrationProviderServiceEnum
```
Dschoordsch, line 41: -1 This seems unrelated and changes the enum to also include mattermost and msTeams which are not task integrations

packages/server/graphql/types/IntegrationProviderServiceEnum.ts
```diff
@@ -10,7 +10,8 @@ const IntegrationProviderServiceEnum = new GraphQLEnumType({
     mattermost: {},
     jiraServer: {},
     azureDevOps: {},
-    msTeams: {}
+    msTeams: {},
+    PARABOL: {}
```
Dschoordsch, line 14: -1 This is not an integration provider really.

packages/server/graphql/types/SegmentEventTrackOptions.ts
```diff
@@ -24,12 +25,15 @@ const SegmentEventTrackOptions = new GraphQLInputObjectType({
     reflectionsCount: {type: GraphQLInt},
     action: {type: GraphQLString},
     searchQueryString: {type: GraphQLString},
-    service: {type: TaskServiceEnum},
+    service: {type: IntegrationProviderServiceEnum},
```
Dschoordsch, line 28: -1 Let's switch back to `TaskServiceEnum` here.

packages/client/mutations/CreateTaskIntegrationMutation.ts
```diff
@@ -187,7 +193,21 @@ const CreateTaskIntegrationMutation: StandardMutation<TCreateTaskIntegrationMuta
...
+    onCompleted: (data, errors) => {
+      if (onCompleted) {
+        onCompleted(data, errors)
+      }
+      if (data.createTaskIntegration && !data?.createTaskIntegration?.error) {
+        SendClientSegmentEventMutation(atmosphere, 'Task Published', {
+          taskId: data.createTaskIntegration.task?.id,
+          teamId: data.createTaskIntegration.task?.teamId,
+          meetingId: data.createTaskIntegration.task?.meeting?.id,
+          meetingType: data.createTaskIntegration.task?.meeting?.meetingType,
```
Dschoordsch, line 205: -1 Ok, now that we established that we don't want wrong information in the 'Task Published' event, let's remove `meetingId` and `meetingType` which come from task. It would be fine if it would come from the current meeting. Following test scenario:
* create task in retrospective
```
analytics.track({
  userId: 'google-oauth2|e3SPYN9W8w',
  event: 'Task Created',
  properties: {
    email: 'georg@parabol.co',
    inMeeting: true,
    meetingId: 'fWD5aT9OHC',
    meetingType: 'retrospective',
    taskId: 'g1PbwncC8E',
    teamId: 'e3SPYTYjg5'
  }
});
```
* publish said task in check-in meeting
```
analytics.track({
  userId: 'google-oauth2|e3SPYN9W8w',
  event: 'Task Published',
  properties: {
    email: 'georg@parabol.co',
    inMeeting: true,
    meetingId: 'fWD5aT9OHC',
    meetingType: 'retrospective',
    service: 'github',
    taskId: 'g1PbwncC8E',
    teamId: 'eOWRlaiWL6'
  }
});
```

Dschoordsch (changes requested):
Sorry, missed some things last time round

packages/client/mutations/CreateTaskIntegrationMutation.ts
tianrunhe: I'm having some difficulties getting the `meeting` object here. Two ways I can think of:

- Parse the URL to get the meetingId and then grab the meeting object through Relay
- Inject `meetingId` as a variable for CreateTaskIntegration mutation

I like neither. Do you have better way?

Dschoordsch: I don't have a better way. Both are not ideal. I would lean slightly more towards adding `meetingId` to the `createTaskIntegration` mutation to be in line with `createTask` mutation. It feels slightly cleaner.

That being said, there is also
https://github.com/ParabolInc/parabol/blob/47ca01164ad036c5fc1c4773811f82adc094b800/packages/client/utils/meetings/getMeetingPathParams.ts#L25
which might simplify your first option. In the end it is up to you. I also missed a meeting context in the past where I had to do the same path parsing.

tianrunhe: Alright I chose to use the first option because:
1. it's less code (if I add `meetingId` to `CreateTaskIntegration` mutation then I need to add `meetingId` to all the places where we call `CreateTaskIntegration`)
2. even if we add `meetingId` to `CreateTaskIntegration`, from client side we still need to get the `meetingId` from somewhere. Since `CreateTaskIntegrationMutation` is fired through `TaskFooterIntegrateMenu`, which could be placed everywhere. So we are back to parsing `meetingId` from URL.

Dschoordsch (changes requested):

packages/client/mutations/CreateTaskIntegrationMutation.ts
```diff
@@ -187,7 +191,24 @@ const CreateTaskIntegrationMutation: StandardMutation<TCreateTaskIntegrationMuta
...
       }
     },
-    onCompleted,
+    onCompleted: (data, errors) => {
+      if (onCompleted) {
+        onCompleted(data, errors)
+      }
+      const {meetingId} = getMeetingPathParams()
+      const store = atmosphere.getStore()
+      const meetingType = (store as any)?._recordSource?._records?.get(meetingId)?.meetingType
```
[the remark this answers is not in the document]
tianrunhe: I was lazy and just copied the code from [`ErrorBoundary`](https://github.com/ParabolInc/parabol/blob/master/packages/client/components/ErrorBoundary.tsx#L40) 😂. I updated that too!

Dschoordsch (approved):

Issue #7068, opened 2022-08-16
Empty Discuss phase after skipping Vote phase


nickoferrall:
To reproduce:

- Create a retro and add reflections
- Proceed to the Group phase
- Then click on the Discuss phase in the sidebar without going to the Vote phase
- See that the Discuss phase looks like this:

![discuss](https://user-images.githubusercontent.com/39854876/184837484-10e9b45f-931d-49c3-abb3-0d85db080fb7.png)

https://www.loom.com/share/d7bbef5e5dbe432a80b0ee3540a63c73

### AC

- [ ] Discuss phase shows discussion items even if the user skips the Vote phase
- [ ] All discussion items have zero votes

PR #7070, opened 2022-08-16
fix: empty discuss phase

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7068

https://www.loom.com/share/c1e68a8e0ddc4147a942f65d8680584c

### To test

- [ ] Can't proceed to the Discuss phase before completing the Vote phase, e.g. jumping from Reflect to Discuss
- [ ] Can proceed to the Discuss phase without voting in Vote phase
- [ ] When the user proceeds to the Discuss phase, the Discuss phase shows all the groups and each has 0 votes

### DIFF
--- a/packages/server/graphql/mutations/helpers/handleCompletedStage.ts
+++ b/packages/server/graphql/mutations/helpers/handleCompletedStage.ts
@@ -1,10 +1,12 @@
 import {AUTO_GROUPING_THRESHOLD, GROUP, REFLECT, VOTE} from 'parabol-client/utils/constants'
-import {AnyMeeting} from '../../../postgres/types/Meeting'
+import unlockAllStagesForPhase from 'parabol-client/utils/unlockAllStagesForPhase'
+import {r} from 'rethinkdb-ts'
 import groupReflections from '../../../../client/utils/smartGroup/groupReflections'
 import getRethink from '../../../database/rethinkDriver'
 import GenericMeetingStage from '../../../database/types/GenericMeetingStage'
 import MeetingRetrospective from '../../../database/types/MeetingRetrospective'
 import insertDiscussions from '../../../postgres/queries/insertDiscussions'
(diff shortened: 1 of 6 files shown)


packages/server/graphql/mutations/helpers/handleCompletedStage.ts
```diff
@@ -60,6 +62,17 @@ const handleCompletedRetrospectiveStage = async (
...
+      const {phases} = meeting
+      unlockAllStagesForPhase(phases, 'discuss', true)
+      await r
+        .table('NewMeeting')
+        .get(meeting.id)
+        .update({
+          phases
+        })
+        .run()
+      data.meeting = meeting
```
mattkrick, line 75: +1 is this necessary? looks like unlockAllStagesForPhase _should_ mutate the phases object directly `data.meeting === meeting`, but I could be missing something

mattkrick (approved):
clean! just 1 little question

packages/server/graphql/mutations/helpers/handleCompletedStage.ts
nickoferrall: `unlockAllStagesForPhase` mutates the phase object, but we need to add the updated meeting to the payload, otherwise we get: `Cannot return null for non-nullable field GroupPhaseCompletePayload.meeting`

nickoferrall: `unlockAllStagesForPhase` mutates the phase object, but we need to add the updated meeting to the payload, otherwise we get: `Cannot return null for non-nullable field GroupPhaseCompletePayload.meeting`

nickoferrall: `unlockAllStagesForPhase` mutates the phase object, but we need to add the updated meeting to the payload, otherwise we get: `Cannot return null for non-nullable field GroupPhaseCompletePayload.meeting`

Issue #7032, opened 2022-08-09
Disable retrospective anonymity


igorlesnenko:
## Issue - Enhancement

Allow users to disable retrospective reflections anonymity

![image](https://user-images.githubusercontent.com/466991/183664269-a0aa3334-e6a7-467e-82f2-3c4bbdac0501.png)

<img width="768" alt="image" src="https://user-images.githubusercontent.com/466991/183665957-0bdadb09-9121-4817-bcb6-d18baf25f470.png">

Mockups: https://www.figma.com/file/o3q946BB5UlYZw0RJeNHdx/Non-anonymous-retros?node-id=0%3A1

Design issue: #5419

### Acceptance Criteria 

- [ ] Add new settings dropdown with Icebreaker and Anonymous reflections checkboxes to the Retrospective new meeting dialog.
- [ ] Anonymity is enabled by default
- [ ] A name is displayed below the reflection card content

**Estimated effort:** 16 points

PR #7076, opened 2022-08-17
feat: allow disable anonymity for retrospective meeting reflections

igorlesnenko:
Fixes: #7032

image

image

Demo:

https://www.loom.com/share/5f10ba990517438ca4b6048342bf2c65

How to test:

- Create a retro meeting, see that reflections are anonymous by default
- In new meeting dialog remove checkbox from Anonymous reflections option, create a retro meeting, see reflections now have name attached
- Try to create various reflections and interact with them
- Refresh page, go to meeting dialog, see latest settings are saved
- Smoke test that other meetings are created normally and icebreaker checkbox works too


igorlesnenko:
Added a name below reflection input too 

image

packages/client/components/NewMeetingSettingsToggleAnonymity.tsx
```diff
@@ -0,0 +1,86 @@
...
+import {useFragment} from 'react-relay'
+import {NewMeetingSettingsToggleAnonymity_settings$key} from '~/__generated__/NewMeetingSettingsToggleAnonymity_settings.graphql'
+import useAtmosphere from '../hooks/useAtmosphere'
+import useMutationProps from '../hooks/useMutationProps'
+import SetMeetingSettingsMutation from '../mutations/SetMeetingSettingsMutation'
+import {PALETTE} from '../styles/paletteV3'
+import Checkbox from './Checkbox'
+import PlainButton from './PlainButton/PlainButton'
+
+const ButtonRow = styled(PlainButton)({
```
JimmyLv, line 13: -1 both `NewMeetingSettingsToggleCheckIn` and `NewMeetingSettingsToggleAnonymity` rely on the exact same `ButtonRow`, and the code for the `ButtonRow` component is repeated 2 times in both files, would you consider reusing one component [^1] since they are basically identical?

[^1]: https://github.com/ParabolInc/parabol/commit/73eb08e13d1ba8568567326af964294b2677576c#diff-987709bac0f7c25b6e84f98c2e709d498da60b668cdf85ae9bfe0565ecd09022R13

packages/server/graphql/types/RetroReflection.ts
```diff
@@ -45,6 +46,26 @@ const RetroReflection = new GraphQLObjectType<Reflection, GQLContext>({
...
+      resolve: async ({creatorId, meetingId}, _args: unknown, {dataLoader}) => {
+        const meeting = await dataLoader.get('newMeetings').load(meetingId)
+
+        const {meetingType} = meeting
+
+        if (meetingType !== 'retrospective') {
+          return null
+        }
+
+        if (!meeting.disableAnonymity) {
```
JimmyLv, line 61: +1 `not disabled anonymity` means the same as anonymous, it just feels a bit strange (aha, my head reversed three times 😂), maybe give it a more reasonable name, depending on you.

packages/server/graphql/mutations/setMeetingSettings.ts
```diff
@@ -0,0 +1,76 @@
...
+    }
+    const {teamId} = settings
+
+    // RESOLUTION
+    await r
+      .table('MeetingSettings')
+      .get(settingsId)
+      .update((row: RValue) => {
+        const updatedSettings: {[key: string]: any} = {}
+        if (checkinEnabled !== null && checkinEnabled !== undefined) {
```
JimmyLv, line 54: -1 Do we have a Util like `isNil`? Just like Lodash does, to checks if value is null or undefined. [^1] Or we can extract one ourselves.

[^1]: https://lodash.com/docs/4.17.15#isNil

packages/server/graphql/mutations/setMeetingSettings.ts
```diff
@@ -0,0 +1,76 @@
...
+          updatedSettings.phaseTypes = r.branch(
+            row('phaseTypes').contains('checkin'),
+            checkinEnabled ? row('phaseTypes') : row('phaseTypes').difference(['checkin']),
+            checkinEnabled ? row('phaseTypes').prepend('checkin') : row('phaseTypes')
+          )
+        }
+
+        if (disableAnonymity !== null && disableAnonymity !== undefined) {
```
JimmyLv, line 62: -1 same as above

packages/client/mutations/SetMeetingSettingsMutation.ts
```diff
@@ -0,0 +1,75 @@
...
+    mutation,
+    variables,
+    onCompleted,
+    onError,
+    optimisticUpdater: (store) => {
+      const {checkinEnabled, disableAnonymity, settingsId} = variables
+      const settings = store.get<Settings>(settingsId)
+      if (!settings) return
+
+      // relay
```
JimmyLv, line 54: +1 to remove the unneeded comment?

packages/client/mutations/SetMeetingSettingsMutation.ts
```diff
@@ -0,0 +1,75 @@
...
+        const phaseTypes = settings.getValue('phaseTypes').slice() as NewMeetingPhaseTypeEnum[]
+        if (checkinEnabled && !phaseTypes.includes('checkin')) {
+          phaseTypes.unshift('checkin')
+        } else if (!checkinEnabled && phaseTypes.includes('checkin')) {
+          phaseTypes.splice(
+            phaseTypes.findIndex((type) => type === 'checkin'),
+            1
+          )
+        }
+        settings.setValue(phaseTypes, 'phaseTypes')
```
JimmyLv, line 65: -1 would prefer an immutable method that doesn't modify the arrays? like:

```TypeScript
const phaseTypes = settings.getValue('phaseTypes') as NewMeetingPhaseTypeEnum[]

if (checkinEnabled && !phaseTypes.includes('checkin')) {
  settings.setValue(['checkin', ...phaseTypes], 'phaseTypes')
} else if (!checkinEnabled && phaseTypes.includes('checkin')) {
  settings.setValue(phaseTypes.filter(type => type !== 'checkin'), 'phaseTypes')
}
```

JimmyLv (changes requested):
I've tested the functionality locally and it works great! Well done! 

For the functionality, I have 2 questions about the `DropdownMenu` options: 
1. The arrow icon on the right, should it be directional? 
in the Figma design, 🔽 The **_downward_** arrow, corresponding to the one that shows the selected item _**below**_. [^1]
Google Chrome-Team Dashboard  Jimmy Lv’s Team-000008-20220825

2. The direction of the arrows, 🔼 🔽 should it switch with clicking, as in the effect of this GIF.
Google Chrome-httpswww google comsearchqunshift+js


For the code, please see my specific comments. 😁

[^1]: https://www.figma.com/file/o3q946BB5UlYZw0RJeNHdx/Non-anonymous-retros?node-id=7%3A628

gcrickman:
A [user](https://app.hubspot.com/contacts/3888472/contact/92714701) 🔒 asks:

> I love your product but the Retro meetings suffer from one huge drawback: people can't find the reflections they wrote when they have to present them!
> 
> So once everyone has written them, instead of jumping into the discussion and grouping, we have to write out which one's whose and spend time reordering them.
> 
> Instead, can you:
> Show who authored the reflection (initial, avatar, name, anything)
> Organise the reflections in each column by the same author - so they are all in one place (until we start grouping them, of course)
> Provide a filter by author?
> This would really make it a lot easier to use and useful!
> 

Thanks very much!

Dschoordsch:
@JimmyLv I fixed the issues. I also added the dropdown icon modification you requested
Dropdown animation

@ackernaut  Igor modified the enter reflection dialog and added the author's name there, could you take a look if that's ok? That should help address users not catching it
Author below reflection input

ackernaut:
>Igor modified the enter reflection dialog and added the author's name there, could you take a look if that's ok? That should help address users not catching it

I tried this out locally. Wow, I think this turned out great. I think seeing the more of the final state from the start is a nice detail, should orient the user better. Thank you @igorlesnenko  and thanks for the ping @Dschoordsch 

💯

JimmyLv (approved):
LGTM! 👍

It's just that Sonar detected a duplicate code rate of more than 3%, so the check failed. I've collected such cases as well https://github.com/ParabolInc/parabol/issues/7123

Issue #6975, opened 2022-07-29
Standups: hyperlink default focus is on the highlighted text


nickoferrall:
In standups, when adding a hyperlink, the focus is on the "text" input rather than the "link" input. Usually, I don't need to edit the text, but I always need to add a link, so I'd prefer the focus to be on "link".

![hyperlink](https://user-images.githubusercontent.com/39854876/181790424-233ef136-dc43-4b7f-adc0-8b0386f3c307.png)

### AC

- [ ] Hyperlink modal focuses on "link" input by default

PR #7078, opened 2022-08-17
fix: hyperlink default focus

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6975

https://www.loom.com/share/3cb5f3c909474f95be24fa92ab299281

### To test

- [ ] In standups, if the user selects some text and presses cmd + k, the hyperlink modal focuses on the link input rather than the text input
- [ ] When using the hyperlink modal in any meeting type, if the text the user highlights is trimmed
- [ ] If the user highlights empty text, the focus remains on the text input

### DIFF
--- a/packages/client/components/EditorLinkChanger/EditorLinkChangerModal.tsx
+++ b/packages/client/components/EditorLinkChanger/EditorLinkChangerModal.tsx
@@ -125,8 +125,8 @@ const EditorLinkChangerModal = (props: Props) => {
           <InputBlock>
             <BasicInput
               {...fields.link}
+              autoFocus={!link && !!text}
               value={fields.link.value === null ? '' : fields.link.value}
-              autoFocus={link === null && text !== ''}
               onChange={onChange}
               name='link'
               spellCheck={false}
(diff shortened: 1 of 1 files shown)


packages/client/components/EditorLinkChanger/EditorLinkChangerModal.tsx
```diff
@@ -126,7 +127,7 @@ const EditorLinkChangerModal = (props: Props) => {
             <BasicInput
               {...fields.link}
               value={fields.link.value === null ? '' : fields.link.value}
-              autoFocus={link === null && text !== ''}
```
nickoferrall, line 129: `link` could be an empty string which is why there wasn't any autofocus

jmtaber129 (changes requested):
Thanks for looking into this! It's hard to think of a case when someone would intentionally create a link with a preceding space.

-1 There is one edge case that I think needs to be addressed - if I select text with spaces that would be trimmed and then add a link, those spaces are deleted after I hit submit.  E.g. if I have text `"cool thing"`, highlight the text `" thing"` (with space), add a link, and submit, the text becomes `"coolthing"`.

This probably means doing the trimming in the Draftjs-/TipTap-specific link changer modals by changing the selection to exclude spaces before grabbing the link text.

nickoferrall:
Good find! That looks like a separate issue that's happening in prod, so I've created an issue for that: https://github.com/ParabolInc/parabol/issues/7088

jmtaber129:
Oof, you're right!

Issue #7016, opened 2022-08-07
Icon looks clickable in timeline


karmatosed:
The timeline avatar icon choice is the expected one for clicking and adding an icon which makes this expected to be clickable. I would recommend that this should either be made clickable or swapped out. Probably swapping out makes more sense for an easier catch.

![2022-08-07 at 13 05](https://user-images.githubusercontent.com/253067/183289716-557861d9-4b1b-4ab9-8a21-df803b2df28f.jpg)

As far as the level of bug, this feels like a low-hanging fix. It could be an annoying hitch to experience for someone if they did feel they should click and it wasn't clickable, but I have no data for how often that happens.

jordanh:
AC
- Icon should perform the same action as the CTA

EE
- 1 hours

PR #7082, opened 2022-08-17
fix: icon looks clickable in timeline

gitstart:
### Description
Fixes/Partially fixes [Parabol Issue 7016](https://github.com/ParabolInc/parabol/issues/7016)

### Demo
[Loom Video](https://www.loom.com/share/9f0b90ad6aed425990037784e34452ed)

### Testing Scenarios

- Icon looks less clickable to prompt user to click on the "Invite Teammate" button.
- Upon hovering over Icon, it becomes more visible with a pointer cursor indicating "clickability".
- Upon Clicking, Icon should have the same functionality as "Invite Teammate" button.

### Final Checklist

- [x] I checked the [code review guidelines](https://github.com/GitStartHQ/client-parabol/docs/codeReview.md)    
- [x] I have added [Metrics Representative](https://github.com/GitStartHQ/client-parabol/docs/codeReview.md#metrics-representative) as reviewer(s) if my PR involves metrics/data/analytics related changes.
- [x]  I have performed a self-review of my code, the same way I'd do it for any other team member.
- [x] I have tested all cases I listed in the testing scenarios and I haven't found any issues or regressions.
- [x] Whenever I took a non-obvious choice I added a comment explaining why I did it this way.
- [x] I added the label One Review Required if the PR introduces only minor changes, does not contain any architectural changes or does not introduce any new patterns and I think one review is sufficient.
- [x] PR title is human readable and could be used in changelog.


packages/client/components/SuggestedActionCard.tsx
```diff
@@ -46,14 +47,19 @@ const CancelIcon = styled(Icon)({
...
   background: PALETTE.SLATE_300,
   borderRadius: '100%',
   boxShadow: buttonShadow,
   padding: 8,
   position: 'absolute',
   fontSize: ICON_SIZE.MD36,
   top: 100,
-  userSelect: 'none'
+  userSelect: 'none',
+  opacity: 0.7,
```
Dschoordsch, line 59: -2 this makes the problem only worse. `SuggestedActionCard` is used in multiple places and now looks clickable in all of those regardless of the icon used.

@ackernaut The issue was not well defined what the best solution would be. I don't mind having the icon repeat the primary action of a suggestion, but maybe there are some guideline against this?

Dschoordsch (changes requested):
This needs some clarification first if making the icon clickable (in which case it should be clickable in all suggested actions) or if we change the visual to not look like a button.

packages/client/components/SuggestedActionCard.tsx
gitstart: @Dschoordsch  thanks for the feedback. You can let me know what you decide

Dschoordsch:
@ackernaut could you check the comments above?

ackernaut:
Here’s my 2 cents:
- I’m only aware of a single piece of feedback that those icons look clickable. While I can see the point (it makes good design feedback, too) I don’t think there’s enough signal here to chase this single point of feedback at the moment over other better known issues. We could probably use a label like ‘feedback ready to solve’ that we stick on user feedback that has enough signal and is therefore available to pick up and work on, but I don’t really want to decide process in a comment 😄 that probably is a product management concern of deciding what can be worked on and what needs to wait and collect more feedback
- I could see these suggested actions being deprecated for other onboarding CTA patterns, so I also feel less inclined to put effort here
- If I were to put effort here I would not make the icon clickable to match the button. I would make the icon look less clickable, maybe remove the drop shadow and use a more muted tone for the icon that makes this more like a visual decorator than a floating icon-only button (as that was the intent for it to be, purely a visual decorator when we didn’t have time to reach for good, decorating illustrations etc. like we do today — this was designed a long time ago)
- I would be okay calling this a sunk cost, closing the issues, and we all go focus on more valuable things
- Feel free to throw me off the island 🏖️ but please send me to another beautiful one 😎

gitstart:
closing this PR, thanks @ackernaut

Issue #7088, opened 2022-08-18
Hyperlink removes whitespace


nickoferrall:
https://www.loom.com/share/8738670f67da40e5b0db2c5a5f5924c4

To reproduce:

- Enter two words into the discussion thread like `"super cool"`
- Highlight the whitespace and the second word, so `" cool"`
- Add a link and press submit
- See that the whitespace has been removed so the text is now `supercool`

### AC

- [ ] Can hyperlink a word with a space without the space being removed

PR #7089, opened 2022-08-18
fix: Hyperlink removes whitespace

87prashant:
# Description
Fixes: #7088 

## Demo
New

## Final checklist

### DIFF
--- a/packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
+++ b/packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
@@ -56,9 +56,9 @@ const EditorLinkChangerDraftjs = (props: Props) => {
     const nonWhitespaceFromEnd = text.search(/\s*$/)
     const startStr = nonWhitespaceFromStart === -1 ? '' : text.slice(0, nonWhitespaceFromStart)
     const endStr = nonWhitespaceFromEnd === -1 ? '' : text.slice(nonWhitespaceFromEnd)
-    const hasTextTitle = text.trim()
+    const hasTextTitle = !!text.trim()
     // trim link text if it contains any non-whitespace characters, otherwise keep it verbatim
-    const trimmedText = hasTextTitle || text
+    const trimmedText = hasTextTitle ? text.trim() : text
     const focusedEditorState = EditorState.forceSelection(editorState, selectionState)
     let newEditorState = focusedEditorState
     if (hasTextTitle) {
--- a/packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
+++ b/packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
@@ -2,7 +2,7 @@ import {ContentState, EditorState, Modifier, SelectionState} from 'draft-js'
 import React, {RefObject} from 'react'
 import {UseTaskChild} from '../../hooks/useTaskChildFocus'
 import {BBox} from '../../types/animations'
-import completeEntity, {getExpandedSelectionState} from '../../utils/draftjs/completeEntity'
+import completeEntity from '../../utils/draftjs/completeEntity'
 import EditorLinkChangerModal from './EditorLinkChangerModal'
 
 interface Props {
@@ -51,21 +51,19 @@ const EditorLinkChangerDraftjs = (props: Props) => {
   } = props
   useTaskChild('editor-link-changer')
   const handleSubmit = ({text, href}: {text: string; href: string}) => {
-    const keepSelection = true
     // don't include leading and trailing whitespace in the link text
     const nonWhitespaceFromStart = text.search(/\S/)
     const nonWhitespaceFromEnd = text.search(/\s*$/)
     const startStr = nonWhitespaceFromStart === -1 ? '' : text.slice(0, nonWhitespaceFromStart)
     const endStr = nonWhitespaceFromEnd === -1 ? '' : text.slice(nonWhitespaceFromEnd)
-    const trimmedText = text.trim() ? text.trim() : text
+    const hasTextTitle = text.trim()
+    // trim link text if it contains any non-whitespace characters, otherwise keep it verbatim
+    const trimmedText = hasTextTitle || text
     const focusedEditorState = EditorState.forceSelection(editorState, selectionState)
     let newEditorState = focusedEditorState
-    // For whitespace only links
-    if (text.trim()) {
+    if (hasTextTitle) {
       const contentState = focusedEditorState.getCurrentContent()
-      const expandedSelectionState = keepSelection
-        ? focusedEditorState.getSelection()
-        : getExpandedSelectionState(focusedEditorState)
+      const expandedSelectionState = focusedEditorState.getSelection()
       const contentStateAfterStartStr: ContentState = expandedSelectionState.isCollapsed()
         ? Modifier.insertText(contentState, expandedSelectionState, startStr)
         : Modifier.replaceText(contentState, expandedSelectionState, startStr)
@@ -76,11 +74,11 @@ const EditorLinkChangerDraftjs = (props: Props) => {
       )
     }
     newEditorState = completeEntity(newEditorState, 'LINK', {href}, trimmedText, {
-      keepSelection
+      keepSelection: true
     })
     const selectionStateAfterTrimmedText = newEditorState.getSelection()
     let contentStateAfterEndStr = newEditorState.getCurrentContent()
-    if (text.trim()) {
+    if (hasTextTitle) {
       const contentWithEntity = newEditorState.getCurrentContent()
       contentStateAfterEndStr = Modifier.insertText(
         contentWithEntity,
--- a/packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
+++ b/packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
@@ -61,7 +61,7 @@ const EditorLinkChangerDraftjs = (props: Props) => {
     const focusedEditorState = EditorState.forceSelection(editorState, selectionState)
     let newEditorState = focusedEditorState
     // For whitespace only links
-    if (trimmedText) {
+    if (text.trim()) {
       const contentState = focusedEditorState.getCurrentContent()
       const expandedSelectionState = keepSelection
         ? focusedEditorState.getSelection()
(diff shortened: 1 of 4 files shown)


87prashant:
@nickoferrall 
Spaces are not removed (AC)  but they become underlined as in SS. I am not sure if that should be acceptable.

nickoferrall:
@87prashant thanks for the PR. I think we should aim to only hyperlink the text rather than the whitespace though

87prashant:
Ok, would work on that

87prashant:
@nickoferrall could you take a look again?

JimmyLv:
I just reviewed the Security issue reported by sonar and marked it as "Safe", so now "All checks have passed"

87prashant:
Hi @BartoszJarocki
It is ready for review

packages/client/utils/draftjs/completeEntity.tsx
```diff
@@ -80,29 +94,60 @@ const completeEntity = (
   const {keepSelection} = options
```
Dschoordsch, line 94: -1 I think since this function is used also by other functionality (like adding emojis) and the omitting space logic is limited to link handling, I think we should move the logic to https://github.com/ParabolInc/parabol/blob/7a409665830e2b00166f3306e611ad31759f0b32/packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx#L39
What do you think?

Then we might be able to basically do:
* store current selection
* shrink selection to text only part (might collapse into just the input cursor)
* replace with link
* expand selection to include previous whitespace

What do you think?

Dschoordsch (changes requested):
@nickoferrall I think omitting the whitespace from the link is nicer, but it also complicates things. I checked in Slack and they:
* include whitespace in the link verbatim
* allow whitespace only links
I don't think it's the best solution, but it simplifies things and might match user expectation

nickoferrall:
@Dschoordsch that makes sense to me 👍

packages/client/utils/draftjs/completeEntity.tsx
```diff
@@ -80,29 +94,60 @@ const completeEntity = (
...
   const {editorChangeType, entityType} = operationTypes[entityName]
   const contentState = editorState.getCurrentContent()
-  const contentStateWithEntity = contentState.createEntity(entityName, entityType, entityData)
-  const entityKey = contentStateWithEntity.getLastCreatedEntityKey()
+  // for https://github.com/ParabolInc/parabol/issues/7088
```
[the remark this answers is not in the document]
87prashant: Ok

packages/client/utils/draftjs/completeEntity.tsx
87prashant: Ok, thanks!

87prashant: Done

packages/client/components/EditorLinkChanger/EditorLinkChangerModal.tsx
```diff
@@ -66,7 +66,6 @@ const EditorLinkChangerModal = (props: Props) => {
       getDefault: () => text,
       validate: (value) =>
         new Legitity(value)
-          .trim()
```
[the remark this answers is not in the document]
87prashant: Done, added whitespace only links

packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
```diff
@@ -37,11 +51,49 @@ const EditorLinkChangerDraftjs = (props: Props) => {
   } = props
   useTaskChild('editor-link-changer')
   const handleSubmit = ({text, href}: {text: string; href: string}) => {
+    const keepSelection = true
+    // don't include leading and trailing whitespace in the link text
+    const nonWhitespaceFromStart = text.search(/\S/)
+    const nonWhitespaceFromEnd = text.search(/\s*$/)
+    const startStr = nonWhitespaceFromStart === -1 ? '' : text.slice(0, nonWhitespaceFromStart)
+    const endStr = nonWhitespaceFromEnd === -1 ? '' : text.slice(nonWhitespaceFromEnd)
+    const trimmedText = text.trim() ? text.trim() : text
```
Dschoordsch, line 60: -1 could you please add a comment why you're doing this
```ts
    // trim link text if it contains any non-whitespace characters, otherwise keep it verbatim
```
could also be shortened to
```ts
    const trimmedText = text.trim() || text
```
if you like

packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
```diff
@@ -37,11 +51,49 @@ const EditorLinkChangerDraftjs = (props: Props) => {
...
     const focusedEditorState = EditorState.forceSelection(editorState, selectionState)
-    const nextEditorState = completeEntity(focusedEditorState, 'LINK', {href}, text, {
-      keepSelection: true
+    let newEditorState = focusedEditorState
+    // For whitespace only links
+    if (text.trim()) {
```
Dschoordsch, line 64: -1 This is the non-whitespace branch, so the comment above is a bit confusing. Since we branch more often, maybe it's helpful to put this in an variable to make it explicit, so we can avoid some of the comments. This should then also be used in l60
```ts
const hasTextTitle = !!text.trim()
```

packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
```diff
@@ -37,11 +51,49 @@ const EditorLinkChangerDraftjs = (props: Props) => {
   } = props
   useTaskChild('editor-link-changer')
   const handleSubmit = ({text, href}: {text: string; href: string}) => {
+    const keepSelection = true
```
Dschoordsch, line 54: -1 please remove and manually switch to the true branch later

packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
```diff
@@ -37,11 +51,49 @@ const EditorLinkChangerDraftjs = (props: Props) => {
...
+    const endStr = nonWhitespaceFromEnd === -1 ? '' : text.slice(nonWhitespaceFromEnd)
+    const trimmedText = text.trim() ? text.trim() : text
     const focusedEditorState = EditorState.forceSelection(editorState, selectionState)
-    const nextEditorState = completeEntity(focusedEditorState, 'LINK', {href}, text, {
-      keepSelection: true
+    let newEditorState = focusedEditorState
+    // For whitespace only links
+    if (text.trim()) {
+      const contentState = focusedEditorState.getCurrentContent()
+      const expandedSelectionState = keepSelection
```
Dschoordsch, line 66: -1 this is always `true`

Dschoordsch (changes requested):
Works like a charm! Only some minor cleanup required.

packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
87prashant: I am not getting what you mean by branch here (maybe it is a naive question😑). Could you define this term?

packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
87prashant: > what you mean by branch

Here also.

packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
Dschoordsch: You're branching the execution flow based on a condition
```
if (condition) {
  // "true branch"
} else {
  // "false branch"
}
```
So anytime something happens on a condition, there are multiple execution branches.

What I meant is, because the `text.trim()` appeared in 3 different conditions (`if` and `?`) let's just put it in a variable with a meaningful name so we can omit comments.

packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
Dschoordsch: Here I mean, because `keepSelection` is always true, we can just write
```ts
      const expandedSelectionState = focusedEditorState.getSelection()
```
directly

packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
87prashant: Using
```
hasTextTitle = text.trim()
```

packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
87prashant: yeah

packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
```diff
@@ -51,21 +51,19 @@ const EditorLinkChangerDraftjs = (props: Props) => {
...
   useTaskChild('editor-link-changer')
   const handleSubmit = ({text, href}: {text: string; href: string}) => {
-    const keepSelection = true
     // don't include leading and trailing whitespace in the link text
     const nonWhitespaceFromStart = text.search(/\S/)
     const nonWhitespaceFromEnd = text.search(/\s*$/)
     const startStr = nonWhitespaceFromStart === -1 ? '' : text.slice(0, nonWhitespaceFromStart)
     const endStr = nonWhitespaceFromEnd === -1 ? '' : text.slice(nonWhitespaceFromEnd)
-    const trimmedText = text.trim() ? text.trim() : text
+    const hasTextTitle = text.trim()
```
Dschoordsch, line 59: +1 Name is not ideal as it implies to be a boolean.
It would be nicer to do
```ts
const hasTextTitle = !!text.trim()
// ...
const trimmedText = hasTextTitle ? text.trim() : text
```
This introduces back a bit of redundant code, but improves readability. My fault for leaving the comment in there with `||` after adding the other one about the variable.
```

Dschoordsch (approved):
Would be nice if you could still improve this last bit.

packages/client/components/EditorLinkChanger/EditorLinkChangerDraftjs.tsx
87prashant: Ok

Dschoordsch (approved):
Thanks a lot!

Issue #7072, opened 2022-08-16
Sort team members preview in new meeting team selector


igorlesnenko:
Prev PR: #7001
Currently, we randomize team members in preview. To make team more recognizable do the following:

AC:

* always showing the team lead, since they're probably the most recognizable user on the team ("Oh yeah, that's Igor's team")
* removing the viewer from the shuffle when the team **size > 4** & they're not the team lead (that way, we can provide as much useful info as possible)

PR #7091, opened 2022-08-18
fix: show the team lead first in the preview of team selector

igorlesnenko:
Fixes: #7072

image

How to test:

- Create a team with 4 members, see that team lead is always shown first, other members are shuffled
- Create a team with more than 4 members, see that you are excluded from the preview if you are not the team lead


Dschoordsch (approved):
Works as advertised.

Issue #7038, opened 2022-08-10
Make Azure Work Item id unique


nickoferrall:
Currently, the Azure Work Item graphql id is not unique: https://github.com/ParabolInc/parabol/blob/26ec310dcc9e2b35be2aad5922da62aa7cc5ea59/packages/server/graphql/types/AzureDevOpsWorkItem.ts#L15-L18. 

See thread for further context: https://github.com/ParabolInc/parabol/pull/7007#discussion_r940040671

### AC

- Add a resolver that makes the id unique using `AzureDevOpsIssueId`
- Update the description to reflect the change
- Update areas of the app that were using the id as a key, e.g. https://github.com/ParabolInc/parabol/blob/04dd285132de56059a30142f4e0b69991ff56645/packages/client/components/AzureDevOpsScopingSearchResultItem.tsx#L104

Issue #7073, opened 2022-08-17
Refactor teamProject in Azure integration


nickoferrall:
In the Azure integration, it looks like `teamProject` can be a name or an id. 

In https://github.com/ParabolInc/parabol/blob/26ec310dcc9e2b35be2aad5922da62aa7cc5ea59/packages/server/graphql/types/AzureDevOpsWorkItem.ts#L23-L26 it suggests it's only a name, but it looks like it's also being used as an id: https://github.com/ParabolInc/parabol/blob/04e4f7939a8677df05b946b8b46161fd4ba86099/packages/server/dataloader/azureDevOpsLoaders.ts#L530-L533

### AC

- [ ] Verify whether `teamProject` is being used as a name, id, or both
- [ ] Settle on a `name` or `id` and rename `teamProject` to something like `teamProjectName` or `teamProjectId`
- [ ] Refactor all uses of `teamProject` in our codebase to use a consistent type
- [ ] Fix inconsistencies with `projectId` and `projectKey`. See https://github.com/ParabolInc/parabol/pull/7093#discussion_r955093072

PR #7093, opened 2022-08-19
fix: azure work item id & change item link to project name

nickoferrall:
Predecesor: https://github.com/ParabolInc/parabol/pull/7053

Fix https://github.com/ParabolInc/parabol/issues/7038 & fix https://github.com/ParabolInc/parabol/issues/7048

This PR changes the Azure Work Item id from an issue key of say `7` to a unique id of `instanceId:projectKey:issueKey`. 

The id was being used in the client, see: https://github.com/ParabolInc/parabol/blob/04dd285132de56059a30142f4e0b69991ff56645/packages/client/components/AzureDevOpsScopingSearchResultItem.tsx#L98-L106

So I've changed this to show the issue key & project name:
azure-link

### To test

- [ ] See that Work Item id is unique
- [ ] The work item link description is the `issueKey` and the `project.name`
- [ ] Can create a new issue and nothing breaks


packages/client/components/AzureDevOpsScopingSearchResultItem.tsx
```diff
@@ -1,111 +0,0 @@
-import styled from '@emotion/styled'
```
nickoferrall, line 1: This wasn't being used anywhere as we're using `ScopingSearchResultItem`, so I've removed it

packages/server/graphql/types/AzureDevOpsWorkItem.ts
```diff
@@ -14,15 +17,39 @@ const AzureDevOpsWorkItem = new GraphQLObjectType<any, GQLContext>({
...
       type: new GraphQLNonNull(GraphQLString),
       description: 'Title of the work item'
     },
+    // TODO: change teamProject name: https://github.com/ParabolInc/parabol/issues/7073
     teamProject: {
       type: new GraphQLNonNull(GraphQLString),
-      description: 'Name of the Team Project the work item belongs to'
+      description: 'Name or id of the Team Project the work item belongs to'
+    },
+    project: {
```
nickoferrall, line 35: It feels expensive grabbing the `project` just for the `projectName`, but we can't guarantee the project name from `teamProject` as it's often an `id`. Having access to the `project` in the `workItem` could well be useful in the future, and we currently do this in Jira, so I think it's the right approach

mattkrick: agreed!

mattkrick (changes requested):
couple minor readability things, other than that LGTM!

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -347,6 +349,45 @@ export const allAzureDevOpsProjects = (
   )
 }
 
+export const azureDevOpsProject = (
+  parent: RootDataLoader
+): DataLoader<AzureDevOpsRemoteProjectKey, AzureDevOpsRemoteProject | never[], string> => {
+  return new DataLoader<AzureDevOpsRemoteProjectKey, AzureDevOpsRemoteProject | never[], string>(
```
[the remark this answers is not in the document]
nickoferrall: Thanks for the explanation, that's helpful! I've updated the type. The `never[]` type happened because the `return results.map(...` line had an empty array in the ternary. I've changed that to `null` and removed the `never[]`

packages/client/components/AzureDevOpsScopingSearchResults.tsx
```diff
@@ -127,22 +131,23 @@ const AzureDevOpsScopingSearchResults = (props: Props) => {
...
-            }}
-            summary={node.title}
-            url={node.url}
-            linkText={`${node.type} #${node.id}`}
-            linkTitle={`Azure DevOps Work Item #${node.id}`}
-          />
-        ))}
+        {edges.map(({node}) => {
+          const {id: workItemId, title, project, url} = node
+          const {issueKey} = AzureDevOpsIssueId.split(workItemId)
```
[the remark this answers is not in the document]
nickoferrall: Makes sense! Updated

packages/server/graphql/types/AzureDevOpsWorkItem.ts
```diff
@@ -14,15 +17,39 @@ const AzureDevOpsWorkItem = new GraphQLObjectType<any, GQLContext>({
   fields: () => ({
     id: {
       type: new GraphQLNonNull(GraphQLID),
-      description: 'GUID instanceId:issueKey'
+      description: 'GUID instanceId:projectKey:issueKey',
```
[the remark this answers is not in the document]
nickoferrall: `projectKey` is being used in https://github.com/ParabolInc/parabol/blob/584cb1d210b8f05dd5b2cd0f2886fa84e7f1afd1/packages/client/shared/gqlIds/AzureDevOpsIssueId.ts#L11 and https://github.com/ParabolInc/parabol/blob/584cb1d210b8f05dd5b2cd0f2886fa84e7f1afd1/packages/server/database/types/TaskIntegrationAzureDevOps.ts#L18 I'll add this to the `teamProject` refactor issue https://github.com/ParabolInc/parabol/issues/7073

packages/server/graphql/types/AzureDevOpsWorkItem.ts
```diff
@@ -23,6 +23,14 @@ const AzureDevOpsWorkItem = new GraphQLObjectType<any, GQLContext>({
         return AzureDevOpsIssueId.join(instanceId, teamProject, id)
       }
     },
+    issueKey: {
+      type: new GraphQLNonNull(GraphQLString),
+      description: 'The id of the issue from Azure, e.g. 7',
+      resolve: async ({id}: {id: string}) => {
+        const {issueKey} = AzureDevOpsIssueId.split(id)
```
mattkrick, line 30: wait a minute... L21 shows that `source.id === `issueKey` and this should that `source.id` is the composite key. gotta test this one out...

mattkrick: ah, the split method was written to return the issueKey if the full ID didn't contain any colons. Neat!
Gonna change this to `return id` for simplicity sake

packages/server/utils/AzureDevOpsServerManager.ts
```diff
@@ -602,6 +628,15 @@ class AzureDevOpsServerManager implements TaskIntegrationManager {
     return {error: firstError, process: result.name}
   }
 
+  async getProject(instanceId: string, projectId: string) {
```
mattkrick, line 631: -1 just to keep this manager looking like other Manager types, let's return `Error | ProjectRes` and let the calling function handle that

mattkrick:
committed directly to branch & merging since Nick is OOO for awhile

Issue #7006, opened 2022-08-05
Azure integration doesn't automatically refresh


nickoferrall:
If I integrate with Azure and leave the app for a day, I need to reintegrate to have access to my Azure integration.

### AC

- [ ] Azure integration automatically refreshes the token

mattkrick:
I'd suggest that all azure queries are fetched from dataloaders & they all first use a dataloader to fetch a fresh access token.
Pattern: https://github.com/ParabolInc/parabol/blob/072cdfcbe5cf1a5eed3a9c72ec1da51756d2b9fe/packages/server/dataloader/atlassianLoaders.ts#L45

PR #7097, opened 2022-08-23
fix: azure refresh auth

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7006

When refreshing the token, Azure returned the error `invalid_grant`. The error description said: 
`The refresh token was issued to a single page app (SPA), and therefore has a fixed, limited lifetime of 1.00:00:00, which cannot be extended. It is now expired and a new sign in request must be sent by the SPA to the sign in page`

This issue https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/4104 and this [Azure doc](https://docs.microsoft.com/en-us/azure/active-directory/develop/refresh-tokens#refresh-token-lifetime) confirm that the user needs to re-auth after 24 hours. 

In this PR, I'm removing the user's stale auth if the refresh token error message is `invalid_grant`. In Sprint Poker, the user will now see Parabol's auth button instead of an empty list of issues. 

### To test

- It's a difficult one to test, but you could integrate with Azure, wait 24 hours, go to a Poker meeting and see that you need to re-auth
- After re-authenticating, you can see the issues

### DIFF
--- a/packages/server/dataloader/azureDevOpsLoaders.ts
+++ b/packages/server/dataloader/azureDevOpsLoaders.ts
@@ -153,7 +153,7 @@ export const freshAzureDevOpsAuth = (
             )
             const oauthRes = await manager.refresh(refreshToken)
             if (oauthRes instanceof Error) {
-              // Azure refresh token only lasts 24 hrs for SPAs. User needs to manually re-auth after that: https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/4104
+              // Azure refresh token only lasts 24 hrs for SPAs. User must manually re-auth after that: https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/4104
               if (oauthRes.message === 'invalid_grant') {
                 await removeTeamMemberIntegrationAuthQuery('azureDevOps', teamId, userId)
               }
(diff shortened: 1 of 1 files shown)


Dschoordsch:
I think that could be a configuration issue, we shouldn't be in the SPA but in the Web App category [^1]. I don't know if that makes a difference, but we should modify our app anyways.

[^1]: https://docs.microsoft.com/en-us/azure/active-directory/develop/authentication-flows-app-scenarios

nickoferrall:
Shouldn't we be in the SPA category though as we're using React? Also, do you happen to know where we can update the configuration in Azure - I haven't come across anything in the [azure portal](https://portal.azure.com/#view/Microsoft_AAD_RegisteredApps/ApplicationMenuBlade/~/Authentication/appId/46eda4b2-e9f2-4c3e-acc0-4e23c09d81f1/isMSAApp~/false) yet

Dschoordsch:
> Shouldn't we be in the SPA category though as we're using React? Also, do you happen to know where we can update the configuration in Azure - I haven't come across anything in the [azure portal](https://portal.azure.com/#view/Microsoft_AAD_RegisteredApps/ApplicationMenuBlade/~/Authentication/appId/46eda4b2-e9f2-4c3e-acc0-4e23c09d81f1/isMSAApp~/false) yet

I thought so too (so did the MS code with folks since they especially implemented OAuth2 with PKCE for it), but reading through the linked page since we're not requesting the token from the browser but a separate backend, we fall into the regular Web App category. This might also give us greater freedom with regards to the refresh token lifetime.
I know where to change it, but it seems not to be a simple replace the SPA with Web app, maybe you could give it a try?

nickoferrall:
I played around with the auth briefly but didn't find a solution. As I'm away from tomorrow, I've created this issue to document it: https://github.com/ParabolInc/parabol/issues/7114

We could move this PR to stuck in the meantime, but I think it's an improvement over what we currently have, so I'll request a review

mattkrick:
:flushed: a 1 day refresh token!?!?

Yep, agreed with the conclusion, we are a "web app" as defined by them.
tbh I can't blame them, when i built the github integration I thought we wanted an oauth2 integration & we really need a "GitHub  App"

mattkrick:
merging since this PR is better than what we currently have. the fix is gonna be moving to a webapp per #7114

Issue #6883, opened 2022-07-14
Standups: Emoji support 


BartoszJarocki:
Somehow we missed the emoji support. We should add it the same way it works currently in our draftjs implementation. 
Existing tiptap extensions: https://tiptap.dev/api/nodes/emoji  (paid). We should figure out what's the best way of adding support (existing vs implementing our own) 

### Acceptance Criteria (optional)
- [ ] figure out the solution existing extensions vs implementing our own
- [ ] create implementation issues

**Estimated effort:** 4h

PR #7102, opened 2022-08-23
feat(tiptap): Emoji Support

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/6883

Add emoji support by typing `:`.  Reuse the emoji menu used in our Draft.js editor to show emojis in TipTap.

Changes:
* Refactor existing emoji menu to decouple the menu from the draft.js editor
* Add a new tiptap emoji component that uses the [Suggestion](https://github.com/ueberdosis/tiptap/tree/main/packages/suggestion) tiptap utility to render the emoji menu.

This uses a pattern used by many TipTap React components in which the component is rendered by simply passing in the editor - no added configuration in the config (see the [React Bubble Menu](https://github.com/ueberdosis/tiptap/blob/9a254bf9daf6d839bd02c968e14837098b903b38/packages/react/src/BubbleMenu.tsx)).

**note**: one test case this fails is when the cursor is part-way through the emoji query - the emoji suggestions are only based on the query from `:` to the cursor, and selecting an emoji will only replace the text from `:` to the cursor.  E.g. if I type out `:joy`, and place my cursor between `o` and `y` and click the `:joy:` emoji, the resultant text will look like `😂y`.  **This is an issue with the tiptap `Suggestion` utility** (see the example on their [Emojis package docs](https://tiptap.dev/api/nodes/emoji)), so we'll need to either PR a fix into that package, or do some custom range selection to fix this. This likely isn't enough of a degradation to prioritize fixing in the near-term, but we should stay attentive to feedback about this.

## Demo
https://www.loom.com/share/fa5cb5c6b84445e9bbbdf877df8c926b

## Testing scenarios
Do the following on both standups and Draft.js:
* Type out part of an emoji by typing `:<emoji`, then click an emoji
* Type out part of an emoji, select elsewhere in the textbox, then select the end of the emoji query, and click an emoji
* Type out part of an emoji, click out of the textbox to lose focus, then select the end of the emoji query, and click an emoji
* Type out part of an emoji, then type a space and see that the menu disappears.

## Final checklist

### DIFF
--- a/packages/client/components/TaskEditor/withEmojis.tsx
+++ b/packages/client/components/TaskEditor/withEmojis.tsx
@@ -1,3 +1,4 @@
+import {EditorState} from 'draft-js'
 import React, {Component} from 'react'
 import {autoCompleteEmoji} from '../../utils/draftjs/completeEntity'
 import getDraftCoords from '../../utils/getDraftCoords'
@@ -10,6 +11,7 @@ const withEmojis = (ComposedComponent) => {
     cachedCoords: any
 
     state = {
+      focusedEditorState: null,
       isOpen: false,
       query: ''
     }
@@ -29,9 +31,22 @@ const withEmojis = (ComposedComponent) => {
 
     menuRef = React.createRef<any>()
 
-    onSelectEmoji = (emoji, editorState) => {
+    static getDerivedStateFromProps(nextProps, prevState) {
+      const {editorState} = nextProps
+      return {
+        // clicking on a menu will cause the editorStateSelection to lose focus, so we persist the last state before that point
+        focusedEditorState: editorState.getSelection().getHasFocus()
+          ? editorState
+          : prevState.focusedEditorState
+      }
+    }
+
+    onSelectEmoji = (emoji) => {
       const {setEditorState} = this.props
-      const nextEditorState = autoCompleteEmoji(editorState, emoji)
+      const nextEditorState = autoCompleteEmoji(
+        this.state.focusedEditorState! as EditorState,
+        emoji
+      )
       setEditorState(nextEditorState)
     }
 
@@ -74,7 +89,6 @@ const withEmojis = (ComposedComponent) => {
 
     renderModal = () => {
       const {query} = this.state
-      const {editorState} = this.props
       this.cachedCoords = getDraftCoords() || this.cachedCoords
       return (
         <EmojiMenuContainer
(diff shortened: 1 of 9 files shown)


packages/client/components/TaskEditor/withEmojis.tsx
```diff
@@ -29,10 +31,22 @@ const withEmojis = (ComposedComponent) => {
...
-      e.preventDefault()
+    static getDerivedStateFromProps(nextProps, prevState) {
+      const {editorState} = nextProps
+      return {
+        // clicking on a menu will cause the editorStateSelection to lose focus, so we persist the last state before that point
+        focusedEditorState: editorState.getSelection().getHasFocus()
+          ? editorState
+          : prevState.focusedEditorState
+      }
+    }
```
jmtaber129, line 42: I'm not sure if this exact logic is necessary, but I've copied this from `withEmojis` to maintain identical behavior

packages/client/components/promptResponse/EmojiMenuTipTap.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const [emojiQuery, setEmojiQuery] = useState('')
+  const [range, setRange] = useState<Range | null>(null)
+
+  useEffect(() => {
+    if (tiptapEditor.isDestroyed) {
+      return
+    }
+
+    const plugin = Suggestion({
+      pluginKey: pluginKey as any,
```
BartoszJarocki, line 26: +1 just curious, why do we need to cast here?

BartoszJarocki (approved):
LGTM! the only improvement I'd love to see is being able to navigate the emoji menu with the keyboard (just like on slack). right now, when I want to select the emoji from the menu, I have to use a mouse, which isn't perfect.

packages/client/components/promptResponse/EmojiMenuTipTap.tsx
jmtaber129: The types between the `PluginKey` I'm using and the `PluginKey` `Suggestion` expects don't quite match up exactly, so we need to cast to work around that.

Added a comment + TODO for this.

jmtaber129:
> I'd love to see is being able to navigate the emoji menu with the keyboard

This would be awesome! Our existing emoji menu on draft.js doesn't seem to support this, but that would be a nice enhancement to take on later.

BartoszJarocki:
yeah, I didn't mean this PR, I'll create a task for it. nice work!

PR #7105, opened 2022-08-24
feat(standups): Starting and stopping recurrence from meeting menu

### LINKED ISSUE #6920: Recurrence MVP: Start+stop recurrence in current meeting view
See designs - https://github.com/ParabolInc/parabol/issues/6747#issuecomment-1176683854

Blocked on https://github.com/ParabolInc/parabol/issues/6916

Support starting+cancelling recurrence from within the current meeting via the options menu.

- [ ] User can start recurrence from the options menu in active meetings
- [ ] User can end recurrence from the options menu in active meetings
- [ ] Option to start recurrence is disabled for ended meetings
- [ ] User can end recurrence from the options menu in ended meetings if no other meetings in the series are active

BartoszJarocki:
# Description

Fixes #6920 

## Demo

https://user-images.githubusercontent.com/1017620/187133852-5d924fa4-71ec-458f-9ee3-c23f8a3700c0.mp4

## Testing scenarios

- [ ] Start recurrence, meeting menu should show 'Stop repeating'
- [ ] Click 'Stop repeating', menu should show 'Repeat M-F'
- [ ] Start recurrence, end the meeting, go back to the meeting, and click 'Stop repeating', menu should show 'Repeat M-F' but it's supposed to be disabled. 

## Final checklist


packages/server/graphql/public/typeDefs/stopRecurrence.graphql
```diff
@@ -17,10 +17,9 @@ union StopRecurrencePayload = ErrorPayload | StopRecurrenceSuccess
 
 type StopRecurrenceSuccess {
   """
-  The updated meeting
-  :TODO: Change this to be any NewMeeting.
+  The updated meeting series
   """
-  meeting: TeamPromptMeeting!
+  meetingSeries: MeetingSeries!
```
BartoszJarocki, line 22: This is the biggest change in this PR. To handle all the cases, I think it makes much more sense to return `MeetingSeries` from `stopRecurrence` instead of just a `Meeting`. The main reason for this is that technically there's a possibility that none of the meetings are active, but we should still be able to disable the recurrence (last AC from the task). In this case, there's no meeting to return, as there are no active meetings. In general, `start/stop` recurrence mutations create/update `MeetingSeries` so I think it's a bit better to structure it that way.

BartoszJarocki:
@jmtaber129 do you mind taking a look? thank you!

packages/client/components/TeamPrompt/TeamPromptOptionsMenu.tsx
```diff
@@ -62,14 +65,17 @@ const TeamPromptOptionsMenu = (props: Props) => {
   const atmosphere = useAtmosphere()
   const {onCompleted, onError} = useMutationProps()
   const {history} = useRouter()
+  const isEnded = !!endedAt
   const hasRecurrenceEnabled = meetingSeries && !meetingSeries.cancelledAt
+  const hasActiveMetings = !!meetingSeries?.activeMeetings?.length
+  const canToggleRecurrence = isEnded ? hasActiveMetings : true
```
jmtaber129, line 71: +1 `!isEnded || hasActiveMeetings`?

packages/server/graphql/public/typeDefs/stopRecurrence.graphql
jmtaber129: I'm not sure I agree with this, since it breaks the symmetry of starting recurrence (action on a meeting) and ending recurrence (now on a meeting series).

> In this case, there's no meeting to return, as there are no active meetings

I'd say the meeting to accept as an argument and then return would just be an ended meeting

packages/server/graphql/types/NewMeeting.ts
```diff
@@ -48,6 +48,14 @@ export const newMeetingFields = () => ({
     type: GraphQLISO8601Type,
     description: 'The timestamp the meeting officially ended'
   },
+  meetingSeriesId: {
+    type: new GraphQLNonNull(GraphQLID),
```
jmtaber129, line 52: -1 This isn't non-null

packages/server/graphql/public/typeDefs/MeetingSeries.graphql
```diff
@@ -41,4 +41,9 @@ type MeetingSeries {
   When the meeting series was cancelled
   """
   cancelledAt: DateTime
+
+  """
+  The meetings that belong to the meeting series
+  """
+  activeMeetings: [NewMeeting!]!
```
jmtaber129, line 48: +1 Since recurrence is currently scoped only to team prompt meetings, we may need to keep this as TeamPromptMeeting for now.

jmtaber129 (changes requested):
(forgot to review this until a little while after my normal work day, so didn't test locally or anything)

We may want to discuss a bit more on what the recurrence API should be. IMO the semantics should either be "start/stop recurrence for a meeting", or "create a recurring meeting series from a meeting, and then stop/restart recurrence for the series". "Start/restart recurrence on a meeting and stop recurrence on a series" feels a bit odd to me

packages/server/graphql/public/typeDefs/stopRecurrence.graphql
BartoszJarocki: yeah, `startRecurrence` would have to be aligned to this. The biggest problem I have with passing `meetingId` in `stopRecurrence` input is the fact it can be used on the already ended meetings. what's more, there might be 0 active meetings, what should we return then, as a result? I think it'd be best to have some sync time and make sure we're on the same page with both PRs, shouldn't take too long. wdyt?

packages/server/graphql/types/NewMeeting.ts
BartoszJarocki: fixed

packages/client/components/TeamPrompt/TeamPromptOptionsMenu.tsx
BartoszJarocki: this won't allow ending the recurrence of ended meeting (!true || false). my condition was wrong as well, and as we use the same button for a start and stop, the condition to disable it is quite complex. 

the correct and a little verbose but easy-to-read version would look like that
```
  const canStartRecurrence = !isEnded
  const canEndRecurrence = !isEnded || !hasActiveMetings
  const canToggleRecurrence = hasRecurrenceEnabled ? canEndRecurrence : canStartRecurrence
```

wdyt?

packages/server/graphql/public/typeDefs/MeetingSeries.graphql
BartoszJarocki: as now we have recurrence-related properties in a `NewMeeting` i think it's ok to leave it as is and avoid unnecessary casts. once MVP is done, I'd like to create a recurrence spec discussion task to move us forward

BartoszJarocki:
@jmtaber129 would you mind taking a look in the meantime? would love to push recurrence forward!

jmtaber129 (changes requested):
Mostly looks good, but we should allow stopping recurrence even if there are no active meetings. Otherwise a user would have to wait for a recurring meeting to start in order to end recurrence for the series.

packages/client/components/TeamPrompt/TeamPromptOptionsMenu.tsx
```diff
@@ -62,14 +65,19 @@ const TeamPromptOptionsMenu = (props: Props) => {
   const atmosphere = useAtmosphere()
   const {onCompleted, onError} = useMutationProps()
   const {history} = useRouter()
+  const isEnded = !!endedAt
   const hasRecurrenceEnabled = meetingSeries && !meetingSeries.cancelledAt
+  const hasActiveMeetings = !!meetingSeries?.activeMeetings?.length
+  const canStartRecurrence = !isEnded
+  const canEndRecurrence = !isEnded || !hasActiveMeetings
```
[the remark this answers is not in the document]
BartoszJarocki: Seems like you've misunderstood this conditional. `canEndRecurrence` is `true` when the meeting is not ended or if there are no other active meetings, so it works exactly how you described. See the scenario below:
1. I open a Standup meeting
2. I start the recurrence
3. I end the meeting (there are no other active meetings in the series)
4. I go back to the meeting
5. I stop the recurrence


https://user-images.githubusercontent.com/1017620/192195074-970eaf32-d8de-4f95-aa27-85dabf1acf28.mp4

considering this, would you mind taking a look again? thanks!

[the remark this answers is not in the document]
jmtaber129: Ahh yea you're right I totally misread this 🤦

jmtaber129 (approved):

packages/client/components/TeamPrompt/TeamPromptOptionsMenu.tsx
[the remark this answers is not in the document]
Dschoordsch: You could add a comment why you don't want to cancel recurrence from an ended meeting while there are active ones still open.

packages/server/graphql/public/typeDefs/stopRecurrence.graphql
Dschoordsch: I think ending on a meeting series is a safe bet. I can imagine that once we extend this to other meeting types we might want to have an overview for all meeting series and then be able to start/stop them.

Starting a meeting series should take a meeting as argument to take as a prototype to know which structure it has (i.e. which template to use etc.).

Restarting should take again a meeting series as that's the object we're mutating.

packages/server/graphql/public/types/StopRecurrenceSuccess.ts
```diff
@@ -1,13 +1,14 @@
...
 export type StopRecurrenceSuccessSource = {
-  meetingId: string
+  meetingSeriesId: string
 }
 
 const StopRecurrenceSuccess: StopRecurrenceSuccessResolvers = {
-  meeting: async ({meetingId}, _args, {dataLoader}) => {
-    return dataLoader.get('newMeetings').load(meetingId) as Promise<MeetingTeamPrompt>
+  meetingSeries: async ({meetingSeriesId}, _args, {dataLoader}) => {
+    const id = MeetingSeriesId.split(meetingSeriesId)
```
Dschoordsch, line 10: +1 I don't know what you've discussed before with regards to GraphQL IDs, but I personally would return database id's in the source objects from the mutations and do the ID conversion in the GraphQLObject resolvers as I perceive these as a conversion layer.

Dschoordsch (approved):
Did not test

packages/server/graphql/public/types/StopRecurrenceSuccess.ts
BartoszJarocki: very true, good catch!

PR #7111, opened 2022-08-25
chore: add pull-request-stats github workflow

JimmyLv:
# Description

Google Chrome-LinearB Engineering Operational Excellence-000005-20220825

According to the [Cycle Time](https://en.wikipedia.org/wiki/Cycle_time_(software)) statistics from [LinearB](https://linearb.io/), we're spending a longer time indeed in the Review phase, so let's figure out how to speed it up, and the first step might be to show more detailed Review data:
This PR adds usage of [pull-request-stats workflow](https://github.com/flowwer-dev/pull-request-stats), use this Github Action to print relevant stats about Pull Request reviewers. Let's continuously measure and do more improvements!

The objective of this action is to:

- Reduce the time taken to review the pull requests.
- Encourage quality on reviews.
- Help deciding which people to assign as reviewers.

Please feel free to comment, let's just see how it works in this PR, please let me know what you think!

## Demo

image

### DIFF
--- a/.github/workflows/stats.yml
+++ b/.github/workflows/stats.yml
@@ -12,4 +12,4 @@ jobs:
         uses: flowwer-dev/pull-request-stats@master
         with:
           charts: true
-          sort-by: 'COMMENTS'
+          sort-by: 'REVIEWS'
--- a/.github/workflows/stats.yml
+++ b/.github/workflows/stats.yml
@@ -13,4 +13,4 @@ jobs:
         uses: flowwer-dev/pull-request-stats@master
         with:
           charts: true
-          sort-by: 'REVIEWS'
+          sort-by: 'TIME'
--- a/.github/workflows/stats.yml
+++ b/.github/workflows/stats.yml
@@ -1,16 +1,17 @@
 name: Pull Request Stats
 
 on:
-  pull_request:
-    types: [labeled]
+  schedule:
+    - cron: '0 * * * 5'
 
 jobs:
   stats:
-    if: ${{ github.event.label.name == 'pr_stats' }}
     runs-on: ubuntu-latest
     steps:
       - name: Run pull request stats
         uses: flowwer-dev/pull-request-stats@master
         with:
+          slack-channel: '#t_product_github_notifications'
+          slack-webhook: ${{ secrets.SLACK_WEBHOOK }}
           charts: true
           sort-by: 'TIME'
--- a/.github/workflows/stats.yml
+++ b/.github/workflows/stats.yml
@@ -2,10 +2,11 @@ name: Pull Request Stats
 
 on:
   pull_request:
-    types: [opened]
+    types: [labeled]
 
 jobs:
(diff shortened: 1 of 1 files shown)


Dschoordsch (changes requested):
Statistic looks nice, but I don't want to see this on every PR. Can we have this on a separate site?

JimmyLv:
> Statistic looks nice, but I don't want to see this on every PR.

Thanks for such quick feedback! It's a good point and one that's worth discussing.

I think the goal of this statistic is to ultimately shorten Cycle Time, to encourage Reviewers to do it faster, and to encourage the quality of reviews; Reviewers are also PR creators, so they can objectively see the process of each Reviewer (including themselves) when submitting a PR.

So it's a matter of "timing", and if it's put on a separate page, my concern is that people won't have as much of a chance to look at it, and thus the motivation will be less effective?

However, in the meantime, the other thing I'm considering: Will this put **_pressure_** on our process of doing a Review? 
I'm not sure, it might be worth trying, let me post to the `#engineering` channel to hear what others think ☺️

JimmyLv:
Thanks @BartoszJarocki for the feedback, yeah, I agree and support the idea that PR Review can't be a competition. 
Seeing the stats weekly or monthly, need to find a place to generate it, but if not do Pull Request inside, then I can't figure out where. 🤣 

I feel like we can make it optional, through PR's Label, like when manually adding `"pr_stats"` label, this GitHub Action Bot 🤖 will add a comment.
 From there it can become more flexible, and following this pattern, we can also add more bots to PR later.

Dschoordsch:
I would like to see those stats in Slack once a week in fridayship or at the end of every sprint before the retro. Having these every Friday is probably easier.

nickoferrall:
Feels safe to try! If the aim is to increase cycle time, perhaps we should sort by "Time to review" rather than "Total comments"?

JimmyLv:
@nickoferrall Yeah, I've updated before for ["TIME"](https://github.com/flowwer-dev/pull-request-stats#action-inputs) as a sorting option. But it seems that the Bot recognizes the existing stats comment and stops regenerating again, so let me delete the comment and try running workflow again. 😂

JimmyLv:
@Dschoordsch After reconfirming, the [Slack integration](https://github.com/flowwer-dev/pull-request-stats#slack-integration) requires monthly sponsorship (min $20 usd / month) [^1], so it's not available to integrate with Slack to send weekly reports for now. 🤣

> we can make it optional, through PR's Label, like when manually adding the "pr_stats" label, this GitHub Action Bot 🤖 will add a comment.

So, I'd still recommend making this workflow optional, and I've changed the relevant configuration in the `yml` file. So that we can look at it manually when we want to, and then I'll make it automated later if necessary. 😁

[^1]: https://github.com/flowwer-dev/pull-request-stats#premium-features-

Dschoordsch:
Let's try it with Slack.

PR #7118, opened 2022-08-26
Client-side mentions

jmtaber129:
# Description

## Demo

## Testing scenarios

- [ ] Scenario A
  - Step 1
  - Step 2...

- [ ] Scenario B
  - Step 1
  - Step 2....

## Final checklist


packages/server/graphql/queries/notifications.ts
```diff
@@ -31,7 +31,7 @@ export default {
       .limit(first + 1)
       .run()
 
-    const nodes = nodesPlus1.slice(0, -1)
+    const nodes = nodesPlus1.slice(0, first)
```
jmtaber129, line 34: This is a bug - previously, we were looking up the first 6 notifications, then dropping the last element to get down to 5, then using the original to figure out some pagination.  But if a user has fewer than 5 notifications, we'll drop the last one.

jmtaber129:
@tianrunhe not sure whether we want to add tracking for mention notifications so I left them out for now, but let me know if you think they're worth including

tianrunhe:
> @tianrunhe not sure whether we want to add tracking for mention notifications so I left them out for now, but let me know if you think they're worth including

Thanks for tagging! Sounds good to ignore atm. No much tension here!

packages/client/components/promptResponse/PromptResponseEditor.tsx
```diff
@@ -102,14 +109,15 @@ const StyledEditor = styled('div')`
 
 interface Props {
   autoFocus?: boolean
+  teamId?: string
```
BartoszJarocki, line 112: +1 is there any case where team id would not be defined?

packages/server/graphql/public/typeDefs/NotifyResponseMentioned.graphql
```diff
@@ -0,0 +1,42 @@
...
+  status: NotificationStatusEnum!
+
+  """
+  The datetime to activate the notification & send it to the client
+  """
+  createdAt: DateTime!
+  type: NotificationEnum!
+
+  """
+  *The userId that should see this notification
```
BartoszJarocki, line 19: +1 lets remove `*`

BartoszJarocki (approved):
lgtm!

packages/client/components/promptResponse/PromptResponseEditor.tsx
jmtaber129: Not necessarily, but this is only required for editing, so it seems wrong to make it required for cases when it's read-only (e.g. in the discussion drawer or summary)

packages/client/components/ResponseMentioned.tsx
```diff
@@ -0,0 +1,54 @@
...
+  )
+  const {history} = useRouter()
+  const {meeting, response} = notification
+  const {picture: authorPicture, preferredName: authorName} = response.user
+
+  const {id: meetingId, name: meetingName} = meeting
+  const goThere = () => {
+    // :TODO: (jmtaber129): Link directly to card once we support that.
+    history.push(`/meet/${meetingId}`)
+  }
```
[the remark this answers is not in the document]
jmtaber129: Ooo yea, you're totally right. Might be a worthy refactor later on.

packages/server/graphql/public/mutations/helpers/publishTeamPromptMentions.ts
```diff
@@ -0,0 +1,54 @@
+import {JSONContent} from '@tiptap/core'
+import getRethink from '../../../../database/rethinkDriver'
+import NotificationResponseMentioned from '../../../../database/types/NotificationResponseMentioned'
+import {TeamPromptResponse} from '../../../../postgres/queries/getTeamPromptResponsesByIds'
+
+const getMentionedUserIdsFromContent = (content: JSONContent): string[] => {
+  if (content.type === 'mention' && content.attrs?.id) {
+    return [content.attrs?.id]
```
mattkrick, line 8: +1 can remove optional chaining operator, the conditional above already narrowed the type

packages/server/graphql/public/mutations/helpers/publishTeamPromptMentions.ts
```diff
@@ -0,0 +1,54 @@
...
+  }
+
+  if (!content.content) {
+    return []
+  }
+
+  return content.content.map(getMentionedUserIdsFromContent).flat()
+}
+
+const publishTeamPromptMentionNotifications = async (
```
mattkrick, line 18: +1 this function isn't actually calling `publish`, maybe call it `insert`?

packages/server/graphql/public/types/NotifyResponseMentioned.ts
```diff
@@ -0,0 +1,20 @@
+import NotificationResponseMentioned from '../../../database/types/NotificationResponseMentioned'
+import {NotifyResponseMentionedResolvers, ResolversTypes} from '../resolverTypes'
+
+export type NotifyResponseMentionedSource = Pick<
```
mattkrick, line 4: -1 can remove & in codegen.json reference the database type directly like https://github.com/ParabolInc/parabol/blob/6fae361e2fa77b7c21e33100c60621303b236ac0/codegen.json#L50

unless i'm missing something?

packages/server/graphql/queries/notifications.ts
mattkrick: +2 i'm a moron

mattkrick (changes requested):
looks GREAT! only big question is if we could borrow more patterns from the tiptap example repo & clean up 2 of the components

packages/client/components/EditorSuggestions/EditorSuggestions.tsx
```diff
@@ -25,7 +25,7 @@ const MentionMenu = styled('div')({
 
 interface Props {
   active: number
-  handleSelect: (idx: number) => (e: React.MouseEvent) => void
+  handleSelect: (item: DraftSuggestion) => void
```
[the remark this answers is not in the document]
jmtaber129: Hmm I'm a little confused:
1. How does changing to `BaseSuggestion` allow us to now case the `handleSelect` function of `SuggestMentionableUsers.tsx`?
2. Doesn't this make for _less_ type safety in context? I.e. if this is only dealing with `BaseSuggestion`s, that means this could call `handleSelect` with a `BaseSuggestion` that _doesn't_ represent a `MentionSuggestion` or `TagSuggestion`. So even if we're maintaining some state about whether the suggestion passed to the select handler should be a mention or a tag, there's no longer a guarantee that it's either.

I think if this was all more object-oriented then making these types the base class would make sense, but not with how this is currently written.

packages/server/graphql/public/mutations/helpers/publishTeamPromptMentions.ts
jmtaber129: Renamed to `create`

packages/client/components/promptResponse/MentionsTipTap.tsx
```diff
@@ -0,0 +1,156 @@
...
+          }
+        ])
+        .run()
+    },
+    [tiptapEditor, range]
+  )
+
+  const keyHandlerRef = useRef<any>(null)
+
+  // :HACK: We can't just pass these into the 'Suggestion' plugin because these change based on
```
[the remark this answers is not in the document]
jmtaber129: Done!

jmtaber129:
@mattkrick I resolved most of your comments. The only other question is around `DraftSuggestion` vs. `BaseSuggestion`

packages/client/components/EditorSuggestions/EditorSuggestions.tsx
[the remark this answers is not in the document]
mattkrick: 1. gah i did a shit job of explaining.

Here's the current scenario:
- prop type: `handleSelect: (item: DraftSuggestion) => void`
- the above breaks down into: `handleSelect: (item: MentionSuggestion | TagSuggestion) => void`
-  passed in value: `(item: MentionSuggestion) => void`

The reason why typescript gets angry at this is because `EditorSuggestions` wants a function that can handle `MentionSuggestion | TagSuggestion`. `SuggestMentionableUsers` is only passing in a function that can handle `MentionSuggestions`. In other words, it's passing in something that can only fulfill half the required contract... and that's true! the function being passed in cannot handle a `TagSuggestion`. In _most_ code, this control flow would be inverted, where the thing being passed in can handle _more_ possibilities than the thing requiring it (e.g. "I can handle Mentions & Tags, but my child only needs me to handle Mentions").

So, we have 2 options: 

- we can either say "yes, it can in fact handle `TagSuggestion`, which is what the casting says (although from a pedantic standpoint that's not really true)
- or we can say "I'll accept any type of suggestion, I'm not going to discriminate based on the different types", which is what using the `BaseSuggestion` says

There's a word for this in type systems & I can't for the life of me remember what it is... but I hope this kinda helps! 

2. 100%! it does offer less type safety, The casting put in place says "I can handle Mentions or Tags", which isn't true, that function doesn't even know Tags are a thing! Using BaseSuggestion says  "I can handle _any_ Suggestion, but I'll treat them all the same". Less type safe, but more accurate. 

you're totally on point about OOP, we're writing functional code using a prototypical language with an OOP type system. it's a wild ride :sweat_smile:

packages/client/components/EditorSuggestions/EditorSuggestions.tsx
```diff
@@ -25,7 +25,7 @@ const MentionMenu = styled('div')({
...
   originCoords: BBox
   suggestions: DraftSuggestion[]
```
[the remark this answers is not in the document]
jmtaber129: I'm not sure this is true - if the `suggestion` is of type `BaseSuggestion`, it'll still be missing the types that may be needed for SuggestionItem

packages/client/components/EditorSuggestions/EditorSuggestions.tsx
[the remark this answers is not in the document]
jmtaber129: Okay, I think I understand a little better now. Still, in practice, we'd need to downcast from `BaseSuggestion` to `MentionSuggestion` in the handler. Instead of simply changing this to `BaseSuggestion`, I added generics (`T extends BaseSuggestion`) so passing in a bunch of `MentionSuggestion`s will mean the `handleSelect` gets called with a `MentionSuggestion`.

packages/client/components/promptResponse/MentionsTipTap.tsx
```diff
@@ -0,0 +1,156 @@
...
+            keyHandlerRef.current.upHandler()
+            return true
+          }
+
+          if (event.key === 'ArrowDown') {
+            keyHandlerRef.current.downHandler()
+            return true
+          }
+
+          if (event.key === 'Enter') {
```
mattkrick, line 125: -1 to match Draft's behavior can we add `|| 'Tab'`?

mattkrick (changes requested):

PR #7147, opened 2022-09-02
feat: show users who added a reactji

BartoszJarocki:
# Description

Fixes no issue, I just think we need that! 

## Demo

Hover over a reacji
Screenshot 2022-09-28 at 17 39 29

## Testing scenarios

- [ ] open any discussion thread, add emoji, hover over it, and see if a tooltip appears
- [ ] sign in as another user, add another emoji and see if hovering over it from another browser shows the correct user
- [ ] open standup meeting and repeat the scenarios above

## Final checklist

### DIFF
--- a/packages/client/components/ReflectionCard/EmojiUsersReaction.tsx
+++ b/packages/client/components/ReflectionCard/EmojiUsersReaction.tsx
@@ -3,6 +3,7 @@ import graphql from 'babel-plugin-relay/macro'
 import React from 'react'
 import {useFragment} from 'react-relay'
 import {EmojiUsersReaction_reactji$key} from '~/__generated__/EmojiUsersReaction_reactji.graphql'
+import {PALETTE} from '../../styles/paletteV3'
 
 const EmojiUsersReactionRoot = styled('div')({
   display: 'flex',
@@ -17,6 +18,10 @@ const EmojiUsersReactionRoot = styled('div')({
   whiteSpace: 'normal'
 })
 
+const DarkerGrayPart = styled('span')({
+  color: PALETTE.SLATE_400
+})
+
 const LIST_FORMATTER =
   Intl.ListFormat !== undefined
     ? new Intl.ListFormat('en', {style: 'long', type: 'conjunction'})
@@ -24,9 +29,10 @@ const LIST_FORMATTER =
 
 interface Props {
   reactjiRef: EmojiUsersReaction_reactji$key
+  reactjiShortName?: string
 }
 
-const EmojiUsersReaction = ({reactjiRef}: Props) => {
+const EmojiUsersReaction = ({reactjiRef, reactjiShortName}: Props) => {
   const reactji = useFragment(
     graphql`
       fragment EmojiUsersReaction_reactji on Reactji {
@@ -43,6 +49,7 @@ const EmojiUsersReaction = ({reactjiRef}: Props) => {
   return (
     <EmojiUsersReactionRoot>
       {LIST_FORMATTER.format(reactji.users.map(({preferredName}) => preferredName))}
+      {reactjiShortName && <DarkerGrayPart>reacted with :{reactjiShortName}:</DarkerGrayPart>}
     </EmojiUsersReactionRoot>
   )
 }
--- a/packages/client/components/ReflectionCard/EmojiUsersReaction.tsx
+++ b/packages/client/components/ReflectionCard/EmojiUsersReaction.tsx
@@ -4,7 +4,7 @@ import React from 'react'
 import {useFragment} from 'react-relay'
 import {EmojiUsersReaction_reactji$key} from '~/__generated__/EmojiUsersReaction_reactji.graphql'
 
-const EmojiUsersReactionParent = styled('div')({
+const EmojiUsersReactionRoot = styled('div')({
   display: 'flex',
   flexDirection: 'column',
   justifyContent: 'center',
@@ -42,9 +42,9 @@ const EmojiUsersReaction = ({reactjiRef}: Props) => {
   )
 
   return (
-    <EmojiUsersReactionParent>
+    <EmojiUsersReactionRoot>
       {LIST_FORMATTER.format(reactji.users.map(({preferredName}) => preferredName))}
-    </EmojiUsersReactionParent>
+    </EmojiUsersReactionRoot>
   )
 }
 
--- a/packages/client/components/ReflectionCard/EmojiUsersReaction.tsx
+++ b/packages/client/components/ReflectionCard/EmojiUsersReaction.tsx
@@ -1,0 +1,74 @@
+import styled from '@emotion/styled'
+import graphql from 'babel-plugin-relay/macro'
+import React from 'react'
+import {useFragment} from 'react-relay'
+import {EmojiUsersReaction_reactji$key} from '~/__generated__/EmojiUsersReaction_reactji.graphql'
+
+//TODO: remove after migrating to es2021 - https://github.com/microsoft/TypeScript/issues/46907
+declare namespace Intl {
+  type ListType = 'conjunction' | 'disjunction'
+
+  interface ListFormatOptions {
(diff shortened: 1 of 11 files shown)


enriquesanchez:
@BartoszJarocki This is nice!

I have one request and one question:

Request: Can the hover area include the reactji number? Currently the tooltip only appears when hovering the reactji, but if you hover the number count nothing happens. I assume the hover action can be set to the whole `button` element?

https://user-images.githubusercontent.com/3276087/191547624-9609979e-b972-4850-816d-b3566469b3ce.mp4

Question: How does it behave when there are a lot of user names in the tooltip? Is there a max number of users? Does it wrap?

BartoszJarocki:
@enriquesanchez thank you for looking into it! 

>Can the hover area include the reactji number? Currently the tooltip only appears when hovering the reactji, but if you hover the number count nothing happens. I assume the hover action can be set to the whole button element?

for sure!

>Question: How does it behave when there are a lot of user names in the tooltip? Is there a max number of users? Does it wrap?
well, it'll probably grow horizontally right now, which isn't perfect. do you have any ideas on how we can make the UI better? 🙏

enriquesanchez:
I like what Slack does, where once it reaches a specific width (800px), it starts to wrap and grow vertically:

Captura de Pantalla 2022-09-21 a la(s) 1 49 30 p m

But I still think we should plan for extreme cases, for example, what would happen if 50 people react with the same reactji? What about 100?

This feels ugly, but setting a max-height and adding scrolling would solve it. What do you think? 

For our case, maybe something like 600px wide by 300px tall would work best?

BartoszJarocki:
@enriquesanchez thank you! here's what I came up with
Screenshot 2022-09-22 at 16 55 35
wdyt?

enriquesanchez:
@BartoszJarocki Nice! I think we can get rid of the emoji in this tooltip, no need to repeat it and it will let us keep it small. How does that sound?

BartoszJarocki:
@enriquesanchez no strong opinions either way. I borrowed the design from slack 😅

BartoszJarocki:
> This PR exceeds the recommended size of 1000 lines. Please make sure you are NOT addressing multiple issues with one PR. Note this PR will be delayed and might be rejected due to its size.

hey, be fair and stop counting yarn.lock lines

BartoszJarocki:
@nickoferrall could you please take a look at this one? thank you!

packages/client/components/ReflectionCard/ReactjiCount.tsx
```diff
@@ -51,15 +54,38 @@ const Count = styled('div')({
...
+    MenuPosition.UPPER_CENTER,
+    {delay: SHOW_REACTJI_USERS_DELAY}
+  )
+  const reactji = useFragment(
+    graphql`
+      fragment ReactjiCount_reactji on Reactji {
+        id
+        count
+        isViewerReactji
+        users {
```
nickoferrall, line 78: -1 looks like we're not using `users` here so it's safe to remove

packages/client/components/ReflectionCard/ReactjiCount.tsx
```diff
@@ -51,15 +54,38 @@ const Count = styled('div')({
   paddingLeft: 4
 })
 
+export const SHOW_REACTJI_USERS_DELAY = 100 // ms
```
nickoferrall, line 57: +1 could add this to `constEnums` but I don't think it matters too much https://github.com/ParabolInc/parabol/blob/26cb10c809739921f3c65e04d36ddd6e238585fb/packages/client/types/constEnums.ts#L398

packages/server/graphql/types/Reactji.ts
```diff
@@ -20,6 +27,13 @@ const Reactji = new GraphQLObjectType<ReactjiDB, GQLContext>({
       type: new GraphQLNonNull(GraphQLInt),
       description: 'The number of users who have added this reactji'
     },
+    users: {
+      type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(require('./User').default))),
+      description: 'The users who added a reactji',
+      resolve: (parent, _args: any, {dataLoader}) => {
```
nickoferrall, line 33: +1 could destructure `parent` to `{userId`} and add the type, then change `_args` from `any` to `unknown`

packages/client/components/ReflectionCard/EmojiUsersReaction.tsx
```diff
@@ -0,0 +1,51 @@
...
+interface Props {
+  reactjiRef: EmojiUsersReaction_reactji$key
+}
+
+const EmojiUsersReaction = ({reactjiRef}: Props) => {
+  const reactji = useFragment(
+    graphql`
+      fragment EmojiUsersReaction_reactji on Reactji {
+        id
+        users {
```
nickoferrall, line 34: -2 looks like the demo will need to be updated. It crashes for me as it doesn't have any reactji users

packages/client/components/ReflectionCard/EmojiUsersReaction.tsx
```diff
@@ -0,0 +1,51 @@
...
+          picture
+        }
+      }
+    `,
+    reactjiRef
+  )
+
+  return (
+    <EmojiUsersReactionParent>
+      {LIST_FORMATTER.format(reactji.users.map(({preferredName}) => preferredName))}
```
nickoferrall, line 46: +1 I'd love to see the reactji type, e.g. `:heart:`, here as well as the user's name. When I see the name of the emoji, it helps me search for the emoji next time with my keyboard rather than needing to scroll to find it 

reactji

nickoferrall (changes requested):
I love this feature! Excited to see it in prod

packages/client/components/ReflectionCard/EmojiUsersReaction.tsx
BartoszJarocki: good catch! I updated the demo but seems like it didn't like the unused fragment query. I removed it and it's all good on my end. Could you please double check? 🙏

packages/client/components/ReflectionCard/EmojiUsersReaction.tsx
BartoszJarocki: @nickoferrall good idea! see below an improved copy
Screenshot 2022-10-03 at 16 04 48

packages/client/components/ReflectionCard/ReactjiCount.tsx
BartoszJarocki: moved const to `Times`, although I think it's better when the definition is as close as possible to the place of usage. not a big deal, let's stick to the current pattern!

nickoferrall (approved):
LGTM!

packages/client/components/ReflectionCard/EmojiUsersReaction.tsx
```diff
@@ -0,0 +1,57 @@
...
+          preferredName
+        }
+      }
+    `,
+    reactjiRef
+  )
+
+  return (
+    <EmojiUsersReactionRoot>
+      {LIST_FORMATTER.format(reactji.users.map(({preferredName}) => preferredName))}
```
Dschoordsch, line 51: +1 Possible improvement: I like how Slack etc. replace the viewers name with "You".
Screenshot from 2022-10-04 17-16-12

packages/client/mutations/AddReactjiToReactableMutation.ts
```diff
@@ -14,6 +14,10 @@ graphql`
         id
```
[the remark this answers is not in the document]
BartoszJarocki: done!

packages/client/components/ReflectionCard/EmojiUsersReaction.tsx
BartoszJarocki: good suggestion! for now, I'd want to avoid introducing additional complexity and land it as is. I'll create a task for that!

Dschoordsch (approved):
Did not test again

PR #7155, opened 2022-09-06
feat: translation

### LINKED ISSUE #7103: i18n PoC
As we might try to do a PoC soon here are some things to consider.

## Library
There are 2 main libraries out there [react-i18next](https://github.com/i18next/react-i18next) and [react-intl](https://github.com/formatjs/formatjs). The former is relative easy to use with adding only little clutter[^1] while the latter seems to require quite a bit more boilerplate for every string[^2]. Having also worked with react-i18next and not ran into major issues, we should use this library going forward.
Supposedly i18next also works on server side, which would allow us to localize the summaries [^3]

[^1]: https://react.i18next.com/getting-started#simple-content
[^2]: https://formatjs.io/docs/react-intl/#formatting-data
[^3]: https://www.i18next.com/

## Translatable texts
There are multiple areas where we use text which would need to be translated
1) **Texts used directly in the app**
  This would be the easiest to translate and apart from the library and translations only requires some small layout tweaks and probably a UI to choose the language. This should also be the first step when trying a PoC
2) **Meeting summary**
  Once a user can choose a language, it should be stored on the `User` table and also used for summaries. This could be a stretch goal.
3) **Error messages**
  Error messages are strings on the server side. These should not be translated on the server side, but since we have a limited number of fixed error messages we show, we can translate those on client side.
4) **Retro and Poker templates**
  This would require changing the data structure so that we can query templates for a certain language and probably fall back to English if a template is not available as translation. This can be worked around by the users themselves by creating their own templates
5) **Marketing page and marketing emails**
  This is out of scope apart maybe of marking the users preferred language in HubSpot

## Acceptance criteria for PoC
* [ ] language is detected by browser locale and a UI element allows switching language
* [ ] language setting is stored in `User` table
* [ ] strings in app are translated

## Acceptance criteria production
* [ ] PoC
* [ ] estimation of long term effects on development effort taking into consideration the number of translations and members speaking those languages

Dschoordsch:
# Description

Fixes #7103 

Added a jscodeshift template to extract a lot of strings from existing component and wrap them in i18next translation component. The browser locale is used to switch between 'en', 'es' and 'de' at the moment. If translations are missing, it falls back to 'en'.

The branch is based on replacing the icon font with SVG icons to avoid extracting those texts as well.

## Demo

https://www.loom.com/share/0ebeb7d442e4449f88dcd1241077b5bb

## Testing scenarios

- [ ] Scenario A
  - Step 1
  - Step 2...

- [ ] Scenario B
  - Step 1
  - Step 2....

## Final checklist


Dschoordsch:
Terribly out of date and not a priority atm. If we want to move forward with this we should base it on top of https://github.com/BartoszJarocki/jscodeshift-react-i18next and fix issues there.

PR #7181, opened 2022-09-14
feat: allow meeting to be reset to the reflect phase

### LINKED ISSUE #4881: people accidentally advancing through retro phases then can't go back
A [user](https://app.hubspot.com/contacts/3888472/contact/10336201) 🔒  writes: 

> Since her team is not used to the tool yet, they often advance to the next phase too soon and then are unable to go back. I believe what they do is that the facilitator clicks “Next” assuming he is just doing it for themself, without being aware they will be forcing everyone to move to the next retrospective phase right away, and then I guess they can’t go back.

igorlesnenko:
# Description

Fixes #4881

Allows retro meeting to be reset to the reflect phase

## Demo

Loom: https://www.loom.com/share/edf4861480be42d18c25a2d25411fb57

image

image

## Testing scenarios

- Create a meeting with at least 2 participants
- Add reflections for both participants, go to group phase, create some groups
- Go back to reflect phase, reset meeting
- Check that meeting was correctly reset for the second participant in real time
- Check that all meeting functionality works correctly for both participants and you can go through all phases and finish the meeting.
- Try to reset meeting from discussion phase or vote phase too

### DIFF
--- a/packages/client/components/RetroReflectPhase/RetroReflectPhase.tsx
+++ b/packages/client/components/RetroReflectPhase/RetroReflectPhase.tsx
@@ -46,7 +46,7 @@ const RetroReflectPhase = (props: Props) => {
           </PhaseHeaderDescription>
         </MeetingTopBar>
         <PhaseWrapper>
-          <StageTimerDisplay meeting={meeting} />
+          <StageTimerDisplay meeting={meeting} canUndo={true} />
           <ColumnWrapper
             setActiveIdx={setActiveIdx}
             activeIdx={activeIdx}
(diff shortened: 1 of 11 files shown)


mattkrick:
My .02:
- From a psychological standpoint, when a meeting goes backwards instead of forwards, it feels bad. Speaking for myself, whenever the person leading a meeting says "let's go back and redo..." I disengage. The root cause here is that the facilitator moved forward before the team was ready. Preventing them from making a mistake is always preferable to allowing them to undo the mistake!
- Why does the facilitator move forward before everybody is ready? If everyone else hasn't clicked the "Ready" button, then when the facilitator clicks "next" they get a tooltip and have to click "next" again to move forward. _This_ is the area that's broken! Is the tooltip unclear? How could we make the words more meaningful? Do we say "Is everyone ready? If so click next again". Should we specify that if they move forward they can't move back? Do they know that the team can still add reflections in the group phase?

Issue #7179, opened 2022-09-14
Add auth section to user profile


nickoferrall:
See Notion for more context: https://www.notion.so/parabol/Settings-Admin-IA-7d6e7e8fecf84a72a15d4ef7c7dab555#078753a8797648cb8c3f3b98d50d5741

Figma designs: https://www.figma.com/file/2Nj1y1KUfE2swHZuDY5Zxn/My-User-Account-Management?node-id=110%3A1520

### AC

- [ ] Add an Authentication panel to the profile page which includes a password reset link
- [ ] The Authentication page only shows up for users with a username/password. It doesn't show up for users using Google
- [ ] Update the user menu to say "My Settings" instead of "Profile"
- [ ] In profile, "My information" is changed to "Profile"
- [ ] We drop the all-caps to title casing in the admin section cards
- [ ] When clicking to send the reset password link, there’s a confirmation state

PR #7183, opened 2022-09-15
feat: add password reset to user profile

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7179

https://www.loom.com/share/36eaa7d6701b4ab19161012032deda5e

This PR also includes an error state that gives the user feedback if there's a problem sending the password reset.

### To test

- [ ] Make sure that you have added the Mailgun env variables from https://github.com/ParabolInc/action-devops/blob/4aa09acb2ddc8376aaabd11bfc5bcfd7380b9bf8/environments/local#L41-L43
- [ ]  Add an Authentication panel to the profile page which includes a password reset link
- [ ]  The Authentication page only shows up for users with a username/password. It doesn't show up for users using Google
- [ ]  Update the user menu to say "My Settings" instead of "Profile"
- [ ]  In profile, "My information" is changed to "Profile"
- [ ]  We drop the all-caps to title casing in the admin section cards
- [ ]  When clicking to send the reset password link, there’s a confirmation state


packages/client/components/StandardHubUserMenu.tsx
```diff
@@ -67,7 +67,7 @@ const StandardHubUserMenu = (props: Props) => {
         label={
           <MenuItemLink to={'/me/profile'}>
```
nickoferrall, line 68: I haven't changed the URL, so this will still be `/me/profile` unless we want to change it to `/me/settings`?

nickoferrall:
Hey @igorlesnenko, would you mind reviewing this as Terry is away for the rest of the week?

igorlesnenko (changes requested):
When I copying and pasting a reset link I see the following error:

image

Works after refresh.

Does it work for you?

packages/client/components/StandardHubUserMenu.tsx
igorlesnenko: I think it is good to keep old link or add a redirect

nickoferrall:
@igorlesnenko I haven't been able to reproduce that. Does it happen when you click the reset password link in the email?

igorlesnenko:
@nickoferrall to reproduce:

- Open incognito window and login
- Request password reset link
- Open a new incognito tab
- Copy and paste password reset link to this tab

Does it work for you?

However, I don't see any changes to existing password reset link, so at least this PR is safe to merge

igorlesnenko (approved):

nickoferrall:
I followed those steps in an incognito window, but I didn't get the error 🤔  Might be worth creating an issue if you can consistently reproduce it? 

Thanks for the review! I agree that no changes have been made to the link so it feels safe to merge

Issue #7178, opened 2022-09-14
Unsubscribe from meeting summary emails


nickoferrall:
In `/me/profile`, show a checkbox that lets users unsubscribe from meeting summary emails. 

See Figma designs: https://www.figma.com/file/2Nj1y1KUfE2swHZuDY5Zxn/My-User-Account-Management?node-id=110%3A1520

See Notion doc with further context: https://www.notion.so/parabol/Settings-Admin-IA-7d6e7e8fecf84a72a15d4ef7c7dab555?d=d81e2049dd804aa3822d5c9822bfd58d#8b767ca2449e4476b37ddd72c4d3943f

### AC

- [ ] Add an email Notification Panel to the profile page
- [ ] The Notification Panel includes a checkbox that says "Send meeting summary emails"
- [ ] By default, the checkbox is checked and meeting summary emails are sent
- [ ] If it's unchecked, meeting summary emails are not sent to that user
- [ ] In the footer of the meeting summary, include an unsubscribe link

PR #7190, opened 2022-09-16
feat: unsubscribe from summary emails

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7178

This builds on top of https://github.com/ParabolInc/parabol/pull/7183 as there's similar functionality.

https://www.loom.com/share/9b2ac67531124ec88936334a614853c8

### To test

- [ ]  Make sure that you have added the Mailgun env variables from https://github.com/ParabolInc/action-devops/blob/4aa09acb2ddc8376aaabd11bfc5bcfd7380b9bf8/environments/local#L41-L43
- [ ]  Add an email Notification Panel to the profile page
- [ ]  The Notification Panel includes a checkbox that says "Send meeting summary emails"
- [ ]  By default, the checkbox is checked and meeting summary emails are sent
- [ ]  If it's unchecked, meeting summary emails are not sent to that user
- [ ]  In the footer of the meeting summary, include an unsubscribe link


tianrunhe:
Can you add metric to the mutation? 😎 I'm sure we want to know how many users turned that off.

packages/client/components/EmailNotifications.tsx
```diff
@@ -0,0 +1,63 @@
...
+  )
+
+  const handleClick = () => {
+    ToggleSummaryEmailMutation(atmosphere, {}, {onError, onCompleted})
+  }
+
+  return (
+    <Wrapper>
+      <StyledCheckbox active={sendSummaryEmail} onClick={handleClick} />
+      <Text>{'Send meeting summary emails'}</Text>
```
igorlesnenko, line 57: Would be nice if we can make the whole checkbox label clickable

igorlesnenko (approved):

packages/client/components/EmailNotifications.tsx
nickoferrall: True! Updated 👍

nickoferrall:
> Can you add metric to the mutation? 😎 I'm sure we want to know how many users turned that off

Added!

tianrunhe (changes requested):

packages/client/mutations/ToggleSummaryEmailMutation.ts
```diff
@@ -0,0 +1,40 @@
...
+        error {
+          message
+        }
+      }
+      ...ToggleSummaryEmailMutation_team @relay(mask: false)
+    }
+  }
+`
+
+const ToggleSummaryEmailMutation: StandardMutation<any> = (
```
Dschoordsch, line 27: -1 this should be typed
```ts
const ToggleSummaryEmailMutation: StandardMutation<TToggleSummaryEmailMutation> = (
```

packages/server/graphql/mutations/helpers/endMeeting/sendNewMeetingSummary.ts
```diff
@@ -10,12 +11,17 @@ export default async function sendNewMeetingSummary(newMeeting: Meeting, context
...
+  const viewerId = getUserId(authToken)
+  const [teamMembers, team, user] = await Promise.all([
     dataLoader.get('teamMembersByTeamId').load(teamId),
     dataLoader.get('teams').loadNonNull(teamId),
+    dataLoader.get('users').load(viewerId),
     r.table('NewMeeting').get(meetingId).update({summarySentAt: now}).run()
   ])
+  if (!user) return
+  const {sendSummaryEmail} = user
+  if (sendSummaryEmail === false) return
```
Dschoordsch, line 24: -2 I did not test, but as I understand the setting it allows each user to set their preferences themselves. In the current implementation only the setting of the viewer who ends the meeting is considered, not of any other meeting participant. The check should be down in line 33.
```ts
  const emailAddresses = users.filter(isValid).filter(({sendSummaryEmail}) => sendSummaryEmail !== false).map(({email}) => email)
```

packages/server/graphql/public/typeDefs/toggleSummaryEmail.graphql
```diff
@@ -0,0 +1,18 @@
+extend type Mutation {
+  """
+  Toggles the sendSummaryEmail value on the User object which determines whether summary emails are sent to the user
+  """
+  toggleSummaryEmail: ToggleSummaryEmailPayload!
```
Dschoordsch, line 5: +1 probably an edge case, but did you consider adding the setting on team level? This way users could have this setting per team

Dschoordsch (changes requested):

packages/server/graphql/public/typeDefs/toggleSummaryEmail.graphql
nickoferrall: Yeah, there are plans to make these settings more advanced in the near future, but we just wanted to build something quick for now

packages/server/graphql/mutations/helpers/endMeeting/sendNewMeetingSummary.ts
nickoferrall: Good catch, thank you!

packages/server/graphql/public/mutations/toggleSummaryEmail.ts
```diff
@@ -0,0 +1,29 @@
...
+  if (!viewer) return standardError(new Error('User not found'), {userId: viewerId})
+
+  // RESOLUTION
+  const {sendSummaryEmail} = viewer
+  await updateUser({sendSummaryEmail: !sendSummaryEmail}, viewerId)
+  segmentIo.track({
+    userId: viewerId,
+    event: 'Toggle subscribe to summary email',
+    properties: {subscribeToSummaryEmail: !sendSummaryEmail}
+  })
```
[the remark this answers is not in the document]
nickoferrall: I was not aware! Thanks for letting me know and keeping a watchful eye on the PRs. That pattern is way better  🙌

tianrunhe (approved):
Sign-off from data side. Thanks for making the suggested changes!

Dschoordsch (approved):
Did a short smoke test and seems to work.

packages/server/utils/analytics/analytics.ts
```diff
@@ -306,6 +308,10 @@ class Analytics {
     this.track(userId, 'Task Estimate Set', taskEstimateProperties)
   }
 
+  toggleSubToSummaryEmail = (userId: string, subscribeToSummaryEmail: boolean) => {
+    this.track(userId, 'Summary Email Setting Changed', subscribeToSummaryEmail)
```
tianrunhe, line 312: The 3rd argument of `this.track` expects a dictionary so this should be `this.track(userId, 'Summary Email Setting Changed', {subscribeToSummaryEmail})`

PR #7192, opened 2022-09-16
fix(noImplicitAny): Fixup 200 ts rules (Part 1 of 2)

### LINKED ISSUE #7148: Add noImplicitAny to client config (Part 1 of X)
Now that the server config has `noImplicitAny` turned on, we'll want to do the same for the client

So, I propose we share the love & chip away at it.
Instead of turning the rule off & on, @jmtaber129 suggested we do what the tsc team does: https://code.visualstudio.com/blogs/2019/05/23/strict-null

- [x] create a tsconfig.noImplcitAny.json in the client
- [x] that config has noImplicitAny & a list of files (or an include section)
- [x] we slowly add to the files that get checked
- [ ] our git hook & CI runs a check using this config in parallel to current typechecks 
- [x] create a copy of this issue named Part (2 of x) so we can re-prioritize for the next person. We'll keep doing this until it's done

AC
- [x] config file setup & 200 noImplicitAny errors fixed on client

Estimate: 4 hours

mattkrick:
# Description

Fixes/Partially Fixes #7148 

fixes a couple hundred `noImplicitAny` rules.
The rest are fixed in #7193

## Testing scenarios

run `yarn tsc --noEmit -p  tsconfig.noImplicitAny.json` and see that all the rules in the mutations directory are fixed


packages/client/components/BillingLeaderActionMenu.tsx
```diff
@@ -45,7 +45,7 @@ const BillingLeaderActionMenu = (props: Props) => {
       if (submitting) return
       submitMutation()
       const variables = {orgId, userId, role}
-      SetOrgUserRoleMutation(atmosphere, variables, {}, onError, onCompleted)
+      SetOrgUserRoleMutation(atmosphere, variables, {onError, onCompleted})
```
mattkrick, line 48: updated old mutations to use the `StandardMutation<T>` generic so they've all got typesafety now

packages/client/mutations/RemoveOrgUserMutation.ts
```diff
@@ -92,19 +102,25 @@ const mutation = graphql`
...
-export const removeOrgUserOrganizationUpdater = (payload, {atmosphere, store}) => {
+export const removeOrgUserOrganizationUpdater: SharedUpdater<RemoveOrgUserMutation_organization> = (
+  payload,
+  {atmosphere, store}
+) => {
   const {viewerId} = atmosphere
-  const removedUserId = getInProxy(payload, 'user', 'id')
-  const removedOrgUserId = getInProxy(payload, 'organizationUserId')
-  const orgId = getInProxy(payload, 'organization', 'id')
+  const removedUserId = payload.getLinkedRecord('user').getValue('id')
```
mattkrick, line 110: `getInProxy` was an old helper i made a loong time ago, before I PRed relay to add strong types & before the `?` existed in javascript. Now that we have those, it's cleaner to use the native methods

packages/client/mutations/RemoveOrgUserMutation.ts
```diff
@@ -113,61 +129,71 @@ export const removeOrgUserTeamUpdater: SharedUpdater<RemoveOrgUserMutation_team>
...
+export const removeOrgUserTeamOnNext: OnNextHandler<RemoveOrgUserMutation_team> = (
+  payload,
+  context
+) => {
   const {atmosphere} = context
   const {teams} = payload
+  if (!teams) return
   teams.forEach((team) => {
-    const {newMeeting} = team
-    if (!newMeeting) return
```
mattkrick, line 145: caught a bug! `newMeeting` doesn't even exist on the payload, so this bit of code was never reached

packages/client/mutations/UpdateReflectionGroupTitleMutation.ts
```diff
@@ -37,7 +37,7 @@ const UpdateReflectionGroupTitleMutation: StandardMutation<TUpdateReflectionGrou
     onError,
     optimisticUpdater: (store) => {
       const {reflectionGroupId, title} = variables
-      const reflectionGroupProxy = store.get(reflectionGroupId)
+      const reflectionGroupProxy = store.get(reflectionGroupId)!
```
mattkrick, line 40: in many places I opted to keep the runtime behavior the same because this PR affects so many parts

packages/client/subscriptions/TaskSubscription.ts
```diff
@@ -32,56 +32,40 @@ const subscription = graphql`
...
   return requestSubscription<TTaskSubscription>(atmosphere, {
     subscription,
     variables,
     updater: (store) => {
       const payload = store.getRootField('taskSubscription') as any
       if (!payload) return
-      const type = payload.getValue('__typename')
+      const type = payload.getValue('__typename') as keyof typeof updateHandlers
       const context = {atmosphere, store: store as RecordSourceSelectorProxy<any>}
-      switch (type) {
```
mattkrick, line 51: now that all the updaters have the same signature `(payload, context)` we can use a lookup table above

packages/client/utils/sortByTier.ts
```diff
@@ -4,7 +4,7 @@ const sortByTier = <T extends readonly {tier: TierEnum | string; name: string}[]
   teamsOrOrgs: T
 ) => {
   const teamsSlice = teamsOrOrgs.slice()
-  const tierVal = (team) => (team.tier === 'enterprise' ? -2 : team.tier === 'pro' ? -1 : 1)
+  const tierVal = (team: T[0]) => (team.tier === 'enterprise' ? -2 : team.tier === 'pro' ? -1 : 1)
```
JimmyLv, line 7: #TIL 😁 indexed access types, also called [lookup types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html#keyof-and-lookup-types). 
```ts
type P5 = string[][0]; // string
```

packages/client/subscriptions/TaskSubscription.ts
```diff
@@ -32,56 +32,40 @@ const subscription = graphql`
...
-        case 'RemoveOrgUserPayload':
-          removeOrgUserTaskUpdater(payload, store, viewerId)
-          break
-        case 'UpdateTaskPayload':
-          updateTaskTaskUpdater(payload, context)
-          break
-        case 'UpdateTaskDueDatePayload':
-          break
-        default:
-          console.error('TaskSubscription case fail', type)
```
JimmyLv, line 75: -1 should we keep the default behavior here?

JimmyLv, line 75: -1 should we keep the default behavior here?

packages/client/tsconfig.noImplicitAny.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "noImplicitAny": true,
```
JimmyLv, line 4: +1 Do we still need it? or will it be removed in part 2?

JimmyLv (approved):
Nice work! 🎉

packages/client/utils/sortByTier.ts
mattkrick: heh, never knew they were called that! such a great way to sneak into a typed array

packages/client/subscriptions/TaskSubscription.ts
mattkrick: i probably should've avoided that runtime refactor here, the "new" way of handling these is just not throwing an error because it's not technically an error if we didn't build a handler for it!

packages/client/tsconfig.noImplicitAny.json
mattkrick: removed in part 2!

mattkrick:
closing in favor of part 2 since both have been reviewed independently

PR #7203, opened 2022-09-19
fix: snackbar messages are not bottom centered during standups

### LINKED ISSUE #7158: Snackbar messages are not bottom centered during standups
## Issue - Bug
Screenshot 2022-09-07 125238

- **How serious is the bug:**  P4
- **Tested on:**  Windows
- **Browser:** Chrome / firefox

Snackbar messages should be bottom centered

gitstart:
# Description

Fixes/Partially Fixes #7158

After digging, I discovered that the snackbar was positioned relative to the sidebar on the left, so this affected the positioning on pages that the sidebar isn't in view.

I modified the code to add the width of the sidebar to the position of the snackbar only when the screen is desktop when the page is not a meeting i.e when the sidebar is showing.

## After the bug fixed
Screenshot 2022-09-08 at 07 46 36
Screenshot 2022-09-07 at 20 00 42
Screenshot 2022-09-08 at 07 47 11
Screenshot 2022-09-07 at 20 03 54

## Testing scenarios

- [x] Scenario A
  - Step 1
  - Starting a meeting while on  another meeting. 

## Final checklist


gitstart:
Hi @Dschoordsch, @igorlesnenko  This PR is up for review

gitstart:
> @gitstart thank you for contributing. do you mind filling the PR template accordingly? I see the default values for testing scenarios. thank you!
Hi [BartoszJarocki](https://github.com/BartoszJarocki), the PR description was been updated, you can give another look to the PR.

packages/client/components/Snackbar.tsx
```diff
@@ -57,6 +60,7 @@ const Snackbar = React.memo(() => {
   const {openPortal, terminatePortal, portal} = usePortal({id: 'snackbar', noClose: true})
   const {location} = useRouter()
   const isMeetingRoute = location.pathname.startsWith('/meet/')
+  const isStandupRoute = isMeetingRoute && !location.pathname.includes('/checkin')
```
87prashant, line 63: I think using `relay` to add dependency of `meetingType` is also an option.

gitstart: I am seeking for more clarity on this @87prashant.

gitstart: Hi @87prashant, I am not really clear on this, can you explain further?

gitstart-parabol:
Hi @BartoszJarocki, this PR is ready for review.

packages/client/components/Snackbar.tsx
```diff
@@ -25,7 +25,10 @@ const Modal = styled('div')<{isMeetingRoute: boolean; isDesktop: boolean}>(
     paddingBottom: isMeetingRoute ? 64 : 8,
     position: 'fixed',
     top: 0,
-    width: isMeetingRoute && isDesktop ? `calc(100% + ${NavSidebar.WIDTH}px)` : '100%',
+    width:
+      isMeetingRoute && !isStandupRoute && isDesktop
```
BartoszJarocki, line 29: -1 the initial idea here was that only meeting routes have sidebar, so we could assume, we need to position the snackbar differently in that case. Right now, standup meeting is an exception, so not all meeting have the same layout. instead of introducing another boolean, I'd suggest replacing `isMeetingRoute` and `isStandupRoute` with just one variable, `hasSidebar`, and initializing it accordingly.

packages/client/components/Snackbar.tsx
BartoszJarocki, line 63: -2 this will be true even for non-standup routes, like `/meet/i0m8N96why/reflect`, which is a retrospective meeting. to make sure we're on the standup route, you need to check against pattern like `/meet\/.*\/responses/g`

BartoszJarocki (changes requested):
Thank you for looking into this. I left a couple of comments.

BartoszJarocki (approved):

PR #7208, opened 2022-09-21
feat: user presence supports k8s

### LINKED ISSUE #7204: feat(k8s): Flush redis Presence on shutdown
AC
- [ ] when node receives the shutdown signal, it calls redis to remove all user presence for users connected to that SERVER_ID
- [ ] remove the chunk of code from postDeploy that removes user presence

Estimate: 2 hours

mattkrick:
# Description

fix #7204

User presence is currently reset during a `postDeploy` step. This won't work when we move to k8s and have multiple servers deploying at different times.

This PR will wipe out user presence for all users connected to a given socket server during shutdown. In the event of an ungraceful shutdown, when a server starts up it performs a health check on all other servers. It then removes any user presence for servers that were not present during the health check.

## Testing scenarios

To see the tests in action, watch the loom: https://www.loom.com/share/79221b9fb47d42ddbcb7564a705a3e55

- [ ] In pm2.dev.config, increase number of socket servers to 2+ and autorestart to `true`
- [ ] in handleUpgrade, insert the following line `  console.log('USER CONNECT', userId, process.env.SERVER_ID, process.pid)`
- [ ] with 2 different users in 2 different browsers, connect to the app & make sure they're on the same team (both on same team dash)
- [ ] refresh the page on user 1, make sure it looks like they go offline in user 2's browser
- [ ] continue refreshing until the 2 users are on 2 different PIDs (see console logs)
- [ ] kill the server that user 1 is connected to via `kill -SIGTERM <PID>`
- [ ] user 2 sees user 1 go offline for a second, then come back online


packages/server/utils/ServerHealthChecker.ts
```diff
@@ -0,0 +1,74 @@
...
+    this.subscriber.on('message', (_channel, message) => {
+      this.pendingPongs.delete(message)
+    })
+    this.subscriber.subscribe(`socketServerPong:${SERVER_ID}`)
+    this.publisher.sadd('socketServers', SERVER_ID)
+  }
+  async ping() {
+    const socketServers = await this.publisher.smembers('socketServers')
+    this.pendingPongs = new Set(...socketServers.filter((id) => id !== SERVER_ID))
+    this.publisher.publish('socketServerPing', SERVER_ID)
```
Dschoordsch, line 27: +1 we should await here

packages/server/utils/ServerHealthChecker.ts
```diff
@@ -0,0 +1,74 @@
...
+    await this.reportDeadServers(deadServerIds)
+    return deadServerIds
+  }
+
+  async reportDeadServers(deadServerIds: string[]) {
+    if (deadServerIds.length === 0) return
+    const authToken = new ServerAuthToken()
+    // remove the dead servers from the set of socket servers in redis
+    const writes = deadServerIds.map((id) => ['srem', 'socketServers', id])
+    await this.publisher.multi(writes).exec()
```
Dschoordsch, line 40: +1 `await this.publisher.srem('socketServers', deadServerIds)` should work just fine here

packages/server/utils/ServerHealthChecker.ts
```diff
@@ -0,0 +1,74 @@
...
+    const userPresenceStream = this.publisher.scanStream({match: 'presence:*'})
+    userPresenceStream.on('data', async (keys) => {
+      if (!keys?.length) return
+      const reads = keys.map((key: string) => {
+        return ['lrange', key, '0', '-1']
+      })
+      userPresenceStream.pause()
+
+      const presenceBatch = (await this.publisher.multi(reads).exec()) as [null, string[]][]
+      presenceBatch.map((record, idx) => {
```
Dschoordsch, line 51: +1 we should probably await `publishInternalGQL` here

Dschoordsch (commented):
Did not test, but looks like it should work. Anything blocking this?

mattkrick:
@Dschoordsch you're too kind! thanks for the review. I'm stuck on testing that SIGINT handler. i could call `kill -s SIGINT`, but i'd like to just ctrl+c, but that's not triggering it! I'm not sure what is swallowing the signal...

mattkrick:
i misunderstood the SIGINT handler, it doesn't handle promises.
instead of doing async work in the SIGINT handler, I'll have the dying node publish a message to all living nodes asking them to kill it in redis :grimacing:

mattkrick:
@Dschoordsch would you mind giving this old thing a fresh review?

packages/server/utils/ServerHealthChecker.ts
```diff
@@ -0,0 +1,88 @@
...
+          const promise = publishInternalGQL({
+            authToken,
+            query: disconnectQuery,
+            socketId,
+            variables: {userId}
+          })
+          disconnectPromises.push(promise)
+        })
+      })
+      await Promise.all(disconnectPromises)
```
Dschoordsch, line 77: +1 untested, but couldn't we use `flatMap` to avoid having this promise storing logic?
```ts
     await presenceBatch.flatMap((record, idx) => {
        const key = keys[idx]!
        const userId = key.slice(key.indexOf(':') + 1)
        const connections = record[1]
        return connections.map((connection) => {
          const presence = JSON.parse(connection) as UserPresence
          const {socketServerId, socketId} = presence
          if (!deadServerIds.includes(socketServerId)) return
          // let GQL handle the disconnect logic so it can do special handling like notify team memers
          return publishInternalGQL({
            authToken,
            query: disconnectQuery,
            socketId,
            variables: {userId}
          })
        })
      })
```

Dschoordsch (changes requested):
Looks good, did not test yet. Couple of smaller nits

packages/server/server.ts
```diff
@@ -22,21 +22,31 @@ import SSEConnectionHandler from './sse/SSEConnectionHandler'
...
   service: `Web ${process.env.SERVER_ID}`,
   appsec: process.env.DD_APPSEC_ENABLED === 'true',
   plugins: false
 })
 
-const PORT = Number(PROD ? process.env.PORT : process.env.SOCKET_PORT)
+const onKill = async () => {
+  r.getPoolMaster()?.drain()
+  const healthChecker = new ServerHealthChecker()
+  await healthChecker.reportDeadServers([process.env.SERVER_ID!])
```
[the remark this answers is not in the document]
mattkrick: i thought so! oddly enough it works as long as we don't `await r.getPoolMaster().drain()`

packages/server/utils/ServerHealthChecker.ts
```diff
@@ -0,0 +1,88 @@
...
+    this.subscriber.subscribe('socketServerPing', `socketServerPong:${SERVER_ID}`)
+    this.publisher.sadd('socketServers', SERVER_ID)
+  }
+
+  // get a list of servers who should be alive
+  // ping all servers
+  // if there are servers who say they're alive but they have responded, flag them as dead
+  async ping() {
+    const socketServers = await this.publisher.smembers('socketServers')
+    this.pendingPongs = new Set(...socketServers.filter((id) => id !== SERVER_ID))
```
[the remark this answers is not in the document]
mattkrick: oh that's good, will fix

mattkrick:
I think this is good to go, we're got a lot of risky PRs going on right now, so I'll hold off on merging this until things calm down in a week or so.

PR #7226, opened 2022-09-23
feat: let super users invite to team

nickoferrall:
See this Slack thread for further context: https://parabol.slack.com/archives/C4JAUUZ9P/p1663860269145519

A user would like to be added to a team that doesn't have any team members. We would like to be able to execute the `inviteToTeam` mutation, but can't do it because it requires the viewer to be on the team.

This PR enables super users to access the mutation.

### To test

- [ ] A super user can invite a user to a team even when the super user does not belong to that team
- [ ] After executing the mutation, the user receives an email invite and is able to join the team

### DIFF
--- a/packages/server/graphql/mutations/inviteToTeam.ts
+++ b/packages/server/graphql/mutations/inviteToTeam.ts
@@ -13,7 +13,7 @@ import teamInviteEmailCreator from '../../email/teamInviteEmailCreator'
 import {getUsersByEmails} from '../../postgres/queries/getUsersByEmails'
 import removeSuggestedAction from '../../safeMutations/removeSuggestedAction'
 import {analytics} from '../../utils/analytics/analytics'
-import {getUserId, isTeamMember} from '../../utils/authorization'
+import {getUserId, isSuperUser, isTeamMember} from '../../utils/authorization'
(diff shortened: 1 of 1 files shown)


packages/server/graphql/mutations/inviteToTeam.ts
```diff
@@ -57,8 +57,8 @@ export default {
 
       // AUTH
       const viewerId = getUserId(authToken)
-      if (!isTeamMember(authToken, teamId)) {
-        return standardError(new Error('Team not found'), {userId: viewerId})
+      if (!isTeamMember(authToken, teamId) && !isSuperUser(authToken)) {
+        return standardError(new Error(`Viewer does not belong to that team`), {userId: viewerId})
```
Dschoordsch, line 61: -1 I think we put "Team not found" here for security reasons. At least we used this consistently everywhere.

Dschoordsch (changes requested):

packages/server/graphql/mutations/inviteToTeam.ts
nickoferrall: True, I guess we want to make that message unclear for potential attackers. I can see a few places where we're not doing that, but it seems safer to keep the old error message. Update pushed!

PR #7234, opened 2022-09-28
feat: speed up repo integrations query

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6007

The `repoIntegrations` query is very expensive as it reaches out to all of our integrations even though the user probably wants a few of their most recently used repoIntegrations.

In this PR, I've removed the `allRepoIntegrations` query, the `useAllIntegrations` hook, and the query to our db in `repoIntegrationHelpers` which returned the previously used integrations.

Instead, I've implemented a cache in Redis so devs can just query `repoIntegrations`. 

The Redis cache consists of `allRepoIntegrations` and `prevUsedRepoIntegrations`:

- `allRepoIntegrations` is a key value pair with a 90-day TTL that gets updated every time we reach out to the network
- `prevUsedRepoIntegrations` is a sorted set which uses the most recently used unix timestamp as the score

Sorry this is such a large PR! Refactoring the way `repoIntegrations` work requires changing a bunch of different files. I also had to upgrade Redis to get the types I needed which meant changing several additional files. 

### To test

- [ ] Can push a task to GitHub, GitLab, Jira, Jira Server, and Azure DevOps
- [ ] Can create a new GitHub, GitLab, Jira, Jira Server, and Azure DevOps issue in Sprint Poker
- [ ] After pushing a task to an integration or creating an issue, refresh the page, open the task footer integration menu, and see that the most recently used `repoIntegration` is now at the top of the menu
- [ ] After pushing a few tasks to integrations, remove one of the integrations, e.g. GitHub, refresh the page, open the task footer menu again, and see that the menu no longer includes that integration
- [ ] Add a new integration, e.g. GitLab, refresh the page, and see that the integration shows up in the task footer menu
- [ ] In the task footer menu, search for a repoIntegration that doesn't exist. See that `repoIntegrations` reaches out to the APIs 

Ideally, we wouldn't need to refresh to see the changes, but that's the way it was previously working. As this PR is large enough, I considered changing that to be out of scope.

Creating a new issue in Azure results in an undefined project title, see issue: https://github.com/ParabolInc/parabol/issues/7266

https://www.loom.com/share/05f8a6d4147e415cac8cf732f233ad9f


packages/server/graphql/queries/helpers/repoIntegrationHelpers.ts
```diff
@@ -1,49 +1,33 @@
...
-type PrevRepoIntegrationRes = PrevGitHubRepoIntegration | PrevJiraRepoIntegration
-
 export const getPrevRepoIntegrations = async (
   userId: string,
   teamId: string,
   permLookup: Unpromise<ReturnType<typeof getPermsByTaskService>>
 ) => {
   const r = await getRethink()
-  const res = (await (
+  const prevIntegrationsRes = (await (
```
nickoferrall, line 30: Hey @mattkrick, yesterday during your office hours we discussed removing prev repo integrations from rethinkdb and storing them in Redis. 

You suggested that I check the speed of this nasty query in prod in Rethinkdb and see if it's slow. Here are the results:

- 277ms for me in the Product team
- 48ms for me in the Product Squad 3's team
- 334ms for you in the Product team
- 371ms for Jordan in the Product team
- 337 ms for Jordan in the Growth team

That doesn't seem that slow to me given that you and Jordan have probably created a lot more tasks than the average user.  

Do you think I should go ahead with the Redis refactor? If so, the plan for that was:

- when we call `repoIntegrations`, cache all the values in Redis
- yesterday I think we said that the key would be `teamId` + `integrationServiceName`, but I think we'd need to include the `userId` in that key too
- we would set a TTL of 3 months on the Redis key. Every time we fetch the key, we'd update the TTL
- the Redis values would be an array of objects which include all of the fields we get in `repoIntegrations`
- we would include a `lastUsedAt` field in the object which we would sort the results by

mattkrick: Up to you! Are there other users that might have slower queries? Instead of targeting the average user wait time, could we say something like "99% of users have a latency under 500ms"? Do we cache on the client side so it's quicker after the first time I open the menu?

On the side, if we don't cache in redis is there a way to keep the query from growing in complexity with every new integration that we add? If we get rid of the query, could we also remove other fields like a label or display name kept in Task.integration? If the goal is to have a really fast initial fetch, is there a way to avoid reaching out to each integration to get the repo name?

packages/server/graphql/queries/helpers/fetchAllRepoIntegrations.ts
```diff
@@ -12,51 +33,45 @@ const fetchAllRepoIntegrations = async (
...
-        ? -1
-        : a.service < b.service
-        ? -1
-        : a.id.toLowerCase() < b.id.toLowerCase()
-        ? -1
-        : 1
-    )
+
+  // alternate the repoIntegrations by service so the user is aware of all available services
+  const repoIntegrationsAltServices: RemoteRepoIntegration[] = []
```
nickoferrall, line 53: Previously, in the menu, we'd show everything from one service followed by everything from another. The problem is that we only show 10 repo integrations to start with, so the menu might show 10 GitHub repos even though they've also integrated Jira.

alternating

I'm now alternating between the available services so the user knows what's available.

packages/client/components/TaskFooterIntegrateMenuList.tsx
```diff
@@ -34,39 +34,110 @@ const Label = styled('div')({
...
-    !!hasMore,
-    teamId,
-    userId
-  )
+
+  useEffect(() => {
+    // if searching for a repoIntegration that doesnt exist in the cache, it could be stale so use the network
+    if (!networkOnly && filteredIntegrations.length === 0) {
+      setNetworkOnly(true)
+      setKeepParentFocus(false)
```
nickoferrall, line 129: Unable to focus on search input without this change

packages/client/shared/gqlIds/IntegrationRepoId.ts
```diff
@@ -21,38 +26,32 @@ type AzureDevOpsRepoIntegration = {
...
-      | AzureDevOpsRepoIntegration
-  ) => {
+  join: (integration: RepoIntegration) => {
     const {service} = integration
     switch (service) {
       case 'github':
         return integration.nameWithOwner
       case 'jira':
-        return JiraProjectId.join(integration.cloudId, integration.projectKey)
+        return JiraProjectId.join(integration.cloudId, integration.projectKey ?? integration.key!)
```
nickoferrall, line 48: `projectKey` doesn't exist in remote Jira projects, i.e. projects that we get directly from Jira's API. We use `IntegrationRepoId` for both remote Jira projects and our internal gql Jira projects, so I suggest checking for `projectKey` and `key`

packages/server/dataloader/RedisCache.ts
```diff
@@ -51,7 +51,7 @@ export default class RedisCache<T extends keyof CacheType> {
   // }
   private getRedis() {
     if (!this.redis) {
-      this.redis = new Redis(process.env.REDIS_URL, {connectionName: 'redisCache'})
+      this.redis = new Redis(6379, process.env.REDIS_URL!, {connectionName: 'redisCache'})
```
nickoferrall, line 54: I had to upgrade Redis to be able to sort the sorted set in reverse order by timestamp. After upgrading: 

- the typings have improved which caught a few potential bugs
- the `Redis.Redis` type became `Redis`
- `new Redis()` requires a port number as its first arg. We could add this to the env vars, but I don't think it needs to be kept hidden so I added it here. Alternatively, I could add the number to `constEnums`, for example, and we could always import it from there

packages/server/dataloader/azureDevOpsLoaders.ts
```diff
@@ -117,6 +117,8 @@ export interface AzureUserInfo {
 export interface AzureAccountProject extends TeamProjectReference {
   userId: string
   teamId: string
+  instanceId: string
+  projectId: string
```
nickoferrall, line 121: We call `allAzureDevOpsProjects` in `fetchAllRepoIntegrations`, which get used by `IntegrationRepoId` which needs the `instanceId` and `projectId`: https://github.com/ParabolInc/parabol/blob/bb6bbf9b7bcb06fb266d036396f1490e7822e31d/packages/client/shared/gqlIds/IntegrationRepoId.ts#L52

packages/server/graphql/mutations/helpers/createTaskInService.ts
```diff
@@ -42,13 +43,15 @@ const createTaskInService = async (
...
+    const integrationRepoId = IntegrationRepoId.join({cloudId, projectKey, service})
     return {
       integration: new TaskIntegrationJira({
         accessUserId,
         cloudId,
         issueKey
       }),
-      integrationHash: JiraIssueId.join(cloudId, issueKey)
+      integrationHash: JiraIssueId.join(cloudId, issueKey),
+      integrationRepoId
```
nickoferrall, line 54: When we create a new integration issue in a Sprint Poker meeting, we call `createTask` which calls `createTaskInService`. The `integration` and `integrationHash` provide info about the issue, but not some about the `repoIntegration`. I'm now returning the `repoIntegrationId` so that we can update the prev used repo integration cache and update the timestamp.

igorlesnenko:
@nickoferrall I have a very cryptic name when pushing to Azure, is it expected behaviour?
image

nickoferrall:
@igorlesnenko yep, it's like that in prod. I've created an issue to fix it: https://github.com/ParabolInc/parabol/issues/7274

igorlesnenko:
@nickoferrall 

> In the task footer menu, search for a repoIntegration that doesn't exist. See that repoIntegrations reaches out to the APIs

- I opened a dashboard
- Created a new project in azure
- Tried to find this new project and was unable to
- It only works after refresh.
- Am I doing this right, is it expected behaviour?

https://user-images.githubusercontent.com/466991/195320327-5aba9fea-2d00-4a90-b436-b5fc88407be3.mov

Also I feel that experience a bit weird, as the whole dialog reloads as I type, and I can't continue typing while it loads and after that focus is lost.
Just wanted to mention it, if it is not a bug, I think it is ok to keep it like this for now

nickoferrall:
> It only works after refresh.
> Am I doing this right, is it expected behaviour?

Yep, that's expected. I mentioned in the PR description that it's not ideal that the refresh is needed, but that's how it's currently working in prod, and as this PR is already very large, I didn't want to increase the scope any further.

> Also I feel that experience a bit weird, as the whole dialog reloads as I type, and I can't continue typing while it loads and after that focus is lost.

Yeah, I agree! It's a tough one. We want to use the Redis cache as much as possible, but if the cache is stale, e.g. a user has created a new GitHub repo, we want to use the network. Two ideas we discussed in https://github.com/ParabolInc/parabol/issues/6007 were:

1. If the user searches for a repoIntegration and it doesn't exist, they may be searching for a newly added repoIntegration, so use the network
2. Show a "Load more" button that the user can click to invalidate the cache

We could do 1 & 2, or something completely different, but I'd like to start with just option 1. If I had created a new GitHub repo, I'd search for it. Hopefully, the "Load more" button would be redundant. 

At the moment, after searching for something that doesn't exist, the menu shows the loading spinner as the new query is happening. I get that it feels weird when you're searching. Perhaps the component order should be refactored so the search input is a parent of the `TaskFooterIntegrateMenuRoot` query? That way the user would still see the search input while the menu list is loading. Ideally, I'd like to make that change in a separate PR though as it would require touching several more files.

packages/server/package.json
```diff
@@ -33,7 +33,7 @@
     "@types/dotenv": "^6.1.1",
     "@types/fbjs": "^3.0.2",
     "@types/graphql": "^14.5.0",
-    "@types/ioredis": "^4.0.10",
+    "@types/ioredis": "^4.28.10",
```
igorlesnenko, line 36: The [migration guide](https://github.com/luin/ioredis/wiki/Upgrading-from-v4-to-v5) says we can now remove `@types/ioredis`. I think we can create an issue about this if make sense.

nickoferrall: Oh nice! I'd prefer to handle it in a separate PR. I've created an issue here: https://github.com/ParabolInc/parabol/issues/7280

packages/client/mutations/AddGitHubAuthMutation.ts
```diff
@@ -15,7 +15,6 @@ graphql`
         }
       }
       # after adding, check for new integrations (populates the menu)
```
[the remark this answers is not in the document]
nickoferrall: Good point, removed!

packages/server/integrations/updatePrevUsedRepoIntegrationsCache.ts
```diff
@@ -0,0 +1,31 @@
+import IntegrationRepoId from 'parabol-client/shared/gqlIds/IntegrationRepoId'
+import getAllCachedRepoIntegrations from '../graphql/queries/helpers/getAllCachedRepoIntegrations'
+import getPrevUsedRepoIntegrations from '../graphql/queries/helpers/getPrevUsedRepoIntegrations'
+import getPrevUsedRepoIntegrationsRedisKey from '../utils/getPrevUsedRepoIntegrationsRedisKey'
+import getRedis from '../utils/getRedis'
+
+const updatePrevUsedRepoIntegrationsCache = async (teamId: string, repoIntegrationId: string) => {
```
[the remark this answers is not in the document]
nickoferrall: > is it possible that other teammates would see a list of my repositories instead of their own? 

Yeah, I think that's a valid tension, but that's how the integrations are currently working in prod. If one user adds the GitHub integration, we can all see their repos in a Sprint Poker meeting. That might feel strange to some users, but it does mean that each team member doesn't need to add an integration to use the meeting - just one person does 

> will teammates clear each other caches?

Yep, they will as the cache is by team. I think that's the right move as everyone on the team should have access to the team's repo that they want to use

packages/server/integrations/updatePrevUsedRepoIntegrationsCache.ts
```diff
@@ -0,0 +1,31 @@
...
+  const redis = getRedis()
+  const prevUsedRepoIntegrationsKey = getPrevUsedRepoIntegrationsRedisKey(teamId)
+  const [prevUsedRepoIntegrations, allCachedRepoIntegrations] = await Promise.all([
+    getPrevUsedRepoIntegrations(teamId),
+    getAllCachedRepoIntegrations(teamId)
```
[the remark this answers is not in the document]
nickoferrall: What scenario are you thinking of where that might happen?

When `repoIntegrations` is initially called, it validates the cache. `updatePrevUsedRepoIntegrationsCache` is called when the user has selected a `repoIntegration` from the menu.

It is possible that the user: 
- opens the menu
- `allRepoIntegrationsCache` is updated
- someone then deleted a repo in GitHub, for example
- the user selects the deleted repo from the cache, but it doesn't exist

However, in that case (I just tested it), we'd get an error message on the task saying that the repo doesn't exist. This function wouldn't be called so it wouldn't be added to the prevUsed cache, and when `repoIntegrations` next fetches from the network, it would remove that repo.

nickoferrall:
> This make sense. Should we create an issue about it?

Yeah, I think we should - here's an issue: https://github.com/ParabolInc/parabol/issues/7283

packages/server/integrations/updatePrevUsedRepoIntegrationsCache.ts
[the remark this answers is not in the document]
mattkrick: yaaa i think the thinking was (back in the day) that if a user didn't have access to write to a certain repo, they could still push to it & we'd use the other person's credentials.
for example, if Nick had access to a repo I didn't, I could push it it & then it would show Nick adding a task & there'd be a footer saying "on behalf of Matt". ...tbh I'm not sure if we kept that logic or not :sweat_smile:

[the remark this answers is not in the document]
igorlesnenko: My concern is that 2 users, even if they both have an access to the same repo, will have different set of repos. So for the second user I would see cached **private repos** of the first user. He might not want to expose it.

What do you think?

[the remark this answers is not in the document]
igorlesnenko: I tried to test GitHub integrations with 2 different users with 2 different GitHub  accounts. I notice few issues and tried to explain the privacy issue. Here is the loom — https://www.loom.com/share/5cb94e6f8a97495aa568d0e2be5c9e74 

In case of poker I unable to see your private repos, however in case of shared cache I will able to

image

What do you think? Are you able to reproduce issues from the loom?

packages/server/graphql/queries/helpers/removeStalePrevUsedRepoIntegations.ts
```diff
@@ -0,0 +1,15 @@
...
+import {RemoteRepoIntegration} from './fetchAllRepoIntegrations'
+
+const removeStalePrevUsedRepoIntegrations = async (
+  prevUsedRepoIntegrations: RemoteRepoIntegration[] | null,
+  prevUsedRepoIntegrationsKey: string
+) => {
+  if (!prevUsedRepoIntegrations) return
+  const sixMonthsAgo = Date.now() - ms('180d')
+  const redis = getRedis()
+  await redis.zremrangebyscore(prevUsedRepoIntegrationsKey, '-inf', sixMonthsAgo)
```
mattkrick, line 12: -2 does the previously used repos have a TTL? let's say a user stops using us for a month or a year. how will that key get removed from redis? if redis died, how would we recover the previous used repos?
for persisted data, postgres works really well because we keep that on disk where it's cheap & we have replicas, perform backups, etc. 
we have to treat redis like its data is ephemeral & could die at any moment without a loss in UX.

if we stick with redis, perhaps we have a function that calculates the previously used repos by polling the tasks on that team if the value in redis isn't found? then we can put a TTL on it

mattkrick (changes requested):

packages/client/components/TaskFooterIntegrateMenuList.tsx
```diff
@@ -34,39 +34,109 @@ const Label = styled('div')({
...
-  return jiraItemName || githubName
+type Item = NonNullable<
+  NonNullable<
+    TaskFooterIntegrateMenuListLocalQueryResponse['viewer']['teamMember']
+  >['repoIntegrations']['items']
+>[0]
+
+const getValue = (item: Item) => {
+  const {service} = item
+  if (service === 'jira' || service === 'azureDevOps' || service === 'jiraServer')
```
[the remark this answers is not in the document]
nickoferrall: I think that would be helpful as we add new integrations, but we don't need to include msTeams here as we can't push tasks to it?

packages/server/graphql/queries/helpers/fetchAllRepoIntegrations.ts
```diff
@@ -12,51 +33,45 @@ const fetchAllRepoIntegrations = async (
...
+  let loopCount = 0
+  while (repoIntegrationsAltServices.length < fetchedRepoIntegrations.length) {
```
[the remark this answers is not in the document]
nickoferrall: Love the explanation and the imperative loop! 🔥 Thanks!

packages/server/graphql/queries/helpers/sortRepoIntegrations.ts
```diff
@@ -0,0 +1,20 @@
+import IntegrationRepoId from 'parabol-client/shared/gqlIds/IntegrationRepoId'
+import {RemoteRepoIntegration} from './fetchAllRepoIntegrations'
+
+const sortRepoIntegrations = async (
+  allRepoIntegrations: RemoteRepoIntegration[] | null,
+  prevUsedRepoIntegrations: RemoteRepoIntegration[] | null
+) => {
+  if (!prevUsedRepoIntegrations) return allRepoIntegrations ?? []
```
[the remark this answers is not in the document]
nickoferrall: I don't think so - this is just removing duplicate repoIntegrations and showing the prevUsed cached ones before the allRepo caches ones

packages/server/graphql/queries/helpers/removeStalePrevUsedRepoIntegations.ts
nickoferrall: We don't have a TTL as it's a sorted set using the timestamp as the score. We could use a key-value pair with a TTL but updating that cache would be more complex. Currently, the stale values are removed when anyone in the team reads repoIntegrations. If a team stops using us for a year, the values wouldn't be removed. How much of a concern is that?

> we have to treat redis like its data is ephemeral & could die at any moment without a loss in UX

For sure, if Redis died, there wouldn't be any real loss to the UX. repoIntegrations checks the prevUsed cache and allRepos cache. If allRepos cache is empty, it'll reach out to the network. If prevUsed cache is empty, it'll show the allRepos cache or allRepos results from the network

mattkrick: >  How much of a concern is that?

if the values never expire, we'll want to move this to postgres. redis is just for ephemeral data-- think of it like an really big in-memory javascript object. if it grows unbounded, then we'd call that a memory leak.

> wouldn't be any real loss to the UX.

wouldn't we lose the previously used repos? for example, on the product team say we have 2 previous used repos: parabol & action-chronos. if redis died, the user would now see all the repos instead of the 2 that they've been using. as a user, would that feel odd?

packages/client/components/TaskFooterIntegrateMenuList.tsx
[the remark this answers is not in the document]
mattkrick: ah! i wasn't sure if teams could handle tasks yet or if that was only devops.

packages/server/graphql/queries/helpers/removeStalePrevUsedRepoIntegations.ts
nickoferrall: I've pushed an update that expires the prevRepos sorted set after 6 months of inactivity 👍 

To recap our chat yesterday, we felt that using Redis was the right solution because if Redis were to go down, we would fetch all repos from the network. The user wouldn't see their prevUsed repos at the top of the list, but they would be able to easily search for them

nickoferrall:
I haven't been able to reproduce [that error](https://github.com/ParabolInc/parabol/pull/7234#issuecomment-1284403880). I'm thinking it could be stale node modules causing problems as I upgraded Redis.

@MindEvo would you mind testing this PR and seeing if you can reproduce the error Matt shared? 🙏

packages/server/integrations/updatePrevUsedRepoIntegrationsCache.ts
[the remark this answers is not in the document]
igorlesnenko: > For the bug you mentioned where the user is unable to see any issues in Poker, does your second GitHub account have any issues?

@nickoferrall second account only have an access to Parabol repo, but it have no issues created by himself on GitHub. Is that what you mean?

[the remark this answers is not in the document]
nickoferrall: Does the second GitHub account belong to any repos that contain issues? Does it belong to the Parabol repo, for instance?

I tested the issue by creating a brand new GitHub account and I created a couple of repos. I integrated with Parabol, but couldn't see any issues in Sprint Poker (just like the bug you mentioned), but that was because I hadn't added any issues to that repo.

If your second GitHub account does belong to a repo that contains issues, would you mind double-checking if the bug persists after my latest changes?

mattkrick:
doh! looks like `REV` requires redis v6.2, I wasn't on latest.
LGTM!

PR #7237, opened 2022-09-29
chore: boot message changed to include Server ID and task. 

rafaelromcar-parabol:
# Description

- Adds the Server ID to the boot message of the Web Server.
- Changes where the Server ID is shown in the GQL Executor boot message.

## Demo

Screenshot from 2022-09-29 08-57-59

## Testing scenarios

There are no particular test scenarios. The application has been launched locally and it runs, showing the expected message.

## Final checklist

### DIFF
--- a/packages/server/listenHandler.ts
+++ b/packages/server/listenHandler.ts
@@ -4,8 +4,9 @@ import getGraphQLExecutor from './utils/getGraphQLExecutor'
 
 const listenHandler = (listenSocket: us_listen_socket) => {
   const PORT = Number(PROD ? process.env.PORT : process.env.SOCKET_PORT)
+  const SERVER_ID = Number(process.env.SERVER_ID)
   if (listenSocket) {
-    console.log(`\n🔥🔥🔥 Ready for Sockets: Port ${PORT} 🔥🔥🔥`)
+    console.log(`\n🔥🔥🔥 Server ID: ${SERVER_ID}. Ready for Sockets: Port ${PORT} 🔥🔥🔥`)
     getGraphQLExecutor().subscribe()
   } else {
     console.log(`❌❌❌    Port ${PORT} is in use!    ❌❌❌`)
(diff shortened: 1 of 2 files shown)


packages/server/listenHandler.ts
```diff
@@ -4,8 +4,9 @@ import getGraphQLExecutor from './utils/getGraphQLExecutor'
 
 const listenHandler = (listenSocket: us_listen_socket) => {
   const PORT = Number(PROD ? process.env.PORT : process.env.SOCKET_PORT)
+  const SERVER_ID = Number(process.env.SERVER_ID)
```
Dschoordsch, line 7: +1 Why do you parse it as number if you just want to print it?

Dschoordsch (approved):

packages/server/listenHandler.ts
Dschoordsch: Did not test

rafaelromcar-parabol: TBH, because I didn't know if it was needed and it was done that way the precedent line. I can change it if you want, I just tested it locally and has no impact.

Dschoordsch: Please change it then. If for whatever reason this is not a number, it might be useful to see what it's set to. In your version it would simply print `NaN`.

rafaelromcar-parabol: Actually, yesterday I had a similar problem with `const PORT = Number(PROD ? process.env.PORT : process.env.SOCKET_PORT)` 

I launched the application (in Kubernetes) without PORT env var configured and it just showed `NaN` in the message `Ready for Sockets: NaN` and then crashed. It might be worth to check somewhere if the mandatory parameters are set. Not in this PR, but what do you think @Dschoordsch?

Dschoordsch: Yes, I'm all for it. Result will still be the application crashing, but at least with a better message :+1:

rafaelromcar-parabol: Change done. I let you resolve it if it's OK

Dschoordsch (approved):
Tested, works

PR #7030, opened 2022-08-09
fix: Hover over avatar to see names not working for some users

87prashant:
# Description

Fixes: #6139

## Demo

https://user-images.githubusercontent.com/106697681/183618928-8abe9dff-dade-4c20-9fbc-43ddb0161e15.mp4

## Testing scenarios
Check the functionality in every meeting type

## Final checklist

- [x] I checked the [code review guidelines](../docs/codeReview.md)
- [x] I have added [Metrics Representative](../docs/codeReview.md#metrics-representative) as reviewer(s) if my PR invovles metrics/data/analytics related changes
- [x] I have performed a self-review of my code, the same way I'd do it for any other team member
- [x] I have tested all cases I listed in the testing scenarios and I haven't found any issues or regressions
- [x] Whenever I took a non-obvious choice I added a comment explaining why I did it this way
- [ ] I added the label `One Review Required` if the PR introduces only minor changes, does not contain any architectural changes or does not introduce any new patterns and I think one review is sufficient'
- [x] PR title is human readable and could be used in changelog

PR #7239, opened 2022-09-30
fix: Hover over avatar to see names not working for some users

Dschoordsch:
# Description

Rebase of #7030


Dschoordsch:
Was already reviewed in #7030 and I tested the change locally.

PR #7241, opened 2022-09-30
fix: number of stories in summary stats is one less than actual stories

### LINKED ISSUE #7092: Sprint Poker: number of stories in summary stats is one less than actual stories
## Issue - Bug

When a Sprint Poker meeting has ended, the number of stories in the summary stats is one less than the actual number of stories that were estimated.

Captura de Pantalla 2022-08-18 a la(s) 3 36 51 p m

Note that the stats at the top says `8` but the list of stories estimated below totals `9`.

- **OS:** Unspecified
- **Browser:** Unspecified

- Number of stories in the summary stats is equal to the number of stories estimated

87prashant:
# Description

Fixes/Partially Fixes #7092

## Demo
Screenshot 2022-09-30 234026

## Final checklist

### DIFF
--- a/packages/server/graphql/mutations/endSprintPoker.ts
+++ b/packages/server/graphql/mutations/endSprintPoker.ts
@@ -67,11 +67,11 @@ export default {
     const {stage} = currentStageRes
     const phase = getMeetingPhase(phases)
     stage.isComplete = true
+    stage.endAt = now
     const storyCount = new Set(
       estimateStages.filter(({isComplete}) => isComplete).map(({taskId}) => taskId)
     ).size
     const discussionIds = estimateStages.map((stage) => stage.discussionId)
-    stage.endAt = now
 
     const completedMeeting = (await r
       .table('NewMeeting')
--- a/packages/server/graphql/mutations/endSprintPoker.ts
+++ b/packages/server/graphql/mutations/endSprintPoker.ts
@@ -64,13 +64,13 @@ export default {
     if (!currentStageRes) {
       return standardError(new Error('Cannot find facilitator stage'), {userId: viewerId})
     }
+    const {stage} = currentStageRes
+    const phase = getMeetingPhase(phases)
+    stage.isComplete = true
     const storyCount = new Set(
       estimateStages.filter(({isComplete}) => isComplete).map(({taskId}) => taskId)
     ).size
     const discussionIds = estimateStages.map((stage) => stage.discussionId)
-    const {stage} = currentStageRes
-    const phase = getMeetingPhase(phases)
-    stage.isComplete = true
     stage.endAt = now
 
     const completedMeeting = (await r
--- a/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/makePokerStats.ts
+++ b/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/makePokerStats.ts
@@ -20,7 +20,7 @@ const makePokerStats = (meetingRef: makePokerStats_meeting$key) => {
   const meetingMembersCount = meetingMembers.length
   return [
     {value: '', label: ''},
-    {value: storyCount, label: plural(storyCount, 'Story', 'Stories')},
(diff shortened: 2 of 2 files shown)


nickoferrall (changes requested):
Thanks for the PR! The main bug in that issue is the number of stories aren't being calculated correctly because the final story isn't being [marked as complete](https://github.com/ParabolInc/parabol/issues/7092#issuecomment-1247194705). I think we should fix that before closing the linked issue

nickoferrall (changes requested):
It's not working for me - if I vote on three stories, the summary only counts two:

story-count

Is it working for you?

87prashant:
@nickoferrall 
Yes, but I did `yarn start` also to make it work.

packages/server/graphql/mutations/endSprintPoker.ts
```diff
@@ -64,13 +64,13 @@ export default {
     if (!currentStageRes) {
       return standardError(new Error('Cannot find facilitator stage'), {userId: viewerId})
     }
+    const {stage} = currentStageRes
+    const phase = getMeetingPhase(phases)
+    stage.isComplete = true
```
87prashant, line 69: After these changes, I also did `yarn build`

nickoferrall (changes requested):
This still isn't working for me. @MindEvo would you mind testing this PR to see if it works for you, please?

87prashant:
It is only working for me, that's weird.
@nickoferrall 
Demo video:

https://user-images.githubusercontent.com/106697681/195330917-d5792c72-4915-4613-ada3-a0648bd5c76a.mp4

87prashant:
@nickoferrall 
> Is the Story Count always correct for you?

Yes, I ran over 20 meetings and everyone had positive results. It seems like it is counting the 'completed stage' before marking the stage—on which we click the end meeting button—complete.
I don't know if it is related: You used the GitHub issues (in your video) and I used the Parabol issues as stories.

Also, what made you think or how did you know that the fix might have the bug, so, you should test it in a lot of meetings because I only checked it in 2-3 meetings and thought it works. Is it part of the process of how you test things or something that made you suspicious during reviewing? I am just curious and want to improve my process.

nickoferrall:
Thanks, @MindEvo!

@87prashant 
> how did you know that the fix might have the bug, so, you should test it in a lot of meetings

I wanted to test several meetings to make sure that it was fixed - nothing to do with the code, I just noticed after 2/3 meetings that the count was incorrect. 

Have you been able to reproduce the bug that @MindEvo shared https://github.com/ParabolInc/parabol/pull/7241#issuecomment-1276556851?

87prashant:
@nickoferrall 
> Have you been able to reproduce the bug that @MindEvo shared https://github.com/ParabolInc/parabol/pull/7241#issuecomment-1276556851?

No, even if I don't vote on any of the stages, still I get the correct result.

@MindEvo 
> I think it's counting the stories as completed after loading them, instead of waiting for them to be scored.

Correct me if I am wrong. I think it should count the stories as completed if we visit them regardless of vote/score.

87prashant:
@Dschoordsch could you also help in reviewing this PR?

packages/server/graphql/mutations/endSprintPoker.ts
Dschoordsch: +1 I ran it through the debugger and this part of the fix works. In my test I had 3 stages, completed 1 and ended on 2. However when I stepped into this function, 1 wasn't completed so I ended up with a miscount here.

-1 Please move stage.endAt here as well to keep the stage modifying statements together.

Dschoordsch (changes requested):

87prashant:
Ok, thanks!
@Dschoordsch

Dschoordsch:
To be clear, let's focus on fixing this current bug first instead of investigating why sometimes the stage was not completed. We should do that in a separate PR.

Dschoordsch (approved):
Did a smoke test

Dschoordsch:
Thanks for the PR!

PR #7248, opened 2022-10-05
feat: Add confirmation tooltip on the retro meeting sidebar

igorlesnenko:
# Description

#4881

- Added a confirmation tooltip on the sidebar
- Changed 'Tap next' message everywhere.

## Demo

https://user-images.githubusercontent.com/466991/193296964-9ad2733e-b17f-41ed-b52a-d42f5003dfaa.mov

### DIFF
--- a/packages/client/components/RetroMeetingSidebar.tsx
+++ b/packages/client/components/RetroMeetingSidebar.tsx
@@ -1,5 +1,5 @@
 import graphql from 'babel-plugin-relay/macro'
-import React, {Fragment} from 'react'
+import React, {Fragment, useState} from 'react'
 import {createFragmentContainer} from 'react-relay'
 import useRouter from '~/hooks/useRouter'
 import isDemoRoute from '~/utils/isDemoRoute'
@@ -39,7 +39,8 @@ const RetroMeetingSidebar = (props: Props) => {
     localPhase,
     localStage,
     phases,
-    settings
+    settings,
+    meetingMembers
   } = meeting
   const {phaseTypes} = settings
   const localPhaseType = localPhase ? localPhase.phaseType : ''
@@ -48,6 +49,7 @@ const RetroMeetingSidebar = (props: Props) => {
   const isViewerFacilitator = facilitatorUserId === viewerId
   const isUnsyncedFacilitatorPhase = facilitatorPhaseType !== localPhaseType
   const isUnsyncedFacilitatorStage = localStage ? localStage.id !== facilitatorStageId : undefined
+  const [confirmingPhase, setConfirmingPhase] = useState<NewMeetingPhaseTypeEnum | null>(null)
   return (
     <NewMeetingSidebar
       handleMenuClick={handleMenuClick}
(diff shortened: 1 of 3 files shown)


igorlesnenko:
@tianrunhe does it make sense to add some metrics here?

packages/client/components/RetroMeetingSidebar.tsx
```diff
@@ -48,24 +49,47 @@ const RetroMeetingSidebar = (props: Props) => {
...
           const canNavigate = isViewerFacilitator ? isNavigableByFacilitator : isNavigable
           const handleClick = () => {
-            gotoStageId(itemStageId).catch()
-            handleMenuClick()
+            const prevPhaseType = phaseTypes[index - 1]
+            const prevItemStage = prevPhaseType
+              ? getSidebarItemStage(prevPhaseType, phases, facilitatorStageId)
+              : null
+
+            const {isComplete: isPrevItemStageComplete = false, readyCount = 0} =
```
nickoferrall, line 75: Why do we need to check whether the previous and current stages are complete? Don't we only want to know whether the current stage is complete and, if it is, we can proceed to the next stage without showing the user the tooltip warning?

nickoferrall (approved):
LGTM! Tested and it's working well

packages/client/components/RetroMeetingSidebar.tsx
igorlesnenko: @nickoferrall the reason is that I can navigate to the any of the previous phases and go back to the current phase.

* For example, I on the current phase which is not completed.
* I go back to some of the completed phases I should not see any tooltip
* Then I go back to the current phase that is still in progress, I shouldn't see any tooltip too.

Alternative approach is to not show a tooltip if phase is completed as is.
But also do not show a tooltip on current phase. But I need to determine the "current phase" somehow for which I didn't find implemented function, so I checking previous instead.

tianrunhe:
> @tianrunhe does it make sense to add some metrics here?

I feel like reviewing the tooltip is too niche to add a metric so I'm fine omitting that for now.

packages/client/components/RetroMeetingSidebar.tsx
Dschoordsch: -1 Logic seems flawed, I had some cases where the tooltip occurred where it shouldn't, for example clicking on the reflect phase right at the start
Screenshot from 2022-10-10 23-15-34

Dschoordsch (changes requested):

packages/client/components/RetroMeetingSidebar.tsx
igorlesnenko: @Dschoordsch added a fix

Dschoordsch: I think you could have defaulted `isPrevItemStageComplete = true` here to solve this particular case.

Dschoordsch (changes requested):
Change looks good.
I ran into a bug when
* advancing to "Vote"
* go back to "Group"
* go to "Discuss"
I could advance, but no discussions were available
Peek 2022-10-11 12-54

Can you please double check if this is an existing bug or introduced by your change?

igorlesnenko:
@Dschoordsch I reproduced it in master, so it is existing bug.
image

Created an issue here : https://github.com/ParabolInc/parabol/issues/7265

Dschoordsch (approved):

PR #7255, opened 2022-10-07
fix: unsubscribe analytics bug

nickoferrall:
See https://github.com/ParabolInc/parabol/pull/7190#discussion_r989552229

### DIFF
--- a/packages/server/utils/analytics/analytics.ts
+++ b/packages/server/utils/analytics/analytics.ts
@@ -317,7 +317,6 @@ class Analytics {
   }
 
   toggleSubToSummaryEmail = (userId: string, subscribeToSummaryEmail: boolean) => {
-    //
     this.track(userId, 'Summary Email Setting Changed', {subscribeToSummaryEmail})
   }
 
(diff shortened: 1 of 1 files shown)


packages/server/utils/analytics/analytics.ts
```diff
@@ -317,7 +317,7 @@ class Analytics {
   }
 
   toggleSubToSummaryEmail = (userId: string, subscribeToSummaryEmail: boolean) => {
-    this.track(userId, 'Summary Email Setting Changed', subscribeToSummaryEmail)
+    this.track(userId, 'Summary Email Setting Changed', {subscribeToSummaryEmail})
   }
 
   private track = (userId: string, event: AnalyticsEvent, properties?: any) =>
```
Dschoordsch, line 323: -1 If you see a type mismatch bug like this, you should always ask yourself if this could have been catched by typescript. In this case it could have been if the typing here would be a bit more strict.
```ts
  private track = (userId: string, event: AnalyticsEvent, properties?: Record<string, any>) =>
```

Dschoordsch (changes requested):

packages/server/utils/analytics/analytics.ts
tianrunhe: I agree with Georg, let's take this chance and improve this!

nickoferrall: Thanks, good call - updated!

Dschoordsch (approved):

PR #7257, opened 2022-10-10
fix(Jira): show only available fields in Sprint Poker

### LINKED ISSUE #7206: Jira: refactor voting field
As a Sprint Poker facilitator I want to be able to vote to the correct field. I only want to see fields which are actually available, even if these differ between issue types.

1) Right now we're reading the "available" fields from the schema with `expand=schema` on [get issue](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issues/#api-rest-api-3-issue-issueidorkey-get)). This also contains fields which might not be available (like "Story Points" or "Work Ration") we switch to explicitly read the `editmeta` information with [get editmeta](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issues/#api-rest-api-3-issue-issueidorkey-editmeta-get) which contains only fields which are available to edit.
These can also be retrieved with `expand=editmeta`.
This approach will show far fewer fields and will also hide "Story Points"/"Story Point estimates" if it is not available.
Special handling is necessary for time tracking fields (Original estimate, remaining estimate) which are treated special by Jira and need to be checked separately (one of these variations is set by roughly 60 users atm.)
2) Since we hopefully hide "Story Points"/"Story point estimate" with this new implementation, we should modify the drop down to add a more explaining "Are you missing a story point field? We can try to fix it" to run the current fix flow.
3) The automatic fix screen should link to the manual guide

Relates to #6887, #6589

## Acceptance criteria
* [ ] read field metadata via editmeta
* [ ] store voting field per issue type
  * [ ]  move Jira voting field out of Team in separate table
* [ ] allow user to trigger autofix if no variation of "Story Points" is available
* [ ] link Jira guides how to enable estimate features

Dschoordsch:
# Description

Partially Fixes #7206

Part 1/3
* store the estimation field per issue type and in a separate table than team (aligned with other integrations)
* if no estimation field is stored for the specific issue type, fall back to fields from other issue types from the same project (if these match). This also allows transitioning from the current state were no issue types are stored
* available fields are read from `editmeta` instead of `schema` which supposedly only shows fields that can be edited. This change will hide most fields from the menu.

### new JiraDimensionFieldMap table
https://www.loom.com/share/c87e4bc3fa2740e199885401c7cf429e

### reading and voting to different fields
https://www.loom.com/share/69526435b3504b98b02c7b91242cfdd6

## Testing scenarios

- See demos

- [ ] Update a field after migration
  - check Postgres JiraDimensionFieldMap table before the change, it should contain an entry for the `{teamId, cloudId, projectKey}` tuple with `issueType===''`
  - start a sprint poker with and estimate a matching Jira issue
  - [ ] check DB again and now see the correct `issueType`
  - estimate another issue type
  - [ ] see new entry in DB

- [ ] Fallback field is used correctly
  - setup a Jira project with 2 issue types, both with the same field "Story Points"
  - estimate one issue in a poker meeting and set the field to "Story Points"
  - start a new poker meeting and add the second issue type
  - [ ] the selected field should also be "Story Points"
  - [ ] after voting on this issue, a second entry is added to the DB

- [ ] Different voting fields for different issue types
  - set up a Jira project with 2 issue types, both have "Story Points" but one has an additional field "Exclusive Task Points"
  - run a poker meeting and choose different fields to estimate for both
  - [ ] run a second meeting and see that different fields are stored for different issue types

## Final checklist


nickoferrall:
Hey @Dschoordsch, are there any testing steps that you'd like us to follow?

packages/server/dataloader/atlassianLoaders.ts
```diff
@@ -215,28 +227,21 @@ export const jiraIssue = (
...
-                    )
-                  )
-                  .map(([fieldId]) => {
-                    return issueRes.names[fieldId]
-                  })
-                  .sort()
-              )
-            )
+            const possibleEstimationFieldNames = [] as string[]
+            Object.entries<any>(issueRes.editmeta?.fields)?.forEach(([fieldId, {schema}]) => {
```
nickoferrall, line 231: +1 could add a type that includes `schema` here

packages/server/graphql/mutations/setTaskEstimate.ts
```diff
@@ -102,24 +104,85 @@ const setTaskEstimate = {
...
+        const dimensionField = dimensionFields.find(({fieldName}) =>
+          validFields.includes(fieldName)
         )
-        const fieldName = dimensionField?.fieldName ?? SprintPokerDefaults.SERVICE_FIELD_NULL
+
+        // If we're using a field stored for a different issueType, update the DB to store the new match
+        if (dimensionField && dimensionField.issueType !== issueType) {
+          // Legacy unknown field type, replace it with an actual issueType
+          if (dimensionField.issueType === '') {
+            dimensionField.issueType = issueType
```
nickoferrall, line 145: +2 nice way of handling anything Jira throws at us

packages/server/graphql/public/mutations/updateJiraDimensionField.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  // RESOLUTION
+  const data = {teamId, meetingId}
+  const dimensionFields = await dataLoader
+    .get('jiraDimensionFieldMap')
+    .load({teamId, cloudId, projectKey, issueType, dimensionName})
+  const existingDimensionField = dimensionFields[0]
+  if (
+    existingDimensionField?.fieldName === fieldName &&
```
nickoferrall, line 64: +1 conditional can be removed here

packages/server/postgres/queries/getJiraDimensionFieldMap.ts
```diff
@@ -0,0 +1,17 @@
...
+} from './generated/getJiraDimensionFieldMapQuery'
+
+export interface JiraDimensionFieldMap extends IGetJiraDimensionFieldMapQueryResult {}
+export interface GetJiraDimensionFieldMapParams extends IGetJiraDimensionFieldMapQueryParams {}
+
+const getJiraDimensionFieldMap = async (params: GetJiraDimensionFieldMapParams) => {
+  // pg-typed doesnt' support records, so we can't use multiple composite keys
+  // https://github.com/adelsz/pgtyped/issues/317
+  const res = await getJiraDimensionFieldMapQuery.run(params as any, getPg())
+  return res as JiraDimensionFieldMap[]
```
nickoferrall, line 15: Why do we need the casting here and `as any` on the line above?

packages/client/components/JiraFieldMenu.tsx
```diff
@@ -23,7 +23,7 @@ const JiraFieldMenu = (props: Props) => {
   const {meetingId, dimensionRef, serviceField, task} = stage
   if (task?.integration?.__typename !== 'JiraIssue') return null
   const {integration} = task
-  const {cloudId, projectKey, possibleEstimationFieldNames} = integration
```
nickoferrall, line 26: +1 this isn't the line of code where the problem is happening, but I wanted to note it down:

If a project doesn't have a Story Points field, the menu should probably not show the divider or the padding at the top:
jira-menu

nickoferrall (approved):
Great work! This passed all the testing steps and the code LGTM. I imagine this was a painful one to work on - I'm excited that there's light at the end of the Jira bug tunnel 🙌

Dschoordsch:
@tianrunhe you were added as a reviewer automatically because I added a migration. Do you think we should add any analytics to any of the logic?

tianrunhe (changes requested):
Thanks for tagging me! Or thanks the robot for tagging me 🤣

While we are here, can we take the chance to add the missing metrics for the mutation `updatePokerScope`? Thank you!

- Event name: `Poker Meeting Scope Updated`
- Properties:
  - `meetingId`
  - `service`: the `TaskServiceEnum`
  - `taskId`
  - `action`: `ADD` or `DELETE`

Dschoordsch:
@igorlesnenko could you please also take a look? Because this is a bigger change, I would feel more comfortable if 2 people would review it.

igorlesnenko (approved):

Dschoordsch:
@tianrunhe We already have some tracking for this
https://github.com/ParabolInc/parabol/blob/49df2b9aaea8f645ccd81ec3b3202c2c286a8dae/packages/client/mutations/UpdatePokerScopeMutation.ts#L300
It was added on the client side, because only there we know whether or not the user used "select all"

tianrunhe (approved):
Sign off from metrics side

igorlesnenko (changes requested):
Posting the issue in the correct PR:
Getting an error while running migration
```
> Migrating files:
> - 1663850062796_JiraDimensionFieldMap
> Rolling back attempted migration ...
error: duplicate key value violates unique constraint "JiraDimensionFieldMap_pkey"
```

igorlesnenko (approved):
The new migration works

PR #7261, opened 2022-10-10
feat(metrics): Send isPatient0 property to Google Analytics

### LINKED ISSUE #7182: Add `isPatient0` in `Account Created` events
## Issue - Enhancement
We'd like to track the conversation rate for patient-0 sign-ups. Just like today we add label for `isInvited` [here](https://github.com/ParabolInc/parabol/blob/master/packages/server/graphql/mutations/helpers/bootstrapNewUser.ts#L39), we need another label/value pair there to determine if the sign-up is a patient-0 user.

- [ ] When a `Account Created` event is emit, a new label/value pair of `isPatient0` is included.

tianrunhe:
# Description

Fixes #7182 

## Demo
Screen Shot 2022-10-10 at 11 28 29 AM

## Testing scenarios

- [ ] Patient 0
  - Create an account with a new domain
  - In Segment verify the value for label `isPatient0` is 1

- [ ] Non-patient 0
  - Create an account with an existing domain
  - In Segment verify the value for label `isPatient0` is 0

## Final checklist

### DIFF
--- a/packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
+++ b/packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
@@ -8,44 +8,17 @@ import User from '../../../database/types/User'
 import generateUID from '../../../generateUID'
 import insertUser from '../../../postgres/queries/insertUser'
 import IUser from '../../../postgres/types/IUser'
+import {analytics} from '../../../utils/analytics/analytics'
 import segmentIo from '../../../utils/segmentIo'
 import addSeedTasks from './addSeedTasks'
 import createNewOrg from './createNewOrg'
 import createTeamAndLeader from './createTeamAndLeader'
 import isPatientZero from './isPatientZero'
 
-// no waiting necessary, it's just analytics
-const handleSegment = async (user: User, isInvited: boolean) => {
-  const {id: userId, createdAt, email, featureFlags, tier, segmentId, preferredName} = user
-  const domain = email.split('@')[1]
-  segmentIo.identify({
-    userId,
-    traits: {
-      createdAt,
-      email,
-      name: preferredName,
-      isActive: true,
-      featureFlags,
-      highestTier: tier,
-      isPatient0: await isPatientZero(userId, domain)
-    },
-    anonymousId: segmentId
-  })
-  segmentIo.track({
-    userId,
-    event: 'Account Created',
-    properties: {
-      isInvited,
-      // properties below required for Google Analytics destination
-      category: 'All',
-      label: 'isInvited',
-      value: isInvited ? 1 : 0
-    }
-  })
-}
-
 const bootstrapNewUser = async (newUser: User, isOrganic: boolean) => {
-  const {id: userId, preferredName, email} = newUser
+  const {id: userId, createdAt, preferredName, email, featureFlags, tier, segmentId} = newUser
+  const domain = email.split('@')[1]
+  const isPatient0 = await isPatientZero(userId, domain)
   const r = await getRethink()
   const joinEvent = new TimelineEventJoinedParabol({userId})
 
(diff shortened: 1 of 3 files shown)


packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
```diff
@@ -8,16 +8,29 @@ import User from '../../../database/types/User'
...
+
+  await Promise.all([
+    r({
+      user: r.table('User').insert(newUser),
+      event: r.table('TimelineEvent').insert(joinEvent)
+    }).run(),
+    insertUser(newUser)
+  ])
+
+  // Identify the user so user properties are set before any events are sent
```
tianrunhe, line 33: This is an attempt to fix the issue where user identification happens after the `Account Created` event. For example, [this user](https://analytics.amplitude.com/parabol/project/224689/search/amplitude_id%3D472337852118?collapsed=true&sessionHandle=YOxmKx2_YOxmLBi_G35h6bW_22x_c56a5850-6e89-41b6-8acf-d72bac53aff7R&eventId=795b708b-461b-11ed-ba9e-49b6f46c1139) 's `Account Created` event is not tagged with isPatient0 property.

packages/server/graphql/mutations/helpers/isPatientZero.ts
```diff
@@ -1,6 +1,16 @@
 import getPatientZeroByDomain from '../../../postgres/queries/getPatientZeroByDomain'
 
-const isPatientZero = async (userId: string, domain?: string | null) =>
-  domain ? userId === (await getPatientZeroByDomain(domain)).id : false
+const isPatientZero = async (userId: string, domain?: string | null) => {
+  if (domain) {
+    const patient0 = await getPatientZeroByDomain(domain)
+    if (!patient0) {
```
tianrunhe, line 6: Fix a bug where when the new user is actually indeed a patient 0, `getPatientZeroByDomain(domain)` would return `undefined` and that causes accessing `.id` throw an error.

packages/server/utils/analytics/analytics.ts
```diff
@@ -320,6 +321,16 @@ class Analytics {
     this.track(userId, 'Summary Email Setting Changed', subscribeToSummaryEmail)
   }
 
+  accountCreated = (userId: string, isInvited: boolean, isPatient0: boolean) => {
+    this.track(userId, 'Account Created', {
+      isInvited,
+      // properties below needed for Google Analytics goal setting
+      category: 'All',
+      label: 'isPatient0',
+      value: isPatient0 ? 1 : 0
```
tianrunhe, line 330: You might notice that in the current version, this part of the code is:

```
      category: 'All',
      label: 'isInvited',
      value: isInvited ? 1 : 0
```

This has been overwritten instead of preserved because we are not actually using it anywhere.

jmtaber129, line 330: Looking at the [GA docs](https://support.google.com/analytics/answer/1033068#Anatomy&zippy=%2Cin-this-article), it's unclear whether this is actually how we should pass p0 information to GA. I.e. it sounds like `value` here is associated with the event, rather than the label. If I understand the GA docs correctly, we should actually only be sending `label`s that are either `isPatientZero` or `isNotPatientZero` (or similar).

Are we able to filter by both label and value in GA to do this analysis? What will the GA chart look like using this information?

jmtaber129 (commented):
The code looks good, but I think this may be misusing the event props that get passed to GA - see inline comment.

packages/server/utils/analytics/analytics.ts
tianrunhe: I think you are right! I was misled by the existing code. Thanks for pointing this out!

jmtaber129 (approved):
LGTM! It might be good to have someone on Growth confirm that these are the labels we want, but I have no tensions if you want to just merge this.

Issue #6144, opened 2022-03-01
Bug: billing leader NOT reassigned if the current one's account is deleted


tianrunhe:
## Issue - Bug

If the user is the billing leader for an org and his/her account is deleted, the org will have no billing leaders and no one else in the org can claim that. 

We should fix it.

### Acceptance Criteria (optional)
- [ ] When a billing leader's account is deleted, reassigning the role to someone else in the org

jordanh:
Keeping this one open

PR #7267, opened 2022-10-11
fix: assign new billing leader after deleting billing leader

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/6144

https://www.loom.com/share/81c633b5d1304c6fb8e8a76ee07c916a

### DIFF
--- a/packages/server/graphql/mutations/helpers/removeFromOrg.ts
+++ b/packages/server/graphql/mutations/helpers/removeFromOrg.ts
@@ -87,7 +87,6 @@ const removeFromOrg = async (
         await resolveDowngradeToPersonal(orgId, organization.stripeSubscriptionId!, userId)
       }
     }
-    // }
   }
   try {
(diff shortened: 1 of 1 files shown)


tianrunhe (changes requested):
The server side test [`Delete user`](https://app.circleci.com/pipelines/github/ParabolInc/parabol/7123/workflows/8f33e8e1-f858-40cc-a1e7-140838a7ed30/jobs/15927/tests) seems to fail consistently. Maybe you need to update the test code?

nickoferrall:
Thanks, @tianrunhe. The rethinkdb query had an additional bug - if there were no other users in the team, it would return an error. This wasn't getting picked up by the test as the rethink query would only be run for non-free users, and the test user is on the free tier. Update pushed!

tianrunhe (approved):

Dschoordsch (approved):
Tested, works.

Issue #7104, opened 2022-08-24
"copy invite link" moves on scroll


87prashant:
## Issue - Bug

https://user-images.githubusercontent.com/106697681/186346510-7bf1cc41-5f44-4bc0-af2a-99b85774dd9a.mp4

### Acceptance Criteria (optional)
It should be fixed below 3 dots

PR #7275, opened 2022-10-12
fix: copy invite link moves on scroll

gitstart:
# Description

Fixes [#7104]
I fixed the card dropdown menus to ensure their positions are fixed to their parent cards as required

## Demo
[loom video](https://www.loom.com/share/e799d2ad062744a9ab514c364cbb53b5)

## Testing scenarios

- [ ] On the meetings page
  - Click the menu button on any of the meeting cards
  - Scroll up or down
  - The menu should not be fixed to the same screen position, rather, it should move along with its parent card

## Final checklist


gitstart:
Hi @Dschoordsch, @igorlesnenko, @JimmyLv  this PR is ready for review

packages/client/hooks/useMenu.ts
```diff
@@ -53,11 +63,30 @@ const useMenu = <T extends HTMLElement = HTMLButtonElement>(
...
+      portalStatus === PortalStatus.Mounted ||
+      portalStatus === PortalStatus.Entering ||
+      portalStatus === PortalStatus.Entered
+    ) {
+      setTopOffset(cardContainerRef?.current?.scrollTop || 0)
+    }
+    return (coords as {right: number; top: number}).top
+      ? {
+          ...coords,
+          top: (coords as {right: number; top: number}).top - 64 + topOffset
```
JimmyLv, line 80: -1 what's this magic number? 64

packages/client/hooks/useMenu.ts
```diff
@@ -53,11 +63,30 @@ const useMenu = <T extends HTMLElement = HTMLButtonElement>(
...
+
+  const menuCoords = useMemo(() => {
+    if (
+      portalStatus === PortalStatus.Mounted ||
+      portalStatus === PortalStatus.Entering ||
+      portalStatus === PortalStatus.Entered
+    ) {
+      setTopOffset(cardContainerRef?.current?.scrollTop || 0)
+    }
+    return (coords as {right: number; top: number}).top
```
JimmyLv, line 77: -1 it's better to extract a type for coords firstly.

packages/client/hooks/useMenu.ts
```diff
@@ -53,11 +63,30 @@ const useMenu = <T extends HTMLElement = HTMLButtonElement>(
...
   }, [originRef.current /* eslint-disable-line react-hooks/exhaustive-deps */])
   const {loadingDelay, loadingDelayRef} = useLoadingDelay()
+
+  const [topOffset, setTopOffset] = useState<number>(0)
+
+  const menuCoords = useMemo(() => {
+    if (
+      portalStatus === PortalStatus.Mounted ||
+      portalStatus === PortalStatus.Entering ||
+      portalStatus === PortalStatus.Entered
```
JimmyLv, line 73: +1 this could better simplify to `[PortalStatus.Mounted, PortalStatus.Entering, PortalStatus.Entered].includes(portalStatus)`

packages/client/hooks/useMenu.ts
JimmyLv: -2 I think you can use `'key' in object`, ref: https://www.freecodecamp.org/news/how-to-check-if-an-object-has-a-key-in-javascript/

packages/client/hooks/usePortal.tsx
```diff
@@ -38,7 +38,7 @@ export interface UsePortalOptions {
   onClose?: () => void
   id?: PortalId
   // if you nest portals, this should be the id of the parent portal
-  parentId?: PortalId
+  parentId?: PortalId | string
```
JimmyLv, line 41: -1 the parent portal is enumerable, if you need to change it to `string` it becomes generic, need to rethink about it

JimmyLv (changes requested):
Thanks! No problem with the functionality, but the code needs to be improved 😁

packages/client/hooks/useMenu.ts
gitstart: That was actually a hack we had to use to get the menu in the right position. I will store it in a constant variable for better readability.

packages/client/hooks/usePortal.tsx
gitstart: Never mind, I was able to get a workaround, pushing soon.

packages/client/hooks/useMenu.ts
```diff
@@ -53,11 +65,28 @@ const useMenu = <T extends HTMLElement = HTMLButtonElement>(
...
+      [PortalStatus.Mounted, PortalStatus.Entering, PortalStatus.Entered].includes(portalStatus)
+    ) {
+      setTopOffset((cardContainerRef?.current?.scrollTop || 0) - TOP_POSITION_EXCESS)
+    }
+    return 'top' in coords
+      ? {
+          ...coords,
+          top: coords.top + topOffset
+        }
+      : coords
```
JimmyLv, line 82: -1 it's better to refactor as a `guard clauses`, ref: https://refactoring.guru/replace-nested-conditional-with-guard-clauses

JimmyLv (changes requested):

JimmyLv (approved):

mattkrick:
Thank you for the PR!
Would it be weird in the menu gets scrolled offscreen & the keyboard events still targeted the menu?
Looking around at other menu implementations, I noticed that they try to emulate native behavior by preventing scrolling when a menu is open:

- https://mui.com/material-ui/react-menu/
- https://www.radix-ui.com/docs/primitives/components/dropdown-menu

what do you think about an implementation that did something similar?

gitstart-parabol:
> Thank you for the PR! Would it be weird in the menu gets scrolled offscreen & the keyboard events still targeted the menu? Looking around at other menu implementations, I noticed that they try to emulate native behavior by preventing scrolling when a menu is open:
> 
> * https://mui.com/material-ui/react-menu/
> * https://www.radix-ui.com/docs/primitives/components/dropdown-menu
> 
> what do you think about an implementation that did something similar?

I think it can be done if required, but it's probably not a good UX in my opinion.

gitstart-parabol:
What is the verdict on this? @mattkrick

Dschoordsch:
> Thank you for the PR! Would it be weird in the menu gets scrolled offscreen & the keyboard events still targeted the menu? Looking around at other menu implementations, I noticed that they try to emulate native behavior by preventing scrolling when a menu is open:
> 
> * https://mui.com/material-ui/react-menu/
> * https://www.radix-ui.com/docs/primitives/components/dropdown-menu
> 
> what do you think about an implementation that did something similar?

There are 2 solutions possible and I saw both in the material ui examples:
1) what Matt said: prevent scrolling of the page when the menu is open
2) do not prevent scrolling, but if the focus on the menu changes, scroll it back into view (probably default browser behaviour?)

At the moment keyboard focus seems to be broken on master and on this PR. It would be good to know if this change works well together with #7350.
I did not check yet, but if that would already do 2), then that would be good enough. I feel like 1) is more suited for bigger/more important menus and it would need to give some visual clue why scrolling is not possible anymore so it's not surprising for the user. All this seems overkill in this situation.

gitstart-parabol:
> > Thank you for the PR! Would it be weird in the menu gets scrolled offscreen & the keyboard events still targeted the menu? Looking around at other menu implementations, I noticed that they try to emulate native behavior by preventing scrolling when a menu is open:
> > 
> > * https://mui.com/material-ui/react-menu/
> > * https://www.radix-ui.com/docs/primitives/components/dropdown-menu
> > 
> > what do you think about an implementation that did something similar?
> 
> There are 2 solutions possible and I saw both in the material ui examples:
> 
> 1. what Matt said: prevent scrolling of the page when the menu is open
> 2. do not prevent scrolling, but if the focus on the menu changes, scroll it back into view (probably default browser behaviour?)
> 
> At the moment keyboard focus seems to be broken on master and on this PR. It would be good to know if this change works well together with #7350. I did not check yet, but if that would already do 2), then that would be good enough. I feel like 1) is more suited for bigger/more important menus and it would need to give some visual clue why scrolling is not possible anymore so it's not surprising for the user. All this seems overkill in this situation.

@Dschoordsch I think it would be simpler to merge this PR with #7350 and then, if 2) needs to be implemented, we can raise a separate issue to handle that. What do you think?

Dschoordsch:
Sounds good.

Dschoordsch:
Waiting on #7350 to be merged to ensure these two work well together

mattkrick:

Issue #7221, opened 2022-09-22
Retro template: Drop Add Keep Improve (DAKI) is missing Keep stage


Dschoordsch:
https://parabol.slack.com/archives/C4JAUUZ9P/p1662736860901099

## Acceptance criteria
* [ ] create mutation to add Keep stage
* [ ] ensure the stage has no duplicate key with other stages

PR #7279, opened 2022-10-12
fix: missing daki prompt

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7221

To test:

- Run the up migration, open the retro template menu, see that the DAKI template includes the "Keep" prompt
- Running the down migration removes the prompt

I'm not really sure why/how the prompt doesn't exist in the db when it was included here https://github.com/ParabolInc/parabol/blob/ec1cc40f177b02a0a5f37b8e7cd7b422b3f7961e/packages/server/database/migrations/20210602120505-addAdditionalRetroTemplates.ts#L80 and the other three prompts are all there 🤔

### DIFF
--- a/packages/server/database/migrations/20221012142324-fixDakiMigration.ts
+++ b/packages/server/database/migrations/20221012142324-fixDakiMigration.ts
@@ -1,0 +1,24 @@
+import {R} from 'rethinkdb-ts'
+
+export const up = async function (r: R) {
+  const now = new Date()
+  const missingKeepPrompt = {
+    createdAt: now,
+    description: 'What should we continue doing?',
+    groupColor: '#D9D916',
+    id: 'keepPrompt',
+    isActive: true,
+    question: 'Keep',
+    sortOrder: 2,
+    teamId: 'aGhostTeam',
+    templateId: 'dropAddKeepImproveDAKITemplate',
+    title: 'Keep',
+    updatedAt: now
+  }
+
+  await r.db('actionDevelopment').table('ReflectPrompt').insert(missingKeepPrompt).run()
+}
+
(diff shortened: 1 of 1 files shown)


igorlesnenko (approved):
Yeah, this is super weird that this prompt is missing. As it does not exists in both local and prod DB, I guess it could be some migration that removed it?

packages/server/database/migrations/20221012142324-fixDakiMigration.ts
```diff
@@ -0,0 +1,23 @@
...
+    id: 'keepPrompt',
+    isActive: true,
+    question: 'Keep',
+    sortOrder: 2,
+    teamId: 'aGhostTeam',
+    templateId: 'dropAddKeepImproveDAKITemplate',
+    title: 'Keep',
+    updatedAt: now
+  }
+  await r.db('actionDevelopment').table('ReflectPrompt').insert(missingKeepPrompt).run()
```
Dschoordsch, line 18: -1 this is the wrong database. Usually we don't put the database here at all because it is configured somewhere.

packages/server/database/migrations/20221012142324-fixDakiMigration.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export const down = async function (r: R) {
+  await r.db('actionDevelopment').table('ReflectPrompt').get('keepPrompt').delete().run()
```
Dschoordsch, line 22: -1 see above

Dschoordsch (changes requested):

Dschoordsch (approved):

Issue #7249, opened 2022-10-05
Show 'You' instead of current user name when displaying users who added a reaction


BartoszJarocki:
+1 Possible improvement: I like how Slack etc. replace the viewers name with "You".
![Screenshot from 2022-10-04 17-16-12](https://user-images.githubusercontent.com/7331043/193858720-15116fe5-1edb-4a43-896d-c3267ebf0724.png)

_Originally posted by @Dschoordsch in https://github.com/ParabolInc/parabol/pull/7147#discussion_r987015644_

PR #7285, opened 2022-10-12
feat: show you in emoji list instead of viewer name

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7249

### To test

- [ ] If the viewer clicks on an emoji, the list says `You` instead of their name
- [ ] If multiple people including the viewer click on the emoji, the `You` appears first in the list, just like in Slack, i.e. `You and Susan reacted with :fire:` instead of `Susan and you reacted with :fire:`


BartoszJarocki (approved):
lgtm!

PR #7310, opened 2022-10-18
POC: Generic notification publishing

jmtaber129:
POC for making the "publish notification" flow more generic. Instead of associating new notifications specifically with the mutation that triggered them (e.g. "someone posted a standup response with your name in it, so here's a notification triggered by that mutation"), new notifications can just be published via a generic `AddedNotification` type (e.g. "here's a notification telling you about how you were mentioned in someone's standup response").

Benefits of this approach:
* Most boilerplate associated with adding a new notification is abstracted away
  * New notifications don't need changes to the `NotificationSubscription` or the triggering mutation
  * New notifications don't need to implement an updater to handle adding the notification in relay
  * See these changes needed for standup mention notifications:
    * https://github.com/ParabolInc/parabol/pull/7118/files#diff-21aeeb7bbbb0e6f8da9f5d79ce1a47e92a1b7f72f01782977288af585cda119e
    * https://github.com/ParabolInc/parabol/pull/7118/files#diff-4ae049d67f824570f8c2a3f8d68fd1ceb8a332209961392f00cd2e7a8f94955b
    * https://github.com/ParabolInc/parabol/pull/7118/files#diff-a23acb52670a08c06e8f8e7bdf09bd98d7983a0ae5566d8b463720f762682e6f
    * https://github.com/ParabolInc/parabol/pull/7118/files#diff-a6d5286e029f906a1d5fb60f8847bb2d6730cc8617c3be60a95d239e68ff946b
* Creating a new notification only requires:
  * Adding the notification component + fragment to the notification picker
  * Adding the notification's GQL + Database models
  * Calling `publish` _somewhere_ on the server.

Another way you might think about this at a high level is going from "this mutation just triggered a specific type of notification, so here's that notification, make sure you handle it when you handle the mutation" to "here's the ID for a new notification, we don't care where it came from, just resolve it and load it".

Other raw notes: https://www.notion.so/parabol/James-Notes-Notifications-publish-refactor-fc28fe02d76d4b3da1b21ebe0a0f278c

### DIFF
--- a/packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
+++ b/packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
@@ -94,8 +94,10 @@ const upsertTeamPromptResponse: MutationResolvers['upsertTeamPromptResponse'] = 
     publish(
       SubscriptionChannel.NOTIFICATION,
       notification.userId,
-      'UpsertTeamPromptResponseSuccess',
-      data,
+      'AddedNotification',
+      {
+        addedNotificationId: notification.id
+      },
       subOptions
     )
   })
--- a/packages/server/graphql/public/types/AddedNotification.ts
+++ b/packages/server/graphql/public/types/AddedNotification.ts
@@ -1,0 +1,18 @@
+import {getUserId} from '../../../utils/authorization'
+import {AddedNotificationResolvers} from '../resolverTypes'
+
+export type AddedNotificationSource = {addedNotificationId: string}
+
+const AddedNotification: AddedNotificationResolvers = {
+  addedNotification: async ({addedNotificationId}, _args: unknown, {dataLoader, authToken}) => {
+    const viewerId = getUserId(authToken)
+    const notification = (await dataLoader.get('notifications').load(addedNotificationId)) as any
+    if (notification.userId === viewerId) {
+      return notification
+    } else {
+      return null
+    }
+  }
+}
+
+export default AddedNotification
--- a/codegen.json
+++ b/codegen.json
@@ -33,6 +33,7 @@
         "contextType": "../graphql#GQLContext",
         "mappers": {
           "AddApprovedOrganizationDomainsSuccess": "./types/AddApprovedOrganizationDomainsSuccess#AddApprovedOrganizationDomainsSuccessSource",
+          "AddedNotification": "./types/AddedNotification#AddedNotificationSource",
           "AddFeatureFlagPayload": "./types/AddFeatureFlagPayload#AddFeatureFlagPayloadSource",
           "Company": "./types/Company#CompanySource",
           "CreateImposterTokenPayload": "./types/CreateImposterTokenPayload#CreateImposterTokenPayloadSource",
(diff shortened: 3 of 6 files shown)


packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
```diff
@@ -94,8 +94,10 @@ const upsertTeamPromptResponse: MutationResolvers['upsertTeamPromptResponse'] =
...
       SubscriptionChannel.NOTIFICATION,
       notification.userId,
-      'UpsertTeamPromptResponseSuccess',
-      data,
+      'AddedNotification',
+      {
+        addedNotificationId: notification.id
+      },
       subOptions
     )
```
jmtaber129, line 102: This could be abstracted into something like `publishNotification(notification, subOptions)`.

packages/server/graphql/public/types/AddedNotification.ts
```diff
@@ -0,0 +1,18 @@
...
+export type AddedNotificationSource = {addedNotificationId: string}
+
+const AddedNotification: AddedNotificationResolvers = {
+  addedNotification: async ({addedNotificationId}, _args: unknown, {dataLoader, authToken}) => {
+    const viewerId = getUserId(authToken)
+    const notification = (await dataLoader.get('notifications').load(addedNotificationId)) as any
+    if (notification.userId === viewerId) {
+      return notification
+    } else {
+      return null
```
mattkrick, line 13: -1 this auth check shouldn't be necessary because the `publish` call is only sending 1 message to 1 user. if we do hit the null condition, that tells us we're sending an unnecessary message to the pubsub, and we should try to avoid that! no sense in having the server do all that work just to send `null` to the client.

if we make that change, we can make the return type non-null

packages/server/graphql/public/mutations/upsertTeamPromptResponse.ts
mattkrick: i like it! waaay long ago we used to use a similar pattern for all entities (e.g. AddedTeam, UpdatedTeam, RemovedTeam) it worked really well! Eventually we had to switch because the client side logic sometimes needed to know which event triggered the change, which fields specifically were updated, etc.

I think something like this where it's only handling the C in CRUD & only handling the Notification entity is a great compromise on simplicity & flexibility. The question we have to ask ourselves is, "If 2 different mutations can create the same notification, will the client care which mutation caused it?" For example, if someone gets invited to a team via a push invitation, does the client treat it differently vs. if another mutation triggered the invited to team notification? If that case arises, is there a workaround (e.g. fall back to using the SuccessPayload?)

tl;dr if you can't think of any cases where the client needs to know the mutation that triggered it, it's probably a safe pattern!

jmtaber129: I think the way most notifications get loaded generally means that the client can't care which mutation triggered it - when notifications are loaded on page load, we just load the notification object with no context into where the notification came from.

There _may_ be certain exceptions to this for e.g. team invitations or org removals, but those are less "hey, just letting you know about this thing!" and more "something very specific just changed and we need to change some state". For notifications that only need us to add the notification to the notifications drawer (and maybe pop a toast based on the notification type), this should work fine.

mattkrick: totally agree!

Issue #7297, opened 2022-10-14
Reflection card visually disappears when you try to drag it from an expanded group of cards


BartoszJarocki:
First seen in 6.77.0, since then it occurred > 900 times. 

https://sentry.io/organizations/parabol/issues/3649823309/?referrer=github_plugin

```
TypeError: Cannot read properties of null (reading 'cloneNode')
  at cloneReflection (./packages/client/utils/retroGroup/cloneReflection.ts:13:25)
  at current (./packages/client/hooks/useDraggableReflectionCard.tsx:335:20)
  at apply (./packages/client/hooks/useEventCallback.ts:14:16)
```

igorlesnenko:
Looks like happens when you try to drag a card from an expanded group of cards:

https://user-images.githubusercontent.com/466991/196685027-acc04743-8689-4ee2-adc0-9387834f72a7.mov

Dschoordsch:
Might be related to #6910 `packages/client/hooks/useDraggableReflectionCard.tsx`

PR #7312, opened 2022-10-19
fix: avoid overriding custom group name with a single card

### LINKED ISSUE #7207: Bug: Auto Group titles should not erase manually input title
## Issue - Bug

[A user](https://app.hubspot.com/contacts/3888472/contact/77510551) 🔒 reports:
> There was one particular thing that's annoying, when grouping cards, it's trying to be smart and auto-suggest the name of the group, but that also overrides the "custom" name that was given, this was the most annoying thing we noticed as we often try to incrementally group things and each time adding a new card to the group made us lose the name.

I quickly test this on our demo team and was unable to reproduce it single player. I wonder if this only happens in multiplayer (e.g. if the Viewer isn't the one to set the title)?

I suggest the first step with the bug is testing it multiplayer and seeing if it can be reproduced... This would be very annoying indeed!

Performed triage to find the root cause of the bug, timeboxed to ~1 hour. 

igorlesnenko:
Fixes #7207

## Demo

https://user-images.githubusercontent.com/466991/196692055-d2529cbe-c823-4c7f-b651-25f820e08e21.mov

## Testing scenarios

A:

- Give a custom title to a single card group.
- Drag that card to another card (with no group title) to create a group.
- See that the custom title was transferred to a new group correctly

B:
- Smoke test other drag & drop scenarios with smart title and custom titles


igorlesnenko:
@MindEvo can you please help testing this PR? I want to make sure that these changes didn't break any other logic

packages/server/graphql/mutations/helpers/updateReflectionLocation/addReflectionToGroup.ts
```diff
@@ -63,7 +71,25 @@ const addReflectionToGroup = async (
     }).run()
 
     const nextTitle = getGroupSmartTitle(nextReflections)
-    await updateSmartGroupTitle(reflectionGroupId, nextTitle)
+    if (
+      reflectionGroup.title === reflectionGroup.smartTitle &&
+      nextReflections.length === 2 &&
```
nickoferrall, line 76: -1 why do we want to update the smart title only if the nextReflection length is 2? If the nextReflection group has 3 reflections but has a smart title, and the oldReflection group has a custom title, would we want to update the title of the new group?

Perhaps something like this would work?
```
    const newGroupHasCustomTitle = reflectionGroup.title !== nextTitle
    const oldGroupHasCustomTitle = oldReflectionGroup.title !== oldReflectionGroup.smartTitle

    if (!newGroupHasCustomTitle && oldGroupHasCustomTitle) {
```

nickoferrall (changes requested):

MindEvo:
There's a bug where if you select a group of cards then try to drag one of them it throws the TypeError message. 

[Loom Message - 19 October 2022 - Watch Video](https://www.loom.com/share/a43ee41426fe42488b086cfd63035fbe)

igorlesnenko:
@MindEvo thank you! we have a ticket for this bug: #7297

packages/server/graphql/mutations/helpers/updateReflectionLocation/addReflectionToGroup.ts
igorlesnenko: @nickoferrall here I only handle a case where you drag a single card with a title on another **single card** with no visible title.

`nextReflections.length === 2` is a bit confusing, as when we do this check the dragged card is already added to the group, so length `2` includes the dragged card.

But I think it make sense to override any smart title if you drag a card with a custom title over it.

I updated the code and changed the logic

packages/server/graphql/mutations/helpers/updateReflectionLocation/addReflectionToGroup.ts
```diff
@@ -63,7 +71,23 @@ const addReflectionToGroup = async (
     }).run()
 
     const nextTitle = getGroupSmartTitle(nextReflections)
-    await updateSmartGroupTitle(reflectionGroupId, nextTitle)
+    const oldGroupHasSingleReflectionCustomTitle =
+      oldReflectionGroup.title !== oldReflectionGroup.smartTitle && oldReflections.length === 0
```
nickoferrall, line 75: +1 why do you prefer to only update the new group's title if the old group contains 1 reflection? If the old group has two reflections and a custom title and the new group has a smart title, wouldn't we still want to use the custom title from the old group?

nickoferrall (approved):

packages/server/graphql/mutations/helpers/updateReflectionLocation/addReflectionToGroup.ts
igorlesnenko: @nickoferrall My logic is the following:
If a group has > 1 reflection and a custom title, e.g. `Oranges`. And we move one reflection out of that group, we consider that the reflection does not belong to that title (`Oranges`), and we don't need to apply same title to one more group, i.e. have two `Oranges` groups.

Dschoordsch (approved):
Did a short test and it works

Issue #7301, opened 2022-10-18
Can end a meeting from meetings view


nickoferrall:
Sometimes meetings are buggy and when you click on them, you get redirected to the meetings view. When that happens, it would be helpful if you could end the meeting from the meetings view. Currently, the buggy meeting will live on forever in the meetings view. 

We could add an "End meeting" option to this menu:
![meetings-view](https://user-images.githubusercontent.com/39854876/196437748-c30ad878-5e3a-4a9f-a2bb-1c5626ec77c5.png)

See for https://github.com/orgs/ParabolInc/projects/6/views/8 further context.

### AC

- [ ] Can end a meeting from the meeting card dropdown menu in the meetings view

PR #7313, opened 2022-10-19
feat: end meeting from meetings view

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7301

https://www.loom.com/share/f2c8aaab283c4d2d9d9a90d05b1bc132

### To test

- [ ] Open dropdown menu of a meeting card where the viewer is the facilitator. Click the "End the meeting" button and end the meeting
- [ ] If the viewer isn't the facilitator, the "End the meeting" option doesn't show up in the menu
- [ ] If it's a standup meeting, the "End the meeting" option always shows up

### DIFF
--- a/packages/client/components/MeetingCardOptionsMenu.tsx
+++ b/packages/client/components/MeetingCardOptionsMenu.tsx
@@ -61,8 +61,8 @@ const query = graphql`
       }
       meeting(meetingId: $meetingId) {
         id
-        facilitatorUserId
         meetingType
+        facilitatorUserId
       }
     }
   }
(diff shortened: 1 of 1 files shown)


igorlesnenko (approved):
Looks good to me. Will finally clear my list of opened meetings easily :)

Could maybe be improved by adding snackbar and confirmation in future

Dschoordsch (approved):
I think it looks good but might need some amplitude event, @tianrunhe what do you think?

tianrunhe:
> I think it looks good but might need some amplitude event, @tianrunhe what do you think?

Thanks for tagging! I don't have much tension at the moment. We already have `Meeting Ended` events today. Although it'd be interesting to distinguish between a "normal meeting end" vs. "buggy meeting kill" but it feels a too niche use case.

Maybe, maybe we can add a property into the `Meeting Ended` event, to indicate where the meeting is ended from? Within a meeting or from the meeting card? But I realize that involves modifying 4 mutations so I'll defer to @nickoferrall

nickoferrall:
Hmm I think because Sentry should still track the buggy meeting and because we'd need to update 4 mutations (and remember to update any new meeting mutations created in the future), I'd prefer to continue without it. Plus ending a meeting from the meetings view wouldn't necessarily mean it's a buggy meeting end - they might just prefer to end it from the meetings view

PR #7323, opened 2022-10-21
feat: rename christmas template

nickoferrall:
This PR renames the Christmas template following this request: https://parabol.slack.com/archives/C836NA350/p1666361083087839

The id of meeting templates is based on the template name, which isn't ideal. I've updated the template name which means its id is out of sync with the name. I think that's OK, but an alternative solution could be removing the template from the db and inserting a new template with the updated name & id. 

### To test

- [ ] Run the migration and see that the `A Christmas Carol Retrospective` name is now `Christmas Retrospective`

### DIFF
--- a/packages/server/database/migrations/20221021163318-rename-christmas-template.ts
+++ b/packages/server/database/migrations/20221021163318-rename-christmas-template.ts
@@ -1,0 +1,18 @@
+import {R} from 'rethinkdb-ts'
+
+export const up = async function (r: R) {
+  await r
+    .table('MeetingTemplate')
+    .get('aChristmasCarolRetrospectiveTemplate')
+    .update({name: 'Christmas Retrospective'})
+    .run()
+}
+
+export const down = async function (r: R) {
+  //
+  await r
+    .table('MeetingTemplate')
+    .get('aChristmasCarolRetrospectiveTemplate')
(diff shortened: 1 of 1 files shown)


packages/server/database/migrations/20221021163318-rename-christmas-template.ts
```diff
@@ -0,0 +1,17 @@
+import {R} from 'rethinkdb-ts'
+
+export const up = async function (r: R) {
+  await r
+    .table('MeetingTemplate')
+    .get('aChristmasCarolRetrospectiveTemplate')
+    .update({name: 'Christmas Retrospective'})
```
Dschoordsch, line 7: +1 I would assume we want to keep Santa, I asked clarification question in Slack.

Dschoordsch (changes requested):

packages/server/database/migrations/20221021163318-rename-christmas-template.ts
Dschoordsch: Yes, confirmed, we want to keep Santa.

nickoferrall: Thanks, Santa's back!

Dschoordsch (approved):

PR #7325, opened 2022-10-21
fix: put isPatient0 on User row

mattkrick:
# Description

When prod went down today, every single connection was stuck on the `isPatient0` query.
This query looks pretty harmless

```sql
SELECT * FROM "User"
WHERE "domain" = :domain ORDER BY "createdAt" LIMIT 1;
```

However, it was hanging, and at times took up over 11 seconds!
My suspicion is the `ORDER BY` clause. For example, if you login with gmail, it has to take every other gmail user & sort by the createdAt field. during a mass reconnect, this can get pretty taxing.
Instead of optimizing the query, we can just remove it. `isPatient0` is determined when a user is created. it's never updated after that! 

This PR adds the column, sets the value, makes it non-null, and then removes the complex logic in the code.
Now, when a user gets created, it simply queries if there is any user with the same domain that exists. no ORDER BY needed.
It no longer runs this query on socket connection, so that's faster, too.
 
## Testing scenarios

[ ] run the migration & then look at the User table. there's only 1 `true` isPatient0 per domain
- [ ] create a new user with an existing domain. isPatient0 is false for them
- [ ] create a new user with a new domain. isPatient0 is true for them


packages/server/postgres/queries/src/getPatientZeroByDomainQuery.sql
```diff
@@ -2,4 +2,4 @@
   @name getPatientZeroByDomainQuery
```
tianrunhe, line 2: +1 rename this query and the corresponding .ts code, as it no longer returns the patient 0

tianrunhe (approved):

tghanken (approved):

PR #7335, opened 2022-10-24
fix(docker): do not expose ports to the local network

igorlesnenko:
How to test:

- Checkout to master
- Run yarn dev
- Figure out your laptop IP in the local network
- Using this IP and device connected to the same network access http://YOUR_IP:8080/, see that it is accessible
- Checkout to this branch (fix-docker-ports)
- **Warning! This might wipe local database data, I was not able to find alternative way to do this**: Run `docker-compose down` `docker-compose up -d`
- Run `docker ps` see that containers now have `127.0.0.1` address specified in the PORTS column
- Using this IP and device connected to the same network access http://YOUR_IP:8080/, see that it is not accessible anymore
- Check other ports too


docker-compose.yml
```diff
@@ -38,7 +38,7 @@ services:
       target: prod
     env_file: .env
     ports:
-      - '3000:3000'
+      - '127.0.0.1:3000:3000'
```
igorlesnenko, line 41: I still have port 3000 accessible from the local network. Why could it be?

adaniels-parabol:
I'd suggest another network for internal data and set the value of the network `internal: true` then apply that network to the components for backend services. This way the web app can still be exposed if needed.

adaniels-parabol:
There are so many edge cases though:
- network traffic from app to db, redis, postgres
- network traffic from app and postgres, but not db and redis
- network traffic to everything for troubleshooting
My recommendation would be to let the local firewall handle the blocking of external access.

igorlesnenko:
Thanks @adaniels-parabol! Closing this PR for now then

Issue #7334, opened 2022-10-24
Add isFree to MeetingTemplate


nickoferrall:
We want to make 7 public templates accessible to all users, while the remaining templates are available to only pro & enterprise users.

To do this, we should add an `isFree` field to the `MeetingTemplate` object. If `isFree` is not true, we know it's a premium template.

### AC

- [ ] Implement a migration that adds an `isFree` boolean to the `MeetingTemplate` object
- [ ] The [following 7 templates](https://www.notion.so/parabol/Template-limits-on-starter-plan-limited-number-and-or-limit-sharing-customization-e5231a1dcfa5450a99b2dc5c09480a0a#09585fd0b0c54ba08d5820adceb8741c) should have `isFree: true`

tianrunhe:
Please remember to also update the [metrics](https://github.com/ParabolInc/parabol/blob/master/packages/server/utils/analytics/helpers.ts#L25)

nickoferrall:
@tianrunhe I'm thinking we'll add `meetingTemplateIsPremium: template?.isPremium` to that object. How does that sound?

PR #7345, opened 2022-10-25
feat: add isFree to meeting template

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7334

### To Test

- [ ] Run the migration and see that `isFree` is true in `MeetingTemplate` for the 7 templates included [here](https://www.notion.so/parabol/Template-limits-on-starter-plan-limited-number-and-or-limit-sharing-customization-e5231a1dcfa5450a99b2dc5c09480a0a#09585fd0b0c54ba08d5820adceb8741c) 🔒

### DIFF
--- a/packages/server/database/migrations/20221025161717-addIsFreeMeetingTemplate.ts
+++ b/packages/server/database/migrations/20221025161717-addIsFreeMeetingTemplate.ts
@@ -1,6 +1,6 @@
 import {R} from 'rethinkdb-ts'
 
-const premiumTemplateIds = [
+const freeTemplateIds = [
   'startStopContinueTemplate',
   'gladSadMadTemplate',
   'whatWentWellTemplate',
@@ -11,16 +11,8 @@ const premiumTemplateIds = [
 ]
 
 export const up = async function (r: R) {
-  await r
-    .table('MeetingTemplate')
-    .getAll(r.args(premiumTemplateIds))
-    .update({isPremium: true})
-    .run()
+  await r.table('MeetingTemplate').getAll(r.args(freeTemplateIds)).update({isFree: true}).run()
 }
 export const down = async function (r: R) {
-  await r
-    .table('MeetingTemplate')
-    .getAll(r.args(premiumTemplateIds))
-    .update({isPremium: false})
-    .run()
+  await r.table('MeetingTemplate').getAll(r.args(freeTemplateIds)).update({isFree: false}).run()
 }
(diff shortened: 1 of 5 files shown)


nickoferrall:
Requesting a Reviewer Review from @igorlesnenko and tagging in @tianrunhe to see if this is what you had in mind with the metrics 🙏

packages/server/database/migrations/20221025161717-addIsPremiumMeetingTemplate.ts
```diff
@@ -0,0 +1,26 @@
...
+  'sailboatTemplate',
+  'original4Template',
+  'fourLsTemplate'
+]
+
+export const up = async function (r: R) {
+  await r
+    .table('MeetingTemplate')
+    .getAll(r.args(premiumTemplateIds))
+    .update({isPremium: true})
```
tianrunhe, line 17: I thought these 7 templates should be FREE? i.e., not premium? Did I interpret the requirement wrong?

tianrunhe (changes requested):

packages/server/database/migrations/20221025161717-addIsPremiumMeetingTemplate.ts
nickoferrall: You're completely right! 🤦‍♂️ I've changed it to `isFree` instead of `isPremium` as I think it makes the logic easier to read

packages/server/database/migrations/20221025161717-addIsFreeMeetingTemplate.ts
```diff
@@ -0,0 +1,18 @@
...
+  'sailboatTemplate',
+  'original4Template',
+  'fourLsTemplate'
+]
+
+export const up = async function (r: R) {
+  await r.table('MeetingTemplate').getAll(r.args(freeTemplateIds)).update({isFree: true}).run()
+}
+export const down = async function (r: R) {
+  await r.table('MeetingTemplate').getAll(r.args(freeTemplateIds)).update({isFree: false}).run()
```
tianrunhe, line 17: +1 feels like `down()` should clear out the `isFree` properties vs. setting them to false

packages/server/utils/analytics/helpers.ts
```diff
@@ -23,6 +23,7 @@ export const createMeetingProperties = (
     meetingTemplateName: template?.name,
     meetingTemplateScope: template?.scope,
     meetingTemplateIsFromParabol: template?.isStarter,
+    meetingTemplateIsFree: template?.isFree,
```
tianrunhe, line 26: -1 Would prefer a default `false` over `undefined` to make analysis in Amplitude easier to understand, e.g., [Aviva recently was confused](https://parabol.slack.com/archives/C03C3J42WDU/p1666296869022669). While we are here, can we also fix the line above:

```suggestion
    meetingTemplateIsFromParabol: template?.isStarter ?? false,
    meetingTemplateIsFree: template?.isFree ?? false,
```

tianrunhe (changes requested):

tianrunhe (approved):

Issue #7341, opened 2022-10-25
Implement setIsFreeMeetingTemplate private mutation


nickoferrall:
Predecessor: https://github.com/ParabolInc/parabol/issues/7334

Create a private mutation so that Growth team members can update the `isFree` field on templates. This will make templates accessible to all users or restrict them to paying users.

### AC

- [ ] Create a private mutation that updates `isFree` on `MeetingTemplate`
- [ ] The mutation could take the `templateId` as an argument and/or the template `name`

Dschoordsch:
I think we shouldn't normalize any names and instead only check if these match exactly.
Instead of `togglePremiumTemplate` I would prefer `setPremiumTemplate` with an additional `isPremium: Boolean!` so the result doesn't change if the mutation is run multiple times.

PR #7346, opened 2022-10-26
feat: implement setIsFreeMeetingTemplate mutation

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7341

### To test

- [ ] Run the mutation with `templateIds` as the arg and see that the templates `isFree` field is updated
- [ ] Run the mutation with `templateNames` as the arg and see that the templates `isFree` field is updated


packages/server/graphql/private/mutations/setIsFreeMeetingTemplate.ts
```diff
@@ -0,0 +1,43 @@
...
+    const updatedTemplateIds = await r
+      .table('MeetingTemplate')
+      .filter((row) => r.expr(templateNames).contains(row('name')))
+      .update(
+        {isFree},
+        {returnChanges: true}
+      )('changes')('new_val')('id')
+      .default(null)
+      .run()
+    return {templateIds: updatedTemplateIds}
```
igorlesnenko, line 39: If no rows updated, for example they are already `isFree: true`, no ids returned and no result returned in mutation even if templates exists:

image

Is it expected behaviour?

igorlesnenko (approved):
If no update was made, no templates re

packages/server/graphql/private/mutations/setIsFreeMeetingTemplate.ts
nickoferrall: Yep! I wanted to let the user know if a template has been updated, so I only wanted to return the updated templates

codegen.json
```diff
@@ -20,6 +20,7 @@
           "PingableServices": "./types/PingableServices#PingableServicesSource",
           "ProcessRecurrenceSuccess": "./types/ProcessRecurrenceSuccess#ProcessRecurrenceSuccessSource",
           "RemoveAuthIdentitySuccess": "./types/RemoveAuthIdentitySuccess#RemoveAuthIdentitySuccessSource",
+          "SetIsFreeMeetingTemplate": "./types/SetIsFreeMeetingTemplate#SetIsFreeMeetingTemplateSource",
```
Dschoordsch, line 23: -1 This should be `"SetIsFreeMeetingTemplatePayload"` to match the GraphQL type.

Dschoordsch (changes requested):
Did not test

packages/server/graphql/private/types/SetIsFreeMeetingTemplateSuccess.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+const SetIsFreeMeetingTemplateSuccess: SetIsFreeMeetingTemplateSuccessResolvers = {
+  updatedTemplates: async ({updatedTemplateIds}) => {
+    if (!updatedTemplateIds) return null
+    const r = await getRethink()
+    const updatedTemplates = (await r
+      .table('MeetingTemplate')
+      .getAll(r.args(updatedTemplateIds))
+      .run()) as ResolversTypes['MeetingTemplate'][]
```
nickoferrall, line 15: I'm not sure why the casting is required here as `updatedTemplates` is defined as `[MeetingTemplate!]` in gql and the returned value from this rethinkdb query is `MeetingTemplate[]`

Dschoordsch: -1 This is the TS error I get
```
 Type 'MeetingTemplate' is not assignable to type 'Omit<ReflectTemplate, "team" | "prompts">'.
        Types of property 'isFree' are incompatible.
          Type 'boolean | undefined' is not assignable to type 'boolean'.
            Type 'undefined' is not assignable to type 'boolean'. (tsserver 2322)
```
so it might be worth fixing the type of `isFree` now before we have to add more such casts.

packages/server/graphql/private/mutations/setIsFreeMeetingTemplate.ts
```diff
@@ -8,10 +8,10 @@ const setIsFreeMeetingTemplate: MutationResolvers['setIsFreeMeetingTemplate'] =
   const r = await getRethink()
   // VALIDATION
   if (!templateIds?.length && !templateNames?.length) {
-    return {error: {message: 'Must provide template ids or names'}}
+    throw new Error('Must provide template ids or names')
```
Dschoordsch, line 11: -1 This will result in the query failing and returning a GraphQL error, so instead of
```
data: {
  error: {
    title: ...
    message: ...
  }
}
```
we will get
```
data: undefined
error: [some backtrace]
```
which is not what we want.

Dschoordsch (changes requested):

packages/server/graphql/private/types/SetIsFreeMeetingTemplateSuccess.ts
nickoferrall: Hey @mattkrick, this is the TS error from yesterday's office hours. After updating the `isFree` field, I get that ugly error message starting with:

```
(property) updatedTemplates?: Resolver<Maybe<(ResolverTypeWrapper<Omit<PokerTemplate, "team" | "dimension" | "dimensions"> & {
dimension: ResolverTypeWrapper<Omit<TemplateDimension, "team" | "template" | "selectedScale"> & {
selectedScale: ResolverTypeWrapper<Omit<TemplateScale, "team" | "dimensions"> & {
```

You mentioned that I should send over a link and you might spend a few minutes digging into it? No worries if not!

mattkrick: sure, i'll take a look now

mattkrick: here's the fix, let me know if that makes sense!
https://www.loom.com/share/4aea62d9a3ce4c918766ec824ad10bfa

nickoferrall: Fantastic, thanks for the explanation on mappers and why it's better to return an empty array vs null in the success case! It's always helpful to see how you got to the solution

packages/server/graphql/private/types/SetIsFreeMeetingTemplateSuccess.ts
```diff
@@ -0,0 +1,19 @@
...
+}
+
+const SetIsFreeMeetingTemplateSuccess: SetIsFreeMeetingTemplateSuccessResolvers = {
+  updatedTemplates: async ({updatedTemplateIds}) => {
+    const r = await getRethink()
+    const updatedTemplates = await r
+      .table('MeetingTemplate')
+      .getAll(r.args(updatedTemplateIds))
+      .run()
+    return updatedTemplates
```
Dschoordsch, line 15: +1 No need to change as this is not called often, but you could have just returned the `updatedTemplates` from `setIsFreeMeetingTemplate` directly instead of just the id and then reading it again here.

Dschoordsch (approved):
Tested, works.

packages/server/graphql/private/types/SetIsFreeMeetingTemplateSuccess.ts
nickoferrall: I went with this approach as I was following a pattern that's used in some other mutations, e.g. [editTask](https://github.com/ParabolInc/parabol/blob/9177909d9b3bc9cf2ba5d88a32c0ce53b1da177f/packages/server/graphql/mutations/editTask.ts) has the `task` but returns the `taskId` so [EditTaskPayload](https://github.com/ParabolInc/parabol/blob/9177909d9b3bc9cf2ba5d88a32c0ce53b1da177f/packages/server/graphql/types/EditTaskPayload.ts) queries the task, or [flagConversionModel](https://github.com/ParabolInc/parabol/blob/fa767a6f719cec03a966d41cc45d1a077a1857fb/packages/server/graphql/private/mutations/flagConversionModal.ts) has the `organization` but returns the `orgId` so [FlagConversionModalPayload](https://github.com/ParabolInc/parabol/blob/ae11377bf1f529f954cbbe80064775b2e6182056/packages/server/graphql/private/types/FlagConversionModalPayload.ts) grabs the `organization`.  

I guess the difference is in `SetIsFreeMeetingTemplateSuccess`, we're querying directly to rethinkdb rather than using a dataLoader, so the other payloads will use the cache whereas this one won't.

Am I right in saying that the advantage of returning ids from the mutation rather than the object is because it adds an extra sanity check?

Dschoordsch: Usually you should return ids to avoid overfetching, i.e. if the mutation did not fetch all the data that is needed for the success payload resolver, then it shouldn't fetch it just to return it. The mutation doesn't know[^1] what the client requested and should not fetch data which might not be needed.
`editTask` has a different use case with subscriptions, we don't want to serialize the data to send it over redis if it's not needed in the end. The return value could be different, but it's better to be the same for consistency, otherwise the success payload resolver would need to check for different inputs.
Why `flagConversionModal` does it this way, I don't know. I assume it is just to be aligned with the others.

[^1]: There is a 4th argument of type `GraphQLResolveInfo` to the resolver which contains this information, but it shouldn't really be used for this.

nickoferrall: Thanks! The overfetching and subscription use cases both make sense to me. 

For the [flagConversionModal](https://github.com/ParabolInc/parabol/blob/fa767a6f719cec03a966d41cc45d1a077a1857fb/packages/server/graphql/private/mutations/flagConversionModal.ts), @mattkrick it'd be great to know why you return the `id` rather than the `organization` object here?

mattkrick: In a perfect world, we'd use dataloaders for reads & the writes wouldn't have to return anything (e.g. we wouldn't have to `RETURN *` in pg or `{returnChanges: true}` in rethinkdb.

in that scenario, you're still only issuing 1 write & 1 read, albeit the read is non-atomic & requires a separate round trip to the DB.

my reason for returning just an ID is to send as little data as possible through the data transform pipeline. we don't know which fields, if any, the user wants from the organization. so why send the whole org? it's not a huge issue either way, It's just simpler to reason about-- the SourceStream has the foregin keys, the ResopnseStream transforms that SourceStream into everything the client wants.

PR #7350, opened 2022-10-26
fix: Contents of some menu dropdowns cannot be accessed with keyboard

gitstart-parabol:
# Description

Update Menu elements to be fully operated with a keyboard, and retains as expected

## Demo

https://user-images.githubusercontent.com/45232708/197995194-22147b9e-cfae-49e9-ac37-6648b53281c5.mov

https://user-images.githubusercontent.com/45232708/199522380-91918b92-8bc4-4836-a6b3-4028441b0f84.mov

## Testing scenarios

- [ ] Scenario A: Testing menu items which are links
  - Step 1: Using the keyboard, navigate to the user avatar on the navbar
  - Step 2: Toggle open the menu
  - Step 3: Verify that you are able to fire the link actions by clicking `Enter` on any active menu item
  - Step 4: Verify that you are not able to `Tab` away while the menu is opened

- [ ] Scenario B: Testing menu items that are not links
  - Step 1: Using the keyboard, navigate to a meeting card's menu trigger (3 dot icon)
  - Step 2: Toggle open the menu
  - Step 3: Verify that you are able to fire the actions by clicking `Enter` on any active menu item
  - Step 4: Verify that you are not able to `Tab` away while the menu is opened

## Final checklist


gitstart-instrumental:
Hi @Dschoordsch,

With regards to the sidebar items I believe the convention for navigating through them is with `Tab` (Forward), and `SHIFT + TAB` (Backward), which seem to work as expected at the moment.

Regarding some dropdowns which cannot be navigated with the arrow keys, there's an edge case that I believe would need to be clarified on the expected behavior.
The edge case here is that the dropdowns which are not working at the moment have a prop that ensures that the focus is kept on the parent and not on the dropdown menu (if the focus is not on the menu, we cannot navigate it).

Screenshot 2022-10-26 at 3 10 15 PM

There are many of these use cases on the codebase, so I would like to ask about the expected behavior on the menu usage with prop `keepParentFocus`.

BTW this is the behavior when we removed that prop.

https://user-images.githubusercontent.com/45232708/198050824-85bd0593-3050-4b59-9bbd-fbbdb98a8f37.mov

gitstart-parabol:
Hi @Dschoordsch,

With regards to the sidebar items I believe the convention for navigating through them is with `Tab` (Forward), and `SHIFT + TAB` (Backward), which seem to work as expected at the moment.

Regarding some dropdowns which cannot be navigated with the arrow keys, there's an edge case that I believe would need to be clarified on the expected behavior.
The edge case here is that the dropdowns which are not working at the moment have a prop that ensures that the focus is kept on the parent and not on the dropdown menu (if the focus is not on the menu, we cannot navigate it).

Screenshot 2022-10-26 at 3 10 15 PM

There are many of these use cases on the codebase, so I would like to ask about the expected behavior on the menu usage with prop `keepParentFocus`.

BTW this is the behavior when we removed that prop.

https://user-images.githubusercontent.com/45232708/198050824-85bd0593-3050-4b59-9bbd-fbbdb98a8f37.mov

Dschoordsch:
That is a very good question!
From what I can tell from the sparse git history this was initially introduced to work around something in the EmojiMenu. Since then the use is more widespread to workaround an issue we get with input boxes in menus, here https://github.com/ParabolInc/parabol/pull/5945#issuecomment-1024510319 is a nice video of what happens without `keepParentFocus`: As soon as the input blurs it steals the focus from the menu again and prevents the action this way.
This happens because
* the input tries to steal focus here https://github.com/ParabolInc/parabol/blob/7c38a8846152a352ad12b06adfc1950c80d6e776/packages/client/components/MenuSearch.tsx#L35
* but we prevent it when `keepParentFocus` here https://github.com/ParabolInc/parabol/blob/7c38a8846152a352ad12b06adfc1950c80d6e776/packages/client/components/Menu.tsx#L83

So in my opinion this stealing focus behaviour of `MenuSearch` component should be removed. This requires on the other hand to remove the focus stealing behaviour of `Menu` here https://github.com/ParabolInc/parabol/blob/7c38a8846152a352ad12b06adfc1950c80d6e776/packages/client/components/Menu.tsx#L71
With most of these removed, I'd think that most cases should work by default, but need to be double checked, especially the EmojiMenu.

@mattkrick Do you remember why refocusing `MenuSearch` on blur was necessary?

mattkrick:
IIRC all focus-related things with the menu are because some menus have a search bar at the top of them & as you search you want the rows to filter but still be able to keep typing.

last week i did a little deep dive on new tech frameworks that the FOSS community is using.
One that I thought was really interesting was radix UI, which is a bunch of headless UI components that are aria compliant. Being headless, they offer no opinions on styles, so we're free to use emotion with them, they just attached all the aria & a11y features for us.
I'd LOVE to see a PR that tried incorporating one of those components. e.g. https://www.radix-ui.com/docs/primitives/components/dropdown-menu

there are a few others that i found (healess UI by the tailwindcss folks & react-aria by adobe) but IMO radix looked like the most mature

gitstart-parabol:
> > IIRC all focus-related things with the menu are because some menus have a search bar at the top of them & as you search you want the rows to filter but still be able to keep typing.
> 
> Ok, this appeared to work with my proposed changes, @gitstart-parabol please double check it continues to work after the change.

It works after the proposed changes

gitstart-parabol:
Screenshot 2022-11-02 at 8 47 24 AM

This behavior remains beneficial, to preserve the autofocus behavior of `MenuSearchItem`.

Dschoordsch:
@BartoszJarocki James is out, can you do the review please?

BartoszJarocki:
@jmtaber129 may I ask you to review this one?

gitstart-parabol:
Any update on this? @BartoszJarocki @jmtaber129

packages/client/components/MenuItem.tsx
```diff
@@ -27,19 +28,21 @@ interface Props {
...
-    '&:hover,:focus': {
-      backgroundColor: isActive ? PALETTE.SLATE_200 : PALETTE.SLATE_100,
-      outline: 0
-    }
-  })
-)
+const MenuItemStyles = styled('div')<{
+  isActive: boolean
+  isDisabled: boolean | undefined
+  to?: string
```
jmtaber129, line 34: +1 Doesn't seem to be used here

packages/client/components/Menu.tsx
```diff
@@ -167,9 +173,13 @@ const Menu = forwardRef((props: Props, ref: any) => {
         if (activeIdx !== null) {
           const itemHandle = itemHandles.current[activeIdx]
           itemHandle?.onClick?.(e)
+          if (itemHandle?.to) {
+            history.push(itemHandle?.to)
+          }
```
jmtaber129, line 178: Can you please elaborate on why this is needed, and why `to` is added to types elsewhere? Shouldn't `onClick` be enough?

packages/client/components/MenuItem.tsx
```diff
@@ -112,6 +118,7 @@ const MenuItem = forwardRef((props: Props, ref: any) => {
       role='menuitem'
       ref={itemRef}
       isActive={isActive}
+      to={to}
```
jmtaber129, line 121: +1 Doesn't seem to be used here.

jmtaber129 (commented):
Can you elaborate on why `to` needs to be handled explicitly? Wouldn't the `onClick` handler handle `to` links implicitly?

gitstart:
Hi @jmtaber129 thanks for digging into this.
We needed to explicitly handle `to`, to support a use cases of `MenuItem`, where the `onClick` event is not binded to the MenuItem, in favour of using a `Link` component as MenuItem's `Label` property, see sample below.

```shell
// StandardHubUserMenu.tsx

<MenuItem
  label={
    <MenuItemLink to={'/me/organizations'}>
      <MenuItemIcon>
        <AccountBalance />
      </MenuItemIcon>
      {'Organizations'}
    </MenuItemLink>
  }
/>
```

This use case works fine when we use the mouse to trigger click, although it does not suffice for keyboard click since specifically look out for `onClick` property binded to the MenuItem.

You are right though about this use case not being very obvious and can certainly use some improvement. 
We could add a new component `MenuLink` which uses `to` property instead of `onClick`. This might reduce the ambiguity with handling both `to` and `onClick` on MenuItems. WDYT?

packages/client/components/MenuItem.tsx
gitstart: This was added to provide type support, that allows passing `to`, to the `MenuItemStyles` element for use in `Menu` componet.

packages/client/components/MenuItem.tsx
gitstart: Added some context below.

gitstart-parabol: Added some context below. This helps to support the `to` property on the `MenuItemLink` label in the sample use case below.

packages/client/components/Menu.tsx
gitstart-parabol: Added some context below.

packages/client/components/MenuItem.tsx
gitstart-parabol: This was added to provide type support, that allows passing `to`, to the `MenuItemStyles` element for use in `Menu` component.

gitstart-parabol:
> Hi @jmtaber129 thanks for digging into this. We needed to explicitly handle `to`, to support a use cases of `MenuItem`, where the `onClick` event is not binded to the MenuItem, in favour of using a `Link` component as MenuItem's `Label` property, see sample below.
> 
> ```shell
> // StandardHubUserMenu.tsx
> 
> <MenuItem
>   label={
>     <MenuItemLink to={'/me/organizations'}>
>       <MenuItemIcon>
>         <AccountBalance />
>       </MenuItemIcon>
>       {'Organizations'}
>     </MenuItemLink>
>   }
> />
> ```
> 
> This use case works fine when we use the mouse to trigger click, although it does not suffice for keyboard click since specifically look out for `onClick` property binded to the MenuItem.
> 
> You are right though about this use case not being very obvious and can certainly use some improvement. Perhaps we could add a new component `MenuLink` which uses `to` property instead of `onClick`. This might reduce the ambiguity with handling both `to` and `onClick` on MenuItems. What do you think?

@jmtaber129 what do you think about this proposition?

Dschoordsch:
@jmtaber129 are you going to continue the review?

jmtaber129:
Sorry, must have missed this. Re-self-requesting review

packages/client/components/MenuItem.tsx
jmtaber129: Looked at this locally - we shouldn't need this here, since `to` is exposed to the parent via `useImperativeRef` and isn't used as part of `MenuItemStyles`.

packages/client/components/MenuItem.tsx
jmtaber129: Looked at this locally - we shouldn't need this here, since `to` is exposed to the parent via `useImperativeRef` and isn't used as part of `MenuItemStyles`.

jmtaber129 (approved):
Okay, I think I understand the approach. It feels a bit smelly to be breaking the `label` abstraction by handling the `to` property as part of `MenuItem` and `Menu`. I'm not sure the right way to handle this without a larger refactor, so I'll defer to maintainers.

Regardless, the `to` prop in `MenuItemStyles` isn't necessary - we only need to expose `to` to `MenuItem`'s parent via `useImperativeRef`.

Dschoordsch (changes requested):
I tested without the `to` logic in `Menu` and it seems to work just fine with the keyboard. Can you elaborate why this is needed exactly?

gitstart-parabol:
> I tested without the `to` logic in `Menu` and it seems to work just fine with the keyboard. Can you elaborate why this is needed exactly?

Hi @Dschoordsch, the menu functions well in most of the use cases without the newly implemented `to` logic. Excluding those that involve the use of `MenuItemLink` due to their lack of an onClick handler. This is particularly evident in the `StandardHubUserMenu` menu, as demonstrated in the attached demo video. Kindly take a look at the video to observe the behavior with and without the implemented logic.

https://user-images.githubusercontent.com/45232708/219369633-95238430-09cf-4035-871e-487c27e26d06.mov

Dschoordsch:
@gitstart-parabol please always re-request review if you want a response, otherwise it's easy to miss.

gitstart-parabol:
@Dschoordsch this PR is ready for review

mattkrick:

PR #7351, opened 2022-10-26
fix: fix stripe quantity mismatch error

### LINKED ISSUE #5325: Error: Stripe Quantity Mismatch
https://sentry.io/organizations/parabol/issues/2436725787/?referrer=github_plugin

Not fatal, but we're getting it more and more since 6.14.0. Affected users so far: 246, happened 273 times

```
Error: Stripe Quantity Mismatch
  File "./packages/server/billing/helpers/processInvoiceItemHook.ts", line 116, in c
    sendToSentry(new Error('Stripe Quantity Mismatch'), {
  ?, in runMicrotasks
```

igorlesnenko:
Partially solves: #5325

See detailed explanation here: https://github.com/ParabolInc/parabol/issues/5325#issuecomment-1292336866

**How to test:**

- Create organization and upgrade to PRO
- Put some sleep `await new Promise(r => setTimeout(r, 10000));` in the code [here](https://github.com/ParabolInc/parabol/blob/master/packages/server/billing/helpers/processInvoiceItemHook.ts#L101)
- Create and login with 2 users in 2 different browsers
- Copy and paste the invite link for those users at the same time
- See no error quantity mismatch error
- Remove sleep, smoke test that stripe invoices are working fine when adding/removing users

### DIFF
--- a/packages/server/billing/helpers/processInvoiceItemHook.ts
+++ b/packages/server/billing/helpers/processInvoiceItemHook.ts
@@ -69,6 +69,16 @@ const processInvoiceItemHook = async (stripeSubscriptionId: string) => {
     return
   }
   const {id: hookId, type, prorationDate, isProrated, orgId, userId} = hook
+
+  const orgUsers = await r
+    .table('OrganizationUser')
+    .getAll(orgId, {index: 'orgId'})
+    .filter({
+      inactive: false,
+      removedAt: null
+    })
+    .run()
+
   const tentativeProrationDate = isProrated
     ? prorationDate ?? toEpochSeconds(new Date())
     : undefined
@@ -108,14 +118,6 @@ const processInvoiceItemHook = async (stripeSubscriptionId: string) => {
     .run()
 
   if (isFlushed) {
-    const orgUsers = await r
-      .table('OrganizationUser')
-      .getAll(orgId, {index: 'orgId'})
-      .filter({
-        inactive: false,
-        removedAt: null
-      })
-      .run()
     const orgUserCount = orgUsers.length
     if (orgUserCount !== nextQuantity) {
       insertStripeQuantityMismatchLogging(
--- a/packages/server/billing/helpers/processInvoiceItemHook.ts
+++ b/packages/server/billing/helpers/processInvoiceItemHook.ts
@@ -70,23 +70,22 @@ const processInvoiceItemHook = async (stripeSubscriptionId: string) => {
   }
   const {id: hookId, type, prorationDate, isProrated, orgId, userId} = hook
 
-  const orgUsers = await r
-    .table('OrganizationUser')
-    .getAll(orgId, {index: 'orgId'})
-    .filter({
-      inactive: false,
-      removedAt: null
-    })
-    .run()
-
   const tentativeProrationDate = isProrated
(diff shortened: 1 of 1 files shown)


packages/server/billing/helpers/processInvoiceItemHook.ts
```diff
@@ -69,6 +69,16 @@ const processInvoiceItemHook = async (stripeSubscriptionId: string) => {
     return
   }
   const {id: hookId, type, prorationDate, isProrated, orgId, userId} = hook
+
+  const orgUsers = await r
```
nickoferrall, line 73: I haven't been able to repro the mismatch error yet after following the testing instructions. Just wondering if you're able to consistently reproduce it?

I think the bug could well be the result of a race condition where multiple users accept the org invite at the same time, but we do have a Redis Lock on line 45. Wouldn't that prevent this function from being called at the exact same time by two different users?

If we do want to move the `orgUsers` query, would we want to move it to the same `Promise.all` where we're getting the `stripeSubscriptionItem`?

nickoferrall (changes requested):

packages/server/billing/helpers/processInvoiceItemHook.ts
igorlesnenko: > I haven't been able to repro the mismatch error yet after following the testing instructions. Just wondering if you're able to consistently reproduce it?

Yeah, I always able to reproduce it. I recorded a loom: https://www.loom.com/share/a4b0a054743f441781c31c9a5e950074

Did you follow my testing instruction in `master` branch?

> Wouldn't that prevent this function from being called at the exact same time by two different users?

@nickoferrall we only set lock for `3000` ms . The issue occurs if the function executes longer that time.

> If we do want to move the orgUsers query, would we want to move it to the same Promise.all where we're getting the stripeSubscriptionItem?

Make sense, moved to Promise.all

nickoferrall (approved):
Thanks for the Loom! I've just tested it and it's working well for me 🙌

Dschoordsch (commented):
I did not test this yet, it looks good.
However in addition I would prefer if we would slightly refactor this hook:
* just do locking/unlocking and increase the TTL to maybe 30s
* move the rest of the logic into a different function

So instead of having
https://github.com/ParabolInc/parabol/blob/master/packages/server/billing/helpers/processInvoiceItemHook.ts#L44-L52
we would replace it with
```ts
const processInvoiceItemHook = async (stripeSubscriptionId: string) => {
  const redisLock = new RedisLock(stripeSubscriptionId, 30000)
  await redisLock.lock(Infinity)
  try {
    await lockedProcessInvoiceItemHook(stripeSubscriptionId)
  } finally {
    await redisLock.unlock()
  }
}
```
This should solve the current problem and also avoid some `return` cases without unlock so we can safely increase the lock time.

Dschoordsch:
For extra style points we could add a higher order function to RedisLock to encourage this use case

Dschoordsch (changes requested):
Tested current change and it works, I still would like to have the lock last the whole function duration.

igorlesnenko:
Closing the PR, because it probably will not help. See a comment in the issue https://github.com/ParabolInc/parabol/issues/5325#issuecomment-1311604650

PR #7352, opened 2022-10-26
feat: make prev used premium templates available on free tier

### LINKED ISSUE #7336: Add accessiblePremiumTemplateIds to RetrospectiveMeetingSettings
We're limiting some public templates to Pro & Enterprise teams, but we don't want to take away templates that free teams have already used.

It would be a bad experience for a team to discover that they need to upgrade to access a template that they've been using for a while.

I suggest that we implement a migration that finds all the `MeetingTemplate`s that free teams have previously used (i.e. if the `lastUsedAt` is non-null) and add those ids to a new `accessiblePremiumTemplateIds` field that exists in `RetrospectiveMeetingSettings`.

The `isPremium` resolver in `MeetingTemplate` can then check whether the template id exists in `accessiblePremiumTemplateIds`. If it does, `isPremium` is false, and the user can continue to access the template.

### AC

- [ ] Add `accessiblePremiumTemplateIds` to the `RetrospectiveMeetingSettings` db & gql object
- [ ] Implement a migration that adds the template ids that free teams have previously used to the `accessiblePremiumTemplateIds` field

nickoferrall:
Fix #7336 

### To test

- [ ] If a team has previously used a premium (i.e. `MeetingTemplate.isFree` is false / undefined) template within the past 90 days, `MeetingTemplate.isFree` is true


packages/server/graphql/types/MeetingTemplate.ts
```diff
@@ -26,7 +27,12 @@ export const meetingTemplateFields = () => ({
     type: new GraphQLNonNull(GraphQLBoolean),
     description:
       'True if template is available to all teams including non-paying teams, else false',
-    resolve: ({isFree}: {isFree?: boolean}) => !!isFree
+    resolve: async ({id: templateId, isFree}: {id: string; isFree?: boolean}) => {
+      const endTimes = await db.readMany('endTimesByTemplateId', [templateId])
```
igorlesnenko, line 31: - Looks like we are checking if the template was used by **any user** in parabol database and make it free. Should we only check if the viewer **organization** was using it? 

- Do we need to use dataLoader here? If we list templates, for every template we would do that query? If cache is missing we will do a [rethinkdb query](https://github.com/ParabolInc/parabol/blob/89228ca101fea2efcb38dcfadbf989f676d7e50e/packages/server/dataloader/customRedisQueries.ts#L13-L22).

igorlesnenko (changes requested):

nickoferrall:
Decided to keep things simple and give the same experience to all free users, regardless of whether they've used a premium template before. We can build this is any users request it after we've shipped template limits.

See Slack thread for further context: https://parabol.slack.com/archives/C042E8PGPD4/p1667388355109689

PR #7354, opened 2022-10-27
fix: task integration demo bug

nickoferrall:
@MindEvo discovered a bug in the demo where you couldn't push tasks to integrations (see Rafa's message in Slack [here](https://parabol.slack.com/archives/C836NA350/p1666857871119549?thread_ts=1666780209.440339&cid=C836NA350)). 

This was because https://github.com/ParabolInc/parabol/pull/7234 added `TaskFooterIntegrateMenuListLocalQuery` which wasn't added to the demo. 

The same PR also requested the `service` field from the repoIntegration, which wasn't being returned in the `DemoJiraRemoteProject` demo class. 

### To test

- [ ] Can open the task integration menu in the demo
- [ ] Can push a task to GitHub & Jira

### DIFF
--- a/packages/client/modules/demo/ClientGraphQLServer.ts
+++ b/packages/client/modules/demo/ClientGraphQLServer.ts
@@ -275,6 +275,18 @@ class ClientGraphQLServer extends (EventEmitter as GQLDemoEmitter) {
         }
       }
     },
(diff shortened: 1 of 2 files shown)


Dschoordsch (approved):
Works

PR #7356, opened 2022-10-27
feat(metrics): Enable Google Analytics client side page_view tracking

tianrunhe:
# Description

Fixes https://github.com/ParabolInc/parabol-data-sanctum/issues/450

## Demo

Screenshot of GA4 successfully tracked my `page_view` events and geo info (yes, I'm working by seaside now 😎)

Screenshot 2022-10-27 at 2 36 06 PM

Here's another GA4 screenshot showing it successfully detected my userId:

Screenshot 2022-10-27 at 2 36 33 PM

## Testing scenarios

- [ ] `page_view` events tracked
  - Add `GA_TRACKING_ID` to your `.env`. See [action-devops/environments/local](https://github.com/ParabolInc/action-devops/blob/master/environments/local#L72)
  - Start the server locally and go to any app page
  - Go to the Staging [GA4 realtime](https://analytics.google.com/analytics/web/#/p327845365/realtime/overview) dashboard and verify the `page_view` events come through

## Final checklist


packages/client/components/AnalyticsPage.tsx
```diff
@@ -80,6 +88,25 @@ if (datadogEnabled) {
...
+        gaOptions: {
+          userId: id,
+          clientId: segmentId
+        }
+      })
+    }
+    if (gaMeasurementId) {
+      initializeGA4().catch()
+    }
+  })
```
tianrunhe, line 108: @Dschoordsch I'm not sure if this is the right place to put it, looking for suggestion! 🙏

Dschoordsch: Looks like the correct place to me.

tianrunhe:
@Dschoordsch Can you review this one for me please? Did I assign the PR to you, instead of requesting you to review? My intention was to ask you for a review. Thanks!

packages/client/components/AnalyticsPage.tsx
```diff
@@ -16,6 +17,7 @@ const query = graphql`
   query AnalyticsPageQuery {
     viewer {
       id
+      segmentId
```
Dschoordsch, line 20: We also initialize segment on this page but never read the `segmentId` before, which strikes me as odd. Are you sure we're writing this before we query it?

Dschoordsch (changes requested):
Yes, you assigned me, so I wasn't sure if you just wanted an answer.
I'm not sure about the `segmentId` because I don't know where it comes from, could you check it also works for new signups etc.?
I wasn't able to see my data in the analytics board.

tianrunhe:
@Dschoordsch Sorry this PR was pending for a while due to my time-off and separate discussion happened within data team. 

I think the null `segmentId` error only happened in local where we don't initialize Segment. I removed the `__PRODUCTION__` check in AnalyticsPage and I think it should be fine (as in local testing won't pollute production environment) because:
1. DataDog is [still gated](https://github.com/ParabolInc/parabol/blob/master/packages/client/components/AnalyticsPage.tsx#L50-L51) by `__PRODUCTION__` check
2. Segment and GA have different keys for local & staging environments

That being said, I did see we have a lot of production user records with NULL `segmentId` so I make the field optional within `User` GraphQL type.
New account should have `segmentId` setup. Can you test again with the local stage [`SEGMENT_WRITE_KEY`](https://github.com/ParabolInc/action-devops/blob/master/environments/local#L59) and [`GA_TRACKING_ID`](https://github.com/ParabolInc/action-devops/blob/master/environments/local#L72)? Thank you!

packages/client/components/AnalyticsPage.tsx
```diff
@@ -80,9 +88,26 @@ if (datadogEnabled) {
...
+        gaOptions: {
+          userId: id,
+          clientId: segmentId
+        }
+      })
+    }
+    if (gaMeasurementId && !isGAInitialized) {
+      initializeGA().catch()
+      setIsGAInitialized(true)
+    }
```
Dschoordsch, line 109: -1 We never uninitialize GA on logout. Adding
```ts
    if (!atmosphere.viewerId) {
      ReactGA.set({userId: null, clientId: null})
      setIsGAInitialized(false)
    }
```
seems to do the trick

Dschoordsch (changes requested):

Dschoordsch (approved):
Looks good

PR #7361, opened 2022-10-31
fix: Send button is disabled in retro discussion on mobile

### LINKED ISSUE #7165: Send button is disabled in retro discussion on mobile
## Issue - Bug
Send button is disabled on mobile. However, we can send using the keyboard arrow/next-line button. I think it should be similar to the desktop version and the next-line button should work as next line button 

Apart from that clicking on the send button also closes the keyboard which is also annoying. 

#### Desktop
Screenshot 2022-09-09 224844

#### Mobile
https://user-images.githubusercontent.com/106697681/189408259-3ed76f93-08f0-47df-9ce0-306b0c3027e1.mp4

- **OS:**  Android
- **Browser:** Chrome 

- User can send the comment using the send button
- Clicking the send button should not close the keyboard
- Next-line button should change the line (instead of sending comment) as there is no other way on mobile to change the line.

gitstart-parabol:
# Description

Partially Fixes [issue-link](https://github.com/ParabolInc/parabol/issues/7165) Partially Fixes #7165
- [x] User can send the comment using the send button

## Demo
https://user-images.githubusercontent.com/34602861/199014360-62b3df67-8322-4a8b-91f8-db44523fce69.mov

## Testing scenarios

- [ ] Scenarios
  - Start the server locally and go to any app page from an  android device
  - Create a meeting and add a comment to the meeting
  - Then reply to that comment of the meeting

## Final checklist

- [x] I checked the [code review guidelines](https://github.com/ParabolInc/parabol/docs/codeReview.md)


gitstart-parabol:
Hi @BartoszJarocki, this PR is ready for review

packages/client/components/SendCommentButton.tsx
```diff
@@ -41,29 +42,52 @@ interface Props {
...
+        })
+      } else {
+        onSubmit()
+      }
+    }
+    btn?.addEventListener('click', submitListener, true)
+    return () => {
+      btn?.removeEventListener('click', submitListener, true)
+    }
+  }, [onSubmit])
```
BartoszJarocki, line 75: -2 there's no need for `useEffect` here. what's more, when on android you're adding `touchend` event listener which isn't cleared and will cause a memory leak. In that kind of situation, it's worth asking why is this button disabled, when the text is already there? 

The root cause seems to be the function, that checks if the editor has text. On android we use fallback editor, and this function doesn't take that into account. see the diff below:

Screenshot 2022-11-04 at 07 29 30

changing the way we get the current text, fixes the issue with only a few lines of code without a need for that kind of `useEffects`.

BartoszJarocki (changes requested):
Thank you for looking into this! I added a comment with a solution that fixes the root cause of this. Please take a look and adjust the changes. the fix is actually pretty straightforward.

packages/client/components/SendCommentButton.tsx
BartoszJarocki: there's a good read from react team about when you _don't_ need `useEffect`: https://beta.reactjs.org/learn/you-might-not-need-an-effect

gitstart-parabol: @BartoszJarocki Thank you for your feedback and suggestions; the other changes were made to account for the remaining acceptance criteria; please take another look.

gitstart-parabol:
> Thank you for looking into this! I added a comment with a solution that fixes the root cause of this. Please take a look and adjust the changes. the fix is actually pretty straightforward.

@BartoszJarocki Thank you for your feedback and suggestions; the other changes were made to account for the remaining acceptance criteria; please take another look.

packages/client/components/DiscussionThreadInput.tsx
```diff
@@ -159,8 +159,16 @@ const DiscussionThreadInput = forwardRef((props: Props, ref: any) => {
...
+    const editorEl = editorRef.current
+    if (isAndroid) {
+      if (!editorEl || editorEl.type !== 'textarea') return ''
+      return editorEl.value
+    }
 
-  const hasText = ensureHasText(editorState.getCurrentContent().getPlainText())
+    return editorState.getCurrentContent().getPlainText()
+  }
+  const hasText = ensureHasText(getCurrentText())
```
BartoszJarocki, line 171: -1 do you mind keeping only these changes in this particular PR? if something isn't working, it's easier to track the change which is causing the issue when PRs are related to only one functionality/issue.

BartoszJarocki (changes requested):

packages/client/components/DiscussionThreadInput.tsx
gitstart-instrumental: @BartoszJarocki These changes only account for the visibility of the button when there is no content in the input for Android devices; according to the issue description, we only pass the first check.

- [X] User can send the comment using the send button
- [ ] Clicking the send button should not close the keyboard
- [ ] Next-line button should change the line (instead of sending a comment) as there is no other way on mobile to change the line.

Does this mean the other acceptance criteria should be implemented in other PRs?

gitstart-parabol: @BartoszJarocki These changes only account for the visibility of the button when there is no content in the input for Android devices; according to the issue description, we only pass the first check.

- [X] User can send the comment using the send button
- [ ] Clicking the send button should not close the keyboard
- [ ] Next-line button should change the line (instead of sending a comment) as there is no other way on mobile to change the line.

Does this mean we should implement the other acceptance criteria in other PRs?

BartoszJarocki: Good question! There are no strict rules here, but definitely, there's no rule about 1 task = 1 PR. Our code review policy states, `Once you're assigned to a task, it's your responsibility to split it the way you can submit reasonably sized PRs. It’s ok to have multiple PRs for a single issue.`. 

The main reason I ask to split PR is that I don't really understand why for example, part `if (e.key !== 'Enter' || e.shiftKey) return` was removed down below. And while we can discuss that further, the main issue here (send button is disabled on mobile) would be already merged.

BartoszJarocki (approved):

PR #7372, opened 2022-11-03
chore: Fix isPatient0 migration runtime

Dschoordsch:
# Description

The previous migration was too slow to run on production data and timed out. Solving this by creating a temporary table to read the patient 0 values from.

Previous version timed out after > 30 min while current version runs in < 2min on staging.

## Demo

Screenshot from 2022-11-03 10-32-06

## Testing scenarios

- Run migration on staging data
  - [ ] smoke test "isPatient0" values look correct

## Final checklist

### DIFF
--- a/packages/server/postgres/migrations/1666387804873_userPatient0.ts
+++ b/packages/server/postgres/migrations/1666387804873_userPatient0.ts
@@ -9,7 +9,7 @@ export async function up() {
     ADD COLUMN IF NOT EXISTS "isPatient0" BOOLEAN DEFAULT FALSE;
 
     DROP TABLE IF EXISTS "tmp_patient0";
-    SELECT DISTINCT ON ("domain") "domain", "id" INTO "tmp_patient0" FROM "User" ORDER BY "domain", "createdAt";
+    SELECT DISTINCT ON ("domain") "domain", "id" INTO TEMP "tmp_patient0" FROM "User" ORDER BY "domain", "createdAt";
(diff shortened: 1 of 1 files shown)


packages/server/postgres/migrations/1666387804873_userPatient0.ts
```diff
@@ -5,14 +5,18 @@ export async function up() {
...
-  `)
-  await client.query(`
-  UPDATE "User" u
-  SET "isPatient0" = COALESCE("createdAt" = (SELECT "createdAt" FROM "User" WHERE "domain" = u.domain ORDER BY "createdAt" LIMIT 1), false);
-  `)
-  await client.query(`
+    ALTER TABLE "User"
+    ADD COLUMN IF NOT EXISTS "isPatient0" BOOLEAN DEFAULT FALSE;
+
+    DROP TABLE IF EXISTS "tmp_patient0";
```
BartoszJarocki, line 11: +! out of curiosity, does it make sense to use `TEMPORARY` table here? 
https://www.postgresql.org/docs/current/sql-createtable.html 

they're session specific and were created for this exact use case

BartoszJarocki (approved):
looks good!

packages/server/postgres/migrations/1666387804873_userPatient0.ts
Dschoordsch: Good suggestion! I will keep the `DROP` anyways, as the temp table isn't cleared up when running the query manually in pgadmin.

PR #7377, opened 2022-11-03
fix: update tier when leaving org

### LINKED ISSUE #7319: Tier is not updated when a user is removed from an org
## Issue - Bug
Simplest way to reproduce is to start fresh:
1. Create two accounts: abby@test.com and bob@test.com
2. Abby invites Bob to her team/org
3. At this time, both Abby and Bob's tier in the `User` table are `personal`, no problem there
4. Abby upgrades her org to `pro`
5. Now both Abby and Bob's tier in the `User` table changed to `pro`, which are expected
6. Abby removes Bob from her org
7. Bob's tier should be `personal` now, but it stays at `pro` in the `User` table

I suspect this could also happen when the user is automatically marked as `inactive`.

- [ ] User's table is correctly reflected in the `User` table when he/she is removed from an org

nickoferrall:
Fix #7319 

### Testing steps

- [ ] Reproduce the bug in `master` using the steps outlined in https://github.com/ParabolInc/parabol/issues/7319
- [ ] After following those 7 steps, check out this branch and run the migration. See that Bob's tier is now `personal` in the `User` table in pgadmin
- [ ] Follow the same 7 steps but now see that once Abby removes Bob from her org, Bob's tier is now personal


packages/server/database/migrations/20221103134908-adjustUserTiers.ts
```diff
@@ -0,0 +1,101 @@
...
+import getPg, {closePg} from '../../postgres/getPg'
+import IUser from '../../postgres/types/IUser'
+import segmentIo from '../../utils/segmentIo'
+import {TierEnum} from '../types/Invoice'
+import OrganizationUser from '../types/OrganizationUser'
+import User from '../types/User'
+
+export const up = async function (r: R) {
+  const removedOrgUserIds = await r
+    .table('OrganizationUser')
```
nickoferrall, line 11: I'm not sure why the typechecker is hanging 🤔 I get the following error: `Migration error Error: Cannot find module './generated/getUsersByIdsQuery'`, but I'm not importing `getUsersByIdsQuery`. I was importing it in the previous commit. I tried re-running it but no dice

tianrunhe: I encountered this issue recently too, fixed it by running `yarn pg:build` before `yarn dev`

tianrunhe:
Note: The Github action picks me as a reviewer not fully randomly, or because it thinks we are buddies 🤣. It's [one rule](https://github.com/ParabolInc/parabol/blob/master/.github/reviewers.yml#L35) Georg set that says any changes in DB migration should include either me or Taylor as a reviewer, because it could have downstream impact to our data warehouse, e.g., schema changes. So in the future please do not remove me as it will break my glass heart 😭

That being said I'll be on PTO for a while so I suggest you include @tghanken on this one 😃

tghanken:
>It's [one rule](https://github.com/ParabolInc/parabol/blob/master/.github/reviewers.yml#L35) Georg set that says any changes in DB migration should include either me or Taylor as a reviewer

For what it's worth, I think it makes sense for us to be added as an additional reviewer, not as the sole reviewer (as the action appears to be currently setup).  I'm happy to review migrations and PRs from a data lens, but think it makes sense for someone on the Product team to function as the "main" reviewer.

Dschoordsch:
@tghanken The data representative should not do the review on their own, but we don't add developers to review changes automatically for internals, because the PR author might want to choose a specific reviewer for various reasons. So your review is additional and you can focus just on the data collection/analytics impact of the change.

packages/server/database/migrations/20221103134908-adjustUserTiers.ts
nickoferrall: Good to know! But the error for me is happening in GitHub Actions, rather than locally, where the typecheck is working successfully. Running `yarn pg:build` before `yarn dev` doesn't update any files for me, so I don't think it'll fix the Actions typechecker issue. 

@Dschoordsch or @rafaelromcar-parabol do you have any ideas as to why the Actions typechecker would [continue endlessly and show an error when I terminate it](https://github.com/ParabolInc/parabol/actions/runs/3394832362/jobs/5645174770) but work locally? I've tried rerunning it once

nickoferrall:
> So in the future please do not remove me as it will break my glass heart 😭

😄 for sure! The reason I removed you was because the PR is in self-review as I want to fix the build issue before requesting a review. I'll be sure to tag a data representative once that's fixed

packages/server/database/migrations/20221103134908-adjustUserTiers.ts
```diff
@@ -0,0 +1,101 @@
+import {R, RDatum, RValue} from 'rethinkdb-ts'
+import getPg, {closePg} from '../../postgres/getPg'
+import IUser from '../../postgres/types/IUser'
+import segmentIo from '../../utils/segmentIo'
```
Dschoordsch, line 4: -1 seems this is the culprit. We need to be careful with migrations and ideally avoid referring to application logic types most of the time. The reason is that these are going to change while the migration should be a stable snapshot in time.
Your particular issue is that you're indirectly referencing `getUsersByIds` which is a generated SQL query. For generating the query, the generator needs to know the database schema which it queries directly from the database. The database is only ready after all the migrations were run, thus migrations cannot refer to such generated files.

packages/server/database/migrations/20221103134908-adjustUserTiers.ts
```diff
@@ -0,0 +1,101 @@
...
+        `SELECT * FROM "User"
+        WHERE id = $1;`,
+        [userId]
+      )
+    })
+  )
+  const users = usersRes.map((user) => user.rows[0]) as unknown as IUser[]
+  users.forEach((user) => {
+    user &&
+      segmentIo.identify({
```
Dschoordsch, line 89: -2 Doing an external API call in a database migration seems really strange to me and I'm sure there are many reasons against it, failing to deploy because segment is not available is just one of them. Please remove this.

Dschoordsch (changes requested):
Did only check your typecheck issue

tghanken (approved):
Looks good from a data perspective.  Excited to be able to trust this field again!

packages/server/database/migrations/20221103134908-adjustUserTiers.ts
Dschoordsch: I didn't follow it either, it was in the error message
```
requireStack: [
    '/home/runner/work/parabol/parabol/packages/server/postgres/queries/getUsersByIds.ts',
    '/home/runner/work/parabol/parabol/packages/server/utils/segmentIo.ts',
    '/home/runner/work/parabol/parabol/packages/server/database/migrations/202211031348-adjustUserTiers.ts',
    '/home/runner/work/parabol/parabol/node_modules/rethinkdb-ts-migrate/lib/migrate.js',
    '/home/runner/work/parabol/parabol/scripts/migrate.js'
  ]
```

packages/server/database/migrations/20221107180234-adjustUserTiers.ts
```diff
@@ -0,0 +1,80 @@
...
+import OrganizationUser from '../types/OrganizationUser'
+import User from '../types/User'
+
+export const up = async function (r: R) {
+  const removedOrgUserIds = await r
+    .table('OrganizationUser')
+    .hasFields('removedAt')
+    .filter((row) => row('tier').ne('personal'))('userId')
+    .distinct()
+    .run()
```
Dschoordsch, line 13: I checked and this results in about 6000 rows which seems manageable.

packages/server/database/migrations/20221107180234-adjustUserTiers.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  await r
+    .table('User')
+    .getAll(r.args(removedOrgUserIds))
+    .update(
```
Dschoordsch, line 18: +1 Did you check if you could just return the changes to omit the next query in l43?

packages/server/database/migrations/20221107180234-adjustUserTiers.ts
```diff
@@ -0,0 +1,80 @@
...
+        'personal'
+      )
+    }))
+    .run()) as {id: string; tier: TierEnum}[]
+
+  const pg = getPg()
+  await Promise.all(
+    userTiers.map((userTier) => {
+      return pg.query(
+        `UPDATE "User" AS u SET "tier" = $1::"TierEnum"
```
Dschoordsch, line 68: +1 `AS u` seems redundant because `WHERE $2 = "id"` should work just fine because it's not ambiguous.

Dschoordsch (approved):
Did a short smoke test and it worked.

Issue #7143, opened 2022-08-31
Meetings: improve top bar layout on mobile


enriquesanchez:
## Issue - Bug

The layout of the top bar (meeting title, team members) can be improved on small screens, particularly when the meeting title is long:

<img width="448" alt="Captura de Pantalla 2022-08-31 a la(s) 5 54 25 p m" src="https://user-images.githubusercontent.com/3276087/187800023-bae0d246-d59f-4389-aa94-6ce2441d38f4.png">

In addition to this, there is no indication when the user reaches the limit of number of characters allowed in the title. Users are left to wonder what is happening as they type and see that nothing is happening.

### Acceptance Criteria

Design a solution that:
- [ ] Improves the mobile layout when the meeting title is long
- [ ] Improves the UX when editing the title and reaching the max number of characters
- [ ] Made sure fix applies to all meeting types (check-in, retro, sprint poker, and standups)

## Estimated effort:
- 4 hrs

jordanh:
Noting related issue: #7167

PR #7398, opened 2022-11-04
fix: Meetings: improve top bar layout on mobile

gitstart:
# Description

Fixes/Partially Fixes [issue](https://github.com/ParabolInc/parabol/issues/7143)
Issue #7143

- Improves the standup meeting topbar user experience when the meeting title is too long on mobile view.
- Helps to notify user when the max number of characters allowable has been reached for mobile and normal view.
- Improves general look of the standup topbar on mobile view.

## Demo
https://user-images.githubusercontent.com/116144318/221913817-34a7b1f2-d32f-4a94-81ba-5f982e8bb303.mov

## Testing scenarios

- [x] Improve mobile view layout for Standup Topbar.   
- [x] Notify the user when the standup meeting name limit has been reached.
- [x] Notify the user when the organization name limit has been reached.

## Final checklist


gitstart-parabol:

packages/client/components/BackButton.tsx
```diff
@@ -17,6 +17,8 @@ const IconButton = styled(FlatButton)({
 
 const BackIcon = styled(ArrowBack)({
   color: 'inherit'
+  position: 'fixed',
+  bottom: '94%'
```
BartoszJarocki, line 21: This doesn't look good, why hardcode something that looks like a very arbitrary number?

BartoszJarocki (changes requested):

gitstart-parabol:
Hi @BartoszJarocki, what's the verdict on this?

BartoszJarocki (changes requested):

packages/client/validation/Legitity.ts
```diff
@@ -39,7 +39,7 @@ class Legitity {
     // this.value.length gives us the count of UTF-16 units, so 🔥 has a length of 2
     // Spreading the string into an array gives us the desired length in codepoints (characters): https://stackoverflow.com/a/54369605
     const value = [...this.value]
-    if (!this.error && value.length > len) {
+    if (!this.error && value.length >= len) {
```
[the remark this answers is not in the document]
gitstart-parabol: @BartoszJarocki the current code only checks if the length of the input is greater than the specified error case length, which is why the error indicator does not appear when the `max` for the standup 'title' is reached; however, this change will be improved.

gitstart-parabol:
@BartoszJarocki this PR is ready for review.

BartoszJarocki (approved):
looks good! @Dschoordsch do you mind checking it on Android? thanks!

packages/client/validation/Legitity.ts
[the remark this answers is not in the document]
Dschoordsch: -2 Changing this breaks the intend of the function. I expect `new Legitity('123').max(3)` to pass, not to fail. I made this explicit in #7863.

Dschoordsch (changes requested):
See Legitity comment

packages/client/validation/Legitity.ts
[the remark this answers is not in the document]
gitstart-parabol: Okay, we'll revert this change to reinstate the desired behaviour 🙌

Dschoordsch (changes requested):
The warning does not appear anymore

gitstart-parabol:
> The warning does not appear anymore

Hi @Dschoordsch, after a bit of investigation, we discovered that the existing validator `Legitity.max` will never catch any value greater than the maximum length because the `maxLength` is also being passed down to the `TextArea` and `Input` components. Because of this, the input field constraints prevent the user from entering any text that would trigger the warning.

Is there any other strategy you can suggest to get a more preferred outcome?

Dschoordsch:
The behaviour we want is:
* the user can input characters up to maxlength without any warning
* if the user writes the character maxlength+1 we show the warning, that means we should remove maxlength from the text input
* we could either forbid saving or truncate the text to maxlength on saving. I would argue the latter is more convenient and thus we should go this way

gitstart-parabol:
@Dschoordsch, This PR awaits another review.

packages/client/components/EditableText.tsx
```diff
@@ -166,7 +166,7 @@ const EditableText = forwardRef((props: Props, ref: any) => {
     const inProps = {
       autoFocus: true,
       ref: inputRef as any,
-      maxLength,
+      maxLength: maxLength + 1,
```
Dschoordsch, line 169: -2 This component is used elsewhere, please don't change global components to fix a local issue without retesting all other usages of those components.

Dschoordsch (changes requested):

packages/client/components/EditableText.tsx
gitstart-parabol: Hello @Dschoordsch , regarding the general fix in this PR, we figured that the bug mentioned  in this issue seemed to be occurring in other areas where this component was being used as well. For instance, when updating the `organization name`.

Dschoordsch: Can you please update the testing scenarios to the new changes?

gitstart-parabol: Hi @Dschoordsch, we are unable to locate any ui tests that require updating on this instance. This recent change only result in a visual difference, without affecting the current unit test suite for the `Legality` file.

Dschoordsch: @gitstart-parabol sorry for not being more clear what I meant: Could you please update the manual test cases mentioned in the description of this PR to also test for regressions in other places where your changes apply?

mattkrick:

PR #7400, opened 2022-11-04
chore: Cleanup packages to support future bundled server build

mattkrick:
# Description

Following up from https://github.com/ParabolInc/parabol/pull/6965.
The previous PR attempted to bundle node_modules into the server.
This change had to be reverted because our current digital ocean deployment doesn't have sharp tools installed on the container. It was also difficult because it took a bunch of memory to create those builds in CI.

A large client needs us to bundle everything together.
To accomplish this, we'll maintain a separate fork that does the bundling. That requires pulling in all the recent changes from productino whenever we bump their version. In the future, we might make this a flag.
The goal of this PR is to reduce the delta between that separate fork & the production branch.
This PR re-adds the chores like package upgrades & using non-native packages, _but does not change any of the build steps_.


scripts/webpack/dev.servers.config.js
```diff
@@ -31,16 +31,15 @@ module.exports = {
   output: {
     filename: '[name].js',
     path: path.join(PROJECT_ROOT, 'dev'),
-    libraryTarget: 'commonjs',
-    publicPath: `http://localhost:${process.env.PORT}/static/`
```
mattkrick, line 35: publicPath is used for things like jira images that the cache locally. this has been moved to the image file loader below for specificity & to avoid the future use case where `.node` native modules are loaded from publicPath

scripts/webpack/prod.client.config.js
```diff
@@ -182,35 +180,6 @@ module.exports = ({isDeploy, isStats}) => ({
           }
         ]
       },
-      {
-        test: /GitHubManager\.ts/,
```
mattkrick, line 186: this file was removed long ago, I just never updated the webpack config

scripts/webpack/prod.servers.config.js
```diff
@@ -88,9 +87,22 @@ module.exports = ({isDeploy}) => ({
...
+            loader: 'file-loader',
+            options: {
+              publicPath
+            }
           }
         ]
+      },
+      {
+        include: [/node_modules/],
+        test: /\.node$/,
```
mattkrick, line 99: this rule doesn't do anything since all the node_modules are ignored via the `nodeExternals` call above
(it does a whole lot for the bundling fork though)

packages/server/utils/AzureDevOpsServerManager.ts
```diff
@@ -1,4 +1,3 @@
-import AbortController from 'abort-controller'
```
mattkrick, line 1: node v16 has a native AbortController that we can use instead of the ponyfill

Dschoordsch (approved):

PR #7402, opened 2022-11-04
feat: create bundles without node_modules

mattkrick:
This is what parabol looks like without node_modules.  We can use this for our docker containers to keep our images lightweight. The source maps aren't necessary, but we can discuss uploading them to sentry if we like.

Screenshot 2023-02-10 at 12 37 05 PM

To build:
- `yarn build --no-deps`

To run migrations:
- `node dist/migrate.js`

To run postdeploy:
- `node dist/postDeploy.js`

note: the builds on this branch WILL FAIL in CI due to out of memory errors. that's OK! we'll build them elsewhere 
@adaniels-parabol


mattkrick:
@adaniels-parabol calling `node dist/migrate.js` should not longer have external dependencies

adaniels-parabol (approved):

Issue #7344, opened 2022-10-25
Implement restrictions on public templates


nickoferrall:
Predecessors https://github.com/ParabolInc/parabol/issues/7332 & https://github.com/ParabolInc/parabol/issues/7341 & https://github.com/ParabolInc/parabol/issues/7336

### AC

- [ ] Implement the design that comes out of https://github.com/ParabolInc/parabol/issues/7332
- [ ] If `MeetingTemplate.isPremium === true && tier === 'personal'`, the template shows the restricted UI
- [ ] If  `MeetingTemplate.isPremium === false`, show the free, publically accessible template UI
- [ ] If `MeetingTemplate.isPremium === true && tier !== 'personal'`, show the unlocked UI
- [ ] If a user clicks on the "Upgrade Now" CTA in the template picker, track the event in Segment
- [ ] Update the [selectTemplate](https://github.com/ParabolInc/parabol/blob/209eb714c484f074cedc92dddc3cb2fcb2960823/packages/server/graphql/mutations/selectTemplate.ts) mutation to check whether the user should have access to the template
- [ ] update the template ordering to take accessibility into account https://github.com/ParabolInc/parabol/blob/3545d153c1d0260f836646ce5c6a7beec79d7292/packages/server/graphql/queries/helpers/getPublicScoredTemplates.ts#L4

Dschoordsch:
I came across the `isStarter` property in the templates and found it's only used for sorting. We should modify the sorting to punish locked templates, I added an AC.

Issue #4099, opened 2020-08-21
Refine public template sharing


mattkrick:
managing user generated content is quite possibly the hardest technical problem in the world today.
To manage public template, we have a few options:

1. Wait for a public template to get used 10 times. Only then may it appear on the list. The more often it gets used & the newer it is, the higher it ranks. (This is how it works today)
2. Human moderation. Certain users (e.g. Parabol employees) see all public templates & can approve/reject them
3. Go full blown tik-tok. send a random sampling of 20 templates to all users. let them vote them up/down. templates that get upvoted/used are more likely to appear for more users

I think the gatekeeping of option 1 kinda sucks. a template can be good without it being used 10 times.
Option 2 is just a dumbed down version of Option 3.

So, let's segment users into a couple buckets: early access & standard.
Early access:
- See content as soon as it's released (it could be ugly)
- Can flag content (inappropriate, low quality, etc.)
- Can mark content as viewable for standard

Standard
- See only content that is good
- Can flag content (inappropriate, low quality, etc). Their votes probably count for less

So, when a template gets marked as public:
- look at the publisher. do they have other good stuff? if so, let it go straight to standard. else, it needs to go to just early access
- when content gets flagged by early access, it will never be seen by standard & it appears less to other early access folks. maybe it has to get flagged a couple times.
- when content gets marked viewable (again, maybe a couple times) it'll be visible to standard folks
- when something gets flagged, the approved drops, causing it to be seen less, maybe at all
- if someone approves a template & then it gets flagged, we learn about the approver and the flagger. should this impact their weight?

avivapinchas:
Clarifying questions: 
> Wait for a public template to get used 10 times. Only then may it appear on the list. The more often it gets used & the newer it is, the higher it ranks. (This is how it works today)
This is fully how it's implemented today? I was out last week and saw some discussion about some first few public templates, so assuming this is how it works now after latest fixes? 

>  let's segment users into a couple buckets
How do we determine who is early access and who is standard? 

> look at the publisher. do they have other good stuff?
How do we define 'other good stuff' here?

mattkrick:
> This is fully how it's implemented today? I was out last week and saw some discussion about some first few public templates, so assuming this is how it works now after latest fixes?

correct! there was a bug that didn't filter them out properly

> How do we determine who is early access and who is standard?

to start with, only employees are early access. as we grow our list of superusers/cheerleaders, we can add them, too.

> How do we define 'other good stuff' here?

[...]

PR #7405, opened 2022-11-07
feat: restrict public templates

nickoferrall:
Fix https://github.com/parabolinc/parabol/issues/7344

### DIFF
--- a/packages/client/utils/makeTemplateDescription.ts
+++ b/packages/client/utils/makeTemplateDescription.ts
@@ -34,11 +34,10 @@ const makeTemplateDescription = (
     `,
     viewerRef
   )
-  if (!viewer) return ''
-  const {featureFlags} = viewer
+  const showTemplateLimit = viewer?.featureFlags.templateLimit
   const {lastUsedAt, team, isFree} = template
   const {name: teamName, tier} = team
-  if (lowestScope === 'PUBLIC' && featureFlags.templateLimit) {
+  if (lowestScope === 'PUBLIC' && showTemplateLimit) {
     if (isFree) return 'Free template'
     return `Premium template ${tier === 'personal' ? '🔒' : '✨'}`
   }
--- a/packages/client/utils/makeTemplateDescription.ts
+++ b/packages/client/utils/makeTemplateDescription.ts
@@ -3,12 +3,14 @@ import {useFragment} from 'react-relay'
 import {readInlineData} from 'relay-runtime'
 import {makeTemplateDescription_template$key} from '../__generated__/makeTemplateDescription_template.graphql'
 import {makeTemplateDescription_viewer$key} from '../__generated__/makeTemplateDescription_viewer.graphql'
+import {TierEnum} from '../__generated__/SendClientSegmentEventMutation.graphql'
 import relativeDate from './date/relativeDate'
 
 const makeTemplateDescription = (
   lowestScope: string,
   templateRef: makeTemplateDescription_template$key,
-  viewerRef: makeTemplateDescription_viewer$key | null
+  viewerRef?: makeTemplateDescription_viewer$key | null,
+  tier?: TierEnum
 ) => {
   const template = readInlineData(
     graphql`
@@ -18,7 +20,6 @@ const makeTemplateDescription = (
         isFree
         team {
           name
-          tier
         }
       }
     `,
@@ -32,14 +33,13 @@ const makeTemplateDescription = (
         }
       }
     `,
-    viewerRef
+    viewerRef ?? null
   )
   const showTemplateLimit = viewer?.featureFlags.templateLimit
   const {lastUsedAt, team, isFree} = template
-  const {name: teamName, tier} = team
+  const {name: teamName} = team
   if (lowestScope === 'PUBLIC' && showTemplateLimit) {
-    if (isFree) return 'Free template'
-    return `Premium template ${tier === 'personal' ? '🔒' : '✨'}`
+    return isFree ? 'Free template' : `Premium template ${tier === 'personal' ? '🔒' : '✨'}`
   }
   if (lowestScope === 'TEAM')
     return lastUsedAt
--- a/packages/client/utils/makeTemplateDescription.ts
+++ b/packages/client/utils/makeTemplateDescription.ts
@@ -1,26 +1,47 @@
 import graphql from 'babel-plugin-relay/macro'
+import {useFragment} from 'react-relay'
 import {readInlineData} from 'relay-runtime'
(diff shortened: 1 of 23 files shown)


packages/server/graphql/queries/helpers/getPublicScoredTemplates.ts
```diff
@@ -2,19 +2,21 @@ import db from '../../../db'
...
+  templates: {createdAt: Date; id: string; isStarter?: boolean; isFree?: boolean}[]
 ) => {
   const sharedTemplateIds = templates.map(({id}) => id)
   const sharedTemplateEndTimes = await db.readMany('endTimesByTemplateId', sharedTemplateIds)
   const scoreByTemplateId = {} as {[templateId: string]: number}
   templates.forEach((template, idx) => {
-    const {id: templateId, createdAt, isStarter} = template
+    const {id: templateId, createdAt, isStarter, isFree} = template
     const endTimes = sharedTemplateEndTimes[idx]!
+    const isFreeBonus = isFree ? 1000 : 0
```
nickoferrall, line 13: Hey @mattkrick, we have the `isStarter` field in `TemplateScale` from https://github.com/ParabolInc/parabol/pull/4220. It looks like all public templates have `isStarter === true`, but as `getPublicScoredTemplates` is checking if `isStarter` is true, I imagine it wasn't intended to be this way?

Perhaps it was meant to indicate which templates should feature higher up the list?

Now that we're restricting templates, templates that are accessible to everyone have the `isFree` field. We want those to feature at the top of the list.

I want to check:
1) That my understanding of the `isStarter` templates is correct 
2) Whether you think adding a large bonus to the `isFree` templates is the right move?
3) Should `isStarter` be removed if it's true for all public templates?

packages/client/utils/makeTemplateDescription.ts
```diff
@@ -1,26 +1,46 @@
...
 import {makeTemplateDescription_template$key} from '../__generated__/makeTemplateDescription_template.graphql'
+import {makeTemplateDescription_viewer$key} from '../__generated__/makeTemplateDescription_viewer.graphql'
+import {TierEnum} from '../__generated__/SendClientSegmentEventMutation.graphql'
 import relativeDate from './date/relativeDate'
 
 const makeTemplateDescription = (
   lowestScope: string,
-  templateRef: makeTemplateDescription_template$key
+  templateRef: makeTemplateDescription_template$key,
+  viewerRef?: makeTemplateDescription_viewer$key | null,
```
nickoferrall, line 12: In Poker, there are no premium templates. We could pass the `viewerRef` to this function from the poker files so that the Poker public template subtitles say "Free template" instead of "Created by Parabol", but that would involve passing the `viewerRef` through many components, as you can see in this PR on the retro side. 

Instead, I'd rather keep the Poker public template subtitles as they are for now. When we remove the feature flag (https://github.com/ParabolInc/parabol/issues/7413), the Poker subtitles will change to "Free template", just like the retro templates.

nickoferrall: Hmm we'll actually need access to the `featureFlags` and therefore the `User` object in Poker to show the custom template team & org copy. I'll add it to this PR - it'll probably bump it up to a large, but as other PRs will build off of this one, I think it's the right move. Removing my request for a review until this is done

igorlesnenko (commented):
> Open the template picker in Sprint Poker and see that both public templates have the "Free template" subtitle

When I open poker as paid user I see both templates are premium, is it expected?

image


When I open poker as a free user, they both premium but I can still use `Weighted Shortest Job First` but there is no any button on  `Estimated Effort`

image
image

Is it expected behaviour?

nickoferrall:
Thanks for the review, @igorlesnenko. Just wondering if you've pulled the latest commit from this branch and you're still seeing that behaviour? 

> No button on one template

That's the expected behaviour if the template has been selected

> And "use template" button on another, but it does not work

What happens when you click on the button?

> Also see different icons, in sidebar it is stars, and on the right panel there is a lock

I haven't been able to reproduce that. Let me know if you're still getting the same behaviour & if you have any tips on how to repro it. Here's a Loom video of what I'm seeing: https://www.loom.com/share/6007f266ec9a4096affc732f2ca63744

igorlesnenko:
@nickoferrall looks like I see the issue only on free teams: https://www.loom.com/share/8e3bea0c48ca4d3ab35dc8629a755c5f

Let me know if the loom is helpful or I need to provide more info

nickoferrall:
I think [this migration](https://github.com/ParabolInc/parabol/blob/c0b16ca938ca98bf110a72f4852163b5bf3f9375/packages/server/database/migrations/20221107145022-makePokerPublicTemplatesFree.ts) may not have run (even though you ran `yarn db:migrate` 🤔 ). Would you mind checking whether `isFree` is true [here](https://github.com/ParabolInc/parabol/blob/547cf93bc6cf7af362e5de12d61367f702721782/packages/client/utils/makeTemplateDescription.ts#L38)? Here's what I see when I create a new org & team: https://www.loom.com/share/1fbe7e778a0b46bc87c458dc5bf59372

I'd be happy to Zoom if that'd be easier!

igorlesnenko:
@nickoferrall you are correct, migration didn't run for some reason. I manually updated templated and now it works.
Not sure why migration didn't run, but it is probably not related to the PR

igorlesnenko (approved):

packages/server/graphql/queries/helpers/getPublicScoredTemplates.ts
nickoferrall: @mattkrick just bumping this question

mattkrick: yep, you got it. the old logic was to allow folks to mark their custom templates as public & then as the templates gain traction, other folks start to use them & we create a kind of marketplace for templates. We stopped building that (https://github.com/ParabolInc/parabol/issues/4099), but the groundwork is still there.

Since moderating user-generated templates is so difficult & that's no longer the direction we're heading, I'd say it's probably safe to remove that type of logic entirely. `getPublicScoredTemplates` could be removed &  `publicTemplates` could simply return everything created by parabol (ie `teamId: 'aGhostTeam'`)

adding `isFree` sounds like a good move!

nickoferrall: Sounds good! I've implemented that change. Now that we're no longer moving in the user-generated templates direction, shall we remove the "share with the world" option as custom templates can only be shared with the team/org?

share-with-world

Dschoordsch (changes requested):
Did not test

PR #7424, opened 2022-11-11
fix: typo in Earth 2.0 check-in question

jordanh:
# Description

Trivial fix of check-in question typo


jordanh:
Bah. The silly editor on GitHub touched WAY too many lines on commit. Going to revert this commit and fix up this commit. So much for it being done in under 30 seconds.

Issue #5745, opened 2021-12-03
Select dropdown can't be scrolled with keyboard


enriquesanchez:
## Issue - Bug

When interacting with a select dropdown, such as the team select dropdown for team in task cards, I can't see all the available items if I'm navigating with a keyboard. The keyboard selection does no scroll the list.

https://user-images.githubusercontent.com/3276087/144638646-3c420ca2-34bc-4f98-a54b-92dfaebc8795.mp4

In the video above, you'll notice that I can't tell if `Product` is selected because the list won't scroll with my selections. I have to scroll with a mouse in order to see which item I have selected.

- **How serious is the bug?** P4
- **OS:** macOS
- **Browser:** Firefox 94.0.1

### Acceptance Criteria
Users can:
   - View all items in a select dropdown when navigating with a keyboard. Scroll should follow keyboard selection.

### Estimated effort:
1 hour to triage. More if root cause is already identified.

enriquesanchez:
Un-stale it!

PR #7426, opened 2022-11-14
fix: select dropdown can't be scrolled with keyboard

gitstart:
# Description 
Select dropdown can't be scrolled with keyboard

Fixes  [IssueLink](https://github.com/ParabolInc/parabol/issues/5745)  #5745

## Demo
https://user-images.githubusercontent.com/1501599/202843311-f0dcf82f-881e-4e7e-99f2-31e048897259.mov

## Testing scenarios

- [x] Scenario A
  - Create a team and add more than 7 members to the team then
  - Open the select dropdown and  Interact over the  list items using the keyboard.

## Final checklist


gitstart-parabol:

packages/client/components/Menu.tsx
```diff
@@ -102,17 +104,17 @@ const Menu = forwardRef((props: Props, ref: any) => {
...
       if (idx === null) setActiveIdx(firstIndex)
       else if (menuItemIdxs.includes(idx)) setActiveIdx(idx)
-      else if (idx < firstIndex) setActiveIdx(lastIndex)
-      else if (idx > lastIndex) setActiveIdx(firstIndex)
-      else if (activeIdx && idx > activeIdx) {
-        for (let ii = idx; ii <= lastIndex; ii++) {
+      else if (idx > firstIndex) setActiveIdx(lastIndex)
+      else if (idx < lastIndex) setActiveIdx(firstIndex)
+      else if (activeIdx && idx < activeIdx) {
+        for (let ii = idx; ii >= lastIndex; ii--) {
```
Dschoordsch, line 110: -2 What happened here? Why is this change necessary?

Dschoordsch (changes requested):

BartoszJarocki:
@gitstart what's more, this would be great to have in every PR you submit. it'll save us a lot of review time.

packages/client/components/Menu.tsx
```diff
@@ -157,10 +157,8 @@ const Menu = forwardRef((props: Props, ref: any) => {
   const handleKeyDown = useCallback(
     (e: React.KeyboardEvent) => {
       if (e.key === 'ArrowDown') {
-        e.preventDefault()
         setSafeIdx(activeIdx === null ? null : activeIdx + 1)
       } else if (e.key === 'ArrowUp') {
-        e.preventDefault()
```
gitstart-instrumental, line 163: @BartoszJarocki After some testing, we concluded that the default action to the event `onKeyDown` allows the parent element to automatically scroll to any focusable children elements. Since `e.preventDefault()` was nuked on `ArrowDown` and `ArrowUp`, the default behavior described in the block above was reinstated. This allows for smooth scrolling down and up with the keyboard arrow keys.

gitstart-parabol, line 163: @BartoszJarocki After some testing, we concluded that the default action to the event `onKeyDown` allows the parent element to automatically scroll to any focusable children elements. Since `e.preventDefault()` was nuked on `ArrowDown` and `ArrowUp`, the default behavior described in the block above was reinstated. This allows for smooth scrolling down and up with the keyboard arrow keys.

gitstart-parabol:
Hi @BartoszJarocki @Dschoordsch, this PR is ready for review

packages/client/components/Menu.tsx
Dschoordsch: I couldn't find a reason these are here, so I think you're right in removing these.

Dschoordsch (approved):
Works, also does not break the emoji dialog

Issue #7342, opened 2022-10-25
Implement custom template upgrade feedback


nickoferrall:
Predecessor: https://github.com/ParabolInc/parabol/issues/7330

### AC

- [ ] Implement the design in https://github.com/ParabolInc/parabol/issues/7330
- [ ] We send a Segment event when the user clicks the CTA button

PR #7427, opened 2022-11-14
feat: custom template upgrade feedback

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7342

### DIFF
--- a/packages/client/modules/meeting/components/AddNewPokerTemplate.tsx
+++ b/packages/client/modules/meeting/components/AddNewPokerTemplate.tsx
@@ -1,14 +1,16 @@
 import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
 import React, {useEffect, useRef} from 'react'
-import {createFragmentContainer} from 'react-relay'
+import {useFragment} from 'react-relay'
 import LinkButton from '../../../components/LinkButton'
 import TooltipStyled from '../../../components/TooltipStyled'
 import useAtmosphere from '../../../hooks/useAtmosphere'
 import useMutationProps from '../../../hooks/useMutationProps'
 import AddPokerTemplateMutation from '../../../mutations/AddPokerTemplateMutation'
 import {Threshold} from '../../../types/constEnums'
-import {AddNewPokerTemplate_pokerTemplates} from '../../../__generated__/AddNewPokerTemplate_pokerTemplates.graphql'
+import {AddNewPokerTemplate_pokerTemplates$key} from '../../../__generated__/AddNewPokerTemplate_pokerTemplates.graphql'
+import {AddNewPokerTemplate_team$key} from '../../../__generated__/AddNewPokerTemplate_team.graphql'
+import {AddNewPokerTemplate_viewer$key} from '../../../__generated__/AddNewPokerTemplate_viewer.graphql'
 
 const ErrorLine = styled(TooltipStyled)({
   margin: '0 0 8px'
@@ -28,13 +30,43 @@ const AddPokerTemplateLink = styled(LinkButton)({
 
 interface Props {
   gotoTeamTemplates: () => void
-  pokerTemplates: AddNewPokerTemplate_pokerTemplates
-  teamId: string
+  pokerTemplatesRef: AddNewPokerTemplate_pokerTemplates$key
+  viewerRef: AddNewPokerTemplate_viewer$key
+  teamRef: AddNewPokerTemplate_team$key
+  setShowUpgradeDetails: (showUpgradeDetails: boolean) => void
 }
 
 const AddNewPokerTemplate = (props: Props) => {
-  const {gotoTeamTemplates, teamId, pokerTemplates} = props
+  const {gotoTeamTemplates, teamRef, pokerTemplatesRef, viewerRef, setShowUpgradeDetails} = props
   const atmosphere = useAtmosphere()
+  const pokerTemplates = useFragment(
+    graphql`
+      fragment AddNewPokerTemplate_pokerTemplates on PokerTemplate @relay(plural: true) {
+        name
+      }
+    `,
+    pokerTemplatesRef
+  )
+  const {featureFlags} = useFragment(
+    graphql`
+      fragment AddNewPokerTemplate_viewer on User {
+        featureFlags {
+          templateLimit
+        }
+      }
+    `,
+    viewerRef
+  )
+  const team = useFragment(
+    graphql`
+      fragment AddNewPokerTemplate_team on Team {
+        id
+        tier
+      }
+    `,
+    teamRef
+  )
+  const {id: teamId, tier} = team
   const {onError, onCompleted, submitMutation, submitting, error} = useMutationProps()
   const errorTimerId = useRef<undefined | number>()
   useEffect(() => {
@@ -44,6 +76,10 @@ const AddNewPokerTemplate = (props: Props) => {
   }, [])
   const addNewTemplate = () => {
     if (submitting) return
+    if (featureFlags.templateLimit && tier === 'personal') {
+      setShowUpgradeDetails(true)
+      return
+    }
     if (pokerTemplates.length >= Threshold.MAX_RETRO_TEAM_TEMPLATES) {
       onError(
         new Error(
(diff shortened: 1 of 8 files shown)


nickoferrall:
@acressall I'd love to get your thoughts on the design in this one please 🙏

acressall:
@nickoferrall I tried testing this locally, but got an error after checking out to this PR. Here's the use case I wanted to try:
- I am a current free user
- I have one or more custom templates on my account/team
- The template restriction is added
- I can still see my previously created custom templates
- I can no longer create new custom templates

^^ that's assuming we _want_ these templates to still be available to existing users

Everything else in the Loom looks great!

igorlesnenko (approved):
looks good to me

packages/client/modules/meeting/components/SelectTemplate.tsx
```diff
@@ -41,6 +41,10 @@ const Button = styled(FloatingActionButton)({
   pointerEvents: 'all'
 })
 
+const UpgradeButton = styled(Button)({
+  padding: '10px 24px'
```
nickoferrall, line 45: Increased size of upgrade button following feedback [here](https://parabol.slack.com/archives/C042E8PGPD4/p1668552680439969?thread_ts=1668522709.276439&cid=C042E8PGPD4)

packages/client/modules/meeting/components/AddNewPokerTemplate.tsx
```diff
@@ -74,16 +110,10 @@ const AddNewPokerTemplate = (props: Props) => {
     <div>
       {error && <ErrorLine>{error.message}</ErrorLine>}
       <AddPokerTemplateLink palette='blue' onClick={addNewTemplate} waiting={submitting}>
-        {'Create New Template'}
+        {`Create New Template ${featureFlags.templateLimit && tier === 'personal' ? '🔒' : ''}`}
```
Dschoordsch, line 113: +1 Could be simplified to
```ts
        Create New Template {featureFlags.templateLimit && tier === 'personal' && '🔒'}
```

packages/client/modules/meeting/components/AddNewPokerTemplate.tsx
```diff
@@ -28,13 +30,43 @@ const AddPokerTemplateLink = styled(LinkButton)({
 
 interface Props {
   gotoTeamTemplates: () => void
-  pokerTemplates: AddNewPokerTemplate_pokerTemplates
-  teamId: string
+  pokerTemplatesRef: AddNewPokerTemplate_pokerTemplates$key
+  viewerRef: AddNewPokerTemplate_viewer$key
+  teamRef: AddNewPokerTemplate_team$key
+  setShowUpgradeDetails: (showUpgradeDetails: boolean) => void
```
Dschoordsch, line 36: -1 it's only ever called with `true` so it should be `showUpgradeDetails: () => void` instead

packages/client/modules/meeting/components/AddNewReflectTemplate.tsx
```diff
@@ -28,13 +30,43 @@ const AddRetroTemplateLink = styled(LinkButton)({
 
 interface Props {
   gotoTeamTemplates: () => void
-  reflectTemplates: AddNewReflectTemplate_reflectTemplates
-  teamId: string
+  reflectTemplatesRef: AddNewReflectTemplate_reflectTemplates$key
+  viewerRef: AddNewReflectTemplate_viewer$key
+  setShowUpgradeDetails: (showUpgradeDetails: boolean) => void
```
Dschoordsch, line 35: -1 it's only ever called with `true` so it should be `showUpgradeDetails: () => void` instead

Dschoordsch (changes requested):
There is some flickering before showing/hiding the message
Flickering when showing the message
(GIF did not record it well enough, but it's visible in both cases)

nickoferrall:
@Dschoordsch I don't see that flickering on mine, but I also don't have that persistent vertical scrollbar, so perhaps it's browser specific. Which browser are you using? 

Do you see any flickering when switching between Team/Org/Public tabs or clicking "Create New Template" on master?

packages/client/modules/meeting/components/AddNewPokerTemplate.tsx
nickoferrall: It's called with `false` in the modal: https://github.com/ParabolInc/parabol/blob/2eaee63fe719d48b781de220e40cc88ba395c8a4/packages/client/modules/meeting/components/PokerTemplateModal.tsx#L85. In this file, it's only ever called with `true`, so I could change it to `showUpgradeDetails: (showUpgradeDetails: true) => void`?

Dschoordsch: It should be changed to `showUpgradeDetails: () => void` in all the components who receive the callback. The fact it is a true/false state in PokerTemplateModal is not relevant for the other components.

nickoferrall: `showUpgradeDetails` is the boolean value, so do you mean it should be changed to `setShowUpgradeDetails: () => void`? That wouldn't work in this file ,`AddNewPokerTemplate`, as we're passing true into `setShowUpgradeDetails` on line 80.

Or are you thinking that in `PokerTemplateModal` I do the following:

```

  const displayUpgradeDetails = () => {
    setShowUpgradeDetails(true)
  }

  const hideUpgradeDetails = () => {
    setShowUpgradeDetails(false)
  }

  useEffect(() => {
    if (showUpgradeDetails) hideUpgradeDetails()
  }, [activeTemplate])
```

and pass `displayUpgradeDetails` down to the child components? Then the prop type would be `displayUpgradeDetails: () => void`

Dschoordsch:
I see the flickering on Firefox and Chromium. It's only happening when showing/hiding the new overlay of yours. It looks like an issue with the graphic of the template not being available on a render pass while switching.

packages/client/modules/meeting/components/AddNewPokerTemplate.tsx
Dschoordsch: Yes, the second part. Sorry if that wasn't clear. Create a function you pass down to the other dependencies which is `() => void`.

Dschoordsch:
> I see the flickering on Firefox and Chromium. It's only happening when showing/hiding the new overlay of yours. It looks like an issue with the graphic of the template not being available on a render pass while switching.

I retried a bit more and could also see the flickering by just switching between templates in some situations, so I assume it's something which was there before.

packages/client/modules/meeting/components/AddNewPokerTemplate.tsx
nickoferrall: Makes sense! I've pushed an update with that change 👍

nickoferrall:
> I retried a bit more and could also see the flickering by just switching between templates in some situations, so I assume it's something which was there before.

Thanks for retrying. Are you happy for me to create an issue for that bug and merge this PR? 

Are you using Linux or Windows by any chance?

Dschoordsch:
> Thanks for retrying. Are you happy for me to create an issue for that bug and merge this PR?
> 
> Are you using Linux or Windows by any chance?

I'm using linux. No need to create an extra issue for it.

packages/client/modules/meeting/components/AddNewPokerTemplate.tsx
```diff
@@ -74,16 +110,10 @@ const AddNewPokerTemplate = (props: Props) => {
     <div>
       {error && <ErrorLine>{error.message}</ErrorLine>}
       <AddPokerTemplateLink palette='blue' onClick={addNewTemplate} waiting={submitting}>
-        {'Create New Template'}
+        {`Create New Template ${featureFlags.templateLimit && tier === 'personal' && '🔒'}`}
```
Dschoordsch, line 113: -1 This is not what I had suggested and will produce "Create New Template false". You're doing string interpolation here which will print the result of `${}`. I had suggested to move this one level up (i.e. remove the surrounding {} and backticks), then React would take the `false` component and ignore it. Either way, now it's broken.

packages/client/modules/meeting/components/AddNewReflectTemplate.tsx
```diff
@@ -71,16 +107,10 @@ const AddNewReflectTemplate = (props: Props) => {
     <div>
       {error && <ErrorLine>{error.message}</ErrorLine>}
       <AddRetroTemplateLink palette='blue' onClick={addNewTemplate} waiting={submitting}>
-        {'Create New Template'}
+        {`Create New Template ${featureFlags.templateLimit && tier === 'personal' && '🔒'}`}
```
Dschoordsch, line 110: -1 see above

Dschoordsch (changes requested):

Dschoordsch (approved):

PR #7429, opened 2022-11-14
fix: Organization Image is on top of default

### LINKED ISSUE #6822: Organization Image is on top of default
## Issue - Bug
When uploading a profile image for my organization, if the image has a transparent background the default is still visible behind
Screen Shot 2022-06-23 at 9 44 16 AM

Users can:
   - When an image is uploaded it _replaces_ the default
   - The border around the profile is maintained

gitstart-parabol:
# Description

Fixes #6822
When uploading a profile image for `my organization`, if the image has a transparent background the default is still visible behind. 

## Demo

https://www.loom.com/share/a2845f6db1ec438c829682f2742377a2

## Testing scenarios

- [ ] Scenario A
  - Pull the branch and run the app
  - Got to `me/organizations` url
  - Click on `your user name`s organization
  - Click on the profile photo

## Final checklist


packages/client/components/Avatar/Avatar.tsx
```diff
@@ -3,14 +3,18 @@ import React, {forwardRef} from 'react'
...
-type ImageBlockProps = Pick<Props, 'sansRadius' | 'sansShadow' | 'picture' | 'size' | 'onClick'>
+type ImageBlockProps = Pick<
+  Props,
+  'sansRadius' | 'sansShadow' | 'hasBorder' | 'picture' | 'size' | 'onClick'
+>
 
 const ImageBlock = styled('div')<ImageBlockProps>(
-  ({sansRadius, sansShadow, picture, size, onClick}) => ({
-    backgroundImage: `url(${picture}), url(${defaultUserAvatar})`,
+  ({sansRadius, sansShadow, hasBorder, picture, size, onClick}) => ({
```
gitstart-parabol, line 12: The `hasBorder` prop is used to indicate whether the profile image should retain its border.

packages/client/components/Avatar/Avatar.tsx
```diff
@@ -3,14 +3,18 @@ import React, {forwardRef} from 'react'
...
+    backgroundImage: `url(${picture || defaultUserAvatar})`,
     backgroundPosition: 'center center',
     backgroundRepeat: 'no-repeat',
     backgroundSize: 'cover',
+    border: hasBorder ? '4px solid #B6AEC6' : undefined,
```
gitstart-parabol, line 17: The border hex code was gotten from the placeholder image so it can retain a similar border.

gitstart-parabol:

igorlesnenko (changes requested):
Now we have a double border on the default image:

image

Because image itself has a border, and we add an additional border.

igorlesnenko (changes requested):
Actually, I suggest to remove this manually added border at all. 

Reasons:
- We don't have it elsewhere around the organization image in the app, why add this on that particular screen?
- Matching it manually with the image is fragile, what if we change the default image later? Ideally, this shouldn't be a part of the default image too, but made with css.
- We currently does not show a border anyway if you upload a non transparent image. So it is a new change introduced rather than a fix.

cc @acressall

igorlesnenko:
> @igorlesnenko are we cropping it do be a circle? 

We crop it only using CSS:

image

> . Do you think we should make the background round, so it would be a closer design to the profile settings?

Do you mean organization page on my screenshot above? Yes, I think it would be more consistent to make it round there too.
Yes, this would be a CSS change

> My only issue is that the white background is a square, so it might be confusing without the border to show you where it will be cropped.

Make sense.
What if we add some shadow? This will work both with default avatar and any uploaded avatar. Could this be a solution?

image

acressall:
Looks good @igorlesnenko! Let's make it round and add the shadow

igorlesnenko (changes requested):
Looks good! Just one more small fix:

When applying the shadow to the default avatar, there is a weird sub-pixel appears

image

image

Maybe we can use some lighter shadow color like '#DADADA' ?

gitstart-parabol:
@igorlesnenko  Please re-evaluate, review suggestions have been implemented.

Dschoordsch (approved):

PR #7438, opened 2022-11-17
fix: Line breaks in comments now work on Android

### LINKED ISSUE #7378: Line breaks are not supported in comments
## Issue - Bug

When a user makes a list in the comment using the line break (which should take the items to the next line) and sends it, the list still appears in a single line.
The defect captured in the loom video below [works as expected on the webview]

https://www.loom.com/share/243265d13666499884cc2f845dec33ac

- **OS:** Windows
- **Browser:** Chrome 107.0.5304.88
- **Node version:** `16.16.0`
- **NPM version:** `8.11.0`

Users can:
   - Write a comment with line breaks and see the line breaks after the comments are submitted

gitstart-parabol:
# Description
After adding new line on a Mobile (Android) devices, content shows up as a in a single line character

Fixes  #7378

## Step to reproduce Bug
- Open app on a Mobile (Android) devices
- Navigate to any meeting, add a comment and reply
- In the replies try entering text with new line
- after submitting u will notice the new line was added

## Demo
[Loom Video](https://www.loom.com/share/c358cf04079045d5971c1128e94f4973)

## Testing scenarios

- [x] Test if user can add new line on Mobile (Android) devices
  - Reply to replies with a texts containing new line

## Final checklist


gitstart-parabol:

igorlesnenko (changes requested):
If possible, let's fix the original feature so it supports all devices

packages/client/utils/draftjs/convertMobileValueToTaskContent.ts
```diff
@@ -0,0 +1,10 @@
+import {ContentState} from 'draft-js'
+import {convertStateToRaw} from './convertToTaskContent'
+
+const convertAndroidValueToTaskContent = (spacedText: string) => {
```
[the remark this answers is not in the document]
gitstart-parabol: Thanks for catching this @igorlesnenko, we actually built this function specifically for Android devices since these were the only devices experiencing this. Hence, we have updated the file name accordingly.

packages/client/components/DiscussionThreadInput.tsx
```diff
@@ -226,7 +226,7 @@ const DiscussionThreadInput = forwardRef((props: Props, ref: any) => {
       if (!editorEl || editorEl.type !== 'textarea') return
       const {value} = editorEl
       if (!ensureHasText(value)) return
-      addComment(convertToTaskContent(value))
+      addComment(convertAndroidValueToTaskContent(value))
```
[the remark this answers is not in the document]
gitstart-parabol: - Actually, this fix is only meant for Android devices since they are the only ones experiencing this issue.
- We found out that this comment is already being called within a conditional statement in line 225 of the `DiscussionThreadInput.tsx` file checking if it is an Android device, so we guessed this function was originally built only for cases where Android devices are used.

packages/client/utils/draftjs/convertMobileValueToTaskContent.ts
```diff
@@ -0,0 +1,10 @@
...
+  const text = spacedText.trim()
```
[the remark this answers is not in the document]
gitstart-parabol: Just as you mentioned, the previous implementation deleted spaces within the comment, this also meant it was removing all line breaks entered into the comments, which resulted in non-support of line breaks within the comments. Since this was only happening on Android devices, we decided to solve the problem by only removing the spaces around the comments on Android devices, thereby preserving any line breaks entered by the user within the comments.

gitstart-parabol:
Hi @igorlesnenko, your feedback comments have been addressed, please let us know if we need to clarify anything, thanks.

gitstart-parabol:
@igorlesnenko,  This PR is ready for another review.

packages/client/utils/draftjs/convertToTaskContent.ts
```diff
@@ -1,11 +1,17 @@
...
+import isAndroid from './isAndroid'
 
-export const removeSpaces = (str: string) =>
-  str
-    .split(/\s/)
-    .filter((s) => s.length)
-    .join(' ')
+export const removeSpaces = (str: string) => {
+  if (isAndroid) {
+    return str.replace(/ +(?= )/g, '').trim()
```
igorlesnenko, line 7: Would this function give different result on non-android devices? 

If this function works identically on all devices, then let's leave only this function for all cases and remove `isAndroid` check.

igorlesnenko (changes requested):
I'm getting the following error when trying to run `yarn dev`.

```
20:29:18 5|GraphQL Schema Updater  | ReferenceError: navigator is not defined
20:29:18 5|GraphQL Schema Updater  |     at isAndroid (webpack://parabol-action/./packages/client/utils/draftjs/isAndroid.ts?:6:3)
```

Also the build for this PR fails too with the same error.

packages/client/utils/draftjs/convertToTaskContent.ts
gitstart-parabol: @igorlesnenko This function only works for Android devices; otherwise, we thought it would be ideal to have a single function, but we couldn't.

igorlesnenko: @gitstart-parabol what result it will return on non-android devices? How would it be different from android devices?

packages/client/utils/draftjs/convertToTaskContent.ts
```diff
@@ -1,11 +1,17 @@
 import {ContentState, convertToRaw} from 'draft-js'
 import entitizeText from './entitizeText'
+import isAndroid from './isAndroid'
 
-export const removeSpaces = (str: string) =>
-  str
-    .split(/\s/)
-    .filter((s) => s.length)
-    .join(' ')
+export const removeSpaces = (str: string) => {
```
igorlesnenko, line 5: @gitstart-parabol I looked a little deeper into the problem. I see that we already check for `isAndroid` on higher lever here
https://github.com/ParabolInc/parabol/blob/78e0a8386a325cac3bcdb30608b271eec955e304/packages/client/components/DiscussionThreadInput.tsx#L225

And we do not use `convertToTaskContent` function on non-android devices, which means we do not strip any spaces at all.

It also says `convert to task` but there is no `task`, which is a bit weird.

We probably don't need to fix `removeSpaces` function.

What if we do the following:

Replace 

https://github.com/ParabolInc/parabol/blob/78e0a8386a325cac3bcdb30608b271eec955e304/packages/client/components/DiscussionThreadInput.tsx#L229

With 

https://github.com/ParabolInc/parabol/blob/0dbd37ee1cefaeeee27a11cf547cf710f321a7fd/packages/client/utils/draftjs/convertToTaskContent.ts#L25-L26

Would something like this work?

igorlesnenko (changes requested):

packages/client/utils/draftjs/convertToTaskContent.ts
gitstart-parabol: @igorlesnenko, Thank you for making the suggestions; they were really helpful in resolving the problems.

gitstart-parabol:
@igorlesnenko, what's the verdict on this?

igorlesnenko:
@gitstart-parabol looks good, can I ask you what tool did you use to test on mobile?

gitstart-parabol:
> Loom Video

Android studio emulator and `chrome://inspect/#devices`

igorlesnenko:
@gitstart-parabol Thansk! It works for me now

image

The only issue I see is that the PR build is failing, looks like some prettier issue, can you check? 

image

gitstart-parabol:
> @gitstart-parabol Thanks! It works for me now
> 
> image
> 
> The only issue I see is that the PR build is failing, looks like some prettier issue, can you check?
> 
> image

gitstart-parabol:
> Details

 @igorlesnenko  Let us double-check so that this can be resolved as soon as possible.

gitstart-parabol:
> @gitstart-parabol Thanks! It works for me now
> 
> image
> 
> The only issue I see is that the PR build is failing, looks like some prettier issue, can you check?
> 
> image

 @igorlesnenko  Let us double-check so that this can be resolved as soon as possible.

gitstart-parabol:
Hi @igorlesnenko, the PR build CI error has been fixed, you can check again now.

Dschoordsch (approved):

PR #7441, opened 2022-11-17
feat(teams-limit): add limit exceeded and locking flags

### LINKED ISSUE #7385: Teams limit: Add tier limit exceeded flags
Let's add some flags in the database and API as a separate issue, so we can use them to implement various functionality in parallel.

See #7379

AC:
- Add and expose in API `tierLimitExceededAt ` flag 
- Add and expose in API `scheduledLockAt ` flag
- `lockedAt`

igorlesnenko:
Closes #7385

How to test:

Set flags on any org

http://localhost:8080/#dataexplorer
```
r.db('actionDevelopment').table('Organization').filter({
  id: 'ijVADO81eo'
}).update({
  tierLimitExceededAt: new Date('2021-01-01'),
  scheduledLockAt: new Date('2021-02-02') ,
  lockedAt: new Date('2021-03-03') 
})
```

Run query http://localhost:3000/admin/graphql

```
{
  viewer {
    organization(orgId: "ijVADO81eo") {
      tierLimitExceededAt
      scheduledLockAt
      lockedAt
    }
  }
}
```

### DIFF
--- a/packages/server/database/types/Organization.ts
+++ b/packages/server/database/types/Organization.ts
@@ -14,6 +14,9 @@ interface Input {
   updatedAt?: Date
   showConversionModal?: boolean
   payLaterClickCount?: number
+  tierLimitExceededAt?: Date
+  scheduledLockAt?: Date
+  lockedAt?: Date
 }
 
 export default class Organization {
@@ -32,6 +35,9 @@ export default class Organization {
   stripeSubscriptionId?: string | null
   upcomingInvoiceEmailSentAt?: Date
   tier: TierEnum
+  tierLimitExceededAt?: Date
+  scheduledLockAt?: Date
+  lockedAt?: Date
   updatedAt: Date
   constructor(input: Input) {
     const {
@@ -45,7 +51,10 @@ export default class Organization {
       showConversionModal,
       payLaterClickCount,
       picture,
-      tier
+      tier,
+      tierLimitExceededAt,
+      scheduledLockAt,
+      lockedAt
     } = input
     this.id = id || generateUID()
     this.activeDomain = activeDomain
@@ -58,5 +67,8 @@ export default class Organization {
     this.picture = picture
     this.showConversionModal = showConversionModal === null ? undefined : showConversionModal
     this.payLaterClickCount = payLaterClickCount || 0
+    this.tierLimitExceededAt = tierLimitExceededAt
+    this.scheduledLockAt = scheduledLockAt
+    this.lockedAt = lockedAt
   }
 }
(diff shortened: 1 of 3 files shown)


packages/server/graphql/public/typeDefs/Organization.graphql
```diff
@@ -72,6 +72,21 @@ type Organization {
...
+
+  """
+  Schedule the organization to be locked at
+  """
+  scheduledLockAt: DateTime
+
+  """
+  Organization locked at
+  """
+  lockedAt: DateTime
```
nickoferrall, line 88: Just wondering why we need both `lockedAt` and `scheduledLockAt`?

igorlesnenko: @nickoferrall We plan have a cron that will actually lock an organization, lock all related teams and send an email. This is a flag that locking process is finished, emails were sent etc.

nickoferrall (approved):
LGTM!

packages/server/graphql/types/Organization.ts
```diff
@@ -85,6 +85,18 @@ const Organization: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<a
...
     },
+    tierLimitExceededAt: {
+      type: GraphQLISO8601Type,
+      description: 'Flag the organization as exceeding the tariff limits by setting a datetime'
+    },
+    scheduledLockAt: {
+      type: GraphQLISO8601Type,
+      description: 'Schedule the organization to be locked at'
+    },
+    lockedAt: {
```
Dschoordsch, line 96: +1 Might be good to add this to the DB type already as well
https://github.com/ParabolInc/parabol/blob/9e4726a29de46fad8869a155a9ee8228c1e405fc/packages/server/database/types/Organization.ts#L16

Dschoordsch (approved):

packages/server/database/types/Organization.ts
```diff
@@ -14,6 +14,9 @@ interface Input {
   updatedAt?: Date
   showConversionModal?: boolean
   payLaterClickCount?: number
+  tierLimitExceededAt?: Date
+  scheduledLockAt?: Date
+  lockedAt?: Date
```
Dschoordsch, line 19: +1 I would imagine we never set these on construction and thus could omit them from the Input object

Dschoordsch (approved):

packages/server/database/types/Organization.ts
igorlesnenko: Make sense

Dschoordsch: You just removed the assignment, now it's broken because you could add it to the `Input` object but it would not get assigned anymore in the constructor

Issue #7240, opened 2022-09-30
Auto focus on new template title


nickoferrall:
[This chart](https://analytics.amplitude.com/parabol/chart/inrjy9s?source=search) in amplitude suggests that users may not know how to edit some template titles.

There are several templates that are being regularly used with the word "Copy" at the end, e.g. `Start, Stop, Continue Copy`.

The title `*New Template` is being used by almost 20 users each week in total across dozens of teams, [see chart](https://analytics.amplitude.com/parabol/chart/z4jdghx?source=redirect%3A%20chart%20saved).

I suggest that we auto-focus on the title after clicking "Create custom template" or copying a template. See Loom for further info: https://www.loom.com/share/a315a07f09b94fcf9ba373b7c80d52ab

### AC

- [ ] After clicking "Create New Template" or clone & copy a template, the title is auto-focused and the text is empty
- [ ] The `Please enter a template name` error message doesn't show up initially on auto-focus, but if the user doesn't enter a title and they click away, it does show-up onBlur

jordanh:
Way to use data!

PR #7442, opened 2022-11-17
feat: focus on newly created template

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7240

Demo: https://www.loom.com/share/f72494cfa44b4b8185de636df04c3b45

### To test

- [ ] Create a new template and see that the newly created template item (the item in the list on the left) is highlighted
- [ ] See that the newly created template's title is autofocused
- [ ] Clone a template and see that the template item is highlighted and the title is autofocused
- [ ] Templates that are not newly created don't have their titles autofocused

### DIFF
--- a/packages/client/hooks/useActiveTopTemplate.ts
+++ b/packages/client/hooks/useActiveTopTemplate.ts
@@ -29,7 +29,7 @@ const useActiveTopTemplate = (
         })
       }
     }, 300)
-  }, [isActive, selectedTemplateId])
+  }, [isActive, selectedTemplateId, edges])
 }
 
 export default useActiveTopTemplate
--- a/packages/client/modules/meeting/components/EditableTemplateName.tsx
+++ b/packages/client/modules/meeting/components/EditableTemplateName.tsx
@@ -30,6 +30,7 @@ const EditableTemplateName = (props: Props) => {
   const {name, templateId, teamTemplates, isOwner} = props
   const atmosphere = useAtmosphere()
   const {onError, error, onCompleted, submitMutation, submitting} = useMutationProps()
+  const autoFocus = name === '*New Template' || name.slice(-4) === 'Copy'
 
   const handleSubmit = (rawName: string) => {
     if (submitting) return
@@ -66,6 +67,7 @@ const EditableTemplateName = (props: Props) => {
   return (
     <InheritedStyles>
       <StyledEditableText
+        autoFocus={autoFocus}
         disabled={!isOwner}
         error={error ? error.message : undefined}
         handleSubmit={handleSubmit}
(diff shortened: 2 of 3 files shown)


packages/client/modules/meeting/components/EditableTemplateName.tsx
```diff
@@ -30,6 +30,7 @@ const EditableTemplateName = (props: Props) => {
   const {name, templateId, teamTemplates, isOwner} = props
   const atmosphere = useAtmosphere()
   const {onError, error, onCompleted, submitMutation, submitting} = useMutationProps()
+  const autoFocus = name === '*New Template' || name.slice(-4) === 'Copy'
```
nickoferrall, line 33: Feels like a brittle way of detecting whether it's a newly created template, but we are doing a similar thing elsewhere: https://github.com/ParabolInc/parabol/blob/f21825c30161d761d49df00a0f6fe2e24ac25792/packages/client/modules/meeting/components/AddNewReflectTemplate.tsx#L80

packages/client/hooks/useActiveTopTemplate.ts
```diff
@@ -29,7 +29,7 @@ const useActiveTopTemplate = (
         })
       }
     }, 300)
-  }, [isActive, selectedTemplateId])
+  }, [isActive, selectedTemplateId, edges])
```
nickoferrall, line 32: This was the bug. The `selectedTemplateId` would initially be the optimistic id, but once it changed to the id from the server, `edges` wouldn't update - it would still contain the optimistic id so `isSelectedInView` would be false

packages/client/utils/relay/setActiveTemplate.ts
```diff
@@ -6,14 +6,14 @@ import Atmosphere from '../../Atmosphere'
 const setActiveTemplateInRelayStore = (
   store: RecordSourceProxy,
   teamId: string,
-  templateId: string | null,
+  templateId: string,
```
nickoferrall, line 9: This isn't related to the fix - while digging around, I just noticed that this should be `string` instead of `string | null`

igorlesnenko (approved):
Great fix!

igorlesnenko (approved):
Looks good to me!

packages/client/modules/meeting/components/EditableTemplateName.tsx
Dschoordsch: I think it's safe to go forward with this change. I think `name.endsWith(' Copy')` would be nicer to read though.

Dschoordsch (approved):
Did not test

packages/client/modules/meeting/components/EditableTemplateName.tsx
nickoferrall: Nice, yeah I prefer `endsWith` 👍

PR #7446, opened 2022-11-17
feat:  oauth2 redirect to cloudflare

### LINKED ISSUE #7445: Support OAuth2 across all PPMIs
- Our integrations require using oauth2
- Many integrations only support a single callback url, e.g. action.parabol.co/auth/atlassian
- that single works for folks using prod, but not for dev, staging, PPMIs, etc.
- until now, we've worked around this by creating a brand new app (e.g. a new atlassian 3LO app) for each PPMI
- a better way is to have a single endpoint that redirects to the right origin

So, a user wants to integrate with jira. They send an oauth2 request with a state that looks like this:
```ts
{hash: randomBits, origin: 'http://localhost:3000', service: 'atlassian'}
```

that state gets forwarded on to our proxy, which will be a cloudflare worker. it could be anything, but cloudflare gives you 100k jobs/day for FREE.

that worker reads the state & redirects the request to the proper origin.
~~the vuln exists in the unverified `state.origin`. if an attacker gets a user to go through the oauth2 flow where the state.origin is controlled by an attacker, e.g. untrusted-hacker.com, they they could get the user's code since the identity provider will redirect to untrusted-hacker.com/?code=mysecretcode. that would allow the attacker to trade in the code for an oauth2 token where they could make requests on behalf of the victim.~~

nevermind, the attacker would need our atlassian server secret to get that token :tada:

mattkrick:
# Description

Fixes #7445

This allows us to have a single redirect URL for all integration oauth2 flows for all deployments (local, dev, staging, prod, PPMIs, etc).

## Demo

https://www.loom.com/share/8334c429e48246eda9678c060377f452

TEST
- [ ] in your local .env, set OAUTH2_REDIRECT=https://oauth2redirect.parabol.workers.dev/
- [ ] integrate to jira to make sure it works
- [ ] check for security vulns

Before Merging
- [x] Add OAUTH2_REDIRECT to env
- [x] Add GitHub & Slack support

Immediately after deploying:
- [ ] goto https://github.com/organizations/ParabolInc/settings/applications/504449
- [ ] update redirect url to https://oauth2redirect.parabol.workers.dev
- [ ] goto https://api.slack.com/apps/A4ZM70S93/oauth?
- [ ] remove the top 3 update urls (all except the new one)
- [ ] goto https://developer.atlassian.com/console/myapps/debd66fe-456c-4901-bf2a-91e1ac5e8b72/authorization/auth-code-grant
- [ ] update the redirect url to https://oauth2redirect.parabol.workers.dev


packages/client/utils/AtlassianClientManager.ts
```diff
@@ -14,8 +13,11 @@ class AtlassianClientManager extends AtlassianManager {
     scopes: JiraPermissionScope[] = AtlassianManager.SCOPE
   ) {
     const {submitting, onError, onCompleted, submitMutation} = mutationProps
-    const providerState = Math.random().toString(36).substring(5)
-    const redirect = makeHref('/auth/atlassian')
+    const hash = Math.random().toString(36).substring(5)
+    const providerState = btoa(
+      JSON.stringify({hash, origin: window.location.origin, service: 'atlassian'})
```
mattkrick, line 18: this is the magic, the client is passing its origin & service to atlassian, which then after auth passes it to our cloud function, which then forwards it to the origin

packages/server/oauth2redirects.js
```diff
@@ -0,0 +1,34 @@
...
+  let destinationURL
+  try {
+    destinationURL = new URL(origin)
+  } catch (e) {
+    return new Response('Invalid origin provided in state')
+  }
+  destinationURL.pathname = pathname
+  destinationURL.search = search
+  const destinationStr = destinationURL.toString()
+  return Response.redirect(destinationStr, 301)
```
Dschoordsch, line 29: -1 Permanent redirect seems wrong. If anyone is ever using multiple PPMIs this could fail.

mattkrick: 302 it is!

packages/client/utils/AtlassianClientManager.ts
mattkrick: `appOrigin` should just be for the server going off the server's `HOST, PORT` env vars. At least I hope we haven't used it on the client anywhere :sweat_smile: 

update: we use the name `appOrigin` in SSR components like emails (https://github.com/ParabolInc/parabol/blob/b656e7d44061aabd78abadfb05e79d61de9bd181/packages/client/modules/summary/components/NewMeetingSummary.tsx#L65), but still pass in `location.origin`

packages/server/oauth2redirects.js
```diff
@@ -0,0 +1,34 @@
...
+    parsedState = JSON.parse(atob(state))
+  } catch (e) {
+    // nope
+  }
+  if (!parsedState || typeof parsedState !== 'object') {
+    return new Response('No state received in response')
+  }
+
+  // can remove fallback in prod
+  const {origin, service} = parsedState.origin
```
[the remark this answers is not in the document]
mattkrick: ahhh i clean up my code before committing & break everything. good catch!

mattkrick:
Screenshot from 2022-11-28 11-07-43

confirmed working for slack & github in local dev, too.

Issue #7451, opened 2022-11-18
Prevent cloning templates


nickoferrall:
Now that we're restricting users in "Starter tiers" from creating new templates, we need to prevent cloning existing templates too.

### AC

- [ ] Users in the free Starter tier are unable to clone a template
- [ ] Users in the paying Team tier can

PR #7452, opened 2022-11-18
feat: prevent cloning templates

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7451

Demo: https://www.loom.com/share/de77bb5303c84a50b185de71fbce6505

- [ ] Pro teams can continue to clone templates
- [ ] Free teams cannot clone templates


igorlesnenko (approved):
Looks good to me!

Dschoordsch (approved):
I think we should check this on server side, too. So there should be some check in `addReflectTemplate` and `addPokerTemplate` mutations. Feel free to create a follow up issue for this.

Did smoke test.

nickoferrall:
> I think we should check this on server side, too. So there should be some check in `addReflectTemplate` and `addPokerTemplate` mutations. Feel free to create a follow up issue for this.
> 
> Did smoke test.

Makes sense, I've created an issue: https://github.com/ParabolInc/parabol/issues/7487

PR #7461, opened 2022-11-22
fix(dx): Run all yarn clean commands, even if one fails

jmtaber129:
# Description
`yarn clean` will fail if certain expected files are missing due to the use of the `&&` operator. Instead, use `;`, which will run all commands regardless of the result of the preceding commands.

## Testing scenarios
Run `yarn clean`, then running `yarn clean` again and see that all commands are still executed.

## Final checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "start": "pm2-runtime start pm2.config.js",
     "heroku-postbuild": "echo 'Skipping build'",
     "build": "node scripts/prod.js",
-    "clean": "rm -rf ./packages/client/__generated__; rm ./queryMap.json; rm ./packages/server/graphql/public/schema.graphql; rm ./packages/server/graphql/private/schema.graphql; rm -rf ./dev; rm -rf ./dist; rm -rf ./build",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
     "start": "pm2-runtime start pm2.config.js",
     "heroku-postbuild": "echo 'Skipping build'",
     "build": "node scripts/prod.js",
-    "clean": "rm -rf ./packages/client/__generated__ && rm ./queryMap.json && rm ./packages/server/graphql/public/schema.graphql && rm ./packages/server/graphql/private/schema.graphql && rm -rf ./dev && rm -rf ./dist && rm -rf ./build",
+    "clean": "rm -rf ./packages/client/__generated__; rm ./queryMap.json; rm ./packages/server/graphql/public/schema.graphql; rm ./packages/server/graphql/private/schema.graphql; rm -rf ./dev; rm -rf ./dist; rm -rf ./build",
```
Dschoordsch, line 27: +1 Any reason not to combine everything into one `rm` command?

Dschoordsch (approved):

package.json
jmtaber129: Nope! That makes way more sense, changed to just one `rm -rf` command

Issue #7487, opened 2022-11-23
Prevent cloning templates from the server


nickoferrall:
See https://github.com/ParabolInc/parabol/pull/7452#pullrequestreview-1191459538 

### AC

- [ ] `addReflectTemplate` mutation checks whether the user should be able to clone the template
- [ ] `addPokerTemplate` mutation checks whether the user should be able to clone the template

PR #7489, opened 2022-11-23
feat: prevent creating new templates from server

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7487

### To test

- [ ] Update the client so a user on the free team can create a new retro template or clone an existing one, i.e. remove https://github.com/ParabolInc/parabol/blob/a7f18329d9a5047831b28a06feb572918f08e478/packages/client/modules/meeting/components/AddNewReflectTemplate.tsx#L81 and set `showClone` to `true` in https://github.com/ParabolInc/parabol/blob/8043bce3f29688e84d51fc7f57d35bec79142560/packages/client/modules/meeting/components/ReflectTemplateDetails.tsx#L160
- [ ] Try to create a new template and see that it's not possible as `addReflectTemplate` returns an error
- [ ] Do the same for Poker and see that `addPokerTemplate` returns an error

### DIFF
--- a/packages/server/graphql/mutations/addPokerTemplate.ts
+++ b/packages/server/graphql/mutations/addPokerTemplate.ts
@@ -4,6 +4,7 @@ import getRethink from '../../database/rethinkDriver'
 import {RDatum} from '../../database/stricterR'
 import PokerTemplate from '../../database/types/PokerTemplate'
 import TemplateDimension from '../../database/types/TemplateDimension'
+import {getUserById} from '../../postgres/queries/getUsersByIds'
 import {getUserId, isTeamMember} from '../../utils/authorization'
 import publish from '../../utils/publish'
 import standardError from '../../utils/standardError'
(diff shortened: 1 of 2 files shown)


igorlesnenko (approved):

packages/server/graphql/mutations/addPokerTemplate.ts
```diff
@@ -53,6 +54,11 @@ const addPokerTemplate = {
     if (!viewerTeam) {
       return standardError(new Error('Team not found'), {userId: viewerId})
     }
+    const viewer = await getUserById(viewerId)
```
Dschoordsch, line 57: -1 Let's use the dataloader here `await dataloader.get('users').loadNonNull(viewerId)` just in case we load the viewer for other reasons in the same resolve call.

packages/server/graphql/mutations/addReflectTemplate.ts
```diff
@@ -54,6 +55,11 @@ const addReflectTemplate = {
     if (!viewerTeam) {
       return standardError(new Error('Team not found'), {userId: viewerId})
     }
+    const viewer = await getUserById(viewerId)
```
Dschoordsch, line 58: -1 Please use the dataloader here.

Dschoordsch (changes requested):

Dschoordsch (approved):
Works

Issue #7471, opened 2022-11-22
AI Summary: implement topic & discussion summary


nickoferrall:
In the retrospective meeting summary, include a GPT-3 generated ~2 sentence summary of all the reflections in the topic and another summary sentence of all of the comments in the topic.

### AC

- [ ] Add a `topicSummary` field (this is a summary of the reflections in the topic) to the `Discuss` phase GraphQL object and db table
- [ ] Add a `discussionSummary` field (this is a summary of the comments in the topic) to the `Discuss` phase GraphQL object and db table
- [ ] When the meeting proceeds to the Discuss phase, use GPT-3 to generate the `topicSummary`
- [ ] When the meeting ends, use GPT-3 to generate the `discussionSummary`
- [ ] Include the `topicSummary` and `discussionSummary` in the meeting summary, following the designs in https://github.com/ParabolInc/parabol/issues/7468

jordanh:
Please update AC for what happens when GPT-3 is unavailable (e.g. if Parabol is running on an air-gapped network)

nickoferrall:
@jordanh issue description updated 👍

PR #7494, opened 2022-11-29
feat: topic summary in meeting summary

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7471

Loom demo: https://www.loom.com/share/9b1c4eff94fd45b8bea98aa75e379af4

### To test

- [ ] Accept the invite to Parabol's OpenAI account
- [ ] Add the `OPEN_AI_API_KEY` & `OPEN_AI_ORG_ID` keys to the `.env` file. You can find this [here](https://github.com/ParabolInc/action-devops/blob/master/environments/local#L46-L47) 🔒 
- [ ] Add the `aiSummary` feature flag 
- [ ] Create a retro meeting, add reflections, group them together, and end the meeting. In the meeting, see a topic summary, which is a GPT-3 generated summary of the reflection groups
- [ ] If there is only one reflection in the reflection group, there is no topic summary as it's redundant
- [ ] If the user does not have an `OPEN_AI_API_KEY`, e.g. if Parabol is running on an air-gapped network, it will fail gracefully. The user will see the regular meeting summary that currently exists in prod

### DIFF
--- a/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/RetroTopic.tsx
+++ b/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/RetroTopic.tsx
@@ -89,13 +89,13 @@ const RetroTopic = (props: Props) => {
   const stage = useFragment(
     graphql`
       fragment RetroTopic_stage on RetroDiscussStage {
-        topicSummary
         reflectionGroup {
           title
           voteCount
           reflections {
             ...EmailReflectionCard_reflection
           }
+          summary
         }
         discussion {
           commentCount
@@ -118,9 +118,9 @@ const RetroTopic = (props: Props) => {
     `,
     meetingRef
   )
-  const {reflectionGroup, discussion, topicSummary} = stage
+  const {reflectionGroup, discussion} = stage
   const {commentCount} = discussion
-  const {reflections, title, voteCount} = reflectionGroup!
+  const {reflections, title, voteCount, summary} = reflectionGroup!
   const imageSource = isEmail ? 'static' : 'local'
   const icon = imageSource === 'local' ? 'thumb_up_18.svg' : 'thumb_up_18@3x.png'
   const src = `${ExternalLinks.EMAIL_CDN}${icon}`
@@ -133,7 +133,7 @@ const RetroTopic = (props: Props) => {
       : `See ${commentCount} ${plural(commentCount, 'Comment')}`
   const commentLinkStyle = commentCount === 0 ? noCommentLinkStyle : someCommentsLinkStyle
   const corsOptions = isEmail ? EMAIL_CORS_OPTIONS : APP_CORS_OPTIONS
-  const showSummary = viewerMeetingMember?.user.featureFlags.aiSummary && topicSummary
+  const showSummary = viewerMeetingMember?.user.featureFlags.aiSummary && summary
   return (
     <>
       <tr>
@@ -155,7 +155,7 @@ const RetroTopic = (props: Props) => {
               >{`AI generated summaries are a premium feature. We'll share them with you in your first few retros so you can see what they're like.`}</td>
             </tr>
             <tr>
-              <td style={textStyle}>{topicSummary}</td>
+              <td style={textStyle}>{summary}</td>
             </tr>
           </td>
         </tr>
--- a/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/RetroTopic.tsx
+++ b/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/RetroTopic.tsx
@@ -133,7 +133,7 @@ const RetroTopic = (props: Props) => {
       : `See ${commentCount} ${plural(commentCount, 'Comment')}`
   const commentLinkStyle = commentCount === 0 ? noCommentLinkStyle : someCommentsLinkStyle
   const corsOptions = isEmail ? EMAIL_CORS_OPTIONS : APP_CORS_OPTIONS
-  const showAiSummary = viewerMeetingMember?.user.featureFlags.aiSummary && topicSummary
+  const showSummary = viewerMeetingMember?.user.featureFlags.aiSummary && topicSummary
   return (
     <>
       <tr>
@@ -143,7 +143,7 @@ const RetroTopic = (props: Props) => {
           </AnchorIfEmail>
         </td>
       </tr>
-      {showAiSummary && (
+      {showSummary && (
         <tr>
           <td align='left' style={{lineHeight: '22px', fontSize: 14}}>
             <tr>
--- a/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/RetroTopic.tsx
+++ b/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/RetroTopic.tsx
@@ -46,6 +46,30 @@ const someCommentsLinkStyle = {
   textDecoration: 'none'
 }
 
+const topicTitleStyle = {
+  color: PALETTE.SLATE_700,
+  fontFamily: FONT_FAMILY.SANS_SERIF,
+  fontWeight: 600,
+  lineHeight: '22px',
+  fontSize: 14,
+  padding: '8px 48px'
+}
+
+const subtitleStyle = {
+  color: PALETTE.SLATE_700,
+  fontFamily: FONT_FAMILY.SANS_SERIF,
+  fontStyle: 'italic',
+  padding: '8px 48px',
+  fontSize: 14
+}
+
+const textStyle = {
+  color: PALETTE.SLATE_700,
+  fontFamily: FONT_FAMILY.SANS_SERIF,
+  padding: '0px 48px 8px 48px',
+  fontSize: 14
+}
+
 const noCommentLinkStyle = {
   ...someCommentsLinkStyle,
   color: PALETTE.SLATE_600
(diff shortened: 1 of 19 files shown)


packages/server/graphql/mutations/helpers/handleCompletedStage.ts
```diff
@@ -73,6 +74,7 @@ const handleCompletedRetrospectiveStage = async (
         })
         .run()
       data.meeting = meeting
+      generateGroupSummaries(meeting.id, dataLoader, facilitatorUserId)
```
nickoferrall, line 77: We're not awaiting here so that the meeting doesn't wait for the OpenAI API when proceeding to the Vote stage. It's possible that the user could race through the Vote & Discuss stage, reach the meeting summary, and not see the GPT-3 topic summary as the API hasn't responded in time. 

I think that's fine. They can always refresh the page once the API has responded and they'll then see the topic summary.

packages/server/graphql/mutations/helpers/generateGroupSummaries.ts
```diff
@@ -0,0 +1,34 @@
...
+    dataLoader.get('users').loadNonNull(facilitatorUserId)
+  ])
+  if (!facilitator.featureFlags.includes('aiSummary')) return
+  const r = await getRethink()
+  const manager = new OpenAIServerManager()
+  for (const group of reflectionGroups) {
+    const reflectionsByGroupId = reflections.filter(
+      ({reflectionGroupId}) => reflectionGroupId === group.id
+    )
+    if (reflectionsByGroupId.length <= 1) continue
```
nickoferrall, line 23: If there's only one reflection in the group, I don't think there's any value in summarising it. Doing this will reduce our API costs too

packages/server/database/updateStage.ts
```diff
@@ -8,7 +9,7 @@ import rMapIf from './rMapIf'
 const updateStage = async (
   meetingId: string,
   stageId: string,
-  phaseType: string,
+  phaseType: NewMeetingPhaseTypeEnum,
```
nickoferrall, line 12: Not related to this feature, but while working on it, I noticed this should be `NewMeetingPhaseTypeEnum` instead of `string`

packages/server/utils/OpenAIServerManager.ts
```diff
@@ -0,0 +1,37 @@
...
+    const configuration = new Configuration({apiKey: process.env.OPEN_AI_API_KEY})
+    this.openAIApi = new OpenAIApi(configuration)
+  }
+
+  async getSummary(text: string | string[]) {
+    if (!this.openAIApi) return null
+    try {
+      const response = await this.openAIApi.createCompletion({
+        model: 'text-davinci-003',
+        prompt: `Summarize this for a second-grade student in one or two sentences: ${text}`,
```
nickoferrall, line 21: Modified example from here: https://beta.openai.com/examples/default-summarize

igorlesnenko (commented):
It worked fine first time.

On the second meeting with only one group of more than 1 card I got the following error in console, and saw no summaries:

`Error: Request failed with status code 429`

After that it stopped working and I see no errors.

Does it work for you now? Did we run out of credits?

nickoferrall:
> Did we run out of credits?

Nope, we're still well within the credit limit. I'll take a look and see if I can reproduce it 👍

igorlesnenko:
> After that it stopped working and I see no errors.
Ah I see, I got a race condition for the third time, after refreshing the page — AI summery is here.

However, 429 error is still an issue. Weird that it appeared.

nickoferrall:
> 429 error is still an issue

It's a rate limit error. Maybe because we're both using the same token in different countries while on a personal plan. We can see if the same error shows up once we upgrade to a company plan

packages/server/graphql/mutations/helpers/generateGroupSummaries.ts
```diff
@@ -0,0 +1,34 @@
+import extractTextFromDraftString from 'parabol-client/utils/draftjs/extractTextFromDraftString'
+import getRethink from '../../../database/rethinkDriver'
+import OpenAIServerManager from '../../../utils/OpenAIServerManager'
+import {DataLoaderWorker} from '../../graphql'
+
+const generateGroupSummaries = async (
+  meetingId: string,
+  dataLoader: DataLoaderWorker,
```
igorlesnenko, line 8: nit: maybe put dataloader param last? (Could be my personal feeling)

packages/server/graphql/mutations/helpers/generateGroupSummaries.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (!facilitator.featureFlags.includes('aiSummary')) return
+  const r = await getRethink()
+  const manager = new OpenAIServerManager()
+  for (const group of reflectionGroups) {
+    const reflectionsByGroupId = reflections.filter(
+      ({reflectionGroupId}) => reflectionGroupId === group.id
+    )
+    if (reflectionsByGroupId.length <= 1) continue
+    const reflectionTextByGroupId = reflectionsByGroupId.map((reflection) =>
+      extractTextFromDraftString(reflection.content)
```
igorlesnenko, line 25: -1 We have `plaintextContent` field. Would that work instead?

igorlesnenko (changes requested):

packages/server/graphql/mutations/helpers/generateGroupSummaries.ts
nickoferrall: I also like the `dataLoader` to be last! Here I've put the `facilitatorUserId` last only because we use it to check the feature flag. Once the feature is shipped to everyone, we'll be able to remove that param so `dataLoader` will be last

igorlesnenko (approved):

packages/server/graphql/mutations/helpers/generateGroupSummaries.ts
igorlesnenko: Make sense

packages/server/graphql/mutations/helpers/generateGroupSummaries.ts
```diff
@@ -0,0 +1,33 @@
...
+  const manager = new OpenAIServerManager()
+  for (const group of reflectionGroups) {
+    const reflectionsByGroupId = reflections.filter(
+      ({reflectionGroupId}) => reflectionGroupId === group.id
+    )
+    if (reflectionsByGroupId.length <= 1) continue
+    const reflectionTextByGroupId = reflectionsByGroupId.map(
+      ({plaintextContent}) => plaintextContent
+    )
+    if (reflectionTextByGroupId.length === 0) continue
```
Dschoordsch, line 26: -1 This cannot happen because `.map` does not change the length of the array and `reflectionsByGroupId` has length >= 2.

packages/server/graphql/mutations/helpers/generateGroupSummaries.ts
```diff
@@ -0,0 +1,33 @@
...
+    const summary = await manager.getSummary(reflectionTextByGroupId)
+    if (!summary) continue
+    r.table('RetroReflectionGroup').get(group.id).update({summary}).run()
```
Dschoordsch, line 29: +1 Even if we don't await from the calling side, it would be nice to have the ability to do so, i.e. we should properly await all async calls in this function. Best would be to replace the line 18 with a
```ts
return Promise.all(reflectionGroups.map(async (group) => {
```
and await or return here.

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/RetroTopic.tsx
```diff
@@ -75,6 +133,7 @@ const RetroTopic = (props: Props) => {
       : `See ${commentCount} ${plural(commentCount, 'Comment')}`
   const commentLinkStyle = commentCount === 0 ? noCommentLinkStyle : someCommentsLinkStyle
   const corsOptions = isEmail ? EMAIL_CORS_OPTIONS : APP_CORS_OPTIONS
+  const showSummary = viewerMeetingMember?.user.featureFlags.aiSummary && summary
```
Dschoordsch, line 136: +1 because we generate the summary conditionally, I think there is no need to check for the feature flag here again.

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/RetroTopic.tsx
```diff
@@ -84,6 +143,23 @@ const RetroTopic = (props: Props) => {
...
+      {showSummary && (
+        <tr>
+          <td align='left' style={{lineHeight: '22px', fontSize: 14}}>
+            <tr>
+              <td style={topicTitleStyle}>{'Topic Summary:'}</td>
+            </tr>
+            <tr>
+              <td
+                style={subtitleStyle}
+              >{`AI generated summaries are a premium feature. We'll share them with you in your first few retros so you can see what they're like.`}</td>
```
Dschoordsch, line 155: -1 I don't see any check for the tier, so the message could be misleading for team/enterprise users.

Dschoordsch (changes requested):
I did not test yet.

Dschoordsch (approved):
The results are mixed, so maybe we should add a message it's experimental?
* strange formatting:
Screenshot from 2022-12-08 13-00-54
* this one was nice
Screenshot from 2022-12-08 13-01-33
"GPT is a computer program which understands what you write. If you write in German, GPT will still understand it."
So it was answering the questions rather than summarizing it.

Maybe we could give the AI a more personal touch to make the summary less absolute. Instead of "Topic Summary" write something like "🤖 Sam the meeting summarizer thinks"

nickoferrall:
I think that's a great idea. I've created an issue for it: https://github.com/ParabolInc/parabol/issues/7544

PR #7509, opened 2022-11-30
feat: upgrade node to 16.18.1

adaniels-parabol:
# Description

Upgrade node from 16.16.0 to 16.18.1


mattkrick:
let's bump to v18 instead so we can hang out on the newer LTS

Dschoordsch:
We moved to 18 in #7780

Issue #7220, opened 2022-09-22
Add "Heard Seen Respected" Template


garethaledavies:
To explore new product areas, I'd like to add a "Liberating Structures" activity into the product and see if it gets used. This can be used more as a structured conflict resolution or meeting culture retrospective. 

Liberating Structures are licensed under a non-commercial CC license where we also need to give attribution. I think we could show that attribution here (as suggested by @jordanh):

![Group 3 (1)](https://user-images.githubusercontent.com/15236825/191709919-63578313-b924-452b-8fb4-ef682115cad6.png)

> Originally created by Liberating Structures

Here are the prompts for the template:

**Heard, Seen, Respected (HSR)**

_Heard_
Think about times or forums in which you felt your voice was not heard. What happened? How did it feel?

_Seen_
Think about times or forums in which you felt you were not seen or your efforts were not recognized. What happened? How did it feel?

_Respected_
Think about times or forums in which you felt you, your boundaries, or your contributions were not respected. What happened? How did it feel?

For reference, here is the original template: [](https://www.liberatingstructures.com/19-heard-seen-respected-hsr/).

> "You can foster the empathetic capacity of participants to “walk in the shoes” of others. Many situations do not have immediate answers or clear resolutions. Recognizing these situations and responding with empathy can improve the “cultural climate” and build trust among group members. HSR helps individuals learn to respond in ways that do not overpromise or overcontrol. It helps members of a group notice unwanted patterns and work together on shifting to more productive interactions. Participants experience the practice of more compassion and the benefits it engenders."

garethaledavies:
This template page is now live on the website: [](https://www.parabol.co/templates/sprint-retrospectives/heard-seen-respected/)

Can someone please help me get it into the product?

PR #7510, opened 2022-12-01
feat: add heard seen respected template

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7220

### To test

- [ ] Run migration and see the newly added template

### DIFF
--- a/packages/server/database/migrations/20221201121536-addHeardSeenRespectedTemplate.ts
+++ b/packages/server/database/migrations/20221201121536-addHeardSeenRespectedTemplate.ts
@@ -87,6 +87,7 @@ const promptsInfo = [
 
 const template = makeTemplate(templateName)
 const reflectPrompts = promptsInfo.map((promptInfo) => makePrompt(promptInfo))
+
 export const up = async function (r: R) {
   try {
     await Promise.all([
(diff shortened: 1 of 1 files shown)


igorlesnenko (approved):

packages/server/database/migrations/20221201121536-addHeardSeenRespectedTemplate.ts
```diff
@@ -0,0 +1,113 @@
...
+}
+
+const makePrompt = (promptInfo: PromptInfo) => {
+  const {question, description, groupColor, templateId, sortOrder} = promptInfo
+  return {
+    createdAt,
+    description,
+    groupColor,
+    // FIXME this may create duplicated ids
+    id: nameToId(question, 'prompt'),
```
Dschoordsch, line 48: -1 we should not create the ids just based on the question as it can easily create duplicate ids which will overwrite an existing prompt which then is missing (someone even left the FIXME from where you've copied this ;-)
It's not an issue this time, but lets create it from the templateId + question to be sure and create another good example to copy from

Dschoordsch (changes requested):

Dschoordsch (approved):
Works

PR #7513, opened 2022-12-01
Fix/oauth2 slack revert

mattkrick:

# Description

Slack requires approval for changing a redirect URI. So we revert the change until slack approves it.


mattkrick:
whoops, based off master

PR #7517, opened 2022-12-02
feat(teams-limit): flag an organization when it has exceeded teams limit

### LINKED ISSUE #7379: Teams limit: Implement tier limits check function
We need to check an organization tier limits, flag the organization and notify members if the limits were exceeded

**AC**:
- Check if organization exceeded its limit on "meeting end" and other appropriate events
- The function checks the number of sticky teams in the organization 
- **Sticky team** is the team with more than 1 member that completed 3 meetings with more than 1 attendee and have had at least 1 meeting in the last 30 days
- If there are 3 or more sticky teams, the organization must be flagged (current free plan allows maximum 2 teams)
- Set `tier_limit_exceeded_at` flag as `CURRENT_DATE_TIME`
- Set `scheduled_block_at` flag as `CURRENT_DATE_TIME + 30 days`
- Check if organization fixed its limits on "remove team" event, and remove all flags in that case
- Upgrading organization should remove all flags

Also see notion document: https://www.notion.so/parabol/Implement-time-bound-stick-for-enforcing-2-team-limit-in-Usage-Stats-7b6cb3446267449f9269cbc2aa944ae8

igorlesnenko:
Fixes #7379

**How to test:**

- Create a new organization 
- Add `teamLimits` feature flag to the organization `featureFlags` field (Edit rethinkdb table or run `addFeatureFlagToOrg` mutation from #7526)
- Create 3 teams (For simplicity you can adjust numbers in `teamLimitsCheck.ts` https://github.com/ParabolInc/parabol/pull/7517/files#diff-0ebb6687f73dcbeef95db77a23ed552ac97457d45368062605a87c17330b9f9dR7-R10)
- For each team run 3 different meetings with at least 2 attendees, see that **only** after finishing a 3rd meeting for the 3rd team, the organization got flagged with `tierLimitExceededAt` and `scheduledLockAt` values
- You can check flags by calling `r.db('actionDevelopment').table('Organization').get('PUT_ORG_ID_HERE')`
- Create a new team, remove that new team, see flags were **not** removed
- Remove one sticky team (3 meetings, 2 attendees) check that all flags were removed
- Bring back flags (you can edit the database)
- Upgrade organization to PRO, check that all flags were removed from the organization
- Create a new team instead of removed, run 3 meetings with 2 attendee, see that no flags were set on PRO org.


packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -0,0 +1,107 @@
...
+  if (!hasFlags(organization)) {
+    return
+  }
+
+  if (!(await isLimitExceeded(orgId, dataLoader))) {
+    await removeFlags(orgId)
+  }
+}
+
+export async function checkTeamsLimitMaybeFlagOrganization(
```
nickoferrall, line 79: +1 I think `checkTeamsLimit` would suffice

nickoferrall (changes requested):
Looking good! Comments are mostly just related to naming

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -0,0 +1,107 @@
...
+import {RDatum, RValue} from '../../database/stricterR'
+import Organization from '../../database/types/Organization'
+import {DataLoaderWorker} from '../../graphql/graphql'
+
+const STICKY_TEAM_MIN_MEETING_ATTENDEES = 2
+const STICKY_TEAM_MIN_MEETINGS = 3
+const PERSONAL_TIER_MAX_TEAMS = 2
+const PERSONAL_TIER_LOCK_AFTER_DAYS = 30
+
+function hasFlags(organization: Organization) {
```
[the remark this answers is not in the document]
igorlesnenko: Maybe `isOrganizationRestricted` ?

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -0,0 +1,107 @@
...
+  return organization.tierLimitExceededAt || organization.scheduledLockAt || organization.lockedAt
```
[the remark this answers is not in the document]
igorlesnenko: @nickoferrall Right now it is not possible, however it is possible in theory. Conceptually organization could be locked without scheduling it. Do you think we need to remove `lockedAt` from the check?

packages/server/billing/helpers/teamLimitsCheck.ts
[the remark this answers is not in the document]
nickoferrall: I like it!

packages/server/billing/helpers/teamLimitsCheck.ts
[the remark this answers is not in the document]
nickoferrall: If you think it's fairly likely that in the future `lockedAt` can be non-null while `scheduledLockAt` is null, I'd keep it the way it is. But if that probably won't happen, I think we should remove it

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -0,0 +1,102 @@
...
+        .group('teamId')
+        .ungroup()
+        .filter((row) => row('reduction').count().ge(STICKY_TEAM_MIN_MEETINGS))
+        .count()
+        .gt(PERSONAL_TIER_MAX_TEAMS)
+    })
+    .run()
+}
+
+export const removeFlags = async (orgId: string) => {
```
nickoferrall, line 50: +1 I think `flags` should only refer to feature flags and logic related to locking should use locking/restricted as functions. So, I think we should change this `removeFlags` name

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -0,0 +1,102 @@
...
+    .get(orgId)
+    .update({
+      tierLimitExceededAt: null,
+      scheduledLockAt: null,
+      lockedAt: null
+    })
+    .run()
+}
+
+export const maybeRemoveLimitExceededFlags = async (
```
nickoferrall, line 62: +1 like above, I suggest that we don't include `flags` in the name. Maybe `maybeUnlockOrg` or `maybeRemoveTeamsLimit`?

nickoferrall (approved):

packages/server/billing/helpers/teamLimitsCheck.ts
igorlesnenko: updated

packages/server/billing/helpers/teamLimitsCheck.ts
igorlesnenko: make sense, updated

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  const now = new Date()
+
+  await r
+    .table('Organization')
+    .get(orgId)
+    .update({
+      tierLimitExceededAt: now,
+      scheduledLockAt: new Date(now.getTime() + ms(`${PERSONAL_TIER_LOCK_AFTER_DAYS}d`))
```
Dschoordsch, line 92: -1 Please also update `updatedAt` field

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  if (!(await isLimitExceeded(orgId, dataLoader))) {
+    await r
+      .table('Organization')
+      .get(orgId)
+      .update({
+        tierLimitExceededAt: null,
+        scheduledLockAt: null,
+        lockedAt: null
```
Dschoordsch, line 64: -1 please also update `updatedAt` field

Dschoordsch (changes requested):
Only a small thing. Works good!

Dschoordsch (approved):

PR #7520, opened 2022-12-02
feat(teams-limit): add upgrade CTA to the meeting sidebar if teams limit is exceeded

igorlesnenko:
image

### DIFF
--- a/packages/client/components/NewMeetingSidebarUpgradeBlock.tsx
+++ b/packages/client/components/NewMeetingSidebarUpgradeBlock.tsx
@@ -1,0 +1,74 @@
+import styled from '@emotion/styled'
+import React from 'react'
+import {useHistory} from 'react-router'
+import {PALETTE} from '../styles/paletteV3'
+import FlatPrimaryButton from './FlatPrimaryButton'
+
+const RootBlock = styled('div')({
+  padding: '16px 8px'
+})
+
+const Container = styled('div')({
+  background: PALETTE.SLATE_200,
+  border: `2px solid ${PALETTE.GRAPE_500}`,
+  padding: 12,
+  borderRadius: 4
+})
+
+const Heading = styled('div')({
+  color: PALETTE.SLATE_700,
+  fontSize: '14px',
+  lineHeight: '16px',
+  fontWeight: 600
+})
+
+const Description = styled('div')({
+  color: PALETTE.SLATE_900,
+  fontSize: '12px',
+  lineHeight: '16px',
+  marginTop: 8
+})
+
+const Action = styled('div')({
+  marginTop: 8
+})
+
+const UpgradeButton = styled(FlatPrimaryButton)({
+  width: '100%'
+})
+
+interface Props {
+  onClick?: () => void
+  orgId: string
+}
+
+const NewMeetingSidebarUpgradeBlock = (props: Props) => {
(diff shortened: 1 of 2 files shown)


packages/client/components/NewMeetingSidebarUpgradeBlock.tsx
```diff
@@ -0,0 +1,74 @@
...
+const NewMeetingSidebarUpgradeBlock = (props: Props) => {
+  const {onClick, orgId} = props
+  const history = useHistory()
+  return (
+    <RootBlock>
+      <Container>
+        <Heading>🎉 We’re glad you love Parabol!</Heading>
+
+        <Description>
+          To make sure you don’t lose access, make sure to upgrade to the Team plan so you can have
```
nickoferrall, line 54: +1 It might be helpful to explicitly state why they might lose access. While the end of the sentence explains that a benefit of upgrading is getting "as many teams as you need", it's not super clear to me why the user might lose access. Perhaps: "You've exceeded the two-team limit. To make sure you don't lose access, upgrade to the Team plan so you can have as many teams as you need."

I realise that this sentence was included in the designs, so feel free to ignore. Curious to get your thoughts though

packages/client/components/NewMeetingSidebarUpgradeBlock.tsx
```diff
@@ -0,0 +1,74 @@
...
+          as many teams as you need.
+        </Description>
+
+        <Action>
+          <UpgradeButton
+            onClick={() => {
```
nickoferrall, line 60: +1 teeny nit: as this function is doing two things, I'd lean towards creating a `handleClick` function and using it here to make the JSX more readable

packages/client/components/NewMeetingSidebarUpgradeBlock.tsx
igorlesnenko: Make sense. I updated the copy. 
cc @acressall 

image

packages/client/components/NewMeetingSidebarUpgradeBlock.tsx
```diff
@@ -0,0 +1,74 @@
...
+  onClick?: () => void
+  orgId: string
+}
+
+const NewMeetingSidebarUpgradeBlock = (props: Props) => {
+  const {onClick, orgId} = props
+  const history = useHistory()
+
+  const handleClick = () => {
+    onClick && onClick()
```
Dschoordsch, line 50: +1 you could shorten this to `onClick?.()`

Dschoordsch (changes requested):
Code looks good. Only on tiny screens (horizontal phone) the meeting stages can't be reached at all anymore. The current scrollable area only includes the stages, but I think also the facilitator needs to be in the same area to make it work.
Screenshot from 2022-12-08 18-20-11
Screenshot from 2022-12-08 18-18-16

igorlesnenko:
@Dschoordsch if I made scrollable area bigger, it will work, but it will be tiny anyway. Maybe add close button on mobile only? 
image

igorlesnenko:
@Dschoordsch added close button here is the demo
https://www.loom.com/share/404cb9784b0a4eb0a8898e462297643b

Dschoordsch (approved):

PR #7521, opened 2022-12-03
fix: upgrade daypicker

mattkrick:

# Description

Daypicker icons weren't showing up for the last month since we removed the material icons font.
There was no way to inject an SVG inside the component without upgrading daypicker to v8.
After upgrading, the overall style changed, but tbh I like it more than before, so I just kept it & changed the colors to match our palette.

I moved the style sheet to the Action component so we only have to worry about it once.

## Demo

Works on tasks & in meetings

Screenshot from 2022-12-02 16-41-13
Screenshot from 2022-12-02 16-40-42


enriquesanchez (approved):
This looks so much better than what we currently have 👍 

I want to suggest two minor tweaks, but these are not critical and I don't think they should stop this fix from going out:

1. The selected date number is not correctly centered inside the highlight circle
2. This also happens in the current picker, but the title above the month and year (Change Due Date) should have a little more top-margin so there's more breathing room and is vertically centered.

mattkrick:
OMG you're so fast! the un-centered circle bothers me, it _might_ just be on my machien though. linux has a bug where things get off-centered by 1 pixel

the padding i can fix! gave it an extra 8px on the top

PR #7523, opened 2022-12-05
fix(metrics): Always include email in traits for identify calls

tianrunhe:
# Description

While investigating Segment <-> HubSpot integration, I saw this error pop out from Segment that complains `email` is missing from identify calls:

Screenshot 2022-12-05 at 10 03 46 AM

## Demo

Screenshot 2022-12-05 at 2 07 12 PM

## Testing scenarios
- [ ] Make sure `email` is included in `traits` for all `identify` calls

## Final checklist

### DIFF
--- a/packages/server/billing/helpers/adjustUserCount.ts
+++ b/packages/server/billing/helpers/adjustUserCount.ts
@@ -8,6 +8,8 @@ import insertOrgUserAudit from '../../postgres/helpers/insertOrgUserAudit'
 import {OrganizationUserAuditEventTypeEnum} from '../../postgres/queries/generated/insertOrgUserAuditQuery'
 import {getUserById} from '../../postgres/queries/getUsersByIds'
 import updateUser from '../../postgres/queries/updateUser'
+import IUser from '../../postgres/types/IUser'
+import {analytics} from '../../utils/analytics/analytics'
 import {toEpochSeconds} from '../../utils/epochTime'
 import getActiveDomainForOrgId from '../../utils/getActiveDomainForOrgId'
 import getDomainFromEmail from '../../utils/getDomainFromEmail'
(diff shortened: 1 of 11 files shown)


Dschoordsch:
Sorry I haven't looked into it yet. Will review it later today

packages/server/billing/helpers/adjustUserCount.ts
```diff
@@ -150,8 +146,12 @@ export default async function adjustUserCount(
   const r = await getRethink()
   const orgIds = Array.isArray(orgInput) ? orgInput : [orgInput]
 
+  const user = await getUserById(userId)
+  if (!user) {
+    throw new Error(`User does not exist: ${userId}`)
+  }
   const dbAction = dbActionTypeLookup[type]
-  await dbAction(orgIds, userId)
+  await dbAction(orgIds, user)
```
Dschoordsch, line 154: +1 I would prefer if we'd pass just `userId, email` to make clear that's the only thing we're using from the object. No big deal though.

packages/server/billing/helpers/adjustUserCount.ts
```diff
@@ -150,8 +146,12 @@ export default async function adjustUserCount(
   const r = await getRethink()
   const orgIds = Array.isArray(orgInput) ? orgInput : [orgInput]
 
+  const user = await getUserById(userId)
```
Dschoordsch, line 149: 11 I think we can safely use the dataloader here. While we might mutate the `OrganizationUser` we don't touch the `User` object here or in the mutations, so there is no risk of reading stale data.

Dschoordsch (changes requested):
Something seems amiss with `adjustUserCount`. If I remove a member from my org, I see a segment event for their new identity but none for the remove event or `adjustUserCount`. Do you know what's happening there?

packages/server/billing/helpers/adjustUserCount.ts
tianrunhe: I have other refactor coming up that needs the whole user object. You can check it out later!

tianrunhe:
> I see a segment event for their new identity but none for the remove event or adjustUserCount

The `removeFromOrg` triggers db action type of `REMOVE_USER` which goes to `deleteUser` and we didn't have any metrics emitted there. I added a new one.

Dschoordsch (changes requested):
You missed adding the dataloader to `adjustUserCount` in 2 places
https://github.com/ParabolInc/parabol/blob/c7f68af4813448e4ec8ccbdcaa45a9e999260cff/packages/server/safeMutations/acceptTeamInvitation.ts#L96
https://github.com/ParabolInc/parabol/blob/c7f68af4813448e4ec8ccbdcaa45a9e999260cff/packages/server/graphql/private/mutations/autopauseUsers.ts#L29
No idea why this isn't caught by `yarn typecheck`

Dschoordsch (approved):

PR #7526, opened 2022-12-05
feat(teams-limit): add addFeatureFlagToOrg mutation and support teams limit feature flag

### LINKED ISSUE #7519: Teams limit: add an ability to add a feature flag to the whole organization
We need to have a feature flag on the whole organization, not a single user, as the warnings-blocking flow would be triggered for the whole org.

AC:
- [ ] Support adding a feature flag to organization in add feature flag mutation
- [ ] Remove teamLimits feature flag from user type
- [ ] add teamLimits feature flag to organization type

igorlesnenko:
Fixes #7519

- Added a new mutation `addFeatureFlagToOrg` that allows to add a feature flag to the `Organization` type
- Removed `teamsLimit` feature flag from the `User` type and added it to the `Organization` type

## Testing scenarios

- Go to http://localhost:3000/admin/graphql
- Select private schema
- Run the following mutation putting your org ids

```
mutation {
  addFeatureFlagToOrg(orgIds: ["ivQxFEZ7tC", "ijVADO81eo"], flag: teamsLimit) {
  	...on ErrorPayload {
      error {
        title
        message
      }
    }
    ...on AddFeatureFlagToOrgSuccess {
      updatedOrganizations {
        id
        featureFlags {
          teamsLimit
        }
      }
  	}
  }
}
```

- See that result returned correctly


packages/server/graphql/public/mutations/addFeatureFlagToOrg.ts
```diff
@@ -0,0 +1,34 @@
+import getRethink from '../../../database/rethinkDriver'
+import {getUserId, isSuperUser} from '../../../utils/authorization'
+import standardError from '../../../utils/standardError'
+import {MutationResolvers} from '../resolverTypes'
+
+const addFeatureFlagToOrg: MutationResolvers['addFeatureFlagToOrg'] = async (
```
nickoferrall, line 6: +1 now that we have this mutation, I think we should rename the other one `addUserFeatureFlag` or `addFeatureFlagToUser` to clarify the difference

packages/server/graphql/public/mutations/addFeatureFlagToOrg.ts
```diff
@@ -0,0 +1,34 @@
...
+  if (!isSuperUser(authToken)) {
+    return standardError(new Error('Not authorised to add feature flag'), {
+      userId: viewerId
+    })
+  }
+
+  // RESOLUTION
+  await r
+    .table('Organization')
+    .getAll(r.args(orgIds), {index: 'id'})
```
nickoferrall, line 25: +1 we can just do `.getAll(orgIds)` here. Here's the ChatGPT explanation 😉 :
chatgpt-query

packages/server/graphql/public/mutations/addFeatureFlagToOrg.ts
```diff
@@ -0,0 +1,34 @@
...
+    .update({
+      featureFlags: r.row('featureFlags').default([]).setInsert(flag)
+    })
+    .run()
+
+  return {orgIds}
```
nickoferrall, line 31: -1 I think we should return the updated `orgIds` from the query above rather than returning the unchanged `orgIds` input. Outputting the input doesn't tell us much, but returning the updated orgIds tells us which orgs now have the feature flag

packages/server/graphql/public/types/AddFeatureFlagToOrgPayload.ts
```diff
@@ -0,0 +1,15 @@
+import isValid from '../../isValid'
+import {AddFeatureFlagToOrgPayloadResolvers} from '../resolverTypes'
+
+export type AddFeatureFlagToOrgPayloadSource = {orgIds: string[]} | {error: {message: string}}
+
+const AddFeatureFlagToOrgPayload: AddFeatureFlagToOrgPayloadResolvers = {
+  organizations: async (source, _args, {dataLoader}) => {
```
nickoferrall, line 7: +1 see comment above. This could become `updatedOrganizations` if we return the updated `orgIds` from the mutation

packages/server/graphql/public/types/Organization.ts
```diff
@@ -14,6 +14,9 @@ const Organization: OrganizationResolvers = {
   company: async ({activeDomain}, _args, {authToken}) => {
     if (!activeDomain || !isSuperUser(authToken)) return null
     return {id: activeDomain}
+  },
+  featureFlags: ({featureFlags}) => {
```
nickoferrall, line 18: Why do we need to add this? What does it do?

packages/server/graphql/public/typeDefs/addFeatureFlagToOrg.graphql
```diff
@@ -0,0 +1,41 @@
...
+
+    """
+    the flag that you want to give to the organization
+    """
+    flag: OrganizationFlagEnum!
+  ): AddFeatureFlagToOrgPayload!
+}
+
+type AddFeatureFlagToOrgPayload {
+  error: StandardMutationError
```
nickoferrall, line 33: +1 we have a newer pattern where we return a success object and a union or an error payload. That way, we know that if it's not an error payload, `organizations` definitely exist. See example [here](https://github.com/ParabolInc/parabol/blob/13cabc3c6bd6b2082c2d7e37a46695a1498f6221/packages/server/graphql/private/typeDefs/changeEmailDomain.graphql)

nickoferrall (changes requested):

packages/server/graphql/public/types/Organization.ts
igorlesnenko: it exposes `featureFlags` on organization. So we can query

```
 organization {
  featureFlags {
    teamsLimit
  }
}
```

packages/server/graphql/public/mutations/addFeatureFlagToOrg.ts
igorlesnenko: @nickoferrall When I do the update operation I can specify fields to return. In this particular case I would see the resulting `teamsLimit` value for each org, I can also query `updatedAt` field if needed.

If I return only actually updated organizations, in some cases there may be some orgs missing in the answer and I will not know why, either I made a mistake in the id, or nothing has changed.

What do you think?

packages/server/graphql/public/mutations/addFeatureFlagToOrg.ts
igorlesnenko: @nickoferrall In case of renaming I would need to update various documentation, and people should update their saved queries. Is it ok to introduce that breaking change?

packages/server/graphql/public/mutations/addFeatureFlagToOrg.ts
nickoferrall: I think it could be confusing if the Parabol employee runs the query with several orgIds, one of them fails to update, but they see that organization in the results without any error message. If they don't query `teamsLimit` and just query the org `name`, for example, they would assume the org had updated.

Also, even if the Parabol employee does query the `teamsLimit`, but they're updating 100 org ids, it would be a pain reading through all of the responses to see if one failed.

Perhaps we could return `updatedOrgIds` and `notUpdatedOrgIds`? Like [this example](https://github.com/ParabolInc/parabol/blob/13cabc3c6bd6b2082c2d7e37a46695a1498f6221/packages/server/graphql/private/types/ChangeEmailDomainSuccess.ts)

nickoferrall (changes requested):

packages/server/graphql/public/types/Organization.ts
igorlesnenko: @nickoferrall I just followed same pattern we use for the User, I saw we moved featureFlags from just types to public types: https://github.com/ParabolInc/parabol/pull/6664/files#diff-020fe869f7d8e452f6fc5eee264fd6082069b29a2c13cb182230a972e7da8872L84-L94

I guess it is just a new pattern to define types? I specify it in `public/typeDefs` then write resolver like this.

packages/server/graphql/public/mutations/addFeatureFlagToOrg.ts
igorlesnenko: Make sense, made `updatedOrganizations` instead of `organizations`

nickoferrall (approved):

Dschoordsch (changes requested):
Running the mutation with one valid and one invalid id
```graphql
mutation AddFeatureFlagToOrg {
  addFeatureFlagToOrg(orgIds: ["georg@parabol.co","e3SPYTYjg4"], flag:teamsLimit) {
    ...on ErrorPayload {
      error {
        title
        message
      }
    }
    ...on AddFeatureFlagToOrgSuccess {
      updatedOrganizations {
        id
      }
    }
  }
}
```
gives unexpected output
```
{
  "data": {
    "addFeatureFlagToOrg": {
      "updatedOrganizations": []
    }
  }
}
```
The valid org already had the feature flag. As a user I cannot distinguish whether an `orgId` was invalid or the organization already had the feature flag.

igorlesnenko:
@Dschoordsch added validation

Dschoordsch (approved):

Issue #7488, opened 2022-11-23
AI Summary: implement discussion summary in the meeting summary


nickoferrall:
In the retrospective meeting summary, include a GPT-3 generated ~2 sentence summary of all of the comments in the discussion thread of the topic.

### AC

- [ ] Add a `discussionSummary` field (this is a summary of the comments in the topic) to the `Discuss` phase GraphQL object and db table
- [ ] When the meeting ends, use GPT-3 to generate the `discussionSummary`
- [ ] Include the `discussionSummary` in the meeting summary, following the designs in https://github.com/ParabolInc/parabol/issues/7468
- [ ] Make the GPT-3 API keys an optional environment variable. If the key doesn't exist, don't show the discussion summary. This way, if Parabol is running on an air-gapped network, we will not reach out to the GPT-3 API

nickoferrall:
@jordanh issue description updated 👍

PR #7534, opened 2022-12-06
feat(ai-summary): discussion summary in meeting summary

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7488

Demo: https://www.loom.com/share/9dabdb277aca410bacf66fa85dcd5666

### To test

- [ ] Make sure you've added both the `OPEN_AI_API_KEY` and `OPEN_AI_ORG_ID` env vars 
- [ ] Add the `aiSummary` feature flag to the user
- [ ] Create a retro, add a bunch of comments and tasks in the discussion threads
- [ ] When you end the meeting, you should see a `Discussion Summary` in the meeting summary


packages/server/graphql/mutations/endRetrospective.ts
```diff
@@ -93,6 +94,7 @@ export default {
       return standardError(new Error('Cannot find facilitator stage'), {userId: viewerId})
     }
     const {stage} = currentStageRes
+    await handleCompletedStage(stage, meeting, dataLoader)
```
nickoferrall, line 97: When we end the meeting, `handleCompletedStage` isn't called from `navigateMeeting` so the `discussionSummary` isn't generated. To fix, I've added this here in `endRetrospective`

nickoferrall:
> Looks good! I wonder should we show some kind of spinner or loading indicator if summary is not ready yet?

Yeah, I think you're right! I've created an issue as I don't think it'll be straightforward: https://github.com/ParabolInc/parabol/issues/7537

Dschoordsch (approved):
I think it's good for now. I would like to update the summary once the API requests finished. Maybe we should add these to the meeting subscription and just send them once ready?

I haven't checked the email, but that would also miss these fields, right?

nickoferrall:
>  I would like to update the summary once the API requests finished. Maybe we should add these to the meeting subscription and just send them once ready?

Agreed! This issue should resolve that: https://github.com/ParabolInc/parabol/issues/7537

> I haven't checked the email, but that would also miss these fields, right?

In a later PR (https://github.com/ParabolInc/parabol/pull/7543), I only send the email once the whole summary has been generated which means the discussion summary should also have been generated by then

Issue #7498, opened 2022-11-29
Recurrence: Add information when the next meeting is going to start to the ended meeting view


BartoszJarocki:
Design: https://github.com/ParabolInc/parabol/issues/7420

AC:
- if the meeting has already ended but recurrence is still active information when the next meeting is going to start is visible

PR #7542, opened 2022-12-08
feat(recurrence): Added current meeting link to ended meeting badge in standups

### LINKED ISSUE #7500: Recurrence: Show current activity link when browsing through already ended meetings in the series
Design: https://github.com/ParabolInc/parabol/issues/7457

AC:
- [ ] when browsing through already ended meeting in the series a link to the currently active meeting is showing (if there's any)

BartoszJarocki:
# Description

Fixes #7500 #7498 

## Demo

localhost_3000_meet_iUDGs1q4tG_responses
localhost_3000_meet_iUHmzSRkuk_responses

## Testing scenarios

- [ ] check if ended meeting badge is showing a link to the current meeting, if available

This scenario requires some manual db tweaking as there's no direct way to create meeting in the meeting series
-  create two standups meetings, start recurrence in one of them
-  manually update the second meeting with `meetingSeriesId` associated with the one that has recurrence started
-  now, end one of the meetings and check the ended meeting badge, there should be a link to the current meeting

- [ ] check if next meeting start time is showing, if there are no active meetings and recurrence is active
- create standup meeting, enable recurrence, end the meeting and go back to the meeting view

## Final checklist

### DIFF
--- a/packages/client/components/TeamPrompt/TeamPromptEndedBadge.tsx
+++ b/packages/client/components/TeamPrompt/TeamPromptEndedBadge.tsx
@@ -1,7 +1,9 @@
 import styled from '@emotion/styled'
 import React from 'react'
 import {Link} from 'react-router-dom'
+import useBreakpoint from '../../hooks/useBreakpoint'
 import useRefreshInterval from '../../hooks/useRefreshInterval'
+import {Breakpoint} from '../../types/constEnums'
 import {humanReadableCountdown} from '../../utils/date/relativeDate'
 import {TeamPromptBadge} from './TeamPromptBadge'
 
@@ -21,11 +23,18 @@ interface NextMeetingCountdownProps {
 export const NextMeetingCountdown = (props: NextMeetingCountdownProps) => {
   const {nextMeetingDate} = props
   useRefreshInterval(1000)
+
   const fromNow = humanReadableCountdown(nextMeetingDate)
   if (!fromNow) return null
 
   return <span>Next one starts in {humanReadableCountdown(nextMeetingDate)}.</span>
 }
+
+const TeamPromptEndedBadgeRoot = styled('div')({
+  display: 'flex',
+  justifyContent: 'center',
+  alignItems: 'start'
+})
 
 const StyledLink = styled(Link)({
   textDecoration: 'underline',
@@ -33,13 +42,22 @@ const StyledLink = styled(Link)({
 })
 
 const EmojiContainer = styled('span')({
-  paddingRight: 4
+  paddingRight: 8
 })
+
+const TeamPromptEndedTextContainer = styled('span')<{isDesktop: boolean}>(({isDesktop}) => ({
+  display: 'inline-block',
+  width: isDesktop ? undefined : 220,
+  overflow: 'hidden',
+  overflowWrap: 'break-word'
+}))
 
 // here we just want one of the props to be present, never both
 type Props = {closestActiveMeetingId: string} | {nextMeetingDate: Date} | Record<string, never>
 
 export const TeamPromptEndedBadge = (props: Props) => {
+  const isDesktop = useBreakpoint(Breakpoint.SIDEBAR_LEFT)
+
   const renderAdditionalInfo = () => {
     if ('closestActiveMeetingId' in props) {
       return <NextMeetingLink closestActiveMeetingId={props.closestActiveMeetingId} />
@@ -54,9 +72,12 @@ export const TeamPromptEndedBadge = (props: Props) => {
 
   return (
     <TeamPromptBadge>
-      <div>
-        <EmojiContainer>✅</EmojiContainer> This activity has ended. {renderAdditionalInfo()}
-      </div>
+      <TeamPromptEndedBadgeRoot>
+        <EmojiContainer>✅</EmojiContainer>{' '}
+        <TeamPromptEndedTextContainer isDesktop={isDesktop}>
+          This activity has ended. {renderAdditionalInfo()}
+        </TeamPromptEndedTextContainer>
+      </TeamPromptEndedBadgeRoot>
     </TeamPromptBadge>
   )
 }
--- a/packages/client/components/TeamPrompt/TeamPromptEndedBadge.tsx
+++ b/packages/client/components/TeamPrompt/TeamPromptEndedBadge.tsx
@@ -1,6 +1,27 @@
+import styled from '@emotion/styled'
 import React from 'react'
+import {Link} from 'react-router-dom'
 import {TeamPromptBadge} from './TeamPromptBadge'
 
-export const TeamPromptEndedBadge = () => {
-  return <TeamPromptBadge>✅ This activity has ended.</TeamPromptBadge>
+const StyledLink = styled(Link)({
+  textDecoration: 'underline'
+})
+
+interface Props {
+  closestActiveMeetingId?: string
+}
+
+export const TeamPromptEndedBadge = ({closestActiveMeetingId}: Props) => {
+  return (
+    <TeamPromptBadge>
+      <div>
+        ✅ This activity has ended.{' '}
+        {closestActiveMeetingId && (
+          <StyledLink to={`/meet/${closestActiveMeetingId}`}>
+            Go to the currenct activity.
+          </StyledLink>
+        )}
+      </div>
+    </TeamPromptBadge>
+  )
 }
--- a/packages/client/components/TeamPrompt/TeamPromptEndedBadge.tsx
+++ b/packages/client/components/TeamPrompt/TeamPromptEndedBadge.tsx
@@ -5,7 +5,12 @@ import {humanReadableCountdown} from '../../utils/date/relativeDate'
 import {TeamPromptBadge} from './TeamPromptBadge'
 
 const StyledLink = styled(Link)({
-  textDecoration: 'underline'
+  textDecoration: 'underline',
+  fontWeight: 400
+})
+
+const EmojiContainer = styled('span')({
+  paddingRight: 4
 })
 
 // here we just want one of the props to be present, never both
@@ -30,7 +35,9 @@ export const TeamPromptEndedBadge = (props: Props) => {
 
   return (
     <TeamPromptBadge>
-      <div>✅ This activity has ended. {renderAdditionalInfo()}</div>
+      <div>
+        <EmojiContainer>✅</EmojiContainer> This activity has ended. {renderAdditionalInfo()}
+      </div>
     </TeamPromptBadge>
   )
 }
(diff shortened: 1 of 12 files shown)


BartoszJarocki:
@enriquesanchez I left the link in the same style as the first part of the text (in bold). please lmk if you'd rather make the link not bold, thank you!

packages/server/dataloader/customLoaderMakers.ts
```diff
@@ -526,6 +526,7 @@ export const activeMeetingsByMeetingSeriesId = (parent: RootDataLoader) => {
             .table('NewMeeting')
             .getAll(key, {index: 'meetingSeriesId'})
             .filter({endedAt: null}, {default: true})
+            .orderBy(r.asc('createdAt'))
```
BartoszJarocki, line 529: added this to make sure we have a stable meeting order in the response, ie. the newest meeting should be first in the list

packages/server/graphql/private/mutations/processRecurrence.ts
```diff
@@ -17,9 +18,12 @@ import {MutationResolvers} from '../resolverTypes'
...
   startTime: Date,
-  dataLoader,
-  r,
-  subOptions
+  dataLoader: DataLoaderWorker,
+  r: ParabolR,
+  subOptions: {
+    mutatorId: string | undefined
+    operationId: string
+  }
```
BartoszJarocki, line 26: this is just some code cleanup, added types to be a good TS citizen

packages/server/graphql/private/mutations/processRecurrence.ts
```diff
@@ -53,15 +57,13 @@ const startRecurringTeamPrompt = async (
 const processRecurrence: MutationResolvers['processRecurrence'] = async (
   _source,
   {},
-  {authToken, dataLoader, socketId: mutatorId}
+  {dataLoader, socketId: mutatorId}
```
BartoszJarocki, line 60: removed unused variable

packages/server/graphql/private/mutations/processRecurrence.ts
```diff
@@ -53,15 +57,13 @@ const startRecurringTeamPrompt = async (
...
 ) => {
   const r = await getRethink()
   const now = new Date()
   const operationId = dataLoader.share()
   const subOptions = {mutatorId, operationId}
 
-  // VALIDATION
-
```
BartoszJarocki, line 64: ✂️

packages/client/components/TeamPrompt/TeamPromptEndedBadge.tsx
```diff
@@ -1,6 +1,27 @@
...
+}
+
+export const TeamPromptEndedBadge = ({closestActiveMeetingId}: Props) => {
+  return (
+    <TeamPromptBadge>
+      <div>
+        ✅ This activity has ended.{' '}
+        {closestActiveMeetingId && (
+          <StyledLink to={`/meet/${closestActiveMeetingId}`}>
+            Go to the currenct activity.
```
BartoszJarocki, line 21: `Go to the current activity.` or `Go to current activity.`? 😅

BartoszJarocki:
@jmtaber129 may I ask you to take a look? thanks!

enriquesanchez:
@BartoszJarocki Thanks for the ping!

>please lmk if you'd rather make the link not bold, thank you!

Yes, let's keep the link not bold and use the default link color, which I think it's Slate 700.

May I ask if we can also add a little more space between the checkmark emoji and the text?
Captura de pantalla 2022-12-08 a la(s) 11 20 18 a m

I have a margin of 8px on the mockups.

Also, when the text inside the pill gets too long like in your screenshots, we might have to wrap it to a new line, particularly in mobile views. I'm hoping we can make it look something like this:

Captura de pantalla 2022-12-08 a la(s) 11 22 49 a m

BartoszJarocki:
@enriquesanchez thank you! maybe we should put all the additional data like `Next one starts in X` in thew line entirely? 🤔

enriquesanchez:
I don't mind how it looks if there's enough space (like on desktop), I actually prefer to keep a single line.

jmtaber129 (changes requested):
Code looks good! However, there are two bugs I'm seeing:
1. -1 When process recurrence executes, the active meetings for the ended meeting aren't updated, so it shows "N hours until next meeting" instead of "go to current meeting". This is because the active meetings are updated by the end meeting mutation, before the new meeting is created. We'll need to either change the process recurrence logic to start new meetings and _then_ end old meetings (probably bad for slack notifications), or modify the start meeting mutation to publish updates for active meetings on the meeting series.
2. -1 If I navigate directly to an ended activity (or refresh on an ended activity), click to go to the new activity, and then click back, all the cards disappear. This seems to be a bug in `useTransition` where previously-exited node have a hard time re-entering. This patch seems to fix it:
```
diff --git a/packages/client/hooks/useAnimatedCard.ts b/packages/client/hooks/useAnimatedCard.ts
index ecc1e752f..2d18b55fa 100644
--- a/packages/client/hooks/useAnimatedCard.ts
+++ b/packages/client/hooks/useAnimatedCard.ts
@@ -36,6 +36,8 @@ const useAnimatedCard = (displayIdx: number, status: TransitionStatus) => {
       // if exiting, keep it where it is, but don't let it take up space
       el.style.position = 'absolute'
       el.style.transform = `translate(${offsetLeft - 8}px,${offsetTop - 8}px)`
+    } else {
+      el.style.position = 'static'
     }
     lastIdxRef.current = displayIdx
     lastRectRef.current = {top: offsetTop, left: offsetLeft}
diff --git a/packages/client/hooks/useTransition.ts b/packages/client/hooks/useTransition.ts
index 8301a6000..033075589 100644
--- a/packages/client/hooks/useTransition.ts
+++ b/packages/client/hooks/useTransition.ts
@@ -73,7 +73,10 @@ const useTransition = <T extends {key: Key}>(children: T[]) => {
 
   return useMemo(() => {
     const currentTChildren = [] as TransitionChild<T>[]
-    const {current: prevTChildren} = previousTransitionChildrenRef
+    const {current: prevTChildrenTemp} = previousTransitionChildrenRef
+    const prevTChildren = prevTChildrenTemp.filter(
+      (prevTChild) => prevTChild.status !== TransitionStatus.EXITING
+    )
 
     let touched = false
     // add mounted nodes + update new orderings
```

BartoszJarocki:
@jmtaber129 thank you for finding that out! the solution seems to be pretty straightforward as `processRecurrence` publishes updates for both types `EndTeamPromptSuccess` and `StartTeamPromptSuccess`. there's already a team subscription for these two, so the only part missing was fetching the `meetingSeries` whenever we get the update. See the fix [here. ](https://github.com/ParabolInc/parabol/pull/7542/commits/7ba068f326c4dbea78c6cebf7fa5b39fefbaff93)

Here's the screen recording of that case (there's a delay between showing the countdown and next meeting link because I had to manually run the process recurrence mutation)
https://user-images.githubusercontent.com/1017620/207005287-12812729-d278-4624-8742-c08e5866251a.mp4

And btw, I've also added refreshing the `Next meeting starts in X` countdown automatically as it wasn't working before.

@enriquesanchez I think it should be good now, do you mind taking a look?
localhost_3000_meet_j0Zv4qz0fm_responses
localhost_3000_meet_j0Zv4qz0fm_responses(iPhone SE) (1)

packages/server/dataloader/customLoaderMakers.ts
jmtaber129: +1 This should be `r.desc` right? I.e. if `createdAt` is in ascending order, the first meeting will be the oldest.

BartoszJarocki: I probably used the wrong words here but what I had in mind was to have the closest active meeting first. we currently don't support multiple instances opened at the same time, but if we ever change that, having `asc` order here makes the most sense to me. Imagine the following scenario

The meeting series has three active meetings - the first one ending tomorrow, the second one ending the day after tomorrow, and the last one ending the day after the second one. I'd say the link to `current activity` should take me to the first one, as other meetings are in the future. That's why I used `asc` here to have the `current` one as the first elem by default. In general, there are no good or bad answers here, it's only a question how define business requirements. For now, I'd leave it as is. @jmtaber129 lmk if that makes sense

packages/client/subscriptions/TeamSubscription.ts
```diff
@@ -140,6 +140,7 @@ const TeamSubscription = (
       }
     },
     onNext: (result) => {
+      console.log('TeamSubscription onNext', result)
```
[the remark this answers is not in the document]
BartoszJarocki: 🤦‍♂️

packages/client/hooks/useTransition.ts
```diff
@@ -73,14 +73,17 @@ const useTransition = <T extends {key: Key}>(children: T[]) => {
 
   return useMemo(() => {
     const currentTChildren = [] as TransitionChild<T>[]
-    const {current: prevTChildren} = previousTransitionChildrenRef
+    const filteredPrevTChildren = previousTransitionChildrenRef.current.filter(
+      (prevTChild) => prevTChild.status !== TransitionStatus.EXITING
+    )
```
BartoszJarocki, line 78: @mattkrick do you mind taking a look at the changes in this file? we had a case where elements were not showing after back & forth navigation, more details here: https://github.com/ParabolInc/parabol/pull/7542#pullrequestreview-1210944164 

thank you!

enriquesanchez:
@BartoszJarocki LGTM! 👍

packages/client/hooks/useTransition.ts
mattkrick: i can test this later today after the ship, but just from reading the code, it looks like animations that are exiting will get set to mounted, which means the exit animations will cut short. if that's not the case, LGTM!

i know framer-motion does this really well under the hood, if it's easier to use that, I wouldn't be opposed

jmtaber129: @mattkrick The case you mentioned would specifically happen for children that are currently present. I.e. a child that _was_ exiting as a result of no longer being present, but is now once again present, would be treated as though there’s no previous child state.

The specific issue I found while debugging is as follows:
* Set X of children are mounted
* Transition to new meeting occurs, so set X transitions to exiting, and set Y transitions through entering, entered, mounted. Set X is still kept in prev children with state exiting.
* Transition to the previous meeting occurs, so set Y transitions to exiting, and set X _should_ transition back to mounted, but because they already exist on prevChildren, we keep their state, which is exiting.

Normally, the exiting child would be removed from the prev children when its transition ends, but there seems to be a react-y DOM reuse thing happening so the `onTransitionEnd` event doesn't fire, because the would-be exiting cards in the DOM are reused to render the new meeting's cards.

The solution implemented here is treating children that were in the exiting state previously as "out of the picture".

mattkrick: > set X should transition back to mounted, but because they already exist on prevChildren, we keep their state, which is exiting.

ahh, that's the part i needed to read in prose. Got it, LGTM!

mattkrick (approved):

Issue #7470, opened 2022-11-22
AI Summary: implement whole meeting summary in meeting summary


nickoferrall:
In the meeting summary, include a GPT-3 generated ~2 sentence summary of all the reflections and comments that happened during the meeting.

### AC

- [ ] Add a `summary` field to the `NewMeeting` gql object and db
- [ ] Once a meeting has ended, use GPT-3 to generate a summary of all of the reflections and comments in the meeting
- [ ] Include that summary at the top of the meeting summary following the designs in https://github.com/ParabolInc/parabol/issues/7468
- [ ] Make the GPT-3 API keys an optional environment variable. If the key doesn't exist, don't show the whole meeting summary and show the regular meeting summary instead. This way, if Parabol is running on an air-gapped network, we will not reach out to the GPT-3 API

jordanh:
Please add AC for gracefully degrading experience when Parabol is running on an air-gapped network

nickoferrall:
@jordanh issue description updated 👍

Issue #7537, opened 2022-12-07
AI Summary: update meeting summary automatically 


nickoferrall:
If a user clicks through a Discuss Stage and quickly ends the meeting, the user may see the meeting summary before we've received a response from OpenAI with the discussion summary. At the moment, the user doesn't get any feedback that this is happening, and if it does happen, they need to refresh the page to see the discussion summary. 

### AC

- [ ] Show a loading spinner in the meeting summary if a discussion summary or whole meeting summary has been requested but not yet generated
- [ ] Once we get the response, the meeting summary updates without the user needing to refresh the page
- [ ] The meeting summary that is sent by email doesn't include a loading spinner

PR #7543, opened 2022-12-08
feat(ai-summary): whole meeting summary

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7470

A problem I ran into is that we can't generate the whole meeting summary until the meeting ends. However, if we send a request to OpenAI to summarise the meeting, there can be a few-second delay while we're waiting for the response. If you end the meeting and see a loading spinner for a few seconds, it can feel like something is broken.

So, I think we should show the user the meeting summary right away, show a loading spinner for the GPT-3 generated whole summary, and then update the summary automatically once we have the response (originally Igor's suggestion [here](https://github.com/ParabolInc/parabol/pull/7534#pullrequestreview-1208711721)).

I've created an issue https://github.com/ParabolInc/parabol/issues/7537 to update the summary automatically once you get a response, but for this issue, you have to refresh the page to see the whole meeting summary.

An alternate solution could be generating a whole meeting summary every time the meeting progresses through a Discuss Stage. We would then have something we could immediately show users when they see the meeting summary. Once we get the response from OpenAI for the final Discuss Stage, we could show the updated whole meeting summary. I'm not in favour of doing this as it would require more API calls which is costly and showing a whole meeting summary that gets updated after a few seconds is a weird user experience IMO. 

Demo: https://www.loom.com/share/b8f70aba81e84042b0e3ded854738916

### To test

- [ ] Make sure that you've added the `OPEN_AI_API_KEY` and `OPEN_AI_ORG_ID` env vars
- [ ] Add the `aiSummary` feature flag to your user
- [ ] Create a retro meeting where you add several reflections and comments/tasks
- [ ] End the meeting, see the summary, refresh the page, and see the whole meeting summary at the top of the page
- [ ] Check your emails and see that the meeting summary email looks good and includes the whole meeting summary


packages/server/graphql/types/NewMeeting.ts
```diff
@@ -119,6 +119,10 @@ export const newMeetingFields = () => ({
     description: 'true if should show the org the conversion modal, else false',
     resolve: ({showConversionModal}: {showConversionModal: boolean}) => !!showConversionModal
   },
+  summary: {
```
nickoferrall, line 122: I decided to add `summary` here instead of `MeetingRetrospective` because we'll probably want to add this to other meeting types if the feature is successful

packages/server/graphql/mutations/helpers/generateWholeMeetingSummary.ts
```diff
@@ -0,0 +1,36 @@
...
+  const reflectionsContent = reflections.map((reflection) => reflection.plaintextContent)
+  const commentsContent = commentsByDiscussions
+    .filter(isValid)
+    .flatMap((commentsByDiscussion) =>
+      commentsByDiscussion.map(({plaintextContent}) => plaintextContent)
+    )
+  const tasksContent = tasksByDiscussions
+    .filter(isValid)
+    .flatMap((tasksByDiscussion) => tasksByDiscussion.map(({plaintextContent}) => plaintextContent))
+  const contentToSummarize = [...commentsContent, ...tasksContent, ...reflectionsContent]
```
igorlesnenko, line 29: Just wonder if AI can summarize better, if the content is somehow divided into groups, so he know that the comments belong to the specific topic

packages/server/graphql/mutations/helpers/generateWholeMeetingSummary.ts
```diff
@@ -0,0 +1,36 @@
+import getRethink from '../../../database/rethinkDriver'
+import OpenAIServerManager from '../../../utils/OpenAIServerManager'
+import {DataLoaderWorker} from '../../graphql'
+import isValid from '../../isValid'
+
+const generateWholeMeetingSummary = async (
```
igorlesnenko, line 6: A couple of times I got the summary cut off at the beginning
image

image

I assume this is an AI glitch?

igorlesnenko (approved):

packages/server/graphql/mutations/helpers/generateWholeMeetingSummary.ts
nickoferrall: We already have the discussion summaries (which are made up of all the comments and tasks), and we have the topic summaries (which are the summaries of the reflections), so instead of summarising all comments, tasks, and reflections, we could just summarise all of the topic and discussion summaries? 

We could see how this works in production with real meeting content and then make the change if the summaries don't seem very good?

packages/server/graphql/mutations/helpers/generateWholeMeetingSummary.ts
nickoferrall: > I assume this is an AI glitch?

I think so as we're not slicing the content. Out of interest, did this happen with realistic meeting content or a few test reflections?

Dschoordsch: Are we getting multiple messages and omitting some?

igorlesnenko: @nickoferrall these were test reflections, however the content was realistic, not a random

packages/server/graphql/mutations/helpers/generateWholeMeetingSummary.ts
igorlesnenko: Summarizing summaries is an interesting idea.

> We could see how this works in production with real meeting content and then make the change if the summaries don't seem very good?

Make sense!

packages/server/graphql/mutations/helpers/generateWholeMeetingSummary.ts
nickoferrall: > Are we getting multiple messages and omitting some?

We're returning the [top choice](https://github.com/ParabolInc/parabol/blob/d7ed4a359e2fdb4b0c9c4d46760c64db43cdb3f3/packages/server/utils/OpenAIServerManager.ts#L31) which is the way that most examples I've found online handle the response. Normally, I only see one `choice` in the response. 

A small UI improvement could be making sure that the first letter is capitalised? I've also seen examples where the summary begins with a comma, so `, The team is doing great`, so I could clean that up too

Dschoordsch (approved):

packages/server/graphql/mutations/helpers/generateWholeMeetingSummary.ts
nickoferrall: Just tried adding this, but weirdly I got worse results! All of the summaries included "second grade" after making the change, e.g. `The second-grade team did a great job and everyone should be proud of their accomplishments.`. 

I'm sure that the prompt and meeting content order (or using topic & discussion summaries instead of individual reflections, comments, and tasks) can be improved, but I'll go ahead and merge this so we can see how it behaves with real meeting content and then iterate on the prompt & text provided.

PR #7558, opened 2022-12-12
fix(standups): Update discussion preview when discussion isn't open

jmtaber129:
# Description
Currently, if a standup discussion has a comment added, the subscription updater won't update the "N replies" footer on the standup response card unless the card's discussion is currently open. Repro:
* Sign in as two Parabol users, and navigate each to the same standup meeting
* As user A, add a comment to a discussion
* Notice that as user B, the "N replies" preview+button doesn't update to reflect the added comment
* Clicking the "reply" button as user B leads to the preview updating when the discussion loads

This is because the current utils used to update the discussion only look for the discussion on the `viewer` object, even when the discussion may exist elsewhere in the store.

Check both the viewer object and the rest of the store to find the discussion.

## Testing scenarios
Perform the repro described above, and confirm that the card updates.

## Final checklist


BartoszJarocki (approved):
good find!

Issue #7383, opened 2022-11-04
Teams limit: Trigger usage stats


igorlesnenko:
When tariff limits exceeded lets enable Usage stats for billing leaders, so they can check what is going on

Previous issue:
#7379

Designs: https://www.figma.com/file/aYc5fKU09GawnsQwKTjcDv/Time-bound-stick?node-id=0%3A1&t=mTNGlHiYcWKypsHC-0

AC:
- Flagging a team as exceeding the limits should enable `Usage stats` for all billing leaders
- Enable usage stats only if `Organization` has `activeDomain`
- To enable: 
  - Set `insignts` feature flag for them
  - Set `suggestedTier: 'pro'` for `OrganizationUser` entries 
  (_With suggestedTier=pro, user is only able to see its own organizations in the statistics, which is good. Enabling 'enterprise' without human validation cause security issues, so we are not doing that_)
- Keep Usage stats enabled even if organization upgraded or fixed its limits
- In case Usage Stats enabled because of `limitExceeded`, **do not** show a persistent snackbar: `useUsageSnackNag`
- Instead show one-time closable snackbar:
<img width="358" alt="image" src="https://user-images.githubusercontent.com/466991/203119191-09e9fad1-051d-4c45-b3b5-4e56a74e92f9.png">

acressall:
@igorlesnenko here are the designs:

## Snackbar & profile nav
The snackbar can be closed by the user and will not come back. The link in the profile nav will be present for as long as usage stats are present
![Snackbar CTA - Usage stats](https://user-images.githubusercontent.com/18582225/203099966-8fa4fe61-8921-4eee-837b-5d54ac56c04c.png)

## Usage stats page
There is a notification for when the user's org is beyond their tier. This will go away if they upgrade or reduce the number of teams to within the limit
![Usage stats CTA](https://user-images.githubusercontent.com/18582225/203100229-fcd18432-80a0-46c0-947c-65d93b25e64d.png)

igorlesnenko:
Moved usage stats design changes to a separate issue #7567

PR #7547, opened 2022-12-08
feat(standups): Standup response replied-to notification

jmtaber129:
# Description
~DO NOT MERGE: Blocked on https://github.com/ParabolInc/parabol/pull/7540, and should incorporate changes from https://github.com/ParabolInc/parabol/pull/7546~. Done

Fixes https://github.com/ParabolInc/parabol/issues/7485

Adds a notification for when a user replies to a different user's standup response. The notification includes who made the comment, a link to the meeting, and an all-comment preview (i.e. no truncation for long comments). Implements both email and in-app notification.

## Demo
In-app:
Screen Shot 2022-12-08 at 5 28 00 PM

In email:
Screen Shot 2022-12-08 at 5 28 13 PM

## Testing scenarios
In a standup meeting with non-empty responses
- [ ] User B comments on User A's standup response via the reply button
- [ ] User A receives an in-app notification without refresh
- [ ] In-app notification shows correct commenter, preview and link to meeting
- [ ] Trigger a notification email
- [ ] Notification in User A's email is correct
- [ ] User A comments on User A's standup response
- [ ] No notification is triggered for User A or User B

## Final checklist

- [X] I checked the [code review guidelines](../docs/codeReview.md)
- [X] I have added [Metrics Representative](../docs/codeReview.md#metrics-representative) as reviewer(s) if my PR invovles metrics/data/analytics related changes
- [X] I have performed a self-review of my code, the same way I'd do it for any other team member
- [X] I have tested all cases I listed in the testing scenarios and I haven't found any issues or regressions
- [X] Whenever I took a non-obvious choice I added a comment explaining why I did it this way
- [X] I added the label `One Review Required` if the PR introduces only minor changes, does not contain any architectural changes or does not introduce any new patterns and I think one review is sufficient'
- [X] PR title is human readable and could be used in changelog

PR #7563, opened 2022-12-13
feat(teams-limit): enable usage stats when teams limit exceeded

igorlesnenko:
Partially solves: #7383

**How to test:**

- Have a free tier org with a couple of billing leaders
- Org should have an `activeDomain` set (set manually in db or create an org from a user with custom email domain)
- Add org feature flag (see: https://github.com/ParabolInc/parabol/pull/7526#issue-1477287954)
- Adjust numbers [here](https://github.com/ParabolInc/parabol/pull/7517/files#diff-0ebb6687f73dcbeef95db77a23ed552ac97457d45368062605a87c17330b9f9dR6-R9) for easier test
- Run a meeting on that free tier org
- If the limit exceeded, usage stats enabled for all billing leaders, and notification created for all billing leaders (see screenshot):

Screenshot 2022-12-13 at 19 34 17

Snackbar is not included in this PR.


nickoferrall:
Hey @igorlesnenko, just checking which issue this PR is solving?

igorlesnenko:
@nickoferrall this partially solves #7383

packages/client/subscriptions/NotificationSubscription.ts
```diff
@@ -91,6 +104,9 @@ const subscription = graphql`
       ...NotificationSubscription_meetingStageTimeLimitEnd @relay(mask: false)
       ...NotificationSubscription_paymentRejected @relay(mask: false)
 
+      # Generic notifications
+      ...NotificationSubscription_newNotification @relay(mask: false)
```
nickoferrall, line 108: +1 why `newNotification` instead of `notification` like the others?

nickoferrall (changes requested):
If I open on the notification menu on the non-facilitator's screen, I get the following error: 

```
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.

Check the render method of `NotificationPicker`.
```

Here's a Loom going through it: https://www.loom.com/share/4ed26da02c5f4ece84d80908da43fb51

The bug doesn't happen if both users are in the meeting. If the non-facilitator isn't in the meeting, doesn't refresh the page, and goes to the notification menu, they see that bug

jmtaber129 (commented):
Just popping in to promote some new patterns TOFU has introduced with notifications:
* The notification subscription picks up any notifications in the `NotificationPicker` fragment, so you _probably_ won't need any changes to `NotificationSubscription`. This landed [within the last few days](https://github.com/ParabolInc/parabol/pull/7540), so you'll probably need to rebase/merge master.
* We now have notification previews in our emails - if it makes sense to show this notification in emails, you should add a corresponding component to `packages/client/modules/email/components/EmailNotifications`. The emails handle unimplemented notifications gracefully so this could also be done in a follow-up PR, or even not at all.

https://github.com/ParabolInc/parabol/pull/7547/files can serve as an example of what a new notification needs/doesn't need (without snackbar).

packages/client/subscriptions/NotificationSubscription.ts
```diff
@@ -63,6 +63,19 @@ graphql`
...
+      id
+      ... on NotifyTeamsLimitExceeded {
+        organization {
+          id
+        }
+      }
+    }
+  }
+`
+
```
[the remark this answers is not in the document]
igorlesnenko: @jmtaber129 thanks! I updated the PR according to the new approach

packages/client/subscriptions/NotificationSubscription.ts
igorlesnenko: Removed this, as with the recent changes it is not needed

igorlesnenko:
@nickoferrall I updated the PR according to the new approach for generic notifications. It works for me with 2 users. 
image

Can you reproduce it now?

packages/client/components/TeamsLimitExceededNotification.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <NotificationTemplate
+      avatar={orgPicture || defaultOrgAvatar}
+      message={`Your account is on a roll! Check out "${orgName}" usage`}
+      action={<NotificationAction label={'See Usage'} onClick={onActionClick} />}
+      notification={notification}
+    />
+  )
+}
+
+export default createFragmentContainer(TeamsLimitExceededNotification, {
```
nickoferrall, line 33: +1 when creating new functions, we should use the newer `useFragment` approach rather than `createFragmentContainer`

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -1,13 +1,61 @@
...
+import publishNotification from '../../graphql/public/mutations/helpers/publishNotification'
+import getPg from '../../postgres/getPg'
+import {appendUserFeatureFlagsQuery} from '../../postgres/queries/generated/appendUserFeatureFlagsQuery'
 
 const STICKY_TEAM_MIN_MEETING_ATTENDEES = 2
 const STICKY_TEAM_MIN_MEETINGS = 3
 const PERSONAL_TIER_MAX_TEAMS = 2
 const PERSONAL_TIER_LOCK_AFTER_DAYS = 30
 
+async function getBillingLeaders(orgId: string, dataLoader: DataLoaderWorker) {
```
nickoferrall, line 16: +1 `const getBillingLeaders` instead of `function`

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -74,20 +122,21 @@ export const maybeRemoveRestrictions = async (orgId: string, dataLoader: DataLoa
...
-  }
-
-  if (organization.tierLimitExceededAt || organization.tier !== 'personal') {
-    return
-  }
-
-  if (!(await isLimitExceeded(orgId, dataLoader))) {
-    return
-  }
+  // if (!organization.featureFlags?.includes('teamsLimit')) {
```
nickoferrall, line 125: +1 can remove these lines?

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -97,4 +146,15 @@ export const checkTeamsLimit = async (orgId: string, dataLoader: DataLoaderWorke
...
+
+  const billingLeaders = await getBillingLeaders(orgId, dataLoader)
+  const billingLeadersIds = billingLeaders.map((billingLeader) => billingLeader.id)
+
+  // Enable usage stats
+  if (organization.activeDomain) {
+    await enableUsageStats(billingLeadersIds, orgId)
+
+    // Send push notification
+    await sendWebsiteNotifications(orgId, billingLeadersIds, dataLoader)
```
nickoferrall, line 158: Does the `sendWebsiteNotifications` need to wait for `enableUsageStats` to complete? If not, we can use `Promise.all` to speed things up

packages/client/components/TeamsLimitExceededNotification.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const {name: orgName, picture: orgPicture} = organization
+
+  const onActionClick = () => {
+    history.push(`/usage`)
+  }
+
+  return (
+    <NotificationTemplate
+      avatar={orgPicture || defaultOrgAvatar}
+      message={`Your account is on a roll! Check out "${orgName}" usage`}
```
nickoferrall, line 26: +1 I think it should be `Check out "${orgName}"'s usage`, as it's Parabol's usage instead of Parabol usage

nickoferrall (approved):
Works for me! I can't repro that previous bug

packages/server/billing/helpers/teamLimitsCheck.ts
igorlesnenko: I think it needs to wait, as it sends a notification to usage stats link, so it needs to be enabled first

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -97,4 +146,15 @@ export const checkTeamsLimit = async (orgId: string, dataLoader: DataLoaderWorke
...
     .run()
+
+  const billingLeaders = await getBillingLeaders(orgId, dataLoader)
+  const billingLeadersIds = billingLeaders.map((billingLeader) => billingLeader.id)
+
+  // Enable usage stats
+  if (organization.activeDomain) {
+    await enableUsageStats(billingLeadersIds, orgId)
+
+    // Send push notification
```
Dschoordsch, line 157: +1 slightly too many comments. If you're just repeating the next function call, you can omit it

Dschoordsch (approved):

Dschoordsch:
Something I saw during testing is that the usage stats showed the wrong organization:
Screenshot from 2022-12-16 16-50-29
Screenshot from 2022-12-16 16-50-24
Both billing users saw a different org. The one with georg+test1 was not even a member of said org.
I saw you've didn't change anything there, but this can still be confusing.

packages/client/components/TeamsLimitExceededNotification.tsx
```diff
@@ -0,0 +1,48 @@
...
+      }
+    `,
+    notificationRef
+  )
+  const {history} = useRouter()
+  const {organization} = notification
+  const {name: orgName, picture: orgPicture} = organization
+
+  const onActionClick = () => {
+    history.push(`/usage`)
```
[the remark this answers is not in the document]
tghanken: A `Notification Clicked` event w/ a `notificationCategory: 'usage'` property would be ideal!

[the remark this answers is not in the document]
igorlesnenko: @tghanken we have the following type for notifications

```
export type NotificationEnum =
  | 'KICKED_OUT'
  | 'MEETING_STAGE_TIME_LIMIT_END'
  | 'PAYMENT_REJECTED'
  | 'PROMOTE_TO_BILLING_LEADER'
  | 'RESPONSE_MENTIONED'
  | 'RESPONSE_REPLIED'
  | 'TASK_INVOLVES'
  | 'TEAM_ARCHIVED'
  | 'TEAM_INVITATION'
  | 'TEAMS_LIMIT_EXCEEDED'
  ```
  
  Do we need a `notificationCategory` anyway, or we can use the above type?

igorlesnenko:
@Dschoordsch this seems to be a bug — we pick all organizations from a domain and just show a biggest one. User might not be a member of that org. I think we changed this part in next PR by @nickoferrall  #7587 and do not show name on the button anymore

packages/client/components/TeamsLimitExceededNotification.tsx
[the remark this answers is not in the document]
tghanken: Using that enum is perfect!

tghanken (approved):

PR #7572, opened 2022-12-14
feat(ai-summary): auto update meeting summary

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7537

Demo: https://www.loom.com/share/6ffa144f1d2a4df9bd67ae4245a01a79

### To test

- [ ] Create a retro meeting with two users, add reflections and comments
- [ ] End the meeting and see the whole meeting summary loading feedback
- [ ] After a second or so, see that the loading feedback gets replaced by the AI-generated whole meeting summary text
- [ ] See that the final discussion summary is also included (if you add 2+ comments in the final Discuss Stage) in the meeting summary
- [ ] Look at your emails and see both the whole meeting summary and discussion summary


packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/WholeMeetingSummary.tsx
```diff
@@ -41,17 +42,59 @@ const WholeMeetingSummary = (props: Props) => {
...
-  const {summary, team} = meeting
-  if (!summary) return null
+  const {summary: wholeMeetingSummary, team, reflectionGroups, phases} = meeting
+  const discussPhase = phases.find((phase) => phase.phaseType === 'discuss')
+  const {stages} = discussPhase ?? {}
   const explainerText =
     team?.tier === 'personal' ? AIExplainer.STARTER : AIExplainer.PREMIUM_MEETING
+  const hasTopicSummary = reflectionGroups.some((group) => group.summary)
+  const hasDiscussionSummary = !!stages?.some((stage) => stage.discussion?.summary)
+  const hasOpenAISummary = hasTopicSummary || hasDiscussionSummary
```
nickoferrall, line 72: It's possible that a topic or discussion summary was generated during the meeting but we've run out of tokens to generate the whole meeting summary, but this is highly unlikely. Worst case scenario, the loading feedback would persist

igorlesnenko (approved):
Looks good!

Dschoordsch (changes requested):

packages/client/mutations/EndRetrospectiveMutation.ts
```diff
@@ -52,6 +52,23 @@ graphql`
   }
 `
 
+graphql`
+  fragment EndRetrospectiveMutation_meeting on EndRetrospectiveSuccess {
+    meeting {
+      summary
```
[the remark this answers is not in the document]
nickoferrall: Makes sense, updated!

packages/server/utils/publish.ts
```diff
@@ -1,7 +1,7 @@
 import getPubSub from './getPubSub'
 
 export interface SubOptions {
-  mutatorId?: string
+  mutatorId: string | null // if the viewer's client has received a mutation response from the server, but a subsequent update happens in the mutation, update the client by publishing a subscription. To do this, set mutatorId to null/undefined: https://github.com/ParabolInc/parabol/blob/88a801d80d0c51c38b6c9722dfa80fbca8f7bebd/packages/server/graphql/ResponseStream.ts#L26
```
Dschoordsch, line 4: +1 Sorry, I missed that it was optional already 🤦
I suggest making it optional again, adding a simpler comment like: "Passing the user id of the mutator will omit sending a message to that user" or similar. Then we can keep this change smaller.

Dschoordsch (approved):

packages/server/utils/publish.ts
Dschoordsch: And remove the `| null` again

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/WholeMeetingSummary.tsx
Dschoordsch: Would this delay the email forever as well?

Dschoordsch (approved):

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/WholeMeetingSummary.tsx
nickoferrall: No, the email would send; it would just show a loading spinner

PR #7576, opened 2022-12-15
chore: Add comments to rootSchema

Dschoordsch:
# Description

Add some comments to the `rootSchema` to not get lost next time.

## Demo

https://www.loom.com/share/387739634cde4f709882e6b36d11526f

## Testing scenarios

Build passes and app starts

## Final checklist

### DIFF
--- a/packages/server/graphql/public/rootSchema.ts
+++ b/packages/server/graphql/public/rootSchema.ts
@@ -1,3 +1,9 @@
+/*
+ * Combine type definitions and resolvers from multiple sources
+ *  - legacy GraphQLObjectType types from `../types` which are defined TypeScript first
+ *  - new SDL first types with definitions in `typeDefs` and resolvers in `types`
+ *  - GitHub and GitLab schemas
+ */
 import {addResolversToSchema, mergeSchemas} from '@graphql-tools/schema'
 import {GraphQLSchema} from 'graphql'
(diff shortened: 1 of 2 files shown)


packages/server/graphql/public/rootSchema.ts
```diff
@@ -62,17 +70,19 @@ const {schema: withGitLabSchema, gitlabRequest} = nestGitLabEndpoint({
...
   typeDefs
 })
 
 // IMPORTANT! mergeSchemas has a bug where resolvers will be overwritten by the default resolvers
 // See https://github.com/ardatan/graphql-tools/issues/4367
-const withNestedResolversSchema = addResolversToSchema({
-  schema: withNestedSchema,
+const parabolWithNestedResolversSchema = addResolversToSchema({
+  schema: legacyParabolWithNestedSchema,
   resolvers: composeResolvers(resolvers, permissions)
```
Dschoordsch, line 86: @mattkrick we're only applying permissions to SDL first resolvers. Is that intentional?

nickoferrall (approved):

packages/server/graphql/public/rootSchema.ts
mattkrick: It was, but I'm all for changing it if you prefer.
My logic was the old types use the old pattern. If we refactor the permissions into the new permissionMap then we should also refactor that resolver into the new SDL pattern, too.

Issue #7577, opened 2022-12-15
Count min and max characters correctly when using emojis


nickoferrall:
@jordanh noted in [Slack](https://parabol.slack.com/archives/C4JAUUZ9P/p1671080091525449) 🔒 that the length of emojis can be incorrectly calculated. 

![image](https://user-images.githubusercontent.com/39854876/207844813-bb4c3c70-70b8-4c40-9b17-3efb58e7c3af.png)

### AC

- [ ] When calculating the min and max length of an input, account for emojis

PR #7578, opened 2022-12-15
fix: emoji length check

nickoferrall:
Fix: https://github.com/ParabolInc/parabol/issues/7577

Demo: https://www.loom.com/share/e50282aaa8294da28070bb530ea31df3

### To test

- [ ] Paste an emoji, e.g. 🔥, into the template picker value field. See that it counts the emoji as one character instead of two
- [ ] Other min/max string length calculations continue to work as expected

### DIFF
--- a/packages/client/validation/Legitity.ts
+++ b/packages/client/validation/Legitity.ts
@@ -38,7 +38,6 @@ class Legitity {
   max(len: number, msg?: string) {
     // this.value.length gives us the count of UTF-16 units, so 🔥 has a length of 2
     // Spreading the string into an array gives us the correct string length in codepoints (characters): https://stackoverflow.com/a/54369605
-
     const value = [...this.value]
     if (!this.error && value.length > len) {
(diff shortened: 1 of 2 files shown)


igorlesnenko (changes requested):
Looks like it works on frontend but not backend, I can type 2 fire emoji, but can't save


https://user-images.githubusercontent.com/466991/208103655-5c999c18-ccba-432c-b328-e2e3cc5ff077.mov

igorlesnenko (changes requested):
I have my branch updated, but still can't save two emoji. What am I doing wrong?

https://user-images.githubusercontent.com/466991/208515687-bcaee2f8-74e1-4460-8d01-ce6c6d718ec5.mov

igorlesnenko (approved):
Some glitch on my side, code was not refreshed! Works now!

Dschoordsch (approved):

PR #7580, opened 2022-12-15
chore: Fix missing public types in private schema

Dschoordsch:
# Description

Private schema required the repetition of public schema types. Adding the public types manually avoids this repetition, but also adds all public queries and mutations to the private schema. This is not an issue though, as users who have access to the private schema are allowed access to the public one already.

## Demo

https://loom.com/share/c5b59edab6f04418826074adb51b807c

## Testing scenarios

* check the app
* run something on the private schema in GraphiQL

## Final checklist


mattkrick:
i like it! i wonder if we could get rid of all of `/private/_legacy.graphql` :thinking: i'll poke around & see if we can, it'd be great to only have 1

mattkrick (approved):

PR #7584, opened 2022-12-15
fix: bump sanitize-svg version

mattkrick:

# Description

Fixes a security hole allowing for XSS via SVG upload.
The way we use it today, you'd only be able to XSS yourself, meaning the attacker would have to get the victim to upload a nefarious profile picture.
 
## Demo

Not gonna publish too much about this until it is in prod :sweat_smile:


mattkrick:
merging because it's just a version bump on a dep

Issue #7567, opened 2022-12-14
Teams limit: display a limit exceeded warning under usage stats


igorlesnenko:
AC: 

If the limit exceeded, usage stats should display a warning:

![image](https://user-images.githubusercontent.com/466991/203120053-94a49a05-b30c-4288-a0bf-e67167f33f95.png)

PR #7511, opened 2022-12-01
chore: Limit meeting history on timeline

Dschoordsch:
# Description

Fixes #7374
Show a locked message on the timeline for meetings older than 30 days of free tiers. If there are locked meetings from multiple free teams, only for the first one the CTA is shown. All locked meetings have a different icon and link to the upgrade page.

## Demo

[If possible, please include a screenshot or gif/video, it'll make it easier for reviewers to understand the scope of the changes and how the change is supposed to work. If you're introducing something new or changing the existing patterns, please share a Loom and explain what decisions you've made and under what circumstances]

## Testing scenarios

* check the timeline with meetings older than 30 days
* must include all 4 meeting types: check-in, retro, poker, stand-up
* [ ] meetings of free tier older than 30 days show a lock and Upgrade CTA
* [ ] clicking CTA links to upgrade page and sends segment event
* [ ] clicking CTA on the message does the same

## Final checklist

- [ ] I checked the [code review guidelines](../docs/codeReview.md)
- [ ] I have added [Metrics Representative](../docs/codeReview.md#metrics-representative) as reviewer(s) if my PR invovles metrics/data/analytics related changes
- [ ] I have performed a self-review of my code, the same way I'd do it for any other team member
- [ ] I have tested all cases I listed in the testing scenarios and I haven't found any issues or regressions
- [ ] Whenever I took a non-obvious choice I added a comment explaining why I did it this way
- [ ] I added the label `One Review Required` if the PR introduces only minor changes, does not contain any architectural changes or does not introduce any new patterns and I think one review is sufficient'
- [ ] PR title is human readable and could be used in changelog

PR #7587, opened 2022-12-16
feat(teams-limit): usage stats warning

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7567

### To test

- [ ] Add the teams-limit org feature flag to an org on the personal tier
- [ ] Change the TeamsLimit flags to 1 [here](https://github.com/ParabolInc/parabol/blob/4b133d2e6b5c616f2fba3397025bba5d34bb0720/packages/client/types/constEnums.ts#L490) so it's easier to test
- [ ] Create a meeting to exceed the limit
- [ ] Open "Usage Stats" in the profile menu (see screenshot below)
- [ ] See the warning message (see screenshot below)

usage

teams-limit


packages/client/components/InsightsDomainPanel.tsx
```diff
@@ -147,10 +198,24 @@ const InsightsDomainPanel = (props: Props) => {
...
+  const toBeLockedOrg = organizations.find((org) => org.scheduledLockAt)
   const isDesktop = useBreakpoint(Breakpoint.NEW_MEETING_SELECTOR)
+  const atmosphere = useAtmosphere()
+  const {history} = useRouter()
+
+  const handleClick = () => {
+    if (!toBeLockedOrg) return
+    const {id: orgId} = toBeLockedOrg
+    SendClientSegmentEventMutation(atmosphere, 'Upgrade CTA Clicked', {
+      upgradeCTALocation: 'usageStats',
```
nickoferrall, line 213: @tghanken just checking if this is how you'd like to record the event when a user clicks the "Upgrade" button?

tghanken: For tracking Meeting History (#7511), we started using `Upgrade Intent` rather than `Upgrade CTA Clicked` to track across more areas.

packages/client/components/InsightsDomainPanel.tsx
```diff
@@ -147,10 +198,24 @@ const InsightsDomainPanel = (props: Props) => {
...
+      orgId
+    })
```
tghanken, line 215: ~~+1 Can we tweak this to match the Meeting History (#7511) pattern?~~

```suggestion
    SendClientSegmentEventMutation(atmosphere, 'Upgrade Intent', {
      source: 'usageStats',
      upgradeTier: 'pro',
      orgId
    })
```

tghanken: Actually, upon further review I think it's better that we keep your `Upgrade CTA Clicked` event.  We already have logged some of these in production, and haven't registered any `Upgrade Intent` events yet, so it's probably cleaner to unify it under `Upgrade CTA Clicked`.

However, I would like to add in the `upgradeTier` parameter, as it'll help us as we expand our plan offerings in the future.

You've also done some nice work formalizing it under analytics.ts, thanks!  I'll put together a PR to migrate the events introduced by #7511 to match this pattern.

tghanken (changes requested):
-1 Can we add `upgradeTier` to the `Upgrade CTA Clicked` event?

packages/client/components/InsightsDomainPanel.tsx
nickoferrall: Sure thing - I've added `upgradeTier` 👍

tghanken (approved):

packages/client/components/InsightsDomainPanel.tsx
```diff
@@ -147,10 +198,24 @@ const InsightsDomainPanel = (props: Props) => {
     activeOrganizationCount,
     activeTeamCount,
     activeUserCount,
-    meetingCount
+    meetingCount,
+    organizations
   } = domain
-
+  const toBeLockedOrg = organizations.find((org) => org.scheduledLockAt)
```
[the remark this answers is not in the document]
igorlesnenko: > One domain could have several orgs that are locked. I guess we should show the first org and send them to the first org's billing page when they click Upgrade?

Can we iterate through orgs and show a warning with upgrade button for every org with `scheduledLockAt`? What do you think?

I also think @Dschoordsch found an additional issue [here]:(https://github.com/ParabolInc/parabol/pull/7563#issuecomment-1355573783) 

Looks like we fetch all the organizations for the domain, even if the user does not a member of the org. 

Should we fix it here or create a separate issue?

igorlesnenko (changes requested):

nickoferrall:
@igorlesnenko thanks for sharing that comment: https://github.com/ParabolInc/parabol/pull/7563#issuecomment-1355573783.

It looks like the Insights Nudge is doing a very similar thing to this Teams Limit warning. Ideally, we'd use one or the other, but as Teams Limit is still an experiment, I think we should accommodate both.

Perhaps in the new year, we find out how often Inisghts Nudge is being used? If it's not being used often, which is what I suspect, then we could replace it with the teams limit warning. That would hopefully simplify the logic and could resolve [this issue](https://github.com/ParabolInc/parabol/pull/7601#pullrequestreview-1226348606) too.

> Can we iterate through orgs and show a warning with upgrade button for every org with scheduledLockAt?

I think we should just show one CTA button. It would look confusing to have several CTAs (also, I imagine there aren't many domains that have 2+ orgs that have exceeded the limit). It looks like Matt was thinking the same thing when he wrote `InsightsNudge` as he grabbed the biggest org rather than showing many orgs ([see here](https://github.com/ParabolInc/parabol/blob/13462ed9ec92f97258eb2f12c368750718f78f38/packages/client/components/InsightsDomainNudge.tsx#L91)).

> Looks like we fetch all the organizations for the domain, even if the user does not a member of the org.

It looks like this is intentional. @Dschoordsch is a super user so all orgs were returned [here](https://github.com/ParabolInc/parabol/blob/13462ed9ec92f97258eb2f12c368750718f78f38/packages/server/graphql/public/types/Company.ts#L246). 

I've added a `viewerOrganizations` field to `Company` so that even super users only see the orgs the viewer belongs to.

Dschoordsch (approved):
Looks good and works. I had to refresh though because it seems there is no subscription update for `checkTeamsLimit`

Issue #7387, opened 2022-11-04
Teams limit: If organization is blocked, prevent new teams creation


igorlesnenko:
If organization was blocked prevent new team creation 

AC: 
- Check if `lockedAt` is set on organization
- When blocked organization selected in dropdown, disable other form fields 
- Show the warning in UI that organization is blocked, they need to remove some teams or upgrade

PR #7591, opened 2022-12-16
feat(teams-limit): prevent new team creation

nickoferrall:
Fix: https://github.com/ParabolInc/parabol/issues/7387

Demo: https://www.loom.com/share/6d8efbc8a9fa41d2a6974c0399907c81

locked-team

### To test

- [ ] Add the `lockedAt` field to your org with a rethinkdb query
- [ ] Go to create a new team
- [ ] Select the locked org and see the warning message 
- [ ] The new team input field and Create Team button are disabled
- [ ] Clicking the Upgrade link takes you to the org billing page


packages/client/modules/newTeam/components/NewTeamForm/NewTeamForm.tsx
```diff
@@ -141,6 +185,15 @@ const NewTeamForm = (props: Props) => {
     }
   }
 
+  const goToBilling = () => {
+    SendClientSegmentEventMutation(atmosphere, 'Upgrade CTA Clicked', {
+      upgradeCTALocation: 'createTeam',
```
nickoferrall, line 190: Hey @tghanken, does this segment event look OK to you?

tghanken (approved):

packages/server/graphql/mutations/addTeam.ts
```diff
@@ -62,9 +64,11 @@ export default {
...
         return standardError(new Error('Failed input validation'), {userId: viewerId})
       }
-      if (orgTeams.length >= Threshold.MAX_FREE_TEAMS) {
-        const organization = await dataLoader.get('organizations').load(orgId)
-        const {tier}: {tier: TierEnum} = organization
+      const {tier, featureFlags} = organization
+      const maxTeams = featureFlags?.includes('teamsLimit')
+        ? Threshold.MAX_PERSONAL_TIER_TEAMS
+        : Threshold.MAX_FREE_TEAMS
+      if (orgTeams.length >= maxTeams) {
```
igorlesnenko, line 71: I don't think this is a correct validation, because for the limit we count only **sticky teams** and restrict account only after it was flagged. 
i.e. `orgTeams.length >= maxTeams` could be true, but the limit might not be exceeded

I think we just need to check here for `limitExceededAt` flag

Checking for feature flag becomes redundant in this case

igorlesnenko (changes requested):

Dschoordsch (approved):
Works as expected. Only nitpick is that the disabled "Create Team" button still looks quite vibrant and thus does not appear disabled visually.

Dschoordsch (changes requested):
Missed that I had a -1 comment

nickoferrall:
> the disabled "Create Team" button still looks quite vibrant and thus does not appear disabled visually

Interesting - it is still colourful, but when comparing it to the primary action button "Add Meeting", which is visible on the same page, I think there's a good contrast. I'm using the disabled primary action button design created by Terry [here](https://github.com/ParabolInc/parabol/blob/9177909d9b3bc9cf2ba5d88a32c0ce53b1da177f/packages/client/components/PrimaryButton.tsx#L21), so I'd rather keep it if you don't feel strongly

Dschoordsch:
> > the disabled "Create Team" button still looks quite vibrant and thus does not appear disabled visually
> 
> Interesting - it is still colourful, but when comparing it to the primary action button "Add Meeting", which is visible on the same page, I think there's a good contrast. I'm using the disabled primary action button design created by Terry [here](https://github.com/ParabolInc/parabol/blob/9177909d9b3bc9cf2ba5d88a32c0ce53b1da177f/packages/client/components/PrimaryButton.tsx#L21), so I'd rather keep it if you don't feel strongly

It's fine, no strong feelings. If I compare it to the "Add meeting" button, you're correct, but I shouldn't have to compare elements with each other to deduce their state, it should be obvious.

Dschoordsch (approved):

Issue #7486, opened 2022-11-22
Notifications: Notify mentions in discussion threads


jmtaber129:
When a user is at-mentioned in a discussion thread, they receive a notification.

Order of importance in meetings:
* Poker - discussions on tickets
* Standups - threads on responses
* Async retros - discussions on topics
* (action meetings usually happen sync, and don’t currently allow discussions to happen before the agenda stage is started, so this is less relevant)

AC:
- [ ] When a user at-mentions another user, the other user receives a notification
- [ ] The notification's message is `<User> mentioned you in a discussion in <Meeting name>`
- [ ] The notification includes a link to the meeting (minimum), or to the phase+stage where the discussion is (preferred)

PR #7596, opened 2022-12-16
feat(notifications): Discussion Mention notifications

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/7486
Sends a notification when a user is at-mentioned in a discussion thread.

Lot of brand new code for the notification, plus some additions to some GQL models, hence the large diff.

Our existing `getUsersToIgnore` logic doesn't seem to work particularly well (at least not locally), so this version will send a notification even if the recipient is in the meeting/stage of the discussion comment.

## Demo
Screen Shot 2022-12-19 at 5 39 13 PM
Screen Shot 2022-12-19 at 5 40 36 PM

## Testing scenarios
- [ ] Log in as users A and B
- [ ] Start one of each meeting type, and progress to the discussion phase of the meeting
- [ ] For each meeting type, at-mention user B as user A
- [ ] Trigger a notification email
- [ ] Confirm the email notifications look good, and confirm email links to the specific discussion stage (not just the meeting)
- [ ] Without refreshing user B's window, open the notifications menu, confirm each notification is present, and confirm links to each meeting go to the specific stage.
- [ ] As user A, at-mention user A in any discussion thread
- [ ] Confirm that no notification is triggered
- [ ] As user A, at-mention user B on user B's standup response
- [ ] Confirm that user B receives a notification that someone replied to their standup, but not that someone at-mentioned them in the discussion.

## Final checklist


packages/server/graphql/types/GenericMeetingStage.ts
```diff
@@ -2,15 +2,7 @@ import {GraphQLObjectType} from 'graphql'
...
 import NewMeetingStage, {newMeetingStageFields} from './NewMeetingStage'
 
-const phaseTypes = [
-  'reflect',
-  'group',
-  'vote',
-  'firstcall',
-  'lastcall',
-  'SCOPE',
-  'RESPONSES'
```
jmtaber129, line 12: This is removing `'RESPONSES'`, since that type has its own GQL phase type.

packages/server/graphql/types/Discussion.ts
```diff
@@ -42,6 +49,54 @@ const Discussion = new GraphQLObjectType<any, GQLContext>({
...
+      description: 'The stage that the discussion is located',
+      resolve: async (
+        {discussionTopicId, discussionTopicType, meetingId},
+        _args: unknown,
+        {dataLoader}
+      ) => {
+        const meeting = await dataLoader.get('newMeetings').load(meetingId)
+        const {phases, teamId} = meeting
+        switch (discussionTopicType) {
+          case 'agendaItem': {
```
jmtaber129, line 63: I took a stab at reducing duplication between these cases, but gave up after failing to wrangle the types and templating :/

packages/client/modules/email/components/EmailNotifications/EmailDiscussionMentioned.tsx
```diff
@@ -0,0 +1,120 @@
...
+import {useFragment} from 'react-relay'
+import {cardShadow} from '../../../../styles/elevation'
+import {PALETTE} from '../../../../styles/paletteV3'
+import {FONT_FAMILY} from '../../../../styles/typographyV2'
+import makeAppURL from '../../../../utils/makeAppURL'
+import fromStageIdToUrl from '../../../../utils/meetings/fromStageIdToUrl'
+import {notificationSummaryUrlParams} from '../NotificationSummaryEmail'
+import EmailNotificationTemplate from './EmailNotificationTemplate'
+
+const editorStyles = {
```
BartoszJarocki, line 15: +1 you can use `React.CSSProperties` to get typesafety

packages/server/graphql/mutations/addComment.ts
```diff
@@ -89,6 +91,44 @@ const addComment = {
...
+          const {userId: responseUserId} = TeamMemberId.split(discussion.discussionTopicId)
+          if (responseUserId === mentionedUserId) {
+            // The mentioned user will already receive a 'RESPONSE_REPLIED' notification for this
+            // comment
+            return false
+          }
+        }
+
+        // :TODO: (jmtaber129): Consider limiting these to when the mentionee is *not* on the
+        // relevant page.
```
BartoszJarocki, line 111: +1 it'd be good to create a follow up task for that

packages/server/graphql/mutations/addComment.ts
```diff
@@ -89,6 +91,44 @@ const addComment = {
       }
     }
 
+    const {entityMap} = JSON.parse(content)
```
BartoszJarocki, line 94: -1 this technically can throw, so I'd wrap it with a try-catch, maybe extract a function to publish notifications as this resolver is getting quite big?

packages/server/graphql/public/types/NotifyDiscussionMentioned.ts
```diff
@@ -0,0 +1,21 @@
...
+  __isTypeOf: ({type}) => type === 'DISCUSSION_MENTIONED',
+  meeting: async ({meetingId}, _args, {dataLoader}) => {
+    const meeting = await dataLoader.get('newMeetings').load(meetingId)
+    return meeting as MeetingTeamPrompt
+  },
+  author: ({authorId}, _args: unknown, {dataLoader}) => {
+    return dataLoader.get('users').loadNonNull(authorId)
+  },
+  comment: ({commentId}, _args: unknown, {dataLoader}) => {
+    return dataLoader.get('comments').load(commentId)
```
BartoszJarocki, line 14: +1 comment isn't nullable, should we use `loadNonNull` here as well?

packages/server/graphql/types/Discussion.ts
```diff
@@ -42,6 +49,70 @@ const Discussion = new GraphQLObjectType<any, GQLContext>({
...
+            const phase = phases.find((phase) => phase.phaseType === AGENDA_ITEMS) as
+              | AgendaItemsPhase
+              | undefined
```
BartoszJarocki, line 66: +1 if you'd like to get away without casting, you can create type predicates like
```
export const isAgendaItemsPhase = (phase: GenericMeetingPhase): phase is AgendaItemsPhase =>
  phase.phaseType === 'agendaitems'
```

and use it like 
```
const phase = phases.find(isAgendaItemsPhase)
```

and you can do all the phase types that way. we already have some predicates in `meetingTypePredicats.ts` file. you can add more there

BartoszJarocki (changes requested):
one tiny thing related to error handling, a few potential non-blocking improvements, nice work!

packages/server/graphql/public/types/NotifyDiscussionMentioned.ts
jmtaber129: The `comments` dataloader is already non-nullable - `load` returns a non-null Promise, and it doesn't have a `loadNonNull` method.

packages/server/graphql/types/Discussion.ts
jmtaber129: TIL!

packages/server/graphql/mutations/addComment.ts
jmtaber129: https://github.com/ParabolInc/parabol/issues/7627

packages/server/graphql/mutations/addComment.ts
jmtaber129: Fixed

packages/server/graphql/public/types/NotifyDiscussionMentioned.ts
mattkrick: ah yeah, this'll be fixed when we move away from rethinkdb. the rethinkdb PK loaders don't use the new pattern so it assumes everything is non-null, even when that's not true

packages/server/graphql/mutations/addComment.ts
```diff
@@ -89,6 +139,21 @@ const addComment = {
       }
     }
 
+    const notificationsToAdd = getMentionNotifications(
```
mattkrick, line 142: -1 what about updating/deleting comments? if i add/remove a mention in an update, should we diff them to make sure new mentionees get a notification?

mattkrick (changes requested):

packages/server/graphql/types/Discussion.ts
```diff
@@ -42,6 +51,62 @@ const Discussion = new GraphQLObjectType<any, GQLContext>({
       description:
         'The partial foregin key that describes the type of object that is the topic of the discussion. E.g. AgendaItem, TaskId, ReflectionGroup, GitHubIssue'
     },
+    stage: {
+      type: NewMeetingStage,
```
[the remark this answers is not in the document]
jmtaber129: Not necessarily, e.g. if the stage was removed

packages/client/utils/meetings/fromStageIdToUrl.ts
```diff
@@ -42,17 +43,16 @@ const fromStageIdToUrl = (
...
-  const {phases} = meeting
+  const {phases, id: meetingId} = meeting
   const stageRes = findStageById(phases, stageId) || findStageById(phases, fallbackStageId)
   if (!stageRes) return '/'
   const {phase, stageIdx} = stageRes
   const {phaseType} = phase
   const phaseSlug = phaseTypeToSlug[phaseType]
-  const {meetingId} = getMeetingPathParams()
-  if (!meetingId) return '/'
+  const {meetingId: maybeDemoMeetingId} = getMeetingPathParams()
```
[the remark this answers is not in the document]
jmtaber129: `getMeetingPathParams` checks the actual browser location, which doesn't exist on the server.

However, `meetingId === maybeDemoMeetingId` if it's the demo ID on the browser, so I just removed the call to `getMeetingPathParams` entirely

packages/server/graphql/mutations/addComment.ts
jmtaber129: We probably should, but the edge cases might get a bit hairy - e.g. if a user is mentioned, unmentioned, and mentioned again, we'd only want one notification. I think keeping it simple for now is the way to go, but I added a TODO to `updateCommentContent`.

mattkrick: yep, i think GitHub does a good job of managing that whenever you @-mention someone in some text. happy to merge to keep the PRs small, I'd encourage adding that feature though. adding someone in an edit & not having that turn into a notification would feel like a bug

Dschoordsch: That's how Slack does it by the way. If you @-mention in an edit, then no notification is sent.

mattkrick: whoa! TIL. makes sense though, more ephemeral thing = created faster = no notification. something longer lasting like a GH issue = longer lasting = notification.

Issue #7593, opened 2022-12-16
User can't user their custom scale in poker


drewhousman1010:
A [user ](https://app.hubspot.com/contacts/3888472/contact/33379651) 🔒 writes: 

> Subject: bug when selecting template for poker scale
> Message: Hi! When creating a new poker meeting, I am trying to select my template for the scale ([user first name]'s scale), but it switches automatically to the Estimated Effort template. How do I fix that? It's been happening for 1 week for us now.

This is a non-paying user.

PR #7600, opened 2022-12-19
fix: access custom template

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7593

Demo: https://www.loom.com/share/c0702add99d249ceb40c227d7faf5b39

### Test

- [ ] Create a new user
- [ ] Create a custom poker/retro template 
- [ ] Add the `templateLimit` feature flag
- [ ] See that they can still use their custom poker/retro template


Dschoordsch (approved):
Did not test

PR #7601, opened 2022-12-20
feat(teams-limit): Add snackbar to usage stats enabled notification

### LINKED ISSUE #7383: Teams limit: Trigger usage stats

igorlesnenko:
# Description

Some notifications can now be accompanied by a snackbar, because previously they were only displayed as a red dot on top of the bell icon. In the case of lock notifications, we need to show a snackbar to make them more visible

The functionality will also be used for 7 reminder notification

Fixes #7383

Previous PR: #7563

## Demo

Loom: https://www.loom.com/share/ae0023682c0247fcb6fa4219d4361a94

Screenshot 2022-12-20 at 16 00 30

## Testing scenarios

- Go through testing scenario from #7563
- But also see that now a snackbar appears
- Snackbar don't disappear when page is refreshed
- Snackbar disappears when you click on it or read the notification by clicking a bell icon in the header
- You should not see nagging usage stats snackbar in case usage stats were enabled with limit exceeded flag


nickoferrall (changes requested):
I haven't been able to see the snackbar message in this PR yet. At the moment, it's showing the "View your usage stats" snackbar. Here's a Loom going through the steps I've taken: https://www.loom.com/share/384f3ffd24454c7a96379362f21beda6

igorlesnenko:
@nickoferrall From your video I see that your organization is already flagged so the function that created notification + snackbar didn't run

image

Can you remove `limitExceededAt` flag and try again?

I also see that you have non-closing usage stats snackbar. It should not appear at all when I enable usage stats using my function. So it is probably a bug, if you haven't changed any usage stats related configuration manually. I'll check

nickoferrall:
Thanks @igorlesnenko, I've updated `limitExceededAt` to `null`, but I'm still seeing the same "View your usage stats" snackbar. 

I've tried removing the `insights` feature flag and ran through the testing steps in this Loom, but no luck: https://www.loom.com/share/762cafcccd0a447d9e494c4c9c2d7aa9

igorlesnenko:
@nickoferrall after running the meeting, do you see the flags are now set for the org in the database?

nickoferrall:
@igorlesnenko no they weren't set, they remain `null`

igorlesnenko:
@nickoferrall looks like the function thinks that you didn't exceed the limit. Trying to understand why.

There is a check:

```

  if (teamIds.length <= PERSONAL_TIER_MAX_TEAMS) {
    return false
  }
```

If you have `PERSONAL_TIER_MAX_TEAMS` set to `1` then you need 2 sticky teams for it to work.

Do you have 2 teams?

Can you set `PERSONAL_TIER_MAX_TEAMS` to `0` ?

packages/client/components/TopBarNotifications.tsx
```diff
@@ -43,6 +48,9 @@ const TopBarNotifications = ({queryRef}: Props) => {
   const {viewer} = data
   const notifications = viewer?.notifications || {edges: []}
   const {edges} = notifications
+  const snackbarNotifications = edges.filter(
```
nickoferrall, line 51: +2 I like this idea. It'll be easy to add further snackbar notifications in the future if we want to

nickoferrall (changes requested):
That's working for me now! I have a second team and thought that I had ran a meeting with it, but perhaps not. 

The "View your usage snackbar" still pops up first and only after I click it away do I see the snackbar added in this PR, so I think we'll need to resolve that before merging this PR?

Also, perhaps it would be useful to comment out the hardcoded values to speed up testing? E.g.

```
const STICKY_TEAM_MIN_MEETING_ATTENDEES = 2
const STICKY_TEAM_MIN_MEETINGS = 3
const PERSONAL_TIER_MAX_TEAMS = 2
const PERSONAL_TIER_LOCK_AFTER_DAYS = 30
// const STICKY_TEAM_MIN_MEETING_ATTENDEES = 1
// const STICKY_TEAM_MIN_MEETINGS = 1
// const PERSONAL_TIER_MAX_TEAMS = 0
// const PERSONAL_TIER_LOCK_AFTER_DAYS = 1
```



Also, perhaps we should

jmtaber129:
Hey @nickoferrall @igorlesnenko! Just popping in to let y'all know that I'm actually proposing a different pattern for notification toasts over in https://github.com/ParabolInc/parabol/pull/7603. 

Also, I noticed that this will emit a snackbar for any unread notifications when the top bar renders, not just newly published notifications (i.e. a notification that just occurred). Is this intentional? This is different from our existing notification patterns. Additionally, if a user signs in with a bunch of notifications implemented in `NotificationSnackbarPicker` (e.g. 10+ instances), the snackbar queue will be pretty full, and newly published notifications won't be seen unless the user dismisses all of previous unread notifications.

igorlesnenko:
@jmtaber129  

>  Is this intentional?

This is intentional.

For my case I need the following behaviour for the snackbar notification:

- Notification could be created when user is not online (e.g. "Your organization is going to be blocked in 30 days!")
- When user opens the app he should see the unread notification as a snackbar
- Notification snackbar should stay across refreshing the page

Would this be possible with your pattern?

jmtaber129:
@igorlesnenko Is there a notion doc or GH epic with additional context I can look at? I feel like I may be missing something here - looking at this PR's issue + Figma file, the snackbar CTA seems like something that would be nagged using `useSnackNag` (similar to usage stats), conditioned on whether the org's `scheduledLockAt` is non-null. As-is, this is depending on the notifications as its source of truth (which feels like a smell), so e.g. one billing leader could upgrade a team, and another billing leader would still see the persistent snackbar notification saying they'll be locked out if their notification is unread.

Regardless, this raises a different tension - IMO implementing this as an abstraction on notifications is in conflict with our existing notification abstractions. Currently, when a notification is created we 1) create + publish the notification to the notification dropdown, and 2) show a toast _if_ the user is online. The way I think about these is that the purpose of (2) is basically to say "hey, this notification _just_ happened" in a transient way, while (1) serves as a more permanent presentation of the notification. This would change (2) to serve the same purpose as (1) for some notifications.

Since this is new logic for a specific notification (which could potentially cause problems if applied to several other notifications, as stated earlier), I'd suggest using something the hook/component for this directly for this notification, rather than implementing this as an abstraction for notifications in general.

igorlesnenko:
@jmtaber129 TLDR: I refactored the code.

> Is there a notion doc or GH epic with additional context I can look at?

Here is the full context:

- https://www.notion.so/parabol/Implement-time-bound-stick-for-enforcing-2-team-limit-in-Usage-Stats-7b6cb3446267449f9269cbc2aa944ae8
- https://www.figma.com/file/aYc5fKU09GawnsQwKTjcDv/Time-bound-stick?node-id=0%3A1&t=VZ8PWu74LBVOy0qD-0
- https://github.com/ParabolInc/parabol/issues/7383

>  the snackbar CTA seems like something that would be nagged using useSnackNag (similar to usage stats)

When using `useSnackNag`, it is not possible to close that snackbar. It will appear every time when user refresh the page. 

We don't want to be that nagging, we want to allow to close the snackbar, but we want it to stay until closed.

So instead of a single nagging snackbar, we want to have a set of notifications — limit exceeded, 14 days left, 7 days left, you're blocked, etc.

This would work with your pattern. 

However, we want the billing leader to see the snackbar anyway, even if he missed it when he wasn't online. 

So when he logins to the app, he should see this notification as a snackbar like if it just arrived.

That what I'm trying to achieve with my PR. If notification are one of the specific types and "unread", show it as a snackbar when page loads.

>  one billing leader could upgrade a team, and another billing leader would still see the persistent snackbar notification saying they'll be locked out if their notification is unread.

That is true, however another billing leader would still see an email saying his organization will be locked. But I think it would be easy to understand that this is a just an "old" message.

However, If we need to handle that case, we can just remove such notifications on organization upgrade (or mark them as "read"). 

> This would change (2) to serve the same purpose as (1) for some notifications.

I would look at it in a different angle. For important notifications we suggest that a small dot near the bell icon have not much visibility, and we want a snackbar to be shown anyway until closed, even if the user was offline. Or different angle — sometimes we just want snackbars to be delivered even if user is offline — "Hey look there was a new notification while you was offline"

---

I refactored my code and now using your approach with a separate component to make `pinned` notifications. (I would still need to improve it a bit)
What do you think?

jmtaber129:
@igorlesnenko thanks for the additional context! I think this makes sense. I do still have some minor forward-looking tensions, but they can be addressed as they come up - these changes are safe to try in my book 👍

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -25,7 +33,7 @@ const enableUsageStats = async (userIds: string[], orgId: string) => {
     .table('OrganizationUser')
     .getAll(r.args(userIds), {index: 'userId'})
     .filter({orgId})
-    .update({suggestedTier: 'pro'})
+    .update({suggestedTier: 'team'})
```
igorlesnenko, line 36: Fixed tier here, that is probably the reason I was not able to see usage stats nagging snackbar

packages/client/hooks/useUsageSnackNag.ts
```diff
@@ -60,14 +60,25 @@ const useUsageSnackNag = (insights: boolean) => {
...
       }
     `,
     {isUserNaggable}
   )
+
   const {domains} = viewer
-  const nag = domains?.find(({tier, suggestedTier}) => shouldNag(tier, suggestedTier))
+
+  // Do not use snack nag if the user has an organization with the limit exceeded, in this case we use a different warnings approach
+  const limitExceeded = domains?.find(({viewerOrganizations}) =>
```
igorlesnenko, line 76: Fixed usage stats nagging snackbar. Now I don't show it in case limit exceeded flag is set.

igorlesnenko:
@nickoferrall Fixed usage stats snackbar, now it shouldn't appear

packages/client/mutations/toasts/mapTeamsLimitExceededToToast.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  return {
+    autoDismiss: 0,
+    key: `newNotification:${notificationId}`,
+    message: `Your account is on a roll! Check out "${orgName}"'s usage`,
+    action: {
+      label: 'See Usage',
+      callback: () => {
+        history.push(`/usage`)
+        SendClientSegmentEventMutation(atmosphere, 'Clicked usage snackbar CTA')
```
tghanken, line 32: +1 I'd like to refactor this event to `Feature CTA Clicked`, with `feature` = "Usage" and `featureCtaLocation` = "Snackbar", but know that it is in use elsewhere in the codebase.

If we decide to merge this as is (which we should do unless we refactor everywhere), I'll open a new issue refactor similar events after I finalize our Segment event naming policy.

tghanken (approved):

packages/client/mutations/toasts/mapTeamsLimitExceededToToast.ts
igorlesnenko: @tghanken I suggest to create a new refactoring issue, to keep this PR small

nickoferrall (approved):

packages/client/components/PinnedSnackbarNotifications.tsx
```diff
@@ -0,0 +1,109 @@
...
+
+    if (notificationSnack.autoDismiss === 0) {
+      const callback = notificationSnack.onDismiss
+      notificationSnack.onDismiss = () => {
+        const {id: notificationId} = node
+        SetNotificationStatusMutation(
+          atmosphere,
+          {
+            notificationId,
+            status: 'CLICKED'
```
Dschoordsch, line 74: -1 I don't understand why this is necessary. It's the same event as calling the action. In fact I see both events are fired here.

packages/client/mutations/toasts/popNotificationToast.ts
```diff
@@ -50,6 +53,23 @@ export const popNotificationToastOnNext: OnNextHandler<
...
 
+  if (notificationSnack.autoDismiss === 0) {
+    const callback = notificationSnack.onDismiss
+    notificationSnack.onDismiss = () => {
+      const {id: notificationId} = addedNotification
+      SetNotificationStatusMutation(
+        atmosphere,
+        {
+          notificationId,
+          status: 'CLICKED'
```
Dschoordsch, line 64: -1 Why was this `onDismiss` event added? It's also sending the same `'CLICKED'` event as the action.

Dschoordsch (changes requested):
Tested, works except for additional mutation calls

packages/server/graphql/private/typeDefs/_legacy.graphql
```diff
@@ -1280,7 +1280,7 @@ type User {
   """
   all the notifications for a single user
   """
-  notifications(first: Int!, after: DateTime): NotificationConnection!
+  notifications(first: Int!, after: DateTime, types: [NotificationEnum!]): NotificationConnection!
```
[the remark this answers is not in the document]
igorlesnenko: Removed the whole User type

packages/client/components/PinnedSnackbarNotifications.tsx
igorlesnenko: @Dschoordsch snackbar is attached to a specific notification. We want to mark notification clicked, if snackbar is clicked. However, snackbar could also be dismissed automatically, in this case we don't want `clicked` mutation to be executed.

Refactored code a bit, and added `onManualDismiss` that should cover necessary cases.

packages/client/mutations/toasts/popNotificationToast.ts
```diff
@@ -50,24 +53,20 @@ export const popNotificationToastOnNext: OnNextHandler<
     return
   }
 
-  if (notificationSnack.action) {
```
Dschoordsch, line 53: -1 Why did you remove the action part here? Doesn't that break existing notifications?

Dschoordsch (changes requested):

packages/client/components/PinnedSnackbarNotifications.tsx
igorlesnenko: @Dschoordsch

>  but there is some precedence in the code where we set CLICKED just to set the notification inactive

That is exactly what I'm trying to do here too.

If the user manually dismiss a snackbar that is attached to a specific notification, I just mark corresponding notification as `clicked`, so it will not appear as unread and will not be highlighted in notifications dropdown.

packages/client/mutations/toasts/popNotificationToast.ts
igorlesnenko: @Dschoordsch this part is not needed anymore. The previous code marked the notification as read only if we pressed the action button. That's why we checked if button exists.

But it could be a snackbar without any button. Or user could just click on other parts of the snackbar to close it.

So now I handle these cases by marking notification as read if snackbar was manually dismissed.

This is the same logic we follow for notifications in header, if you click on it it will be marked as clicked. 

cc @jmtaber129

Dschoordsch (approved):

Issue #7559, opened 2022-12-13
Notifications: Pop toasts for new notifications


jmtaber129:
Our new notifications don't currently include toasts. Add toast-popping functionality to the generic `AddedNotification` updaters such that a dev can map a notification fragment to an `addSnackbar` call.

AC:
- [ ] Generic utils for popping toasts for notifications
- [ ] Standup mentions, replied-to's, and any other new notifications pop toasts when they fire

PR #7603, opened 2022-12-20
feat(standups/notifications): Generalized notification toasts + new toasts

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/7559
This does a couple things:
* Brings the generalized `AddedNotification` pattern into our toasts, such that a new notification simply needs to map the notification fragment to toast's `Snack` object, and add the fragment to the `popNotificationToast` fragment
  * This automatically marks the notification as `CLICKED` when the user clicks into the notification via the action button (similar to a subset of existing toasts).
* Implement the mapping for a few of the new notifications:
  * Standup response replied-to toasts
  * Standup response mentioned toasts

Once either this PR or https://github.com/ParabolInc/parabol/pull/7596 lands, we'll want to incorporate relevant changes into the not-merged PR.

## Demo
Screen Shot 2022-12-20 at 4 09 04 PM
Screen Shot 2022-12-20 at 4 08 47 PM

## Testing scenarios
- [ ] Generalized logic
  - [ ] Unimplemented toasts + toasts implemented elsewhere are not affected
  - [ ] When a notification with an implemented toast mapping is triggered, the notification appears as a toast
  - [ ] Clicking the action button on the notification both performs the correct action, and marks the notification as read
- [ ] New notifications
  - [ ] Log in as user A and user B
  - [ ] User A responds to user B's standup response
  - [ ] User B sees a toast, and can click directly to the response
  - [ ] User A mentioned user B in their standup response
  - [ ] User B sees a toast, and can click directly to the response

## Final checklist

### DIFF
--- a/packages/client/components/NotificationPicker.tsx
+++ b/packages/client/components/NotificationPicker.tsx
@@ -1,14 +1,13 @@
 import graphql from 'babel-plugin-relay/macro'
 import React, {Suspense} from 'react'
 import {createFragmentContainer} from 'react-relay'
-import {ValueOf} from '~/types/generics'
 import lazyPreload, {LazyExoticPreload} from '~/utils/lazyPreload'
 import {
   NotificationEnum,
   NotificationPicker_notification
 } from '~/__generated__/NotificationPicker_notification.graphql'
 
-const typePicker = {
+const typePicker: Record<NotificationEnum, LazyExoticPreload<any>> = {
   KICKED_OUT: lazyPreload(() => import(/* webpackChunkName: 'KickedOut' */ './KickedOut')),
   PAYMENT_REJECTED: lazyPreload(
     () => import(/* webpackChunkName: 'PaymentRejected' */ './PaymentRejected')
@@ -36,7 +35,7 @@ const typePicker = {
   RESPONSE_REPLIED: lazyPreload(
     () => import(/* webpackChunkName: 'ResponseReplied' */ './ResponseReplied')
   )
-} as Record<NotificationEnum, LazyExoticPreload<any>>
+}
 
 interface Props {
   notification: NotificationPicker_notification
@@ -45,7 +44,7 @@ interface Props {
 const NotificationPicker = (props: Props) => {
   const {notification} = props
   const {type} = notification
-  const SpecificNotification = typePicker[type] as ValueOf<typeof typePicker>
+  const SpecificNotification = typePicker[type]
   return (
     <Suspense fallback={''}>
       <SpecificNotification notification={notification} />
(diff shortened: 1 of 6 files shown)


packages/client/mutations/toasts/mapResponseMentionedToToast.ts
```diff
@@ -0,0 +1,46 @@
+import graphql from 'babel-plugin-relay/macro'
```
jmtaber129, line 1: Future notifications will only need a file like this + a couple lines in `popNotificationToast` to implement the toast - no need to implement an `onNext` handler and get into the internals of the notification subscription, this should Just Work ™️

packages/client/mutations/toasts/popNotificationToast.ts
```diff
@@ -0,0 +1,80 @@
...
+
+const typePicker = {
+  KICKED_OUT: null,
+  PAYMENT_REJECTED: null,
+  TASK_INVOLVES: null,
+  PROMOTE_TO_BILLING_LEADER: null,
+  TEAMS_LIMIT_EXCEEDED: null,
+  TEAM_ARCHIVED: null,
+  TEAM_INVITATION: null,
+  MEETING_STAGE_TIME_LIMIT_END: null,
```
BartoszJarocki, line 21: @jmtaber129 I assume we didn't have all these notifications below, and it's just a placeholder to introduce them in the future?

jmtaber129: This was initially sort of some copy-pasta from `NotificationPicker`, but I realized it's better to just declare the type as a `Partial<Record<>>` instead of explicitly marking these as `null`.

Went ahead and updated `NotificationPicker` to reflect this, too.

BartoszJarocki (changes requested):
@jmtaber129 code looks good! there's one case that I found not working

1. User A logs in, opens a standup meeting
2. User B logs in, opens the same standup meeting
3. User B responds to User A response
4. User A sees the response, clicks the `See the discussion`, nothing happens (ie. the discussion side panel won't open)

Everything works correctly when User A isn't in the meeting, then clicking `See the discussion` opens the meeting correctly _with_ the sidebar opened.

My guess there's somewhere `useEffect` without deps, not sure

jmtaber129:
@BartoszJarocki nice catch! Implemented a fix

BartoszJarocki (approved):
lgmt!

packages/client/components/NotificationPicker.tsx
```diff
@@ -45,7 +44,7 @@ interface Props {
 const NotificationPicker = (props: Props) => {
   const {notification} = props
   const {type} = notification
-  const SpecificNotification = typePicker[type] as ValueOf<typeof typePicker>
```
mattkrick, line 48: +2 TIL!

mattkrick (approved):

PR #7606, opened 2022-12-21
feat(teams-limit): lock organization if limit the is exceeded for 30 days

### LINKED ISSUE #7386: Teams limit: Lock organization after 30 days
We need to block all organization teams and send notification if organization was not upgraded within 30 days or some teams were not removed to fix the limit

Email designs TBD

**AC:**
- Cron should lock an organization if `scheduledLockAt` is set (Try to use existing `ScheduledJob` functionality)
- Ignore archived organizations
- Set `lockedAt = new Date()`
- User should still be able to remove teams
- If limit is fixed, block should be removed
- Upgrading organization should remove block from all teams

igorlesnenko:
Fixes #7386

How to test:

- [ ] Uncomment the following lines [here](https://github.com/ParabolInc/parabol/pull/7606/files#diff-0ebb6687f73dcbeef95db77a23ed552ac97457d45368062605a87c17330b9f9dR13)

```
// const enum Threshold {
//   MAX_PERSONAL_TIER_TEAMS = 0,
//   MIN_STICKY_TEAM_MEETING_ATTENDEES = 1,
//   MIN_STICKY_TEAM_MEETINGS = 1,
//   PERSONAL_TIER_LOCK_AFTER_DAYS = 0
// }
```

- [ ] Run and complete a meeting to trigger limits check function that will create a lock job
- [ ] Run a job by performing the following query here http://localhost:3000/admin/graphql

```
mutation {
 runScheduledJobs(seconds: 5000) 
}
```

- [ ] See organization now has `lockedAt` flag set (No UI changes in this PR)
- [ ] Upgrade the org and see that `lockedAt` flag is gone

**Scenario 2.**

In some cases we may skip the locking. 

- [ ] Before running a lock job, remove `scheduledLockAt` from the organization. Run lock job. See that org has not become locked.
- [ ] Before running a lock job, set any custom `lockedAt` on the organization. Run lock job. See that org has not become re-locked, `lockedAt` value has remained the same
- [ ]  Before running a lock job, change `scheduledLockAt` to any custom date (which means locking was rescheduled), see that org has not become locked.


packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -158,4 +175,34 @@ export const checkTeamsLimit = async (orgId: string, dataLoader: DataLoaderWorke
     await enableUsageStats(billingLeadersIds, orgId)
     await sendWebsiteNotifications(orgId, billingLeadersIds, dataLoader)
   }
+
+  await scheduleJobs(scheduledLockAt, orgId)
+}
+
+export const processLockOrganizationJob = async (
+  job: ScheduledJobOrganizationLock,
+  {dataLoader}: {dataLoader: DataLoaderWorker}
```
nickoferrall, line 184: +1 could pass in the `dataLoader` without putting it in an object, so 

```
  job: ScheduledJobOrganizationLock,
  dataLoader: DataLoaderWorker
```

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -158,4 +175,34 @@ export const checkTeamsLimit = async (orgId: string, dataLoader: DataLoaderWorke
...
+) => {
+  const {orgId, runAt} = job
+
+  const organization = await dataLoader.get('organizations').load(orgId)
+
+  // Skip the job if unlocked or already locked or scheduled lock date changed
+  if (
+    !organization.scheduledLockAt ||
+    organization.lockedAt ||
+    organization.scheduledLockAt.getTime() !== runAt.getTime()
```
nickoferrall, line 194: -1 here we're returning if the `runAt` timestamp is different to the `scheduledLockAt` timestamp, but as `runScheduledJobs` happens every 10 minutes, it's likely that the timestamps will never match. 

I think we need to check if the date is the same rather than the timestamp

nickoferrall (changes requested):

packages/server/billing/helpers/teamLimitsCheck.ts
igorlesnenko: @nickoferrall We set `runAt` value at the job creation, so it is not actual real-time value, but the value when job is supposed to be run.

https://github.com/ParabolInc/parabol/pull/7606/files#diff-0ebb6687f73dcbeef95db77a23ed552ac97457d45368062605a87c17330b9f9dR46

So If `runAt !== scheduledLockAt` that means that lock time was changed for some reason, and we don't need to run this job.

nickoferrall: OK I see, that makes sense 👍

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -158,4 +175,34 @@ export const checkTeamsLimit = async (orgId: string, dataLoader: DataLoaderWorke
...
+
+  const now = new Date()
+
+  return r
+    .table('Organization')
+    .get(orgId)
+    .update({
+      lockedAt: now
+    })
+    .run()
```
nickoferrall, line 207: +1 I think the file should be broken up into separate files as it's becoming quite long and difficult to follow

nickoferrall (approved):

packages/server/graphql/private/mutations/runScheduledJobs.ts
```diff
@@ -35,19 +36,24 @@ const processMeetingStageTimeLimits = async (
...
-  const processor = jobProcessors[job.type]
-  processor(job, {dataLoader}).catch(console.log)
+
+  if (job.type === 'MEETING_STAGE_TIME_LIMIT_END') {
+    return processMeetingStageTimeLimits(
+      job as ScheduledJobMeetingStageTimeLimit,
+      dataLoader
+    ).catch(console.log)
+  } else if (job.type === 'LOCK_ORGANIZATION') {
+    return processLockOrganizationJob(job as ScheduledJobOrganizationLock, dataLoader).catch(
```
nickoferrall, line 53: +1 we can avoid the casting here by checking if `orgId` exists in the job:

```
  if ('orgId' in job) {
    return processLockOrganizationJob(job, dataLoader).catch(console.log)
  } else {
    return processMeetingStageTimeLimits(job, {
      dataLoader
    }).catch(console.log)
  }
```

packages/server/billing/helpers/teamLimitsCheck.ts
igorlesnenko: changed here #7637

packages/server/graphql/private/mutations/runScheduledJobs.ts
igorlesnenko: changed here #7637

packages/server/graphql/private/mutations/runScheduledJobs.ts
```diff
@@ -35,19 +36,24 @@ const processMeetingStageTimeLimits = async (
...
+      console.log
```
Dschoordsch, line 54: +1 `console.error`?

packages/server/graphql/private/mutations/runScheduledJobs.ts
```diff
@@ -35,19 +36,24 @@ const processMeetingStageTimeLimits = async (
...
   // prevent duplicates. after this point, we assume the job finishes to completion (ignores server crashes, etc.)
   if (res.deleted !== 1) return
-  const processor = jobProcessors[job.type]
-  processor(job, {dataLoader}).catch(console.log)
+
+  if (job.type === 'MEETING_STAGE_TIME_LIMIT_END') {
+    return processMeetingStageTimeLimits(
+      job as ScheduledJobMeetingStageTimeLimit,
+      dataLoader
+    ).catch(console.log)
```
Dschoordsch, line 51: +1 `console.error`?

packages/server/billing/helpers/teamLimitsCheck.ts
Dschoordsch: -1 You're loading the organization with the dataloader in l188, so we should invalidate or update it in the cache because we don't know what will be run after this function, there might be more org related things scheduled.

Dschoordsch (changes requested):
Did not test

igorlesnenko:
@tghanken do we need any additional events here to track other than added in the next pr here #7637?

Dschoordsch (approved):
Looks good, did not test

tghanken:
> @tghanken do we need any additional events here to track other than added in the next pr here #7637?

I think we're fine without explicitly tracking events here.

tghanken (approved):

Issue #7560, opened 2022-12-13
Jira voting does not work although the field is on the correct screen


Dschoordsch:
This [customer](https://app.hubspot.com/contacts/3888472/contact/87227101) has an issue where they cannot vote to "Story Points" although the field is visible in the drop down and on the correct screen in Jira.

Because everything seems correct at first glance, we should add some more logging around this issue so we can resolve it.

Dschoordsch:
There seems to be an issue with selecting the correct field.

Dschoordsch:
Sentry issue: [ACTION-PRODUCTION-DAF](https://sentry.io/organizations/parabol/issues/3822484391/?referrer=github_plugin)

Dschoordsch:
Issue confirmed: There are multiple fields with name "Story Points" and thus multiple ids, but only 1 is correct.

Dschoordsch:
It can be easily reproduced and is most likely the root cause for this Jira issue:
* create a company managed team in Jira with the default issue types (Bug, Story)
* check that Story has "Story Points" field
* create a new custom field of type number and name it "Story Points" as well
* assign it to the Bug screen

Now when trying to switch voting for Bugs to "Story Points" it will try to vote to the Story field and fail.

PR #7607, opened 2022-12-21
fix(Jira): fix Sprint Poker voting for projects with duplicate fields

Dschoordsch:
Resolves: #7560 

Jira allows to have multiple fields with the same name in a project, even on the same issue. While this does not make sense from a usability standpoint, this backfired for us, because we were selecting fields based on their name and could end up selecting fields which belonged to a different issue type altogether. Now even multiple fields called "Story Points" on one issue are allowed and work.

## Demo

https://www.loom.com/share/b56ce436c22b4eeb870957e3d3592a94

## Testing scenarios

See demo

- Create at least 2 fields with the name "Story Points" in a Jira project
- Assign 1 field to the issue type "Bug" and the other one to the issue type "Story"
- [ ] Start a Sprint Poker with an issue of each type and vote to both "Story Points" fields and see the vote works

### DIFF
--- a/packages/client/components/JiraFieldDimensionDropdown.tsx
+++ b/packages/client/components/JiraFieldDimensionDropdown.tsx
@@ -57,7 +57,10 @@ const JiraFieldDimensionDropdown = (props: Props) => {
         task {
           integration {
             ... on JiraIssue {
-              possibleEstimationFieldNames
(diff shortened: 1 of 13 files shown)


nickoferrall (approved):
Tested and it LGTM!

Issue #6286, opened 2022-03-25
Show warning when switching away from a branch with a migration


Dschoordsch:
When switching development branches with migrations, it is easy to forget to run the down migration before checking out another branch, which can mess up the local database.

## Acceptance criteria
* [ ] if local branch has migration not present on target branch, alert the user
* [ ] (optional) stop the checkout until the user confirms the previous warning "Your current branch has additional migrations which you might want to run down, do you want to continue with the checkout?"​

jordanh:
@Dschoordsch and I discussed. He wants to keep it open. He says, with confidence, "I will do it during slack week." 😉

PR #7616, opened 2023-01-03
chore(DX): Show a warning on switching branch with migrations

Dschoordsch:
Fixes: #6286 

When switching branches where the target branch is missing some migrations, print a warning with the migration files which will not be present there and steps to run the down migration. The down migrations are not run automatically.

## Demo

https://www.loom.com/share/f05d16bb86ee432984e4a86d6591759d

## Testing scenarios

See demo
* [ ] with the hook change switch to a branch which does not have all current migrations
  * [ ] check with different postgres and rethink migrations
* [ ] warning message is printed in this case

### DIFF
--- a/.husky/post-checkout
+++ b/.husky/post-checkout
@@ -3,10 +3,10 @@
 
 yarn postcheckout
 
-CURRENT_BRANCH=$(git branch --show-current)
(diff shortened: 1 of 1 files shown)


BartoszJarocki (approved):
good stuff!

PR #7618, opened 2023-01-03
fix: emails landing in spam

mattkrick:

# Description

We've had reports that emails land in spam.
Digging into it, I tracked it down to poor color contrast in our emails. Spam filters thought were were trying to hide invisible text in the emails. The 2 problem areas were the link colors & footer colors.

I also saw some spam filters look for a physical address & phone number, so I added those to the footer. 

### BEFORE
Screenshot from 2023-01-03 12-13-24
Screenshot from 2023-01-03 10-23-45

### AFTER
Screenshot from 2023-01-03 12-44-00
Screenshot from 2023-01-03 12-35-34

## Testing scenarios

- [  ] Make sure the screenshots look OK!


mattkrick:
that is a very good question! i have no idea :rofl:

Issue #7467, opened 2022-11-22
AI Summary: track relevant events


nickoferrall:
To understand whether the AI summary feature has been used and is useful, we'll want to track relevant events. 

Speak with the data team to find out which events should be tracked and update the AC below.

### Proposed AC

- [ ] Track when the meeting summary has been viewed

nickoferrall:
Hey @tghanken, are there any events that we should start tracking now for this feature?

tghanken:
I'd like to track both views and engagement:

### Views
- Event: `Viewed AI Summary`
  - Property `source`: either `Meeting Summary` or `Discussion`
  - Property `tier`: user's tier, so we can track if the trial works
  - Property `meetingId`
 
### Engagement
I'm a little less certain of what we can/should track here.  Do we have any finalized designs yet?  I think at the bare minimum, we can track reactjis in the discussion, but maybe we could have a "learn more" CTA or other way for them to interact?  Regardless, we can use our existing tracking to see if this drives conversions, it'll just be hard to see if people find the actual feature useful without some kind of non-paid conversion action.

PR #7619, opened 2023-01-04
chore(ai-summary): track when the summaries are viewed

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7467

### To test

- [ ] When the meeting summary is viewed with an AI Summary, an event is tracked in Segment
- [ ] When the discussion thread is viewed with an AI Topic Summary, an event is tracked in Segment
- [ ] Track adding an emoji to the AI comment in the discussion thread

### DIFF
--- a/packages/client/components/ThreadedCommentBase.tsx
+++ b/packages/client/components/ThreadedCommentBase.tsx
@@ -1,7 +1,7 @@
 import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
 import {convertToRaw, EditorState} from 'draft-js'
-import React, {ReactNode, useRef, useState} from 'react'
+import React, {ReactNode, useEffect, useRef, useState} from 'react'
 import {commitLocalUpdate, createFragmentContainer} from 'react-relay'
 import useAtmosphere from '~/hooks/useAtmosphere'
 import useEditorState from '~/hooks/useEditorState'
@@ -14,8 +14,10 @@ import isTempId from '~/utils/relay/isTempId'
 import {ThreadedCommentBase_comment} from '~/__generated__/ThreadedCommentBase_comment.graphql'
 import {ThreadedCommentBase_discussion} from '~/__generated__/ThreadedCommentBase_discussion.graphql'
 import {ThreadedCommentBase_viewer} from '~/__generated__/ThreadedCommentBase_viewer.graphql'
+import SendClientSegmentEventMutation from '../mutations/SendClientSegmentEventMutation'
 import anonymousAvatar from '../styles/theme/images/anonymous-avatar.svg'
 import deletedAvatar from '../styles/theme/images/deleted-avatar-placeholder.svg'
+import {PARABOL_AI_USER_ID} from '../utils/constants'
 import {DiscussionThreadables} from './DiscussionThreadList'
 import CommentEditor from './TaskEditor/CommentEditor'
 import ThreadedAvatarColumn from './ThreadedAvatarColumn'
@@ -87,6 +89,17 @@ const ThreadedCommentBase = (props: Props) => {
       editorRef.current?.focus()
     })
   }
+
+  useEffect(() => {
+    if (createdByUserNullable?.id === PARABOL_AI_USER_ID) {
+      SendClientSegmentEventMutation(atmosphere, 'AI Summary Viewed', {
+        source: 'Discussion',
+        tier: viewer.tier,
+        meetingId,
+        commentId
+      })
+    }
+  }, [])
 
   const onToggleReactji = (emojiId: string) => {
     if (submitting) return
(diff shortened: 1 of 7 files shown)


packages/client/components/ThreadedCommentBase.tsx
```diff
@@ -88,6 +90,20 @@ const ThreadedCommentBase = (props: Props) => {
...
 
+  useEffect(() => {
+    if (createdByUserNullable?.id === PARABOL_AI_USER_ID) {
+      const contentState = editorState.getCurrentContent()
+      const summary = contentState.getPlainText()
+      SendClientSegmentEventMutation(atmosphere, 'AI Summary Viewed', {
+        source: 'Discussion',
+        tier: viewer.tier,
+        meetingId,
+        discussionTopicId,
```
nickoferrall, line 101: @tghanken I added the `discussionTopicId` to the Discussion event as the `AI Summary Viewed` event will be triggered for every Discuss Stage. With `discussionTopicId`, we can see which topic is being summarised

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/WholeMeetingSummaryResult.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const {stages} = discussPhase ?? {}
+  const explainerText = team?.tier === 'starter' ? AIExplainer.STARTER : AIExplainer.PREMIUM_MEETING
+  useEffect(() => {
+    const discussionSummaries = stages?.map((stage) => stage.discussion?.summary).filter(isNotNull)
+    SendClientSegmentEventMutation(atmosphere, 'AI Summary Viewed', {
+      source: 'Meeting Summary',
+      tier: meeting.team.tier,
+      meetingId: meeting.id,
+      discussionSummaries,
+      summary: wholeMeetingSummary
```
nickoferrall, line 78: I thought we could add the summaries to the segment events. This way, we can read through the summaries to see if they make sense or if the OpenAI prompt needs adjusting. 

In the meeting summary, I'm tracking both the `wholeMeetingSummary` and all of the `discussionSummaries`, i.e. the summary of the discussion threads. The topic summary is being tracked in the discussion thread.

packages/server/graphql/mutations/addReactjiToReactable.ts
```diff
@@ -164,13 +165,16 @@ const addReactjiToReactable = {
...
     analytics.reactjiInteracted(
       viewerId,
       meetingId,
       meetingType,
       reactableId,
       reactableType,
-      !!isRemove
+      reactji,
+      !!isRemove,
+      isAIComment
```
nickoferrall, line 177: I suggest that we record which `reactji` is being interacted with, and I've added a boolean `isAIComment` field. 

Emojis isn't a perfect way to measure whether users like the summaries, but the types of emojis they use should give us some indication of whether they like it or not.

tghanken: I like this a lot!

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/WholeMeetingSummaryResult.tsx
tghanken: -1 I'm not sure we really need to send the full summaries to the Segment events, and should generally try to avoid sending possible PII.

If we're storing them as normal discussion items in the DB, maybe we can just pass the ID into Segment?  We can still do the deeper analysis, without worrying about how this propagates through our tools.

tghanken (changes requested):
-1 We should avoid passing large strings into Segment for performance/cost/privacy reasons.  I'd rather pass the ID and do our "Summary Triage" in data-sanctum.

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/WholeMeetingSummaryResult.tsx
nickoferrall: That makes sense! Updated 👍

tghanken (approved):

packages/server/graphql/mutations/addReactjiToReactable.ts
```diff
@@ -164,13 +165,16 @@ const addReactjiToReactable = {
 
     const data = {reactableId, reactableType}
     const {meetingType} = await dataLoader.get('newMeetings').load(meetingId)
+    const isAIComment = reactable instanceof Comment && reactable.createdBy === PARABOL_AI_USER_ID
```
Dschoordsch, line 168: -1 Does not work for me:
Screenshot from 2023-01-09 17-19-49
Screenshot from 2023-01-09 17-20-03

Dschoordsch (changes requested):

Dschoordsch (approved):

Issue #7622, opened 2023-01-05
Check-In: Cannot move or unpin agenda items


Dschoordsch:
## How to reproduce
Start a Check-In with already pinned agenda items

## Expected behaviour
I can create, move and pin and unpin agenda items

## Actual behaviour
Moving, pinning and unpinning does not work

PR #7623, opened 2023-01-05
fix: Legitity min/max failed on undefined value

Dschoordsch:
Fixes: #7622 

Legitity has a separate `required` test which checks for non-undefined, thus all other checks need to be able to cope with undefined values.

Added tests to simplify verification of the fix.

## Demo
Peek 2023-01-05 13-57

## Testing scenarios

- in a Check-In meeting
  - [ ] reorder, pin and unpin agenda items

- run client tests
  - `yarn workspace parabol-client test`

## Final checklist

### DIFF
--- a/packages/client/validation/__tests__/Legitity.test.ts
+++ b/packages/client/validation/__tests__/Legitity.test.ts
@@ -1,0 +1,21 @@
+import Legitity from '../Legitity'
+
+test('Legitity.max succeeds with undefined', async () => {
+  const legitity = new Legitity(undefined).max(12, 'too long')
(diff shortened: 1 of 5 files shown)


packages/client/validation/__tests__/Legitity.test.ts
```diff
@@ -0,0 +1,21 @@
+import Legitity from '../Legitity'
+
+test('Legitity.max succeeds with undefined', async () => {
```
nickoferrall, line 3: +2 great example of something to test

nickoferrall (approved):
Works for me. Thanks for the fix 🙏

PR #7633, opened 2023-01-09
fix: server error undefined template

### LINKED ISSUE #7632: Error: Cannot destructure property 'isFree' of 'p' as it is undefined.
https://sentry.io/organizations/parabol/issues/3862453178/?referrer=github_plugin

```
Error: Cannot destructure property 'isFree' of 'p' as it is undefined.
```

mattkrick:
# Description

Fixes #7632

Bug was caused by accessing a prop when the parent object was undefined.


nickoferrall (approved):
LGTM! Thanks for the fix 🙏

PR #7635, opened 2023-01-10
chore(metrics): emit sign_up events from GA4 client side

tianrunhe:
# Description

Today, we sent `sign_up` events from Segment to GA4. However, due to limitation of Segment, these events have no session information (or any info):

Screenshot 2023-01-09 at 3 56 12 PM

As a result, `sign_up` events cannot be attributed to previous `page_view` events in GA4:

211431799-ecd7bbd7-eef0-4a23-89a2-0353f06454ee

Screenshot 2023-01-09 at 4 01 24 PM

Basically no useful analysis can be performed...😢

This PR emits `sign_up` events directly from GA4 client side. This way all relevant info (session, geo, ...) are automatically populated.

## Demo

The GA4 client side emitted `sign_up` events look like these in BigQuery:

Screenshot 2023-01-09 at 4 06 29 PM

Attribution seems to work afterwards:

Screenshot 2023-01-09 at 4 08 06 PM

Screenshot 2023-01-09 at 4 07 26 PM

## Testing scenarios

- [ ] Sign up with username and password
  - Verify a `sign_up` event is emitted to GA4

- [ ] Sign up with Google
  - Verify a `sign_up` event is emitted to GA4

- [ ] Log in with Google
  - Verify **NO** `sign_up` event is emitted to GA4

## Final checklist

### DIFF
--- a/packages/client/components/AnalyticsPage.tsx
+++ b/packages/client/components/AnalyticsPage.tsx
@@ -3,7 +3,7 @@
 import {datadogRum} from '@datadog/browser-rum'
 import * as Sentry from '@sentry/browser'
 import graphql from 'babel-plugin-relay/macro'
-import {useEffect, useRef, useState} from 'react'
+import {useEffect, useRef} from 'react'
 import ReactGA from 'react-ga4'
 import useAtmosphere from '~/hooks/useAtmosphere'
 import {LocalStorageKey} from '~/types/constEnums'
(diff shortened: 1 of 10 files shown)


packages/client/components/AnalyticsPage.tsx
```diff
@@ -89,39 +89,39 @@ const TIME_TO_RENDER_TREE = 100
 
 const AnalyticsPage = () => {
   const atmosphere = useAtmosphere()
-  const [isGAInitialized, setIsGAInitialized] = useState(false)
+  if (gaMeasurementId) {
+    ReactGA.initialize(gaMeasurementId, {
```
tianrunhe, line 93: There won't be duplicate initializations as it's [protected](https://github.com/PriceRunner/react-ga4/blob/master/src/ga4.js#L183) inside `ReactGA`

packages/client/components/AnalyticsPage.tsx
```diff
@@ -89,39 +89,39 @@ const TIME_TO_RENDER_TREE = 100
...
+        const {viewer} = res
+        const {id, segmentId} = viewer
+        ReactGA.set({
+          userId: id,
+          clientId: segmentId ?? getAnonymousId()
         })
+      } else {
+        ReactGA.set({
+          userId: null,
+          clientId: getAnonymousId()
```
tianrunhe, line 119: We need the `clientId` before login & after logout to attribute the entire user journey

Dschoordsch (approved):
I had some trouble testing this, after logging out it did not reliably send any events anymore, even with a new user. I trust you that you've tested this properly and this is a local issue.

Dschoordsch:
Also, is it intentional that we don't track SAML signups?

tianrunhe:
> Also, is it intentional that we don't track SAML signups?

I completely overlooked this use case. Thanks for pointing it out! What do you think my implementation here? It's a bit different than sign up with password or Google.

Dschoordsch (commented):
Code looks good, I just haven't come around setting up SAML for testing yet, will do tomorrow.

Dschoordsch (approved):
Debugging GA4 still feels a bit like a gamble, but the SAML event works just fine, good work!

Issue #7518, opened 2022-12-02
Teams limit: send warning email when the limit is exceeded


igorlesnenko:
Prev issue: #7379

Designs TBD

AC:
- [ ] Send a warning email to all organization billing leaders, saying that the tier limit was exceeded and account will be blocked in 30 days

igorlesnenko:
Worth checking notification preview functionality: https://github.com/ParabolInc/parabol/pull/7563#pullrequestreview-1218409335 
Maybe it could be used for this email notification

acressall:
@avivapinchas can you proofread the copy for these emails? @igorlesnenko I have three emails here -- are we missing any use cases? https://docs.google.com/document/d/1XUGU2Ozhdapc4dSA2raNbi6gA0FexxUJRw2xlQQcxNM/edit?usp=sharing

avivapinchas:
I've made some non-trivial edits here. I wanted to: 
- Share more info, if possible, to make the scenarios more clear - @igorlesnenko lmk if we aren't able to pull in Org name and Number of teams 
- Make the emails more different from one another, increase the urgency with each email, and focus on loss aversion (don't lose something you're used to) as the main persuasive tool 

LMK if this isn't what you had in mind!

PR #7637, opened 2023-01-10
chore(teams-limit): send locked & warning email

### LINKED ISSUE #7388: Teams limit: Additional email reminders on day 7
When the limit is exceeded we need to send additional reminders that they need to fix or upgrade until fully blocked

See cron issue: #7386

Designs: https://www.figma.com/file/aYc5fKU09GawnsQwKTjcDv/Time-bound-stick?node-id=0%3A1&t=mTNGlHiYcWKypsHC-0

**AC**

- Send additional email reminder on day 7

### LINKED ISSUE #7639: Teams limit: Send email notification when organization is locked
Prev issue: #7386

AC

- Send email notification to all billing leaders when organization is locked

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7518

Fix #7639

Fix #7388

Demo: https://www.loom.com/share/06704d9b05444531b4bc07dadbc9ddfd

Email examples (30 day email says there's 1 sticky team because it was a test - in prod, they would need 3 or more sticky teams to get an email):

30-day-limit

seven-day-warning

deactivated

### To test

- [ ] Make sure the org has the teamLimit feature flag switched on
- [ ] Uncomment the Threshold section in teamLimitsCheck [here](https://github.com/ParabolInc/parabol/blob/596f030bdb3e16c61512e3b442e94ca88475ed86/packages/server/billing/helpers/teamLimitsCheck.ts#L16) for easier testing
- [ ] Run a retro meeting and see that you receive the Team Limit Reached email
- [ ] Update an org in the data explorer so that the `scheduledLockAt` field is a date in six days, e.g. `r.db('actionDevelopment').table('Organization').get(ORG_ID).update({scheduledLockAt: new Date(new Date().getTime() + 6 * 24 * 60 * 60 * 1000)})`
- [ ] Go to `/admin/graphql` and run `runScheduledJobs(seconds: 10000)`. See that you receive a warning email
- [ ] Update an org in the data explorer so that the `scheduledLockAt` field is a date in the past and `lockedAt` is null
- [ ] Go to `/admin/graphql` and run `runScheduledJobs(seconds: 10000)`. See that you receive the Parabol Account Deactivated email


packages/client/modules/email/components/LimitsEmails/SevenDayWarningEmail.tsx
```diff
@@ -0,0 +1,78 @@
...
+          {'This is a follow-up notification to remind you that '}
+          <span style={{fontWeight: 600}}>
+            {`${orgName}'s Parabol account is at risk of deactivation`}
+          </span>
+          {' because you’ve exceeded the two team limit on your '}
+          <a style={linkStyle} href={ExternalLinks.PRICING_LINK}>
+            {'Starter Plan'}
+          </a>
+          {
+            '. Once your account is deactivated, you will lose access to your teams and won’t be able to run Retrospective, Sprint Poker or Standup meetings with Parabol.'
```
nickoferrall, line 50: We could say `won't be able to run any meetings with Parabol` to make this more concise or is this a way of reminding them of the meeting types we offer?

packages/client/modules/email/components/LimitsEmails/ThirtyDayWarningEmail.tsx
```diff
@@ -0,0 +1,88 @@
...
+}
+
+type Props = LimitsEmailProps & {
+  stickyTeamCount: number
+}
+
+export default function ThirtyDayWarningEmail(props: Props) {
+  const {appOrigin, preferredName, orgId, orgName, stickyTeamCount} = props
+  const billingURL = makeAppURL(appOrigin, `/me/organizations/${orgId}/billing`, {
+    searchParams: {
```
nickoferrall, line 32: @tghanken I'm adding these search params so that we can tell where the user is coming from when the land on the billing page. Is that what you have in mind? Do you know if we're tracking the URL params in the billing page yet? And would you like me to edit these params at all?

packages/server/utils/analytics/analytics.ts
```diff
@@ -369,6 +371,10 @@ class Analytics {
     this.track(userId, 'Summary Email Setting Changed', {subscribeToSummaryEmail})
   }
 
+  teamsLimitEmailSent = (userId: string, orgId: string, type: TeamLimitsEmailType) => {
```
nickoferrall, line 374: @tghanken I added this analytics event so that we track when emails are sent. Would you like this to be edited at all?

packages/server/graphql/private/mutations/processScheduledLocks.ts
```diff
@@ -0,0 +1,107 @@
...
+  _source,
+  _args,
+  {dataLoader}
+) => {
+  const r = await getRethink()
+  const now = new Date().getTime()
+  const inOneWeek = now + ms('7d')
+  const inSixDays = now + ms('6d')
+
+  const orgsScheduledForLockWithinWeek = await r
```
nickoferrall, line 17: As the mutation is run daily, there shouldn't be many `orgsScheduledForLockWithinWeek`. It also doesn't matter too much if the mutation takes a while to execute because it's a chron job. So, while there's a lot of mapping and filtering below, I _aimed_ for readability over performance, although it may still be difficult to read 🤷‍♂️

nickoferrall:
@avivapinchas the PR description has examples of what the emails currently look like. Let me know if you'd like me to make any changes!

avivapinchas:
@nickoferrall LGTM! Updated the copy to reflect we'll be fully locking the account ([here](https://docs.google.com/document/d/1XUGU2Ozhdapc4dSA2raNbi6gA0FexxUJRw2xlQQcxNM/edit#).

packages/server/graphql/private/mutations/processScheduledLocks.ts
```diff
@@ -0,0 +1,107 @@
+import ms from 'ms'
+import sendTeamsLimitEmail from '../../../billing/helpers/sendTeamsLimitEmail'
+import getRethink from '../../../database/rethinkDriver'
+import isValid from '../../isValid'
+import {MutationResolvers} from '../resolverTypes'
+
+const processScheduledLocks: MutationResolvers['processScheduledLocks'] = async (
```
igorlesnenko, line 7: @nickoferrall I happened to make a slightly different solution in my PR. It uses the existing scheduled job mechanism. 

The idea is to create scheduled jobs every time the limit is exceeded. It could be multiple jobs created at once - block after 30 days, alert after 7, etc.

Here is the PR: https://github.com/ParabolInc/parabol/pull/7606/files

What do you think about this solution?

packages/client/modules/email/components/LimitsEmails/ThirtyDayWarningEmail.tsx
tghanken: I think these params make sense based on what we've collected historically.  We're collecting URL params globally throughout the site using Segment pageviews, so we should be setup to capture this!

packages/server/utils/analytics/analytics.ts
tghanken: +1 Could we change this to a generic `Notification Email Sent`? I'd rather have it be usable for other similar types of conversion messaging in the future without changing the name.  We can always expand what we pass into the `type` parameter for other email types.

tghanken (approved):
Looks good!  One small renaming wishlist item.

packages/server/graphql/private/mutations/processScheduledLocks.ts
igorlesnenko: @nickoferrall I think yes. There is already a job that locks an organization. You can add sending email to it here:

https://github.com/ParabolInc/parabol/pull/7606/files#diff-0ebb6687f73dcbeef95db77a23ed552ac97457d45368062605a87c17330b9f9dR201-R207

Then you can create a reminder email job object here: 

https://github.com/ParabolInc/parabol/pull/7606/files#diff-0ebb6687f73dcbeef95db77a23ed552ac97457d45368062605a87c17330b9f9dR50

And add a job processor for it here:

https://github.com/ParabolInc/parabol/pull/7606/files#diff-88438075aa9f4efe2c41c3ee8437756a2c7a6dfd1a50e2e3718cba2660330f4bR46-R55

nickoferrall: Sounds good, I'll work on implementing this today 👍

packages/server/utils/analytics/analytics.ts
nickoferrall: For sure - updated!

packages/server/graphql/private/mutations/processScheduledLocks.ts
nickoferrall: Update pushed! I've also started splitting up the `teamLimitChecks` file and changed the base of the PR to your `lock-scheduled-job` PR

packages/server/graphql/private/mutations/runScheduledJobs.ts
```diff
@@ -36,23 +36,17 @@ const processMeetingStageTimeLimits = async (
...
-  if (job.type === 'MEETING_STAGE_TIME_LIMIT_END') {
-    return processMeetingStageTimeLimits(
-      job as ScheduledJobMeetingStageTimeLimit,
-      dataLoader
-    ).catch(console.log)
-  } else if (job.type === 'LOCK_ORGANIZATION') {
-    return processLockOrganizationJob(job as ScheduledJobOrganizationLock, dataLoader).catch(
-      console.log
-    )
+  if ('orgId' in job) {
```
igorlesnenko, line 46: I was checking a `type` here because in future, we may have more job types added and they may also have `orgId`. What do you think?

packages/client/modules/email/components/LimitsEmails/LockedEmail.tsx
```diff
@@ -0,0 +1,78 @@
...
+  return (
+    <Layout maxWidth={544}>
+      <EmailBlock innerMaxWidth={innerMaxWidth}>
+        <Header align='left' appOrigin={appOrigin} corsOptions={EMAIL_CORS_OPTIONS} />
+        <p style={{...copyStyle, marginBottom: '0px'}}>{`Hi ${preferredName} 👋`}</p>
+        <EmptySpace height={16} />
+        <p style={{...copyStyle, marginBottom: '0px'}}>
+          {'Unfortunately, '}
+          <span style={{fontWeight: 600}}>
+            {`${orgName} has exceeded the two-team limit on the Starter Plan for more than 30 days, and your account has been deactivated.`}
```
igorlesnenko, line 49: Is it ok to hardcode numbers or should we take it from Threshold?

packages/server/email/teamLimitsEmailCreator.tsx
```diff
@@ -0,0 +1,113 @@
...
+      utm_source: 'notification email',
+      utm_medium: 'email',
+      utm_campaign: 'locked-email'
+    }
+  })
+
+  const emailTextLookup = {
+    thirtyDayWarning: `Hi ${preferredName} 👋
+
+    This is a friendly note to let you know that ${orgName} has officially reached ${stickyTeamCount} active teams on Parabol - congrats! We love to see organizations finding value in Parabol and improving their teams in the process.
```
igorlesnenko, line 26: I think we don't need to use real-time `stickyTeamCount` value, as everywhere else we are talking about 2 teams limit. I think it is ok to just say that they are over 2 teams, without specifying exact number. I would use `MAX_PERSONAL_TIER_TEAMS`

 What do you think?

packages/server/database/types/processTeamsLimitsJob.ts
```diff
@@ -0,0 +1,33 @@
...
+const processTeamsLimitsJob = async (job: ScheduledTeamLimitsJob, dataLoader: DataLoaderWorker) => {
+  const {orgId, runAt, type} = job
+  const [organization, orgUsers] = await Promise.all([
+    dataLoader.get('organizations').load(orgId),
+    dataLoader.get('organizationUsersByOrgId').load(orgId)
+  ])
+  const {name: orgName, scheduledLockAt, lockedAt} = organization
+
+  // Skip the job if it's already been locked or the scheduled lock date changed
+  if (!scheduledLockAt || lockedAt || scheduledLockAt.getTime() !== runAt.getTime()) return
```
igorlesnenko, line 16: `scheduledLockAt.getTime() !== runAt.getTime()` 
This check wouldn't work as we have `runAt = scheduledLockAt.getTime() - ms('7d')`
Maybe we need to store `scheduledFirstWarningAt` value

igorlesnenko (changes requested):

packages/server/graphql/private/mutations/runScheduledJobs.ts
nickoferrall: Are there any jobs you have in mind that will be added soon? If so, I'm happy to cast the type, but if not, I think it's best to check for the `orgId`

packages/client/modules/email/components/LimitsEmails/LockedEmail.tsx
nickoferrall: Thanks - updated!

packages/server/database/types/processTeamsLimitsJob.ts
nickoferrall: Makes sense. I've updated this so that we have an `expectedRunAt` timestamp based on the job type

packages/client/modules/email/components/LimitsEmails/SevenDayWarningEmail.tsx
igorlesnenko: I personally feel that it is sounds good as a reminder 🤔

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -172,37 +163,18 @@ export const checkTeamsLimit = async (orgId: string, dataLoader: DataLoaderWorke
...
+      billingLeaders.map((billingLeader) =>
+        sendTeamsLimitEmail({
+          user: billingLeader,
+          orgId,
+          orgName,
+          emailType: 'thirtyDayWarning'
+        })
+      )
+    ]),
+      scheduleTeamLimitsJobs(scheduledLockAt, orgId)
```
igorlesnenko, line 178: Should it be outside of Promise.all?

packages/server/graphql/private/mutations/runScheduledJobs.ts
igorlesnenko: Make sense, no other jobs for now, could be kept as is and changed later if needed

packages/server/email/teamLimitsEmailCreator.tsx
```diff
@@ -0,0 +1,107 @@
...
+    `,
+    locked: `Hi ${preferredName} 👋
+
+    Unfortunately, ${orgName} has exceeded the two-team limit on the Starter Plan for more than 30 days, and your account has been deactivated.
+
+    Your teams will not be able to run meetings with Parabol, add new teams to your account, or access previous teams.
+
+    You can re-activate your teams by upgrading your account: ${billingURL}
+
+    If you have any questions feel free to contact us - we’re here to help.
```
igorlesnenko, line 55: Should we use a longer dash instead of "-"? For example "–' or "—"?

packages/server/email/teamLimitsEmailCreator.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+    Parabol Team
+  `,
+    sevenDayWarning: `Hi ${preferredName} 👋
+
+    This is a follow-up notification to remind you that ${orgName}'s Parabol account is at risk of deactivation because you’ve exceeded the two team limit on your Starter Plan.
+
+    Once your account is deactivated, you will lose access to your teams and won’t be able to run Retrospective, Sprint Poker or Standup meetings with Parabol.
+
+    You'll need to upgrade your account within the next 7 days to avoid losing access to your agile meetings: ${billingURL}
```
igorlesnenko, line 41: Should we use a variable instead of '7 days' ?

packages/server/email/teamLimitsEmailCreator.tsx
```diff
@@ -0,0 +1,107 @@
...
+  })
+
+  const emailTextLookup = {
+    thirtyDayWarning: `Hi ${preferredName} 👋
+
+    This is a friendly note to let you know that ${orgName} has officially reached ${Threshold.MAX_PERSONAL_TIER_TEAMS} active teams on Parabol - congrats! We love to see organizations finding value in Parabol and improving their teams in the process.
+
+    As a reminder: Parabol's Starter Plan has a limit of two teams. Please upgrade your account to continue using Parabol with all of your teams: ${billingURL}
+
+    If you aren’t ready to upgrade, then in 30 days your account will revert to the two teams allowed on the Starter Plan. Feel free to contact us with any questions - we’re here to help!
```
igorlesnenko, line 31: Should we use variable instead of '30 days' and variable instead of 'two' in the text?

igorlesnenko (approved):

packages/server/billing/helpers/teamLimitsCheck.ts
nickoferrall: Why is that?

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -172,37 +163,18 @@ export const checkTeamsLimit = async (orgId: string, dataLoader: DataLoaderWorke
...
-  // Skip the job if unlocked or already locked or scheduled lock date changed
-  if (
-    !organization.scheduledLockAt ||
-    organization.lockedAt ||
-    organization.scheduledLockAt.getTime() !== runAt.getTime()
-  ) {
-    return
+    await Promise.all([
+      enableUsageStats(billingLeadersIds, orgId),
+      sendWebsiteNotifications(orgId, billingLeadersIds, dataLoader),
```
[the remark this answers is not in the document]
nickoferrall: Would it be a problem if they get a notification slightly before the usage stats are enabled? I imagine that by the time they click on the notification, usage stats would be enabled

packages/server/email/teamLimitsEmailCreator.tsx
nickoferrall: Sounds good 👍

packages/server/email/teamLimitsEmailCreator.tsx
nickoferrall: Sounds good 👍

packages/server/email/teamLimitsEmailCreator.tsx
nickoferrall: Sure - updated

packages/server/billing/helpers/teamLimitsCheck.ts
igorlesnenko: @nickoferrall I mean it is currently outside of `Promise.all`, should we put it inside?

nickoferrall: Oh thanks! I thought it was already included. Update pushed

packages/server/billing/helpers/teamLimitsCheck.ts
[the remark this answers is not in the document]
igorlesnenko: I think if they get notification before, they will click and would see an error on usage stats page, however, I thin possibility is small and refreshing a page would fix it.

Dschoordsch (changes requested):
Did not test

packages/server/database/types/processTeamsLimitsJob.ts
```diff
@@ -0,0 +1,38 @@
...
+    dataLoader.get('organizationUsersByOrgId').load(orgId)
+  ])
+  const {name: orgName, scheduledLockAt, lockedAt} = organization
+
+  if (!scheduledLockAt || lockedAt) return
+
+  const expectedRunAt =
+    type === 'LOCK_ORGANIZATION' ? scheduledLockAt.getTime() : scheduledLockAt.getTime() - ms('7d')
+
+  if (expectedRunAt !== runAt.getTime()) return
```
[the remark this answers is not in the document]
nickoferrall: @igorlesnenko I think I wrote this following your example - I think we were checking if the `runAt` time is the same as `expectedRunAt` because the scheduled job may have been updated, e.g. if the user upgrades, but wouldn't we cancel the scheduled job instead of changing the `runAt` timestamp? I'll remove this bit for now

packages/server/email/teamLimitsEmailCreator.tsx
```diff
@@ -0,0 +1,107 @@
...
+  userId: string
+  preferredName: string
+  orgId: string
+  orgName: string
+  emailType: TeamLimitsEmailType
+}
+
+const teamLimitsEmailCreator = (props: Props) => {
+  const {userId, preferredName, orgId, emailType, orgName} = props
+  const Email =
```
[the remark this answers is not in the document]
nickoferrall: `Email` is a JSX variable which is why it's capitalised. This follows the naming convention, see https://github.com/ParabolInc/parabol/blob/1d6423b4b380c9d4980b6dee60aea93b86a546ec/packages/client/components/ActionMeeting.tsx#L53

packages/client/components/TopBarNotifications.tsx
```diff
@@ -32,6 +32,7 @@ const TopBarNotifications = ({queryRef}: Props) => {
               node {
                 id
                 status
+                type
```
Dschoordsch, line 35: -1 Somehow this was sneaking in recently, but it should not be here since it's not used in the file directly. Instead if it's needed somewhere else, then a fragment should be spread here. @igorlesnenko can this be removed here safely?

igorlesnenko, line 35: Needs to be deleted, looks like merge artifact

packages/server/graphql/private/typeDefs/_legacy.graphql
```diff
@@ -1049,6 +1049,397 @@ interface NewMeeting {
   locked: Boolean!
 }
 
+"""
+The user account profile
+"""
+type User {
```
igorlesnenko, line 1055: These changes needs to be removed, looks like merge artifact

Dschoordsch, line 1055: -1 We don't need to repeat public types in the private schema anymore. This only leads to code duplication. I will do a cleanup of all these duplicates soon, in the meantime please don't add any back.

packages/server/email/teamLimitsEmailCreator.tsx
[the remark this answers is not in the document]
Dschoordsch: You're right, my fault.

packages/server/database/types/processTeamsLimitsJob.ts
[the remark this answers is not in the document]
igorlesnenko: @nickoferrall Currently we don't cancel Jobs if user upgrades or fixes the limit. Shall we? 

This check might also prevent race condition:

- Job fetched from DB
- User upgraded
- Job executed

Thought it is a very rare case

What do you think?

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -175,7 +172,18 @@ export const checkTeamsLimit = async (orgId: string, dataLoader: DataLoaderWorke
...
+      billingLeaders.map((billingLeader) =>
+        sendTeamsLimitEmail({
+          user: billingLeader,
+          orgId,
+          orgName,
+          emailType: 'thirtyDayWarning'
+        })
+      ),
+      scheduleTeamLimitsJobs(scheduledLockAt, orgId)
```
Dschoordsch, line 185: -1 I missed this before, but this seems to be the wrong place. This limit is about teams limit, so why would we only send this when an organization has an `activeDomain` (i.e. qualifies for enterprise)?

In addition it seems to me like `scheduleTeamLimitsJobs` and `scheduleJobs` below should be merged.

Dschoordsch (changes requested):
Had some trouble with testing, the instructions are out of date and there seems to be a logic error in `teamLimitsCheck`

packages/client/components/TopBarNotifications.tsx
igorlesnenko: It could be removed, looks like some merge side effect

packages/server/billing/helpers/teamLimitsCheck.ts
igorlesnenko: I think we just need to move this check above everything.
Changed it here: https://github.com/ParabolInc/parabol/pull/7682/files/d2b3b5e426ec2537f2352ce3af49f4ec5d42db00#diff-0ebb6687f73dcbeef95db77a23ed552ac97457d45368062605a87c17330b9f9dR144

packages/server/database/types/processTeamsLimitsJob.ts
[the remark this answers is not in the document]
nickoferrall: @igorlesnenko yeah I think we'll need to cancel the job if the user upgrades, no? What currently happens to the job if the user upgrades?

packages/server/billing/helpers/teamLimitsCheck.ts
nickoferrall: @igorlesnenko you're thinking I add `if (!organization.activeDomain) return` to line 144?

igorlesnenko: @nickoferrall yes, basically do not run any limits checks if organization does not have a domain.

packages/server/database/types/processTeamsLimitsJob.ts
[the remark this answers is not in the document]
igorlesnenko: @nickoferrall Currently, nothing happens to the job if user upgrades. I thought that `expectedRunAt` check would "cancel" job when it starts. But we can remove job in a real time when user upgrades or deletes a team.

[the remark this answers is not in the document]
nickoferrall: As in this check: `if (expectedRunAt !== runAt.getTime()) return`?

Does the `expectedRunAt` date change when they upgrade? It seems more intuitive to me that the job is removed once the user upgrades, but I'm open to other ideas.

I've created an issue to capture this: https://github.com/ParabolInc/parabol/issues/7698

nickoferrall:
> Had some trouble with testing, the instructions are out of date and there seems to be a logic error in teamLimitsCheck

I've updated the description and testing steps. I ran through the tests myself and everything worked as expected

packages/server/billing/helpers/teamLimitsCheck.ts
Dschoordsch: You did not answer my question of **why** we want to have the team limit only for organizations with an `activeDomain`.

Dschoordsch (changes requested):

packages/server/billing/helpers/teamLimitsCheck.ts
nickoferrall: I don't see why we'd need to check `activeDomain` as, according to [this message](https://parabol.slack.com/archives/C042E8PGPD4/p1675267220136159?thread_ts=1675266998.778289&cid=C042E8PGPD4) 🔒, we have a Team tier team that doesn't have an active domain. 

I've removed this check, but @igorlesnenko please let me know if I'm missing something

packages/client/modules/email/components/LimitsEmails/LockedEmail.tsx
```diff
@@ -0,0 +1,78 @@
...
+          {`You can re-activate your teams by `}
+          <a style={linkStyle} href={billingURL}>
+            {`upgrading your account`}
+          </a>
+          {'.'}
+        </p>
+        <Button url={billingURL}>{'Reactivate Teams'}</Button>
+        <EmptySpace height={16} />
+        <p style={copyStyle}>
+          {`If you have any questions, or if you'd like to discuss enterprise plans, feel free to contact us —— `}
```
nickoferrall, line 64: I've added this note on enterprise plans following [this conversation](https://parabol.slack.com/archives/C042E8PGPD4/p1675292067447849?thread_ts=1675290895.962719&cid=C042E8PGPD4) 🔒 in Slack.

I've included this in the locked email and 7 day warning email, but not in the thirty day warning email as I think that email doesn't flow as well if we include an enterprise plans note

packages/server/billing/helpers/teamLimitsCheck.ts
igorlesnenko: @nickoferrall The check was added because of 2 reasons:

1) Currently, it is not possible to show usage stats for organizations without a domain.

I think for now we need to keep this check at least in `enableUsageStats`.

2) We want to exclude deals that is in progress on HubSpot. We use the domain to check this.

@tghanken is it possible to determine if the deal is in progress on HubSpot if we don't know an organization domain? In this case, can we assume that the deal is not in progress and safely trigger the limit warnings?

nickoferrall: Hmm but if we just add the check to `enableUsageStats`, wouldn't `sendWebsiteNotifications` still execute, which would trigger a notification telling the user to check their usage (see https://github.com/ParabolInc/parabol/pull/7601)?

Perhaps a temporary solution could be adding the `if (!organization.activeDomain) return` check again, but creating a new issue to either a) enable usage stats for teams without an active domain or b) show a different notification if there isn't an active domain?

Also, feel free to respond after your holiday!

tghanken: > is it possible to determine if the deal is in progress on HubSpot if we don't know an organization domain?

Organizations with a null `activeDomain` field will not have a corresponding company in HubSpot, so there is effectively no way a deal can automatically be created for an organization with no domain.  It feels like a safe assumption that any org with a null `activeDomain` will not have a deal in progress.  The only exception would be if a user on the domain initiates contact with our sales team.  I'll check with the team if we have a process for handling inbound sales requests from users on null domains.

> Perhaps a temporary solution could be adding the if (!organization.activeDomain) return check again, but creating a new issue to either a) enable usage stats for teams without an active domain or b) show a different notification if there isn't an active domain?

I think this is the way to go.  The safe, fast approach is to keep the check, and we can optimize for the small tail of qualified null `activeDomain` orgs after the fact.  Right now these orgs are a VERY small subset of our qualified orgs.

igorlesnenko: > Perhaps a temporary solution could be adding the if (!organization.activeDomain) return check again, but creating a new issue to either a) enable usage stats for teams without an active domain or b) show a different notification if there isn't an active domain?

Yes, I agree. We can run experiment only on organizations with active domain for now and improve later.

In the next new issue we also need a way to check for an active deal on HubSpot for orgs without a domain.

nickoferrall: Sounds good. I've created an issue here https://github.com/ParabolInc/parabol/issues/7723 and pushed an update with a comment linking to the issue

Dschoordsch (changes requested):
There seems to be a formatting issue (- -) with the locked email
Screenshot from 2023-02-09 11-47-36
Same for the 7 day warning

Screenshot from 2023-02-09 11-51-58

For the locked message it also appears in the text version

Screenshot from 2023-02-09 11-54-53
Screenshot from 2023-02-09 11-54-01

There is a unicode long dash if that is the intention.

In the locked message there is also "30 days, and" in both versions where I find the comma oddly placed. If that's supposed to be an Oxford comma, I think this wouldn't make sense for a 2 item list.

Otherwise it looks good and works as expected!

nickoferrall:
Interesting, I don't see that double dash on Chrome. Perhaps it's browser or OS-specific. I've pushed an update 👍 

For the comma, Grammarly suggests that we do need it:

comma

And as it was included in the original copy [here](https://docs.google.com/document/d/1XUGU2Ozhdapc4dSA2raNbi6gA0FexxUJRw2xlQQcxNM/edit#heading=h.wiffjptluzk9), I've left it in

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -1,5 +1,5 @@
 import ms from 'ms'
-import {Threshold} from 'parabol-client/types/constEnums'
+// import {Threshold} from 'parabol-client/types/constEnums'
```
Dschoordsch: -1 please revert testing change

Dschoordsch (changes requested):

packages/server/billing/helpers/teamLimitsCheck.ts
nickoferrall: @Dschoordsch I think you're looking at the penultimate commit. I changed that here: https://github.com/ParabolInc/parabol/pull/7637/commits/561b54982025b0f8752446b3fa0674faf02583f2

Dschoordsch (approved):
Sorry, you're right

PR #7645, opened 2023-01-11
feat(teams-limit): implement organization locked modal

### LINKED ISSUE #7638: Teams limit: implement organization locked modal
Prev issue #7386

AC

- When organization is locked, also lock all teams and display a modal similar to `UnpaidTeamModal`
- Modal should suggest to contact support for deletion of teams or upgrade (if billing leader)
- If limit is fixed, lock should be removed
- Upgrading organization should remove lock from all teams

igorlesnenko:
Fixes #7638

Billing leader:

image

Non billing leader:

image

Meeting creation error:

image

How to test

- Set `lockedAt` value on organization 
- See that all teams locked and modal is shown
- See that modal is different for billing leader and non billing leader user
- See that you can't create a new meeting
- See that you can still create a meeting on non-locked teams
- See that after upgrading org the lock is gone

### DIFF
--- a/packages/client/components/NewMeetingActions.tsx
+++ b/packages/client/components/NewMeetingActions.tsx
@@ -1,8 +1,10 @@
 import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
-import React from 'react'
+import React, {useEffect} from 'react'
 import {useFragment} from 'react-relay'
 import {NewMeetingActions_team$key} from '~/__generated__/NewMeetingActions_team.graphql'
+import useAtmosphere from '../hooks/useAtmosphere'
+import SendClientSegmentEventMutation from '../mutations/SendClientSegmentEventMutation'
 import {Breakpoint, Threshold} from '../types/constEnums'
 import FlatPrimaryButton from './FlatPrimaryButton'
 import NewMeetingActionsCurrentMeetings from './NewMeetingActionsCurrentMeetings'
@@ -72,6 +74,7 @@ const NewMeetingActions = (props: Props) => {
         ...NewMeetingActionsCurrentMeetings_team
         id
         organization {
+          id
           name
           lockedAt
         }
@@ -80,11 +83,22 @@ const NewMeetingActions = (props: Props) => {
     teamRef
   )
 
+  const atmosphere = useAtmosphere()
+
   const {organization} = team
 
-  const {lockedAt, name: organizationName} = organization
+  const {lockedAt, name: organizationName, id: orgId} = organization
 
   const isLocked = !!lockedAt
+
+  useEffect(() => {
+    if (isLocked) {
+      SendClientSegmentEventMutation(atmosphere, 'Upgrade CTA Viewed', {
+        upgradeCTALocation: 'startNewMeetingOrganizationLockedError',
+        orgId
+      })
+    }
+  }, [isLocked])
 
   return (
     <ActionRow>
--- a/packages/client/components/NewMeetingActions.tsx
+++ b/packages/client/components/NewMeetingActions.tsx
@@ -52,7 +52,9 @@ const StartButton = styled(FlatPrimaryButton)({
 })
 
 const ErrorBlock = styled(StyledError)({
-  paddingBottom: 24
+  paddingBottom: 24,
+  width: '100%',
+  textAlign: 'center'
 })
 
 interface Props {
--- a/packages/client/components/NewMeetingActions.tsx
+++ b/packages/client/components/NewMeetingActions.tsx
@@ -51,6 +51,10 @@ const StartButton = styled(FlatPrimaryButton)({
   height: 50
 })
 
+const ErrorBlock = styled(StyledError)({
+  paddingBottom: 24
+})
+
 interface Props {
   error?: {message: string}
   teamRef: NewMeetingActions_team$key
@@ -72,9 +76,9 @@ const NewMeetingActions = (props: Props) => {
 
   return (
     <ActionRow>
+      {error && <ErrorBlock>{error.message}</ErrorBlock>}
       <ActiveMeetingsBlock>
         <NewMeetingActionsCurrentMeetings team={team} />
-        {error && <StyledError>{error.message}</StyledError>}
       </ActiveMeetingsBlock>
       <ButtonBlock>
         <StartButton onClick={onStartMeetingClick} waiting={submitting}>
--- a/packages/client/components/NewMeetingActions.tsx
+++ b/packages/client/components/NewMeetingActions.tsx
@@ -3,7 +3,7 @@ import graphql from 'babel-plugin-relay/macro'
 import React from 'react'
 import {useFragment} from 'react-relay'
 import {NewMeetingActions_team$key} from '~/__generated__/NewMeetingActions_team.graphql'
-import {Breakpoint} from '../types/constEnums'
+import {Breakpoint, Threshold} from '../types/constEnums'
 import FlatPrimaryButton from './FlatPrimaryButton'
 import NewMeetingActionsCurrentMeetings from './NewMeetingActionsCurrentMeetings'
 import StyledError from './StyledError'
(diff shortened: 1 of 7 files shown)


acressall:
This looks good! One q: does this appear everywhere or only on specific screens? ie. if I am a team lead, can I navigate to team settings to delete a team myself?

igorlesnenko:
@acressall currently it works similar to unpaid modal, and will appear everywhere on any page related to the specific team. So deleting a team functionality is also blocked. 

image


As far as I understand, we want users to contact us by email, rather than deleting teams manually. Is this correct @avivapinchas ?

avivapinchas:
In future, I do think it would be nice to let users have the autonomy to do this themselves, but for now, I prefer we just ask them to contact us because: 
- Explaining how to delete teams is kind of confusing 
- Only leads can delete their teams, so they would need to coordinate this amongst themselves, which feels complicated - we don't have a way to do this on an account level in the app today 
- If they have to contact us, then we can learn more about what they're wanting and how much this is coming up 
- Frankly, it's just a smaller thing to build

So yes I think asking them to contact us to delete teams makes sense

igorlesnenko:
Also updated an error message when user tries to create meeting

image

igorlesnenko:
@acressall make sense, made the button disabled

image

acressall (approved):

packages/client/modules/teamDashboard/components/UnpaidTeamModal/UnpaidTeamModal.tsx
```diff
@@ -54,6 +67,44 @@ const UnpaidTeamModal = (props: Props) => {
...
+        <LockedAtContent>
+          Unfortunately, <strong>{organizationName}</strong> has exceeded the{' '}
+          {Threshold.MAX_PERSONAL_TIER_TEAMS} teams limit on the Starter Plan for more than{' '}
+          {Threshold.PERSONAL_TIER_LOCK_AFTER_DAYS} days, and your account has been locked.
+          <br />
+          <br />
+          You can re-activate your teams by upgrading your account.
+          <br />
+          <br />
+          If you’d like to keep using Parabol with Starter Plan, please{' '}
```
nickoferrall, line 93: +1 I think `using Parabol on the Starter Plan` would read better

packages/server/graphql/mutations/helpers/isStartMeetingLocked.ts
```diff
@@ -1,7 +1,16 @@
+import {Threshold} from '~/types/constEnums'
 import {DataLoaderWorker} from '../../graphql'
 
 const isStartMeetingLocked = async (teamId: string, dataLoader: DataLoaderWorker) => {
   const team = await dataLoader.get('teams').loadNonNull(teamId)
+  const organization = await dataLoader.get('organizations').load(team.orgId)
+
+  const {lockedAt, name: organizationName} = organization
+
+  if (lockedAt) {
```
nickoferrall, line 10: +1 I think this should come after the `lockMessageHTML`. That way, someone from sales can add a custom lock message even if `lockedAt` is true

nickoferrall (approved):

packages/client/modules/teamDashboard/components/UnpaidTeamModal/UnpaidTeamModal.tsx
```diff
@@ -54,6 +69,50 @@ const UnpaidTeamModal = (props: Props) => {
...
+  if (organization.lockedAt) {
+    return (
+      <DashModal>
+        <DialogTitle>{'Organization Locked'}</DialogTitle>
+        <LockedAtContent>
+          Unfortunately, <strong>{organizationName}</strong> has exceeded the{' '}
+          {Threshold.MAX_PERSONAL_TIER_TEAMS} teams limit on the Starter Plan for more than{' '}
+          {Threshold.PERSONAL_TIER_LOCK_AFTER_DAYS} days, and your account has been locked.
+          <br />
+          <br />
```
Dschoordsch, line 97: +1 having 2 `<br />` looks like we should put the blocks in `<p>` and style it accordingly

packages/client/modules/teamDashboard/components/UnpaidTeamModal/UnpaidTeamModal.tsx
```diff
@@ -54,6 +69,50 @@ const UnpaidTeamModal = (props: Props) => {
...
+          You can re-activate your teams by upgrading your account.
+          <br />
+          <br />
+          If you’d like to keep using Parabol on the Starter Plan, please{' '}
+          <ContactUsLink href={ExternalLinks.CONTACT} target='_blank' rel='noopener noreferrer'>
+            contact us
+          </ContactUsLink>{' '}
+          to let us know which teams you’d like to delete to fit within the two-team limit.
+          {isALeader && (
+            <StyledButton size='medium' onClick={() => goToBilling('organizationLockedModal')}>
```
Dschoordsch, line 107: +1 I don't like that non-billing users have no CTA button, not even a description to contact a billing leader

Dschoordsch (approved):
Is the change for the error message intentional? It looks a bit lost

Screenshot from 2023-01-16 15-06-26

packages/client/modules/teamDashboard/components/UnpaidTeamModal/UnpaidTeamModal.tsx
igorlesnenko: Updated, using the same message we use for the unpaid modal

image

packages/client/components/NewMeetingActions.tsx
```diff
@@ -65,19 +71,37 @@ const NewMeetingActions = (props: Props) => {
...
     <ActionRow>
+      {error && <ErrorBlock>{error.message}</ErrorBlock>}
+      {isLocked && (
+        <ErrorBlock>
+          Unfortunately, {organizationName} has exceeded the {Threshold.MAX_PERSONAL_TIER_TEAMS}{' '}
+          teams limit on the Starter Plan for more than {Threshold.PERSONAL_TIER_LOCK_AFTER_DAYS}{' '}
+          days, and your account has been locked. You can re-activate your teams by upgrading your
+          account.
+        </ErrorBlock>
+      )}
```
tghanken, line 99: -1 Can we add an `Upgrade CTA Viewed` event when this error block is returned?  We also need to add this location to `UpgradeCTALocationEnumType`.

packages/client/modules/teamDashboard/components/UnpaidTeamModal/UnpaidTeamModal.tsx
```diff
@@ -54,6 +72,52 @@ const UnpaidTeamModal = (props: Props) => {
...
+  const isALeader = billingLeaders.findIndex((leader) => leader.id === viewerId) !== -1
+
+  const goToBilling = (upgradeCTALocation: UpgradeCTALocationEnum) => {
+    SendClientSegmentEventMutation(atmosphere, 'Upgrade CTA Clicked', {
+      upgradeCTALocation
+    })
+    history.push(`/me/organizations/${orgId}`)
+  }
+
+  if (organization.lockedAt) {
```
tghanken, line 91: -1 Can we add an `Upgrade CTA Viewed` event here as well with `upgradeCTALocation` set?

tghanken (changes requested):
-1 I'd like to add in `Upgrade CTA Viewed` events to mirror how we're tracking our other conversion pipelines.

packages/client/components/NewMeetingActions.tsx
igorlesnenko: Added

packages/client/modules/teamDashboard/components/UnpaidTeamModal/UnpaidTeamModal.tsx
igorlesnenko: @tghanken added, can you check?

tghanken (approved):
LGTM!

Issue #7615, opened 2023-01-03
AI Summary: improve quality of summaries


nickoferrall:
Currently, summaries can have some formatting issues e.g. sentences can begin with an unnecessary comma.

![image](https://user-images.githubusercontent.com/39854876/210347152-df4d9401-5daa-40b7-b326-f970c8a7b055.png)

### AC

- [ ] Update the prompt and/or the way we pass text to the prompt to improve the quality of the summaries and reduce the formatting issues

PR #7646, opened 2023-01-11
chore(ai-summary): improve prompt

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7615

After following suggestion number 2 in this article (https://help.openai.com/en/articles/6654000-best-practices-for-prompt-engineering-with-openai-api), I'm getting better results from the summary.

Here is a comparison of the most recent Squad retro in prod vs the summaries with this new prompt. The first image is from prod and the second with the new prompt:

Example 1:
prod
new-prompt

Example 2:
prod-2
prompt-2

Example 3:
prod-3
prompt-3

### DIFF
--- a/packages/server/utils/OpenAIServerManager.ts
+++ b/packages/server/utils/OpenAIServerManager.ts
@@ -18,7 +18,6 @@ class OpenAIServerManager {
 
   async getSummary(text: string | string[]) {
     if (!this.openAIApi) return null
-
     try {
       const response = await this.openAIApi.createCompletion({
         model: 'text-davinci-003',
--- a/packages/server/utils/OpenAIServerManager.ts
+++ b/packages/server/utils/OpenAIServerManager.ts
@@ -18,10 +18,15 @@ class OpenAIServerManager {
 
   async getSummary(text: string | string[]) {
     if (!this.openAIApi) return null
+
     try {
       const response = await this.openAIApi.createCompletion({
         model: 'text-davinci-003',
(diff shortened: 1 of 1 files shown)


packages/server/utils/OpenAIServerManager.ts
```diff
@@ -21,7 +21,11 @@ class OpenAIServerManager {
     try {
       const response = await this.openAIApi.createCompletion({
         model: 'text-davinci-003',
-        prompt: `Summarize this for a second-grade student in one or two sentences: ${text}`,
+        prompt: `Below is a comma-separated list of text. Summarize the text for a second-grade student in one or two sentences.
```
Dschoordsch, line 24: +1 We could give it even more hints for summarising reflections:

> Below is a list of topics with bullet points. Summarize these for a second-grade student in one or two sentences:
> 
> What went well?
> - We improved the prompt quality
> 
> What was difficult?
> - getting everyone on the same page

Dschoordsch (approved):
Did not test, but your examples look good

packages/server/utils/OpenAIServerManager.ts
tianrunhe: FYI my original hack day [implementation](https://github.com/ParabolInc/parabol/blob/feat/gpt3Summaries/packages/client/utils/OpenAIManager.ts) on the prompt is similar to what Georg suggested.

nickoferrall: Sounds good - merging this PR for now so Georg can continue the ship and will play with the suggested prompt later

Issue #7647, opened 2023-01-11
AI Summary: remove AI comment from comment count


nickoferrall:
The AI-generated comment is currently being included in the comment count, at least in the meeting summary.

I noticed that when looking at the meeting summary images in this PR: https://github.com/ParabolInc/parabol/pull/7646

### AC

- [ ] Comment count does not include AI comments

PR #7648, opened 2023-01-11
fix(ai-summary): remove AI comment from comment count

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7647

### To test

- [ ] Create a retro meeting, and add two reflection groups, add comments to one reflection topic, and don't add comments to the other
- [ ] In the meeting summary, see that the comment count is correct. The reflection topic with no comments says "No comments"
- [ ] Go to the timeline and see the correct comment count

### DIFF
--- a/packages/server/dataloader/customLoaderMakers.ts
+++ b/packages/server/dataloader/customLoaderMakers.ts
@@ -82,7 +82,6 @@ export const commentCountByDiscussionId = (parent: RootDataLoader) => {
         r
           .table('Comment')
           .getAll(r.args(discussionIds as string[]), {index: 'discussionId'})
-
           .filter((row: RDatum) =>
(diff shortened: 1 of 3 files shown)


igorlesnenko (commented):
I see Ai summary in the summary, but don't see any AI comments
image

Am I missing something?

nickoferrall:
@igorlesnenko do you have the AI Summary feature flag on? Also, make sure you don't proceed through the Vote phase too quickly. It takes a couple of seconds to get the response from Open AI so if you click Next right away, you won't see the AI Comment

nickoferrall:
Thanks @igorlesnenko. I've pushed an update so hopefully you can see the AI comment in the thread now 👍

igorlesnenko (approved):
Now it works!

Issue #7501, opened 2022-11-29
AI Summary: limit access to feature on the Starter plan


nickoferrall:
We want to give free users access to the AI Summary feature so they can decide whether they'd like to upgrade to access it.

One option would be to show the summary for the first three "real" retros. A "real" retro could be a retro with three or more meeting members and five or more reflections. See discussion in Notion [here](https://www.notion.so/AI-summary-0db139fbccbe4a45a25c9b51437e4bda?d=b2d0b764f6bd4f2ab729a6cdad2dd85b#3d8837e325604e78913040b4c67b46a1) 🔒 .

### AC

- [ ] As a Squad, decide on what the right limit is
- [ ] Implement the limit

nickoferrall:
Will start by showing the summary in the first three “real” retros, a “real” retro being a retro with three or more meeting members and five or more reflections.

See Slack discussion: https://parabol.slack.com/archives/C042E8PGPD4/p1673457764994249

I'll add a field to the Team object to keep count of how many "real" retros a team has run

PR #7658, opened 2023-01-13
chore(ai-summary): limit access to AI summaries

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7501

Demo: https://www.loom.com/share/4e122fce510d48df87f68b114ef45a92

### To test

- [ ] Look at a team in pgadmin by right clicking on the Team table, opening Query Tool, and entering this query:
```
SELECT * FROM public."Team"
WHERE id = 'TEAM_ID';
```
- [ ] Create a retro, see the AI-generated summaries, and end the meeting. If the retro has 5 or more reflections and 3 or more meeting members, see that the `qualAIMeetingsCount` has increased by 1. Otherwise, it remains the same
- [ ] With a team on the Starter tier, update `qualAIMeetingsCount` to 3. Run a retro meeting and see that the AI summaries are no longer generated
- [ ] With a team on the Team tier, update `qualAIMeetingsCount` to 3. Run a retro meeting and see that they can still access their AI summaries


packages/server/graphql/mutations/helpers/generateWholeMeetingSummary.ts
```diff
@@ -33,7 +37,12 @@ const generateWholeMeetingSummary = async (
   if (contentToSummarize.length <= 1) return
   const [r, summary] = await Promise.all([getRethink(), manager.getSummary(contentToSummarize)])
   if (!summary) return
-  await r.table('NewMeeting').get(meetingId).update({summary}).run()
+  const [meeting] = await Promise.all([
+    dataLoader.get('newMeetings').load(meetingId),
+    r.table('NewMeeting').get(meetingId).update({summary}).run()
+  ])
+  // mutate the cache
+  meeting.summary = summary
```
nickoferrall, line 45: Alternatively, I could return the summary here and pass it to `updateQualAIMeetingsCount`, but I prefer to pass fewer params around and fetch the `meeting` with the `summary` in `updateQualAIMeetingsCount`

packages/server/postgres/migrations/1673629660481_addQualAIMeetingsCount.ts
```diff
@@ -0,0 +1,22 @@
...
+  `)
+  await client.end()
+}
+
+export async function down() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+  ALTER TABLE "qualAIMeetingsCount"
+  DROP COLUMN IF EXISTS "summary";
```
igorlesnenko, line 19: -1 Looks like a wrong column here

packages/server/graphql/mutations/helpers/generateDiscussionSummary.ts
```diff
@@ -5,19 +5,21 @@ import updateDiscussions from '../../../postgres/queries/updateDiscussions'
...
+  const {id: meetingId, endedAt, facilitatorUserId, teamId} = meeting
+  const [facilitator, comments, tasks, team] = await Promise.all([
     dataLoader.get('users').loadNonNull(facilitatorUserId),
     dataLoader.get('commentsByDiscussionId').load(discussionId),
-    dataLoader.get('tasksByDiscussionId').load(discussionId)
+    dataLoader.get('tasksByDiscussionId').load(discussionId),
+    dataLoader.get('teams').load(teamId)
   ])
-  if (!facilitator.featureFlags.includes('aiSummary')) return
+  if (!canAccessAISummary(team, facilitator.featureFlags)) return
```
igorlesnenko, line 22: As far as I understand, in most of cases a team can not access AI summary, should we add some checks before loading bunch of data in `Promise.all`?

packages/server/postgres/migrations/1673629660481_addQualAIMeetingsCount.ts
nickoferrall: Thanks, updated!

packages/server/graphql/mutations/helpers/generateDiscussionSummary.ts
nickoferrall: Yeah, I was thinking about this, but to check `canAccessAISummary` we need to grab the `facilitator` and the `team`. We could await those two fields, and if they pass the `canAccessAISummary` check, we grab the `comments` and `tasks` in the discussion thread, but as we're only getting them from one `discussionId`, they should be relatively cheap queries. 

It's possible that Rethink is super slow with one of the requests, but then separating the `comments` and `tasks` requests will normally mean a slightly slower response time for AI Summary users, which will hopefully be a majority of users soon. I don't have a strong preference but lean towards the current implementation. What do you think?

igorlesnenko: Make sense, then keeping it as is is fine I think

igorlesnenko (approved):

packages/server/graphql/mutations/helpers/generateWholeMeetingSummary.ts
Dschoordsch: -1 I'm not really happy with the load an object just to update the cache approach. If it wasn't loaded or isn't needed later anymore, then we've just wasted some bandwidth and time. I think I've mentioned this before, but I would prefer if we'd had something like
```
dataLoader.get('newMeetings').updateCache(meetingId, {summary})
```
which we could put in the `Promise.all` call and which would update the cache only if it's in there.

Marking it as -1 to avoid the current approach to become a pattern.

packages/server/graphql/mutations/helpers/updateQualAIMeetingsCount.ts
```diff
@@ -0,0 +1,23 @@
...
+    dataLoader.get('teams').load(teamId),
+    dataLoader.get('retroReflectionsByMeetingId').load(meetingId),
+    dataLoader.get('newMeetings').load(meetingId)
+  ])
+  if (meetingMembers.length < 3 || !team || !meeting.summary || reflections.length < 5) return
+  const {qualAIMeetingsCount} = team
+  const updates = {
+    qualAIMeetingsCount: qualAIMeetingsCount + 1
+  }
+  await updateTeamByTeamId(updates, teamId)
```
Dschoordsch, line 20: -1 since we load team by dataloader we should update or invalidate the cache here for it. We should do this in all cases where we aren't sure if the object will be used subsequently. Since this is a helper we don't know anything about the calling site.

packages/server/graphql/mutations/helpers/generateDiscussionSummary.ts
Dschoordsch: +1 I'm not so sure about this as this repeats in the other files. It looks like we wouldn't load most of the data if they don't qualify for AI summaries, also not in the context where this function is called. In this case I would err on the side of caution and load it only for qualified users.

Dschoordsch (changes requested):
Did not test

packages/server/graphql/mutations/helpers/updateQualAIMeetingsCount.ts
nickoferrall: Makes sense - updated!

packages/server/graphql/mutations/helpers/generateWholeMeetingSummary.ts
nickoferrall: I think that's a better approach but can be handled in a separate issue to help get this shipped. I've created an issue here: https://github.com/ParabolInc/parabol/issues/7687

Dschoordsch (approved):

Issue #7656, opened 2023-01-12
AI Summary: add whole meeting summary to Slack notification


nickoferrall:
Here's the Slack thread where the idea came from: https://parabol.slack.com/archives/C042E8PGPD4/p1673550771419099 🔒 

### AC

- [ ] When a retro ends, include the AI-generated whole meeting summary in the Slack notification

PR #7659, opened 2023-01-16
feat(ai-summary): add summary to end meeting notification 

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7656

New end meeting Slack notification example:
ai-summary-slack-notif

MS Teams: 
ms-teams

Mattermost:
mattermost-ai-summary

### Testing steps

- [ ] Integrate a team with Slack, Mattermost, and MS Teams. See this [Notion doc](https://www.notion.so/parabol/Add-Parabol-notifications-to-MS-Teams-a7b3f1f0b8c447d5b475072280d3e01f) 🔒 to learn how to add MS Teams (remember to add the feature flag)
- [ ] End a retro meeting that has an AI summary
- [ ] See that the end meeting notifications include the AI-generated summary


packages/server/graphql/mutations/helpers/generateWholeMeetingSummary.ts
```diff
@@ -33,7 +33,12 @@ const generateWholeMeetingSummary = async (
   if (contentToSummarize.length <= 1) return
   const [r, summary] = await Promise.all([getRethink(), manager.getSummary(contentToSummarize)])
   if (!summary) return
-  await r.table('NewMeeting').get(meetingId).update({summary}).run()
+  const [meeting] = await Promise.all([
+    dataLoader.get('newMeetings').load(meetingId),
+    r.table('NewMeeting').get(meetingId).update({summary}).run()
+  ])
+  // mutate the cache
+  meeting.summary = summary
```
nickoferrall, line 41: Need to update the cache for the notifications to have access to the summary

packages/server/graphql/mutations/helpers/notifications/MattermostNotifier.ts
```diff
@@ -178,37 +179,43 @@ const MattermostNotificationHelper: NotificationIntegrationHelper<MattermostNoti
...
+        value: team.name
+      },
+      {
+        short: true,
+        title: 'Meeting',
+        value: meeting.name
+      },
+      {
+        short: false,
+        title: 'Stats',
```
nickoferrall, line 200: The title was previously `Summary` but I thought it'd be confusing to have a `Summary` and an `AI Summary` so I changed it to `Stats`

igorlesnenko (approved):

packages/server/graphql/mutations/helpers/generateWholeMeetingSummary.ts
Dschoordsch: Not ideal, but probably the best. Let's add the load call to the `Promise.all` in l13.

As updating the cache is more and more common, it would be nice to extend the Dataloader with an `update` function where you can update the object (if it is loaded already).

packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
```diff
@@ -162,6 +162,21 @@ export const MSTeamsNotificationHelper: NotificationIntegrationHelper<MSTeamsNot
     const meetingEndedReviewColumn = new AdaptiveCards.Column()
     meetingEndedReviewColumn.width = 'auto'
 
+    if (summary) {
+      const aiSummaryColumnSet = new AdaptiveCards.ColumnSet()
+      const aiSummaryColumn = new AdaptiveCards.Column()
+      aiSummaryColumn.width = 'stretch'
+      const aiSummaryTitle = new AdaptiveCards.TextBlock('AI Summary 🤖')
+      aiSummaryTitle.weight = AdaptiveCards.TextWeight.Bolder
+      const cleanedAISummary = summary.replace(/(\r\n|\n|\r)/gm, '') // remove line breaks from the summary
```
Dschoordsch, line 171: -1 shouldn't line breaks be replaced with spaces?

packages/server/graphql/mutations/helpers/notifications/MattermostNotifier.ts
```diff
@@ -178,37 +179,43 @@ const MattermostNotificationHelper: NotificationIntegrationHelper<MattermostNoti
...
+        value: summaryText
+      }
+    ]
+    if (summary) {
+      const cleanedAISummary = summary.replace(/(\r\n|\n|\r)/gm, '') // remove line breaks from the summary
```
Dschoordsch, line 205: -1 line breaks should be replaced by a space

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
```diff
@@ -174,12 +174,17 @@ export const SlackSingleChannelNotifier: NotificationIntegrationHelper<SlackNoti
...
+    const blocks: Array<{type: string}> = [
       makeSection(title),
       makeSections([createTeamSectionContent(team), createMeetingSectionContent(meeting)]),
-      makeSection(summaryText),
-      makeEndMeetingButtons(meeting)
+      makeSection(summaryText)
     ]
+    if (meeting.summary) {
+      const aiSummaryTitle = 'AI Summary :robot_face:'
+      const summary = meeting.summary.replace(/(\r\n|\n|\r)/gm, '') // remove line breaks from the summary
```
Dschoordsch, line 184: -1 line breaks should be replaced by spaces

Dschoordsch (changes requested):
Did not test yet

packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
nickoferrall: The line breaks happen at the beginning of the summary, so we can just remove them, e.g. `\n\nThe text is saying that something is great and that it was done well.'`

Dschoordsch: There is no guarantee that the bot will reply this way. It can as well write multiple paragraphs where we wouldn't want to replace it with nothing. If it's about leading/trailing whitespace just call `.trim()`

Dschoordsch (changes requested):
Use `.trim()` instead of possibly merging paragraphs without any space.

packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
nickoferrall: Sounds good - updated!

packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
```diff
@@ -162,6 +162,21 @@ export const MSTeamsNotificationHelper: NotificationIntegrationHelper<MSTeamsNot
     const meetingEndedReviewColumn = new AdaptiveCards.Column()
     meetingEndedReviewColumn.width = 'auto'
 
+    if (summary) {
+      const aiSummaryColumnSet = new AdaptiveCards.ColumnSet()
+      const aiSummaryColumn = new AdaptiveCards.Column()
+      aiSummaryColumn.width = 'stretch'
+      const aiSummaryTitle = new AdaptiveCards.TextBlock('AI Summary 🤖')
+      aiSummaryTitle.weight = AdaptiveCards.TextWeight.Bolder
+      const trimmedAISummary = summary.trim()
```
Dschoordsch, line 171: +1 It would be nice if we could clean the summary before storing it, so we don't need to remember trimming it each time we're presenting it

Dschoordsch (approved):
Only tested Mattermost and MS Teams

packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
nickoferrall: I like it - I've moved the trim to the OpenAIServerManager 👍

Issue #7466, opened 2022-11-22
Missing prompts on Starfish Retrospective


garethaledavies:
The Starfish retrospective template is missing prompts. I know this has happened before for SaMoLo and perhaps some others. It would be great to get a fix for this, but also to understand how/why this keeps happening.

Here's the template with missing prompts:

<img width="903" alt="Screenshot 2022-11-22 at 14 50 14" src="https://user-images.githubusercontent.com/15236825/203330807-09b46d0b-c7e8-4586-81fd-7aaf791b1efe.png">

Here are the prompts to add:

Keep Doing 🌀: What behaviors are working and adding value?
Less of ➖: What are we over-doing or doing too much of?
More of ➕: What are we not taking enough advantage of?
Stop ⏹️: What isn't adding value or helping the team?
Start ⏯️: What new things should we explore or add to our team?

jordanh:
Wow. Super interesting. I agree with p2 label here. Let's figure this out

PR #7660, opened 2023-01-16
fix: missing prompts in starfish retro template

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7466

It looks like the Starfish template had missing prompts because it was using the duplicated prompt id [here](https://github.com/ParabolInc/parabol/issues/7466). It was inserted at the same time as the DAKI template, which also had the same bug. I've checked the other templates that were added, Lean Coffee and What Went Well, and those are working as expected.

### Testing steps

- [ ] Run the migration and see the template now includes the two missing prompts

### DIFF
--- a/packages/server/database/migrations/20230116165620-fixStarfishTemplate.ts
+++ b/packages/server/database/migrations/20230116165620-fixStarfishTemplate.ts
@@ -1,0 +1,40 @@
+import {R} from 'rethinkdb-ts'
+
+export const up = async function (r: R) {
+  const now = new Date()
+  const missingLessOfPrompt = {
+    createdAt: now,
+    description: 'What are we over-doing or doing too much of?',
+    groupColor: '#D9D916',
+    id: 'starfishTemplate:lessOfPrompt',
+    isActive: true,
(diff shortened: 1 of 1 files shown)


Dschoordsch (approved):

packages/server/database/migrations/20230116165620-fixStarfishTemplate.ts
```diff
@@ -0,0 +1,41 @@
...
+    isActive: true,
+    question: 'More Of ➕',
+    sortOrder: 2,
+    teamId: 'aGhostTeam',
+    templateId: 'starfishTemplate',
+    title: 'More Of ➕',
+    updatedAt: now
+  }
+
+  await r.table('ReflectPrompt').insert([missingLessOfPrompt, missingMoreOfPrompt]).run()
```
Dschoordsch, line 32: +1 what we should do here is
```ts
  await r.table('ReflectPrompt').insert([missingLessOfPrompt, missingMoreOfPrompt], {conflict: 'error'}).run()
```
This would have avoided the error in the past

PR #7671, opened 2023-01-18
feat(team-health): calculate & write sentiment scores for reflections & retro meetings

tianrunhe:
# Description

I revised the previous PR and stripped out the front-end stuff. I think we still need some design work on how to surface this data but I feel like we can move ahead with back-end implementation as it should be _safe-to-try_ ™️:
1. It uses the same Google Cloud Natural Language we are already using for entities extraction
2. With the data written into DB, we can start to look at them in our data warehouse
3. The additional cost is relatively cheap. According to the [price sheet](https://cloud.google.com/natural-language/pricing), the cost of sentiment analysis == the cost of entity analysis. Today we are spending ~$450 per month on entity analysis:

Screenshot 2023-03-08 at 08 36 21

 
## Demo
https://www.loom.com/share/491e7de866a44d849168b25f45c18edc

## Testing scenarios

- [ ] Reflections
  - Add a new reflection or update an existing reflection
  - Verify `sentimentScore` is written for those reflections in RethinkDB

- [ ] Retro meetings
  - Run through a retro meeting with some reflections
  - Verify `sentimentScore` is written for the meeting in RethinkDB

## Final checklist

### DIFF
--- a/packages/server/database/types/Reflection.ts
+++ b/packages/server/database/types/Reflection.ts
@@ -10,6 +10,7 @@ export interface ReflectionInput {
   content: string
   plaintextContent?: string // the plaintext version of content
   entities: GoogleAnalyzedEntity[]
+  sentimentScore?: number
   meetingId: string
   reactjis?: Reactji[]
   reflectionGroupId?: string
@@ -26,6 +27,7 @@ export default class Reflection {
   content: string
   plaintextContent: string
   entities: GoogleAnalyzedEntity[]
+  sentimentScore: number
   isActive: boolean
   meetingId: string
   reactjis: Reactji[]
@@ -40,6 +42,7 @@ export default class Reflection {
       createdAt,
       creatorId,
       entities,
+      sentimentScore,
       id,
       meetingId,
       reactjis,
@@ -55,6 +58,7 @@ export default class Reflection {
     this.content = content
     this.plaintextContent = plaintextContent || extractTextFromDraftString(content)
     this.entities = entities
+    this.sentimentScore = sentimentScore || 0.0
     this.isActive = true
     this.meetingId = meetingId
     this.reactjis = reactjis || []
--- a/packages/server/database/types/Reflection.ts
+++ b/packages/server/database/types/Reflection.ts
@@ -27,7 +27,7 @@ export default class Reflection {
   content: string
   plaintextContent: string
   entities: GoogleAnalyzedEntity[]
-  sentimentScore: number
+  sentimentScore?: number
   isActive: boolean
   meetingId: string
   reactjis: Reactji[]
@@ -58,7 +58,7 @@ export default class Reflection {
     this.content = content
     this.plaintextContent = plaintextContent || extractTextFromDraftString(content)
     this.entities = entities
-    this.sentimentScore = sentimentScore || 0.0
+    this.sentimentScore = sentimentScore
     this.isActive = true
     this.meetingId = meetingId
     this.reactjis = reactjis || []
--- a/packages/server/graphql/mutations/createReflection.ts
+++ b/packages/server/graphql/mutations/createReflection.ts
@@ -16,7 +16,7 @@ import standardError from '../../utils/standardError'
 import {GQLContext} from '../graphql'
 import CreateReflectionInput, {CreateReflectionInputType} from '../types/CreateReflectionInput'
 import CreateReflectionPayload from '../types/CreateReflectionPayload'
-import getReflectionEntities from './helpers/getReflectionEntities'
+import {getReflectionEntities, getReflectionSentimentScore} from './helpers/getReflectionEntities'
 
 export default {
   type: CreateReflectionPayload,
@@ -58,6 +58,7 @@ export default {
     // RESOLUTION
     const plaintextContent = extractTextFromDraftString(normalizedContent)
     const entities = await getReflectionEntities(plaintextContent)
+    const sentimentScore = await getReflectionSentimentScore(plaintextContent)
     const reflectionGroupId = generateUID()
 
     const reflection = new Reflection({
@@ -65,6 +66,7 @@ export default {
       content: normalizedContent,
       plaintextContent,
       entities,
+      sentimentScore,
       meetingId,
       promptId,
       reflectionGroupId,
--- a/packages/server/graphql/mutations/createReflection.ts
+++ b/packages/server/graphql/mutations/createReflection.ts
@@ -43,6 +43,7 @@ export default {
     if (!reflectPrompt) {
       return standardError(new Error('Category not found'), {userId: viewerId})
     }
+    const {question} = reflectPrompt
     const meeting = await r.table('NewMeeting').get(meetingId).default(null).run()
     if (!meeting) return standardError(new Error('Meeting not found'), {userId: viewerId})
     const {endedAt, phases, teamId} = meeting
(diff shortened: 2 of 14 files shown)


nickoferrall (approved):
This looks great! Excited to have this data, and I'm loving your many hack projects 

Out of interest, did you opt for the Google API because we already had that pattern in place so it was faster to build? I wonder how the performance and price compare with OpenAI 🤔

tianrunhe:
> Out of interest, did you opt for the Google API because we already had that pattern in place so it was faster to build? I wonder how the performance and price compare with OpenAI 🤔

I did think about these questions!

From implementation perspective, yes Google NLP is an existing pattern but also I like the closed-form API format: you ask some text, it gives you a number. OpenAI could be more powerful but in the mean time it's also more open-ended so the response is unpredictable. We need to be very specific on the prompt to get what we need but it's still hard to parse the response, e.g.:

Screenshot 2023-03-09 at 10 57 39

From cost perspective, Google NLP is also cheaper! Here's the comparison ([assuming 4.7 characters per word](https://www.google.com/search?q=average+character+per+word&rlz=1C5CHFA_enUS1042US1042&oq=aver&aqs=chrome.1.69i57j69i59j46i131i199i433i465i512j0i433i512j0i433j0i433i512j0i131i433i512j46i131i199i433i465j0i433i512j0i271.2318j1j7&sourceid=chrome&ie=UTF-8)) 

Google NLP ([price sheet](https://cloud.google.com/natural-language/pricing)):
1 Unit ~= 1000 chars => 1 Unit ~= 212 words
$1 / 1000 Units => $1 / 212,000 words => $0.00472 / 1K words

OpenAI ([price sheet](https://openai.com/pricing)):
1K tokens ~= 750 words
Davinci(the one we are using now for AI summary): $0.02 / 1K tokens => $0.02 / 750 words => 0.02666667 / 1K words

nickoferrall:
Very interesting! Thanks for sharing

packages/server/graphql/mutations/helpers/generateWholeMeetingSentimentScore.ts
```diff
@@ -2,14 +2,9 @@ import {DataLoaderWorker} from '../../graphql'
...
 const generateWholeMeetingSentimentScore = async (
   meetingId: string,
-  facilitatorUserId: string,
   dataLoader: DataLoaderWorker
 ) => {
-  const [facilitator, reflections] = await Promise.all([
-    dataLoader.get('users').loadNonNull(facilitatorUserId),
-    dataLoader.get('retroReflectionsByMeetingId').load(meetingId)
-  ])
-  if (!facilitator.featureFlags.includes('aiSummary')) return undefined
```
tianrunhe: I removed this check with the thinking:
1. It's not expensive at all to calculate an average score
2. We are just writing that score to DB. If there would be any limit in the future, it'd be on the front-end where we decide when we can surface this data.

nickoferrall: @tianrunhe my concern is some users don't want their data to be used by AI, so I think it would be safer to leave the feature flag check in here. 

Ideally, we roll out the AI Summary feature soon and then invert the feature flag, so we would just have a small exempt list of users/orgs

nickoferrall (changes requested):
I think we should keep the feature flag check here

packages/server/graphql/mutations/helpers/generateWholeMeetingSentimentScore.ts
tianrunhe: Fair enough!

nickoferrall (approved):

packages/server/database/types/Reflection.ts
```diff
@@ -55,6 +58,7 @@ export default class Reflection {
     this.content = content
     this.plaintextContent = plaintextContent || extractTextFromDraftString(content)
     this.entities = entities
+    this.sentimentScore = sentimentScore || 0.0
```
Dschoordsch, line 61: -1 I think we should distinguish neutral meetings from having no data, i.e. leave it undefined if we don't have a score.

packages/server/graphql/mutations/createReflection.ts
```diff
@@ -57,14 +58,18 @@ export default {
 
     // RESOLUTION
     const plaintextContent = extractTextFromDraftString(normalizedContent)
-    const entities = await getReflectionEntities(plaintextContent)
+    const [entities, sentimentScore] = await Promise.all([
+      getReflectionEntities(plaintextContent),
+      getReflectionSentimentScore(plaintextContent)
```
Dschoordsch, line 63: +1 did you try combining both entity and sentiment analyzis? https://cloud.google.com/natural-language/docs/analyzing-entity-sentiment
It doesn't make a cost difference, but would save one roundtrip.

packages/server/graphql/mutations/helpers/getReflectionSentimentScore.ts
```diff
@@ -0,0 +1,12 @@
+import getGoogleLanguageManager from '../../../getGoogleLanguageManager'
+import manageGoogleNLPErrorResponse from './manageGoogleNLPErrorResponse'
+
+const getReflectionSentimentScore = async (plaintextContent: string) => {
+  if (!plaintextContent) return 0.0
```
Dschoordsch, line 5: -1 It makes sense for a quick prototype, but because we're storing these long term to present the data nicely, let's distinguish neutral score from no data

packages/server/graphql/types/NewMeeting.ts
```diff
@@ -154,6 +155,10 @@ export const newMeetingFields = () => ({
     type: GraphQLString,
     description: `The GPT-3 generated summary of all the content in the meeting, such as reflections, tasks, and comments. Undefined if the user doesnt have access to the feature or it's unavailable in this meeting type`
   },
+  sentimentScore: {
+    type: GraphQLFloat,
+    description: `The overall sentiment score for the meeting. Undefined if the user doesnt have access to the feature or it's unavailable in this meeting type`
```
Dschoordsch, line 160: +1 Range of values would be nice to know here

Dschoordsch (changes requested):

packages/server/graphql/mutations/createReflection.ts
tianrunhe: I think that API returns entities + **entity-level** sentiment score, whereas here we'd like a sentiment score for the entire document. The aggregations of the score from each entity doesn't necessarily represent the score for the entire document, e.g.

Screenshot 2023-03-13 at 11 12 46

Dschoordsch (approved):
2 thoughts:
1) we could make this a premium feature and only run the sentiment analysis retrospectively once they upgrade to avoid the cost for the starter tier
1.1) we could use a free offline library for the free tier, sth. like https://winkjs.org/wink-sentiment/, not as good but free and works in air-gapped env.
2) I think we should also consider the retrospective groups for the analysis, see for example
Screenshot from 2023-03-14 12-50-53
which gives scores of 0 and -0.7 while effectively being the same

tianrunhe:
@Dschoordsch I like your thoughts so I have made two more changes:
1. I combined the prompt question with the reflection text when calling sentiment analysis
2. Realizing 👆 this change will increase our API cost and in amid of the uncertain financial scenario, I limited the call to only team and enterprise users

packages/server/graphql/mutations/updateReflectionContent.ts
```diff
@@ -43,12 +44,19 @@ export default {
     if (!reflection) {
       return standardError(new Error('Reflection not found'), {userId: viewerId})
     }
-    const {creatorId, meetingId, reflectionGroupId} = reflection
+    const {creatorId, meetingId, reflectionGroupId, promptId} = reflection
+    const reflectPrompt = await dataLoader.get('reflectPrompts').load(promptId)
+    if (!reflectPrompt) {
+      return standardError(new Error('Category not found'), {userId: viewerId})
+    }
```
Dschoordsch, line 51: +1 As we're not expecting this to happen, you could just call `loadNonNull`

PR #7673, opened 2023-01-18
fix: delete empty reflection groups

### LINKED ISSUE #7669: Delete empty RetroReflectionGroup rows
RethinkDB Size:
We have about 5 million rows in the reflection group table
About 2 million of those are empty groups
That means we could save about 800MB if we nuke those rows
I’ve tested out deleting them in staging, nothing bad happens

I'm gonna go ahead & wipe them as well as put a PR in place to delete them when a retro meeting ends so they don't build up.

mattkrick:
# Description

Fixes #7669 frees up ~800MB

## Demo

Running this on staging the read query took about 80 seconds. When done in batches, it took ~7 seconds for 10k records, so i figured a single query is probably better.


nickoferrall (approved):
Tested ending a retro and it LGTM!

PR #7675, opened 2023-01-19
Fix/6674/fix meetings with invalid facilitator stage

### LINKED ISSUE #6674: No one can join a sprint poker meeting
## Issue - Bug

I spent quite a bit of time trying to figure out why in some rare cases users can't join the sprint poker meeting by clicking a meeting card in the dashboard. It is possible to join the meeting by providing the URL of a specific phase, but nothing is really clickable then anyway. More context [here.](https://parabol.slack.com/archives/C4JAUUZ9P/p1654117050738069)

The root cause is `facilitatorStageId` is set to the id of a non-existing stage. I haven't found the exact race condition that could cause this. Sprint poker is a specific meeting where everyone who participates can modify the scope. This allows for the following case: 

1. Alice is a facilitator, adds issue X to the scope, and advances the meeting to the selected issue phase (and possibly leaves the meeting)
2. Bob joins the meeting and removes issue X from the scope. 

I fixed one of the issues here: https://github.com/ParabolInc/parabol/pull/6667. everything else seems to be working correctly as we update the `facilitatorStageId` when that cases happens: https://github.com/ParabolInc/parabol/blob/master/packages/server/graphql/mutations/updatePokerScope.ts#L95 

I'm not sure how many times this happens to our users as we don't log this. When this happens https://github.com/ParabolInc/parabol/blob/41dd54baf5adfba1b84134fb78bd4645ff34ec49/packages/client/hooks/useInitialSafeRoute.ts#L72 returns `/` and user is redirected back to the dashboard. Unless there's something obvious I've missed I think it'd be worth to add sentry logging this case to see if it's worth to further investigate it. 

- [ ] Add logging when we throw users back the dashboard from the meeting view

Dschoordsch:
# Description

Fixes #6674
Add a fallback for cases when the facilitator stage is invalid
    
The facilitatorStageId can get invalid at times if the right
circumstances happen (facilitator moving while someone else removes a
stage etc.). Without this fix the client would not open the meeting
anymore because it always redirects to `/`. Now it falls back to the
first open stage or the last closed stage that's navigable.
    
The end mutations now also continue even if the facilitator stage is
invalid.

## Demo

## Testing scenarios

- [ ] Scenario A
  - Step 1
  - Step 2...

- [ ] Scenario B
  - Step 1
  - Step 2....

## Final checklist


BartoszJarocki (approved):
lgtm!

PR #7677, opened 2023-01-20
feat(teams-limit): add 7 days reminder notification and snackbar

### LINKED ISSUE #7568: Teams limit: add missing segment events
AC:

- Check if we have all necessary data for following the Measurement Plan 

https://www.notion.so/parabol/Implement-time-bound-stick-for-enforcing-2-team-limit-in-Usage-Stats-7b6cb3446267449f9269cbc2aa944ae8

- Add missing events if needed

### LINKED ISSUE #7678: Teams limit: add 7 days reminder notification and snackbar
image

AC:

- 7 days reminder snackbar and notification added 
- Add for billing leaders only

igorlesnenko:
# Description

Fixes #7678

Fixes #7568

## Demo

image

image

## Testing scenarios

- [ ] enable feature flag on specific org:

```
  .getAll(r.args(['jWQU6W39S0']), {index: 'id'})
  .update({
    tierLimitExceededAt: null,
    scheduledLockAt: null,
    lockedAt: null
});

```
- [ ] uncomment variables (search `const enum Threshold {`)
- [ ] run and complete a meeting on the org
- [ ] run a scheduled jobs http://localhost:3000/admin/graphql 

```
mutation {
 runScheduledJobs(seconds: 5000) 
}
```

- [ ] see snackbar appeared for billing leaders
- [ ] Refresh a page, see snackbar is not gone
- [ ] Close snackbar see notification is marked as read
- [ ] Open notification - see notification exists and read


packages/client/mutations/toasts/mapTeamsLimitReminderToToast.ts
```diff
@@ -0,0 +1,43 @@
...
+    autoDismiss: 0,
+    key: `newNotification:${notificationId}`,
+    message: `"${orgName}" is over the limit of ${
+      Threshold.MAX_STARTER_TIER_TEAMS
+    } Free Teams. Your free access will end on ${makeDateString(scheduledLockAt)}`,
+    action: {
+      label: 'Upgrade',
+      callback: () => {
+        history.push(`/me/organizations/${orgId}`)
+        SendClientSegmentEventMutation(atmosphere, 'Clicked teams limit reminder snackbar')
```
tghanken, line 37: -1 Can we change this to a `Snackbar Clicked` event with a `snackbarType: 'teamLimits'` property?

packages/client/components/TeamsLimitReminderNotification.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const {history} = useRouter()
+  const {organization, scheduledLockAt} = notification
+  const {name: orgName, picture: orgPicture, id: orgId} = organization
+
+  const onActionClick = () => {
+    SendClientSegmentEventMutation(atmosphere, 'Notification Clicked', {
+      notificationType: 'TEAMS_LIMIT_REMINDER'
+    })
+    history.push(`/me/organizations/${orgId}`)
+  }
```
tghanken, line 45: -1 Since this is an explicit upgrade CTA, we should also emit `Upgrade CTA Viewed` and `Upgrade CTA Clicked` events here.

tghanken (changes requested):

packages/client/mutations/toasts/mapTeamsLimitReminderToToast.ts
igorlesnenko: @tghanken we have 2 possible actions on a snackbar.

1) You can click on it anywhere and it will be dismissed
2) You can click on the upgrade button, you will be redirected to the upgrade page and snackbar will be dismissed

I guess in this case we need 3 events
`Upgrade CTA Viewed` and `Upgrade CTA Clicked` and `Snackbar Clicked` ?

igorlesnenko: @tghanken refactored events a bit, can you check?

packages/client/components/NotificationDropdown.tsx
```diff
@@ -80,12 +82,15 @@ const NotificationDropdown = (props: Props) => {
...
       )}
       {edges.map(({node}) => {
-        const {id: notificationId, status} = node
+        const {id: notificationId, status, type} = node
         const onViewFn = () => {
           SetNotificationStatusMutation(atmosphere, {notificationId, status: 'READ'}, {})
         }
         const onClickFn = () => {
           SetNotificationStatusMutation(atmosphere, {notificationId, status: 'CLICKED'}, {})
+          SendClientSegmentEventMutation(atmosphere, 'Notification Clicked', {
```
igorlesnenko, line 91: Just added a generic event for all notifications

packages/client/components/NewMeetingSidebarUpgradeBlock.tsx
```diff
@@ -59,14 +61,21 @@ const UpgradeButton = styled(FlatPrimaryButton)({
...
 
 const NewMeetingSidebarUpgradeBlock = (props: Props) => {
-  const {onClick, orgId} = props
+  const {onClick, orgId, meetingId} = props
   const history = useHistory()
   const [closed, setClosed] = React.useState(false)
+  const atmosphere = useAtmosphere()
 
   const handleUpgradeClick = () => {
+    SendClientSegmentEventMutation(atmosphere, 'Upgrade CTA Clicked', {
```
igorlesnenko, line 74: Added missing events here too

tghanken (approved):
LGTM!

nickoferrall (approved):

packages/server/graphql/private/typeDefs/_legacy.graphql
```diff
@@ -1049,397 +1049,6 @@ interface NewMeeting {
   locked: Boolean!
 }
 
-"""
```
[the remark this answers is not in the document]
igorlesnenko: @nickoferrall I removed it in previous PR, looks like because of some merge issues, you added it back in your PR.
I removed it because now we merge public types with private. And instead of doing modifications in 2 places I just removed User type completely from here

packages/server/graphql/public/types/NotifyTeamsLimitReminder.ts
```diff
@@ -0,0 +1,10 @@
+import {NotifyTeamsLimitReminderResolvers} from '../resolverTypes'
+
+const NotifyTeamsLimitReminder: NotifyTeamsLimitReminderResolvers = {
+  __isTypeOf: ({type}) => type === 'TEAMS_LIMIT_REMINDER',
+  organization: ({orgId}, _args, {dataLoader}) => {
```
Dschoordsch, line 5: -1 I have slight concerns here: We're not checking organization membership here and in fact once the notification is sent, I can remove the user from the org and they will still see the organization name, meaning they could have access to the org through this member.
Because we're only allowing predefined queries, it's not that big of a deal, but we should set a good example here:
* add a rule in https://github.com/ParabolInc/parabol/blob/83a4a39fb5fab61787ee3e432d6e75d7bec37982/packages/server/graphql/public/permissions.ts#L25 for this field
* (optional) modify `removeOrgUser` to also remove notifications with the `orgId`

Dschoordsch (changes requested):

packages/server/graphql/public/types/NotifyTeamsLimitReminder.ts
Dschoordsch: I think we can do sth. like
```
Organization: {
  '*': or(isSuperUser, isViewerOrgMember)
}
```
and create the `isViewerOrgMember` rule.

Because that's not an issue right now, we can also do it in a separate ticket, but should add it to this sprint.

igorlesnenko: @Dschoordsch What if we store all necessary information for notification as hardcoded values inside a notification object?
I think we can consider notifications as text messages which should not change dynamically.

Dschoordsch: Fair enough. Then an organization name change would not be reflected, but I think that's not a real life issue

Dschoordsch (approved):
Did not test again

PR #7679, opened 2023-01-20
chore: Migrate MeetingTemplate to PG (Phase 1 of 3)

mattkrick:
# Description

Squad 3 is doing some work on the MeetingTemplate table, so I figured we should move it to PG to minimize throwaway work. This refactor isn't adding any extra functionality, we'll worry about that that after phase 3 is complete. I hope to use this migration as a working example that we can base other table migrations off of. If you see any way we can improve it, let me know!

Scope
- Adds a MeetingTemplate table to postgres
- Every write to the MeetingTemplate table now happens in both RethinkDB and PG
- If the write fails in PG (e.g. it can't find the document because that row only exists in rethinkDB), it fails silently. 
- Effectively, the only writes that will succeed are inserts & updates on those brand new records
- We don't copy the data from RethinkDB to PG in this step! Why not? see example below

Example of how a data integrity error could be caused if we migrated existing data during this step:
- In RethinkDB, the following record exists: `{id: 123, name: 'foo'}`
- A migration copies that record to postgres
- The old graphql server only writes to RethinkDB. After the migration, **but before the graphql server gets shut down**, it updates the name to `'bar'`. Now in RethinkDB we have `{id: 123, name: 'bar'}` but in pg we still have `{id: 123, name: 'foo'}`. 
- The chances of this happening are slim. The window is typically < 10 seconds. Better to be safe than spend all week updating broken records! 
- By waiting 1 deployment cycle, we guarantee that the graphql server knows to send the update to both databases, ensuring consistency. 
- There's still a risk for PPMIs since they might batch updates together! We upgrade these app during times on low use, so risk is minimal

## Testing scenarios

- [x] updateScope
- [x] updateOrgId
- [x] update name
- [x] update lastUsedAt
- [x] insert
- [x] delete template
- [x] archive team

These tests will have to wait until phase 2 since they require the data to be migrated first:
- [ ] free status by name
- [ ] free status by id

This DB constraint will have to wait till Phase 2 as well:
- `parentTemplateId -> REFERENCES "MeetingTemplate"("id") ON DELETE SET NULL`


packages/server/dataloader/customLoaderMakers.ts
```diff
@@ -406,6 +406,7 @@ export const meetingTemplatesByType = (parent: RootDataLoader) => {
       const resultsByType = await Promise.all(
         entries.map((entry) => {
           const [meetingType, teamIds] = entry
+          // Will convert to PG by Mar 1, 2023
```
mattkrick, line 409: flagging all table reads with the same comment to alert other devs & make it easier to find instances later

packages/server/graphql/private/types/SetIsFreeMeetingTemplateSuccess.ts
```diff
@@ -1,17 +1,15 @@
...
 }
 
 const SetIsFreeMeetingTemplateSuccess: SetIsFreeMeetingTemplateSuccessResolvers = {
-  updatedTemplates: async ({updatedTemplateIds}) => {
-    const r = await getRethink()
-    const updatedTemplates = await r
-      .table('MeetingTemplate')
-      .getAll(r.args(updatedTemplateIds))
-      .run()
+  updatedTemplates: async ({updatedTemplateIds}, _args, {dataLoader}) => {
```
mattkrick, line 9: this isn't a write, but i changed it anyways since we don't want to be reading directly from the DB in a mutation payload resolver.

packages/server/postgres/migrations/1674237759799_meetingTemplates.ts
```diff
@@ -0,0 +1,47 @@
+import {Client} from 'pg'
+import getPgConfig from '../getPgConfig'
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+    DROP TYPE IF EXISTS "SharingScopeEnum";
+    CREATE TYPE "SharingScopeEnum" AS ENUM (
+      'USER',
```
mattkrick, line 10: I added User to the ReBAC model here even though we don't use it yet. The goal is to use the same `SharingScopeEnum ` across many different entities in our system

packages/server/postgres/queries/insertMeetingTemplate.ts
```diff
@@ -0,0 +1,18 @@
+import getRethink from '../../database/rethinkDriver'
+import MeetingTemplate from '../../database/types/MeetingTemplate'
+import getPg from '../getPg'
+
+const insertMeetingTemplate = async (meetingTemplate: MeetingTemplate) => {
+  const r = await getRethink()
+  const pg = getPg()
+  const {id, name, teamId, orgId, parentTemplateId, type} = meetingTemplate
+  await Promise.allSettled([
```
mattkrick, line 9: the `allSettled` pattern seemed cleaner than a try/catch because we almost never return a value from a mutation. In Phase 3 when we stop writes to RethinkDB we can just delete a couple lines & we're done!

packages/server/postgres/queries/updateMeetingTemplateFreeStatusByName.ts
```diff
@@ -0,0 +1,28 @@
...
+      .getAll('aGhostTeam', {index: 'teamId'})
+      .filter((row) => r.expr(templateNames).contains(row('name')))
+      .update(
+        {isFree, updatedAt: now},
+        {returnChanges: true}
+      )('changes')('new_val')('id')
+      .default([])
+      .run(),
+    pg.query(
+      `UPDATE "MeetingTemplate" SET "isFree" = $1 WHERE "teamId" = 'aGhostTeam' AND name LIKE ANY($2) RETURNING id;`,
```
mattkrick, line 20: UNTESTED! requires copying the public templates to PG first, which we'll do in phase 2

packages/server/postgres/queries/insertMeetingTemplate.ts
Dschoordsch: -1 I'm not sure I can follow here. This would just silently ignore all errors, wouldn't it?

packages/server/postgres/queries/insertMeetingTemplate.ts
```diff
@@ -0,0 +1,18 @@
...
+    r.table('MeetingTemplate').insert(meetingTemplate).run(),
+    pg.query(
+      `INSERT INTO "MeetingTemplate" (id, name, "teamId", "orgId", "parentTemplateId", type) VALUES ($1, $2, $3, $4, $5, $6)`,
```
Dschoordsch, line 12: -1 Both entries would differ in `createdAt` and `updatedAt` date. Also in theory `meetingTemplate` could contain a non-default `scope`, `lastUsedAt`, `isStarter` and `isFree` field.

Dschoordsch (changes requested):
I'm not sure why we would ignore all db errors now.

For the issue with running all migrations at once: If we implement the whole migration in a short time period there is the risk that we reach a point where we delete the rethink table. Because we run all rethink migrations before the postgres migrations it could happen that we delete the table before copying the data to postgres. So basically copying to postgres needs to be a rethink migration.
It would be nice to have some more control over migrations, like only running a rethink migration if a certain postgres migration already ran, basically some synchronization points between these two. Ideally we would just run all at once intermingled by timestamp.

packages/server/dataloader/customLoaderMakers.ts
mattkrick: I'm treating older migrations as immutable & we'll just fix things up in a new migration, unless I'm missing something?

packages/server/postgres/queries/insertMeetingTemplate.ts
mattkrick: > RE: `createdAt, updatedAt`

yep, timestamps are gonna be off by <5ms. When we check for equality in phase 2 we'll have to account for that clock shift, it shouldn't be a problem.

> RE extra fields

You're absolutely correct! We currently don't allow setting those fields during creation & I didn't want to add any extra functionality during this refactor, but can see that it might act as a footgun if I don't. Happy to do it either way as it's pretty minimally invasive, just let me know what you think is best.

mattkrick:
>For the issue with running all migrations at once: If we implement the whole migration in a short time period there is the risk that we reach a point where we delete the rethink table. Because we run all rethink migrations before the postgres migrations it could happen that we delete the table before copying the data to postgres. So basically copying to postgres needs to be a rethink migration.

I'm not sure I follow... we run all legacy rethinkdb migrations, then all migrations in the postgres directory. This migration is in the postgres directory, so if we decide to delete a RethinkDB table, it should be OK, unless I'm missing something?

packages/server/dataloader/customLoaderMakers.ts
Dschoordsch: I think it's fine. We only need to be careful how we delete the table. That's what I tried to describe. We **cannot** do
**rethink migrations**
* create `MeetingTemplate`
* multiple fill `MeetingTemplate`
* delete `MeetingTemplate`

**postgres migrations**
* create `MeetingTemplate`
* copy rethink `MeetingTemplate` data over

but instead we **need** to do
**rethink migrations**
* create `MeetingTemplate`
* multiple fill `MeetingTemplate`

**postgres migrations**
* create `MeetingTemplate`
* copy rethink `MeetingTemplate` data over
* delete rethink `MeetingTemplate`

mattkrick: ohhh! yep, i get what you're saying. 💯  the `database/migrations` can no longer be touched, all migrations should live in `postgres/migrations`, even if they only touch rethinkdb in order to assure linearity. perhaps I should get rid of the rethinkdb `create migration` function to ensure it 😜

packages/server/postgres/queries/insertMeetingTemplate.ts
mattkrick: We can't `Promise.all` because then it'll throw if either promise rejects. We can treat it as 2 separate promises, that's what I did in a previous commit:
 https://github.com/ParabolInc/parabol/pull/7679/commits/bd580a74ab72ad930454f1b06a7da189c4dcf75a#diff-b5b9e532d5e86ec912acda5064778d79f5c9ce4390f66acc77b5720882b5143aL8-L26
 
 I didn't like that very much because it means we have to create the rethink promise, then await the pg promise, then await the rethink promise to make sure they work in parallel. I thought that pattern might reduce readability for the team, but I'm OK with doing whatever makes sense!

Dschoordsch: In cases where we expect postgres to fail (so not for inserts), we can use `Promise.allSettled` but need to verify and throw if rethink did throw an error. So adding a
```
  if (result.status === 'rejected') throw new Error(result.reason)
```
would work for me.

I wasn't so concerned about the performance around `MeetingTemplate` and would simply await both promises (rethink and postgres) one after the other for readability, but checking and throwing is cleanest I think (again only where we expect postgres to fail)

mattkrick: that is super reasonable. will fix!

mattkrick: one thing i realized-- rethinkdb doesn't throw if an update fails. it'll only throw if it can't reach the DB, or if we didn't write the query correctly, etc. Still nice to throw in those conditions, it'll just be pretty rare

mattkrick:
@Dschoordsch ready for another review!

packages/server/postgres/queries/insertMeetingTemplate.ts
Dschoordsch: -1 There is at least one place where we also have `scope` and `lastUsedAt`
https://github.com/ParabolInc/parabol/blob/ed7aae85321542057e0069946dc38b89a883e5ca/packages/server/graphql/mutations/updateTemplateScope.ts#L81-L88

Dschoordsch (changes requested):

packages/server/postgres/queries/insertMeetingTemplate.ts
```diff
@@ -0,0 +1,19 @@
+import getRethink from '../../database/rethinkDriver'
+import MeetingTemplate from '../../database/types/MeetingTemplate'
+import getPg from '../getPg'
+
+const insertMeetingTemplate = async (meetingTemplate: MeetingTemplate) => {
+  const r = await getRethink()
+  const pg = getPg()
+  const {id, name, teamId, orgId, parentTemplateId, type} = meetingTemplate
+  const [rRes] = await Promise.allSettled([
```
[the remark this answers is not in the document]
mattkrick: Ah gotcha. Could we keep it the same & just do a `console.error` instead of throwing? I'd hate if a write to PG broke the app.

packages/server/postgres/queries/insertMeetingTemplate.ts
mattkrick: added those 2 + isFree, isStarter. future proof!

PR #7681, opened 2023-01-24
chore(teams-limit): avoid showing nagging snackbar if the limit is fixed

igorlesnenko:
If the user fixed its exceeding limit, we need to change `suggestedTier` to `starter` to avoid showing nagging snackbar.

How to test:

- Test on a fresh account with fresh active domain without upgraded teams. (e.g. create new user user@freshdomain123.com)
- Add a feature flag on the org

```
  .getAll(r.args(['PUT_YOUR_ORG_ID_HERE']), {index: 'id'})
  .update({
    tierLimitExceededAt: null,
    scheduledLockAt: null,
    lockedAt: null
});
```

- Uncomment https://github.com/ParabolInc/parabol/blob/99e3106cc942b8b74dea6668487fcc6658bede1e/packages/server/billing/helpers/teamLimitsCheck.ts#L14
- Change values to:
```
const enum Threshold {
  MAX_STARTER_TIER_TEAMS = 1,
  MIN_STICKY_TEAM_MEETING_ATTENDEES = 1,
  MIN_STICKY_TEAM_MEETINGS = 1,
  STARTER_TIER_LOCK_AFTER_DAYS = 0
}
```
- Create at least 2 teams
- Run and complete a meeting for each team
- See usage stats enabled but there is no nagging snackbar (only closable snackbar that will not appear again on page refresh)
- Remove one team, see `tierLimitExceeded` flag is gone
- See that you still can access usage stats, but you still don't see nagging snackbar
- Refresh the page, see that everything is fine
- Upgrade to PRO, see that everything is fine


packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -35,7 +35,7 @@ const enableUsageStats = async (userIds: string[], orgId: string) => {
     .table('OrganizationUser')
     .getAll(r.args(userIds), {index: 'userId'})
     .filter({orgId})
-    .update({suggestedTier: 'team'})
+    .update({suggestedTier: 'starter'})
```
nickoferrall, line 38: The org has exceeded the starter tier limits, so we call `enableUsageStats`. I'm not sure why the `suggestedTier` would be `starter` rather than `team` as we're suggesting that they upgrade?

nickoferrall (changes requested):

packages/server/billing/helpers/teamLimitsCheck.ts
igorlesnenko: @nickoferrall I see a couple of issues with specifying `team` as a `suggestedTier` :

1. We have hardcoded logic that shows a nagging snackbar in this case, which we don't want: https://github.com/ParabolInc/parabol/blob/master/packages/client/hooks/useUsageSnackNag.ts#L22
2. As a workaround I checked if there is an org with `limitExceededAt` flag. It works, but not perfectly:
3. If the user fixed its limit by removing a team, he will see nagging snackbar again, unless we disable usage stats for him.
4. But we can't just disable usage stats, we also need to remove feature flag. Because usage stats does not work without `suggestedTier`.

I think we just want to enable usage stats without suggesting anything, so the above approach feels a bit hacky.

Setting `personal` solves these issues in a quick way without much refactoring.

What do you think?

nickoferrall: I think it makes sense as a quick solution that helps us to ship the feature faster, but I think we should create an issue to refactor it soon

nickoferrall (approved):

packages/server/billing/helpers/teamLimitsCheck.ts
igorlesnenko: Agree, I can create as soon as I get back

Dschoordsch (changes requested):
Without digging too deep, I don't see what the intended behaviour for this change is and why it's needed.

igorlesnenko:
@Dschoordsch This is a quick solution for enabling usage stats without showing a nagging snackbar.

- Currently, to enable usage stats we need to do two things — add a feature flag and set `suggestedTier`
- We have hardcoded logic that shows a nagging snackbar if `suggestedTier === 'team' && tier === 'starter'` https://github.com/ParabolInc/parabol/blob/master/packages/client/hooks/useUsageSnackNag.ts#L22
- We don't want to show a nagging snackbar. Previously, as a workaround I checked if there is an org with `limitExceededAt` flag. It works, but not perfectly:
- If the user fixed its limit by removing a team, he will see a nagging snackbar again, unless we disable usage stats for him by removing `suggestedTier`
- But we can't just remove `suggestedTier`, we also need to remove feature flag. Because currently usage stats does not work without `suggestedTier`.
- As a quick solution I want to be able to set `suggestedTier === 'starter'`. In this case usage stats will be enabled, but nagging snackbar will not be shown.
- I'm going to create refactoring issue to improve this

Do you think is it ok for a quick solution? Or should we implement more robust solution from the beginning?

packages/server/billing/helpers/teamLimitsCheck.ts
Dschoordsch: -2 Ok, I think I get now what you're trying to achieve. I still believe setting `suggestedTier: 'starter'` when they run over the limit is wrong, we should still set `suggestedTier: 'team'` here. However we can change `maybeRemoveRestrictions` to set the `suggestedTier: 'starter'` if they fall below the limit again. This should still accomplish what you wanted to do, shouldn't it?

Dschoordsch (changes requested):

packages/server/billing/helpers/teamLimitsCheck.ts
igorlesnenko: @Dschoordsch this would work, updated the description and pushed new changes

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -114,16 +114,27 @@ export const maybeRemoveRestrictions = async (orgId: string, dataLoader: DataLoa
...
+          scheduledLockAt: null,
+          lockedAt: null,
+          updatedAt: new Date()
+        })
+        .run(),
+      r
+        .table('OrganizationUser')
+        .getAll(r.args(billingLeadersIds), {index: 'userId'})
+        .filter({orgId})
+        .update({suggestedTier: 'starter'})
```
Dschoordsch, line 134: -1 Because you're loading the billingLeaders via dataloader, we need to update or reset them in the cache.

Dschoordsch (changes requested):

Dschoordsch (approved):
Did not test again

PR #7682, opened 2023-01-25
chore(teams-limit): query HubSpot API to filter out in progress deals

### LINKED ISSUE #7436: Teams limit: Exclude organizations and domains with active deal in progress
We need to exclude deals we're actively working from getting the time-bound trigger.

This information is available in HubSpot.

There are 2 possible approaches that I see:

1) Dynamically query HubSpot API to check if the organization or domain should be excluded.
2) Setup HubSpot webhook that will update some flag on organization accordingly, for example (`has_active_deal`)

Second option  seems to be better in terms of architecture, but I would suggest to pick something that would be faster to implement for now.

AC:
- [ ] Determine which approach for excluding organizations is better
- [ ] Do not trigger limit flow for organizations that are WIP by sales
- [ ] Do not trigger limit flow for organizations that are under a domain that is WIP by sales

igorlesnenko:
Fixes #7436

In this PR we query HubSpot API according to this [guide](https://www.notion.so/parabol/How-to-identify-in-progress-deals-in-HubSpot-via-API-9563a8994dce44eda7ae0c691d376a5c) to skip organizations with active deal in progress

I went with this approach as the fastest to implement, considering we are doing an experiment

I also disabled `teamLimitsCheck` function for organizations without active domains at all.

**How to test:**

- Add `HUBSPOT_API_KEY` env variable to `.env` file
- Add `HUBSPOT_SALES_PIPELINE_ID`
- Add `HUBSPOT_SALES_PIPELINE_ACTIVE_STAGES`
- Have an organization with a fake active domain (warning — do not use real customer domains)
- Reset org and add feature flag:
```
  .getAll(r.args(['PUT_YOUR_ORG_ID_HERE']), {index: 'id'})
  .update({
    tierLimitExceededAt: null,
    scheduledLockAt: null,
    lockedAt: null
});
```
- Uncomment https://github.com/ParabolInc/parabol/blob/99e3106cc942b8b74dea6668487fcc6658bede1e/packages/server/billing/helpers/teamLimitsCheck.ts#L14
- Trigger limits check function by running a meeting and completing it
- See that because the domain does not exists on HubSpot, the limit was successfully triggered (user see a snackbar etc)
- Now find a real deal in progress on a HubSpot, copy the domain and edit the code here: https://github.com/ParabolInc/parabol/blob/05fb6ae752918fb91faf2a10817002ab7020abcd/packages/server/billing/helpers/teamLimitsCheck.ts#L145 
- Reset the organization 
- Trigger limits check function by running a meeting and completing it
- See that no limit was triggered
- Reset the organization, remove edits to the domain (use the fake one)
- Put wrong value in `HUBSPOT_ACCESS_TOKEN` (but not empty value)
- Restart app `yarn dev`
- Trigger limits check function by running a meeting and completing it
- See that no limit was triggered


packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -139,6 +141,18 @@ export const checkTeamsLimit = async (orgId: string, dataLoader: DataLoaderWorke
 
   if (!(await isLimitExceeded(orgId, dataLoader))) return
 
+  if (!organization.activeDomain) return
```
igorlesnenko, line 144: @tghanken We have few things that we can run only if organization has an active domain, including querying HubSpot API.
So I completely disabled the limits check for organizations without domain, is it okay?

packages/server/billing/helpers/teamLimitsCheck.ts
```diff
@@ -139,6 +141,18 @@ export const checkTeamsLimit = async (orgId: string, dataLoader: DataLoaderWorke
...
+
+  const hasActiveDeals = await domainHasActiveDeals(organization.activeDomain)
+
+  if (hasActiveDeals) {
+    if (hasActiveDeals instanceof Error) {
```
igorlesnenko, line 149: If we can't determine if the deal is in progress, let's not trigger the check.

tghanken (approved):
Looks good to me!

packages/server/hubspot/hubSpotApi.ts
```diff
@@ -0,0 +1,143 @@
...
+
+type DealsResponse = {
+  total: number
+}
+
+type CompaniesResponse = {
+  results: Company[]
+}
+
+const getCompaniesByDomain = async (domain: string): Promise<CompaniesResponse | Error> => {
```
nickoferrall, line 17: +1 it looks like the HubSpot Client package offers types, so we wouldn't need to cast them: https://www.npmjs.com/package/@hubspot/api-client. It's probably not worth refactoring though unless we expect to add more logic to the hubspot API soon

packages/server/hubspot/hubSpotApi.ts
```diff
@@ -0,0 +1,143 @@
+import fetch from 'node-fetch'
```
nickoferrall, line 1: +1 this wasn't introduced in this PR, but the folder is called `hubspot`, but the files within it are spelt `hubSpot`. It'd be nice if they were consistent. Could change the folder name to `hubSpot` to ease my OCD

packages/server/hubspot/hubSpotApi.ts
```diff
@@ -0,0 +1,143 @@
...
+  const resJson = await res.json()
+
+  if (resJson.status === 'error') {
+    return new Error(resJson.message)
+  }
+
+  return resJson
+}
+
+const SALES_PIPELINE_ID = '14fa7db3-cc82-4562-8603-90d792014b42'
```
nickoferrall, line 61: +1 I think these ids should probably be stored in the `.env` file. They're useless without the `HUBSPOT_API_KEY`, but if someone got access to that, they could quickly pull all of our sales data

nickoferrall (approved):
Tested and it's working for me! 

@Dschoordsch you can find the HubSpot API key [here](https://github.com/ParabolInc/action-devops/blob/03e90c37b8f2700411a38db73cdd32059385f8e0/environments/production#L72) and active deals being worked on [here](https://app.hubspot.com/contacts/3888472/objects/0-3/views/all/board) 🔒

packages/server/billing/helpers/teamLimitsCheck.ts
Dschoordsch: +1 I think this should be a bit more explicit discussion. While there are not many teams which qualify with no `activeDomain`, I don't see why we should exclude them from the teams limit.

packages/server/hubspot/hubSpotApi.ts
Dschoordsch: -1 Yes, let's prefix this with HUBSPOT and put in the `.env`

packages/server/hubspot/hubSpotApi.ts
```diff
@@ -0,0 +1,143 @@
...
+const PQL_ACCEPTED_STAGE_ID = '2410510'
+const PQL_REPLIED_STAGE_ID = 'd394bd39-f31c-4cac-afba-2d9b85b57e2b'
+const SQL_STAGE_ID = '48c94b2b-1d51-47af-b563-cf490d799ec2'
+const SQL_IN_PROCUREMENT_STAGE_ID = '611924'
```
Dschoordsch, line 65: -1 Let's also put these into `.env`. We don't need to distinguish between those, so let's just do
```
HUBSPOT_SALES_PIPELINE_ACTIVE_STAGES='2410510,d394bd39-f31c-4cac-afba-2d9b85b57e2b,48c94b2b-1d51-47af-b563-cf490d799ec2,611924'
```
and then `.split(',')` them before usage.

Dschoordsch (changes requested):
Only minor `.env` cleanup. Looks good, did not test

packages/server/hubspot/hubSpotApi.ts
igorlesnenko: Make sense, It's completely slipped my mind that it's possible to use a library here.
I will add a comment in the code, so we can take this library next time.

igorlesnenko:
@Dschoordsch fixed, created a PR with variables: https://github.com/ParabolInc/action-devops/pull/103

Dschoordsch (approved):
Works

Issue #7643, opened 2023-01-10
SAML UI: Create `updateSAML` mutation


Dschoordsch:
Needs:
* #7642 

Updating SAML for a domain should be possible by billing leaders of an organization for themselves.

## Acceptance criteria

* [ ] Create `updateSAML` mutation similar to [`enableSAMLForDomain`](https://github.com/ParabolInc/parabol/blob/d75533910d56f4358a08f7477fa3df3a72013522/packages/server/graphql/private/mutations/enableSAMLForDomain.ts#L70)
* [ ] arguments: `orgId`, `metadata`
* [ ] `metadata` can be `null` for disabling SAML again (should also update `url` in the `SAML` table)
* [ ] only callable by a `billingLeader` on an `enterprise` organization
* [ ] requires an existing `SAML` entry, otherwise fails

gitstart-parabol:
@Dschoordsch, we suggest including `slug` or `name` as part of the mutation arguments to facilitate querying the `SAML` table, as the `orgId` field may potentially be `null`.

Dschoordsch:
@gitstart-parabol This mutation should only be callable by a billing leader of an organization with a verified domain. That means there must be a SAML entry with the `orgId` set, otherwise they should not have access to this feature.

Modifying existing entries without an `orgId` via `domain` and `slug` can only be done by us using the existing, private `enableSAMLForDomain` mutation.

PR #7685, opened 2023-01-27
feat: SAML UI: Create updateSAML mutation

gitstart-parabol:
Fixes https://github.com/ParabolInc/parabol/issues/7643

* Create `updateSAML` mutation similar to [enableSAMLForDomain](https://github.com/ParabolInc/parabol/blob/d75533910d56f4358a08f7477fa3df3a72013522/packages/server/graphql/private/mutations/enableSAMLForDomain.ts#L70)
* arguments: `orgId`, `metadata`
* `metadata` can be `null` for disabling SAML again (should also update` url` in the `SAML` table)
* only callable by a `billingLeader` on an `enterprise` organization
* requires an existing `SAML` entry, otherwise fails

 ## Testing scenarios
 * Go to http://localhost:3000/admin/graphql
 * Run the following mutation and update the variables.
```
mutation updateSaml($orgId: ID!, $metadata: String) {
  updateSAML(orgId: $orgId, metadata:$metadata) {
    ... on UpdateSAMLSuccess {
      success
    }
    ... on ErrorPayload {
      error {
        title
        message
      }
    }
  }
}


```


gitstart-parabol:

gitstart-parabol:
Hi @Dschoordsch, I just wanted to find out how soon will we be getting reviews on this. Thanks.

packages/server/graphql/public/typeDefs/updateSAML.graphql
```diff
@@ -0,0 +1,28 @@
...
+}
+
+"""
+Return value for updateSAML, which could be an error
+"""
+union UpdateSAMLPayload = ErrorPayload | UpdateSAMLSuccess
+
+type UpdateSAMLSuccess {
+  """
+  Describe the first return field here
```
BartoszJarocki, line 25: -1 no generic description please

packages/server/graphql/public/typeDefs/updateSAML.graphql
```diff
@@ -0,0 +1,28 @@
...
+  """
+  success: Boolean!
```
BartoszJarocki, line 27: @Dschoordsch I'm not 100% sure how this is going to work, do we need to return newly created/updated object id here and be able to resolve it? I'm asking as I'm not sure if frontend needs more data than `true/false`

packages/server/graphql/public/mutations/updateSAML.ts
```diff
@@ -0,0 +1,36 @@
...
+  const samlRecord = await r
+    .table('SAML')
+    .getAll(orgId, {index: 'orgId'})
+    .limit(1)
+    .nth(0)
+    .default(null)
+    .run()
+
+  if (!samlRecord) {
+    return {error: {message: 'SAML has not been created in Parabol yet.'}}
```
BartoszJarocki, line 18: +1 as this is something called by the user, I'd say we should log it to sentry, check the usage of `standardError` https://github.com/ParabolInc/parabol/blob/a1ca44dda2254fe6747e64e2c358c9dd53f7237b/packages/server/utils/standardError.ts#L3

packages/server/graphql/public/mutations/updateSAML.ts
```diff
@@ -0,0 +1,36 @@
...
+  }
+
+  const url = metadata ? getSignOnURL(metadata, samlRecord.id) : null
+  if (url instanceof Error) return {error: {message: url.message}}
```
BartoszJarocki, line 22: +1 [`standardError`](https://github.com/ParabolInc/parabol/blob/a1ca44dda2254fe6747e64e2c358c9dd53f7237b/packages/server/utils/standardError.ts#L3)

packages/server/graphql/public/mutations/updateSAML.ts
```diff
@@ -0,0 +1,36 @@
+import getRethink from '../../../database/rethinkDriver'
+import {MutationResolvers} from '../resolverTypes'
+import getSignOnURL from './helpers/SAMLHelpers/getSignOnURL'
+
+const updateSAML: MutationResolvers['updateSAML'] = async (_source, {orgId, metadata}, {}) => {
```
BartoszJarocki, line 5: @Dschoordsch just making sure as I don't see anything subscriptions related, is this mutation supposed to publish any data on `org` level? Or is the UI only for the `billing leader` and there's no need to do anything subscriptions related?

codegen.json
```diff
@@ -69,6 +70,7 @@
           "UpdateDimensionFieldSuccess": "./types/UpdateDimensionFieldSuccess#UpdateDimensionFieldSuccessSource",
           "UpdateGitLabDimensionFieldSuccess": "./types/UpdateGitLabDimensionFieldSuccess#UpdateGitLabDimensionFieldSuccessSource",
           "UpdateMeetingPromptSuccess": "./types/UpdateMeetingPromptSuccess#UpdateMeetingPromptSuccessSource",
+          "UpdateSAMLSuccess": "./types/UpdateSAMLSuccess#UpdateSAMLSuccessSource",
```
BartoszJarocki, line 73: -1 no need to add mapping if there's no such file as graphql codegen will generate incorrect types

BartoszJarocki (changes requested):

packages/server/graphql/public/mutations/updateSAML.ts
Dschoordsch: The UI is only for the billing leader. There could be multiple, but that's an edge case we don't need to fix right now.

BartoszJarocki: got it

packages/server/graphql/public/typeDefs/updateSAML.graphql
gitstart-parabol: @BartoszJarocki, here is a  [draft](https://github.com/ParabolInc/parabol/pull/7788) for frontend, where we use this `mutation`  for enabling and disabling SSO. we have opted to perform a refetch to inform future decisions based on the query result.

codegen.json
gitstart-parabol: @BartoszJarocki  this was  autogenerated by `codegen` at the point of creating this `mutation`,  I would appreciate your input on whether or not it is necessary to remove it?

gitstart-parabol:
@BartoszJarocki this PR is ready for review.

BartoszJarocki (approved):

Dschoordsch (approved):
Works. A could of minor nits which can be done in a follow up.

PR #7686, opened 2023-01-27
feat: SAML UI: Create a verifyDomain mutation

### LINKED ISSUE #7642: SAML UI: Create a `verifyDomain` mutation
An organization has an associated domain which we derive from their email address. If a user wants to setup SAML we need to verify that the domain really belongs to them as all users under that domain will be affected. For starters we want to be able to do this manually with a custom domain.

## Acceptance criteria
* [ ] add a new optional field `orgId` to the SAML table
* [ ] create a new private mutation `verifyDomain(orgId: ID!, domains: [String!]!, slug: ID!)` [^1] which inserts a new SAML value with `domain`, `slug` and `orgId` set; `url` and `metadata` will be `null`
* [ ] ensure the current code can handle `url` and `metadata` set to `null`

[^1]: the idea https://github.com/ParabolInc/parabol/issues/6811#issuecomment-1180913613 also mentions a `method` field, but we don't need it because it will always be "manual" atm.

gitstart-parabol:
# Description

Fixes #7642

* Add a new optional field orgId to the SAML table
* Create a new private mutation verifyDomain(domains: [parabol.co], slug: parabol) [1](https://developers.gitstart.com/ticket/PARA-7642#user-content-fn-1) which inserts a new SAML value with domain, slug and orgId set; url and metadata will be null
* Ensure the current code can handle url and metadata set to null

 ## Testing scenarios
 * Go to http://localhost:3000/admin/graphql
 * Select private schema
 * Run the following mutation and update the variables.
```
mutation VerifyDomainMutation($orgId:String! $domains:[ID!]! $slug: String!) {
  verifyDomain(orgId:$orgId domains: $domains slug: $slug) {
     ... on VerifyDomainSuccess {
       success
    }
    ... on ErrorPayload {
      error {
        message
      }
    }   
  }
}
```


gitstart-parabol:

packages/server/graphql/private/mutations/verifyDomain.ts
```diff
@@ -0,0 +1,38 @@
...
+  const normalizedDomains = domains.map((domain) => domain.toLowerCase())
+
+  // VALIDATION
+  const slugName = normalizeName(slug)
+  if (slugName instanceof Error) return {error: {message: slugName.message}}
+
+  const organizationSAMLExist = await r.table('SAML').get(slugName).run()
+  if (organizationSAMLExist) {
+    await r.table('SAML').get(slugName).update({orgId: null}).run()
+  }
```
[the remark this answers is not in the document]
gitstart-parabol: @igorlesnenko, following the discussion with @Dschoordsch in this thread (https://github.com/ParabolInc/parabol/issues/7642#issuecomment-1407456281). It was confirmed that any previous values associated with the orgId should be nullified.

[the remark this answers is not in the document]
igorlesnenko: @gitstart-parabol In this case he was only talking about migration, he meant that the default value should be null and we should not fill it in migration for existing records.

packages/server/database/migrations/20230206131155-updateOrgIdOnSAML.ts
```diff
@@ -0,0 +1,16 @@
...
+    .update((row) => ({
+      orgId: row('orgId').default(null)
+    }))
+    .run()
+
+  await r.table('SAML').indexCreate('orgId').run()
+}
+
+export const down = async function (r: R) {
+  await r.table('SAML').indexDrop('orgId').run()
```
igorlesnenko, line 15: +1 I think we don't need that migration if we just set a column to null, because rethinkdb is schemaless no need to explicitly add a new column.

I also don't see the index is used in this PR. I suggest to just remove this migration

igorlesnenko (changes requested):

packages/server/database/migrations/20230206131155-updateOrgIdOnSAML.ts
gitstart-parabol: @igorlesnenko,  the index is being used in issue  #7643 to search a `saml` record based on the `orgId`.  however   adding a new optional field `orgId` to the SAML table is part of this PR.  you can  please advice.

igorlesnenko: @gitstart-parabol I see, lets keep the index then

gitstart-parabol: Alright @igorlesnenko, the other comments above have been implemented.

packages/server/graphql/private/mutations/enableSAMLForDomain.ts
```diff
@@ -3,6 +3,7 @@ import {v4 as uuid} from 'uuid'
 import zlib from 'zlib'
 import getRethink from '../../../database/rethinkDriver'
 import {MutationResolvers} from '../resolverTypes'
+import {normalizeSlugName} from './helpers/SAMLHelpers'
```
igorlesnenko, line 6: As I see, we currently use separate files for each helper, can we rename `SAMLHelpers.ts` file to `normalizeSlugName.ts` and do `export default`?

packages/server/graphql/private/mutations/verifyDomain.ts
```diff
@@ -0,0 +1,28 @@
...
+  const slugName = normalizeSlugName(slug)
+  if (slugName instanceof Error) return {error: {message: slugName.message}}
+
+  const organizationSAMLExist = await r.table('SAML').get(slugName).run()
+  if (organizationSAMLExist) return {error: {message: 'SAML exist for organization'}}
+
+  await r
+    .table('SAML')
+    .insert({
+      orgId,
```
igorlesnenko, line 19: Is it possible to have multiple entries for same `orgId`? If not I suggest to validate that too.

packages/server/graphql/private/mutations/verifyDomain.ts
```diff
@@ -0,0 +1,28 @@
...
+
+const verifyDomain: MutationResolvers['verifyDomain'] = async (_source, {slug, domains, orgId}) => {
+  const r = await getRethink()
+  const normalizedDomains = domains.map((domain) => domain.toLowerCase())
+
+  // VALIDATION
+  const slugName = normalizeSlugName(slug)
+  if (slugName instanceof Error) return {error: {message: slugName.message}}
+
+  const organizationSAMLExist = await r.table('SAML').get(slugName).run()
```
igorlesnenko, line 13: We can check if record exists without fetching the whole data by  ` .count().eq(1)`

packages/server/graphql/private/typeDefs/verifyDomain.graphql
```diff
@@ -0,0 +1,33 @@
...
+
+    """
+    organization id
+    """
+    orgId: ID!
+
+    """
+    Company's slug
+    """
+    slug: String!
```
igorlesnenko, line 19: I think we can make it ID! too, as it is actually a primary key as I see

igorlesnenko (changes requested):

igorlesnenko:
@gitstart-parabol build is failing for some reason, can you merge master?

packages/server/database/migrations/20230206131155-updateOrgIdOnSAML.ts
```diff
@@ -0,0 +1,16 @@
+import {R} from 'rethinkdb-ts'
+
+export const up = async function (r: R) {
```
igorlesnenko, line 3: As of [7720](https://github.com/ParabolInc/parabol/pull/7720) we now store all migrations in  `sevrer/postgres/migrations`. For this reason the build fails. Can you move the migration there?

igorlesnenko: You would need to run `yarn pg:migrate create updateOrgIdOnSAML.ts`.
Then copy your migration there
Update migration code by adding `connectRethinkDB` see example here: https://github.com/ParabolInc/parabol/blob/eeb6d0282a8e92688a4fab92a2338d76b5d3c6f3/packages/server/postgres/migrations/1674074684762_deleteEmptyGroups.ts

packages/server/graphql/private/mutations/verifyDomain.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  // VALIDATION
+  const slugName = normalizeSlugName(slug)
+  if (slugName instanceof Error) return {error: {message: slugName.message}}
+
+  const slugNameExist = await r.table('SAML')('id').count(slugName).eq(1).run()
+  if (slugNameExist) return {error: {message: 'SAML exist with slug name'}}
+
+  const orgIdExist = await r.table('SAML')('orgId').count(orgId).eq(1).run()
+  if (orgIdExist) return {error: {message: 'SAML exist for organization'}}
```
igorlesnenko, line 17: +1 can we make it as a single query or use `Promise.all`?

igorlesnenko (changes requested):
Small tweaks left

packages/server/graphql/private/mutations/verifyDomain.ts
```diff
@@ -0,0 +1,32 @@
...
+  const r = await getRethink()
+  const normalizedDomains = domains.map((domain) => domain.toLowerCase())
+
+  // VALIDATION
+  const slugName = normalizeSlugName(slug)
+  if (slugName instanceof Error) return {error: {message: slugName.message}}
+
+  const [slugNameExist, orgIdExist] = await Promise.all([
+    r.table('SAML')('id').count(slugName).eq(1).run(),
+    r.table('SAML')('orgId').count(orgId).eq(1).run()
```
igorlesnenko, line 15: +1 I noticed that the queries in this form this do not use indexes

We can write something like this instead:

```
await Promise.all([
  r.table('SAML').get(slugName).run(),     
    r.table('SAML').getAll(orgId, {index: 'orgId'}).limit(1)
    .count()
    .eq(1).run()
])

**Review by igorlesnenko — CHANGES_REQUESTED** [review]
Just a couple of more tweaks

**Review by igorlesnenko — APPROVED** [review]
LGTM

File: packages/server/graphql/private/typeDefs/verifyDomain.graphql [hunk 10]
```diff
@@ -0,0 +1,33 @@
...
+}
+
+"""
+Return value for verifyDomain, which could be an error
+"""
+union VerifyDomainPayload = ErrorPayload | VerifyDomainSuccess
+
+type VerifyDomainSuccess {
+  """
+  Describe the first return field here
```
Dschoordsch, line 30: +1 "True if the operation was a success"

Dschoordsch (approved):

Dschoordsch:
Works

packages/server/graphql/private/typeDefs/verifyDomain.graphql
```diff
@@ -0,0 +1,33 @@
+extend type Mutation {
+  """
+  Describe the mutation here
```
igorlesnenko, line 3: +1 We also can update this in next PR

Issue #7474, opened 2022-11-22
AI Summary: add live chatbot 


nickoferrall:
To get qualitative feedback from users, add the live chat bot to the AI summary feature.

Check with the Growth team to find out where the best place to include this is.

### Proposed AC

- [ ] Live chatbot added to meeting summary

nickoferrall:
This issue is in the Done column so I'll close it

PR #7705, opened 2023-01-31
chore(ai-summary): update meeting summary url

nickoferrall:
We want to add the HubSpot chatbot to meeting summaries that have the AI Summary. To do that, we can update the URL so HubSpot knows when it has an AI Summary.

Related to https://github.com/ParabolInc/parabol/issues/7474

### To test

- [ ] End a retro meeting where the facilitator has access to the AI Summary. See that the URL ends in `?ai=true`
- [ ] End a retro meeting where the facilitator does not have access to the AI Summary. See that the URL ends in `/new-summary/${meetingId}`

### DIFF
--- a/packages/client/mutations/EndRetrospectiveMutation.ts
+++ b/packages/client/mutations/EndRetrospectiveMutation.ts
@@ -28,6 +28,13 @@ graphql`
       reflectionCount
       taskCount
       topicCount
+      facilitator {
+        user {
+          featureFlags {
+            aiSummary
+          }
+        }
+      }
     }
     team {
       id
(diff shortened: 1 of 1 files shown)


packages/client/mutations/EndRetrospectiveMutation.ts
```diff
@@ -28,6 +28,13 @@ graphql`
       reflectionCount
       taskCount
       topicCount
+      facilitator {
```
nickoferrall, line 31: Checking the facilitator feature flag instead of the `summary` field in the retro meeting because it may not have been generated by the time the meeting ends

Dschoordsch (changes requested):
Does not work. I get `ai=true` even when I ran out of free ai meetings

Dschoordsch (approved):

Issue #7700, opened 2023-01-31
Recurrence Stable Link: Link Redirection


jmtaber129:
When a user navigates to `/meeting-series/:id`, redirect to the most recent meeting in the series.

AC:
- [ ] For a meeting series with an active meeting, `/meeting-series/:id` redirects to that active meeting
- [ ] For a meeting series with no active meetings, `/meeting-series/:id` redirects to the most recent meeting (same sort order as History)
- [ ] For a meeting series ID the user does not have access to, use same behavior as when a user doesn't have access to a meeting
  - [ ] This includes invalid meeting IDs

PR #7707, opened 2023-02-01
feat(recurrence): Stable link for meeting series

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/7700

Support stable link meeting redirection for meeting series links of the form `/meeting-series/:meetingId`, e.g. `/meeting-series/kzTbFLeRY4` for series containing a meeting with ID `kzTbFLeRY4`, which will redirect to the most recent meeting in `kzTbFLeRY4`'s series.

If the user does not have access to the meeting (either because it doesn't exist or because it's on a different team), we'll redirect to `/invitation-required`, which will request an invite for the user to the meeting series' team, then redirect once the invite request is accepted.

This also fixes a bug where the `/invitation-required` path won't successfully redirect to a `/meet/:meetingId` route when `meetingId` is inactive (i.e. when the path points to a meeting that already ended).

## Demo
https://www.loom.com/share/bb4ec3c7dad54fcb923b970ab359f2f3

## Testing scenarios
- [ ] Create two users - User A that is on a meeting series' team, and User B that is not.
- [ ] As user A, navigate to `/meeting-series/:meetingId`, where `meetingId` is any meeting ID in the series, and confirm that it redirects to the most recent meeting
  - [ ] Do this when the most recent meeting is active and inactive, and when the `meetingId` is active or inactive.
- [ ] As user B and with user A online, navigate to `/meeting-series/:meetingSeriesId`.
- [ ] Confirm that user B redirects to `/invitation-required` and user A sees an invitation request toast for user B
- [ ] Accept the request as user A
- [ ] Confirm that user B redirects to the most recent meeting in the series

## Final checklist


jmtaber129:
@BartoszJarocki I'm marking this as "One Review Required" since this doesn't introduce any new patterns and is relatively straightforward, but feel free to loop in maintainers if you think it's worth doing so.

packages/client/components/MeetingSeriesRedirector.tsx
```diff
@@ -0,0 +1,56 @@
...
+      })
+    } else {
+      const {mostRecentMeeting} = meetingSeries
+      const {id: meetingId} = mostRecentMeeting
+
+      history.replace({
+        pathname: `/meet/${meetingId}`
+      })
+    }
+  }, [])
```
BartoszJarocki, line 51: +1 as this is a one time thing, there's no need to use effect here. you can simply use react router [redirect component](https://v5.reactrouter.com/web/api/Redirect). declarative and cleaner

packages/client/components/MeetingSeriesRoot.tsx
```diff
@@ -0,0 +1,32 @@
...
+const MeetingRoot = () => {
+  const {history, match} = useRouter<{meetingSeriesId: string}>()
+  const {params} = match
+  const {meetingSeriesId: encodedMeetingSeriesId} = params
+  const meetingSeriesId = decodeURIComponent(encodedMeetingSeriesId)
+  useEffect(() => {
+    if (!meetingSeriesId) {
+      history.replace('/meetings')
+    }
+  }, [])
```
BartoszJarocki, line 18: +1 same here, better to use `<Redirect />`, although it'll have to be below usage of `useQueryLoaderNow` but that's ok

packages/server/graphql/types/User.ts
```diff
@@ -525,17 +526,22 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
           description:
             'The meetingId to check for the invitation, if teamId not available (e.g. on a meeting route)'
         },
+        meetingSeriesId: {
+          type: GraphQLID,
+          description:
+            'The meetingSeriesId to check for the invitation, if teamId not available (e.g. on a meeting series route)'
+        },
```
BartoszJarocki, line 533: +1 where do we use it? do you mind sharing a little bit more what's the use case here? having troubles understanding what's the problem we're solving 🙏

BartoszJarocki (commented):
A couple of comments, will do the testing and final round after they're resolved. nice work!

packages/client/components/MeetingSeriesRedirector.tsx
jmtaber129: TIL about the `Redirect` component!

BartoszJarocki (commented):
@jmtaber129 I tested and it's all good. the only worry I have right now is we shouldn't be using postgres int ids in links. I believe link should look like `meeting-series/kxvmKg4St2` instead `meeting-series/meetingSeries:120`. The reasons: is easy to figure out meeting ids outside of their teams for our users, it's also easy to see how often this feature is used. Do you have any thoughts about that?

jmtaber129:
@BartoszJarocki hmm yea you're right.  Not sure what alternative approaches there are though 🙁  meetings look like `/meet/kxvmKg4St2` largely because rethink IDs look like `kxvmKg4St2`.

BartoszJarocki:
@jmtaber129 probably a wild idea but what if...
1. ID of the first meeting in the meeting series would be used in the url `meeting-series/:firstMeetingId`
2. Then the query in `MeetingSeriesRedirector` would be something like 
```
  const data = usePreloadedQuery<MeetingSeriesRedirectorQuery>(
    graphql`
      query MeetingSeriesRedirectorQuery($meetingId: ID!) {
        viewer {
          isConnected
          meeting(meetingId: $meetingId) {
            meetingSeries {
               mostRecentMeeting {
                  id
               }
            }
          }
        }
      }
    `,
    queryRef,
    {UNSTABLE_renderPolicy: 'full'}
  )
```
and then... everything else stays the same? am I missing something or would that work just fine?

jmtaber129:
@BartoszJarocki I actually really like that! So `/meeting-series/:meetingId` would redirect to the most recent meeting in the meeting's series.

I'll change this to that behavior.

BartoszJarocki:
awesome!

packages/server/graphql/public/mutations/acceptTeamInvitation.ts
```diff
@@ -51,8 +51,6 @@ const acceptTeamInvitation: MutationResolvers['acceptTeamInvitation'] = async (
   const {invitation} = invitationRes
   const {meetingId, teamId, invitedBy: inviterId} = invitation
   const acceptAt = invitation.meetingId ? 'meeting' : 'team'
-  const meeting = meetingId ? await dataLoader.get('newMeetings').load(meetingId) : null
-  const activeMeetingId = meeting && !meeting.endedAt ? meetingId : null
```
jmtaber129, line 55: @mattkrick I'd like to make this change as part of a fix for `/invitation-required` not redirecting when `meetingId` points to a meeting that has already ended - are there any negative side effects I might not be seeing here?

BartoszJarocki (approved):
lgtm!

Issue #7528, opened 2022-12-06
Checkout Flow: add feature flag


nickoferrall:
Add a feature flag for the checkout flow. We may not A/B test this feature, but it would be helpful to have a feature flag while we're building it so we can play with it internally.

### AC

- [ ] Feature flag added

PR #7709, opened 2023-02-02
chore(checkout-flow): add checkout feature flag

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7528


packages/server/graphql/types/UserFlagEnum.ts
```diff
@@ -1,7 +1,5 @@
 import {GraphQLEnumType} from 'graphql'
 
-export type UserFeatureFlagEnum = 'azureDevOps' | 'msTeams'
```
nickoferrall, line 3: This wasn't being used anywhere

Dschoordsch (approved):

Issue #7381, opened 2022-11-03
RemoveOrgUserMutation error


nickoferrall:
Affected 35 users in the past 21 days

https://sentry.io/organizations/parabol/issues/3667878032/?referrer=github_plugin

```
TypeError: Cannot read properties of undefined (reading 'getValue')
  at apply (./packages/client/mutations/RemoveOrgUserMutation.ts:174:60)
  at Set.forEach (<anonymous>)
  at Set.forEach (<anonymous>)
  at Set.forEach (<anonymous>)
...
(28 additional frame(s) were not displayed)
```

jordanh:
Watermark: 

EVENTS
[48](https://sentry.io/organizations/parabol/issues/3667878032/events/?project=107196&referrer=github_plugin)
USERS
[40](https://sentry.io/organizations/parabol/issues/3667878032/tags/user/?project=107196&referrer=github_plugin)

PR #7721, opened 2023-02-07
Feat/7409/remove proration

Dschoordsch:
# Description

## Demo

## Testing scenarios

- [ ] Scenario A
  - Step 1
  - Step 2...

- [ ] Scenario B
  - Step 1
  - Step 2....

## Final checklist


Dschoordsch:
@nickoferrall FYI I think this should also be reviewed by Matt afterwards.

nickoferrall (approved):
Tested and it's working well!

For Matt: you may notice that when removing a user from an org, that user hits a fatal bug. There's already an issue for that (https://github.com/ParabolInc/parabol/issues/7381) and a PR too

packages/server/graphql/queries/invoices.ts
```diff
@@ -54,8 +54,19 @@ export default {
...
+        .filter({
+          inactive: false,
+          removedAt: null
+        })
+        .count()
+        .run()
     ])
+    const upcomingInvoice = after
+      ? undefined
+      : await makeUpcomingInvoice(orgId, orgUserCount, stripeId)
```
mattkrick, line 69: +1 can this be called in parallel with the calls above?

mattkrick (changes requested):
looks great! just 1 little safety check to handle concurrent quantity writes + reads

packages/server/billing/helpers/updateSubscriptionQuantity.ts
```diff
@@ -0,0 +1,62 @@
...
+      .getAll(orgId, {index: 'orgId'})
+      .filter({removedAt: null, inactive: false})
+      .count()
+      .run()
+  ])
+  if (!org) throw new Error(`org not found for invoice`)
+  const {stripeSubscriptionId} = org
+  if (!stripeSubscriptionId) return
+
+  const subscription = await manager.retrieveSubscription(stripeSubscriptionId)
```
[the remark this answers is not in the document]
Dschoordsch: :facepalm:

packages/server/graphql/queries/invoices.ts
Dschoordsch: No, it requires `orgUserCount` from those calls.

packages/server/billing/helpers/updateSubscriptionQuantity.ts
```diff
@@ -0,0 +1,68 @@
...
+  const manager = getStripeManager()
+
+  const org = await dataLoader.get('organizations').load(orgId)
+  if (!org) throw new Error(`org not found for invoice`)
+  const {stripeSubscriptionId} = org
+  if (!stripeSubscriptionId) return
+
+  const redisLock = new RedisLockQueue(`updateSubscriptionQuantity:${orgId}`, 5000)
+  try {
+    await redisLock.lock(10000)
```
mattkrick, line 27: -1 i may be misinterpreting this API, but I think this says each lock will last a max of 5 seconds & wait up to 10 seconds to achieve a lock.

if a call to stripe takes 1 second max & we have 2 calls to stripe per function, that means this'll timeout if we have > 5 users joining the org at the same time.
I know in the past we've had something like 50 users migrated to an org. we may want to bump this up by 10-100x just because if it fails this'll be really hard to track down.

Dschoordsch: You're correct with the timeouts for the locks, but I don't think it's an issue. If we have 50 users migrated, we don't need to run this 50 times but actually only once at the end. Because the locked section will read the last user count from the subscription **and** the db, it doesn't matter which of the 50 jobs comes through as long as there is one job which runs after all changes are done. The current setting should achieve this.

I will add it as a comment.

Dschoordsch:
@mattkrick can you check if I made a mistake in my chain of thought?

mattkrick:
The good news: It works!
The bad news: it throws an error if the entire operation lasts > 10 seconds. the error is harmless, but it doesn't look harmless. I'd like to prevent it from being thrown if at all possible.

The error: `Error: Could not achieve lock for lock:queue:updateSubscriptionQuantity:luolDUtsU8`.

possible solutions:
- increase the lock wait limit (it could seriously be 10 mins if we wanted, doesn't really matter)
- add a debounce to that function so that it only calls stripe after the `orgUserCount` has stopped changing (it just needs to run once per org, but we're calling it once per orguser)
- ???

PR #7722, opened 2023-02-07
chore(team-limits): Remove scheduled jobs when upgrade or delete a team

### LINKED ISSUE #7698: Teams limit: cancel scheduled job is user upgrades
If an org reaches the teams limit, we add the `scheduledLockAt` date. If they upgrade before that date, we need to cancel the scheduled job to prevent them from receiving warning & locked emails.

See comment here for more detail: https://github.com/ParabolInc/parabol/pull/7637#discussion_r1089589010

### AC

- [ ] Upgrading to Team or Enterprise automatically cancels the scheduled job to send warning and locked emails

igorlesnenko:
fixes #7698


nickoferrall (changes requested):
I'm getting the following error in the console: 

```
SEND TO SENTRY [Error: Cannot return null for non-nullable field NotifyTeamsLimitExceeded.orgName.] undefined
```
I checked out the parent branch and then checked out this one but got the same result. I also removed node modules. 

Perhaps I'm missing a migration?

igorlesnenko:
@nickoferrall that is because we previously created notifications with another set of data.

Can you delete them?

```
r.db('actionDevelopment')
    .table('Notification')
   .filter({type:'TEAMS_LIMIT_REMINDER' })
   .delete()
   
    r.db('actionDevelopment')
    .table('Notification')
   .filter({type:'TEAMS_LIMIT_EXCEEDED' })
   .delete()
     
     ````

**Review by nickoferrall — APPROVED** [review]
That's working for me 👍 

Although if I delete the org rather than the team, it looks like the scheduled jobs aren't removed. Perhaps another small issue for that use case?

**Comment by igorlesnenko** [comment]
@nickoferrall added to this PR

File: packages/server/billing/helpers/teamLimitsCheck.ts [hunk 1]
```diff
@@ -110,6 +110,19 @@ const isLimitExceeded = async (orgId: string, dataLoader: DataLoaderWorker) => {
     .run()
 }
 
+export const removeTeamLimitsJobs = async (orgId: string) => {
+  const removeJobTypes = ['LOCK_ORGANIZATION', 'WARN_ORGANIZATION']
+  return r
+    .table('ScheduledJob')
+    .filter((row: RValue) => {
```
Dschoordsch, line 117: -1 I'm a bit worried about performance here. Right now we're not having many `ScheduledJob`s, but this will change once we enable the team limit. For this case we should make better use of indexes. I think creating one for `orgId` should be sufficient, so we can do
```ts
r
    .table('ScheduledJob')
    .getAll(orgId, {index: 'orgId'})
    .filter((row: RValue) => (r.expr(removeJobTypes).contains(row('type')))_
    .delete()
    .run()
```

packages/server/graphql/mutations/archiveOrganization.ts
```diff
@@ -1,5 +1,6 @@
 import {GraphQLID, GraphQLNonNull} from 'graphql'
 import {SubscriptionChannel} from 'parabol-client/types/constEnums'
+import {removeTeamLimitsJobs} from '../../billing/helpers/teamLimitsCheck'
```
Dschoordsch, line 3: +1 I think this indicates that `teamLimitsCheck` should be in its own file

Dschoordsch (changes requested):
Did not test

packages/server/graphql/mutations/archiveOrganization.ts
igorlesnenko: @Dschoordsch do you mean move ‘removeTeamLimitsJobs’ in a separate file?

Dschoordsch: Yes, that's what I meant and almost what I wrote.

Dschoordsch:
@igorlesnenko can you please rebase the change so it does not contain 47 changed files?

igorlesnenko:
@Dschoordsch done

igorlesnenko:
fixed migration

Dschoordsch (approved):
Sorry for taking so long. I confused Notifications with ScheduledJobs. I created #7746 because we're not clearing notifications.

I only tested deleting a team.

Issue #7695, opened 2023-01-30
Checkout Flow: add left sidebar


nickoferrall:
Navigating to the org page (`me/organizations/${orgId}`) results in the left sidebar being replaced with the left sidebar in the designs [here](https://www.figma.com/file/SyJdkzumWsUJo58cmBGgus/Checkout-concepts?node-id=209%3A3221&t=89Ho501858UU34wG-0).

<img width="274" alt="left-sidebar" src="https://user-images.githubusercontent.com/39854876/215601735-fef04efa-c73e-41d2-89ad-67b3f1922d71.png">

### AC

- [ ] Org page shows the left sidebar that looks like the designs
- [ ] Only shows the Members, Plans & Billing, and Usage items (and only shows Usage if this is enabled)
- [ ] Clicking the Members menu item takes the user to the Members page, `/members`
- [ ] Clicking Plans & Billing takes the user to the new Checkout Flow, which has the same URL `/billing`
- [ ] Clicking Usage takes them to the same `/usage` page that exists in our current app
- [ ] Clicking the back Organizations button takes them back to the `/me/organizations` page

acressall:
@nickoferrall a few notes on the first iteration:

- We'll exclude "Teams" for now. When/if we decide to introduce public & private teams, we can bring in this view
- Do you think we should move usage stats here at some point? Rather than going somewhere else for it? 

[here is the design](https://www.figma.com/file/SyJdkzumWsUJo58cmBGgus/Checkout-concepts?node-id=161%3A2088&t=gwfEUONL7iNnfmn6-4) for the organization details view.
![Desktop - org details](https://user-images.githubusercontent.com/18582225/215784344-44782ead-83dd-40c9-9952-e8bdbc9fd55b.png)


Also, [here's](https://www.figma.com/file/SyJdkzumWsUJo58cmBGgus/Checkout-concepts?node-id=165%3A4221&t=gwfEUONL7iNnfmn6-4) how it will look on mobile. Clicking the hamburger in the main nav will open/close the sidebar. 
![Mobile - billing - nav open](https://user-images.githubusercontent.com/18582225/215785324-82141e9e-84fb-4164-ac5e-0cbd41a76bc9.png)

nickoferrall:
> We'll exclude "Teams" for now

Yeah, in the AC I suggested that we only show: `the Members, Plans & Billing, and Usage items`. Does that sound right?

> Do you think we should move usage stats here at some point? Rather than going somewhere else for it?

[...]

Issue #6062, opened 2022-02-15
Use MaterialUI SvgIcons instead of Font


blimmer:
## Issue - Enhancement

Today, we ship two quite large font icons for Material UI:

```console
> ll -h packages/client/styles/theme/fonts/MaterialIcons*
-rw-r--r-- 1 blimmer staff 103K Feb  1 09:49 packages/client/styles/theme/fonts/MaterialIcons-Regular.woff2
-rw-r--r-- 1 blimmer staff 113K Feb  1 09:49 packages/client/styles/theme/fonts/MaterialIconsOutlined-Regular.woff2
```

This pattern means that we ship every icon (used or unused) to the client, which is inefficient. If we switched over to the React `@mui/icons-material` package, we could use the SVG versions of just the icons we use in the app. This should result in a far smaller payload to deliver the icons we use.

Additionally, on slow internet connections, there's a strange behavior where you see the "name" of the icon in the DOM while the font is loading. This breaks the layout and overall looks a bit odd. Check out this Loom recording to see what I'm talking about: https://www.loom.com/share/9bde6426362d4c8ca97ff7a3d6039a83

## Accessibility

I noticed this issue while working on the follow-up to #5961. When I requested the text content of a submitted retrospective item, the string I entered oddly had `"cancelsearch"` appended to the end of string:

```
Error: expect(received).toHaveText(expected)

Expected string: "We should start doing this!"
Received string: "We should start doing this!cancelsearch"
Call log:
  - expect.toHaveText with timeout 5000ms
  - waiting for selector "[data-cy=reflection-column-Start] [data-cy=reflection-stack-Start]"
```

When I inspected the HTML element, indeed, the `<i>` element has the name of the element:

![Screen Shot 2022-02-15 at 12 59 18](https://user-images.githubusercontent.com/630449/154139442-fa270fd8-4fc9-4a37-905d-a96ee79a4a3c.png)

I listened to this with VoiceOver on my Mac and it wasn't announcing the `cancel` text, so I don't think there's necessarily an accessibility issue with the current pattern. However, some screen readers might also behave differently than VoiceOver.

### Acceptance Criteria

- SVG material icons are used instead of the MUI font icons.
- Unused icons aren't included in the final bundle (see https://mui.com/components/icons/#usage).

**Estimated effort:** 12 points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

[...]

PR #7733, opened 2023-02-08
feat(checkout-flow): left sidebar

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7695

Loom demo: https://www.loom.com/share/7287b3edba524f4cb532e829e6e6e541

### To test

- [ ] Add the `checkoutFlow` feature flag
- [ ] Go to an Organization page, i.e. `/me/organization/${orgId}`, and see that the sidebar contents is replaced with `Plans & Billing` and `Members`
- [ ] On mobile, see the new sidebar


packages/client/modules/teamDashboard/components/TeamTasksHeader/TeamTasksHeader.tsx
```diff
@@ -125,7 +125,7 @@ const TeamTasksHeader = (props: Props) => {
                   <NavLink
                     className={css(linkStyles)}
                     title={orgName}
-                    to={`/me/organizations/${orgId}`}
+                    to={`/me/organizations/${orgId}/billing`}
```
nickoferrall, line 128: `/me/organizations/${orgId}` defaults to billing. I think we should be explicit

packages/client/modules/userDashboard/components/Organization/OrganizationPage.tsx
```diff
@@ -36,7 +36,7 @@ const OrganizationPage = (props: Props) => {
           <Route
             exact
             path={`${match.url}`}
-            render={(p) => <OrgBilling {...p} organization={organization} />}
+            render={() => <Redirect to={`${match.url}/${BILLING_PAGE}`} />}
```
nickoferrall, line 39: As `/me/organizations/${orgId}` defaults to Billing, we may as well redirect them to the billing page. This makes highlighting the active left nav item easier

packages/client/modules/userDashboard/components/UserDashboard/UserDashboard.tsx
```diff
@@ -23,11 +23,11 @@ const UserProfile = lazy(
 
 interface Props {
   match: any
-  notifications: any
```
nickoferrall, line 26: This wasn't being used

nickoferrall:
Hey @acressall, would you mind checking out the Loom in the description to see if this looks good to you? See this comment [here](https://github.com/ParabolInc/parabol/issues/7695#issuecomment-1422484217) on the problem I ran into with Usage

packages/client/components/Dashboard/LeftDashNavItem.tsx
```diff
@@ -44,13 +54,25 @@ const Label = styled('div')({
...
+  group: <Group />,
+  warning: <Warning />
+}
+
 interface Props {
   className?: string
   onClick?: () => void
   label: string
   href: string
   //FIXME 6062: change to React.ComponentType
```
igorlesnenko, line 74: Do we still need this comment?

packages/client/components/Dashboard/DashSidebar.tsx
```diff
@@ -57,8 +54,78 @@ const Wrapper = styled('div')({
...
+              <NavItem icon={'arrowBack'} href={'/me'} label={'Back'} />
+              <OrgName>{name}</OrgName>
+              <NavItem
+                icon={'creditScore'}
+                href={`/me/organizations/${orgId}/billing`}
+                label={'Plans & Billing'}
+              />
+              <NavItem
+                icon={'group'}
+                href={`/me/organizations/${orgId}/member`}
```
igorlesnenko, line 119: -1 The url should be 'members', otherwise it won't open member tab

igorlesnenko (changes requested):

packages/client/components/Dashboard/DashSidebar.tsx
nickoferrall: Thanks, updated 👍

packages/client/components/Dashboard/LeftDashNavItem.tsx
nickoferrall: I'm pretty sure we don't as issue https://github.com/ParabolInc/parabol/issues/6062 was fixed by @Dschoordsch's PR https://github.com/ParabolInc/parabol/pull/7154. 

This comment appears 14 times in the app, so I'll wait for Georg to confirm that we don't need it before removing them

igorlesnenko (approved):

packages/client/components/Dashboard/LeftDashNavItem.tsx
Dschoordsch: Good point, thanks for bringing up! That's a cleanup we should do but I didn't back then because the PR was already quite big. I'll create an issue for it.

packages/client/components/Dashboard/DashSidebar.tsx
```diff
@@ -57,8 +55,78 @@ const Wrapper = styled('div')({
...
 })
 
+const OrgName = styled('div')({
+  paddingTop: 8,
+  paddingLeft: 8,
+  fontWeight: 600,
+  fontSize: 12,
+  lineHeight: '24px',
+  color: PALETTE.SLATE_500,
+  [makeMinWidthMediaQuery(Breakpoint.SIDEBAR_LEFT)]: {
```
Dschoordsch, line 65: -1 this breakpoint seems unnecessary, see later comment

packages/client/components/Dashboard/DashSidebar.tsx
```diff
@@ -57,8 +55,78 @@ const Wrapper = styled('div')({
...
+    `,
+    viewerRef
+  )
+
+  if (!viewer) return null
+  const {featureFlags, organizations} = viewer
+  const showOrgSidebar = featureFlags.checkoutFlow && pathname.startsWith(`/me/organizations`)
+
+  if (showOrgSidebar) {
+    const orgIdFromPathname = pathname.split('/')[3]
```
Dschoordsch, line 102: -1 if we switch to `useMatch` we can read this param via
```ts
const {orgId} = orgRouteMatch.params
```

packages/client/components/Dashboard/DashSidebar.tsx
```diff
@@ -57,8 +55,78 @@ const Wrapper = styled('div')({
...
+          name
+        }
+      }
+    `,
+    viewerRef
+  )
+
+  if (!viewer) return null
+  const {featureFlags, organizations} = viewer
+  const showOrgSidebar = featureFlags.checkoutFlow && pathname.startsWith(`/me/organizations`)
```
Dschoordsch, line 99: -1 instead of `pathname.startsWith` we should use
```ts
const orgRouteMatch = useMatch<{orgId: string}>('/me/organizations/:orgId')
const showOrgSidebar = featureFlags.checkoutFlow && orgRouteMatch
```
You have to check if you need to set `strict` or `exact` in addition.

packages/client/components/Dashboard/MobileDashSidebar.tsx
```diff
@@ -70,7 +85,66 @@ const Footer = styled('div')({
...
+        }
+      }
+    `,
+    viewerRef
+  )
+  const location = useLocation()
+  const {pathname} = location
+  if (!viewer) return null
+  const {featureFlags, organizations} = viewer
+  const showOrgSidebar = featureFlags.checkoutFlow && pathname.startsWith(`/me/organizations`)
```
Dschoordsch, line 110: -1 Let's use `useMatch` here as well

packages/client/components/Dashboard/MobileDashSidebar.tsx
```diff
@@ -70,7 +85,66 @@ const Footer = styled('div')({
...
+
+  if (showOrgSidebar) {
+    const orgIdFromPathname = pathname.split('/')[3]
```
Dschoordsch, line 113: -1 see above

Dschoordsch (changes requested):
* I noticed there is some highlight issue when going back and forth via the "back" link or back/forward buttons in the browser
Peek 2023-02-09 15-13
* How about not navigating to a different page when pressing the "back" link, but instead just showing the global sidebar?
* It feels odd that "Organizations" is only reachable directly via the profile menu on desktop and not at all on mobile. There the "Back to organizations" link seems to be the only way to get there. cc: @acressall

acressall:
> * How about not navigating to a different page when pressing the "back" link, but instead just showing the global sidebar?
Ohh this is an interesting idea.. I lean towards going back a page still, but I have a 1:1 with Terry soon and will chat with him about it

> * It feels odd that "Organizations" is only reachable directly via the profile menu on desktop and not at all on mobile. There the "Back to organizations" link seems to be the only way to get there. cc: @acressall
Yeah, it's not ideal that the profile nav goes away on mobile. It might be time to take a broader look at our navigation as a whole -- I'll have a chat with Terry about this as well.

For now, I think it's ok to keep as-is.
@nickoferrall I do agree with you about keeping the add meeting button for consistency. I don't imagine many people will click it from here, but it might be more inline with muscle memory of where the nav is

packages/client/components/Dashboard/MobileDashSidebar.tsx
```diff
@@ -45,6 +48,18 @@ const Nav = styled('nav')({
...
 })
 
+const OrgName = styled('div')({
+  paddingTop: 8,
+  paddingLeft: 8,
+  fontWeight: 600,
+  fontSize: 12,
+  lineHeight: '24px',
+  color: PALETTE.SLATE_500,
+  [makeMinWidthMediaQuery(Breakpoint.SIDEBAR_LEFT)]: {
```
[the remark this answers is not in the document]
nickoferrall: Good catch - I copied this from another part of the nav. The breakpoint isn't needed

packages/client/components/Dashboard/DashSidebar.tsx
nickoferrall: Oh cool, `useRouteMatch` is way cleaner. Updated!

nickoferrall:
Moving this comment https://github.com/ParabolInc/parabol/issues/7695#issuecomment-1424420101 to this PR so all conversations are in the same place.

> I'm ok with the btn saying "back" as long as it goes back to the last page the user was on

@acressall I think that's a good call, but as I'm building it, I realise that it can mean the user is taken away from our app. For example, if they're on Google, type the URL of the billing page, and click back, they'll be taken back to Google rather than a Parabol page.

Also, if they were on the Billing page, then clicked on the Members page, and then clicked Back, they'd be taken to the Billing page again, so the page wouldn't change. 

Ideally, I think the button should just take the user back to the `/meetings` page as they're probably clicking that button because they want to see the old sidebar.

Perhaps something like this?

meetings

or this?

home

acressall:
Ahh okay, good catch. 

To avoid this getting messy or needing another design review, lets backtrack to the original design. 

If it says "All Organizations" and goes to the `/organizations` view, the user will get their main navigation back, as well as the option to see other orgs if they need to. This keeps with the current pattern we have right now. 

I don't think this is the _ideal_ flow, but won't hurt to keep as it currently stands. I've raised the need for a more robust nav system with the designers, so it could be something we work on soon before we have a bunch more admin controls

nickoferrall:
Sounds good! I'll implement that shortly 👍

packages/client/components/Dashboard/LeftDashNavItem.tsx
```diff
@@ -26,7 +36,7 @@ const NavItem = styled(PlainButton)<{isActive: boolean}>(({isActive}) => ({
   [makeMinWidthMediaQuery(Breakpoint.SIDEBAR_LEFT)]: {
     borderRadius: '0 4px 4px 0'
   },
-  ':hover,:focus': {
+  ':hover': {
```
nickoferrall, line 39: This fixed the buggy hovering logic for me mentioned here: https://github.com/ParabolInc/parabol/pull/7733#pullrequestreview-1291325958

packages/client/components/Dashboard/LeftDashNavItem.tsx
```diff
@@ -44,48 +54,38 @@ const Label = styled('div')({
...
-  return slashlessPath.startsWith(href)
+  icon: keyof typeof iconLookup
 }
 
 const LeftDashNavItem = (props: Props) => {
   const {className, label, icon, href, onClick} = props
-  const {history} = useRouter()
-  const isActive = getIsActive(href)
+  const history = useHistory()
+  const match = useRouteMatch(href)
```
nickoferrall, line 81: Looks like we can use `useRouteMatch` here as well to simplify the `isActive` logic

Dschoordsch (approved):
Looks good, I did not have the focus issue anymore

Issue #7735, opened 2023-02-08
Recurrence: Meetings on archived teams continue to restart


jmtaber129:
See [Slack](https://parabol.slack.com/archives/C4JAUUZ9P/p1675853193628709) 🔒 

Expected behavior: after archiving a team, any recurring meetings on the team will stop recurring
Actual behavior: after archiving a team, any recurring meetings on the team will continue to restart

We should either a) cancel all meeting series when a team is archived, b) check whether a meeting series' team is archived before starting a new meeting via recurrence, or c) both of these.

PR #7747, opened 2023-02-09
fix(recurrence): Don't restart meetings on archived teams

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/7735

When a team is archived, we don't close any meetings or meeting series (presumably to make undoing that action possible by Parabol employees for support cases). This has a negligible impact for one-off meetings, but because we don't check if a team is active when processing recurrence, recurring meetings on archived teams will continue to recur.

## Testing scenarios
- [ ] Set up a recurring meeting such that the meeting will be stopped + restarted on the next `processRecurrence` run (I just temporarily changed my local machines date to tomorrow after starting a recurring meeting)
- [ ] Archive the team with the recurring meeting
- [ ] Run `processRecurrence`
- [ ] Confirm that the number of meetings started + ended reflects that the meeting was ended but not restarted
- [ ] Optionally, step forward an additional day, run `processRecurrence` again, and confirm that no email summary is sent for the recurring meeting on the archived team.

## Final checklist

### DIFF
--- a/packages/server/graphql/private/mutations/processRecurrence.ts
+++ b/packages/server/graphql/private/mutations/processRecurrence.ts
@@ -82,7 +82,7 @@ const processRecurrence: MutationResolvers['processRecurrence'] = async (_source
   await Promise.all(
     activeMeetingSeries.map(async (meetingSeries) => {
       const seriesTeam = await dataLoader.get('teams').loadNonNull(meetingSeries.teamId)
-      if (seriesTeam.isArchived) {
+      if (seriesTeam.isArchived || !seriesTeam.isPaid) {
         return
       }
 
--- a/packages/server/graphql/private/mutations/processRecurrence.ts
+++ b/packages/server/graphql/private/mutations/processRecurrence.ts
@@ -81,6 +81,11 @@ const processRecurrence: MutationResolvers['processRecurrence'] = async (_source
   const activeMeetingSeries = await getActiveMeetingSeries()
   await Promise.all(
     activeMeetingSeries.map(async (meetingSeries) => {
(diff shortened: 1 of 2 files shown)


packages/server/graphql/private/mutations/processRecurrence.ts
```diff
@@ -81,6 +81,11 @@ const processRecurrence: MutationResolvers['processRecurrence'] = async (_source
   const activeMeetingSeries = await getActiveMeetingSeries()
   await Promise.all(
     activeMeetingSeries.map(async (meetingSeries) => {
+      const seriesTeam = await dataLoader.get('teams').loadNonNull(meetingSeries.teamId)
+      if (seriesTeam.isArchived) {
```
BartoszJarocki, line 85: good catch! now, should we also check if organization is locked?
CC @nickoferrall

jmtaber129: Good question! Probably not, but curious what @nickoferrall thinks.

I just pushed a commit that does this, but I can revert if that's not what we want.

nickoferrall: As we're preventing other meetings from starting if it's locked, I'd say we should also prevent standups from recurring if the org is locked or the team's `isPaid` field is false. See logic [here](https://github.com/ParabolInc/parabol/blob/b99805925134225e941241fa459cdf0d68b2ab49/packages/server/graphql/mutations/helpers/isStartMeetingLocked.ts)

BartoszJarocki (approved):
lgtm!

PR #7749, opened 2023-02-09
chore: fix case-sensitive

mattkrick:
# Description

fix the case of a file. hopefully github is case-sensitive? i forgot to reformat my local machine to be case sensitive


mattkrick:
the failing case is an unreachable .deb, should probably work today?

Issue #7434, opened 2022-11-16
Add seasonal retrospective templates to the product


garethaledavies:
I've got a few more templates I'd like to add into the product:

---

**Holi retrospective**

**Colors**
_What color represents our recent work? Why?_

**Music** 
_Are we all dancing to the same rhythm? Where were we in sync? When did we fall out of sync?_

**Dance** 
_What big wins make us want to dance? What moves or actions should we try next?_

**Sweets** 
_Who deserves something sweet, kudos, or a word of thanks?_

---

**Easter retrospective**

**Easter eggs** 
_What was hidden from sight or caught you by surprise? What did you learn or find out?_

**Seeds** 
_What new processes would we like to cultivate? Is there something that faded away that we should bring back?_

**Hope** 
_What did you hope for? What are you hoping for next?_

**Chocolate**
_Who deserves some kudos or a word of thanks?_

---

**Midsummer retrospective**

**Wreaths** 
_What processes and practices are binding us together and making us stronger?_

**Schnapps**
_What’s been hard to swallow, admit, or realize about our work?_

**Longest day**
_What feels like it took too long or was never-ending?_

**Strawberry cake**
_Who deserves some kudos or a word of thanks?_

---

**Lunar new year retrospective**

**🧧 Hongbao** 
_Who deserves a red envelope with kudos and some words of appreciation?_ 

**🎊 Couplets** 
_What important messages should we remember for guidance?_

**🧨 Firecracker** 
_What caught us by surprise or gave us a fright?_

I would love to get @tianrunhe's feedback on the lunar new year retro before we put these into the product!

tianrunhe:
Wooo I love a lunar new year theme template! My thoughts:
1. We stick with English translations for broader audience, so I'll replace "Hongbao" with "Red Envelope"
2. Lanterns are more of culture symbols for [Lantern Festival](https://en.wikipedia.org/wiki/Lantern_Festival) over Lunar new year. A more Lunar NY themed symbol would be [couplets](https://en.wikipedia.org/wiki/Antithetical_couplet). The prompt could be sth like "What should we hang up on our door as our guidance?". I'm not a good writer 😂 @garethaledavies come up with sth better than mine please

garethaledavies:
> We stick with English translations for broader audience, so I'll replace "Hongbao" with "Red Envelope"

[...]

PR #7760, opened 2023-02-13
feat: add seasonal retros

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7434

Changes I've made from the issue:

- Capitalised retrospective to be consistent with the others, so `Holi Retrospective` instead of `Holi retrospective`
- Capitalise question titles, so `Easter Eggs` instead of `Easter eggs`
- Move the emoji to the end of the question title as the formatting can look strange on 

@garethaledavies do those changes work for you and would you mind reviewing the PR to make sure the copy is correct?

Here's a screenshot of one of them:
new-retro

### To test

- [ ] Run the migration and find the retro templates that were mentioned in the linked issue all exist
- [ ] Check that the prompts looks correct 
- [ ] Start a retro meeting with one of the new templates and see it works

### DIFF
--- a/packages/server/postgres/migrations/1676299334703_addSeasonalTemplates.ts
+++ b/packages/server/postgres/migrations/1676299334703_addSeasonalTemplates.ts
@@ -1,0 +1,198 @@
+import {R} from 'rethinkdb-ts'
+import {PALETTE} from '../../../client/styles/paletteV3'
+
+const createdAt = new Date()
+
+const makeId = (name: string, type: 'template' | 'prompt') => {
+  const cleanedName = name
+    .replace(/[^0-9a-z-A-Z ]/g, '') // remove emojis and apostrophes
+    .split(' ')
+    .map(
+      (name, idx) =>
+        (idx === 0 ? name.charAt(0).toLowerCase() : name.charAt(0).toUpperCase()) + name.slice(1)
+    )
+    .join('')
+    .trim()
+  return `${cleanedName}${type === 'template' ? 'Template' : 'Prompt'}`
+}
+
+const makeTemplate = (name: string) => ({
+  createdAt,
+  id: makeId(name, 'template'),
+  isActive: true,
+  name,
+  orgId: 'aGhostOrg',
+  scope: 'PUBLIC',
+  teamId: 'aGhostTeam',
+  type: 'retrospective',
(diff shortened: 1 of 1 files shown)


packages/server/postgres/migrations/1676299334703_addSeasonalTemplates.ts
```diff
@@ -0,0 +1,211 @@
...
+    host,
+    port: parseInt(port, 10),
+    db: pathname.split('/')[1]
+  })
+  return r as any as ParabolR
+}
+
+export const up = async function () {
+  try {
+    await connectRethinkDB()
```
igorlesnenko, line 189: I see that everywhere else in the end of the function we call the `await r.getPoolMaster()?.drain()`
Do we need this here too or not necessary?

igorlesnenko (approved):
Works!

packages/server/postgres/migrations/1676299334703_addSeasonalTemplates.ts
```diff
@@ -0,0 +1,213 @@
...
+    await Promise.all([
+      r.table('MeetingTemplate').insert(templates).run(),
+      r.table('ReflectPrompt').insert(reflectPrompts).run()
+    ])
+    await r.getPoolMaster()?.drain()
+  } catch (e) {
```
Dschoordsch, line 195: -1 Any reason why we ignore errors here?

Dschoordsch (changes requested):

packages/server/postgres/migrations/1676299334703_addSeasonalTemplates.ts
nickoferrall: That's the pattern we have in many of our other migrations, see [here](https://github.com/ParabolInc/parabol/blob/ec1cc40f177b02a0a5f37b8e7cd7b422b3f7961e/packages/server/database/migrations/20220602164119-fixRetroTemplates.ts) for example. Would you prefer to send the error to Sentry?

Dschoordsch: Let's go with it, it will not break the app if we're missing some templates. Otherwise we should not ignore errors in migrations.

Dschoordsch (approved):

Issue #7764, opened 2023-02-13
Remove user feature flag


nickoferrall:
At the moment, we have an `addFeatureFlag` mutation, but we don't have a mutation to remove the flag.

Some users in the AI Summary A/B test might want to turn the feature off, so we should build this sooner rather than later.

### AC

- [ ] Can remove a user feature flag with a mutation

PR #7765, opened 2023-02-13
chore: remove user feature flag

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7764

If the PR is approved as it is, I'll update our Notion docs to reflect the change.

### To test

- [ ] Run the `updateFeatureFlag` mutation with `addFlag` as true. Go to pgadmin and see that a flag was added
- [ ] Run the `updateFeatureFlag` mutation with `addFlag` as false. Go to pgadmin and see that the flag was removed


packages/client/mutations/UpdateFeatureFlagMutation.ts
```diff
@@ -0,0 +1,41 @@
+import graphql from 'babel-plugin-relay/macro'
+import {commitMutation} from 'react-relay'
+import {StandardMutation} from '../types/relayMutations'
+import {UpdateFeatureFlagMutation as TUpdateFeatureFlagMutation} from '../__generated__/UpdateFeatureFlagMutation.graphql'
+
+graphql`
+  fragment UpdateFeatureFlagMutation_notification on UpdateFeatureFlagPayload {
+    user {
+      id
+      # add flag here
```
igorlesnenko, line 10: I'm not sure what that comment means? Do we use this mutation on client? Maybe remove it completely, or do we plan to use it in future?

igorlesnenko (approved):
Works!

packages/client/mutations/UpdateFeatureFlagMutation.ts
nickoferrall: I took this from the `AddFeatureFlagMutation` - I think if we want to notify users, we can include the flag here

nickoferrall:
@igorlesnenko I meant to include the `One Review Required` label in this PR. Any objections to me merging this PR or do you think I should request a Maintainer Review?

igorlesnenko:
@nickoferrall No objections

packages/client/mutations/UpdateFeatureFlagMutation.ts
igorlesnenko: Yeah, I mean, I don't see we use `AddFeatureFlagMutation` or `UpdateFeatureFlagMutation` on client. We export it but don't import

nickoferrall: I think we used to let users add themselves to a feature using a cheat code, like if they typed in "Spotlight", they'd get access to it.

I could see that happening again in the future. For instance, if they could request access to the Insights beta feature and we'd automatically turn it on. So I think it's safe to keep it

PR #7766, opened 2023-02-14
fix: add args to dd-trace

mattkrick:
# Description

We had downtime today & we're still not sure what causes it.
We saw many of the long running queries occurred on the same team. 
It was possible to find the team by taking an intersection of all the userIds, but it'd be much easier to just have that teamId ready in datadog APM.

However, we don't want to store sensitive arguments like password.
This PR sends arguments to datadog & allows you to exclude arguments that may be sensitive.
Screenshot 2023-03-17 at 3 33 04 PM

# Test
- [ ] run a query on datadog using the staging key & see the arguments show up. make sure nothing sensitive shows up! https://app.datadoghq.com/apm/traces?query=%40_top_level%3A1%20env%3Alocal&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&historicalData=false&messageDisplay=inline&sort=desc&spanID=2268273604486304113&spanType=service-entry&timeHint=1679092353722&trace=740778531903268531740778531903268531&traceID=740778531903268531&start=1679091466124&end=1679092366124&paused=false


packages/server/graphql/CompiledQueryCache.ts
```diff
@@ -3,9 +3,23 @@ import {GraphQLSchema, parse} from 'graphql'
...
+}
+
 const compileQuery = tracedCompileQuery(tracer, {
+  excludeArgs: {
+    loginWithPassword: ['password'],
+    resetPassword: ['token', 'newPassword'],
+    signUpWithPassword: ['password', 'invitationToken'],
+    updateCreditCard: ['stripeToken'],
+    verifyEmail: ['verificationToken']
+  } as ExcludedArgs,
```
mattkrick, line 22: this is just to provide autocomplete & prevent typos when adding new ones

docker/dev.yml
```diff
@@ -1,6 +1,18 @@
 version: "3.7"
 
 services:
+  datadog:
```
mattkrick, line 4: we're no longer exposing our DD agent to the public internet, so we must spin one up on our own machine

packages/server/graphql/traceGraphQL.ts
```diff
@@ -220,7 +240,8 @@ function wrappedResolve(
     }
 
     const field = assertField(tracer, context, info, path)
-
+    const tags = addVariableTags(args, info.fieldName, config.excludeArgs)
+    if (tags) field.span.addTags(tags)
```
mattkrick, line 244: I'm putting them here on the resolve function vs. putting them on the top of the query.
This is because the mutation name (AKA operationName) is defined by the client (e.g. LoginMutation). If we change/add/remove that name on the client, i don't want to accidentally send sensitive info to dd.

It's also nice because on extra large queries with a TON of variables you only see the variables that make it to the specific field.

mattkrick:
ready for review @Dschoordsch

packages/server/graphql/CompiledQueryCache.ts
```diff
@@ -3,9 +3,23 @@ import {GraphQLSchema, parse} from 'graphql'
...
+type ExtractArgs<T> = T extends Resolver<any, any, any, infer Args> ? Args : never
+type ExcludedArgs = {
+  [P in keyof OperationResolvers]?: Array<keyof ExtractArgs<OperationResolvers[P]>>
+}
+
 const compileQuery = tracedCompileQuery(tracer, {
+  excludeArgs: {
+    loginWithPassword: ['password'],
+    resetPassword: ['token', 'newPassword'],
+    signUpWithPassword: ['password', 'invitationToken'],
```
Dschoordsch, line 19: +1 If we want to exclude the `invitationToken` here, do we also want to exclude it from `acceptTeamInvitation`? It does not appear to be that sensible, but we should be consistent.

packages/server/graphql/CompiledQueryCache.ts
Dschoordsch: +2 nice

docker/dev.yml
Dschoordsch: I had no luck using this agent. Do we need updated API keys?

mattkrick: yes! i don't think action-devops has the latest key. Rafa is gonna update them, i didn't want to meddle.
i grabbed mine directly from DD. Key ending in `eda33`

PR #7772, opened 2023-02-14
fix: do not run removeOrgUserTaskUpdater for undefined tasks

### LINKED ISSUE #7381: RemoveOrgUserMutation error

BartoszJarocki:
# Description

Fixes #7381 

## Demo

No demo

## Testing scenarios

- [ ] User removing during the meeting
 - user A is an org billing leader, user B is a user in the user's A org
 - start a check in meeting
 - from a different browser window, join with user B, set a stage to user's B update
 - as a user A, go to org settings and remove user B from org
 - user B should be redirected to the dashboard

- [ ] User with no tasks removed 
 - user B has no tasks
 - user A is an org billing leader
 - user A removes user B
 - users B app doesn't crash

## Final checklist

### DIFF
--- a/packages/client/subscriptions/NotificationSubscription.ts
+++ b/packages/client/subscriptions/NotificationSubscription.ts
@@ -267,6 +267,8 @@ const NotificationSubscription = (
           break
         case 'UpdateFeatureFlagPayload':
           break
+        case 'AuthTokenPayload':
+          break
         case 'AddNewFeaturePayload':
           addNewFeatureNotificationUpdater(payload, context)
           break
--- a/packages/client/mutations/RemoveOrgUserMutation.ts
+++ b/packages/client/mutations/RemoveOrgUserMutation.ts
@@ -131,6 +131,7 @@ export const removeOrgUserTeamUpdater: SharedUpdater<RemoveOrgUserMutation_team>
 ) => {
   const removedUserId = payload.getLinkedRecord('user').getValue('id')
   const {viewerId} = atmosphere
+
   if (removedUserId === viewerId) {
     const teams = payload.getLinkedRecords('teams')
     const teamIds = teams.map((team) => team.getValue('id'))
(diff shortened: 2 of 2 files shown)


packages/client/subscriptions/NotificationSubscription.ts
```diff
@@ -257,6 +257,8 @@ const NotificationSubscription = (
           break
         case 'UpdateFeatureFlagPayload':
           break
+        case 'AuthTokenPayload':
```
BartoszJarocki, line 260: This fixes `NotificationSubscription case fail AuthTokenPayload` error showing in the logs. @mattkrick is there anything specific app should do here? just want to make sure it's not a bug and I'm not silencing it

mattkrick: nope, this is just a reeeeeally old pattern that we haven't replaced yet. all the other subscriptions use a lookup table.

jmtaber129:
@BartoszJarocki could you rebase onto latest master and run through the testing scenarios? This doesn't seem to work if I cherry-pick onto master, though there may be a different bug at play here too

BartoszJarocki:
@jmtaber129 I did the testing and here are the results
1. Removing from org without refreshing (same as user without tasks scenario) - works on my end. is something wrong with that scenario on your end? asking because when relay store is involved there might be some race conditions to fix and I just want to make sure it works/doesn't work for everyone

https://user-images.githubusercontent.com/1017620/225063290-cd79b803-03e4-4ecc-9a91-6c3dc770242d.mp4


2. Removing user during the meeting - indeed broken, will take a look if something has changed here

jmtaber129:
@BartoszJarocki after pulling your rebased branch, now things seem to be working 🤔 Not sure what was going on with my local dev with the cherry-picked diff. LGTM!

jmtaber129 (approved):

BartoszJarocki:
not sure, but since there's no harm in merging that, let's do it! thank you for the review 😄

PR #7775, opened 2023-02-15
fix: Fix server error when joining a team by invitation link

### LINKED ISSUE #7771: Error: The loader.load() function must be called with a value,but got: undefined.
https://parabol.sentry.io/issues/1697059572/?referrer=github_plugin

```
Error: The loader.load() function must be called with a value,but got: undefined.
```

Reproduction:
* create a new team
* create an invite link from the team dashboard
* follow the invite link with a different user

igorlesnenko:
Fixes #7771

How to test:

- create a new team
- create an invite link from the team dashboard
- follow the invite link with a different user

### DIFF
--- a/packages/server/graphql/types/AcceptTeamInvitationPayload.ts
+++ b/packages/server/graphql/types/AcceptTeamInvitationPayload.ts
@@ -38,7 +38,7 @@ const AcceptTeamInvitationPayload = new GraphQLObjectType<any, GQLContext>({
       type: NewMeeting,
       description: 'The requested meeting',
       resolve: async (
-        {meetingId}: {meetingId: string},
+        {meetingId}: {meetingId?: string},
         _args: unknown,
         {dataLoader, authToken}: GQLContext
       ) => {
--- a/packages/server/graphql/types/AcceptTeamInvitationPayload.ts
+++ b/packages/server/graphql/types/AcceptTeamInvitationPayload.ts
@@ -42,6 +42,9 @@ const AcceptTeamInvitationPayload = new GraphQLObjectType<any, GQLContext>({
         _args: unknown,
         {dataLoader, authToken}: GQLContext
       ) => {
+        if (!meetingId) {
+          return null
+        }
         const viewerId = getUserId(authToken)
         const meeting = await dataLoader.get('newMeetings').load(meetingId)
         if (!meeting) {
--- a/packages/client/mutations/AcceptTeamInvitationMutation.ts
+++ b/packages/client/mutations/AcceptTeamInvitationMutation.ts
@@ -116,9 +116,10 @@ export const acceptTeamInvitationNotificationUpdater: SharedUpdater<
   const viewer = store.getRoot().getLinkedRecord('viewer')
   if (viewer) {
     const requestedMeeting = payload.getLinkedRecord('meeting')
-    const requestedMeetingId = requestedMeeting.getValue('id')
-    viewer.setLinkedRecord(requestedMeeting, 'meeting', {meetingId: requestedMeetingId})
-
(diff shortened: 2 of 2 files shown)


packages/server/graphql/types/AcceptTeamInvitationPayload.ts
```diff
@@ -42,6 +42,9 @@ const AcceptTeamInvitationPayload = new GraphQLObjectType<any, GQLContext>({
         _args: unknown,
         {dataLoader, authToken}: GQLContext
       ) => {
+        if (!meetingId) {
+          return null
+        }
         const viewerId = getUserId(authToken)
         const meeting = await dataLoader.get('newMeetings').load(meetingId)
         if (!meeting) {
```
nickoferrall, line 50: I think we should remove the error message on line 48 because it's not an error if there's no meetingId

packages/server/graphql/types/AcceptTeamInvitationPayload.ts
```diff
@@ -42,6 +42,9 @@ const AcceptTeamInvitationPayload = new GraphQLObjectType<any, GQLContext>({
         _args: unknown,
```
nickoferrall, line 42: +1 on line 41, let's change it to `meetingId?: string` so we know it can be undefined

nickoferrall (changes requested):
Two small changes requested. Even if they're not ready, I think we should merge this PR before the ship

packages/server/graphql/types/AcceptTeamInvitationPayload.ts
Dschoordsch: I disagree, it's an error if we pass an invalid `meetingId`

nickoferrall: `meetingId` is coming from a [TeamInvitation](https://github.com/ParabolInc/parabol/blob/9177909d9b3bc9cf2ba5d88a32c0ce53b1da177f/packages/server/database/types/TeamInvitation.ts) where `meetingId` can be undefined

Dschoordsch: Yes. We check for null `meetingId` in l45 and return if null was passed. If a supposedly valid `meetingId` is passed but we cannot find a meeting associated with it, that seems to be an error to me.

Dschoordsch: Calling `loadNonNull` would do the trick as well.

igorlesnenko: Yes, I think this is an error about meeting expected but it is not found for some reason.

Dschoordsch (changes requested):
Looks good to me, does not fix the problem completely. While I don't see this server error anymore, I don't get redirected when accepting the invite either.

igorlesnenko:
@Dschoordsch I was trying to reproduce redirect issue, but it works for me. Here is what I was doing: https://www.loom.com/share/e5a64e8cb84547c4a620a629c90c78f9

Is there something you do differently? Do you see any errors in console?

Dschoordsch:
I have `requestedMeeting` being undefined here
https://github.com/ParabolInc/parabol/blob/377b1128dee1f13bbe82329afe15d7d2b34574da/packages/client/mutations/AcceptTeamInvitationMutation.ts#L119
The accepting user has no active meetings in this org.

igorlesnenko:
@Dschoordsch interesting, I never reach that code, because `viewer` is always undefined for me. But I clearly see a bug here. Added a fix

Dschoordsch (approved):
Works!

PR #7781, opened 2023-02-16
chore(teams-limit): remove teams limit notifications after upgrade or removing a team

### LINKED ISSUE #7746: Teams limit: Upgrading does not close nagging bar
## How to reproduce:
Have a Starter team with the teams limit nagging bar showing and upgrade to Team tier

## Expected behaviour:
The snackbar disappears.

## Actual behaviour:
Nagging snackbar stays visible until it's manually dismissed
Screenshot from 2023-02-09 12-01-54

igorlesnenko:
Fixes #7746

**In this PR:**

* Added an ability to update notifications (mark them as clicked and close related snackbar)
* Unpublish teams limit notifications if team was upgraded or removing a team fixed the limit

**How to test:**

- Uncomment `const enum Threshold {`
- Reset an org
```
  .getAll(r.args(['PUT ORG ID HERE']), {index: 'id'})

  .update({
    tierLimitExceededAt: null,
    scheduledLockAt: null,
    lockedAt: null
});
```

- Complete a meeting to exceed teams limit 
- See snackbar appeared and new notifications dot on the bell icon in header
- Do not open notifications center, do not close a snackbar
- Upgrade a team (or remove a team to fix the limit)
- See that teams limit snackbar was closed
- See that notification marked as clicked


packages/client/mutations/toasts/mapTeamsLimitReminderToToast.ts
```diff
@@ -23,7 +24,7 @@ const mapTeamsLimitReminderToToast = (
 
   return {
     autoDismiss: 0,
-    key: `newNotification:${notificationId}`,
+    key: makeNotificationToastKey(notificationId),
```
igorlesnenko, line 27: `notificationId` is a unique value, lets unify the `key` of toasts that mirrors specific notifications. This way we can remove them easily

jmtaber129 (commented):
I'm not sure _deleting_ notifications is the right thing to do here, since deleting a notification entirely could potentially lead to confusion amongst users ("wait, where's that notification go?", especially if the notification was sent to their email). I would suggest the following approach instead:
* Instead of deleting out-of-date notifications, mark them as "read" instead
* Push a new notification that contradicts the previous notification. E.g. if the old notification is "your team is over the free limit, please upgrade", you might send a new notification that says "your team is no longer above the free limit". Then the user could see both the current state and the historical state.

igorlesnenko:
@jmtaber129 make sense, I updated the PR to mark notification as clicked.
For additional `success` notifications I'm going to create a separate issue

nickoferrall:
@igorlesnenko it looks like James is the Reviewer Reviewer for this PR. Would you also like me to review it?

packages/server/graphql/public/types/UpdatedNotification.ts
```diff
@@ -0,0 +1,22 @@
...
+  updatedNotification: async ({updatedNotificationId}, _args: unknown, {dataLoader, authToken}) => {
+    const viewerId = getUserId(authToken)
+    const notification = await dataLoader.get('notifications').load(updatedNotificationId)
+    if (notification.userId !== viewerId) {
+      throw new Error(
+        `Viewer ID does not match notification user ID: notification ${updatedNotificationId} for user ${notification.userId} published to user ${viewerId}`
+      )
+    }
+
+    console.log('returning notification', notification)
```
jmtaber129, line 16: +1 Remove debug line

packages/client/mutations/toasts/updateNotificationToast.ts
```diff
@@ -0,0 +1,27 @@
...
+  }
+`
+
+export const updateNotificationToastOnNext: OnNextHandler<
+  updateNotificationToast_notification,
+  OnNextHistoryContext
+> = (payload, {atmosphere}) => {
+  const {updatedNotification} = payload
+
+  if (updatedNotification.status === 'CLICKED') {
```
jmtaber129, line 21: +1 This should probably include `READ`, too.

jmtaber129 (approved):
+1 It might be good to clear these notification snackbars when the notification is clicked through the notification dropdown, but I don't have a strong preference.

Couple other minor things, otherwise LGTM!

Dschoordsch (changes requested):
Tested, works.

packages/server/graphql/public/types/UpdatedNotification.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export type UpdatedNotificationSource = {updatedNotificationId: string}
+
+const UpdatedNotification: UpdatedNotificationResolvers = {
+  updatedNotification: async ({updatedNotificationId}, _args: unknown, {dataLoader, authToken}) => {
+    const viewerId = getUserId(authToken)
+    const notification = await dataLoader.get('notifications').load(updatedNotificationId)
+    if (notification.userId !== viewerId) {
+      throw new Error(
+        `Viewer ID does not match notification user ID: notification ${updatedNotificationId} for user ${notification.userId} published to user ${viewerId}`
```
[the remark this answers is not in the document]
Dschoordsch: For future reference: Errors are not user facing because they're sanitized: https://github.com/ParabolInc/parabol/blob/master/packages/server/utils/sanitizeGraphQLErrors.ts

PR #7785, opened 2023-02-16
fix: Added automatic docker host platform detection to yarn db:start, db:stop

jordanh:
I got tired of needing to modify dev.yml for my M1 based mac. I changed the dev.yml to accept environment variables that can specify which platform to use, create a script that emits these environment variables, and modified package.json to tie these two elements together.

Tested on M1. It would be good to verify these changes work on an x86 architecture (Linux, OSX, or otherwise)

### M1 architecture machine
- [ ] Run `yarn db:start`, docker-compose starts as expected
- [ ] Run `yarn db:stop`, docker-compose stops as expected

### x86 architecture machine
- [x] Run `yarn db:start`, docker-compose starts as expected
- [x] Run `yarn db:stop`, docker-compose stops as expected

## Final checklist

- [x] I checked the [code review guidelines](../docs/codeReview.md)
- [x] I have added [Metrics Representative](../docs/codeReview.md#metrics-representative) as reviewer(s) if my PR invovles metrics/data/analytics related changes
- [x] I have performed a self-review of my code, the same way I'd do it for any other team member
- [x] I have tested all cases I listed in the testing scenarios and I haven't found any issues or regressions

   (Note: I haven't been able to test x86, since I no longer own a machine with this architecture)

- [x] Whenever I took a non-obvious choice I added a comment explaining why I did it this way
- [x] I added the label `One Review Required` if the PR introduces only minor changes, does not contain any architectural changes or does not introduce any new patterns and I think one review is sufficient'
- [x] PR title is human readable and could be used in changelog

PR #7786, opened 2023-02-16
chore: select best platform for the job

mattkrick:
# Description

fix #7785 

All dev containers run on the architecture as the host machine. no rosetta required.

## Demo

On an M1:
```sh
yarn db:stop
docker image rm redis && docker image rm rethinkdb && docker image rm rediscommander/redis-commander
yarn db:start
```

check docker desktop, see all images are ARM.

### DIFF
--- a/docker/dev.yml
+++ b/docker/dev.yml
@@ -2,7 +2,6 @@ version: "3.7"
 
 services:
   db:
-    platform: linux/x86_64
     image: rethinkdb:latest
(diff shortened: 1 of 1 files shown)


mattkrick:
I realized I'm probably the only one who can test both AMD64 & ARM64 locally so I'm gonna do that & if it works i'll merge

mattkrick:
@jordanh think this would work for ya?

Issue #7529, opened 2022-12-06
Checkout Flow: Add the Plans UI


nickoferrall:
Following the Figma designs [here](https://www.figma.com/file/SyJdkzumWsUJo58cmBGgus/Checkout-concepts?node-id=209%3A3219&t=89Ho501858UU34wG-0), create the UI for the Plans section of the plans page.

<img width="484" alt="Screenshot 2023-01-30 at 15 11 05" src="https://user-images.githubusercontent.com/39854876/215595936-2882d2b8-2894-4c36-a46b-f1a180163e89.png">

### AC

- [ ] Plans UI looks like Figma designs
- [ ] Don't include the teams limit warning message - this will be added here: https://github.com/ParabolInc/parabol/issues/7692
- [ ] UI uses hardcoded data, and clicking CTA button doesn't do anything. Functionality will be added in https://github.com/ParabolInc/parabol/issues/7791

PR #7793, opened 2023-02-20
feat(checkout-flow): plans UI

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7529

I added info icons to Active Teams and Active Members to match Usage Stats:

plan-ui

This PR doesn't include responsive work. The mobile UI will be implemented here: https://github.com/ParabolInc/parabol/issues/7794

Questions: 

- What text should the info icon show when the user hovers over it (in the Team plan subtitle)?
- Are there designs for mobile?

### To test

- [ ] Make sure you've turned on the checkout-flow user flag
- [ ] Click on the organization page, so `/me/organizations/${orgId}/billing` and see the Plans UI
- [ ] Check if the UI reflects the designs [here](https://www.figma.com/file/SyJdkzumWsUJo58cmBGgus/Checkout-concepts?node-id=252%3A4201&t=lGveHwAZpTVeEc3O-0)


packages/client/modules/userDashboard/components/OrgBilling/OrgPlans.tsx
```diff
@@ -0,0 +1,277 @@
...
+        'Essential templates',
+        'Retrospectives, Sprint Poker, Standups, Check-Ins',
+        'Unlimited team members'
+      ],
+      buttonStyle: 'disabled',
+      buttonLabel: 'Current Plan'
+    },
+    {
+      tier: 'team',
+      stats: ['Everything in Starter', 'Premium templates', 'Custom templates', 'Unlimited teams'],
```
nickoferrall, line 216: In the designs, it says `Unlimited team members`, but I've changed that to `Unlimited teams`

nickoferrall:
Hey @acressall, I added a couple of questions to the PR descriptions that it'd be great to get your help on 🙏  Also, would you mind reviewing the PR to see if the UI reflects your designs?

igorlesnenko (changes requested):
I see few difference from the mockup, not sure how important they are:

Can we fix responsiveness?

image

image

- Feels like we also using smaller font sized on the mockup. 

- There is no gap in the mockup

image

nickoferrall:
> Can we fix responsiveness?

This PR doesn't include responsive UI as I'm not sure if there are designs ready for that. I've created an issue for it here: https://github.com/ParabolInc/parabol/issues/7794

> Feels like we also using smaller font sized on the mockup.

I've reduced the font size by 2 pixels. What do you think?

> There is no gap in the mockup

Good catch - fixed!

igorlesnenko:
@nickoferrall make sense

I would also do other adjustments to make this look closer to mockup
- reduce left right paddings around block content
- reduce space between lines
- make font size of count numbers bigger

igorlesnenko (approved):
Works! Minor adjustments to css to make it closer to mockup would be good to have

acressall:
@nickoferrall let's steal the copy and tooltip functionality from usage stats: https://action.parabol.co/usage

On mobile, tapping on the icon should make the tooltip appear.

Taking a task to update the mobile designs for this panel! Will update #7794 when ready!

nickoferrall:
> let's steal the copy and tooltip functionality from usage stats: https://action.parabol.co/usage

I'm referring to the `$6 per active member paid monthly` tooltip, which doesn't exist in Usage stats. I am using the same tooltip copy for Active Members and Active Teams as the Usage stats. 

> Taking a task to update the mobile designs for this panel!

Sounds good! I'll implement your comment `On mobile, tapping on the icon should make the tooltip appear` once those designs are ready.

Have you had the chance to review the PR, @acressall? Once you have, and if you're happy with the design, I'll move it to Maintainer Review.

Alternatively, if you don't have the time, as I know you're partially away, I can move this to Maintainer Review and you can always request changes to the design at a later date.

acressall:
Thanks for the clarification @nickoferrall! Let's use the same language from the credit card section `Active users are anyone who uses Parabol within a billing period`

I haven't had a chance to pull this PR to my local, and don't trust my wifi connection to update it well. If you're ok with it, I'll review with the mobile design PR later

nickoferrall:
Sounds good! I've added the tooltip. I changed the subtitle from "$6 per active member" to "$6 per active user" to be consistent with the tooltip and as I think it's easier to understand. If that doesn't work, let me know! I'll move this along to Maintainer Review

packages/client/modules/userDashboard/components/OrgBilling/OrgPlansAndBillingHeading.tsx
```diff
@@ -0,0 +1,95 @@
...
+import {OrgPlansAndBillingHeading_organization$key} from '../../../../__generated__/OrgPlansAndBillingHeading_organization.graphql'
+
+const Wrapper = styled('div')({
+  alignItems: 'center',
+  color: PALETTE.SLATE_700,
+  display: 'flex',
+  lineHeight: '24px',
+  padding: '16px 0px',
+  flexWrap: 'wrap',
+  maxWidth: 976, // same as usage stats to be consistent
```
Dschoordsch, line 18: +1 Do we want to move it to constEnums then?

packages/client/modules/userDashboard/components/Organization/OrgNav.tsx
```diff
@@ -0,0 +1,67 @@
...
+    <Wrapper>
+      <NavItemLabel onClick={() => history.push('/meetings')}>Dashboard</NavItemLabel>
+      <StyledIcon>
+        <NavigateNextIcon />
+      </StyledIcon>
+      <NavItemLabel onClick={() => history.push('/me/organizations')}>Organization</NavItemLabel>
+      <StyledIcon>
+        <NavigateNextIcon />
+      </StyledIcon>
+      <NavItemLabel isBold>{`${name}'s Org`}</NavItemLabel>
```
Dschoordsch, line 62: +1 This ends up in "Georg Bremer's Org's Org" for me with the starter team. Even with a real name like "Parabol's Org" I feel we could omit the "'s Org"

packages/client/modules/userDashboard/components/Organization/OrgNav.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+const NavigateNextIcon = styled(NavigateNext)({
+  height: 18,
+  color: PALETTE.SLATE_900
+})
+
+const NavItemLabel = styled('span')<{isBold?: boolean}>(({isBold}) => ({
+  fontWeight: isBold ? 600 : 400,
+  '&:hover': {
+    cursor: isBold ? 'default' : 'pointer'
```
Dschoordsch, line 31: -1 `isBold` does not explain why the cursor needs to be different. I think `isCurrent` would show the intent better.

Dschoordsch (changes requested):
Mostly clarifying:
the design looks a bit off on standard HD screens with the big empty space on the right
Screenshot from 2023-02-24 12-14-42
The left side navigation does not work anymore.

nickoferrall:
Thanks for the screenshot. 

Completely agree that it could look better on wide screens. I had followed the width of Usage Stats which has a similar empty space on the right:
usage

If we increase the width of the Plan's section, the boxes' content looks bad. Centering the list makes things look worse.
wide

Instead, I suggest we keep the same width but move the plans section to the centre. There's space on the left and right, but it's less severe, and the Plans content still looks OK:
center-everything

This approach is also consistent with the Billing page in prod:
billing

What do you think @acressall? If you'd prefer to take a different approach, would you mind creating designs for wide monitors, and I'll create a GitHub issue? 

> The left side navigation does not work anymore.

Thanks, fixed!

acressall:
Ahh yes, extra wide is always tricky. I lean towards sticking with the sticky to the left, awkward space on the right option. The main reason, is this keeps the actions close to the navigation, so you shouldn't have to move your cursor way across the screen too much. 
There's also plenty of space for the "Plan Details" panel to come in on the right.

nickoferrall:
Thanks @acressall. 

@Dschoordsch I've followed Alicia's advice and moved the content back to the left so there's plenty of space for the right sidebar to slide into

Dschoordsch (approved):

Issue #7791, opened 2023-02-20
Checkout flow: Plans functionality


nickoferrall:
Predecessor https://github.com/ParabolInc/parabol/issues/7529

### AC

- [ ] Show the Active Teams, Active Members, and Total Meetings stats for the org
- [ ] For users on the Starter tier, The Team plan is currently selected, Starter plan button is disabled, and clicking the Contact button opens a new email tab with `love@parabol.co` as the recipient 
- [ ] For users on the Team tier, the Team button says `Current Plan` and is disabled, while the Starter button now reads `Downgrade`. Downgrade logic will be handled here: https://github.com/ParabolInc/parabol/issues/7697
- [ ] For users on the Enterprise tier, the Team and Starter CTA buttons says "Contact" and open an email addressed to `love@parabol.co`
- [ ] Don't include the teams limit warning message - this will be added here: https://github.com/ParabolInc/parabol/issues/7692

acressall:
I agree with you @nickoferrall ! Option 3 is best

PR #7799, opened 2023-02-21
feat(checkout-flow): plans functionality 

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7791

This PR uses the org data in the Plans UI rather than hardcoded values. 

### To test

- [ ] Make sure the checkout flow flag is on for your user
- [ ] Go to the org page and see that the stats are accurate for your org


packages/server/graphql/public/types/helpers/getActiveTeamCountByOrgIds.ts
```diff
@@ -0,0 +1,76 @@
...
+import {RValue} from '../../../../database/stricterR'
+import TeamMember from '../../../../database/types/TeamMember'
+import {DataLoaderWorker} from '../../../graphql'
+import isValid from '../../../isValid'
+import {InputMaybe} from '../../resolverTypes'
+import {getTeamsByOrgIds} from '../Company'
+
+const THIRTY_DAYS = 1000 * 60 * 60 * 24 * 30
+
+const getActiveTeamCountByOrgIds = async (
```
nickoferrall, line 11: Moved this to its own file so we can use it in Usage Stats and for these org stats

packages/server/graphql/types/Organization.ts
```diff
@@ -52,6 +54,16 @@ const Organization: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<a
...
       }
     },
+    totalMeetingCount: {
+      type: new GraphQLNonNull(GraphQLInt),
+      description: 'The number of meetings that have occurred in the org',
+      resolve: async ({id: orgId}, _args: unknown, {dataLoader}) => {
+        const allTeamsOnOrg = await dataLoader.get('teamsByOrgIds').load(orgId)
+        const teamIds = allTeamsOnOrg.map(({id}) => id)
+        const allMeetings = await dataLoader.get('allMeetingsByTeamId').loadMany(teamIds)
+        return allMeetings.filter(isValid).flatMap((meetings) => meetings).length
```
igorlesnenko, line 64: Looks like we are fetching a lot of objects from the database to show a count. Some organizations might have a lot of meetings in total.
Does it make sense to run a single count query or cache the value in org table and increment?

packages/server/graphql/public/types/helpers/getActiveTeamCountByOrgIds.ts
```diff
@@ -0,0 +1,76 @@
...
+  orgIds: string | string[],
+  dataLoader: DataLoaderWorker,
+  after?: InputMaybe<Date>
+) => {
+  // teams with at least 2 team members who have logged in within the last 30 days & within an active organization that has had a meeting that has an max(createdAt, endedAt) newer than 30 days ago
+  const metAfter = after ? new Date(after) : new Date(Date.now() - THIRTY_DAYS)
```
igorlesnenko, line 17: We have now slightly different algorithms to determine active teams count here and in teams limit. Do you think we should do something about it?

igorlesnenko (changes requested):

packages/server/graphql/public/types/helpers/getActiveTeamCountByOrgIds.ts
nickoferrall: How are we currently defining it in teams limits? I agree that it should be consistent across teams limits, Usage Stats, and the org view

packages/server/graphql/public/types/Company.ts
```diff
@@ -6,11 +6,11 @@ import errorFilter from '../../errorFilter'
...
 import isValid from '../../isValid'
 import {CompanyResolvers} from '../resolverTypes'
-export type CompanySource = {id: string}
+import getActiveTeamCountByOrgIds from './helpers/getActiveTeamCountByOrgIds'
 
-const THIRTY_DAYS = 1000 * 60 * 60 * 24 * 30
+export type CompanySource = {id: string}
 
-const getTeamsByOrgIds = async (
+export const getTeamsByOrgIds = async (
```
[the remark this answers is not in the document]
nickoferrall: Sounds good 👍

packages/client/modules/userDashboard/components/OrgBilling/OrgPlans.tsx
```diff
@@ -175,22 +215,32 @@ const OrgPlans = () => {
...
-      buttonStyle: 'secondary',
-      buttonLabel: 'Contact'
+      buttonStyle: getButtonStyle(tier, 'enterprise'),
+      buttonLabel: getButtonLabel(tier, 'enterprise')
     }
   ] as const
 
+  const handleClick = (label: 'Contact' | 'Select Plan' | 'Downgrade' | 'Current Plan') => {
+    if (label === 'Contact') {
+      window.open('mailto:love@parabol.co', '_blank')
```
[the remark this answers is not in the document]
nickoferrall: I don't see an example of us doing this elsewhere in the app, but there are plenty of places where we open the `mailto:love@parabol.co` email, so I lean towards keeping it this way to be consistent

packages/server/graphql/types/Organization.ts
nickoferrall: Hey @acressall, just wanted to bump this comment ^ as I think it got hidden in the barrage of GH notifications I've sent you

igorlesnenko: @nickoferrall if we can drop it, this would be a solution. In other case if we want to avoid expensive migration, do you think it make sense to run a `.count()` query to rethink db instead of fetching all objects?

packages/client/modules/userDashboard/components/OrgBilling/OrgPlans.tsx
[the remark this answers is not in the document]
igorlesnenko: @nickoferrall I see few places if you search by `ExternalLinks.CONTACT`, but no strong preference from me

packages/server/graphql/public/types/helpers/getActiveTeamCountByOrgIds.ts
igorlesnenko: @nickoferrall for team limits we count sticky teams. Here is the query: 

https://github.com/ParabolInc/parabol/blob/b00aec49e224f499d53ccaca0ef5bf4b5a719b4f/packages/server/billing/helpers/teamLimitsCheck.ts#L85

Sticky team is the team that completed 3 meetings with more than 1 attendee and have had at least 1 meeting in the last 30 days

nickoferrall: Thanks, @igorlesnenko. Which definition do you think we should change? Where did the teams limit definition come from? 

It looks like the main difference is the Usage Stats definition says the 2 team members logged in within the last 30 days. However, if the Team Limits query is checking if there were 2 team members in the meeting that ended within 30 days, then that's essentially the same thing, right?

If that's correct, then perhaps we just update the Teams Limit definition with the Usage Stats definition, as Usage Stats was the original?

Also, I see the Teams Limit definition says there was `at least 1 meeting in the last 30 days` but I don't see anything in the query that's checking if the meeting did end within the past 30 days?

packages/server/graphql/types/Organization.ts
nickoferrall: To keep things moving, I've gone ahead and removed `Total Meetings` as it doesn't affect the price, which could be confusing for users, and it's expensive to calculate. 

@acressall if you feel strongly that we need to display `Total Meetings`, let me know and we can always add it back

igorlesnenko:
@nickoferrall can you rebase please?

packages/server/graphql/public/types/helpers/getActiveTeamCountByOrgIds.ts
nickoferrall: Thanks for explaining that @igorlesnenko. 

> Maybe keep everything as is for now and decide if we need to change after the experiment? 

I agree, I don't think it's something we need to spend time on now as we may change the teams limit after the experiment, e.g. to 3 teams. 

Would you mind creating an issue in the teams limit milestone so that we can revisit this topic once the experiment is complete and if we decide to roll out the current limit to all users?

igorlesnenko: Make sense, created #7831

igorlesnenko (approved):

packages/client/modules/userDashboard/components/OrgBilling/OrgPlans.tsx
```diff
@@ -156,23 +159,58 @@ const UpgradeButton = styled(FlatPrimaryButton)<{
...
+  }
+}
+
+const getButtonLabel = (tier: TierEnum, plan: TierEnum) => {
+  if (tier === 'starter') {
+    return plan === 'starter' ? 'Current Plan' : plan === 'team' ? 'Select Plan' : 'Contact'
+  } else if (tier === 'team') {
+    return plan === 'team' ? 'Current Plan' : plan === 'starter' ? 'Downgrade' : 'Contact'
+  } else {
+    return plan === 'enterprise' ? 'Current Plan' : 'Contact'
```
igorlesnenko, line 178: +1 Feels like could be simplified:

Something like this:

```
if (tier === plan) { 
  return 'Current Plan' 
} else if  (plan === 'enterprise') {
 return 'Contact'
} else if (plan === 'starter') {
return 'Downgrade'
} else { ... }
```

packages/client/modules/userDashboard/components/OrgBilling/OrgPlans.tsx
```diff
@@ -233,9 +281,15 @@ const OrgPlans = () => {
               </UL>
             </Content>
             <ButtonBlock>
-              <UpgradeButton buttonStyle={plan.buttonStyle} size='medium'>
-                {plan.buttonLabel}
-              </UpgradeButton>
+              {plan.buttonStyle && (
```
igorlesnenko, line 284: +1 I see it is defined for every plan, does it make sense to check it?

Dschoordsch (approved):

packages/client/modules/userDashboard/components/OrgBilling/OrgPlans.tsx
nickoferrall: True, thanks! Originally, that wasn't the case as we weren't going to show Plans for Enterprise users. Updating it now

packages/client/modules/userDashboard/components/OrgBilling/OrgPlans.tsx
nickoferrall: We have to check both `plan` and `tier`, so I don't think that'll work unfortunately

igorlesnenko: @nickoferrall `Current Plan` is always when `plan === tier`. So could be simplified like this:

```
  if (tier === plan) {
    return 'Current Plan'
  }

  if (tier === 'starter') {
    return plan === 'team' ? 'Select Plan' : 'Contact'
  } else if (tier === 'team') {
    return plan === 'starter' ? 'Downgrade' : 'Contact'
  } else {
    return 'Contact'
  }
```

But if not `current plan`, then enterprise is always `contact` as you can't downgrade to enterprise. If `starter` and not `current plan` then it is always donwgrade. 

So would be like this:

```
const getButtonLabel = (tier: TierEnum, plan: TierEnum) => {
  if (tier === plan) {
    return 'Current Plan'
  } else if (plan === 'enterprise') {
    return 'Contact'
  } else if (plan === 'starter') {
    return 'Downgrade'
  } else {
    return 'Select Plan'
  }
}
```

Does it make sense?

PR #7800, opened 2023-02-21
chore: Migrate MeetingTemplate table (Phase 2 of 3)

mattkrick:
# Description

- [x] Copies the RethinkDB MeetingTemplate table to PG
- [x] Does it all in a single all-or-nothing transaction
- [x] Adds an equality checker to graphql so you can make sure the data in both are.... 

## Demo

https://www.loom.com/share/6b4676756d7445a19a73ac92d6b33045

## Testing scenarios

- see demo
- run migration
- run checkEquality(table: "MeetingTemplate") mutation. see both DBs have the same # of rows


packages/server/postgres/migrations/1676461074304_NotificationOrgIdIndex.ts
```diff
@@ -3,7 +3,10 @@ import connectRethinkDB from '../../database/connectRethinkDB'
 
 export const up = async function () {
   await connectRethinkDB()
-  await r.table('Notification').indexCreate('orgId').run()
+  try {
```
mattkrick, line 6: wrapped in try/catch to keep these idempotent

packages/server/postgres/migrations/1677272969994_meetingTemplatesMove.ts
```diff
@@ -0,0 +1,107 @@
+import {FirstParam} from 'parabol-client/types/generics'
```
mattkrick, line 1: now that it's written I hope we can more or less copy/paste this for every table that we migrate

packages/server/graphql/private/mutations/checkRethinkPgEquality.ts
```diff
@@ -1,55 +1,60 @@
...
 
-  // RESOLUTION
-  if (!writeToFile) {
-    const errors = await tableResolver(maxErrors)
-    return JSON.stringify(errors)
+  if (tableName === 'MeetingTemplate') {
+    const rowCountResult = await checkRowCount(tableName)
+    const rethinkQuery = r.table('MeetingTemplate').orderBy('updatedAt', {index: 'updatedAt'})
+    const errors = await checkTableEq(rethinkQuery, getMeetingTemplatesByIds, {
+      createdAt: defaultEqFn,
```
mattkrick, line 42: i wanted a cleaner way to see what equality means for each column. the old code had special case handing for undefined, null, array ordering, etc. 

we may come across tables where arrays must be ordered & others where array order doesn't matter. we may have objects where extra props don't matter & some cases where they do, etc.

packages/server/postgres/migrations/1677272969994_meetingTemplatesMove.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  await pg.tx('meetingTemplateTable', (task) => {
+    const fetchAndProcess: FirstParam<typeof task.sequence> = async (
+      _index,
+      leftBoundCursor: undefined | Date
+    ) => {
+      const nextData = await getNextData(leftBoundCursor)
+      if (!nextData) return undefined
+      const insert = pgp.helpers.insert(nextData, columnSet)
+      const update = ` ON CONFLICT(id) DO UPDATE SET
```
mattkrick, line 82: this could happen in the following scenario:
- an old template gets migrated to PG in batch 1
- a user modifies that template while batch 2 is processing
- the modified template is picked up in the last batch & updates are made

packages/server/postgres/migrations/1677272969994_meetingTemplatesMove.ts
```diff
@@ -0,0 +1,107 @@
...
+    )
+    if (lastMatchingUpdatedAt === -1) {
+      throw new Error(
+        'batchSize is smaller than the number of items that share the same cursor. Increase batchSize'
+      )
+    }
+    return nextBatch.slice(0, lastMatchingUpdatedAt + 1)
+  }
+
+  await pg.tx('meetingTemplateTable', (task) => {
```
mattkrick, line 74: this is all happening in a single transaction to make sure we don't have any 1-off problems

jmtaber129:
@mattkrick Code LGTM (apart from failing tests), but could you elaborate a bit more on the potential risks here, as well as how those risks might be mitigated + what options we have if we need to revert? I'm especially interested in:
* What do we do if `checkEquality` returns a few errors?
* What do we do if `checkEquality` returns a _lot_ of errors (e.g. 10+% of the table)?
* How long do we expect the full migration to take on the prod dataset considering the size of the `MeetingTemplate` table? Is it possible a migration like this could interfere with the normal release process (e.g. if it takes a long time to execute or fails entirely)?
* Would it be feasible to decouple the actual data migration from the next release? E.g. executing the migration after the relevant code is deployed?

mattkrick:
> What do we do if checkEquality returns a few errors?

the troubleshooting begins!
- if we're missing records, then there's probably some app logic that is only writing to RethinkDB & not to PG. 
- if there is a delta between the data, then the write to PG is somehow different than the write to RethinkDB so we'll need to fix up a query
- most common is there's an error in how we measure equality, which is thankfully the easiest & we'll just have to fix the equality checker itself. this slowed us down a bunch last time so that's why I did some refactoring so each field now has it's own way to measure equality

> What do we do if checkEquality returns a lot of errors (e.g. 10+% of the table)?

- we'll write the result to disk so we can get a better picture. typically it's just the same type of error over & over again (e.g. `pg.updatedAt !== r.updatedAt` because they're off by a couple ms

> How long do we expect the full migration to take on the prod dataset considering the size of the MeetingTemplate table? Is it possible a migration like this could interfere with the normal release process (e.g. if it takes a long time to execute or fails entirely)?

- PG is fantastic at handling concurrency. RethinkDB... not so much. Thankfully we're just reading from RethinkDB, and I kept batch sizes pretty small to prevent any bottlenecks or memory spikes.
- PG transactions are a row-level lock, so reads & even inserts aren't a problem, but updates to existing rows in PG will get queued behind the transaction. 
- I don't have a great estimate on how long it'll take. I'd be surprised if it was more than 2 minutes!

> Would it be feasible to decouple the actual data migration from the next release? E.g. executing the migration after the relevant code is deployed?

- it's possible, but any special handling will also have to apply to all of our privately managed instances
- migrations are currently run as a `predeploy` instead of `postdeploy` because often times migrations must be in place before the new code is launched. moving from RethinkDB to PG is the exception! In the past we attempted to have both predeploy & postdeploy, but that turned out to be a bad idea due to linearizability-- since all predeploy migrations happen first & then all postdeploy migrations happen second, the outcome is dependent on when we release-- which is fine for the cloud app, but could wreck havoc on PPMIs that only get upgraded every so often.

jmtaber129 (approved):
LGTM! (pending tests are fixed) Excited to get this migration closer to done

PR #7804, opened 2023-02-22
fix(emails): Upgrade mailgun to 7.0.4

jmtaber129:
# Description
With some recent package upgrades, mailgun requests will fail with `from parameter is missing`, even though `from` is definitely present.

This issue seems to be fixed with `v7.0.4`, so upgrade to that version.

I took a look through the [changelog](https://github.com/mailgun/mailgun.js/blob/master/CHANGELOG.md) to see if we may be impacted by the breaking changes + smoke tested emails, and there don't seem to be any issues, but I encourage reviewer to also take a look.

## Testing scenarios
Smoke test emails

## Final checklist

### DIFF
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -114,7 +114,7 @@
     "jsdom": "^20.0.0",
     "jsonwebtoken": "^9.0.0",
     "mailcomposer": "^4.0.1",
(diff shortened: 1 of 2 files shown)


Dschoordsch (approved):
Works locally

Issue #7736, opened 2023-02-08
Standups: Some email summaries are blank


jmtaber129:
See [slack](https://parabol.slack.com/archives/C4JAUUZ9P/p1675853193628709) 🔒 

Some email summaries for standups are blank:

![Image](https://user-images.githubusercontent.com/9013217/217577664-2c3176b8-d07b-4721-a0cd-ef505517c487.png)

PR #7805, opened 2023-02-22
fix(recurrence/emails): Generate email summaries using facilitator auth

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/7736

The `processRecurrence` mutation generally runs with an admin credential provided by `action-chronos`, which doesn't actually have direct access to meetings for which we're sending summaries as part of `processRecurrence`.

Instead of using whichever auth token was used to execute the mutation to generate the email summary, use a generated auth token that impersonates the meeting facilitator. This is similar to what we're already doing for [notification emails](https://github.com/ParabolInc/parabol/blob/47dfb31cbdac7bb07fbcb90dd3ca5124f799ce76/packages/server/graphql/private/mutations/sendBatchNotificationEmails.ts#L56-L60).

Notably, this was personally difficult to reproduce locally because the GQL admin console uses the auth token of the logged-in user (myself), which was generally a user that had access to the meetings that would be started+stopped.

## Demo
N/A

## Testing scenarios
- [ ] Start a meeting series as user A
- [ ] As a superuser B who does _not_ have access to the meeting series started by user A, run `processRecurrence` in the GQL admin console:
```
mutation {
  processRecurrence {
    ... on ProcessRecurrenceSuccess {
      meetingsStarted
      meetingsEnded
    }
  }
}
```
- [ ] Confirm that user A receives a non-blank email summary.
- [ ] Smoke test meeting summaries in-app and in emails.

## Final checklist


BartoszJarocki (approved):
good find!

PR #7807, opened 2023-02-22
chore(retros-in-disguise): Added retros in disguise feature flag and empty route

BartoszJarocki:
# Description

This PR adds `retrosInDisguise` feature flag and an empty frontend component for activity library. Empty component is accessible under `/activity-library` route, if the `retrosInDisguise` feature flag was set. If not, it'll redirect to 404.

Component names, route urls and everything else are TDB and intention of this PR is only add stub components so we can start working on the details. 

## Demo

No demo

## Testing scenarios

- [ ] Add a `retrosInDisguise` feature flag, check if `/activity-library` is accessible

- [ ] Check if a new route will redirect to 404 if user doesn't have feature flag set 

## Final checklist

### DIFF
--- a/packages/client/components/ActivityLibrary/ActivityLibrary.tsx
+++ b/packages/client/components/ActivityLibrary/ActivityLibrary.tsx
@@ -1,0 +1,32 @@
+import React from 'react'
+import graphql from 'babel-plugin-relay/macro'
+import {PreloadedQuery, usePreloadedQuery} from 'react-relay'
+import {ActivityLibraryQuery} from '~/__generated__/ActivityLibraryQuery.graphql'
+import {Redirect} from 'react-router'
+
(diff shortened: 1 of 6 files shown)


mattkrick:
yeah, this is a totally new view for us. It looks like a full page modal with the "X" in the top right.
Personally, I don't like it when modals have their own URLs, so I'm all for treating this like a full page instead of a modal.
If we want to get fancy we could use framer-motion & add a page transition to make it fade in like other modals, but definitely not necessary for now

BartoszJarocki:
I mean, I can fix anything here, but the idea was to just have a stub to work on, not a final design or a component/route. @jmtaber129 what changes are you actually requesting?

jmtaber129 (approved):

PR #7808, opened 2023-02-22
chore: add css extract

mattkrick:
# Description

Base PR: #7597

- Extracts css files 
- the first one is loaded directly from the initial HTML
- Only 1 css file per unique content (things like emoji-mart css will not have to be fetched on new versions)

## Testing scenarios

 - run `yarn build` look in the `/build` directory. 
 - see that every `.css` has unique content & there aren't multiple css files with the same content
 - see that there are no css sourcemaps
 - see `/build/index.html` includes a tag to load the initial css file
 - load the page & there is no FOUC


BartoszJarocki (approved):
works like a charm! I was missing excluding sourcemaps 🤦

PR #7809, opened 2023-02-23
chore: Added a tailwind preset file

BartoszJarocki:
# Description

This PR adds a Tailwind [preset](https://tailwindcss.com/docs/presets) that can be used as a single source of truth for old `Palette` file and new tailwind config. 

## Demo

No demo

## Testing scenarios

- [ ] See if app looks like before the changes

## Final checklist


packages/client/tailwind.preset.js
```diff
@@ -0,0 +1,157 @@
+/**
```
BartoszJarocki, line 1: This file can be published to npm and used in other places where we use tailwind as a [preset](https://tailwindcss.com/docs/presets), for example parabol.co, icebreakers.parabol.co

packages/client/tailwind.preset.d.ts
```diff
@@ -0,0 +1,30 @@
+type ThemeColors =
```
BartoszJarocki, line 1: I'm not sure if that's the best way, but since we want to import the file into tailwind.config.js we need to have colors also in the `.js` file. This, file adds automatic types when we import it into our typescript codebase. I didn't go with `json` as I wanted to use Tailwind presets functionality, in case we'd like a single source of truth for all our products.

packages/client/styles/paletteV3.ts
```diff
@@ -4,79 +4,84 @@
...
 
-export const enum PALETTE {
-  AQUA_400 = '#55C0CF',
+import tailwindPreset from '../tailwind.preset.js'
 
-  FOREST_400 = '#73A153',
+const {colors} = tailwindPreset.theme
 
-  FUSCIA_400 = '#DB70DB',
+export const PALETTE = {
```
BartoszJarocki, line 15: It can't be an enum anymore as TS complains we can't use computed values in enums

packages/client/utils/taskStatus.ts
```diff
@@ -10,7 +10,7 @@ export const taskStatusLabels: Record<TaskStatus, TaskStatusLabel> = {
   [TaskStatus.PRIVATE]: TaskStatusLabel.PRIVATE
 } as const
 
-export const taskStatusColors: Record<TaskStatus, PALETTE> = {
+export const taskStatusColors: Record<TaskStatus, string> = {
```
BartoszJarocki, line 13: as PALETTE is not an enum anymore, we'll loose a type safety here 😞

mattkrick: good riddance to const enums! long live string unions 🎉

mattkrick (commented):
I normally wouldn't mention this, but since this PR requires manually writing a declaration file  there is an alternative where you could get rid of your declaration file AND keep your type safety. 

if you're curious, i wrote a quick little comparison. feel free to use it or not. your call!
https://github.com/ParabolInc/parabol/pull/7811/files

BartoszJarocki:
@mattkrick I merged your PR so this one should be ready, would you mind taking final look? thank you!

PR #7811, opened 2023-02-24
chore: keep types on colors

mattkrick:
# Description

An alternative to manual types  that preserves type safety.

sucrase (or babel, swc, etc) can be called inline & monkeypatch the `require` statement to process typescript, kinda like a mini webpack.

Up to you if you wanna use this strategy!


packages/client/styles/paletteV3.ts
```diff
@@ -84,4 +84,4 @@ export const PALETTE = {
   WHITE: colors.white,
 
   SUCCESS_LIGHT: '#2db553'
-}
+} as const
```
mattkrick, line 87: since `colors` has a const assertion on it, PALETTE can have one, too

packages/client/tailwindTheme.ts
```diff
@@ -0,0 +1,156 @@
...
+        '400': '#C3C0D8',
+        '500': '#A7A3C2',
+        '600': '#82809A',
+        '700': '#444258',
+        '800': '#2D2D39',
+        '900': '#1C1C21'
+      }
+    }
+  }
+} as const
```
mattkrick, line 156: since this is a .ts, we can use a const assertion now

packages/client/utils/taskStatus.ts
```diff
@@ -10,7 +11,7 @@ export const taskStatusLabels: Record<TaskStatus, TaskStatusLabel> = {
   [TaskStatus.PRIVATE]: TaskStatusLabel.PRIVATE
 } as const
 
-export const taskStatusColors: Record<TaskStatus, string> = {
+export const taskStatusColors: Record<TaskStatus, ValueOf<typeof PALETTE>> = {
```
mattkrick, line 14: this is a pretty tame example, but it's showing that we have a string union here, not just `string`.

mattkrick:
merge it to yours if you like! otherwise you can merge yours to maser

Issue #7530, opened 2022-12-06
Checkout Flow: credit card UI


nickoferrall:
Figma designs [here](https://www.figma.com/file/SyJdkzumWsUJo58cmBGgus/Checkout-concepts?t=p3f7trldNdOyScXZ-0)

<img width="623" alt="credit-card-ui" src="https://user-images.githubusercontent.com/39854876/215596345-59e7d328-8497-4f6a-9c7b-4729cf05607b.png">

### AC

- [ ] Implement the UI for the credit card section of the checkout flow
- [ ] The billing section text is hardcoded. We'll make it dynamic here: https://github.com/ParabolInc/parabol/issues/7693

nickoferrall:
Hey @acressall, are you thinking the heading "Your first payment..." would only show for new users, and then everyone else would just see "Your next payment is due on..."? 

I lean towards just saying, "Your next payment is due on..." for all users to keep things simple and consistent. WDYT?

Also, do you have any designs for an error state? If the user enters incorrect card details, where would we should "Invalid CVC" text, for example?

acressall:
I'm ok with the consistent language you proposed! [Here's the error state design](https://www.figma.com/file/SyJdkzumWsUJo58cmBGgus/Checkout-concepts?node-id=474%3A4319&t=f1ZxoeQjHE2AkLLA-4). Do we have content from the existing form we can use, or do we need some?

nickoferrall:
> [Here's the error state design](https://www.figma.com/file/SyJdkzumWsUJo58cmBGgus/Checkout-concepts?node-id=474%3A4319&t=f1ZxoeQjHE2AkLLA-4).

Thanks!

> Do we have content from the existing form we can use, or do we need some?

Yeah, atm, in the credit card model we have errors that say "invalid CVV" or "invalid credit card". 

![Screenshot 2023-02-23 at 12 44 17](https://user-images.githubusercontent.com/39854876/220909452-0fd38eaa-29d3-4f7b-8768-1ee7814ff8b4.png)

I guess I could make the error state look like your design and replace the "Woops! Something looks off" with "Invalid CVV"? Or maybe "Woops! That's an invalid CVV"

acressall:
> I guess I could make the error state look like your design and replace the "Woops! Something looks off" with "Invalid CVV"?

Yes, that's good! I just threw in some dummy copy, so don't worry about pulling in any of that

nickoferrall:
[...]

Issue #7693, opened 2023-01-30
Checkout Flow: Make credit card info dynamic


nickoferrall:
Predecessor: https://github.com/ParabolInc/parabol/issues/7530

### AC

- [ ] Next payment text and credit card info is dynamic, i.e. it's calculated using the org data
- [ ] After submitting credit card info and clicking Upgrade, the confetti animation happens
- [ ] The org is upgraded to the Team plan
- [ ] Handle error states as discussed [here](https://github.com/ParabolInc/parabol/issues/7530#issuecomment-1441707327)

nickoferrall:
Pausing work on this issue until I know whether we'd like to use Stripe Elements - see comment [here](https://github.com/ParabolInc/parabol/pull/7812#discussion_r1119261639)

nickoferrall:
Hey @acressall, when using Stripe Elements, the billing address is required. This adds one/two extra fields depending on the country:

![Screenshot 2023-04-05 at 15 16 29](https://user-images.githubusercontent.com/39854876/230108939-8eb57e26-57a1-49ec-a32e-766b8a2128d2.png)

![Screenshot 2023-04-05 at 15 16 37](https://user-images.githubusercontent.com/39854876/230108949-33eb70e4-2532-4e33-b5d4-0f6738523ddf.png)

I wanted to check if that works for you? Perhaps I should add extra spacing to the Team Plan Pricing so that it fills more of the height of the panel?

Also, Stripe expects to send the user to a payment completion page once they've entered their card details. Perhaps this would be a good idea as it would make it clearer that they've upgraded vs just seeing the confetti animation and the right drawer opening?

Here's a Loom with how it's currently looking: https://www.loom.com/share/8ae42db6610147f6946d06ecfc31e7b1

acressall:
> I wanted to check if that works for you? Perhaps I should add extra spacing to the Team Plan Pricing so that it fills more of the height of the panel?

I have no issue with the extra field! As for the pricing side, I don't mind terribly if the upgrade btn is hanging, but I trust your judgement here.

> Also, Stripe expects to send the user to a payment completion page once they've entered their card details. Perhaps this would be a good idea as it would make it clearer that they've upgraded vs just seeing the confetti animation and the right drawer opening?

Does Stripe expect us to send them to a _Stripe confirmation_ page? Or just any page? I left some comments in the Loom, but want to make sure I understand. Do they want a new URL? Or is a change of state enough?

nickoferrall:
[...]

Issue #4460, opened 2020-12-17
Spruce up credit card modal


nickoferrall:
To improve readability and reduce the chance that a user incorrectly enters their card number, we should add a space after every 4th number, so `4242424242424242` becomes `4242 4242 4242 4242`.

We could also add padding above the Upgrade button and consider using [Stripe Elements](https://stripe.com/docs/stripe-js) so that there's only one input for the card number, expiry date & CVC rather than 3.  

![upgrade-to-pro](https://user-images.githubusercontent.com/39854876/102505537-06405800-407a-11eb-9d34-22ebf2c4750d.png)

mattkrick:
this feels like a good item to tackle when we do a pricing update:
- add in option to pay annually
- add in option to pay via ACH
- add some spaces between every 4 CC numbers

jordanh:
Scrubbed. Iceboxing

nickoferrall:
Our credit card conversion rate seems really low: https://analytics.amplitude.com/parabol/chart/7mexl8a

Sprucing it up might help a bit. I suggest that we use Stripe Elements in our modal. As you can see in their example [here](https://stripe.com/docs/payments/payment-element), we could build something like this:
![stripe-elements](https://user-images.githubusercontent.com/39854876/193289012-87bdf127-7e57-438a-b1cb-2418afa8b733.png)

To keep it in line with our current functionality, from the example above, we could:

- Remove the card options
- Remove the country & postal code (although these help increase card acceptance rate, I don't believe that's an issue for us atm)
- Include Parabol's price count & button from our current modal

The benefits of doing this are:

- Stripe elements look better, in my opinion, as they show what the CVC is and what cards we accept
- Having a space separating every 4th character in the card number makes it easier to read & use
- You can find Stripe elements across many popular apps. Having a familiar checkout experience might give users a sense of security

igorlesnenko:
I found that is is not very easy to use Stripe Elements in our case.

There are [2 possible ways](https://github.com/stripe-samples/accept-a-payment#accept-a-payment) to use stripe elements:

**Payment Element**

A single element that embeds a fancy payment form using iframe. The one that is shown on the [screenshot](https://github.com/ParabolInc/parabol/issues/4460#issuecomment-1263642257) above

It handles a lot of things automatically.

However, it is not just a set of inputs that we can use instead of our inputs. It requires complete refactoring and some backend work too, as it uses `PaymentIntents` that we currently don't use.

**Custom payment flow**

[...]

PR #7812, opened 2023-02-24
Feat(checkout-flow): credit card UI

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7530

In this PR, I'm building the credit card UI. It's not responsive as those designs are being worked on and the values are hardcoded as they'll be added in https://github.com/ParabolInc/parabol/issues/7693

### To test

- [ ] Make sure your user has the checkout-flow feature flag turned on
- [ ] Go to your organization page
- [ ] See the credit card UI looks like the designs

### DIFF
--- a/packages/client/modules/userDashboard/components/OrgBilling/Billing.tsx
+++ b/packages/client/modules/userDashboard/components/OrgBilling/Billing.tsx
@@ -1,12 +1,13 @@
 import styled from '@emotion/styled'
+import {Error as ErrorIcon} from '@mui/icons-material'
 import React, {useEffect, useState} from 'react'
 import Panel from '../../../../components/Panel/Panel'
+import PrimaryButton from '../../../../components/PrimaryButton'
 import Row from '../../../../components/Row/Row'
 import useForm from '../../../../hooks/useForm'
 import useScript from '../../../../hooks/useScript'
 import {PALETTE} from '../../../../styles/paletteV3'
 import StripeClientManager from '../../../../utils/StripeClientManager'
-import CreditCardErrorLine from '../CreditCardModal/CreditCardErrorLine'
 
 const StyledPanel = styled(Panel)({
   maxWidth: 976,
@@ -39,7 +40,6 @@ const Plan = styled('div')({
   padding: '16px',
   height: 400,
   borderRadius: 4,
-  border: `2px solid red`,
   width: '50%',
   overflow: 'hidden'
 })
@@ -53,6 +53,27 @@ const Heading = styled('span')<{isBold?: boolean}>(({isBold}) => ({
 const HeadingBlock = styled('div')({
   padding: '16px'
 })
+
+const LineIcon = styled('div')({
+  svg: {
+    fontSize: 19
+  },
+  display: 'flex',
+  justifyContent: 'center',
+  alignItems: 'center'
+})
+
+const UpgradeButton = styled(PrimaryButton)<{isDisabled: boolean}>(({isDisabled}) => ({
+  flexGrow: 1,
+  background: isDisabled ? PALETTE.SLATE_200 : PALETTE.SKY_500,
+  color: isDisabled ? PALETTE.SLATE_600 : PALETTE.WHITE,
+  boxShadow: 'none',
+  elevation: 0,
+  '&:hover': {
+    boxShadow: 'none',
+    background: isDisabled ? PALETTE.SLATE_200 : PALETTE.SKY_600
+  }
+}))
 
 const Title = styled('div')({
   color: PALETTE.SLATE_800,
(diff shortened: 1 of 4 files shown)


packages/client/modules/userDashboard/components/OrgBilling/Billing.tsx
```diff
@@ -0,0 +1,335 @@
...
+                placeholder='Full Name'
+                onBlur={() => setDirtyField('cardName')}
+                onChange={onChange}
+                id='cardholder-name'
+                name='cardName'
+                type='text'
+                autoComplete='cc-name'
+              />
+              <InputLabel>{'Card Number'}</InputLabel>
+              <CleaveInput
```
BartoszJarocki, line 252: should we use Stripe Elements to not risk being not PCI compliant?
https://github.com/ParabolInc/parabol/issues/4460#issuecomment-1281241237

BartoszJarocki: @nickoferrall small but nice detail, you can set Stripe Elements to use our font by passing `options` prop to `Elements` component

```
    <Elements options={{
      fonts: [
        {
          family: 'IBM Plex Sans',
          src: `url('/static/fonts/IBMPlexSans-Regular.woff2') format('woff2')`,
          weight: '400',
        },
      ],
     }}>
      ...
    </Elements>
```

igorlesnenko:
@nickoferrall do you need another review from me?

nickoferrall:
Yes, please @igorlesnenko. I think Bartosz was just commenting, so I wouldn't count it as a Reviewer Review

igorlesnenko (commented):
I like the idea of using stripe elements. I think I need to look again if we refactor on using them

packages/client/modules/userDashboard/components/OrgBilling/Billing.tsx
acressall: @nickoferrall this is nice! If it doesn't add too much bloat to override the styles to match the design, I'd like to have the customized bg and border. Thinking about form consistency across the app, we're sticking pretty close to the material design standard.

Other _small_ nitpick is the upgrade button. It should be the full width of the form, and disabled (grey) until the form is filled out

nickoferrall:
I've merged the Stripe Elements PR into this one. Alicia has approved the UI for the Stripe Elements so I'm just looking for a final technical review from @Dschoordsch on this one 🙏

igorlesnenko (approved):

packages/client/modules/userDashboard/components/OrgBilling/Billing.tsx
```diff
@@ -0,0 +1,197 @@
...
+import {loadStripe} from '@stripe/stripe-js'
+import {Elements} from '@stripe/react-stripe-js'
+import CreatePaymentIntentMutation from '../../../../mutations/CreatePaymentIntentMutation'
+import useAtmosphere from '../../../../hooks/useAtmosphere'
+import useMutationProps from '../../../../hooks/useMutationProps'
+import {CreatePaymentIntentMutationResponse} from '../../../../__generated__/CreatePaymentIntentMutation.graphql'
+import {CompletedHandler} from '../../../../types/relayMutations'
+
+const StyledPanel = styled(Panel)({
+  maxWidth: 976
```
Dschoordsch, line 17: -1 That number looks familiar, don't we have something in `constEnums` for it?

packages/client/modules/userDashboard/components/OrgBilling/Billing.tsx
```diff
@@ -0,0 +1,197 @@
...
+      const {clientSecret} = createPaymentIntent
+      if (clientSecret) {
+        setClientSecret(clientSecret)
+      }
+    }
+
+    CreatePaymentIntentMutation(atmosphere, {}, {onError, onCompleted: handleCompleted})
+  }, [])
+
+  const appearance = {
```
Dschoordsch, line 114: +1 I think this is the worst way of defining these settings: it's in the component, so it gets cluttered, but it's also untyped and you have to follow the references to see where it's used. I would be happier with pasting all this right in the component.

packages/server/graphql/public/mutations/createPaymentIntent.ts
```diff
@@ -0,0 +1,24 @@
+import {getUserId} from '../../../utils/authorization'
+import {getStripeManager} from '../../../utils/stripe'
+import {MutationResolvers} from '../resolverTypes'
+
+const createPaymentIntent: MutationResolvers['createPaymentIntent'] = async (
+  _source,
+  {},
+  {authToken, dataLoader, socketId: mutatorId}
```
Dschoordsch, line 8: +1 `mutatorId` is unused and will probably remain unused as well, can you please clean that up now?

Dschoordsch (approved):
Would be nice if you'd do some cleanup before merging.

Issue #7697, opened 2023-01-30
Checkout Flow: add downgrade logic


nickoferrall:
<img width="582" alt="Screenshot 2023-01-30 at 15 56 35" src="https://user-images.githubusercontent.com/39854876/215604325-8a03a574-03e3-4a9d-9ca0-1758419f0459.png">

### AC

- [ ] Clicking the "Downgrade" button (as outlined [here](https://github.com/ParabolInc/parabol/issues/7529)) opens a dialog with the Downgrade text specified in the [designs](https://www.figma.com/file/SyJdkzumWsUJo58cmBGgus/Checkout-concepts?node-id=209%3A3221&t=89Ho501858UU34wG-0)
- [ ] Pressing "Keep my Plan" closes the dialog
- [ ] Pressing "Yes, downgrade" downgrades the org from Team to Starter and shows them the second dialog screen asking why they're leaving
- [ ] The user can select several of the available options and click submit, which closes the dialog and causes the Plans UI to update. Their answer is tracked in Segment

tghanken:
I'd like to follow a similar pattern to our Upgrade flow for Segment downgrade events:
- Start with a `Downgrade Clicked` event when the billing leader starts the downgrade.  Include a `downgradeLocation` property
- For menus in the middle of the flow, we can use a `Downgrade Continue Clicked` event with the name of the menu as a property.
- Finally, when the downgrade is finalized, we should use the existing `Organization Downgraded` event.

In practice, I see this as one `Downgrade Clicked` event emitted on the plans UI when the downgrade option is selected, then a `Downgrade Continue Clicked` event emitted when the "Yes, Downgrade" button is clicked on the attached design.  Finally, `Organization Downgraded` would be submitted when "Submit" is clicked, and we'd add the list of feedback as a new property.

PR #7830, opened 2023-02-28
feat(checkout-flow): add downgrade logic

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7697

Loom demo: https://www.loom.com/share/734efabc2ed0439fa0d67022e6f997b2

downgrade

The feedback we give users after they downgrade could be improved. Issue created here (https://github.com/ParabolInc/parabol/issues/7843) to fix this.

### To test

- [ ] Add the checkout-flow feature flag to your user and go to the org page for an org on the Team tier
- [ ] Go to the new billing page and click downgrade
- [ ] Clicking downgrade opens a modal
- [ ] Clicking Keep Plan closes the modal
- [ ] Selecting Submit causes the org to downgrade 
- [ ] The events mentioned in https://github.com/ParabolInc/parabol/issues/7697 are collected as expected


nickoferrall:
Hey @tghanken, it'd be great to get your eyes on this PR when you get the chance as there are a few different events I'm tracking. 

In the `Downgrade Continue Clicked` event, you said that we should include the name of the menu, but as we don't have any other downgrade modals atm, I was wondering if that's needed? I haven't added that yet.

I have included a `downgradeLocation`, which is just `billing` for now. I have added that but I wonder if it could be removed?

packages/server/graphql/mutations/downgradeToStarter.ts
```diff
@@ -42,6 +48,29 @@ export default {
...
+      if (!billingLeaderUser) {
+        const error = new Error('Unable to find billing leader')
+        sendToSentry(error, {userId: viewerId})
+        return
+      }
+      analytics.organizationDowngraded(viewerId, {
+        orgId,
+        oldTier: tier,
+        newTier: 'starter',
+        billingLeaderEmail: billingLeaderUser.email,
```
nickoferrall, line 69: @tghanken `organizationDowngraded` accepts only one billing leader email string, but there could be many billing leaders in an org. I've grabbed the first billing leader. Does that work for you or would you like all the billing leader emails?

packages/server/graphql/mutations/downgradeToStarter.ts
```diff
@@ -42,6 +48,29 @@ export default {
...
+        orgName: organization.name,
+        reasonsForLeaving
```
nickoferrall, line 71: I added `reasonsForLeaving` as this seems the best place to track this field. I don't love passing this as an arg to the mutation though. 

Alternate options could be sending the segment client mutation from the client rather than recording it here. I could also create an enum of downgrade reasons if reviewers feel the string array is too brittle

nickoferrall:
Hey @acressall, it'd be great if you could review this PR or watch the Loom in the description too 🙏  A few things I've changed from the design:

- the "Downgrade" title is bold
- added colour to the checkboxes in the second screen of the modal
- added more padding to the checkbox list

If you don't like any of that, just let me know!

Also, I think we should give the user more visual feedback after they downgrade. At the moment, they click the "Submit" button, the modal closes, and the UI says "Starter" instead of "Team", but it's not super obvious. 

Perhaps there should be a green tick animation after clicking "Submit" so it's clear it's done? I think that should be tackled in a separate issue though

acressall:
Thanks for the demo @nickoferrall ! Looks good. One small bit of feedback:
The text in the modal for the bullet points and checkboxes looks darker than the rest. Aside from the heading, the paragraphs should all be `Slate 900`

> Perhaps there should be a green tick animation after clicking "Submit" so it's clear it's done? I think that should be tackled in a separate issue though

I'm curious what you have in mind for the green tick? I like the idea in theory, as long as it stays visible for long enough, since right now the modal goes away pretty quick. An alternative would be a snackbar that goes away after x seconds. We could easily get a crying emoji in there ;) I agree we can tackle this in a separate issue

packages/client/modules/userDashboard/components/OrgBilling/DowngradeModal.tsx
```diff
@@ -0,0 +1,219 @@
...
+
+const LabelGroup = styled('div')({
+  paddingTop: '32px',
+  display: 'flex',
+  justifyContent: 'space-between',
+  width: '100%'
+})
+
+const Description = styled('div')({
+  color: PALETTE.SLATE_700,
```
acressall, line 42: this should be Slate 900, same as the labels below

nickoferrall: Sure thing!

Here's how the update looks:

Screenshot 2023-03-01 at 18 20 02

Screenshot 2023-03-01 at 18 19 41

acressall (approved):

packages/server/graphql/mutations/downgradeToStarter.ts
igorlesnenko: +1 I would prefer to use enum, otherwise someone can write arbitrary strings?
No strong preference between backend/client.

packages/client/modules/userDashboard/components/OrgBilling/DowngradeModal.tsx
```diff
@@ -0,0 +1,219 @@
...
+  closeModal: () => void
+  organizationRef: DowngradeModal_organization$key
+}
+
+const reasonsToLeave = [
+  'Parabol is too expensive',
+  'Budget changes',
+  'Missing key features',
+  `Not using Parabol's paid features`,
+  'Moving to another tool (please specify)'
```
igorlesnenko, line 121: Do we plan to add textfield with the tool name?

igorlesnenko (changes requested):
I see a little red flicker when clicking on the downgrade button, is it expected behaviour?

https://user-images.githubusercontent.com/466991/222475424-57b7fa10-b010-47ca-bea0-d8d5cdd647e3.mov


I also getting this error in console when trying downgrading. Downgrade works, but I need to refresh the page to see the result. Can you reproduce it?

```
23:38:50 1|GraphQL Executor  | TypeError: Cannot read properties of null (reading 'email')
23:38:50 1|GraphQL Executor  |     at resolveDowngradeToStarter (/Users/igor/www/parabol/dev/gqlExecutor.js:11196:30)
23:38:50 1|GraphQL Executor  |     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
23:38:50 1|GraphQL Executor  |     at async Object.resolve [as MutationdowngradeToStarterResolver] (/Users/igor/www/parabol/dev/gqlExecutor.js:5029:5)
23:38:50 1|GraphQL Executor  | Trace: {
23:38:50 1|GraphQL Executor  |   error: `{"data":{"downgradeToStarter":null},"errors":[{"message":"Cannot read properties of null (reading 'email')","locations":[{"line":5,"column":3}],"path":["downgradeToStarter"]}]}`
23:38:50 1|GraphQL Executor  | }
23:38:50 1|GraphQL Executor  |     at executeGraphQL (
```

packages/client/modules/userDashboard/components/OrgBilling/DowngradeModal.tsx
nickoferrall: Ah great point! Updated 👍

packages/server/graphql/mutations/downgradeToStarter.ts
nickoferrall: Sure thing, I've added an enum instead

nickoferrall:
Thanks for the review @igorlesnenko. I wasn't able to reproduce that red background colour in your video so please let me know if that's still happening for you. Here's a Loom demo of what I'm seeing after the changes: https://www.loom.com/share/25a951c27dc049539b49851a7935c7f4

packages/server/graphql/mutations/downgradeToStarter.ts
```diff
@@ -42,6 +60,26 @@ export default {
     }
 
     // RESOLUTION
+    if (!isSuperUser) {
```
igorlesnenko, line 63: -1 This code is always `false`, should be `!isSuperUser(authToken)`

packages/server/graphql/mutations/downgradeToStarter.ts
```diff
@@ -42,6 +60,26 @@ export default {
...
+        return
+      }
+      analytics.organizationDowngraded(viewerId, {
+        orgId,
+        oldTier: tier,
+        newTier: 'starter',
+        billingLeaderEmail: billingLeaders[0].email,
+        orgName: organization.name,
+        reasonsForLeaving,
+        otherTool
```
igorlesnenko, line 80: -1 Maybe set some limit on that field? If I put to big value, the code fails with an error

```
22:04:35 1|GraphQL Executor  | You have triggered an unhandledRejection, you may have forgotten to catch a Promise rejection:
22:04:35 1|GraphQL Executor  | AssertionError [ERR_ASSERTION]: Your message must be < 32kb.
```

packages/server/graphql/mutations/downgradeToStarter.ts
```diff
@@ -42,6 +60,26 @@ export default {
...
+      const [billingLeaders, organization] = await Promise.all([
+        getBillingLeadersByOrgId(orgId, dataLoader),
+        dataLoader.get('organizations').load(orgId)
+      ])
+      if (!billingLeaders[0]) {
+        const error = new Error('Unable to find billing leader')
+        sendToSentry(error, {userId: viewerId})
+        return
+      }
+      analytics.organizationDowngraded(viewerId, {
```
igorlesnenko, line 73: -1 We recording this event twice, second time it is recorded here: https://github.com/ParabolInc/parabol/blob/9ea8985892b661d4b5806deb0bcb9e06e77ae97e/packages/server/graphql/mutations/helpers/resolveDowngradeToStarter.ts#L45

igorlesnenko (changes requested):
The red background happens because for a fraction of time the button goes to active state, where background is set as red:

image

Maybe remove for more smooth experience

I also found the reason why mutation is failing for me.
Here we query User from rethinkdb table, but I guess we are not using it anymore. So the User is always `null` for me because I use fresh account:
https://github.com/ParabolInc/parabol/blob/9ea8985892b661d4b5806deb0bcb9e06e77ae97e/packages/server/graphql/mutations/helpers/resolveDowngradeToStarter.ts#L26

packages/client/utils/constants.ts
```diff
@@ -158,3 +160,27 @@ export const MAX_SPOTLIGHT_COLUMNS = 3
...
+  'Moving to another tool (please specify)'
+]
+
+export const reasonsToDowngradeLookup = {
+  'Parabol is too expensive': 'tooExpensive',
+  'Budget changes': 'budgetChanges',
+  'Missing key features': 'missingKeyFeatures',
+  "Not using Parabol's paid features": 'notUsingPaidFeatures',
+  'Moving to another tool (please specify)': 'anotherTool'
+} as Record<ReadableReasonToDowngradeEnum, ReasonToDowngradeEnum>
```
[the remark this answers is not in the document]
nickoferrall: We don't get the types with Object.keys unless we cast them, so I thought this would be safer

packages/server/graphql/mutations/downgradeToStarter.ts
nickoferrall: I like it, updated 👍

packages/server/dataloader/customLoaderMakers.ts
```diff
@@ -542,3 +542,25 @@ export const activeMeetingsByMeetingSeriesId = (parent: RootDataLoader) => {
     }
   )
 }
+
+export const billingLeadersIdsByOrgId = (parent: RootDataLoader) => {
```
nickoferrall, line 546: Previously, we were adding the analytics in `downgradeToStarter` and needed to grab the `billingLeadersIds`, so I moved this to a dataloader. We're no longer doing this as the `resolveDowngradeToStarter` mutation already tracks this. 

Still, the dataLoader is being used in teamsLimits so I think it's worth keepingf

nickoferrall:
@igorlesnenko I still haven't been able to reproduce the red background flash on the button, but I noticed that the `FlatPrimaryButton` has a rose backgrounde image on focus, so I've swapped it for the `BaseButton`. Hopefully that fixes it 🤞 

> Here we query User from rethinkdb table, but I guess we are not using it anymore. So the User is always null for me because I use fresh account:

Oh great catch! Looks like that's a bug that's been around for a while. Are you able to use the [updateEmail](https://github.com/ParabolInc/parabol/blob/fa767a6f719cec03a966d41cc45d1a077a1857fb/packages/server/graphql/private/mutations/updateEmail.ts#L26) mutation? If not, I can fix that in a separate PR

igorlesnenko:
> Are you able to use the [updateEmail](https://github.com/ParabolInc/parabol/blob/fa767a6f719cec03a966d41cc45d1a077a1857fb/packages/server/graphql/private/mutations/updateEmail.ts#L26) mutation? If not, I can fix that in a separate PR

This should work because in the worst case it simply won't update anything in User table. However, do we need to update legacy table? I would remove this if not

packages/server/graphql/mutations/helpers/resolveDowngradeToStarter.ts
```diff
@@ -48,7 +52,9 @@ const resolveDowngradeToStarter = async (
     orgName: org.name,
     oldTier: 'team',
     newTier: 'starter',
-    billingLeaderEmail: user.email
+    billingLeaderEmail: user.email,
```
igorlesnenko, line 55: Maybe we don't need to track this property at all? Looks like it matches email, as we take this from `viewerId`.
@tghanken Do we need that property? 

```
analytics.track({
  userId: 'local|jQD5bQyFEs',
  event: 'Organization Downgraded',
  properties: {
    billingLeaderEmail: 'igor@example.com',
    domain: 'example.com',
    email: 'igor@example.com',
    newTier: 'starter',
    oldTier: 'team',
    orgId: 'jXKJCeJ9ok',
    orgName: 'archive test'
  }
});
```

igorlesnenko (approved):
Works!

tghanken (approved):
Looks good!

packages/server/graphql/mutations/helpers/resolveDowngradeToStarter.ts
nickoferrall: @igorlesnenko where does the email field get appended to the properties object?

igorlesnenko: @nickoferrall Looks like it is coming from this kind of calls: https://github.com/ParabolInc/parabol/blob/3c90a6facc7691ec482fee1a9e5504508eb49c18/packages/server/graphql/mutations/helpers/bootstrapNewUser.ts#L33

packages/client/utils/constants.ts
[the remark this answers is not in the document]
Dschoordsch: +1 If you're going for type safety, then you should not cast with `as` at the end. You can try this and add an invalid key and it will still work. If you instead declare the type of the variable, then this would be detected
```ts

export const reasonsToDowngradeLookup: Record<ReadableReasonToDowngradeEnum, ReasonToDowngradeEnum> = {
  'Parabol is too expensive': 'tooExpensive',
  'Budget changes': 'budgetChanges',
  'Missing key features': 'missingKeyFeatures',
  "Not using Parabol's paid features": 'notUsingPaidFeatures',
  'Moving to another tool (please specify)': 'anotherTool'
}
```

Dschoordsch (approved):
Would be nice if you could solve the double email issue in segment events before merging.

Issue #7832, opened 2023-03-01
RiD Activity Library: Basic grid of stubbed activity cards


jmtaber129:
Basic grid of stubbed activity cards, intended to support parallel work on impl of activity cards, search, and categories

AC:
- [ ] All available templates appear in the activity library view as unstyled stubs (e.g. just divs with the template name), including check-ins and standups

PR #7836, opened 2023-03-01
feat(retros-in-disguise): Base activity library + stubbed cards

jmtaber129:
# Description
Create the basic layout for the activity library + stubbed activity "cards".

From here, we can implement activity card designs in parallel with categories + search (which effectively just filter the activity cards shown).

Fixes https://github.com/ParabolInc/parabol/issues/7832

## Demo
Screen Shot 2023-03-02 at 4 39 26 PM

## Testing scenarios
- [ ] Confirm that team, org, and public retro + poker templates appear, and that Standup and Check-in cards are present.

## Final checklist

### DIFF
--- a/packages/client/components/ActivityLibrary/ActivityLibrary.tsx
+++ b/packages/client/components/ActivityLibrary/ActivityLibrary.tsx
@@ -1,16 +1,18 @@
 import graphql from 'babel-plugin-relay/macro'
-import React, {useEffect, useRef} from 'react'
+import React from 'react'
 import {PreloadedQuery, usePreloadedQuery} from 'react-relay'
 import {Redirect} from 'react-router'
 import {ActivityLibraryQuery, MeetingTypeEnum} from '~/__generated__/ActivityLibraryQuery.graphql'
-import useRouter from '../../hooks/useRouter'
-import sortByTier from '../../utils/sortByTier'
 import ActivityLibrarySideBar from './ActivityLibrarySideBar'
 import ActivityLibraryCard from './ActivityLibraryCard'
 
 graphql`
   fragment ActivityLibrary_template on MeetingTemplate {
     id
+    teamId
+    team {
+      name
+    }
     name
     type
   }
@@ -19,58 +21,15 @@ graphql`
 const query = graphql`
   query ActivityLibraryQuery {
     viewer {
+      availableTemplates(first: 100) @connection(key: "ActivityLibrary_availableTemplates") {
+        edges {
+          node {
+            ...ActivityLibrary_template @relay(mask: false)
+          }
+        }
+      }
       featureFlags {
         retrosInDisguise
-      }
-      teams {
-        id
-        name
-        tier
-        retroSettings: meetingSettings(meetingType: retrospective) {
-          ... on RetrospectiveMeetingSettings {
-            teamTemplates {
-              ...ActivityLibrary_template @relay(mask: false)
-            }
-            organizationTemplates(first: 20)
-              @connection(key: "ReflectTemplateListOrg_organizationTemplates") {
-              edges {
-                node {
-                  ...ActivityLibrary_template @relay(mask: false)
-                }
-              }
-            }
-            publicTemplates(first: 50)
-              @connection(key: "ReflectTemplateListPublic_publicTemplates") {
-              edges {
-                node {
-                  ...ActivityLibrary_template @relay(mask: false)
-                }
-              }
-            }
-          }
-        }
-        pokerSettings: meetingSettings(meetingType: poker) {
-          ... on PokerMeetingSettings {
-            teamTemplates {
-              ...ActivityLibrary_template @relay(mask: false)
-            }
-            organizationTemplates(first: 20)
-              @connection(key: "PokerTemplateListOrg_organizationTemplates") {
-              edges {
-                node {
-                  ...ActivityLibrary_template @relay(mask: false)
-                }
-              }
-            }
-            publicTemplates(first: 50) @connection(key: "PokerTemplateListOrg_publicTemplates") {
-              edges {
-                node {
-                  ...ActivityLibrary_template @relay(mask: false)
-                }
-              }
-            }
-          }
-        }
       }
     }
   }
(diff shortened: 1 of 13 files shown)


packages/client/components/ActivityLibrary/ActivityLibrary.tsx
```diff
@@ -23,10 +44,37 @@ export const ActivityLibrary = (props: Props) => {
...
     UNSTABLE_renderPolicy: 'full'
   })
+  const {viewer} = data
+  const {featureFlags, availableTemplates} = viewer
+
+  const templates = [
+    {id: 'action', type: 'action', name: 'Check-in', team: {name: 'Parabol'}},
+    {id: 'teamPrompt', type: 'teamPrompt', name: 'Standup', team: {name: 'Parabol'}},
+    ...availableTemplates.edges.map((edge) => edge.node)
+  ] as {
```
BartoszJarocki, line 54: +1 to have a better typesafety in this case it'd be better to specify type in the variable declaration
```
const templates: {
    id: string
    name: string
    type: MeetingTypeEnum
    team: {name: string}
  }[] = ...
```

as what casting does is basically saying, `not gonna check anything, I'll assume the structure is correct`

packages/server/graphql/public/types/User.ts
```diff
@@ -57,6 +63,50 @@ const User: UserResolvers = {
...
+        : user.tms.filter((teamId: string) => authToken.tms.includes(teamId))
+
+    // Get the team templates.
+    const teamTemplates = await dataLoader
+      .get('meetingTemplatesByType')
+      .loadMany([
+        ...teamIds.map((teamId) => ({teamId, meetingType: 'poker' as MeetingTypeEnum})),
+        ...teamIds.map((teamId) => ({teamId, meetingType: 'retrospective' as MeetingTypeEnum}))
+      ])
+    const scoredTeamTemplates = await getScoredTemplates(teamTemplates.filter(isValid).flat(), 0.9)
```
BartoszJarocki, line 82: -1 what does `0.9` mean here and why this particular value? it'd be great to either extract a const with some comment so it's easier to understand what's happening here

packages/server/graphql/public/types/User.ts
```diff
@@ -57,6 +63,50 @@ const User: UserResolvers = {
...
+    const orgIds = [...new Set(teams.filter(isValid).map((team) => team.orgId))]
+    const orgTemplatesResult = await dataLoader.get('meetingTemplatesByOrgId').loadMany(orgIds)
+    const organizationTemplates = orgTemplatesResult
+      .filter(isValid)
+      .flat()
+      .filter(
+        (template: MeetingTemplate) =>
+          template.scope !== 'TEAM' && !teamIds.includes(template.teamId)
+      )
+    const scoredOrgTemplates = await getScoredTemplates(organizationTemplates, 0.8)
```
BartoszJarocki, line 95: -1 same here

BartoszJarocki (changes requested):
Looks good! Left a few comments, it's ready to go afterwards 👍

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
```diff
@@ -1,12 +1,33 @@
...
+    }
+    name
+    type
+  }
+`
 
 const query = graphql`
   query ActivityLibraryQuery {
     viewer {
+      availableTemplates(first: 100) @connection(key: "ActivityLibrary_availableTemplates") {
```
BartoszJarocki, line 24: +1 github lost my comment, or I somehow deleted it. it's a little bit of overkill to have a connection  here, as we won't have too many, no harm though

packages/server/graphql/public/types/User.ts
```diff
@@ -57,6 +64,53 @@ const User: UserResolvers = {
     const stripeLineItems = await fetchAllLines(invoiceId)
     const invoice = await manager.retrieveInvoice(invoiceId)
     return generateInvoice(invoice, stripeLineItems, orgId, invoiceId, dataLoader)
+  },
+  availableTemplates: async ({id: userId}, {first, after}, {authToken, dataLoader}) => {
```
BartoszJarocki, line 68: +1 is it possible to combine type definition via SDL and the old way? if yes, we should try to use the new way when we can

BartoszJarocki (approved):
lgtm! moving to maintainer review, @mattkrick do you mind taking a look?

packages/server/graphql/public/types/User.ts
jmtaber129: Could you elaborate? This is on the new SDL `User` type.

BartoszJarocki: oh god, long day, sorry 🤦

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
jmtaber129: Hmm I don't disagree, but since the existing `organizationTemplates` and `publicTemplates` fields on the meeting settings objects are connections, I have a weak preference to either a) make this a connection, too, or b) make this an array and make those fields arrays.

mattkrick (changes requested):
a few minor things we can fix later, but only 1 thing blocking the merge.

packages/client/components/ActivityLibrary/ActivityLibraryCard.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react'
+
+interface Props {
+  name: string
+  type: string
+  teamName: string
+}
+
+const ActivityLibraryCard = (props: Props) => {
```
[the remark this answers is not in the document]
jmtaber129: I don't foresee us using this full component in the `/meetings` dashboard, since it will eventually have onClick handlers, etc. for entering the activity details view. 

We can refactor the relevant parts of this into its own component once we add the UI portion of this to the dashboard.

packages/server/graphql/queries/helpers/getScoredTemplates.ts
```diff
@@ -1,8 +1,8 @@
 import db from '../../../db'
 import getTemplateScore from '../../../utils/getTemplateScore'
 
-const getScoredTemplates = async (
-  templates: {createdAt: Date; id: string}[],
+const getScoredTemplates = async <T extends {createdAt: Date; id: string}>(
```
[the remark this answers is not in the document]
jmtaber129: Interesting context! However, to limit the scope of this PR, I'd prefer to leave this as-is for now, and we can update the ranking logic at a later date.

packages/server/utils/getTemplateScore.ts
```diff
@@ -12,8 +12,13 @@ const SATURTION = 1 // max score, we want a score that is 0-1
 const MIDPOINT = 45 // after this many days, the score will be SATURATION / 2
 const GROWTH_INTERVAL = 30 // 80% of the decline happens in this many days
 
+// Default hotness factors for ranking templates
+export const ORG_HOTNESS_FACTOR = 0.8
+export const TEAM_HOTNESS_FACTOR = 0.9
+
 const getAgeScore = (age: number) => {
-  return SATURTION / (1 + Math.exp((Math.log(81) / GROWTH_INTERVAL) * (age - MIDPOINT)))
+  const ageDays = age / 1000 / 60 / 60 / 24
```
jmtaber129, line 20: @mattkrick this came up when I was regression testing removing the team templates DB call - like you mentioned we'll probably change how we rank templates at some point, but I figured we should fix make this fix, since otherwise all templates will score 0 due to age being millis when the logic assumes it's days.

packages/server/graphql/public/types/User.ts
```diff
@@ -57,6 +64,53 @@ const User: UserResolvers = {
...
+      ])
+    const scoredTeamTemplates = await getScoredTemplates(
+      teamTemplates.filter(isValid).flat(),
+      TEAM_HOTNESS_FACTOR
+    )
+
+    // Get the org templates.
+    const teams = await dataLoader.get('teams').loadMany(teamIds)
+    const orgIds = [...new Set(teams.filter(isValid).map((team) => team.orgId))]
+    const orgTemplatesResult = await dataLoader.get('meetingTemplatesByOrgId').loadMany(orgIds)
```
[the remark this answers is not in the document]
jmtaber129: Good catch! Removed the team templates call and pulled those templates from these results instead.

packages/client/components/ActivityLibrary/ActivityLibraryCard.tsx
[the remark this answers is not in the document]
BartoszJarocki: will allow myself to nitpick a little

1. Parts of this component should be reusable as they're exactly the same (ie image part with this rounded thingy in the corners)
Screenshot 2023-03-09 at 07 23 11

2. imo activity details should be a route like `/activity-library/:activity-id` which preferably makes this component a link (this also will be good for accessibility and keyboard navigation reasons), see the video below

https://user-images.githubusercontent.com/1017620/223938475-b73a31c0-7f62-47b2-ba18-0221a742ac96.mp4

if it's a link we can easily use tab + enter to navigate that view, which is a win.

it's not really related to this pr though 🙃

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
```diff
@@ -23,10 +44,32 @@ export const ActivityLibrary = (props: Props) => {
   const data = usePreloadedQuery<ActivityLibraryQuery>(query, queryRef, {
     UNSTABLE_renderPolicy: 'full'
   })
+  const {viewer} = data
+  const {featureFlags, availableTemplates} = viewer
+
+  const templates = [
+    {id: 'action', type: 'action', name: 'Check-in', team: {name: 'Parabol'}},
```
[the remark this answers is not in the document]
jmtaber129: I'm thinking since we'll probably need to do some special-casing on the client-side for standups+check-ins, it _might_ be better to add these on the client-side - i.e. the server gives the client all the meeting _templates_, and it's up to the client to use those templates to show the available activities.

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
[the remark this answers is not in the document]
jmtaber129: I kind of disagree with this for the reasons mentions below - we don't have an "activity" object in the server-side data model, but we do have templates + meetings, so the server should return templates, and then the client uses those templates to display available activities.

packages/server/utils/getTemplateScore.ts
mattkrick: sounds good!

PR #7848, opened 2023-03-02
chore(metrics): Add content_group in GA4 page_view event

tianrunhe:
# Description

Fixes https://github.com/ParabolInc/parabol-data-sanctum/issues/544

## Demo
Screenshot 2023-03-01 at 16 14 34

Screenshot 2023-03-01 at 16 14 51

## Testing scenarios

Install [this Chrome extension](https://chrome.google.com/webstore/detail/gtmga-debugger/ilnpmccnfdjdjjikgkefkcegefikecdc?hl=en&utm_source=gtm-ga) to make testing easier

  - Start the local server
  - Open up Chrome developer tools and click the "Analytics Debugger", then "Start debugging"
  - Go to any page in the app
  - Expand the item in Analytics Debugger and click the "GA4" tab, there should be:
    1. An page_view event
    2. Expand the EVENT PARAMETERS, you should see an associated value with `content_group`  

## Final checklist

### DIFF
--- a/packages/client/components/AnalyticsPage.tsx
+++ b/packages/client/components/AnalyticsPage.tsx
@@ -100,7 +100,6 @@ const AnalyticsPage = () => {
     }
   }, [ReactGA.isInitialized, gaMeasurementId])
 
-  const {href, pathname} = location
   useEffect(() => {
     const configGA = async () => {
       if (!ReactGA.isInitialized || !isSegmentLoaded) {
@@ -108,9 +107,6 @@ const AnalyticsPage = () => {
       }
 
       const {viewerId} = atmosphere
-      ReactGA.set({
-        content_group: getContentGroup(pathname)
-      })
       if (viewerId) {
         const res = await atmosphere.fetchQuery<AnalyticsPageQuery>(query)
         if (!res) return
@@ -128,9 +124,10 @@ const AnalyticsPage = () => {
       }
     }
     configGA()
-  }, [ReactGA.isInitialized, atmosphere.viewerId, pathname])
+  }, [ReactGA.isInitialized, atmosphere.viewerId])
 
   /* eslint-disable */
+  const {href, pathname} = location
   const pathnameRef = useRef(pathname)
   const segmentKey = window.__ACTION__.segment
   useEffect(() => {
--- a/packages/client/components/AnalyticsPage.tsx
+++ b/packages/client/components/AnalyticsPage.tsx
@@ -11,6 +11,7 @@ import safeIdentify from '~/utils/safeIdentify'
 import {AnalyticsPageQuery} from '~/__generated__/AnalyticsPageQuery.graphql'
 import useScript from '../hooks/useScript'
 import getAnonymousId from '../utils/getAnonymousId'
+import getContentGroup from '../utils/getContentGroup'
 import makeHref from '../utils/makeHref'
 
 const query = graphql`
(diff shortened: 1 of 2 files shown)


packages/client/components/AnalyticsPage.tsx
```diff
@@ -93,7 +94,7 @@ const AnalyticsPage = () => {
     if (gaMeasurementId) {
       ReactGA.initialize(gaMeasurementId, {
         gtagOptions: {
-          send_page_view: true
```
tianrunhe, line 96: By removing this line, we fall back to the [default setting](https://github.com/PriceRunner/react-ga4/blob/f31ed502adec6dc9dd68e3b59d39c1857880f433/src/ga4.js#L190), which is `send_page_view = false`. This is because we are going to manually send the `page_view` event on Line 196.

packages/client/utils/getContentGroup.ts
```diff
@@ -0,0 +1,24 @@
+export default function getContentGroup(pathname: string) {
```
tianrunhe, line 1: This is copied from the current setting in [UA](https://analytics.google.com/analytics/web/#/a74109115w112182245p117099558/admin/content-groupings/m-content-contentGroupingTable.rowShow=10&m-content-contentGroupingTable.rowStart=0&m-content-contentGroupingTable.sortColumnId=dateModified&m-content-contentGroupingTable.sortDescending=true)

packages/client/components/AnalyticsPage.tsx
Dschoordsch: -1 I'm not sure if it's this part, but I get each page view event twice, only once with the `content_group` set
https://www.loom.com/share/62de4f1c77b448d28c3fcf956598e34d

Dschoordsch (changes requested):

tianrunhe:
I can't reproduce the duplicate events on my local... I could confirm duplicate events were emitted from your environment though. Comparing these two events, let's call the one with `content_group` set as "the correct event", and the duplicate one as "the duplicate event", I found the following:

1. The duplicate event has `page_referrer = http://localhost:3000/team/hN5okO02Zi`; the correct event does not have this field
2. The duplicate event's `page_location` is `http://localhost:3000/team/hN5o0LzMEU` whereas the correct event's value is `http://localhost/team/hN5o0LzMEU` 🤷‍♂️
3. The duplicate event reports 641ms engagement time. The correct event reports 5.7s

It looks to me the "union" of these two events are what really happened? i.e., you switched from one team page to another and stayed there for a while.

Can you reproduce this in an incognito window?

Dschoordsch:
I could not find a way to see the events 1 by 1, but also from an incognito tab the events are sent twice
Screenshot from 2023-03-03 10-00-44

tianrunhe:
I finally figured out: previously we do not send `page_view` events manually. GA4 did this automatically for us because we have the "Page changes based on browser history events" turned on in Enhanced measurement:

Screenshot 2023-03-06 at 13 46 18

With this change, we are now manually emitting `page_view` events so we should turned off this feature, otherwise we will get duplicate events. Why did you see duplicate events but I didn't? It's because you were using `GA_TRACKING_ID='G-B937HFKWWD'` which points to [Staging GA4](https://analytics.google.com/analytics/web/#/p327845365/) that has the option turned on. I have been testing on [Local GA4](https://analytics.google.com/analytics/web/#/p346024326) that has the option turned off.

I have updated [`action-devops`](https://github.com/ParabolInc/action-devops/commit/f3d3263c5adcec3e41c90d117357f2a2ff1996ac) to reflect the correct GA4 and Segment keys for local environment.

Can you update your `.env` with `GA_TRACKING_ID='G-HJYW8K1HQW'` and test again? Thank you!

Dschoordsch (approved):

PR #7854, opened 2023-03-02
feat: add team filter in meeting view

### LINKED ISSUE #5518: Filter meetings view by team
I would like to be able to filter the meetings view by team. This would help me find the meeting that I'm looking for faster.

We're able to filter tasks by team. I suggest adding a similar drop-down filter to the meetings view.

meetings-view

tasks-view

### AC

- [ ] Meetings view includes a drop-down menu that includes all of the teams that the user is in
- [ ] Selecting a team in the drop-down menu filters the meeting cards to only show the meetings that belong to that team
- [ ] If the user selects a team that has no meetings, show an empty state card. Alternatively, we could only show teams in the drop-down that have meetings, but that could be confusing.

tianrunhe:
# Description

Fixes #5518 

## Testing scenarios

- [ ] Scenario A
  - Step 1
  - Step 2...

- [ ] Scenario B
  - Step 1
  - Step 2....

## Final checklist

### DIFF
--- a/packages/client/components/MeetingsDash.tsx
+++ b/packages/client/components/MeetingsDash.tsx
@@ -3,29 +3,16 @@ import graphql from 'babel-plugin-relay/macro'
 import React, {RefObject, useMemo} from 'react'
 import {createFragmentContainer} from 'react-relay'
 import {MeetingsDash_viewer} from '~/__generated__/MeetingsDash_viewer.graphql'
-import useAtmosphere from '../hooks/useAtmosphere'
 import useBreakpoint from '../hooks/useBreakpoint'
 import useCardsPerRow from '../hooks/useCardsPerRow'
-import {MenuPosition} from '../hooks/useCoords'
 import useDocumentTitle from '../hooks/useDocumentTitle'
-import useMenu from '../hooks/useMenu'
 import useTransition from '../hooks/useTransition'
-import {
-  Breakpoint,
-  EmptyMeetingViewMessage,
-  Layout,
-  UserTaskViewFilterLabels
-} from '../types/constEnums'
+import {Breakpoint, EmptyMeetingViewMessage, Layout} from '../types/constEnums'
 import getSafeRegex from '../utils/getSafeRegex'
-import lazyPreload from '../utils/lazyPreload'
-import makeMinWidthMediaQuery from '../utils/makeMinWidthMediaQuery'
-import {useUserTaskFilters} from '../utils/useUserTaskFilters'
-import DashSectionControls from './Dashboard/DashSectionControls'
-import DashSectionHeader from './Dashboard/DashSectionHeader'
-import DashFilterToggle from './DashFilterToggle/DashFilterToggle'
 import DemoMeetingCard from './DemoMeetingCard'
 import MeetingCard from './MeetingCard'
 import MeetingsDashEmpty from './MeetingsDashEmpty'
+import MeetingsDashHeader from './MeetingsDashHeader'
 import StartMeetingFAB from './StartMeetingFAB'
 import TutorialMeetingCard from './TutorialMeetingCard'
 
@@ -33,30 +20,6 @@ interface Props {
   meetingsDashRef: RefObject<HTMLDivElement>
   viewer: MeetingsDash_viewer | null
 }
-
-const desktopBreakpoint = makeMinWidthMediaQuery(Breakpoint.SIDEBAR_LEFT)
-
-const UserDashTeamMenu = lazyPreload(
-  () =>
-    import(
-      /* webpackChunkName: 'UserDashTeamMenu' */
-      './UserDashTeamMenu'
-    )
-)
-
-const StyledDashFilterToggle = styled(DashFilterToggle)({
-  margin: '4px 16px 4px 0',
-  [desktopBreakpoint]: {
-    margin: '0 24px 0 0'
-  }
-})
-
-const UserTasksHeaderDashSectionControls = styled(DashSectionControls)({
-  justifyContent: 'flex-start',
-  flexWrap: 'wrap',
-  width: '100%',
-  overflow: 'initial'
-})
 
 const Wrapper = styled('div')<{maybeTabletPlus: boolean}>(({maybeTabletPlus}) => ({
   padding: maybeTabletPlus ? 0 : 16,
@@ -76,8 +39,6 @@ const EmptyContainer = styled('div')({
 })
 
 const MeetingsDash = (props: Props) => {
-  const atmosphere = useAtmosphere()
-  const {viewerId} = atmosphere
   const {meetingsDashRef, viewer} = props
   const {teams = [], preferredName = '', dashSearch, activeMeetings: meetings} = viewer ?? {}
   const activeMeetings = useMemo(() => {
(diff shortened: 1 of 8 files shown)


jmtaber129 (changes requested):
-1 There's a bug with meeting updates getting published - if I start a meeting as a different user, I need to refresh for that meeting to appear under "All Teams". It may be better to keep the queries+API the same, and only filter down the specific teams on the client-side. That would simplify this and cut down the diff of this PR.

+1 Showing the existing "no meeting" state when filtering to a specific team feels odd - maybe we replace the team-filtered empty state with something like "this team has no meetings - click here to see all meetings"? As-is, it says "Looks like you have no upcoming meetings" which isn't necessarily accurate. (cc @enriquesanchez)

+1 The filter is misaligned for me - looks like a `maxWidth: Layout.TASK_COLUMNS_MAX_WIDTH` in `DashSectionHeader`. It only manifests when debug tools are closed or in a separate window:
Screen Shot 2023-03-06 at 5 08 45 PM

enriquesanchez:
> Showing the existing "no meeting" state when filtering to a specific team feels odd - maybe we replace the team-filtered empty state with something like "this team has no meetings - click here to see all meetings"? 

@tianrunhe and I had talked about this and decided to postpone it to keep the scope small, particularly given how Bruce has been working on this "on the side". That said, I fully agree with you, and I think even just a simple paragraph would be better than the current behavior. What do you think Bruce? Can we change the empty state as part of this effort?

tianrunhe:
> > Showing the existing "no meeting" state when filtering to a specific team feels odd - maybe we replace the team-filtered empty state with something like "this team has no meetings - click here to see all meetings"?
> 
> @jmtaber129 @tianrunhe and I had talked about this and decided to postpone it to keep the scope small, particularly given how Bruce has been working on this "on the side". That said, I fully agree with you, and I think even just a simple paragraph would be better than the current behavior. What do you think Bruce? Can we change the empty state as part of this effort?

For sure, will do it after we sort out the implementation details

tianrunhe:
@jmtaber129 

>  It may be better to keep the queries+API the same, and only filter down the specific teams on the client-side. That would simplify this and cut down the diff of this PR.

Can you explain this a bit more? I'm not sure I fully get it. Thank you! 🙏

tianrunhe:
@jmtaber129 I think I got it working from client side and was able to reduce the size of this PR! Let me know if this is what you have suggested!

packages/client/components/MeetingsDash.tsx
```diff
@@ -113,7 +122,7 @@ const MeetingsDash = (props: Props) => {
           <MeetingsDashEmpty
             name={preferredName}
             message={
-              dashSearch
+              dashSearch || teamFilter
                 ? EmptyMeetingViewMessage.NO_SEARCH_RESULTS
```
jmtaber129, line 126: +1 I don't think the existing `NO_SEARCH_RESULTS` copy is effective when the user is only filtering by team, since it specifies "query" (as in, search query), and it may not be clear that you're filtering down to a specific team.

IMO ideally, this would read something like `we could not find any meetings on this team`, with a link that looks like `See meetings on all teams` immediately after.

packages/client/components/MeetingsDashTeamMenu.tsx
```diff
@@ -0,0 +1,104 @@
...
+      teamFilter {
+        id
+      }
+      teams {
+        id
+        name
+      }
+    }
+  `
+})
```
jmtaber129, line 104: -1 This needs to use `useFragment`

jmtaber129 (changes requested):
-1 This implements functionality very similar to https://github.com/ParabolInc/parabol/blob/7ca78bd94ce6c2e601faa854e4bd5bc5ca31591a/packages/client/components/UserDashTeamMenu.tsx, but with a different approach (storing team filter in client GQL schema vs. URL query param). What's the rationale for doing it this way vs. reusing the existing pattern used in the user dash?

-1 Related to ^, we should refactor the menu to be reused between the user and meeting dashboards. As-is, we could have two identical-looking team filter menu components with subtly different behavior. I'd suggest extracting a presentational component that takes the list of teams, the selected team, an `onSelectTeam` callback, a `showAllTeams` bool, and `menuProps`.

+1 The filter misalignment mentioned in https://github.com/ParabolInc/parabol/pull/7854#pullrequestreview-1326844555 is still an issue for me.

tianrunhe:
@jmtaber129 Thanks for meeting with me last week! I finally get some time to revise it. Let me know if this revision matches what you have suggested. Thank you!

packages/client/components/MeetingsDash.tsx
```diff
@@ -75,15 +81,18 @@ const MeetingsDash = (props: Props) => {
...
       })
-    const filteredMeetings = dashSearch
-      ? meetings.filter(({name}) => name && name.match(getSafeRegex(dashSearch, 'i')))
-      : meetings
+    const searchedMeetings = dashSearch
+      ? sortedMeetings.filter(({name}) => name && name.match(getSafeRegex(dashSearch, 'i')))
+      : sortedMeetings
+    const filteredMeetings = searchedMeetings.filter((node) =>
+      teamIds ? teamIds.includes(node.teamId) : true
+    )
```
jmtaber129, line 89: +1 Move the ternary outside the `filter`? I.e. `teamIds ? searchedMeetings.filter(...) : searchedMeetings`, similar to the ternary for `searchedMeetings`

jmtaber129 (changes requested):
Code looks good! Besides one +1 inline, two things:
* -1 that misalignment bug is still present for me. Are you able to reproduce? I'm on a 16-inch MacBook.
* +1 We should add that paragraph for empty state per https://github.com/ParabolInc/parabol/pull/7854#issuecomment-1456843213

jmtaber129 (changes requested):
-1 As-is, this changes the empty state for when "all teams" are selected and the search query matches no meetings. Additionally, saying "no meetings match your _query_" when the user didn't type anything in feels a bit odd. I'd suggest the following copy for the different cases:
* No active meetings on the team: "Looks like you have no upcoming meetings on this team. Click here to see meetings on all teams"
* No active meetings matching query on team: "Sorry we could not find any meetings matched with your query on this team. Click here to see meetings on all teams"
* No active meetings when "All teams" selected - unchanged
* No active meetings matching query when "all teams" selected - unchanged

tianrunhe:
@jmtaber129 Sorry I misunderstood your comments earlier. Now I think I get it: we would like to match with existing behavior. Now I think it should be good:

- No active meetings when "All teams" selected - unchanged
Screenshot 2023-05-31 at 10 48 30


- No active meetings matching query when "all teams" selected - unchanged
Screenshot 2023-05-31 at 10 47 15


- No active meetings on the team: "Looks like you have no upcoming meetings on this team. Click here to see meetings on all teams"
Screenshot 2023-05-31 at 10 47 53


- No active meetings matching query on team: "Sorry we could not find any meetings matched with your query on this team. Click here to see meetings on all teams"
Screenshot 2023-05-31 at 10 48 08

jmtaber129 (changes requested):
@tianrunhe ehh this still isn't quite right, since the empty state for "no meetings existing across all teams" and "no meetings across all teams match the query" should be different. The "no meetings match the query" case should be "Sorry we could not find any meetings matched with your query. Would you like to checkout these tips and tricks?". See screenshot from `master`:

Screen Shot 2023-05-31 at 2 09 15 PM

tianrunhe:
If you haven't noticed yet, I'm competing for the title of "longest PR ever" 😢. @jmtaber129 Thanks so much for your patience!

jmtaber129 (approved):
One opportunity to simplify some components through conditional rendering, otherwise LGTM!

packages/client/components/MeetingsDashHeader.tsx
```diff
@@ -0,0 +1,106 @@
...
+    menuProps: teamFilterMenuProps
+  } = useMenu(MenuPosition.UPPER_RIGHT, {
+    isDropdown: true
+  })
+  const oldTeamsRef = useRef<MeetingsDashHeader_viewer$data['teams']>([])
+  const nextTeams = viewer?.teams ?? oldTeamsRef.current
+  if (nextTeams) {
+    oldTeamsRef.current = nextTeams
+  }
+  const teams = oldTeamsRef.current
```
Dschoordsch, line 79: -1 It's not obvious what you're trying to achieve here, please add a comment

Dschoordsch (changes requested):
Tested, works

packages/client/components/MeetingsDashEmpty.tsx
```diff
@@ -30,15 +34,43 @@ const Copy = styled('p')({
 interface Props {
   name: string
   message: string
+  filteredResult?: boolean
```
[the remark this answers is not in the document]
tianrunhe: The boolean is used to check if a team filter is selected, so I renamed it to `isTeamFilterSelected`

packages/client/components/MeetingsDashEmpty.tsx
```diff
@@ -30,15 +34,43 @@ const Copy = styled('p')({
...
+  }
+
+  const {name, message, filteredResult} = props
   return (
     <Section>
       <Heading>{`Hi ${name},`}</Heading>
-      <Copy>{`${message}`}</Copy>
+      <Copy>
+        {message}
+        {filteredResult ? (
```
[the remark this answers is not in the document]
tianrunhe: 3rd time is the charm. I'll wait for when we have more than 2 components 😅

Dschoordsch (approved):

Issue #5130, opened 2021-07-13
add audio cue when timer runs out 


drewhousman1010:
A [user](https://app.hubspot.com/contacts/3888472/contact/17542651) 🔒  writes:

> When the timer stops it doesn't do anything. An audio cue would be nice.

This is related to [issue 5010](https://github.com/ParabolInc/parabol/issues/5010)

jordanh:
AC
- [ ] design exploration for adding sound when time expires

EE
- [ ] 16 hours

mattkrick:
my vote is for this sound: https://vm.tiktok.com/ZMd7Guy9Y/

Issue #5808, opened 2021-12-17
Missing check-in summary


ackernaut:
Responding to @igorlesnenko in https://github.com/ParabolInc/parabol/issues/5720#issuecomment-992352481_

### Steps
- In production, we ended a check-in meeting (ID = 9aJRkkfs6k)
- I received an empty email message

### Acceptance criteria
- Triaged the bug and fixed, or updated or created a new issue

### Estimated effort
- 3 hours

### Artifacts

![Screen Shot 2021-12-17 at 12 07 19](https://user-images.githubusercontent.com/307286/146589083-d13c44e0-5a0d-4345-b056-db6b2e923947.png)

<details>
  <summary>click for email html</summary>

```Delivered-To: terry@parabol.co
Received: by 2002:a05:6a10:fc0b:0:0:0:0 with SMTP id fy11csp796264pxb;
        Thu, 9 Dec 2021 09:22:17 -0800 (PST)
X-Google-Smtp-Source: ABdhPJwDvNROYc6+96dAoV9whnW3BkNxaeAL0FlCrnS5qdCjkaHBmGN18bneLIX0mIx3+pHOLRSN
X-Received: by 2002:a25:7b05:: with SMTP id w5mr7345226ybc.655.1639070537280;
        Thu, 09 Dec 2021 09:22:17 -0800 (PST)
ARC-Seal: i=1; a=rsa-sha256; t=1639070537; cv=none;
        d=google.com; s=arc-20160816;
        b=aNNG6AeRFDuiBr5pUKr5VVBwd6OFVCpIHLG20qiPtZEpzQceeAvBMcSwdDYZYHj8EQ
         oGP/jHZ/sXsJiUxP9m9NyltSsMyZjrAKjLFVKo3UFixXyfkVpILleqSMeiRtRM7oLD84
         9TxsGqJuPy1jcafLHH+aafdrLdOohw1+GjvE7Tgb5V7Xs2G1Pc5AfmguH+5ug24g7qop
         mxbwyMpwrHVqbMuLzgBig3m1D2DXvUv2GREtDK9BsXiM0G1tKHteC7BPa75u/LTnb6Nv
         e4GwAkHHf++PHaLZzee6lbNvj2R2E2ck2LmY8SVjkb38BbFgongoAvH0I37y2DccCu/U
         J2pw==
ARC-Message-Signature: i=1; a=rsa-sha256; c=relaxed/relaxed; d=google.com; s=arc-20160816;
[...]

PR #7858, opened 2023-03-03
feat: add sound effect for stage timer

paul-vd:
# Description

This pull request fixes an issue that we have been having during our retrospectives with the timer runs out silently causing us to overrun time dedicated to a topic. The solution is to add a sound cue that plays when the timer reaches zero, alerting everyone in the session that the time is up. This way, the participants can be more aware of the time limit and act accordingly. The sound queue is implemented using the Audio API 

In the future it can be improved by allowing to enable/disable.

Fixes https://github.com/ParabolInc/parabol/issues/5130

## Final checklist


packages/client/components/StageTimerDisplayGauge.tsx
```diff
@@ -30,11 +31,33 @@ const Gauge = styled('div')<{isTimeUp: boolean; isDesktop: boolean}>(({isTimeUp,
...
+  // a page when the timer is already up.
+  const hasPlayed = useRef(false)
+
+  useEffect(() => {
+    hasPlayed.current = isTimeUp
+  }, [])
+
+  useEffect(() => {
+    if (isTimeUp && !hasPlayed.current) {
+      new Audio(sound).play()
```
nickoferrall, line 45: +1 this is creating the audio element everytime the hook is called so we could move it out of the useEffect and memoize it, say on line 35: `const audioElement = useMemo(() => new Audio(sound), [])`

It's a small file though so not a big deal at all. It might matter if we reuse this hook with larger audio files

packages/client/components/StageTimerDisplayGauge.tsx
```diff
@@ -1,5 +1,6 @@
 import styled from '@emotion/styled'
-import React from 'react'
+import React, {useEffect, useRef} from 'react'
+import sound from '../../../static/sounds/tic-tac.mp3'
```
nickoferrall, line 3: +1 I like the sound although it's not super obvious to me that it signals the end of a timer. Perhaps the countdown game show sound would be more obvious? 

Also, very nit-picky, but we could change the variable name from `sound` to `endTimerSound`

nickoferrall (approved):
Love the PR! Excited to see this in the app

packages/client/components/StageTimerDisplayGauge.tsx
Dschoordsch: I like the sound as it's not very distracting. I would love to try it for some time to get a feel how it behaves in a real meeting. My first reaction was to put it behind a feature flag, but I don't think it's unsafe to try it as is.
@ackernaut As this is introducing sound into our app, I think you might want to check it.

packages/client/components/StageTimerDisplayGauge.tsx
paul-vd: Agreed, I will update that.

ackernaut (commented):
I think we should iterate on the sound.
1) When the timer was up, then you hear a timer sound like when a timer is running, instead of a ringing sound like when a timer reaches 0
2) The sound seemed a little quiet. Being quiet and ambiguous (running sound instead of ending sound) just caused for a sub-par experience

The baseline for quality I would compare to would be FigJam:
1) They have a timer module in the upper left
2) At the last 4-5 seconds it starts lightly beeping (a timer is running out kind of sound) and then when it hits zero it has a brief ringing sound. The ringing sound is louder than the brief countdown beeping sound. I think the volume levels are good.
3) When it hits zero the timer does a little side-to-side shake or shimmy. Nice touch

My vote is for iteration here, but I don’t want to block forward progress. I would like to hear opinions on sound design and UX from @enriquesanchez and @acressall 

I’d be happy for one of them to partner with you if you want extra design input and one of them wants to help here.

I’m glad we’re experimenting here!!

enriquesanchez:
Hey @paul-vd, thank you so much for working on this! 🙌 

[Here's a youtube video that demos the FigJam timer](https://youtu.be/B0M0HtxDGo8?t=44) that both @ackernaut and @acressall have mentioned. I think we all agree that something very similar to this is what we want.

I just tested your PR and I think you already have the countdown nailed, it just a matter of adding the ring at the end of the timer.

Let me know if this helps.

paul-vd:
> Hey @paul-vd, thank you so much for working on this! 🙌
> 
> [Here's a youtube video that demos the FigJam timer](https://youtu.be/B0M0HtxDGo8?t=44) that both @ackernaut and @acressall have mentioned. I think we all agree that something very similar to this is what we want.
> 
> I just tested your PR and I think you already have the countdown nailed, it just a matter of adding the ring at the end of the timer.
> 
> Let me know if this helps.

Thanks for the suggestions, I'm looking at this now!

paul-vd:
I don't know if my development environment is not setup correctly, but it's a real pain developing, I get a let of hot reloading for no reason, and failed fetch's which is really strange, I have not taken the time to investigate how the stack has been setup. I will try to get back to this a bit later. 🦥

I'm a taker for anyone that would like to pair on this? a simple reproduction is to add the sound to an audio element.
```tsx
import endTimerSound from '../../../static/sounds/tic-tac.mp3'

const StageTimerDisplayGauge = (props: Props) => {
  const {endTime} = props
  useRefreshInterval(1000)
  const isDesktop = useBreakpoint(Breakpoint.SINGLE_REFLECTION_COLUMN)
  const timeLeft = endTime && countdown(endTime)

  const fromNow = timeLeft || 'Time’s Up!'
  return (
    <>
      <Gauge isDesktop={isDesktop} isTimeUp={!timeLeft}>
        {fromNow}
      </Gauge>
      <audio id='timer-end-sound-effect' aria-hidden className='hidden' autoPlay={false}>
        <source src={endTimerSound} type='audio/mp3' />
      </audio>
    </>
  )
}
```

I first noticed this with the `new Audio` that's why I tested with an audio element, any ideas?

Dschoordsch:
@paul-vd I'm not sure I can reproduce the issue. I get some reloads due to this error
```
09:39:55 3|Dev Server        | TypeError: Cannot set property closed of #<Readable> which has only a getter
09:39:55 3|Dev Server        |     at FsReadStream.close (/home/man/Work/parabol/parabol/node_modules/memfs/lib/volume.js:2047:17)
09:39:55 3|Dev Server        |     at FsReadStream._destroy (/home/man/Work/parabol/parabol/node_modules/memfs/lib/volume.js:2032:10)
09:39:55 3|Dev Server        |     at _destroy (node:internal/streams/destroy:109:10)
09:39:55 3|Dev Server        |     at class_1.destroy (node:internal/streams/destroy:71:5)
09:39:55 3|Dev Server        |     at class_1.<anonymous> (/home/man/Work/parabol/parabol/node_modules/memfs/lib/volume.js:1962:22)
09:39:55 3|Dev Server        |     at class_1.emit (node:events:525:35)
09:39:55 3|Dev Server        |     at endReadableNT (node:internal/streams/readable:1359:12)
09:39:55 3|Dev Server        |     at process.processTicksAndRejections (node:internal/process/task_queues:82:21)
09:39:55 PM2                 | App [Dev Server:3] exited with code [1] via signal [SIGINT]
09:39:55 PM2                 | App [Dev Server:3] starting in -fork mode-
09:39:55 PM2                 | App [Dev Server:3] online
09:39:56 3|Dev Server        | 📘 Using cached DLL
09:40:05 3|Dev Server        | webpack compiled successfully
```
Is that what you're seeing?

Dschoordsch:
updating [`webpack-dev-server`](https://github.com/paul-vd/parabol/blob/8ce3d12ae3cc936955ad8cd368afd2f2e1054ab9/packages/client/package.json#L64) to `^4.11.1` fixes the issue for me

Dschoordsch:
@paul-vd are you still working on it? Did you try my proposed fix?

paul-vd:
@Dschoordsch Thanks I will give that a go, I have not yet had time to test it out! I should have some time later today or tomorrow.

static/sounds/tic-tac.mp3
Dschoordsch, line 1: For future reference, the sound appears to originate from https://bigsoundbank.com/sound-2816-tic-tac-mechanical-alarm-clock-5.html which is under CC0

Dschoordsch:
Closing in favour of #5808 where I do some cleanup on this PR. @paul-vd thanks for initiating this!

Issue #7692, opened 2023-01-30
Checkout Flow: Show teams limit warning message


nickoferrall:
Predecessor: https://github.com/ParabolInc/parabol/issues/7529

<img width="478" alt="Screenshot 2023-01-30 at 16 06 37" src="https://user-images.githubusercontent.com/39854876/215606311-690604a2-cbf9-4151-8f6a-632a876e0c8f.png">

### AC

- [ ] Show the teams limit warning message in the plans sections of the new checkout flow

PR #7879, opened 2023-03-07
feat(checkout-flow): add teams limit warning

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7692

In this PR, I'm using the UI that we have in Insights as I think we should be consistent with the UI & messaging.

Org with the `scheduledToLock`:
warning-1

Org with the `lockedAt`:
warning-2

### To test

- [ ] View the billing page for an org that has the `scheduledToLock` field and see that the warning UI is the same as Insight
- [ ] View the billing page for an org that has the `lockedAt` field and see that the warning UI is the same as Insight
- [ ] Usage Stats page still works as it did previously
- [ ] The warning message UI looks good

### DIFF
--- a/packages/client/modules/userDashboard/components/OrgBilling/OrgPlans.tsx
+++ b/packages/client/modules/userDashboard/components/OrgBilling/OrgPlans.tsx
@@ -14,6 +14,7 @@ import {OrgPlans_organization$key} from '../../../../__generated__/OrgPlans_orga
 import {ElementWidth, Threshold} from '../../../../types/constEnums'
 import {TierEnum} from '../../../../__generated__/SendClientSegmentEventMutation.graphql'
 import OrgStats from './OrgStats'
+import LimitExceededWarning from '../../../../components/LimitExceededWarning'
 
 const StyledPanel = styled(Panel)({
   maxWidth: ElementWidth.PANEL_WIDTH,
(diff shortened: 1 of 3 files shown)


nickoferrall:
Hey @acressall, what do you think of the UI in the screenshots in the PR description?

I've used the same UI as Usage Stats as I think it makes sense for the UI & messaging to be consistent in both places. That means there are two changes:

1. It doesn't include the cute lil bird in an egg emoji. We could add this in although the tone of the message might conflict with the emoji
2. The copy is slightly different. I like the way the Usage Stats UI specifies why the org/domain is over the Starter tier limit

acressall (approved):
Makes sense to keep the copy consistent! 👍

packages/client/modules/userDashboard/components/OrgBilling/OrgPlans.tsx
```diff
@@ -245,6 +247,7 @@ const OrgPlans = (props: Props) => {
   return (
     <StyledPanel label='Plans'>
       <StyledRow>
+        <LimitExceededWarning organizationRef={organization} />
```
igorlesnenko, line 250: -1 The warning is always shown, even if the limit is not exceeded

igorlesnenko (changes requested):

igorlesnenko (approved):

Issue #7896, opened 2023-03-09
Jira: not possible to update story points in some cases 


igorlesnenko:
In some cases, the points update looks successful, the error is not returned, but on the jira side the update does not happen

AC:

- [ ] Add some logging to find out in which cases it happens 
- [ ] Analyze logs
- [ ] Remove logging
- [ ] Fix the issue

PR #7897, opened 2023-03-09
chore: add temporary story points update result logging

igorlesnenko:
part of #7896


nickoferrall (approved):
Could add a comment to say why this console.log is here. If I stumbled across it in development, I'd probably remove it assuming it was accidentally added

PR #7908, opened 2023-03-10
feat(retros-in-disguise): Added activity library cards

### LINKED ISSUE #7833: Implement activity card designs

BartoszJarocki:
# Description

Fixes #7833 

This PR introduces a couple of new things
1. Radix scroll area to keep consistent scroll styles and prevent layout shifts
2. CSS container queries via Tailwind plugin to help with responsive grid 

## Demo

Desktop
desktop

IPad Air
ipad air

IPhone 12 pro
iphone 12 pro

IPhone SE
iphone se

## Testing scenarios

- [ ] Compare to design and check various screen sizes

## Final checklist


packages/client/components/ActivityLibrary/ActivityCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+  teamPrompt: {primary: 'bg-aqua-400', secondary: 'bg-[#CBECF0]'}
+}
+
+interface Props {
+  className?: string
+  type: MeetingTypeEnum
+  children: React.ReactNode
+}
+
+export const ActivityCard = (props: Props) => {
```
BartoszJarocki, line 23: This is non interactive activity card that could be reused on meeting dashboard later on

packages/client/components/ActivityLibrary/ActivityIllustrations.tsx
```diff
@@ -0,0 +1,115 @@
+import aChristmasCarolRetrospectiveTemplate from '../../../../static/images/illustrations/aChristmasCarolRetrospectiveTemplate.png'
```
BartoszJarocki, line 1: Extracted a file with all the illustrations to prevent copy paste and having multiple places to update while adding a new template

packages/client/components/ActivityLibrary/ActivityLibraryCard.tsx
```diff
@@ -1,21 +1,88 @@
...
-      <div className='text-lg font-bold'>{name}</div>
+    <Component
+      className={clsx(
+        'z-10 block pr-6 text-sm font-semibold leading-5 text-slate-800 focus:outline-none sm:text-base',
+        className
+      )}
+      {...rest}
+    >
+      {/* Extend touch target to entire activity card if it's a link */}
+      {'to' in props && <span className='absolute inset-0' aria-hidden='true' />}
```
BartoszJarocki, line 36: `a` tags can't contain divs so this is a way of extending the touch/link area to the whole card while still maintaining all the accessibility features

packages/client/components/ActivityLibrary/ActivityLibraryCard.tsx
```diff
@@ -1,21 +1,88 @@
...
+  type: MeetingTypeEnum
+  children: React.ReactNode
+}
+
+const Card = (props: CardProps) => {
+  const {className, type, children} = props
+
+  return (
+    <ActivityCard
+      className={clsx(
```
BartoszJarocki, line 73: Here I'm adding some interactivity to the base activity card

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
```diff
@@ -88,17 +117,35 @@ export const ActivityLibrary = (props: Props) => {
                 <div className='mt-0.5'>TODO: create custom activity</div>
               </div>
             </div>
+          ) : (
+            <div className='mx-auto grid max-w-7xl grid-cols-2 gap-4 p-4 @2xl:grid-cols-3 @4xl:grid-cols-4 @6xl:grid-cols-5'>
```
BartoszJarocki, line 121: Container queries are useful as we want to know what's the size of the parent and use that as a baseline for column count. Due to various paddings/margins using normal media queries won't work well

BartoszJarocki:
@jmtaber129 would you mind taking a look? thanks!

packages/client/components/ActivityLibrary/ActivityCard.tsx
jmtaber129: +1 Is this pulled out specifically for reuse later on, or is the core intent clean code in this PR? It seems like there's a _lot_ more that we'd want to reuse for the meeting dashboard, so if this is intended for reuse, we should probably extract almost all of the `ActivityLibraryCard` - basically everything except the badges

packages/client/components/Dashboard/DashSidebar.tsx
```diff
@@ -79,6 +79,7 @@ const DashSidebar = (props: Props) => {
         ...DashNavList_viewer
         featureFlags {
           checkoutFlow
+          retrosInDisguise
```
jmtaber129, line 82: +1 Is this used somewhere?

jmtaber129 (changes requested):
Mostly looks good, but one blocking issue with the `Link` implementation

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
```diff
@@ -70,13 +94,18 @@ export const ActivityLibrary = (props: Props) => {
   }
 
   return (
-    <div className='flex h-full bg-white'>
-      <ActivityLibrarySideBar />
-      <div>
+    <div className='flex h-full w-full flex-col'>
```
jmtaber129, line 97: Was removing `bg-white` intentional here? White background is what's in the designs.

mattkrick (changes requested):

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
BartoszJarocki: not intentional, I was playing with some stuff and forgot to add it, thanks for catching it!

packages/client/components/Dashboard/DashSidebar.tsx
BartoszJarocki: nope, removed

packages/client/components/ActivityLibrary/ActivityCard.tsx
```diff
@@ -0,0 +1,48 @@
...
+import React from 'react'
+import {MeetingTypeEnum} from '../../__generated__/ActivityLibraryQuery.graphql'
+
+interface CardTheme {
+  primary: string
+  secondary: string
+}
+
+export const MeetingThemes: Record<MeetingTypeEnum, CardTheme> = {
+  action: {primary: 'bg-aqua-400', secondary: 'bg-[#CBECF0]'},
```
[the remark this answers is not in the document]
BartoszJarocki: I pulled it all from the designs, should we/can we change the colors without asking? 😅

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
```diff
@@ -47,13 +54,30 @@ const getTemplateValue = (template: {name: string}) => template.name
 export const ActivityLibrary = (props: Props) => {
   const {queryRef} = props
   const data = usePreloadedQuery<ActivityLibraryQuery>(query, queryRef)
+  const {history} = useRouter()
   const {viewer} = data
   const {featureFlags, availableTemplates} = viewer
 
+  const handleCloseCLick = () => {
```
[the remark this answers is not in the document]
BartoszJarocki: oh, good catch!

packages/client/components/ActivityLibrary/ActivityLibraryCard.tsx
```diff
@@ -1,21 +1,88 @@
...
+        'group transition-shadow focus-within:ring-2 focus-within:ring-primary hover:shadow-xl motion-reduce:transition-none',
```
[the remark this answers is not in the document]
BartoszJarocki: yeah, changed that to `shadow-md`. and I agree with moving towards more flat design. in this case I asked our design folks and they decided to go with elevation. I'm happy to change that anytime thought ;)

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
BartoszJarocki: in general, I agree but what I don't like about _not_ using is `max-w` is the fact it'll look bad on wide screen monitors. and yeah, there are other ways of doing this but if you have truly wide monitor you'd need to move your head from left to right to see the content 😋 

that said, you're right about using grid template columns, it's a better fit here. I need to up my grid skills! thanks for the suggestion, fixed 👍

packages/client/components/ActivityLibrary/ActivityLibraryHeader.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <div className={clsx('mx-1', className)}>
+      <div className='flex basis-[15%] items-center justify-start'>
+        <LogoBlock className='shrink-0' />
+        <div className='hidden shrink-0 pr-2 text-lg font-semibold lg:block lg:text-xl'>
+          Start Activity
+        </div>
+      </div>
+
+      <div className='flex flex-1 items-center'>
+        <div className='mx-auto w-full max-w-7xl md:px-4'>{children}</div>
```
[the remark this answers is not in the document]
BartoszJarocki: changed, even though I think it makes sense to have it due to wide screen monitors support 😛

packages/client/components/ActivityLibrary/ActivityLibraryCard.tsx
```diff
@@ -1,21 +1,88 @@
...
 
-interface Props {
-  name: string
-  type: string
-  teamName: string
+const Image = (props: PropsWithChildren<React.ImgHTMLAttributes<HTMLImageElement>>) => {
+  const {className, src} = props
+
+  return (
+    <div className='absolute inset-0 top-5 flex'>
```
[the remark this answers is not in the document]
BartoszJarocki: I changed that, it works OK but now, when image depends on the card height, we have all sorts of layout shifts. Another tension I have is it shouldn't really be a `-1`. Current solution works, just differently. Also, if you look at the designs, image isn't _below_ the title, it's just centered (see the last grid item). 

Screenshot 2023-03-29 at 18 09 47

Don't get me wrong, I like pixel perfect implementations, but we need to find a balance between what's acceptable, what are our personal preferences and the time spent, as we don't really know if this code won't be thrown away in a couple of weeks.

@mattkrick I'd appreciate if you could test the current solution (it's broken when the screen is wide and there's only one or a couple of columns as they stretch to fill the container) but at least it'd give me an idea if I'm heading into right direction. Thank you!

BartoszJarocki:
@jmtaber129 @mattkrick may I ask you to take another look? thanks!

packages/client/components/ActivityLibrary/ActivityCard.tsx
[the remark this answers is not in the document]
jmtaber129: Most of these actually match up with one of the colors in the theme (the only exception being retros)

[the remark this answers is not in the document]
BartoszJarocki: ok, no, retro has to be hardcoded, `grape-100` looks terrible

jmtaber129 (approved):
LGTM! 🚢

mattkrick:
Grids looking good!
There are 2 issue, neither blocking.
First is the cards can look a little long at a given viewport, instead of stretching past a certain point we'll probably want to just show white space to the right.
Screenshot 2023-04-03 at 4 37 27 PM


Second is the card height for all cards in view changes  based on the tallest card. This means card heights change as you click through categories, which feels funny. Perhaps adding a height so cards with single-line title are the same height as those with 2-line titles? Then we add a text overflow ellipsis to make sure it doesn't extend to 3 lines?

good to merge either way 
Screenshot 2023-04-03 at 4 36 57 PM

PR #7911, opened 2023-03-10
chore: add libvips to dist

mattkrick:
# Description

Removes dependency on a globally available VIPS

## Testing scenarios

- `yarn build --no-deps`
- `rm -rf node_modules`
- `node dist/gqlExecutor & node dist/web`
- in the app, change your profile picture to something fun (it has to be fun 👀 )


mattkrick:
keeping the merge train moving, this is limited to just building using `--no-deps`

Issue #7900, opened 2023-03-09
Can't create custom template


nickoferrall:
Reported by Grayson here: https://parabol.slack.com/archives/C4JAUUZ9P/p1678371595957819

To reproduce, click "Create custom template" and see the bug.

This seems to be caused by the recent useFragment change in `ReflectTemplateItem`. If I revert the recent change, I no longer get the bug.

### AC

- [ ] Fix the bug while still using useFragment

https://parabol.sentry.io/issues/3982303770/?referrer=github_plugin

```
TypeError: Cannot destructure property 'id' of 'p' as it is undefined.
  at Ta (./packages/client/modules/meeting/components/ReflectTemplateItem.tsx:99:14)
  at Set.forEach (<anonymous>)
  at setActiveTemplate (./packages/client/utils/relay/setActiveTemplate.ts:30:3)
  at timer.current (./packages/client/hooks/useActiveTopTemplate.ts:26:9)
...
(25 additional frame(s) were not displayed)
```

PR #7917, opened 2023-03-13
fix(templates): Fix 'Create new template' button error for retros

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/7900

See [slack](https://parabol.slack.com/archives/C4JAUUZ9P/p1678371595957819) 🔒 
Attempting to click "create new template" for retros currently throws an error due to weirdness manifesting as a result of the `createFragmentContainer` -> `useFragment` migration in `ReflectTemplateItem`.

Fix this by changing the fragments to the same structure that Poker uses, which doesn't contain this bug.

## Demo
N/A

## Testing scenarios
- [ ] Click "create new template" in the retro template modal and confirm now error is thrown

## Final checklist

### DIFF
--- a/packages/client/modules/meeting/components/ReflectTemplateList.tsx
+++ b/packages/client/modules/meeting/components/ReflectTemplateList.tsx
@@ -108,20 +108,20 @@ const ReflectTemplateList = (props: Props) => {
     graphql`
       fragment ReflectTemplateList_settings on RetrospectiveMeetingSettings {
         ...ReflectTemplateSearchBar_settings
-        ...ReflectTemplateListTeam_settings
         id
+        templateSearchQuery
         team {
           ...AddNewReflectTemplate_team
+          ...ReflectTemplateListTeam_team
           id
         }
         activeTemplate {
           ...getTemplateList_template
           id
-          teamId
-          orgId
         }
         teamTemplates {
           ...AddNewReflectTemplate_reflectTemplates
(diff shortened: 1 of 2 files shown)


mattkrick:
any idea on why this pattern works & the other one doesn't?

mattkrick:
Notes in case we revisit this...

Confirmed that this is still an issue with Relay v14.1.0.
I like this fix, it's the most minimally invasive fix that doesn't sacrifice UX, but it does sacrifice DX because we have to be aware of this bug & work around it by breaking out arrays that contain an optimistic object. TBH this bug will probably bite me again because i'm not going to be thinking of it.
I tried `@relay(plural: true)` and `@relay(mask:false)` on the fragment spread but that didn't do anything. 
The reference to the fragment is valid. But relay can't find the fragment itself. that can only mean that it's GCing asynchronously. The expected behavior is that the relay subscriber on the parent component removes the optimistic object from the array before it removes the optimistic object from its store. 


The error is

```
warning.js:30 Warning: Relay: Expected to have been able to read non-null data for fragment `ReflectTemplateItem_template` declared in `useFragment()`, since fragment reference was non-null. Make sure that that `useFragment()`'s parent isn't holding on to and/or passing a fragment reference for data that has been deleted.

ReflectTemplateItem.tsx?568e:99 Uncaught TypeError: Cannot destructure property 'id' of 'template' as it is undefined.
```

Issue #7835, opened 2023-03-01
RiD Activity Library: Implement categories


jmtaber129:
Implement category selection in activity library, including quick start.

[Designs](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?node-id=337%3A20681&t=6LcgGSoOGDTFBYJK-1)

AC:
- [ ] Category buttons match designs
- [ ] Clicking a category button shows the templates in that category
- [ ] Quick start includes a naive set of templates (possibly hard-coded for now)
- [ ] Clicking a category while a search query is entered clears the search query and shows the templates in that category.

PR #7927, opened 2023-03-15
feat(rid): Activity categories

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/7835

Add category filtering to the activity library. Categories are hard-coded for now.

## Demo
https://www.loom.com/share/a06853e17e244a43ad4731ab15eb578f

## Testing scenarios
- [ ] Click a few categories and confirm filtered templates belong to the category according to [the sheet](https://docs.google.com/spreadsheets/d/1yhsu2dWvbgp_Dbftkw_E1Ilwk7iE85iUYHj7m9NxT6I/edit#gid=1843064613) 🔒 
- [ ] Start typing a search query, and confirm activities are only filtered according to the search query and no category is selected
- [ ] Delete the search query, and confirm the most recent category is selected again
- [ ] Start typing a search query, then click a category, and confirm the search query is cleared and templates in the category are shown

## Final checklist

### DIFF
--- a/packages/client/components/ActivityLibrary/ActivityLibrary.tsx
+++ b/packages/client/components/ActivityLibrary/ActivityLibrary.tsx
@@ -1,5 +1,5 @@
 import graphql from 'babel-plugin-relay/macro'
-import React, {useMemo} from 'react'
+import React, {useMemo, useState} from 'react'
 import {PreloadedQuery, usePreloadedQuery} from 'react-relay'
 import {Redirect} from 'react-router'
 import {ActivityLibraryQuery} from '~/__generated__/ActivityLibraryQuery.graphql'
@@ -8,6 +8,7 @@ import ActivityLibraryCard from './ActivityLibraryCard'
 import SearchBar from './SearchBar'
 import useSearchFilter from '../../hooks/useSearchFilter'
 import halloweenRetrospectiveTemplate from '../../../../static/images/illustrations/halloweenRetrospectiveTemplate.png'
+import clsx from 'clsx'
 
 graphql`
   fragment ActivityLibrary_template on MeetingTemplate {
@@ -18,6 +19,8 @@ graphql`
     }
     name
     type
+    category
+    isRecommended
   }
 `
 
@@ -44,6 +47,24 @@ interface Props {
 
 const getTemplateValue = (template: {name: string}) => template.name
 
+const CATEGORY_ID_TO_NAME = {
+  recommended: 'Quick Start',
+  retro: 'Retrospective',
+  estimation: 'Estimation',
+  standup: 'Standup',
+  feedback: 'Feedback',
+  strategy: 'Strategy'
+}
+
+const CATEGORY_ID_TO_COLOR_CLASS = {
+  recommended: 'bg-grape-700',
+  retro: 'bg-grape-500',
+  estimation: 'bg-tomato-500',
+  standup: 'bg-aqua-400',
+  feedback: 'bg-jade-400',
+  strategy: 'bg-rose-500'
+}
+
 export const ActivityLibrary = (props: Props) => {
   const {queryRef} = props
   const data = usePreloadedQuery<ActivityLibraryQuery>(query, queryRef)
@@ -52,8 +73,22 @@ export const ActivityLibrary = (props: Props) => {
 
   const templates = useMemo(
     () => [
-      {id: 'action', type: 'action', name: 'Check-in', team: {name: 'Parabol'}},
-      {id: 'teamPrompt', type: 'teamPrompt', name: 'Standup', team: {name: 'Parabol'}},
+      {
+        id: 'action',
+        type: 'action',
+        name: 'Check-in',
+        team: {name: 'Parabol'},
+        category: 'standup',
+        isRecommended: true
+      },
+      {
+        id: 'teamPrompt',
+        type: 'teamPrompt',
+        name: 'Standup',
+        team: {name: 'Parabol'},
+        category: 'standup',
+        isRecommended: true
+      },
       ...availableTemplates.edges.map((edge) => edge.node)
     ],
     [availableTemplates]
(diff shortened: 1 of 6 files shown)


packages/client/components/ActivityLibrary/ActivityLibrary.tsx
```diff
@@ -18,6 +19,8 @@ graphql`
     }
     name
     type
+    category
+    isRecommended
```
jmtaber129, line 23: I'm choosing to make `isRecommended` (i.e. included in Quick Start) distinct from `category` because 1) quick start will likely require some amount of special-casing that makes sense to keep separate from other category-like types, 2) making `category` plural adds complexity now, and 3) it should be easy enough to collapse these into a single `categories` field in the future.

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
```diff
@@ -44,6 +47,28 @@ interface Props {
...
+// :TODO: (jmtaber129): Fold this into the 'MeetingThemes' to be added in
+// https://github.com/ParabolInc/parabol/pull/7908.
+const CATEGORY_ID_TO_COLOR_CLASS = {
+  recommended: 'bg-grape-700',
+  retro: 'bg-grape-500',
+  estimation: 'bg-tomato-500',
+  standup: 'bg-aqua-400',
+  feedback: 'bg-jade-400',
+  strategy: 'bg-rose-500'
+}
```
jmtaber129, line 70: This is duplicated in part in https://github.com/ParabolInc/parabol/pull/7908, so we should merge the two mappings either as part of whichever PR lands last, or after both PRs land.

packages/client/hooks/useSearchFilter.ts
```diff
@@ -3,16 +3,17 @@ import useForm from '~/hooks/useForm'
...
       getDefault: () => ''
     }
   })
+  const resetQuery = () => setValue('search', '')
   const {search} = fields
   const {value} = search
   const query = value.toLowerCase()
   const filteredItems = useFilteredItems(query, items, (item) => getValue(item).toLowerCase())
-  return {query: value, filteredItems, onQueryChange: onChange}
+  return {query: value, filteredItems, onQueryChange: onChange, resetQuery}
```
jmtaber129, line 16: We could also make this a more general `setQuery` function, but `resetQuery` seems more applicable for most cases (e.g. "X"s for clearing the search query), even as just a convenience method for a `setQuery` method we might expose in the future.

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
```diff
@@ -62,20 +101,57 @@ export const ActivityLibrary = (props: Props) => {
...
+  }
+
   return (
     <div className='flex h-full bg-white'>
       <ActivityLibrarySideBar />
       <div>
         <SearchBar searchQuery={searchQuery} onChange={onQueryChange} />
+        <div className='ml-2 flex'>
+          {(Object.keys(CATEGORY_ID_TO_NAME) as Array<CategoryID>).map((category) => (
+            <button
```
BartoszJarocki, line 139: +1 we probably want each tab to be it's own URL so we can link to it from landing pages etc. If that's the case, then it'd be better to use `a` instead of `button` here and use `nav` as a parent to be a good internet citizen.

@mattkrick any thoughts?

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
```diff
@@ -62,20 +101,57 @@ export const ActivityLibrary = (props: Props) => {
...
+              className={clsx(
+                'mr-2 cursor-pointer rounded-full py-2 px-4 text-xs font-semibold text-slate-700',
```
BartoszJarocki, line 141: +1 instead of `mr-2` it's a little bit nicer to use `gap-x-2` in the parent

BartoszJarocki (commented):
Looks good but I'd make sure what approach we want to move forward with navigation (state vs url) CC @mattkrick

jmtaber129:
@BartoszJarocki good point re: navigation! I'll modify the approach here to use the URL instead of just the internal state (cc @mattkrick )

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
```diff
@@ -74,8 +142,25 @@ export const ActivityLibrary = (props: Props) => {
       <ActivityLibrarySideBar />
       <div>
         <SearchBar searchQuery={searchQuery} onChange={onQueryChange} />
+        <div className='ml-2 flex gap-x-2'>
+          {(Object.keys(CATEGORY_ID_TO_NAME) as Array<CategoryID>).map((category) => (
+            <Link
```
BartoszJarocki, line 147: +1 there's a `NavLink` that can be styled depending if the link is active or not automatically, more [here](https://v5.reactrouter.com/web/api/NavLink). just a fyi, not a merge blocker by any means

BartoszJarocki (approved):
lgtm!

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
```diff
@@ -62,9 +105,34 @@ export const ActivityLibrary = (props: Props) => {
...
+  const history = useHistory()
+  const {match} = useRouter<{categoryId?: string}>()
+  const {params} = match
+  const {categoryId: selectedCategory} = params
+
+  useEffect(() => {
+    if (!selectedCategory || !Object.keys(CATEGORY_ID_TO_NAME).includes(selectedCategory)) {
+      history.replace(`/activity-library/category/${QUICK_START_CATEGORY_ID}`)
+    }
+  }, [selectedCategory, history])
```
BartoszJarocki, line 121: +1 one thing, as I'm not a fan of `useEffect` I'd use `<Redirect />` here 😛

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
jmtaber129: TIL! Unfortunately there seems to be an issue with our version of @types/react-router-dom that prevents us from using the `isActive` prop, and I don't want that the necessary upgrade to creep into this PR.

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
jmtaber129: Nice call! Updated

PR #7928, opened 2023-03-15
chore(upgradeRelay): add $data suffix to relay data fragments

mattkrick:
# Description

Before v12, Relay generated typescript types for data fragments using the fragmentName as the export name.
Relay v12 generates typescript types for data fragments using both the naming convention `fragment` and `fragment$data`. 
Relay v13+ only exports `fragment$data`.

This PR updates all fragment references to use the `$data` suffix.

**There is no impact to runtime in this PR, just types!**

## Demo

 N/A

## Testing scenarios

CI/CD passes


mattkrick (commented):
reviewed each file. since this is one big "Find and Replace" I don't want to waste anyone else's times going through these

PR #7929, opened 2023-03-16
chore(upgradeRelay): refactor response and variables from relay mutation types

mattkrick:
# Description

Let's say there exists a mutation called "AddTask".
Relay v12 generates types for "AddTask" and exports them. they look like this:

```ts
export type AddTaskMutationResponse {};
export type AddTaskMutationVariables {};
export type AddTaskMutation = {
    readonly response: AddTaskMutationResponse;
    readonly variables: AddTaskMutationVariables;
};
```
Relay v13 generates the following types:

```ts
export type AddTaskMutation$response {};
export type AddTaskMutation$variables {};
export type AddTaskMutation = {
    readonly response: AddTaskMutation$response;
    readonly variables: AddTaskMutation$variables;
};
```

Since both v12 & v13 export the same `AddTaskMutation` type, I changed all references from e.g. `AddTaskMutationResponse` to `AddTaskMutation['response']`. 

Why not wait until Relay v13 & rename it to `AddTaskMutation$response`? Because I want to keep that PR as small as possible 🤷 

## Demo

N/A

## Testing scenarios

This PR does not affect runtime, just types! As long as `tsc` on the CI runs, we know it works 😉

### DIFF
--- a/packages/client/components/GitHubScopingSearchFilterMenu.tsx
+++ b/packages/client/components/GitHubScopingSearchFilterMenu.tsx
@@ -8,10 +8,7 @@ import useGetRepoContributions from '../hooks/useGetRepoContributions'
 import {MenuProps} from '../hooks/useMenu'
 import SearchQueryId from '../shared/gqlIds/SearchQueryId'
 import getReposFromQueryStr from '../utils/getReposFromQueryStr'
-import {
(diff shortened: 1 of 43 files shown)


mattkrick (commented):
manually reviewed each file. when CI passes it's g2g

PR #7931, opened 2023-03-16
chore(metrics): Consolidate GA4 `sign_up` events emission

tianrunhe:
## Description
- Moved mutation `loginSAML` to SDL pattern
- Consolidated fields needed for analytics inside `UserLoginPayload` into a new fragment `GA4Frag`
- Moved GA4 `sign_up` event emission into `handleSuccessfulLogin` method
- Fixed a couple of missing arguments for SAML login

## Demo
Screenshot 2023-03-16 at 14 47 32

## Testing scenarios
- [ ] Sign up from Google
  - Sign up a new account from Google
  - Verify the `sign_up` event is emitted with `is_patient_0` property set

- [ ] Sign up using password
  - Sign up a new account user password
  - Verify the `sign_up` event is emitted with `is_patient_0` property set

- [ ] Sign up with SAML
  - Sign up a new account with SAML
  - Verify the `sign_up` event is emitted with `is_patient_0` property set

- [ ] Login
  - Login with Google, password or SAML
  - Verify no `sign_up` events are emitted

## Final checklist


Dschoordsch (approved):
I think it's working, but I only see the sign_up event occasionally when checking with the web interface "DebugView", i.e. I would delete my SSO account and create a new one, but not see the event most of the time. I can see the identify call having an effect, but I only saw that one once or twice.

Dschoordsch:
To be clear, I'm hitting https://github.com/ParabolInc/parabol/blob/2256e2bf3e9cdf125aee4f7d294c946ed670d774/packages/client/components/SAMLRedirect.tsx#L23 but cannot see the effect in the "DebugView"

packages/client/utils/getTokenFromSSO.ts
```diff
@@ -16,6 +17,19 @@ const getTokenFromSSO = (url: string) => {
...
+      const isNewUser = params.get('isNewUser') === 'true'
+      const isPatient0 = params.get('isPatient0') === 'true'
+      if (isNewUser && !error) {
+        ReactGA.event('sign_up', {
+          userId,
+          user_properties: {
+            is_patient_0: isPatient0
+          }
+        })
+      }
```
tianrunhe, line 31: I know this is kinda weird to put it here... but this appears the only place we can access the SAML redirect URL in the popup window. Maybe we can rename this method? Or make it return more data than `token` and handle the GA4 stuff in locations where this method is called, i.e., `EmailPasswordAuthForm` and `TeamInvitationSSO`?

Dschoordsch: -1 I agree. We're not deciding here that it was a successful login, but on the calling side https://github.com/ParabolInc/parabol/blob/05e7f9336c9a94b37d8e2589db51ba773a95c90a/packages/client/components/EmailPasswordAuthForm.tsx#L180 so that's a better place to send the event

Dschoordsch (changes requested):
I'll wait for the next iteration before retesting

packages/server/graphql/public/typeDefs/UserLogInPayload.graphql
```diff
@@ -7,12 +7,12 @@ type UserLogInPayload {
...
   """
-  authToken: ID
-  userId: ID
+  authToken: ID!
+  userId: ID!
   """
   if a new user is created
   """
-  isNewUser: Boolean
+  isNewUser: Boolean!
```
Dschoordsch, line 15: -1 This is not correct, we're returning null here on login error: https://github.com/ParabolInc/parabol/blob/62792bda08c4a1fb56200944569513445cc3aeae/packages/server/graphql/public/mutations/loginWithPassword.ts#L31

packages/client/utils/getTokenFromSSO.ts
```diff
@@ -18,27 +24,19 @@ const getTokenFromSSO = (url: string) => {
       if (event.origin !== window.location.origin) return
 
       const params = new URLSearchParams(popup.location.search)
-      const userId = params.get('userId')
+      const userId = params.get('userId')!
```
Dschoordsch, line 27: -1 I don't like asserting here just for analytics. If we get this wrong, we'll throw an error and something breaks

packages/client/components/EmailPasswordAuthForm.tsx
```diff
@@ -172,12 +173,13 @@ const EmailPasswordAuthForm = forwardRef((props: Props, ref: any) => {
...
     }
     submitMutation()
-    const {token, error} = await getTokenFromSSO(url)
+    const {token, error, ga4Args} = await getTokenFromSSO(url)
     if (!token) {
       onError(new Error(error || 'Error logging in'))
       return true
     }
     atmosphere.setAuthToken(token)
+    emitGA4SignUpEvent(ga4Args!)
```
Dschoordsch, line 182: +1 I don't particular like this, but I see because the token is valid, these are as well. Could you please add a comment stating this?

packages/client/components/TeamInvitationSSO.tsx
```diff
@@ -26,12 +27,13 @@ const TeamInvitationSSO = (props: Props) => {
...
       const invitationToken = localStorage.getItem(LocalStorageKey.INVITATION_TOKEN)!
       submitMutation()
-      const {token, error} = await getTokenFromSSO(ssoURL)
+      const {token, error, ga4Args} = await getTokenFromSSO(ssoURL)
       if (!token) {
         onError(new Error(error || 'Error logging in'))
         return
       }
       atmosphere.setAuthToken(token)
+      emitGA4SignUpEvent(ga4Args!)
```
Dschoordsch, line 36: +1 Could you please add a comment that this is ok because it's valid if token is valid?

Dschoordsch (changes requested):
Did not test

packages/client/utils/getTokenFromSSO.ts
```diff
@@ -1,30 +1,52 @@
...
+type SucessReturnType = {
+  token: string
+  ga4Args: GA4SignUpEventEmissionRequiredArgs
+}
+
+type ErrorReturnType = {
+  error: string
+}
+
+type ReturnType = SucessReturnType | ErrorReturnType
```
tianrunhe, line 13: Let me know if this is a better way to handle your last 2 comments, or it's overkill

Dschoordsch: Yes, looks good!

Dschoordsch (approved):
Works!

packages/client/utils/getTokenFromSSO.ts
```diff
@@ -1,30 +1,52 @@
...
+
       window.clearInterval(closeCheckerId)
       popup?.close()
       window.removeEventListener('message', handler)
-      resolve({token, error})
+
+      if (userId) {
+        resolve({token, ga4Args: {userId, isNewUser, isPatient0}})
+      } else {
+        resolve({error: 'userId is null'})
```
mattkrick, line 43: we're swallowing the useful error from the server here, I think that's why we've been getting some extra support tickets around this message

PR #7932, opened 2023-03-16
chore: Throw and catch GraphQL errors on the client

### LINKED ISSUE #7339: Misleading "Invitation required" message on meeting loading error
When a meeting fails to load due to some error (e.g. returning null for a non-null field, #7337) the app shows an "Invitation required" message. This is misleading and also delay us discovering this error.

## Acceptance criteria
* [ ] check GraphQL error message when loading a meeting fails
  * [ ] if it is invitation required, show the current invitation required dialog (might just be no error, needs checking first)
  * [ ] if there is a graphql error, show a proper error message and send to sentry

Dschoordsch:
# Description

Fixes #7339
Relay does not report graphql errors (like "Cannot return null for non-nullable field...").

## Demo

https://www.loom.com/share/dfb01e81ab47437ea68843cab0d296f1

### No regressions
- [ ] visit a team link where you're not on, see the invitation needed dialog

- [ ] visit a meeting link where you're not on, see the invitation needed dialog

- [ ] query a team you're not on from `admin/graphql` and see it still works
  ``` 
  query Viewer {
    viewer {
      team(teamId: "igME1fVwLm") {
        id
      }
    }
  }
  ```
### New behaviour
- [ ] intentionally produce a graphql error by return `null` for a [non-nullable field]( https://github.com/ParabolInc/parabol/blob/70268dc58eb91c84a01bed76ccca650057adf933/packages/server/graphql/types/PokerMeeting.ts#L37)
  - [ ] visit the meeting
  - [ ] see error popping up (on master this would be "invitation required")

## Final checklist


nickoferrall (approved):

packages/client/Atmosphere.ts
```diff
@@ -278,6 +278,22 @@ export default class Atmosphere extends Environment {
       }
     }
     const transport = uploadables ? this.baseHTTPTransport : this.transport
+    const errorCheckerSink: Sink<any> | undefined = sink
+      ? {
+          next: (value) => {
+            if (value.errors) {
+              throw new Error(value.errors[0]?.message)
```
mattkrick, line 285: -2 I'd like to work together to figure out another way to do this. A test case where this wouldn't work is a poker meeting where a user can't access a jira issue. The meeting is good. The phases are good. 99% of the data is good except the 1 jira issue that couldn't get fetched. If we throw here, then would the whole meeting object get thrown out? (I didn't test this specific use case, so I could be wrong!).

packages/client/Atmosphere.ts
```diff
@@ -304,6 +304,22 @@ export default class Atmosphere extends Environment {
...
     const transport = uploadables ? this.baseHTTPTransport : this.transport
+    const errorCheckerSink: Sink<any> | undefined = sink
+      ? {
+          next: (value) => {
+            if (value.errors) {
+              console.error(value.errors)
+            }
+            return sink.next(value)
+          },
+          error: (error: Error, isUncaughtThrownError?: boolean) =>
```
mattkrick, line 315: +1 nit, took me awhile to see that the only thing changed is the next handler. could we do something like 

```ts
const errorSink = sink ? {...sink, next: (val) => {}} : undefined
```

mattkrick (approved):
LGTM, just 1 little stylistic thing

packages/client/mutations/handlers/handleRemoveTeams.ts
```diff
@@ -5,6 +5,7 @@ import pluralizeHandler from './pluralizeHandler'
 const handleRemoveTeam = (teamId: string, store: RecordSourceSelectorProxy<any>) => {
   const viewer = store.getRoot().getLinkedRecord('viewer')!
   safeRemoveNodeFromArray(teamId, viewer, 'teams')
+  viewer.setValue(false, 'canAccess', {entity: 'Team', id: teamId})
```
mattkrick, line 8: this is interesting! i figured if the team object is missing then that'd trigger a query to the server that would include `canAccess` in the query. how'd you find out that the cache wasn't getting invalidated?

Dschoordsch: It does, but not fast enough. If I remove the `&& !team` part in the checks `if (!canAccessTeam && !team)` the redirect happens before the refetch. So this is a bit redundant, but I like it better than possibly running into issues later.

PR #7933, opened 2023-03-16
feat(summaries): Bulk task export button

jmtaber129:
# Description
Add a bulk task export button to meeting summaries where tasks may have been created. This would:
* Enhance integration discoverability
* Reduce friction for moving meeting action items into a backlog (e.g. GitHub, Jira, etc.)

For simplicity, I'm calling the existing `createTaskIntegration` mutation multiple times in parallel instead of creating a new mutation or modifying the existing mutation to handle multiple tasks. We could refactor this approach into one of those two options, but it's best to use this simpler approach for now and pay down that tech debt sometime in the future.

## Demo
https://www.loom.com/share/349595bcea0646f19b5202bb4a6e5a1e

## Testing scenarios
- [ ] Smoke test pushing task to integration from task card
- [ ] Smoke test meeting summary with no tasks
- [ ] Smoke test email summary with + without tasks
- [ ] Test bulk export for one task
- [ ] Test bulk export for multiple tasks

## Final checklist

### DIFF
--- a/packages/client/hooks/useIsIntegrated.ts
+++ b/packages/client/hooks/useIsIntegrated.ts
@@ -1,0 +1,97 @@
+import graphql from 'babel-plugin-relay/macro'
+
+import {useIsIntegrated_integrations$key} from '../__generated__/useIsIntegrated_integrations.graphql'
+import {useFragment} from 'react-relay'
+
+type IntegrationLookup = {
+  hasGitHub: boolean
+  hasAtlassian: boolean
+  hasGitLab: boolean
+  hasJiraServer: boolean
+  hasAzureDevOps: boolean
+}
+
+graphql`
+  fragment useIsIntegratedJiraServerIntegration on JiraServerIntegration {
+    auth {
+      isActive
+    }
+  }
+`
+graphql`
+  fragment useIsIntegratedAtlassianIntegration on AtlassianIntegration {
+    isActive
+  }
+`
+graphql`
+  fragment useIsIntegratedGitHubIntegration on GitHubIntegration {
+    isActive
+  }
+`
+graphql`
+  fragment useIsIntegratedGitLabIntegration on GitLabIntegration {
+    auth {
+      isActive
+    }
+  }
+`
+graphql`
+  fragment useIsIntegratedAzureDevOpsIntegration on AzureDevOpsIntegration {
+    auth {
+      isActive
+    }
+  }
+`
+
+export const makePlaceholder = (integrationLookup: IntegrationLookup) => {
+  const {hasGitHub, hasAtlassian, hasGitLab, hasAzureDevOps} = integrationLookup
+  const names = [] as string[]
+  if (hasGitHub) names.push('GitHub')
+  if (hasAtlassian) names.push('Jira')
+  if (hasGitLab) names.push('GitLab')
+  if (hasAzureDevOps) names.push('Azure DevOps')
+  return `Search ${names.join(' & ')}`
+}
+
+export const useIsIntegrated = (integrationsRef?: useIsIntegrated_integrations$key) => {
+  const integrations = useFragment(
+    graphql`
+      fragment useIsIntegrated_integrations on TeamMemberIntegrations {
+        jiraServer {
+          ...useIsIntegratedJiraServerIntegration @relay(mask: false)
+        }
+        atlassian {
+          ...useIsIntegratedAtlassianIntegration @relay(mask: false)
+        }
+        github {
+          ...useIsIntegratedGitHubIntegration @relay(mask: false)
+        }
+        gitlab {
+          ...useIsIntegratedGitLabIntegration @relay(mask: false)
+        }
+        azureDevOps {
+          ...useIsIntegratedAzureDevOpsIntegration @relay(mask: false)
+        }
+      }
+    `,
+    integrationsRef ?? null
+  )
+  if (!integrations) {
+    return null
+  }
+  const {atlassian, github, jiraServer, gitlab, azureDevOps} = integrations
+  const hasAtlassian = atlassian?.isActive ?? false
+  const hasGitHub = github?.isActive ?? false
+  const hasGitLab = gitlab?.auth?.isActive ?? false
+  const hasJiraServer = jiraServer?.auth?.isActive ?? false
+  const hasAzureDevOps = azureDevOps?.auth?.isActive ?? false
+  return hasAtlassian || hasGitHub || hasJiraServer || hasGitLab || hasAzureDevOps
+    ? {
+        hasAtlassian,
+        hasGitHub,
+        hasJiraServer,
+        hasGitLab,
+        hasAzureDevOps
+      }
+    : null
+}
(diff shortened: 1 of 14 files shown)


packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/ExportAllTasks.tsx
```diff
@@ -0,0 +1,188 @@
...
+              const error = getOnCompletedError(res, errors)
+              if (error) {
+                reject(error)
+              } else {
+                SendClientSegmentEventMutation(atmosphere, 'Bulk Tasks Published', {
+                  teamId,
+                  meetingId,
+                  meetingType,
+                  service: integrationProviderService
+                })
```
jmtaber129, line 126: @tianrunhe these are the metrics changes relevant for your review

tianrunhe: I don't see much value of having this metric sent from client side. We already have the metric [`Task Published` emitted](https://github.com/ParabolInc/parabol/blob/e24c9bea36fd59920408b043142227b3dc40a6fd/packages/server/utils/analytics/analytics.ts#L352), just not from the `createTaskIntegration` mutation. If it's not too much extra work, would you please add it there instead?

jmtaber129: We want to specifically track when a bulk export finishes, which is where this is called. Could you elaborate on where you're thinking this should go?

tianrunhe: Took a second look and test locally. Actually this metric right now is kinda duplicate of what's already been emitted in CreateTaskIntegrationMutation [here](https://github.com/ParabolInc/parabol/blob/e24c9bea36fd59920408b043142227b3dc40a6fd/packages/client/mutations/CreateTaskIntegrationMutation.ts#L252).
I do think there's some value tracking that the button was clicked and user batch published some tasks from this particular location, so maybe after the loop we emit 1 overall metric. The name and attributes can stay the same.

jmtaber129: Ah yea you're right, I meant for this to be called once at the end but this was just in the wrong place. I moved this to the callback for when all mutations complete, rather than each time a mutation completes.

tianrunhe (approved):
Looks good from metrics side!

packages/client/hooks/useIsIntegrated.ts
```diff
@@ -0,0 +1,97 @@
...
+`
+
+export const makePlaceholder = (integrationLookup: IntegrationLookup) => {
+  const {hasGitHub, hasAtlassian, hasGitLab, hasAzureDevOps} = integrationLookup
+  const names = [] as string[]
+  if (hasGitHub) names.push('GitHub')
+  if (hasAtlassian) names.push('Jira')
+  if (hasGitLab) names.push('GitLab')
+  if (hasAzureDevOps) names.push('Azure DevOps')
+  return `Search ${names.join(' & ')}`
```
BartoszJarocki, line 53: +1 consider using `Intl.ListFormat` for such things, [an example here](https://github.com/ParabolInc/parabol/blob/34080336e5be425eae060bfa7a916d333fff47bb/packages/client/components/ReflectionCard/EmojiUsersReaction.tsx#L26)

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/ExportAllTasks.tsx
```diff
@@ -0,0 +1,188 @@
...
+    )
+  )
+
+  const handlePushToIntegration = (
+    integrationRepoId: string,
+    integrationProviderService: Exclude<TaskServiceEnum, 'PARABOL'>,
+    integrationLabel?: string
+  ) => {
+    submitMutation()
+    Promise.allSettled(
```
BartoszJarocki, line 106: clarifying questions before I continue with the review
1. why client side? this should be done on server, unless I'm missing something
2. if this has to be done on client side, what about hitting rate limiter?

packages/client/hooks/useIsIntegrated.ts
jmtaber129: This is actually just copy pasted as part of the refactor, so I'd prefer to leave this as-is for now

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/ExportAllTasks.tsx
jmtaber129: 1. You're right; this should be moved to the server in the long term. However, leaving this on the frontend seems like a reasonable 80/20 to avoid a large refactor of the `createTaskIntegration` mutation right now (either to modify that mutation to accept multiple task IDs or refactoring the code to be reused by a new bulk mutation). 
2. Looking at our server-side code, this mutation doesn't seem to be protected by a ratelimiter. There's a potential for hitting rate limiters in downstream integrations, but this is an edge case I'm fine leaving unaddressed for now. The code is structured such that if a few tasks fail to push due to e.g. a downstream rate-limiting error, the sync could be retried easily.

Overall, since this is a semi-experimental new feature, I'd prefer to put this into the wild and see that it gets some usage before investing additional effort into making this more robust. If we find users are using this and getting value from it, but it has stability issues, that's when I'd want to invest effort into the server-side refactor. But if this is unstable and gets no usage I'd rather just tear it out than make it robust.

BartoszJarocki (approved):

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/ExportAllTasks.tsx
mattkrick: Jira's rate limiter for reading issues surprised me with how low it is. It's lower than they say in their docs & it's the reason we have to store all jira issues in redis-- we would often hit it during poker meetings, and that was for _reads_!

I don't see a rate limit for creating issues (https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-issues/#api-rest-api-3-issue-post) but they do have a bulk-create API, which leads me to believe one exists.

can you run a stress test & see how many issues it can push before it breaks?

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/ExportAllTasks.tsx
```diff
@@ -0,0 +1,188 @@
...
+              if (error) {
+                reject(error)
+              } else {
+                resolve()
+              }
+            }
+          })
+        })
+      })
+    ).then((results) => {
```
mattkrick, line 127: +1 i haven't sene one of these in awhile! why not async/await?

jmtaber129: > why not async/await?

Because I was mentally in promise-mode when I wrote this 😅 Switched this to use async/await for the result of `Promise.allSettled`

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/ExportAllTasks.tsx
jmtaber129: I was able to sync 20 tasks without issue, but not 22 tasks. However, the user can simply retry the sync and any unsynced tasks will sync correctly.

I also realized when stress testing that this actually just fails silently, so I also added a snackbar when this fails. There's probably a better failure case UX, but I think this is fine for now:
Screen Shot 2023-05-19 at 2 49 00 PM

PR #7937, opened 2023-03-17
fix: fix non-enterprise usage stats, use teams limit algorithm

### LINKED ISSUE #7831: Teams limit: Use a single method for counting active teams everywhere
Currently, we count active teams differently in usage stats and teams limit

* For usage stats we count teams with 2 team members.
* For teams limit, team may have 2 team members or more, but it is important that these 2 members are actually attended the meetings, not just added to the team and login.
* For usage stats, org should have at leasе one fresh meeting
* For teams limit team must have 3 meetings (with more than 1 attendee)

We need to use a single algorithm everywhere

Context: https://github.com/ParabolInc/parabol/pull/7799#discussion_r1121250099

### LINKED ISSUE #7916: Usage stats shows incorrect count of active teams
Based on data reported [here](https://parabol.slack.com/archives/C042E8PGPD4/p1678718317634859), we have one organization that has only 3 active teams, while the usage stats shows 8 active teams

AC:

- Determine the root cause
- Show the correct number of active teams

igorlesnenko:
fixes #7916 
fixes #7831

### DIFF
--- a/packages/server/graphql/public/types/Company.ts
+++ b/packages/server/graphql/public/types/Company.ts
@@ -1,27 +1,55 @@
 import getRethink from '../../../database/rethinkDriver'
 import {RDatum, RValue} from '../../../database/stricterR'
 import TeamMember from '../../../database/types/TeamMember'
-import {getUserId, isSuperUser} from '../../../utils/authorization'
+import {getUserId} from '../../../utils/authorization'
 import errorFilter from '../../errorFilter'
 import isValid from '../../isValid'
 import {CompanyResolvers} from '../resolverTypes'
 import getActiveTeamCountByOrgIds from './helpers/getActiveTeamCountByOrgIds'
 import {getTeamsByOrgIds} from './helpers/getTeamsByOrgIds'
+import {DataLoaderWorker} from '../../graphql'
+import AuthToken from '../../../database/types/AuthToken'
 
 export type CompanySource = {id: string}
 
+const getSuggestedTierOrganizations = async (
+  domain: string,
+  authToken: AuthToken,
+  dataLoader: DataLoaderWorker
+) => {
+  const organizations = await dataLoader.get('organizationsByActiveDomain').load(domain)
+  const orgIds = organizations.map(({id}) => id)
+  const viewerId = getUserId(authToken)
+  const allOrganizationUsers = (
+    await Promise.all(
+      orgIds.map((orgId) => {
+        return dataLoader.get('organizationUsersByUserIdOrgId').load({orgId, userId: viewerId})
+      })
+    )
+  ).filter(isValid)
+  // If suggestedTier === enterprise, that means the user is allowed to see across
+  // all organizations, even the ones they are not a member of!
+  const isViewerAllowedToSeeAll = allOrganizationUsers.some(
+    ({suggestedTier}) => suggestedTier === 'enterprise'
+  )
+  if (isViewerAllowedToSeeAll) return organizations
+  // Pro-qualified or unqualified users can only see orgs that they are apart of
+  const allowedOrgIds = allOrganizationUsers.map(({orgId}) => orgId)
+  return organizations.filter((organization) => allowedOrgIds.includes(organization.id))
+}
+
 const Company: CompanyResolvers = {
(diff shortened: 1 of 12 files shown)


packages/server/graphql/public/types/Company.ts
```diff
@@ -174,43 +202,7 @@ const Company: CompanyResolvers = {
...
-    const isViewerAllowedToSeeAll = allOrganizationUsers.some(
-      ({suggestedTier}) => suggestedTier === 'enterprise'
-    )
-    if (isViewerAllowedToSeeAll) return organizations
-    // Pro-qualified or unqualified users can only see orgs that they are apart of
-    const allowedOrgIds = allOrganizationUsers.map(({orgId}) => orgId)
-    return organizations.filter((organization) => allowedOrgIds.includes(organization.id))
-  },
-
-  viewerOrganizations: async ({id: domain}, _args, {authToken, dataLoader}) => {
```
igorlesnenko, line 201: I removed `viewerOrganizations` and removed `if (isSuperUser(authToken)) return organizations` check from `organizations`

`viewerOrganizations` seems to be a "workaround" to avoid this check. But do we need this check at all? It feels confusing as it breaks expected logic for super users.

If we need this later I think its better to create some admin tools or `allDomainOrganizations` prop

packages/server/graphql/public/types/helpers/getActiveTeamCountByOrgIds.ts
```diff
@@ -1,76 +1,15 @@
...
-import {InputMaybe} from '../../resolverTypes'
-import {getTeamsByOrgIds} from './getTeamsByOrgIds'
+import getTeamIdsByOrgIds from '../../../../postgres/queries/getTeamIdsByOrgIds'
+import getActiveTeamCountByTeamIds from './getActiveTeamCountByTeamIds'
 
-const THIRTY_DAYS = 1000 * 60 * 60 * 24 * 30
+// Active team is the team that completed 3 meetings with more than 1 attendee
+// and have had at least 1 meeting in the last 30 days
+// Warning: the query is very expensive
+// TODO: we need to store all calculations in the database, e.g. attendeeCount
```
igorlesnenko, line 7: It is easier to store many of needed calculations in the database. We have very expensive queries otherwie.

packages/server/graphql/public/types/helpers/getActiveTeamCountByTeamIds.ts
```diff
@@ -0,0 +1,59 @@
+import getRethink from '../../../../database/rethinkDriver'
+import {RDatum, RValue} from '../../../../database/stricterR'
+import {Threshold} from '~/types/constEnums'
+// Uncomment for easier testing
+//import { ThresholdTest as Threshold } from "~/types/constEnums";
+
+// Active team is the team that completed 3 meetings with more than 1 attendee
+// and have had at least 1 meeting in the last 30 days
+// Warning: the query is very expensive
+// TODO: store all calculations in the database, e.g. meeting.attendeeCount
```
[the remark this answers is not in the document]
igorlesnenko: Issue created: https://github.com/ParabolInc/parabol/issues/7975

Dschoordsch (changes requested):
Looks good! Did not test yet.
@tghanken do you run any GraphQL queries? Could you be affected when we remove some of the fields?

tghanken:
> @tghanken do you run any GraphQL queries? Could you be affected when we remove some of the fields?

None of our core processes are using GraphQL today, so this should be safe to merge from a data perspective.

Dschoordsch (changes requested):
To me it seems #7831 is not completely solved. I see 1 active team, but supposedly I'm over the 2 free teams limit?
Screenshot from 2023-04-03 12-35-17

igorlesnenko:
@Dschoordsch The message you see is based on `tierLimitExceededAt` flag, rather than on a real time data. 
Is it possible that you have tested teams limits before based on test configuration and the `tierLimitExceededAt` flag has remained on since then?

If this is the case, If you remove the flag manually from the DB, run and complete a meeting, the warning should not appear anymore

Dschoordsch (approved):
Nevermind, it's because of the testing conditions: being on starter and having enterprise suggested.

PR #7950, opened 2023-03-28
fix: hanging recurrence

### LINKED ISSUE #7951: Processing recurrence hangs the application
Two issues primarily cause the problem:
1. Not validating user input correctly (ie. we should not accept rrule with interval set to non number value)
2. Bug in `RRule` library that's causing infinite loop when interval is `NaN` - https://github.com/jakubroztocil/rrule/issues/321 

- [ ] validate user input 
- [ ] fix existing db rows that have interval as `NaN`

BartoszJarocki:
# Description

Fixes #7951 

## Testing scenarios

- [ ] Insert recurrence rule with interval set to `NaN` and run the `processRecurrence` mutation, should _not_ timeout

## Final checklist

### DIFF
--- a/packages/server/graphql/private/mutations/processRecurrence.ts
+++ b/packages/server/graphql/private/mutations/processRecurrence.ts
@@ -101,6 +101,11 @@ const processRecurrence: MutationResolvers['processRecurrence'] = async (_source
       // For meetings that should still be active, start the meeting and set its end time.
       // Any subscriptions are handled by the shared meeting start code
       const rrule = RRule.fromString(meetingSeries.recurrenceRule)
+      // technically, RRULE should never return NaN here but there's a bug in the library
+      // https://github.com/jakubroztocil/rrule/issues/321
+      if (isNaN(rrule.options.interval)) {
+        return
+      }
 
       // Only get meetings that should currently be active, i.e. meetings that should have started
       // within the last 24 hours, started after the last meeting in the series, and started before
--- a/packages/server/graphql/private/mutations/processRecurrence.ts
+++ b/packages/server/graphql/private/mutations/processRecurrence.ts
@@ -65,9 +65,9 @@ const processRecurrence: MutationResolvers['processRecurrence'] = async (_source
     .run()) as MeetingTeamPrompt[]
 
   const res = await Promise.all(
-    teamPromptMeetingsToEnd.map(async (meeting) => {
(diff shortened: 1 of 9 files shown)


packages/server/graphql/public/types/RRule.ts
```diff
@@ -5,25 +5,31 @@ import {RRuleScalarConfig} from '../resolverTypes'
 const isRRuleValid = (rrule: RRule) => {
   if (rrule.options.freq !== Frequency.WEEKLY) {
     return {
-      error: 'Query error: Weekly is the only supported frequency'
+      error: 'RRule frequency must be WEEKLY'
+    }
+  }
+
+  if (isNaN(rrule.options.interval)) {
```
BartoszJarocki, line 12: added validation so that won't happen again, ie user will see an error when input is wrong

packages/server/graphql/private/mutations/processRecurrence.ts
```diff
@@ -101,6 +101,11 @@ const processRecurrence: MutationResolvers['processRecurrence'] = async (_source
       // For meetings that should still be active, start the meeting and set its end time.
       // Any subscriptions are handled by the shared meeting start code
       const rrule = RRule.fromString(meetingSeries.recurrenceRule)
+      // technically, RRULE should never return NaN here but there's a bug in the library
+      // https://github.com/jakubroztocil/rrule/issues/321
+      if (isNaN(rrule.options.interval)) {
+        return
+      }
```
BartoszJarocki, line 108: we should _not_ need that but I'll sleep better

Dschoordsch (approved):
I cannot reproduce the underlying issue via the UI, do you know how this happened?

Dschoordsch:
It would be nice to handle this case better on the UI side instead of just showing server error
Peek 2023-03-28 14-21

BartoszJarocki:
> It would be nice to handle this case better on the UI side instead of just showing server error Peek 2023-03-28 14-21 [ Peek 2023-03-28 14-21 ](https://user-images.githubusercontent.com/7331043/228234590-48aed6df-11bd-460c-af67-01530a4f6aaf.gif) [ ](https://user-images.githubusercontent.com/7331043/228234590-48aed6df-11bd-460c-af67-01530a4f6aaf.gif)

@Dschoordsch for sure, but I'd rather keep the hotfix as small as possible

Issue #7976, opened 2023-03-29
Org members UI bug


nickoferrall:
![image](https://user-images.githubusercontent.com/39854876/228498412-8f4fbeb0-f21d-4c2a-b1bc-f5345833e081.png)

Bug reported by @enriquesanchez & @lorena-cure [here](https://parabol.slack.com/archives/C836NA350/p1680042157006259) 🔒 

### AC

- [ ] Org members panel UI looks reasonable

PR #7978, opened 2023-03-29
fix: org members panel UI

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7976

### To test

- [ ] Check the org members page with the checkoutFlow feature flag and see that the UI looks as expected
- [ ] Check the org members page without the checkoutFlow feature flag and see that the panel no longer stretches out to the right


igorlesnenko (approved):
Works!

Issue #7696, opened 2023-01-30
Checkout Flow: implement right sidebar


nickoferrall:
<img width="278" alt="Screenshot 2023-01-30 at 15 48 15" src="https://user-images.githubusercontent.com/39854876/215602810-e7f623b0-0713-4d43-91de-245e3ee00bff.png">

### AC

- [ ] Can click on the Plan Details icon and open a right sidebar
- [ ] Can click to close the right sidebar
- [ ] Right sidebar shows information about the org's current plan
- [ ] Right sidebar includes info for Parabol Pros

acressall:
A few extra notes:

- The sidebar is _closed_ by default
- After upgrade, the sidebar opens automatically, like a confirmation

Copy: https://docs.google.com/document/d/1BrYlpXi1XGSuUOIyTWHpr5yzQFNMgoOcF-Yq5yIW0zs/edit?usp=sharing

nickoferrall:
Thanks, that's helpful!

nickoferrall:
Hey @acressall, what should the content say if the team is on the Starter plan? [This link](https://docs.google.com/document/d/1BrYlpXi1XGSuUOIyTWHpr5yzQFNMgoOcF-Yq5yIW0zs/edit?usp=sharing) includes copy for Team & Enterprise, and the screenshot says Starter, but I think it's the info for the Team tier.

acressall:
@avivapinchas do you have time to whip something up for the starter copy, or I can ask Gareth?

avivapinchas:
Sure thing! It'll be Monday, since I am off tomorrow

nickoferrall:
Also, one question on the copy - we say: `In addition to the Parabol features you’re used to, you now have access to: `, but the sidebar will exist even if the user upgraded years ago. 

Perhaps we should say: "In addition to the features on the Starter (or Team if they're now on Enterprise) tier, you have access to:`?

avivapinchas:
Copy added in the doc! 

I decided to focus on: 
- Benefits: all the meeting types (since that's 'new' in theory) 
- Actions: copied what we highlight in the onboarding email flow
- Resources: added some resources that focus on new teams or new facilitators. My thinking is that most folks could see this at some point, so I wanted it to be different from Team & Enterprise 

I like your suggestion, @nickoferrall! 

LMK if you need anything else from me

PR #7986, opened 2023-03-30
feat(checkout-flow): implement right drawer

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7696

Screenshots of the three different tiers:

Screenshot 2023-04-06 at 15 23 34

Screenshot 2023-04-06 at 15 23 15

Screenshot 2023-04-04 at 15 17 43

### To Test

- [ ] Make sure the right drawer looks good on all three tiers
- [ ] It looks good on mobile
- [ ] All of the links send the user to the correct URL

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,6 @@
     "mini-css-extract-plugin": "^2.7.2",
     "minimist": "^1.2.5",
     "node-loader": "^2.0.0",
-    "nx": "14.7.5",
     "pg-promise": "^11.2.0",
     "postcss": "^8.4.21",
     "postcss-loader": "^7.0.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4590,13 +4590,6 @@
     read-package-json-fast "^2.0.3"
     which "^2.0.2"
 
-"@nrwl/cli@14.7.5":
-  version "14.7.5"
-  resolved "https://registry.yarnpkg.com/@nrwl/cli/-/cli-14.7.5.tgz#35315e55388de792b7c0a4d4513ba92351a19774"
-  integrity sha512-hkkavBDHPZKuxG9q8bcib9/TYnTn13t8CaePjx1JvYqWTYblWVLrzlPhJKFC44Dkch+rtvZ/USs5Fih76se25g==
-  dependencies:
-    nx "14.7.5"
-
 "@nrwl/cli@15.6.3":
   version "15.6.3"
   resolved "https://registry.yarnpkg.com/@nrwl/cli/-/cli-15.6.3.tgz#999531d6efb30afc39373bdcbd7e78254a3a3fd3"
@@ -4614,13 +4607,6 @@
     ignore "^5.0.4"
     semver "7.3.4"
     tslib "^2.3.0"
-
-"@nrwl/tao@14.7.5":
-  version "14.7.5"
-  resolved "https://registry.yarnpkg.com/@nrwl/tao/-/tao-14.7.5.tgz#9b10c97f9efc558a651c98b27c29197ecc2d5d28"
-  integrity sha512-MzfJMqVbiMitYjWXaL5/7dDKw1hDG7acciGeu5SyUX8J2J0ymKzXhqjshPvn/Ga1E9QtnMckd6aKmLlvochVag==
-  dependencies:
-    nx "14.7.5"
 
 "@nrwl/tao@15.6.3":
   version "15.6.3"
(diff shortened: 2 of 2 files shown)


acressall (approved):

packages/client/modules/userDashboard/components/OrgBilling/OrgPlanDrawer.tsx
```diff
@@ -0,0 +1,118 @@
...
+  )
+  const {id: orgId, tier, showDrawer} = organization
+  const atmosphere = useAtmosphere()
+  const isDesktop = useBreakpoint(Breakpoint.SIDEBAR_LEFT)
+
+  const toggleSidebar = () => {
+    commitLocalUpdate(atmosphere, (store) => {
+      const org = store.get(orgId)
+      if (!org) return
+      const showDrawer = org.getValue('showDrawer')
```
igorlesnenko, line 93: What is the reason for storing this in store rather than state? Feels like adds extra complexity

packages/client/modules/userDashboard/components/OrgBilling/OrgPlanDrawerContent.tsx
```diff
@@ -0,0 +1,255 @@
...
+})
+
+const UL = styled('ul')({
+  margin: 0
+})
+
+const LI = styled('li')<{isBlue?: boolean}>(({isBlue}) => ({
+  fontSize: 16,
+  lineHeight: '28px',
+  color: isBlue ? PALETTE.SKY_500 : PALETTE.SLATE_900,
```
igorlesnenko, line 43: image

Sometimes it is blue for links, sometimes it black. Do we really need different colors?

packages/client/modules/userDashboard/components/OrgBilling/OrgPlanDrawerContent.tsx
```diff
@@ -0,0 +1,255 @@
...
+                  {resource.title}
+                </Link>
+              </LI>
+            ))}
+          </UL>
+        </List>
+        <List>
+          <Subtitle>{'Other resources for effective agile teams: '}</Subtitle>
+          <UL>
+            {additionalStarterRewsources.map((resource) => (
```
igorlesnenko, line 179: +1 small typo: `additionalStarterRewsources`

igorlesnenko (approved):
Is there a reason to scroll each list individually?

image

"Plan details" button works like toggle, would be nice if it could change its state

image

Space between the link lines is not clickable, would be nice if it could be clickable too


https://user-images.githubusercontent.com/466991/230388683-8d380218-3bc8-45ed-ac60-66e485cad712.mov

packages/client/modules/userDashboard/components/OrgBilling/OrgPlanDrawerContent.tsx
nickoferrall: The designs have bullets in black when there's just text and blue when they're links: 

image

The newer content, like the one in your screenshot, has a mix of regular text and links, so I went with black bullets. I think you're probably right that we should be consistent, so I'll just make them all black. 

@acressall let me know if you have any objections 🙏 Here's the update:

Screenshot 2023-04-06 at 15 43 55

packages/client/modules/userDashboard/components/OrgBilling/OrgPlanDrawerContent.tsx
nickoferrall: Thanks, fixed!

packages/client/modules/userDashboard/components/OrgBilling/OrgPlanDrawer.tsx
nickoferrall: Keeping it in the client schema means we can avoid prop drilling, i.e. passing the `showDrawer` state around a bunch of components. 

It's currently being used in the OrgPlanDrawer and OrgPlansAndBillingHeading components and could be used in more components in the future. If we change it to state, I'd need to pass it through a few components. This is following the same pattern that we have for a few other drawers in the app.

packages/client/modules/userDashboard/components/OrgBilling/OrgPlanDrawerContent.tsx
acressall: I like this update! Thanks Igor for catching it and Nick for making the change 🎉

Issue #7960, opened 2023-03-28
RiD Details View: Retro Template Details (no sidebar)


jmtaber129:
Implement the template portion of RiD Details View for Retros (i.e. no sidebar). For v0, reuse existing retro template component(s)

[Figma](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?node-id=310%3A23597&t=fROVwP1DKIkIBZJX-1)

AC:
- [ ] Clicking a retro activity in the activity library navigates to the details view for that retro template
- [ ] Retro template is displayed using existing component(s)
- [ ] URL is unique to the activity and can be copied+shared+used on landing pages

PR #7990, opened 2023-04-03
feat(rid): Activity Details (no sidebar)

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/7960

Implements the details view for retros primarily by reusing existing components, mutations, and GQL models.

I'm explicitly leaving mobile views out of scope for now.

Note that this doesn't match designs precisely, as some feature details are still somewhat ambiguous (e.g. cloning - to which team?) and/or requires disproportionate effort (e.g. distinct edit mode, modifying categories, etc.). We should punt on those features until the next iteration.

## Demo
https://www.loom.com/share/147d92e0c39b4d4f9c16ace0904d1823

## Testing scenarios
- [ ] Navigate to details view for each of the following and confirm UI is reasonable
  - [ ] Free public template
  - [ ] Premium public template
  - [ ] Org template (from team the user is _not_ on)
  - [ ] Team template
- [ ] On a team template
  - [ ] Edit the name, and confirm validation error if name matches a different template on the team
  - [ ] Edit a prompt
  - [ ] Add a prompt
  - [ ] Change visibility
  - [ ] Delete the template
- [ ] Navigation
  - [ ] Invalid or missing template ID - navigates back to activity library
  - [ ] back navigation - goes back to previous category, or falls back to template category
- [ ] Regression on existing template view
  - [ ] Go to team, org, and public templates in the current view and confirm no UI regressions are present.

## Final checklist

### DIFF
--- a/packages/client/components/ActivityLibrary/ActivityCard.tsx
+++ b/packages/client/components/ActivityLibrary/ActivityCard.tsx
@@ -28,23 +28,9 @@ interface CardTheme {
   secondary: string
 }
 
-export const QUICK_START_CATEGORY_ID = 'recommended'
+export type CategoryID = 'retrospective' | 'estimation' | 'standup' | 'feedback' | 'strategy'
 
-export const CATEGORY_ID_TO_NAME = {
-  [QUICK_START_CATEGORY_ID]: 'Quick Start',
-  retrospective: 'Retrospective',
-  estimation: 'Estimation',
-  standup: 'Standup',
-  feedback: 'Feedback',
-  strategy: 'Strategy'
-}
-
-export type CategoryID = keyof typeof CATEGORY_ID_TO_NAME
-
-export const MeetingThemes: Record<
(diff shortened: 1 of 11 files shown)


BartoszJarocki:
@jmtaber129 what's the main reason for skipping the mobile? with tailwind it should be fairly easy to start with mobile and go for bigger screens with small tweaks to flex or grids and I believe we should try to make that a habit. I'm ok with iterating but with that kind of stuff, it probably would take ~same time to do it with mobile from the beginning

jmtaber129:
@BartoszJarocki Very few of our current users open the New Meeting view on mobile (<2%, see https://analytics.amplitude.com/parabol/chart/t29besq?source=copy+url 🔒 ), so focusing on desktop seems like a reasonable way to minimize the scope of the initial iteration of RiD

BartoszJarocki (approved):

PR #7993, opened 2023-04-04
chore: remove nx from package

Dschoordsch:
# Description

NX is a dependency lerna and adding it manually causes problems when adding new packages. Because we're not running lerna anymore, remove it.

## Testing scenarios

- [ ] run `yarn lint` to verify commands using nx still work
- [ ] try `yarn workspace parabol-server add graphql` to verify installing packages works

## Final checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,6 @@
     "mini-css-extract-plugin": "^2.7.2",
     "minimist": "^1.2.5",
     "node-loader": "^2.0.0",
-    "nx": "14.7.5",
     "pg-promise": "^11.2.0",
     "postcss": "^8.4.21",
     "postcss-loader": "^7.0.2",
--- a/yarn.lock
+++ b/yarn.lock
@@ -4590,13 +4590,6 @@
     read-package-json-fast "^2.0.3"
     which "^2.0.2"
 
-"@nrwl/cli@14.7.5":
-  version "14.7.5"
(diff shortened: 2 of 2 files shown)


BartoszJarocki:
@Dschoordsch aren't we using `lerna` to bump the version numbers during the release [here](https://www.notion.so/parabol/Release-Playbook-c3d1b146ea7c4f2fbf9501ae578ced25?pvs=4#5c95a16c868149a58e1f9221d6f86b4d)?

Dschoordsch:
You're right, I missed this external dependency. Then let's do it the other way round

BartoszJarocki:
@Dschoordsch it's gonna be a little bit harder then and I'm not sure if we want to do it. Eventually, I think the idea was to move from lerna to NX. 

now, if you look at our configs, NX is responsible for pretty much all the things in the monorepo, see the [package.json](https://github.com/ParabolInc/parabol/blob/183f543b74b49f40d19eb1829b79e0c8d2a119a2/package.json#L39). 

if you'd like to remove nx, then we need to revert a whole bunch more.

maybe @mattkrick knows some easy way to fix the issue?

Dschoordsch:
@BartoszJarocki We're good. Lerna is built on top of nx and supported by the same company now: https://nx.dev/recipes/adopting-nx/lerna-and-nx
So I think it's fine to just remove the manual nx dependency although not being the cleanest maybe.

BartoszJarocki (approved):
Now I'm even more confused as I thought lerna is separate, older and much less capable tool to manage monorepos. That said, I tested and it all works, but I have no idea why ;)

PR #8000, opened 2023-04-04
feat: pick a fun name for the default team when user signs up

tianrunhe:
# Description
Coming from #terrrible-ideas [here](https://parabol.slack.com/archives/C03NGEJUKCN/p1679935808804259). The idea is we have a list of funny names, when user signs up, we pick one randomly and make that the name for the default team. Also add a new metric to track when user changes team name.

## Demo

Metrics: 
Screenshot 2023-04-04 at 15 50 50

## Testing scenarios

- [ ] Default team name
  - Create a new account
  - See a default team is created, with a fun name chosen randomly from the list

- [ ] `Team Name Changed` metrics
  - Create a new account
  - See a default team is created, with a fun name chosen randomly from the list
  - Change the default team name to something else
  - Go to [Segment Debugger](https://app.segment.com/parabol-jordan/sources/action_local/debugger) and verify a metric called `Team Name Changed` is populated and the `isOldNameDefault` property is `true`

## Final checklist

### DIFF
--- a/packages/client/utils/makeDefaultTeamName.ts
+++ b/packages/client/utils/makeDefaultTeamName.ts
@@ -1,0 +1,8 @@
+const defaultTeamNames = ['Bug Writers', 'Long Meeting Lovers', 'Work Procrastinators']
+
+export const makeDefaultTeamName = (teamId: string) => {
+  const seed = [...teamId].reduce((prev, cur) => prev + cur.charCodeAt(0), 0)
+  const idx = seed % defaultTeamNames.length
+  // Guaranteed not out of bounds
+  return defaultTeamNames[idx]!
+}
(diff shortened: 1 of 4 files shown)


packages/client/utils/makeDefaultTeamName.ts
```diff
@@ -0,0 +1,54 @@
...
+  'Work Procrastinators ⏱️',
+  'Eat Lunch at 11AM 🥪',
+  'Midday Nap 😴',
+  'Show Us Your Cat 🐱',
+  'Comb Your Hair for Zoom 💅',
+  'Pajama Pants🌛',
+  'Highly Caffeinated ☕',
+  'Mute Slack & Chill 😎',
+  'Friday Afternoon Meetings Should Be Illegal 🙈',
+  "MacGuyver's of Fixing Bugs 🛠️",
```
Dschoordsch, line 13: That should be MacGuyvers, shouldn't it? As it's just the plural

nickoferrall (approved):
Tested, and it LGTM!

packages/client/utils/makeDefaultTeamName.ts
tianrunhe: Fixed! 🙏

PR #7958, opened 2023-03-28
feat(ai-summary): replace GPT-3 with ChatGPT

tianrunhe:
# Description
ChatGPT or GPT-3.5 performs at a similar capacity to GPT-3 but at only 10% of the price. This PR upgrades our OpenAI models to GPT-3.5. See the [official documentation](https://platform.openai.com/docs/guides/chat/chat-vs-completions) from OpenAI on this.

I would also like to make our prompts better, using all I've learnt from ["prompt engineering" course](https://learnprompting.org/docs/intro) but let's keep this one small and simple for now.

## Demo
This is from production:

Screenshot 2023-03-28 at 18 28 19

This is from my local test, using same set of reflection text:
Screenshot 2023-03-28 at 18 28 28

## Testing scenarios

[Please list all the testing scenarios a reviewer has to check before approving the PR]

- [ ] Topic summary
  - Start a retro meeting and advance to discussion phase with more than 1 reflection for the topic
  - Observe the topic summary

- [ ] Meeting summary
  - End a retro meeting
  - Observer the meeting summary

## Final checklist

- [x] I checked the [code review guidelines](../docs/codeReview.md)
- [x] I have added [Metrics Representative](../docs/codeReview.md#metrics-representative) as reviewer(s) if my PR invovles metrics/data/analytics related changes
- [x] I have performed a self-review of my code, the same way I'd do it for any other team member
- [x] I have tested all cases I listed in the testing scenarios and I haven't found any issues or regressions
- [x] Whenever I took a non-obvious choice I added a comment explaining why I did it this way
- [x] I added the label `One Review Required` if the PR introduces only minor changes, does not contain any architectural changes or does not introduce any new patterns and I think one review is sufficient'
- [x] PR title is human readable and could be used in changelog

PR #8001, opened 2023-04-05
chore(ai-summary): revert chatgpt change

nickoferrall:
I found a couple of small issues with the AI Summary following the change to ChatGPT in https://github.com/ParabolInc/parabol/pull/7958

I think we should use ChatGPT instead of the davinci model, but we should fix the following issues first.

I'd like to revert the ChatGPT change for now as the roll-out will happen today/tomorrow.  CC @tianrunhe 

Loom demo: https://www.loom.com/share/d0685ece343241578391432c42e192f7

Issues: 

- [ ] If there is some text that the model is unable to summarise, it doesn't create a whole meeting summary, meaning the loading state remains forever. This doesn't happen with the old davinci set-up:

Screenshot 2023-04-05 at 11 01 46

- [ ] After writing nonsensical reflections, one summary said it was inappropriate

Screenshot 2023-04-05 at 10 56 36

- [ ] The following [error message](https://github.com/ParabolInc/parabol/blob/57c0211e3e45bd3e7f207ce46b19defd4c43c1ec/packages/server/graphql/mutations/endRetrospective.ts#L52) is triggered when ChatGPT is unable to send a whole meeting summary, which happened quite a lot while I was testing
- [ ] We should increase the `max_tokens` for ChatGPT to its new limit of 4096 so it can digest a largest quantity of test

### DIFF
--- a/packages/server/utils/OpenAIServerManager.ts
+++ b/packages/server/utils/OpenAIServerManager.ts
@@ -20,28 +20,21 @@ class OpenAIServerManager {
     if (!this.openAIApi) return null
     try {
       const location = summaryLocation ?? 'retro meeting'
-      const response = await this.openAIApi.createChatCompletion({
-        model: 'gpt-3.5-turbo',
-        messages: [
-          {
-            role: 'user',
-            content: `Below is a comma-separated list of text from a ${location}.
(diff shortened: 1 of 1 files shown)


Dschoordsch (approved):
Did not test as this is just a revert.

PR #8004, opened 2023-04-06
feat(checkout-flow): right drawer

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7696

Screenshots of the three different tiers:

Screenshot 2023-04-06 at 15 23 34

Screenshot 2023-04-06 at 15 23 15

Screenshot 2023-04-04 at 15 17 43

I ran into a merge issue that I've never seen before on https://github.com/ParabolInc/parabol/pull/7986, so I closed that and created this new PR. Moving this to Maintainer Review as it passed Reviewer Review.

### To Test

- [ ] Make sure the right drawer looks good on all three tiers
- [ ] It looks good on mobile
- [ ] All of the links send the user to the correct URL


packages/client/modules/userDashboard/components/OrgBilling/OrgPlanDrawer.tsx
```diff
@@ -0,0 +1,118 @@
...
+        id
+        showDrawer
+        tier
+      }
+    `,
+    organizationRef
+  )
+  const {id: orgId, tier, showDrawer} = organization
+  const atmosphere = useAtmosphere()
+  const isDesktop = useBreakpoint(Breakpoint.SIDEBAR_LEFT)
```
Dschoordsch, line 87: +1 I like reusing breakpoints, but here it seems this is not the best one. Because we disable vertical scrolling, it would seem more appropriate to me if we'd gray out everything as soon as the details would not fit anymore.
Gray at sidebar left breakpoint:
Screenshot from 2023-04-11 11-10-45

Not gray when details don't hide the `Plans`:
Screenshot from 2023-04-11 11-11-14

The edge case between where some part of the Plans is hidden, we cannot scroll vertically and it's not grayed out:
Screenshot from 2023-04-11 11-11-01

Dschoordsch (approved):

Dschoordsch:
@acressall tagging you because I had some design questions

packages/client/modules/userDashboard/components/OrgBilling/OrgPlanDrawer.tsx
nickoferrall: @acressall here's a Loom on what I currently see: https://www.loom.com/share/194140ccc25344feb4e67b13e30d2cf5

@Dschoordsch mentioned that we cannot scroll vertically when some part of the Plans is hidden, but I haven't been able to reproduce that. However, the draw can cover some of the content. Alicia, do you think we should grey out the background when the draw covers any content or keep things as they are now? You can see how it currently looks in the Loom

acressall: Thanks so much for the Loom! So the difference with the agenda panel is it's over a task board so we have a horizontal scroll behind that lets you see the rest of the content. I'm leaning towards adding the grey background as soon as it covers the content, for the sake of contrast, so anyone with a visual disability (or even just bad screen calibration) can distinguish the panel from the background content

packages/client/modules/userDashboard/components/OrgBilling/OrgPlanDrawer.tsx
```diff
@@ -0,0 +1,118 @@
...
+import {BezierCurve, Breakpoint, DiscussionThreadEnum, ZIndex} from '../../../../types/constEnums'
+import LabelHeading from '../../../../components/LabelHeading/LabelHeading'
+import {desktopSidebarShadow} from '../../../../styles/elevation'
+import useBreakpoint from '../../../../hooks/useBreakpoint'
+import PlainButton from '../../../../components/PlainButton/PlainButton'
+import useAtmosphere from '../../../../hooks/useAtmosphere'
+import {ICON_SIZE} from '../../../../styles/typographyV2'
+import OrgPlanDrawerContent from './OrgPlanDrawerContent'
+
+const DrawerHeader = styled('div')({
```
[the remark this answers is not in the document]
nickoferrall: For sure! I was going to write it in tailwind, but remembered that we have the responsive drawer styles set-up, so most of this was copied and pasted from existing drawer components

packages/client/modules/userDashboard/components/OrgBilling/OrgPlanDrawer.tsx
```diff
@@ -0,0 +1,118 @@
...
+  '@supports (height: 1svh) and (height: 1lvh)': {
+    height: isDesktop ? '100lvh' : '100svh'
+  }
+}))
+
+const CloseIcon = styled(Close)({
+  color: PALETTE.SLATE_600,
+  cursor: 'pointer',
+  '&:hover': {
+    opacity: 0.5
```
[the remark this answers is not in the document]
nickoferrall: We seem to have this close button style in quite a few other places (~7 in total), e.g. https://github.com/ParabolInc/parabol/blob/c7b72ac1a722c4f155af8f6cdf49d1042ad3fefd/packages/client/components/TeamPrompt/TeamPromptDiscussionDrawer.tsx#L64-L66. 

I think the style in your gif looks good, though. I imagine it's using a `FlatButton` instead of a `PlainButton`. Do you know what component it's from? I'd be happy to change the style here as long as it's consistent with the rest of the app

packages/client/modules/userDashboard/components/OrgBilling/OrgPlanDrawer.tsx
nickoferrall: Makes sense! I've updated the breakpoints for the drawer. Here's a Loom with the update: https://www.loom.com/share/00b2462577a14e178c4835de97388eac

What do you think @acressall?

acressall: Looks great! Thanks Nick!

Issue #7864, opened 2023-03-06
Checkout Flow: Support 3D Secure


nickoferrall:
For card payments in the EU, it's important that we support 3D secure payments. Further context here: https://github.com/ParabolInc/parabol/pull/7852#pullrequestreview-1325600146

### AC

- [ ] Support 3D Secure payments

PR #8005, opened 2023-04-06
feat(checkout-flow): handle credit card functionality

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7693

Demo: https://www.loom.com/share/66aedcba66f14d6ea1ce1dcd82217c0e

I had a look on the Stripe Dashboard to see if Igor & Georg have access, but I couldn't see any other users. I think that's because I have restricted permissions where I can't see or invite other users. So if you don't have Stripe access, please tag someone in Slack that can invite you.

I'll improve the UI after the user has upgraded in the following issue: https://github.com/ParabolInc/parabol/issues/8006

A subscription will not be created for 3D Secure payment methods. This will be handled here: https://github.com/ParabolInc/parabol/issues/7864

### To test

- [ ] Add the checkout flow feature flag to an org on the Starter tier
- [ ] Enter dummy card details. Click "Upgrade" before you've entered all the details and see the error messages
- [ ] Enter dummy credit card details correctly, click Upgrade, and see the confetti animation
- [ ] Go to the Stripe dashboard, click "Test" mode, and see that the subscription was successfully created
- [ ] Refresh the page and see that the org has now been upgraded to the Team tier and the credit card panel no longer shows up


codegen.json
```diff
@@ -40,7 +40,7 @@
           "Comment": "../../database/types/Comment#default as CommentDB",
           "Company": "./types/Company#CompanySource",
           "CreateImposterTokenPayload": "./types/CreateImposterTokenPayload#CreateImposterTokenPayloadSource",
-          "CreatePaymentIntentSuccess": "./types/CreatePaymentIntentSuccess#CreatePaymentIntentSuccessSource",
+          "CreateSetupIntentSuccess": "./types/CreateSetupIntentSuccess#CreateSetupIntentSuccessSource",
```
nickoferrall, line 43: SetupIntent is used for subscriptions whereas createPaymentIntent is used for one-off payments

packages/server/graphql/mutations/upgradeToTeamTier.ts
```diff
@@ -53,7 +62,12 @@ export default {
     const viewer = await dataLoader.get('users').load(viewerId)
     const {email} = viewer!
     try {
-      await upgradeToTeamTier(orgId, stripeToken, email, dataLoader)
+      // TODO: remove upgradeToTeamTierOld once we rollout the new checkout flow: https://github.com/ParabolInc/parabol/milestone/150
+      if (paymentMethodId) {
+        await upgradeToTeamTier(orgId, paymentMethodId, email, dataLoader)
```
nickoferrall, line 67: `upgradeToTeamTier` changed quite a bit now that we're using Stripe Elements and use the paymentMethodId rather than creating a stripe token using the credit card details the user gives us. 

I felt that the simplest move was to rename `upgradeToTeamTier` to `upgradeToTeamTierOld` and create a new `upgradeToTeamTier` helper function. We can remove the old one when we rollout the new checkout flow.

nickoferrall:
Thanks @igorlesnenko - I could reproduce the `payment_method` console error. I fixed this by attaching the payment to the customer before updating the default payment method.

igorlesnenko (approved):

packages/client/modules/userDashboard/components/OrgBilling/BillingForm.tsx
```diff
@@ -20,56 +26,74 @@ const StyledForm = styled('form')({
...
-
 const UpgradeButton = styled(PrimaryButton)<{isDisabled: boolean}>(({isDisabled}) => ({
   background: isDisabled ? PALETTE.SLATE_200 : PALETTE.SKY_500,
   color: isDisabled ? PALETTE.SLATE_600 : PALETTE.WHITE,
   boxShadow: 'none',
   marginTop: 16,
   width: '100%',
   elevation: 0,
-  '&:hover': {
+  '&:hover, &:focus': {
```
Dschoordsch, line 36: Just a side note, the "Plans" panel does not have any focus behaviour which makes it impossible to use with just the keyboard

packages/client/modules/userDashboard/components/OrgBilling/BillingForm.tsx
```diff
@@ -20,56 +26,74 @@ const StyledForm = styled('form')({
...
+    const {setupIntent, error} = await stripe.confirmSetup({
+      elements,
+      redirect: 'if_required'
+    })
     setIsLoading(false)
+    if (error) return
+    const {payment_method: paymentMethodId, status} = setupIntent
+    if (status === 'succeeded' && typeof paymentMethodId === 'string') {
+      setIsPaymentSuccessful(true)
+      const handleCompleted = () => {}
```
Dschoordsch, line 75: -1 this one is failing linting

packages/client/modules/userDashboard/components/OrgBilling/BillingForm.tsx
```diff
@@ -20,56 +26,74 @@ const StyledForm = styled('form')({
...
+    if (!stripe || !elements) return
+    setIsLoading(true)
+    const {setupIntent, error} = await stripe.confirmSetup({
+      elements,
+      redirect: 'if_required'
+    })
     setIsLoading(false)
+    if (error) return
+    const {payment_method: paymentMethodId, status} = setupIntent
+    if (status === 'succeeded' && typeof paymentMethodId === 'string') {
```
Dschoordsch, line 73: -1 What about if not? If I test with a [declining card](https://stripe.com/docs/testing#declined-payments) 4000000000000002 or with a [3D Secure card](https://stripe.com/docs/testing#authentication-and-setup) 4000002760003184 and decline, then there is no message at all

Dschoordsch (changes requested):

packages/client/modules/userDashboard/components/OrgBilling/BillingForm.tsx
nickoferrall: Thanks, updated!

By the way, 3D Secure cards won't create subscriptions in this PR. It'll be handled in https://github.com/ParabolInc/parabol/issues/7864

packages/client/modules/userDashboard/components/OrgBilling/BillingForm.tsx
nickoferrall: Yeah, that would be nice! I've created an issue to work on that: https://github.com/ParabolInc/parabol/issues/8085

Dschoordsch (approved):
Before we forget, can you please also add new testing scenarios to the release test?
Could also be done a bit later, but should be logged in an issue so we don't forget.

nickoferrall:
> can you please also add new testing scenarios to the release test?

Sure thing! I've added an issue here: https://github.com/ParabolInc/parabol/issues/8089

Issue #7962, opened 2023-03-28
RiD Details View: Retro Sidebar


jmtaber129:
Implement the sidebar for the RiD details view for retros

[Figma for sidebar](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?node-id=449-24127&t=PRHULhykN1IzuXNk-0) (TODO: confirm when these designs are finalized)

[Thread on team selection on non-public templates](https://parabol.slack.com/archives/C836NA350/p1680724905390839) (TODO: confirm when these designs are finalized)

AC:
- [ ] Sidebar matches designs
- [ ] Team selection behavior for non-public templates is reasonable
- [ ] Can start a new retro meeting

PR #8008, opened 2023-04-06
feat(rid): Retro Details Sidebar

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/7962

Implement retro details sidebar + improve responsiveness for smaller desktop screens.

The team selection dropdown case when the template is limited to team (vs org) is out of scope for this PR and will be implemented later - team selection here is reasonable for now.

Parts of the diff would benefit from `?w=1`

Needs rebase once https://github.com/ParabolInc/parabol/pull/7990 lands

## Demo
https://www.loom.com/share/f9f857bf626d42acbd560c89e8dc0511

## Testing scenarios
- [ ] Smoke test responsiveness
- [ ] Smoke test active meetings
- [ ] Team-scoped template
  - [ ] Confirm no other teams included in team dropdown
  - [ ] toggle a setting
  - [ ] Start retro, and confirm team, setting, and template
- [ ] Org-scoped template on team that user is on
  - [ ] Confirm only teams in org included in team dropdown
  - [ ] Select a non-owner team
  - [ ] Start retro, and confirm team, setting, and template
- [ ] Org-scoped template on team that user is not on
  - [ ] Confirm only teams in org included in team dropdown
  - [ ] Start retro, and confirm team, setting, and template
- [ ] Public template
  - [ ] Confirm all teams the user is on are included in team dropdown
  - [ ] Start retro, and confirm team, setting, and template
- [ ] Smoke test regressions
  - [ ] Team picker in existing flow
  - [ ] current meetings in existing flow
  - [ ] Editable template name
  - [ ] Template selection

## Final checklist

### DIFF
--- a/packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
+++ b/packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
@@ -1,0 +1,125 @@
+import React, {useState} from 'react'
+import graphql from 'babel-plugin-relay/macro'
+import {useFragment} from 'react-relay'
+import StartRetrospectiveMutation from '~/mutations/StartRetrospectiveMutation'
+import {ActivityDetailsSidebar_template$key} from '~/__generated__/ActivityDetailsSidebar_template.graphql'
+import {ActivityDetailsSidebar_teams$key} from '~/__generated__/ActivityDetailsSidebar_teams.graphql'
+import NewMeetingTeamPicker from '../NewMeetingTeamPicker'
+import {MenuPosition} from '../../hooks/useCoords'
+import sortByTier from '../../utils/sortByTier'
+import NewMeetingSettingsToggleCheckIn from '../NewMeetingSettingsToggleCheckIn'
+import NewMeetingSettingsToggleAnonymity from '../NewMeetingSettingsToggleAnonymity'
+import NewMeetingActionsCurrentMeetings from '../NewMeetingActionsCurrentMeetings'
+import FlatPrimaryButton from '../FlatPrimaryButton'
+import SelectTemplateMutation from '../../mutations/SelectTemplateMutation'
+import useAtmosphere from '../../hooks/useAtmosphere'
+import useMutationProps from '../../hooks/useMutationProps'
+import {useHistory} from 'react-router'
+
+interface Props {
+  selectedTemplateRef: ActivityDetailsSidebar_template$key
+  teamsRef: ActivityDetailsSidebar_teams$key
+}
+
+const ActivityDetailsSidebar = (props: Props) => {
+  const {selectedTemplateRef, teamsRef} = props
+  const selectedTemplate = useFragment(
+    graphql`
+      fragment ActivityDetailsSidebar_template on MeetingTemplate {
+        id
+        teamId
+        orgId
+        scope
+      }
+    `,
+    selectedTemplateRef
+  )
+
+  const teams = useFragment(
+    graphql`
+      fragment ActivityDetailsSidebar_teams on Team @relay(plural: true) {
+        id
+        lastMeetingType
+        name
+        tier
+        orgId
+        retroSettings: meetingSettings(meetingType: retrospective) {
+          ...NewMeetingSettingsToggleCheckIn_settings
+          ...NewMeetingSettingsToggleAnonymity_settings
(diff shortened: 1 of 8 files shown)


packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
```diff
@@ -0,0 +1,125 @@
...
+        orgId
+        scope
+      }
+    `,
+    selectedTemplateRef
+  )
+
+  const teams = useFragment(
+    graphql`
+      fragment ActivityDetailsSidebar_teams on Team @relay(plural: true) {
```
mattkrick, line 40: Question: are the 2 fragments because of that weird relay bug with optimistic objects getting GC'd?

packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
```diff
@@ -0,0 +1,125 @@
...
+      }
+    )
+  }
+
+  return (
+    <div className='flex w-96 flex-col border-l border-solid border-slate-300 px-4 pb-9 pt-14'>
+      <div className='mb-6 text-xl font-semibold'>Settings</div>
+
+      <div className='flex grow flex-col gap-2'>
+        {availableTeams && (
```
mattkrick, line 99: -1 `availableTeams.length > 0`? i think availaableTeams is always gonna be an array

mattkrick (changes requested):
-1 If the template title ends with the word "Copy" then when you click it & enter the details view the title is autoFocused in edit mode. Feels like this shouldn't be

packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
jmtaber129: Not at all, this is just the component API that makes the most sense here. Otherwise we'd be passing in a viewer ref + template ID and duplicating a lot of the logic in the parent `ActivityDetails` component.

jmtaber129:
> If the template title ends with the word "Copy" then when you click it & enter the details view the title is autoFocused in edit mode

This is an artifact of the existing `EditableTemplateName` component, which autofocuses when the template name suggests a new template whose name should be edited (i.e. [`*New Template` or ends with ` Copy`](https://github.com/ParabolInc/parabol/blob/9e04756874afacf72569ae98a1b91b2624411134/packages/client/modules/meeting/components/EditableTemplateName.tsx#L43)). This can be reproduced in the existing template selection modal, too.

I've largely been strictly reusing components from the existing template details view to minimize the scope of this view for the initial iteration, but this should become a non-issue when we implement the dedicated "edit mode" for RiD, so I'd strongly prefer to leave it as-is for now.

mattkrick:
Yeah, we can leave as-is for internally testing, but we'll want to make sure that doesn't creep into our GA release. How would you like to track this? An individual issue or part of a big GA checklist issue?

jmtaber129:
TBH I'm not sure we want to "fix" this - the logic in https://github.com/ParabolInc/parabol/blob/9e04756874afacf72569ae98a1b91b2624411134/packages/client/modules/meeting/components/EditableTemplateName.tsx#L43 suggests newly cloned templates should have their name updated (instead of just being "XYZ Copy" forever). I imagine we'll want to maintain that status quo when navigating to newly cloned templates, whether that's entering into a dedicated edit mode, or autofocusing the template name.

In any case, this issue is tangential to the specific changes introduced in this PR, and this issue should be resolved as part of edit mode regardless.

enriquesanchez:
This is looking great! 

Will the mobile view also be in scope within this PR?

Also, I'm not sure if you saw this, but we want to have a subtle animation when the user moves into and out of the edit mode, something like this:

https://user-images.githubusercontent.com/3276087/231293409-576c74fc-ebba-451a-b0f4-c10baa5d0ffc.mp4

Also notice how the back arrow next to "Start Activity" changes to a close icon and says "Edit Activity".

jmtaber129:
@enriquesanchez I'm calling mobile out of scope for now to minimize the scope of the initial iteration.

We also haven't implemented a dedicated edit mode yet (everything is editable if the user has perms, just like the existing template details view), but that animation would be included as part of that.

PR #8018, opened 2023-04-10
chore: refactor notifications to SDL (example PR)

mattkrick:
# Description

Maintaining our schema in both JavaScript & SDL has caused some DX problems as reported in our DX retro as well as some bugs in prod (#8016). The task is far too big to tackle all at once, but I'd like to get some good patterns in place so we can work towards buying down this debt instead of adding to it.

Most of this work is a codemod that just moves the resolve function from 1 file to another. No need to review line-by-line!

## Demo

N/A All DX


packages/server/graphql/mutations/archiveTeam.ts
```diff
@@ -10,10 +10,14 @@ import publish from '../../utils/publish'
 import segmentIo from '../../utils/segmentIo'
 import standardError from '../../utils/standardError'
 import {GQLContext} from '../graphql'
-import ArchiveTeamPayload from '../types/ArchiveTeamPayload'
 
 export default {
-  type: new GraphQLNonNull(ArchiveTeamPayload),
+  type: new GraphQLNonNull(
+    new GraphQLObjectType({
```
mattkrick, line 16: This is the only novel pattern in the whole PR.
Since `ArchiveTeamPayload` is created using the SDL, it's not available to the legacy JS objects.
To use it, we just created a stub  inline & that stub will get filled in when the SDL gets merged into schema.
This is useful for things that we're not ready to migrate yet.
For example, in this PR I wanted to migrate the Notification types, their corresponding payload types, but not the mutations themselves.

jmtaber129 (approved):

PR #8019, opened 2023-04-10
fix: update recurrenceSettings call

mattkrick:
# Description

Fun merge conflict happened because 2 PRs landed back to back, both passing, neither touching the same files, but it didn't fail the GH actions check until both landed on master.
Once on master, we don't run GH actions as part of the release process, so this didn't pop up during the release cycle, but it did pop up in PRs based off master.


BartoszJarocki (approved):
thank you for fixing this!

Issue #8025, opened 2023-04-11
RiD: Create retro activity from "create custom {category} activity" button


jmtaber129:
From a non-quick start retro category view, create a fresh retro template.

["Create custom {category} activity" figma](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?node-id=337%3A20681&t=15PPYKq0ZGAuobG2-1)

[Modal Figma](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?node-id=915%3A28265&t=7C8RKY6hckMOKINM-1)

AC:
- [ ] When a user clicks "create custom {category} activity" (e.g. "create custom retro activity"), a team selection modal appears
- [ ] When a user clicks the "create activity" button on the modal, a new template is created on that team, and the window navigates to that modal

PR #8027, opened 2023-04-11
feat(rid): Create retro template from retro categories view

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8025

~Needs rebase once https://github.com/ParabolInc/parabol/pull/8008 lands~

Create a new retro template from the "Create New {category} Activity" button under retro categories in the activity lib.

Note that templates don't keep the category they're created from - this will be added when we move categories to the DB and allow them to be changed.

## Demo

https://www.loom.com/share/10160e33bb334a2ab55f8ae0d3687267

## Testing scenarios
- [ ] Create a new template from the "create custom {category} activity" button
- [ ] Select a team from the dropdown
- [ ] Click "select team", and confirm the window navigates to a fresh new retro template
- [ ] Give the template a name + prompts
- [ ] Go back without refreshing, and confirm the template is present
- [ ] Errors
  - [ ] Create a new template, then go back without updating name
  - [ ] Create a new template on the same team, and confirm an error message appears

## Final checklist

### DIFF
--- a/packages/client/components/ActivityLibrary/ActivityCard.tsx
+++ b/packages/client/components/ActivityLibrary/ActivityCard.tsx
@@ -42,12 +42,13 @@ export interface ActivityCardProps {
   className?: string
   category: CategoryID
   title?: string
-  imageSrc: string
+  imageSrc?: string
   badge: React.ReactNode | null
+  children?: React.ReactNode
 }
 
 export const ActivityCard = (props: ActivityCardProps) => {
-  const {className, category, title, imageSrc, badge} = props
+  const {className, category, title, imageSrc, badge, children} = props
   return (
     <div
       className={clsx(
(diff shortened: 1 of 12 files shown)


mattkrick:
-1 there's a regression in the /new-meeting view where the menu doesn't span full width. aside from that LGTM
Screenshot 2023-04-17 at 12 16 25 PM

jmtaber129:
@mattkrick I'm not sure this is a regression, as the repro steps I've found also reproduce this issue on master + in prod. This seems to only be the case when the dropdown loads as one size, then the window size changes such that it should be wider.  I.e. 1) load the `/new-meeting` view in a small window and open the dropdown, 2) increase window size, 3) open the dropdown again.

mattkrick:
weird! ah well not a showstopper, LGTM

PR #8030, opened 2023-04-12
feat(pipeline): Docker build pipeline

rafaelromcar-parabol:
# Description

TLDR: **Adds a pipeline to build the Docker image of the application and pushes it to our private Docker repository.**

- **Authentication:** It authenticates in a safe way without any service account token stored in the repository, always using temporary tokens. It uses a [Github Action](https://github.com/google-github-actions/auth) provided by Google for the authentication against Google Cloud Platform.
- **Built images:** it builds a new image with every push to the repository. The push can be two things:
  - **Commit to any branch:** it builds a Docker image and tags it with the commit hash and also the name of the branch in the form `docker_repository_development:COMMIT_HASH` and  `docker_repository_development:BRANCH_NAME`.
  - **Tag:** it build a Docker image and tags it with the commit hash and the git tag, pushing it to two places, development for the commit hash and production repository for the tag. There is no filter of the tag format. Docker images are `docker_repository_development:COMMIT_HASH` and  `docker_repository_production:GIT_TAG`
- **Secrets:** some secrets are required, and they were already added to the repository.

## Notes
- **Docker build**: the file `dockerignore` should be improved. Also, the Dockerfile could be improved, but it works right now and it is good enough.
- **Clean-up:** a clean-up is needed on the available dockerfiles and docker-compose files, as there are too many right now.
- **Pushing to the production Docker registry:** we might want to restrict the condition to push to the production Docker repository in the future. For now, only the the images built when a tag is pushed are pushed there.
- **Github Action filters on run:** we might want to filter and execute the action only for some tags and some branches with a given format, but I think we could start just with this simple action and iterate over it.


adaniels-parabol (approved):

.dockerignore
```diff
@@ -1,12 +1,45 @@
...
-/dist
-/.circleci
-/.git
-/.github
-/.vscode
-/packages/client/__generated__
-/.husky
-/.prettierrc
-/docs
+.git
```
mattkrick, line 1: -1 instead of this file imperatively including the contents of `.gitignore` could we do a `cat .gitignore > .dockerignore` in the docker-build.yml below? that way we don't have to worry about making sure to update this file whenever we update `.gitignore`.

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,84 @@
...
+          - 8080:8080
+          - 28015:28015
+          - 29015:29015
+      redis:
+        image: redis:6.2.6
+        ports:
+          - 6379:6379
+      postgres:
+        image: postgres:12.10-alpine
+        env:
```
mattkrick, line 28: +1 could we do a `env_file: ./environments/pipeline` to keep a single source of truth?

Dschoordsch (approved):
I built the image locally, but had trouble running it locally.

.github/workflows/docker-build.yml
rafaelromcar-parabol: Sadly we cannot do that. Please check the official documentation of Github Workflows for [job services](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idservices). The only reference to environment variables is just [env](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idservicesservice_idenv). And we cannot pass the env vars to the [PostgreSQL image](https://hub.docker.com/_/postgres) mounting a file somewhere.

.dockerignore
rafaelromcar-parabol: This file isn't used only during the pipelines, it is also used if anyone builds the image locally, using any Dockerfile (some built by them or some of the ones we have). This file must always be kept up to date with the files that need to stay out of the Docker images produced by our repository, as a security measure.

I think your solution is enough for the pipeline, but do not work for the rest.

rafaelromcar-parabol:
@mattkrick you told me about the --no-deps option that could be used and could then save quite a lot of space on the produced images. I don't know if it could also reduce the build time. Should we test it before merging or iterate after?

There is also room for improvement on the size of the image. A very simple article talking about it [here](https://blog.devgenius.io/reduce-the-size-of-your-node-js-docker-image-by-up-to-90-53aad23890e2). At the end of it there are some more articles. I would need to work with you on this, but I'm positive we could reduce the size of the image by a lot. Maybe @Dschoordsch would like to spend some time on this?

.dockerignore
mattkrick: can you tell me more about supporting local builds? I don't think any of the devs have a need to do that, so it might be a feature that we don't need to support.  if we do want to support it, could we concat the files in that dockerfile, too?

.github/workflows/docker-build.yml
mattkrick: could we use a github action to do that job?
e.g.
```yml
- uses: cardinalby/export-env-action@v2
  with:
    envFile: 'constants.env'  
```

another option (untetsed) is https://github.com/marketplace/actions/dotenv-action

rafaelromcar-parabol: We could, but that action (untested on my side as well), uses an env file to set environment variables on the Github runner host. The services are executed BEFORE any action. They run on the host, but before you can run anything else.

What is the tension here? This variables are unlikely to be changed as they are just temporary credentials for build. I don't see why would we want to change or adapt them. At least not very often.

mattkrick: if the env vars live on the host that seems OK with me!

The main concern is readability & knowledge transfer. There is an implicit constraint that the password used here must equal the password used in the env file. You know that & I know that, **_but the new developer we hire won't know that._** That constraint isn't documented anywhere. We could add a comment saying the two must be equal, but a better solution is self-documenting code. If this file reads the variables from the env, it communicates that dependency without extra words.

rafaelromcar-parabol: 100% with you on that. I tried solving this using a pipeline.env. 

I think I didn't explain myself correctly about the action you proposed. It wouldn't work. It sets the vars on the runner AFTER the services are running. It is either not using services, which are the easiest way of running the dbs in the pipelines, or having this vars hardcoded (and add a comment, which looks like a fine solution for a simple case like this one to me)

.dockerignore
mattkrick: gotcha! in this case I'm not sure if ignoring things like `/dist` is the right call because that would exclude all the built server assets, right?

could you provide some quick steps on how we should build this locally? With the `--no-deps` flag I imagine `/dist` is the _only_ thing we want to include in the build!

.github/workflows/docker-build.yml
mattkrick: oh! gotcha, `uses` isn't available under `services`? maaaaan that's lame. ah well, a comment works!

.dockerignore
rafaelromcar-parabol: Our Dockerfile (at least this one) do not use LOCAL files to build. That is the idea. Even if you build from your workstation, should not use already generated files.

When Docker starts to build, it takes the context as it is and start running the commands. If you check the lines 20 to 25 of the pipeline.dockerfile, you will see that the commands `yarn install --frozen-lockfile` and `yarn build` are executed. The dockerignore is there to ensure that no local /dist files are seen by the Docker daemon while building and it generates its own files, to be consistent between different docker builds done by different people in different machines using the same version of the code.

To be fair, this is only necessary while building on the workstations. Pipelines build are always done from scratch, so most of the .dockerignore isn't really useful. Probably we would be fine just with:

```
.git
.vscode
.github
.circleci
.husky
docs
```

More on [dockerignore](https://docs.docker.com/engine/reference/builder/#dockerignore-file)

There is no official way to specify a particular dockerignore file and that is why, for the pipeline, it must be on the root folder. Also for anyone building from local machine. And it must be then shared.

There are some [undocumented hacks](https://stackoverflow.com/questions/40904409/how-to-specify-different-dockerignore-files-for-different-builds-in-the-same-pr/57774684#57774684) we could test.

rafaelromcar-parabol:
I have been testing in my fork, I have pushed all the changes and I have some notes to share:
- **--no-deps:** the pipeline is prepared to, with a small change, start producing images using --no-deps. It takes more time to produce the image using that option, and it requires way more memory. We cannot use the free runners we are using right now. I'll check having our own big runners vs paying to use Github's. We really need to work on this, as current images are huge.
- **Cache:** I tried configuring the cache for the Docker images using our own registry as cache, but it doesn't work as expected yet. Also, even locally, I cannot really manage to split the commands yarn install and yarn clean cache from the yarn build. We will need to work together to do that. Fixing that might save up to 3 minutes on build time.
- **Security:** as building the image with additional security implies about 3 more minutes of build time, I disabled that for non-production images. Only images that are produced after a tag are built with additional security.
- **Build time:** it takes between 9 (without security and no-deps) and 20 minutes (with security + no-deps) minutes for the pipeline to end. I cannot make it smaller than 9 minutes right now, without much caching. We need to work together to get this to be faster, as I'm positive we can get that to even 3 minutes, if we do the work correctly. 

I think we should merge this, start building docker images and iterate over it, improving whatever we can improve. What do you think? @mattkrick @Dschoordsch @adaniels-parabol

rafaelromcar-parabol:
@Dschoordsch and I spent some time on sync discussing about how to improve the build time and how to cache things and here are the conclusions:
- There are multiple package.json and a single yarn.lock that needs to be copied and cached. This will be done in separate steps from the rest of the build process and we should be able to cached that. This might save up to 2 minutes. Though it might not be possible as we are not pushing the base image to the repository (and our cache is right now configured to be our Docker repository).
- The additional security might be cached if executed before the copy of the app code from the base image. This might save up to 4 minutes.
- The database migrations and pg:build steps might also be reduced soon. This isn't a priority as it will only save around 1 minute. We could try to copy over only the required files for the migrations and the pg:build in a separate step, and reproduce the same logic as with the package.json files and the yarn install. It might save time as well.
- Using --no-deps would reduce our image size by a lot, but it would also imply spending way more time on build. Space vs speed. TBD.

mattkrick:
sure, harmless to merge, but i still can't get it to build 🤔.
If you'd like me to review, I'll need some steps to run the build process & then some steps to run to verify that the build is operational. 

This is as far as I got

```
➜  parabol git:(rafaelromcar-parabol-docker-build) DOCKER_BUILDKIT=0 docker build -f docker/parabol-ubi/docker-build/dockerfiles/pipeline.dockerfile .
Sending build context to Docker daemon  67.34MB
Step 1/35 : ARG _NODE_VERSION=${_NODE_VERSION}
Step 2/35 : FROM node:${_NODE_VERSION} as base
invalid reference format
```

diegocarrillo:
> sure, harmless to merge, but i still can't get it to build 🤔. If you'd like me to review, I'll need some steps to run the build process & then some steps to run to verify that the build is operational.
> 
> This is as far as I got
> 
> ```
> ➜  parabol git:(rafaelromcar-parabol-docker-build) DOCKER_BUILDKIT=0 docker build -f docker/parabol-ubi/docker-build/dockerfiles/pipeline.dockerfile .
> Sending build context to Docker daemon  67.34MB
> Step 1/35 : ARG _NODE_VERSION=${_NODE_VERSION}
> Step 2/35 : FROM node:${_NODE_VERSION} as base
> invalid reference format
> ```

I think you're missing some arguments including: ```--build-arg _NODE_VERSION=18.14.2``` (take the value from the action)

rafaelromcar-parabol:
@diegocarrillo is right @mattkrick :)

I have to add the documentation, and I'm still trying to work on the cache before merging, but here you have some instructions:

From the root folder of the repository execute:

```
export postgresql_tag=12.10-alpine; \
export rethinkdb_tag=2.4.0; \
export redis_tag=6.2.6; \
export _BUILD_ENV_PATH=docker/parabol-ubi/docker-build/environments/pipeline; \
export _NODE_VERSION=$(jq -r -j '.engines.node|ltrimstr("^")' package.json); \
export _DOCKERFILE=./docker/parabol-ubi/docker-build/dockerfiles/pipeline.dockerfile
export _DOCKER_REPOSITORY=parabol; \
export _NO_DEPS=false/true; \
export _SECURITY_ENABLED=false/true; \
export _DOCKER_TAG=WHATEVER_YOU_WANT
```

Only the last three would be changed 99% of the time.

Then, to build the image:

```
docker run --name temp-postgres --network=host -e POSTGRES_PASSWORD=temppassword -e POSTGRES_USER=tempuser -e POSTGRES_DB=tempdb -d -p 5432:5432 postgres:${postgresql_tag} && \
docker run --name temp-rethinkdb --network=host -d -p 28015:28015 -p 29015:29015 -p 8080:8080 rethinkdb:${rethinkdb_tag} && \
docker run --name temp-redis --network=host -d -p 6379:6379 redis:${redis_tag} && \
docker build --network=host -t ${_DOCKER_REPOSITORY}:${_DOCKER_TAG} -f $_DOCKERFILE --build-arg _NODE_VERSION=$_NODE_VERSION --build-arg _BUILD_ENV_PATH=${_BUILD_ENV_PATH} --build-arg _SECURITY_ENABLED=${_SECURITY_ENABLED} --build-arg _NO_DEPS=${_NO_DEPS} . && \
docker stop temp-postgres temp-rethinkdb temp-redis && docker rm temp-postgres temp-rethinkdb temp-redis -f || docker stop temp-postgres temp-rethinkdb temp-redis && docker rm temp-postgres temp-rethinkdb temp-redis -f
```

- If you do not want to cache anything locally, please add `--no-cache` right after `docker build`.
- Please be aware that this docker build uses A LOT of disk space and might need some clean-up. The most brutal way to do it would be `docker system prune -a -f` but this will delete **all images**. You might want to manually delete them instead.

rafaelromcar-parabol:
The PR is ready to be merged IMHO.

- It will build an image on every push (tag or branch). For tags, it will push to our production private Docker repository. The Docker tag will be equal to the Git tag. For branches, it will push and produce two images in our development Docker repository: one with the commit hash as Docker tag and another with the branch name as Docker tag.
- It is prepared for, in the future, test and build using the option --no-deps. IMHO it should be the default option for all branches and tags once we have the runners we require for it, but that is for another PR.
- Documentation has been added on how to build with this new dockerfile BUT the process isn't really recommended for local workstation. This PR is really FOR the pipeline. That said, you can test to build following the instructions on the README.
- It isn't ultra fast, but it builds in between 7 and 9 minutes. Adding Docker cache makes things even slower as our image size is just to big. We might want to try that again once we build with --no-deps. For now, no cache for Docker.
- The produced images work perfectly when running the web-server and the gql-executor using `./dist/SCRIPT` and the migrations and postdeploy through yarn commands. It doesn't work when running migrations and postdeploy through ./dist scripts, but that IMHO should be for a different PR. The objective here is to get a pipeline that builds docker images that we can use to deploy the application, and it does it.

Please, let me know what you think @mattkrick @Dschoordsch @adaniels-parabol

Issue #7943, opened 2023-03-27
Bug(Sprint Poker): Error in ParabolScopingSearchResultItem


Dschoordsch:
https://parabol.sentry.io/issues/3402262694/?referrer=github_plugin

```
TypeError: Cannot destructure property 'id' of 'u' as it is undefined.
  at Ft (./packages/client/components/ParabolScopingSearchResultItem.tsx:78:14)
  at Set.forEach (<anonymous>)
...
(31 additional frame(s) were not displayed)
```

## How to reproduce
* Create a Sprint Poker meeting
* add parabol tasks
* advance the meeting
* delete one of the tasks (the task itself from the task dashboard by removing its content)
* go back to the sprint meeting scoping session
* press the new task button

Dschoordsch:
2.2k events, 355 users, thus this should be a p1

Dschoordsch:
Sentry issue: [ACTION-PRODUCTION-E0T](https://parabol.sentry.io/issues/3987304563/?referrer=github_plugin)

Dschoordsch:
Sentry issue: [ACTION-PRODUCTION-DWZ](https://parabol.sentry.io/issues/3983042913/?referrer=github_plugin)

Dschoordsch:
Sentry issue: [ACTION-PRODUCTION-DWY](https://parabol.sentry.io/issues/3983042911/?referrer=github_plugin)

Dschoordsch:
Linked additional issues which have different backtrace but appear in the same area and timespan thus probably have the same root cause.

Dschoordsch:
Yes. I discussed this yesterday with Matt. It's an issue resulting from adding the optimistic task to the edges of the task search result. Omitting the optimistic task would work, but Matt also suggested rearranging the GraphQL fragments, because James recently found out that something similar happened because of @plural items nested in a fragment.

lilicaruso:
Enterprise [User](https://app.hubspot.com/live-messages/3888472/inbox/4222831376#email) 🔒 followed up on resolution of this one.

Dschoordsch:
The workaround should be released by the end of this week.

Dschoordsch:
Just removed the p1 label to not convey unnecessary urgency since the workaround solves it for now.

jmtaber129:
@Dschoordsch @BartoszJarocki can this be closed in favor of a task for the actual fix since the bandaid fix is merged?

Dschoordsch:
[...]

PR #8032, opened 2023-04-12
fix: Workaround for adding Parabol tasks in Sprint Poker

Dschoordsch:
# Description

Workaround for #7943 

There is a bug with Relay: adding an optimistic result to an array will cause it to be undefined once the real result arrives. This workaround just skips the optimistic update adding to unblock users until there is a better fix.

## Testing scenarios

- [ ] Start Sprint Poker
- [ ] Add a Parabol task in the scope phase

## Final checklist


BartoszJarocki (approved):
works!

Issue #8024, opened 2023-04-11
RiD: Clone a retro template


jmtaber129:
Users can click a button on a template to clone that template to a team they're on. 

Note: substitute "copy" with "clone" almost everywhere in designs

[Figma for non-owner details view](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?node-id=310%3A23597&t=15PPYKq0ZGAuobG2-1)

[Figma for owner details view](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?node-id=308%3A20217&t=15PPYKq0ZGAuobG2-1) - since edit-mode isn't implemented, assume "delete" button is where edit button is.

[Figma for modal](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?node-id=915%3A27871&t=7C8RKY6hckMOKINM-1) - TODO: Confirm these designs are approved

AC:
- [ ] "Clone and edit" button on templates the user does not own.
- [ ] Clone icon button on templates the user does own.
- [ ] Clicking the clone button opens a team selection modal
- [ ] Clicking "clone to team" on the modal copies the template to that team, and navigates to the newly created template

PR #8036, opened 2023-04-12
feat(rid): Clone template

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8024

~Needs rebase once https://github.com/ParabolInc/parabol/pull/8027 lands~

Enable cloning templates from the Activity Library Details view.

**Note to maintainer reviewer:** This slightly relaxes the backend permissions for cloning. Previously, a user could only clone a template if a) it's team-scoped and they're on that team, or b) it's org-scoped and the cloning team is in the org, or c) it's public. This PR changes (b) to only require that the user is in the org. This means templates could be cloned across orgs. Alternatively, we could revert the backend permission changes, and only support cloning within an org.

## Demo
https://www.loom.com/share/ce0ef593692c42258a3dd2da1152c445

## Testing scenarios

- [ ] Clone edit templates for the following scenarios:
  - [ ] Team-scoped template copied to the same team
  - [ ] Team-scoped template copied to a team within the same org
  - [ ] Team-scoped template copied to a team in a different org
  - [ ] Org-scoped template copied to the same team
  - [ ] Org-scoped template copied to a team within the same org
  - [ ] Org-scoped template copied to a team in a different org
  - [ ] Public template
- [ ] Smoke test local updates (e.g. cloned templates appear in the grid view without refresh)
- [ ] (regression test) Create a retro activity from a category view.

## Final checklist

### DIFF
--- a/packages/client/components/ActivityLibrary/ActivityDetails.tsx
+++ b/packages/client/components/ActivityLibrary/ActivityDetails.tsx
@@ -1,4 +1,5 @@
 import graphql from 'babel-plugin-relay/macro'
+import {ContentCopy} from '@mui/icons-material'
 import React, {useCallback} from 'react'
 import {PreloadedQuery, usePreloadedQuery} from 'react-relay'
 import {Redirect, useHistory} from 'react-router'
@@ -28,6 +29,7 @@ import ActivityDetailsSidebar from './ActivityDetailsSidebar'
 import CloneTemplate from '../../modules/meeting/components/CloneTemplate'
 import useModal from '../../hooks/useModal'
 import TeamPickerModal from './TeamPickerModal'
+import FlatButton from '../FlatButton'
 
 graphql`
   fragment ActivityDetails_template on MeetingTemplate {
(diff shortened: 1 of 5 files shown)


mattkrick:
good call on the laxed permissions!

I think we could improve the UX in 1 of 2 ways:
- Assume the same teamId as the parent & skip the pick team modal, they can change that after creation
- Keep the team picker modal, but default to the current team instead of defaulting to the top one in the list

When we clone should we default the sharing level to "Organization"?

jmtaber129:
@mattkrick We don't currently have a mutation to transfer ownership of a template, so supporting (1) would increase the scope here substantially.

Re: (2) - I'd generally expect users to want to clone templates to _different_ teams, rather than copying to the same team and then tweaking. FWIW the default team logic here is the same as the team picker on the `/new-meeting` view.

I do think we should apply one of these _eventually_, but we should kick that can down the road until we have a working thing that we can iterate on.

> When we clone should we default the sharing level to "Organization"?
I think probably yes, but since that adds a small amount of complexity (we'd need to either modify the mutation or add a call to `UpdateTemplateScope` immediately after creation) and we're still discussing the broader UX around this, I'd prefer to keep the logic as-is to mirror the existing New Meeting view's behavior around template cloning.

mattkrick:
> I think probably yes, but since that adds a small amount of complexity (we'd need to either modify the mutation or add a call to UpdateTemplateScope immediately after creation) and we're still discussing the broader UX around this, I'd prefer to keep the logic as-is to mirror the existing New Meeting view's behavior around template cloning.

Totally agree, if there's any question it's fine to hold off. In this case I think making new/cloned templates default to org permissions is pretty stable though, but I'll default to @ackernaut. Should only be 1 line of code to fix

ackernaut:
Yes, defaulting the template sharing to the org level is still the plan. If that slows us down in the short term, I’m okay deferring. In that case I think we would open an implementation issue to make the changes to the default that we can tackle later?

mattkrick:
yeah, as long as we don't forget to get it in before GA I'm fine with whatever! there's a merge conflict, but withstanding that this PR is g2g

Issue #6051, opened 2022-02-14
GitLab v3: Self-Hosted Architectural Spec Written


nickoferrall:
Now that we've integrated Parabol with GitLab cloud, let's create an architectural spec for the self-hosted integration.  

### AC

- [ ] Spec is written in GitHub with a high-level proposal on how implementation will work. [See example here](https://github.com/ParabolInc/parabol/issues/4412#issuecomment-948003051).
- [ ] After spec is reviewed by Reviewer & Maintainer, the Epic and Issues are created (including design and implementation issues).

acressall:
A [user](https://app.hubspot.com/contacts/3888472/contact/94183751) has requested this integration:

> Hi, is there any way to integrate with Self-hosted Gitlab and Parabol ?

enriquesanchez:
[Another user](https://app.hubspot.com/live-messages/3888472/inbox/4057812496#email) 🔒 has also inquired about the self-hosted version:

>My team host gitlab on our own server so it's it differ url then the main gitlab. Can I still integrate parabol with my team's gitlab?

jordanh:
Removing stale as a user asked about this last week. Keeping open for now to collect more signal

jordanh:
(chances are we'll do this for one of our DoD folks)

jordanh:
Taking this out of the icebox

Dschoordsch:
We're going to support only 1 GitLab integration per team member. This is already possible with the current architecture.

PR #8046, opened 2023-04-13
feat: make gitlab url dynamic for onprem gitlab

adaniels-parabol:
Make gitlab url part of env vars so it can be set for onprem instance of gitlab


mattkrick:
Would this expose our githlab client secret with the PPMI? 
we've got a mutation for adding a self-hosted gitlab: https://github.com/ParabolInc/parabol/blob/9177909d9b3bc9cf2ba5d88a32c0ce53b1da177f/packages/server/graphql/mutations/addIntegrationProvider.ts. 

I'm not sure if it's all the way hooked up yet, but that's the pattern.

adaniels-parabol:
> Would this expose our githlab client secret with the PPMI? we've got a mutation for adding a self-hosted gitlab: https://github.com/ParabolInc/parabol/blob/9177909d9b3bc9cf2ba5d88a32c0ce53b1da177f/packages/server/graphql/mutations/addIntegrationProvider.ts.
> 
> I'm not sure if it's all the way hooked up yet, but that's the pattern.

Thus far no one has request gitlab integration on PPMI. This was for self-hosted, I wanted an option to set the URL for this reason. It seems like this would be an easier method? However, I don't mind trying the other integration. Can you give an example of how we'd enable it?

Note on self-hosted or PPMI if they wanted to use gitlab they would add an app to their instance and get client_id and client secret for their environment.

mattkrick:
yeah, either webhook or oauth2, I forget which one gitlab likes to use. it takes a few extra fields, but it looks like this:

```gql
mutation {
  addIntegrationProvider(input: {oAuth2ProviderMetadataInput: {serverBaseUrl: "htpp://foo.co", clientId: "abc", clientSecret: "Shhh"}})
}
```

mattkrick:
closing this in favor of https://github.com/ParabolInc/parabol/issues/6051. Uncle Sam is asking for this, so it's gonna happen!

PR #8052, opened 2023-04-14
fix: require accepted invites before sending more

mattkrick:
# Description

We've had some spam invites going out.
To fix, we'll exclude trouble domain.
Next, we'll trust the first 5 invites.
After 5, we'll require:
- At least 1 invite to have been accepted
- At least 15% of all invites sent via email have been accepted

### DIFF
--- a/packages/server/graphql/mutations/inviteToTeam.ts
+++ b/packages/server/graphql/mutations/inviteToTeam.ts
@@ -16,6 +16,7 @@ import removeSuggestedAction from '../../safeMutations/removeSuggestedAction'
 import {analytics} from '../../utils/analytics/analytics'
 import {getUserId, isSuperUser, isTeamMember} from '../../utils/authorization'
 import getBestInvitationMeeting from '../../utils/getBestInvitationMeeting'
+import getDomainFromEmail from '../../utils/getDomainFromEmail'
 import publish from '../../utils/publish'
 import standardError from '../../utils/standardError'
 import {GQLContext} from '../graphql'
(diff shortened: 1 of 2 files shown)


packages/server/graphql/mutations/inviteToTeam.ts
```diff
@@ -63,6 +64,34 @@ export default {
       }
 
       // RESOLUTION
+      const getInviteTrustScore = async (userId: string, teamId: string) => {
+        const untrustedDomains = ['tempmail.cn']
```
tianrunhe, line 68: +1 put in a constant and maybe add a comment to our Slack thread so other dev understands why

packages/server/graphql/mutations/inviteToTeam.ts
```diff
@@ -63,6 +64,34 @@ export default {
...
+          r.table('TeamInvitation').getAll(teamId, {index: 'teamId'}).count().run(),
+          r
+            .table('TeamInvitation')
+            .getAll(teamId, {index: 'teamId'})
+            .filter({acceptedAt: null})
+            .count()
+            .run()
+        ])
+        // trust their first 5 invites
+        if (total <= 5) return 0.95
```
tianrunhe, line 84: I wonder if 5 is too small, shall we try 10 to start?

tianrunhe (approved):

packages/server/graphql/mutations/inviteToTeam.ts
mattkrick: what does put in a constant mean here?

tianrunhe: Create a server side file like we have for `packages/client/utils/constants.ts` and add a line like:
```TypeScript
export const UNTRUSTED_DOMAINS = ['tempmail.cn']
```

PR #8059, opened 2023-04-17
fix(recurrence): prevent repeating date in standups name

BartoszJarocki:
# Description

Fixes an issue where the standup title would be like `Friday Ship - April 17 - April 17`. The issue was caused by the implementation detail, where `MeetingTeamPrompt` constructor would always append a date to a title, no matter if it was already there or not. I moved creating title outside of the constructor for all cases as eventually we'll use a different timezone when creating a single instance meeting vs meeting series.

## Demo

No demo

## Testing scenarios

- [ ] Run process recurrence a couple of times to make sure it's not happening anymore

## Final checklist

### DIFF
--- a/packages/server/database/types/MeetingTeamPrompt.ts
+++ b/packages/server/database/types/MeetingTeamPrompt.ts
@@ -57,8 +57,7 @@ export default class MeetingTeamPrompt extends Meeting {
       phases,
       facilitatorUserId,
       meetingType: 'teamPrompt',
-      //TODO: use client timezone here (requires sending it from the client and passing it via gql context most likely)
-      name: createTeamPromptTitle(name || 'Standup', new Date(), 'UTC'),
+      name,
       meetingSeriesId,
       scheduledEndTime
     })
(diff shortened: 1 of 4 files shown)


packages/server/database/types/MeetingTeamPrompt.ts
```diff
@@ -57,8 +57,7 @@ export default class MeetingTeamPrompt extends Meeting {
       phases,
       facilitatorUserId,
       meetingType: 'teamPrompt',
-      //TODO: use client timezone here (requires sending it from the client and passing it via gql context most likely)
-      name: createTeamPromptTitle(name || 'Standup', new Date(), 'UTC'),
+      name,
```
jmtaber129, line 60: +1 We should make `name` on the `Input` interface required.

jmtaber129 (approved):
LGTM aside from one minor type issue

packages/server/database/types/MeetingTeamPrompt.ts
BartoszJarocki: It'd require frontend changes, as currently, we don't require users to provide a meeting series name. We could always use `Standup` by default but this seems to be a different question. I'll merge this one, we'll can discuss further improvements before jumping into impl

PR #8066, opened 2023-04-18
fix: Rename makeTemplateDescription -> useTemplateDescription to satisfy hooks rules

jmtaber129:
# Description
FIxes errors in https://github.com/ParabolInc/parabol/actions/runs/4726296812/jobs/8385744221?pr=8036

`makeTemplateDescription` (not a hook) calls `useFragment` (a hook), which breaks React hook rules, since hooks can only be called from function components and other hooks.

Rename `makeTemplateDescription` to `useTemplateDescription` to make it a hook + rearrange code that calls it conditionally.

Also includes a minor fix where `useAtmosphere` was called conditionally.

Also fixes prettier in a file (what's actually breaking code style on master)

## Demo
N/A

## Testing scenarios
Smoke test places where template descriptions are used.

## Final checklist

### DIFF
--- a/packages/client/components/ActivityLibrary/ActivityDetails.tsx
+++ b/packages/client/components/ActivityLibrary/ActivityDetails.tsx
@@ -11,7 +11,7 @@ import AddTemplatePrompt from '../../modules/meeting/components/AddTemplatePromp
 import {ActivityCard, CategoryID, MeetingThemes} from './ActivityCard'
 import {activityIllustrations} from './ActivityIllustrations'
 import customTemplateIllustration from '../../../../static/images/illustrations/customTemplate.png'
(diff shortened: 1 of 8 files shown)


packages/client/components/ActivityLibrary/ActivityDetails.tsx
```diff
@@ -109,26 +109,27 @@ const ActivityDetails = (props: Props) => {
     )
   }, [templateId, submitting, submitMutation, onError, onCompleted])
 
-  if (!selectedTemplate) {
```
jmtaber129, line 112: `makeTemplateDescription` was being called conditionally because of this, so we need to reorder some of this logic.

BartoszJarocki (approved):

PR #8071, opened 2023-04-19
feat: push meeting templates to CDN

### LINKED ISSUE #8070: Priming DB with image URLs
Today, each meeting template has an illustration. 
Instead of handling this in the DB, we have a large lookup table on the client: https://github.com/ParabolInc/parabol/blob/19b12d4203f2ca41d49b83bfb8dafde25bc483cb/packages/client/components/ActivityLibrary/ActivityIllustrations.tsx. It's grown into a little database!

### Acceptance Criteria
As we allow for new, custom templates, we want to support folks being able to upload their own illustrations in addition to the illustrations that we already have. Heck, maybe we even generate some AI illustrations & let them pick. To do this, we'll need to store the image in the CDN & keep a URL to that image in postgres' MeetingTemplate table. The tricky part is doing this in a way that makes it easy to support private, air-gapped instances.

Bonus points if the strategy can support priming our CDN with our default avatars (AA, AB, AC, etc...). We don't do that today.

### Strategy

We cannot push to the CDN during the webpack build because the build will only happen once in CI & then distributed to many places as a docker image

That means just like we have database migrations for postgres, we'll likely need a CDN migration for illustrations. It will need to make sure that all the illustrations that we may need exist in the CDN.
We also need to make it easy to add new meeting templates with images. that means the image name must be deterministic when the migration is run.

alternatively, we could support a hybrid approach where known templates have an image stored on the client & unknown templates have an `illustrationUrl` stored in the DB.

another alternative is for known templates we keep a filename in the `illustrationUrl` without a path & then it'll just be part of the build directory, e.g. /static/v6.99.0/.

Ultimately, the "right" way to do it is to keep all the data together & to do no special casing.That means all image URLs live in `illustrationUrl` and they are all full URLs.

### Proposal

- On the server, we require the entire static/images/illustrations directory. That way all illustrations will make it into the server bundle. We plop them in /illustrations.
- we have a migrateImages script that gets called on predeploy. it makes sure those images land in the CDN, but outside of the build directory. that way they can be referenced by the database

mattkrick:
# Description

Fixes #8070

- Template illustrations get bundled with the server assets
- a script gets called predeploy which looks for those assets & pushes them to a CDN if they don't exist
- it works for both S3 and local image hosting
- i had to bump the aws-sdk. It was reeeeeeeally old & calling `headObject` was sometimes failing
- After this is done we can start adding these URLs to rows in the DB

## Demo

Nothing to see :-(

## Testing scenarios

- run `yarn build` and look at /dist/templates
- run `node dist/migrateImages` and see the files have been pushed to S3 (e.g. https://action-files.parabol.co/development/store/Organization/aGhostOrg/template/gladSadMadTemplate.png)
- change your .env `FILE_STORE_PROVIDER='local'` and re-run the above 2 commands. see the files live in /self-hosted
- change your user avatar, just to make sure it works

### DIFF
--- a/packages/server/fileStorage/S3Manager.ts
+++ b/packages/server/fileStorage/S3Manager.ts
@@ -1,52 +1,60 @@
-import aws from 'aws-sdk'
+import {HeadObjectCommand, PutObjectCommand, S3Client} from '@aws-sdk/client-s3'
 import mime from 'mime-types'
-import {APP_CDN_USER_ASSET_SUBDIR} from 'parabol-client/utils/constants'
-import protocolRelativeUrl from 'parabol-server/utils/protocolRelativeUrl'
 import path from 'path'
+import url from 'url'
 import FileStoreManager from './FileStoreManager'
 
 export default class S3Manager extends FileStoreManager {
-  private baseUrl: {hostname: string; pathname: string}
+  // e.g. development, production
+  private envSubDir: string
+  // e.g. action-files.parabol.co
   private bucket: string
-  private s3: AWS.S3
+  private s3: S3Client
   constructor() {
     super()
-    const {CDN_BASE_URL, AWS_S3_BUCKET} = process.env
+    const {CDN_BASE_URL, AWS_S3_BUCKET, AWS_REGION} = process.env
     if (!CDN_BASE_URL || CDN_BASE_URL === 'key_CDN_BASE_URL') {
       throw new Error('CDN_BASE_URL ENV VAR NOT SET')
     }
     if (!AWS_S3_BUCKET) {
       throw new Error('AWS_S3_BUCKET ENV VAR NOT SET')
     }
-    const baseUrl = protocolRelativeUrl.parse(CDN_BASE_URL) as any
-    if (!baseUrl.hostname || !baseUrl.pathname) {
+    const baseUrl = url.parse(CDN_BASE_URL.replace(/^\/+/, 'https://'))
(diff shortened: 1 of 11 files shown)


BartoszJarocki:
everything looks good except for the fact I'm getting 404 when I change the avatar after setting file store provider to local. I can see it in `/self-hosted` so should be good but `selfHostedHandler` doesn't seem to be called. 

Screenshot 2023-04-21 at 09 13 31

Screenshot 2023-04-21 at 09 13 39

@mattkrick would you mind checking if this case works for you? I just want to rule out my local env quirks

mattkrick:
@BartoszJarocki right you are! it's actually a bug in master that only affects dev. i went ahead & fixed it here

packages/server/fileStorage/S3Manager.ts
```diff
@@ -1,52 +1,60 @@
...
+  // e.g. development, production
+  private envSubDir: string
+  // e.g. action-files.parabol.co
   private bucket: string
-  private s3: AWS.S3
+  private s3: S3Client
   constructor() {
     super()
-    const {CDN_BASE_URL, AWS_S3_BUCKET} = process.env
+    const {CDN_BASE_URL, AWS_S3_BUCKET, AWS_REGION} = process.env
```
BartoszJarocki, line 15: `AWS_REGION` is already present in all the env configs, you just added the usage here, right?

BartoszJarocki (approved):
lgtm! 🚢

PR #8072, opened 2023-04-19
fix(stripe): Fix the bug when there's failed payment we charge customer extra amount

### LINKED ISSUE #8061: Multiples charges incurred on Team subscription when customer failed payments
## Issue - Bug

We have seen issues when customer failed their payments and then got duplicate charges on future invoices (e.g., [this one](https://parabol.slack.com/archives/C029ZTH0EL9/p1681386083144879)). Today we we receive payment_fail event, we delete the subscription but somehow we are leaving a balance on that customer account.

- [ ] User won't be extra charged on failed payments

tianrunhe:
# Description

Fixes #8061. To reproduce the bug, on `master` branch:

1. [Install Stripe CLI](https://stripe.com/docs/stripe-cli) if you haven't
2. Setup Stripe webhook to forward to your local server: `stripe listen --forward-to localhost:3000/stripe`, and copy the webhook singing secret to replace the one in [your .env](https://github.com/ParabolInc/action-devops/blob/master/environments/local#L73)
3. Create a new team/org and go to the org billing page
4. Upgrade with this Stripe payment failure test card: `4000 0000 0000 3063`, expiration date any month in the future, CVC can be anything
Screenshot 2023-04-19 at 12 26 52

5. Observe the org is upgraded
6. Wait a moment and then refresh the page, you should see that one invoice is in the state of PENDING (because the payment wasn't successful) but another invoice for next month with 2x of the amount
Screenshot 2023-04-19 at 12 28 00

7. Now update the credit card number to `4242 4242 4242 4242`

8. Wait a moment and then refresh the page, you should see a new invoice generated and paid with 2x of the amount, indicating the customer has been double charged.
Screenshot 2023-04-19 at 12 29 37

## Testing scenarios
With this branch pulled and server started, repeat the above-mentioned steps 3 - 5.

6. Wait a moment and then refresh the page, you should see one invoice in the state of FAILED:
Screenshot 2023-04-19 at 12 34 58

7. Now update the credit card number to `4242 4242 4242 4242`
8. Wait a moment and then refresh the page, you should see a new invoice generated and paid the correct amount, and an upcoming invoice also with the correct amount

Screenshot 2023-04-19 at 12 37 15

## Final checklist

### DIFF
--- a/packages/server/billing/helpers/terminateSubscription.ts
+++ b/packages/server/billing/helpers/terminateSubscription.ts
@@ -18,7 +18,7 @@ const terminateSubscription = async (orgId: string) => {
             stripeSubscriptionId: null
           },
           {returnChanges: true}
-        )('changes')(0)('old_val')
+        )('changes')(0)('old_val')('stripeSubscriptionId')
         .default(null) as unknown as string
     }).run(),
     updateTeamByOrgId({isPaid: false}, orgId)
@@ -32,7 +32,7 @@ const terminateSubscription = async (orgId: string) => {
     try {
       await manager.deleteSubscription(stripeSubscriptionId)
     } catch (e) {
-      // noop
+      console.error(`cannot delete subscription ${stripeSubscriptionId}`, e)
     }
   }
   return stripeSubscriptionId
--- a/packages/server/billing/helpers/terminateSubscription.ts
+++ b/packages/server/billing/helpers/terminateSubscription.ts
@@ -27,8 +27,6 @@ const terminateSubscription = async (orgId: string) => {
   const {organization} = rethinkResult
   const {stripeSubscriptionId} = organization
 
-  // stripe already does this for us (per account settings) but we do it here so we don't have to wait an hour
-  // if this function is called by a paymentFailed hook, then the sub may not exist, so catch and release
   if (stripeSubscriptionId) {
     const manager = getStripeManager()
     try {
--- a/packages/server/billing/helpers/terminateSubscription.ts
+++ b/packages/server/billing/helpers/terminateSubscription.ts
@@ -1,4 +1,5 @@
 import getRethink from '../../database/rethinkDriver'
+import Organization from '../../database/types/Organization'
 import updateTeamByOrgId from '../../postgres/queries/updateTeamByOrgId'
 import {getStripeManager} from '../../utils/stripe'
 
@@ -8,7 +9,7 @@ const terminateSubscription = async (orgId: string) => {
   // flag teams as unpaid
   const [rethinkResult] = await Promise.all([
     r({
-      stripeSubscriptionId: r
+      organization: r
         .table('Organization')
         .get(orgId)
         .update(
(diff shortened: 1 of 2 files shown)


nickoferrall (approved):
Really appreciate the testing steps with screenshots! It made my life as a reviewer really easy.  It works, and LGTM!

Dschoordsch:
> webhook singing secret

I wonder which tune it sings

packages/server/billing/helpers/terminateSubscription.ts
```diff
@@ -18,7 +18,7 @@ const terminateSubscription = async (orgId: string) => {
             stripeSubscriptionId: null
           },
           {returnChanges: true}
-        )('changes')(0)('old_val')
+        )('changes')(0)('old_val')('stripeSubscriptionId')
```
Dschoordsch, line 21: -1 this seems to contradict how we're reading the `rethinkResult` in line l26

packages/server/billing/helpers/terminateSubscription.ts
```diff
@@ -32,7 +32,7 @@ const terminateSubscription = async (orgId: string) => {
     try {
       await manager.deleteSubscription(stripeSubscriptionId)
     } catch (e) {
-      // noop
+      console.error(`cannot delete subscription ${stripeSubscriptionId}`, e)
```
Dschoordsch, line 35: +1 Reading the comment in l28 it appears as if we intentionally do not track errors here. Are you sure this is necessary?

Dschoordsch (changes requested):
Looks good and it's the best type of fix where we just remove broken code :sunglasses: 
However the changes in terminateSubscription appear wrong, are these covered by the test case?

packages/server/billing/helpers/terminateSubscription.ts
tianrunhe: I wasn't sure what you meant by contradiction, but I made a few changes to make it more clear & easy to read. Let me know if that's what you were thinking?

packages/server/billing/helpers/terminateSubscription.ts
tianrunhe: I would highly suspect that it never did what it was supposed to do... the issue was hidden as we didn't have that many self upgrades in the past, let alone the even rare cases where the payment failed.

As a quick test, I commented out this part as the comment seemed to indicate, i.e., it wouldn't harm if the deletion failed or not even performed, as Stripe is configured to do so automatically. Well, not the case here. If the customer failed to perform the first payment and just did nothing, this is gonna happen:
1. A failed payment notification was sent
2. The invoice stays in PENDING state
3. The org still stays in the Team tier, i.e., the customer is enjoying the Team tier for free

Screenshot 2023-04-25 at 11 47 29

Looking at Stripe side, after 24 hours, this is going to happen:
1. The invoice is void
2. The payment was cancelled 
3. The subscription was incomplete
But again, none of those prevent us from giving the customers Team tier for free, from the app side.

Screenshot 2023-04-25 at 11 48 02

So, all in all, we definitely need that subscription deletion to finish and succeed, in order for the app to cancel that upgrade. Therefore, we should capture the error.

Dschoordsch: Thanks for investigating this and removing the comment!

packages/server/billing/helpers/terminateSubscription.ts
Dschoordsch: Thanks for the change, now I understand what you were doing.

Dschoordsch (approved):
Did not test again, but looks good now!

PR #8074, opened 2023-04-19
feat: add mainCategory and illustrationUrl to MeetingTemplate

### LINKED ISSUE #7714: Retros in disguise: Categories
1 template can have many categories.
Within each category, a template will have a sort order. 
seasonal templates should show up at certain seasons.

To do this, I'd propose the following:

- creating a `TemplateCategory` table `{id, name}`.
- creating an intersection table like `TemplateCategoryTemplate`. `{templateId, categoryId, sortOrder}`.
- we could just put an array of category names on `MeetingTemplate`, but then we'd have to do sortOrder differently. We could sort alphabetically, or have an overall popularity instead of a category-specific sortOrder. I like this more. we could use enums for meeting categories, or we could just use strings & trust that we verify that on the application side. problem is how to sort by popularity? 
- we also should account for hiding seasonal templates (e.g. Christmas Retro). 
   - We could do this by setting the sortOrder to -1, but then we'd have something to set. 
   - probably cleaner to have a `hideStartingAt, hideEndingAt` on the template itself.
   - we could have a "seasonal" category, but we might have a "summer vacation" retro or similar
   - we could add in some logic locally to hide the templates within a certain date range. the christmas retro will probably go to the client anyways. 

To filter seasonal templates, we could do something like the follow in MySQL:

```sql

# MySQL
SELECT * from table where
MONTH(fieldA) > MONTH(fieldB) OR(
MONTH(fieldA) = MONTH(fieldB) AND DAY(fieldA) >= DAY(fieldB))

# PG has something like 
DATE_PART('day', fieldA)

# alternatively see https://wiki.postgresql.org/wiki/MONTH()_equivalent
```

AC (WIP)
- [ ] add `illustrationURL: VARCHAR(255)` to table  
- [ ] add a `string` called mainCategory on the MeetingTemplate table. For now, every template will only have 1 category in the DB. In GraphQL, we'll probably make a `categories` field which will include the main category + other categories like "your favorites", "recently played", "from your org". If we do this, then we lose the sortOrder for the carousel vs. having a category with an array of templates.
- [ ] add `hideStartingAt, hideEndingAt` on the MeetingTemplate table. hide seasonal templates using this
- [ ] After #7713 is complete we can worry about populating the "for you" category as well as the sort order. 
- [ ] I presume sortOrder will be a combination of overall popularity, A/B testing, and team/org popularity. For now, we can keep it simple & alphabetical

mattkrick:
# Description

Fix #7714

- Adds application logic to ensure new templates have `illustraitonUrl, mainCategory` rows
- Adds nullable `illustrationUrl, mainCategory` to MeetingTemplate table (will make non-nullable in next migration, requires above application logic to be in effect)
- Adds `hideStartingAt, hideEndingAt` to be used for hiding seasonal templates
- adds a new table so users can mark select templates as their favorites

## Demo

Creating a new template results in the following:
Screenshot 2023-04-19 at 4 18 26 PM

## Testing scenarios

- create a new retro/poker template. see the new columns are non-null
- clone one of the new retro/poker templates that you just created. see the new columns are non-null

note: if you clone an old template, these columns will still be null on the clone. that's OK! in the next migration we'll prime all the rows.


jmtaber129 (commented):
What's the reasoning behind adding all these fields for unrelated features in a single PR?

I generally have a strong preference to only add columns alongside the code that actually uses the columns for the following reasons:
1. Adding these columns don't provide any value on their own - adding them now doesn't less work, and has the potential to create more work down the line (see below)
2. We're sort of locking ourselves into a specific data model without knowing what the feature code will actually look like. I.e. a dev could start implementation and then realize a different data model would be better, creating more work (either by doing a migration or working around the data model)
3. If any of these features end up getting deprioritized for any reason, we're left with a bunch of unused DB cruft that could deteriorate over time and need to be re-migrated to be useful.

mattkrick:
it sounds like you have some concerns with adding a DB column before the application code uses that column. is there a particular column that you have some concerns about & we can address that one specifically?

mattkrick:
Following up from our slack convo & your comment here:

Descriptions:
- `illustrationUrl` is here to build on #8071 and will be primed by #8081. It allows folks to upload custom illustrations, but more importantly we'll be able to add meeting templates via migration without updating the lookup table in the client code
- `mainCategory` is primed by #8081, it'll be how we segment activities across the top of the activity library. we may need to  have a `categories` field in GraphQL, or we'll make a `Categories` object. Architecture is in #8080
- `hideStartingAt, hideEndingAt` is primed in #8081. Showing e.g. Christmas templates in July is a longstanding bug that distracts users with templates that they're not interested in.

PR #8081, opened 2023-04-20
feat: prime new columns on MeetingTemplate [merge after #8074 is in prod]

mattkrick:
# Description

After #8074 lands in production, the application will guarantee that all new meeting templates will have an illustrationUrl & mainCategory.
With that guarantee in place, we can backfill all the rest of the templates & set those columns to non-null.
This also adds hide dates to seasonal templates. No more Christmas templates in July 🎉 

## Demo

https://www.loom.com/share/2814231052b44a37975f7492bd2ce45b

## Testing scenarios

- Go to the templates section & you don't see a christmas template.
- run the migration. nothing breaks.


mattkrick:
Got confirmation from Gareth that the time ranges are OK, we'll just have to be sure the permalinks work during off-season.

PR #8086, opened 2023-04-21
feat(prompt-to-join): Add teammate UI

### LINKED ISSUE #7802: Prompt to join org: Add teammate UI
Clicking Review request button should open a dialog that allows to add a user to specific teams on the domain

Designs

image

- Blue check -> when user has added (active)
- Slate 600 outline -> available, unchecked
- Slate 600 check -> disabled, user has already been added to team

**AC:**

- Implement add teammate dialog
- Dialog should appear when the Review button is clicked on the notification or snackbar or notification (#7797)
- "Cancel" button should close the dialog
- Dialog shows **all** teams under the domain where the user is a team lead
- Enterprise organizations are excluded

**Not included:**

- Add to teams mutation

igorlesnenko:
Fixes #7802

image

**How to test:**

- Re-run the migration packages/server/postgres/migrations/1680865284180_addDomainJoinRequestTable.ts from previous PR, as I changed it (if you haven't)
- Delete existing `REQUEST_TO_JOIN_ORG` notifications 
```
r.db('actionDevelopment')
    .table('Notification')
   .filter({type:'REQUEST_TO_JOIN_ORG' })
   .delete()   
```

- Follow testing guide from  #8003 and #8044 to create necessary notifications
- Click on `Review` button in the notification in header, see that dialog appeared
- Click on `Review` button in the snackbar, see that dialog appeared
- Check that teams are listed correctly (see #7802)


packages/client/components/ReviewRequestToJoinOrgModal.tsx
```diff
@@ -0,0 +1,127 @@
...
+
+const query = graphql`
+  query ReviewRequestToJoinOrgModalQuery {
+    viewer {
+      teams {
+        id
+        name
+        teamMembers(sortBy: "preferredName") {
+          userId
+          preferredName
```
nickoferrall, line 30: +1 is preferredName being used here?

packages/client/components/ReviewRequestToJoinOrgModal.tsx
```diff
@@ -0,0 +1,127 @@
...
+        return (
+          <div className='mb-2 flex items-center text-base' key={teamId}>
+            <Checkbox
+              active={active}
+              disabled={disabled}
+              onClick={handleClick}
+              color={active && !disabled ? PALETTE.SKY_500 : undefined}
+            />
+            <label className={`ml-2 ${disabled ? 'opacity-[.38]' : ''}`}>
+              {teamName} | {orgName}
```
nickoferrall, line 83: +1 would each label include the same `orgName`? If so, I think we should remove it here and include it in the subtitle, so: `Which teams in ${orgName} would you like to...`

Screenshot 2023-04-25 at 16 27 20

packages/client/components/ReviewRequestToJoinOrgModal.tsx
```diff
@@ -0,0 +1,127 @@
...
+          if (active) {
+            setSelectedTeams((prevSelectedTeams) => prevSelectedTeams.filter((id) => id !== teamId))
+          } else {
+            setSelectedTeams((prevSelectedTeams) => [...prevSelectedTeams, teamId])
+          }
+        }
+
+        return (
+          <div className='mb-2 flex items-center text-base' key={teamId}>
+            <Checkbox
```
nickoferrall, line 76: +1 perhaps there should be a select all button? We want the user to be added to as many teams as possible, so we should make this easy for the team leads

nickoferrall (approved):
I've tested it, and it's looking great!

packages/client/components/ReviewRequestToJoinOrgModal.tsx
igorlesnenko: Not each label will include org name, because you may have multiple different orgs on the same domain
However, I agree that his could be improved. Added to the list of ideas for v2 if it is ok: https://www.notion.so/parabol/Do-you-want-to-join-x-org-prompt-V2-d09a562280b14b81bad54b87de31d3b0

packages/client/components/ReviewRequestToJoinOrgModal.tsx
igorlesnenko: Added to the list of ideas for v2 if that is ok: https://www.notion.so/parabol/Do-you-want-to-join-x-org-prompt-V2-d09a562280b14b81bad54b87de31d3b0

packages/client/components/ReviewRequestToJoinOrgModal.tsx
```diff
@@ -0,0 +1,127 @@
...
+  )
+}
+
+interface Props {
+  closePortal: () => void
+  requestCreatedBy: string
+  email: string
+}
+
+const ReviewRequestToJoinOrgModal = (props: Props) => {
```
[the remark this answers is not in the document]
igorlesnenko: In this particular case the second component is just a way to use `useQueryLoaderNow`. Usually we were creating a separate `Root` component and importing the query from the child component. I felt it is overcomplicated to just run a single graphql query. I'll keep it as is for now, maybe @Dschoordsch has strong opinion in this.

packages/client/components/ReviewRequestToJoinOrgModal.tsx
```diff
@@ -0,0 +1,127 @@
...
+  const {closePortal, email, requestCreatedBy} = props
+
+  const queryRef = useQueryLoaderNow<ReviewRequestToJoinOrgModalQuery>(query)
+
+  return (
+    <DialogContainer>
+      <DialogTitle>{'Add teammate'}</DialogTitle>
+      <DialogContent>
+        <div className={'mb-4 text-base'}>
```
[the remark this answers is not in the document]
igorlesnenko: I added more tailwind here, but still reusing existing Dialog components. I created issue about creating generic tailwind dialog components: #8107

packages/client/components/ReviewRequestToJoinOrgModal.tsx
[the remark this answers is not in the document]
igorlesnenko: I had to change this in next PR, will be classic Root component scheme

packages/client/components/ReviewRequestToJoinOrgModal.tsx
```diff
@@ -0,0 +1,117 @@
...
+import PrimaryButton from './PrimaryButton'
+import SecondaryButton from './SecondaryButton'
+import {ReviewRequestToJoinOrgModalQuery} from '../__generated__/ReviewRequestToJoinOrgModalQuery.graphql'
+import useQueryLoaderNow from '../hooks/useQueryLoaderNow'
+import Checkbox from './Checkbox'
+
+const query = graphql`
+  query ReviewRequestToJoinOrgModalQuery {
+    viewer {
+      teams {
```
Dschoordsch, line 16: -1 We're querying all teams, not just the relevant ones. I see that all of them probably have the matching domain, but I think it should be filtered at least to orgs with the feature flag set.
We might even just limit it to the orgId from the request.

Dschoordsch (changes requested):

packages/client/components/RequestToJoinOrgNotification.tsx
```diff
@@ -19,23 +25,44 @@ const RequestToJoinOrgNotification = (props: Props) => {
...
+  const {togglePortal, modalPortal, closePortal} = useModal({
+    id: 'reviewRequestToJoinOrgModal',
+    parentId: 'topBarNotificationsMenu'
+  })
+  const {name, email, picture, requestCreatedBy} = notification
 
-  const onActionClick = () => {
-    // TODO: implement review dialog
+  const onActionClick = (e: React.MouseEvent) => {
+    e.stopPropagation()
```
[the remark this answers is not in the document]
igorlesnenko: Change approach in the next PR: https://github.com/ParabolInc/parabol/pull/8136/files#diff-8d7169a8176d285e712b14866607557aaaefec3c86afcabad180ee3424e302f3R34-R37

packages/client/components/RequestToJoinOrgNotification.tsx
```diff
@@ -19,23 +25,44 @@ const RequestToJoinOrgNotification = (props: Props) => {
...
+      <NotificationTemplate
+        avatar={picture}
+        message={`${name} has requested to join your organization`}
+        action={<NotificationAction label={`Review ${email}`} onClick={onActionClick} />}
+        notification={notification}
+      />
+      {modalPortal(
+        <div
+          onClick={(e) => {
+            e.stopPropagation()
```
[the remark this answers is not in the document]
igorlesnenko: Changed approach in the next PR: https://github.com/ParabolInc/parabol/pull/8136/files#diff-8d7169a8176d285e712b14866607557aaaefec3c86afcabad180ee3424e302f3R34-R37

packages/client/components/RequestToJoinOrgNotification.tsx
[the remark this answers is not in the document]
igorlesnenko: Merged into this PR

packages/client/components/RequestToJoinOrgNotification.tsx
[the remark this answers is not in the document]
igorlesnenko: Merged into this PR

packages/client/mutations/toasts/mapRequestToJoinOrgToToast.ts
```diff
@@ -25,7 +28,9 @@ const mapRequestToJoinOrgToToast = (
     action: {
       label: 'Review',
       callback: () => {
-        // TODO: Implement review window and add segment events
```
tghanken, line 28: +1 I assume we're adding tracking in another PR once we finish the overall implementation?  I'd like to see something like a "Join Request Reviewed" event with an "Action" parameter ("Deny", "Add"), and maybe a list of teams the user was added to?

tghanken (approved):

packages/client/mutations/toasts/mapRequestToJoinOrgToToast.ts
igorlesnenko: Thanks, I will add this in next PR #8160

packages/client/components/ReviewRequestToJoinOrgModal.tsx
```diff
@@ -0,0 +1,124 @@
...
+  const {domainJoinRequest} = data.viewer
+
+  if (!domainJoinRequest) {
+    return null
+  }
+
+  const {createdBy, createdByEmail, domain} = domainJoinRequest
+
+  const teams = data.viewer.teams
+  const sortedTeams = teams
```
Dschoordsch, line 58: +1 Might be worth putting in
```ts
const sortedTeams = useMemo(() => teams
...
, [teams])

File: packages/client/components/ReviewRequestToJoinOrgModal.tsx [hunk 11]
```diff
@@ -0,0 +1,124 @@
...
+import DialogTitle from './DialogTitle'
+import PrimaryButton from './PrimaryButton'
+import SecondaryButton from './SecondaryButton'
+import {ReviewRequestToJoinOrgModalQuery} from '../__generated__/ReviewRequestToJoinOrgModalQuery.graphql'
+import Checkbox from './Checkbox'
+
+const query = graphql`
+  query ReviewRequestToJoinOrgModalQuery($requestId: ID!) {
+    viewer {
+      domainJoinRequest(requestId: $requestId) {
```
Dschoordsch, line 15: +1 I think we should put eligible teams directly on the request and do the server side filtering as you've suggested.

Dschoordsch (changes requested):
I think the popup logic does not work completely yet:
https://www.loom.com/share/e0e5c0950cda476fa585fb3fd0a761c4

igorlesnenko:
> I think the popup logic does not work completely yet:
https://www.loom.com/share/e0e5c0950cda476fa585fb3fd0a761c4

@Dschoordsch Missed this case.

I think it would be tricky to fix and probably requires implementing new approaches.

Currently I have to specify single background component for the modal to work. Probably we need to implement another approach for handling **global** modals

https://github.com/ParabolInc/parabol/pull/8086/files#diff-983791c0e4cc294a108b4e96900d0dde3d8f860b81f377b17a2c762900ff593dR53-R56

I can receive a snackbar notification while I'm on any page, I also can navigate between pages while my snackbar is active. We don't have a way to hide/show/receive particular snackbars on some particular routes.

Do you think this could be an issue created for v2, or should we implement solutions for v1?

packages/client/components/Dashboard.tsx
```diff
@@ -169,6 +173,10 @@ const Dashboard = (props: Props) => {
           </Switch>
           <Switch>
             <Route path='/new-meeting/:teamId?' component={NewMeetingRoot} />
+            <Route
```
Dschoordsch, line 176: -1 I think we should move this one up to the `PrivateRoute` level as the "modal" is shown at that level (i.e. also on top of meetings). Basically add a second `<Switch` there like here and modify the top one to also consider the `backgroundLocation`. This way we would also render it on top of meetings.

Dschoordsch (changes requested):
I think we just need to

Dschoordsch (approved):
Tested, works

Issue #7979, opened 2023-03-29
Public teams: make teams public


Dschoordsch:
To simplify discovery and increase awareness of Parabol usage, users should be able to see all teams of their company.

## Acceptance criteria
* [ ] add an [organization feature flag](https://github.com/ParabolInc/parabol/blob/57c0211e3e45bd3e7f207ce46b19defd4c43c1ec/packages/server/graphql/public/types/OrganizationFeatureFlags.ts#L3) for public teams
* [ ] modify the [`Organization.teams`](https://github.com/ParabolInc/parabol/blob/57c0211e3e45bd3e7f207ce46b19defd4c43c1ec/packages/server/graphql/types/Organization.ts#L72) resolver to return all teams when the feature flag is set
* [ ] modify `User.teams` to return all organization teams visible to the user based on the same feature flag
* [ ] ensure permissions in `Team` still only return data for members of that team

PR #8087, opened 2023-04-23
feat: add a feature flag for public teams

alphaofficial:
# Description
This adds a feature flag to make public teams visible to users

#7979 

## Demo

## Testing scenarios

- [ ] Scenario A
  - Add feature flag `publicTeam` to organisation
  - Users in any team in the organisation should be able to view other teams in the organisation

- [ ] Scenario B
  - Organisations without publicTeams flag should still have existing view permission

## Final checklist

### DIFF
--- a/packages/server/graphql/types/Organization.ts
+++ b/packages/server/graphql/types/Organization.ts
@@ -70,10 +70,12 @@ const Organization: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<a
     },
     teams: {
       type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(Team))),
-      description: 'all the teams the viewer is on in the organization',
+      description: 'all the teams the viewer is on and the public teams in the organization',
       resolve: async ({id: orgId}, _args: unknown, {authToken, dataLoader}) => {
-        const allTeamsOnOrg = await dataLoader.get('teamsByOrgIds').load(orgId)
-        const org = await dataLoader.get('organizations').load(orgId)
+        const [allTeamsOnOrg, org] = await Promise.all([
+          dataLoader.get('teamsByOrgIds').load(orgId),
+          dataLoader.get('organizations').load(orgId)
+        ])
 
         if (isSuperUser(authToken)) return allTeamsOnOrg
         else if (org.featureFlags?.includes('publicTeams')) {
--- a/packages/server/graphql/types/Organization.ts
+++ b/packages/server/graphql/types/Organization.ts
@@ -73,9 +73,13 @@ const Organization: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<a
       description: 'all the teams the viewer is on in the organization',
       resolve: async ({id: orgId}, _args: unknown, {authToken, dataLoader}) => {
         const allTeamsOnOrg = await dataLoader.get('teamsByOrgIds').load(orgId)
-        return isSuperUser(authToken)
-          ? allTeamsOnOrg
-          : allTeamsOnOrg.filter((team) => authToken.tms.includes(team.id))
+        const org = await dataLoader.get('organizations').load(orgId)
+
+        if (isSuperUser(authToken)) return allTeamsOnOrg
+        else if (org?.featureFlags?.includes('publicTeams')) {
+          return allTeamsOnOrg
+        }
+        return allTeamsOnOrg.filter((team) => authToken.tms.includes(team.id))
       }
     },
     tier: {
--- a/packages/server/graphql/types/Organization.ts
+++ b/packages/server/graphql/types/Organization.ts
@@ -76,7 +76,7 @@ const Organization: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<a
         const org = await dataLoader.get('organizations').load(orgId)
 
         if (isSuperUser(authToken)) return allTeamsOnOrg
-        else if (org?.featureFlags?.includes('publicTeams')) {
+        else if (org.featureFlags?.includes('publicTeams')) {
           return allTeamsOnOrg
         }
         return allTeamsOnOrg.filter((team) => authToken.tms.includes(team.id))
--- a/packages/server/graphql/types/User.ts
+++ b/packages/server/graphql/types/User.ts
@@ -562,7 +562,7 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
       ) => {
         const viewerId = getUserId(authToken)
         const [user, organizationUsers] = await Promise.all([
-          dataLoader.get('users').load(userId),
+          dataLoader.get('users').loadNonNull(userId),
           dataLoader.get('organizationUsersByUserId').load(userId)
         ])
         const orgIds = organizationUsers.map(({orgId}) => orgId)
--- a/packages/server/graphql/types/User.ts
+++ b/packages/server/graphql/types/User.ts
@@ -554,37 +554,41 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
     },
     teams: {
       type: new GraphQLNonNull(new GraphQLList(new GraphQLNonNull(Team))),
-      description: 'all the teams the user is on that the viewer can see.',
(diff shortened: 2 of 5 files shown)


packages/server/graphql/types/User.ts
```diff
@@ -562,11 +562,27 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
...
+        }
+
+        const isViewerOrSuperUser = viewerId === userId || isSuperUser(authToken)
+
+        const teamIds = isViewerOrSuperUser
+          ? user.tms
+          : user.tms.filter((teamId: string) => authToken.tms.includes(teamId))
+
         const teams = (await dataLoader.get('teams').loadMany(teamIds)).filter(isValid)
+        teams.push(...orgsWithPublicTeam)
```
nickoferrall, line 585: +1 `teams` are all of the teams the user belongs to, whereas `orgsWithPublicTeam` are all of the teams in the orgs that the user belongs to.

This wasn't immediately obvious to me - perhaps the variable names could be updated to make this clearer? I'd also rather have a final variable called something like `const allVisibleTeams = [...teams, ...orgsWithPublicTeam]` rather than using `.push`.

packages/server/graphql/types/User.ts
```diff
@@ -562,11 +562,27 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
       ) => {
```
nickoferrall, line 562: +1 The `teams` description is: `all the teams the user is on that the viewer can see.`, but we're returning all the teams in the orgs that the user belongs to.

I think it would be worth checking if the `User.teams` field is being used anywhere where the expected values are just the teams that the user is on. 

If that is the case, we should probably create a new field on the User object. If it's not the case, I'd suggest updating the description.

packages/server/graphql/types/User.ts
```diff
@@ -562,11 +562,27 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
...
+            const tms = await dataLoader.get('teamsByOrgIds').load(org.id)
+            orgsWithPublicTeam.push(...tms.filter((team) => !user.tms.includes(team.id)))
+          }
+        }
+
+        const isViewerOrSuperUser = viewerId === userId || isSuperUser(authToken)
+
+        const teamIds = isViewerOrSuperUser
+          ? user.tms
+          : user.tms.filter((teamId: string) => authToken.tms.includes(teamId))
```
nickoferrall, line 582: +1 we know `teamId` is a string so no need to add the type here

nickoferrall (changes requested):
This is looking good! I know the PR is just a draft, so my review might be overkill, but I wanted to share as much feedback as possible. 

Great stuff making progress so quickly, and thanks for sharing your work early.

packages/server/graphql/types/User.ts
nickoferrall: I like it! I'd be tempted to change the name of `teams` too to signal the difference, e.g. `teamsUserBelongsTo`, but I'm probably being too pedantic about naming

nickoferrall (approved):
LGTM! I'll request a review from Georg, who is a Maintainer. He's away until Tuesday, so you won't get a response until then, but feel free to pick up one of the other issues if you haven't already

packages/server/graphql/types/Organization.ts
```diff
@@ -73,9 +73,13 @@ const Organization: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<a
       description: 'all the teams the viewer is on in the organization',
       resolve: async ({id: orgId}, _args: unknown, {authToken, dataLoader}) => {
         const allTeamsOnOrg = await dataLoader.get('teamsByOrgIds').load(orgId)
-        return isSuperUser(authToken)
-          ? allTeamsOnOrg
-          : allTeamsOnOrg.filter((team) => authToken.tms.includes(team.id))
+        const org = await dataLoader.get('organizations').load(orgId)
```
Dschoordsch, line 76: +1 We could improve this part by putting both loads into a `Promise.all` and await both at once.

packages/server/graphql/types/User.ts
```diff
@@ -554,21 +554,41 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
...
+          const organization = await dataLoader.get('organizations').load(orgId)
+          if (organization.featureFlags?.includes('publicTeams')) {
+            const tms = await dataLoader.get('teamsByOrgIds').load(organization.id)
+            organizationTmsWithPublicFlag.push(
+              ...tms.filter((team) => !user?.tms.includes(team.id))
+            )
+          }
+        }
+
+        const isViewerOrSuperUser = viewerId === userId || isSuperUser(authToken)
```
Dschoordsch, line 582: +1 putting this into a separate variable is ok, but does not add much readability as it's only used once.

packages/server/graphql/types/User.ts
```diff
@@ -562,11 +562,27 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
...
-        const teamIds =
-          viewerId === userId || isSuperUser(authToken)
-            ? user.tms
-            : user.tms.filter((teamId: string) => authToken.tms.includes(teamId))
+        const organizationsForUser = await dataLoader.get('organizationUsersByUserId').load(userId)
+        const orgIds = organizationsForUser.map(({orgId}: OrganizationUserType) => orgId)
+
+        const orgsWithPublicTeam = []
+
+        for (const orgId of orgIds) {
```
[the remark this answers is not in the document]
Dschoordsch: -1 I think what Nick meant is doing something like
```ts
const publicTeams = await Promis.all(orgIds.map(async (orgId) => {
    ...
}))
```

packages/server/graphql/types/User.ts
```diff
@@ -554,21 +554,41 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
...
-        const user = (await dataLoader.get('users').load(userId))!
-        const teamIds =
-          viewerId === userId || isSuperUser(authToken)
-            ? user.tms
-            : user.tms.filter((teamId: string) => authToken.tms.includes(teamId))
-        const teams = (await dataLoader.get('teams').loadMany(teamIds)).filter(isValid)
-        teams.sort((a, b) => (a.name > b.name ? 1 : -1))
-        return teams
+        const [user, organizationUsers] = await Promise.all([
+          dataLoader.get('users').load(userId),
```
Dschoordsch, line 565: +1 we could use `loadNonNull` here so we always get a valid user (or it throws)

packages/server/graphql/types/User.ts
```diff
@@ -554,21 +554,41 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
...
+          dataLoader.get('organizationUsersByUserId').load(userId)
+        ])
+        const orgIds = organizationUsers.map(({orgId}) => orgId)
+
+        const organizationTmsWithPublicFlag = []
+
+        for (const orgId of orgIds) {
+          const organization = await dataLoader.get('organizations').load(orgId)
+          if (organization.featureFlags?.includes('publicTeams')) {
```
Dschoordsch, line 574: -2 I did not test this, but for permissions we previously checked that the `viewer` (the one making the request) can only see the teams they're already on.
Right now it seems that I could see the teams I'm already on plus all the teams from organizations with the public feature flag, regardless of whether or not I'm on that org.

Dschoordsch (changes requested):

Dschoordsch (changes requested):

Dschoordsch (approved):

packages/server/graphql/types/User.ts
Dschoordsch: I couldn't let it go and tried again and could reproduce the error I was seeing in the code: https://www.loom.com/share/e3fb1b29958d4da08166a3f965a17759

Dschoordsch (changes requested):

mattkrick:

Issue #8092, opened 2023-04-24
Standup Recurrence: Reuse last meeting's prompt


jmtaber129:
We currently use the same default prompt for every new team prompt meeting (see https://github.com/ParabolInc/parabol/blob/80ceece66013f5f3b976a41fedd66e7f3b806d1d/packages/server/graphql/mutations/helpers/safeCreateTeamPrompt.ts#L40).

For recurring standups, we should reuse the last standup's prompt.

We _may_ want to reuse the team's last standup prompt for non-recurring meetings, too.

AC:
- [ ] When a recurring standup meeting restarts, it keeps the previous meeting's prompt
- [ ] (Optional, TBD) When a non-recurring standup is created, default to the team's most recent standup's prompt.

PR #8095, opened 2023-04-24
feat(recurrence): Use the prompt from the most recent meeting in the series

jmtaber129:
# Description
Fixes https://github.com/ParabolInc/parabol/issues/8092

Currently, we use the hard-coded default prompt for all standup meetings. Instead, when restarting a recurring meeting, just use the prompt from the most recent meeting in the series.

## Demo
https://www.loom.com/share/fe44272fbaaa4b0a9f416138cd7ed689

## Testing scenarios
- [ ] Smoke test creating standups
- [ ] Create recurring standup meeting
- [ ] Change the prompt
- [ ] Time travel and run process recurrence
- [ ] Confirm new meeting uses previous meeting's prompt.

## Final checklist

### DIFF
--- a/packages/server/graphql/private/mutations/processRecurrence.ts
+++ b/packages/server/graphql/private/mutations/processRecurrence.ts
@@ -17,6 +17,7 @@ import {MutationResolvers} from '../resolverTypes'
 
 const startRecurringTeamPrompt = async (
   meetingSeries: MeetingSeries,
+  lastMeeting: MeetingTeamPrompt | null,
   startTime: Date,
   dataLoader: DataLoaderWorker,
   r: ParabolR,
@@ -37,7 +38,8 @@ const startRecurringTeamPrompt = async (
   )
   const meeting = await safeCreateTeamPrompt(meetingName, teamId, facilitatorId, r, dataLoader, {
     scheduledEndTime: nextMeetingStartDate,
-    meetingSeriesId: meetingSeries.id
+    meetingSeriesId: meetingSeries.id,
+    meetingPrompt: lastMeeting ? lastMeeting.meetingPrompt : undefined
   })
 
   await r.table('NewMeeting').insert(meeting).run()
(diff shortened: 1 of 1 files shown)


packages/server/graphql/private/mutations/processRecurrence.ts
```diff
@@ -37,7 +38,8 @@ const startRecurringTeamPrompt = async (
   )
   const meeting = await safeCreateTeamPrompt(meetingName, teamId, facilitatorId, r, dataLoader, {
     scheduledEndTime: nextMeetingStartDate,
-    meetingSeriesId: meetingSeries.id
+    meetingSeriesId: meetingSeries.id,
+    meetingPrompt: lastMeeting ? lastMeeting.meetingPrompt : undefined
```
BartoszJarocki, line 42: +1 I wonder, how the `lastMeeting` can be null here? If recurrence is running, it means there's already at least one meeting and there will always be a prompt to reuse. 

If that's true, I think it'd be a good time to cleanup this, as it'd be used when first meeting is created. It'd be cleaner to have a const somewhere with default prompt and always provide it when creating an initial meeting ie. make `prompt` required parameter of `safeCreateTeamPrompt`.
https://github.com/ParabolInc/parabol/blob/287167069c78ac7f32b0f8ad9d0e05b020c5a410/packages/server/graphql/mutations/helpers/safeCreateTeamPrompt.ts#L40 

Also, if the `lastMeeting` is `null` in the case above, we'd override the prompt provided by constructor with `undefined`. So it kinda works, because it'll never happen

jmtaber129: There _should_ always be a `lastMeeting`, but there are no DB constraints that enforce this, so it's not guaranteed.

I extracted the default prompt and used that instead of `undefined`.

BartoszJarocki (approved):

PR #8103, opened 2023-04-26
chore: remove pg.d.ts from git, build locally

mattkrick:
# Description

`pg.d.ts` is a file that is automatically generated from our postgres schema.
generated files shouldn't be kept in version control. they should be built just-in-time so we don't waste time with merge conflicts. I goofed up & checked this file in.

This PR builds that file whenever `yarn dev` is called. it also builds it in our GH pipeline.

## Testing scenarios

- [ ] run `yarn dev` see `pg.d.ts` is created
- [ ] see the CI. `pg.d.ts` is created & tsc doesn't yell at us

TODO:
- the `source` bash command doesn't exist in GH CI...


Dschoordsch:
I would like to keep this file checked in. This will decrease build complexity in the long run because we don't need the DB anymore during build.
We removed the generated files from pg-typed because they were not deterministic and resulted in diffs without changes. I would keep the definitions in for some time to observe if we run into the same issue.
If we can keep it, I would just add another build check to ensure the checked-in version is up to date.

Dschoordsch:
Yes, let's go forward with kysely, convert one or two of the existing database queries and observe how it goes for a month or so. If there are no red flags, let's get rid of pg-typed.

mattkrick:
Closing this in favor of https://github.com/ParabolInc/parabol/issues/8113. I'll create a new PR that only includes the `pg:generate` call in pm2.dev

Issue #8099, opened 2023-04-25
RiD: Distinct Edit Mode


jmtaber129:
Currently, we follow the existing pattern for editing templates, in which all fields are implicitly editable if the user is an owner of the template.

The RiD designs call for an edit mode, in which the user clicks an edit button to make the fields editable and the template deletable, and then clicks "Done Editing" when the user no longer wants to edit the template.

[Figma](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?node-id=571%3A25684&t=eaBy1MH5b54fmu7Q-1)

For simplicity, any edits made in edit mode will persist immediately, rather than persisting when the user exits edit mode.

AC:
- [ ] No changes to activity details for non-owned templates
- [ ] For owned templates, nothing is editable by default
- [ ] For owned templates, there's a pencil button matching designs that enters edit mode
- [ ] When entering edit mode, that sidebar animates out smoothly
- [ ] In edit mode:
  - [ ] all fields are editable, and all edits persist immediately
  - [ ] a delete button appears
  - [ ] a "done editing" button is present in a bottom bar
- [ ] Clicking "done editing" exits edit mode

PR #8108, opened 2023-04-26
feat(rid): Edit Mode

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8099

Needs rebase once https://github.com/ParabolInc/parabol/pull/8111 lands.

Edit mode for RiD Details View. Clicking the pencil icon on owned templates makes all fields editable.

Note that all edits persist immediately (rather than after exiting edit mode). We may want to change this behavior in a future iteration.

## Demo
https://www.loom.com/share/257162a6f3eb4287a40ad80645d631a0

## Testing scenarios
- [ ] No changes to activity details for non-owned templates
- [ ] For owned templates, nothing is editable by default
- [ ] For an owned template, click the pencil button to enter edit mode
- [ ] When entering edit mode, that sidebar animates out smoothly
- [ ] In edit mode, edit all fields
- [ ] Click "done editing" to exit edit mode
- [ ] Enter edit mode, and delete the template

## Final checklist

### DIFF
--- a/packages/client/components/ActivityLibrary/ActivityDetails.tsx
+++ b/packages/client/components/ActivityLibrary/ActivityDetails.tsx
@@ -301,12 +301,14 @@ const ActivityDetails = (props: Props) => {
               </div>
             </div>
           </div>
-          {!isEditing && (
-            <ActivityDetailsSidebar selectedTemplateRef={selectedTemplate} teamsRef={teams} />
-          )}
+          <ActivityDetailsSidebar
+            selectedTemplateRef={selectedTemplate}
+            teamsRef={teams}
+            isOpen={!isEditing}
+          />
         </div>
         {isEditing && (
-          <div className='flex h-20 items-center justify-center bg-slate-200'>
+          <div className='absolute bottom-0 flex h-20 w-full items-center justify-center bg-slate-200'>
             <button
               onClick={() => setIsEditing(false)}
               className='w-max cursor-pointer rounded-full bg-sky-500 px-10 py-3 text-center font-sans text-base text-lg font-semibold text-white hover:bg-sky-600'
--- a/packages/client/components/ActivityLibrary/ActivityDetails.tsx
+++ b/packages/client/components/ActivityLibrary/ActivityDetails.tsx
@@ -172,122 +172,132 @@ const ActivityDetails = (props: Props) => {
               </Link>
               <div className='w-max text-xl font-semibold'>Start Activity</div>
             </div>
-            <div className='flex w-full flex-col justify-start pl-4 pr-14 xl:flex-row xl:justify-center xl:pl-14'>
-              <ActivityCard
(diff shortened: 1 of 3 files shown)


packages/client/components/ActivityLibrary/ActivityDetails.tsx
```diff
@@ -156,115 +158,166 @@ const ActivityDetails = (props: Props) => {
...
           </div>
+          <ActivityDetailsSidebar
+            selectedTemplateRef={selectedTemplate}
+            teamsRef={teams}
+            isOpen={!isEditing}
+          />
         </div>
-        <ActivityDetailsSidebar selectedTemplateRef={selectedTemplate} teamsRef={teams} />
+        {isEditing && (
+          <div className='absolute bottom-0 flex h-20 w-full items-center justify-center bg-slate-200'>
```
BartoszJarocki, line 311: -1 there's no explicit `relative` element in this view so we shouldn't be using `absolute` and `bottom-0` here. this element sits in `flex` container so it's better to just use `mt-auto` without `absolute` and `bottom-0`

packages/client/components/ActivityLibrary/ActivityDetails.tsx
```diff
@@ -156,115 +158,166 @@ const ActivityDetails = (props: Props) => {
...
+            <button
+              onClick={() => setIsEditing(false)}
+              className='w-max cursor-pointer rounded-full bg-sky-500 px-10 py-3 text-center font-sans text-base text-lg font-semibold text-white hover:bg-sky-600'
```
BartoszJarocki, line 314: -1 you're using two different utility classes for the same css prop here `text-base text-lg`

BartoszJarocki (commented):
two little things to clean up, looks good otherwise!

packages/client/components/ActivityLibrary/ActivityDetails.tsx
jmtaber129: This is intentional so that this element sits above the sidebar during the animation. If this is positioned normally, the sidebar will jump a bit as this element displaces it.

packages/client/components/ActivityLibrary/ActivityDetails.tsx
jmtaber129: Oops, nice catch!

packages/client/components/ActivityLibrary/ActivityDetails.tsx
BartoszJarocki: ah I see. what about setting root div to have positon `relative`? without that element will be implicititly positioned to nearest element with position `relative` or root (body). it'll work until someone adds a div that'll be somewhere in the hierarchy and could break this view without knowing

jmtaber129: Ah gotcha. Added `relative` to the root div in this component

BartoszJarocki (approved):
lgtm! 🚢

PR #8114, opened 2023-04-26
chore: generate kysely schema in dev

mattkrick:
# Description

Successor to #8103

Call kysely codegen whenever dev starts


mattkrick:
this part of #8103 got approved so going forward with it

PR #8116, opened 2023-04-27
fix(chore): properly count emojis length

### LINKED ISSUE #8098: 🐈‍⬛ is too many characters for Sprint Poker, but 🐕 is not
## Issue - Bug

When trying to create a custom scale in Sprint Poker, I wasn't able to add 🐈‍⬛ as a value because it is too many characters. I wasn't sure which emoji could be added and which couldn't. 

Screenshot 2023-04-24 at 2 24 33 PM

- **How serious is the bug?** P0 - P4 - This is not a serious bug, probably very few folks are running into this
- **OS:** macOS / Linux / Windows / ?
- **Browser:** Chrome / Safari / Lynx / ? + version
- **Node version:** `node --version`
- **NPM version:** `npm --version`
- **RethinkDB version:** `rethinkdb --version`

Users can add all emoji as a value, OR can understand easily which emoji could be added. 

tianrunhe:
# Description

Fixes #8098 

## Demo

Screenshot 2023-04-27 at 14 47 04

## Testing scenarios

- [ ] Emojis
  - Create a new Poker Scale
  - Add a new value 🐈‍⬛
  - Verify there is no client or server side error

- [ ] Allow max 3 characters
  - Create a new Poker Scale
  - Add a new value XXS
  - Verify there is no client or server side error

## Final checklist


tianrunhe:
@nickoferrall This works on my desktop but I'm not sure why it fails on build, probably something wrong with package dependencies. Can you help me build on top of this PR and move it forward (as I'm out for next week)? Thank you!

nickoferrall:
Love the quick fix! 

> Can you help me build on top of this PR and move it forward (as I'm out for next week)

@tianrunhe I'm away tomorrow, Monday, and Wednesday and still have a fair bit of work left in the Sprint. As this is p3, I think it's fine to leave it until you're back!

tianrunhe:
@nickoferrall I fixed the dependency issue and now it's ready for review! 🙏

nickoferrall:
Some other work took longer than expected today, so I'll review this PR with a fresh mind tomorrow

packages/client/validation/Legitity.ts
```diff
@@ -37,9 +38,8 @@ class Legitity {
 
   max(len: number, msg?: string) {
     if (this.value !== undefined) {
-      // this.value.length gives us the count of UTF-16 units, so 🔥 has a length of 2
-      // Spreading the string into an array gives us the desired length in codepoints (characters): https://stackoverflow.com/a/54369605
-      const value = [...this.value]
+      // https://stackoverflow.com/a/46085147
+      const value = _.toArray(this.value)
```
nickoferrall, line 42: I lean towards using `Array.from` as it also accurately counts emoji length and would mean we wouldn't need to install the fairly chunky lodash library. 

One edge case though seems to be combined emojis, like 👪, which is really three emojis in one. For that, lodash seems like the better option. 

No strong preference from me!

nickoferrall (approved):
LGTM!

packages/client/validation/Legitity.ts
tianrunhe: The black cat doesn't work either...
```
> Array.from('🐈‍⬛')
< (3) ['🐈', '‍', '⬛']
```

nickoferrall: Interesting! Black cat is actually multiple emojis too 😂

tianrunhe:
@nickoferrall I was able to only import the needed `lodash` function without installing the entire package. However that seems to make `yarn.lock` changes quite a lot. I'm not a professional in dependency management. WDYT?

nickoferrall (approved):
I think that just importing toArray from lodash is the right move, and it seems like there are enough edge cases to warrant using a third-party package. No need to worry about yarn.lock, it'll be regenerated using package.json everytime people run `yarn`

tianrunhe:
> @nickoferrall I was able to only import the needed `lodash` function without installing the entire package. However that seems to make `yarn.lock` changes quite a lot. I'm not a professional in dependency management. WDYT?

It's an issue with `yarn.lock` on master branch. Matt fixed it and I merged with master again. Now it looks good!

Dschoordsch (approved):
It would be nice if we could use [`Intl.Segmenter`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter) which is sadly not available on Firefox yet and would require a polyfill.
Your version works, however there is some small UI issue with very long dimensions
Screenshot from 2023-05-22 20-05-55
I would let this slip as it's very unlikely that the hidden part is necessary for the team to distinguish the dimensions.

Issue #8090, opened 2023-04-24
TypeError: Cannot read properties of undefined (reading 'getLinkedRecord')


Dschoordsch:
https://parabol.sentry.io/issues/4079926867/?referrer=github_plugin

```
TypeError: Cannot read properties of undefined (reading 'getLinkedRecord')
  at acceptTeamInvitationNotificationUpdater (./packages/client/mutations/AcceptTeamInvitationMutation.ts:110:24)
  at value (./packages/client/subscriptions/subscriptionUpdater.ts:1:440)
  at fn (./packages/client/subscriptions/subscriptionUpdater.ts:17:51)
  at _optionalChain (./packages/client/subscriptions/subscriptionUpdater.ts:1:418)
  at updater (./packages/client/subscriptions/subscriptionUpdater.ts:17:5)
...
(17 additional frame(s) were not displayed)
```

Dschoordsch:
88 events, 48 users

jmtaber129:
I can reproduce this locally with the following steps:
* Open two windows with Parabol logged into User A's account
* Navigate to a meeting User A doesn't have access to.
* As user B, approve of User A joining the team
* The main window of User A behaves properly, but the other window throws the error.'

I haven't tracked down the root cause, but this likely only recently surfaced because we were previously always checking that the `payload` passed to `acceptTeamInvitationNotificationUpdater` was non-null, but this is not guaranteed with the current subscription updater added as part of the relay upgrade (see https://github.com/ParabolInc/parabol/blob/05f27d1510368eec2d60e0b861071ed51f0c7ba6/packages/client/subscriptions/subscriptionUpdater.ts#L15-L17).

We could change the `subscriptionUpdater` to check that `field` is non-null as a bandaid solution

PR #8118, opened 2023-04-27
fix: Accepting team invitation updates correctly on other tabs

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8090

See https://github.com/ParabolInc/parabol/issues/8090#issuecomment-1525861778 for additional context.

## Testing scenarios
See https://github.com/ParabolInc/parabol/issues/8090#issuecomment-1525861778 for repro steps.
- [ ] Run repro steps on `master` and confirm issue is present
- [ ] Run repro steps on this PR and confirm issue is resolved.

## Final checklist

### DIFF
--- a/packages/client/subscriptions/subscriptionUpdater.ts
+++ b/packages/client/subscriptions/subscriptionUpdater.ts
@@ -13,6 +13,7 @@ const subscriptionUpdater =
     if (!payload) return
     const fieldName = payload.getValue('fieldName')
     const field = payload.getLinkedRecord(fieldName)
+    if (!field) return
     const handler = updateHandlers[fieldName]
     handler?.(field, {atmosphere, store})
   }


mattkrick:
yikes! what a bug. `field` should always be present, if it's not, we've got a bigger problem on our hands & this would hide that problem.

Would you mind digging into this a little more so we can find a root cause?

mattkrick:
tried to reproduce this, looks like user A (mutator) gets a server error, user A (subscriber) gets the error mentioned, and user B gets another error! 

The server error looks like this: `Cannot return null for non-nullable field Team.(agendaItems,agendaItems)` which means the server isn't using the updated JWT

mattkrick:
whoaaaa root cause was the fragment `AcceptTeamInvitationMutation_notification` was never included in the notification subscription even though we called its updater!
It never threw an error before the relay upgrade because it was silently failing. That means anyone who accepted a team invitation would only accept it for their first tab, not their other open tabs/browsers/clients! This must've been an error for _years_!
 
Now with the relay upgrade & the return payload is nested, this error made it past the nullish `payload` check.

jmtaber129:
Closing in favor of https://github.com/ParabolInc/parabol/pull/8127

Issue #8026, opened 2023-04-11
RiD: Team selection on team-scoped templates


jmtaber129:
Design issue - https://github.com/ParabolInc/parabol/issues/7955

Team-scoped templates can only be used to create activities on the owner team, so we want a special case for the team selection dropdown when the template is team-scoped explaining the issue + prompting the user to share to the org.

As part of this, we should allow all team members (not just team lead) to change template visibility

[Figma](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?node-id=915%3A26502&t=7C8RKY6hckMOKINM-1)

AC:
- [ ] When a template is team-scoped, clicking the team dropdown shows a popover instead of the normal team selection menu
- [ ] The popover includes the desired copy + a button prompting the user to change the scope to "ORGANIZATION"
- [ ] Clicking the button updates the template scope

mattkrick:
Convo doc: https://www.notion.so/parabol/UX-for-changing-sharing-scope-of-templates-e428dc2538de411c94a34b81adfe9f88

PR #8120, opened 2023-04-27
feat(rid): Prompt user to change template scope to org

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8026

When a template is team-scoped, the activity can only be started on the owning team, so the team dropdown in the details view would only show the owning team.

Instead, show a popover that prompts the user to change the scope to "org".

This loosens the template scope update mutation such that any team member can update the scope (not just team leads)

## Demo
https://www.loom.com/share/0f4486618b4344a2a5924b281befb361

## Testing scenarios
- [ ] Team-scoped template:
  - [ ] Clicking the team selector shows the popover
  - [ ] Clicking the button in the popover updates the scope, and replaces the popover with the team selector menu for teams in the org.
- [ ] Org-scoped templates always show the normal team dropdown
- [ ] Non-team-leads can change template scope.

## Final checklist

### DIFF
--- a/packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
+++ b/packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
@@ -120,6 +120,29 @@ const ActivityDetailsSidebar = (props: Props) => {
     )
   }
 
+  const teamScopePopover = templateTeam && selectedTemplate.scope === 'TEAM' && (
+    <div className='w-[352px] p-4'>
+      <div>
+        This custom activity is private to the <b>{templateTeam.name}</b> team.
+      </div>
+      <br />
+      <div>
+        As a member of the team you can share this activity with the{' '}
+        <b>{templateTeam.organization.name}</b> organization so that any team can also use the
+        activity.
+      </div>
+      <button
+        onClick={handleShareToOrg}
+        className={
+          'mt-4 flex w-max cursor-pointer items-center rounded-full border border-solid border-slate-400 bg-white px-3 py-2 text-center font-sans text-sm font-semibold text-slate-700 hover:bg-slate-100'
+        }
+      >
+        <LockOpen style={{marginRight: '8px', color: PALETTE.SLATE_600}} />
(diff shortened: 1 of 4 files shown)


packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
```diff
@@ -93,6 +99,37 @@ const ActivityDetailsSidebar = (props: Props) => {
...
+
+  const teamScopePopover = templateTeam && selectedTemplate.scope === 'TEAM' && (
+    <div className='w-[352px] p-4'>
+      <div>
+        This custom activity is private to the <b>{templateTeam.name}</b> team.
+      </div>
+      <br />
+      <div>
+        As a member of the team you can share this activity with the{' '}
+        <b>{templateTeam.organization.name}</b> organization so that any team can also use the
```
mattkrick, line 118: +1 nit on the word "any" and below on "all", which some less technically-inclined folks may interpret to mean "all teams [in the app]".
Maybe "allow teams in the X organization..."?

jmtaber129: Hmm, I could go either way.

@ackernaut this is the copy from your designs, WDYT?

ackernaut: What do you think about this [iteration on copy in Notion](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?type=design&node-id=1381%3A32031&t=ChKuSk6LsUmoIzoT-1)

jmtaber129: @ackernaut that copy LGTM! I'd just change "can use it" to "can use the activity" to be specific, similar to the previous copy.

jmtaber129:
New copy:
Screen Shot 2023-05-04 at 1 55 10 PM

PR #8127, opened 2023-04-28
fix: subscribe to acceptInvite paylods in subscription

### LINKED ISSUE #8090: TypeError: Cannot read properties of undefined (reading 'getLinkedRecord')

mattkrick:
# Description

fixes #8090 
builds on top of #8118 & the great repro that @jmtaber129 found

I also added in a fix to remove null values from the payload. I must've had a bad merge conflict because that used to be there 😅 Makes this stuff easier to debug & generally keeps payload size to 1 MTU, so mobile clients should be happy.

Worth noting that there could be other similar bugs like this out there!

## Demo

https://www.loom.com/share/733bcaafd1aa4f9a972cda8f62de4c48

## Testing scenarios

- [ ] See #8090 for a great repro!
- [ ] Look at the incoming subscription payload. it doesn't include an object with a bunch of nulls, it's just the fieldName & then 1 other prop
- [ ] both tabs of the user accepting the invite have access to the team without a refresh

### DIFF
--- a/packages/client/subscriptions/NotificationSubscription.ts
+++ b/packages/client/subscriptions/NotificationSubscription.ts
@@ -1,13 +1,14 @@
 import graphql from 'babel-plugin-relay/macro'
 import {RouterProps} from 'react-router'
 import {requestSubscription} from 'relay-runtime'
-import {archiveTimelineEventNotificationUpdater} from '~/mutations/ArchiveTimelineEventMutation'
-import {endCheckInNotificationUpdater} from '~/mutations/EndCheckInMutation'
-import {endRetrospectiveNotificationUpdater} from '~/mutations/EndRetrospectiveMutation'
 import {InvalidateSessionsMutation_notification$data} from '~/__generated__/InvalidateSessionsMutation_notification.graphql'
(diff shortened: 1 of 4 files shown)


jmtaber129 (approved):
+1 Should we log and/or throw an error when the field for `fieldName` is absent so we can catch this sort of invariant violation earlier? I can imagine a similar bug being reintroduced in the future, so a clear warning in dev might prevent this.

mattkrick:
that is a great idea!
added an error for both updaters & onNext handlers

Issue #7966, opened 2023-03-28
RiD Details View: Poker Sidebar


jmtaber129:
Implement the sidebar for the RiD details view for poker

[Figma](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?t=xlIwfN60WzTIJ9te-0) (TODO: confirm when these designs are finalized)

AC:
- [ ] Sidebar matches designs
- [ ] Team selection behavior for non-public templates is reasonable
- [ ] Can start a new poker meeting

PR #8131, opened 2023-05-02
feat(rid): Poker details view

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/7966 fixes https://github.com/ParabolInc/parabol/issues/7965

Implement Poker Details view, including:
* Non-editing view
* Edit-mode, including changing dimensions, scales, etc., as well as deletion, sharing, etc.
* Cloning
* Sidebar settings
* Starting poker meetings

There are a number of refactors that could simplify this code as a whole (e.g. combining poker+reflect template mutations, moving global state into component state, etc.), but none are worth doing in the near term. We should consider these refactors post-GA.

## Demo
https://www.loom.com/share/020962d346934f3eb194f8c1649cfa16

## Testing scenarios
- [ ] Designs are correct when not in edit mode for owner + non-owner
- [ ] Can clone for owner + non-owner, and cloned activity appears in activity grid view without refresh
- [ ] Edit-mode
  - [ ] Dimensions
    - [ ] Add new dimension
    - [ ] Rename dimension
    - [ ] Move dimensions via drag
    - [ ] Remove dimension
    - [ ] Scales
      - [ ] Change scale
      - [ ] Create scale
      - [ ] Clone scale
      - [ ] Edit scale
      - [ ] Delete scale
  - [ ] Change sharing
  - [ ] Delete template
- [ ] Sidebar
  - [ ] Change team
  - [ ] Change icebreaker setting
  - [ ] Start poker meeting

## Final checklist


packages/client/components/ActivityLibrary/ActivityDetails.tsx
```diff
@@ -108,24 +121,63 @@ const ActivityDetails = (props: Props) => {
...
+  const {
+    openPortal: openPokerTemplateScaleDetailsPortal,
+    modalPortal: pokerTemplateScaleDetailsPortal,
+    closePortal: closePokerTemplateScaleDetailsPortal
+  } = useModal({
+    id: 'pokerTemplateScaleDetailsModal'
+  })
+
+  useEffect(() => {
+    if (selectedTemplate?.team.editingScaleId) {
```
jmtaber129, line 170: It might be nice to keep track of `editingScaleId` in component state instead of global client-side relay state, but a refactor for this seems premature, so I'm simply reusing the existing global state + logic for simplicity. We should consider this refactor if/when we fully adopt the activity library.

packages/client/components/ActivityLibrary/ActivityDetails.tsx
```diff
@@ -144,11 +196,18 @@ const ActivityDetails = (props: Props) => {
 
   const [isEditing, setIsEditing] = useState(false)
 
+  useEffect(
+    () =>
+      selectedTemplate &&
+      setActiveTemplate(atmosphere, selectedTemplate.teamId, templateId, selectedTemplate.type),
```
jmtaber129, line 202: There's some logic around dimensions + scale modifications that assume an active template in client-side global relay state. We should consider refactoring this in the long term, but simply setting the active template when the template changes is a reasonable quick-and-dirty solution in the near-term.

mattkrick:
In editing mode, the "Done editing" footer floats on top of the dimensions (for limited viewport heights).
Instead could we make it a `flex-end`, and just take up the remaining space? that way we can scroll dimensions.

Screenshot 2023-05-03 at 10 56 47 AM

Should clicking "Clone and edit" go into edit mode after selecting the team? If not, should we just call it "clone"? 

I created https://github.com/ParabolInc/parabol/issues/8145 because creating a new scale should default to using that new scale they just made. You did it right by keeping the behavior unchanged

when deleting an estimation template, could we redirect to /estimation? instead of /recommended?

packages/client/components/ActivityLibrary/ActivityDetails.tsx
```diff
@@ -320,17 +415,23 @@ const ActivityDetails = (props: Props) => {
...
+          templatesRef={availableTemplates.edges.map((edge) => edge.node)}
+          closePortal={closeTeamPickerPortal}
           parentTemplateId={selectedTemplate.id}
+          type={type}
         />
       )}
+      {type === 'poker' &&
+        selectedTemplate.team.editingScaleId &&
+        pokerTemplateScaleDetailsPortal(
+          <div className='w-[520px]'>
```
mattkrick, line 431: +1 do we have an issue for making these view mobile friendly? Anything over ~320px will probably take some extra consideration

packages/client/mutations/AddPokerTemplateMutation.ts
```diff
@@ -34,7 +35,7 @@ export const addPokerTemplateTeamUpdater: SharedUpdater<AddPokerTemplateMutation
   const template = payload.getLinkedRecord('pokerTemplate')
   if (!template) return
   const templateId = template.getValue('id')
-  handleAddPokerTemplate(template, store)
+  handleAddMeetingTemplate(template, 'poker', store)
```
mattkrick, line 38: +2 nice refactor

packages/client/components/ActivityLibrary/ActivityDetails.tsx
jmtaber129: I'm not too worried about mobile here, since so few `new-meeting` views occur on mobile, and the existing mobile UX for editing templates like this is pretty bad.

jmtaber129:
@mattkrick 
> Instead could we make it a flex-end, and just take up the remaining space?

This is tricky because of the animation between non-editing and editing - when the sidebar animates out, we want the bottom bar to appear without either component "jumping". See original edit mode conversation: https://github.com/ParabolInc/parabol/pull/8108#discussion_r1179526841

To address this, I made the following changes:
* Changed the bottom bar position to `fixed`
* Added bottom padding for the details equal to the height of the bottom bar (not especially elegant, but it works)
  * I made this unconditional so dimensions/prompts don't go all the way to the very bottom of the window outside of edit-mode.
* Applied similar changes to the sidebar to keep it in place when scrolling

I played around a bit with some overflow tailwind classes, but I couldn't quite get anything to work. Def open to alternatives though.

> Should clicking "Clone and edit" go into edit mode after selecting the team?

Hmm probably, though this is more related to edit mode in general than this details view specifically. I've created https://github.com/ParabolInc/parabol/issues/8151 to track this.

> when deleting an estimation template, could we redirect to /estimation? instead of /recommended?

Yep! Fixed

mattkrick:
> Added bottom padding for the details equal to the height of the bottom bar

Great approach. This is how the left nav drawer works. Also plays nicely using `transform` instead of `height` so framerate stays high

Issue #8123, opened 2023-04-27
Suggest Groups: click button to form groups


nickoferrall:
In the group phase of a retro meeting, a button called "Suggest Groups" exists. Clicking the button groups the reflections together. 

The user will need to refresh the page to see the groups update. 

### AC

- [ ] "Suggest Groups" button exists in the Group phase of a retro
- [ ] Click the button, refresh the page, and see the reflection groups are formed using OpenAI
- [ ] Users that don't have access to OpenAI don't see the button

Issue #8126, opened 2023-04-27
Suggest Groups: update groups without refresh


nickoferrall:
Predecessor: https://github.com/ParabolInc/parabol/issues/8123

### AC 

- [ ] Add the subscription
- [ ] After the viewer clicks "Suggest Groups", the groups update for all other users without needing to refresh the page

Issue #8124, opened 2023-04-27
Suggest Groups: add reset groups button


nickoferrall:
Predecessor: https://github.com/ParabolInc/parabol/issues/8123

After clicking the "Suggest Groups" button, the user should be able to reset the groups if they don't like them

### AC

- [ ] After clicking "Suggest Groups" button, it is replaced by a "Reset Groups" button
- [ ] Clicking the button reverts the AI groups

PR #8132, opened 2023-05-02
feat(suggest-groups): click button to group reflections

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8123

Demo: https://www.loom.com/share/c453d1b30fca44c7b74066b64fe81343

Note:
- If you quickly click the "Suggest Groups" button after proceeding to the Group phase, it won't work as ChatGPT won't have responded. In a later PR, I'll add some error/loading feedback text. See: https://github.com/ParabolInc/parabol/issues/8133
- Non-viewers will need to refresh the page to see the change. See https://github.com/ParabolInc/parabol/issues/8126

### AC

- [ ] Add the org feature flag 
- [ ] Create a retro and add several reflections
- [ ] Move to the Group phase and see the "Suggest Groups" button
- [ ] Wait ~30 seconds, click on "Suggest Groups", and see that the reflections are grouped

### DIFF
--- a/packages/client/components/RetroGroupPhase.tsx
+++ b/packages/client/components/RetroGroupPhase.tsx
@@ -63,7 +63,6 @@ const RetroGroupPhase = (props: Props) => {
   const {id: meetingId, endedAt, showSidebar, organization} = meeting
   const {featureFlags} = organization
   const {suggestGroups} = featureFlags
-  console.log('🚀 ~ organization:', organization)
 
   const handleAutoGroupClick = () => {
     AutogroupMutation(atmosphere, {meetingId}, {onError, onCompleted})
@@ -82,11 +81,11 @@ const RetroGroupPhase = (props: Props) => {
         </MeetingTopBar>
         <PhaseWrapper>
           <StageTimerDisplay meeting={meeting} canUndo={true} />
-          {suggestGroups && (
-            <ButtonWrapper>
-              <PrimaryButton onClick={handleAutoGroupClick}>{'Suggest Groups ✨'}</PrimaryButton>
-            </ButtonWrapper>
-          )}
+          {/* {suggestGroups && ( */}
+          <ButtonWrapper>
+            <PrimaryButton onClick={handleAutoGroupClick}>{'Suggest Groups ✨'}</PrimaryButton>
+          </ButtonWrapper>
+          {/* )} */}
           <MeetingPhaseWrapper>
             <GroupingKanban meeting={meeting} phaseRef={phaseRef} />
           </MeetingPhaseWrapper>
--- a/packages/client/components/RetroGroupPhase.tsx
+++ b/packages/client/components/RetroGroupPhase.tsx
@@ -3,7 +3,7 @@ import graphql from 'babel-plugin-relay/macro'
  * Renders the UI for the reflection phase of the retrospective meeting
  *
  */
-import React from 'react'
+import React, {useState} from 'react'
 import {useFragment} from 'react-relay'
 import useCallbackRef from '~/hooks/useCallbackRef'
 import {RetroGroupPhase_meeting$key} from '~/__generated__/RetroGroupPhase_meeting.graphql'
@@ -23,12 +23,17 @@ import styled from '@emotion/styled'
 import AutogroupMutation from '../mutations/AutogroupMutation'
 import useAtmosphere from '../hooks/useAtmosphere'
 import useMutationProps from '../hooks/useMutationProps'
+import {Elevation} from '../styles/elevation'
 
 const ButtonWrapper = styled('div')({
   display: 'flex',
-  width: '100%',
-  paddingBottom: 32,
-  paddingLeft: 338 // TODO: super hacky, only for demo
+  padding: '16px 0px 8px 0px'
+})
+
+const StyledButton = styled(PrimaryButton)({
+  '&:hover, &:focus': {
+    boxShadow: Elevation.Z2
+  }
 })
 
 interface Props extends RetroMeetingPhaseProps {
@@ -57,15 +62,18 @@ const RetroGroupPhase = (props: Props) => {
   )
   const [callbackRef, phaseRef] = useCallbackRef()
   const atmosphere = useAtmosphere()
-
   const {onError, onCompleted} = useMutationProps()
-
+  const [hasSuggestedGroups, setHasSuggestedGroups] = useState(false)
   const {id: meetingId, endedAt, showSidebar, organization} = meeting
   const {featureFlags} = organization
   const {suggestGroups} = featureFlags
 
   const handleAutoGroupClick = () => {
-    AutogroupMutation(atmosphere, {meetingId}, {onError, onCompleted})
+    if (!hasSuggestedGroups) {
+      AutogroupMutation(atmosphere, {meetingId}, {onError, onCompleted})
+      // TODO: show ungroup button instead
+      setHasSuggestedGroups(true)
+    }
   }
 
   return (
@@ -78,14 +86,16 @@ const RetroGroupPhase = (props: Props) => {
         >
           <PhaseHeaderTitle>{phaseLabelLookup.group}</PhaseHeaderTitle>
           <PhaseHeaderDescription>{'Drag cards to group by common topics'}</PhaseHeaderDescription>
+          {suggestGroups && (
+            <ButtonWrapper>
+              <StyledButton disabled={hasSuggestedGroups} onClick={handleAutoGroupClick}>
+                {'Suggest Groups ✨'}
+              </StyledButton>
+            </ButtonWrapper>
+          )}
         </MeetingTopBar>
         <PhaseWrapper>
           <StageTimerDisplay meeting={meeting} canUndo={true} />
-          {/* {suggestGroups && ( */}
-          <ButtonWrapper>
-            <PrimaryButton onClick={handleAutoGroupClick}>{'Suggest Groups ✨'}</PrimaryButton>
-          </ButtonWrapper>
-          {/* )} */}
           <MeetingPhaseWrapper>
             <GroupingKanban meeting={meeting} phaseRef={phaseRef} />
           </MeetingPhaseWrapper>
--- a/packages/client/components/RetroGroupPhase.tsx
+++ b/packages/client/components/RetroGroupPhase.tsx
@@ -46,6 +46,11 @@ const RetroGroupPhase = (props: Props) => {
         id
         endedAt
         showSidebar
+        organization {
+          featureFlags {
+            suggestGroups
+          }
+        }
       }
     `,
     meetingRef
@@ -55,7 +60,10 @@ const RetroGroupPhase = (props: Props) => {
 
   const {onError, onCompleted} = useMutationProps()
 
-  const {id: meetingId, endedAt, showSidebar} = meeting
+  const {id: meetingId, endedAt, showSidebar, organization} = meeting
+  const {featureFlags} = organization
+  const {suggestGroups} = featureFlags
+  console.log('🚀 ~ organization:', organization)
 
   const handleAutoGroupClick = () => {
     AutogroupMutation(atmosphere, {meetingId}, {onError, onCompleted})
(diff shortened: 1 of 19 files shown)


packages/client/components/RetroReflectPhase/RetroReflectPhase.tsx
```diff
@@ -17,6 +19,8 @@ import StageTimerDisplay from '../StageTimerDisplay'
 import PhaseItemColumn from './PhaseItemColumn'
 import ReflectWrapperMobile from './ReflectionWrapperMobile'
 import ReflectWrapperDesktop from './ReflectWrapperDesktop'
+import AutogroupMutation from '../../mutations/AutogroupMutation'
```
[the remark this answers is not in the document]
igorlesnenko: Looks like not relevant anymore

packages/server/utils/OpenAIServerManager.ts
```diff
@@ -40,6 +40,68 @@ class OpenAIServerManager {
...
+              Don't change the input text at all, even if it's spelt incorrectly.
+              If you're unable to group the reflections, simply say "No".
+
+      Example Input:
+      ['The retreat went well', 'The project might not be ready in time', 'The deadline is really tight', 'I liked that the length of the retreat', 'I enjoyed the hotel', 'Feeling overworked']
+
+      Example Output:
+      [
+        {
+          "The Retreat": [
```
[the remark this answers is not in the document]
nickoferrall: Great question! I've added an issue for this as I think we can test it internally before this is resolved: https://github.com/ParabolInc/parabol/issues/8153

packages/server/utils/OpenAIServerManager.ts
```diff
@@ -40,6 +40,68 @@ class OpenAIServerManager {
...
+          top_p: 1,
+          frequency_penalty: 0,
+          presence_penalty: 0
+        },
+        {
+          timeout: 60000
+        }
+      )
+      const answer = (response.data.choices[0]?.message?.content?.trim() as string) ?? null
+      const nullableAnswer = /^No\.*$/i.test(answer) ? null : answer
```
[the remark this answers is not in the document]
nickoferrall: Updated! 👍

igorlesnenko (approved):
Works!

packages/server/graphql/public/mutations/autogroup.ts
```diff
@@ -0,0 +1,107 @@
...
+  const {authToken} = context
+  const viewerId = getUserId(authToken)
+  const r = await getRethink()
+  const meeting = await r.table('NewMeeting').get(meetingId).run()
+
+  if (!meeting) {
+    return standardError(new Error('Meeting not found'), {userId: viewerId})
+  }
+  if (meeting.meetingType !== 'retrospective') {
+    return standardError(new Error('Incorrect meeting type'), {userId: viewerId})
```
Dschoordsch, line 28: +1 Maybe also just return "Meeting not found"?

packages/server/graphql/public/mutations/autogroup.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+  const {groupedReflectionsJSON, teamId} = meeting
+  if (!isTeamMember(authToken, teamId)) {
+    return standardError(new Error('Team not found'), {userId: viewerId})
```
Dschoordsch, line 32: +1 I think we should return "Meeting not found" here so you cannot use this mutation to test for Meeting ids.

packages/server/graphql/mutations/helpers/generateGroups.ts
```diff
@@ -0,0 +1,28 @@
...
+  if (!hasSuggestGroupsFlag) return
+  const r = await getRethink()
+  const groupReflectionsInput = reflections.map((reflection) => reflection.plaintextContent)
+  const manager = new OpenAIServerManager()
+  const groupedReflectionsJSON = await manager.groupReflections(groupReflectionsInput)
+  if (!groupedReflectionsJSON) {
+    console.warn('ChatGPT was unable to group the reflections')
+    return
+  }
+  await r.table('NewMeeting').get(meetingId).update({groupedReflectionsJSON}).run()
```
Dschoordsch, line 25: -1 I don't think we should store the groups on the meeting, but if we did, then it should be as an object directly, not as an unparsed JSON string.

Dschoordsch (changes requested):
Did not test

packages/server/graphql/public/mutations/autogroup.ts
nickoferrall: Makes sense, I've updated the error message 👍

packages/client/components/RetroGroupPhase.tsx
```diff
@@ -50,6 +86,13 @@ const RetroGroupPhase = (props: Props) => {
         >
           <PhaseHeaderTitle>{phaseLabelLookup.group}</PhaseHeaderTitle>
           <PhaseHeaderDescription>{'Drag cards to group by common topics'}</PhaseHeaderDescription>
+          {suggestGroups && (
+            <ButtonWrapper>
+              <StyledButton disabled={hasSuggestedGroups} onClick={handleAutoGroupClick}>
+                {'Suggest Groups ✨'}
```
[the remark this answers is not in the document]
nickoferrall: Originally, I called it "Auto Group", but Aviva shared feedback ([here](https://parabol.slack.com/archives/C042E8PGPD4/p1682545436491539?thread_ts=1682533744.533489&cid=C042E8PGPD4)) that "Suggest Groups" felt friendlier and more optional. 

I suspect users may feel concerned that the change is irreversible, so I also think "Suggest Groups" may encourage users to click the button. 

The naming in the UI is more of a marketing thing, so I think we should go with Aviva's suggestion. 

I have an issue here to implement a "Reset Groups" button: https://github.com/ParabolInc/parabol/issues/8124

packages/server/graphql/types/RetrospectiveMeeting.ts
```diff
@@ -52,6 +53,11 @@ const RetrospectiveMeeting: GraphQLObjectType<any, GQLContext> = new GraphQLObje
       description: 'The number of comments generated in the meeting',
       resolve: ({commentCount}) => commentCount || 0
     },
+    groupedReflectionsJSON: {
```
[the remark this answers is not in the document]
nickoferrall: The type will depend on the answer to this q: https://github.com/ParabolInc/parabol/pull/8132#discussion_r1193988811

packages/server/utils/OpenAIServerManager.ts
```diff
@@ -40,6 +40,68 @@ class OpenAIServerManager {
...
+              If you're unable to group the reflections, simply say "No".
+
+      Example Input:
+      ['The retreat went well', 'The project might not be ready in time', 'The deadline is really tight', 'I liked that the length of the retreat', 'I enjoyed the hotel', 'Feeling overworked']
+
+      Example Output:
+      [
+        {
+          "The Retreat": [
+            "The retreat went well",
```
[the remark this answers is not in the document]
tianrunhe: I like this prompt!

packages/server/graphql/mutations/helpers/generateGroups.ts
Dschoordsch: Because `NewMeeting` is still in rethink, you can just store the object directly in a field on `NewMeeting`, no reason to add another relation.
I think even in Postgres in the future this would be a good candidate for just a JSON field.

nickoferrall: The JSON includes both the new reflection group title and the text of the reflections in the group, so we couldn't add just an `autogroupReflectionsIds` field to the `Meeting` object if that's what you had in mind?

Here's the format of the JSON:

```
      [
        {
          "The Retreat": [
            "The retreat went well",
            "I liked that the length of the retreat",
            "I enjoyed the hotel"
          ]
        },
        {
          "Deadlines": [
            "Feeling overworked",
            "The project might not be ready in time",
            "The deadline is really tight"
          ]
        }
      ]
```

Dschoordsch: What I had in mind is rather a format like
```
[
  {
    groupTitle: "The Retreat",
    reflectionIds: [12,23,45]
  },
  {
    groupTitle: "Deadlines",
    reflectionIds: [67,89,0]
  }
]
```
So do the processing and then store the object. That would make most sense to me, so it's easy to digest the data without knowing it's coming from openai

packages/server/utils/OpenAIServerManager.ts
```diff
@@ -40,6 +41,86 @@ class OpenAIServerManager {
       return null
     }
   }
+
+  async groupReflections(reflectionsText: string[]) {
+    if (!this.openAIApi) return null
+    try {
+      const response = await this.openAIApi.createChatCompletion(
+        {
+          model: 'gpt-3.5-turbo',
```
nickoferrall, line 50: I tried testing this with the reflections in our previous Squad's retro, but it really struggles to group reflections when they're not similar. This often leads to a few groups being created with many reflections but several with just one reflection.

I've tried many different iterations of saying: "Never create groups with just one reflection", but it struggles with this.

I have tried it with the gpt-4 model, and it works significantly better. However, the token limit is too low. Also, the gpt-3.5-turbo model can take 10-20 seconds to respond, so the gpt-4 model can hit the 60 seconds timeout limit.

We could group all of the solo groups together into one "Other" group. I suspect it'll be hard to launch the feature to all users until gpt-4 comes out of beta, which I would imagine will happen within a month or two, given the rate of recent progress. 

Screenshot 2023-05-19 at 16 34 06

packages/server/graphql/public/mutations/autogroup.ts
```diff
@@ -1,104 +1,48 @@
...
-      sendToSentry(error, {tags: {problematicReflection}})
-      return
-    }
-
-    const reflectionPromises = reflectionsTextInGroup
-      .slice(1)
-      .map((reflectionTextInGroup) =>
-        processReflectionInGroup(firstReflectionInGroup, reflectionTextInGroup, group)
+  await Promise.all(
+    autogroupReflectionGroups.flatMap((group) => {
```
Dschoordsch, line 32: -1 if autogrouping failed, `autogroupReflectionGroups` will be undefined

Dschoordsch (changes requested):
Did test and found small issue

packages/server/utils/OpenAIServerManager.ts
```diff
@@ -42,51 +43,78 @@ class OpenAIServerManager {
...
+      (total, reflectionText) => total + Math.ceil(reflectionText.length / AVG_CHARS_PER_TOKEN),
+      0
+    )
+    const avgReflectionLength = totalLengthInTokens / reflectionsText.length
+
+    const avgReflectionsPerGroup = 3 // assume each group will have around 3 reflections
+    const numGroups = Math.ceil(reflectionsText.length / avgReflectionsPerGroup)
+
+    const outputTokens = Math.round(
+      numGroups * (avgGroupNameLength + avgReflectionsPerGroup * avgReflectionLength)
```
Dschoordsch, line 59: +1 isn't this equal to `numGroups * avgGroupNameLength + totalLengthInTokens`?

Dschoordsch (approved):
I did not test generated groups

packages/server/database/types/MeetingRetrospective.ts
```diff
@@ -1,3 +1,4 @@
+import {AutogroupReflectionGroupType} from '../../graphql/types/AutogroupReflectionGroup'
```
Dschoordsch, line 1: -1 Our database types should not depend on GraphQL types. When we change our API, the underlying data does not change automatically. I missed this one before and it's causing your typecheck issue.

Dschoordsch (changes requested):

packages/server/utils/OpenAIServerManager.ts
nickoferrall: Good point, updated!

nickoferrall:
@Dschoordsch this PR was previously approved until the build error came up. Now that that's fixed, are you happy for me to merge it so that it makes it into this release?

Dschoordsch (approved):

PR #8136, opened 2023-05-03
chore(prompt-to-join): Allow to open add teammate dialog from the snackbar

igorlesnenko:
**How to test:**

- Follow testing steps from #8086
- But now make sure you can open popup both from notification and snackbar


packages/client/components/ReviewRequestToJoinOrgModal.tsx
```diff
@@ -27,79 +30,74 @@ const query = graphql`
...
+          Which teams would you like to add <strong>{createdByEmail}</strong> to?
+        </div>
+
+        <div>
+          {sortedTeams.map((team) => {
+            const {id: teamId, name: teamName, organization, teamMembers} = team
+            const {name: orgName} = organization
+
+            // TODO: implement filter on API side
+            const disabled = !!teamMembers.find((member) => member.userId === createdBy)
```
nickoferrall, line 72: +1 I think something like `isUserCreatedTeam` would be easier to understand. Also, `some` instead of `find` would mean we can remove the `!!`

packages/client/components/ReviewRequestToJoinOrgRoot.tsx
```diff
@@ -0,0 +1,57 @@
...
+import useQueryLoaderNow from '../hooks/useQueryLoaderNow'
+import reviewRequestToJoinOrgModalQuery, {
+  ReviewRequestToJoinOrgModalQuery
+} from '../__generated__/ReviewRequestToJoinOrgModalQuery.graphql'
+
+const ReviewRequestToJoinOrgRoot = () => {
+  const {match} = useRouter<{requestId: string}>()
+  const {params} = match
+  const {requestId} = params
+  useSubscription('NewMeetingRoot', NotificationSubscription)
```
nickoferrall, line 20: -1 why are these `NewMeetingRoot`s required?

nickoferrall (changes requested):

packages/client/components/ReviewRequestToJoinOrgRoot.tsx
igorlesnenko: Removed, works fine without

igorlesnenko:
Merging this PR into #8086 as it includes many fixes for #8086, and it might be easier to review together

PR #8154, opened 2023-05-04
chore: remove __jsxRuntime path

### LINKED ISSUE #8148: audit paths that only exist at build time
where we build is not going to be where we run, so we'll need to make sure any absolute paths that make it into our bundle are still valid outside of the build machine.

mattkrick:
# Description

Fixes #8148

When JSX gets compiled, it will include a line like `const __jsxRuntime = "/Users/matt/home/code/parabol/packages/foo/bar/baz.ts"`. This is only needed in development. In production, it just bloats the build. More importantly, it's a leaky abstraction because that path is only valid for the machine that builds the app, not the machine that runs the app.

## Testing scenarios

- [ ] Run `yarn clean && yarn build`, then perform a find in the build directory & the dist directory. The only absolute paths in the filesystem exist in `pushToCDN`. (`pushToCDN` rewrites the absolute paths when it runs, but if you know how to get rid of them in the compiled code let me know!)


igorlesnenko (approved):
Works!

igorlesnenko:
>  The only absolute paths in the filesystem exist in pushToCDN. (pushToCDN rewrites the absolute paths when it runs, but if you know how to get rid of them in the compiled code let me know!)

Quick investigation: If I change here:

https://github.com/ParabolInc/parabol/blob/418c5cc8f46e39a6e3668b78d51fccac768f77d4/scripts/webpack/prod.servers.config.js#L142

this `publicPath: distPath` to this `publicPath: '/dist/'`, it solves the issue for me

mattkrick:
ahhh yeah! doing `/dist` works as long as we're not using a CDN, but typically the public path is going to be point to our CDN on S3 e.g. action-files.parabol.co

Issue #8151, opened 2023-05-04
RiD: Edit Mode - automatically enter edit mode from "clone and edit" or new template creation


jmtaber129:
Currently, when we create a new activity or clone an existing activity, we default to non-editing mode. We should enter editing mode automatically for activities we just created/cloned.

AC:
- [ ] When creating or cloning an activity, we enter edit mode automatically
- [ ] When clicking on any activity in the grid, we don't enter edit mode automatically (even if the title suggests a new template or copy)

PR #8156, opened 2023-05-04
feat(rid): Enter edit mode automatically for new templates

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8151

Automatically enter edit mode from template cloning+creation via `edit` nav state.

## Demo
https://www.loom.com/share/4ebf1bb32d554e8c9ba27d480801ca17

## Testing scenarios
- [ ] Create new retro and poker templates, and confirm edit mode is entered automatically
- [ ] Clone retro and poker templates, and confirm edit mode is entered automatically
- [ ] Confirm edit mode is not entered when clicking into template from activity grid, even for templates with "New Template" and "X Copy" titles.

## Final checklist

### DIFF
--- a/packages/client/components/ActivityLibrary/ActivityDetails.tsx
+++ b/packages/client/components/ActivityLibrary/ActivityDetails.tsx
@@ -205,6 +205,11 @@ const ActivityDetails = (props: Props) => {
 
   const [isEditing, setIsEditing] = useState(false)
 
+  useEffect(() => {
+    const params = new URLSearchParams(history.location.search)
+    setIsEditing(!!params.get('edit'))
+  }, [history.location.search, setIsEditing])
+
   useEffect(
     () =>
       selectedTemplate &&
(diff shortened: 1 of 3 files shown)


packages/client/components/ActivityLibrary/ActivityDetails.tsx
```diff
@@ -205,6 +205,10 @@ const ActivityDetails = (props: Props) => {
 
   const [isEditing, setIsEditing] = useState(false)
 
+  useEffect(() => {
+    setIsEditing(!!history.location.state?.edit)
+  }, [history.location.state?.edit, setIsEditing])
```
BartoszJarocki, line 210: -1 no need to use `useEffect` here, you can init the state like 
```
  const [isEditing, setIsEditing] = useState(() => !!history.location.state?.edit)
```

unless I'm missing some case when we actually want to observe location state

BartoszJarocki (changes requested):
I believe there's no need to use `useEffect`, otherwise looks good!

packages/client/components/ActivityLibrary/ActivityDetails.tsx
jmtaber129: This is what I tried originally, but if we clone a template, React will reuse this component, keeping the previous template's edit state, so we need to set this manually

BartoszJarocki: ah interesting! let's go with that solution, I'll investigate it a bit more in the meantime

BartoszJarocki (approved):

Issue #8152, opened 2023-05-04
Recurrence: Navigate to previous and next meetings


jmtaber129:
Currently, the only way to navigate to previous and next meetings in a meeting series is by finding the desired meeting in the meeting history.

Instead, we'd like to add support for navigating to certain meetings from within a recurring meeting.

Specifically, we want to add back+next arrows to the meeting title that navigate to the previous and next meetings in the series.

[Figma](https://www.figma.com/file/azeQHaPX6yVsTHCq1aFPvK/Standups-Stable-Link?type=design&node-id=13%3A10786&t=jDi6aUiMYFH7uHXE-1)

AC:
- [ ] Recurring standup meetings have back+next buttons that match designs
- [ ] If no previous meeting is present in the series, the button is disabled
- [ ] If no next meeting is present in the series, the button is disabled
- [ ] No change to non-recurring meetings.

PR #8157, opened 2023-05-04
feat(recurrence): Prev/next meeting navigation

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8152

Adds prev+next buttons to recurring meetings to link to the previous or next meeting in the series, if applicable.

## Demo
https://www.loom.com/share/450680aacaaf44bca9b1a2d211e6cb29

## Testing scenarios
- [ ] For a recurring meeting with other meetings in the series, navigate back and forth through meetings.
- [ ] For non-recurring meetings, no buttons appear
- [ ] For recurring meetings with no other meetings in the series, no buttons appear.
- [ ] Prev/next buttons behave as links, and can be opened in new tabs

## Final checklist

### DIFF
--- a/packages/server/graphql/public/types/TeamPromptMeeting.ts
+++ b/packages/server/graphql/public/types/TeamPromptMeeting.ts
@@ -12,57 +12,54 @@ const TeamPromptMeeting: TeamPromptMeetingResolvers = {
     return null
   },
   meetingSeries: async ({meetingSeriesId}, _args, {dataLoader}) => {
-    if (meetingSeriesId) {
-      const series = await dataLoader.get('meetingSeries').load(meetingSeriesId)
-      if (!series) {
-        return null
-      }
+    if (!meetingSeriesId) return null
 
-      return series
+    const series = await dataLoader.get('meetingSeries').load(meetingSeriesId)
+    if (!series) {
+      return null
     }
-    return null
+
+    return series
   },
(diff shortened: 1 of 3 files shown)


packages/server/graphql/public/types/TeamPromptMeeting.ts
```diff
@@ -19,6 +21,48 @@ const TeamPromptMeeting: TeamPromptMeetingResolvers = {
       return series
     }
     return null
+  },
+  prevMeeting: async ({meetingSeriesId, createdAt}, _args, {dataLoader}) => {
+    if (meetingSeriesId) {
```
BartoszJarocki, line 26: this isn't even a +1 but just a suggestion. I'm personally a big fan of flattening code hierarchy as it's easier to read. 

Doing this
```
if (!meetingSeriesId) return null

// everything else goes here
```

makes the code even easier to read. If you would like to dig more, here's a good overview of the early return pattern: https://medium.com/swlh/return-early-pattern-3d18a41bba8

BartoszJarocki (approved):
Not a fan of element visually changing the position when there's no more prev meeting but we can improve the UI later, let's 🚢

packages/server/graphql/public/types/TeamPromptMeeting.ts
jmtaber129: Ah yea, nice catch. I usually prefer this, too

PR #8159, opened 2023-05-04
feat: mutable category

mattkrick:
Stuck behind #8081. use that as your base branch before doing a code review, or just ignore the DB migrations.

# Description

- Allows the viewer to change the category of their activity
- I introduced radix-ui for menus. It's FANTASTIC. I'd like to use this as a pattern that the gitstart folks can emulate so they can refactor our old, non-aria-compliant menus to this.
- I refactored `MeetingTemplate` to the SDL pattern on the server. mixing & matching was too confusing.

## Demo

https://www.loom.com/share/f11c939499254548a6776fb79def71fe

## Testing scenarios

- [ ] In edit mode, can trigger the dropdown by clicking the down arrow or the badge
- [ ] can use the keyboard to pick the category using up/down arrows
- [ ] can use the keyboard to exit the menu using Esc
- [ ] can use the keyboard to pick a category using typeahead (e.g. type "est" and it'll highlight Estimation)
- [ ] can use the keyboard to select a category by hitting Enter
- [ ] the down arrow is only visible & the badge is only clickable while editing
- [ ] make the updateTemplateCategory server mutation return an error message `return {error: {message: 'foo'}}`. It shows up as a snack

OUT OF SCOPE
- [ ] open/close menu animations. I'd like to introduce framer motion. that will be another PR 
- [ ] Bring elevations into tailwinds so box shadows look the same as legacy (advice welcome @BartoszJarocki)


codegen.json
mattkrick, line 1: `yarn newMutation` sorts this automatically

packages/client/components/ActivityLibrary/ActivityDetails.tsx
```diff
@@ -89,20 +86,6 @@ const query = graphql`
   }
 `
 
-interface DetailsBadgeProps {
```
mattkrick, line 92: Moved this & its parents into a their own files. This component is getting a little wild

packages/client/components/ActivityLibrary/ActivityDetailsBadges.tsx
```diff
@@ -0,0 +1,46 @@
...
+        ...ActivityDetailsCategoryBadge_template
+        isFree
+        orgId
+      }
+    `,
+    templateRef
+  )
+  const {isFree, orgId} = template
+  const isParabolTemplate = orgId === 'aGhostOrg'
+  const isPremium = isParabolTemplate && !isFree
```
mattkrick, line 27: I changed this logic to make it a little more readable

packages/client/components/ActivityLibrary/ActivityDetailsCategoryBadge.tsx
```diff
@@ -0,0 +1,95 @@
...
+      fragment ActivityDetailsCategoryBadge_template on MeetingTemplate {
+        id
+        category
+      }
+    `,
+    templateRef
+  )
+  const {id: templateId} = template
+  const category = template.category as CategoryID
+  const [commit] = useMutation<TUpdateTemplateCategoryMutation>(UpdateTemplateCategoryMutation)
```
mattkrick, line 34: thoughts on using `useMutation`? It has an `isInFlight` value that would allow us to stop using `useMutationProps`. I kinda dig it, but happy to switch back to old school, too.

packages/client/components/ActivityLibrary/ActivityDetailsCategoryBadge.tsx
```diff
@@ -0,0 +1,95 @@
...
+          atmosphere.eventEmitter.emit('addSnackbar', {
+            key: 'updateCategory',
+            message,
+            autoDismiss: 5
+          })
+      }
+    })
+  }
+  return (
+    <DropdownMenu.Root>
```
mattkrick, line 56: radix menus are 🔥 . let me know whatcha think!

packages/server/graphql/public/typeDefs/_legacy.graphql
```diff
@@ -2005,6 +2005,11 @@ type ReflectTemplate implements MeetingTemplate {
   """
   isActive: Boolean!
 
+  """
+  True if template is available to all teams including non-paying teams, else false
+  """
+  isFree: Boolean!
```
mattkrick, line 2011: these only existed in the JS version!

packages/server/graphql/public/types/PokerTemplate.ts
```diff
@@ -3,11 +3,17 @@ import {PokerTemplateResolvers} from '../resolverTypes'
 const RECOMMENDED_TEMPLATES = ['estimatedEffortTemplate']
 
 const PokerTemplate: PokerTemplateResolvers = {
-  category: (_source, _args, _context) => {
-    return 'estimation'
+  __isTypeOf: ({type}) => type === 'poker',
```
jmtaber129, line 6: -1 Let's pull the typedefs out of `_legacy.graphql` too if we're migrating these to SDL.

packages/server/graphql/public/mutations/updateTemplateCategory.ts
```diff
@@ -0,0 +1,38 @@
...
+) => {
+  const operationId = dataLoader.share()
+  const subOptions = {mutatorId, operationId}
+  const pg = getKysely()
+
+  // VALIDATION
+  if (mainCategory.length < 1) return {error: {message: 'Category name cannot be blank'}}
+  if (mainCategory.length > 100) return {error: {message: 'Category name is too long'}}
+
+  // I'd like to relax this later, hence the separate validation from string length
```
jmtaber129, line 19: -1 Let's just add a TODO for the string length validation to make this simpler, since we don't yet have plans for custom categories.

Even for custom categories, we'd probably have a separate "create category" mutation, such that we'd just take a category ID here.

packages/server/graphql/public/mutations/updateTemplateCategory.ts
```diff
@@ -0,0 +1,38 @@
...
+  const stockCategories = ['retrospective', 'estimation', 'standup', 'feedback', 'strategy']
```
jmtaber129, line 20: +1 extract to constant array, i.e. `STOCK_CATEGORIES` outside of the resolver?

+1 `CATEGORY_OPTIONS`?

packages/client/components/ActivityLibrary/ActivityDetailsCategoryBadge.tsx
jmtaber129: I kinda like it, too, but I'm hesitant to introduce it without a solid understanding of how it would fit in with our existing mutations, especially since this pattern would move the updater, onComplete, etc. methods into the component that calls them.

Let's use the existing pattern for now and look more deeply at this pattern separately.

jmtaber129 (changes requested):
First pass, will take another look on monday.

-1 As mentioned previously, mutable categories, hiding seasonal templates, and mutable illustration URLs are three distinct and unrelated features, so please split up this PR.

Additionally, the squad is likely only prioritizing mutable categories (and _maybe_ hiding seasonal templates) in the near-term.

packages/client/components/ActivityLibrary/ActivityDetailsCategoryBadge.tsx
mattkrick: > especially since this pattern would move the updater, onComplete, etc. methods into the component that calls them.

This is true for `optimisticUpdater, updater`. We could move those back to the mutation file though since they don't have a scope.

`onError, onCompleted, variables` were already defined in the component, so there's no change for those, unless i'm misunderstanding what you meant?

packages/client/components/ActivityLibrary/ActivityDetailsCategoryBadge.tsx
```diff
@@ -0,0 +1,95 @@
...
+            {CATEGORY_ID_TO_NAME[category]}
+          </ActivityDetailsBadge>
+          {isEditing && <KeyboardArrowDownIcon />}
+        </PlainButton>
+      </DropdownMenu.Trigger>
+      <DropdownMenu.Portal>
+        <DropdownMenu.Content className='border-rad rounded bg-white shadow-lg' sideOffset={5}>
+          <DropdownMenu.RadioGroup value={category} onValueChange={updateTemplateCategory}>
+            {Object.keys(CATEGORY_THEMES).map((v) => {
+              const value = v as CategoryID
```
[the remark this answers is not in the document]
mattkrick: good catch, `category` is already used in the outer scope, so I can do something like `categoryId`

packages/server/graphql/public/mutations/updateTemplateCategory.ts
mattkrick: will do

packages/server/graphql/public/mutations/updateTemplateCategory.ts
mattkrick: will do

packages/server/graphql/public/mutations/updateTemplateCategory.ts
```diff
@@ -0,0 +1,38 @@
...
+  if (!stockCategories.includes(mainCategory))
+    return {error: {message: 'Custom categories not available'}}
+
+  // RESOLUTION
+  await pg
+    .updateTable('MeetingTemplate')
+    .set({mainCategory})
+    .where('id', '=', templateId)
```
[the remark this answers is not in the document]
mattkrick: this is a good DX tension...
`templateId` is guaranteed to be valid from `permissions.ts` because we compare `viewer.tms` to `template.teamId`. I like separating permissions from resolution, but I agree it's not intuitive that `templateId` has already been validated. I don't know a great way to communicate that! I added a comment for now

packages/server/graphql/public/types/PokerTemplate.ts
mattkrick: I'm all for that!

PR #8160, opened 2023-05-05
feat(prompt-to-join): Add teammate mutation

### LINKED ISSUE #7803: Prompt to join org: Accept request mutation
Clicking "Add to Teams" button should run a mutation that adds a user to selected teams

**AC:** 

- Implement a new mutation that add a user to selected teams
- Mutation accepts a `requestId` and array of team ids:

```
 requestId: 123
 teamIds: ['345', '678']
```

- Mutation adds a user to provided `teamIds`
- Validate the input
  - Viewer should be a team leader of the provided teams
  - Teams should belong to the request domain
  - The request should not be expired  
- If the request is expired, return request expired error
- If user is already on a team, just ignore the team 
- Add mutation to the UI here #7802

igorlesnenko:
- Fixes #7803
- In this PR I also remove unnecessary enterprise restriction on sending join request 
- Added segment events

https://user-images.githubusercontent.com/466991/236413243-0bd08389-183a-451c-bd06-7426d6269315.mov

**How to test:** 
- Follow previous PR testing steps to see add teammate UI (#8086)
- Test that adding a user works correctly
- Check that segment events sent correctly


packages/client/components/ReviewRequestToJoinOrgModal.tsx
```diff
@@ -42,24 +62,39 @@ interface Props {
...
+
   const data = usePreloadedQuery<ReviewRequestToJoinOrgModalQuery>(query, queryRef)
 
   const [selectedTeams, setSelectedTeams] = useState<string[]>([])
 
   const {domainJoinRequest} = data.viewer
 
+  useEffect(() => {
+    if (domainJoinRequest) {
+      SendClientSegmentEventMutation(atmosphere, 'Join Request Reviewed')
```
tghanken, line 75: +1 Would change this name:
```suggestion
      SendClientSegmentEventMutation(atmosphere, 'Join Request Viewed')
```

It also might make sense to move this out of the modal logic, and instead emit the event when they see the Join Request notification.  That way the funnel we'd be tracking is "Join Request Viewed" (emitted for every user who receives a request when they view it) -> "Join Request Reviewed" when they either "Deny" or add the user to teams.  I'm not sure we learn a lot from people who are just opening the modal.

packages/client/mutations/toasts/mapRequestToJoinOrgToToast.ts
```diff
@@ -36,7 +37,7 @@ const mapRequestToJoinOrgToToast = (
     secondaryAction: {
       label: 'Deny',
       callback: () => {
-        // Do nothing
+        SendClientSegmentEventMutation(atmosphere, 'Join Request Denied')
```
tghanken, line 40: +1 Would like to see us use a single "Reviewed" event.
```suggestion
        SendClientSegmentEventMutation(atmosphere, 'Join Request Reviewed', {
          action: 'Deny'
        })
```

packages/client/mutations/AcceptRequestToJoinDomainMutation.ts
```diff
@@ -0,0 +1,57 @@
...
+      const error = res.acceptRequestToJoinDomain.error
+      if (!error) {
+        atmosphere.eventEmitter.emit('addSnackbar', {
+          message: '🎉 Success! User added',
+          autoDismiss: 5,
+          key: 'acceptRequestToJoinDomainSuccess'
+        })
+        SendClientSegmentEventMutation(atmosphere, 'Join Request Accepted', {
+          teamIds: variables.teamIds
+        })
```
tghanken, line 45: +1 Would like to see us use a single "Reviewed" event.
```suggestion
        SendClientSegmentEventMutation(atmosphere, 'Join Request Reviewed', {
          action: 'Accept',
          teamIds: variables.teamIds
        })
```

packages/server/utils/analytics/analytics.ts
```diff
@@ -102,6 +102,10 @@ export type AnalyticsEvent =
   // snackbar
   | 'Snackbar Clicked'
   | 'Snackbar Viewed'
+  // Join request
+  | 'Join Request Reviewed'
+  | 'Join Request Denied'
+  | 'Join Request Accepted'
```
tghanken, line 108: +1 Would change this to match my previous feedback:
```suggestion
  | 'Join Request Viewed'
  | 'Join Request Reviewed'
```

tghanken (approved):
Looks good!  I made some metric recommendations that would allow us slightly better visibility into how users are interacting with the feature, but they aren't blocking.

packages/client/components/ReviewRequestToJoinOrgModal.tsx
igorlesnenko: We have existing `Snackbar Viewed` event, should we use it when snackbar with the request viewed?

igorlesnenko: Made some changes

tghanken (approved):
Looks good!

packages/client/components/ReviewRequestToJoinOrgModal.tsx
```diff
@@ -41,12 +52,17 @@ interface Props {
...
+  const [selectedTeams, setSelectedTeams] = useState<string[]>([])
 
   const data = usePreloadedQuery<ReviewRequestToJoinOrgModalQuery>(query, queryRef)
+  const viewer = useFragment<ReviewRequestToJoinOrgModal_viewer$key>(
+    ReviewRequestToJoinOrgModalViewerFragment,
+    data.viewer
+  )
 
-  const [selectedTeams, setSelectedTeams] = useState<string[]>([])
+  const {domainJoinRequest} = viewer
```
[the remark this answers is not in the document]
igorlesnenko: `teams` are now part of the `domainJoinRequest`, so I can't put it here

packages/client/components/ReviewRequestToJoinOrgModal.tsx
```diff
@@ -60,7 +76,14 @@ const ReviewRequestToJoinOrgModal = (props: Props) => {
...
   }
 
-  const {createdBy, createdByEmail} = domainJoinRequest
+  const {createdBy, createdByEmail, id: requestId} = domainJoinRequest
+
+  const onAdd = () => {
+    AcceptRequestToJoinDomainMutation(atmosphere, {
+      requestId,
+      teamIds: selectedTeams
+    })
```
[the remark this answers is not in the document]
igorlesnenko: I'm handling error and success inside mutation 

https://github.com/ParabolInc/parabol/blob/66c191c46c96d336156ff05aa7369b170b2dda7f/packages/client/mutations/AcceptRequestToJoinDomainMutation.ts#L36

packages/server/graphql/public/mutations/acceptRequestToJoinDomain.ts
```diff
@@ -0,0 +1,133 @@
...
+    return standardError(new Error('Invalid organizations'))
+  }
+
+  const validTeams = leadTeams.filter((team) => validOrgIds.includes(team.orgId))
+
+  if (!validTeams.length) {
+    return standardError(new Error('Invalid teams'))
+  }
+
+  // make sure that same request can't be accepted at the same moment
```
Dschoordsch, line 72: -1 Please elaborate why we can't accept the request multiple times

packages/server/graphql/types/SnackbarTypeEnum.ts
```diff
@@ -1,13 +1,18 @@
...
+  | 'requestToJoinOrg'
 
 const SnackbarTypeEnum = new GraphQLEnumType({
   name: 'SnackbarTypeEnum',
   values: {
     teamsLimitExceeded: {},
     teamsLimitReminder: {},
-    promptToJoinOrg: {}
+    promptToJoinOrg: {},
+    requestToJoinOrg: {}
```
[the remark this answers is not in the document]
igorlesnenko: This is not a feature flag, but 2 different snackbars one is "Would you like to join org?", another is "a user requesting to join your org."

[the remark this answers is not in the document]
Dschoordsch: My fault

packages/server/graphql/public/mutations/acceptRequestToJoinDomain.ts
```diff
@@ -0,0 +1,131 @@
...
+    return standardError(new Error('Invalid organizations'))
+  }
+
+  const validTeams = leadTeams.filter((team) => validOrgIds.includes(team.orgId))
+
+  if (!validTeams.length) {
+    return standardError(new Error('Invalid teams'))
+  }
+
+  // Avoid race conditions: we make changes based on the data fetched first, see insertNewTeamMember
```
Dschoordsch, line 72: +1 I think we would be better served in making `insertNewTeamMember` safe by refactoring the rethink query to read and write in one go using `branch`, maybe sth. like this
```ts
r.table('TeamMember').get(teamMemberId).branch(
  r.table('TeamMember').get(teamMemberId).update({isNotRemoved: true, updatedAt: new Date()}),
  r.table('TeamMember').insert({
    ...teamMember,
    isLead: r.table('TeamMember').getAll(teamId, {index: 'teamId'}).filter({isNotRemoved: true}).count().eq(0)
  })
).run()
```

Dschoordsch (changes requested):
Dialog does not close on adding a user
Peek 2023-05-17 16-58

packages/server/graphql/public/mutations/acceptRequestToJoinDomain.ts
```diff
@@ -0,0 +1,131 @@
...
+    const [organizationUser] = await Promise.all([
+      dataLoader.get('organizationUsersByUserIdOrgId').load({userId, orgId}),
+      insertNewTeamMember(user, teamId),
+      addTeamIdToTMS(userId, teamId)
+    ])
+
+    if (!organizationUser) {
+      // clear the cache, adjustUserCount will mutate these
+      dataLoader.get('organizationUsersByUserIdOrgId').clear({userId, orgId})
+      dataLoader.get('users').clear(userId)
```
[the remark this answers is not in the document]
igorlesnenko: I copy pasted this from `acceptTeamInvitation`.

I think we need to remove dataloaders here at all and avoid using dataloaders for fetching data in mutations, as it makes everything very confusing

We modify user here: `addTeamIdToTMS(userId, teamId)`

Then we execute some complex functions that call other functions that are using `User`. So they might or might not require updated user or might require in future.

These functions creates new `OrganizationUser` and modify `User`. So we need to clear this after execution.

packages/server/graphql/public/mutations/acceptRequestToJoinDomain.ts
igorlesnenko: @Dschoordsch I think this is not the only issue, then we call `adjustUserCount` and  `setUserTierForUserIds` which both also make actions based on fetched data
I feel too much refactoring would be here if it is possible

packages/server/graphql/public/mutations/acceptRequestToJoinDomain.ts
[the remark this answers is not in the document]
igorlesnenko: Removed dataloaders. Feels like dangerous pattern to me: we have separate functions that are used in multiple places. However, by looking at the function alone we are unable to say — will it write in cache, will it take fresh data from cache, will it take stale data from cache.

packages/server/graphql/private/mutations/connectSocket.ts
```diff
@@ -44,7 +44,7 @@ const connectSocket: MutationResolvers['connectSocket'] = async (
       .getAll(userId, {index: 'userId'})
       .filter({removedAt: null, inactive: true})('orgId')
       .run()
-    adjustUserCount(userId, orgIds, InvoiceItemType.UNPAUSE_USER, dataLoader).catch(console.log)
+    adjustUserCount(userId, orgIds, InvoiceItemType.UNPAUSE_USER).catch(console.log)
```
Dschoordsch, line 47: +1 this one we will call most often, but that's still only once per session, so I think it's fine to remove the dataloader as you've did. I wouldn't like to do that as a general pattern though.

igorlesnenko: I think in very few `update/insert` cases dataloader would safe us an additional db query, however we always have to remember to clear the cache. No strong opinion. Merging this for now, without taking this as a general pattern.

Issue #7914, opened 2023-03-13
Checkout Flow: fix initial jumpiness 


nickoferrall:
See https://github.com/ParabolInc/parabol/pull/7910#pullrequestreview-1337131552

This is probably caused by the Credit Card section taking a split second to load because we're sending a payment request to the server. Perhaps it would be best to show a credit card empty state while it's loading.

### AC

- [ ] Go the new checkout flow and see that there's no jumpiness in the UI with the credit card section showing up slightly after the other sections load

PR #8162, opened 2023-05-05
feat(checkout-flow): handle 3D Secure cards

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7864

Demo: https://www.loom.com/share/3159b9e87a384e7f8358ca50834a69c4

To do:

- [ ] The UI needs to be changed to handle the new `CardElement` UI and the initial jumpiness while Stripe Elements are loading: https://github.com/ParabolInc/parabol/issues/7914.
- [ ] The `UpgradeToTeamTierMutation` needs to be split up. At the moment, it includes the subscription logic and upgrading the org, but with 3D Secure cards, if the user fails to authenticate, it'll still upgrade the org.

I'll create issues for the above assuming no significant changes are requested when reviewing this PR.

### AC

- [ ] Enter a 3D secure card (see examples [here](https://stripe.com/docs/payments/3d-secure?locale=en-GB#three-ds-cards)) such as 4000000000003220
- [ ] Complete the 3D Secure pop-up auth and see that the subscription is successfully created
- [ ] For a different org, enter the same card number but click "Fail" in the 3D Secure pop-up. See the error message in the UI and see that the payment failed in Stripe (although the org will still be upgraded for now).

---

Stuff I learnt while creating this:

- Typically, to charge 3D Secure cards, you need to create a payment intent and then pass that to Stripe Elements. A set-up intent is used for subscription businesses that don't charge customers right away, e.g. a car rental that charges the customer after the rental period ends (see [docs](https://support.stripe.com/questions/payment-intents-api-vs-setup-intents-api#:~:text=Use%20the%20Setup%20Intents%20API,are%20ready%20to%20charge%20them.)).
- However, when creating a subscription, it creates its own paymentIntent.
- For 3D secure card payments, we need to add the `expand: ['latest_invoice.payment_intent']` field when creating the subscription. By doing this, we can get the `clientSecret` needed to confirm the payment after the 3D Secure card has been authorised by the user.

CC @mattkrick. While the new card number worked during office hours, it failed afterwards 🤷‍♂️  not sure why it worked on that call. Changing `paymentIntent` for `setupIntent` and then `confirmCardPayment` using the subscription secret worked.

### DIFF
--- a/packages/server/graphql/mutations/helpers/upgradeToTeamTier.ts
+++ b/packages/server/graphql/mutations/helpers/upgradeToTeamTier.ts
@@ -11,6 +11,7 @@ import {DataLoaderWorker} from '../../graphql'
 const upgradeToTeamTier = async (
   orgId: string,
   paymentMethodId: string,
+  email: string,
   dataLoader: DataLoaderWorker
 ) => {
   const r = await getRethink()
@@ -30,7 +31,7 @@ const upgradeToTeamTier = async (
   const {stripeId} = organization
   const customer = stripeId
     ? await manager.retrieveCustomer(stripeId)
-    : await manager.createCustomer(orgId)
+    : await manager.createCustomer(orgId, email)
   const {id: customerId} = customer
   await manager.attachPaymentToCustomer(customerId, paymentMethodId)
   // wait until the payment is attached to the customer before updating the default payment method
--- a/packages/server/utils/stripe/StripeManager.ts
+++ b/packages/server/utils/stripe/StripeManager.ts
@@ -21,7 +21,7 @@ export default class StripeManager {
     return this.stripe.paymentMethods.attach(paymentMethodId, {customer: customerId})
   }
 
-  async createCustomer(orgId: string, email?: string, source?: string) {
+  async createCustomer(orgId: string, email: string, source?: string) {
     return this.stripe.customers.create({
       email,
       source,
--- a/packages/server/graphql/mutations/upgradeToTeamTier.ts
+++ b/packages/server/graphql/mutations/upgradeToTeamTier.ts
@@ -65,7 +65,12 @@ export default {
     try {
       // TODO: remove upgradeToTeamTierOld once we rollout the new checkout flow: https://github.com/ParabolInc/parabol/milestone/150
       if (paymentMethodId) {
-        stripeSubscriptionClientSecret = await upgradeToTeamTier(orgId, paymentMethodId, dataLoader)
+        stripeSubscriptionClientSecret = await upgradeToTeamTier(
+          orgId,
+          paymentMethodId,
+          email,
+          dataLoader
+        )
       } else if (stripeToken) {
         await upgradeToTeamTierOld(orgId, stripeToken, email, dataLoader)
       }
--- a/packages/server/graphql/public/mutations/createSetupIntent.ts
+++ b/packages/server/graphql/public/mutations/createSetupIntent.ts
@@ -10,14 +10,13 @@ const createSetupIntent: MutationResolvers['createSetupIntent'] = async (
   {dataLoader, authToken}
 ) => {
   const viewerId = getUserId(authToken)
-  const [viewer, organization] = await Promise.all([
+  const [viewer, organization, organizationUser] = await Promise.all([
     dataLoader.get('users').loadNonNull(viewerId),
-    dataLoader.get('organizations').load(orgId)
+    dataLoader.get('organizations').load(orgId),
+    dataLoader.get('organizationUsersByUserIdOrgId').load({userId: viewerId, orgId})
   ])
-  const {tms} = viewer
-  const viewerTeams = await dataLoader.get('teams').loadMany(tms)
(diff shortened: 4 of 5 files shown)


mattkrick:
@nickoferrall did the other card 3D number work? Very odd stripe behavior...

packages/client/modules/userDashboard/components/OrgBilling/BillingForm.tsx
```diff
@@ -63,33 +85,54 @@ const BillingForm = (props: Props) => {
...
+    UpgradeToTeamTierMutation(
+      atmosphere,
+      {orgId, paymentMethodId: paymentMethod.id},
+      {onError, onCompleted: handleCompleted}
+    )
   }
 
   return (
     <StyledForm id='payment-form' onSubmit={handleSubmit}>
-      <PaymentElement id='payment-element' options={{layout: 'tabs'}} />
```
igorlesnenko, line 92: I'm surprised that there is no 3d secure support. Not sure I understand why.
I'm [reading](https://stripe.com/docs/payments/payment-card-element-comparison) about PaymentElement, stripe says it is recommended solution, and shows 3d secure support

image

igorlesnenko (commented):
I have couple of questions

packages/client/modules/userDashboard/components/OrgBilling/BillingForm.tsx
nickoferrall: You can use `PaymentElement` with 3D Secure cards, but it looks like you can't with the `createPaymentMethod` method: https://stripe.com/docs/js/payment_methods/create_payment_method#stripe_create_payment_method-paymentMethodData-card

We need to use `createPaymentMethod` to create the payment method on the client using the card details the user inputs, then create the subscription on the server using those card details, and then confirm the card payment on the client once the 3D Secure card has been authorised by the user and we have the subscription client secret.

nickoferrall:
> @nickoferrall did the other card 3D number work?

Nope, neither card number worked. I tried the other test cards [here](https://stripe.com/docs/payments/3d-secure?locale=en-GB#three-ds-cards), but no luck. Confirming the card payment using the subscription client secret is now working consistently though

packages/server/graphql/mutations/helpers/upgradeToTeamTier.ts
```diff
@@ -67,6 +73,7 @@ const upgradeToTeamTier = async (
   ])
 
   await Promise.all([setUserTierForOrgId(orgId), setTierForOrgUsers(orgId)])
+  return clientSecret
```
igorlesnenko, line 76: With the current approach we upgrade tier to `team` before the successful payment.

I can interrupt payment or 3D secure will fail and I still get the upgrade

https://www.loom.com/share/05260018c4264c57b7d278b412ac8aa1

igorlesnenko (changes requested):

packages/server/graphql/mutations/helpers/upgradeToTeamTier.ts
igorlesnenko: @nickoferrall sorry, missed it

igorlesnenko (approved):

packages/server/graphql/mutations/helpers/upgradeToTeamTier.ts
nickoferrall: All good! I wanted to share this PR even though `UpgradeToTeamTierMutation` needs refactoring because I've already made several changes, and that refactor will add quite a bit more code

Dschoordsch (changes requested):

packages/server/utils/stripe/StripeManager.ts
```diff
@@ -21,7 +21,7 @@ export default class StripeManager {
     return this.stripe.paymentMethods.attach(paymentMethodId, {customer: customerId})
   }
 
-  async createCustomer(orgId: string, email: string, source?: string) {
+  async createCustomer(orgId: string, email?: string, source?: string) {
```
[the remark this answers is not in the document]
nickoferrall: Originally, I was also trying to save us from creating too many Stripe customers, but then I discovered the `stripeId` in `Organization` is the Stripe `customerId`, so now I'm just checking if `stripeId` already exists. Updated!

packages/client/modules/userDashboard/components/OrgBilling/PaymentDetails.tsx
```diff
@@ -176,8 +146,7 @@ const PaymentDetails = (props: Props) => {
           <Content>
             <Elements
```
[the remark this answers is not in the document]
nickoferrall: I still get the jumpiness as there's a 1/2 second delay from the page loading to getting the client secret from the createSetupIntent mutation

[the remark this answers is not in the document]
Dschoordsch: I see the components load, but no jumpiness as the width and height is already reserved in this case. Forgot to mention that I removed the return in l170.

PR #8169, opened 2023-05-08
chore: Add admin/login page to SSO for retool

Dschoordsch:
# Description

Add a separete login page which redirects the auth token.
There might be some security concerns I'm not seeing atm. At the very least we need to verify the redirect URI before blindly following it.

## Demo

https://www.loom.com/share/455f244ff23d43738ada67dc6660ec9d

## Testing scenarios

- [ ] Scenario A
  - Step 1
  - Step 2...

- [ ] Scenario B
  - Step 1
  - Step 2....

## Final checklist


packages/server/graphql/intranetGraphQLHandler.ts
```diff
@@ -35,7 +35,7 @@ const intranetHttpGraphQLHandler = uWSAsyncHandler(async (res: HttpResponse, req
       ip,
       query,
       variables,
-      isPrivate,
+      isPrivate: true,
```
Dschoordsch, line 38: If we treat everything coming to intranet-graphql as isPrivate, we can avoid the problems with setting the property manually and use the GraphQL resource type from retool.

packages/client/mutations/handlers/handleAuthenticationRedirect.ts
```diff
@@ -14,7 +14,13 @@ const handleAuthenticationRedirect: OnNextHandler<
   SendClientSegmentEventMutation(atmosphere, 'User Login')
   const redirectTo = getValidRedirectParam()
   if (redirectTo) {
-    history.push(redirectTo)
+    const nonce = new URLSearchParams(window.location.search).get('nonce')
+    const {authToken} = atmosphere
+    if (nonce) {
```
Dschoordsch, line 19: That's not the best criteria, but right now we don't pass the `'admin-login'` page type to the mutation

tianrunhe (approved):

packages/client/components/AuthenticationPage.tsx
```diff
@@ -23,11 +23,18 @@ const AuthenticationPage = (props: Props) => {
...
-  const {authObj} = atmosphere
+  const {authObj, authToken} = atmosphere
   useDocumentTitle('Sign Up for Free Online Retrospectives | Parabol', 'Sign Up')
   useCanonical(page)
   if (authObj) {
     const nextUrl = getValidRedirectParam() || '/meetings'
+    if (page === 'admin-login') {
+      if (window.__ACTION__.AUTH_ALLOWED_REDIRECTS.includes(nextUrl)) {
+        const nonce = new URLSearchParams(window.location.search).get('nonce')
+        window.location.href = `${nextUrl}?nonce=${nonce}&token=${authToken}`
```
mattkrick, line 34: You did great identifying the attack vector!
Putting on my hacker hat here...If I want control over parabol, my goal is going to be to get a salesperson to click my link that looks like `action.parabol.co/admin/login?redirectTo=https://yougothacked.com`.
The only thing preventing me from getting their key is a single mutable global variable 😬 . So now the question is, how do I change that global variable? 
- i could get them to install an extension or bookmarklet, but that's a longshot
- i could have a link like`<a href="javascript:__ACTION__ AUTH_ALLOWED_REDIRECTS=['https://yougothacked.com'].;">Click me</a>`. But that would require a salesperson seeing an untrusted link. Maybe I invite them to my team & get them to click that link on a task card? I guess that vector could be used even without this new code...
- there's _probably_ another attack vector I'm not thinking of. I'm not an expert! 

IMO, I can't see a vuln, but I could be wrong, and the potential downside is an attacker becoming a superuser, so I want to give this some more thought.

- if the variable wasn't in the global scope i'd feel a _little_ safer
- if we handled this on the server that could be better, still. e.g. write an /auth route that checks just for this case & if it doesn't match, fall back to the default GET
- If we use Retool's Google OAuth2 it wants our Google secret, which i don't love
- IMO, the safest option is doing the following 3 steps in Retool:
  - API call to Google (https://accounts.google.com/o/oauth2/v2/auth)
    - For the redirect_uri, you can add a new one in google console. I think you already tried this? So I'm curious if it didn't work... The better option is to update google to use our `env.OAUTH2_REDIRECT` like we do with Slack/GitHub
  - API call to parabol.co/graphql calling `loginWithGoogle(code)`
  - Define a variable using `http3.body.data.loginWithGoogle.authToken`

Normally I wouldn't care much, but the stakes are pretty high on this one, so I want to make sure we're as safe as possible!

mattkrick (changes requested):

packages/client/components/AuthenticationPage.tsx
Dschoordsch: Yes, the stakes are high. I might be wrong, but if an attacker is able to modify the global variable on the parabol site before redirect, they're most likely already able to read the token from the session storage.

Ideally I would prefer a scenario where we partially handle this on the server side:
* normal app token never has `su` rights, a separate login is required for admin and impersonate
* we include the referer in the token and check if it matches
This would mean logging in more often, but I think would be more secure.

mattkrick: yep, i'm right there with you. as we become more successful, more hackers are going to target us, so our goal should be to keep tokens out of reach of JS. i haven't written an issue for it, but security-wise I'd like to put the JWT in an http-only cookie so we can get it out of the websocket `GET` request as well as remove it from localStorage. As a bonus, we can share this cookie with the landing page, too (e.g. provide a "Login" button instead of "Sign up"). 

unfortunately we can't reliably add a referer to a JWT because since that is provided by the client it can be spoofed, but a special login for elevated credentials is a good idea!
Only extra thing I'd add is additional roles. having every employee be a superuser is scary

what are your thoughts on the retool auth flow above? it should result in less app code & no need for env vars!

Dschoordsch: What if we just move the same logic with the separate login endpoints to the server side:
* requestGoogleLogin will forward to google for signin and then redirect back to the second endpoint
* redirectGoogleLogin will send the code to the retool redirect
* in order to not have to keep state between these two, we could sign the state object with the server secret

Dschoordsch: This flow would not work anymore with the access token as httpOnly cookie. Thus we should implement a refresh token flow and store the refresh token in a httpOnly cookie.

mattkrick: Great findings! You're correct, today we're sending the `code` to a potentially untrusted url & our `loginWithGoogle` endpoint will trade in any `code` for a `token` without verifying that origin. Yikes!

Whitelisting on the server will work, but maintaining a growing whitelist of PPMIs isn't fun.

Your idea of signing a state object is great! Technically we don't need to sign the whole state object, we just need to sign the `origin`, and we only need to do that when the server starts up. What if we just send that signedOrigin to the client?

```ts
// on our server
const signedOrigin = crypto.createHmac('sha256', SERVER_SECRET!).update(process.env.HOST).digest('base64')

// on Client GoogleClientManager.ts
const providerState = btoa(JSON.stringify({origin: window.location.origin, signedOrigin: window.__ACTION__.SIGNED_ORIGIN, service: 'google'}))

// on oauth2Redirect
if (signedOrigin !== crypto.createHmac('sha256', SERVER_SECRET!).update(state.origin).digest('base64')) throw Error('hacker!')
```

It's 1 extra client key, but no state & no extra env vars. Whatcha think?

mwermuth:
Looked at during backlog grooming.

PR #8178, opened 2023-05-10
chore: remove node-fetch from app code

mattkrick:
# Description

- Removes node-fetch since Node now has fetch natively
- Fixes the update avatar call that used `.buffer`, which doesn't exist on native fetch

## Testing scenarios

- [ ] change your user avatar


packages/client/jest.config.js
```diff
@@ -3,14 +3,14 @@ const tsJestPresets = require('ts-jest/presets')
...
-  globals: {
-    'ts-jest': {
-      diagnostics: false
-    }
-  },
   transform: {
-    '\\.(gql|graphql)$': 'jest-transform-graphql',
-    ...tsJestPresets.jsWithBabel.transform
+    '\\.(gql|graphql)$': '../server/__tests__/jest-transform-graphql-shim.js',
+    '^.+\\.tsx?$': [
```
mattkrick, line 8: copied from https://kulshekhar.github.io/ts-jest/docs/getting-started/options/

packages/server/graphql/mutations/helpers/updateUserProfile.ts
```diff
@@ -38,7 +37,8 @@ const updateUserProfile = async (
 
   if (validUpdatedUser.picture && validUpdatedUser.picture.endsWith('.svg')) {
     const res = await fetch(validUpdatedUser.picture)
-    const buffer = await res.buffer()
+    const arrayBuffer = await res.arrayBuffer()
```
mattkrick, line 40: this is the only real change

packages/client/package.json
```diff
@@ -55,7 +55,7 @@
     "eslint-plugin-emotion": "^10.0.14",
     "eslint-plugin-react": "^7.16.0",
     "eslint-plugin-react-hooks": "^1.6.1",
-    "jest": "^27.0.6",
+    "jest": "^29.5.0",
```
mattkrick, line 58: native `fetch` isn't available to jest until v28 so i bumped to latest

packages/client/jest.config.js
Dschoordsch: +1 Should we remove the `preset` as the warning box on the page linked suggests?

Dschoordsch (approved):

packages/client/jest.config.js
mattkrick: good catch! fixed

packages/server/graphql/mutations/helpers/updateUserProfile.ts
mattkrick: right you are! we got some good spaghetti code going on there.
shall i remove that `.svg` conditional clause in this PR or should we save it for later? OK with either option.

Dschoordsch: Let's leave it for later and then let's remove the `picture` input from this mutation completely.

mattkrick: Issue made! https://github.com/ParabolInc/parabol/issues/8207

Issue #6642, opened 2022-05-30
Unify the use of the global unique id for PostgreSQL objects


igorlesnenko:
GraphQL specification requires objects to have global unique id across the whole app: https://graphql.org/learn/global-object-identification/

Currently, we are creating individual global id generator function for every object type, for example:

https://github.com/ParabolInc/parabol/blob/7c60878ee1740fd9b60a1d62c10b584d04802d11/packages/client/shared/gqlIds/IntegrationProviderId.ts#L1-L4

For obejcts stored in postgresql, I suggest to do the following:

- Use `fromGlobalId` and `toGlobalId` functions from https://github.com/graphql/graphql-relay-js package
- Always use `string` type for `id` column in postgresql tables (Use `uuid_generate_v4()` for generating unique id)
- Always have an `id` column

### Acceptance Criteria
   - Change type of existing `id` columns (use `uuid_generate_v4()` to generate unique id)
   - Replace individual functions with `fromGlobalId` and `toGlobalId` where possible
   - Edit PR review guide to include above convention

**Estimated effort:** 21h

Dschoordsch:
I think we could replace the id resolver instead with [`globalIdField`](https://github.com/graphql/graphql-relay-js/blob/d4c533fa0ebe866abe655262bf01999a1e2c984c/src/node/node.ts#L114).

Dschoordsch:
**Story Points: 3**
  [See the discussion](https://action.parabol.co/meet/dNtK0rjNFS/estimate/4) in Sprint Poker #​​1

  *Powered by [Parabol](https://www.parabol.co/resources/sprint-poker-meetings)*

tghanken:
Could whoever picks this up tag me once implementation begins?  I want to ensure there's no impact to our downstream data pipelines (although there shouldn't be as currently outlined).

PR #8191, opened 2023-05-12
chore(prompt-to-join): domainJoinRequest replace string id with auto increment

igorlesnenko:
Based on discussion here I change string id back to auto-increment

How to test:

- Run migration (it will clear all old data)
- Click on `request to join org` in the snackbar, see that new request is created, see that you can open add teammate modal

Old notifications might not work, but I decided to not include a fix, as the feature is not released yet

### DIFF
--- a/packages/server/postgres/migrations/1683902222006_domainJoinRequestAutoIncrementId.ts
+++ b/packages/server/postgres/migrations/1683902222006_domainJoinRequestAutoIncrementId.ts
@@ -1,0 +1,39 @@
+import {Client} from 'pg'
+import getPgConfig from '../getPgConfig'
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+    DROP TABLE IF EXISTS "DomainJoinRequest";
+
+    CREATE TABLE IF NOT EXISTS "DomainJoinRequest" (
+      "id" SERIAL PRIMARY KEY,
+      "createdBy" VARCHAR(100) NOT NULL,
+      "domain" VARCHAR(100) CHECK (lower(domain) = domain) NOT NULL,
+      "expiresAt" TIMESTAMP WITH TIME ZONE DEFAULT NULL,
+      "createdAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+      "updatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+      FOREIGN KEY("createdBy")
+        REFERENCES "User"("id")
+        ON DELETE CASCADE
+    );
+    
+    CREATE UNIQUE INDEX IF NOT EXISTS "DomainJoinRequest_createdBy_domain_unique"
+    ON "DomainJoinRequest" ("createdBy", "domain");
+    
+    DROP TRIGGER IF EXISTS "update_DomainJoinRequest_updatedAt" ON "DomainJoinRequest";
+    CREATE TRIGGER "update_DomainJoinRequest_updatedAt" BEFORE UPDATE ON "DomainJoinRequest" FOR EACH ROW EXECUTE PROCEDURE "set_updatedAt"();
+  `)
+
+  await client.end()
+}
+
+export async function down() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+    DROP TABLE IF EXISTS "DomainJoinRequest";
+  `)
+  await client.end()
+}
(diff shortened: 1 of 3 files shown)


packages/server/postgres/migrations/1683902222006_domainJoinRequestAutoIncrementId.ts
```diff
@@ -0,0 +1,39 @@
+import {Client} from 'pg'
+import getPgConfig from '../getPgConfig'
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+    DROP TABLE IF EXISTS "DomainJoinRequest";
```
igorlesnenko, line 8: Clearing old data, as the feature was not released yet

packages/server/postgres/pg.d.ts
```diff
@@ -181,6 +181,15 @@ export interface MeetingTemplate {
   type: 'action' | 'poker' | 'retrospective' | 'teamPrompt'
   isStarter: Generated<boolean>
   isFree: Generated<boolean>
+  illustrationUrl: string | null
```
Dschoordsch, line 184: -1 l64 needs to be updated, i.e. you should rerun `yarn pg:generate`

Dschoordsch (changes requested):
Thanks for updating!

packages/server/graphql/public/typeDefs/acceptRequestToJoinDomain.graphql
```diff
@@ -6,7 +6,7 @@ extend type Mutation {
     """
     DomainJoinRequest id
     """
-    requestId: ID!
+    requestId: Int!
```
Dschoordsch, line 9: -1 We should keep this `ID!` as it's still an id. If we plan to use it in a connection, we should also make it globally unique by following the gqlId join/split pattern.

packages/server/utils/fromGlobalId.ts
```diff
@@ -0,0 +1,17 @@
+const charSet = 'abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-_:'
+
+function fromGlobalId(globalId: string) {
```
igorlesnenko, line 3: I took approach from  https://github.com/graphql/graphql-relay-js, but wrote my own functions that produces more beautiful global ids

igorlesnenko: Also see #6642

igorlesnenko: Examples:

`Company:123`

graphql-relay: `Q29tcGFueToxMjM=`
own: `8USVGT4Fuvw`

`DomainJoinRequest:99999`

graphql-relay: RG9tYWluSm9pblJlcXVlc3Q6OTk5OTk=
own: `9USGOTcUOTkKW0KYZFCCCCC`

igorlesnenko: own functions produces slightly more shorter ids
graphql-relay function uses base64 and it is probably easier to convert back manually

packages/server/graphql/public/mutations/acceptRequestToJoinDomain.ts
```diff
@@ -31,7 +32,7 @@ const acceptRequestToJoinDomain: MutationResolvers['acceptRequestToJoinDomain']
   const request = await pg
     .selectFrom('DomainJoinRequest')
     .selectAll()
-    .where('id', '=', requestId)
+    .where('id', '=', fromGlobalId(requestId).id)
```
Dschoordsch, line 35: +1 I like the `fromGlobalId` and `toGlobalId` functions. I still think we should create a `gqlIds/DomainJoinRequestId` to wrap these and call `fromGlobalId` and `toGlobalId` internally. This has the advantage to be consistent with the other places and probably easier to find when figuring out how to decode the id.

packages/server/utils/fromGlobalId.ts
Dschoordsch: +1 Not sure this is worth introducing our own encoding schema. It's easy enough to look this up in our code. Shorter ids are nice though.

Dschoordsch (approved):

packages/server/utils/fromGlobalId.ts
igorlesnenko: updated to use `graphql-relay`

packages/server/graphql/public/mutations/acceptRequestToJoinDomain.ts
igorlesnenko: Added

igorlesnenko:
Merging this with parent branch as it will be easier to test

Issue #7674, opened 2023-01-19
Meeting completed slack notification missing meeting details


adaniels-parabol:
## Issue - Bug

Meeting completed slack notification on PPMI is missing meeting details. For example, see the meeting details below. Deployed version is v6.68.0

Meeting Summary
![image](https://user-images.githubusercontent.com/71724289/213338117-c8d3e229-0cdd-4b29-a17b-b9a7b80d5176.png)

Slack Notification
![image](https://user-images.githubusercontent.com/71724289/213338138-46b52127-b811-493f-a69b-73667932302c.png)

The Meeting Summary includes information about reflections and topics, but that is not shown in the slack notification.

- **How serious is the bug?** P3

**Estimated effort:** T.B.D

adaniels-parabol:
We recently incorporated a single integrations auth endpoint with dedicated callback url, does this have anything to do with the issue? Is the auth callback url included in the notifications from slack or possible a different url is used for notifications?

adaniels-parabol:
Per the customer, this started working with no fix in place. Slack issue maybe?

jordanh:
This is very odd and feels "racy" to me. @Dschoordsch, who is scrubbing with me, wonders if this may have to do with the recent adding of AI to meeting summaries (as it's a change that occurred in this area recently).

Keeping this open to see if it happens again

avivapinchas:
Pinging @nickoferrall who's been working on AI summary

adaniels-parabol:
@nickoferrall this is still occurring randomly. I noticed you had an issue merged to resolve (https://github.com/ParabolInc/parabol/pull/7759) . One note, this customer is a PPMI and doesn't have OPENAI configured, is it required to not see this bug? At last check the customer did not want to use OPENAI.

They are on v6.94.0

adaniels-parabol:
@nickoferrall here's a screenshot for reference of the bug.
![image](https://user-images.githubusercontent.com/71724289/225328195-b7f857d1-0d5b-4522-b010-8cfc2ced679b.png)

nickoferrall:
Thanks for the update. I've created a PR https://github.com/ParabolInc/parabol/pull/7926 that _may_ help. I'll look into this more after the retreat.

PR #8197, opened 2023-05-15
fix: slack notification race condition

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7674

Occasionally, the Slack stats are all 0. This could be because:
- The `generateWholeMeetingSummary` is updating the meeting object at the same time the Slack stats are being inserted into the meeting object.
- The meeting object cache is stale when the `SlackNotifier` requests it

### To test

- [ ] End a retro and see an AI Summary and a Slack notification with the accurate meeting stats


igorlesnenko (approved):
Works!

Dschoordsch (approved):
Only did a smoke test. Code looks fine.

Issue #8200, opened 2023-05-16
Fix missing AI Summary


nickoferrall:
@tianrunhe reported in a recent meeting there were no AI Summaries (see [Slack message](https://parabol.slack.com/archives/C042E8PGPD4/p1683822527222089)).

In Sentry, we're getting some 429 errors from Open AI: https://parabol.sentry.io/issues/3912924421/events/?project=107196&referrer=issue-stream&statsPeriod=14d&stream_index=0

Here's a link to OpenAI's rate limits: https://platform.openai.com/account/rate-limits

It looks like there are two different types of rate limits:
1. Rate per minute (RPM)
2. Token per minute (TPM)

I think we have a few options:

- Apply for an increase in our limits
- Change models from `text-davinci-003` to `gpt-3.5-turbo`. This would increase the RPM threshold from 3,000 to 3,500, but the TPM limit would decrease from 250,000 to 90,000.
- See if there's a way to refactor the code to send fewer requests
- Implement [exponential backoff](https://platform.openai.com/docs/guides/rate-limits/retrying-with-exponential-backoff). I think this approach would be tricky for us as we want the AI Summaries soon after we've sent the request

I've emailed OpenAI requesting an increase in our limits.

While waiting for a response, I think we should improve our error handling to help us find out whether the 429 error is due to the RPM or TPM. If it's RPM, we should change models to ChatGPT. If it's TPM, we shouldn't.

If the limit increase request is rejected, we should look into refactoring the code to send fewer requests.

### AC

- [ ] Improve OpenAI error handling to see what type of 429 error we're getting

tianrunhe:
My best guess is that we are suffering from the TPM issue. According to OpenAI's [doc](https://platform.openai.com/docs/models/gpt-3-5), `text-davinci-003` should also be able to handle 4096 tokens. However if I change our config from 256 to 4096, I get consistent 400 errors from OpenAI. I experienced the same issue when I tried to use GPT-3.5.

tianrunhe:
[...]

PR #8201, opened 2023-05-16
chore: improve rate limit error handling

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8200

The Sentry error logs suggest it's an axios error. By sending the `data.error` to Sentry, we should be able to find out if it's an RPM or TPM limit.

I haven't included testing steps in this PR as I'm not sure how you can test it unless you hit a 429 error.

### DIFF
--- a/packages/server/utils/OpenAIServerManager.ts
+++ b/packages/server/utils/OpenAIServerManager.ts
@@ -1,5 +1,6 @@
 import {Configuration, OpenAIApi} from 'openai'
 import sendToSentry from './sendToSentry'
+import {AxiosError} from 'axios'
 
 class OpenAIServerManager {
   private openAIApi: OpenAIApi | null
@@ -35,7 +36,20 @@ class OpenAIServerManager {
       })
       return (response.data.choices[0]?.text?.trim() as string) ?? null
     } catch (e) {
-      const error = e instanceof Error ? e : new Error('OpenAI failed to getSummary')
+      if (e instanceof Error) {
+        const rateLimitStatusCode = 429
+        if (
+          (e as AxiosError).response?.status === rateLimitStatusCode &&
+          (e as AxiosError).response?.data?.error
+        ) {
+          sendToSentry((e as AxiosError).response?.data?.error)
+        } else {
+          sendToSentry(e)
+        }
+        return null
+      }
+
+      const error = new Error('OpenAI failed to getSummary')
       sendToSentry(error)
       return null
     }
--- a/packages/server/utils/OpenAIServerManager.ts
+++ b/packages/server/utils/OpenAIServerManager.ts
@@ -48,7 +48,6 @@ class OpenAIServerManager {
         }
         return null
       }
-
       const error = new Error('OpenAI failed to getSummary')
       sendToSentry(error)
       return null


packages/server/utils/OpenAIServerManager.ts
```diff
@@ -35,7 +36,19 @@ class OpenAIServerManager {
       })
       return (response.data.choices[0]?.text?.trim() as string) ?? null
     } catch (e) {
-      const error = e instanceof Error ? e : new Error('OpenAI failed to getSummary')
+      if (e instanceof Error) {
```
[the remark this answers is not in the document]
nickoferrall: The current Sentry error logs include a ton of info, but not much of it is useful other than the 429 error. It looks like an axios error where we should be able to get more valuable info from that response data. OpenAI says their error response specifies the type of 429 error, so I suspect we can find it in that response data. 

Here are the Sentry logs: https://parabol.sentry.io/issues/3912924421/events/?project=107196&referrer=issue-stream&statsPeriod=14d&stream_index=0

packages/server/utils/OpenAIServerManager.ts
```diff
@@ -35,7 +36,19 @@ class OpenAIServerManager {
...
+        const rateLimitStatusCode = 429
+        if (
+          (e as AxiosError).response?.status === rateLimitStatusCode &&
+          (e as AxiosError).response?.data?.error
+        ) {
+          sendToSentry((e as AxiosError).response?.data?.error)
```
Dschoordsch, line 45: -1 I assume `.error` is a string message and should be wrapped in an error, right?

packages/server/utils/OpenAIServerManager.ts
```diff
@@ -35,7 +36,19 @@ class OpenAIServerManager {
...
+        } else {
+          sendToSentry(e)
+        }
+        return null
+      }
+      const error = new Error('OpenAI failed to getSummary')
```
Dschoordsch, line 51: +1 It would be nice to have the short branch first with an early return, or an else here.

Dschoordsch (changes requested):

nickoferrall:
Having spent more time digging through the OpenAI community support pages, I haven't seen any evidence of more detailed logs available than the "Too many requests" error that we can currently see in Sentry (see [here](https://parabol.sentry.io/issues/3912924421/events/25340a1c0fe94062b30ba08d83cdf390/?project=107196)), so I'll close this PR.

It's still unclear whether we're hitting a token per minute limit (TPM) or a rate per minute (RPM) limit. Hopefully, we can find this out by:
- Looking at the frequency of usage of the API. If we're sending more than 3,000 requests per minute (the davinci limit [here](https://platform.openai.com/account/rate-limits)), we know it's a RPM limit. We should be able to see this in OpenAI's usage page ([here](https://platform.openai.com/account/usage)), but it's broken at the moment, saying "temporarily unavailable".
- Looking into our own meeting data to see how many summaries were created during a minute where we have a Sentry error. 
- Changing `max_tokens` [here](https://github.com/ParabolInc/parabol/pull/8267). If it results in significantly less 429 errors, we know it was a TPM problem.

I sent a request to OpenAI to increase our limits last week. They say they usually take 7-10 business days to respond, so we should hear from them this week or next.

PR #8227, opened 2023-05-18
feat(aiSummary): Replace GPT-3 with GPT-3.5 for aiSummary

tianrunhe:
# Description

Attempt 2 for ChatGPT. In this version, I try to handle cases when ChatGPT couldn't provide a summary.

## Demo

An example meeting with 1 topic with summary and another without. 

Screenshot 2023-05-18 at 15 43 08

## Testing scenarios

- [ ] Happy case
  - Write some valid reflections
  - Verify both of the topic summary and meeting summary is present

- [ ] Junk case
  - Write some nonsense reflections
  - Verify none of the topic summary or meeting summary is present

- [ ] Mixed case
  - Write some nonsense reflections, and some valid reflections
  - Verify topic summary and meeting summary

## Final checklist


nickoferrall (commented):
My concern is that we're currently being rate limited by OpenAI. If it's a RPM limit, ChatGPT should help, but if it's a TPM, it might make the problem worse. See issue here: https://github.com/ParabolInc/parabol/issues/8200

If we can get out limits increased by OpenAI, I would definitely be in favour of switching to ChatGPT. I sent them an email earlier this week, but I haven't heard back from them yet.

tianrunhe:
> If it's a RPM limit, ChatGPT should help, but if it's a TPM, it might make the problem worse. See issue here: #8200

I left a comment on that issue...

packages/server/utils/OpenAIServerManager.ts
```diff
@@ -20,20 +26,28 @@ class OpenAIServerManager {
...
-        model: 'text-davinci-003',
-        prompt: `Below is a comma-separated list of text from a ${location}. Summarize the text for a second-grade student in one or two sentences.
+      const response = await this.openAIApi.createChatCompletion({
+        model: 'gpt-3.5-turbo',
+        messages: [
+          {
+            role: 'user',
+            content: `Below is a comma-separated list of text from a ${location}.
+            Summarize the text for a second-grade student in one or two sentences.
+            If you can't provide a summary, simply write the word "${OpenAIMagicWords.NO_SUMMARY_PROMPT_INSTRUCTION_WORDS}".
```
nickoferrall, line 36: @tianrunhe when you get the chance, would you mind adding the new gender pronoun line to this prompt: `When referring to people in the summary, do not assume their gender and default to using the pronouns "they" and "them".`?

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/RetroTopic.tsx
```diff
@@ -121,7 +121,10 @@ const RetroTopic = (props: Props) => {
 
   const {reflectionGroup, discussion, id: stageId} = stage
   const {commentCount, discussionSummary} = discussion
+  const validDiscussionSummary =
+    discussionSummary && discussionSummary !== OpenAIMagicWords.NO_SUMMARY_RESPONSE
   const {reflections, title, voteCount, topicSummary} = reflectionGroup!
+  const validTopicSummary = topicSummary && topicSummary !== OpenAIMagicWords.NO_SUMMARY_RESPONSE
```
nickoferrall, line 127: -1 I think the server should filter out any invalid summaries rather than checking if it says "No" on the client. 

Rather than the server returning "No", perhaps it could return `null`, so we just need to check if the summary exists?

packages/server/utils/OpenAIServerManager.ts
```diff
@@ -21,20 +27,29 @@ class OpenAIServerManager {
...
-        prompt: `Below is a comma-separated list of text from a ${location}. Summarize the text for a second-grade student in one or two sentences. When referring to people in the summary, do not assume their gender and default to using the pronouns "they" and "them".
+      const response = await this.openAIApi.createChatCompletion({
+        model: 'gpt-3.5-turbo',
+        messages: [
+          {
+            role: 'user',
+            content: `Below is a newline delimited text from a ${location}.
+            Summarize the text for a second-grade student in one or two sentences.
+            When referring to people in the summary, do not assume their gender and default to using the pronouns "they" and "them".
+            If you can't provide a summary, simply write the word "${OpenAIMagicWords.NO_SUMMARY_PROMPT_INSTRUCTION_WORDS}".
```
nickoferrall, line 38: -1 I've copied and pasted the reflections from two previous Retros. For one of them, it was perfect, for another, it returned "No" for a few of them, when it would normally attempt a summary. 

Below is an example of a group of reflections that it would always be able to create a summary for previously but now consistently returns "No".

Screenshot 2023-07-24 at 15 58 34


Perhaps it'd be best to remove this line?

nickoferrall (changes requested):

packages/server/utils/OpenAIServerManager.ts
tianrunhe: If we remove that command from the prompt, sometimes ChatGPT would return like "the text does not provide any information":
Screenshot 2023-07-24 at 14 51 34
I also got "I'm sorry, but the given text does not provide any information that can be summarized for a second-grade student" once.
So I think we need a deterministic null state response so on the customer side, we can provide the same apology.

I can also reproduce your example locally but when I try the exactly same prompt in ChatGPT GUI, I got consistent meaningful answers: 
Screenshot 2023-07-24 at 14 54 22
Does the GUI use GPT-4?
Because with the same set of parameters we have in the app, using `gpt-4` in the playground produces meaningful answer too:
Screenshot 2023-07-24 at 14 57 22

nickoferrall: > sometimes ChatGPT would return like "the text does not provide any information"

I feel like the summary that says, "This is just a bunch of random letters", is a great summary. I often enjoy seeing what it comes up with when there are random reflections like this. 

> "I'm sorry, but the given text does not provide any information that can be summarized for a second-grade student" once.

I think that's a bad summary as it mentions "second-grade student" which is confusing. Maybe we could remove "second-grade student" from the prompt 🤔 

I lean towards not including the `simply write the word "No"` bit of the prompt to encourage more summaries, but feel free to include it if you feel strongly. 

> Does the GUI use GPT-4?

I think you can specify whether ChatGPT should use 3.5 or 4. Maybe the other settings are throwing it off?

nickoferrall (approved):

Dschoordsch (approved):
If you're confident about possible rate limitations, it's good to go.

> 🤖 Topic Summary
The text provided is not sufficient to summarize the meeting for the facilitator as it consists of random characters and does not contain any meaningful information.

PR #8229, opened 2023-05-19
feat(docker-pipeline): arm64 platform built alongside amd64

rafaelromcar-parabol:
# Description

Adds AMD64 images to our pipeline, making life easier for developers with Mac and allowing us to run our application in ARM instances in some Cloud providers.

## Testing scenarios

- [x] A Mac user can docker pull an image produced by this pipeline.
- [ ] A Mac user can run the docker image locally.
- [x] A Linux user can docker pull an image produced by this pipeline.
- [ ] A Linux user can run the docker image locally.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -114,6 +114,7 @@ jobs:
           allow: network.host
           file: ${{ env.PARABOL_DOCKERFILE }}
           context: .
+          platforms: linux/amd64,linux/arm64
           build-args: |
             "_NODE_VERSION=${{ env.NODE_VERSION }}"
             "_SECURITY_ENABLED=true"
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -86,9 +86,11 @@ jobs:
           yarn pg:migrate up && \
           yarn pg:build && \
           yarn build --no-deps
-
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
+      #
+      # Multi-platform build. Uncomment to be able t build multi-platform images.
+      #
+      # - name: Set up QEMU
+      #   uses: docker/setup-qemu-action@v2
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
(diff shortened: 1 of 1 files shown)


mattkrick:
This will require building for 2 archs & then storing both of those images, right?
Could we wait until the need for this arises? Developers do not need a performant docker image & we don't have any customers requiring an ARM build yet

rafaelromcar-parabol:
To be 100% honest, I gave it a try because I thought it would just work smoothly... it isn't the case. I might give it another try in some "free time" this week and close it if I cannot make it work. 

We might want to use ARM machines in GCP (t2a) as they are cheaper than regular ones (only E2 are cheaper and way less convenient. This isn't going to save tons of $$$ but some, and it would be better for the long term. I have yet to try the ARM GCP machines, as I had only tried the AWS ones, but it might be great.

- [GCP instance types](https://cloud.google.com/compute/docs/machine-resource#general-purpose_machine_family_guide)
- [GCP instance pricing](https://cloud.google.com/compute/all-pricing)

Also, I think it would be good to test if you all can docker build in your own machines and architecture. For that, please check [this PR](https://github.com/ParabolInc/parabol/pull/8230)

rafaelromcar-parabol:
It is working.

It added about 2 mins to the docker build process ([example](https://github.com/ParabolInc/parabol/actions/runs/5046716942/jobs/9052623972)). 

WDYT? As it isn't required now, I would understand if we just disable it for now. Should we leave at least commented stuff so we can enable it if we want?

mattkrick:
My .02 is to leave it disabled until we need it. If we're gonna trade in speed or simplicity I want to make sure we're getting something good in return!

rafaelromcar-parabol:
I agree @mattkrick. I'm going to change that just to leave the comments of what needs to be changed in order to build arm64 images for the future, so the time spent here, even if not big, can save some time in the future.

rafaelromcar-parabol:
If you find that useful, merge it. If not, close the PR answering me @mattkrick

mattkrick:
merged in #8403

PR #8230, opened 2023-05-19
chore(docker): documentation on how to build the docker image locally improved

rafaelromcar-parabol:
# Description

Documentation on how to build the docker image locally is improved.

## Testing scenarios

- [ ] Mac user can build an image from M1 or M2 models.
- [ ] Linux user can build an image.

### DIFF
--- a/docker/parabol-ubi/docker-build/README.md
+++ b/docker/parabol-ubi/docker-build/README.md
@@ -55,14 +55,50 @@ export _DOCKER_TAG=test-image
 
 The application must be already built locally using the command `yarn build --no-deps` mode.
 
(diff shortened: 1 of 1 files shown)


rafaelromcar-parabol:
@mattkrick @Dschoordsch could you follow the documentation and tell me if you find it clear enough?

The how-to run the application should also be improved, but as I don't see a lot of use cases right now on running the application just with docker commands manually, I'll skip it for today and do another PR whenever we need it or I have time to spend on it.

mattkrick:
not tested but LGTM. We can revisit when building docker images locally becomes more of a priority. For now I'm happy to just build in CI

Issue #7813, opened 2023-02-24
Checkout Flow: add tracking


nickoferrall:
Discuss with @tghanken which events we would like to capture. 

Previously we were tracking when the credit card model was opened, but landing on the Organization page might give us less signal that a user may be interested in upgrading.

Perhaps we measure when a user:
- visits the org page
- begin adding credit card details
- click on a plan
- click upgrade

### AC

- [ ] Discuss what to track with Taylor
- [ ] Implement tracking for the relevant events

tghanken:
I think you have the right events outlined already! One clarifying question around the design: Can payment details be added independently of adding/changing plans? If so, we'd want events when they start adding payment details, and when they save the card details.

We'd also want to track a "Modify Organization Billing Leaders" event when users add/remove/edit the list of billing leaders.

I think the list of events looks like this:
- [x] Organization page view (automatically handled by Segment)
- [ ] "Plan Tier Selected" event w/ a tier parameter showing the selected plan tier
- [ ] "Payment Details Started" event
- [ ] "Payment Details Completed" even on CC details save
- [x] "Organization Upgraded" - should already be emitted when the upgrade completes, don't need to add anything
- [ ] "Billing Leader Modified" w/ a modification type enum (add/remove/update) and the userId of the modified leader

tghanken:
> Yeah, I think we'll need the option to edit card details that have previously been submitted, if that's what you're referring to?

Yeah, exactly.  I think if we emit those payment details events each time they edit, that should be sufficient!

PR #8235, opened 2023-05-19
chore: add checkout flow tracking

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/7813

### Testing Steps

See that we're tracking the following event in Segment:

- [ ] Select a plan tier
- [ ] Started entering card details
- [ ] Finished entering card details
- [ ] Billing leader added/removed

### DIFF
--- a/packages/server/utils/analytics/analytics.ts
+++ b/packages/server/utils/analytics/analytics.ts
@@ -85,6 +85,7 @@ export type AnalyticsEvent =
   | 'Downgrade Clicked'
   | 'Downgrade Continue Clicked'
   | 'Organization Downgraded'
+  | 'Billing Leader Modified'
 
   // task
   | 'Task Created'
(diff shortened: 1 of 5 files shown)


packages/server/utils/analytics/analytics.ts
```diff
@@ -387,6 +388,20 @@ class Analytics {
       newName
     })
 
+  billingLeaderModified = (
+    userId: string,
+    viewerId: string,
+    orgId: string,
+    modificationType: 'add' | 'remove'
```
nickoferrall, line 395: The [issue](https://github.com/ParabolInc/parabol/issues/7813#issuecomment-1495078347) mentioned including a `modificationType` with `add`, `update`, and `remove`, but we're only able to add and remove billing leaders at the moment, so I've included these two options

packages/server/utils/analytics/analytics.ts
```diff
@@ -387,6 +388,20 @@ class Analytics {
       newName
     })
 
+  billingLeaderModified = (
+    userId: string,
```
nickoferrall, line 392: In `analytics.ts`, `userId` is usually an argument, but we're passing in the `viewerId`, e.g. [here](https://github.com/ParabolInc/parabol/blob/4d892690c8f5a252e884a80b49ff3f31be9bbe2f/packages/server/graphql/mutations/startCheckIn.ts#L101) and [here](https://github.com/ParabolInc/parabol/blob/e17d1124d06c553751f6262be2d1ad0477b034c2/packages/server/graphql/mutations/removeAtlassianAuth.ts). 

When adding/removing the billing leader, the viewer can be different from the user that is being updated, and we probably want to track both.

tghanken (approved):
Looks good!

Issue #8243, opened 2023-05-22
Refactor addFeatureFlagToOrg to updateOrgFeatureFlag


nickoferrall:
At the moment, we can add feature flags to orgs but we can't remove them. See Slack discussion where this came up [here](https://parabol.slack.com/archives/C042E8PGPD4/p1684777061102139).

### AC

- [ ] `addFeatureFlagToOrg` is refactored to `updateOrgFeatureFlag`
- [ ] Can remove org feature flags using the mutation in the same way `updateFeatureFlag` currently works

PR #8244, opened 2023-05-22
chore: refactor addFeatureFlagToOrg to updateOrgFeatureFlag

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8243

I don't see any documentation in Notion about an org feature flag, so I created my own: https://www.notion.so/parabol/How-to-enable-or-remove-an-org-feature-flag-4b4be190ed674bf8903340d6b5fa50e4

### To test

- [ ] Can add an org feature flag
- [ ] Can remove an org feature flag

### DIFF
--- a/codegen.json
+++ b/codegen.json
@@ -10,7 +10,7 @@
       "config": {
         "contextType": "../graphql#InternalContext",
         "mappers": {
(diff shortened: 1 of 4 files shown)


igorlesnenko (approved):
Works!

PR #8246, opened 2023-05-22
fix: NewMeeting dialog rendered the dashboard twice

Dschoordsch:
The backgroundLocation approach does not work properly when nesting, so move the NewMeeting dialog a level up as well.
This caused the background dashboard to be rendered twice

## Demo

https://www.loom.com/share/c18f23bd2c4444bda7dda015b4096b0b

## Testing scenarios

Easiest way for me to visually reproduce:
* on meeting dashboard press new meeting button
* go to sprint poker
* press on the template
* [ ] on master the whole app is scrolled up, dialog is half visible

Also check that the dashboard is not added twice to the DOM when the new meeting dialog is shown.

## Final checklist

### DIFF
--- a/packages/client/components/Dashboard.tsx
+++ b/packages/client/components/Dashboard.tsx
@@ -1,9 +1,8 @@
 import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
-import {Location} from 'history'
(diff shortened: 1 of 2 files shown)


igorlesnenko (approved):
Nice catch! Didn't notice at all

Issue #8247, opened 2023-05-22
Unable to edit comments in threads (for Standups, Check-Ins)


ackernaut:
### Reproduction steps
- Open a Standup or Check-In
- Proceed to a thread (on a Standup response card, or on a Check-In agenda item)
- Add a comment to the thread
- You can make edits, but upon submission, the text reverts to the original comment text
- This doesn’t seem to affect Retros and Sprint Poker, but it may be good to double check

![2023-05-22 15 06 03](https://github.com/ParabolInc/parabol/assets/307286/0e72c4c6-1695-49cc-9368-9fb2847d5516)

PR #8164, opened 2023-05-05
feat: can edit & delete AI comment

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8096

Screenshot 2023-05-05 at 17 16 38

### To test

- [ ] Create a retro, add comments, go to the Discuss phase and see an AI Summary in the discussion thread
- [ ] Can edit the comment
- [ ] Can delete the comment

PR #8248, opened 2023-05-22
fix(discussions): Fix editing comments in standups + check-ins

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8247

In https://github.com/ParabolInc/parabol/pull/8164, we added some logic where we needed to confirm that a discussion was part of a meeting. However, the logic that checked if a discussion is part of a meeting was only implemented for poker and retros, so editing / deleting comments in check-ins and standups is broken.

## Testing scenarios
- [ ] Edit comments in all meeting types
- [ ] Delete comments in all meeting types

## Final checklist


packages/server/graphql/mutations/deleteComment.ts
```diff
@@ -50,7 +49,8 @@ const deleteComment = {
       return {error: {message: `Not a member of the meeting`}}
     }
     const {createdBy, discussionId} = comment
-    if (!isDiscussionFromMeeting(discussionId, meeting)) {
+    const discussion = await dataLoader.get('discussions').loadNonNull(discussionId)
+    if (discussion.meetingId !== meetingId) {
```
jmtaber129, line 53: `discussion.meetingId === meetingId` is probably the best and simplest source of truth for whether a discussion is part of the meeting, which is why I fixed this by simply fetching the discussion here instead of trying to implement `isDiscussionFromMeeting` for check-ins and standups

nickoferrall (approved):
LGTM!

PR #8257, opened 2023-05-23
fix: checkbox color overriden by styled components

igorlesnenko:
After rewriting `Checkbox` to tailwind, we can't simply override color with style component.

This happens because tailwind defines clasees like `body .text-slate-600` which is more specific than classname added by styled components.

By using `&&` it is possible to make styled components class more specific, it will look like `.css-ry6kcf.css-ry6kcf`

image

**How to test:**
- see that checkbox shows the correct color

**Alternative possible approach:** refactor `Checkbox` and introduce `size` prop and `color` prop

Because of order of classes in tailwind stylesheet, one class would always be preferable:
image

**Another approach:**
Remove default color at all

### DIFF
--- a/packages/client/components/Checkbox.tsx
+++ b/packages/client/components/Checkbox.tsx
@@ -4,7 +4,6 @@ import clsx from 'clsx'
 
 interface Props {
   active: boolean | null
-  color?: string
   className?: string
   disabled?: boolean
   onClick?: (e: React.MouseEvent) => void
--- a/packages/client/components/EmailNotifications.tsx
+++ b/packages/client/components/EmailNotifications.tsx
@@ -27,10 +27,12 @@ const ErrorMessage = styled(StyledError)({
 })
 
 const StyledCheckbox = styled(Checkbox)({
-  width: 24,
-  height: 24,
-  marginRight: 8,
-  color: PALETTE.SKY_500
+  '&&': {
+    width: 24,
+    height: 24,
+    marginRight: 8,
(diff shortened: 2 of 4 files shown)


Dschoordsch:
Just forwarding this one to @BartoszJarocki as he's the tailwind expert

BartoszJarocki:
@igorlesnenko well, some time ago we had a conversation about overriding components styled using styled components with tailwind [here](https://parabol.slack.com/archives/C608QL4RH/p1682002666015809). Since then, we have the solution that adds `body tailwind-class` that increases specificity of tailwind utility classes. It's configured [here](https://github.com/ParabolInc/parabol/blob/9ed3947f7ed9a147f5b79633d49af05ae96ebffc/tailwind.config.js#L5). If you remove that part of the config, it'll be easy to override tailwind styles but it might break other places. I still think it's a right choice to allow override stuff with tailwind, than other way around as we're migrating away from styled components. I'll spend some time and try to figure out if there's any other way. Will let you know.

BartoszJarocki (approved):
I couldn't find a better way except for rewriting given components with tailwind. Let's go with the solution you proposed

igorlesnenko:
I think in future, additional props is a good way to go, because as far as I know tailwind is not allowing
```
svg: {
      fontSize: 28
    }
```
It is also easier to specify one number instead of 3

However the question is how `color` prop should look like

igorlesnenko:
created an issue #8273

Issue #7885, opened 2023-03-08
Standups: Add meeting ended notifications


BartoszJarocki:
AC
- [ ] when meeting ends send a meeting ended notification just like     for other meeting types (slack, mattermost etc)

[:lock: Slack Message](https://parabol.slack.com/archives/C03THUWKVQ8/p1678245670827799)

gcrickman:
A [customer](https://app.hubspot.com/contacts/3888472/company/7348403968) 🔒 writes:

> It wasn't clear that everyone on the team would need to connect to slack via the team view --> integrations link. We are piloting async standup via Parabol with this small 5 person team. Now that they've done that and I've removed the recurrences for the previous meetings and started a new one. They get a notification via the Parabol app in slack when meetings start but haven't gotten anything when a meeting ends. Given we spin up teams and dismantle them frequently based on project lifespans it's a bit more overhead than we were hoping for to make everyone have to do that every time we create a team in Parabol. Given that's required, it seems, it would be rad if they only had to do that once at their account level and then any team they were allocated to, they'd then get notifications via slack.
> 
> Why do we choose a channel to have Parabol post to if nothing is posted to the channel? We were hopeful that the async standup summaries would post to the designated channel to make it easy on the production team folks managing the teams. We used to use another tool called Status Hero that did this and were stoked to see this

PR #8258, opened 2023-05-23
feat(standups): Meeting ended integration notification

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/7885

Implements `getSummaryText` and integration-specific buttons for standups, and adds `IntegrationNotifier.endMeeting` call to `safeEndTeamPrompt` (same place where we send meeting summaries)

## Demo

Screen Shot 2023-05-23 at 3 31 01 PM

## Testing scenarios
- [ ] Smoke test standup ended notifications

## Final checklist

### DIFF
--- a/packages/server/graphql/mutations/helpers/notifications/getSummaryText.ts
+++ b/packages/server/graphql/mutations/helpers/notifications/getSummaryText.ts
@@ -6,8 +6,9 @@ import {isMeetingPoker} from '../../../../database/types/MeetingPoker'
 import {isMeetingRetrospective} from '../../../../database/types/MeetingRetrospective'
 import {isMeetingTeamPrompt} from '../../../../database/types/MeetingTeamPrompt'
 import sendToSentry from '../../../../utils/sendToSentry'
+import {getTeamPromptResponsesByMeetingId} from '../../../../postgres/queries/getTeamPromptResponsesByMeetingIds'
 
-const getSummaryText = (meeting: Meeting) => {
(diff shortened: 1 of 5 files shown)


packages/server/graphql/mutations/helpers/notifications/getSummaryText.ts
```diff
@@ -39,7 +40,11 @@ const getSummaryText = (meeting: Meeting) => {
       'comment'
     )}.`
   } else if (isMeetingTeamPrompt(meeting)) {
-    return 'TODO: Implement teamPrompt summary text'
+    const responseCount = (await getTeamPromptResponsesByMeetingId(meeting.id)).filter(
```
BartoszJarocki, line 43: +1 this will do db hit for every integration configured, we may want to use similar pattern as retro ie. store the response count in the meeting object that'd be set when meeting ends and just read it here

BartoszJarocki (approved):
looks good, one suggestion left

packages/server/graphql/mutations/helpers/notifications/getSummaryText.ts
jmtaber129: We decided against that pattern as part of CR in https://github.com/ParabolInc/parabol/pull/6529. We could reduce the DB calls by moving `getTeamPromptResponsesByMeetingId` to a dataloader, but I don't think doing so is important enough for this PR, especially since there's already an index for this table on `meetingId`.

Issue #7953, opened 2023-03-28
Recurring meetings may start at a wrong time


BartoszJarocki:
For some reason, recurring meetings may start at a different time than configured.
See a discussion [🔒 here.](https://parabol.slack.com/archives/C03THUWKVQ8/p1677768374408449)

### Acceptance Criteria (optional)
- [ ] root cause identified
- [ ] depending on the root cause either fix implemented or solution proposed

**Estimated effort:** 4 hour to triage. More if root cause is already identified.

PR #8269, opened 2023-05-24
fix(recurrence): Incorrect start times due to timezone weirdness

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/7953

The `rrule` library we're using handles times + time zones in a way that's very... strange. Basically, `rrule` assumes the UTC equivalent of all dates it sees are actually just in the system's time zone. So a time of `12:00:00Z` for a system in PST is actually interpreted by `rrule` as 12 PST, or `21:00:00Z` true UTC.

The immediate fix is simple but tedious - convert these "UTC" local time zone times returned by `rrule` to the true JS `Date`, and convert JS `Date`s to the weird "local as UTC" time before passing them to any `rrule` methods.

The long-term fix is probably to rip out the `rrule` lib and roll our own (or fork the lib and make changes for improved time zone support), but that's probably not worth it for now.

## Demo
https://www.loom.com/share/b15157c0c1714513bfa1547c2669caf5

## Testing scenarios
Full test of recurrence, across time zones:
- [ ] Set up the environment by starting your dev server, changing your system time zone, then reloading the client
  - [ ] This will put your server in a different time zone from your client, which is when these bugs manifest
- [ ] Start a new team prompt meeting, and start recurrence
- [ ] Confirm that the next meeting start time is correct
- [ ] Open the edit recurrence modal, and confirm the default start time is the same as what was selected when starting recurrence
- [ ] Change the start time, confirm the update, and confirm the next meeting start time is updated to the correct time
- [ ] Process recurrence
  - [ ] Make sure no other meeting series are active except the one created earlier
  - [ ] Change your system time to just before the restart time, run the `processRecurrence` mutation, and confirm no meetings were started or ended
  - [ ] Change your system time to just after the restart time, run the `processRecurrence` mutation, and confirm the meeting was restarted (one meeting started, one meeting ended)
  - [ ] Confirm the new meeting has the correct end time.

## Final checklist

### DIFF
--- a/packages/client/components/TeamPrompt/Recurrence/RecurrenceSettings.tsx
+++ b/packages/client/components/TeamPrompt/Recurrence/RecurrenceSettings.tsx
@@ -2,7 +2,7 @@ import clsx from 'clsx'
 import dayjs from 'dayjs'
 import utcPlugin from 'dayjs/plugin/utc'
 import React, {PropsWithChildren, useEffect} from 'react'
-import {Frequency, RRule, datetime} from 'rrule'
+import {Frequency, RRule} from 'rrule'
 import {MenuPosition} from '../../../hooks/useCoords'
 import useMenu from '../../../hooks/useMenu'
 import {PortalId} from '../../../hooks/usePortal'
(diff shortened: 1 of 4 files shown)


packages/client/components/TeamPrompt/Recurrence/RecurrenceSettings.tsx
```diff
@@ -251,7 +252,13 @@ export const RecurrenceSettings = (props: Props) => {
             freq: Frequency.WEEKLY,
             interval: recurrenceInterval,
             byweekday: recurrenceDays.map((day) => day.rruleVal),
-            dtstart: dayjs(recurrenceStartTime).utc().toDate(),
+            dtstart: datetime(
```
BartoszJarocki, line 255: +1 should we reuse method you created [here](https://github.com/ParabolInc/parabol/pull/8269/files#diff-021df5c0b52eff71f137510aa3223218ab5cf2d9ef15e1fb04e218210322965eR3)?

BartoszJarocki (approved):
great find, as far as I was able to test, should be all good!

packages/client/components/TeamPrompt/Recurrence/RecurrenceSettings.tsx
jmtaber129: Yep! Fixed

Issue #8272, opened 2023-05-24
Bug: RiD Custom Retro Template no longer allows me to choose prompt colors


jordanh:
## Issue - Bug

This is a "bug-hancement": the new RiD custom retro activity flow no longer allows me to choose a color for each prompt label

![image](https://github.com/ParabolInc/parabol/assets/103000/cd57f78b-908a-4d93-a66f-7f518aa7080e)

Repro steps:
- Create a custom retro activity template
- Configure a prompt
- Hover or click the color, note there is no longer a dropdown to choose the prompt color

### Acceptance Criteria (optional)
Users can:
   - Set custom prompt colors when creating a retro template

Triage is performed to find the root cause of the bug, timeboxed to ~1 hour. 

**Estimated effort:** 1 hour to triage. More if root cause is already identified.

PR #8279, opened 2023-05-25
fix(rid): Fix changing prompt colors in activity library

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8272

The prompt color picker assumed that it's in the old template picker modal, so it threw an error when it couldn't find the template picker modal in the activity library.

Refactor a bit to pass the parent modal ID down from the parent.

## Demo
https://www.loom.com/share/6c69352d8fda44bfbde26f39dbe28346

## Testing scenarios
- [ ] Change template color in old template picker
- [ ] Change template color in activity library.

## Final checklist

### DIFF
--- a/packages/client/modules/meeting/components/EditableTemplatePromptColor.tsx
+++ b/packages/client/modules/meeting/components/EditableTemplatePromptColor.tsx
@@ -11,11 +11,13 @@ import PalettePicker from '../../../components/PalettePicker/PalettePicker'
 import {MenuPosition} from '../../../hooks/useCoords'
 import useMenu from '../../../hooks/useMenu'
 import {PALETTE} from '../../../styles/paletteV3'
+import {PortalId} from '../../../hooks/usePortal'
 
 interface Props {
   isOwner: boolean
   prompt: EditableTemplatePromptColor_prompt$key
   prompts: EditableTemplatePromptColor_prompts$key
+  parentId?: PortalId
 }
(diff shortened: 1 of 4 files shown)


packages/client/modules/meeting/components/EditableTemplatePromptColor.tsx
```diff
@@ -78,7 +80,7 @@ const EditableTemplatePromptColor = (props: Props) => {
   const {groupColor} = prompt
   const {menuProps, menuPortal, originRef, togglePortal} = useMenu<HTMLButtonElement>(
     MenuPosition.UPPER_LEFT,
-    {parentId: 'templateModal'}
+    {parentId: parentId}
```
BartoszJarocki, line 83: +1 are there any use cases where `parentId` is undefined? if yes, will it work correctly? I see previously it was always defined as `templateModal` now, it's nullable

BartoszJarocki (approved):

packages/client/modules/meeting/components/EditableTemplatePromptColor.tsx
jmtaber129: Yes, when the component is not rendered in a modal, `parentId` will be `undefined`. Previously, `parentId` was always `templateModal`, which caused an error when this was rendered outside of a modal.

Issue #8250, opened 2023-05-23
Checkout Flow: add org details


nickoferrall:
Add Organization Details as shown in this Figma design: https://www.figma.com/file/SyJdkzumWsUJo58cmBGgus/Checkout-concepts?type=design&node-id=161-2088&t=PyHGRNDput2zbLZg-0

### AC

- [ ] Add Organization Details to the sidebar
- [ ] On the page, the user can edit the org name
- [ ] The user can delete the org just like they can in the current org billing page

PR #8280, opened 2023-05-25
feat(checkout-flow): add org details 

nickoferrall:
Fix: https://github.com/ParabolInc/parabol/issues/8250

org-details

### To test

- [ ] Add the checkout-flow user feature flag
- [ ] Click on Organization Details in the sidebar
- [ ] Can edit the org name
- [ ] Can delete the org

### DIFF
--- a/packages/client/components/Dashboard/MobileDashSidebar.tsx
+++ b/packages/client/components/Dashboard/MobileDashSidebar.tsx
@@ -5,7 +5,7 @@ import {useFragment} from 'react-relay'
 import {useRouteMatch} from 'react-router'
 import {PALETTE} from '../../styles/paletteV3'
 import {NavSidebar} from '../../types/constEnums'
-import {BILLING_PAGE, MEMBERS_PAGE} from '../../utils/constants'
+import {BILLING_PAGE, MEMBERS_PAGE, ORG_DETAILS_PAGE} from '../../utils/constants'
 import {DashSidebar_viewer$key} from '../../__generated__/DashSidebar_viewer.graphql'
 import DashNavList from '../DashNavList/DashNavList'
 import StandardHub from '../StandardHub/StandardHub'
@@ -133,6 +133,12 @@ const MobileDashSidebar = (props: Props) => {
                 icon={'group'}
                 href={`/me/organizations/${orgId}/${MEMBERS_PAGE}`}
                 label={'Members'}
+              />
+              <LeftDashNavItem
+                onClick={handleMenuClick}
+                icon={'work'}
+                href={`/me/organizations/${orgId}/${ORG_DETAILS_PAGE}`}
+                label={'Organization Details'}
               />
             </NavItemsWrap>
           </Nav>
--- a/packages/client/components/Dashboard/MobileDashSidebar.tsx
+++ b/packages/client/components/Dashboard/MobileDashSidebar.tsx
@@ -5,7 +5,7 @@ import {useFragment} from 'react-relay'
 import {useRouteMatch} from 'react-router'
 import {PALETTE} from '../../styles/paletteV3'
 import {NavSidebar} from '../../types/constEnums'
-import {BILLING_PAGE, MEMBERS_PAGE, ORG_DETAILS_PAGE} from '../../utils/constants'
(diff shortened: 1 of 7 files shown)


packages/client/components/Dashboard/MobileDashSidebar.tsx
```diff
@@ -134,6 +134,12 @@ const MobileDashSidebar = (props: Props) => {
                 href={`/me/organizations/${orgId}/${MEMBERS_PAGE}`}
                 label={'Members'}
               />
+              <LeftDashNavItem
```
nickoferrall, line 137: The [designs](https://www.figma.com/file/SyJdkzumWsUJo58cmBGgus/Checkout-concepts?type=design&node-id=161-2088&t=xvkm9h4kV0Or6GQi-0) have Organization Details at the top of the sidebar, but I think the billing page should be the top priority, and this should be the lowest priority, so I've put it at the bottom of the sidebar.

Let me know if you disagree @acressall 🙏

acressall: Oooh I like this idea. Maybe later we can change the name to "organization settings" since it will include the AI toggle and potentially other things

acressall (approved):

igorlesnenko (approved):
Works! Deleting org feels a bit weird, redirect without notification

igorlesnenko (approved):
Works! Deleting org feels a bit weird, redirect without notification

packages/client/modules/userDashboard/components/OrgBilling/OrgDetails.tsx
```diff
@@ -0,0 +1,67 @@
...
+        createdAt
+        picture
+        tier
+        name
+      }
+    `,
+    organizationRef
+  )
+  const {orgId, createdAt, isBillingLeader, picture: orgAvatar, name, tier} = organization
+  const pictureOrDefault = orgAvatar || defaultOrgAvatar
```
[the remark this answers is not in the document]
nickoferrall: Yep, the picture field is nullable: https://github.com/ParabolInc/parabol/blob/5435f855bf85ee880001209e139b4c47f903e55d/packages/server/graphql/types/Organization.ts#L61

packages/client/components/Dashboard/MobileDashSidebar.tsx
nickoferrall: Great! Updated to Organization Settings 👍

packages/client/utils/constants.ts
```diff
@@ -92,6 +92,7 @@ export const NOTIFICATIONS = 'notifications'
 /* Org Settings */
 export const BILLING_PAGE = 'billing'
 export const MEMBERS_PAGE = 'members'
+export const ORG_SETTINGS_PAGE = 'org-settings'
```
Dschoordsch, line 95: +1 because it's on a subpage of `organizations/:orgId/` the `org-` part is redundant and could be removed.

packages/client/modules/userDashboard/components/OrgBilling/OrgBillingDangerZone.tsx
```diff
@@ -54,15 +58,23 @@ const OrgBillingDangerZone = (props: Props) => {
         ...ArchiveOrganization_organization
         isBillingLeader
         tier
+        viewerOrganizationUser {
```
Dschoordsch, line 61: -1 I see that if the user has the feature flag, then this component is used in a different component tree. I think this component should not know how it's used. Instead I see 2 possible cleaner solutions:
* add a prop `isWide` to the component. hacky but good enough for now
* add a style/className property and set the style directly on the parent

packages/client/modules/userDashboard/components/OrgBilling/OrgDetails.tsx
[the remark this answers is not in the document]
Dschoordsch: If an empty string would be valid, then we should check with `??` instead of `||`

Dschoordsch (changes requested):
Only minor issues

Dschoordsch (approved):

PR #8285, opened 2023-05-26
fix: do not cancel stripe subscription in case of failed payment

### LINKED ISSUE #8166: Billing: Failed Payments should not immediately cancel Stripe Subscriptions
## Issue - Bug

Currently, failed payments for `Team` subscriptions automatically cancel the Stripe subscription.  This can lead to some unexpected billing behavior, such as unpaid balances failing to carry forward, as seen [here](https://parabol.slack.com/archives/C4JAUUZ9P/p1683154782014419) 🔒.  Instead of automatically cancelling the subscriptions, we should allow them to go "Past Due", and allow users to change their payment information on the original subscription.

- Allow payments to fail without the subscription being automatically cancelled for previously stable subscriptions
- Changing payment information retries billing on failed invoices
- Continue to ensure new `Team` subscriptions cannot be created without a valid form of payment

igorlesnenko:
Fix #8166

I changed the following [subscriptions settings](https://dashboard.stripe.com/settings/billing/automatic) on stripe, it should not affect prod, as we currently terminate subscriptions from the code on failed payments anyway:

image

image

How to test:

Case 1:

- Setup webhook (see #8072)
- Create a new org
- Upgrade using `4000000000000341` card (It will fail later after upgrade)
- Payment failed event should be received on the webhook
- Refresh page, see teams are locked
- Open stripe, see that subscription is canceled
- Upgrade again with working card `4242 4242 4242 4242`
- See that a new `active` subscription was created

Case 2: 

* make sure the webhook is active
* create a new org
* Uncomment: `// trial_end: toEpochSeconds(new Date(Date.now() + 1000 * 10))`, in `createTeamSubscription`
* Upgrade with [4000 0000 0000 0341](https://stripe.com/testing#cards) card (will fail after trial)
* Wait few minutes
* Go to the subscription https://dashboard.stripe.com/test/subscriptions
* Find draft invoice (or wait for it to be appeared)
* Finalize the invoice — edit and click charge (to avoid waiting 1 hour)
* See that teams were locked
* See that the subscription went to past due state but not canceled
* Update card on organization: `4242 4242 4242 4242`
* See all teams unlocked, subscription became `active`, no new subscription created, invoice is paid


packages/server/graphql/private/mutations/stripeFailPayment.ts
```diff
@@ -56,10 +57,19 @@ const stripeFailPayment: MutationResolvers['stripeFailPayment'] = async (
...
 
-  if (paid || stripeSubscriptionId !== subscription) return {orgId}
+  if (paid || stripeSubscriptionId !== subscription || stripeSubscriptionId === null) return {orgId}
 
   // RESOLUTION
-  await terminateSubscription(orgId)
+  const subscriptionObject = await manager.retrieveSubscription(stripeSubscriptionId)
+
+  if (subscriptionObject.status === 'incomplete' || subscriptionObject.status === 'canceled') {
+    // Terminate subscription if it is the first failed payment, or it is already canceled
```
tianrunhe, line 66: Why terminate the subscription on **1st** failed payment?

tianrunhe (changes requested):
I don't quite follow why we are still canceling subscriptions on first failed payment? I thought the test case 1 is what's happening today and we want to change against it?

packages/server/graphql/private/mutations/stripeFailPayment.ts
igorlesnenko: @tianrunhe I keep what is happening today only for case 1: 
User setup the subscription for the first time, he adds a card but then almost immediately first payment fails.

I keep this for two reasons:
- I think it is safe to cancel subscription in this case.
- Stripe [documentation](https://stripe.com/docs/billing/subscriptions/overview#payment-window) suggest to cancel such subscriptions if no payment was made within 23 hours. 
To not handle this 23 hours case specifically I think it is easier to just cancel right now:

```
Payment window
Customers have about 23 hours to make a successful payment. The subscription remains in status incomplete and the invoice is open during this time. If your customer pays the invoice, the subscription updates to active and the invoice to paid. If they don't make a payment, the subscription updates to incomplete_expired and the invoice becomes void.

This window exists because your customer usually makes the first payment for a subscription while on-session. If the customer returns to your application after 23 hours, create a new subscription for them.
```

igorlesnenko: Added more comments in code. What do you think?

tianrunhe:
I don't like that it cancels the subscription and lock the team immediately, for 3DS authentication needed card (test card number: `4000002500003155` from [here](https://stripe.com/docs/testing?testing-method=card-numbers#regulatory-cards)). We have [customer contacted us](https://parabol.slack.com/archives/C4JAUUZ9P/p1681487258938689) for this use case before.

tianrunhe (approved):
All good then!

mattkrick:
Ooooo this is some tricky stuff, thank you for tackling it!

So as I understand the business logic, we cancel the subscription in stripe if it has been alive for < 1 day (failed payment, delayed failed payment).

If a subsequent payment fails (e.g. CC canceled/expired) then we lock them out of their team, but keep the subscription running. That means if it takes them a month to update their CC, _they owe us for that month even though they haven't used us._ Is that OK? Not saying it's good or bad, just want to make sure that's the intended effect.

### Code Review

We use `stripeSubscriptionId` as a proxy for an org being paid, so we'll need to make sure that across the codebase we no longer do that. For example we need to fix https://github.com/ParabolInc/parabol/blob/7a99a7cb71d82db85381d70d7d4a5f1bd362dc30/packages/server/graphql/mutations/moveTeamToOrg.ts#L73

igorlesnenko:
@mattkrick 
> So as I understand the business logic, we cancel the subscription in stripe if it has been alive for < 1 day (failed payment, delayed failed payment).

Yes, we cancel subscription if it is in `incomplete` state and payment fails (no other payments were made in past)

> If a subsequent payment fails (e.g. CC canceled/expired) then we lock them out of their team, but keep the subscription running. That means if it takes them a month to update their CC, they owe us for that month even though they haven't used us. Is that OK? Not saying it's good or bad, just want to make sure that's the intended effect.

Yes, this is the expected behavior at this time. We want to improve by giving some period before locking the org, andonly then cancel the subscription: #8341

> We use stripeSubscriptionId as a proxy for an org being paid, so we'll need to make sure that across the codebase we no longer do that. For example we need to fix

Thanks! This seems to be the only case. Fixed!

Issue #8278, opened 2023-05-25
Checkout Flow: refactor UpgradeToTeamTierMutation to separate subscription and upgrade logic


nickoferrall:
`UpgradeToTeamTierMutation` currently creates the Stripe subscription, and it upgrades the user to the Team tier. 

With 3D Secure cards, this is a problem because we need to use the response from the created but incomplete Stripe subscription to confirm the subscription in the client. If the user fails to authorise their 3D Secure card, the Stripe subscription won't be created, but they will be upgraded.

See https://github.com/ParabolInc/parabol/pull/8162 for further context.

### AC

- [ ] Split the mutation up into two mutations
- [ ] If a user fails to authorise their 3D Secure card, their team is not upgraded to Team

Issue #8006, opened 2023-04-06
Checkout Flow: Handle UI changes after a user has upgraded


nickoferrall:
Predecessor: https://github.com/ParabolInc/parabol/issues/7693

### AC 

After a user has upgraded to the Team tier, update the following UI:

- [ ] Show the Plans section at the bottom of the page and show Invoices at the top. See comment on what happens after upgrading [here](https://www.figma.com/file/SyJdkzumWsUJo58cmBGgus?node-id=209:3567#345588008)
- [ ] After downgrading, if the user goes to upgrade again, their credit card info, e.g. CC number, is no longer visible. See [discussion](https://parabol.slack.com/archives/C042E8PGPD4/p1680277374115349) 🔒 
- [ ] After upgrading, the right draw opens by default. See https://github.com/ParabolInc/parabol/pull/7986

PR #8289, opened 2023-05-30
chore(checkout-flow): handle failing 3D Secure card auth

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8278

Demo: https://www.loom.com/share/e5bdd83d0ab84b0bad5ac622af958ff5

### To Test

- [ ] Enter a 3D secure card (see examples [here](https://stripe.com/docs/payments/3d-secure?locale=en-GB#three-ds-cards)) such as 4000000000003220
- [ ] Fail the authorisation
- [ ] Refresh the page and see that the org hasn't upgraded
- [ ] Now enter the card, successfully complete the authorisation, and see that the org successfully upgrades


igorlesnenko (changes requested):
1) I unable to upgrade, getting this error:

image

```
18:41:31 1|GraphQL Executor | TypeError: Cannot destructure property 'last4' of 'creditCard' as it is undefined.
18:41:31 1|GraphQL Executor |     at stripeFailPayment (/Users/igor/www/parabol/dev/gqlExecutor.js:24584:5)
```

2) If I fail 3D secure authorization, my org is blocked (You need to enable webhook to see this, might be useful: #8285)

image

3) Is it possible that because of internet connection issues, my card is charged but `UpgradeToTeamTierMutation` didn't run?

nickoferrall:
@igorlesnenko I haven't been able to reproduce this - see Loom: https://www.loom.com/share/f3321634c87645c489ebe11f75f824a9

I tried throttling the network but no luck. Could you record a Loom and let me know any steps I should follow to reproduce this please?

packages/server/graphql/public/mutations/createStripeSubscription.ts
```diff
@@ -34,6 +34,12 @@ const createStripeSubscription: MutationResolvers['createStripeSubscription'] =
 
   const manager = getStripeManager()
   const {stripeId} = organization
+  if (stripeId) {
+    const existingSubscriptions = await manager.listActiveSubscriptions(stripeId)
+    if (existingSubscriptions.data.length > 0) {
+      return standardError(new Error('Organization already has a subscription'), {userId: viewerId})
+    }
+  }
```
igorlesnenko, line 42: Not sure I fully understand what we trying to achieve here.

If we have `stripeId` saved we should have `subscriptionId` saved too.

Here we check if `stripeId` saved, subscription exist on stripe, but not saved. How is it possible?

igorlesnenko:
@nickoferrall I tried to reproduce it again and it works now. Not sure what it was. The only thing is that I need to refresh the page to see the UI changes after upgrade

packages/server/graphql/public/mutations/createStripeSubscription.ts
nickoferrall: The `stripeId` exists if the Stripe customer exists, but a subscription could become invalid afterwards. For example, if their card expires or we cancel their subscription from the Stripe dashboard. 

That's why I thought that we shouldn't return an error if the `stripeId` exists but only if an active subscription exists.

nickoferrall:
> I tried to reproduce it again and it works now. Not sure what it was. The only thing is that I need to refresh the page to see the UI changes after upgrade

OK great! I've got an issue to handle UI changes after upgrading: https://github.com/ParabolInc/parabol/issues/8006

packages/server/graphql/public/mutations/upgradeToTeamTier.ts
```diff
@@ -0,0 +1,93 @@
...
+  // AUTH
+  const viewerId = getUserId(authToken)
+  const organization = await dataLoader.get('organizations').load(orgId)
+  const {stripeId, tier, activeDomain, name: orgName} = organization
+
+  if (!stripeId) {
+    return standardError(new Error('Organization does not have a stripe id'), {
+      userId: viewerId
+    })
+  }
```
Dschoordsch, line 33: -1 We should check that the `stripeId` is attached to a valid subscription. Having a random `stripeId` and then calling `upgradeToTeamTier` shouldn't be enough to remove all restrictions and updating the tier.

packages/client/modules/userDashboard/components/OrgBilling/BillingForm.tsx
```diff
@@ -121,11 +124,12 @@ const BillingForm = (props: Props) => {
         setIsLoading(false)
         return
       }
+      UpgradeToTeamTierMutation(atmosphere, {orgId}, {onError, onCompleted})
```
Dschoordsch, line 127: -1 I'm not a big fan of having client logic responsible of tying `createSubscription` and `upgradeToTeamTier` mutation together. Can we call the latter internally based on a stripe event we get? This would be much more reliable, otherwise there could be a mismatch the other way round, we could have customers paying still on the free tier.

igorlesnenko (changes requested):
Sorry, I wasn't clear enough.

Issues in the comment [here](https://github.com/ParabolInc/parabol/pull/8289#pullrequestreview-1455694646) are 3 different issues.
I able to reproduce the issue 1 again

I recorded a loom about all of them https://www.loom.com/share/5a8725d7b9864de8aa347ede2dfa217e

packages/client/modules/userDashboard/components/OrgBilling/BillingForm.tsx
nickoferrall: Me neither, but for 3D Secure cards, the client needs to confirm the payment once the subscription has been confirmed. See https://stripe.com/docs/issuing/3d-secure?locale=en-GB and https://stackoverflow.com/questions/71631786/stripe-confirmcardpayment-frontend-vs-paymentintent-confirm-backend for more info.

nickoferrall:
Thanks for the Loom, Igor! I didn't have the webhooks set up, so I never received the `stripeFailPayment` error. The 1st & 2nd errors are caused by the same issue where `CreditCard` doesn't exist in `stripeFailPayment`. I've resolved that issue.

The third issue I've commented on [here](https://github.com/ParabolInc/parabol/pull/8289#discussion_r1222967673). I don't like having two client mutations for the reason you mentioned, but I don't see another way of processing 3D Secure cards. In the unlikely event someone successfully subscribes, loses their internet connection, and doesn't upgrade, we would just upgrade them ourselves. 

There are a couple of other parts of the app where we're using the `CreditCard` object, so I've created an issue to fix that: https://github.com/ParabolInc/parabol/issues/8350

codegen.json
```diff
@@ -25,6 +24,7 @@
           "SetIsFreeMeetingTemplateSuccess": "./types/SetIsFreeMeetingTemplateSuccess#SetIsFreeMeetingTemplateSuccessSource",
           "SignupsPayload": "./types/SignupsPayload#SignupsPayloadSource",
           "StripeFailPaymentPayload": "./mutations/stripeFailPayment#StripeFailPaymentPayloadSource",
+          "UpdateOrgFeatureFlagSuccess": "./types/UpdateOrgFeatureFlagSuccess#UpdateOrgFeatureFlagSuccessSource",
```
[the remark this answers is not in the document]
nickoferrall: I'm using Matt's `yarn newMutation` which automatically sorts codgen

packages/server/graphql/public/mutations/createStripeSubscription.ts
igorlesnenko: @nickoferrall this is correct, it is possible that `stripeId` exists but subscription is not. 

However you are checking another case. In your case `stripeId` exists and subscription exists, but data about existing subscription somehow is not stored in our database.

I wonder how is this case possible.
And if we need to handle it, we probably should record the subscription data back in our db instead of throwing an error.
Or instead of querying stripe we need to just check `organization.stripeSubscriptionId`

packages/server/graphql/public/mutations/upgradeToTeamTier.ts
```diff
@@ -0,0 +1,100 @@
...
+
+  const manager = getStripeManager()
+  const existingSubscriptions = await manager.listActiveSubscriptions(stripeId)
+  if (!existingSubscriptions.data.length) {
+    return standardError(new Error('Organization already has a subscription'), {userId: viewerId})
```
igorlesnenko, line 39: The error text looks incorrect, should say about missing subscription

igorlesnenko (changes requested):
Small change and a couple of questions

packages/server/graphql/private/mutations/stripeFailPayment.ts
```diff
@@ -29,32 +29,34 @@ const stripeFailPayment: MutationResolvers['stripeFailPayment'] = async (
...
 
   if (!org) {
     // org no longer exists, can fail silently (useful for all the staging server bugs)
     return {error: {message: 'Org does not exist'}}
   }
-  const {creditCard, stripeSubscriptionId} = org
+
+  const {stripeSubscriptionId, tier} = org
+
+  // 3D Secure auth failed, so the subscription was initiated but not complete
```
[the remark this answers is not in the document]
nickoferrall: The subscription object is already "Incomplete" in Stripe and says, "3D Secure attempt failed", so I don't think we need to terminate it

packages/server/graphql/public/mutations/createStripeSubscription.ts
nickoferrall: I was checking Stripe instead of our db as I think it's the more important source of truth, but I'm happy to check `stripeSubscriptionId` instead. It'll help speed up the mutation.

packages/server/billing/stripeWebhookHandler.ts
```diff
@@ -30,6 +30,20 @@ const eventLookup = {
         }
       `
     },
+    payment_succeeded: {
```
nickoferrall, line 33: `upgradeToTeamTier` is now called by Stripe's webhook rather having two mutations on the client. This should handle the use case where the client creates the stripe subscription but loses their internet connection before the upgrade happens

packages/client/mutations/UpgradeToTeamTierMutation.ts
```diff
@@ -37,10 +36,12 @@ graphql`
 `
 
 const mutation = graphql`
-  mutation UpgradeToTeamTierMutation($orgId: ID!, $stripeToken: ID, $paymentMethodId: ID) {
-    upgradeToTeamTier(orgId: $orgId, stripeToken: $stripeToken, paymentMethodId: $paymentMethodId) {
-      error {
-        message
+  mutation UpgradeToTeamTierMutation($invoiceId: ID!) {
```
nickoferrall, line 39: `upgradeToTeamTier` is now on the private schema, so we don't need a client mutation, but we do need the fragments for the subscription. Perhaps the fragments could be created elsewhere instead of here so we don't have this mutation 🤷

packages/server/utils/stripe/StripeManager.ts
```diff
@@ -54,19 +75,14 @@ export default class StripeManager {
...
-  }
-
   async createTeamSubscription(customerId: string, orgId: string, quantity: number) {
     return this.stripe.subscriptions.create({
       // USE THIS FOR TESTING A FAILING PAYMENT
       // https://stripe.com/docs/billing/testing
       // trial_end: toEpochSeconds(new Date(Date.now() + 1000 * 10)),
       customer: customerId,
       proration_behavior: 'none',
+      payment_behavior: 'default_incomplete',
```
nickoferrall, line 85: We need this for 3D Secure cards so that the payment is considered incomplete before the user authorises it

packages/server/utils/stripe/StripeManager.ts
```diff
@@ -121,7 +144,9 @@ export default class StripeManager {
   }
 
   async retrieveInvoice(invoiceId: string) {
-    return this.stripe.invoices.retrieve(invoiceId)
+    return this.stripe.invoices.retrieve(invoiceId, {
+      expand: ['payment_intent']
```
nickoferrall, line 148: This is required so that we get the `paymentIntent.status` in `stripeFailPayment`. The payment is considered a failure until the user authorises it.

packages/client/mutations/UpgradeToTeamTierMutation.ts
nickoferrall: We have a fragment folder, so I've moved this there 👍

nickoferrall:
I've fixed the `loader.load()` error, and the second error message says you can retry the API request if you see it intermittently (which is the case for me). If it happens consistently, it's worth digging into in a separate issue.

>  would it be possible to show some loading indicator or something

Happy to add this, but I think it should be in a different issue and might need design work

Dschoordsch (changes requested):
I cannot reproduce it right now, but a couple of times when I failed a payment and then succeeded shortly after, I got an error from stripe that the element was already accessed by another API call. Probably an edge case though.

packages/server/billing/stripeWebhookHandler.ts
nickoferrall: @Dschoordsch would it be OK if I create an issue to handle this, ship the checkout flow, and then work on it right away?

nickoferrall:
> I cannot reproduce it right now, but a couple of times when I failed a payment and then succeeded shortly after, I got an error from stripe that the element was already accessed by another API call. Probably an edge case though.

I get that occasionally too, but _I think_ it's happening when we're both testing the Stripe payments at the same time. At least when I saw the error, I could see you were creating payments in the Stripe Dash too. I'll keep an eye out for this.

packages/server/graphql/public/typeDefs/_legacy.graphql
```diff
@@ -9212,7 +9212,29 @@ type UpdateTemplateScopeSuccess {
...
+  A link to the user’s profile image.
+  """
+  picture: URL
+
+  """
+  The name, as confirmed by the user
+  """
+  preferredName: String
+}
+
```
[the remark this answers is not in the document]
nickoferrall: It was, thanks! Removed 👍

Dschoordsch (approved):

PR #8292, opened 2023-05-30
feat: add FixedActivity to GQL Schema

mattkrick:
# Description

partially fixes #8293 (the backend piece)

- This PR creates a new GQL type called `FixedActivity`, which is an activity type that does not have templates (check-in, standups).
- This `FixedActivity` allows us to stop special casing check-ins & standups. `Activity = FixedActivity | MeetingTemplate`
- It also adds `viewerLowestScope` to the GQL schema. This allows the client to request the lowest scope of a template's permissions. In the past, we had this defined a couple different ways & it required fetching the organization IDs. These frontend changes will be implemented in #8293
- the migration fixes the illustrationUrl used in development & self-hosted instances
- GQL `User.availableTemplates` has been refactored to be waaay faster & sort better

## Demo

No front-end changes in this

## Testing scenarios

- [ ] fetch a list of templates using activity library & legacy UI, they all work & sort order seems reasonable


jmtaber129 (commented):
@mattkrick it looks like this PR is doing a few distinct things. Would you mind splitting this up so it's easier to understand which parts are relevant to other parts? E.g. move `viewerLowestScope` into the PR with that frontend refactor, move the seasonal template hiding changes to a separate PR, move the added `illustrationUrl` GQL fields to a PR where that field is used, etc.. Grab-bag PRs like these are inherently difficult to review and slow down the CR process, increasing our team's WIP in the process.

If you're worried about merge conflicts, stacking the resulting PRs should resolve that tension.

PR #8293, opened 2023-05-30
chore: remove useActivityDetails

### LINKED ISSUE #8262: refactor ActivityDetails to decrease overfetch
`ActivityDetails` has a graphql fragment that has a few extra fragments & fields than it uses. This can be problematic in reading & refactoring the code because it's difficult to know which fields can be removed.

AC
- [ ] in the gql fragment, every field requested in ActivityDetails is used by Activity details
- [ ] the fragment spreads included are for components that are direct children of the ActivityDetails
- [ ] sort order of templates deprioritizes older templates

Estimate: 2 hours

mattkrick:
# Description

Fixes #8262 

- #8408 added backend support for treating FixedActivity types (e.g. standup, check-in) like meeting templates. With that in place, we can remove the mocks on the client.
- on the details view we now only fetch the single activity in question, which is pretty great if folks are coming from the landing page. 
- we now make use of `illustrationUrl`, which means adding illustrations to the DB does not require a change to the front-end code

## Demo

https://www.loom.com/share/a234f3a4da5e4eaf890d8f0019711e0a?sid=c7a47093-e56b-40e8-8187-a367c7952918

No UI changes! Just code refactor!

## Testing

- [ ] run `yarn build && node dist/pushToCDN`. This will push all the illustrations to your local /self-hosted directory or to S3, depending on your env


packages/client/components/ActivityLibrary/ActivityDetails/ActivityDetails.tsx
```diff
@@ -22,33 +20,26 @@ graphql`
     category
     orgId
     teamId
+    illustrationUrl
     isFree
     scope
-    team {
-      editingScaleId
-      ...PokerTemplateScaleDetails_team
```
mattkrick, line 29: I think at some point this got refactored to TemplateDetails? Either way, removed it since it wasn't used in this specific component

packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
```diff
@@ -32,7 +32,7 @@ import {RecurrenceSettings} from '../TeamPrompt/Recurrence/RecurrenceSettings'
 import {ActivityDetailsRecurrenceSettings} from './ActivityDetailsRecurrenceSettings'
 
 interface Props {
-  selectedTemplateRef: ActivityDetailsSidebar_template$key | null
+  selectedTemplateRef: ActivityDetailsSidebar_template$key
```
mattkrick, line 35: follow-up from your hotfix, the template is now non-nullable, which makes the logic below a little easier

packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
```diff
@@ -99,7 +99,8 @@ const ActivityDetailsSidebar = (props: Props) => {
       : []
 
   const [selectedTeam, setSelectedTeam] = useState(
-    availableTeams.find((team) => team.id === preferredTeamId) ??
+    () =>
```
mattkrick, line 102: by wrapping in a thunk it only gets called once instead of once per render.

packages/server/fileStorage/FileStoreManager.ts
```diff
@@ -6,19 +6,23 @@ export default abstract class FileStoreManager {
...
+    const dotfreeExt = ext.replace(/^\./, '')
+    const partialPath = `Organization/${orgId}/picture/${filename}.${dotfreeExt}`
     return this.putFile(file, partialPath)
   }
 
   async putTemplateIllustration(file: Buffer, orgId: string, ext: string, name?: string) {
     const filename = name ?? generateUID()
-    const partialPath = `Organization/${orgId}/template/${filename}.${ext}`
+    const dotfreeExt = ext.replace(/^\./, '')
+    const partialPath = `Organization/${orgId}/template/${filename}.${dotfreeExt}`
```
mattkrick, line 25: i realized some files were getting double dots 😬

scripts/toolboxSrc/pushToCDN.ts
```diff
@@ -11,7 +11,7 @@ const pushToCDN = async () => {
   const context = (require as any).context(
     '../../static/images/illustrations',
     false,
-    /Template.png$/
+    /\/action.png$|\/teamPrompt.png$|Template.png$/
```
mattkrick, line 14: we don't suffix these fixed activities with `Template`, which is fine, but it wasn't grabbing them to pushToCDN

PR #8297, opened 2023-05-31
chore(prompt-to-join): disable buttons when running accept request mutation

igorlesnenko:
image

**How to test:**

- Trigger prompt to join org
- See that add teammate dialog buttons become disabled when mutation is running

### DIFF
--- a/packages/client/mutations/useAcceptRequestToJoinDomainMutation.ts
+++ b/packages/client/mutations/useAcceptRequestToJoinDomainMutation.ts
@@ -1,11 +1,11 @@
 import graphql from 'babel-plugin-relay/macro'
 import {useMutation, UseMutationConfig} from 'react-relay'
 import useAtmosphere from '../hooks/useAtmosphere'
-import {AcceptRequestToJoinDomainMutation as TAcceptRequestToJoinDomainMutation} from '../__generated__/AcceptRequestToJoinDomainMutation.graphql'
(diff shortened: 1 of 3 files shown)


nickoferrall (approved):
This looks good, but as it's using/introducing a new Relay hook pattern, I think it requires a Maintainer Review. I can see that there was a conversation about using Relay hooks [here](https://parabol.slack.com/archives/C608QL4RH/p1684455091188979), but I'm not sure if there's an agreed upon pattern

igorlesnenko:
I see we already using `useMutation` hook in `useTemplateCategoryMutation`, if we introduce another pattern, we would need to refactor it anyway someday, that is why I feel safe to use it here too

packages/client/mutations/useAcceptRequestToJoinDomainMutation.ts
```diff
@@ -0,0 +1,71 @@
...
+            key: 'acceptRequestToJoinDomainError'
+          })
+          handlers?.onError && handlers.onError()
+        }
+      },
+      // allow components to override default handlers
+      ...config
+    })
+  }
+  return [execute, submitting] as const
```
Dschoordsch, line 68: +1 I'm not certain having these in an array rather than an object is the best way forward, but happy to see how it evolves first.

Dschoordsch (approved):
Looks good, did not test

packages/client/mutations/useAcceptRequestToJoinDomainMutation.ts
igorlesnenko: Merging this for now, some discussion about possible improvement here #8299

PR #8299, opened 2023-05-31
chore: add useSimpleMutation

igorlesnenko:


packages/client/mutations/useGenericMutation.ts
```diff
@@ -0,0 +1,48 @@
...
+  mutation: any,
+  dataField: string,
+  successMessage: string,
+  errorMessage: string
+  // TODO some segment config?
+) => {
+  const [commit, submitting] = useMutation(mutation)
+  const atmosphere = useAtmosphere()
+
+  const execute = (config: UseMutationConfig<any>, handlers?: Handlers) => {
```
mattkrick, line 20: `config` already has `onError, onSuccess, variables`. Could we remove `handlers`?

packages/client/mutations/AcceptRequestToJoinDomainMutation.ts
```diff
@@ -25,47 +22,14 @@ const mutation = graphql`
...
-      },
-      // allow components to override default handlers
-      ...config
-    })
-  }
-  return [execute, submitting] as const
+const useAcceptRequestToJoinDomainMutation = () => {
+  return useGenericMutation(
+    mutation,
+    'acceptRequestToJoinDomain',
```
mattkrick, line 29: we can remove this as a param & use `getRequest(mutation).operation.selections[0].name`

packages/client/mutations/useGenericMutation.ts
```diff
@@ -0,0 +1,48 @@
...
+type Handlers = {
+  onSuccess?: (res: any, variables: any) => void
+  onError?: (error: any) => void
+}
+
+export const useGenericMutation = (
+  mutation: any,
+  dataField: string,
+  successMessage: string,
+  errorMessage: string
```
BartoszJarocki, line 14: this assumes every mutation needs that, which isn't the case, I'd remove it

packages/client/mutations/useGenericMutation.ts
```diff
@@ -0,0 +1,48 @@
...
+import useAtmosphere from '../hooks/useAtmosphere'
+//import SendClientSegmentEventMutation from './SendClientSegmentEventMutation'
+
+type Handlers = {
+  onSuccess?: (res: any, variables: any) => void
+  onError?: (error: any) => void
+}
+
+export const useGenericMutation = (
+  mutation: any,
```
BartoszJarocki, line 11: we should aim to use generics here instead of any

packages/client/mutations/useGenericMutation.ts
BartoszJarocki: +1

packages/client/mutations/useGenericMutation.ts
```diff
@@ -0,0 +1,48 @@
...
+          handlers?.onSuccess && handlers.onSuccess(res, variables)
+        } else {
+          atmosphere.eventEmitter.emit('addSnackbar', {
+            message: errorMessage || error.message,
+            autoDismiss: 5,
+            key: `${dataField}Error`
+          })
+          handlers?.onError && handlers.onError(error)
+        }
+      },
```
BartoszJarocki, line 42: imo this assumes too much about the potential use cases. not every mutation wants to display toasts. if mutation wants to do that, it should provide a custom handler

igorlesnenko:
@mattkrick @BartoszJarocki If I implement all these changes, then `useGenericMutation` becomes no different from just `useMutation`.

Initially, I wanted to simplify something that often come ups when using mutations. 

I updated the approach and now only handle errors (no snackbars etc), because we have few different ways to handle errors in the code:

1. getGraphQLError: https://github.com/ParabolInc/parabol/blob/ff99d594ce6d9058be2ec29477e3a1e0594fee1e/packages/client/mutations/AddTeamMutation.ts#L89

2. getOnCompletedError
https://github.com/ParabolInc/parabol/blob/4174249ee6c7a2de9257110b50ec602c80e35c0c/packages/client/components/TeamPrompt/Recurrence/UpdateRecurrenceSettingsModal.tsx#L123

3. Manual handling
https://github.com/ParabolInc/parabol/blob/f3361ab138d76ecbd2fa233e518790226414709e/packages/client/mutations/EmailPasswordResetMutation.ts#L38

4. Often, there is no error handling at all. Probably, because having both `onCompleted` and `onError` might be confusing. Even though `onError` exists, we should also should handle an error in `onCompleted`.
https://github.com/ParabolInc/parabol/blob/5414f4fc88447283eb5b030e1cdcbc4e64be00e3/packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx#L131

Let me know what you think

mwermuth:
@igorlesnenko is this PR still ongoing? There doesn't seem to be any action since august.

igorlesnenko:
@mwermuth I'm planning to do something

igorlesnenko:
Closing for now

PR #8300, opened 2023-05-31
fix: fix loader.load with undefined in accept team invitation

igorlesnenko:
Probably a solution for #7771 
(At least some of errors)

image

**How to test:**

- Invite a user using email
- Click accept invitation, see it works
- Click accept invitation again, see no error in logs


nickoferrall (approved):
Tested and it works for me. Nice fix!

PR #8302, opened 2023-05-31
fix(prompt-to-join): fix add teammate routing when opened on a new tab

igorlesnenko:
image

When copy pasting add teammate modal url in a new tab, it shows 404 on the background. 

Works on just page refresh, that is why I didn't notice it.

How to test:

- trigger prompt to join org
- open add teammate dialog
- copy the URL
- paste in a new tab
- see everything is fine


Dschoordsch (approved):
Works

PR #8310, opened 2023-06-01
feat(DX): Infer parent portal IDs through React context

jmtaber129:
# Description
POC for inferring parent portal IDs through a React context.

Currently, we manually configure portals with a `parentId` if they're nested, and then we use that parent ID to attach the new portal to the parent portal. This is tricky, since many components rendering portals are decoupled from the rendered parent portal, which leads to bugs like https://github.com/ParabolInc/parabol/issues/8272 when a component is reused without updating child portals' `parentId`s (which can often be deep inside the component tree).

Instead, track the parent portal ID through a React context managed within the `usePortal` hook. This should eliminate almost all needs for passing `parentId` to `usePortal`-like hooks (e.g. `useMenu`, etc.), and either eliminate this class of bug or (at minimum) make fixes for these bugs much simpler.

## Demo
Nested portal without `parentId` passed:
https://www.loom.com/share/b9669e2ca9e6484f8145014131d14d2c

Nested portal with incorrect `parentId` passed:
https://www.loom.com/share/e5766720bff04c398bd0a743607556ed

New POC - no `parentId` needed:
https://www.loom.com/share/37ec71d5c5114c6ea587ef3417ec039f

## Testing scenarios
Smoke test modals throughout the app

## Final checklist

### DIFF
--- a/packages/client/hooks/usePortal.tsx
+++ b/packages/client/hooks/usePortal.tsx
@@ -1,4 +1,4 @@
-import React, {ReactNode, useCallback, useEffect, useRef} from 'react'
+import React, {ReactNode, createContext, useCallback, useContext, useEffect, useRef} from 'react'
 import {createPortal} from 'react-dom'
 import requestDoubleAnimationFrame from '../components/RetroReflectPhase/requestDoubleAnimationFrame'
 import hideBodyScroll from '../utils/hideBodyScroll'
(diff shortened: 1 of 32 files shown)


packages/client/hooks/usePortal.tsx
```diff
@@ -184,7 +188,10 @@ const usePortal = (options: UsePortalOptions = {}) => {
       const targetEl = portalRef.current
       return !targetEl || portalStatusRef.current === PortalStatus.Exited
         ? null
-        : createPortal(reactEl, targetEl)
+        : createPortal(
+            <PortalContext.Provider value={options.id}>{reactEl} </PortalContext.Provider>,
```
mattkrick, line 192: +2 nice pattern!

mattkrick (approved):
LGTM! haven't tested them, but the code looks great!

Issue #8316, opened 2023-06-02
Checkout Flow: update credit card UI


nickoferrall:
In https://github.com/ParabolInc/parabol/pull/8162, I needed to use Stripe's `CardElement` instead of their `PaymentElement` to handle 3D Secure cards.

This `CardElement` now needs to be styled properly.

Figma designs [here](https://www.figma.com/file/SyJdkzumWsUJo58cmBGgus/Checkout-concepts?t=p3f7trldNdOyScXZ-0)

<img width="623" alt="credit-card-ui" src="https://user-images.githubusercontent.com/39854876/215596345-59e7d328-8497-4f6a-9c7b-4729cf05607b.png">

Note: we don't need to collect the card holder's name, so this doesn't need to match the design.

### AC

- [ ] Credit card input looks similar to the design

PR #8318, opened 2023-06-02
feat(checkout-flow): credit card UI

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8316 & fix https://github.com/ParabolInc/parabol/issues/7914

Loom demo: https://www.loom.com/share/120bf2df861d4cf78eef86a015bfbe93

### To test

- [ ] Add the checkout flow feature flag
- [ ] See that the credit card UI looks similar to the designs
- [ ] Refresh the page and see that there's no jumpiness. The credit card panel shows up at the same time as the others
- [ ] Error messages shows up when entering invalid card detais
- [ ] Can submit valid card details and upgrade the org

### DIFF
--- a/packages/client/modules/userDashboard/components/OrgBilling/PaymentDetails.tsx
+++ b/packages/client/modules/userDashboard/components/OrgBilling/PaymentDetails.tsx
@@ -110,7 +110,6 @@ const PaymentDetails = (props: Props) => {
     graphql`
       fragment PaymentDetails_organization on Organization {
         id
-        tier
         orgUserCount {
           activeUserCount
         }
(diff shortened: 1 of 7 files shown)


packages/client/modules/userDashboard/components/OrgBilling/PaymentDetails.tsx
```diff
@@ -127,29 +124,16 @@ const PaymentDetails = (props: Props) => {
...
-  if (!clientSecret.length) return null
   return (
     <StyledPanel label='Credit Card'>
       <StyledRow>
         <Plan>
           <Title>{'Credit Card Details'}</Title>
           <Content>
-            <Elements
-              options={{
-                clientSecret
```
nickoferrall, line 149: `clientSecret` was is no longer needed with `CardElement` and `confirmCardPayment`. This means we no longer need to return null before we get the `clientSecret` which fixes the jumpiness

igorlesnenko (approved):
I unable to see the error message until I click on upgrade. Maybe it make sense to show it as soon as field became red

image

acressall (approved):

PR #8325, opened 2023-06-06
fix: dumpHeap to /tmp directory

Dschoordsch:
The node process drops all user rights and thus needs to write to a directory writable by all.

## Testing scenarios

run `mutation Dump { dumpHeap(isDangerous: true) }` locally and check the file exists

## Final checklist


mattkrick:
LGTM! I think we only allow 60s to respond, so hopefully it completes in that time otherwise we're going on a hunt 😝

Dschoordsch:
I removed the `mkdtemp` part as it would only append some random characters. We already have the timestamp, that should be unique enough. `os.tempdir` usually is `/tmp`, so hunting the file down shouldn't be that hard.

PR #8337, opened 2023-06-06
feat(rid): Add new users to feature flag

jmtaber129:
# Description
See [spec](https://www.notion.so/parabol/miniRFC-Adding-new-users-to-RiD-feature-flag-f1d3fa0524664c47aa9cfe12f5426b6a?pvs=4#a7f9fc955dad4c8c8dac816e055e54d7) 🔒 

For new P0 users (users for which no existing users share the domain), add 5% of them to the RiD feature flag.

## Testing scenarios
- [ ] Domain affinity for non-P0s
  - [ ] If some other user in the domain has the flag, always add the flag to the new user (suggest changing the `PERCENT_ADDED_TO_RID` to 0 to test this)
  - [ ] If there are other users in the domain, and no other users have the flag, never add the flag to the new user (suggest changing the `PERCENT_ADDED_TO_RID` to 1 to test this)
- [ ] P0s (no other users in the domain)
  - [ ] For low `PERCENT_ADDED_TO_RID`, most new signups won't be added
  - [ ] For high `PERCENT_ADDED_TO_RID`, most new signups will be added
  - [ ] For `PERCENT_ADDED_TO_RID` of 0.5, ~half of new signups will be added

## Final checklist

### DIFF
--- a/packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
+++ b/packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
@@ -18,6 +18,8 @@ import getUsersbyDomain from '../../../postgres/queries/getUsersByDomain'
 import sendPromptToJoinOrg from '../../../utils/sendPromptToJoinOrg'
 import {makeDefaultTeamName} from 'parabol-client/utils/makeDefaultTeamName'
 
+const PERCENT_ADDED_TO_RID = 0.05
+
 const bootstrapNewUser = async (newUser: User, isOrganic: boolean, searchParams?: string) => {
   const r = await getRethink()
   const {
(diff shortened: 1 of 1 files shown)


mattkrick:
hoooooly moly i had no idea we were querying all users for a given domain. 
This is pretty bad for e.g. `gmail.com`, which could return 100,000 records.

Your code is great. Would you mind adding some logic to first check `isCompanyDomain` before running that big ol query?


Screenshot 2023-06-06 at 12 47 14 PM

jmtaber129:
@mattkrick Oooof totally missed that's what was happening! Added a check for `isCompanyDomain` before fetching domain users.

PR #8339, opened 2023-06-07
fix(standups): allow selecting text within the standup response drawer

### LINKED ISSUE #7866: Standups: Response text in a discussion thread cannot be selected
## Issue - Bug

When opening a discussion thread in a Standup meeting, the response text cannot be selected. Please note that comments can be selected.

Captura de pantalla 2023-03-06 a la(s) 11 25 37 a m

## Acceptance Criteria:
Users can:
   - Select (and copy) the Standup response text in a discussion thread
   
BartoszJarocki:
# Description

Fixes #7866 

## Demo

No demo

## Testing scenarios

- [ ] Start a standup meeting, open discussion thread for a response and check if the response can be selected

## Final checklist


packages/client/components/TeamPrompt/TeamPromptDiscussionDrawer.tsx
```diff
@@ -34,7 +34,6 @@ const Drawer = styled('div')<{isDesktop: boolean; isOpen: boolean}>(({isDesktop,
   top: 0,
   right: isDesktop ? 0 : undefined,
   transition: `all 200ms ${BezierCurve.DECELERATE}`,
-  userSelect: 'none',
```
BartoszJarocki, line 37: @jmtaber129 do you remember why it was added here in the first place? I checked and I haven't found any regression so it should be good. another fix would be to add `userSelect: 'auto'` to the response editor itself, but this feels more appropriate, unless I'm missing why it's here

jmtaber129: I think this was largely copied from another drawer in the app, I think for the case when the drawer is draggable (smaller screens where the drawer appears on top of the main view IIRC). So without this, a user trying to drag the drawer closed might end up selecting text.

We can probably just condition this style on `isDesktop` being false

BartoszJarocki: done, @jmtaber129 could you take a look now?

PR #8340, opened 2023-06-07
chore: upgrade react to 18.2.0

BartoszJarocki:
# Description

In order to use Framer Motion in the future, we need to update React to 18.2.0. Framer motion uses `useId` hook and other stuff, which is not available in React 17.

Official upgrade guide: [How to Upgrade to React 18 – React](https://react.dev/blog/2022/03/08/react-18-upgrade-guide)

TODO
- [ ] `useTransition` hook isn't working properly ie. meeting cards are not showing on initial render on dashboard, avatar list isn't showing etc; here the solution could be simply removing that and using #8371
- [ ] due to ☝️ standup response cards are not showing as well
- [ ] adding more than one agenda item seems to broken; I suspect it could be something with react-dnd, worth checking if updating it to the newest version solves the issue
- [ ] as now we're using `<StrictMode/>` all parts of the app relying on the `useInitialRender` have to be checked 
- [ ] check if ```;(RelayFeatureFlags as any).ENABLE_RELAY_CONTAINERS_SUSPENSE = false
;(RelayFeatureFlags as any).ENABLE_PRECISE_TYPE_REFINEMENT = true``` are still needed (why we had them in the first place?) 

## Demo

No demo

## Testing scenarios

- [ ] Smoke test the frontend app
- [ ] Smoke test the emails rendering

## Final checklist


packages/client/types/modules.d.ts
```diff
@@ -19,6 +19,7 @@ declare module 'emoji-mart/dist-modern/utils/data.js'
 declare module 'emoji-mart/dist-modern/components/picker/nimble-picker'
 declare module 'react-textarea-autosize'
 declare module 'react-copy-to-clipboard'
+declare module 'react-dom/client'
```
BartoszJarocki, line 22: not sure why I'm getting even though the types were updated to the latest version
```
Could not find a declaration file for module 'react-dom/client'. '/Users/bjarocki/Work/parabol/web-parabol/node_modules/react-dom/client.js' implicitly has an 'any' type.
  If the 'react-dom' package actually exposes this module, consider sending a pull request to amend 'https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-dom'ts(7016)
```

BartoszJarocki: oh, now I know. version was locked in our root package.json resolutions!

mattkrick:
I've got a sneaking suspicion that I've probably done some hacky things that break if components are suspended mid-render. Could you add `<StrictMode>` to the app & we can smoke test? That will cause components to mount twice. Unfortunately we'll have to poke around just about every feature of the app & see what console logs arise

BartoszJarocki:
will do!

BartoszJarocki:
@mattkrick i've added a strict mode and indeed, seems like you need to refresh manually the page to see the data rendered correctly. there are no error/warnings thought 😵‍💫 I'll see what can I do, but I'd appreciate any hints where and what parts could be hacky, thanks!

BartoszJarocki:
I tracked down not displaying meeting cards problem to `useTransition` hook in strict mode, not a relay issue luckily. I'll post more if I learn something

BartoszJarocki:
```
;(RelayFeatureFlags as any).ENABLE_RELAY_CONTAINERS_SUSPENSE = false
;(RelayFeatureFlags as any).ENABLE_PRECISE_TYPE_REFINEMENT = true
```
I haven't noticed any change in behavior after removing this from `Atmosphere`, @mattkrick any ideas any particular issue that this was solving so I can retest?

mattkrick:
The glitches will occur anywhere that we do special handling on the first render cycle because that will get called twice before paint. We typically do that to measure the DOM in a before state & then animate it to a post state. The answer you don't want to hear is we'll have to audit every `useEffect` and `useLayoutEffect` call to see if it's doing something exclusively on the first render. 

I think auditing the usage of this hook would probably catch 80% of the use cases!
https://github.com/ParabolInc/parabol/blob/8c68bf771a350b31c7948e6bd11c5c110e30c977/packages/client/hooks/useInitialRender.ts

adaniels-parabol:
Before merging this in can we verify no impact with --no-deps at build time?

BartoszJarocki:
> Before merging this in can we verify no impact with --no-deps at build time?

@adaniels-parabol sure thing, sadly there's still a lot to be done here before we can merge

mwermuth:
Going to close this as we will not be working on this anytime "soon". See [Slack](https://parabol.slack.com/archives/C608QL4RH/p1705071913617129).

Issue #8353, opened 2023-06-09
Checkout Flow: implement update credit card logic


nickoferrall:
### AC

- [ ] Users can update their default credit card

Issue #8350, opened 2023-06-08
Checkout Flow: fetch card details from Stripe


nickoferrall:
We currently have the `CreditCard` object stored in `Organization` and `Invoice`. With the new checkout flow, we're using Stripe Elements and won't store any card details. Instead, we need to retrieve them from Stripe.

### AC

- [ ] Fetch credit card details from Stripe 
- [ ] Don't remove `CreditCard`. We'll need this for old users

PR #8354, opened 2023-06-09
feat(checkout-flow): update checkout flow after upgrade

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8006 & https://github.com/ParabolInc/parabol/issues/8350

Loom Demo: https://www.loom.com/share/40873069b792431d97edc4ad3351e221

This PR doesn't handle implementing updating credit card details. That will be handled in: https://github.com/ParabolInc/parabol/issues/8353

After upgrading, the user's invoice and credit card sections don't show up until the page is refreshed. I think it's OK to ship it like this and then update it as the user already gets feedback that they've upgraded.

### To test

- [ ] Add the checkout flow user feature and upgrade to Team
- [ ] See that the right drawer slides in. You still need to refresh the page to see the other changes. If https://github.com/ParabolInc/parabol/pull/8289 doesn't fix that then I'll create another small PR to do so
- [ ] Refresh the page and see that the UI looks like https://github.com/ParabolInc/parabol/issues/8006's description
- [ ] See that users that don't have the feature flag enabled can upgrade using the old UI


nickoferrall:
@acressall  I wonder whether we need a close icon for the "Update Credit Card" bit? I touch on it in the Loom video

acressall:
Thanks for the loom @nickoferrall !

> I wonder whether we need a close icon for the "Update Credit Card" bit?

Good shout! Let's add a cancel button on the far right, in the outline btn style

acressall:
Here's a quick mock
CC (1)

acressall (approved):

nickoferrall:
Thanks, @acressall! I've pushed an update with a "Cancel" button 👍 

Screenshot 2023-06-12 at 16 32 06

igorlesnenko (approved):
Works!

packages/client/modules/userDashboard/components/OrgBilling/UpdatePayment.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+type Props = {
+  setIsUpdating: (isUpdating: boolean) => void
+}
+
+const UpdatePayment = (props: Props) => {
+  const {setIsUpdating} = props
+
+  return (
+    <StyledForm id='payment-form'>
```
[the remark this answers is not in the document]
nickoferrall: Changed it to Tailwind 👍

packages/server/graphql/types/Invoice.ts
```diff
@@ -60,7 +61,29 @@ const Invoice = new GraphQLObjectType<any, GQLContext>({
...
     creditCard: {
       type: CreditCard,
-      description: 'the card used to pay the invoice'
+      description: 'the card used to pay the invoice',
+      resolve: async ({creditCard, orgId}, _args, {dataLoader}) => {
+        // we used to store credit card details in the db, so we need to check there first
+        if (creditCard) {
+          return creditCard
+        }
+        const organization = await dataLoader.get('organizations').load(orgId)
```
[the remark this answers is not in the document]
nickoferrall: I believe our engineering practice is to wait until it's repeated three times before extracting it, so I didn't move it to a separate place

packages/server/graphql/mutations/helpers/getCCFromCustomer.ts
```diff
@@ -1,24 +1,50 @@
...
+  if (customer.deleted) {
+    throw new Error(`Customer ${customer.id} has been deleted`)
   }
-  const manager = getStripeManager()
-  const card = await manager.retrieveSource(customer.id, customer.default_source)
-
-  if (!card) {
-    throw new Error(`Could not find a card for ${customer.id}`)
+  if (
+    typeof customer.default_source !== 'string' &&
```
Dschoordsch, line 11: -1 Checking the types, it could be of type `Card` as well, which should suit us. Same below in other places where we only accept `string`

packages/client/modules/userDashboard/components/OrgBilling/OrgBillingCreditCardInfo.tsx
```diff
@@ -83,13 +83,29 @@ const OrgBillingCreditCardInfo = (props: Props) => {
...
   if (!creditCard) return null
-  const {activeUserCount} = orgUserCount
   const {brand, last4, expiry} = creditCard
+
+  if (isUpdating) {
+    return (
+      <StyledPanel label='Credit Card'>
+        <StyledRow>
+          <Elements stripe={stripePromise}>
+            <UpdatePayment setIsUpdating={setIsUpdating} />
```
Dschoordsch, line 96: +1 would be nice to indicate this is just temporary

Dschoordsch (changes requested):
Tested, works

packages/client/modules/userDashboard/components/OrgBilling/OrgBillingInvoices.tsx
```diff
@@ -59,7 +64,7 @@ const OrgBillingInvoices = (props: Props) => {
   }
   if (!invoices || !invoices.edges.length) return null
   return (
-    <Panel label='Invoices'>
+    <StyledPanel label='Invoices' isWide={!!hasCheckoutFlowFlag}>
```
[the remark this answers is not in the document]
nickoferrall: Makes sense, updated!

packages/client/modules/userDashboard/components/OrgBilling/OrgBillingCreditCardInfo.tsx
nickoferrall: 🤔 why would it be temporary?

packages/client/modules/userDashboard/components/OrgBilling/UpdatePayment.tsx
```diff
@@ -0,0 +1,97 @@
...
+        </div>
+      </div>
+      <div className='flex w-full flex-nowrap items-center justify-between'>
+        <div className='w-1/4'>
+          <UpgradeButton isDisabled={false} size='medium' type={'submit'}>
+            {'Update'}
+          </UpgradeButton>
+        </div>
+        <div className='mt-4 flex w-1/4 justify-end'>
+          <SecondaryButton size='medium' type={'submit'} onClick={() => setIsUpdating(false)}>
```
Dschoordsch, line 88: -1 Missed this the last time round, type should be cancel `type='cancel'`, also the superfluous `{` could be removed

packages/client/modules/userDashboard/components/OrgBilling/UpdatePayment.tsx
```diff
@@ -0,0 +1,97 @@
...
+      fontSize: '16px',
+      '::placeholder': {
+        color: PALETTE.SLATE_600
+      }
+    }
+  }
+}
+
+type Props = {
+  setIsUpdating: (isUpdating: boolean) => void
```
Dschoordsch, line 36: -1 If it's not temporary, then we shouldn't pass a state setter of a parent in here directly. Instead we should have an `onCancel` handler (and an `onSubmit` if that shouldn't be a noop)

Dschoordsch (changes requested):
Did not test again

packages/client/modules/userDashboard/components/OrgBilling/OrgBillingCreditCardInfo.tsx
Dschoordsch: It's not updating anything yet, is it? I don't see any action wired up for 'submit'

nickoferrall: When the user clicks "Cancel", it updates the `isUpdating` state to false so that the update payment UI closes. I've created a `handleCancel` function and passed that here so the name changes. In the following PR, I add logic that handle the "submit" case.

Dschoordsch (approved):
Did not test again

PR #8362, opened 2023-06-13
fix: Bump undici to latest

mattkrick:
# Description

Hopefully fixes the memory leak found in `global.fetch`. Underlying fix is here: https://github.com/nodejs/undici/pull/1824


packages/server/monkeyPatchFetch.ts
```diff
@@ -0,0 +1,8 @@
+// CAN REMOVE WHEN WE UPGRADE TO NODE V20
+// There's a memory leak in fetch that shipped with Node v18.
+// The leak is fixed in v20, but we have clients who cannot run v20 yet
+// For now, we monkeypatch the native fetch with the fixed one
+
+import {fetch} from 'undici'
+
+global.fetch = fetch as any
```
mattkrick, line 8: @adaniels-parabol here she be

Issue #7265, opened 2022-10-11
Discussion topics are missing when going from Group to Discuss


igorlesnenko:
Steps to reproduce:

* Create a retro meeting
* advancing to "Vote"
* go back to "Group"
* go to "Discuss"

I could advance, but no discussions were available
![Peek 2022-10-11 12-54](https://user-images.githubusercontent.com/7331043/195072300-863f3900-702c-4698-92c9-66a1674a43a4.gif)

### Acceptance Criteria
- Users can see discussion topics if they exists when advancing from any phase

**Estimated effort:**  5h

jmtaber129:
I did a bit of investigation into this and found that this is due to the `vote` phase never being marked as "completed" by our logic. The current navigation logic will mark the facilitator's current stage as completed iff the navigation reflects forward movement in the meeting, and then navigate to the new stage. So if you navigate backward, then skip ahead, the phase the facilitator is implicitly completing is not marked complete.

Without the `vote` phase being marked complete, the side effect of creating discussion stages doesn't happen.

A similar bug is if you navigate from reflect -> group -> reflect -> vote -> group: The groups should be frozen because voting has started, but this is not the case because we didn't trigger the side effects of completing the `group` phase.

87prashant:
@jmtaber129 
So, it means every phase should be marked as completed if we reached it at least once.

87prashant:
Hi, @igorlesnenko @jmtaber129  I am working on this issue.
Could you define the difference between the phase and stage by considering a retrospective meeting? Does every phase have at least one stage?

Also, could you check out the PR for once if it could be the appropriate/potential solution?

jordanh:
@87prashant are you still working on this one per chance, or may I reassign it to a Parabol team member?

87prashant:
Yes, I am working. But I was waiting for a review of the work I did till now, if it could be a potential solution.

igorlesnenko:
@87prashant Sorry for the late reply, I completely missed it. Feel free to mention again if there is no response for a long time.

> Could you define the difference between the phase and stage by considering a retrospective meeting? Does every phase have at least one stage?

Phase is like discussion, stage is like discussion topic. Each phase has at least one stage

I looked at your PR. Not sure I fully understand how would you use the `maxPhaseIndex `to solve the issue. Can you explain a little bit more please?

[...]

PR #8374, opened 2023-06-14
fix(retros): When skipping phases, mark interim stages as complete

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/7265

Our meeting flow allows users to skip between phases, as long as they navigated to the previous phase before. For example, a user could navigate group -> vote -> group -> discuss. However, our current navigation logic only executes "completion" side effects for stages/phases we just navigated from. In the above example, we wouldn't execute side effects for completing the "vote" phase because we navigated from group to discuss.

To fix this, check if we're skipping phases when navigating to a new stage, and execute side effects for any skipped stages.

## Demo
https://www.loom.com/share/a467a9b8add045d28c1417e05c6e0070

## Testing scenarios
- [ ] Navigate like so in a retro: Reflect -> Group -> Reflect -> Vote -> Group (confirm completed) -> Discuss (confirm discussions are present) -> Vote (confirm completed)
- [ ] Smoke test navigation

## Final checklist

### DIFF
--- a/packages/client/hooks/useGotoStageId.ts
+++ b/packages/client/hooks/useGotoStageId.ts
@@ -88,9 +88,6 @@ const useGotoStageId = (meetingRef: useGotoStageId_meeting$key) => {
             }
           }
         }
-        if (!isComplete && isForwardProgress(phases, facilitatorStageId, stageId)) {
-          variables.completedStageId = facilitatorStageId
-        }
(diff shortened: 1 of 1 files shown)


igorlesnenko (approved):
Works! One small fix

Issue #8312, opened 2023-06-01
Activity library: when starting an activity we default to the first team on the list instead of last used or route context


enriquesanchez:
>  if you are on a team page and start a meeting there, the old meeting selector will set the team accordingly but the activity library seems to always choose the 1st team sorted alphabetically? ([from Slack](https://parabol.slack.com/archives/C042HTVBH8D/p1685636102060229) 🔒 )

IIRC we also used to default to the last team used, and now we always default to the 1st team on the list

### AC:
- Previous behavior for team selection when starting a new meeting is restored

jordanh:
Looked at this during scrub. It would be great to add to our next sprint.

PR #8376, opened 2023-06-14
feat(rid): After navigating to or selecting a team, use that team by default

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8312

In the activity library, we default to the first alphabetically-ordered team, with preference to premium teams. This is different from the existing flow, in which if a user navigates to a team dashboard, that team is the default.

Add a new global client-side relay field called `defaultTeamId` that gets updated any time the user navigates to a team dashboard or selects a team from a dropdown. This team ID is then the default for other team dropdowns.

## Demo
https://www.loom.com/share/c795617873d9437ca1e374dce17fdfee

## Testing scenarios
- [ ] Navigate to a team dashboard, then go to a template details page, and confirm that team is the default (as long as it's in the dropdown)
- [ ] Pick a team from a dropdown, then navigate to a different page, then to an AL page. Confirm that the team is the default

## Final checklist

### DIFF
--- a/packages/client/components/ActivityLibrary/ActivityDetails/ActivityDetails.tsx
+++ b/packages/client/components/ActivityLibrary/ActivityDetails/ActivityDetails.tsx
@@ -42,7 +42,7 @@ export const query = graphql`
   query ActivityDetailsQuery {
     viewer {
       tier
-      defaultTeamId
+      preferredTeamId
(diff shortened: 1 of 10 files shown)


BartoszJarocki (approved):
nicely done. I'd personally prefer `preferredTeamId` instead of `defaultTeamId` though ;)

jmtaber129:
I actually like `preferredTeamId` better, too! Renamed

PR #8383, opened 2023-06-15
chore: Prefill OrganizationApprovedDomain

Dschoordsch:
This is mainly useful for automatic setup for newly spawned instances.

## Testing scenarios

* Set environment variable and run up migration
  `ORGANIZATION_APPROVED_DOMAINS=parabol.co,example.com yarn pg:migrate up`
* Check that creating a new account with these domains will require email verification
* Run down migration  `yarn pg:migrate down` and check it's gone from PG

## Final checklist

### DIFF
--- a/packages/server/postgres/migrations/1686845459075_fillOrganizationApprovedDomain.ts
+++ b/packages/server/postgres/migrations/1686845459075_fillOrganizationApprovedDomain.ts
@@ -1,0 +1,28 @@
+import {Client} from 'pg'
+import getPgConfig from '../getPgConfig'
+
+const organizationApprovedDomains = process.env.ORGANIZATION_APPROVED_DOMAINS?.split(',')
+
+export async function up() {
+  if (!organizationApprovedDomains) return
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await Promise.all(
+    organizationApprovedDomains.map((domain) =>
+      client.query(
+        `INSERT INTO "OrganizationApprovedDomain" (domain, "orgId", "addedByUserId") VALUES ($1, 'aGhostOrg', 'aGhostUser')`,
+        [domain]
+      )
+    )
+  )
+  await client.end()
+}
+
+export async function down() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(
+    `DELETE FROM "OrganizationApprovedDomain" WHERE "orgId" = 'aGhostOrg' AND "addedByUserId" = 'aGhostUser'`
+  )
+  await client.end()
+}
--- a/.env.example
+++ b/.env.example
@@ -43,6 +43,8 @@ MAIL_SMTP_USE_TLS='1' # set to '0' for false
 MAIL_SMTP_CIPHERS='HIGH:MEDIUM:!aNULL:!eNULL:@STRENGTH:!DH:!kEDH'
 MIN_LOG_DURATION='200'
 OAUTH2_REDIRECT=''
+# domains added here will be added to the domains requiring email verification during the first run of the migrations
+ORGANIZATION_APPROVED_DOMAINS=example.com,example.net
 PGADMIN_DEFAULT_EMAIL='pgadmin4@pgadmin.org'
 PGADMIN_DEFAULT_PASSWORD='admin'
 PORT='3000'


packages/server/postgres/migrations/1686845459075_fillOrganizationApprovedDomain.ts
```diff
@@ -0,0 +1,28 @@
...
+const organizationApprovedDomains = process.env.ORGANIZATION_APPROVED_DOMAINS?.split(',')
+
+export async function up() {
+  if (!organizationApprovedDomains) return
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await Promise.all(
+    organizationApprovedDomains.map((domain) =>
+      client.query(
+        `INSERT INTO "OrganizationApprovedDomain" (domain, "orgId", "addedByUserId") VALUES ($1, 'aGhostOrg', 'aGhostUser')`,
```
mattkrick, line 13: Clarifying question: Why are we protecting people from joining `aGhostOrg`? Since no real users are in it, no one can send invites from it, so isn't it already protected?

Dschoordsch: Same table is used for email verification. The reason is on-demand environments. We need to have a way to create a 'su' user there so we can run custom graphql code and add feature flags etc. One easy way is to just have a job which adds it to all parabol.co domain members. To avoid the unlikely case that someone would find the environment and signs up with a fake email, we want email verification for this domain.

Dschoordsch: @mattkrick :arrow_up:

mattkrick: Ah! OK so this is just to turn on email verification for new signups with the given domain? Makes sense.

Questions:
- Where are we granting `su` to the domain? Is that done ad-hoc? in a psql script called from a helm chart?
- Is it safe to apply `su` to an entire domain? Seems like that's a permission that should not be applied on a domain-wide basis, but that's what it looks like will happen from the slack convo
- since this migration only runs once, changing the env var after the migration runs won't do anything, which might be confusing. Should we instead treat this like we prime our integrations & run it as an idempotent DB call that happens after all migrations run? https://github.com/ParabolInc/parabol/blob/7938dd78615230f2a6501594813a43722004e06c/scripts/toolboxSrc/primeIntegrations.ts

For your consideration:
- add an env var like `SUPERUSER_EMAILS=foo@bar.co,baz@bar.co`
- add a script like primeIntegrations that gets called from postDeploy.js
- in that script, it gives `su` to all the emails in SUERUSER_EMAILS

rafaelromcar-parabol: > Where are we granting su to the domain? Is that done ad-hoc? in a psql script called from a helm chart?

We are not doing that right now. The idea was to grant that for **all existing email addresses of a domain** from the Gitlab pipeline of our on-demand environments.

> Is it safe to apply su to an entire domain?

We don't really want to do that, we want to enable e-mail verification for domains so we can avoid people signing up with fake e-mails for those domains. We just want to do that for existing users in the on-demand environment, and only with a manually approved step. I don't think it would be safe to add su to everyone within a domain in any environment. Not even to us.

> since this migration only runs once, changing the env var after the migration runs won't do anything, which might be confusing. Should we instead treat this like we prime our integrations & run it as an idempotent DB call that happens after all migrations run?

I would prefer a script rather than a migration for what you just said. I would actually have even two different scripts:
- **force-org-approved-domains:** using the `ORGANIZATION_APPROVED_DOMAINS` with the list of domains that must have e-mail verification enabled.
- **super-users:** with `SUPERUSER_EMAILS` with a list of users that would get su permissions.

I think both are useful in different use cases, even if the first script `force-org-approved-domains` isn't really required.

TLDR: developing a script that adds su to some users, listed in `SUPERUSER_EMAILS` and remove su from anyone else if executed would be great and would solve our "problem" here.

mattkrick: sure, makes sense! IMO i'd wait until we have a client beg for that first one-- many (most?) PPMIs use SAML, in which case the domain is already locked down. The only case where that wouldn't work is if a PPMI uses SAML for their employees & they want to allow contractors from other companies use their PPMI, but not join their org/teams in the PPMI, which feels like a long shot!

rafaelromcar-parabol: Discussed sync with @Dschoordsch  during our 1-1.

- A script in `scripts/toolboxSRC` added to the `dist` folder after build that adds su to the users listed in a variable is enough. That script does not need to be able to remove su for the rest of users.
- @Dschoordsch  will take this task

Dschoordsch: #8428

Dschoordsch:
Closing in favour of #8428

Issue #8084, opened 2023-04-21
TypeError: Cannot destructure property 'isConnected' of 't' as it is undefined.


Dschoordsch:
https://parabol.sentry.io/issues/3921990898/?referrer=github_plugin

```
TypeError: Cannot destructure property 'isConnected' of 't' as it is undefined.
  at <anonymous> (./packages/client/components/DashboardAvatars/DashboardAvatars.tsx:109:14)
  at Array.forEach (<anonymous>)
  at sortedAvatars (./packages/client/components/DashboardAvatars/DashboardAvatars.tsx:107:17)
  at j (./packages/client/components/DashboardAvatars/DashboardAvatars.tsx:104:25)
  at Set.forEach (<anonymous>)
...
(45 additional frame(s) were not displayed)
```

Dschoordsch:
1k users, 1k events.

avivapinchas:
Closed in favor of https://github.com/ParabolInc/parabol/issues/8088

Dschoordsch:
While they're very similar and might be worked on in conjunction, the symptoms differ. Here the user is nullish while in #8088 the user appears valid and the team member is nullish.

igorlesnenko:
There were 13 events for the last 24 hours, I'll set it as P2 so we can look a bit later.

igorlesnenko:
Now only 3 errors for the last 30 days, closing, unassigning in sentry, if it will appear again then it will be reported by error reporter again

PR #8387, opened 2023-06-16
chore: spread DashboardAvatars_team fragment everywhere

igorlesnenko:
attempt to solve https://github.com/ParabolInc/parabol/issues/8084

Still not sure what is going on, can't reproduce the bug

Trying to spread fragment everywhere and check if it will fix the bug. Like I did [here](https://github.com/ParabolInc/parabol/pull/8160/commits/864c346cfbb71e5f2c5afafd8169df1a7911b7c2)

### DIFF
--- a/packages/client/components/SuggestMentionableUsers.tsx
+++ b/packages/client/components/SuggestMentionableUsers.tsx
@@ -26,6 +26,7 @@ const query = graphql`
           picture
           preferredName
         }
+        ...DashboardAvatars_team
       }
     }
   }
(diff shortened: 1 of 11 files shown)


mattkrick:
Adding an extra fragment to a query won't fix it because if the query does not have the right data, the component won't render until that data is fetched. Even with those fragments as written, they won't help the component because those query fragments are only accessible by the component with the matching name ( you can do a console.log on the returned result to see what i mean, fragments are just obfuscated behind a namespace in the object).

If the bug is in client code, the change _must_ be related to an imperative change via mutation or subscription updater. If we instead add the query fragment to a mutation payload that could possibly solve the problem?

PR #8397, opened 2023-06-20
feat: Add eventType filter in timeline view

tianrunhe:
## Description
There are two renames that contribute to the bigger PR:
1. `packages/client/utils/useUserTaskFilters.ts` → `packages/client/utils/useQueryParameterParser.ts`
2. `packages/client/utils/constructTeamFilterQueryParamURL.ts` → `packages/client/utils/constructFilterQueryParamURL.ts`

## Testing scenarios

- [ ] Filter certain event type
  - Go to timeline page
  - See the filter menu should default to "All Events"
  - Select an event type from the menu and see corresponding events displayed

- [ ] Back to "All Events"
  - When an event type is selected & filtered from the menu, open the menu again to see "All Events" option present

- [ ] Other team filters
  - Verify team & user filters in Team, Tasks & Meetings Page still work as expected

## Final checklist


packages/client/components/TeamlineEventTypeMenu.tsx
```diff
@@ -0,0 +1,70 @@
...
+    return {
+      eventTypeValues,
+      defaultActiveIdx:
+        eventTypeValues.findIndex((eventType) => eventTypeFilters?.includes(eventType)) +
+        (showAllEvents ? 2 : 1)
+    }
+  }, [eventTypeFilters])
+
+  return (
+    <Menu
```
nickoferrall, line 40: -1 two small UI things:

1. A vertical scroll bar

vertical-scroll-bar


2. There's not quite enough space for the final menu item. I think it need some extra padding to match to first menu item 
Screenshot 2023-07-03 at 16 35 58

packages/client/components/TimelineFeedList.tsx
```diff
@@ -28,7 +28,8 @@ const TimelineFeedList = (props: Props) => {
       fragment TimelineFeedList_query on Query
       @refetchable(queryName: "TimelineFeedListPaginationQuery") {
         viewer {
-          timeline(first: $first, after: $after) @connection(key: "TimelineFeedList_timeline") {
+          timeline(first: $first, after: $after, eventTypes: $eventTypes)
```
nickoferrall, line 31: -1 if there are no events for that eventType, there's an empty page. I think some empty state would be helpful here. Could add this in a separate PR 

Screenshot 2023-07-03 at 16 39 12

nickoferrall (changes requested):
This is such a helpful feature 🙌  I've wanted this for a while

packages/client/components/TeamlineEventTypeMenu.tsx
tianrunhe: These are existing issues: if you have 5 teams and open the team filter in team dash or user dash, you will see the vertical scrolling bar & no padding for the last menu item. I think this has something to do with the `Menu` or `MenuItem` components that are reused across the app. I feel like fixing it is out of the scope of this PR. WDYT?

packages/client/components/TimelineFeedList.tsx
tianrunhe: There should always be account creation and default team creation events (you can't archive them as meeting events so they should always be there): 

Screenshot 2023-07-08 at 14 22 31

The fact that you have nothing locally feels like a local data issue. I don't have tension changing this at the moment.

nickoferrall: > There should always be account creation and default team creation events

But if you filter by Sprint Poker, it would only show the Sprint Poker events? So if you don't have any Sprint Poker meetings, I think we need to show some empty state to avoid a blank page

packages/client/components/TeamlineEventTypeMenu.tsx
nickoferrall: Yeah, I agree - it looks like an existing issue, so I think it's out of scope for this PR. A GitHub issue would be helpful though

tianrunhe: Created an issue [here](https://github.com/ParabolInc/parabol/issues/8500)

packages/client/components/SuggestedActionCard.tsx
```diff
@@ -54,7 +54,7 @@ const FloatingSealIcon = styled('div')({
   width: 36,
   position: 'absolute',
   svg: {
-    fontSize: 36
+    fontSize: 20
```
tianrunhe, line 57: Was around here and saw this:

Screenshot 2023-07-10 at 16 27 19

So a quick fix

nickoferrall (approved):

packages/client/utils/constants.ts
```diff
@@ -193,3 +193,25 @@ export const reasonsToDowngradeLookup: Record<
   "Not using Parabol's paid features": 'notUsingPaidFeatures',
   'Moving to another tool (please specify)': 'anotherTool'
 }
+
+export const timelineEventTypeMenuLabels: Record<string, string> = {
```
Dschoordsch, line 197: +1 It would be nice to use `TimelineEventEnum` as key here. We could just import it from
```ts
import {TimelineEventEnum} from '../__generated__/MyDashboardTimelineQuery.graphql'
```
Type needs to be changed in `EventTypeFilterMenuItemLabel#Props` as well, but then it just works.

packages/client/components/EventTypeFilterMenuItemLabel.tsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react'
+import {timelineEventTypeMenuIcons, timelineEventTypeMenuLabels} from '../utils/constants'
+import TimelineEventTypeIcon from './TimelineEventTypeIcon'
+import MenuItemLabel from './MenuItemLabel'
+import {FilterLabels} from '../types/constEnums'
+
+interface Props {
+  eventType?: string
```
Dschoordsch, line 8: +1 I would change this type to `TimeLineEventEnum`, see other comment

packages/client/components/TimelineFeedList.tsx
```diff
@@ -99,6 +101,18 @@ const TimelineFeedList = (props: Props) => {
...
   }, [timeline.edges])
 
+  if (freeHistory === undefined || freeHistory.length === 0) {
+    return (
+      <div className='text-base'>
+        Looks like you have no events of this type.
+        <Link to={'/me'} className='font-sans font-semibold text-sky-500 no-underline'>
+          {' Click here'}
+        </Link>
+        {'  to see all of the events.'}
```
Dschoordsch, line 111: +1 `<Link...>Show all events.</Link>` would avoid the "Click here" link text.

Dschoordsch (changes requested):

packages/client/components/TimelineEventPokerComplete.tsx
```diff
@@ -78,8 +77,7 @@ const TimelineEventPokerComplete = (props: Props) => {
 
   return (
     <TimelineEventCard
-      IconSVG={locked && canUpgrade ? undefined : <CardsSVG />}
-      iconName={locked && canUpgrade ? 'lock' : undefined}
+      iconName={locked && canUpgrade ? 'lock' : 'style'}
```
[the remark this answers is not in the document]
tianrunhe: I know they are the not the same but I feel this does make the code so much simpler (i.e., consolidation on using Material icons vs. customized SVG). I actually got it from searching `cards` in [Material Icons website](https://mui.com/material-ui/material-icons/?query=cards):

Screenshot 2023-07-12 at 14 32 49

@acressall WDYT? Safe to try to change this icon?

What's in production today:
Screenshot 2023-07-12 at 14 30 19

Proposed change in this PR:
Screenshot 2023-07-12 at 14 29 47

acressall:
@tianrunhe is it possible to rotate it so it's technically upside-down? I think it will read as a hand of cards that way. Right now it makes me think of paint chips or pantone swatches.

tianrunhe:
> @tianrunhe is it possible to rotate it so it's technically upside-down? I think it will read as a hand of cards that way. Right now it makes me think of paint chips or pantone swatches.

Technically we can but that would go against the purpose of making the code simpler (to just use MUI icon directly). In that case, I guess I'll just keep it the old way then. Thanks for the input!

packages/client/components/EventTypeFilterMenuItemLabel.tsx
```diff
@@ -0,0 +1,30 @@
...
+  const {eventType} = props
+  const eventTypeLabel = eventType
+    ? timelineEventTypeMenuLabels[eventType]
+    : FilterLabels.ALL_EVENTS
+  const eventTypeIconName = eventType ? timelineEventTypeMenuIcons[eventType] : 'timeline'
+  return (
+    <>
+      <span className='py-1 pl-4'>
+        <TimelineEventTypeIcon iconName={eventTypeIconName} />
+        {!eventTypeIconName && <CardsSVG />}
```
Dschoordsch, line 23: +1 this feels a bit odd. So basically all `eventType`s that aren't in the `timelineEventTypeMenuIcons` map get the `CardSVG` logo. Might be technically correct, still feels a bit odd as logic without a comment.

Dschoordsch (approved):

PR #8401, opened 2023-06-20
fix: del instead of hdel

### LINKED ISSUE #8400: Redis: bad hdel
See https://github.com/ParabolInc/parabol/issues/8176#issuecomment-1599080468

AC
- [ ] no error in segment

mattkrick:
# Description

fixes #8400

HDEL is for deleting a specific key in a hash. DEL is for deleting the entire hash.


mattkrick:
this is so itty bitty i'm just gonna merge

PR #8407, opened 2023-06-20
feat: handle seasonal templates (Part 3 of X)

mattkrick:
# Description

Fixes seasonal template handling


jmtaber129 (commented):
I'm not sure this is actually needed, since the only seasonal templates are public templates, which are already handled in https://github.com/ParabolInc/parabol/pull/8081/files#diff-251498fbe60c06af717972b077aff34c3b813edb6ce6b886789a836481c84ca8R43-R46.

Otherwise code looks good if there is a reason why this is needed.

mattkrick:
ah! yep, customRedisQueries is going away in https://github.com/ParabolInc/parabol/issues/8414. now that templates live in PG we don't need to use redis as an intermediary cache

PR #8408, opened 2023-06-20
feat: FixedActivity in GQL (Part 4 of 4)

mattkrick:
# Description

- Adds FixedActivity in GQL
- Improves sort order of activities

### DIFF
--- a/codegen.json
+++ b/codegen.json
@@ -3,6 +3,7 @@
     "content": "// eslint-disable-next-line @typescript-eslint/ban-ts-comment\n// @ts-nocheck",
     "defaultScalarType": "string",
     "enumsAsTypes": true,
+    "optionalResolveType": true,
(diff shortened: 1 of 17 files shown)


jmtaber129 (changes requested):
-1 This seems to break loading non-public templates (i.e. only public templates load now), not sure why - the code _looks_ like it should work, but I haven't dug into debugging.

Also, as a reviewer, I generally prefer to see new GQL fields + backend resolvers added in the same PR where they're used in the frontend - then it's easier to test the full feature E2E and understand how new GQL fields fit into frontend code. E.g. I can read the code for the new `viewerLowestScope` and `illustrationUrl` fields + resolvers, but as a reviewer I can't understand how they work in context without looking at https://github.com/ParabolInc/parabol/pull/8293, and I can't easily test them without just testing that other PR as well.

jmtaber129 (approved):

PR #8411, opened 2023-06-21
fix: canAccess for organization

Dschoordsch:
# Description

There was an error in the `canAccess` field which should have been caught by `yarn typecheck`. Looking into this, it turned out that we did not run any typecheck on the `server/graphql` sources.

## Testing scenarios

- Full release test

## Final checklist

### DIFF
--- a/packages/server/database/types/NotificationPromptToJoinOrg.ts
+++ b/packages/server/database/types/NotificationPromptToJoinOrg.ts
@@ -6,7 +6,7 @@ interface Input {
 }
 
 export default class NotificationPromptToJoinOrg extends Notification {
-  orgId: string
+  orgId?: string
   activeDomain: string
   constructor(input: Input) {
     const {userId, activeDomain} = input
--- a/packages/server/database/types/NotificationPromptToJoinOrg.ts
+++ b/packages/server/database/types/NotificationPromptToJoinOrg.ts
@@ -6,7 +6,6 @@ interface Input {
 }
 
 export default class NotificationPromptToJoinOrg extends Notification {
-  orgId?: string
   activeDomain: string
   constructor(input: Input) {
     const {userId, activeDomain} = input
--- a/packages/server/graphql/private/mutations/stripeSucceedPayment.ts
+++ b/packages/server/graphql/private/mutations/stripeSucceedPayment.ts
@@ -1,3 +1,4 @@
+import Stripe from 'stripe'
 import getRethink from '../../../database/rethinkDriver'
 import updateTeamByOrgId from '../../../postgres/queries/updateTeamByOrgId'
 import {isSuperUser} from '../../../utils/authorization'
@@ -25,7 +26,7 @@ const stripeSucceedPayment: MutationResolvers['stripeSucceedPayment'] = async (
   const {
     livemode,
     metadata: {orgId}
-  } = await manager.retrieveCustomer(customerId)
+  } = (await manager.retrieveCustomer(customerId)) as Stripe.Customer
   const org = await r.table('Organization').get(orgId).run()
   if (!org || !orgId) {
     if (livemode) {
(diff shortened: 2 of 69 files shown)


packages/server/graphql/private/mutations/stripeSucceedPayment.ts
```diff
@@ -25,7 +26,7 @@ const stripeSucceedPayment: MutationResolvers['stripeSucceedPayment'] = async (
   const {
     livemode,
     metadata: {orgId}
-  } = await manager.retrieveCustomer(customerId)
+  } = (await manager.retrieveCustomer(customerId)) as Stripe.Customer
```
BartoszJarocki, line 29: +1 same as ☝️

BartoszJarocki (approved):
great find and nice work on fixing all the issues! some comments left

packages/server/graphql/public/types/User.ts
```diff
@@ -42,7 +42,7 @@ const User: UserResolvers = {
       case 'Organization':
         const organizationUser = await dataLoader
           .get('organizationUsersByUserIdOrgId')
-          .load({userId: viewerId, id})
+          .load({userId: viewerId, orgId: id})
```
mattkrick, line 45: tsc saves the day 🎉

packages/server/graphql/public/types/MeetingSeries.ts
```diff
@@ -18,7 +18,7 @@ const MeetingSeries: MeetingSeriesResolvers = {
       .orderBy((doc) => (doc('endedAt') ? 0 : 1), r.desc('createdAt'))
       .limit(1)
       .run()
-    return meetings[0]
+    return meetings[0]!
```
mattkrick, line 21: -1 should we change the GQL Schema to make `mostRecentMeeting` nullable? It's possible they've never run one before

packages/server/graphql/public/mutations/acceptTeamInvitation.ts
```diff
@@ -23,11 +23,6 @@ const acceptTeamInvitation: MutationResolvers['acceptTeamInvitation'] = async (
   {invitationToken, notificationId},
   {authToken, dataLoader, socketId: mutatorId}
 ) => {
-  // AUTH WORKAROUND
-  if (!isAuthenticated(authToken) || !invitationToken) {
-    // Workaround for https://github.com/zalando-incubator/graphql-jit/issues/171
-    return {}
-  }
```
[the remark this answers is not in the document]
mattkrick: -1 i did a bad job of linking my comments in these 2 files. can you re-enable the auth check here?
 https://github.com/ParabolInc/parabol/blob/099830077cef534da9d0ff546b0fb3898e7dc3dc/packages/server/graphql/public/permissions.ts#L31-L33

packages/server/graphql/private/queries/user.ts
```diff
@@ -5,7 +5,7 @@ const user: QueryResolvers['user'] = async (_source, {email, userId}, {dataLoade
   if (email) {
     return getUserByEmail(email)
   }
-  return dataLoader.get('users').load(userId)
+  return (userId && (await dataLoader.get('users').load(userId))) || null
```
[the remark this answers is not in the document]
mattkrick: the way it's written it also handles if the dataloader returns `undefined`.
To fix, we should probably make the User dataloader return `User | null` instead of `User | undefined`.
The weird thing is, I don't really see where it's getting `undefined` from 😕

packages/server/database/types/NotificationPromptToJoinOrg.ts
```diff
@@ -6,7 +6,7 @@ interface Input {
 }
 
 export default class NotificationPromptToJoinOrg extends Notification {
-  orgId: string
+  orgId?: string
```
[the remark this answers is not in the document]
Dschoordsch: Seems it's not

packages/server/graphql/public/types/MeetingSeries.ts
Dschoordsch: I'm not too deep into meeting series. I think there cannot be a meeting series without at least one meeting.

mattkrick: oh! fair point, LGTM

PR #8412, opened 2023-06-21
feat(share-summary): Share summary mutation

### LINKED ISSUE #8233: Share Summary: Share topic mutation
image

image

**AC:**

- When button is clicked, message is shared to the selected slack channel (see the screenshot)
- Success snackbar is shown with a link to slack message

**Not included**

- Other than slack integrations

See Bruce work: https://github.com/ParabolInc/parabol/commits/hack/shareTopic
Designs: https://www.figma.com/file/hQRaJxKfSgOQO6KXyclK0Y/Share-Summary-Topics?type=design&node-id=282-300

igorlesnenko:
Fixes #8233

image

How to test: 
- Follow testing steps from PR to see the dialog https://github.com/ParabolInc/parabol/pull/8331
- Share topic

Not included:

- Snackbar


packages/server/graphql/public/typeDefs/shareTopic.graphql
```diff
@@ -0,0 +1,33 @@
+extend type Mutation {
+  """
+  Describe the mutation here
+  """
+  shareTopic(
+    """
+    reflectionGroupId
```
nickoferrall, line 7: +1 these comments should be updated as they're referring to different fields

packages/server/graphql/public/mutations/shareTopic.ts
```diff
@@ -0,0 +1,50 @@
...
+  const {teamId} = meeting
+
+  if (!isTeamMember(authToken, teamId)) {
+    return standardError(new Error('Team not found'), {userId: viewerId})
+  }
+
+  let stage = null
+  let stageIndex = -1
+
+  const discussPhase = meeting.phases.find((phase) => phase.phaseType === 'discuss')
```
nickoferrall, line 23: +1 you can use the getPhase helper here: https://github.com/ParabolInc/parabol/blob/ee91d7feb48aaab18e4a712af135db97700c60aa/packages/server/utils/getPhase.ts

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
```diff
@@ -297,5 +297,121 @@ export const SlackNotifier: Notifier = {
...
+    const slackBlocks = [
+      {
+        type: 'section',
+        text: {
+          type: 'mrkdwn',
+          text: `<@${slackAuth.slackUserId}> has shared reflections about *"${topic}"* from their retrospective`
+        }
+      },
+      {
+        type: 'section',
```
nickoferrall, line 371: +1 you can use the `makeSection` function here to reduce the code: https://github.com/ParabolInc/parabol/blob/88c2bd5e41cce781ca497a6047642f2c46e73680/packages/server/graphql/mutations/helpers/notifications/makeSlackBlocks.ts

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
```diff
@@ -297,5 +297,121 @@ export const SlackNotifier: Notifier = {
...
+
+    if (!slackAuth) {
+      throw new Error('Slack auth not found')
+    }
+
+    const {botAccessToken, slackUserId} = slackAuth
+    const manager = new SlackServerManager(botAccessToken!)
+    const channelInfo = await manager.getConversationInfo(channelId)
+
+    if (channelId !== slackUserId) {
```
nickoferrall, line 337: Why would the `channelId` be the same as the `slackUserId`? Isn't the `slackUserId` the `id` of the user?

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
```diff
@@ -297,5 +297,121 @@ export const SlackNotifier: Notifier = {
...
+      throw new Error('Team not found')
+    }
+
+    if (!slackAuth) {
+      throw new Error('Slack auth not found')
+    }
+
+    const {botAccessToken, slackUserId} = slackAuth
+    const manager = new SlackServerManager(botAccessToken!)
+    const channelInfo = await manager.getConversationInfo(channelId)
```
nickoferrall, line 335: +1 can move this below the conditional on line 337

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
```diff
@@ -297,5 +297,121 @@ export const SlackNotifier: Notifier = {
...
+      {
+        type: 'section',
+        text: {
+          type: 'mrkdwn',
+          text: `*Reflections:* \n${reflectionsText}`
+        }
+      }
+    ]
+
+    const notificationChannel = {
```
nickoferrall, line 406: -1 need to add an id here or a TODO to explain why we don't need it

packages/client/hooks/useModalPortal.tsx
```diff
@@ -81,7 +81,7 @@ const ModalContents = styled('div')<{portalStatus: PortalStatus}>(({portalStatus
   display: 'flex',
   flex: '0 1 auto',
   flexDirection: 'column',
-  position: 'relative',
+  position: 'fixed',
```
nickoferrall, line 84: -1 `useModalPortal` is being used by `useModal`, which is used in 36 files. I don't think we should change the styles here, or we'll need to spend a long time making sure that all the other modals work as expected

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
```diff
@@ -297,5 +297,121 @@ export const SlackNotifier: Notifier = {
...
+          throw new Error('Unable to join slack channel')
+        }
+      }
+    }
+
+    const topic = reflectionGroup.title
+    const discussionUrl = makeAppURL(appOrigin, `meet/${meetingId}/discuss/${stageIndex + 1}`)
+    const meetingUrl = makeAppURL(appOrigin, `meet/${meetingId}`)
+
+    const reflectionsText = reflections
```
nickoferrall, line 358: +1 should there be a limit to the reflection text? If a topic has many very long reflections, should we only show some of it?

nickoferrall (changes requested):
Two things I noticed while testing:

1. If I add many very long reflections to a topic and click Share Topic, the topic isn't shared, and there's no error feedback.
2. After initially posting in the channel, I saw a snackbar pop-up that said a different channel name. I _think_ this happened before I pulled the latest update on this branch, and I know your PR description says the snackbar isn't included.

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
igorlesnenko: I'm not sure if we need to enforce the limit on the number of reflections. Feels like we are trying to read the user's mind, whether he wanted to keep 10 or 20 reflections to be shared. Slack has the limit of 3000 characters per block. I now check this and truncate the text in this case. Will it be sufficient for now?

nickoferrall: Yeah, I was thinking character limit rather than limiting the number of reflections. I think it makes sense to go with the same limit that Slack sets so 3,000 characters sounds good to me 👍

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
```diff
@@ -297,5 +309,85 @@ export const SlackNotifier: Notifier = {
...
+    channelId: string
+  ) {
+    const r = await getRethink()
+    const [user, team, meeting, reflectionGroup, reflections, slackAuth] = await Promise.all([
+      dataLoader.get('users').load(userId),
+      dataLoader.get('teams').load(teamId),
+      dataLoader.get('newMeetings').load(meetingId),
+      dataLoader.get('retroReflectionGroups').load(reflectionGroupId),
+      r.table('RetroReflection').getAll(reflectionGroupId, {index: 'reflectionGroupId'}).run(),
+      r.table('SlackAuth').getAll(userId, {index: 'userId'}).filter({teamId})(0).run()
```
nickoferrall, line 330: -1 if there's no slackAuth, I think this would cause an error. Safer to do: `.nth(0).default(null)`

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
```diff
@@ -297,5 +309,85 @@ export const SlackNotifier: Notifier = {
...
+    teamId: string,
+    meetingId: string,
+    reflectionGroupId: string,
+    stageIndex: number,
+    channelId: string
+  ) {
+    const r = await getRethink()
+    const [user, team, meeting, reflectionGroup, reflections, slackAuth] = await Promise.all([
+      dataLoader.get('users').load(userId),
+      dataLoader.get('teams').load(teamId),
```
nickoferrall, line 326: +1 can use `loadNonNull` and won't need to check if `team` exists below

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
```diff
@@ -297,5 +309,85 @@ export const SlackNotifier: Notifier = {
...
+    userId: string,
+    teamId: string,
+    meetingId: string,
+    reflectionGroupId: string,
+    stageIndex: number,
+    channelId: string
+  ) {
+    const r = await getRethink()
+    const [user, team, meeting, reflectionGroup, reflections, slackAuth] = await Promise.all([
+      dataLoader.get('users').load(userId),
```
nickoferrall, line 325: +1 I don't think we need to check if the `user` (the viewer) exists. We're already checking if the viewer is a teamMember in the `shareTopic` mutation

packages/server/graphql/mutations/helpers/notifications/makeSlackBlocks.ts
```diff
@@ -1,20 +1,40 @@
 import generateUID from '../../../../generateUID'
 
-export const makeSection = (text: string) => ({
-  type: 'section',
-  text: {
-    type: 'mrkdwn',
-    text
+// Slack has the limit of 3000 chars per block
+const SLACK_MAX_BLOCK_CHAR_COUNT = 3000
+const truncationString = `[...]`
```
nickoferrall, line 5: +1 I would have thought `...` is more common for truncation than `[...]` but no strong feelings here

packages/server/graphql/mutations/helpers/notifications/makeSlackBlocks.ts
```diff
@@ -1,20 +1,40 @@
...
 
-export const makeSections = (fields: string[]) => ({
-  type: 'section',
-  fields: fields.map((field) => ({
-    type: 'mrkdwn',
-    text: field
-  }))
-})
+export const makeSection = (text: string) => {
+  text = maybeTruncate(text)
```
nickoferrall, line 15: +1 rather than reassign the input, I think it would be easier to read if it's a new variable, like `const maybeTruncatedText = ...`

nickoferrall (changes requested):

packages/server/graphql/mutations/helpers/notifications/makeSlackBlocks.ts
igorlesnenko: @nickoferrall my thought is that `...` could be confused with real `...`

packages/server/graphql/mutations/helpers/notifications/makeSlackBlocks.ts
igorlesnenko: @nickoferrall my thought is that `...` could be confused with real `...`

nickoferrall (approved):

packages/server/graphql/mutations/helpers/notifications/makeSlackBlocks.ts
nickoferrall: Makes sense 👍

igorlesnenko:
@tghanken Can you please advise what kind of events are needed here?

Dschoordsch (approved):

tghanken (approved):

PR #8416, opened 2023-06-21
fix: use unique server name for fresh namespace

### LINKED ISSUE #8176: Redis Memory Leak
It's clear we have a redis memory leak.
Looks like it started sometime between April 15 - April 23.
To fix, use something like redis commander or similar to find the list of keys & determine which ones aren't being flushed.

Screenshot 2023-05-10 at 10 08 23 AM

AC
- [ ] root cause identified
- [ ] patch written in the code
- [ ] manual flush of offending keys triggered

mattkrick:
# Description

Fix #8176

Follow-up to #7208

Previously, I used the SERVER_ID for each socket server. Since we only have 1 socket server for production, that means both the old server & new server have the same SERVER_ID = 0. That means it looks like the server never went down, which causes user presence to keep growing in redis.

To fix, I attached the pid to the SERVER_ID. Now when one comes online, it's seen as unique & can flush the records associated with the old server. 

## Demo

https://www.loom.com/share/df84a2611c764f53a0eadfad566fe006

## Testing scenarios

- [ ] See demo for full tests. Kill a server abruptly. When a new server comes online, see that the user presence redis namespace is cleaned of the old values


mattkrick:
thanks for the great test & repro!

...i'm pretty ashamed that i didn't catch the bug earlier. that was a big dumb error on my part.

packages/server/utils/ServerHealthChecker.ts
```diff
@@ -64,8 +65,8 @@ export default class ServerHealthChecker {
           const connections = record[1]
           return connections.map((connection) => {
             const presence = JSON.parse(connection) as UserPresence
-            const {socketServerId, socketId} = presence
-            if (!deadServerIds.includes(socketServerId)) return
```
Dschoordsch, line 68: -1 I think how we're detecting dead servers is wrong. We're only considering server present in socketServers
Screenshot from 2023-06-26 12-08-43
but we can still have stray presence entries
Screenshot from 2023-06-26 12-08-54
which will consequently never get cleaned up.
I think the best course of action would be to use `socketServers` to ping, construct a list of alive servers and remove all presence entries which are not in the alive servers list. This will also take care of old entries which have `socketInstanceId === undefined`

Dschoordsch (changes requested):

packages/server/utils/ServerHealthChecker.ts
mattkrick: this sounds great! i much prefer `include` logic vs. `exclude`. will fix tomorrow

mattkrick:
@Dschoordsch ready for another review! I removed the redis key `socketServers` and just compute that on the fly now. Simpler API, simpler code, little more CPU intensive.

It's called on shutdown because there's a chance that a server shuts down & doesn't start back up (ie we downscale a deployment).
It's called on startup because there's a chance that the server exited abruptly & the shutdown code didn't get called.
We could just run it on startup if we want to ignore the edge case of downscaling the # of servers.

packages/server/utils/ServerHealthChecker.ts
```diff
@@ -8,45 +8,41 @@ import sendToSentry from './sendToSentry'
...
+        if (!this.remoteSocketServers) {
+          console.error('unsolicited pong received before getLivingServers was called')
+        } else {
+          this.remoteSocketServers.push(remoteServerId)
+        }
       }
     })
-    this.subscriber.subscribe('socketServerPing', `socketServerPong:${SERVER_ID}`)
-    this.publisher.sadd('socketServers', SERVER_ID)
+    this.subscriber.subscribe('socketServerPing', `socketServerPong:${INSTANCE_ID}`)
```
Dschoordsch, line 30: -1 I did some testing by running `yarn dev` and then running `SERVER_ID=9 node scripts/runSocketServer.js`. It turns out this would clear the presence completely. After some digging, this is the culprit. We need to await the `subscribe` call before doing any pings.

mattkrick: i can't reproduce this locally, but i believe you!

while debugging i also found out it was possible to double subscribe to messages, so i made it a singleton.
Also, since we've never downscaled the number of servers we have, i got rid of calling this when the server goes down & only when they come up. let me know what you think!

packages/server/utils/ServerHealthChecker.ts
```diff
@@ -8,45 +8,47 @@ import sendToSentry from './sendToSentry'
...
-  subscriber = new Redis(REDIS_URL, {connectionName: 'serverHealth_sub'})
-  pendingPongs: null | Set<string> = null
-  constructor() {
+class ServerHealthChecker {
+  private publisher = new Redis(REDIS_URL, {connectionName: 'serverHealth_pub'})
+  private subscriber = new Redis(REDIS_URL, {connectionName: 'serverHealth_sub'})
+  private remoteSocketServers: null | string[] = null
+  private joinPoolPromise: null | Promise<void> = null
+  private async joinPool() {
+    if (this.joinPoolPromise) throw new Error('joinPool can only be called once')
```
Dschoordsch, line 19: -1 I don't like that this read and check of `this.joinPoolPromise` is inside this call, whereas the write is outside. I think the check and assignment should either be completely in `joinPool` or completely in `getLivingServers`.
I think cleanest would be just putting all the logic inside `joinPool`
```ts
private async joinPool() {
  if (!this.joinPoolPromise) {
    this.joinPoolPromise = async () => {
      await this.subscriber.subscribe('socketServerPing', `socketServerPong:${INSTANCE_ID}`)
      ...
    }()
  }
  return this.joinPoolPromise
}
```

packages/server/listenHandler.ts
```diff
@@ -1,17 +1,16 @@
 import {us_listen_socket} from 'uWebSockets.js'
 import PROD from './PROD'
 import getGraphQLExecutor from './utils/getGraphQLExecutor'
-import ServerHealthChecker from './utils/ServerHealthChecker'
+import serverHealthChecker from './utils/serverHealthChecker'
```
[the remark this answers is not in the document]
mattkrick: @Dschoordsch ...this is why reformatting your mac to being case-sensitive is so vital 🤣 . great catch! will trick git

packages/server/utils/ServerHealthChecker.ts
mattkrick: 💯 that is cleaner

Dschoordsch (approved):

PR #8418, opened 2023-06-22
Remove default_incomplete payment behaviour for old subscription flow

nickoferrall:
To handle 3D Secure cards, we're setting the payment behaviour to `default_incomplete` so that it's authorised on the client. While this works for the new payment flow, for the old flow, it results in subscriptions being incomplete. 

Merging this PR right away as it's using the old code, and I want to get this in before the ship.


nickoferrall:
CC @Dschoordsch I merged this PR as it's using the old team subscription code, and I wanted to get it merged before the ship

Issue #8434, opened 2023-06-26
Suggest Groups: improve prompt so that results are more accurate


nickoferrall:
Since experimenting with this feature internally, the results haven't been great as it's only been able to group reflections that are very similar to each other. In most meetings, reflections aren't that similar to one another.

Using GPT-4 instead of 3.5-turbo provides more accurate results, but usually hits a timeout.

### AC

- [ ] Prompt is improved so that pressing the "Suggest Groups" button helpfully groups reflections using real meeting data

PR #8436, opened 2023-06-26
feat(suggest-groups): improve suggest groups prompt

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8434

This PR improves the previous prompt by:
- Splitting the request into two: one request to get themes, the second matches the reflection to one of the themes
- The second request only requests the name of the theme as opposed to a complex JSON response
- As the second request is simple, we can use GPT-4, which provides more accurate results
- The drawback of this approach is we send many OpenAI requests all at once. This might result in rate-limiting problems. Still, it's been working fine locally, so it should at least work internally at Parabol.

Demo: https://www.loom.com/share/be6d477a38314eeea1ce3b711b722717

### To test

- [ ] Add the Suggest Groups org feature flag
- [ ] Create a retro and add some dummy reflections. Go to the Group phase, wait ~10 seconds, click the "Suggest Groups" button and see that the reflections are grouped in a sensible order
- [ ] Create another retro and copy & paste reflections from a real, Parabol meeting in prod (sorry, this is a bit time consuming!). Click "Suggest Groups" and see that the reflections are grouped in a sensible order


tianrunhe:
Your Loom demo seems to be recording the wrong window/desktop 😂

nickoferrall:
😄 I shared the Loom just as I was logging off for the day and didn't have time to update it. I've added a new Loom to the PR description!

packages/server/utils/OpenAIServerManager.ts
```diff
@@ -43,107 +41,101 @@ class OpenAIServerManager {
...
+        frequency_penalty: 0,
+        presence_penalty: 0
+      })
 
-    In the output, "The Retreat" and "Deadlines" are example topic names that you can create to group the reflections.
-    Do not give me code to run. Give me the output as valid JSON in the format above.
+      const theme = (response.data.choices[0]?.message?.content?.trim() as string) ?? null
+      if (!theme || !themes.includes(theme)) {
+        if (!retry) {
+          return getThemeForReflection(reflection, true)
```
nickoferrall, line 104: Sometimes, the model makes up a new theme, so I retry the request once more

packages/server/utils/OpenAIServerManager.ts
```diff
@@ -43,107 +41,101 @@ class OpenAIServerManager {
...
-    const numGroups = Math.ceil(reflectionsText.length / avgReflectionsPerGroup)
-
-    const outputTokens = Math.round(numGroups * avgGroupNameLength + totalLengthInTokens)
+  async generateThemes(reflectionsText: string[]) {
+    if (!this.openAIApi) return null
+    const suggestedThemeCountMin = Math.floor(reflectionsText.length / 5)
+    const suggestedThemeCountMax = Math.floor(reflectionsText.length / 4)
+    const prompt = `Create a short list of common themes given the following reflections: ${reflectionsText.join(
+      ', '
+    )}. Each theme should be no longer than a few words. There should be roughly ${suggestedThemeCountMin} or ${suggestedThemeCountMax} themes. Return the themes as a comma-separated list.`
```
nickoferrall, line 50: I'm requesting roughly this many themes because otherwise it tends to create way too many

packages/server/graphql/mutations/helpers/generateGroups.ts
```diff
@@ -17,18 +17,24 @@ const generateGroups = async (
...
+  if (!groupedReflections) {
     console.warn('ChatGPT was unable to group the reflections')
     return
   }
-  const parsedGroupedReflections = JSON.parse(groupedReflectionsJSON)
   const autogroupReflectionGroups: AutogroupReflectionGroupType[] = []
 
-  for (const group of parsedGroupedReflections) {
-    const groupTitle = Object.keys(group)[0]
+  for (const groupTitle of Object.keys(groupedReflections)) {
```
igorlesnenko, line 34: Alternative approach `for (const [groupTitle, reflectionTexts] of Object.entries(groupedReflections)) {`

packages/server/utils/OpenAIServerManager.ts
```diff
@@ -43,107 +41,101 @@ class OpenAIServerManager {
...
+            content: prompt
+          }
+        ],
+        temperature: 0.7,
+        top_p: 1,
+        frequency_penalty: 0,
+        presence_penalty: 0
+      })
+      const themes = (response.data.choices[0]?.message?.content?.trim() as string) ?? null
+      return themes.split(', ')
```
igorlesnenko, line 67: is it possible that some theme would include `,` in the theme title?

packages/server/utils/OpenAIServerManager.ts
```diff
@@ -43,107 +41,101 @@ class OpenAIServerManager {
...
+      }
+      return theme
+    }
 
-    Here is the list of reflections: """
-    ${reflectionsText}
-    """
-    `
+    try {
+      const promises = reflectionsText.map((reflection) => getThemeForReflection(reflection))
```
igorlesnenko, line 120: Feels heavy making one request per each reflection. I wish in a final version we can make it in a single request or at least few requests. I think otherwise we will hit rate limit quickly.

packages/server/utils/OpenAIServerManager.ts
igorlesnenko: So it ignores this statement for some reason? Maybe somehow change it too? e.g. return the number of the theme

> Your output should just be the theme name, and must be one of the themes I've provided.

igorlesnenko (approved):
Looks suitable for internal testing, need to not forget to do something with big amount of requests in case of more wide release

Didn't work well on real reflections for me
First time it failed with 2 errors
`Couldn't find a suitable theme for reflection:`

I had to refresh page, after refreshing page it created one large group and didn't group all other reflections

I think we need to change request so it return a number of a theme, maybe it will help.

packages/server/utils/OpenAIServerManager.ts
nickoferrall: Yeah, it ignores the explicit instruction 😄 I've tried a few different prompts, but hallucinating is a real problem with ChatGPT. Happy to test requesting the number of the theme, although my gut says that that would be harder for the model as it requires an extra step of looking for the index of the theme. Definitely worth testing though.

nickoferrall:
> Didn't work well on real reflections for me
First time it failed with 2 errors
Couldn't find a suitable theme for reflection:
I had to refresh page, after refreshing page it created one large group and didn't group all other reflections

Interesting! Do you have a screenshot of the reflections that it struggled to group?

packages/server/utils/OpenAIServerManager.ts
nickoferrall: My original approach was to send one request, but because of the complexity, the response was very slow, and the quality of the output wasn't good enough for production. I'd like to get something that works internally and then, if the feature is useful, worry about rate limiting. 

Here are the [OpenAI limits](https://platform.openai.com/account/rate-limits). 

Some possible solutions:
- use the 3.5-turbo model with the new prompts. If the quality is still OK, we might be able to work with a 3,500 request-per-minute limit
- only make it available on the Enterprise tier, for example, to limit the number of requests per minute
- refactor the prompts so that we send fewer requests
- request an increase in our rate limits from OpenAI

igorlesnenko:
Will send screenshots in slack

packages/server/graphql/mutations/helpers/generateGroups.ts
nickoferrall: Oh much cleaner, updated!

packages/server/utils/OpenAIServerManager.ts
nickoferrall: Maybe - I am requesting a comma-separated list, so it shouldn't. I haven't seen it do this after a lot of testing locally, but something to watch out for when testing internally

packages/server/utils/OpenAIServerManager.ts
```diff
@@ -43,107 +41,101 @@ class OpenAIServerManager {
...
+      const themesByReflections = await Promise.all(promises)
```
Dschoordsch, line 121: +1 can we move `await Promise.all` in the line above to get rid of the `promises` variable?

packages/server/utils/OpenAIServerManager.ts
Dschoordsch: -1 can you add this as a comment please, so it's not forgotten?

Dschoordsch (approved):

PR #8438, opened 2023-06-27
fix: Show error message when creating Jira task fails

### LINKED ISSUE #6320: Unable to create a new Jira issue in Poker
When creating a new Jira issue in Poker in prod, I can enter the title but the new issue doesn't submit when pressing enter or clicking outside of the input. I can see in the network requests that the mutation hasn't been executed. 

It does work in GitHub and Parabol.

create-jira-issue

### AC

- [ ] Can create a new Jira issue in Poker

Dschoordsch:
# Description

Fixes #6320 
When creating a Jira task in Sprint Poker failed, the error message wasn't send properly to the client. It also wouldn't be shown to the client.
If task creation failed, show the error message now for 5s.

This does not fix the underlying issue, but presents it to the user, so they hopefully can react accordingly or report more detailed errors.

## Demo

https://www.loom.com/share/d798bb9765394f2a9097a51a1fea41e4?sid=6da48e5b-40ec-4599-88cc-36f3b51e4b8e

## Testing scenarios

- [ ] Try to create a **Jira** issue which produces an error
- [ ] see the error message
- [ ] repeat for **GitHub**
- [ ] repeat for **GitLab**
- [ ] repeat for **Azure DevOps**

Run `node scripts/toolbox/postDeploy.js` if you're missing some of the integrations.
You can test error scenarios by adding a something like
```ts
return {error: new Error('Georg was here')}
```
in https://github.com/ParabolInc/parabol/blob/063a526dbcfd6340f4daa645c5875c679745519b/packages/server/graphql/mutations/helpers/createTaskInService.ts#L59

## Final checklist

### DIFF
--- a/packages/server/graphql/public/types/CreateTaskPayload.ts
+++ b/packages/server/graphql/public/types/CreateTaskPayload.ts
@@ -13,6 +13,7 @@ const CreateTaskPayload: CreateTaskPayloadResolvers = {
   error: async ({error}) => {
     if (!error) return null
     return {
+      // although the error already has the correct shape, without this resolver, the error member in source would not be used as source here, but treated as if it was thrown by graphql
       message: error.message
     }
   },
--- a/packages/server/graphql/public/types/CreateTaskPayload.ts
+++ b/packages/server/graphql/public/types/CreateTaskPayload.ts
@@ -4,13 +4,20 @@ import errorFilter from '../../errorFilter'
 import {CreateTaskPayloadResolvers} from '../resolverTypes'
 
 export type CreateTaskPayloadSource = {
-  taskId: string
+  error?: Error
+  taskId?: string
   notificationIds?: string[]
 }
 
 const CreateTaskPayload: CreateTaskPayloadResolvers = {
+  error: async ({error}) => {
+    if (!error) return null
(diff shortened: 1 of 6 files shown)


packages/server/graphql/public/types/CreateTaskPayload.ts
```diff
@@ -4,13 +4,20 @@ import errorFilter from '../../errorFilter'
...
+  error?: Error
+  taskId?: string
   notificationIds?: string[]
 }
 
 const CreateTaskPayload: CreateTaskPayloadResolvers = {
+  error: async ({error}) => {
+    if (!error) return null
+    return {
+      message: error.message
```
Dschoordsch, line 16: Although the error already has the correct shape, without this resolver the error member in source would not be used as source here, but treated as if it was thrown by graphql.

Dschoordsch:
I tried to add [`status=live`](https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-projects/#api-rest-api-3-project-search-get) filter to the projects listing to filter out the non-existent projects, but that does not do anything.

igorlesnenko (approved):

Issue #8133, opened 2023-05-02
Suggest Groups: add loading feedback


nickoferrall:
### AC

- [ ] When proceeding to the Group phase, it takes several seconds before OpenAI returns the groups. During that time, we need to let the user know that they can't click the Suggest Groups button

PR #8440, opened 2023-06-27
feat(suggest-groups): add loading feedback while waiting for OpenAI response

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8133

Demo: https://www.loom.com/share/47d247ba6dce446fa22917f79f4cc542

### To test

- Add the Suggest Groups org feature flag
- Add a bunch of reflections and proceed to the Group phase
- See that the Suggest Groups button is disabled until `autogroupReflectionGroups` exists in our db


packages/server/graphql/mutations/helpers/generateGroups.ts
```diff
@@ -53,6 +57,10 @@ const generateGroups = async (
 
   const r = await getRethink()
   await r.table('NewMeeting').get(meetingId).update({autogroupReflectionGroups}).run()
+  const data = {meetingId}
+  const operationId = dataLoader.share()
+  const subOptions = {operationId, mutatorId: SERVER_ID}
```
nickoferrall, line 62: We need the `mutatorId` but don't have access to the context here. I'm using the `server_id` so the subscription is published to everyone, but I'm not sure if there's a better approach

igorlesnenko (approved):
Feels a bit confusing without the hint or loading indicator why button is disabled

I also notices that when I reset groups and go back to adding reflections, group button does not pick up new reflections, is it correct?

nickoferrall:
> group button does not pick up new reflections, is it correct?

Yep, that's right. It won't pick up the new reflections added. I meant to mention that in the demo 👍  I think it's an edge case that we don't need to worry about for now

packages/server/graphql/mutations/helpers/generateGroups.ts
Dschoordsch: -1 I would prefer if we'd publish subscription updates in mutations as it's easier to find and reason about. However I found 1 example where we do it in a helper and there we pass the `subOptions` as parameter https://github.com/ParabolInc/parabol/blob/4c607c8481398bc54684bb36ab7f239d1b11eb90/packages/server/graphql/public/mutations/helpers/updateNotification.ts#L5

packages/server/graphql/public/typeDefs/generateGroups.graphql
Dschoordsch, line 1: -1 If a file only contains 1 type, the filename should match it.

packages/client/components/RetroGroupPhase.tsx
```diff
@@ -55,6 +55,9 @@ const RetroGroupPhase = (props: Props) => {
         id
         endedAt
         showSidebar
+        autogroupReflectionGroups {
+          groupTitle
```
Dschoordsch, line 59: -1 I think you should use the fragment `GenerateGroupsFrag_meeting` here (you could even just put it in this file) to indicate these need to match for the update.

Dschoordsch (changes requested):
works

packages/client/components/RetroGroupPhase.tsx
nickoferrall: I can't use `GenerateGroupsFrag_meeting` here as it's called on the `GenerateGroupsSuccess` type. Perhaps I could create the fragment within the component to make it clearer? Something like this?

```
  graphql`
    fragment RetroGroupPhaseGenerateGroups_meeting on RetrospectiveMeeting {
      autogroupReflectionGroups {
        groupTitle
      }
    }
  `
  const meeting = useFragment(
    graphql`
      fragment RetroGroupPhase_meeting on RetrospectiveMeeting {
        ...StageTimerControl_meeting
        ...StageTimerDisplay_meeting
        ...GroupingKanban_meeting
        id
        endedAt
        showSidebar
        ...RetroGroupPhaseGenerateGroups_meeting
        organization {
          tier
          featureFlags {
            suggestGroups
          }
        }
      }
    `,
    meetingRef
  )
```

nickoferrall:
I'll merge this PR as Georg's away this week, there was nothing too concerning in the code review feedback, and it will only be used for an internal experiment

packages/server/graphql/mutations/helpers/generateGroups.ts
Dschoordsch: @nickoferrall 
Even if we don't await publishing to the subscription, we can call it from the mutation
```ts
const updateGeneratedGroups = async () => {
  await generateGroups(...)
  publish(...)
}
// don't await because it takes some time
updateGeneratedGroups()
```
I think this would make the intend clearer.
Regardless, I think omitting the `mutatorId` is the right move given the comment
https://github.com/ParabolInc/parabol/blob/3b3d9dd8f6de6eb8bf66b8d37a4d53ca9c1fa882/packages/server/utils/publish.ts#L4

Issue #8125, opened 2023-04-27
Suggest Groups: handle "Suggest Groups" once grouping has already started


nickoferrall:
If the user has already started grouping reflections and clicks "Suggest Groups" we will need to let them know that they will lose their groups 

https://github.com/ParabolInc/parabol/issues/8123

### AC

- [ ] If a team has already begun grouping reflections and clicks "Suggest Groups", a model pops-up letting them know that their grouping progress will be lost
- [ ] The modal includes a primary action button that says "Continue" and a second action button that says "Cancel"
- [ ] Clicking "Cancel" closes the model
- [ ] Clicking "Continue" groups the reflections together

PR #8445, opened 2023-06-27
feat(suggest-groups): add reset groups button

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8124 & https://github.com/ParabolInc/parabol/issues/8125

Demo: https://www.loom.com/share/e9606cff9454417da79448424cb95217

### To test

- [ ] After clicking the "Suggest Groups" button, see the "Reset Groups" button
- [ ] Clicking it reverts the groups to the state they were in before clicking "Suggest Groups"


packages/server/graphql/types/RetrospectiveMeeting.ts
```diff
@@ -58,6 +58,10 @@ const RetrospectiveMeeting: GraphQLObjectType<any, GQLContext> = new GraphQLObje
       type: new GraphQLList(new GraphQLNonNull(AutogroupReflectionGroup)),
       description: 'The suggested reflection groups created by OpenAI'
     },
+    resetReflectionGroups: {
+      type: new GraphQLList(new GraphQLNonNull(AutogroupReflectionGroup)),
```
nickoferrall, line 62: Using the `AutogroupReflectionGroup` type for both `autogroupReflectionGroups` and `resetReflectionGroups` could be confusing. I'm open to other names

igorlesnenko (approved):

packages/client/mutations/AutogroupMutation.ts
```diff
@@ -7,6 +7,9 @@ graphql`
   fragment AutogroupMutation_meeting on AutogroupSuccess {
     meeting {
       id
+      resetReflectionGroups {
```
Dschoordsch, line 10: +1 As far as I can see it, the client needs this only to know it can reset the autogroups, so it could just be a boolean?

packages/server/graphql/public/mutations/resetReflectionGroups.ts
```diff
@@ -0,0 +1,80 @@
...
+
+  const addReflectionPromises = Object.entries(resetGroupsMapper)
+    .map(([newReflectionGroupId, {reflectionIds, groupTitle}]) => {
+      return reflectionIds.map((reflectionId) => {
+        return addReflectionToGroup(reflectionId, newReflectionGroupId, context, groupTitle)
+      })
+    })
+    .flat()
+
+  await Promise.all(addReflectionPromises)
```
Dschoordsch, line 67: +1 `await Promise.all` could be moved to l59 and make it more readable because it's clearer you're not using the promises for anything else.

Dschoordsch (approved):

packages/client/mutations/AutogroupMutation.ts
nickoferrall: It could be, but we'd have to add an additional field to the `RetrospectiveMeeting` object. We're using `resetReflectionGroups` to reset the reflections to their original groups, so I thought it'd be best to check whether it exists rather than add an additional field. 

I'm open to using a boolean though if you feel this would be inefficient on the client.

packages/client/components/RetroGroupPhase.tsx
```diff
@@ -100,24 +118,37 @@ const RetroGroupPhase = (props: Props) => {
...
+                    onMouseLeave={closeTooltip}
+                    className='ml-2 h-6 w-6 cursor-pointer text-slate-600'
+                    ref={originRef}
+                  >
+                    <InfoIcon />
+                  </div>
+                </ButtonWrapper>
+              ) : (
+                <ButtonWrapper>
+                  <StyledButton onClick={handleResetGroupsClick}>{'Reset Groups'}</StyledButton>
```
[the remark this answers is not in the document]
nickoferrall: I'd like to keep the button text as concise as possible, so I'd prefer to avoid the word `suggestion`. Perhaps `Undo Groups`?

PR #8446, opened 2023-06-27
chore: refactor postDeploy to preDeploy

### LINKED ISSUE #8334: Rename postDeploy to predeploy
AC
- [ ] rename postDeploy to preDeploy
- [ ] run preDeploy in predeploy for in app.json
- [ ] add README.md for all predeploy scripts

will CC @rafaelromcar-parabol in PR review

mattkrick:
# Description

fixes #8334 

- moves `pushToCDN` to `predeploy` so you no longer have to make an individual call to `pushToCDN` or worry about what order things get called in
- removes old postdeploy migrations (these were a mistake, since we can't guarantee the run order of pre/post deploy migrations)
- renames postDeploy to preDeploy. This script can be run at any time after the migrations have run and before the new version of the app gets deployed

## Demo

## Testing scenarios

- [ ] predeploy works in dev
- [ ] predeploy works in prod locally 
- [ ] predeploy works in CI
- [ ] predeploy works in production 🎉


rafaelromcar-parabol:
Request: documentation on a README, with the same idea of what you are doing for Chronos. Description of what the script does, env variables that are specific for it.

docker/parabol-ubi/docker-host-st/README.md
```diff
@@ -7,9 +7,11 @@ To run the Parabol UBI in single tenant mode (e.g. simple docker-compose on a do
...
 5. Check logs via command `docker logs <name>-app-1` and wait for the following output to appear
+
 ```shell
 🔥🔥🔥 Server ID: 0. Ready for Sockets: Port 3000 🔥🔥🔥
 💧💧💧 Server ID: 0. Ready for GraphQL Execution 💧💧💧
 💧💧💧 Server ID: 01. Ready for GraphQL Execution 💧💧💧
 ```
-6. Run command `docker exec <name>-app-1 yarn postdeploy`
+
+6. Run command `docker exec <name>-app-1 yarn predeploy`
```
rafaelromcar-parabol, line 17: This isn't correct. Step 6 should be removed and yarn predeploy added to the [docker-compose` file](https://github.com/ParabolInc/parabol/blob/1a3dcd45d0f71bee2a126e0c801506c3f950246e/docker/parabol-ubi/docker-host-st/docker-compose.yaml#L40).

One more note: this single tenant deployment do not work with our current pipeline Docker images. I will create an issue to adapt that, probably after the GCP migration.

mattkrick: done! I'm betting there's some other pieces that are broken, which is totally OK! We can fix those build steps later

rafaelromcar-parabol: Indeed. I create [this issue](https://github.com/ParabolInc/parabol/issues/8448) to track what we have to do and do it whenever we have the time

mattkrick:
@rafaelromcar-parabol mind taking another look? 
- predeploy section added to readme
- migrations called from predeploy & removed from docker calls
- backwards compatible so you can call migrate before predeploy if you like

rafaelromcar-parabol:
Looks fine! I think you should take this opportunity to generate a Docker image from the branch and test it using the on-demand environments. You can follow the doc [here](https://www.notion.so/parabol/How-to-setup-for-the-new-infrastructure-31f0ae97fcc5486e9fe5f2d8d8c70b70). In this particular case it would be a bit harder, we can work on it together for the last step, but I think we should do it before merging.

Let's meet sync this evening (for me) and do the whole thing (build the docker image, test the new process).

mattkrick:
TODO:
- Logs for each script that runs
- throw when CDN is undefined

PR #8449, opened 2023-06-28
chore: Replace TeamHealth answer emojis with their text representation

### LINKED ISSUE #8402: Should we store emojis as team health votes?
​

Dschoordsch:
# Description

Fixes #8402 
To avoid conversion complications between databases.

## Demo

no demo, looks same as before

## Testing scenarios

- on **master** create a retro with team health
- checkout **PR** and create a new retro
- verify voting, revealing, summary and email summary look fine for old and new retro

## Final checklist


Dschoordsch:
I'm not sure this is the way to go. This would add additional complication when we allow customers to customize the answer options. If we compare this with Poker scales, there we don't do such replacements either.

tianrunhe (commented):
First of all, thanks for jumping in & proposing this change! Here's my 2 cents:

1. We don't have problems storing emojis for poker voting results because... well, we don't intend to do analysis about them at the moment. Moreover, most of the voting are numerical already, so in case one day we would like to provide some stats on, say, average story points for your team, we can do that fine.
2. Team health however, inherently has the analysis attribute and there's clear definition of "better" and "worse". People would want to see if their team health improved/worsen in last quarter, for example. Storing emojis directly in this case doesn't provide any benefits, as I see it. Even if we tackle the tech difficulty and successfully import those emojis to BQ, we still need to map them into some numerical representation in order to conduct any analysis.

So... what do you think of actually storing a number for the vote? e.g., 😁 is 5, 😐 is 3, 😢 is 1. In the future we can allow users to customize the emojis, say 🥥, 🍊 and 🍒, that's fine but I would suppose we need to provide an interface for them to decide which one represent "good", "ok" and "bad" respectively. If that's the case, then at the backend we can still map them to 5,3,1. Does that make sense? WDYT?

Dschoordsch:
I don't like numerical score values as it implies I could just average them. But what if I had happy, sad, angry as answer options? Those are clearly not on the same scale, but valid answers.
I wouldn't mind just indexing the answers and for display purposes we'd just show whatever is stored at that index. For graphs we just treat it as if it's continuous.

tianrunhe:
Oh yeah that works! I ❤️ it.

igorlesnenko (changes requested):
I have something weird happening in this branch when switching between different emoji:

https://github.com/ParabolInc/parabol/assets/466991/d6bcedc6-4213-477b-84da-b02ce0e61f7d

Can you reproduce?

igorlesnenko (approved):

PR #8450, opened 2023-06-28
chore: add radix Dialog component

### LINKED ISSUE #8107: UI: Create generic `Dialog` component
Currently all our modals dialogs are different. 

For example new meeting and start recurrence have different close button implementations:

image

image

We also recreating some basic parts every time:

https://github.com/ParabolInc/parabol/blob/183f543b74b49f40d19eb1829b79e0c8d2a119a2/packages/client/components/TeamPrompt/Recurrence/UpdateRecurrenceSettingsModal.tsx#L40-L44

https://github.com/ParabolInc/parabol/blob/afa812a1fc18e6e8b12b8fde877b7f9c3ed1b2e3/packages/client/components/ReviewRequestToJoinOrgModal.tsx#L104-L110

Would be good to have a generic, responsive dialog component using `tailwind` and for example `radix-ui`.

igorlesnenko:
Fixes #8107

This is a version 1, I think we will improve it as needed when start reusing it in other places
I need this for sharing summaries (#8331)

https://github.com/ParabolInc/parabol/assets/466991/5b97168f-6532-49a2-bc7a-463294418a8e

**How to test:**

Follow testing steps from PR #8331


packages/client/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
```
BartoszJarocki, line 1: -1 can you move it to `client/ui/Dialog`? I'd love to separate, generic components from stuff related to our features. I did that here: [feat: added button component #8264](https://github.com/ParabolInc/parabol/pull/8264) so it'd be consistent

packages/client/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,29 @@
...
+import * as RadixDialog from '@radix-ui/react-dialog'
+
+interface DialogProps {
+  open: boolean
+  onClose: () => void
+  children: React.ReactNode
+}
+
+const Dialog = (props: DialogProps) => {
+  const {open, onClose, children} = props
```
BartoszJarocki, line 11: +1 in case we'd want to override some style, position, whatever, I'd expose `className` here. as overriding tailwind classnames isn't too easy in case there's already a related class used, I'd suggest using [dcastil/tailwind-merge: Merge Tailwind CSS classes without style conflicts](https://github.com/dcastil/tailwind-merge)

packages/client/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,29 @@
...
+  return (
+    <RadixDialog.Root
+      open={open}
+      onOpenChange={(newOpen) => {
+        if (!newOpen) {
+          onClose()
+        }
+      }}
```
BartoszJarocki, line 19: +1 as rule of thumb when creating a generic component, that could be reused in various places, we should allow to pass all the props, if the use case requires that. a good example of exposing prestyled, yet configurable generic components is [Dialog - shadcn/ui](https://ui.shadcn.com/docs/components/dialog) - see installation and `manual` tab, you'll see nicely typed interfaces that allow to override all sorts of things.

packages/client/components/Dialog/DialogActions.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+
+interface DialogActionsProps {
+  children: React.ReactNode
+}
+
+const DialogActions = (props: DialogActionsProps) => {
+  const {children} = props
+  return <div className='mt-[25px] flex justify-end space-x-2'>{children}</div>
```
BartoszJarocki, line 9: +1 it'd be great to _not_ use pixels, but `mt-4` etc. it uses `rem` which helps, when you for example, change the default text size in the browser (it'll scale).

packages/client/components/Dialog/DialogClose.tsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react'
+import {Close} from '@mui/icons-material'
+import * as RadixDialog from '@radix-ui/react-dialog'
+
+const DialogClose = () => {
+  return (
+    <RadixDialog.Close asChild>
+      <button
+        className='absolute top-[16px] right-[10px] inline-flex h-[25px] w-[25px] cursor-pointer appearance-none items-center justify-center bg-transparent'
```
BartoszJarocki, line 9: +1 same, please avoid pixels

packages/client/components/Dialog/DialogContent.tsx
```diff
@@ -0,0 +1,17 @@
...
+import * as RadixDialog from '@radix-ui/react-dialog'
+
+interface DialogContentProps {
+  children: React.ReactNode
+}
+
+const DialogContent = (props: DialogContentProps) => {
+  const {children} = props
+  return (
+    <RadixDialog.Content className='fixed top-[50%] left-[50%] max-h-[85vh] w-[90vw] max-w-[450px] translate-x-[-50%] translate-y-[-50%] rounded-lg bg-white p-[16px] shadow-[hsl(206_22%_7%_/_35%)_0px_10px_38px_-10px,_hsl(206_22%_7%_/_20%)_0px_10px_20px_-15px] focus:outline-none data-[state=open]:animate-contentShow'>
```
BartoszJarocki, line 11: +1 same, avoid pixels

packages/client/components/Dialog/DialogDescription.tsx
```diff
@@ -0,0 +1,17 @@
...
+import * as RadixDialog from '@radix-ui/react-dialog'
+
+interface DialogDescriptionProps {
+  children: React.ReactNode
+}
+
+const DialogDescription = (props: DialogDescriptionProps) => {
+  const {children} = props
+  return (
+    <RadixDialog.Description className='mt-[10px] mb-5 text-[15px] leading-normal'>
```
BartoszJarocki, line 11: +1 same, avoid pixels

packages/client/components/Dialog/DialogTitle.tsx
```diff
@@ -0,0 +1,13 @@
+import React from 'react'
+import * as RadixDialog from '@radix-ui/react-dialog'
+
+interface DialogTitleProps {
+  children: React.ReactNode
+}
+
+const DialogTitle = (props: DialogTitleProps) => {
+  const {children} = props
+  return <RadixDialog.Title className='m-0 text-[20px] font-semibold'>{children}</RadixDialog.Title>
```
BartoszJarocki, line 10: +1 no pixels please ;)

packages/client/components/ShareTopicRoot.tsx
```diff
@@ -14,26 +13,23 @@ const ShareTopicRoot = () => {
...
   }, [location])
 
-  const {openPortal, closePortal, modalPortal} = useModal({
-    id: 'shareTopicModal',
-    onClose
-  })
-
   useEffect(() => {
-    openPortal()
+    setOpen(true)
```
BartoszJarocki, line 24: +1 if I understand correctly, it opens the modal after initial render, why not use `true` by default[here](https://github.com/ParabolInc/parabol/pull/8450/files#diff-803c8f379080453c0db3ef534d257a76a3028246ee6cf1c14602cea5328871eaR16) instead?

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/ShareTopic.tsx
```diff
@@ -33,6 +32,8 @@ const style = {
 const ShareTopic = (props: Props) => {
   const {isDemo, isEmail, meetingId, stageId, appOrigin} = props
 
+  const [open, setOpen] = React.useState(false)
```
BartoszJarocki, line 35: +1 as this code will be used a lot, you can consider adding a little utility like 

```
export const useDialog = () => {
  const [isOpen, setIsOpen] = useState(false)

  const open = useCallback(() => setIsOpen(true), [])
  const close = useCallback(() => setIsOpen(false), [])

  return [isOpen, open, close] as const
```

you'll have stable refrences to helper method like open, close that are a little bit more readable than doing `setOpen(true`

packages/client/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,29 @@
...
+    >
+      <RadixDialog.Portal>
+        <RadixDialog.Overlay className='fixed inset-0 bg-slate-700/[.3] data-[state=open]:animate-overlayShow' />
+        {children}
+      </RadixDialog.Portal>
+    </RadixDialog.Root>
+  )
+}
+
+export default Dialog
```
BartoszJarocki, line 29: +1 I'd love to stop using default exports, they're harder to figure out for IDE and have other cons - [Why we have banned default exports in Javascript and you should do the same | by Kris Kaczor | Neufund](https://blog.neufund.org/why-we-have-banned-default-exports-and-you-should-do-the-same-d51fdc2cf2ad)

BartoszJarocki (changes requested):
there's only one thing that's blocking as we should start cleaning things out and put design system component into a different directory. other than that, I left a bunch of suggestions for your consideration. I personally would love to see them implemented, but they're not blockers. thank you for working on that!

packages/client/components/Dialog/Dialog.tsx
igorlesnenko: @BartoszJarocki I see they store everything in one file, do you think it make sense in our case?

BartoszJarocki: no strong opinions, I think the general consensus is to have one main export from one file, let's keep them separate

packages/client/ui/Dialog/SimpleModalDialog.tsx
```diff
@@ -0,0 +1,32 @@
...
+interface DialogProps {
+  isOpen: boolean
+  onClose: () => void
+  children: React.ReactNode
+  rootProps?: React.ComponentPropsWithoutRef<typeof RadixDialog.Root>
+  overlayProps?: React.ComponentPropsWithoutRef<typeof DialogOverlay>
+  portalProps?: React.ComponentPropsWithoutRef<typeof RadixDialog.Portal>
+}
+
+export const SimpleModalDialog = (props: DialogProps) => {
```
BartoszJarocki, line 14: +1 this could be removed if you take my suggestion from previous comment. and with that in mind, there'll be no need for props like `rootProps`, `overlayProps`, `portalProps` and these components will be directly accessible using composition

packages/client/ui/Dialog/useDialog.tsx
```diff
@@ -0,0 +1,10 @@
+import {useCallback, useState} from 'react'
+
+export const useDialog = () => {
```
BartoszJarocki, line 3: +1 I know I suggested it but maybe the better name would be `useDialogState`?

packages/client/ui/Dialog/DialogContent.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react'
+import * as RadixDialog from '@radix-ui/react-dialog'
+import {twMerge} from 'tailwind-merge'
+
+export const DialogContent = React.forwardRef<
+  HTMLDivElement,
```
BartoszJarocki, line 6: +1 this is technically not correct, as if someone uses `asChild` it could render a different html element  - we should use `React.ElementRef<typeof DialogContent.Title>`. not sure if this type is available in React 17 though, not a blocker though

BartoszJarocki (approved):
Left one comment on how I would improve the ability to compose and create new dialogs, otherwise looks good to me!

BartoszJarocki:
@igorlesnenko I haven't tested the impl, have you checked if all accessibility features are working, for example if you can close the dialog with hitting `esc`? we don't use ` <Dialog.Trigger />` and I just want to make sure we don't loose any features we actually want to have handled when using external lib

packages/client/ui/Dialog/DialogContent.tsx
igorlesnenko: There is no such type `ElementRef`, I wasn't sure what to put here instead

igorlesnenko:
@BartoszJarocki accessibility seems to be working fine

packages/client/ui/Dialog/DialogContent.tsx
```diff
@@ -0,0 +1,26 @@
...
+  React.ComponentPropsWithoutRef<typeof RadixDialog.Content>
+>(({className, children, ...props}, ref) => (
+  <DialogPortal>
+    <DialogOverlay />
+    <RadixDialog.Content
+      ref={ref}
+      className={twMerge(
+        'fixed top-[50%] left-[50%] max-h-[85vh] w-[90vw] max-w-md translate-x-[-50%] translate-y-[-50%] rounded-lg bg-white p-4 shadow-[hsl(206_22%_7%_/_35%)_0px_10px_38px_-10px,_hsl(206_22%_7%_/_20%)_0px_10px_20px_-15px] focus:outline-none data-[state=open]:animate-contentShow',
```
mattkrick, line 17: +1 I see that this is copied from radix-ui, which is great for examples, but this is really hard to read. Since we reuse the same box shadows everywhere (https://github.com/ParabolInc/parabol/blob/9ef744ef2d76878f1cec88b6967cb616ff2a1c13/packages/client/styles/elevation.ts) could we create a custom tailwinds shadow so this becomes something like `shadow-dialog` or `shadow-z3`?

packages/client/tailwindTheme.ts
```diff
@@ -154,6 +154,20 @@ export default {
...
         '900': '#1C1C21'
       }
+    },
+    keyframes: {
+      overlayShow: {
+        from: {opacity: 0},
+        to: {opacity: 1}
+      },
+      contentShow: {
+        from: {opacity: 0, transform: 'translate(-50%, -48%) scale(0.96)'},
```
mattkrick, line 164: +1 i like subtle animations but 2% translation on a dialog that is 150px tall is.... 3px 😄 . Too little!

A fun css rule is that you can pass in 2 translate functions & they'll be applied sequentially. that way you can keep your % based translate & then use one with pixels.

Try this:

```ts
contentShow: {
  from: {opacity: 0, transform: 'translate(-50%, -50%) translateY(32px) scale(0.96)'}
}
```

Since the animation finishes at the same position as the css class, no `to` is needed 🎉

mattkrick (changes requested):
love it! can't wait to get framer motion in here so we can add some exit animations, too!

packages/client/ui/Dialog/DialogContent.tsx
igorlesnenko: I see we have modal shadow defined in this file, will copy that in a custom `shadow-dialog`

mattkrick:
another radix component pattern complete! 🎉

Issue #8447, opened 2023-06-28
Suggest Groups: add metrics


nickoferrall:
We should track a few things to help us determine whether this is a useful feature. Two useful metrics that would give us an indication of whether it's a useful feature:

1. % of meetings where clicking Suggest Groups was not proceeded by Reset Groups. If they're always resetting, it's not useful.
2. % of meetings that reach the Group phase where a team member clicked Suggest Groups. If they're afraid to click it or prefer manually grouping, it's not helpful.

### AC

Track the following events:

- [ ] A meeting proceeds to the Group phase, and the Suggest Groups button is visible
- [ ] The Suggest Groups button has been clicked
- [ ] The Reset Groups button has been clicked

PR #8459, opened 2023-06-29
chore(suggest-groups): add suggest groups metrics

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8447

### To test

- [ ] Sends an event to Segment when clicking the suggest groups and reset groups button
- [ ] Sends an event to Segment when generating groups

### DIFF
--- a/packages/server/graphql/mutations/helpers/generateGroups.ts
+++ b/packages/server/graphql/mutations/helpers/generateGroups.ts
@@ -5,6 +5,7 @@ import {DataLoaderWorker} from '../../graphql'
 import {AutogroupReflectionGroupType} from '../../../database/types/MeetingRetrospective'
 import {SubscriptionChannel} from '../../../../client/types/constEnums'
 import publish from '../../../utils/publish'
+import {analytics} from '../../../utils/analytics/analytics'
 
 const SERVER_ID = process.env.SERVER_ID
 
@@ -56,10 +57,20 @@ const generateGroups = async (
   }
 
   const r = await getRethink()
-  await r.table('NewMeeting').get(meetingId).update({autogroupReflectionGroups}).run()
+  const meetingRes = await r
+    .table('NewMeeting')
+    .get(meetingId)
+    .update(
+      {autogroupReflectionGroups},
+      {returnChanges: true}
+    )('changes')(0)('new_val')
+    .run()
+  const {facilitatorUserId} = meetingRes
   const data = {meetingId}
   const operationId = dataLoader.share()
   const subOptions = {operationId, mutatorId: SERVER_ID}
+
+  analytics.suggestGroupsGenerated(facilitatorUserId, meetingId, teamId)
   publish(SubscriptionChannel.MEETING, meetingId, 'GenerateGroupsSuccess', data, subOptions)
   return autogroupReflectionGroups
 }
--- a/packages/server/graphql/mutations/helpers/generateGroups.ts
+++ b/packages/server/graphql/mutations/helpers/generateGroups.ts
@@ -69,7 +69,6 @@ const generateGroups = async (
   const data = {meetingId}
   const operationId = dataLoader.share()
   const subOptions = {operationId, mutatorId: SERVER_ID}
-
   analytics.suggestGroupsGenerated(facilitatorUserId, meetingId, teamId)
   publish(SubscriptionChannel.MEETING, meetingId, 'GenerateGroupsSuccess', data, subOptions)
   return autogroupReflectionGroups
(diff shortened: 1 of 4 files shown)


packages/server/graphql/mutations/helpers/generateGroups.ts
```diff
@@ -56,10 +57,19 @@ const generateGroups = async (
...
+    .update(
+      {autogroupReflectionGroups},
+      {returnChanges: true}
+    )('changes')(0)('new_val')
+    .run()
+  const {facilitatorUserId} = meetingRes
   const data = {meetingId}
   const operationId = dataLoader.share()
   const subOptions = {operationId, mutatorId: SERVER_ID}
+  analytics.suggestGroupsGenerated(facilitatorUserId, meetingId, teamId)
```
nickoferrall, line 72: We require a userId in the `track` function, but in this case, there isn't really a userId. The meeting proceeds from the Reflect phase to the Group phase, and this function is triggered.

I've included the facilitator user id as I think that's the most accurate user id

packages/server/utils/analytics/analytics.ts
```diff
@@ -109,6 +109,10 @@ export type AnalyticsEvent =
   | 'Snackbar Viewed'
   // Join request
   | 'Join Request Reviewed'
+  // Suggest Groups
+  | 'Suggest Groups Generated'
```
tianrunhe, line 113: +1: Suggest**ed** Groups Generated?

packages/server/utils/analytics/analytics.ts
```diff
@@ -109,6 +109,10 @@ export type AnalyticsEvent =
...
+  | 'Suggest Groups Clicked'
+  | 'Reset Groups Clicked'
```
tianrunhe, line 115: Should this two metrics be emitted from client side on the actual buttons? I didn't look at the mutations carefully but could there be cases when user clicked the button but the mutation errors out before reaching the metrics line?

tianrunhe (changes requested):

packages/server/utils/analytics/analytics.ts
nickoferrall: "Suggest Groups" is the noun; it's the name of the button. Generate is the verb, so that's in the past tense. I think it's the same as "Join Request Reviewed" - we wouldn't say "Join Requested Reviewed"

packages/server/utils/analytics/analytics.ts
nickoferrall: I thought the preference was to include it on the server because it means less code, and if the client loses their internet connection, the analytics will still be recorded? 

If there's an error in the mutation before we get to the analytics bit, the mutation wouldn't have executed, so would we want to record it? We'd know the error happened as we're recording that in Sentry

packages/server/utils/analytics/analytics.ts
tianrunhe: But it's not the button that was generated, right? It's the groups that were generated and the groups are "suggested groups" (from ChatGPT). That's how I interpret at least. It's a +1 comment though so it's up to you!

nickoferrall: Hmm yeah, I think you're right! Updated 👍

packages/server/utils/analytics/analytics.ts
tianrunhe: I still think it's weird to see the metrics emitted from server side when those actions are essentially client side behaviors. Reading through the code though I realize when user clicks the button, inside the mutation it will execute to that line even when the groups were not generated previously so I guess it's essentially the same today? But the code still smells to me. I'll let @Dschoordsch do the arbitration 😄

tianrunhe (approved):
Approving conditionally, see [the comment](https://github.com/ParabolInc/parabol/pull/8459#discussion_r1251139114)

tianrunhe:
Sorry just realized you have one review required label and Georg is on PTO. Since this is still a internal test, with the spirit of releasing & testing fast, I'm fine merging it!

packages/server/utils/analytics/analytics.ts
nickoferrall: `analytics` is used in server mutations that are triggered by client events in many places throughout the app, such as `createTask`, `startRetrospective`, `addComment`, `addAtlassianAuth`, etc. Do you feel that this is different to those examples?

Dschoordsch: @tianrunhe thanks for tagging me, but I think ultimately you're in control of the analytics and are free to define what you want to track, i.e. if it's more important the user clicked a button vs. the user performed the grouping.
In general I expect server side analytics to be more consistent because of ad-blockers etc. So I'm in favour of the current solution, but it's up to you as I said.
The only odd one I see is `Suggested Groups Generated` because that just happens when advancing the meeting, but I guess it's important to filter the funnel generated->grouping clicked->groups reset

PR #8460, opened 2023-06-29
chore: add radix Select component

igorlesnenko:
Added generic Select component based on tailwind and radix

I need this for sharing summaries (https://github.com/ParabolInc/parabol/pull/8331)

https://github.com/ParabolInc/parabol/assets/466991/f65c1338-9c2b-4918-87aa-72cccc1e8259

**How to test:**

Follow testing steps from PR https://github.com/ParabolInc/parabol/pull/8331


packages/client/ui/Select/SelectContent.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+export const SelectContent = React.forwardRef<
+  HTMLDivElement,
+  React.ComponentPropsWithoutRef<typeof RadixSelect.Content>
+>(({className, children, position = 'popper', ...props}, ref) => (
+  <RadixSelect.Portal>
+    <RadixSelect.Content
+      ref={ref}
+      className={twMerge(
+        'relative z-50 min-w-[8rem] overflow-hidden rounded bg-white shadow-card-1',
```
BartoszJarocki, line 13: +1 i'd remove an assumption about the width from here and just provide the value in the place where this component is used, that way you'll probably be able to avoid using min-w

packages/client/components/ShareTopicModal.tsx
```diff
@@ -20,12 +29,32 @@ const ShareTopicModal = (props: Props) => {
...
 
+  const labelStyles = `w-[90px] text-left text-sm font-semibold`
+
   return (
     <Dialog isOpen={isOpen} onClose={onClose}>
       <DialogContent>
         <DialogTitle>Share topic</DialogTitle>
         <DialogDescription>Where would you like to share the topic to?</DialogDescription>
 
+        <fieldset className='mx-0 mb-[15px] mb-2 flex items-center gap-5 p-0'>
```
BartoszJarocki, line 40: mb-4? let's avoid px and hardcoded stuff

BartoszJarocki (approved):

packages/client/ui/Select/index.tsx
```diff
@@ -0,0 +1,6 @@
+export {Select} from './Select'
+export {SelectContent} from './SelectContent'
+export {SelectGroup} from './SelectGroup'
+export {SelectItem} from './SelectItem'
+export {SelectTrigger} from './SelectTrigger'
+export {SelectValue} from './SelectValue'
```
mattkrick, line 6: +1 IMO index files just mean 1 more piece of boilerplate to keep updated when we add/remove components, but if everyone likes the pattern I don't mind that much. The index AND additional file that just renames the radix export seem a bit much though. Maintaining 2 additional files that creates 3 sources of truth means someone could import the same thing 3 ways:

```ts
import {SelectValue} from '../ui/Select'
import {SelectValue} from '../ui/Select/SelectValue'
import {Value as SelectValue} from '@radix-ui/react-select'
```

IMO we should have a single source of truth & just import from radix, but I'm open to your opinion as well as @BartoszJarocki as we nail down the right radix-ui patterns for us.

if you wanted to reduce the 3 sources of truth to 2, you could also do

```ts
// index.tsx
export {Value as SelectValue} from '@radix-ui/react-select'
```

BartoszJarocki: no idea how I missed that file! I'm not w fan of index files as @mattkrick said, I think it's just another place to maintain.

I like the idea of reexporting/renaming radix component that's a part of a bigger solution, just because there's a pretty big chance we'll end up customizing it, as well as for naming consistency reason. If the component is only used internally by some kind of abstraction that we introduce, no need to rename/reexport it. that's just my 2c.

igorlesnenko: @mattkrick For some reason I wanted to make imports less noisy and also thought about this `ui` folder as our UI library which potentially could be a separate package.

No strong preference here, I changed imports back to `import {SelectValue} from '../ui/Select/SelectValue'`

However I would agree with @BartoszJarocki and prefer to reexport some of the components because we may customize it later

mattkrick: works for me!

PR #8451, opened 2023-06-28
fix: Fix optimistic update for create task in Sprint Poker

Dschoordsch:
# Description

Fixes #8094 
Optimistic update was disabled because it threw an error. This happened because one of the fields was undefined while it should have been null.

## Demo

https://www.loom.com/share/8500771c4b3048eaa45724ae553e4350?sid=599ca38f-fc7b-4b06-a775-60bd05e12935

## Testing scenarios

- start a Sprint Poker meeting
- in scoping create a Parabol task

## Final checklist

- [ ] I checked the [code review guidelines](../docs/codeReview.md)
- [ ] I have added [Metrics Representative](../docs/codeReview.md#metrics-representative) as reviewer(s) if my PR invovles metrics/data/analytics related changes
- [ ] I have performed a self-review of my code, the same way I'd do it for any other team member
- [ ] I have tested all cases I listed in the testing scenarios and I haven't found any issues or regressions
- [ ] Whenever I took a non-obvious choice I added a comment explaining why I did it this way
- [ ] I added the label `One Review Required` if the PR introduces only minor changes, does not contain any architectural changes or does not introduce any new patterns and I think one review is sufficient'
- [ ] PR title is human readable and could be used in changelog

PR #8472, opened 2023-07-03
fix: can create tasks

nickoferrall:
Reverts https://github.com/ParabolInc/parabol/pull/8451/files

### To test

- On master, try to create a task and see the error message
- Checkout this branch, and successfully create a task

### DIFF
--- a/packages/client/mutations/CreateTaskMutation.ts
+++ b/packages/client/mutations/CreateTaskMutation.ts
@@ -180,7 +180,6 @@ const CreateTaskMutation: StandardMutation<TCreateTaskMutation, OptionalHandlers
         .setLinkedRecord(userId ? store.get(userId)! : null, 'user')
         .setLinkedRecord(viewer, 'createdByUser')
         .setLinkedRecords([], 'replies')
-        .setLinkedRecord(null, 'integration')
       if (integration) {
         const {service, serviceProjectHash} = integration
         if (service === 'jira') {


packages/client/mutations/CreateTaskMutation.ts
```diff
@@ -180,7 +180,6 @@ const CreateTaskMutation: StandardMutation<TCreateTaskMutation, OptionalHandlers
         .setLinkedRecord(userId ? store.get(userId)! : null, 'user')
         .setLinkedRecord(viewer, 'createdByUser')
         .setLinkedRecords([], 'replies')
-        .setLinkedRecord(null, 'integration')
```
igorlesnenko, line 183: I see we revert previous PR partially should we revert all changes?

igorlesnenko (changes requested):
Creating task does not work for me, app crashes with an error

image

igorlesnenko (approved):
Ah I see, we are speaking about tasks in discussion.

nickoferrall:
> we are speaking about tasks in discussion

With this PR, I can create tasks in the task board and in the discussion thread, whereas on master, I can't create tasks either way. Is that the same experience you have?

igorlesnenko:
@nickoferrall yes, I can't create tasks in master

Dschoordsch:
@nickoferrall Reverting the fix partially reintroduced the bug in Sprint Poker :(

Issue #8452, opened 2023-06-28
Checkout Flow Tech Debt: retry invoice after updating payment method


nickoferrall:
See https://github.com/ParabolInc/parabol/pull/8423#discussion_r1243969934 and https://github.com/ParabolInc/parabol/pull/8285

### AC

- [ ] After updating the payment method, try to charge any invoices that previously failed

igorlesnenko:
@nickoferrall I'm not sure if it should be in v2.
The reason we added this is that sometimes the next payment for a new month does not go through for various reasons, then we lock the teams and offer to update the card so that the payment goes through and the teams are unblocked. Otherwise teams will be still locked

igorlesnenko:
@nickoferrall I think we ideally want to downgrade, I created an issue about this improvement: #8341

PR #8478, opened 2023-07-05
feat: retry invoice after failed payment

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8452

Demo: https://www.loom.com/share/8095302e24ec45ddb4c6df42009eab00

### To test

We need to get a failed invoice and then run `updateCreditCard` to charge the failed invoice. This is a hacky way of testing this; I'm open to any better ways. 

- [ ] Use the old checkout flow by updating this line: https://github.com/ParabolInc/parabol/blob/ef51c26e331f693fe993b9ead2433445c11b6cd2/packages/client/modules/userDashboard/components/Organization/Organization.tsx#L130. I'm doing this because the new checkout flow has a 3D Secure check on the client which shows an error when using the `Decline after attaching` Stripe card
- [ ] Comment out the `terminateSubscription` line in `stripeFailPayment` [here](https://github.com/ParabolInc/parabol/blob/e90eb2278e0e87a9149e20e3192c356c5cb4050a/packages/server/graphql/private/mutations/stripeFailPayment.ts#L77). We need to do this because when the first payment fails, we terminate the subscription and remove `stripeSubscriptionId` from the Organization object, which is needed to find the invoices. This wouldn't happen after the first payment has occurred - the status would be `past_due`. 
- [ ] Use the decline after attaching Stripe [card](https://stripe.com/docs/testing?testing-method=card-numbers#declined-payments) `4000000000000341`
- [ ] Switch to the new checkout flow, updating this line: https://github.com/ParabolInc/parabol/blob/ef51c26e331f693fe993b9ead2433445c11b6cd2/packages/client/modules/userDashboard/components/Organization/Organization.tsx#L130
- [ ] Enter a valid credit card and see that the failed payment has now been successfully charged
- [ ] You'll see a `Organization is not on the starter tier` error because the payment has triggered `upgradeToTeamTier`. I have another issue open to change the webhook that we listen to to remove this code smell.

### DIFF
--- a/packages/server/graphql/private/mutations/upgradeToTeamTier.ts
+++ b/packages/server/graphql/private/mutations/upgradeToTeamTier.ts
@@ -65,52 +65,34 @@ const upgradeToTeamTier: MutationResolvers['upgradeToTeamTier'] = async (
   }
 
   // RESOLUTION
-  await Promise.all([
-    r({
-      updatedOrg: r
-        .table('Organization')
-        .get(orgId)
-        .update({
-          creditCard: await getCCFromCustomer(customer),
-          tier: 'team',
-          tierLimitExceededAt: null,
-          scheduledLockAt: null,
-          lockedAt: null,
-          updatedAt: now
-        })
-    }).run(),
-    updateTeamByOrgId(
-      {
-        isPaid: true,
-        tier: 'team'
-      },
-      orgId
-    )
-  ])
+  await Promise.all(
+    [
+      r({
+        updatedOrg: r
+          .table('Organization')
+          .get(orgId)
+          .update({
+            creditCard: await getCCFromCustomer(customer),
+            tier: 'team',
+            tierLimitExceededAt: null,
+            scheduledLockAt: null,
+            lockedAt: null,
+            updatedAt: now
+          })
+      }).run(),
+      updateTeamByOrgId(
+        {
+          isPaid: true,
+          tier: 'team'
+        },
+        orgId
+      )
+    ],
+    removeTeamsLimitObjects(orgId, dataLoader)
+  )
   organization.tier = 'team'
 
-  // If subscription already exists and has open invoices, try to process them
-  const openInvoices = (await manager.listSubscriptionOpenInvoices(stripeSubscriptionId)).data
-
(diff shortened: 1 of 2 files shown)


packages/server/graphql/private/mutations/upgradeToTeamTier.ts
```diff
@@ -85,12 +85,32 @@ const upgradeToTeamTier: MutationResolvers['upgradeToTeamTier'] = async (
...
       orgId
-    ),
-    removeTeamsLimitObjects(orgId, dataLoader)
+    )
   ])
   organization.tier = 'team'
 
-  await Promise.all([setUserTierForOrgId(orgId), setTierForOrgUsers(orgId)])
+  // If subscription already exists and has open invoices, try to process them
+  const openInvoices = (await manager.listSubscriptionOpenInvoices(stripeSubscriptionId)).data
```
igorlesnenko, line 93: Here we added this to upgrade, but we need to do this on card details update `updateCreditCard`
So for example card is expired, we were unable to charge it and locked the org.
User provided a new card by updating details in org settings, we need to retry invoices

igorlesnenko (changes requested):

packages/server/graphql/private/mutations/upgradeToTeamTier.ts
nickoferrall: Makes sense. Should we still include this in `upgradeToTeamTier`? I can see you added it to `oldUpgradeToTeamTier`?

igorlesnenko: @nickoferrall I think we don't need to include it in `upgradeToTeamTier`. It was added to `oldUpgradeToTeamTier` because a single function was responsible for both update details and upgrade

igorlesnenko (approved):
Looks good

Dschoordsch (changes requested):
Please provide some testing steps

nickoferrall:
Sorry @Dschoordsch, I've added testing steps and a Loom

Dschoordsch (approved):

Issue #8481, opened 2023-07-05
Invite all users team member when creating a team


nickoferrall:
See context here: https://parabol.slack.com/archives/C042E8PGPD4/p1688573157250319?thread_ts=1687802102.137759&cid=C042E8PGPD4

### AC

- [ ] When creating a new team, we're now adding a text field where users can add email addresses. Below that text field, there's a checkbox with a label saying invite all team members
- [ ] After clicking the checkbox, all email addresses from within the org are added to the invite text field

PR #8482, opened 2023-07-05
feat: add invite all checkbox

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8481

Demo: https://www.loom.com/share/fc61572269c5406489c6d692cd514df7

### To test

- [ ] Click the checkbox and see it adds all users
- [ ] Deselect it and see that the newly added users are removed
- [ ] Click "Add Organization" and see that the checkbox & label are removed
- [ ] If there are no team members other than the viewer in the org, the checkbox & label aren't visible
- [ ] The viewer is not added to the text field when clicking the checkbox


igorlesnenko (approved):

Dschoordsch (changes requested):
I like the idea and it works, but I don't think a checkbox is the best interaction model. There are multiple edge cases which are just confusing:
* if I add an org user manually, then click the checkbox and then uncheck, they will be removed
* if I add all org users and then manually remove some, the checkbox stays checked
* if I add all org users manually, the checkbox is not checked

It just doesn't represent the state properly. I would like to see a button (text link?) instead which just fills in all org members and we get rid of the uncheck logic. The user can manually remove the emails they don't want. I think this messes less with user expectation.

Dschoordsch:
> I also think users could be concerned that a button might immediately invite users, whereas it's clear that a checkbox can be undone.
I think for a button a better copy would be "Add team members from <Team>" would clear this up.

acressall:
Chiming in on button vs. checkbox:

I think @Dschoordsch makes a good point about state clarity, but I do think it's relatively low risk. We _could_ hide the checkbox after it's been clicked, to avoid the discrepancies, but I don't think it's entirely necessary.

The alternative of the button I think creates some confusion before the user has a chance to see what it does. Buttons come with more risk of "what will happen when I click this?" where a checkbox can be easily undone. All that to say, I personally prefer the checkbox

Issue #8484, opened 2023-07-06
Fix drawer showing incorrect tier immediately after upgrading


nickoferrall:
@tianrunhe discovered a bug where the right drawer shows the Starter tier info immediately after upgrading. See Slack thread [here](https://parabol.slack.com/archives/C042E8PGPD4/p1688588114295889).

### AC

- [ ] Right drawer shows correct tier info after upgrading

PR #8485, opened 2023-07-06
fix: show correct drawer info after upgrade

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8484

Loom demo: https://www.loom.com/share/c6099d36128a4cd593cd1a49126612b7

The problem in master:
- To handle 3D Secure cards, we have to separate the Stripe subscription logic from the upgrade logic. This is because the client need to authenticate the 3D Secure payment after the subscription has been created.
- After the subscription has been created, Stripe webhooks triggers `upgradeToTeamTier` which upgrades the org 
- After `createStripeSubscription` has complete, we show the right draw and confetti before `upgradeToTeamTier` has finished to give the user faster feedback (it's already pretty slow)
- When the right drawer initially shows, it shows the Starter tier info, not the Team tier info

Solution:
- Add a `upgradeToTeamTierSuccessUpdater` to the `OrganizationSubscription`
- In the updater, set `showDrawer` and `showConfetti` to true
- Wait until `upgradeToTeamTier` has been executed before updating the UI

Downside:
- We're now waiting for the `upgradeToTeamTier` subscription before updating the UI. This can be a long wait
- I've got an issue to give the user feedback while they're upgrading: https://github.com/ParabolInc/parabol/issues/8497

### To test

- [ ] Run `stripe listen --forward-to localhost:3000/stripe` to listen to webhooks
- [ ] Upgrade to the Team tier using a 3D Secure card (see [here](https://stripe.com/docs/payments/3d-secure?locale=en-GB#three-ds-cards)), and see that the right drawer and the confetti show after the user has upgraded
- [ ] Use a 3D Secure card and fail the authentication. See that the right drawer and confetti doesn't show
- [ ] Use a non-3D Secure card and see the expected behaviour

### DIFF
--- a/packages/client/modules/userDashboard/components/OrgBilling/BillingForm.tsx
+++ b/packages/client/modules/userDashboard/components/OrgBilling/BillingForm.tsx
@@ -134,6 +134,7 @@ const BillingForm = (props: Props) => {
       commitLocalUpdate(atmosphere, (store) => {
         const org = store.get(orgId)
         if (!org) return
+        // stripe webhooks will trigger upgradeToTeamTier which will update the tier, but we want to show the confetti and correct drawer info immediately
         org.setValue('team', 'tier')
         org.setValue(true, 'showConfetti')
         org.setValue(true, 'showDrawer')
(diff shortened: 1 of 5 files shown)


packages/client/modules/userDashboard/components/OrgBilling/BillingForm.tsx
```diff
@@ -135,17 +125,11 @@ const BillingForm = (props: Props) => {
         return
       }
       const {error} = await stripe.confirmCardPayment(stripeSubscriptionClientSecret)
-      setIsLoading(false)
       if (error) {
         setErrorMsg(error.message)
+        setIsLoading(false)
```
nickoferrall, line 130: The component changes if the user successfully upgrades, so we only need to set `isLoading` to false if it fails

igorlesnenko (approved):
Works!

Dschoordsch (approved):

Issue #8216, opened 2023-05-17
Bug: running `moveTeamToOrg` fails for a big list of teamIds


Dschoordsch:
Running the mutation `moveTeamToOrg` with roughly 50 `teamIds` or more will fail on production because of timeouts. Splitting it into smaller chunks works. We should handle this case in code.

Dschoordsch:
My thinking was that this doesn't seem like a very expensive operation and I would expect our infrastructure to handle it. I forgot to mention that the same works on staging and seems to have a smaller impact overall. I'm concerned about our rethinkdb health.
How the issue is formulated however, I think you're right and this shouldn't be a P2.

PR #8486, opened 2023-07-06
fix: Limit number of teams that can be moved in one moveTeamToOrg call

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8216

Set a hard limit on the `moveTeamToOrg` mutation to short-circuit to an error when the mutation is likely to time out.

This doesn't solve the underlying issue, but since this is an internal-only mutation, these guardrails should be sufficient. Once this lands, we could optionally convert the GH issue from a "bug" to "enhancement" (or create a new issue), with the AC being to solve the underlying issue and remove this guardrail, but such a task would probably be so low priority it'd likely be marked as "Closed - won't do" pretty quickly.

## Demo
Screen Shot 2023-07-06 at 4 12 45 PM

## Final checklist


Dschoordsch (approved):
Did not test

PR #8487, opened 2023-07-06
chore: bump pg to v12.15

mattkrick:
# Description

pg v12 has some vulns when version is < 12.15.
this bumps pg to 12.15 for all deploys.

@adaniels-parabol can you confirm if 12.15 works for gov clients?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -167,7 +167,7 @@ jobs:
     docker:
       - image: cimg/node:18.16.0
       - image: rethinkdb:2.4.0
-      - image: cimg/postgres:12.9 # TODO: update to 12.10 when https://github.com/CircleCI-Public/cimg-postgres/pull/42 is merged
+      - image: cimg/postgres:12.15
         environment:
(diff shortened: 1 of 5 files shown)


rafaelromcar-parabol:
Please be aware that in GCP we do not control the exact version of PG and for now they are in 12.14. They do automatic upgrades whenever the next version is available and tested, but for now, all PostgreSQL instances in major version 12 are in 12.14.

mattkrick:
would it be better to pin to the major version instead with a `postgres:12`?

rafaelromcar-parabol:
I'm sorry because I missed this. I'm not sure, honestly. Pros and cons in both options. Let's keep it explicit, as you are doing right now, but trying to follow what we have in GCP from now on.

We don't know when 12.15 will be available in GCP. [Here](https://cloud.google.com/sql/docs/postgres/db-versions#major_and_minor_versions_supported_by) is the list of available versions right now. Let's hope 12.15 will be available soon, and also let's hope we don't need to upgrade our app ahead of GCP too many times.

PR #8499, opened 2023-07-10
chore: replace xml-lint due to mem leak

### LINKED ISSUE #8498: Memory leak: xml lint
## Issue - Bug

Our XML validator has a memory leak in it: https://github.com/authenio/samlify-node-xmllint/issues/6
That means every time someone logs on via SSO we leak another 16MB 🤣 .

So, we can use a validator that requires java, python, or node native. if we use the node native one, then it has a giant memory leak.

As an alternative, I'm going to wrap the most popular node-native validator to work with samlify & see if that works 🤞 

Effort: 6 hours

mattkrick:
# Description

fix #8498

## Demo

https://www.loom.com/share/8bb1ace6d4da4c3091daee1ef33046f8

## Testing scenarios

testing performed in demo, probably not necessary to test it yourself, but if you'd like to...

- [ ] create a dev okta account & dev parabol app. locally, call `enableSAMLForDomain` with the dev okta's metadata & your parabol.co email
- [ ] verify that logging in works

To test that the memory leak is gone:

- [ ] login a few times & then run `dumpHeap` in /admin/graphql
- [ ] analyze the heap & notice that there isn't a huge array of 16MB JSArrayBuffers being held onto

### DIFF
--- a/packages/server/graphql/private/mutations/loginSAML.ts
+++ b/packages/server/graphql/private/mutations/loginSAML.ts
@@ -1,4 +1,3 @@
-import * as validator from '@authenio/samlify-node-xmllint'
 import base64url from 'base64url'
 import getSSODomainFromEmail from 'parabol-client/utils/getSSODomainFromEmail'
 import querystring from 'querystring'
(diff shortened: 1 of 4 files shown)


igorlesnenko (approved):
Great investigation!

PR #8501, opened 2023-07-10
chore: drop unused RethinkDB tables

mattkrick:
# Description

Housekeeping for RethinkDB: removing unused tables


mattkrick:
no review needed as it's really just dropping some old tables. will test in staging today!

PR #8504, opened 2023-07-12
chore: add RetroReflectionGroup table to PG (Part 1 of 3)

mattkrick:
# Description

The Reflection table is the largest table in our DB. If we can move that to PG before we move to k8s we can reduce our downtime by about an hour 🤯 

Before I attempt to migrate our largest table, I want to migrate a smaller one & have the process go flawlessly. So, I chose this one.

## Testing scenarios

- [ ] run a retro
- [ ] change a reflection group title

### DIFF
--- a/packages/server/postgres/migrations/1689183718138_RetroReflectionGroup-part1-write.ts
+++ b/packages/server/postgres/migrations/1689183718138_RetroReflectionGroup-part1-write.ts
@@ -1,0 +1,35 @@
+import {Client} from 'pg'
+import getPgConfig from '../getPgConfig'
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+  CREATE TABLE IF NOT EXISTS "RetroReflectionGroup" (
+    "id" VARCHAR(100) PRIMARY KEY,
+    "createdAt" TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    "updatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+    "isActive" BOOLEAN NOT NULL DEFAULT TRUE,
+    "meetingId" VARCHAR(100) NOT NULL,
+    "promptId" VARCHAR(100) NOT NULL,
+    "sortOrder" INT NOT NULL DEFAULT 0,
+    "voterIds" VARCHAR(100)[] NOT NULL DEFAULT '{}',
+    "smartTitle" VARCHAR(255),
+    "title" VARCHAR(255)
+  );
+  CREATE INDEX IF NOT EXISTS "idx_RetroReflectionGroup_meetingId" ON "RetroReflectionGroup"("meetingId");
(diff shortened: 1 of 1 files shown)


packages/server/graphql/mutations/helpers/safelyWithdrawVote.ts
```diff
@@ -13,23 +15,40 @@ const safelyWithdrawVote = async (
...
+        )
+      })('replaced')
+      .eq(1)
+      .run(),
+    pg
+      .updateTable('RetroReflectionGroup')
+      .set({
+        voterIds: sql`array_cat(
+          "voterIds"[1:array_position("voterIds",'google-oauth2|ohlApAJnXy')-1],
+          "voterIds"[array_position("voterIds",'google-oauth2|ohlApAJnXy')+1:]
```
igorlesnenko, line 42: I see you hardcoded some Id here

packages/server/graphql/mutations/helpers/safelyWithdrawVote.ts
```diff
@@ -13,23 +15,40 @@ const safelyWithdrawVote = async (
...
+        )`
+      })
+      .where('id', '=', reflectionGroupId)
+      .where(sql`${userId}`, '=', sql`ANY("voterIds")`)
+      .executeTakeFirst()
+  ])
+  const isVoteRemovedFromGroupPG = voteRemovedResult.numUpdatedRows === BigInt(1)
+  if (isVoteRemovedFromGroup !== isVoteRemovedFromGroupPG)
+    console.log('MISMATCH VOTE REMOVED LOGIC')
```
igorlesnenko, line 51: Should we log this kind of errors in sentry?

igorlesnenko (changes requested):
I see we have this old method. Maybe remove to avoid confusion?

https://github.com/ParabolInc/parabol/blob/f3361ab138d76ecbd2fa233e518790226414709e/packages/server/graphql/mutations/autoGroupReflections.ts#L69

Do we need to update this place?

https://github.com/ParabolInc/parabol/blob/65452dcfdbbfabc8f8d1c2ecafcc0a6b1d2d1fb5/packages/server/graphql/mutations/helpers/updateReflectionLocation/removeReflectionFromGroup.ts#L80-L87

packages/server/graphql/mutations/helpers/safelyWithdrawVote.ts
mattkrick: 🙈 too much debugging, thank you!

packages/server/graphql/mutations/helpers/safelyWithdrawVote.ts
mattkrick: yep! that's probably easier to get to than datadog

mattkrick:
awesome review!

- `autogroupReflections` removed in https://github.com/ParabolInc/parabol/pull/8505, merged master to incorporate those changes
- removeReflectionGroup inactivate logic applied to PG
- fixed hardcoded userId
- sending mismatches to sentry

igorlesnenko (approved):

PR #8508, opened 2023-07-13
feat: Add timer sound effect

Dschoordsch:
# Description

Cleanup of #7858 

When the stage timer runs out, play a ticking sound towards the end.
I've decided to not add a chime at the end for now, as we quite often run over the timer and this might be more distracting than necessary.

## Demo

https://www.loom.com/share/5689c6b63c574f8cab06ef7ac386938e?sid=7ab8b556-cba1-4d91-b7a2-818f341f7cad

## Testing scenarios

You can speed up testing by replacing `1m` with `10s` [here](https://github.com/paul-vd/parabol/blob/8ce3d12ae3cc936955ad8cd368afd2f2e1054ab9/packages/client/components/StageTimerModalTimeLimit.tsx#L90)
* [ ] start timer and wait until it runs out, sound is played towards the end
* [ ] start timer and refresh page, browser will block sound if there was no interaction, but no exception will be thrown
* [ ] start timer and while the sound is playing, add additional time: sound should stop

## Final checklist


packages/client/hooks/useSoundEffect.ts
```diff
@@ -0,0 +1,46 @@
+import {useEffect, useRef} from 'react'
+
+type Props = {startTime: Date | string} | {endTime: Date | string}
```
nickoferrall, line 3: +1 a dev could add both the `startTime` and the `endTime` to this hook which would prevent it from working. 

Could change the type to prevent that from happening:
```
type StartTimeProps = {startTime: Date | string; endTime?: never}
type EndTimeProps = {endTime: Date | string; startTime?: never}
type Props = StartTimeProps | EndTimeProps
```

nickoferrall (approved):
LGTM!

Issue #8491, opened 2023-07-07
bug(ai-summary): summaries occasionally show pronoun gender bias


jordanh:
## Issue - Bug

This prompt to ChatGPT:
> Below is a comma-separated list of text from a retro meeting. Summarize the text for a second-grade student in one or two sentences. If you can't provide a summary, simply write the word "No".
>
> Text: New hire for cloud, SO to Robin, who has jumped right in and is asking great questions, identifying areas for improvement, etc., Welcome Robin!, Robin Joining!, Robin Joining the team

Generated this response:

> Robin has joined a team and is already asking good questions and finding ways to make things better. Everyone is very excited to have him onboard!

Issue with this: Robin identifies as a she. Oof.

This could be avoided by engineering the prompt for the summary slightly differently. For example:

> Below is a comma-separated list of text from a retro meeting. Summarize the text for a second-grade student in one or two sentences. When referring to people in the summary, do not assume their gender and default to using the pronouns "they" and "them." If you can't provide a summary, simply write the word "No".
>
> Text: New hire for cloud, SO to Robin, who has jumped right in and is asking great questions, identifying areas for improvement, etc., Welcome Robin!, Robin Joining!, Robin Joining the team

### Acceptance Criteria (optional)
Users can:
   - Prompt update to avoid gendering individuals when the gender of a subject is ambiguous

**Estimated effort:** 1 hour

PR #8509, opened 2023-07-13
fix: avoid pronoun assumptions in ai summary

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8491

I used a newline delimiter as @jordanh suggested [here](https://parabol.slack.com/archives/C042HTVBH8D/p1688755623776749). 

I didn't include the text `If you can't provide a summary, simply write the word "No".` which is mentioned in the issue because I got the following response when summarising reflections which mentioned a bunch of random artists (something it can usually do without problems): 

Screenshot 2023-07-13 at 15 19 19

### To test

- [ ] Create a retro, and see AI Summaries are working as expected
- [ ] Copy & paste the reflections in the issue and see that the summary does not assume gender

### DIFF
--- a/packages/server/utils/OpenAIServerManager.ts
+++ b/packages/server/utils/OpenAIServerManager.ts
@@ -18,14 +18,15 @@ class OpenAIServerManager {
 
   async getSummary(text: string | string[], summaryLocation?: 'discussion thread') {
     if (!this.openAIApi) return null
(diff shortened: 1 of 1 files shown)


igorlesnenko (approved):
Works!

> If you can't provide a summary, simply write the word "No".

If needed, we can handle it manually in code, `if text === 'No.'`

Issue #8503, opened 2023-07-11
Org Details: Teams View


acressall:
Design: #7983 

Billing leads should have more visibility into when they have exceeded the starter tier. They should not be able to remove a team if they are not the team lead.

AC:
- As a billing lead I can view all the teams on my org
- As a billing lead I can view the team lead of each team
- As a team lead I can click a link to "manage team" which brings me to the team's settings view

PR #8510, opened 2023-07-13
feat: teams view

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8503

Demo: https://www.loom.com/share/62299583ec2d40b886e38c79c6092411

Screenshot 2023-07-13 at 17 39 14

Screenshot 2023-07-13 at 17 50 38

@acressall in the issue AC we have `As a billing lead I can view all the teams on my org`. Just confirming that non-billing leaders can view all teams on the org too?

I made two changes from the design:

1. I haven't included the `6 teams` row at the top. This is because it took a bit of time to start customising the `Panel` component. I then wondered whether we need it as most orgs won't have so many teams that it's hard to count. Not including it saves some screen real estate.
2. `Last met at` isn't included. I'll include this in a later PR. @acressall I wanted to check that this date is the date the team last had a meeting?

### To test

- [ ] Users with the `checkoutFlow` feature flag can see all the teams in the org
- [ ] Billing leaders see "Manage Team". If they click it, they go to the team's settings page
- [ ] Team lead emails are visible on the right of the row

### DIFF
--- a/packages/client/modules/userDashboard/components/OrgTeams/OrgTeamsRow.tsx
+++ b/packages/client/modules/userDashboard/components/OrgTeams/OrgTeamsRow.tsx
@@ -1,0 +1,74 @@
+import React from 'react'
+import graphql from 'babel-plugin-relay/macro'
+import styled from '@emotion/styled'
+import Row from '../../../../components/Row/Row'
+import Panel from '../../../../components/Panel/Panel'
+import {Breakpoint, ElementWidth} from '../../../../types/constEnums'
+import {useFragment} from 'react-relay'
+import plural from '../../../../utils/plural'
+
+const StyledPanel = styled(Panel)({
+  maxWidth: ElementWidth.PANEL_WIDTH
+})
+
+const StyledRow = styled(Row)({
+  padding: '12px 8px 12px 16px',
+  [`@media screen and (min-width: ${Breakpoint.SIDEBAR_LEFT}px)`]: {
+    padding: '16px 8px 16px 16px'
+  }
+})
+
+type Props = {
+  teamRef: any // OrgPage_team$key
+}
+
+const OrgTeamsRow = (props: Props) => {
+  const {teamRef} = props
+  const team = useFragment(
+    graphql`
+      fragment OrgTeamsRow_team on Team {
+        name
+        teamMembers {
+          id
+          isLead
+          isSelf
+          email
+        }
+      }
+    `,
+    teamRef
+  )
+  console.log('🚀 ~ team:', team)
+  const {teamMembers, name} = team
+  const teamsCount = 2
+  const teamMembersCount = teamMembers.length
+  const teamLeadEmail = teamMembers.find((member) => member.isLead).email
+  const isViewerTeamLead = teamMembers.some((member) => member.isSelf && member.isLead)
+  console.log('🚀 ~ teamLeadEmail:', teamLeadEmail)
+  // const teamMembersCount = teams.reduce((acc, team) => acc + team.teamMembers.length, 0)
+  // const teamLeadEmail = teams.console.log('🚀 ~ organization:', {organization, teamMembersCount})
+  return (
+    <StyledRow>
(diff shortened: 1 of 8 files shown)


tianrunhe:
We define `lastMetAt` for Team as: `max(coalesce(meeting.endedAt, meeting.createdAt))`, i.e., the latest time when they completed **or started** a meeting. We also look at meeting creation time to counter "evergreen" meetings

nickoferrall:
> I was thinking to maintain privacy / while all teams are private, only billing leads would see all teams

Makes sense! Only billing leaders can view the teams page now 👍

igorlesnenko (approved):
Works!

Some feelings:
- Left and right panel paddings feels too big to me
- Wanted team name to be clickable
- Should we mailto our own email?

nickoferrall:
> Left and right panel paddings feels too big to me

Cool, I've reduced the padding a bit 👍 

> Wanted team name to be clickable

Where would you like to be taken to when you click on the team name? I'd imagine the team settings page, but that's where the "Manage Team" button takes you. If we removed the "Manage Team" button, it might feel strange that you can only click on the team names where you're a team lead 🤷 

> Should we mailto our own email?

As in, when you click on the Team Lead email, Parabol's email is the "To" email? If I click on an email address, I'd expect that email to be the one in the "To" field

Dschoordsch (changes requested):
Does not work for non-super users:
https://github.com/ParabolInc/parabol/blob/eec22e64c694e988d199c6178d66983f0aca7786/packages/server/graphql/types/Organization.ts#L76-L78

nickoferrall:
> Does not work for non-super users:

Great catch! Now that we want billing leaders to access all teams in the org, I think it's OK to relax the restriction from super users to billing leaders, so I've implemented this update for `teams` and `team members`

packages/client/modules/userDashboard/components/OrgTeams/OrgTeamsRow.tsx
```diff
@@ -0,0 +1,67 @@
...
+                  className='cursor-pointer font-bold text-sky-500 hover:underline'
+                >
+                  {'Manage Team'}
+                </Link>
+              </>
+            )}
+          </div>
+
+          <a
+            href={`mailto:${teamLeadEmail}`}
```
Dschoordsch, line 55: +1 I would add `target='_blank'`

packages/server/graphql/types/Team.ts
```diff
@@ -327,9 +327,13 @@ const Team: GraphQLObjectType = new GraphQLObjectType<ITeam, GQLContext>({
         }
       },
       description: 'All the team members actively associated with the team',
-      async resolve({id: teamId}, args: any, {authToken, dataLoader}) {
+      async resolve({id: teamId, orgId}, args: any, {authToken, dataLoader}) {
+        const viewerId = getUserId(authToken)
         const {sortBy = 'preferredName'} = args as {sortBy?: 'preferredName'}
-        if (!isTeamMember(authToken, teamId) && !isSuperUser(authToken)) return []
+        const isBillingLeader = await isUserBillingLeader(viewerId, orgId, dataLoader)
+        const canViewAllMembers =
```
Dschoordsch, line 334: +1 actually all the billingLeader needs to see is the team lead. Not a big deal, but we could add a separate `teamLeads` field. The billing lead can see all org members anyways, but maybe some team members should be kept private?

Dschoordsch (approved):

packages/client/modules/userDashboard/components/OrgTeams/OrgTeamsRow.tsx
nickoferrall: Updated 👍

packages/server/graphql/types/Team.ts
nickoferrall: We need to show the team member count too, so I think we need access to all team members here

Issue #8511, opened 2023-07-13
Creating a new team in an existing org crashes the page


nickoferrall:
The team is still created, but the page crashes. 

### AC

- [ ] Can create a new team in an existing org

Dschoordsch:
Sentry issue: [ACTION-PRODUCTION-EZE](https://parabol.sentry.io/issues/4312123678/?referrer=github_plugin)

PR #8512, opened 2023-07-13
fix: can create team within existing org

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8511

### To test

- [ ] Create a new team within an existing org without the page crashing

### DIFF
--- a/packages/client/modules/newTeam/components/NewTeamForm/NewTeamForm.tsx
+++ b/packages/client/modules/newTeam/components/NewTeamForm/NewTeamForm.tsx
@@ -101,6 +101,15 @@ interface Props {
 
 const NewTeamForm = (props: Props) => {
   const {isInitiallyNewOrg, organizationsRef} = props
+
+  graphql`
+    fragment NewTeamForm_teams on Team @relay(plural: true) {
+      teamMembers {
+        email
+        isSelf
+      }
+    }
+  `
   const organizations = useFragment(
     graphql`
       fragment NewTeamForm_organizations on Organization @relay(plural: true) {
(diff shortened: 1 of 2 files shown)


packages/client/modules/newTeam/components/NewTeamForm/NewTeamForm.tsx
```diff
@@ -129,7 +129,7 @@ const NewTeamForm = (props: Props) => {
   const disableFields = !!lockedSelectedOrg && !isNewOrg
   const selectedOrg = organizations.find((org) => org.id === orgId)
   const selectedOrgTeamMemberEmails = selectedOrg?.teams.flatMap(({teamMembers}) =>
-    teamMembers.filter(({isSelf}) => !isSelf).map(({email}) => email)
+    teamMembers?.filter(({isSelf}) => !isSelf).map(({email}) => email)
```
igorlesnenko, line 132: Not sure I understand, in what cases `teamMembers` will be null? I see we are fetching them above in graphql query

nickoferrall: Looks like the `teamMembers` haven't populated in the newly created team. I've updated the `AddTeamMutation` to query for the `teamMembers` as that's a cleaner way of doing it

packages/client/mutations/AddTeamMutation.ts
```diff
@@ -19,6 +19,9 @@ graphql`
     team {
       id
       name
+      teamMembers {
```
igorlesnenko, line 22: Can we spread a fragment here? teamMembers are not used in this file directly, so it is unknown why are they here

igorlesnenko (approved):

packages/client/mutations/AddTeamMutation.ts
nickoferrall: I like it! Updated 👍

Issue #8497, opened 2023-07-10
Show loading feedback while upgrading 


nickoferrall:
It can take a while for Stripe to respond, so we should give the user feedback that they're upgrading.

Here's a design from @acressall:
https://parabol.slack.com/archives/C042E8PGPD4/p1688654879235479?thread_ts=1688588114.295889&cid=C042E8PGPD4

### AC

- [ ] Give the user feedback that looks like the design

PR #8514, opened 2023-07-14
feat: show loading feedback while upgrading

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8497

Screenshot 2023-07-14 at 15 12 23
Screenshot 2023-07-14 at 15 12 20
Screenshot 2023-07-14 at 15 12 15

### To test

- [ ] Add the checkout flow flag and run `stripe listen --forward-to localhost:3000/stripe` in your terminal
- [ ] Upgrade to the Team tier
- [ ] See that there's feedback while upgrading with an ellipsis and the text changing from Upgrade to Upgrading


igorlesnenko (approved):
Works!

I would also think of possibility of optimistic upgrade, because loading takes some time for me, looks like stripe webhooks are not very fast.
Also If I refresh the page, I will see the page like it is not upgraded (though upgrade will happen after)

nickoferrall:
> I would also think of possibility of optimistic upgrade

But if we optimistically upgrade and then the payment fails, we'd show the user the Team tier when they're not on the Team tier. _Maybe_ that's still worth it so that the majority of users successfully upgrading get a better experience 🤔

igorlesnenko:
> But if we optimistically upgrade and then the payment fails, we'd show the user the Team tier when they're not on the Team tier. Maybe that's still worth it so that the majority of users successfully upgrading get a better experience 🤔

yeah we would optimistically upgrade, and then downgrade back if payment fails.

Issue #8480, opened 2023-07-05
Include customer billing info in the invoices


tianrunhe:
## Issue - Enhancement

Today the invoices (e.g., [this one](https://action.parabol.co/invoice/in_1NMagjFLaFINmHnBcqd9FfZd)) we generated in the app only include billing leaders' email address. We have a bunch of customer requests to include the billing information in the invoices. I think to the simplest starting point, we can just mimic the invoices generated by Stripe in the app (we had to self-generate and customize the invoice because we used to do proration - now we don't need that anymore)

### Acceptance Criteria (optional)
- [ ] `Team` tier billing leaders can see their invoices in the app that include billing information like address, phone numbers, tax ids etc.

**Estimated effort:** X points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

avivapinchas:
[Discussed in user feedback retro](https://action.parabol.co/meet/nJIAHfOafS/discuss/6). 

The ideal scenario would allow users to: 
- Have custom information included on the invoice 
- Download the invoice as PDF (vs printing it from a web view) 
- Be notified when a new invoice is available 

We should first investigate how much of these needs can be covered by this shift to using Stripe invoices, then log issue for any remaining bits not covered here.

nickoferrall:
- It looks like Stripe has the following field on the `invoice` object: `invoice.hosted_invoice_url`
- If you click on that, you get taken to a page where you can download the invoice. Below I've included screenshots of what that page and the downloaded invoice look like:

![Screenshot 2023-07-12 at 08 42 23](https://github.com/ParabolInc/parabol/assets/39854876/19feffeb-b44c-47c6-97cc-0bcd5c8bb45b)
<img width="738" alt="Screenshot 2023-07-12 at 08 42 39" src="https://github.com/ParabolInc/parabol/assets/39854876/cf97460d-d7f8-4ab7-bdcd-03a3e9663281">


After clicking on the invoice in Parabol (see below), we could take the user to the `invoice.hosted_invoice_url` and show them this Stripe page. 

![Screenshot 2023-07-12 at 08 46 01](https://github.com/ParabolInc/parabol/assets/39854876/4b301501-3864-4f3f-a6d8-12fd5eb56308)


@avivapinchas you mentioned that ideally we'd have `custom information included on the invoice`. Is there anything not included in the screenshots above that you think we need to include?

[...]

PR #8534, opened 2023-07-17
feat: show stripe invoices

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8480

Demo: https://www.loom.com/share/3c9a1cd4f47543869e7abc9f727f25a4

Note: this only works for invoices created after this PR. If that's problematic, we could fix this by fetching the `invoice.hosted_invoice_url` from Stripe in the `invoice.payUrl` type.

### To test

- [ ] Run `tripe listen --forward-to localhost:3000/stripe` in your terminal and upgrade to the Team tier
- [ ] Click on the paid invoice and see that it takes you to Stripe's UI
- [ ] See that you can download the invoice successfully


igorlesnenko (approved):
This looks good to me, but few things I noticed:

In the stripe invoice there is an email: `admin@parabol.co` while in our own invoices `love@parabol.co` and  `billing@parabol.co` should we update email in stripe? 

When I open some old invoice, it first opens in the old UI, but when I go back and click on it again it opens in stripe. Is it expected behaviour?

Once I upgraded an org, and got this error in logs
```
Error: This object cannot be accessed right now because another API request or Stripe process is currently accessing it. If you see this error intermittently, retry the request. If you see this error frequently and are making multiple concurrent requests to a single object, make your requests serially or at a lower rate.
```
After that I only able to see the estimate invoice, not paid:
image

nickoferrall:
> should we update email in stripe?

Good eye! I don't think so as I believe they're all sent to Hubspot 

> When I open some old invoice, it first opens in the old UI, but when I go back and click on it again it opens in stripe

I think that's a caching issue, so it shouldn't be a problem

> Once I upgraded an org, and got this error in logs

Thanks for sharing! I've been getting that too. I created a PR to fix it here: https://github.com/ParabolInc/parabol/pull/8545

packages/client/modules/userDashboard/components/InvoiceRow/InvoiceRow.tsx
```diff
@@ -90,9 +92,19 @@ const InvoiceRow = (props: Props) => {
...
+      window.open(payUrl, '_blank', 'noreferrer')
+    } else {
+      history.push(`/invoice/${invoiceId}`)
+    }
+  }
+
   return (
     <Row>
-      <RowLink rel='noopener noreferrer' target='_blank' to={`/invoice/${invoiceId}`}>
+      <RowLink onClick={goToInvoice}>
```
Dschoordsch, line 107: -1 Previously also internal invoices would open in a new window. I think we could just do a
```ts
      <RowLink rel='noopener noreferrer' target='_blank' to={payUrl || `/invoice/${invoiceId}`}>
```
here

Dschoordsch (changes requested):

packages/client/modules/userDashboard/components/InvoiceRow/InvoiceRow.tsx
```diff
@@ -43,18 +43,17 @@ const InfoRowRight = styled('div')({
...
   justifyContent: 'space-between',
   textDecoration: 'none',
-  width: '100%',
-  '&:hover': {
-    cursor: 'pointer'
-  }
+  width: '100%'
 })
 
+const RowLink = LinkStyles.withComponent(Link)
```
Dschoordsch: -1 Does not work, it redirects to
`http://localhost:3000/me/organizations/oU3Kik4zU4/https://invoice.stripe.com/...` for me

Dschoordsch (changes requested):

packages/client/modules/userDashboard/components/InvoiceRow/InvoiceRow.tsx
nickoferrall: 🤦‍♂️ I've replaced the RowLink element with an achor element using tailwind. It's fixed for me now!

Dschoordsch (approved):

Issue #8544, opened 2023-07-19
Unable to access Stripe object because another API request is using it


nickoferrall:
When upgrading to the Team tier, this console error happens frequently:

```
Error: This object cannot be accessed right now because another API request or Stripe process is currently accessing it. If you see this error intermittently, retry the request. If you see this error frequently and are making multiple concurrent requests to a single object, make your requests serially or at a lower rate.
```

### AC

- [ ] Upgrade to team tier without getting the console error

PR #8545, opened 2023-07-19
fix: stripe console err

nickoferrall:
Fix: https://github.com/ParabolInc/parabol/issues/8544

Credit to @mattkrick for helping me with this during office hours. 

Sorry, this includes the updated changelog from the Ship. It'll end up in master shortly, so I think it's OK leaving it here.

This PR will increase the time it takes to upgrade as there could be a few retries. I've created a GitHub issue here to sped things up: https://github.com/ParabolInc/parabol/issues/8546

### To test

- Run `stripe listen --forward-to localhost:3000/stripe` in your terminal
- Upgrade to Team tier a few times using the new checkout flow
- You don't see the console error message mentioned in the linked issue


mattkrick:
LGTM, I see a changelog conflict, but after that's fixed feel free to merge!

Issue #8547, opened 2023-07-19
Update summary URL with Suggest Groups


nickoferrall:
### AC

- [ ] Summary URL includes Suggest Groups if users had access to it so we can show the Hubspot chatbot

nickoferrall:
Closed by https://github.com/ParabolInc/parabol/pull/8548

PR #8548, opened 2023-07-19
chore: update summary url with Suggest Groups

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8547

I'm requesting to merge this into Staging as I've started the ship, and it would be great to start testing it with users quickly. 

### AC

- [ ] Add Suggest Groups org feature flag
- [ ] End a retro and see that `sg` or `sg-xl` shows up in the URL depending on if there were more than 40 reflections


igorlesnenko (approved):
Works!

PR #8551, opened 2023-07-20
feat: Add meetings to the team dashboard

### LINKED ISSUE #8502: Show meetings in Team dashboard
As a user with multiple teams, I want to see my meetings by team easily.

## Acceptance criteria
* [ ] Add tabs to team dashboard "Activity", "Tasks"
* [ ] "Tasks" contains current task view
* [ ] "Activity" contains the meetings view filtered by this team

Dschoordsch:
# Description

Fixes #8502 
Add a tab for meetings to the team dashboard. The default view is still "Tasks" to not interrupt users habits too much.

## Demo

https://www.loom.com/share/d3a5e74fd16842d387c59648f3539a48?sid=92906de6-0297-4f1e-b09d-bb9c62e9fe48

## Testing scenarios

- [ ] open team dashboard
- [ ] switch between tabs
- [ ] try different filters
- [ ] open a meeting
- [ ] check the sidebar highlight

## Final checklist

### DIFF
--- a/packages/client/components/DashNavList/DashNavList.tsx
+++ b/packages/client/components/DashNavList/DashNavList.tsx
@@ -2,6 +2,7 @@ import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
 import React, {Fragment, useMemo} from 'react'
 import {useFragment} from 'react-relay'
+import {useRouteMatch} from 'react-router'
 import {PALETTE} from '~/styles/paletteV3'
 import {Breakpoint} from '~/types/constEnums'
 import makeMinWidthMediaQuery from '~/utils/makeMinWidthMediaQuery'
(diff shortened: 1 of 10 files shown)


Dschoordsch:
@acressall I used the existing tabs instead of designing the component exactly like in the design. Same for the empty meeting cards. Can you check if that's ok?

acressall:
@Dschoordsch I think using the existing elements is fine! Looks good! My only hesitation is defaulting to the tasks view. When I create a new team, will it still default to tasks? I think activity would be the more useful. 

I agree that showing tasks first creates consistency for existing users, but it doesn't add any value for new users (and might be confusing since you'd be starting on the _second_ tab). We also know that not all teams use tasks, so the activity tab might be the more generally useful for everyone. WDYT?

Dschoordsch:
@acressall No strong feelings either way. I think if we default to the activity, we should explicitly link to the tasks dashboard from the summary page: The meeting just ended, so likely there is nothing to see in the activity, but you might just have some new tasks. I'm not sure if there are other links to the team dashboard to consider.

acressall:
> I think if we default to the activity, we should explicitly link to the tasks dashboard from the summary page: The meeting just ended, so likely there is nothing to see in the activity, but you might just have some new tasks. I'm not sure if there are other links to the team dashboard to consider.

Yes, let's link to the tasks tab from meeting summary. I think the context of that makes sense. 
The only other link to the team dash I can think of is when you're in a meeting it says the team name in the menu and that's clickable. In that case I think it makes sense to go to the activity tab, so you can get back to the meeting if you need to.

Dschoordsch:
@igorlesnenko I changed the default to the activity tab, just like @acressall suggested. Any objections?

Issue #8483, opened 2023-07-05
With a feature flag, show all teams within a domain


nickoferrall:
See @avivapinchas's suggestion [here](https://parabol.slack.com/archives/C042E8PGPD4/p1688574183822049?thread_ts=1688503893.686599&cid=C042E8PGPD4). Related to @acressall's issue here: https://github.com/ParabolInc/parabol/issues/7983

### AC

- [ ] Add a user feature flag
- [ ] If the feature flag is turned on, users can see a toggle at the top of the Team view
- [ ] By clicking the toggle, they can see all of the teams within a domain, not just their org

PR #8554, opened 2023-07-20
feat: show all teams within domain

nickoferrall:
Fix: https://github.com/ParabolInc/parabol/issues/8483

@mattkrick we discussed this PR a while ago, and you mentioned privacy concerns, specifically around showing the Team Lead email. I spoke with the Squad, and the sales team currently send a Google Sheet with this info by email. That doesn't make it OK, but it should make it more secure than what we're currently doing. 

I added a comment to the `canViewTeamsInDomain` feature flag to warn future devs to be careful with this flag, and, if this PR is approved, I'll include a warning in the Notion doc too. 

Demo: https://www.loom.com/share/edaf3ff794714b249b17a3835ef55623

If this PR is approved, I'll create a Notion doc with an example mutation and let the sales team know. 

### To test

- [ ] Add the `canViewTeamsInDomain` and `checkoutFlow` user feature flags
- [ ] Go to the Teams view of an org that has the same `activeDomain` as the viewer's domain, e.g. `Parabol Org` has the `parabol.co` domain, and your email is `parabol.co`
- [ ] See that all teams within the domain are visible
- [ ] You can filter to see all teams within org
- [ ] Go to an org with an active domain that's different to the viewer's domain, and you don't see the filter
- [ ] Users without the `canViewTeamsInDomain` can only see teams within their org and don't see the filter menu


packages/client/modules/userDashboard/components/OrgTeams/OrgTeams.tsx
```diff
@@ -1,49 +1,118 @@
...
+    queryRef
   )
-  const {teams, isBillingLeader} = organization
+
+  const {togglePortal, originRef, menuPortal, menuProps} = useMenu(MenuPosition.UPPER_RIGHT)
+  const {viewer} = data
+  const {featureFlags, domains, organization} = viewer
+  const {activeDomain, teams, isBillingLeader, name} = organization ?? {}
+
+  const ALL_TEAMS_IN_ORG = `All Teams In ${name}`
```
nickoferrall, line 63: Originally, I added these labels to `FilterLabels` in `constEnums`, but I think we should show the org/domain name, so I'm instantiating them here

packages/client/modules/userDashboard/components/OrgTeams/OrgTeams.tsx
```diff
@@ -1,49 +1,118 @@
...
   return (
     <>
       <h1>{'Teams'}</h1>
+      {canViewTeamsInDomain && (
+        <>
+          <DashFilterToggle onClick={togglePortal} label={label} ref={originRef} value={label} />
+          {menuPortal(
+            <Menu
+              keepParentFocus
+              defaultActiveIdx={label === ALL_TEAMS_IN_DOMAIN ? 0 : 1}
```
nickoferrall, line 92: If we turn on the feature flag, I assume we'll be directing users to the teams within the domain, so I chose this as the default active index

nickoferrall:
Hey @acressall, it'd be great to get your eyes on the design 🙏  I went with a filter menu instead the switch button I suggested in the issue because I think it's more consistent with our design, and we already have a dropdown menu built

acressall (approved):
Filter is a good choice!

packages/client/modules/userDashboard/components/OrgTeams/OrgTeams.tsx
```diff
@@ -1,49 +1,118 @@
...
+
+  const {canViewTeamsInDomain} = featureFlags
+
+  const handleMenuItemClick = (newLabel: string) => {
+    setLabel(newLabel)
+    togglePortal()
+  }
+
+  const selectedTeams =
+    (label === ALL_TEAMS_IN_DOMAIN && canViewTeamsInDomain ? teamsByDomain : teams) ?? []
```
igorlesnenko, line 81: There is a rare edge case. If rename org to `parabol.co`, toggle breaks so I can't switch it.

nickoferrall:
Updated! I was able to see the teams when you couldn't because I had `insights` turned on which meant `suggestedTier` was true for me: https://github.com/ParabolInc/parabol/blob/74b7e46b72637a0a3cc7c9c5c349b1f7e0249a49/packages/server/graphql/public/types/User.ts#L62

I'm now querying the data from the `company` field rather than `domains`, so you should be able to see it now 👍

igorlesnenko (changes requested):
I have a team that in example.com domain

But it shows a teams from parabol.co domain

image

image

igorlesnenko (changes requested):
If I now go to my example.com org, I don't see any filter anymore

image

igorlesnenko (approved):
As far as I understand if my own domain is different from org domain, I can't see the filter now

nickoferrall:
> As far as I understand if my own domain is different from org domain, I can't see the filter now

Yep, exactly. Your domain is `parabol.co` so `company.organizations` will only return the orgs in your `parabol.co` domain (see [here](https://github.com/ParabolInc/parabol/blob/0f638bcce554c7adea5ac07043d3d7f373a0250b/packages/server/graphql/public/types/Company.ts#L21)). 

If you have the `canViewTeamsInDomain` feature flag, you can go to `Parabol Org` and see all the teams in the org & the domain.

If you go to `Example Org` created by someone with the `example.com` domain, you shouldn't be able to view all the teams in that domain, so I'm hiding the filter.

packages/server/graphql/public/types/Company.ts
```diff
@@ -20,6 +21,8 @@ const getSuggestedTierOrganizations = async (
   const organizations = await dataLoader.get('organizationsByActiveDomain').load(domain)
   const orgIds = organizations.map(({id}) => id)
   const viewerId = getUserId(authToken)
+  const viewer = await getUserById(viewerId)
```
mattkrick, line 24: +1 how can i test this? to fetch `{viewer { company { id } }` we already have to get the viewer, so I'm not sure why it's needed? if it is needed, could we use the dataloader to reduce a fetch?

PR #8563, opened 2023-07-24
feat: add emoji insights

### LINKED ISSUE #8465: Insights: Favourite Emojis
## Acceptance criteria
* on meeting end collect statistics about used reactions
* add column to `TeamInsights` table to collect data
* add to `TeamInsights` graphql type, set null if condition is not fullfilled, otherwise it contains most used emojis
  * condition: `team has at least 2 emojis that have been used >= 5 times`
* add UI

## Out of scope
* collecting data retroactively
* "Is this helpful?"

Dschoordsch:
# Description

Fixes #8465
At the end of a meeting, count the used emojis and add them to the
insights.

## Demo

https://www.loom.com/share/f040d0e3cb274754bcfb9f2fd0fa9878?sid=ef83dfb1-7105-40d3-a8dc-01ff88d8c31f

## Testing scenarios

- [ ] add `teamInsights` org feature flag
- [ ] run and end different meeting types with reactjis
- [ ] collect at least 2 reactjis used 5 times or more
- [ ] see the insights card in team dash

## Final checklist

### DIFF
--- a/packages/server/graphql/mutations/helpers/collectReactjis.ts
+++ b/packages/server/graphql/mutations/helpers/collectReactjis.ts
@@ -1,0 +1,53 @@
+import Meeting from '../../../database/types/Meeting'
+import {getTeamPromptResponsesByMeetingId} from '../../../postgres/queries/getTeamPromptResponsesByMeetingIds'
+import {DataLoaderWorker} from '../../graphql'
+import isValid from '../../isValid'
+
+const collectReactjis = async (meeting: Meeting, dataLoader: DataLoaderWorker) => {
+  const {id: meetingId, phases} = meeting
+
+  const usedReactjis: Record<string, number> = {}
+
+  // Discussions can happen in many different stage types: discuss, ESTIMATE, reflect, RESPONSES
+  const stages = phases.flatMap(({stages}) => stages)
+  const discussionIds = stages
+    .map((stage) => 'discussionId' in stage && stage.discussionId)
+    .filter(isValid) as string[]
+  const discussions = (
+    await dataLoader.get('commentsByDiscussionId').loadMany(discussionIds)
+  ).filter(isValid)
+
+  discussions.forEach((comments) => {
+    comments.forEach(({reactjis}) => {
+      reactjis.forEach(({id}) => {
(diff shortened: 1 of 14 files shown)


acressall:
> I'm collecting only reactions for the insights right now. I think that's valuable enough.
Should we maybe rename the card to "Favorite Reactions"?

Yes, I think this makes sense

> Also the design shows some placeholder text, but I don't think we need to add anything there.
What about the (i) in the top right, will we have additional information to show there?

Gareth is working on copy [here](https://www.notion.so/parabol/Insights-cakepop-copy-4c20939b6efe4ff9b54a2a4de42d08f9), but will be working on this when he's back next week. I think it would be good to have *something* here so it's consistent with the other cards, but we could add copy updates in a separate issue if you like

packages/server/graphql/mutations/helpers/collectReactjis.ts
```diff
@@ -0,0 +1,53 @@
+import Meeting from '../../../database/types/Meeting'
+import {getTeamPromptResponsesByMeetingId} from '../../../postgres/queries/getTeamPromptResponsesByMeetingIds'
+import {DataLoaderWorker} from '../../graphql'
+import isValid from '../../isValid'
+
+const collectReactjis = async (meeting: Meeting, dataLoader: DataLoaderWorker) => {
```
nickoferrall, line 6: I imagine you're intentionally not checking the feature flag here because it's useful to add this data for all meetings so that if we turn the feature flag on, they get to see the Insight sooner rather than later?

I think that's probably the right approach. The downside is it means storing more data in RethinkDB, and we're querying the last 90 days of meetings in `updateTeamInsights` every time we end a meeting.

nickoferrall (approved):
This looks great! Tested and it's working well 👍

packages/server/graphql/mutations/helpers/collectReactjis.ts
Dschoordsch: These are valid concerns. I will go with a compromise: collecting the emojis is not that much effort as the data is loaded anyways and we're updating the meeting already. I will change the update method to only run for feature flag owners.

Issue #8578, opened 2023-07-27
Can add the Zoom transcription bot during the discuss phase


nickoferrall:
Previously, we'd add the transcription bot in the `setMeetingSettings` mutation. Once you'd paste the Zoom link, the bot was added right away.

I changed this so that the bot is added when the user proceeds to the Discuss phase (see https://github.com/ParabolInc/parabol/pull/8517) so that users can create async Retros. 

Now, in the Discuss phase, you can still paste the Zoom link, but it won't add the bot because you're already in the Discuss phase.

### AC

- [ ] Adding the Zoom link in the transcription tab in the Discuss phase adds the bot right away

PR #8580, opened 2023-07-27
fix: add transcription during discuss phase

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8578

Demo: https://www.loom.com/share/5134288f2eac41699969b11588b6f035?sid=a34e65d4-bd09-4448-a396-27c47718d961

### AC

- [ ] Add the `zoomTranscription` feature flag
- [ ] Start a retro meeting. Go to the Discuss phase and add a Zoom URL in the Transcription tab
- [ ] See that the zoom transcription bot is added to the meeting 
- [ ] End the meeting, go to the Transcription tab, and you can see the transcription


igorlesnenko (approved):
Works! Would be good to have some loading feedback when adding a zoom url, as it takes some time for me

packages/client/components/DiscussionThreadListEmptyState.tsx
```diff
@@ -131,18 +131,18 @@ const DiscussionThreadListEmptyState = (props: Props) => {
       }
     }
   })
+  if (!settings) return null
+  const {teamId, videoMeetingURL} = settings
```
Dschoordsch, line 135: -2 I only noticed in #8588 that the meeting settings are not per meeting. We don't prohibit having 2 retros at the same time, thus the settings would be shared. The correct way to handle this would be moving the settings from the meeting settings to the meeting on meeting creation.

Dschoordsch (changes requested):
I think you need to refactor where to store the videoUrl etc. before making any more improvements to the transcription.
Storing the video url on the settings for starting the meeting is fine, but once we create a meeting with those settings, we should move it to the meeting itself, resetting it in the settings.

packages/client/components/DiscussionThreadListEmptyState.tsx
nickoferrall: I've updated this to move the logic to the meeting object and clear `videoMeetingURL` from the settings once the meeting starts.

The PR still works for me, but I'm interested to know if it still works for you as Zoom recently announced that they're changing their policy on bots so we'll need to create an oauth flow:

Screenshot 2023-09-26 at 16 38 39

Dschoordsch (approved):

PR #8591, opened 2023-07-28
chore: change default width and padding of ui/Dialog

igorlesnenko:
In this PR:

- I changed the default width of a ui/Dialog to make it larger
- Applied to gcal schedule dialog to see how it works in other places

image


packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
```diff
@@ -127,9 +127,7 @@ const ActivityDetailsSidebar = (props: Props) => {
   )
   const {onError, onCompleted, submitting, submitMutation, error} = useMutationProps()
   const history = useHistory()
-  const {togglePortal: toggleModal, modalPortal} = useModal({
-    id: 'createGcalEventModal'
-  })
+  const {isOpen: isScheduleModalOpen, open, close} = useDialogState()
```
nickoferrall, line 130: +1 I think `openDialog` and `closeDialog` would be more descriptive, otherwise I'm not sure what `open()` is opening on line 181

packages/client/ui/Dialog/DialogContent.tsx
```diff
@@ -13,7 +13,7 @@ export const DialogContent = React.forwardRef<
     <RadixDialog.Content
       ref={ref}
       className={twMerge(
-        'fixed top-[50%] left-[50%] max-h-[85vh] w-[90vw] max-w-md translate-x-[-50%] translate-y-[-50%] rounded-lg bg-white p-4 shadow-dialog focus:outline-none data-[state=open]:animate-contentShow',
+        'md:w-xl fixed top-[50%] left-[50%] max-h-[85vh] w-[95vw] max-w-[95vw] translate-x-[-50%] translate-y-[-50%] rounded-lg bg-white p-4 shadow-dialog focus:outline-none md:max-w-xl',
```
nickoferrall, line 16: +1 I think there needs to be more padding to make the modal look the same as the old one. Here are the two next to each other:

Screenshot 2023-07-31 at 12 57 10
Screenshot 2023-07-31 at 12 57 39

nickoferrall (approved):

igorlesnenko:
Updated

image

@nickoferrall should I merge into your PR or wait?

igorlesnenko:
@acressall I have a design question:

I now increased padding in our generic dialog, so it looks like on the [screenshot above](https://github.com/ParabolInc/parabol/pull/8591#issuecomment-1660070940)

However, now paddings might looks a bit big for share topic dialog, because I guess it has smaller controls

image

Or is it good padding? Or should the controls be bigger? Or should I override padding for this dialog specifically and make it smaller?

nickoferrall:
@igorlesnenko feel free to merge it when you're ready!

acressall:
@igorlesnenko let's keep the padding the same as other dialogs, and later we can update the sizing of the select fields. We should have some consistent sizing for those, so I'll take a task to look at that with the design circle

Issue #8588, opened 2023-07-28
Meeting settings affect existing meetings


Dschoordsch:
How to reproduce:
* start a meeting with an icebreaker and just leave the meeting
* go to the start meeting settings for the same meeting type and disable icebreakers
* go back to the meeting
(this also works the other way round, disable first, start, enable)

Expected behaviour:
* the meeting is unaffected

Actual behaviour
* the meeting is in the icebreaker phase, but the navigation on the left does not show the icebreaker

https://www.loom.com/share/6d2975135aec4c19a2fc600a3bb3a185?sid=d110d114-2580-43df-9d2c-9d855216820a

Dschoordsch:
Root cause is the `settings` field in the Meeting type. It just returns the current settings and not the ones the meeting was created with.
As far as I could see it, the settings are only used to construct the sidebar and the phase order. This can be read from the meeting directly, thus it's safest to remove the `settings` field altogether from the meeting.

PR #8592, opened 2023-07-28
fix: Changing meeting settings affected existing meetings

Dschoordsch:
# Description

Fixes: #8588 

Changing the meeting settings after meeting creation would affect the meeting phases shown, i.e. the icebreaker would show up even if the meeting has none set, or it would be missing, although the meeting has one. Same for team health.

Meeting had a field settings which pointed to the meeting settings for that meeting type. These always pointed to the latest ones and not the ones associated with the current meeting. Because this is misleading, I decided it's safer to remove said field completely.

## Demo

See #8588

## Testing scenarios

- [ ] create a meeting with icebreaker
- [ ] start another meeting of the same type without
- [ ] check that each meeting has a different setting
- [ ] run Poker, Retro and Check-In normally and see switching phases works

## Final checklist

### DIFF
--- a/packages/client/mutations/EndRetrospectiveMutation.ts
+++ b/packages/client/mutations/EndRetrospectiveMutation.ts
@@ -47,10 +47,6 @@ graphql`
           }
         }
       }
-      settings {
-        recallBotId
-        videoMeetingURL
-      }
     }
     team {
       id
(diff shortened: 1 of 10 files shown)


Dschoordsch:
@nickoferrall this PR might break zoom transcription. I was fine with it because it's still in internal testing, but maybe you can just fix it if that's the case?

packages/client/mutations/EndRetrospectiveMutation.ts
```diff
@@ -47,10 +47,6 @@ graphql`
           }
         }
       }
-      settings {
-        recallBotId
-        videoMeetingURL
-      }
```
Dschoordsch, line 53: @nickoferrall it's unclear why these are here. I think it should be a fragment if this is necessary.
You could replace it with
```
team {
  meetingSettings(type: retrospective) {
    ...
  }
}
if necessary
```

Issue #8606, opened 2023-08-02
Post Notification of Standup Response shared in Slack


jmtaber129:
When a user submits their standup for the first time, send a message in slack saying that the user submitted their response, with a link to Parabol.

AC:
- [ ] When a user submits their standup for the first time, slack is notified
- [ ] When a user _updates_ their response, nothing happens
- [ ] The response includes:
  - [ ] User's name
  - [ ] Meeting name
  - [ ] Direct link to the user's response (using the `responseId` query param)

PR #8607, opened 2023-08-02
feat(standups): Notify slack when user submits standup response

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8606

When a user submits their standup response, send a message to slack with a direct link to their response.

To minimize scope, only supports slack for now.

## Demo
Screen Shot 2023-08-02 at 11 49 07 AM

## Testing scenarios
- [ ] Start a standup meeting on a team with a slack integration
- [ ] Make sure "Standup response submitted" is checked in the team integrations view
- [ ] Submit a standup response
- [ ] See the notification in the integrated slack channel
- [ ] Open the link in the slack notification, and confirm it opens the side drawer for that response
- [ ] Update the standup response, and confirm no additional slack messages are sent

## Final checklist

### DIFF
--- a/packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
+++ b/packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
@@ -15,6 +15,7 @@ import {DataLoaderWorker} from '../../../graphql'
 import getSummaryText from './getSummaryText'
 import {NotificationIntegrationHelper} from './NotificationIntegrationHelper'
 import {Notifier} from './Notifier'
+import {getTeamPromptResponsesByMeetingId} from '../../../../postgres/queries/getTeamPromptResponsesByMeetingIds'
 
 const notifyMSTeams = async (
   event: EventEnum,
@@ -339,6 +340,11 @@ export const MSTeamsNotificationHelper: NotificationIntegrationHelper<MSTeamsNot
     const adaptiveCard = JSON.stringify(card.toJSON())
     const attachments = MakeACAttachment(adaptiveCard)
     return notifyMSTeams('meetingEnd', webhookUrl, userId, teamId, attachments)
+  },
+
+  async standupResponseSubmitted() {
+    // Not implemented
+    return 'success'
   }
 })
 
@@ -400,6 +406,25 @@ export const MSTeamsNotifier: Notifier = {
 
   async integrationUpdated(dataLoader: DataLoaderWorker, teamId: string, userId: string) {
     ;(await getMSTeams(dataLoader, teamId, userId))?.integrationUpdated()
+  },
+
+  async standupResponseSubmitted(
+    dataLoader: DataLoaderWorker,
+    meetingId: string,
+    teamId: string,
+    userId: string
+  ) {
+    const {meeting, team} = await loadMeetingTeam(dataLoader, meetingId, teamId)
+    const user = await dataLoader.get('users').load(userId)
+    const responses = await getTeamPromptResponsesByMeetingId(meetingId)
+    const response = responses.find(({userId: responseUserId}) => responseUserId === userId)
+    if (!meeting || !team || !response || !user) return
+    ;(await getMSTeams(dataLoader, teamId, userId))?.standupResponseSubmitted(
+      meeting,
+      team,
+      user,
+      response
+    )
   }
 }
 
--- a/packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
+++ b/packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
@@ -414,9 +414,11 @@ export const MSTeamsNotifier: Notifier = {
     teamId: string,
     userId: string
   ) {
-    const {meeting, team} = await loadMeetingTeam(dataLoader, meetingId, teamId)
-    const user = await dataLoader.get('users').load(userId)
-    const responses = await getTeamPromptResponsesByMeetingId(meetingId)
+    const [{meeting, team}, user, responses] = await Promise.all([
+      loadMeetingTeam(dataLoader, meetingId, teamId),
+      dataLoader.get('users').load(userId),
+      getTeamPromptResponsesByMeetingId(meetingId)
+    ])
     const response = responses.find(({userId: responseUserId}) => responseUserId === userId)
     if (!meeting || !team || !response || !user) return
     ;(await getMSTeams(dataLoader, teamId, userId))?.standupResponseSubmitted(
--- a/packages/server/graphql/mutations/helpers/notifications/MattermostNotifier.ts
+++ b/packages/server/graphql/mutations/helpers/notifications/MattermostNotifier.ts
@@ -23,6 +23,7 @@ import {
 } from './makeMattermostAttachments'
 import {NotificationIntegrationHelper} from './NotificationIntegrationHelper'
 import {Notifier} from './Notifier'
+import {getTeamPromptResponsesByMeetingId} from '../../../../postgres/queries/getTeamPromptResponsesByMeetingIds'
 
 const notifyMattermost = async (
   event: EventEnum,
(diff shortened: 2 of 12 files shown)


packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
```diff
@@ -250,6 +251,24 @@ export const SlackSingleChannelNotifier: NotificationIntegrationHelper<SlackNoti
...
+  async standupResponseSubmitted(meeting, team, user, response) {
+    const options = {
+      searchParams: {
+        utm_source: 'slack share',
+        utm_medium: 'product',
+        utm_campaign: 'sharing',
+        responseId: response.id
+      }
+    }
+    const responseUrl = makeAppURL(appOrigin, `meet/${meeting.id}/responses`, options)
```
tianrunhe, line 265: I came here as the metrics police👮‍♂️ and gladly see you have already done what I would have asked for! ❤️

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
```diff
@@ -250,6 +251,24 @@ export const SlackSingleChannelNotifier: NotificationIntegrationHelper<SlackNoti
...
     // Slack sends a system message on its own
     return 'success'
+  },
+
+  async standupResponseSubmitted(meeting, team, user, response) {
+    const options = {
+      searchParams: {
+        utm_source: 'slack standup submission',
+        utm_medium: 'product',
+        utm_campaign: 'standups',
```
jmtaber129, line 261: @tianrunhe mind giving a ✅ on the updated UTM params here?

tianrunhe: Tagging @avivapinchas for a review here

tianrunhe: Aviva [suggested](https://parabol.slack.com/archives/C03C3J42WDU/p1691109288833119?thread_ts=1691095523.822949&cid=C03C3J42WDU) to change the campaign to `notifications`

jmtaber129: @tianrunhe updated!

nickoferrall (approved):

tianrunhe (approved):
LGTM from metrics side!

packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
```diff
@@ -400,6 +406,25 @@ export const MSTeamsNotifier: Notifier = {
...
+
+  async standupResponseSubmitted(
+    dataLoader: DataLoaderWorker,
+    meetingId: string,
+    teamId: string,
+    userId: string
+  ) {
+    const {meeting, team} = await loadMeetingTeam(dataLoader, meetingId, teamId)
+    const user = await dataLoader.get('users').load(userId)
+    const responses = await getTeamPromptResponsesByMeetingId(meetingId)
```
mattkrick, line 419: +1 can we parallelize these 3 calls + maybe that `getMSTeams`?

packages/server/graphql/mutations/helpers/notifications/MattermostNotifier.ts
```diff
@@ -387,5 +392,24 @@ export const MattermostNotifier: Notifier = {
...
     ;(await getMattermost(dataLoader, teamId, userId))?.integrationUpdated()
+  },
+
+  async standupResponseSubmitted(
+    dataLoader: DataLoaderWorker,
+    meetingId: string,
+    teamId: string,
+    userId: string
+  ) {
+    const {meeting, team} = await loadMeetingTeam(dataLoader, meetingId, teamId)
```
mattkrick, line 403: same parallel pattern for this one

packages/server/graphql/mutations/helpers/notifications/Notifier.ts
```diff
@@ -21,4 +21,10 @@ export type Notifier = {
     stageIndex: number,
     channelId: string
   ): Promise<NotifyResponse>
+  standupResponseSubmitted(
+    dataLoader: DataLoaderWorker,
+    meetingId: string,
+    teamId: string,
+    userId: string
+  ): Promise<void>
```
mattkrick, line 29: +1 return val is `'success' | {error: Error}` i think?

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
```diff
@@ -250,6 +251,24 @@ export const SlackSingleChannelNotifier: NotificationIntegrationHelper<SlackNoti
...
+  async standupResponseSubmitted(meeting, team, user, response) {
+    const options = {
+      searchParams: {
+        utm_source: 'slack standup submission',
+        utm_medium: 'product',
+        utm_campaign: 'notifications',
+        responseId: response.id
+      }
+    }
+    const responseUrl = makeAppURL(appOrigin, `meet/${meeting.id}/responses`, options)
```
mattkrick, line 265: +1 if i go to /responses it doesn't open the right drawer for me. it's the same behavior if i leave off the /responses. 
is that expected?

packages/server/graphql/mutations/helpers/notifications/Notifier.ts
jmtaber129: Nope! Most of the methods for the `Notifier` are treated as fire-and-forget, so we don't return anything here (`shareTopic` is different because the user action _is_ notifying, so it returns the result, but the other methods don't). It's the `NotifierIntegrationHelper` that returns something.

To be fair it's kind of a confusing pattern that I considered refactoring to be more readable as part of this, but I decided against investing the effort until we have a clearer sense of what upcoming integration work will look like.

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
jmtaber129: Are you including a `responseId` query param? It gets set as part of the `searchParams` in the `options` var above, so `/responses` alone just goes to the `responses` phase of the meeting, which is the only phase of standups.

If you connect your local dev to our #test-slack-bot channel (or some other test channel) and submit a response, you should see a message with a link that opens the correct drawer.

packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
jmtaber129: Ooh yea nice catch! Fixed.

I left the `getMSTeams` call (and similar calls for other integrations) as-is since the prevailing pattern seems to be "get data -> check data -> get integration"

packages/server/graphql/mutations/helpers/notifications/MattermostNotifier.ts
jmtaber129: Also fixed (and in the slack notifier, too)

packages/server/graphql/mutations/helpers/notifications/Notifier.ts
mattkrick: ohhh that's right. carry on!

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
mattkrick: ahh that must be it. Is it weird that there's no query param when using our UI? e.g. i click "reply" & i goto /responses without a param so i can't share the link. definitely off topic, just sent me down a wormhole

jmtaber129: Yeaaaaa that's something we've talked about on the squad in the past. IIRC there's something non-trivial about keeping the URL in-sync, but I don't quite remember what. One option we discussed was an explicit "share link" button in the discussion drawer to copy the link with the param without the param being in the browser URL. At the end of the day, it's just something we've never prioritized.

Issue #8608, opened 2023-08-02
Post standup responses in end meeting notification thread in Slack


jmtaber129:
When a standup meeting ends, convert standup responses to slack blocks (or tiptap plaintext for initial iteration depending on complexity), and post each standup response in the thread for the end meeting message.

Limit scope to Slack for now.

AC:
- [ ] When a non-empty standup ends, each response is posted in the thread of the "meeting ended" notification.
- [ ] When an empty standup ends, a "no responses submitted" message is posted in the thread.
- [ ] Each response message includes a link directly to the response in the Parabol app

PR #8610, opened 2023-08-02
feat(standups): Post standup responses in 'meeting ended' message thread

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8608

When a standup meeting ends, post the submitted responses in a thread on the "meeting ended" slack message.

Only do this in slack for now to minimize scope.

Needs rebase once https://github.com/ParabolInc/parabol/pull/8607 lands.

## Demo
With responses:
Screen Shot 2023-08-02 at 1 11 51 PM
Without responses:
Screen Shot 2023-08-02 at 1 11 57 PM

## Testing scenarios
- [ ] Start a standup on a team with slack integrated, and fill out a few users' responses, then end the meeting.
- [ ] Confirm that the "meeting ended" message appears, with a message in the thread for each user's response containing the user's name + plaintext response, and a link to the response in the Parabol app
- [ ] Start a standup on the same team, and end the standup without any responses
- [ ] Confirm that the "meeting ended" message appears, with a message in the thread saying that no responses were submitted.
 
## Final checklist

### DIFF
--- a/packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
+++ b/packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
@@ -380,7 +380,11 @@ export const MSTeamsNotifier: Notifier = {
   async endMeeting(dataLoader: DataLoaderWorker, meetingId: string, teamId: string) {
     const {meeting, team} = await loadMeetingTeam(dataLoader, meetingId, teamId)
     if (!meeting || !team) return
-    ;(await getMSTeams(dataLoader, team.id, meeting.facilitatorUserId))?.endMeeting(meeting, team)
+    ;(await getMSTeams(dataLoader, team.id, meeting.facilitatorUserId))?.endMeeting(
+      meeting,
+      team,
+      null
+    )
   },
 
   async startTimeLimit(
--- a/packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
+++ b/packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
@@ -214,9 +214,9 @@ const addStandupResponsesToThread = async (
     standupResponses.map(async ({user, response}) => {
       const options = {
         searchParams: {
-          utm_source: 'slack share',
+          utm_source: 'slack standup summary',
           utm_medium: 'product',
-          utm_campaign: 'sharing',
+          utm_campaign: 'standups',
           responseId: response.id
         }
       }
--- a/packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
+++ b/packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
@@ -417,15 +417,13 @@ export const SlackNotifier: Notifier = {
   async endMeeting(dataLoader: DataLoaderWorker, meetingId: string, teamId: string) {
     const {meeting, team} = await loadMeetingTeam(dataLoader, meetingId, teamId)
     const meetingResponses = await getTeamPromptResponsesByMeetingId(meetingId)
-    const standupResponses: Array<{user: User; response: TeamPromptResponse}> = []
-    await Promise.allSettled(
+    // const standupResponses: Array<{user: User; response: TeamPromptResponse}> = []
+    const standupResponses = await Promise.all(
       meetingResponses.map(async (response) => {
-        const user = await dataLoader.get('users').load(response.userId)
-        if (user) {
-          standupResponses.push({
-            user,
-            response
-          })
+        const user = await dataLoader.get('users').loadNonNull(response.userId)
+        return {
+          user,
+          response
         }
       })
     )
--- a/packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
+++ b/packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
@@ -216,7 +216,7 @@ const addStandupResponsesToThread = async (
         searchParams: {
           utm_source: 'slack standup summary',
           utm_medium: 'product',
-          utm_campaign: 'standups',
+          utm_campaign: 'after-meeting',
           responseId: response.id
         }
       }
--- a/packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
+++ b/packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
@@ -17,10 +17,13 @@ import SlackServerManager from '../../../../utils/SlackServerManager'
 import {DataLoaderWorker} from '../../../graphql'
 import getSummaryText from './getSummaryText'
 import {makeButtons, makeSection, makeSections} from './makeSlackBlocks'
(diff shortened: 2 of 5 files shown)


packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
```diff
@@ -304,9 +405,22 @@ export const SlackNotifier: Notifier = {
 
   async endMeeting(dataLoader: DataLoaderWorker, meetingId: string, teamId: string) {
     const {meeting, team} = await loadMeetingTeam(dataLoader, meetingId, teamId)
+    const meetingResponses = await getTeamPromptResponsesByMeetingId(meetingId)
+    const standupResponses: Array<{user: User; response: TeamPromptResponse}> = []
+    Promise.allSettled(
+      meetingResponses.map(async (response) => {
+        const user = await dataLoader.get('users').load(response.userId)
```
jmtaber129, line 412: We _could_ do a single `loadMany` with all the userIds, but `DataLoader` is good at batching, and these userIds seem to already be in the dataloader cache when a meeting ends here anyway, so this seems like the best way to connect users and their responses.

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
```diff
@@ -173,6 +187,68 @@ const makeStartMeetingNotificationLookup: Record<
...
+    return 'success'
+  }
+
+  const results = await Promise.all(
+    standupResponses.map(async ({user, response}) => {
+      const options = {
+        searchParams: {
+          utm_source: 'slack standup summary',
+          utm_medium: 'product',
+          utm_campaign: 'standups',
```
jmtaber129, line 219: @tianrunhe PTAL at these UTMs!

tianrunhe: Aviva [suggested](https://parabol.slack.com/archives/C03C3J42WDU/p1691109288833119?thread_ts=1691095523.822949&cid=C03C3J42WDU) to change the campaign to `after-meeting`

jmtaber129: @tianrunhe updated!

tianrunhe (approved):
LGTM from metrics side!

packages/client/modules/teamDashboard/components/ProviderRow/SlackNotificationRow.tsx
```diff
@@ -25,7 +25,8 @@ const labelLookup = {
   meetingStart: 'Meeting Start',
   MEETING_STAGE_TIME_LIMIT_END: `Meeting ${MeetingLabels.TIME_LIMIT} Ended`,
   MEETING_STAGE_TIME_LIMIT_START: `Meeting ${MeetingLabels.TIME_LIMIT} Started`,
-  TOPIC_SHARED: `Topic Shared`
+  TOPIC_SHARED: `Topic Shared`,
+  STANDUP_RESPONSE_SUBMITTED: 'Standup Response Submitted'
 } as Record<SlackNotificationEventEnum, string>
```
BartoszJarocki, line 30: +1 not related but it'd be nice to improve types along the way and change that to specifying a type instead of casting, it'll actually be typesafe then
```
const labelLookup: Record<SlackNotificationEventEnum, string> = {
  meetingEnd: 'Meeting End',
  meetingStart: 'Meeting Start',
  MEETING_STAGE_TIME_LIMIT_END: `Meeting ${MeetingLabels.TIME_LIMIT} Ended`,
  MEETING_STAGE_TIME_LIMIT_START: `Meeting ${MeetingLabels.TIME_LIMIT} Started`,
  TOPIC_SHARED: `Topic Shared`
}
```

packages/server/graphql/mutations/helpers/notifications/MSTeamsNotifier.ts
```diff
@@ -339,6 +340,11 @@ export const MSTeamsNotificationHelper: NotificationIntegrationHelper<MSTeamsNot
     const adaptiveCard = JSON.stringify(card.toJSON())
     const attachments = MakeACAttachment(adaptiveCard)
     return notifyMSTeams('meetingEnd', webhookUrl, userId, teamId, attachments)
+  },
+
+  async standupResponseSubmitted() {
```
BartoszJarocki, line 345: do we want to create tasks for that to cover it in the future?

BartoszJarocki (approved):
code looks good! In the future it'd be nice to make it easier to discover this feature, right now it's pretty hidden and you need to know where to look for it. Also, not sure if it's possible but we use in our responses rich content quite a lot, and it'll look like 
Screenshot 2023-08-09 at 09 28 50

so that's something to investigate too

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
mattkrick: 💯 if you wanted to make it a pure function you could use `loadNonNull` since user is guaranteed

```ts
const standupResponses = await Promise.all(
  meetingResponses.map(async (response) => {
    const user = await dataLoader.get('users').loadNonNull(response.userId)
    return {user, response}
  })
)
```

mattkrick (approved):

packages/server/graphql/mutations/helpers/notifications/SlackNotifier.ts
```diff
@@ -304,9 +416,22 @@ export const SlackNotifier: Notifier = {
 
   async endMeeting(dataLoader: DataLoaderWorker, meetingId: string, teamId: string) {
     const {meeting, team} = await loadMeetingTeam(dataLoader, meetingId, teamId)
+    const meetingResponses = await getTeamPromptResponsesByMeetingId(meetingId)
+    const standupResponses: Array<{user: User; response: TeamPromptResponse}> = []
```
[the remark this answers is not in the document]
jmtaber129: I started attempting to pull out a separate `endStandupMeeting` method, but with the existing interfaces this gets kinda hairy, since we'd need a separate `endStandupMeeting` on _all_ `NotificationIntegrationHelper`s, which for MS teams + mattermost means either 
a) a `NotificationIntegrationHelper.endStandupMeeting` identical to `endMeeting`, or
b) a `Notifier.endStandupMeeting` that calls `NotificationIntegrationHelper.endMeeting` under the hood, and a `NotificationIntegrationHelper.endStandupMeeting` that throws an `unimplemented` error
neither of which seems much better.

I agree that a standup-specific arg being passed to a generic method isn't ideal, but overall, this feature is very experimental, so I'd rather not put too much effort into doing this the "right" way until we start iterating. I think when we add this feature to mattermost + ms teams we'll do whatever refactor makes the most sense here.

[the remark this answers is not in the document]
mattkrick: makes sense! as long as we budget some time to fix it up once it's no longer experimental i'm good to go

Issue #8602, opened 2023-08-01
Update facilitator for standup meeting if original facilitator is no longer in the team


mwermuth:
Related to #8470 

In order to fix empty standup summary emails when a facilitator exits the team we need to add the following:

When the facilitator of the meeting is no longer in a given team, as a result null is returned from newMeeting query used in MeetingSummaryEmailRootSSRQuery.

potential fix:

1. while removing team member, check if given team member is a facilitator of any active standup meeting; if yes, update it and default to team lead
2. manually update facilitatorUserId for already affected meetings

PR #8611, opened 2023-08-02
fix: Promote new facilitator when facilitator leaves the team

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8602

When a user is removed from a team on which they're facilitating meetings, reassign one of the meeting members (the first one returned by the DB for simplicity) to be the facilitator. Do nothing for meetings in which the user is the only meeting member.

## Testing scenarios
- [ ] Start a meeting with users A and B, where user A is the facilitator
- [ ] As user A, leave the team
- [ ] Confirm user B is now the facilitator of the meeting
- [ ] Confirm reasonable behavior for other cases:
  - [ ] User is facilitating no meetings
  - [ ] User is facilitating a meeting with no other meeting members

## Final checklist


igorlesnenko (approved):
Works!

Should we create migration that updates old meetings?

jmtaber129:
> Should we create migration that updates old meetings?

We _could_ but I don't think it's worth the effort right now, especially when you factor in the risk of doing updates across a ton of meetings - we can do these updates on a case-by-case basis if we get support messages about this.

PR #8612, opened 2023-08-02
feat(tiptap): Loom embed extension

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8659

Implement a read-only Loom extension - when loom links are present in a standup response, add a Loom TipTap node immediately after the paragraph containing the link.

Because this is read-only, other users will always see the expanded Loom embeds, while the author will only see them in closed meetings and the discussion drawer.

Do nothing for summaries, though we could leverage Loom gif thumbnails if we wanted to.

## Demo
What the author sees:
Screen Shot 2023-08-15 at 1 12 15 PM

What the reader sees:
Screen Shot 2023-08-15 at 1 12 08 PM

Discussion drawer (both author and readers see this):
Screen Shot 2023-08-15 at 1 15 38 PM

## Testing scenarios
- [ ] Write a response with non-loom and loom links, and check the link expansion in various scenarios (bullets, quotes, etc.)
- [ ] Smoke test responses.

## Final checklist

### DIFF
--- a/packages/client/components/promptResponse/loomExtension.ts
+++ b/packages/client/components/promptResponse/loomExtension.ts
@@ -1,13 +1,61 @@
-import {Node, nodeInputRule, nodePasteRule} from '@tiptap/core'
+import {Node} from '@tiptap/core'
+import {JSONContent} from '@tiptap/react'
 
 const LOOM_REGEX = /^(?:https:\/\/)?(?:www\.)?loom\.com\/share\/[a-zA-Z0-9]*\?.*$/g
+
+export const unfurlLoomLinks = (rawContent: JSONContent): JSONContent => {
+  if (!rawContent.content) {
+    return rawContent
+  }
+
+  const newContent: JSONContent[] = []
+  rawContent.content.forEach((content) => {
+    if (content.type === 'paragraph') {
+      newContent.push(content, ...getLoomNodesForParagraph(content))
+    } else {
+      newContent.push(unfurlLoomLinks(content))
+    }
+  })
+
+  return {
+    type: rawContent.type,
+    content: newContent
+  }
+}
+
+const getLoomNodesForParagraph = (content: JSONContent) => {
+  const distinctLoomLinks: Record<string, string> = {}
+  content.content?.forEach((subContent) => {
+    if (subContent.type === 'text') {
+      const link = subContent.marks?.find((mark) => mark.type === 'link')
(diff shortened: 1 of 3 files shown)


BartoszJarocki (approved):

packages/client/components/promptResponse/loomExtension.ts
```diff
@@ -0,0 +1,89 @@
...
+  })
+
+  return {
+    type: rawContent.type,
+    content: newContent
+  }
+}
+
+const getLoomNodesForParagraph = (content: JSONContent) => {
+  const distinctLoomLinks: Record<string, string> = {}
```
mattkrick, line 27: +1 since this is a local var, if i hit enter to create a new paragraph & paste the same link again, it's going to unfurl the video again. is this intended? 

fwiw i saw that slack messages will do a big unfurl for 1 link, 2 small unfurls for 2 links, and they won't unfurl for 3 links. thought that was kinda cool to prevent it from getting too long! (e.g. thumbnail + title for 1, just thumbnail for 2, just links for 3+). i saw loom has the /oembed path for metadata like title/description if you feel like playing with it

packages/client/components/promptResponse/loomExtension.ts
```diff
@@ -0,0 +1,89 @@
+import {Node} from '@tiptap/core'
+import {JSONContent, mergeAttributes} from '@tiptap/react'
+
+const LOOM_REGEX = /^(?:https:\/\/)?(?:www\.)?loom\.com\/share\/[a-zA-Z0-9]*\?.*$/g
```
mattkrick, line 4: -1 i don't know how i managed to get an insecure http link while copying & pasting across task cards, but somehow i lost the s & it didn't unfurl. can we use `https?` instead?

Alternatively, I found the loom official regex, which includes their staging URLs 🤣 . Could use that instead https://www.unpkg.com/browse/@loomhq/loom-embed@1.2.4/dist/esm/common.js

mattkrick (approved):

packages/client/components/promptResponse/loomExtension.ts
jmtaber129: Ooo yes, nice catch!

Including the staging URLs doesn't feel quite right to me 😂 but I made some minor modifications to our regex based on this, so nice find!

PR #8617, opened 2023-08-03
chore: Cleanup old starter tier organizations

### LINKED ISSUE #8507: Update orgs on the deprecated personal tier
@tianrunhe shared [here](https://parabol.slack.com/archives/C608QL4RH/p1689201394497069) that there are some orgs hat are on the `personal` tier. Matt suggested that:

> My guess is the migration to update the names ran while the server was still issuing the old names, so any orgs created while the migration ran would get the old name

### AC

- [ ] Create a migration that moves the old `personal` tier orgs to the new `team` tier

tghanken:
Fixes #8507 

When we changed the names of our plan tiers in #7505, there were two organizations created while the migration `20221220161224-updateTier.ts` was running that are left with the old `personal` tier name.  This migration cleans up those tier values, and migrates them to the new `starter` tier name.

### DIFF
--- a/packages/server/database/migrations/20230801200811-fixOrganizationTier.ts
+++ b/packages/server/database/migrations/20230801200811-fixOrganizationTier.ts
@@ -1,0 +1,9 @@
+import {R} from 'rethinkdb-ts'
+
+export const up = async function (r: R) {
+  r.table('Organization').filter(r.row('tier').eq('personal')).update({tier: 'starter'}).run()
+}
+
+export const down = async function (r: R) {
+  // noop
+}
--- a/packages/server/database/migrations/20230801200811-fixOrganizationTier.ts
+++ b/packages/server/database/migrations/20230801200811-fixOrganizationTier.ts
@@ -1,9 +1,0 @@
-import {R} from 'rethinkdb-ts'
-
-export const up = async function (r: R) {
-  r.table('Organization').filter(r.row('tier').eq('personal')).update({tier: 'starter'}).run()
-}
-
-export const down = async function (r: R) {
-  // noop
-}
--- a/packages/server/postgres/migrations/1691422128820_fixOrganizationTier.ts
+++ b/packages/server/postgres/migrations/1691422128820_fixOrganizationTier.ts
@@ -1,0 +1,18 @@
+import {r} from 'rethinkdb-ts'
+import connectRethinkDB from '../../database/connectRethinkDB'
+
+export async function up() {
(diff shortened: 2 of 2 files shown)


packages/server/database/migrations/20230801200811-fixOrganizationTier.ts
```diff
@@ -0,0 +1,9 @@
+import {R} from 'rethinkdb-ts'
+
+export const up = async function (r: R) {
+  r.table('Organization').filter(r.row('tier').eq('personal')).update({tier: 'starter'}).run()
+}
+
+export const down = async function (r: R) {
+  // noop
```
tghanken, line 8: Didn't implement a `down` function, as any cleanup should be handled by `20221220161224-updateTier.ts`.

nickoferrall (approved):
As this is scanning through all Organizations, and Rethinkdb is so brittle, we might need to batch the query. I'll remove the One Review Required label and tag Matt to see what he thinks.

nickoferrall (changes requested):
As this is scanning through all Organizations, and Rethinkdb is so brittle, we might need to batch the query. I'll remove the One Review Required label and tag Matt to see what he thinks.

Before doing that, it looks like the Build & PR Title checks are failing. The build says:
```
    if (files.length !== 154) throw new Error('New migrations must live in the postgres/migrations directory')
```

although this is a Rethinkdb migration 🤔 Once those are fixed, we can move this to Maintainer Review.

igorlesnenko:
@nickoferrall now we are storing all rethink migrations as PG migrations. Here is the example:
https://github.com/ParabolInc/parabol/blob/7991c595c31b973dbbf621b5b86bdbaa0fd994c7/packages/server/postgres/migrations/1676341596550_updateOrgIdOnSAML.ts

tghanken:
> As this is scanning through all Organizations, and Rethinkdb is so brittle, we might need to batch the query. I'll remove the One Review Required label and tag Matt to see what he thinks.

Interesting!  My intuition from using other DB systems is that with a highly selective predicate, it's better to just issue one query and let the query planner handle it, but happy to batch this if it makes sense!  Alternatively I could filter for the two orgIds we care about, but that makes this somewhat less flexible.

nickoferrall (approved):
LGTM! I'm probably overly anxious and it doesn't need to be batched, but I think it's worth going through Maintainer Review to be sure

mattkrick:
LGTM!

Issue #8621, opened 2023-08-07
Increase the number of suggested groups


nickoferrall:
Taylor [mentioned](https://parabol.slack.com/archives/C042E8PGPD4/p1691082414125569):

> Terrible Idea: In our growth retro, we had a ton of reflections and tried suggest groups, but they weren't very helpful.  I wonder if it makes sense for us to try to tweak the prompt to focus more on many smaller but high-intent groups, which could then serve as a jumping off point for users to create the final groups.

### AC

- [ ] Increase the number of suggested groups
- [ ] Test with the reflections from the [Growth Retro](https://action.parabol.co/meet/p4lomnpkSQ/discuss/1) and see if the results are useful

PR #8625, opened 2023-08-07
chore(suggest-groups): increase number of suggested groups

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8621

Here's a long-winded demo where I look through the reflection groups from the recent Growth Retro: https://www.loom.com/share/41e45efd828d49378db820582d6b4423

### To test

- [ ] Create a couple of Retros, click Suggest Groups, and get suggested groups that are at least as useful as the previous prompt

### DIFF
--- a/packages/server/utils/OpenAIServerManager.ts
+++ b/packages/server/utils/OpenAIServerManager.ts
@@ -109,11 +109,12 @@ class OpenAIServerManager {
   async generateThemes(reflectionsText: string[]) {
     if (!this.openAIApi) return null
     const suggestedThemeCountMin = Math.floor(reflectionsText.length / 5)
-    const suggestedThemeCountMax = Math.floor(reflectionsText.length / 4)
+    const suggestedThemeCountMax = Math.floor(reflectionsText.length / 3)
+
     // Specify the approximate number of themes as it will often create too many themes otherwise
     const prompt = `Create a short list of common themes given the following reflections: ${reflectionsText.join(
       ', '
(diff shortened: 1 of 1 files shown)


tghanken:
@nickoferrall Did you intend to link a demo here?

nickoferrall:
Sorry @tghanken, PR description updated!

tghanken (approved):
Safe to merge as is, I'm getting results at least as good as our prior implementation.

+1 I'd like to see us encourage the AI to not group a reflection where there's a low certainty.  Ex. add some logic to [getThemeForReflection](https://github.com/ParabolInc/parabol/blob/bb481d34338f61ab3758dd699a16b2fb49aadbe7/packages/server/utils/OpenAIServerManager.ts#L151) to allow for a return value that indicates no group should be created for the reflection.

Ex:
```
const prompt = `Given the themes [...] themes I've provided. [NEW] If the reflection does not fit any of the themes, classify the reflection as "noThemeMatch".`
```

nickoferrall:
Originally, I didn't specify how many groups and asked it to only group when they're similar, but it usually results in too many groups IMO.


> If the reflection does not fit any of the themes, classify the reflection as "noThemeMatch".

Here's the results I got after adding the suggested line ^:

Screenshot 2023-08-14 at 15 12 12


I'll merge this PR as it is so that we can get the number of groups improvement. I'd definitely like to spend more time improving the prompt as I'm sure it could be better.

Issue #8561, opened 2023-07-24
One on one: create ad-hoc team on meeting start


igorlesnenko:
**AC:**

- When on-one-one meeting is started create a new team with automatically generated name, something like `Your name / Another person name`
- ~The team should have `autocreated` flag~
- If previous ~`autocreated`~ team with exact these set of members exists, reuse the team
- Send invite to the selected user or email

**Keep in mind**

We will use same backend for ad-hoc teams, so in future it will also have an ability to add multiple users, emails, teams (No need to implement this for now)

igorlesnenko:
I decided to not add `autocreated` and just reuse any kind of existing team

Issue #8562, opened 2023-07-24
One on one: show team exists warning


igorlesnenko:
<img width="249" alt="image" src="https://github.com/ParabolInc/parabol/assets/466991/cbdce69d-8ae8-4420-91cc-49ba4532318d">

**AC:**

- If a new team will be created show — "a new team will be created for you and test@example.com"
- If a team with exact same members exists within the org, show a message that the team "Teamname" will be used cc @acressall

PR #8629, opened 2023-08-08
feat(one-on-one): allow oneOnOne input in startCheckIn mutation

igorlesnenko:
Fixes: #8561

**How to test:**

Enable org `oneOnOne` feature flag.

Edit this file to change variables

https://github.com/ParabolInc/parabol/blob/a3d5b02e6cea504d30136fa395b852d1d59744e0/packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx#L153-L156

1) Change variables to set unknown email, start the `one on one` meeting

```
 StartCheckInMutation(
        atmosphere,
        {gcalInput, oneOnOneTeamInput: {
            email: 'igor+unknown@parabol.co',
            orgId: 'oFJQVmaSE8'
          }},
        {history, onError, onCompleted}
      )
```

- [ ] See that a new team was created and meeting was started
- [ ] See that invite was sent

2) Accept the invite, end the meeting. Keep the same code, start a new meeting again

- [ ] See that the team was reused, no new team was created

3) Test the same cases with an email of existing user

4) See that you still can create a meeting with `teamId`


packages/server/graphql/mutations/startCheckIn.ts
```diff
@@ -17,35 +17,62 @@ import createGcalEvent from './helpers/createGcalEvent'
...
   async resolve(
     _source: unknown,
-    {teamId, gcalInput}: {teamId: string; gcalInput?: CreateGcalEventInputType},
-    {authToken, socketId: mutatorId, dataLoader}: GQLContext
+    {
+      teamId: existingTeamId,
+      gcalInput,
+      oneOnOneTeamInput
+    }: {
+      teamId: string
```
nickoferrall, line 49: -1 `teamId` can now be undefined

packages/server/graphql/mutations/startCheckIn.ts
```diff
@@ -17,35 +17,62 @@ import createGcalEvent from './helpers/createGcalEvent'
...
     const r = await getRethink()
+    const {authToken, socketId: mutatorId, dataLoader} = context
     const operationId = dataLoader.share()
     const subOptions = {mutatorId, operationId}
     // AUTH
     const viewerId = getUserId(authToken)
-    if (!isTeamMember(authToken, teamId)) {
-      return standardError(new Error('Team not found'), {userId: viewerId})
+
+    let teamId = existingTeamId
```
nickoferrall, line 62: +1 I find the let a bit hard to follow. 

The first if check could check `existingTeamId` to help avoid `let`. 

Something like: 

```
    if (existingTeamId) {
      if (!isTeamMember(authToken, existingTeamId)) {
      }
      const unpaidError = await isStartMeetingLocked(existingTeamId, dataLoader)
      if (unpaidError) return standardError(new Error(unpaidError), {userId: viewerId})
    }
    const viewer = await dataLoader.get('users').loadNonNull(viewerId)
    const teamId = oneOnOneTeamInput ? await maybeCreateOneOnOneTeam(viewer, oneOnOneTeamInput, context) : existingTeamId
    if (!teamId) {
      return standardError(new Error('Must provide teamId or ad-hoc team'), {userId: viewerId})
    }
```

But that's just personal preference. Maybe you find that is harder to read.

packages/server/graphql/mutations/startCheckIn.ts
```diff
@@ -17,35 +17,62 @@ import createGcalEvent from './helpers/createGcalEvent'
...
+      if (!isTeamMember(authToken, teamId)) {
+        return standardError(new Error('Team not found'), {userId: viewerId})
+      }
+      const unpaidError = await isStartMeetingLocked(teamId, dataLoader)
+      if (unpaidError) return standardError(new Error(unpaidError), {userId: viewerId})
+    } else if (oneOnOneTeamInput) {
+      const viewer = await dataLoader.get('users').loadNonNull(viewerId)
+      teamId = await maybeCreateOneOnOneTeam(viewer, oneOnOneTeamInput, context)
+    } else {
+      return standardError(new Error('Must provide teamId or ad-hoc team'), {userId: viewerId})
```
nickoferrall, line 74: +1 nit but they would be providing `oneOnOneTeamInput` rather than an `ad-hoc team`

packages/server/graphql/mutations/helpers/maybeCreateOneOnOneTeam.ts
```diff
@@ -0,0 +1,155 @@
...
+  await inviteToTeamHelper(invitees, teamId, undefined, context)
+
+  return teamId
+}
+
+const generateOneOnOneTeamName = (firstUserName: string, secondUserName: string) => {
+  return `${firstUserName} / ${secondUserName}`
+}
+
+const maybeCreteOneOnOneTeamFromEmail = async (
```
nickoferrall, line 96: +1 typo here: `Create` rather than `Crete`

packages/server/graphql/mutations/helpers/maybeCreateOneOnOneTeam.ts
```diff
@@ -0,0 +1,155 @@
...
+const createNewTeamAndInvite = async (
+  viewer: IUser,
+  name: string,
+  orgId: string,
+  invitees: string[],
+  context: GQLContext
+) => {
+  const {authToken} = context
+  const teamId = generateUID()
+  await createTeamAndLeader(viewer!, {id: teamId, isOnboardTeam: false, name, orgId})
```
nickoferrall, line 80: +1 can remove the `!`?

packages/server/graphql/public/types/CreateOneOnOneTeamInput.ts
```diff
@@ -0,0 +1,24 @@
+import {GraphQLInputObjectType, GraphQLNonNull, GraphQLString} from 'graphql'
+
+const CreateOneOnOneTeamInput = new GraphQLInputObjectType({
+  name: 'CreateOneOnOneTeamInput',
+  fields: () => ({
+    userId: {
```
nickoferrall, line 6: What's the use case where we'll provide a `userId` but not an `email`?

packages/server/graphql/mutations/helpers/maybeCreateOneOnOneTeam.ts
```diff
@@ -0,0 +1,155 @@
...
+
+const maybeCreteOneOnOneTeamFromEmail = async (
+  viewer: IUser,
+  email: string,
+  orgId: string,
+  context: GQLContext
+) => {
+  const teamName = generateOneOnOneTeamName(viewer.preferredName, email.split('@')[0]!)
+
+  const existingUsers = await getUsersByEmails([email])
```
nickoferrall, line 104: -1 let's use the `getUserByEmail` function that lives in the same file

packages/server/graphql/mutations/helpers/maybeCreateOneOnOneTeam.ts
```diff
@@ -0,0 +1,155 @@
...
+  return pg
+    .selectFrom('Team')
+    .selectAll()
+    .where('id', 'in', existingTeamIds)
+    .where('orgId', '=', orgId)
+    .where('isArchived', '=', false)
+    .execute()
+}
+
+const getExistingExactOneOnOneTeam = async (
```
nickoferrall, line 62: +1 why `exact` in the name rather than `getExistingOneOnOneTeam`?

packages/server/graphql/mutations/helpers/maybeCreateOneOnOneTeam.ts
```diff
@@ -0,0 +1,155 @@
...
+
+export default async function maybeCreateOneOnOneTeam(
+  viewer: IUser,
+  oneOnOneTeam: OneOnOneTeam,
+  context: GQLContext
+) {
+  const {email, userId, orgId} = oneOnOneTeam
+
+  if (email) {
+    return maybeCreteOneOnOneTeamFromEmail(viewer, email, orgId, context)
```
nickoferrall, line 149: -1 I feel like this file could be simplified by merging `maybeCreteOneOnOneTeamFromEmail` and `maybeCreateOneOnOneTeamFromUserId`.

It looks like the main difference is just getting the user by the userId or the email, which could be done as a ternary statement, like `const user = userId ? await getUserById(userId) : await getUsersByEmail(email!)`.

nickoferrall (changes requested):
Tested and it's working for me!

Two notes:

1. Will we let users know in the UI that a new team will be created? They might be surprised that a Check-in creates a new team, but none of the other meetings do.
2. I need to refresh the page to see the newly created team at the moment

packages/server/graphql/public/types/CreateOneOnOneTeamInput.ts
igorlesnenko: I was thinking that when you select existing user in the user picker, we will provide a userId, however, I think I can simplify and just provide emails

packages/server/graphql/mutations/helpers/maybeCreateOneOnOneTeam.ts
igorlesnenko: removed user option completely

igorlesnenko:
> Will we let users know in the UI that a new team will be created? They might be surprised that a Check-in creates a new team, but none of the other meetings do.

We will show a hint, here is the separate issue for that: https://github.com/ParabolInc/parabol/issues/8562

> I need to refresh the page to see the newly created team at the moment 

Should work now

packages/server/graphql/mutations/helpers/maybeCreateOneOnOneTeam.ts
```diff
@@ -0,0 +1,124 @@
...
+  viewer: IUser,
+  oneOnOneTeam: OneOnOneTeam,
+  context: GQLContext
+) {
+  const {email, orgId} = oneOnOneTeam
+
+  const existingUser = await getUserByEmail(email)
+
+  if (!existingUser) {
+    const teamName = generateOneOnOneTeamName(viewer.preferredName, email.split('@')[0]!)
```
nickoferrall, line 112: +1 NIT: these two lines are the same as the 122-123 except for the second arg. Perhaps `generateOneOnOneTeamName` could handle the email split logic? It could take an object as an arg with `firstUserName: string`, `secondUserName?: string`, `secondUserEmail?: string`?

nickoferrall (approved):

packages/server/graphql/mutations/startCheckIn.ts
```diff
@@ -17,35 +17,64 @@ import createGcalEvent from './helpers/createGcalEvent'
...
+
+    if (existingTeamId) {
+      if (!isTeamMember(authToken, existingTeamId)) {
+        return standardError(new Error('Team not found'), {userId: viewerId})
+      }
+      const unpaidError = await isStartMeetingLocked(existingTeamId, dataLoader)
+      if (unpaidError) return standardError(new Error(unpaidError), {userId: viewerId})
+    }
+    const viewer = await dataLoader.get('users').loadNonNull(viewerId)
+    const teamId = oneOnOneTeamInput
```
Dschoordsch, line 70: -1 Please check that exactly one of `teamId`/`oneOnOneTeamInput` is set

packages/server/graphql/mutations/helpers/maybeCreateOneOnOneTeam.ts
```diff
@@ -0,0 +1,124 @@
...
+    .without('reduction')
+    // Get all team members for each team
+    .eqJoin('group', r.table('TeamMember'), {index: 'teamId'})
+    .zip()
+    .group('teamId')
+    .pluck('userId')
+    .ungroup()
+    // Select only teams with exact given number of team members
+    .filter((row: RValue) => {
+      return row('reduction').count().eq(userIds.length)
```
Dschoordsch, line 42: -1 didn't we check this in l32 already?

packages/server/graphql/mutations/helpers/maybeCreateOneOnOneTeam.ts
```diff
@@ -0,0 +1,124 @@
...
+  orgId: string,
+  invitees: string[],
+  context: GQLContext
+) => {
+  const {authToken, socketId: mutatorId, dataLoader} = context
+  const {id: viewerId} = viewer
+  const operationId = dataLoader.share()
+  const subOptions = {mutatorId, operationId}
+  const teamId = generateUID()
+  await createTeamAndLeader(viewer, {id: teamId, isOnboardTeam: false, name, orgId})
```
Dschoordsch, line 78: +1 I think we should store the fact that it's a 1:1 team as well, so we can query and present it differently in the future.

Dschoordsch (changes requested):
Not tested yet

packages/server/graphql/public/types/CreateOneOnOneTeamInput.ts
```diff
@@ -0,0 +1,20 @@
...
+  name: 'CreateOneOnOneTeamInput',
+  fields: () => ({
+    email: {
+      type: new GraphQLNonNull(GraphQLString)
+    },
+    orgId: {
+      type: new GraphQLNonNull(GraphQLString)
+    }
+  })
+})
```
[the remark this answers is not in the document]
igorlesnenko: I think I still need it to use in legacy `startCheckIn.ts`

packages/server/graphql/mutations/helpers/maybeCreateOneOnOneTeam.ts
Dschoordsch: -1 Now that we store whether or not a team is a 1:1 team, we can use that here as well. We can just get a list of `teamId`s where all `userId`s are present and then just check in PG whether the team is a 1:1 team. This way we don't need to count at all. (These big rethink queries scare me)

Dschoordsch (changes requested):

packages/server/graphql/mutations/helpers/maybeCreateOneOnOneTeam.ts
Dschoordsch: For now I would ignore those 2 cases. If we see decent 1:1 usage, I would later restrict invites for those in the birthday cake version.
I'm open to discuss this.

igorlesnenko: @Dschoordsch I see nothing bad would happen if I ignore these cases for now, so I updated the PR.

Dschoordsch (approved):

Issue #8628, opened 2023-08-08
Checkout Flow: Add visual cues to upgrade flow to prevent confusion


tghanken:
We're seeing some users experience [frustration](https://parabol.slack.com/archives/C042E8PGPD4/p1691445086645969?thread_ts=1691132968.111179&cid=C042E8PGPD4) 🔒  with the new checkout flow that is causing them to abandon their upgrade attempt.

When clicking on the `Team` tier, nothing interactive happens within the interface and the `Starter` plan remains highlighted, causing users to believe the checkout flow is broken, even though they can upgrade using the credit card details section.

![no_upgrage_interactivity](https://github.com/ParabolInc/parabol/assets/51011434/db9331fb-9cbd-4576-9042-a77f0f2e2225)

### Possible Solutions
- Hide credit card module until user selects the `Team` plan
- Highlight the `Enterprise` or `Team` plans while the user remains on the billing page if clicked, reset to true organization tier after leaving the page and revisiting.

### Acceptance Criteria
Users can:
   - Get visual feedback while moving through the checkout flow

**Estimated effort:** X points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

acressall:
[This](https://www.figma.com/file/SyJdkzumWsUJo58cmBGgus/Flow---Checkout?type=design&node-id=216-4562&mode=design&t=pxXJIo1ca61GKUqr-4) is what it's meant to look like after click
![plans](https://github.com/ParabolInc/parabol/assets/18582225/a9a8e215-a72a-4bf4-99da-fcc09b52fe81)

PR #8637, opened 2023-08-09
chore: active plan visual cue

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8628

Loom demo: https://www.loom.com/share/bbfd53b43b59418eb152321be0023a14

### To test

- [ ] Click "Select Plan" and see that the plan's border is now outlined, indicating that it's active


acressall (changes requested):
See comment on Loom

igorlesnenko:
Agree, we need to change button text too, maybe even a color

nickoferrall:
Makes sense! Here's the update: https://www.loom.com/share/7a4f3ccd980e4c2d852ee5105ba7f24d

acressall (approved):

igorlesnenko (approved):

PR #8641, opened 2023-08-09
fix(al): Number of activities exceeds number requested

jmtaber129:
# Description
see [slack](https://parabol.slack.com/archives/C042HTVBH8D/p1691596459048329) 🔒 

If a user has access to more than 200 activities, we won't return all of their activities to the client. Sometimes, this means that even basic activities like check-ins won't be shown on the client.

Bump up the max number of activities we return to 2000, and send an error to sentry if we ever see more than 1000 activities.

Also add an `extras` field to our sentry util for capturing additional non-tag context (in this case, number of activities seen).

## Testing scenarios
Smoke test AL.

## Final checklist

### DIFF
--- a/packages/client/components/ActivityLibrary/ActivityLibrary.tsx
+++ b/packages/client/components/ActivityLibrary/ActivityLibrary.tsx
@@ -74,7 +74,7 @@ graphql`
 const query = graphql`
   query ActivityLibraryQuery {
     viewer {
-      availableTemplates(first: 200) @connection(key: "ActivityLibrary_availableTemplates") {
+      availableTemplates(first: 2000) @connection(key: "ActivityLibrary_availableTemplates") {
         edges {
           node {
             ...ActivityLibrary_template @relay(mask: false)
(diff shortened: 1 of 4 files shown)


packages/client/components/ActivityLibrary/ActivityLibrary.tsx
```diff
@@ -74,7 +74,7 @@ graphql`
 const query = graphql`
   query ActivityLibraryQuery {
     viewer {
-      availableTemplates(first: 200) @connection(key: "ActivityLibrary_availableTemplates") {
+      availableTemplates(first: 2000) @connection(key: "ActivityLibrary_availableTemplates") {
```
mattkrick, line 77: -1 can you also increase it here? https://github.com/ParabolInc/parabol/blob/badbd0ffd7440cb43ad45fef79c9b40c09fa82a6/packages/client/components/ActivityLibrary/CreateNewActivity/CreateNewActivity.tsx#L124

mattkrick (changes requested):

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
jmtaber129: Fixed!

jmtaber129:
Merging per approval in slack - https://parabol.slack.com/archives/C042HTVBH8D/p1691603460315079?thread_ts=1691603228.853929&cid=C042HTVBH8D

PR #8648, opened 2023-08-10
chore: fix a couple vulns

mattkrick:
# Description

Fixes 2 vulns: fast-xml-parser, and vm2.
We can't remove vm2 entirely since pm2 ultimately depends on it.
We did remove it from the build step by bumping the graphql-tools (the prisma extension used it & the new version no longer requires that).


packages/server/graphql/public/typeDefs/_legacy.graphql
```diff
@@ -6002,7 +6002,7 @@ type Mutation {
     The name of the new team
     """
     orgName: String!
-    invitees: [String!]
+    invitees: [Email!]
```
mattkrick, line 6005: the bumped version finally catches this error

mattkrick:
Nothing to test, but CI & local smoke test pass

PR #8652, opened 2023-08-11
fix: fix share topic modal can't see slack integration

igorlesnenko:
Looks like for some users share topic modal can't see the slack integration added

This should probably fix it


igorlesnenko:
@nickoferrall here is the slack thread: https://parabol.slack.com/archives/C042E8PGPD4/p1691614291172849

nickoferrall (approved):

Issue #8531, opened 2023-07-17
Gcal Integration: invite team members from modal 


nickoferrall:
In the gcal modal, you can currently add the title, description, and time of the meeting. You can click a checkbox to invite all team members, but users should be able to choose which team members to invite.

I suggest that we build something very similar to the way we invite users on team creation: https://github.com/ParabolInc/parabol/pull/8482

### AC

- [ ] The gcal modal includes an input where users can specify which emails they want to invite to the event
- [ ] Users can add all of the team members' email addresses to the input by clicking a checkbox

PR #8662, opened 2023-08-14
feat(gcal): invite members from modal

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8531

Loom demo: https://www.loom.com/share/2b0cc1aefb38479fb7e9cbc94a7c6ae6

The "invite team members" checkbox could be on by default, but I think making the user check it is best. It would be a bad UX if they didn't want to invite team members and they accidentally emailed everyone in their team. 

### To test

- [ ] Add the gcal user feature flag
- [ ] Add the gcal integration for your user on a team
- [ ] On that team, open the gcal modal, add emails and see that the email addresses are added to the gcal event
- [ ] Click the checkbox and see that all team members in the team are added
- [ ] The "add team members" checkbox doesn't exist if you're the only user in the team


packages/server/graphql/public/types/CreateGcalEventInput.ts
```diff
@@ -25,13 +25,14 @@ const CreateGcalEventInput = new GraphQLInputObjectType({
...
-    inviteTeam: {
-      type: new GraphQLNonNull(GraphQLBoolean),
-      description: 'Whether the users in the team should be invited in gcal'
-    },
     timeZone: {
       type: new GraphQLNonNull(GraphQLString),
       description: 'The timeZone of the meeting'
+    },
+    invitees: {
+      type: new GraphQLList(GraphQLString),
```
igorlesnenko, line 33: Should we do it like this: `new GraphQLList(new GraphQLNonNull(GraphQLEmailType)))` ?

packages/client/modules/userDashboard/components/GcalModal/GcalModal.tsx
```diff
@@ -93,6 +117,52 @@ const GcalModal = (props: Props) => {
...
+      const lastValidEmail = uniqueInvitees[uniqueInvitees.length - 1]
+      lastValidEmail
+        ? setInviteError(`Invalid email(s) after ${lastValidEmail}`)
+        : setInviteError(`Invalid email(s)`)
+    } else {
+      setInviteError(null)
+    }
+    setRawInvitees(nextValue)
+    setInvitees(uniqueInvitees)
+  }
```
igorlesnenko, line 138: I think I saw a similar code in AddTeam and AddTeamMember. Is it possible to reuse this and `handleToggleInviteAll` ?

packages/client/modules/userDashboard/components/GcalModal/GcalModal.tsx
```diff
@@ -93,6 +117,52 @@ const GcalModal = (props: Props) => {
...
+
+  const handleToggleInviteAll = () => {
+    if (!inviteAll) {
+      const {parsedInvitees} = parseEmailAddressList(rawInvitees)
+      const currentInvitees = parsedInvitees
+        ? (parsedInvitees.map((invitee: any) => invitee.address) as string[])
```
igorlesnenko, line 144: Can we specify a type instead of `any`?

packages/client/modules/userDashboard/components/GcalModal/GcalModal.tsx
```diff
@@ -93,6 +117,52 @@ const GcalModal = (props: Props) => {
     onChange(e)
   }
 
+  const onInvitesChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
+    const nextValue = e.target.value
+    if (rawInvitees === nextValue) return
+    const {parsedInvitees, invalidEmailExists} = parseEmailAddressList(nextValue)
+    const allInvitees = parsedInvitees
+      ? (parsedInvitees.map((invitee: any) => invitee.address) as string[])
```
igorlesnenko, line 125: Can we add more specific type instead of any?

packages/client/modules/userDashboard/components/GcalModal/GcalModal.tsx
```diff
@@ -93,6 +117,52 @@ const GcalModal = (props: Props) => {
...
+      const formattedCurrentInvitees =
+        currentInvitees.length && lastInvitee && !lastInvitee.endsWith(',')
+          ? `${currentInvitees.join(', ')}, `
+          : currentInvitees.join(', ')
+      setRawInvitees(`${formattedCurrentInvitees}${emailsToAdd.join(', ')}`)
+      setInvitees([...currentInvitees, ...emailsToAdd])
+    } else {
+      const {parsedInvitees} = parseEmailAddressList(rawInvitees)
+      const currentInvitees = parsedInvitees
+        ? (parsedInvitees.map((invitee: any) => invitee.address) as string[])
```
igorlesnenko, line 157: Can we specify a type instead of any?

igorlesnenko (commented):
Not tested yet

igorlesnenko (commented):
Not tested yet

packages/server/graphql/public/types/CreateGcalEventInput.ts
nickoferrall: Sounds good, updated 👍

packages/client/modules/userDashboard/components/GcalModal/GcalModal.tsx
nickoferrall: I copied this from `NewTeamForm.tsx`. Which other file is it being used in? The engineering policy is to create a reusable function once it's repeated three times

packages/client/modules/userDashboard/components/GcalModal/GcalModal.tsx
nickoferrall: This was copy & pasted, but I've added a type 👍

igorlesnenko (approved):

PR #8670, opened 2023-08-15
fix: Avoid caching permissions with different arguments

Dschoordsch:
# Description

Permissions are cached by function name. When these depend on the arguments passed, they should have their own cache key, see https://the-guild.dev/graphql/shield/docs/rules#limitations

## Demo

No demo

## Testing scenarios

Works as before

## Final checklist


packages/server/graphql/public/rules/isViewerOnTeam.ts
```diff
@@ -6,15 +6,17 @@ import {GQLContext} from '../../graphql'
...
-    const {authToken, dataLoader} = context
-    const viewerId = getUserId(authToken)
-    const teamId = await getTeamId(source, args, context)
-    if (teamId instanceof Error) return teamId
-    const teamMemberId = TeamMemberId.join(teamId, viewerId)
-    const teamMember = await dataLoader.get('teamMembers').load(teamMemberId)
-    if (!teamMember) return new Error('Viewer in not on team')
-    return true
-  })
+  rule(`isViewerOnTeam-${getTeamId}`, {cache: 'contextual'})(
```
mattkrick, line 9: now that i read this again, since getTeamId _may_ also use `source, args`, should we change this cache to `strict`? 

+1 since `getTeamId` is a function, and we're putting it in a string, JS is gonna call `getTeamId.toString()`, which will result in the body of the function, right? should it instead be `getTeamId.name`?
is it enough to use the name of the `getTeamId` function, or do we need to somehow calculate the teamId?

Dschoordsch: We should switch to strict, you're right.
`.name` should be enough unless we're also passing anonymous functions at times, then it would lead to conflicts, so serializing the body is safer.

BartoszJarocki (approved):
works!

PR #8671, opened 2023-08-15
style: Normalize link styles

ackernaut:
# Description

Linked text can be in Standup response cards, thread comments, task cards, and in Poker item descriptions. We wanted to normalize the minimal styles, links use:
- Body color (Slate 700)
- Text decoration underline
- Have an outline on tab (this is the browser default)

This PR adds the text decoration to the Poker item description and updates the color of links in Standup response cards

## Testing
- [x] When linking text on a Standup response card the text is the same as the body color (Slate 700)
- [x] Using 2 local users in a Standup, you can tab to the link on another’s response card, it has an outline on focus
- [x] Add an item with linked text to a Poker meeting, the linked text is underlined, body color (Slate 700) and has an outline on focus

### DIFF
--- a/packages/client/components/PokerEstimateHeaderCardContent.tsx
+++ b/packages/client/components/PokerEstimateHeaderCardContent.tsx
@@ -41,7 +41,7 @@ const CardTitleWrapper = styled('div')({
   width: '100%'
 })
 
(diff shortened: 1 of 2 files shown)


acressall (approved):
Looks good!

PR #8681, opened 2023-08-15
chore(analytics): Add Amplitude server side event tracking

### LINKED ISSUE #8675: Write server side analytics events to Amplitude directly
As one of the first steps in the migration from Segment to Amplitude, we should write all server side events from analytics.ts to Amplitude in parallel with the Segment tracking events.

- Add typesafe wrapper for Amplitude to server code
- Write all analytics.ts events in parallel to Segment & Amplitude

tghanken:
# Description

Fixes #8675.  This is the first step in our migration from Segment to Amplitude.  This sets up all of our core server-side analytics events, found in analytics.ts, to be written in parallel to Amplitude and Segment.

## Demo

https://www.loom.com/share/89594006017945468f897022a5b46bc8?sid=fb06736e-dd3a-49e7-a401-a6c804728c2e

## Testing scenarios

- Get the development environment Amplitude write key from [here](https://app.amplitude.com/data/parabol/Parabol%20Production/sources/development) and save as the env var `AMPLITUDE_WRITE_KEY`
- Open any page and verify that the `Connect Websocket` event fires in both:
  - [ ] Amplitude - You can use the ingestion debugging page found [here](https://app.amplitude.com/data/parabol/Parabol%20Production/sources/development)
  - [ ] Segment - You can verify [here](https://app.segment.com/parabol-jordan/sources/actiondevelopment/debugger)


packages/server/utils/analytics/amplitude/AmplitudeAnalytics.ts
```diff
@@ -0,0 +1,29 @@
...
+      flushQueueSize: PROD ? 20 : 1,
+      optOut: !AMPLITUDE_WRITE_KEY
+    })
+  }
+
+  async track(userId: string, event: AnalyticsEvent, properties?: Record<string, any>) {
+    // used as a failsafe for PPMIs
+    if (!AMPLITUDE_WRITE_KEY) return
+    const user = await getUserById(userId)
+    const {email} = user ?? {}
```
tianrunhe, line 23: Is `email` required? Reading from this [section](https://www.docs.developers.amplitude.com/data/sdks/typescript-node/#tracking-an-event), it looks like only `user_id` is needed.

packages/server/utils/analytics/amplitude/AmplitudeAnalytics.ts
```diff
@@ -0,0 +1,29 @@
...
+    init(AMPLITUDE_WRITE_KEY || 'x', {
+      flushQueueSize: PROD ? 20 : 1,
+      optOut: !AMPLITUDE_WRITE_KEY
+    })
+  }
+
+  async track(userId: string, event: AnalyticsEvent, properties?: Record<string, any>) {
+    // used as a failsafe for PPMIs
+    if (!AMPLITUDE_WRITE_KEY) return
+    const user = await getUserById(userId)
```
tianrunhe, line 22: If `email` is required, we should probably use `dataloader` here instead of hitting PQ for every single event we track

tianrunhe (changes requested):

nickoferrall (changes requested):
The link that you shared in the PR description (to get the API key) takes me here: 
Screenshot 2023-08-21 at 15 43 46


But I can't click on the source. 

[This doc](https://www.docs.developers.amplitude.com/analytics/find-api-credentials/) says the API key should be under project, but it's not there for me. Perhaps because I don't have admin permissions?


Screenshot 2023-08-21 at 15 50 26

Could you add the API key to the devops repo so I can copy it from there?

packages/server/utils/analytics/amplitude/AmplitudeAnalytics.ts
tghanken: `email` is not required, but we are currently adding email to each segment event we emit server side, so I used the same logic to ensure parity between the two implementations. 
 https://github.com/ParabolInc/parabol/blob/699c1d423796befdd3b1dda9360a1c3a14f2f254/packages/server/utils/segmentIo.ts#L17-L19

packages/server/utils/analytics/amplitude/AmplitudeAnalytics.ts
tghanken: That seems like a good idea.  I'll make that change!

tghanken:
> Could you add the API key to the devops repo so I can copy it from there?

I've added the API keys!

packages/server/utils/analytics/amplitude/AmplitudeAnalytics.ts
```diff
@@ -0,0 +1,29 @@
...
+    init(AMPLITUDE_WRITE_KEY || 'x', {
+      flushQueueSize: PROD ? 20 : 1,
+      optOut: !AMPLITUDE_WRITE_KEY
+    })
+  }
+
+  async track(userId: string, event: AnalyticsEvent, properties?: Record<string, any>) {
+    // used as a failsafe for PPMIs
+    if (!AMPLITUDE_WRITE_KEY) return
+    const user = await getDataLoader().get('users').load(userId)
```
tghanken, line 22: @tianrunhe I've updated this and the relevant Segment code to use the dataloader, but I'm not sure I've done it correctly.  Everywhere else seems to call the dataloader through the GraphQL execution environment, whereas this is bypassing that.  Does this work? Or do we need to run all of our tracking through the GraphQL executors?

nickoferrall (approved):
LGTM!

packages/server/utils/analytics/amplitude/AmplitudeAnalytics.ts
tianrunhe: Unfortunately no... `await getDataLoader()` creates a new DataLoaderCache instance each time it's called so essentially it still hits PG every time. Porting the dataLoader from GraphQL seems overkill.
I saw in `segmentIo.ts` today we are doing `await getUserById(userId)` anyway so maybe I was over concerned. Let's just keep it as-is then!

tianrunhe (approved):

tianrunhe:
@tghanken another thing worth checking: today we emit metrics to Segment and then Segment copy them to Amplitude. After this PR, we also send event directly to Amplitude. Would this cause duplicate events in Amplitude?

packages/server/utils/analytics/amplitude/AmplitudeAnalytics.ts
```diff
@@ -0,0 +1,29 @@
...
+import PROD from '../../../PROD'
+
+const {AMPLITUDE_WRITE_KEY} = process.env
+
+/**
+ * Wrapper for amplitude providing a more typesafe interface
+ */
+export class AmplitudeAnalytics {
+  constructor() {
+    init(AMPLITUDE_WRITE_KEY || 'x', {
```
Dschoordsch, line 13: -1 Why `'x'`? If it's just to satisfy the types, I would expect `''`. If `'x'` has any special meaning, please add a code comment.

packages/server/utils/analytics/amplitude/AmplitudeAnalytics.ts
Dschoordsch: Yes, please make that change

packages/server/utils/analytics/analytics.ts
```diff
@@ -445,8 +448,12 @@ class Analytics {
     this.track(userId, 'Reset Groups Clicked', {meetingId, teamId})
   }
 
-  private track = (userId: string, event: AnalyticsEvent, properties?: Record<string, any>) =>
-    this.segmentAnalytics.track(userId, event, properties)
+  private track = (userId: string, event: AnalyticsEvent, properties?: Record<string, any>) => {
+    Promise.all([
```
Dschoordsch, line 452: +1 this is redundant because we're not `await`ing anything, i.e. just calling
```ts
      this.amplitudeAnalytics.track(userId, event, properties)
      this.segmentAnalytics.track(userId, event, properties)
```
without `Promise.all` would do the same.

Dschoordsch (changes requested):
Did not test

packages/server/utils/segmentIo.ts
```diff
@@ -9,12 +11,17 @@ const segmentIo = new SegmentIo(SEGMENT_WRITE_KEY || 'x', {
...
   if (!SEGMENT_WRITE_KEY) return
   const {userId, event, properties: inProps} = options
   if (!userId) return
-  const user = await getUserById(userId)
+  /*
+   * Since segmentIo is called directly throughout server side code, we can't depend on the dataloader
+   * from analytics.ts and create a new one as needed
+   */
+  const localDataloader = dataloader ?? getDataLoader().get('users')
+  const user = await localDataloader.load(userId)
```
tghanken, line 24: This got a little messy because we call `segmentIo.track()` throughout the server side code base, so I had to create a fallback dataloader if we didn't pass one in, but I _think_ I set it up in a decent way.  I'm cleaning up all the direct segmentIo calls in a follow-on PR, so we should be able to clean up this logic at that point as well.

packages/server/utils/analytics/amplitude/AmplitudeAnalytics.ts
tghanken: I had copied the initialization logic from `segmentIo.ts` without thinking too much about it, which included the ` || 'x'` logic.

After reviewing, we have no real need to initialize the SDK at all if we don't have a write key, so I added an early exit in the constructor if the env var is not set.

packages/server/utils/analytics/amplitude/AmplitudeAnalytics.ts
```diff
@@ -0,0 +1,37 @@
...
+    init(AMPLITUDE_WRITE_KEY, {
+      flushQueueSize: PROD ? 20 : 1,
+      optOut: !AMPLITUDE_WRITE_KEY
+    })
+  }
+
+  async track(
+    userId: string,
+    event: AnalyticsEvent,
+    dataloader: NullableDataLoader<string, IUser, string>,
```
Dschoordsch, line 25: +1 Usually we would pass `dataLoader: DataLoaderWorker` here.

packages/server/utils/segmentIo.ts
Dschoordsch: +1 in almost all places where we call `segmentIo.track` we also have a dataloader available. It would be nice to pass it when calling this

Dschoordsch (approved):

packages/server/utils/analytics/amplitude/AmplitudeAnalytics.ts
tghanken: I got a type error when trying to use DataLoaderWorker as it had a type of `CacheWorker<RootDataLoader>` but `getDataLoader()` returns a type of `CacheWorker<DataLoaderBase>`.

I used `CacheWorker<DataLoaderBase>` for the track() interface to get this merged quickly, but can revisit later if necessary.

PR #8682, opened 2023-08-15
fix: exit early if no segment write key

### LINKED ISSUE #8674: Remove segment when the env var is missing
Got the following error message in KR (PPMI):

```
AssertionError [ERR_ASSERTION]: You must pass either an "anonymousId" or a "userId".

```

AC
- [ ] When segment env var is gone, don't send anything to segment

mattkrick:
# Description

fixes #8674
We had a PPMI who does not use segment get segment-related errors in their server logs. This should fix that.

## Testing scenarios

- [ ] nuke the SEGMENT_WRITE_KEY & run in prod mode. when segmentEventTrack gets called, it exits early.

### DIFF
--- a/packages/server/graphql/mutations/segmentEventTrack.ts
+++ b/packages/server/graphql/mutations/segmentEventTrack.ts
@@ -54,8 +54,14 @@ export default {
     {event, options}: SendClientSegmentEventMutationVariables,
     {authToken, dataLoader}: GQLContext
   ) => {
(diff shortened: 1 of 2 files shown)


tianrunhe (approved):
You might be glad to hear: we will completely deprecate Segment in a month!

mattkrick:
whoaaa that's big news!

PR #8687, opened 2023-08-16
fix: Uncompressing emojies twice causes a failure

Dschoordsch:
TeamDashInsights and ReactjiCount both uncompressed the emojies provided by emoji-mart. This caused occasional exceptions. Moving this to a helper instead used by both.

## Demo

no demo

## Testing scenarios

- have a team which shows emoji insights
- start a meeting and add some reactions
- see no exception

## Final checklist

### DIFF
--- a/packages/client/components/ReflectionCard/ReactjiCount.tsx
+++ b/packages/client/components/ReflectionCard/ReactjiCount.tsx
@@ -1,8 +1,5 @@
 import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
-import data from 'emoji-mart/data/apple.json'
-import {uncompress} from 'emoji-mart/dist-modern/utils/data.js'
-import {unifiedToNative} from 'emoji-mart/dist-modern/utils/index.js'
 import React from 'react'
 import {useFragment} from 'react-relay'
 import PlainButton from '~/components/PlainButton/PlainButton'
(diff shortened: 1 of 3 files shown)


tianrunhe (changes requested):
I'm not sure if this is a regression or something new but I did this sequence of test steps:
1. Manually update the DB for a team: set the `mostUsedEmojis` to `[{"id": "heavy_plus_sign", "count": 13}, {"id": "thinking_face", "count": 5}]`
2. Start a meeting within the team, add some reflections and advance to discuss phase. Add an emoji to a reflection card
3. Now go back to RethinkDB and find that reflection, manually increases its emojis count like this:
```javascript
r.db("actionDevelopment").table('RetroReflection').get('xxxx').update({reactjis: [{ "id": "fire" , "userId": "xxx" } , { "id": "fire" , "userId": "xxx" }, ... ]})
```

Screenshot 2023-08-16 at 11 38 23

4. End the meeting
5. Go back to Team dashboard and see no insights at all. Check the Team table and it looks like the `mostUsedEmojis` has been overwritten to `null`?

Dschoordsch:
@tianrunhe for updating the most used emojis, we're just reading old meetings, not the current value. If there aren't at least 2 different reactjis with count >= 5 in the last 3 month, then the value is set to `null`.

tianrunhe (approved):

Issue #8636, opened 2023-08-09
Gcal Integration: schedule button asks user to integrate if they're not integrated


nickoferrall:
If the user isn't integrate and they click the schedule button, I suggest we open another tab with the gcal integration oauth flow:

![Screenshot 2023-08-09 at 14 37 00](https://github.com/ParabolInc/parabol/assets/39854876/a2f7b0f5-75ef-4577-b167-ea627c43dae2)

### AC

- [ ] Handle users clicking "Schedule" that haven't integrated with gcal

PR #8688, opened 2023-08-16
feat(gcal): unauthed schedule button ux

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8636

Demo: https://www.loom.com/share/30ebfc16554e41b2bdcbece0d9eba1a8

### To Test

- [ ] If a user doesn't have the gcal feature flag, they don't see the "Schedule" button
- [ ] If they do, but they haven't integrated with gcal, they get taken to the gcal auth page when they click "Schedule"
- [ ] After integrating with gcal, the modal opens automatically 
- [ ] If they have already integrated with gcal, the schedule button opens the gcal modal immediately

### DIFF
--- a/packages/client/components/ActivityLibrary/ScheduleMeetingButton.tsx
+++ b/packages/client/components/ActivityLibrary/ScheduleMeetingButton.tsx
@@ -1,0 +1,104 @@
+import React, {useEffect, useState} from 'react'
+import graphql from 'babel-plugin-relay/macro'
+import {useDialogState} from '../../ui/Dialog/useDialogState'
+import SecondaryButton from '../SecondaryButton'
+import GcalModal from '../../modules/userDashboard/components/GcalModal/GcalModal'
+import {CreateGcalEventInput} from '../../__generated__/StartRetrospectiveMutation.graphql'
+import GcalClientManager from '../../utils/GcalClientManager'
+import useAtmosphere from '../../hooks/useAtmosphere'
+import {useFragment} from 'react-relay'
+import {ScheduleMeetingButton_team$key} from '~/__generated__/ScheduleMeetingButton_team.graphql'
+import {ScheduleMeetingButton_viewer$key} from '~/__generated__/ScheduleMeetingButton_viewer.graphql'
+import {MenuMutationProps} from '../../hooks/useMutationProps'
+
+type Props = {
+  mutationProps: MenuMutationProps
+  handleStartActivity: (gcalInput?: CreateGcalEventInput) => void
+  teamRef: ScheduleMeetingButton_team$key
+  viewerRef: ScheduleMeetingButton_viewer$key
+}
+
+const ScheduleMeetingButton = (props: Props) => {
+  const {mutationProps, handleStartActivity, teamRef, viewerRef} = props
+  const {
+    isOpen: isScheduleDialogOpen,
+    open: openScheduleDialog,
+    close: closeScheduleDialog
+  } = useDialogState()
+  const atmosphere = useAtmosphere()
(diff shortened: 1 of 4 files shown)


packages/client/components/ActivityLibrary/ScheduleMeetingButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+    `,
+    viewerRef
+  )
+  const hasGcalFlag = viewer.featureFlags.gcal
+
+  const team = useFragment(
+    graphql`
+      fragment ScheduleMeetingButton_team on Team {
+        id
+        teamMembers {
```
igorlesnenko, line 48: Ideally, would be good to not fetch all teamMembers and their integrations, as we only need one

igorlesnenko (approved):

packages/client/components/ActivityLibrary/ScheduleMeetingButton.tsx
Dschoordsch: -1 you can query `teamMember(teamId: ID!)` on `viewer` instead, can't you? I think introducing a `teamMember` field which returns the viewer's teamMember would also be an option.

packages/client/components/ActivityLibrary/ScheduleMeetingButton.tsx
```diff
@@ -0,0 +1,106 @@
...
+          isSelf
+          integrations {
+            gcal {
+              auth {
+                id
+              }
+              cloudProvider {
```
Dschoordsch, line 55: +1 not blocking yet, but could be an issue for PPMIs once we remove the feature flag. I did have the feature flag but no `cloudProvider` yet, so I saw the "schedule" button which just was a no-op. Maybe it would be cleaner to:
* on server side check if the user has the feature flag and only then return the `cloudProvider`, otherwise return `null
* on client don't check the feature flag at all but just the presence of `cloudProvider`

Dschoordsch (changes requested):
Works, but I don't like fetching all team members to find the auth. I think we wouldn't otherwise load them at this point.
Also I think I shouldn't have access to my team members auth for the calendar as this is personal and not a team resource like a backlog.

packages/client/components/ActivityLibrary/ScheduleMeetingButton.tsx
nickoferrall: Sounds good - I'm now only showing the "Schedule" button if the gcal flag & the `cloudProvider` exist. I've also implemented a `viewerTeamMember` field on the `Team` object so we only query one team member.

Dschoordsch (approved):
Did not test again.

Can you create an issue for the permissions of gcal.auth?

nickoferrall:
> Can you create an issue for the permissions of gcal.auth?

@Dschoordsch sorry, what problem is there with the `gcal.auth` permissions?

Dschoordsch:
@nickoferrall I think only the viewer should have access to their gcal auth, not the team members
> Also I think I shouldn't have access to my team members auth for the calendar as this is personal and not a team resource like a backlog.

nickoferrall:
OK cool - in this PR I'm only querying the viewer to get their gcal integration (see [here](feat/8635/handle-errors-gcal)), but you're suggesting that we shouldn't be able to query all team members and get their gcal integrations?

The reason being that gcal is different from other integrations because it's more personal than Slack, Teams, etc

Dschoordsch:
Yes

nickoferrall:
Issue created 👍 

https://github.com/ParabolInc/parabol/issues/8710

Issue #8679, opened 2023-08-15
Standups: "Your tasks" sidebar for Parabol


jmtaber129:
Implement the "your tasks" sidebar for Parabol tasks. For this task, do _not_ include the "add task" functionality.

Designs: https://github.com/ParabolInc/parabol/issues/8633

AC:
- [ ] Top bar is updated according to designs, including on mobile
- [ ] "Your tasks" button that opens up the sidebar
- [ ] Sidebar shows user's tasks from all teams, segmented by task status
- [ ] Empty state appears when no tasks are present
- [ ] Tasks appear when tasks are added in a different window (subscriptions)
- [ ] Transition between task sidebar and discussion sidebar is reasonable
- [ ] After moving a task to a different status, the task appears under the new status

PR #8690, opened 2023-08-16
feat(standups): "Your tasks" drawer with Parabol tasks

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8679

Implement the "Your tasks" drawer to show a user's Parabol tasks within a standup meeting, and update the topbar with the new "your tasks" button.

This PR does a few things to implement this:
1. Refactor the existing `TeamPromptDiscussionDrawer` into a `TeamPromptDrawer` and a `TeamPromptDiscussionDrawer`. The `TeamPromptDrawer` acts as the container for any components that exist in the drawer, handling things like overlay vs static sidebar, sliding in/out, etc.
  1. This is a sizeable chunk of the overall diff.
3. Adding a new `TeamPromptWorkDrawer` to hold all "your work" components, starting with Parabol tasks.
4. Refactors + updates components in the `TeamPromptTopBar` to match updated designs.

**I recommend reviewers review with whitespace hidden** due to the drawer refactor.

## Demo
https://www.loom.com/share/75f2739a64604199a785e203b8d9cb91?sid=4431ced3-cbed-4a74-913b-c8ef843f3ff3

## Testing scenarios
- [ ] Top bar is updated according to designs, including on mobile
- [ ] Click "Your tasks" button to open the sidebar
- [ ] Sidebar shows user's tasks from all teams. Click through each tab
- [ ] Move all tasks from a specific status into a different status, and check empty state for that status
- [ ] Add a task in a different window, and confirm task appears in sidebar with no refresh
- [ ] With task sidebar open, click "reply" on a response to transition to discussion sidebar, then click "your tasks" to transition back.
- [ ] Move a task to a different status, and confirm the task appears under the new status

## Final checklist

### DIFF
--- a/packages/client/components/TeamPrompt/TeamPromptDrawer.tsx
+++ b/packages/client/components/TeamPrompt/TeamPromptDrawer.tsx
@@ -5,37 +5,42 @@ import {commitLocalUpdate, useFragment} from 'react-relay'
 import useAtmosphere from '~/hooks/useAtmosphere'
 import {TeamPromptDrawer_meeting$key} from '~/__generated__/TeamPromptDrawer_meeting.graphql'
 import {desktopSidebarShadow} from '../../styles/elevation'
-import {BezierCurve, DiscussionThreadEnum, ZIndex} from '../../types/constEnums'
+import {BezierCurve, Breakpoint, DiscussionThreadEnum, ZIndex} from '../../types/constEnums'
 import ResponsiveDashSidebar from '../ResponsiveDashSidebar'
 import TeamPromptDiscussionDrawer from './TeamPromptDiscussionDrawer'
 import TeamPromptWorkDrawerRoot from './TeamPromptWorkDrawerRoot'
+import useBreakpoint from '../../hooks/useBreakpoint'
 
-const Drawer = styled('div')<{isDesktop: boolean; isOpen: boolean}>(({isDesktop, isOpen}) => ({
-  boxShadow: isDesktop ? desktopSidebarShadow : undefined,
-  backgroundColor: '#FFFFFF',
-  display: 'flex',
-  flex: 1,
-  flexDirection: 'column',
-  justifyContent: 'stretch',
-  overflow: 'hidden',
-  position: isDesktop ? 'fixed' : 'static',
-  bottom: 0,
-  top: 0,
-  right: isDesktop ? 0 : undefined,
-  userSelect: isDesktop ? undefined : 'none',
-  transition: `all 200ms ${BezierCurve.DECELERATE}`,
-  transform: `translateX(${
-    isOpen
-      ? `calc(${DiscussionThreadEnum.WIDTH}px - min(${DiscussionThreadEnum.WIDTH}px, 100vw))`
-      : `${DiscussionThreadEnum.WIDTH}px`
-  })`,
-  width: `min(${DiscussionThreadEnum.WIDTH}px, 100vw)`,
-  zIndex: ZIndex.SIDEBAR,
-  height: '100%',
-  '@supports (height: 1svh) and (height: 1lvh)': {
-    height: isDesktop ? '100lvh' : '100svh'
-  }
-}))
+const Drawer = styled('div')<{isDesktop: boolean; isMobile: boolean; isOpen: boolean}>(
+  ({isDesktop, isMobile, isOpen}) => ({
+    boxShadow: isDesktop ? desktopSidebarShadow : undefined,
+    backgroundColor: '#FFFFFF',
+    display: 'flex',
+    flex: 1,
+    flexDirection: 'column',
+    justifyContent: 'stretch',
+    overflow: 'hidden',
+    position: isDesktop ? 'fixed' : 'static',
+    bottom: 0,
+    top: 0,
+    right: isDesktop ? 0 : undefined,
+    userSelect: isDesktop ? undefined : 'none',
+    transition: `all 200ms ${BezierCurve.DECELERATE}`,
+    transform: `translateX(${
+      isOpen
+        ? isMobile
+          ? `calc(${DiscussionThreadEnum.WIDTH}px - 100vw)`
+          : 0
+        : `${DiscussionThreadEnum.WIDTH}px`
+    })`,
+    width: isMobile ? '100vw' : `min(${DiscussionThreadEnum.WIDTH}px, 100vw)`,
+    zIndex: ZIndex.SIDEBAR,
+    height: '100%',
+    '@supports (height: 1svh) and (height: 1lvh)': {
+      height: isDesktop ? '100lvh' : '100svh'
+    }
+  })
+)
 
 interface Props {
   meetingRef: TeamPromptDrawer_meeting$key
@@ -54,6 +59,7 @@ const TeamPromptDrawer = ({meetingRef, isDesktop}: Props) => {
     meetingRef
   )
 
+  const isMobile = !useBreakpoint(Breakpoint.FUZZY_TABLET)
   const atmosphere = useAtmosphere()
   const {id: meetingId, isRightDrawerOpen} = meeting
 
@@ -81,7 +87,7 @@ const TeamPromptDrawer = ({meetingRef, isDesktop}: Props) => {
       onToggle={onToggleDrawer}
       sidebarWidth={DiscussionThreadEnum.WIDTH}
     >
-      <Drawer isDesktop={isDesktop} isOpen={isRightDrawerOpen}>
+      <Drawer isDesktop={isDesktop} isMobile={isMobile} isOpen={isRightDrawerOpen}>
         {renderedInnerDrawer}
       </Drawer>
     </ResponsiveDashSidebar>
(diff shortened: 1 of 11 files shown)


packages/client/components/TeamPrompt/TeamPromptDrawer.tsx
```diff
@@ -0,0 +1,98 @@
...
+    renderedInnerDrawer = (
+      <TeamPromptDiscussionDrawer meetingRef={meeting} onToggleDrawer={onToggleDrawer} />
+    ) ?? <TeamPromptWorkDrawerRoot onToggleDrawer={onToggleDrawer} />
+  } else if (localStageId) {
+    renderedInnerDrawer = (
+      <TeamPromptDiscussionDrawer meetingRef={meeting} onToggleDrawer={onToggleDrawer} />
+    )
+  } else if (showWorkSidebar) {
+    renderedInnerDrawer = <TeamPromptWorkDrawerRoot onToggleDrawer={onToggleDrawer} />
+  }
```
BartoszJarocki, line 78: +1 this logic is a little bit hard to follow, especially considering it relies on `TeamPromptDiscussionDrawer` returning either null or an actual component (`TeamPromptDiscussionDrawer` also checks internally `localStageId`, which is also checked here). Is there any chance to make it more clear?

packages/client/schemaExtensions/clientSchema.graphql
```diff
@@ -44,6 +44,7 @@ extend type RetrospectiveMeeting {
 extend type TeamPromptMeeting {
   localStageId: String
   isRightDrawerOpen: Boolean!
+  showWorkSidebar: Boolean!
```
BartoszJarocki, line 47: +1 can we reduce number of flags to just 2? ie do the following
1. use `localStageId` to control if discussion drawer is showing (if it's set, show, if it's not do not show it)
2. use `isWorkSidebarShowing` to control if task drawer is showing

is there anything I'm missing that'd make this solution no viable? if not, it _should_ simplify the render conditionals , make it easier to understand and more explicit

packages/client/components/TeamPromptMeeting.tsx
```diff
@@ -190,7 +179,7 @@ const TeamPromptMeeting = (props: Props) => {
                 </ResponsesGridContainer>
               </ErrorBoundary>
             </StyledMeetingHeaderAndPhase>
-            <TeamPromptDiscussionDrawer meetingRef={meeting} isDesktop={isDesktop} />
+            <TeamPromptDrawer meetingRef={meeting} isDesktop={isDesktop} />
```
BartoszJarocki, line 182: +1 we could do the conditional rendering here instead of doing it internally in `TeamPromptDrawer` and treat `TeamPromptDrawer` as a style container

BartoszJarocki (approved):
nicer work! everything works correctly so no blockers but the logic related to rendering is a little bit hard to follow. I left some suggestions to consider.

packages/client/schemaExtensions/clientSchema.graphql
jmtaber129: Not quite - on smaller screens it's possible to open the drawer by swiping in from the right (e.g. open a discussion thread, close the drawer, then swipe the drawer back in to open the discussion again), so we need both a) the state describing the drawer contents, and b) whether the drawer is actually open right now.

packages/client/components/TeamPrompt/TeamPromptDrawer.tsx
jmtaber129: I've simplified this a bit, but it's still a bit weird. I've added some comments to hopefully make the logic clearer.

packages/client/components/TeamPromptMeeting.tsx
jmtaber129: I still prefer to keep everything abstracted within `TeamPromptDrawer`, especially since the internal drawer components need the `onToggleDrawer` handler managed by the `TeamPromptDrawer`

mattkrick:
code looks good, i'll do testing first thing tomorrow! want to make sure there's no unintended side effects in iOS since they usually add extra gestures there

packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
```diff
@@ -134,74 +134,103 @@ const TeamPromptTopBar = (props: Props) => {
...
+      meetingProxy.setValue(true, 'showWorkSidebar')
+      meetingProxy.setValue(true, 'isRightDrawerOpen')
+    })
+  }
+
+  const buttons = (
+    <ButtonContainer>
+      <button
+        className='group flex h-max w-max cursor-pointer flex-col items-center px-2 text-sm font-semibold text-sky-500 hover:text-sky-600'
+        onClick={onOpenWorkSidebar}
```
mattkrick, line 151: -1 can we make this a toggle? e.g. if i click it when the right drawer is open, it closes the drawer? that'll keep the menu trigger consistent with the other ones in the app!

packages/client/components/TeamPrompt/TeamPromptDrawer.tsx
```diff
@@ -0,0 +1,87 @@
...
+  justifyContent: 'stretch',
+  overflow: 'hidden',
+  position: isDesktop ? 'fixed' : 'static',
+  bottom: 0,
+  top: 0,
+  right: isDesktop ? 0 : undefined,
+  userSelect: isDesktop ? undefined : 'none',
+  transition: `all 200ms ${BezierCurve.DECELERATE}`,
+  transform: `translateX(${isOpen ? 0 : DiscussionThreadEnum.WIDTH}px)`,
+  width: DiscussionThreadEnum.WIDTH,
```
mattkrick, line 28: +1 I think `DiscussionThreadEnum.WIDTH` is too big for mobile displays, which is why on mobile we have to stack the discussion thread below the card area in retros (I see it's been this width for standups, just haven't poked around til now!). Could we reduce it to the 256-296px range? That way we leave enough of the scrim as a click target (I think 16px of the scrim is for the peek, so we really only have a few pixels where we can click to exit). We could also make the width dependent on device! 

Screenshot 2023-08-22 at 11 42 54 AM

packages/client/components/TeamPrompt/TeamPromptWorkDrawer.tsx
```diff
@@ -0,0 +1,102 @@
...
+    .map((edge) => edge.node)
+    .filter((task) => task.status === selectedStatus)
+
+  return (
+    <>
+      <div className='px-4 pt-4'>
+        <div className='flex justify-between pr-4'>
+          <div className='text-base font-semibold'>Your Tasks</div>
+          <div className='cursor-pointer text-slate-600 hover:opacity-50' onClick={onToggleDrawer}>
+            <Close />
```
mattkrick, line 56: -1 Can we make the X in the same location for both drawers? That will minimize the visual change when clicking between them.

Screenshot 2023-08-22 at 11 51 42 AM
Screenshot 2023-08-22 at 11 51 50 AM

mattkrick (changes requested):

packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
jmtaber129: Updated - definitely an improvement!

packages/client/components/TeamPrompt/TeamPromptWorkDrawer.tsx
jmtaber129: Fixed!

packages/client/components/TeamPrompt/TeamPromptDrawer.tsx
jmtaber129: I went a slightly different direction here - the actual CSS is a bit more complicated due to how the sidebar components are structured, but I've changed the effective behavior to something like `width: DiscussionThreadEnum.WIDTH, max-width: 100vw`, so the width will start shrinking once the viewport width is <`DiscussionThreadEnum.WIDTH`.

I could've preserved a few pixels for "tap outside to close" behavior, but between the close button and "swipe right to close" behavior I think it's fine to allow this to go up to full-width on those smallest screens, but let me know what you think!

mattkrick: other 2 fixes look good! This one is pretty much the same for iPhone 12/Pixel 5, which both have ~390px width. If you don't wanna shrink the width you could make it go full screen & just say screw the scrim. main tension is why show the scrim when my fingers are too fat to click it? 

For rough numbers:
- minimum clickable width: 48px (google material design suggestion)
- peek handler:  20px
- target width: ~390px (I think we actually target a min of 360? but that was from years ago, devices have gotten bigger)
- 390 - 20 - 48 = 322 max width

In other parts we try to keep the right drawer at 256, but with cards in it we may have to go bigger https://github.com/ParabolInc/parabol/blob/69cad570a630f877f2f01aea79cd7fdd1431fb39/packages/client/types/constEnums.ts#L331

jmtaber129: @mattkrick  Ahh gotcha. I prefer to grow the width over shrink it here, so I changed this to just go full-screen for breakpoints under `FUZZY_TABLET` (700px).

mattkrick: LOVE IT!

packages/client/styles/helpers/cardRootStyles.ts
```diff
@@ -7,7 +7,6 @@ const cardRootStyles = {
   borderRadius: Card.BORDER_RADIUS,
   boxShadow: cardShadow,
   minWidth: 256,
-  maxWidth: 296,
```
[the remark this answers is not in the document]
mattkrick: update: gonna revert this change because the # of people affected by the extra long boys is probably greater than the number of people affected by the Task drawer

Issue #8635, opened 2023-08-09
Gcal Integration: handle error state


nickoferrall:
If you click the CTA in the gcal modal, you create both the gcal event & the meeting. However, if there's an error and the gcal event can't be created, we need to give the user feedback.

See Matt's Loom example: https://www.loom.com/share/025b4d6d67734a52afab73c04ddfdb21

A few options I can think of:
- Create the meeting, take the user to the meeting, and show a snackbar letting them know that the gcal event wasn't created
- Create the meeting, but don't take the user to the meeting. Show an error message in the modal letting them know the gcal event wasn't created
- Don't create the meeting, show the error message in the modal
- Separate the button so that you create the gcal event and meeting separately

I prefer the first two options, leaning towards the snackbar.

### AC

- [ ] Design created or agreed upon to handle gcal errors

nickoferrall:
@acressall it'd be great to get your thoughts on this one when you get the chance. It's not urgent, but I'd like to implement it within the next week or so

acressall:
@nickoferrall what do you think of showing the auth flow for gcal if they 1. hit the create meeting & send invite btn and 2. don't have the integration already connected? I think the great thing about showing the schedule btn even when there integration isn't added is the awareness it creates -- so could we connect in this context? I would expect that once I've authed the meeting will start. If, for some reason, I decided _not_ to authenticate when prompted, then the meeting can start and we can let them know via snackbar that the invite was not created because they didn't connect to gcal

PR #8694, opened 2023-08-17
feat: give users feedback if there's a gcal error

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8635

Loom Demo: https://www.loom.com/share/65b909380eb8487e8ff94303fb1f4250

Note: @acressall suggested [here](https://github.com/ParabolInc/parabol/issues/8635#issuecomment-1671690060) that we could give them an option to retry creating a gcal event, but for simplicity & speed, I haven't done that yet. 

### To test

- [ ] Update `createGcalEvent.ts` so that it always returns an error, e.g.:
```
const createGcalEvent = async (input: Input) => {
  const {gcalInput, meetingId, viewerId, dataLoader, teamId} = input
  // if (!gcalInput) {
  //   return {error: null}
  // }
  const viewer = await dataLoader.get('users').loadNonNull(viewerId)
  const {featureFlags} = viewer
  // if (featureFlags.includes('gcal')) {
  if (true) {
    return standardError(new Error('Does not have gcal feature flag'), {userId: viewerId})
  }
```

- [ ] Create any meeting type and see that there's a snackbar that tells the user that the gcal event couldn't be created
- [ ] The user has to click on the snackbar to dismiss it


packages/server/graphql/types/StartSprintPokerPayload.ts
```diff
@@ -24,6 +24,11 @@ export const StartSprintPokerSuccess = new GraphQLObjectType<any, GQLContext>({
     },
     teamId: {
       type: new GraphQLNonNull(GraphQLID)
+    },
+    hasGcalError: {
```
igorlesnenko, line 28: Did you consider sending snackbar notification instead? No strong preference, just thinking

igorlesnenko (approved):

packages/server/graphql/types/StartSprintPokerPayload.ts
nickoferrall: I am sending a snackbar notification if `hasGcalError` is true 😄

Dschoordsch (approved):

PR #8695, opened 2023-08-17
feat: Add "Is this helpful?" for insights

### LINKED ISSUE #8466: Insights: tracking and "Is this helpful?"
## Acceptance criteria
* "Is this helpful?" text and buttons at the bottom of insights with tracking of **viewed**, **thumbs_up**, **thumbs_down**
* generic so it can be used for all insights

Dschoordsch:
# Description

Fixes #8466
Add a team insights card with buttons for "Is this helpful?". Clicks are tracked via `sendSegmentEvent`.
Because this alone would not give any feedback to the user, I thought about how we could preserve the fact that they've voted already. The best solution would be to store the information in the `TeamMember`, but that would require the additional field, it would need to get queried and we'd need to have a custom mutation just for the button highlight.
I've chosen a simpler approach to just store it in local state of the component, so the user sees that something had happened, but next time they're free to vote again.

## Demo

Screenshot from 2023-08-17 16-33-50
https://www.loom.com/share/0530da3c9dd047be81c871a1866f15c2?sid=9ade9d27-45af-4b70-a9ab-5dbb0320b4d3

## Testing scenarios

- [ ] with insights shown, press up or down vote button and see the send segment event mutation firing
- [ ] you can manually set `Team.mostUsedEmojis` to
  ```
  [
    {
        "id": "heavy_plus_sign",
        "count": 13
    },
    {
        "id": "thinking_face",
        "count": 5
    }
  ]
  ```
  to make the card appear

## Final checklist

### DIFF
--- a/packages/client/modules/teamDashboard/components/TeamDashInsights/TeamInsightsCard.tsx
+++ b/packages/client/modules/teamDashboard/components/TeamDashInsights/TeamInsightsCard.tsx
@@ -1,5 +1,5 @@
 import graphql from 'babel-plugin-relay/macro'
-import React, {ReactNode} from 'react'
+import React, {ReactNode, useState} from 'react'
 import {useFragment} from 'react-relay'
 import Tooltip from '../../../../components/Tooltip'
 import {Info as InfoIcon} from '@mui/icons-material'
@@ -9,6 +9,7 @@ import SendClientSegmentEventMutation from '../../../../mutations/SendClientSegm
 import useAtmosphere from '../../../../hooks/useAtmosphere'
 import {TeamInsightsCard_insights$key} from '../../../../__generated__/TeamInsightsCard_insights.graphql'
 import TeamInsightsId from '../../../../shared/gqlIds/TeamInsightsId'
+import clsx from 'clsx'
 
 interface Props {
   title: string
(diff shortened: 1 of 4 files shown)


packages/client/modules/teamDashboard/components/TeamDashInsights/TeamInsightsCard.tsx
```diff
@@ -0,0 +1,77 @@
...
+
+  const atmosphere = useAtmosphere()
+  const {viewerId} = atmosphere
+
+  const trackClick = (isHelpfulInsight: boolean) => {
+    SendClientSegmentEventMutation(atmosphere, title, {
+      viewerId,
+      teamId,
+      isHelpfulInsight
+    })
```
Dschoordsch, line 45: @tghanken I added the isHelpfulInsight property to the segment events and then just send the title of the card. My goal here is to reuse this component for other insight cards as well. What do you think?

tghanken: -1 I'd recommend going with something like this instead.  We want the event itself (Insight Card Feedback Clicked) to be the same across all the various insights cards, then to use properties to segment between the individual types.  My recommended component should be safe to reuse across all insights cards!

```suggestion
    SendClientSegmentEventMutation(atmosphere, 'Insight Card Feedback Clicked', {
      viewerId,
      teamId,
      cardName: title,
      isHelpfulInsight
    })
```

tghanken (changes requested):

igorlesnenko (approved):

Issue #8575, opened 2023-07-27
Gcal Integration: improve date UI & UX


nickoferrall:
### AC

- [ ] Date picker input is consistent with the other inputs, i.e. it doesn't have a black border or blue border on focus
- [ ] You can't select an end date before the start date

PR #8696, opened 2023-08-17
feat(gcal): gcal date UI/UX

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8575

Loom demo: https://www.loom.com/share/d7f01375adc5420f917c0d93aea6faa8

### To test 

- [ ] Open the gcal modal and play around with the datetime picker. See that it looks consistent with the other inputs
- [ ] You can't select an end date before a start date
- [ ] Create a meeting, go to your google calendar, and see that an event was created at the time you selected


packages/client/modules/userDashboard/components/GcalModal/GcalModal.tsx
```diff
@@ -164,14 +187,14 @@ const GcalModal = (props: Props) => {
   }
 
   return (
-    <Dialog isOpen={isOpen} onClose={closeModal}>
+    <StyledDialogContainer>
```
nickoferrall, line 190: I had to use the old modal instead of the radix one as the radix one wouldn't let me select any of the datetime fields. 

I tried updating the z-index in a few different places, and spent a fair amount of time looking into it, but the original modal works with the MUI datepicker so I think it's best to use that.

acressall:
I think you've picked the best of all options. I think the real gcal can do it better because the date and time are separate fields, but I don't want to add extra space for that right now. We can improve later if we get lots of usage and/or complaints.

One q: did you set it to use 24hr clock? Or would that change based on device settings by chance? I _think_ it's mostly north america that would dislike this, so probably safe for a good chunk of users

acressall (approved):

nickoferrall:
> I think the real gcal can do it better because the date and time are separate fields,

I think you're right. I've pushed an update separating the two:

Screenshot 2023-08-21 at 16 29 45
Screenshot 2023-08-21 at 16 29 24
Screenshot 2023-08-21 at 16 29 29


> One q: did you set it to use 24hr clock? Or would that change based on device settings by chance? I think it's mostly north america that would dislike this, so probably safe for a good chunk of users

Interesting, I didn't know that's less popular in North America! I've updated it so that it shows AM/PM.

acressall:
Love the update!

> Interesting, I didn't know that's less popular in North America! I've updated it so that it shows AM/PM.
Yeah, I'm not sure which is better/ covers the majority of users. Would love to have it use device settings in future, not sure if that's possible

packages/client/modules/userDashboard/components/GcalModal/GcalModal.tsx
igorlesnenko: Oh, that's unfortunate. Created an issue about this https://github.com/ParabolInc/parabol/issues/8711

igorlesnenko (changes requested):
I can select end date before start date, in this case I see an error after starting a meeting

image

igorlesnenko (approved):
There is one thing that could be improved, If I starting to delete time, it reset all fields

https://www.loom.com/share/a4b33ca9de694a2ea29831cfd3753887

Could be done in other PR

nickoferrall:
> There is one thing that could be improved, If I starting to delete time, it reset all fields
Could be done in other PR

Sounds good, I'll work on this in a separate PR as I'm away tomorrow and I'd like to keep the review process moving. Issue created here: https://github.com/ParabolInc/parabol/issues/8765

packages/client/modules/userDashboard/components/GcalModal/DateTimePickers.tsx
```diff
@@ -0,0 +1,120 @@
...
+        setEnd(newValue)
+      } else {
+        const newStartValue = newValue.subtract(1, 'hour')
+        setStart(newStartValue)
+        setEnd(newValue)
+      }
+    }
+  }
+
+  const handleMouseDown = (e: React.MouseEvent<HTMLDivElement, MouseEvent>) => {
```
Dschoordsch, line 78: -1 Why is this needed, please add a comment

nickoferrall: Updated! 👍

Dschoordsch (approved):

Issue #8677, opened 2023-08-15
Stop writing analytics events to Segment


tghanken:
As a final migration step, stop writing all events to Segment on both the client and server.

### Acceptance Criteria (optional)
- Disable client Segment events
- Disable server Segment events

PR #8699, opened 2023-08-18
chore(analytics): Route analytics identify calls through analytics.ts

### LINKED ISSUE #8680: Move analytics identify calls to analytics.ts
Move the analytics `identify()` function to analytics.ts, and identify to Segment and Amplitude in parallel.

- Add identify() function to analytics.ts
- Write identify events to Segment
- Write identify events to Amplitude

tghanken:
# Description

Fixes #8680.  As part of our migration to Amplitude for analytics, this PR centralizes all `identify()` calls in analytics.ts instead of calling `segmentIo` directly, and makes `identify()` calls to Segment and Amplitude in parallel.  Segment writes will later be removed when we address #8677.

## Testing scenarios

- Get the development environment Amplitude write key from [here](https://app.amplitude.com/data/parabol/Parabol%20Production/sources/development) and save as the env var `AMPLITUDE_WRITE_KEY`
- Open any page and verify that the `Connect Websocket` event fires in both:
  - [ ] Amplitude - You can use the ingestion debugging page found [here](https://app.amplitude.com/data/parabol/Parabol%20Production/sources/development)
  - [ ] Segment - You can verify [here](https://app.segment.com/parabol-jordan/sources/actiondevelopment/debugger)
- Check that the `identify()` calls work properly by:
  - [ ] Amplitude - Confirm on the [User Lookup Page](https://app.amplitude.com/analytics/parabol/activity) that the connect websocket event has your `email`, `isActive`, `featureFlags`, `highestTier`, and `isPatient0` properties set.
  - [ ] Segment - Use the [debugger](https://app.segment.com/parabol-jordan/sources/actiondevelopment/debugger) to verify the same properties are set via an `identify()` call.

### DIFF
--- a/packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
+++ b/packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
@@ -51,19 +51,17 @@ const bootstrapNewUser = async (newUser: User, isOrganic: boolean, searchParams?
   ])
 
   // Identify the user so user properties are set before any events are sent
-  analytics.identify(
+  analytics.identify({
     userId,
-    {
-      createdAt,
-      email,
-      name: preferredName,
-      isActive: true,
-      featureFlags: experimentalFlags,
-      highestTier: tier,
-      isPatient0
-    },
-    segmentId
-  )
+    createdAt,
+    email,
+    name: preferredName,
+    isActive: true,
+    featureFlags: experimentalFlags,
+    highestTier: tier,
+    isPatient0,
+    anonymousId: segmentId
+  })
 
   const tms = [] as string[]
   if (isOrganic) {
--- a/packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
+++ b/packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
@@ -51,9 +51,9 @@ const bootstrapNewUser = async (newUser: User, isOrganic: boolean, searchParams?
   ])
 
   // Identify the user so user properties are set before any events are sent
-  segmentIo.identify({
+  analytics.identify(
     userId,
-    traits: {
+    {
       createdAt,
       email,
       name: preferredName,
@@ -62,8 +62,8 @@ const bootstrapNewUser = async (newUser: User, isOrganic: boolean, searchParams?
       highestTier: tier,
       isPatient0
     },
-    anonymousId: segmentId
-  })
+    segmentId
+  )
 
   const tms = [] as string[]
   if (isOrganic) {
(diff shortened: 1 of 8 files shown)


packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
```diff
@@ -62,8 +62,8 @@ const bootstrapNewUser = async (newUser: User, isOrganic: boolean, searchParams?
       highestTier: tier,
       isPatient0
     },
-    anonymousId: segmentId
```
tianrunhe, line 65: Can you double check today we are sending `anonymousId` from Segment to Amplitude and Amplitude is using that as the `deviceId`?

I checked your account, in PG your account has the segmentId `f8774b9d-xxxx...` but that doesn't match any deviceId in [Amplitude](https://app.amplitude.com/analytics/parabol/project/224689/search/amplitude_id%3D314837844648)?

tghanken: We're not sending it today.

What this will do is add the `segmentId` as an additional `deviceId` in Amplitude, allowing us to merge accounts and have more of an audit trail.  Once I get client tracking setup, I need to confirm that we're not leaving anonymous `deviceIds` hanging that should be tied to a `userId`, but am planning to address that in a follow-on PR.

nickoferrall (approved):

tianrunhe (approved):

packages/server/utils/analytics/amplitude/AmplitudeAnalytics.ts
```diff
@@ -16,6 +16,21 @@ export class AmplitudeAnalytics {
     })
   }
 
+  identify(userId: string, traits: Record<string, any>, anonymousId?: string) {
```
Dschoordsch, line 19: -1 Let's specify the traits to avoid developer mistakes when using this call. While we're at it, we can also include the `userId` so we just have one options object
```ts
type IdentifyOptions = {
  userId: string
  email: string
  name?: string
  isActive?: boolean
  featureFlags?: string[]
  highestTier?: string
  isPatient0?: boolean
  createdAt?: Date
}
```

Dschoordsch (changes requested):
Did not test

packages/server/utils/analytics/amplitude/AmplitudeAnalytics.ts
```diff
@@ -19,6 +19,26 @@ export class AmplitudeAnalytics {
     })
   }
 
+  identify(options: IdentifyOptions) {
+    // used as a failsafe for PPMIs
+    if (!AMPLITUDE_WRITE_KEY) return
+    const {userId, anonymousId, ...traits} = options
+    const identity = new Identify()
+
+    let trait: keyof Omit<IdentifyOptions, 'userId' | 'anonymousId'>
```
Dschoordsch, line 28: +1 a shorter option would be
```ts
let trait: keyof typeof traits
```

Dschoordsch (approved):
Did not verify it sends the correct amplitude calls

Issue #8657, opened 2023-08-14
Replace invite trust score with InboxReady's validate check


nickoferrall:
InboxReady has a `validate` API endpoint that could help us verify that an email address is not spam before sending the email (see [here](https://documentation.mailgun.com/en/latest/api-email-validation.html#email-verifications)). 

However, I suggest that we keep the domain blocklist.

By using the validate API, we can remove the restriction of sending no more than 10 email invites if none of the first 10 accepted an invite. This would fix this issue: https://github.com/ParabolInc/parabol/issues/8656

### AC

- [ ] Replace the custom trust score logic (see [here](https://github.com/ParabolInc/parabol/blob/ea927991096817ac245abb0fbf17b7e96b130a5e/packages/server/graphql/mutations/helpers/inviteToTeamHelper.ts#L26)) with the InboxReady validate check

Issue #8656, opened 2023-08-14
Can't send email invite to all team members


nickoferrall:
When creating a new team, you can check a checkbox to invite all of the team members in your org (or manually add their email addresses). 

However, our custom trust score logic prevents sending emails to more than 10 email addresses if none of the first 10 have accepted the email invite.

https://parabol.sentry.io/issues/4312988499/?referrer=github_plugin

```
Error: Cannot invite by email. Try using invite link
  File "./packages/server/graphql/mutations/inviteToTeam.ts", line 47, in exports["default"].resolve
    const error = new Error(inviteToTeamHelperRes.error.message);
  File "./packages/server/graphql/traceGraphQL.ts", line 156, in async Object.MutationinviteToTeamResolver
    const result = await scope.activate(field.span, () => resolve(source, args, context, info));
```

### AC

- [ ] When creating a new team, you can successfully invite more than 10 team members by email

PR #8707, opened 2023-08-21
feat(invitation-challenges): implement InboxReady's validate check

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8657 & fix https://github.com/ParabolInc/parabol/issues/8656

Loom Demo: https://www.loom.com/share/8ae8e1a69b29440a9f7f9e9bc5fe7bb7

### To test

- [ ] Enter valid emails in the invite email text input and see that they send successfully
- [ ] Enter invalid emails and see that they are not sent if Mailgun's validate check says `undeliverable` or `do_not_send`. The user gets feedback that those emails were not sent.


packages/server/__tests__/inviteToTeam.test.ts
```diff
@@ -19,7 +18,7 @@ test('Invite to team, works for ordinary email domain', async () => {
     `,
     variables: {
       teamId,
-      invitees: ['foo@example.com', inviteeEmail]
+      invitees: ['nick@parabol.co']
```
nickoferrall, line 21: We need to make sure the email addresses are valid now. The `signUp` function returns invalid emails.

packages/server/__tests__/inviteToTeam.test.ts
```diff
@@ -28,43 +27,7 @@ test('Invite to team, works for ordinary email domain', async () => {
...
-      invitees: ['foo@qq.com', inviteeEmail]
-    },
-    authToken
-  })
-
-  expect(inviteToTeam).toMatchObject({
-    data: {
-      inviteToTeam: {
-        error: {
-          message: 'Cannot invite by email. Try using invite link',
```
nickoferrall, line 65: We're no longer returning this error message from inviteToTeamHelper so I think it's safe to remove

packages/server/graphql/mutations/helpers/inviteToTeamHelper.ts
```diff
@@ -78,11 +37,27 @@ const inviteToTeamHelper = async (
...
+    (invitee) => !untrustedDomains.includes(getDomainFromEmail(invitee).toLowerCase())
+  )
+  const mailManager = new MailManagerMailgun()
+  const validateEmailRes = await Promise.all(
+    invitees.map((invitee) => mailManager.validateEmail(invitee))
+  )
+  const validInvitees = filteredInvitees.filter((invitee, index) => {
+    const validationResult = validateEmailRes[index]
+    if (!validationResult) return false
+    if (validationResult === 'undeliverable' || validationResult === 'do_not_send') {
```
nickoferrall, line 51: We're still trying to send emails if `validationResult` is `unknown`

packages/server/__tests__/inviteToTeam.test.ts
```diff
@@ -1,71 +0,0 @@
...
-      invitees: ['foo@example.com', inviteeEmail]
-    },
-    authToken
-  })
-
-  expect(inviteToTeam).toMatchObject({
-    data: {
-      inviteToTeam: {
-        error: null,
-        invitees: ['foo@example.com', inviteeEmail]
```
nickoferrall, line 31: We need to make sure the email addresses are valid now, but the `inviteeEmail` from `signUp` is invalid. 

If we include a real email address, we'll keep pinging that email when running the test. We could create a valid Parabol email just for this? E.g. tests@parabol.co

tianrunhe (changes requested):
Tested & hit this error:

Screenshot 2023-08-22 at 16 00 15

Server side log:
```
15:59:13 1|GraphQL Executor  | error: syntax error at or near ")"
15:59:13 1|GraphQL Executor  |     at Parser.parseErrorMessage (/Users/brucetian/Projects/parabol/node_modules/pg-protocol/dist/parser.js:287:98)
15:59:13 1|GraphQL Executor  |     at Parser.handlePacket (/Users/brucetian/Projects/parabol/node_modules/pg-protocol/dist/parser.js:126:29)
15:59:13 1|GraphQL Executor  |     at Parser.parse (/Users/brucetian/Projects/parabol/node_modules/pg-protocol/dist/parser.js:39:38)
15:59:13 1|GraphQL Executor  |     at Socket.<anonymous> (/Users/brucetian/Projects/parabol/node_modules/pg-protocol/dist/index.js:11:42)
15:59:13 1|GraphQL Executor  |     at Socket.emit (node:events:514:28)
15:59:13 1|GraphQL Executor  |     at Socket.emit (node:domain:489:12)
15:59:13 1|GraphQL Executor  |     at addChunk (node:internal/streams/readable:324:12)
15:59:13 1|GraphQL Executor  |     at readableAddChunk (node:internal/streams/readable:297:9)
15:59:13 1|GraphQL Executor  |     at Readable.push (node:internal/streams/readable:234:10)
15:59:13 1|GraphQL Executor  |     at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
```

But more importantly, after the above-mentioned issue is resolved, it means I can still send to tens of thousands of users with @qq.com? What I believe (from my limited😜 knowledge about China) about those spamming activities are that those invitee emails are valid email addresses...

nickoferrall:
> does it mean I can still send to tens of thousands of users with @qq.com

Nope, I've still included the untrusted domain check here: https://github.com/ParabolInc/parabol/blob/1dde7d70f6ed40d6d3f0a1facd568d407dab739f/packages/server/graphql/mutations/helpers/inviteToTeamHelper.ts#L40

I'm now away on holiday until next Thursday, but if you or anyone else want to pick this up, I think it'd be a good PR to get into prod sooner rather than later

tianrunhe:
> Nope, I've still included the untrusted domain check here:
> 
> https://github.com/ParabolInc/parabol/blob/1dde7d70f6ed40d6d3f0a1facd568d407dab739f/packages/server/graphql/mutations/helpers/inviteToTeamHelper.ts#L40
> 
Oh I see! Sorry I overlooked that. 

> I'm now away on holiday until next Thursday, but if you or anyone else want to pick this up, I think it'd be a good PR to get into prod sooner rather than later

No worries! I think the bandage I put on previously works for now:
* [Total email invites](https://app.amplitude.com/analytics/parabol/chart/new/uno739gs) are stable after last time we've seen spammers on Aug 6th
* [Zooming in](https://app.amplitude.com/analytics/parabol/chart/xv8lafw7) for invites sent directly to those domains also show zero last few weeks
Enjoy your days off!

tianrunhe (approved):
Tested & worked

Dschoordsch (changes requested):
See comment in your demo: We're not giving adequate feedback if there are valid and invalid emails. As a user I would not be sure if the invites to the valid users went out.

I did not check the code yet, but it seems to have a lot of merge artefacts? Could you rebase this, please?

nickoferrall:
> We're not giving adequate feedback if there are valid and invalid emails. As a user I would not be sure if the invites to the valid users went out.

This is the same UX that exists in prod. This PR just changes the way we check whether the email address is valid. Rather than using our custom `getInviteTrustScore`, we're using InboxReady's validation API.

I think we could change the UX, but that should be a separate design issue. 

> Could you rebase this, please?

Sure, I've just merged with master and it's cleaned up now 👍

packages/server/__tests__/inviteToTeam.test.ts
Dschoordsch: -1 We're not testing with `MailManagerMailgun`, do we?

packages/server/graphql/mutations/helpers/inviteToTeamHelper.ts
```diff
@@ -78,11 +37,30 @@ const inviteToTeamHelper = async (
...
   const subOptions = {mutatorId, operationId}
 
-  const trustScore = await getInviteTrustScore(viewerId, teamId, invitees, dataLoader)
-  if (trustScore < 0.15)
-    return {error: {message: 'Cannot invite by email. Try using invite link'}, invitees: []}
+  const untrustedDomains = ['tempmail.cn', 'qq.com']
+  const filteredInvitees = invitees.filter(
+    (invitee) => !untrustedDomains.includes(getDomainFromEmail(invitee).toLowerCase())
+  )
+  const mailManager = new MailManagerMailgun()
```
Dschoordsch, line 44: -1 We should not hardcode it to this as we don't want to hit the network in tests and local development if not necessary.

packages/server/graphql/mutations/helpers/inviteToTeamHelper.ts
```diff
@@ -78,11 +37,30 @@ const inviteToTeamHelper = async (
...
+  const validInvitees = filteredInvitees.filter((invitee, index) => {
+    const validationResult = validateEmailRes[index]
+    if (!validationResult) return false
+    if (validationResult === 'undeliverable' || validationResult === 'do_not_send') {
+      const error = new Error(`Email validation error: ${validationResult}`)
+      sendToSentry(error, {tags: {invitee}})
+      return false
+    }
+    return true
+  })
```
Dschoordsch, line 57: +1 a slightly cleaner way than iterating through the list twice and matching by index would be to do sth. like this (pseudo code):
```ts
const validInvitess = (await Promise.all(
  invitess.map(async (invitee) => {
    const validationResult = await mailManager.validateEmail(invitee))
    if (!validationResult || validationResult === 'undeliverable'...) {
     ..
     return error
    }
    return invitee
  })
)).filter(isValid)
```

packages/server/graphql/mutations/helpers/inviteToTeamHelper.ts
```diff
@@ -109,6 +87,7 @@ const inviteToTeamHelper = async (
       return getIsEmailApprovedByOrg(email, orgId, dataLoader)
     })
   )
+  console.log('🚀 ~ approvalErrors:', approvalErrors)
```
Dschoordsch, line 90: -1 Do we need this here? Also the rocket seems inappropriate.

Dschoordsch (changes requested):

Dschoordsch:
We should add the `validateEmail` function to the `MailManager` interface so we can continue testing locally without hitting Mailgun.

packages/server/__tests__/inviteToTeam.test.ts
nickoferrall: Could you elaborate, please? 

The test is testing `inviteToTeam`, which is using `inviteToTeamHelper`. In the test, we have the `signUp` helper function with provides a random email address.

Previously, our custom logic would send emails to any email address (excluding tempmail.cn and qq.com domains). 

Now, that we're using Mailgun's validate logic in `inviteToTeamHelper`, we won't send emails to the randomly generated ones if they're undeliverable. 

Here's an example response where the test fails:

```
      Object {
        "data": Object {
          "inviteToTeam": Object {
    -       "error": null,
    -       "invitees": Array [
    -         "foo@example.com",
    -         "eusebio_koelpin82@yahoo.com",
    -       ],
    +       "error": Object {
    +         "message": "No valid emails",
    +         "title": null,
    +       },
    +       "invitees": null,
          },
        },
      }
```

packages/server/graphql/mutations/helpers/inviteToTeamHelper.ts
nickoferrall: What do you suggest instead? Check if it's production and only use `MailManagerMailgun` if it is? Locally & in tests, skip the validation?

nickoferrall: Just following up on this message and [this one](https://github.com/ParabolInc/parabol/pull/8707#discussion_r1326020199) in case you haven't seen them @Dschoordsch

Dschoordsch: We have [`getMailManager`](https://github.com/ParabolInc/parabol/blob/434f85f016c8d7e7896c7578b353fdc95deff2b3/packages/server/email/getMailManager.ts#L17) which returns the correct manager based on environment. I think you should extend the `MailManager` interface with the new `validateEmail(...)` function. You could either make it optional and only call it if it's present, or you could make a dummy implementation in the others which returns true based on some basic filters or similar.

Dschoordsch (changes requested):
All functions except `getMailManager` should work with the `MailManager` interface, see my previous comment.

packages/server/graphql/mutations/helpers/inviteToTeamHelper.ts
nickoferrall: Thanks, that makes sense. Updated 👍

packages/server/email/MailManager.ts
```diff
@@ -15,4 +17,7 @@ export interface MailManagerOptions {
 }
 export default abstract class MailManager {
   abstract sendEmail(options: MailManagerOptions): Promise<boolean>
+  async validateEmail(_email: string): Promise<false | MailgunValidateResult> {
+    return 'deliverable'
```
Dschoordsch, line 21: +1 `unknown` would be the correct state, wouldn't it?

packages/server/email/MailManagerMailgun.ts
```diff
@@ -30,4 +37,15 @@ export default class MailManagerMailgun extends MailManager {
...
   }
+
+  async validateEmail(email: string) {
+    try {
+      const res = await this.mailgunClient.validate.get(email)
+      return res.result as MailgunValidateResult
+    } catch (e) {
+      const error = e instanceof Error ? e : new Error('Mailgun failed to validate emails')
+      sendToSentry(error, {tags: {type: 'Mailgun error'}})
+      return false
```
Dschoordsch, line 48: -1 Why should this method have the return type `Promise<false | MailgunValidationResult>`, we could just return `unknown` here.

packages/server/email/MailManager.ts
```diff
@@ -15,4 +17,7 @@ export interface MailManagerOptions {
 }
 export default abstract class MailManager {
   abstract sendEmail(options: MailManagerOptions): Promise<boolean>
+  async validateEmail(_email: string): Promise<false | MailgunValidateResult> {
```
Dschoordsch, line 20: -1 The return type is not very clear for the caller of this function. I think we should just have a boolean send/do not send and leave the interpretation of the different responses to the mail manager implementations.
If you think that's not enough information, we could have a separate type here in the interface
```ts
export type ValidationResult =
  | 'valid'
  | 'invalid'
  | 'risky'
  | 'unknown'
```

Dschoordsch (changes requested):
The return type of `validateEmail` needs some cleanup, otherwise it looks good. Tested, works. Feel free to merge once that's done.

packages/server/email/MailManager.ts
nickoferrall: Sounds good! I've updated the return type to a boolean which fixes the other comments too

Issue #8643, opened 2023-08-10
Activity Library: same template appears a few times


nickoferrall:
Looking at my activity library, I can see `Original 4` shows up three times, `Start, Stop, Continue` twice, and `Working & Stuck` twice:

![activity-library](https://github.com/ParabolInc/parabol/assets/39854876/c1a51102-1dde-4254-8aca-060fb4d473d5)

If we only show the same template once, we can show a wider variety of templates in the same real estate. This might encourage them to try a new template that they prefer or a premium template. 

### AC

- [ ] Templates only show up once in the activity library

PR #8708, opened 2023-08-21
chore: first pass at lastUsedAt caching

### LINKED ISSUE #8581: Improving Category query in the new Activity Library
Improve `ActivityLibraryQuery ` to prevent lag and RethinkDB taking too many hits.

For Context:
https://parabol.slack.com/archives/C042HTVBH8D/p1690389143937259

mattkrick:
# Description

fix #8581 speeds up queries for recently used activities

Before merging I think we should add in a little extra logic so templates can't appear in both popular and recently used, but I wanna leave out changes until we like the refactor.

## Demo

No visual difference 😕 
https://www.loom.com/share/e64150215e9d4067bd6c8232f48d4b8c?sid=1b6e632a-b893-48d2-97b1-1cecd0442ac1

## Testing scenarios

- [ ] look at the retros in your activity library before you run this branch, then checkout this branch & run the migration. you shouldn't see a change!

### DIFF
--- a/packages/server/graphql/public/types/ReflectTemplate.ts
+++ b/packages/server/graphql/public/types/ReflectTemplate.ts
@@ -57,7 +57,6 @@ const ReflectTemplate: ReflectTemplateResolvers = {
       (tmt) => tmt.templateId === id
     )
 
-    //
     const lastUsedAtOnTeam = Math.max(
       ...orgTeamMeetingTemplatesForTemplateId
         .filter((row) => authToken.tms.includes(row.teamId))
--- a/packages/server/graphql/public/types/ReflectTemplate.ts
+++ b/packages/server/graphql/public/types/ReflectTemplate.ts
@@ -1,6 +1,8 @@
 import ms from 'ms'
 import {getUserId} from '../../../utils/authorization'
 import errorFilter from '../../errorFilter'
+import {DataLoaderWorker} from '../../graphql'
+import isValid from '../../isValid'
 import {ReflectTemplateResolvers} from '../resolverTypes'
 
 const POPULAR_RETROS = [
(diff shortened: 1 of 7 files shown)


mattkrick:
@jmtaber129 lemme know if you like the direction this is headed & we can clean it up. not sure why the build is failing, assuming it has to do with 0 Teams existing in the CI DB?

jmtaber129 (commented):
This approach LGTM! We'll also need to upsert `TeamMeetingTemplate` entries when meetings are created, right?

mattkrick:
doh! yessir, i'll drop that in.

mattkrick:
@jmtaber129 ready for review! What are your thoughts on deduping templates so they only appear in 1 subCategory at a time?

packages/server/graphql/public/types/ReflectTemplate.ts
```diff
@@ -64,17 +46,39 @@ const ReflectTemplate: ReflectTemplateResolvers = {
...
+
+    const lastUsedAtOnOrg = Math.max(
+      ...orgTeamMeetingTemplatesForTemplateId
+        .filter((row) => !authToken.tms.includes(row.teamId))
+        .map((row) => row.lastUsedAt.getTime())
+    )
+
+    if (lastUsedAtOnOrg > Date.now() - ms('30d')) {
       subCategories.push('recentlyUsedInOrg')
     }
```
jmtaber129, line 78: +1 Consider pulling a lot of this back into the `getLastUsedDateForTeams` helper? I _think_ we could move [these lines](https://github.com/ParabolInc/parabol/pull/8708/files#diff-15767a301911509b2ae53ecd359954a07080796e808d229142a243825450fae8R49-R64) into that helper, and then do
```
const [lastUsedAtOnTeam, lastUsedAtOnOrg] = await Promise.All([
  getLastUsedDateForTeams(authToken.tms, id, dataLoader),
  getLastUsedDateForTeams(
      orgTeamIds.filter((orgId) => !authToken.tms.includes(orgId)),
      id,
      dataLoader
    )
])
```
and still have DataLoader batch it into a single DB call (I haven't tested this though).

jmtaber129 (approved):
LGTM aside from one relatively minor suggestion! And thanks for taking care of this for us!

jmtaber129:
> What are your thoughts on deduping templates so they only appear in 1 subCategory at a time?

Nick added an [issue](https://github.com/ParabolInc/parabol/issues/8643) for this not too long ago, but I'm not sure I agree (e.g. in Netflix [an inspiration for this], listings often appear in both "Popular" and "Trending", and other categories) - I'd like to get more feedback supporting this suggestion before taking action.

This would likely be done in the frontend anyway to ensure unrendered templates in one subcategory aren't "removed" from another subcategory where they would be rendered otherwise, so this change would be mostly unrelated to this PR anyway.

packages/server/graphql/public/types/ReflectTemplate.ts
mattkrick: done!

Issue #8566, opened 2023-07-25
One on one: organization picker


igorlesnenko:
<img width="562" alt="image" src="https://github.com/ParabolInc/parabol/assets/466991/1dac72bf-8978-4f56-83f0-490197b77756">

**AC:** 

- If email selected or selected user belongs to 2 or more my different orgs, show org selector.

[Designs](https://www.figma.com/file/Vc09AD3hxef2ypriihfK0T/Flow---ad-hoc-teams?type=design&node-id=0-1&mode=design&t=dU3K6kFeVzv8P33A-0)

PR #8712, opened 2023-08-22
chore(one-on-one): show organiation picker if can't determine org automatically

igorlesnenko:
Fixes: #8566

**How to test:**

- add `oneOnOne` org level feature flag
- Go to activity library and select one on one meeting
- Pick a user with whom you share only one common organization, start a meeting, see that the team is created or (reused) in the correct organization
- Select a user with whom you are together in two or more organizations. See that org picker appeared.
- Enter a custom email, see that org picker appeared
- Test that not one-on-one checkin is still working

image


nickoferrall:
Sorry @igorlesnenko, I won't be able to do a review tonight, and I'm away tomorrow for a week, so it's probably best to request a different reviewer on this one

packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
```diff
@@ -136,12 +147,27 @@ const ActivityDetailsSidebar = (props: Props) => {
...
+  const newOneOnOneTeamOrgId = manuallySelectedOrgId
+    ? manuallySelectedOrgId
+    : firstMutualOrgId ?? defaultOrgId
 
   const oneOnOneTeamInput = selectedUsers[0]
     ? {
         email: selectedUsers[0].email,
-        // TODO: always pick default orgId for now, this will be changed in next PRs
-        orgId: selectedTeam.orgId
+        orgId: newOneOnOneTeamOrgId
```
igorlesnenko, line 170: This PR introduces fully working version of one on one meeting.

Earlier I added `isOneOnOneTeam` flag here: https://github.com/ParabolInc/parabol/pull/8629/commits/f8a2d00188d7a91f183e38eb3e89f67c35bb289a

@tghanken Do we need to track any additional events?

tghanken: Can we add the `isOneOnOneTeam` property to the Segment events `New Team`, `Meeting Started`, `Meeting Joined`, and `Meeting Completed`?  I think that should cover it.

tghanken (changes requested):
Looks good once we add the property to Segment events.

packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
```diff
@@ -260,6 +286,29 @@ const ActivityDetailsSidebar = (props: Props) => {
...
+                  <Select
+                    onValueChange={(orgId) => setSelectedOrgId(orgId)}
+                    value={newOneOnOneTeamOrgId}
+                  >
+                    <SelectTrigger className='bg-white'>
+                      <SelectValue />
+                    </SelectTrigger>
+                    <SelectContent>
+                      <SelectGroup>
+                        {viewerOrganizations.map((org) => (
```
jmtaber129, line 302: -1 This should be `mutualOrgs`, no? Or if we want to allow users to create a team on an org the other user isn't a part of, we should _always_ show this org picker

packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
```diff
@@ -136,12 +147,27 @@ const ActivityDetailsSidebar = (props: Props) => {
...
+  const mutualOrgs = selectedUserOrganizationIds.filter((orgId) => viewerOrganizationIds.has(orgId))
+
+  const showOrgPicker = selectedUser && (mutualOrgs.length > 1 || !mutualOrgs.length)
+
+  const manuallySelectedOrgId = showOrgPicker && selectedOrgId !== '' && selectedOrgId
+  const firstMutualOrgId = mutualOrgs[0]
+  const defaultOrgId = selectedTeam.orgId
+  const newOneOnOneTeamOrgId = manuallySelectedOrgId
+    ? manuallySelectedOrgId
+    : firstMutualOrgId ?? defaultOrgId
```
jmtaber129, line 165: -1 This logic is a bit dense, and I think there's some opportunity to simplify here. Here's a general approach that I'd recommend:
* Determine the default org ID (that is, `mutualOrgsIds[0] ?? selectedTeam.orgId`) and use that as initial state for `selectedOrgId`
* When mutual orgs change, determine a new default `selectedOrgId` iff the current `selectedOrgId` is _not_ in the new mutual orgs array

This way, `selectedOrgId` is clearly the single source of truth for which org is currently selected, and we still update it automatically if it needs to be changed.

packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
```diff
@@ -136,12 +147,27 @@ const ActivityDetailsSidebar = (props: Props) => {
   } = useDialogState()
 
   const [selectedUsers, setSelectedUsers] = React.useState<Option[]>([])
+  const selectedUser = selectedUsers[0]
+  const {organizations: viewerOrganizations} = viewer
+
+  const viewerOrganizationIds = new Set(viewerOrganizations.map((org) => org.id))
+
+  const selectedUserOrganizationIds = selectedUser?.organizationIds ?? []
+  const mutualOrgs = selectedUserOrganizationIds.filter((orgId) => viewerOrganizationIds.has(orgId))
```
jmtaber129, line 156: +1 `mutualOrgs` here is actually `mutualOrgIds`, but I'd actually suggest you refactor this so that this holds org objects, rather than just renaming:
```suggestion
  const selectedUserOrganizationIds = new Set(selectedUser?.organizationIds ?? [])
  const mutualOrgs = viewerOrganizations.filter((orgId) => selectedUserOrganizationIds.has(orgId))
```

jmtaber129 (changes requested):
I left a comment in-line, but it's a little unclear what the desired behavior is w.r.t. which orgs can be used to create an ad-hoc team with another user - should it always be on a mutual org if one exists? do we want to allow creation on non-mutual orgs? what about users being invited by email?

Once this is clear, it'll be easier for me to identify what changes may be needed to get this merged.

packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
igorlesnenko: @tghanken added ✅

tghanken (approved):

igorlesnenko:
@jmtaber129 The approach is the following:
- If you enter a custom email, we allow to select any org 
- If you select a user with more than 1 mutual org, you should select one of the mutual orgs (fixed it)
- If you only have 1 mutual org, we don't show an org picker, and that org will be selected automatically

packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
```diff
@@ -131,12 +141,27 @@ const ActivityDetailsSidebar = (props: Props) => {
...
 
   const [selectedUsers, setSelectedUsers] = React.useState<Option[]>([])
+  const selectedUser = selectedUsers[0]
+  const {organizations: viewerOrganizations} = viewer
+
+  const selectedUserOrganizationIds = new Set(selectedUser?.organizationIds ?? [])
+  const mutualOrgs = viewerOrganizations.filter((org) => selectedUserOrganizationIds.has(org.id))
+  const mutualOrgsIds = mutualOrgs.map((org) => org.id)
+
+  const showOrgPicker = selectedUser && (mutualOrgs.length > 1 || !mutualOrgs.length)
```
jmtaber129, line 151: +1 Simplify to `selectedUser && mutualOrgs.length !== 1`.

packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
```diff
@@ -250,6 +275,30 @@ const ActivityDetailsSidebar = (props: Props) => {
...
+                  <Select onValueChange={(orgId) => setSelectedOrgId(orgId)} value={selectedOrgId}>
+                    <SelectTrigger className='bg-white'>
+                      <SelectValue />
+                    </SelectTrigger>
+                    <SelectContent>
+                      <SelectGroup>
+                        {viewerOrganizations
+                          .filter((org) =>
+                            mutualOrgsIds.length ? mutualOrgsIds.includes(org.id) : true
+                          )
```
jmtaber129, line 291: +1 This has the same result as [this logic](https://github.com/ParabolInc/parabol/pull/8712/files#diff-b9436085c8610b2c6e1919f190475b18559cbbd0a80bc8f229eb740ed177ada2R148) - let's just use `mutualOrgs` here instead of filtering on `viewerOrganizations` a second time.

jmtaber129 (approved):
Overall looks good, but there are a few minor comments I left in-line.

+1 Maybe consider pulling the one-on-one picker logic into its own component, too. I.e. everything in the `selectedTemplate.id === 'oneOnOneAction'` true render block and all the selected user + selected org state, and then use that component to drive `oneOnOneTeamInput`.

packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
```diff
@@ -66,6 +72,10 @@ const ActivityDetailsSidebar = (props: Props) => {
           gcal
         }
         ...AdhocTeamMultiSelect_viewer
+        organizations {
```
Dschoordsch, line 75: +1 this is not updated when a user joins an org, so I assume there is some subscription missing. To test this invite a user with 1 common team so you have 2 in common and then start the meeting without refreshing.

Dschoordsch (approved):
Mostly works. I think it would be nice if we'd treat blurring the team member select the same as enter:
https://github.com/ParabolInc/parabol/assets/7331043/8f14f14e-b66c-4515-a952-37681a9f2dc2
Without this, the input is just floating there without being real input. Also I've noticed I could start a one on one meeting without inviting anyone, so really a one only meeting.

igorlesnenko:
>  Also I've noticed I could start a one on one meeting without inviting anyone, so really a one only meeting.

Fixed in next PR and added email validation #8718

Issue #8717, opened 2023-08-23
One on one: add email validation


igorlesnenko:
we allow to put any custom value to user picker, make sure it is a valid email

PR #8718, opened 2023-08-23
chore(one-on-one): Add team exists warning

igorlesnenko:
Fixes: #8562
Fixes: #8717
Fixes: #8814

- Adds a warning that an one-on-one team already exists within the org and will be reused (or new team will be created if not)
- Adds email validation
- Avatars added to the dropdown menu

Full demo: https://www.loom.com/share/cc57b4b73e3b4ff19ba4cf6c6a30805f

image

image

**How to test:**

- Open one on one meeting page
- Enter custom email, see that a correct warning appears
- Select a user, see that a correct warning appears
- Select a user with whom you already had a meeting in one-on-one team, see that a correct warning appears (Note that now we also check one on one team by `isOneOnOneTeam` flag, so previously created teams might not be counted)
- See that a warning matches the an actual team that will be created/used
- Try to enter incorrect email, see it is not allowed
- See that you can still create simply check-in meeting without issues


igorlesnenko:
@acressall this PR adds last UI fixes to one-on-one, here is the full demo: https://www.loom.com/share/cc57b4b73e3b4ff19ba4cf6c6a30805f

Can you please review?

packages/client/components/ActivityLibrary/IsOneOnOneTeamExists.tsx
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+`
+
+interface ComponentProps {
+  queryRef: PreloadedQuery<IsOneOnOneTeamExistsQuery>
+  name: string
+}
+
+const IsOneOnOneTeamExistsComponent = (props: ComponentProps) => {
```
nickoferrall, line 28: +1 we usually create a separate file for a component like this. I think that'd be more consistent with the rest of the codebase

nickoferrall (approved):
I love this feature!

igorlesnenko:
changed meeting name image

acressall (approved):

packages/client/components/ActivityLibrary/IsOneOnOneTeamExists.tsx
Dschoordsch, line 1: +1 `OneOnOneTeamStatus` might be a better name.

packages/server/graphql/public/typeDefs/User.graphql
```diff
@@ -275,7 +275,12 @@ type User {
   """
   Get the list of all organizations a user belongs to
   """
-  organizations: [Organization!]!
+  organizations(
+    """
+    Filter by organization id
+    """
+    id: ID
+  ): [Organization!]!
```
Dschoordsch, line 283: -1 I think as we only allow for 1 id input, there should be a separate singular field `organization(id: ID!)` like we have `teams` and `team` on `User`

Dschoordsch (changes requested):

packages/server/graphql/public/typeDefs/User.graphql
igorlesnenko: @Dschoordsch make sense, just noticed we already had it implemented, fixed!

tghanken (approved):
Looks good from an analytics standpoint.

packages/server/graphql/public/typeDefs/Organization.graphql
```diff
@@ -137,6 +137,11 @@ type Organization {
   """
   orgUserCount: OrgUserCount!
 
+  """
+  Returns one on one team if exists
+  """
+  oneOnOneTeam(oneOnOneTeamInput: CreateOneOnOneTeamInput!): Team
```
Dschoordsch, line 143: -1 this field should only get the `email` of the other user as input, as the org is already given by the parent

Dschoordsch (changes requested):

Dschoordsch (approved):
Did not retest

PR #8728, opened 2023-08-24
fix: user tasks should not return from archived teams

tianrunhe:
# Description

## Demo

## Testing scenarios

- [ ] Scenario A
  - Step 1
  - Step 2...

- [ ] Scenario B
  - Step 1
  - Step 2....

## Final checklist

### DIFF
--- a/packages/server/graphql/getAccessibleTeamIdsForUser.ts
+++ b/packages/server/graphql/getAccessibleTeamIdsForUser.ts
@@ -14,7 +14,6 @@ export const getAccessibleTeamIdsForUser = async (
   teamIds: null | string[],
   dataLoader: DataLoaderWorker
 ) => {
-  const userTeamMembers = await dataLoader.get('teamMembersByUserId').load(userId)
-  const userTeamIds = userTeamMembers.map(({teamId}) => teamId)
-  return teamIds?.length ? teamIds.filter((teamId) => userTeamIds.includes(teamId)) : userTeamIds
+  const user = await dataLoader.get('users').loadNonNull(userId)
+  return teamIds?.length ? teamIds.filter((teamId) => user.tms.includes(teamId)) : user.tms
 }


tianrunhe:
@Dschoordsch I separated the call so timeline page would include archived teams whereas others have only active teams. However it occurs to me we also need to separate that in the `TeamFilter` component: if the timeline include events from the archived teams, users should expect to see the team filter also include them, right? That's why this PR becomes a bit larger. Let me know if there are better ways to handle it! Thank you!

packages/client/components/MyDashboardTimeline.tsx
```diff
@@ -47,6 +47,7 @@ const MyDashboardTimeline = (props: Props) => {
         $userIds: [ID!]
         $eventTypes: [TimelineEventEnum!]
         $teamIds: [ID!]
+        $includeArchived: Boolean = true
```
Dschoordsch, line 50: +1 Instead of setting a default here, I would prefer to just have the variable defined and then we pass the value explicitly. This way I only have to check in one place and not multiple what this value will be like.

packages/client/components/MyDashboardTasksAndHeader.tsx
```diff
@@ -13,7 +13,12 @@ const MyDashboardTasksAndHeader = (props: Props) => {
   const {queryRef} = props
   const data = usePreloadedQuery<MyDashboardTasksAndHeaderQuery>(
     graphql`
-      query MyDashboardTasksAndHeaderQuery($after: DateTime, $userIds: [ID!], $teamIds: [ID!]) {
+      query MyDashboardTasksAndHeaderQuery(
+        $after: DateTime
+        $userIds: [ID!]
+        $teamIds: [ID!]
+        $includeArchived: Boolean = false
```
Dschoordsch, line 20: +1 see later, I would prefer not having the default here

Dschoordsch (changes requested):
It seems to break a subscription. When I now create (or archive) a team, it does not appear (or disappear) from the sidebar without a refresh

tianrunhe:
@Dschoordsch I would like to make this PR smaller and not to increase the scope so it can be included in this week's release. Hence I reverted my additional changes on showing archived teams in the team filters. We haven't heard any customer feedback on wanting the archived teams to be shown in the team filter so I guess we can postpone that to a later time. We do hear issues on the tasks from archived teams shown up on the board and this PR is going to fix that specific issue so let's focus on that.
WDYT? If you are fine with this, I'll create an issue for that later. Thanks for your time!

packages/server/graphql/types/User.ts
```diff
@@ -595,17 +598,29 @@ const User: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<any, GQLC
...
         {authToken, dataLoader}: GQLContext
       ) => {
         const viewerId = getUserId(authToken)
         const user = (await dataLoader.get('users').load(userId))!
-        const teamIds =
+        const activeTeamIds =
           viewerId === userId || isSuperUser(authToken)
             ? user.tms
             : user.tms.filter((teamId: string) => authToken.tms.includes(teamId))
+        const allTeamIds = (await dataLoader.get('teamMembersByUserId').load(userId)).map(
```
Dschoordsch, line 620: -1 Let's only load this if `includeArchived`

packages/server/graphql/queries/tasks.ts
```diff
@@ -124,7 +123,10 @@ export default {
 
     // if archived is true & no userId filter is provided, it should include tasks for ex-team members
     // under no condition should it show tasks for archived teams
-    const validTeamIds = await getAccessibleTeamIdsForUser(viewerId, teamIds, dataLoader)
+    const accessibleTeamIds = (await dataLoader.get('users').loadNonNull(viewerId)).tms
```
Dschoordsch, line 126: -1 we can just use `authToken.tms`

Dschoordsch (changes requested):
I like the simplification a lot and I agree, there was no request for the filter. I wouldn't create an issue to create it.

Only 2 small fixes and it's good to go. Feel free to merge it as is and do them in a follow up.

PR #8746, opened 2023-08-29
chore: remove __PROJECT_ROOT__

### LINKED ISSUE #8202: `yarn build --no-deps` requires queryMap & yarn.lock
AC
-  [ ] queryMap & yarn.lock get compiled into the build
-  [ ] queryMap should be able to get imported
-  [ ] yarn.lock is probably used to determine the project root. we can instead re-use the `__PROJECT_ROOT__` webpack var.
-  [ ] In fact, let's completely remove `getProjectRoot` in favor of a webpack var

NOTE: government clients still need to scan yarn.lock so we might need to leave that in here, but we should _not_ depend on it

mattkrick:
# Description

Fixes #8202

`__PROJECT_ROOT__` is no longer possible since we are doing so much work in predeploy vs build

## Demo

If it builds in CI, it works!

### DIFF
--- a/packages/server/types/modules.d.ts
+++ b/packages/server/types/modules.d.ts
@@ -25,7 +25,6 @@ declare module 'json2csv/lib/JSON2CSVParser'
 declare module 'object-hash'
 declare module 'string-score'
 
-declare const __PROJECT_ROOT__: string
 declare const __APP_VERSION__: string
 declare const __PRODUCTION__: string
 declare const __SOCKET_PORT__: string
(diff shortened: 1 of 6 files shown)


rafaelromcar-parabol:
Do not forget removing [this line](https://github.com/ParabolInc/parabol/blob/master/docker/parabol-ubi/docker-build/dockerfiles/pipeline.dockerfile#L103) and also maybe adapting the comment of the line for the yarn.lock and also do not forget testing the build pipeline with a commit that starts by "dockerize" 😸

mattkrick:
they could be the same cuz they're so small! They do their own thing, so i figured easier to let em stay separate

PR #8747, opened 2023-08-29
fix: move PRODUCTION bool to build time

### LINKED ISSUE #8627: chore: on server, Replace PROD with __PRODUCTION__
`PROD` is a remnant of the old days before we webpacked our server. Now that we run it through webpack first, we can use __PRODUCTION__ and that'll tree shake a few conditionals.

AC
- [] remove `PROD` and replace it with `__PRODUCTION__` on the server

mattkrick:
# Description

fix #8627

`__PRODUCTION__` is a webpack variable that is calculated at build time. 
Since this var is a constant at runtime, webpack can tree shake conditional branches & remove bits of unreachable code, e.g. `if (!__PRODUCTION__) doThing()` gets removed entirely & `doThing()` may also get removed if it is not used elsewhere.

## Demo

If CI passes it works!

### DIFF
--- a/packages/server/graphql/CompiledQueryCache.ts
+++ b/packages/server/graphql/CompiledQueryCache.ts
@@ -2,7 +2,6 @@ import tracer from 'dd-trace'
 import {GraphQLSchema, parse} from 'graphql'
 import {CompiledQuery} from 'graphql-jit'
 import getRethink from '../database/rethinkDriver'
-import PROD from '../PROD'
 import {MutationResolvers, QueryResolvers, Resolver} from './public/resolverTypes'
 import {tracedCompileQuery} from './traceGraphQL'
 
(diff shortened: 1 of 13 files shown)


tghanken:
I'm going to remove myself as a reviewer, since this seems like a developer change 😂.

mattkrick:
ah good catch, bad github bot

packages/server/graphql/CompiledQueryCache.ts
```diff
@@ -45,7 +44,7 @@ export default class CompiledQueryCache {
     if (compiledQuery) return compiledQuery
     const r = await getRethink()
     let queryString = await r.table('QueryMap').get(docId)('query').default(null).run()
-    if (!queryString && !PROD) {
+    if (!queryString && !__PRODUCTION__) {
```
mattkrick, line 47: Verified working. Here's what it looks like when we run `yarn build`. No reference to `queryMap.json` in prod 🎉 

Screenshot 2023-09-06 at 3 27 15 PM

mattkrick:
verified working when built!

Issue #8714, opened 2023-08-22
Gcal Integration: handle Zoom Add-On


nickoferrall:
@kendradixon reported [here](https://parabol.slack.com/archives/C0161UFBM6W/p1692727306554239?thread_ts=1692723341.774589&cid=C0161UFBM6W) that when you use the Zoom Add-On after creating a gcal event from Parabol, it results in two URLs in the location field:

![Screenshot 2023-08-22 at 19 27 13](https://github.com/ParabolInc/parabol/assets/39854876/657568fe-f5d8-47dd-9c49-e0f54c6406e0)

Perhaps the Parabol meeting URL should be added to the description field in Gcal? The ideal solution is building a Zoom integration so users can do everything from Parabol.

### AC

- [ ] Handle Zoom Add-On so that there aren't two URLs in the location field

PR #8776, opened 2023-09-05
fix(gcal): handle Zoom add-on conflict

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8714

Demo: https://www.loom.com/share/9923a47d690d4e26ad5369faaa6fc94c

### To test

- [ ] Create a Parabol meeting using the gcal modal
- [ ] See that the Parabol meeting link is added to the description field, not the location field

### DIFF
--- a/packages/server/graphql/mutations/helpers/createGcalEvent.ts
+++ b/packages/server/graphql/mutations/helpers/createGcalEvent.ts
@@ -26,7 +26,7 @@ const createGcalEvent = async (input: Input) => {
   if (!featureFlags.includes('gcal')) {
     return standardError(new Error('Does not have gcal feature flag'), {userId: viewerId})
   }
-  const {startTimestamp, endTimestamp, title, description, timeZone, invitees} = gcalInput
+  const {startTimestamp, endTimestamp, title, timeZone, invitees} = gcalInput
 
   const gcalAuth = await dataLoader.get('freshGcalAuth').load({teamId, userId: viewerId})
   if (!gcalAuth) {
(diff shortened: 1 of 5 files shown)


acressall (approved):

packages/server/graphql/public/types/CreateGcalEventInput.ts
Dschoordsch, line 1: +2 I like that we simplified this and got rid of the description

packages/server/graphql/mutations/helpers/createGcalEvent.ts
```diff
@@ -47,10 +47,12 @@ const createGcalEvent = async (input: Input) => {
   const calendar = google.calendar({version: 'v3', auth: oauth2Client})
   const meetingUrl = makeAppURL(appOrigin, `meet/${meetingId}`)
```
Dschoordsch, line 48: +1 Going forward we should think about what to do with the participants. I don't like that right now they're separate from your team members, while on the other hand I like that it's easy to bring others into a parabol meeting this way. Going forward we probably should exchange the invitee input field with the one Igor built to allow a mix of existing and new members. In the meantime we should probably replace the meeting link with an invite link to make it easier for folks to get in.

Dschoordsch (approved):

packages/server/graphql/mutations/helpers/createGcalEvent.ts
nickoferrall: > Going forward we probably should exchange the invitee input field with the one Igor built to allow a mix of existing and new members

Yeah, I agree. I really like the input field Igor & Alicia worked on and we should probably add it to all meeting types and think about how it works alongside the gcal scheduler. 

> In the meantime we should probably replace the meeting link with an invite link to make it easier for folks to get in.

If we include the invite link, it would take the users to the most recent meeting, right? But that might not be the correct meeting, which could be confusing.

We could include both the meeting link and the invite link in the gcal description? 

There could be privacy concerns with including the invite link as you might be able to access someone's gcal calendar, but shouldn't access their Parabol team. For example, there's an Exco team that discusses sensitive info that I don't have access to, but I can view everyone's Parabol gcal calendar, so if we include the invite link in the gcal event, I could join that meeting. 

Still, I think that's probably fine. We could include a checkbox in the gcal modal UI to include the invite link. 

I can merge this PR to make sure it lands in the next release and see what the Squad thinks about including the invite link.

Issue #8777, opened 2023-09-05
Export summary to pdf


nickoferrall:
Can export a meeting summary to a pdf so users can share the content of a meeting without inviting users to their team. For further context: https://www.notion.so/parabol/Exportable-summaries-turn-it-into-PDF-or-similar-for-sharing-with-leadership-team-other-leads-etc-4c7e6c7f9d624dedb1313f9fa0498e1d

Related to: https://github.com/ParabolInc/parabol/issues/2730

### AC

- [ ] Can export meeting summary to pdf

PR #8791, opened 2023-09-06
feat: export summary to pdf

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8777

Loom Demo: https://www.loom.com/share/Download-PDF-Updates-b9602acc0bd549bb9c9719073116d6f9

This only implements the download logic from the web app. It doesn't add the button to email summary. In the spirit of keeping PRs small, I'd like to get this reviewed before moving on to that logic.

You can download any meeting summary except from standups. This follows the CSV logic. Another PR can look at implementing this in standups too.

The buttons: 

Screenshot 2023-09-06 at 15 58 43

Screenshot 2023-09-06 at 15 58 50

### AC

- [ ] Go the the meeting summary of a retro, poker, or check-in meeting and see the `Download PDF` button
- [ ] Click it and see that it successfully downloads the PDF
- [ ] Check that the PDF looks OK


packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/SummarySheet.tsx
```diff
@@ -84,11 +87,57 @@ const SummarySheet = (props: Props) => {
...
+
+    while (remainingHeight >= 0) {
+      const position = remainingHeight - scaledHeight
+      if (!isFirstPage) {
+        pdf.addPage()
+      } else {
+        isFirstPage = false
+      }
+
+      pdf.addImage(imgData, 'PNG', 0, position, scaledWidth, scaledHeight)
```
Dschoordsch, line 125: -2 By creating the pdf this way, we end up with a series of screenshots just added to the pdf. The MacOS preview is misleading as it does some OCR, but checking with another PDF viewer you can see that we did not render text. Despite missing search functionality etc. it also renders unsharp text.

The better way would be to use [`pdf.html`](https://rawgit.com/MrRio/jsPDF/master/docs/module-html.html#~html) function which does not scale correctly for me out of the box, but renders text as such.

Dschoordsch (changes requested):

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/SummarySheet.tsx
nickoferrall: I've updated the PR (and the Loom in the description) to use `puppeteer`, which creates the PDF on the server. The text should now be searchable 👍

packages/server/graphql/public/mutations/createPDF.ts
```diff
@@ -0,0 +1,48 @@
...
+  const browser = await puppeteer.launch({headless: 'new'})
+  const page = await browser.newPage()
+
+  await page.setContent(htmlContent, {
+    waitUntil: 'networkidle0' // wait until there are no more network connections so the images are loaded
+  })
+
+  await page.evaluate(() => {
+    // the buttons looks bad in the pdf and aren't needed, so remove them
+    const buttonsTable = document.querySelector(`#hideButtonsInPDF`)
```
nickoferrall, line 16: The buttons looks bad in the PDF so we need to remove them or update their styles: 

Screenshot 2023-09-12 at 15 58 04

They're only useful in the web app, so I think we should remove them.

I don't like finding the elements by using their IDs, but I'm not sure of a better way.

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/SummarySheet.tsx
Dschoordsch: @nickoferrall without diving too deep into it right now, but did you check the system requirements for puppeteer? I have some concerns about chromium impacting our GQLExecutors when running on the same node. That might require some more work on our infrastructure first. What are your thoughts?

nickoferrall:
Hey @mattkrick, as Georg's away for a few days, would you mind being the Maintainer on this one?

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/SummarySheet.tsx
mattkrick: I'd like to avoid adding puppeteer to our runtime if at all possible, it's BIG with a lot of dependencies & something about trusting a blob of HTML from a client feels like it could open us up to a 0-day. (e.g. does puppeteer sanitize all SVGs? <script> tags? clickHandlers?). It's generally used on trusted elements & I just haven't seen enough about it on untrusted HTML structures.

Before we have to go down that road, I'm curious, and this is just because I don't have all the context here, why not just open the print page dialog? Every windows, mac, linux computer comes with a "Print to PDF" printer installed. It will be faster because there's no server involved, more customizable because they can set the name, and best of all-- keep our codebase simpler!

Screenshot 2023-09-18 at 11 15 33 AM

If we absolutely no question about it MUST generate a PDF on our server, we should generate the HTML on the server & then ideally use a smaller package to do the print to pdf (pdfjs?)

nickoferrall: Love this idea! I've created this PR https://github.com/ParabolInc/parabol/pull/8845 with the print page dialog and will close this one 👍

PR #8797, opened 2023-09-06
chore(metrics): Add Amplitude client-side page view tracking

### LINKED ISSUE #8794: Write app side page view events to Amplitude
## Issue - Enhancement
App side [page view events](https://github.com/ParabolInc/parabol/blob/0856d33c3eac61b252335193089efe45c8cabcdb/packages/client/components/AnalyticsPage.tsx#L190-L201) are sent directly from client to Segment today. We need to do the same thing for Amplitude before we shut off Segment.

tianrunhe:
# Description

Fixes #8794 

## Demo

✅ Logging in the Developer Tools console shows it successfully tracked the event:

Screenshot 2023-09-07 at 14 21 36

✅ `Loaded a Page` events show up in Amplitude Development environment:

Screenshot 2023-09-07 at 14 20 58

✅ Event stream for a user displays the event with proper properties. Notice the tracking library is from `amplitude-ts` (as compared to `Segment` that we have in production today)

Screenshot 2023-09-07 at 14 20 20

## Testing scenarios
1. Go to any page
2. Open up console in Developer Tools and observer loggings indicating the events have been successfully tracked
3. Go to [this real-time Amplitude graph](https://app.amplitude.com/analytics/parabol/chart/13mww75e) and you should see new data points popping out

## Final checklist


tghanken (approved):

scripts/webpack/dev.client.config.js
```diff
@@ -129,7 +129,8 @@ module.exports = {
         prblIn: process.env.INVITATION_SHORTLINK,
         AUTH_INTERNAL_ENABLED: process.env.AUTH_INTERNAL_DISABLED !== 'true',
         AUTH_GOOGLE_ENABLED: process.env.AUTH_GOOGLE_DISABLED !== 'true',
-        AUTH_SSO_ENABLED: process.env.AUTH_SSO_DISABLED !== 'true'
+        AUTH_SSO_ENABLED: process.env.AUTH_SSO_DISABLED !== 'true',
+        AMPLITUDE_API_KEY: process.env.AMPLITUDE_WRITE_KEY
```
Dschoordsch, line 133: -1 We don't have this in `env.example` nor in our environments

packages/client/components/AnalyticsPage.tsx
```diff
@@ -200,8 +212,23 @@ const AnalyticsPage = () => {
         {integrations: {'Google Analytics': {clientId: await getAnonymousId()}}}
       )
       ReactGA.send({hitType: 'pageview', content_group: getContentGroup(pathname)})
+      amplitude.track(
```
Dschoordsch, line 215: -1 This effect has an early return on l191 so this will not be reached of GA or Segment aren't loaded, although this should be independent. Also if any of the track calls before would throw, we'd never reach this point. I suggest splitting this up so each of the track functions could work independently.

packages/client/components/AnalyticsPage.tsx
```diff
@@ -211,24 +211,35 @@ const AnalyticsPage = () => {
...
     }, TIME_TO_RENDER_TREE)
-  }, [isSegmentLoaded, pathname, location.search])
+  }, [isSegmentLoaded, pathname])
+
+  useEffect(() => {
+    ReactGA.send({hitType: 'pageview', content_group: getContentGroup(pathname)})
+  }, [pathname])
+
+  useEffect(() => {
+    const title = document.title || ''
```
Dschoordsch, line 222: -1 I think we need the same `setTimeout` hack like we have in l194 because of the comment in l103. Feel free to reorder these to make it easier to find for the next one.
```
"Loaded a Page",
    {
      "name": "Meetings",
      "referrer": "",
      "title": "Meetings | Parabol",
      "path": "/me/tasks",
      "url": "http://localhost:3000/me/tasks",
      "translated": false,
      "search": ""
    },
```

Dschoordsch (changes requested):

packages/client/components/AnalyticsPage.tsx
```diff
@@ -175,6 +182,10 @@ const AnalyticsPage = () => {
     cacheEmail().catch()
   }, [isSegmentLoaded])
 
+  // page titles are changed in child components via useDocumentTitle, which fires after this
+  // we must guarantee that this runs after useDocumentTitle
+  // we can't move this into useDocumentTitle since the pathname may change without chaging the title
+  const TIME_TO_RENDER_TREE = 100
```
Dschoordsch, line 188: +1 I don't know if you've tested this, but for me with Safari, I need a bigger timeout to make this work correctly, at least on the first 2 renders:
* load `localhost:3000` -> it redirects to /meetings and tracks /meetings with signup title
* go to a team dashboard -> it tracks the dash path with "Meetings" title

After that it works fine. I assume it's because we're lazy loading some components, so it takes more time in the beginning.

Dschoordsch (approved):
I'm fine with this change as it's as good as the old versions, but maybe we should just increase the `TIME_TO_RENDER_TREE` threshold. But this comes at the risk of the user having left the page already and then we have a new wrong title. In the end, I assume we just have to live with the fact that page titles are unreliable.

packages/client/components/AnalyticsPage.tsx
tghanken: In practice, we almost always use path instead of page title when analyzing things.  I'm going to merge this as is and we can revisit the threshold if it becomes an actual problem.

PR #8798, opened 2023-09-07
chore(prompt to join org): restrict to verified users

### LINKED ISSUE #8759: Restrict prompt to join org to verified users
Before rolling out to all users, we want to lock the prompt to join org feature to verified users for security purposes.

Users can only ask to join an org if:
- [x] The user asking to join the org is verified
- [ ] The creator of the org they are asking to join is verified (The first OrganizationUser on the org is the creator)
- [ ] The billing leader of the org they are asking to join is verified
- [x] The domain isn't freemail (We check this currently, but should implement #8760  before rollout)

Dschoordsch:
# Description

Fixes #8759 

Before allowing prompt to join org for a user, they must now
* have a non-freemail domain
* be email verified
and the orgs we consider must have
* an active founder with a verified email matching the domain
*OR*
* one of the billing leads has a verified email matching the domain

## Demo

https://www.loom.com/share/389fd50d24f04d7a9978891dcd995c4c?sid=7ff7889a-0bcb-407c-9475-e60b6b708a76

## Testing scenarios

Most edge cases should be handled by tests, however there are a few that should be tested to verify in real world:

1. eligible org, new user not verified
    * founder of org is verified and has correct email and org has active domain
    * user is not verified
    * no request is sent

2. eligible org, new user is verified
    * same as before, but request is sent

All following tests are with a verified user

3. founder of org has wrong domain -> no request
4. additional billing lead has correct verified address -> request
5. founder has correct address, but left -> no request

## Final checklist


packages/server/utils/isRequestToJoinDomainAllowed.ts
```diff
@@ -1,43 +1,73 @@
...
-            .and(orgUsers.filter((ou: RDatum) => ou('userId').eq(viewerId)).isEmpty())
-        )
+    }))
+    .merge((org: RDatum) => ({
+      founder: org('members').nth(0).default(null),
+      billingLeads: org('members').filter({role: 'BILLING_LEADER', inactive: false}),
+      activeMembers: org('members').filter({inactive: false, removedAt: null}).count()
+    }))
+    .filter((org: RDatum) =>
+      org('activeMembers').gt(0).and(org('members').filter({userId}).isEmpty())
```
igorlesnenko, line 37: Did we intentionally change this from 1 to 0 ?

packages/server/utils/isRequestToJoinDomainAllowed.ts
```diff
@@ -1,43 +1,73 @@
...
     )
-    .limit(limit ?? BIG_ENOUGH_LIMIT)('id')
     .run()
+
+  const eligibleOrgs = await Promise.all(
+    orgs.map(async (org) => {
+      const {founder} = org
+      const importentMembers = org.billingLeads.slice() as TeamMember[]
```
igorlesnenko, line 44: small typo? `important`

packages/server/utils/isRequestToJoinDomainAllowed.ts
```diff
@@ -1,43 +1,73 @@
...
+      if (!founder.inactive && !founder.removedAt && founder.role !== 'BILLING_LEADER') {
+        importentMembers.push(founder)
+      }
+
+      const users = (
+        await dataLoader.get('users').loadMany(importentMembers.map(({userId}) => userId))
```
igorlesnenko, line 50: feels like we can first iterate through orgs to get all user ids, and then make a single dataloader query

igorlesnenko (approved):

packages/server/utils/isRequestToJoinDomainAllowed.ts
Dschoordsch: Yes, but this would make the code less readable and it wouldn't improve anything. This is the unique benefit of a dataloader, although a bit tricky to wrap your head around. JavaScript is single threaded, so the code gets executed until it hits an unresolved promise/await call, then it's added to a queue. The dataloader [ensures](https://github.com/graphql/dataloader/blob/d336bd15282664e0be4b4a657cb796f09bafbc6b/src/index.js#L213-L238) that the result of `loadMany` will be unresolved and it will be added to the end of the queue. Once it's the dataloaders turn, it will execute all load calls at once and resolve all promises at once.
I think it's easiest to understand with the following small example:
```ts
import DataLoader from 'dataloader'

const testLoader = new DataLoader(async (keys: string[]) => {
  console.log('loading ', keys)
  return keys
})

test('Testing loading queue', async () => {
  await Promise.all([1,2,3].map(async (i) => {
    console.log('mapping ', i)
    await testLoader.load(i)
  })
})
```
will print
```
    mapping  1
    mapping  2
    mapping  3
    loading  [ 1, 2, 3 ]
```

packages/server/utils/isRequestToJoinDomainAllowed.ts
Dschoordsch: This was not intentional, I missed it. Was this just to exclude onboarding orgs? Because with this >= 2, there is a gap: how do I get my first user to join my org?

igorlesnenko: @Dschoordsch yes, initially we wanted to exclude orgs with a single member. But I think it is fine like this

Dschoordsch: I changed the logic back and adapted the tests

igorlesnenko (approved):

PR #8811, opened 2023-09-08
chore: Allow google to index /create-account

tghanken:
# Description

Right now, https://action.parabol.co/robots.txt has `Allow: create-account` instead of `Allow: /create-account`, which is blocking the create account page from being indexed by Google (and other search engines).  Fixing this should help drive more organic search signups.

## Testing scenarios

N/A, I believe we only run NGINX on dokku, not when developing locally.

## Final checklist

### DIFF
--- a/nginx.conf.sigil
+++ b/nginx.conf.sigil
@@ -99,7 +99,7 @@ server {
 
   location = /robots.txt {
     add_header Content-Type text/plain;
-    return 200 "User-agent: *\nDisallow: /\nAllow: /$dollar\nAllow: create-account\nAllow: /signin\nAllow: /retrospective-demo\n";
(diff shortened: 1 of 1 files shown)


tghanken:
Current dependencies on/for this PR:
* master
  * **PR #8811** <a href="https://app.graphite.dev/github/pr/ParabolInc/parabol/8811" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ParabolInc/parabol/8811?utm_source=stack-comment).

tghanken:
@mattkrick I believe this should fix our robots.txt file, but am not 100% sure we're still using NGINX as our ingress.  Let me know if I need to make this change elsewhere.

mattkrick:
omg, how long has that been like that?!!? 🙈 great find!

interesting, looks like google could index it, but not read it...  
Screenshot 2023-09-11 at 12 23 44 PM

one way to find out if it works! 🤞

Issue #8826, opened 2023-09-13
Ad-hoc: implement add team button in teams dropdown


igorlesnenko:
<img width="395" alt="image" src="https://github.com/ParabolInc/parabol/assets/466991/13cd3d24-2a57-4eb3-8065-e7be955a0d74">

**AC:** 
- "add team" button is added to the teams dropdown
- Just open existing `Create a New Team` page in a new tab :)

**Not included:** 
- Modal

Issue #8824, opened 2023-09-13
Ad-hoc: add a feature flag


igorlesnenko:
Add `adHocTeams` feature flag

PR #8831, opened 2023-09-14
feat(ad-hoc): allow to create team from teams dropdown

igorlesnenko:
Fixes: #8824
Fixes: #8826

image

How to test:
- add `adHocTeam` user-level feature flag
- go to any meeting in activity library, see that now you can add a team from teams dropdown
- Currently, it just opens a new tab with new team screen

### DIFF
--- a/packages/client/components/NewMeetingTeamPicker.tsx
+++ b/packages/client/components/NewMeetingTeamPicker.tsx
@@ -26,10 +26,12 @@ interface Props {
   onSelectTeam: (teamId: string) => void
   positionOverride?: MenuPosition
   customPortal?: React.ReactNode
+  allowAddTeam?: boolean
 }
 
 const NewMeetingTeamPicker = (props: Props) => {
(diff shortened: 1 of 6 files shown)


packages/client/components/NewMeetingTeamPicker.tsx
```diff
@@ -86,6 +88,13 @@ const NewMeetingTeamPicker = (props: Props) => {
             menuProps={menuProps}
             teams={teams}
             teamHandleClick={handleSelectTeam}
+            onAddTeamClick={
+              allowAddTeam
```
nickoferrall, line 92: +1 NIT: moving this to a `handleClick` function could make it easier to read

nickoferrall (approved):
Tested and it's working well!

Dschoordsch (approved):
This is ok because we're moving the dialog into a tab. However right now the team select would not be updated on creating a new team.

PR #8837, opened 2023-09-14
feat: add support for multi-replica deploys

### LINKED ISSUE #8834: preDeploy-- build index.html inside run command
preDeploy builds assets when its called, but those assets are not a part of the docker image.
if a CDN is used, those assets will get uploaded to the CDN.
However, for index.html & non-CDN uses, we'll need to generate the assets again before calling `start`.

mattkrick:
# Description

fix #8834

since index.html doesn't exist in the docker image, we need to generate it wherever `node dist/web.js` is called.

## Demo

N/A

## Testing scenarios

- [x] check our CI, it works!
- [ ] locally, run `yarn build --no-deps && node dist/preDeploy && node dist/web`
- [ ] locally, run `yarn build --no-deps && node dist/web`

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -150,6 +150,10 @@ jobs:
           path: test-results/
           retention-days: 7
 
+      - name: Cleanup Predeploy for Docker Image
+        if: env.DOCKERIZE == 'true'
+        run: rm build/index.html build/sw.js build/manifest.json
+
       - name: Set up Docker Buildx
         if: env.DOCKERIZE == 'true'
         uses: docker/setup-buildx-action@v2
--- a/scripts/toolboxSrc/applyEnvVarsToClientAssets.ts
+++ b/scripts/toolboxSrc/applyEnvVarsToClientAssets.ts
@@ -98,3 +98,5 @@ const rewriteIndexHTML = async () => {
 export const applyEnvVarsToClientAssets = async () => {
   return Promise.all([rewriteServiceWorker(), rewriteIndexHTML(), writeManifest()])
 }
+
+if (require.main === module) applyEnvVarsToClientAssets()
--- a/scripts/toolboxSrc/applyEnvVarsToClientAssets.ts
+++ b/scripts/toolboxSrc/applyEnvVarsToClientAssets.ts
@@ -14,13 +14,13 @@ const getCDNURL = () => {
   return CDN_BASE_URL ? `${CDN_BASE_URL}/build` : '/static'
 }
 
-const rewriteServiceWorker = async () => {
-  const skeleton = await fs.promises.readFile(path.join(clientDir, 'swSkeleton.js'), 'utf-8')
+const rewriteServiceWorker = () => {
+  const skeleton = fs.readFileSync(path.join(clientDir, 'swSkeleton.js'), 'utf-8')
   const deploySpecificServiceWorker = skeleton.replaceAll('__PUBLIC_PATH__', getCDNURL())
-  await fs.promises.writeFile(path.join(clientDir, 'sw.js'), deploySpecificServiceWorker)
+  fs.writeFileSync(path.join(clientDir, 'sw.js'), deploySpecificServiceWorker)
 }
 
-const writeManifest = async () => {
+const writeManifest = () => {
   // If src is relative, then it will be relative to the manifest location, so manifest.json must be at root /
   const cdn = getCDNURL()
   const manifest = {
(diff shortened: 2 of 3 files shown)


.github/workflows/build.yml
```diff
@@ -150,6 +150,10 @@ jobs:
           path: test-results/
           retention-days: 7
 
+      - name: Cleanup Predeploy for Docker Image
+        if: env.DOCKERIZE == 'true'
+        run: rm build/index.html build/sw.js build/manifest.json
```
mattkrick, line 155: these were generated for testing purposes, remove them before they get included in the base image

scripts/toolboxSrc/applyEnvVarsToClientAssets.ts
```diff
@@ -98,3 +98,5 @@ const rewriteIndexHTML = async () => {
 export const applyEnvVarsToClientAssets = async () => {
   return Promise.all([rewriteServiceWorker(), rewriteIndexHTML(), writeManifest()])
 }
+
+if (require.main === module) applyEnvVarsToClientAssets()
```
mattkrick, line 102: writing the 3 client assets takes about 20ms, so even though it's a bit wasteful to do it every time we call the script, it's simple & grants the opportunity to change env vars easily without storing a checksum somewhere

rafaelromcar-parabol:
It builds the docker image see [here](https://github.com/ParabolInc/parabol/actions/runs/6298034666/job/17096131400)

rafaelromcar-parabol:
The web server isn't able to run from the produced Docker image when running just `node dist/web.js`. It throws the following:
```
node:internal/fs/utils:350
    throw err;
    ^

Error: ENOENT: no such file or directory, open '/home/node/parabol/build/index.html'
    at Object.openSync (node:fs:603:3)
    at Object.readFileSync (node:fs:471:35)
    at 280250 (/home/node/parabol/dist/web.js:137191:18)
    at __webpack_require__ (/home/node/parabol/dist/web.js:1425025:42)
    at 400785 (/home/node/parabol/dist/web.js:156502:18)
    at __webpack_require__ (/home/node/parabol/dist/web.js:1425025:42)
    at /home/node/parabol/dist/web.js:1425221:37
    at Object.<anonymous> (/home/node/parabol/dist/web.js:1425223:12)
    at Module._compile (node:internal/modules/cjs/loader:1256:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1310:10) {
  errno: -2,
  syscall: 'open',
  code: 'ENOENT',
  path: '/home/node/parabol/build/index.html'
}

Node.js v18.17.0
```

Which means it isn't producing the index.html when running that command 🤔. That is the whole log.

rafaelromcar-parabol:
Ok so, it works if I run the containers with the Web Server with the following:
```
node dist/preDeploy.js && node dist/web.js
```

Which is expected. But I see some situations where this could provoke some trouble:

## Starting multiple web servers at the same time
To speed up releases and reduce the times an user will be disconnected from their server we are doing the following during a release:
- **PreDeploy:** a new container is created and it executes only `node dist/preDeploy.js`. Once it ends successfully, the release carries on. The next two steps happen at the exact same time.
- **GraphQL Executors:** the current GQL Executors are killed and new ones replace them, in that order. The replacement takes up to a minute. During that time, no GQL Executor is available.
- **Web Servers:** N new web servers (N= number of web servers currently running) are started and will then execute `node dist/preDeploy.js && node dist/web.js` at almost the same time. Once they are ready, which can take up to 2 minutes between the docker pull (big images) and the moment the app answers, the old web servers are killed.

Would that be fine? I have tested in an on-demand environment and it works just fine.

## One current web server gets moved to another node while the migrations of a new release are running
The process is exactly the same as above BUT at some point, before the end one of the old web servers gets killed by Kubernetes for an unknown reason. Then Kubernetes restarts it WITH THE OLD VERSION. Which means it will then run the preDeploy.js script old the OLD version.

Would that be a problem?

.github/workflows/build.yml
rafaelromcar-parabol: We have to add a line with `RUN rm -f build/index.html build/sw.js build/manifest.json` into the [pipeline.dockerfile](https://github.com/ParabolInc/parabol/blob/master/docker/parabol-ubi/docker-build/dockerfiles/pipeline.dockerfile) instead (and the rest of dockerfiles). That way, if someone builds it manually, it will build it correctly.

Should we then remove this step on the Build job?

rafaelromcar-parabol: This can be removed from the build process and do not need to be included anywhere. The script web.js replaces the existing file with the filled one. I think you should remove it for simplicity.

rafaelromcar-parabol:
It looks like working!!! This _could_ be merged now, just check [my last comment](https://github.com/ParabolInc/parabol/pull/8837#discussion_r1339705426).

If it could be included in the release this week, it would be ❤️

mattkrick:
`node dist/web.js` is all that's needed for starting up the web container.
can we run `preDeploy` in a separate container whenever a new version is about to come live? web.js will be a dependent

rafaelromcar-parabol:
> node dist/web.js is all that's needed for starting up the web container.

That is what we are doing already, so great!

> can we run preDeploy in a separate container whenever a new version is about to come live? web.js will be a dependent

Pre Deploy runs BEFORE the new version of the web server starts. It runs in a different POD (thus a different container ofc). Pre Deploy is actually a pre-instal/upgrade hook on our Helm chart.

When releasing the process is:
- PreDeploy is executed as a Kubernetes Job. If it ends in success, the release continues. If not, the release is aborted.
- At the same time, we start M web servers and N gql-executors (M=current number of web servers and N=current number of gql-executors).
- Once the new components are ready (ready as in Kubernetes), the old ones are killed (usually all at the same time, but this isn't 100% guaranteed).

So, to be 100% clear: we will never start a new version of the web server without a successful execution of the PreDeploy script. The PreDeploy script is a hook that happens before the deployment even starts, and it blocks the deployment if it fails.

PR #8839, opened 2023-09-15
fix: remove memory leak from analytics

mattkrick:
# Description

Fixes a memory leak.
Note, this should not be a long term fix! This is just to keep our app from crashing.
Long term fix is here: https://github.com/ParabolInc/parabol/issues/8840

## Demo

https://www.loom.com/share/7491a7ea6d0247649d700c555ab0c959


nickoferrall (approved):
Enjoyed the Loom! LGTM

PR #8845, opened 2023-09-18
feat: download summary as pdf

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8777

Screenshot 2023-09-06 at 15 58 43

### To test

- [ ] Click "download to pdf" and see that the print page dialog opens
- [ ] Download the summary to a pdf and see that it looks good

### DIFF
--- a/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/SummarySheet.tsx
+++ b/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/SummarySheet.tsx
@@ -93,7 +93,14 @@ const SummarySheet = (props: Props) => {
   }
 
   return (
-    <table width='100%' height='100%' align='center' bgcolor='#FFFFFF' style={sheetStyle}>
+    <table
+      className='print:w-[210mm]'
+      width='100%'
+      height='100%'
+      align='center'
+      bgcolor='#FFFFFF'
+      style={sheetStyle}
+    >
       <tbody>
         <tr>
           <td>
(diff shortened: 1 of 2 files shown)


packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/SummarySheet.tsx
```diff
@@ -107,28 +112,49 @@ const SummarySheet = (props: Props) => {
                       <tbody>
                         <tr>
                           <td align='center' width='100%'>
-                            <div className='flex justify-center gap-4'>
+                            <div className='mb-2 flex justify-center gap-4'>
```
nickoferrall, line 115: I added this margin to give the buttons more room to breathe: 

Screenshot 2023-09-18 at 19 36 15

mattkrick: +1 what do you think about hiding these when they're printed? just add the class `.hide-print` to the parent
you could also make it take up the full page width if you wanted to get fancy (e.g. for an A4 letter size, add the class `print:w-[210mm]`. tailwinds is awesome, basically add the `print:` prefix to anything & it'll only show up in print mode!

nickoferrall: Oh that's cool! Didn't know you could do that with tailwind. Updated!

PR #8846, opened 2023-09-19
feat(ad-hoc): Add create team dialog

### LINKED ISSUE #8827: Ad-hoc: add team dialog
image

image

image

https://www.figma.com/file/Vc09AD3hxef2ypriihfK0T/Flow---ad-hoc-teams?type=design&node-id=0-1&mode=design&t=1EEQQFtT7uLRwdQB-0

AC:
- Clicking on "add team" should open "add team" dialog
- Add teammates field should be same field we use for one-on-one
- Display organization dropdown if organization can't be determined (use logic from one-one-one meetings)
- Automatically suggest team name based on selected members
- No changes on backend are needed, submit emails as usual when creating a team
- When team is added, it should be automatically selected in the team dropdown

igorlesnenko:
Fixes #8827
image

**Not included:** 
- Email invites expires warning

**How to test:**

- add `adHocTeam` user-level feature flag
- go to any meeting in activity library, open teams dropdown
- create a team from teams dropdown by clicking "add team"
- see that after creating a team it is automatically selected, invites are sent and you can successfully 
- test with different set of users and custom emails
- See if you selected users from different orgs an org selection dropdown appears showing all potential orgs correctly
- See if you entered only custom emails, org selection dropdown appears and you can select any of your org


igorlesnenko:
@tghanken can you please advise what events do I need to add here?

packages/client/components/ActivityLibrary/ActivityDetailsSidebar.tsx
```diff
@@ -132,12 +132,24 @@ const ActivityDetailsSidebar = (props: Props) => {
       : // it is a team-scoped template, templateTeam  must exist
         [templateTeam!]
 
+  const availableTeamsRef = useRef(availableTeams)
+
+  useEffect(() => {
+    availableTeamsRef.current = availableTeams
```
nickoferrall, line 138: +1 why use a ref here instead of state?

nickoferrall (changes requested):
I noticed this little bug while testing: https://www.loom.com/share/90dd014360bc44ce8d28312134d5a808

igorlesnenko:
@nickoferrall that is interesting bug, I iterate `viewerOrganizations` to show items in dropdown. I think that for some reason you have 2 orgs with same id returned by 

```
  fragment AddTeamDialog_viewer on User {
    ...AdhocTeamMultiSelect_viewer
    organizations {
      id
      name
    }
  }
  ```
  
  Not sure that is expected behaviour of the app. Is it possible that some of the DB data is wrong and you have yourself added to the org twice?

File: packages/client/components/AddTeamDialog.tsx [hunk 2]
```diff
@@ -0,0 +1,201 @@
...
+
+  const onSelectedUsersChange = (newUsers: Option[]) => {
+    setSelectedUsers(newUsers)
+    const selectedUsersOrganizationIds = new Set()
+    newUsers.forEach((user) => {
+      //add user.organizationIds to selectedUserOrganizationIds
+      user.organizationIds.forEach((organizationId) => {
+        selectedUsersOrganizationIds.add(organizationId)
+      })
+    })
```
tghanken, line 94: -1 We should add a "Teammate Selected" event here with the `selectionLocation` set to `addTeamUserPicker`

tghanken (changes requested):
We should try to keep the analytics here in line with what we did for 1:1 teams, there are two places that I see to add events in.

packages/client/components/NewMeetingTeamPicker.tsx
```diff
@@ -47,7 +50,7 @@ const NewMeetingTeamPicker = (props: Props) => {
   }
 
   const handleAddTeamClick = () => {
-    window.open(`/newteam/1`, '_blank', 'noreferrer')
+    setAddTeamDialogOpen(true)
```
tghanken, line 53: -1 Since we're no longer pushing a page, we're going to lose the page view event for the new team page.  We should add a "Add Team Clicked" event here so we don't lose coverage.

tghanken (commented):
Here's my other comment, I missed hitting add to review the first time 😮‍💨

packages/client/components/AddTeamDialog.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const atmosphere = useAtmosphere()
+
+  const {submitting, onCompleted, onError, error, submitMutation} = useMutationProps()
+  const {history} = useRouter()
+
+  const data = usePreloadedQuery<AddTeamDialogQuery>(query, queryRef)
+  const viewer = useFragment<AddTeamDialog_viewer$key>(AddTeamDialogViewerFragment, data.viewer)
+  const {organizations: viewerOrganizations} = viewer
+
+  const labelStyles = `text-left text-sm font-semibold mb-3`
```
nickoferrall, line 62: +2 I like that you're using a var instead of repeating the code

+1 I don't think it needs to be in the component though. Could define it above `AddTeamDialog` to reduce the code inside it

packages/client/components/AddTeamDialog.tsx
```diff
@@ -0,0 +1,201 @@
...
+  const fieldsetStyles = `mx-0 mb-6 flex flex-col w-full p-0`
+
+  const [selectedUsers, setSelectedUsers] = React.useState<Option[]>([])
```
nickoferrall, line 65: +1 `useState` has already been imported so no need for `React.` here or below

packages/client/components/AddTeamDialog.tsx
```diff
@@ -0,0 +1,201 @@
...
+          />
+
+          {selectedUsers.some((user: Option) => !user.id) && (
+            <div className='mt-3 text-xs font-semibold text-slate-700'>
+              Email invitations expire in 30 days.
+            </div>
+          )}
+        </fieldset>
+
+        {showOrgPicker && (
```
nickoferrall, line 155: +1 I think the org picker should always be visible so the height of the modal doesn't change when you select an item. It would also be reassuring for me that I've selected the right team from the right org

packages/client/ui/Input/Input.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react'
+
+import {twMerge} from 'tailwind-merge'
+
+export interface InputProps extends React.InputHTMLAttributes<HTMLInputElement> {}
+
+export const Input = React.forwardRef<HTMLInputElement, InputProps>(
```
nickoferrall, line 7: +2 great to see another component in the mix!

packages/client/ui/Input/Input.tsx
```diff
@@ -0,0 +1,21 @@
...
+  ({className, type, ...props}, ref) => {
+    return (
+      <input
+        type={type}
+        className={twMerge(
+          'flex h-11 w-full rounded border-2 border-slate-300 bg-transparent px-2 py-1 text-sm font-semibold focus:outline-none focus-visible:border-sky-500 disabled:cursor-not-allowed disabled:opacity-50 data-[placeholder]:text-slate-600',
```
nickoferrall, line 13: -1 on focus, this new input is blue, but the rest of our inputs go from grey to black. Also, the font is bolder in this input vs others.

Screenshot 2023-09-28 at 17 23 58



Screenshot 2023-09-28 at 17 22 10

packages/client/components/AddTeamDialog.tsx
```diff
@@ -0,0 +1,201 @@
...
+            }}
+            value={teamName}
+          />
+        </fieldset>
+        {error && <div className='mt-2 text-sm font-semibold text-tomato-500'>{error.message}</div>}
+        <DialogActions>
+          <SecondaryButton onClick={onClose} size='small'>
+            Cancel
+          </SecondaryButton>
+          <PrimaryButton size='small' onClick={handleAddTeam} disabled={submitting || !isValid}>
```
nickoferrall, line 192: +1 I'd prefer the buttons to be a bit larger as they feel slightly small to me: 

Screenshot 2023-09-28 at 17 01 24

packages/client/components/AddTeamDialog.tsx
```diff
@@ -0,0 +1,201 @@
...
+                    ))}
+                </SelectGroup>
+              </SelectContent>
+            </Select>
+          </fieldset>
+        )}
+
+        <fieldset className={fieldsetStyles}>
+          <label className={labelStyles}>Team name</label>
+          <Input
```
nickoferrall, line 179: -1 we're enforcing a team name length limit of 50 characters, but not giving users feedback that that will happen. So if I create a team with 55 characters, the last 5 characters get removed without me knowing.

We could add `maxLength={MAX_TEAM_NAME_LENGTH}` to the input. Or, for the optimal UX, we could give the user some feedback that there's a limit.

packages/client/components/AddTeamDialog.tsx
```diff
@@ -0,0 +1,201 @@
...
+    selectedUsers.length &&
+    (mutualOrgsIds.length > 1 || !mutualOrgsIds.length)
+  )
+
+  const defaultOrgId = mutualOrgsIds[0]
+  const [selectedOrgId, setSelectedOrgId] = useState(defaultOrgId)
+  const [teamName, setTeamName] = useState('')
+  const [teamNameManuallyEdited, setTeamNameManuallyEdited] = useState(false)
+
+  const MAX_TEAM_NAME_LENGTH = 50
```
nickoferrall, line 78: +1 it looks like the limit in pg is 100 characters so I was wondering why 50 here?

nickoferrall (changes requested):
Two small changes and then it LGTM!

packages/client/components/AddTeamDialog.tsx
```diff
@@ -84,6 +85,12 @@ const AddTeamDialog = (props: Props) => {
   }
 
   const onSelectedUsersChange = (newUsers: Option[]) => {
+    if (newUsers.length && newUsers.length > selectedUsers.length) {
+      SendClientSegmentEventMutation(atmosphere, 'Teammate Selected', {
+        selectionLocation: 'addTeamUserPicker'
+      })
+    }
```
tghanken, line 92: +1 I don't really think it matters, but we typically send events at the end of the relevant function after the work is done.

tghanken (approved):
Analytics look good!

packages/client/components/AddTeamDialog.tsx
igorlesnenko: @nickoferrall we have 50 characters validation set on the backend and when we creating a team from http://localhost:3000/newteam/1. So I kept it as is here too

igorlesnenko:
> I noticed this little bug while testing: https://www.loom.com/share/90dd014360bc44ce8d28312134d5a808
> Two small changes and then it LGTM!

@nickoferrall so was it some wrong data in the DB?

packages/client/components/AddTeamDialog.tsx
igorlesnenko: I just removed this now, and allow backend to validate this. I think rarely someone would type 50 chars

packages/client/ui/Input/Input.tsx
igorlesnenko: I made a new input to be matched with new select and multi select components we introduced previously. Not sure if I need to change it to match our old components, tagging @acressall for this question

packages/client/components/AddTeamDialog.tsx
igorlesnenko: If org picker will be always visible it will show one single org in most of cases. Would this be ok? Tagging @acressall for this design question

packages/client/components/AddTeamDialog.tsx
igorlesnenko: made bigger

image

packages/client/components/AddTeamDialog.tsx
acressall: I actually don't mind the small height change. I suspect that a lot of users aren't aware of the concept of orgs, and it could be especially confusing in this context, so I only want to bring it up when we actually need the information

packages/client/ui/Input/Input.tsx
acressall: The design team has been working add focus states, and they are all sky 500, so this is good!

acressall: I think the weight is incorrect though, in the [designs](https://www.figma.com/file/Vc09AD3hxef2ypriihfK0T/Flow---ad-hoc-teams?type=design&node-id=61-3235&mode=design&t=lJr67VlUQFnsOFGL-4) the label is semibold, but the placeholder and input are regular

igorlesnenko: Updated
image

igorlesnenko: Updated:

image

nickoferrall (approved):
> @nickoferrall so was it some wrong data in the DB?

Yep, that was it. LGTM!

packages/client/components/AddTeamDialog.tsx
```diff
@@ -0,0 +1,210 @@
...
+              </SelectContent>
+            </Select>
+          </fieldset>
+        )}
+
+        <fieldset className={fieldsetStyles}>
+          <label className={labelStyles}>Team name</label>
+          <Input
+            onChange={(e) => {
+              !teamNameManuallyEdited && setTeamNameManuallyEdited(true)
```
Dschoordsch, line 187: +1 We should not overuse this shortcut as it makes the code less readable imo.

packages/client/components/AddTeamDialog.tsx
```diff
@@ -0,0 +1,210 @@
...
+  const viewer = useFragment<AddTeamDialog_viewer$key>(AddTeamDialogViewerFragment, data.viewer)
+  const {organizations: viewerOrganizations} = viewer
+
+  const [selectedUsers, setSelectedUsers] = useState<Option[]>([])
+  const [mutualOrgsIds, setMutualOrgsIds] = useState<string[]>([])
+
+  const showOrgPicker = !!(
+    selectedUsers.length &&
+    (mutualOrgsIds.length > 1 || !mutualOrgsIds.length)
+  )
```
Dschoordsch, line 71: -1 This shows the org picker if I invite by email, even if I only have one org to start with
Screenshot 2023-10-05 at 12 11 59

packages/client/components/AddTeamDialogRoot.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const AddTeamDialogRoot = (props: Props) => {
+  const {onClose, onAddTeam} = props
+
+  const queryRef = useQueryLoaderNow<AddTeamDialogQuery>(addTeamDialogQuery, {}, 'network-only')
+
+  return (
+    <Suspense fallback={renderLoader()}>
+      {queryRef && (
+        <AddTeamDialog onAddTeam={onAddTeam} isOpen={true} onClose={onClose} queryRef={queryRef} />
```
Dschoordsch, line 20: +1 normally one would mount the dialog and then only modify `isOpen` so we can get a transition.

Dschoordsch (approved):

packages/client/components/AddTeamDialogRoot.tsx
igorlesnenko: Transition works for me like this. Can you record a video?

I'll merge this PR in the mean time, and will fix it in next PRs if it is an issue

Issue #8853, opened 2023-09-20
Track download to pdf


nickoferrall:
### AC

- [ ] Track when a user clicks download to pdf

PR #8854, opened 2023-09-20
chore: track download to pdf

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8853

### To test

- [ ] Click download to pdf and see the event is tracked


tghanken (approved):
Looks good!

PR #8861, opened 2023-09-20
feat(standups): Add slack configuration link to standup options

jmtaber129:
# Description
See [slack discussion](https://parabol.slack.com/archives/C042HTVBH8D/p1695159030945889) 🔒 

Add a standup meeting option menu item for "slack configuration" that links to the team's integrations page.

Also add metrics for when this option is clicked (note: this won't fire if the user specifically opens the link in a new tab instead of clicking)

## Demo
Screen Shot 2023-09-21 at 1 45 40 PM

## Testing scenarios
- [ ] UI makes sense
- [ ] Clicking sends the user to integration options in a new tab and sends a metric
- [ ] User can open the link by manually opening in new tab

## Final checklist

### DIFF
--- a/packages/client/components/TeamPrompt/TeamPromptOptionsMenu.tsx
+++ b/packages/client/components/TeamPrompt/TeamPromptOptionsMenu.tsx
@@ -1,7 +1,8 @@
 import styled from '@emotion/styled'
-import {Flag, Link, Replay} from '@mui/icons-material'
+import {Flag, Link as MuiLink, OpenInNew, Replay} from '@mui/icons-material'
 import graphql from 'babel-plugin-relay/macro'
 import React from 'react'
+import {Link} from 'react-router-dom'
 import {useFragment} from 'react-relay'
 import useAtmosphere from '~/hooks/useAtmosphere'
 import {MenuProps} from '~/hooks/useMenu'
@@ -15,8 +16,9 @@ import makeAppURL from '../../utils/makeAppURL'
 import Menu from '../Menu'
 import MenuItem from '../MenuItem'
 import {MenuItemLabelStyle} from '../MenuItemLabel'
+import SlackSVG from '../SlackSVG'
 
-const LinkIcon = styled(Link)({
+const LinkIcon = styled(MuiLink)({
   color: PALETTE.SLATE_600,
   marginRight: 8
 })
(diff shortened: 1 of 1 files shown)


tghanken (approved):
Analytics looks good!

packages/client/components/TeamPrompt/TeamPromptOptionsMenu.tsx
```diff
@@ -130,6 +132,28 @@ const TeamPromptOptionsMenu = (props: Props) => {
...
+        label={
+          <Link
+            to={`/team/${team.id}/settings/integrations`}
+            target='_blank'
+            rel='noopener noreferrer'
+          >
+            <OptionMenuItem>
+              <SlackSVG />
+              <span className='ml-2'>Configure slack</span>
+              <OpenInNew className='ml-1 text-base text-slate-600' />
```
BartoszJarocki, line 146: +1 imo it'd be better to push it to the right, looks a little bit cleaner 
Screenshot 2023-09-21 at 12 30 35

packages/client/components/TeamPrompt/TeamPromptOptionsMenu.tsx
```diff
@@ -130,6 +132,28 @@ const TeamPromptOptionsMenu = (props: Props) => {
           openRecurrenceSettingsModal()
         }}
       />
+      <MenuItem
+        key='slack'
+        label={
+          <Link
+            to={`/team/${team.id}/settings/integrations`}
+            target='_blank'
+            rel='noopener noreferrer'
```
BartoszJarocki, line 141: +1 do we actually want to use a new tab? it's within our app, imo it'd be better to just use a normal link

packages/client/components/TeamPrompt/TeamPromptOptionsMenu.tsx
```diff
@@ -130,6 +132,28 @@ const TeamPromptOptionsMenu = (props: Props) => {
...
+          >
+            <OptionMenuItem>
+              <SlackSVG />
+              <span className='ml-2'>Configure slack</span>
```
BartoszJarocki, line 145: +1 `Slack` instead of `slack`?

BartoszJarocki (approved):
looks good! left some comments to consider

packages/client/components/TeamPrompt/TeamPromptOptionsMenu.tsx
jmtaber129: The thinking here is that even though it's within the app, it's still outside of the meeting, so I'm wary of taking the user completely out of the meeting potentially unexpectedly.

enriquesanchez: 🤔 If the user has added content but not clicked Submit, will they lose their work if they are taken to the Slack config page?

enriquesanchez (approved):
LGTM! 👍

Issue #8865, opened 2023-09-20
Make standups a single column (internal feature flag)


jmtaber129:
We'd like to experiment with showing our standup meeting as a single column of responses (like a news feed) instead of a multi-column grid. See [slack](https://parabol.slack.com/archives/C042HTVBH8D/p1695158459844809) 🔒 

AC:
- [ ] Standup meetings are shown as a single column depending on feature flag

PR #8875, opened 2023-09-21
feat(standups): Display responses in single column (feature flagged)

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8865

Standup responses are pretty narrow for our own internal Friday Ship updates, leading to updates that can take up the entire height of the screen (or more), which can be annoying.

Add a feature flag for showing these responses in a single column. We'll use this internally, and roll out to everyone if we decide that's what we want to do. Alternatively, we could potentially add a "grid vs. list" option for each user, persisted on the client-side.

## Demo
Screen Shot 2023-09-21 at 2 05 51 PM
Screen Shot 2023-09-21 at 2 06 15 PM
Screen Shot 2023-09-21 at 2 05 58 PM

## Testing scenarios

- [ ] Smoke test the grid view without the feature flag
- [ ] Add the `singleColumnStandups` org feature flag, and test the response list.

## Final checklist

### DIFF
--- a/packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
+++ b/packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
@@ -30,23 +30,26 @@ const fiveColumnResponseMediaQuery = `@media screen and (min-width: ${ResponsesG
 
 const ResponseWrapper = styled('div')<{
   status: TransitionStatus
-}>(({status}) => ({
+  isSingleColumn: boolean
+}>(({status, isSingleColumn}) => ({
   opacity: status === TransitionStatus.MOUNTED || status === TransitionStatus.EXITING ? 0 : 1,
   transition: `box-shadow 100ms ${BezierCurve.DECELERATE}, opacity 300ms ${BezierCurve.DECELERATE}`,
   display: 'flex',
   flexDirection: 'column',
   width: '100%',
+  maxWidth: isSingleColumn ? '850px' : undefined,
+  margin: isSingleColumn ? '0 auto' : undefined,
   [twoColumnResponseMediaQuery]: {
-    width: `calc(100% / 2 - ${ResponseCardDimensions.GAP}px)`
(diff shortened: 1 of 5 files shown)


packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
```diff
@@ -30,23 +30,26 @@ const fiveColumnResponseMediaQuery = `@media screen and (min-width: ${ResponsesG
...
   status: TransitionStatus
-}>(({status}) => ({
+  isSingleColumn: boolean
+}>(({status, isSingleColumn}) => ({
   opacity: status === TransitionStatus.MOUNTED || status === TransitionStatus.EXITING ? 0 : 1,
   transition: `box-shadow 100ms ${BezierCurve.DECELERATE}, opacity 300ms ${BezierCurve.DECELERATE}`,
   display: 'flex',
   flexDirection: 'column',
   width: '100%',
+  maxWidth: isSingleColumn ? '850px' : undefined,
```
jmtaber129, line 40: 850px is somewhat arbitrary - it's a little more than the previous single-column max width

enriquesanchez: I had landed on [600px on the mockups ](https://www.figma.com/file/mR6vYT8fUDRv9tBGb8OPCL/Standups-iterations?type=design&node-id=1572-25292&mode=design&t=DGUyiQtSLxHm4eLd-4) because I felt that wider than that the text readability suffers, since long text lines are harder on the eye.

Do you think that's too narrow?

BartoszJarocki (approved):
Looks good, I'd just play with the width to improve readability

packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
jmtaber129: Fair enough! Updated this to 600px instead.

Issue #8843, opened 2023-09-18
Auto-Join: add autoJoin mutation


nickoferrall:
### AC

- [ ] Add the `autoJoin` boolean field to the team object
- [ ] Implement a public mutation that takes `teamIds` and `addFlag` as the args. It changes `autoJoin` boolean field to true/false in the same way our user & org feature flags work

PR #8878, opened 2023-09-22
feat: implement autoJoin mutation

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8843

### To test

- [ ] Run the `updateAutoJoin` mutation and see that the `autoJoin` field on the `team` object updates


packages/server/graphql/types/Team.ts
```diff
@@ -40,6 +40,11 @@ const Team: GraphQLObjectType = new GraphQLObjectType<ITeam, GQLContext>({
       type: new GraphQLNonNull(GraphQLID),
       description: 'A shortid for the team'
     },
+    autoJoin: {
```
nickoferrall, line 43: We want to be moving these to the public types, but when I do so, I get: `Team.autoJoin defined in resolvers, but not in schema`, but we don't need to define it in a resolver yet. The codegen file imports the `Team` type from [here](https://github.com/ParabolInc/parabol/blob/40fe222a6108205768bae3918317a36b49319c01/packages/server/postgres/queries/getTeamsByIds.ts#L6) where `autoJoin` is defined 🤔

igorlesnenko: If I remove `autoJoin` from here and then add it here: 

https://github.com/ParabolInc/parabol/blob/03c20567f59f3d942c640c3c4c893502d1b4932e/packages/server/graphql/public/typeDefs/Team.graphql#L4

It works for me

Can you try?

igorlesnenko (changes requested):

packages/server/graphql/types/Team.ts
nickoferrall: I'm not sure what was happening before, but that is working now. Thank you, Sir!

igorlesnenko (approved):
Not tested again

packages/server/graphql/public/mutations/updateAutoJoin.ts
```diff
@@ -0,0 +1,26 @@
...
+  _source,
+  {teamIds, autoJoin},
+  {authToken}
+) => {
+  const viewerId = getUserId(authToken)
+  if (!isSuperUser(authToken)) {
+    const isNotMember = teamIds.some((teamId) => !isTeamMember(authToken, teamId))
+    if (isNotMember) {
+      return standardError(new Error('Not on team(s)'), {userId: viewerId})
+    }
```
Dschoordsch, line 16: -1 Either we say we make this a manual step and skip verification for now (then it should be a private mutation) or we should ensure they're billing leader in absence of another admin like role.

Dschoordsch (changes requested):
I'm not sure there wouldn't be some architectural changes coming out of the follow up, so I wouldn't want to merge this before that's approved. Apart from the small detail it looks fine though.

Dschoordsch (approved):

nickoferrall:
> I'm not sure there wouldn't be some architectural changes coming out of the follow up, so I wouldn't want to merge this before that's approved. Apart from the small detail it looks fine though.

It looks like we're close on the following PR so I'll go ahead and merge this one

Issue #8844, opened 2023-09-18
Auto-Join: add verified users to team


nickoferrall:
Predecessor: https://github.com/ParabolInc/parabol/issues/8843

### AC

- [ ] If a user signs up and has `isEmailVerified` set to true (e.g. if they sign-up using Google Sign-In), add them to teams with the same active domain that have `autoJoin` set to true
- [ ] Track when a user joins a team using the autoJoin feature
- [ ] If the user auto-joins a team, don't create a default org & team as they will already belong to an org & team. It could be confusing.

PR #8883, opened 2023-09-25
feat: autoJoin verified users to team

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8844

Demo: https://www.loom.com/share/45ce1dc4e5e84f3cb4c6e0cd60ab5e78

In this PR, users that sign up with Google Sign-In will automatically join teams that:
- Have `autoJoin` set to true
- Belong to an org with the same `activeDomain` as the user
- At least one user in the team has the same domain

Assuming you don't have access to two Gmail accounts with the same domain, I suggest testing this like so:

### To test

- [ ] Create an account with an example user, i.e. `example@example.com`. See that it creates a dummy org & team, like normal, e.g. `Example Team`. 
- [ ] With a super user, execute the `updateAutoJoin` mutation on `Example Team`, the `updateEmail` mutation to change `example@example.com` to `example@parabol.co`, and `setOrganizationDomain` to set the example org domain to `parabol.co`.
- [ ] Now create an account using Google Sign-In, e.g. `nick@parabol.co` and see that you automatically join the team(s) that fulfil the criteria above
- [ ] See that we track when a user auto joins a team
- [ ] See that if you autoJoin a team, we no longer create a default org & team


packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
```diff
@@ -95,10 +151,6 @@ const bootstrapNewUser = async (
   }
   analytics.accountCreated(userId, !isOrganic, isPatient0)
 
-  if (isOrganic) {
-    sendPromptToJoinOrg(newUser, dataLoader)
```
nickoferrall, line 99: Looks like this can be moved up inside the `isOrganic` check above?

packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
```diff
@@ -18,6 +18,28 @@ import sendPromptToJoinOrg from '../../../utils/sendPromptToJoinOrg'
...
+import isValid from '../../isValid'
+import {Team} from '../../../postgres/queries/getTeamsByIds'
+
+const getAutoJoinTeams = async (
+  usersWithDomainTeams: Team[] | never[],
+  domain: string,
+  dataLoader: DataLoaderWorker
+) => {
+  const orgIds = usersWithDomainTeams
+    .filter(({autoJoin, isArchived}) => autoJoin && !isArchived)
```
igorlesnenko, line 32: +1 I think we can filter this in database query using kysely. Otherwise, we always fetch all teams of all domain users every new sign up which feels unnecessarily heavy

igorlesnenko (approved):
Works!

2 things I think would be great to add:
- Members of the team don't receive any subscription update, I need to refresh a page to see a new member
- I think there should be a notification for both sides, feels a bit scary for me when someone access your data without notice or you access someone data without notice

nickoferrall:
> I think there should be a notification for both sides

Sounds good - I've created an issue for this: https://github.com/ParabolInc/parabol/issues/8909

packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
```diff
@@ -70,23 +103,46 @@ const bootstrapNewUser = async (
...
-    const teamId = generateUID()
-    tms.push(teamId) // MUTATIVE
-    const validNewTeam = {
-      id: teamId,
-      orgId,
-      name: makeDefaultTeamName(teamId),
-      isOnboardTeam: true
+    const isVerified = identities.some((identity) => identity.isEmailVerified)
+    const teamsWithAutoJoin =
+      isVerified && (await getAutoJoinTeams(usersWithDomainTeams, domain, dataLoader))
```
Dschoordsch, line 108: -1 I don't see why we restrict auto join to users without an invite. My mental picture of the auto join feature is like the general channel in Slack, i.e. I would expect every user to initially being added there, even if invited to another team.

packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
Dschoordsch: -1 I agree, we should do the `autoJoin` check on db side.

packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
```diff
@@ -48,11 +79,13 @@ const bootstrapNewUser = async (
     experimentalFlags.push('retrosInDisguise')
   }
 
-  await Promise.all([
+  const usersWithDomainTeamIds = usersWithDomain.flatMap((user) => user.tms)
+  const [usersWithDomainTeams] = await Promise.all([
```
Dschoordsch, line 83: -2 Starting from users with the same domain is the wrong approach. If I have an auto join team at bigcorp.com, I don't want any random users join just because I have joe@consultants.org on there as well.
We should start from the organization with active domain instead.

packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
```diff
@@ -18,6 +18,28 @@ import sendPromptToJoinOrg from '../../../utils/sendPromptToJoinOrg'
 import {makeDefaultTeamName} from 'parabol-client/utils/makeDefaultTeamName'
 import isCompanyDomain from '../../../utils/isCompanyDomain'
 import {DataLoaderWorker} from '../../graphql'
+import getTeamsByIds from '../../../postgres/queries/getTeamsByIds'
+import acceptTeamInvitation from '../../../safeMutations/acceptTeamInvitation'
+import isValid from '../../isValid'
+import {Team} from '../../../postgres/queries/getTeamsByIds'
+
+const getAutoJoinTeams = async (
+  usersWithDomainTeams: Team[] | never[],
```
Dschoordsch, line 27: -1 There is no reason to have this `| never[]` type here, it could just be a more conventional `| null`.

Dschoordsch (changes requested):
-2 We're missing some verification mechanism for the organizations which have an auto join team. Even if we verify this when enabling the flag, there is still a chance an org could be abandoned and taken over by non-verified users. I think it would be good to verify an org has verified billing leads/founder similar to https://github.com/ParabolInc/parabol/blob/e508bcfeaf79c5ea1d679d7ef75c6fdbfb35302c/packages/server/utils/isRequestToJoinDomainAllowed.ts#L10

packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
nickoferrall: Makes sense. I was checking whether the org the team belongs to has the same `activeDomain` as the user, so `consultants.org` wouldn't have joined `bigcorp.com`, but I agree that starting from the org makes sense. Initially, I thought that would be expensive, but I see it's an index, so it would be efficient.

packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
```diff
@@ -24,10 +26,22 @@ const bootstrapNewUser = async (
...
+    tier,
+    segmentId,
+    identities
+  } = newUser
   // email is checked by the caller
   const domain = email.split('@')[1]!
-  const isCompanyDomain = await dataLoader.get('isCompanyDomain').load(domain)
+  const [isCompanyDomain, organizations] = await Promise.all([
+    dataLoader.get('isCompanyDomain').load(domain),
+    dataLoader.get('organizationsByActiveDomain').load(domain)
```
Dschoordsch, line 43: -1 commenting here, but could be fixed further down: We're trying to update the `activeDomain` of organizations only with company domains. But this is a moving target and was just recently updated. This means there are companies which have an `activeDomain` which is not a company domain. These will be returned here and we need to check `isCompanyDomain` somewhere before doing all the autojoin logic.

packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
```diff
@@ -65,8 +83,36 @@ const bootstrapNewUser = async (
...
+  if (teamsWithAutoJoin.length > 0) {
+    tms.push(...teamsWithAutoJoin.map(({id}) => id))
+    await Promise.all(
+      teamsWithAutoJoin.map((team) => {
+        const teamId = team.id
+        return Promise.all([
+          acceptTeamInvitation(team, userId, dataLoader),
+          isOrganic
+            ? Promise.all([
+                addSeedTasks(userId, teamId),
```
Dschoordsch, line 98: -1 We don't want to have the autojoin teams flooded with default tasks

packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
```diff
@@ -65,8 +83,36 @@ const bootstrapNewUser = async (
     anonymousId: segmentId
   })
 
+  const teamsWithAutoJoin = teamsWithAutoJoinRes.flat().filter(isValid)
   const tms = [] as string[]
-  if (isOrganic) {
+
+  if (teamsWithAutoJoin.length > 0) {
+    tms.push(...teamsWithAutoJoin.map(({id}) => id))
```
Dschoordsch, line 90: +1 This is done in `acceptTeamInvitation` already

packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
```diff
@@ -45,11 +59,15 @@ const bootstrapNewUser = async (
     experimentalFlags.push('retrosInDisguise')
   }
 
-  await Promise.all([
+  const isVerified = identities.some((identity) => identity.isEmailVerified)
+  const orgIds = organizations.map(({id}) => id)
+
+  const [teamsWithAutoJoinRes] = await Promise.all([
+    isVerified ? dataLoader.get('autoJoinTeamsByOrgId').loadMany(orgIds) : [],
```
Dschoordsch, line 66: +1 adding `isVerified && isCompanyDomain` here would do the trick

Dschoordsch (changes requested):
Did not test

Dschoordsch (changes requested):
Did not test, but looks good.
> -2 We're missing some verification mechanism for the organizations which have an auto join team. Even if we verify this when enabling the flag, there is still a chance an org could be abandoned and taken over by non-verified users. I think it would be good to verify an org has verified billing leads/founder similar to
> 
> https://github.com/ParabolInc/parabol/blob/e508bcfeaf79c5ea1d679d7ef75c6fdbfb35302c/packages/server/utils/isRequestToJoinDomainAllowed.ts#L10

☝️ that's still true. At the very minimum we should check this when adding the `autoJoin` flag to a team, but ideally we check it when users are about to join. Happy to move this into a different issue in the milestone, but that's a release blocker.

We also should check the `updateEmail`, because it allows a verified user to change their email to one which isn't verified.

nickoferrall:
> we should check this when adding the autoJoin flag to a team, but ideally we check it when users are about to join

Makes sense. I've created a separate PR for this: https://github.com/ParabolInc/parabol/pull/8989

> We also should check the updateEmail, because it allows a verified user to change their email to one which isn't verified.

With that PR, do you think we need to check `updateEmail`? Users won't be able to autoJoin a team if the org doesn't have a billing leader or founder with the same email domain & a verified email.

Dschoordsch:
> With that PR, do you think we need to check updateEmail? Users won't be able to autoJoin a team if the org doesn't have a billing leader or founder with the same email domain & a verified email.
The issue with `updateEmail` is slightly different: It does not modify the identities array which we check for `isEmailVerified`. So in theory someone could sign up, verify their email and then change it. We would consider the changed one still verified, although it isn't. Not a big deal yet as it's private. I'll just create a bug.

Dschoordsch (approved):

PR #8912, opened 2023-10-03
feat(meeting-inception): implement add an activity button

### LINKED ISSUE #8750: Meeting inception: add "add an activity" button
image

AC: 
- Add org level feature flag
- Add "Add an activity button" to the discussion panel 
- Button should simply redirect to the activity library for now

igorlesnenko:
Fixes #8750

image

How to test:

- Add `meetingInception` org level feature flag
- Open various discussion panels, click on `add an activity` see it redirect to the activity library


packages/client/components/DiscussionThreadInput.tsx
```diff
@@ -290,9 +300,11 @@ const DiscussionThreadInput = forwardRef((props: Props, ref: any) => {
     }
   }, [])
 
-  const isActionsContainerVisible = allowTasks || allowPolls
+  const allowAddActivity = featureFlags.meetingInception
+  const isActionsContainerVisible = allowTasks || allowPolls || allowAddActivity
```
nickoferrall, line 304: +1 to test the PR, we have to make `allowPolls` false, but will this be changed in a later PR or will we always need to do this locally?

We haven't worked on polls for ~2 years, so I think it should be removed from this file

packages/client/components/AddActivityButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+  onClick: () => void
+  dataCy: string
+  disabled?: boolean
+}
+
+const AddActivityButton = (props: Props) => {
+  const {onClick, dataCy, disabled} = props
+
+  return (
+    <StyledPlainButton data-cy={`${dataCy}-add`} onClick={onClick} disabled={disabled}>
```
nickoferrall, line 37: +1 we've removed cypress so no need for `data-cy`

nickoferrall (approved):
Sorry for the slow review - I've been OOO for a few days

packages/client/components/DiscussionThreadInput.tsx
igorlesnenko: Set to be always `false`

packages/client/components/DiscussionThreadInput.tsx
```diff
@@ -335,6 +347,15 @@ const DiscussionThreadInput = forwardRef((props: Props, ref: any) => {
               disabled={isActionsContainerDisabled}
             />
           )}
+          {allowAddActivity && (
+            <AddActivityButton
+              dataCy={`${dataCy}-activity`}
```
Dschoordsch, line 352: +1 This is not needed anymore, we don't use cypress anymore

Dschoordsch (approved):
I'm not a fan of the design. I haven't checked it with the zoom transcription UI update yet, but right now it looks a bit incohesive between the top, middle and bottom part of the panel
Screenshot 2023-10-24 at 17 49 38

tghanken (approved):
Looks good to me.  Were you intending to add analytics as part of this PR or as part of a follow-on PR?

igorlesnenko:
@tghanken I was thinking to add something in this PR so we can track if users actually clicking on the button. Do we need something special for this?

packages/client/components/DiscussionThreadInput.tsx
```diff
@@ -335,6 +347,14 @@ const DiscussionThreadInput = forwardRef((props: Props, ref: any) => {
               disabled={isActionsContainerDisabled}
             />
           )}
+          {allowAddActivity && (
+            <AddActivityButton
+              onClick={() => {
+                window.open(`/activity-library/category/recommended`, '_blank', 'noreferrer')
+              }}
```
tghanken, line 354: I'd add an `Add Activity Button Clicked` event here.

tghanken (commented):
I think this is the only place we really need to add an event.

PR #8907, opened 2023-09-28
feat(your-work): Jira Integration

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8900

Implement the Jira integration for Your Work.

Note that this does _not_ include pagination - the panel will load up to 20 issues, but not more. Pagination will be implemented in a separate follow-up PR, as the behavior + API of the Jira `issues` GQL resolver needs to be modified to correctly support pagination in this panel, which could cause regressions in Sprint Poker.

## Demo
https://www.loom.com/share/7c49530e24794d6fa16232bd97cfa2a3?sid=faeaf396-5f36-4934-8baf-8f51d8069581

## Testing scenarios
- [ ] Start a standup on a team where you do _not_ have Jira integrated
- [ ] Open the "Your Work" drawer, and click the Jira tab
- [ ] Confirm the panel prompts you to connect to Jira, and connect
- [ ] Confirm the panel automatically loads the issues
- [ ] Refresh the page, and open the panel to Jira, and confirm things load correctly
- [ ] Check card styles + behavior:
  - [ ] Icon correctly represents status
  - [ ] Number + title links to Jira issue
  - [ ] Project links to project
  - [ ] Link button shows hover tooltip and correctly copies link to Jira issue
- [ ] Check empty state - I suggest commenting out `jiraResults` in `JiraIntegrationResults` and setting it to an empty array to test
- [ ] Smoke test the Parabol tab
- [ ] Smoke test Jira in sprint poker

## Final checklist

- [X] I checked the [code review guidelines](../docs/codeReview.md)
- [X] I have added [Metrics Representative](../docs/codeReview.md#metrics-representative) as reviewer(s) if my PR invovles metrics/data/analytics related changes
- [X] I have performed a self-review of my code, the same way I'd do it for any other team member
- [X] I have tested all cases I listed in the testing scenarios and I haven't found any issues or regressions
- [X] Whenever I took a non-obvious choice I added a comment explaining why I did it this way
- [X] I added the label `One Review Required` if the PR introduces only minor changes, does not contain any architectural changes or does not introduce any new patterns and I think one review is sufficient'
- [X] PR title is human readable and could be used in changelog

PR #8916, opened 2023-10-03
feat(your-work): Jira Integration pagination

jmtaber129:
# Description
Fixes https://github.com/ParabolInc/parabol/issues/8917

Pagination for Your Work - Jira.

Note that because this modifies the Jira GQL API (both pagination args and behavior) there is a risk of regression in the Jira integration for Sprint Poker, so that should be smoke tested more rigorously than usual.

Needs rebase once https://github.com/ParabolInc/parabol/pull/8907 lands

## Demo
Multiple pages (2 per page):
https://www.loom.com/share/83e737abb0bc4c30a1b61f9d20bd29f3?sid=469f4e61-0ecb-44a4-9c40-f7e1dc92cdc2

Paginate on scroll to bottom (7 per page):
https://www.loom.com/share/7dcf95f1fe6249faad5d9fac9f6609ec?sid=112a19e7-eecd-4a06-96a5-1468a95fd8a2

## Testing scenarios
- [ ] Connect a team to Jira, and open a standup meeting on that team
- [ ] Assign several issues to yourself in Jira
- [ ] Set the page size to a number less than the number of assigned issues
- [ ] Confirm that when the end of the list is seen by the user, the next page of issues is loaded.
- [ ] Confirm that when there are no more assigned issues, the end of the list does not show a loading animation.
- [ ] Smoke test sprint poker for Jira.

## Final checklist

### DIFF
--- a/packages/server/graphql/private/typeDefs/_legacy.graphql
+++ b/packages/server/graphql/private/typeDefs/_legacy.graphql
@@ -912,9 +912,9 @@ A connection to a list of items.
 """
 type JiraIssueConnection {
   """
-  Page info with cursors coerced to ISO8601 dates
+  Page info with cursors
   """
-  pageInfo: PageInfoDateCursor
+  pageInfo: PageInfo
 
   """
   A list of edges.
@@ -935,7 +935,7 @@ type JiraIssueEdge {
   The item at the end of the edge
   """
   node: JiraIssue!
-  cursor: DateTime
+  cursor: String
 }
 
 interface TaskIntegration {
--- a/packages/server/graphql/public/types/AtlassianIntegration.ts
+++ b/packages/server/graphql/public/types/AtlassianIntegration.ts
@@ -2,17 +2,23 @@ import {downloadAndCacheImages, updateJiraImageUrls} from '../../../utils/atlass
 import AtlassianServerManager from '../../../utils/AtlassianServerManager'
 import {getUserId} from '../../../utils/authorization'
 import standardError from '../../../utils/standardError'
-import connectionFromTasks from '../../queries/helpers/connectionFromTasks'
 import {AtlassianIntegrationResolvers} from '../resolverTypes'
 
 const AtlassianIntegration: AtlassianIntegrationResolvers = {
   issues: async ({teamId, userId, accessToken, cloudIds}, args, {authToken}) => {
-    const {first, queryString, isJQL, projectKeyFilters} = args
+    const {first, queryString, isJQL, projectKeyFilters, after} = args
     const viewerId = getUserId(authToken)
     if (viewerId !== userId || !accessToken) {
       const err = new Error('Cannot access another team members issues')
       standardError(err, {tags: {teamId, userId}, userId: viewerId})
-      return connectionFromTasks([], 0, err)
+      return {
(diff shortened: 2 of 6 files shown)


packages/server/graphql/private/typeDefs/_legacy.graphql
```diff
@@ -935,7 +935,7 @@ type JiraIssueEdge {
   The item at the end of the edge
   """
   node: JiraIssue!
-  cursor: DateTime
```
jmtaber129, line 938: This connection wasn't truly implemented previously, since our GQL API only accepted a `first` arg.

Using a `DateTime` cursor here makes things confusing when paginating through items in descending date order, since we'd be querying for everything before the `after` date. This also goes against [relay best-practices](https://relay.dev/graphql/connections.htm#sec-Cursor) that specify that cursors should be opaque to clients.

Using a normal `String` cursor also allows us to forward the `first` + `before` GQL connection args directly to the Atlassian API, eliminating complexity around converting our GQL pagination to the Atlassian API's pagination.

This also increases consistency with how the [Jira Server GQL API](https://github.com/ParabolInc/parabol/blob/2b9a5c9d7f0b7eb0e800264bff999e17a735c18f/packages/server/graphql/private/typeDefs/_legacy.graphql#L1035-L1061) is implemented

packages/server/graphql/public/types/AtlassianIntegration.ts
```diff
@@ -28,7 +34,18 @@ const AtlassianIntegration: AtlassianIntegrationResolvers = {
         projectKeyFiltersByCloudId[cloudId] = []
       })
     }
-    const issueRes = await manager.getIssues(queryString ?? null, isJQL, projectKeyFiltersByCloudId)
+    // Request one extra item to see if there are more results
+    const maxResults = first + 1
+    // Relay requires the cursor to be a string
+    const afterInt = parseInt(after ?? '-1', 10)
```
BartoszJarocki, line 40: +1 technically `after` is a string so could be anything, and then `parseInt` will return `NaN`. would be good to check agains that

BartoszJarocki (approved):
just one little thing, nicely done otherwise!

packages/server/graphql/public/types/AtlassianIntegration.ts
```diff
@@ -28,7 +34,18 @@ const AtlassianIntegration: AtlassianIntegrationResolvers = {
...
+    const startAt = Number.isNaN(afterInt) ? -1 : afterInt + 1
```
Dschoordsch, line 41: +1 if `after === null` then `afterInt === -1` and `startAt === 0`.
If `after === 'invalid'` then `afterInt = NaN` and `startAt === -1`.
This seems a bit odd.

-1 I see we're using this value for cursor calculation, this will be wrong then.

Dschoordsch (changes requested):
Only did short smoke test

Dschoordsch (approved):
Are you going to fix pagination in poker scope as well? Otherwise can you please make sure we have an issue for it?

jmtaber129:
Created https://github.com/ParabolInc/parabol/issues/8962 to track pagination for Jira in Poker

PR #8918, opened 2023-10-03
fix: saml urls

### LINKED ISSUE #8894: Self serve SAML can throw error "Requested query string is too long"
A [customer](https://parabol.slack.com/archives/C4JAUUZ9P/p1695823767339129) 🔒 was having trouble setting up SAML with Microsoft. It appears sending the metadata as payload can be too long.

Proposed fix is:
* [ ] modify `parseSAMLMetadata` to store the new metadata in a temporary table (redis?) and only encode the key
* [ ] in `loginSAML` check the db

### LINKED ISSUE #8908: SAML UI: missing entries for config
Previously we had send the customer 3 URLs:
* ACS URL: https://action.parabol.co/saml/${slug}
* Entity ID: https://action.parabol.co/saml-metadata/${slug}
* Start URL: https://action.parabol.co

Currently we're only showing the first, which leads to wrong configs by the customers.

mattkrick:
# Description

fix #8908 
fix #8894

- Adds additional URLs necessary for users to set up SAML
- Adds support for accepting metadata URL instead of metadata XML blob
- Adds radix-ui tooltip component
- Wraps long URLs on mobile devices

## Demo

Screenshot 2023-10-03 at 2 09 11 PM

Screenshot 2023-10-03 at 2 09 42 PM

## Testing scenarios

- [ ] Upgrade to enterprise, then set up parabol in okta (optional)
- [ ] paste in our dev metadata url: https://dev-244808.okta.com/app/exk2960i5d08RfraK357/sso/saml/metadata
- [ ] see an okta login popup. if you have a user account & you're able to login successfully, the metadata + url will be persisted to the DB
- [ ] do it on mobile, it still looks good
- [ ] can click the URL or the icon to copy the URL & get a nice tooltip


packages/client/ui/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1 @@
+export {Root as Tooltip} from '@radix-ui/react-tooltip'
```
mattkrick, line 1: re-exporting the primitive so future devs don't see radix-ui, rather they just import from the /ui directory

packages/client/ui/Tooltip/TooltipContent.tsx
```diff
@@ -0,0 +1,24 @@
+import {Content, Portal} from '@radix-ui/react-tooltip'
+import * as React from 'react'
+import {twMerge} from 'tailwind-merge'
+import {forwardRadix} from '../fordwardRadix'
+
+export const TooltipContent = forwardRadix<typeof Content>(
```
mattkrick, line 6: added forwardRadix to get strong types without having to dig into the radix components to figure out props & element types

packages/server/graphql/public/types/SAML.ts
```diff
@@ -10,6 +10,11 @@ const SAML: SamlResolvers = {
   lastUpdatedByUser: async ({lastUpdatedBy}, _args, {dataLoader}) => {
     return dataLoader.get('users').loadNonNull(lastUpdatedBy)
   },
+  metadataURL: ({metadataURL, metadata}) => {
+    if (metadataURL) return metadataURL
+    if (metadata) return 'Unknown'
```
mattkrick, line 15: this is for existing users for whom we don't have metadataURL

packages/server/postgres/migrations/1696275124427_samlMetadataURL.ts
```diff
@@ -0,0 +1,21 @@
+import {Client} from 'pg'
+import getPgConfig from '../getPgConfig'
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+  ALTER TABLE "SAML"
+  ADD COLUMN IF NOT EXISTS "metadataURL" VARCHAR(2048);
```
mattkrick, line 9: microsoft azure only allows urls of 2048 chars, so borrowing their limit here
we could still run into problems since this URL gets encoded into another URL, but under normal usage we shouldn't have a URL that is longer than 200 chars.

packages/client/modules/userDashboard/components/OrgAuthentication/OrgAuthenticationMetadata.tsx
```diff
@@ -102,56 +80,42 @@ const OrgAuthenticationMetadata = (props: Props) => {
...
+        <BasicInput
+          name='metadataURL'
+          placeholder={`https://idp.example.com/app/sso/saml/metadata`}
+          value={metadataURL}
+          onChange={(e) => setMetadataURL(e.target.value)}
+          error={undefined}
         />
-      </InputSection>
+      </div>
       <div className={'px-4 text-tomato-500 empty:hidden'}>{error?.message}</div>
```
igorlesnenko, line 108: nit: paddings could be slightly tweaked for the error

image

igorlesnenko (approved):
All looks good!

Just 2 things:
- Test are failing, will you fix them?
- I don't have okta account, so not fully tested. Can you add me?

mattkrick:
> Test are failing, will you fix them?

The way we do server testing doesn't allow us to mock modules (the GQL server is running in a separate process), so I can't mock the `fetch` function so I just disabled that test

> I don't have okta account, so not fully tested. Can you add me?

You're in! https://dev-244808-admin.okta.com/admin/user/profile/view/00uwl9h33rqHdLyzQ357

mattkrick:
merging this directly to staging to avoid a hotfix since i'm releasing the next version today

PR #8923, opened 2023-10-04
fix: fix check-in discussion panel layout

### LINKED ISSUE #8922: Discussion panel layout broken when it has no comments
See the screenshot

image

AC:
- discussion panel works correctly in an empty state

igorlesnenko:
fix #8922
image

How to test:
- See that in check-in, retro and other meeting types the panel works as expected
- See that zoom functionality still works as expected

### DIFF
--- a/packages/client/components/DiscussionThreadListEmptyState.tsx
+++ b/packages/client/components/DiscussionThreadListEmptyState.tsx
@@ -132,14 +132,14 @@ const DiscussionThreadListEmptyState = (props: Props) => {
       }
     }
   })
-  if (!meeting) return null
-  const {id: meetingId, videoMeetingURL} = meeting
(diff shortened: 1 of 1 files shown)


packages/client/components/DiscussionThreadListEmptyState.tsx
```diff
@@ -132,14 +132,14 @@ const DiscussionThreadListEmptyState = (props: Props) => {
       }
     }
   })
-  if (!meeting) return null
```
igorlesnenko, line 135: The issue is that this component is not used only in retro meetings, so if meeting is not retro we don't show beautiful empty state at all

rafaelromcar-parabol:
@mattkrick

jmtaber129 (approved):

ackernaut:
Thanks for the fix!

PR #8926, opened 2023-10-04
chore: test sentry and dd uploads

### LINKED ISSUE #8925: Push to datadog & sentry in GH action `build`
AC

When a new version tag is pushed to GH & we dockerize the build, we should also push sourcemaps to sentry & datadog at the same time. This is currently done in CircleCI. We'll keep doing it in CircleCI & additionally do it in GH action `build`. We will entirely cut ties with CircleCI once we move to the new infra.

mattkrick:
# Description

fixes #8925

## Demo

Demo was performed stubbing in `v6.1220.0-exotic` for version & removing the `if:` clause in the yml

- [x] Can upload to Sentry: https://parabol.sentry.io/settings/projects/action-production/source-maps/release-bundles/6.122.0-exotic/
- [x] Can upload to Datadog: `✅ Uploaded 234 sourcemaps in 7.214 seconds.` (can't confirm in datadog UI 😠 )

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -54,7 +54,7 @@ jobs:
           echo "DOCKERIZE=${DOCKERIZE}" >> $GITHUB_ENV
 
           ACTION_VERSION=$(grep '"version":' package.json | cut -d\" -f4)
-          echo "ACTION_VERSION=6.122.0-exotic" >> $GITHUB_ENV
+          echo "ACTION_VERSION=${ACTION_VERSION}" >> $GITHUB_ENV
 
       - name: Setup Node
         uses: actions/setup-node@v3
@@ -95,6 +95,17 @@ jobs:
 
       - name: Verify source is clean
         run: git diff --quiet HEAD || (echo "Changes in generated files detected"; git diff; exit 1)
+
+      - name: Push Artifacts to Datadog
+        #if: github.ref_type == 'tag'
+        env:
+          DATADOG_API_KEY: "${{secrets.DATADOG_API_KEY}}"
+        run: |
+
+          yarn datadog-ci sourcemaps upload ./build \
+            --service=parabol-saas-production \
+            --release-version=${{env.ACTION_VERSION}} \
+            --minified-path-prefix=https://action-files.parabol.co/production/build/
 
       - name: Check Code Quality
         run: yarn codecheck
@@ -205,11 +216,3 @@ jobs:
           environment: "production"
           sourcemaps: "./build"
           version: ${{env.ACTION_VERSION}}
-      - name: Push Artifacts to Datadog
-        #if: github.ref_type == 'tag'
-        run: |
-          yarn add @datadog/datadog-ci -W
-          yarn datadog-ci sourcemaps upload ./build \
-            --service=parabol-saas-production \
-            --release-version=${{env.ACTION_VERSION}} \
-            --minified-path-prefix=https://action-files.parabol.co/production/build/
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -62,6 +62,7 @@ jobs:
           node-version-file: package.json
           # Caching yarn dir & running yarn install is too slow
           # Instead, we aggressively cache node_modules below to avoid calling install
+
       - name: Get cached node modules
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -176,11 +179,6 @@ jobs:
           registry: ${{ secrets.GCP_DOCKER_REGISTRY }}
           username: "oauth2accesstoken"
           password: "${{ steps.auth.outputs.access_token }}"
-      #
```
mattkrick, line 179: chances are we'll never build for anything but linux, but just in case i moved the comments to an issue here:
documented here https://github.com/ParabolInc/parabol/issues/8929

.github/workflows/build.yml
```diff
@@ -196,4 +194,25 @@ jobs:
...
             "${{ secrets.GCP_AR_PARABOL_DEV }}:${{github.sha}}"
             "${{ env.DOCKER_REPOSITORY_FOR_REF }}:${{ env.GITHUB_REF_NAME_NORMALIZED }}"
-          # platforms: linux/amd64,linux/arm64 # Uncomment to be able to build multi-platform images.
+      - name: Push Artifacts to Sentry
+        if: github.ref_type == 'tag'
+        uses: getsentry/action-release@v1
+        env:
+          SENTRY_AUTH_TOKEN: "${{secrets.SENTRY_AUTH_TOKEN}}"
+          SENTRY_ORG: "parabol"
+          SENTRY_PROJECT: "action-production"
```
rafaelromcar-parabol, line 203: Shouldn't we upload it to both action-production and action-staging projects?

.github/workflows/build.yml
```diff
@@ -196,4 +194,25 @@ jobs:
...
+        if: github.ref_type == 'tag'
+        env:
+          DATADOG_API_KEY: "${{secrets.DATADOG_API_KEY}}"
+        run: |
+          echo "$DATADOG_API_KEY"
+          yarn add @datadog/datadog-ci -W
+          yarn datadog-ci sourcemaps upload ./build \
+            --service=parabol-saas-production \
+            --release-version=${{env.ACTION_VERSION}} \
+            --minified-path-prefix=https://action-files.parabol.co/production/build/
```
rafaelromcar-parabol, line 218: This must be a variable. Once we migrate to GCP GCS the URL changes. Also, we might want to hide it

rafaelromcar-parabol:
Shouldn't we also remove that uploads from CircleCI on this PR? There is only one case where that might be a problem: a hotfix deployed in production without producing a new tag.

And BTW, that would be a problem with our current deployment process and with the future deployment process as well 🤔

.github/workflows/build.yml
mattkrick: I believe the plan is to change the DNS record for action-files to point to GCS instead of S3, right?

.github/workflows/build.yml
mattkrick: the name is a bit of a misnomer, staging is an environment inside the project `action-production`.

mattkrick:
hotfixes create new tags, too. no harm in leaving it in circleCI, assets will just get pushed twice until we get completely off circleCI

.github/workflows/build.yml
rafaelromcar-parabol: That is correct. Still, I feel it better also more secure) to manage that with a variable.

.github/workflows/build.yml
rafaelromcar-parabol: 🤔 And do we want that? Some time ago I created two new projects in Sentry. Right now we have:
- action-production: used by both Dokku environments (production and staging) in Digital Ocean.
- action-staging: used right now only by the GCP Staging.
- local-development: created in case someone wanted to use sentry in their local environment.
- on-demand-environments: isolated project for the on-demand envs.

I don't really know if that make sense. I created it to isolate the new GCP envs I was building, at least while building and testing.

mattkrick: IMO a project refers to a codebase. we have 1 project, which is "action", and that is deployed to many environments. In sentry you can search & filter across envs, but not across projects (at least that was the case, not sure if still true).

PR #8938, opened 2023-10-05
fix: CI=true for build action

mattkrick:
draftEnterpriseInvoice will fail the build action if `process.env.CI !== 'true'`. this occurs whenever an outsider triggers the build because today we have `CI=true`in our repo env vars.

TEST

- [ ] as an outsider (e.g. dependabot) run the build action & see it pass (namely the draftEnterpriseInvoice test should pass)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -8,6 +8,8 @@ env:
 jobs:
   build:
     runs-on: ubuntu-8cores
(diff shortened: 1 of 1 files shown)


Dschoordsch (approved):
We have a couple dependabot PRs, let's merge this and see if they pass.

Issue #8924, opened 2023-10-04
Your Work: Add analytics


jmtaber129:
Add analytics tracking for the Your Work drawer.

AC:
Instrumented the following:
- [ ] Opened your work drawer (either click or impression)
- [ ] Integration panel impressions (i.e. viewed Parabol, GitHub, etc.)
- [ ] Connect integration from panel
- [ ] Integration card interactions
  - [ ] Clicked link
  - [ ] Copied link/title
- [ ] Drawer closed
- [ ] Repo/Project filtered

tghanken:
The only thing I'd add is closing the work drawer, so we can figure out if users are viewing but not getting value out of the feature.

jmtaber129:
@tghanken should tracking for closing the work drawer be for _explicitly_ closing the drawer, or just anytime the drawer closes? For example, if the work drawer closes because the user opened a discussion drawer, should we track that as the user closing the drawer even though the user didn't explicitly close the work drawer?

tghanken:
I only think we need to track explicit closes.  We're looking to gauge engagement with the new feature, and I think explicit closes are going to be a much stronger signal for what we're looking for.

PR #8956, opened 2023-10-10
feat(your-work): Product analytics

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8924

Add analytics for Your Work drawer.

Needs rebase once [gCal PR](https://github.com/ParabolInc/parabol/pull/8939) lands.

## Demo
N/A

## Testing scenarios
- [ ] Test each metric listed in the issue:
  - [ ] Open/close drawer:
    - [ ] Via impression (e.g. default open on load)
    - [ ] Via top bar
    - [ ] Via drawer (e.g. close icon + swipe-to-close on smaller screens)
  - [ ] Clicks into individual integration panels
  - [ ] Click to connect integration
  - [ ] Integration cards for GitHub, Jira, and gCal:
    - [ ] Click on links
    - [ ] Click copy
  - [ ] Change GitHub repo filter
- [ ] Does not track opens/closes for discussion drawer
- [ ] Does not track close when switching from your work to discussion

## Final checklist

### DIFF
--- a/packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
+++ b/packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
@@ -19,6 +19,7 @@ import {TeamPromptMeetingStatus} from './TeamPromptMeetingStatus'
 import TeamPromptOptions from './TeamPromptOptions'
 import {KeyboardArrowLeft, KeyboardArrowRight} from '@mui/icons-material'
 import IconLabel from '../IconLabel'
+import SendClientSegmentEventMutation from '../../mutations/SendClientSegmentEventMutation'
 
 const TeamPromptLogoBlock = styled(LogoBlock)({
   marginRight: '8px',
@@ -95,6 +96,7 @@ const TeamPromptTopBar = (props: Props) => {
       fragment TeamPromptTopBar_meeting on TeamPromptMeeting {
         id
         name
+        teamId
         isRightDrawerOpen
         showWorkSidebar
         facilitatorUserId
@@ -139,13 +141,25 @@ const TeamPromptTopBar = (props: Props) => {
 
   const onOpenWorkSidebar = () => {
     if (meeting.isRightDrawerOpen && meeting.showWorkSidebar && !meeting.localStageId) {
-      // If we're selecting a discussion that's already open, just close the drawer.
+      // If we're clicking on 'Your Work' when it's already open, just close the drawer.
+      SendClientSegmentEventMutation(atmosphere, 'Your Work Drawer Closed', {
+        teamId: meeting.teamId,
+        meetingId: meeting.id,
+        source: 'top bar'
+      })
+
       commitLocalUpdate(atmosphere, (store) => {
         const meetingProxy = store.get(meetingId)
         if (!meetingProxy) return
         meetingProxy.setValue(false, 'isRightDrawerOpen')
       })
     } else {
+      SendClientSegmentEventMutation(atmosphere, 'Your Work Drawer Opened', {
+        teamId: meeting.teamId,
+        meetingId: meeting.id,
+        source: 'top bar'
+      })
+
       commitLocalUpdate(atmosphere, (store) => {
         const meetingProxy = store.get(meetingId)
         if (!meetingProxy) return
--- a/packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
+++ b/packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
@@ -142,30 +142,30 @@ const TeamPromptTopBar = (props: Props) => {
   const onOpenWorkSidebar = () => {
     if (meeting.isRightDrawerOpen && meeting.showWorkSidebar && !meeting.localStageId) {
       // If we're clicking on 'Your Work' when it's already open, just close the drawer.
-      SendClientSegmentEventMutation(atmosphere, 'Your Work Drawer Closed', {
-        teamId: meeting.teamId,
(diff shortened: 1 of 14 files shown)


packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
```diff
@@ -139,13 +141,25 @@ const TeamPromptTopBar = (props: Props) => {
 
   const onOpenWorkSidebar = () => {
     if (meeting.isRightDrawerOpen && meeting.showWorkSidebar && !meeting.localStageId) {
-      // If we're selecting a discussion that's already open, just close the drawer.
+      // If we're clicking on 'Your Work' when it's already open, just close the drawer.
+      SendClientSegmentEventMutation(atmosphere, 'Your Work Drawer Closed', {
+        teamId: meeting.teamId,
+        meetingId: meeting.id,
+        source: 'top bar'
+      })
```
tianrunhe, line 149: -1: emit the metric after the local commit

packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
```diff
@@ -139,13 +141,25 @@ const TeamPromptTopBar = (props: Props) => {
...
         const meetingProxy = store.get(meetingId)
         if (!meetingProxy) return
         meetingProxy.setValue(false, 'isRightDrawerOpen')
       })
     } else {
+      SendClientSegmentEventMutation(atmosphere, 'Your Work Drawer Opened', {
+        teamId: meeting.teamId,
+        meetingId: meeting.id,
+        source: 'top bar'
+      })
```
tianrunhe, line 161: -1 emit the metric after the local commit

packages/client/components/TeamPrompt/TeamPromptWorkDrawer.tsx
```diff
@@ -39,12 +43,22 @@ const TeamPromptWorkDrawer = (props: Props) => {
     `,
     meetingRef
   )
+  const atmosphere = useAtmosphere()
+
+  useEffect(() => {
+    SendClientSegmentEventMutation(atmosphere, 'Your Work Drawer Opened', {
+      teamId: meeting.teamId,
+      meetingId: meeting.id,
+      source: 'impression'
```
tianrunhe, line 52: -1 If I have the discussion panel opened, and then I clicked "Your work" from the top bar, I got two `Your Work Drawer Opened` events, one with the source `'top bar'`, the other with the source `'impression'`. I don't think the latter should be emitted?

packages/client/components/TeamPrompt/TeamPromptWorkDrawer.tsx
```diff
@@ -80,7 +94,14 @@ const TeamPromptWorkDrawer = (props: Props) => {
             {baseTabs.map((tab, idx) => (
               <Tab
                 key={tab.label}
-                onClick={() => setActiveIdx(idx)}
+                onClick={() => {
+                  SendClientSegmentEventMutation(atmosphere, 'Your Work Integration Clicked', {
+                    teamId: meeting.teamId,
+                    meetingId: meeting.id,
+                    integrationLabel: baseTabs[idx]?.label
```
tianrunhe, line 101: +1 Let's use `service` here to align naming as in, e.g., [`Integration Added`](https://github.com/ParabolInc/parabol/blob/00447aa9ac97e52449de23f3284e09cae9ff7bfe/packages/server/utils/analytics/analytics.ts#L387)

tianrunhe (changes requested):
Let's rename all `serviceLabel` to `service`, just to align naming with, e.g., [`Integration Added`](https://github.com/ParabolInc/parabol/blob/00447aa9ac97e52449de23f3284e09cae9ff7bfe/packages/server/utils/analytics/analytics.ts#L387) and let's use the standard service name defined [here](https://github.com/ParabolInc/parabol/blob/jmtaber129/your-work-analytics/packages/server/graphql/types/IntegrationProviderServiceEnum.ts#L19)

packages/client/components/TeamPrompt/TeamPromptWorkDrawer.tsx
jmtaber129: This is tricky, since we need the `'impression'` trigger when it's open by default, and we need the `'top bar'` trigger when the user clicks to open it.

What if we changed the `impression` version to a different event like `'Your Work Drawer Impression'`? That way we can still do analysis based on when a user sees the work drawer.

packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
jmtaber129: Fixed

packages/client/components/TeamPrompt/TeamPromptTopBar.tsx
jmtaber129: Fixed

jmtaber129:
@tianrunhe 
> let's use the standard service name

The current service enums are either integration services (which excludes Parabol) or task services (which excludes gcal), so we can't use either of those out of the box.

I went ahead and created a more generic `ServiceEnum` that includes all services for this.

tianrunhe (approved):

packages/client/components/TeamPrompt/TeamPromptWorkDrawer.tsx
tianrunhe: Yeah that works!

BartoszJarocki (approved):

PR #8965, opened 2023-10-12
feat: upgrade PG to v15.4 for local dev

mattkrick:
# Description

Fixes odd locales errors when building with `yarn db:start --build`

Rumor is something in the locales package changed (https://github.com/pgvector/pgvector/issues/197#issuecomment-1653904858) but that was back in July & I'm pretty sure I've rebuilt since then, so I don't know why it's happening now.

## Demo

run `yarn db:start --build` and it should work. Since this also bumps to PG v15, you'll have to `docker volume rm` your postgres data.


rafaelromcar-parabol:
**I suggest you change the name of this PR to `feat(postgresql): update to 15.4`**. Adding the locale is just an addition, but the big change if the upgrade to 15.4.

I updated all references to PostgreSQL so we have `15.4` everywhere. I removed the `-alpine` to simplify it and use the exact same image, which will help in future upgrades, as we will just need to look for `15.4` and change to the new version.

The [build action](https://github.com/ParabolInc/parabol/actions/runs/6531538009/job/17732950100) works and builds a Docker image.

I upgraded our GCP CloudSQL in the new environments and I will test the data restoration, to explore possible problems. I'll keep you posted.

rafaelromcar-parabol:
This must be merged asap as we are already in 15.4 in both production and staging. Is there anything more to do here?

PR #8970, opened 2023-10-16
feat(redis): upgrade to use Redis 7

rafaelromcar-parabol:
# Description

We are already using Redis 7 in both Digital Ocean and GCP. This PR upgrades all Redis references to 7.0.

- It uses 7.0, without a patch version, because it is what's configured in GCP. We cannot control the patch upgrades. By definition, these images aren't inmutable, but the risk is very low and we need to test things ahead of GCP automated upgrades.
- The image  [Docker official 7.0-alpine](https://hub.docker.com/_/redis/tags?page=1&name=7.0-alpine) is used everywhere.
- Redis TLS documentation has been modified in order to ease the use of that functionality in development if required.

## Testing scenarios

- [x] Build action works and produces a Docker image.
- [x] Developers can execute their local environment without problem.

### DIFF
--- a/docker/dev.yml
+++ b/docker/dev.yml
@@ -51,21 +51,12 @@ services:
       - parabol-network
     restart: unless-stopped
   redis:
-    image: bitnami/redis:7.0-alpine
+    image: redis:7.0-alpine
     restart: unless-stopped
-    environment:
-      - ALLOW_EMPTY_PASSWORD=yes
-      - REDIS_PASSWORD=''
-      - REDIS_TLS_ENABLED=no
-      - REDIS_TLS_AUTH_CLIENTS=no
-      - REDIS_TLS_CERT_FILE=/opt/bitnami/redis/certs/redis.crt
-      - REDIS_TLS_KEY_FILE=/opt/bitnami/redis/certs/redis.key
-      - REDIS_TLS_CA_FILE=/opt/bitnami/redis/certs/redisCA.crt
     ports:
       - "6379:6379"
     volumes:
-      - bitnami-redis-data:/bitnami/redis/data
-      - ../certs:/opt/bitnami/redis/certs
+      - redis-data:/data
     networks:
       - parabol-network
   redis-commander:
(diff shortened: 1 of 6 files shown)


rafaelromcar-parabol:
Build action works and produces the Docker image. See [here](https://github.com/ParabolInc/parabol/actions/runs/6531285688/job/17732222082)

docker/dev.yml
```diff
@@ -51,7 +51,7 @@ services:
       - parabol-network
     restart: unless-stopped
   redis:
-    image: bitnami/redis:6.2
+    image: bitnami/redis:7.0-debian-11
```
Dschoordsch, line 54: -1 I don't see why we should use the debian image locally when we can get away with the smaller alpine everywhere else. I don't quite understand the TLS argument either as we're not using TLS locally in any of our environments.

Dschoordsch (changes requested):
`yarn dev` locally works fine with the new image

docker/dev.yml
rafaelromcar-parabol: We could just rollback to the basic image using `7.0-alpine` but if ever we need to test or use the TLS feature, we would need to adapt the configuration for it. See [this](https://github.com/ParabolInc/parabol/pull/8488#issuecomment-1701209184) and [this](https://github.com/ParabolInc/parabol/pull/8488#issuecomment-1701546727).

That said, I think you are right and I'm changing it to redis:7.0-alpine. If we ever need to test that again, we will see how to do so.

docker/dev.yml
```diff
@@ -51,7 +51,7 @@ services:
       - parabol-network
     restart: unless-stopped
   redis:
-    image: bitnami/redis:6.2
+    image: bitnami/redis:7.0-alpine
```
Dschoordsch, line 54: -1 please remove `bitnami/`

Dschoordsch (changes requested):

docker/dev.yml
rafaelromcar-parabol: Sorry, I was doing more changes. You can test now, and check the modified doc as well.

Dschoordsch (approved):

rafaelromcar-parabol:
I assume you checked `yarn dev` back again @Dschoordsch ?

Dschoordsch:
Yes, I did.

PR #8977, opened 2023-10-16
fix: dailyPulse handle nulls

mattkrick:
# Description

Fixes our internal daily pulse.
On Sept 25 we started using a new dataloader to determine if something is a company domain.
Dataloaders can't handle `null`. https://github.com/ParabolInc/parabol/blob/0508fad31335d0cedb324237fcb6342a83c56793/packages/server/dataloader/customLoaderMakers.ts#L709
We could make the dataloader handle nulls, or we could ensure that only non-nulls make it to the dataloader.
Since none of the other dataloders handle `null`, I figured the latter was the better option.

## Demo

This relies on prod data to verify the fix.

## Testing scenarios

In admin/graphql run this

```gql
query DailyPulse($after: DateTime!, $email: String!, $channelId: ID!) {
  dailyPulse(after: $after, email: $email, channelId: $channelId)
}
```

with these vars (replace X with what you see in the dokku prod env)

```json
{
  "after": "2023-10-15T16:05:39.810Z",
  "email": "X",
  "channelId": "X"
}
```

### DIFF
--- a/packages/server/graphql/private/queries/helpers/authCountByDomain.ts
+++ b/packages/server/graphql/private/queries/helpers/authCountByDomain.ts
@@ -19,6 +19,7 @@ const authCountByDomain = async (
         `SELECT count(*)::float as "total", "domain" from "User"
          WHERE (NOT $1 OR "inactive" = FALSE)
          AND "${filterField}" >= $2
+         AND domain IS NOT NULL
          GROUP BY "domain"
          ORDER BY "total" DESC`,
         [countOnlyActive ?? false, after]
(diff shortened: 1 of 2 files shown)


packages/server/graphql/private/queries/helpers/authCountByDomain.ts
```diff
@@ -19,13 +19,15 @@ const authCountByDomain = async (
         `SELECT count(*)::float as "total", "domain" from "User"
          WHERE (NOT $1 OR "inactive" = FALSE)
          AND "${filterField}" >= $2
+         AND domain IS NOT NULL
```
Dschoordsch, line 22: This basically excludes removed users which is a good thing anyways

Dschoordsch (approved):

packages/server/graphql/private/queries/helpers/authCountByDomain.ts
mattkrick: yep! there are also a handful of users who were created where their email doesn't even include an `@` 🤔. I thought we had a regex protecting against that, so they got tricky.

PR #8985, opened 2023-10-18
feat: adding integrations to team view tabs

acressall:
# Description
Adds an integration tab to the team dashboard, and removes it from settings

## Demo
Screenshot 2023-10-18 at 2 54 56 PM
Screenshot 2023-10-18 at 2 58 33 PM

## Final checklist

### DIFF
--- a/packages/client/modules/teamDashboard/components/TeamDashHeader/TeamDashHeader.tsx
+++ b/packages/client/modules/teamDashboard/components/TeamDashHeader/TeamDashHeader.tsx
@@ -113,6 +113,7 @@ const TeamDashHeader = (props: Props) => {
   const {organization, id: teamId, name: teamName, teamMembers} = team
   const {name: orgName, id: orgId} = organization
   const isTasks = useRouteMatch('/team/:teamId/tasks')
+  const isIntegrations = useRouteMatch('/team/:teamId/integrations')
   const {history} = useRouter()
 
   return (
@@ -143,7 +144,7 @@ const TeamDashHeader = (props: Props) => {
                     title={'Settings & Integrations'}
                     to={`/team/${teamId}/settings/`}
                   >
-                    {'Settings & Integrations'}
+                    {'Settings'}
                   </NavLink>
                 )
               }}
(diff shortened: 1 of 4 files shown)


packages/client/modules/teamDashboard/components/TeamDashHeader/TeamDashHeader.tsx
```diff
@@ -157,11 +158,12 @@ const TeamDashHeader = (props: Props) => {
         </Avatars>
       </TeamHeaderAndAvatars>
       <Tabs
-        activeIdx={isTasks ? 1 : 0}
+        activeIdx={isTasks ? 1 : 0 || isIntegrations ? 2 : 0}
```
igorlesnenko, line 161: +1 I think it could be more readable, maybe something like:

```
const getActiveIndex = () => {
  if (isTasks) return 1;
  if (isIntegrations) return 2;
  return 0;
};
```

igorlesnenko (approved):
LGTM!

image

Looks a bit empty now, but I think it is ok for now

packages/client/modules/teamDashboard/components/TeamSettingsWrapper/TeamSettingsWrapper.tsx
```diff
@@ -23,17 +22,12 @@ const IntegrationPage = styled('div')({
...
-  } = props
+  const {match} = props
   const {
     params: {teamId}
   } = match
-  const areaMatch = matchPath(pathname, {path: `${match.url}/:area?`}) || {params: {area: ''}}
   return (
     <IntegrationPage>
-      <TeamSettingsToggleNav activeKey={(areaMatch.params as any).area || ''} teamId={teamId} />
       <Switch>
```
Dschoordsch, line 31: -1 We don't need this `<Switch>` here anymore. In fact, we can remove the whole file and just use the `TeamSettingsRoot` component in the place where we previously used this `TeamSettingsWrapper` component.

Dschoordsch (changes requested):
@acressall do you still want to move forward with this? This was hanging in limbo for quite some time

acressall:
> @acressall do you still want to move forward with this? This was hanging in limbo for quite some time

Honestly, I forgot about it and then last week I remembered and thought I should finish it up, but ran into problems when rebasing. I think we should either close and do later (since it's small) or someone else can take it over

Dschoordsch (approved):

packages/client/modules/teamDashboard/containers/TeamIntegrationsRoot/TeamIntegrationsRoot.tsx
```diff
@@ -1,3 +1,4 @@
+import styled from '@emotion/styled'
```
igorlesnenko, line 1: We don't import styled anymore and using tailwind for new components

igorlesnenko (approved):
Didn't test again

packages/client/modules/teamDashboard/containers/TeamIntegrationsRoot/TeamIntegrationsRoot.tsx
Dschoordsch: Yes, I just moved this from `TeamIntegrationsWrapper`.

Issue #8955, opened 2023-10-10
Standups: responses permalinks


enriquesanchez:
## Issue - Enhancement

![image](https://github.com/ParabolInc/parabol/assets/3276087/00490745-2539-4178-b234-081a4d526fcf)

Allow users to link to individual responses in a Standup.

## Acceptance Criteria
Users can:
   - Copy a link that directs to an individual response
     - The permalink should open a view were the response is focused and the comment thread open

**Estimated effort:** X points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

PR #8986, opened 2023-10-18
feat(standups): Response permalink sharing

jmtaber129:
# Description
Fixes https://github.com/ParabolInc/parabol/issues/8955

Support copying permalinks of standup responses.

## Demo
https://www.loom.com/share/407d84a2400c4bc1b371b7cbc6db28b8?sid=20ed10d4-0c08-4646-b3a7-77645604eef0

## Testing scenarios
- [ ] Link button not present for responses that haven't been submitted
- [ ] Link button is present for responses that have been submitted
- [ ] Clicking link button copies the response permalink to the clipboard

## Final checklist

### DIFF
--- a/packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
+++ b/packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
@@ -4,6 +4,7 @@ import {JSONContent} from '@tiptap/react'
 import graphql from 'babel-plugin-relay/macro'
 import React, {useMemo} from 'react'
 import {commitLocalUpdate, useFragment} from 'react-relay'
+import CopyToClipboard from 'react-copy-to-clipboard'
 import useAnimatedCard from '~/hooks/useAnimatedCard'
 import useAtmosphere from '~/hooks/useAtmosphere'
 import useEventCallback from '~/hooks/useEventCallback'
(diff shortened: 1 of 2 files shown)


packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
```diff
@@ -210,6 +217,35 @@ const TeamPromptResponseCard = (props: Props) => {
 
   const ref = useAnimatedCard(displayIdx, status)
 
+  const responsePermalink = makeAppURL(window.location.origin, `/meet/${meetingId}/responses`, {
+    searchParams: {
+      source: 'sharing',
```
jmtaber129, line 222: Including a full set of UTM params felt like overkill that users wouldn't like, but I think there's value in being able to see when a shared link was opened, hence a `source` param that's separate from the UTMs.

This is the approach Amplitude uses for its `Copy URL` functionality.

jmtaber129:
@tianrunhe mind taking a look at the tracking params + UTM type changes?

packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
tianrunhe: I think shortening the full set of UTM parameters are fine but instead of `source` we should stay with `utm_source`. I did a quick experiment, it looks like GA4 wouldn't pick up the `source` parameter, but `utm_source` works:

Screenshot 2023-10-18 at 14 36 50

tianrunhe (changes requested):

packages/client/components/TeamPrompt/TeamPromptResponseCard.tsx
jmtaber129: Fair enough! Switched this to `utm_source`.

tianrunhe (approved):

BartoszJarocki (approved):

Issue #8988, opened 2023-10-19
Auto-Join: verify orgs when auto-joining


nickoferrall:
See comment here: https://github.com/ParabolInc/parabol/pull/8883#pullrequestreview-1687710279

An org could be abandoned and taken over by non-verified users. Check that the org is valid before auto-joining. 

### AC

- [ ] Before a new user auto-joins a team in the org, check if the org has a founder or billing lead that both has a verified email and an email domain that matches the org's `activeDomain`.

PR #8989, opened 2023-10-19
chore(auto-join): verify org founder or billing leads

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8988

### To Test

- [ ] Sign-in with your Parabol gmail account and create a team call `AutoJoin team`
- [ ] Run `updateAutoJoin` to add the autoJoin flag to the AutoJoin team
- [ ] Run `addApprovedOrganizationDomains` for the `parabol.co` domain
- [ ] Sign-up with your email + something after it, e.g. `nick+autojoin@parabol.co`
- [ ] Verify your email address and see that you automatically join the `AutoJoin team`


nickoferrall:
Thanks @Dschoordsch, I've updated the testing steps 👍

packages/server/dataloader/customLoaderMakers.ts
```diff
@@ -698,14 +700,79 @@ export const samlByOrgId = (parent: RootDataLoader) => {
...
+        validOrgs.find((validOrg) => validOrg.id === orgId) ? [orgId] : []
+      )
+    },
+    {
+      ...parent.dataLoaderOptions
+    }
+  )
+}
+
+export const autoJoinTeamsByValidOrgId = (parent: RootDataLoader) => {
```
Dschoordsch, line 764: +1 I think it would be fair to still call it `autoJoinTeamsByOrgId`

Dschoordsch (changes requested):
I did some [tests](https://github.com/ParabolInc/parabol/pull/9036) and it's failing for an inactive founder.

nickoferrall:
Thanks for adding the tests! The problem was I was getting the `founder` using: `org('members').filter({inactive: false}).nth(0)`, but I wanted to get the first user and could then check if that user is inactive.

Dschoordsch (approved):

Issue #8678, opened 2023-08-15
Add client side Amplitude event tracking


tghanken:
Add the amplitude browser sdk to our client application, and write analytics events directly to Amplitude in parallel with Segment.

### Acceptance Criteria (optional)
- All client analytics events tracked to both Segment and Amplitude

PR #8992, opened 2023-10-19
chore(metrics): Add client side metric Amplitude tracking

tianrunhe:
# Description

Implements #8678 to also send client side metrics to Amplitude using the Amplitude client library. Note that this PR does not remove the old client -> server -> Segment route just yet. We follow the migration pattern of parallel writing then deletion of the old service.

The real meat of the change is in `packages/client/mutations/SendClientSideEvent.ts`, which is renamed from `packages/client/mutations/SendClientSegmentEventMutation.ts`

## Demo

`Your Work Drawer Impression` is a client-side metric and notice the library used is the Amplitude library, instead of Segment.

Screenshot 2023-10-19 at 16 25 20

## Testing scenarios

- [ ] Verify client side metric is emitted to Amplitude
  1.  Make sure you have the `AMPLITUDE_WRITE_KEY` in your `.env` (see [here](https://github.com/ParabolInc/action-devops/blob/a56919818da46a690cbfec98bf5cf09a6d61992f/environments/local#L1))
  2.  Start the server
  3.  Emit any client side metric (e.g., go to a stand up meeting page and play around with the Your Work Drawer)
  4.  Go to [Amplitude User Activity page](https://app.amplitude.com/analytics/parabol/activity)
  5.  Make sure you select Parabol Development on top left
  6.  Search your `userId`
  7.  On the User Look-Up Page, verify a couple of client side metrics are in the Event Stream

## Final checklist


tghanken (approved):
Looks good to me.  Is the plan to eventually fully deprecate the old Client -> Server -> Analytics Tool route in favor of a direct Client -> Analytics route for client side events?

We might lose some visibility for a few users, but generally I think I'm in favor of doing so to simplify things.

tianrunhe:
> We might lose some visibility for a few users

I assume you meant people with ad blockers etc? I think that was actually the reason we introduced the client -> server -> analytics pattern in the first place, if my memory serves me well. Didn't your research find that that population is around 10% of our user base? I think that's tolerable.

tghanken:
> I assume you meant people with ad blockers etc? ... I think that's tolerable.

Exactly.  Since the number is low, I came to the same conclusion as you, it's worth simplifying our approach here.

packages/client/mutations/UpdatePokerScopeMutation.ts
```diff
@@ -294,7 +294,7 @@ const UpdatePokerScopeMutation: StandardMutation<TUpdatePokerScopeMutation, Hand
       const searchQuery = getSearchQueryFromMeeting(meeting, service)
       if (!searchQuery) return
       const {searchQueryString, searchQueryFilters} = searchQuery
-      SendClientSegmentEventMutation(atmosphere, 'Updated Poker Scope', {
+      SendClientSideEvent(atmosphere, 'Updated Poker Scope', {
```
nickoferrall, line 297: +1 what's your thinking around removing `Mutation` from the name here? It's still a mutation, and as all of the other mutation names end with `Mutation`, I think we should do that here too

nickoferrall (approved):

packages/client/mutations/SendClientSideEvent.ts
```diff
@@ -1,22 +1,27 @@
...
 
 export type SendClientSegmentEventOptions = SegmentEventTrackOptions
 
-const SendClientSegmentEventMutation = (
+const SendClientSideEvent = (
   atmosphere: Atmosphere,
   event: string,
   options?: SendClientSegmentEventOptions
 ) => {
   atmosphere.handleFetchPromise(getRequest(mutation).params, {event, options})
```
tianrunhe, line 20: @nickoferrall I'm planning to remove this part and also Line 7 to Line 11, i.e., the mutation, in the very next PR

Dschoordsch (approved):
Did not test

Dschoordsch (approved):
Smoke tested, works

PR #8993, opened 2023-10-20
feat(zoom-transcription): UI improvements

nickoferrall:
See @enriquesanchez's comment here with designs: https://parabol.slack.com/archives/C0161UFBM6W/p1697754002870749?thread_ts=1697454762.894489&cid=C0161UFBM6W

Figma [here](https://www.figma.com/file/okoRD7I7lZR08V4Zk7O7G6/random?type=design&node-id=1732-6422&mode=design&t=KNYUoHSHmlsjEiRd-4).

This PR:

- Updates the transcription thread
- Adds the subscription so that all users are updates when someone enters a Zoom link in the input

This PR does not:

- Update the discussion thread header underlining
- Fix the transcription UI. I want to change the way we store the transcription in db first, which I'll do in a separate PR.

Screenshot 2023-10-20 at 10 16 35
Screenshot 2023-10-20 at 10 18 06

### To test

- [ ] Run the `updateOrgFeatureFlag` mutation and add the `zoomTranscription`
- [ ] Create a Retro and see that the UI looks like the images above
- [ ] The UI updates for all meeting members when one meeting member enters the Zoom URL


packages/client/components/RetroDiscussPhase.tsx
```diff
@@ -262,12 +263,19 @@ const RetroDiscussPhase = (props: Props) => {
...
-                      isReadOnly={allowedThreadables.length === 0}
-                      meetingRef={meeting}
-                      showTranscription={showTranscription}
-                    />
+                    showTranscription ? (
+                      <DiscussionThreadListEmptyTranscriptState
+                        allowTasks={true}
+                        isReadOnly={allowedThreadables.length === 0}
+                        meetingRef={meeting}
+                        showTranscription={showTranscription}
```
nickoferrall, line 271: We no longer need to include `showTranscription`. I removed this in the next PR

enriquesanchez (changes requested):
This is looking good!

I wasn't able to test this PR because I don't know how to enable feature flags locally 🙈 , but looking at the screenshots you shared I have a couple of requests:

1. Can the description text be darker? Looks like it's set to `Slate 600` and the contrast is pretty low. Can we bump it up to `Slate 700`? The text field placeholder is fine at 600.
2. The button's label looks very small (maybe 12px?), is that the standard button component? There's also a gray border around the button which shouldn't be there.

nickoferrall:
>  I don't know how to enable feature flags locally

[Here's](https://www.notion.so/parabol/How-to-enable-or-remove-a-user-feature-flag-5c41c2075d9e4eca9b17aa2f4a79ace7) how you can add a feature flag if you'd like to! You'd go to `http://localhost:3000/admin/graphql`. If you can't access that page, then you need to add the super user role (`su`) to your user using pgadmin.

> Can the description text be darker? Looks like it's set to Slate 600 and the contrast is pretty low. Can we bump it up to Slate 700?

It's the same colour as the description text in the empty discussion thread, so we'd need to change the colour for both of them if we want to:

Screenshot 2023-10-23 at 11 23 50
Screenshot 2023-10-23 at 11 23 42


> The button's label looks very small (maybe 12px?), is that the standard button component? There's also a gray border around the button which shouldn't be there.

Thanks, I've removed the border and bumped the font size to the default size:

Screenshot 2023-10-23 at 11 27 44

Dschoordsch:
>If you can't access that page, then you need to add the super user role (`su`) to your user using pgadmin.

It's easier to just run
```
yarn node ./scripts/toolbox/assignSURole.js --add you@example.com
```
to give yourself `su` rights. Don't forget to log out and in again.

nickoferrall:
> @nickoferrall I'm getting this error while trying to enable the feature flag: "Value \"zoomTranscription\" does not exist in \"UserFlagEnum\" enum.", I'm not sure what I'm doing wrong.

This one is an org feature flag. It looks like you're trying to add a user feature flag. 

> Ah I see. Let's leave it as it is then. I'll revisit text contrast soon.

Sounds good 👍

packages/client/components/ZoomSVG.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
```
Dschoordsch, line 1: +1 Why do we need this in addition to `zoom-logo.svg`?

packages/client/components/DiscussionThreadListEmptyTranscriptState.tsx
```diff
@@ -0,0 +1,156 @@
...
+  }
+})
+
+const Wrapper = styled('div')({
+  alignItems: 'center',
+  display: 'flex',
+  justifyContent: 'center',
+  flexWrap: 'wrap',
+  flex: 1
+})
```
Dschoordsch, line 64: +1 I don't like that we're introducing more styled components when we're actively trying to get rid of it.

Dschoordsch (approved):

Issue #8994, opened 2023-10-20
Zoom Transcription: transcript is not one block of text


nickoferrall:
See @enriquesanchez's designs [here](https://www.figma.com/file/okoRD7I7lZR08V4Zk7O7G6/random?type=design&node-id=1732-6422&mode=design&t=KNYUoHSHmlsjEiRd-4).

![image](https://github.com/ParabolInc/parabol/assets/39854876/217692c1-f107-43e4-86f5-b89fbcd653d7)

### AC

- [ ] Rather than storing the transcript as a string, it should be an array of objects
- [ ] Each object should have the speaker's name and the text that was said
- [ ] The UI should then reflect the design above, with the speaker's name being bold & in italics

PR #8995, opened 2023-10-20
feat(zoom-transcription): split transcript into speaker & words

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8994

Screenshot 2023-10-20 at 16 11 27

This PR changes the transcription from a string to an array of objects with the speaker & words. I haven't added a migration because only a few transcriptions have been recorded and they're all internal, so it's OK to get some console errors when the transcription is returned as the wrong type.

### To test

- [ ] Transcription UI looks similar to the designs in the linked issue


enriquesanchez (changes requested):
I was looking at these mockups with fresh eyes this morning and realized that we should use the same styling we are using for comments. [I've updated the mockup](https://www.figma.com/file/okoRD7I7lZR08V4Zk7O7G6/random?type=design&node-id=1733-9499&mode=design&t=lE13HSnghGd2Brmp-4) for reference, but basically the only change is on the text styles of the users names.

Another thing I'd like to see but don't know if it's possible is to show the user's name only once when there are several comments one after the other by that same person. Right now you see this:

```
Nick O'Ferral
This is a comment
Nick O'Ferral
This is another comment
```

...and ideally it'll look more like this:

```
Nick O'Ferral
This is a comment
This is another comment
```

nickoferrall:
> [I've updated the mockup](https://www.figma.com/file/okoRD7I7lZR08V4Zk7O7G6/random?type=design&node-id=1733-9499&mode=design&t=lE13HSnghGd2Brmp-4) for reference, but basically the only change is on the text styles of the users names

Sounds good, I've updated the UI.


I've also merged the text from a solo speaker. It's not in separate lines, and it looks like a long block because I was recording a Loom, but it would rarely be that long in a normal meeting:

Screenshot 2023-10-23 at 12 08 27

Loom demo here: https://www.loom.com/share/faaabe872a334fdc869590fe23c199fc

enriquesanchez:
@nickoferrall I was wondering, should we instead say something like "Your Zoom transcription will begin once the 'Parabol Notetaker' Zoom bot joins the call, and the content will be available once the meeting has ended."? 🤔 

It feels a little long, but I like that it clearly describes how/when the transcription begins.

nickoferrall:
> I was wondering, should we instead say something like "Your Zoom transcription will begin once the 'Parabol Notetaker' Zoom bot joins the call, and the content will be available once the meeting has ended."?

Yeah, I think mentioning that it happens when the Parabol Notetake joins could be helpful. Perhaps we'd specify the transcription will be available rather than content? So:

"Your Zoom transcription will begin once the 'Parabol Notetaker' bot joins the call, and the transcription will be available once the meeting has ended." 

The downside is it's longer, as you said, and this says `transcription` twice.

enriquesanchez:
> The downside is it's longer, as you said, and this says `transcription` twice.

@nickoferrall So how about:

"Your Zoom transcription will begin once the 'Parabol Notetaker' bot joins the call, and it will be available once the meeting has ended."

???

Dschoordsch (approved):
Tested, seems to work.

Dschoordsch:
One thing I've noticed which might be confusing for a user: I've copied the invite link from Zoom and put it into the retro, but did not start the Zoom meeting immediately. When I finally came around to start it, the bot would not join and I couldn't re-invite it because the meeting already had a Zoom URL.

johnfelipe:
How can test it?
This need pay subscription or something
I want test on promise...

PR #9001, opened 2023-10-21
fix: server health check comment out

mattkrick:
# Description

hotfixes to fix k8s


Dschoordsch (approved):

Dschoordsch:
Setting the token null fixes the immediate problem, but we still show the client an error. Once they reload they can log in again, so it is an improvement, but it's not enough to do a secret update

mattkrick:
true! the right way to do it is to watch for 401 coming from the websocket/sse request. i'll have to PR trebuchet to do that

PR #9042, opened 2023-10-24
feat(local-llm): first pass

mattkrick:
# Description

Adds a local LLM in a webworker to perform semantic search without needing a server.


mattkrick:
we'll definitely have a local llm in the future, but it's really only good for mediocre results, locally

jordanh:
Closing for now

Issue #8713, opened 2023-08-22
Templates: Fix template illustration URLs


jmtaber129:
All custom retros + pre-/post-mortem templates' illustration URLs have been overwritten to have the `gladSadMad` retro template illustration:

![Image](https://user-images.githubusercontent.com/9013217/262403202-54aaf51f-f542-49f4-b090-5e1b1744bc09.png)
![Image](https://user-images.githubusercontent.com/9013217/262403201-84a98aab-1c0f-4ea3-a2d6-7d556086d769.png)

Fix this such that:
* Pre-mortems have the pre-mortem illustration
* Post-mortems have the post-mortem illustration
* Non-copy custom retros have the custom retro illustration
* Copied custom retros have their parent's illustration
* Public retros have the illustration specific to that retro template

PR #9043, opened 2023-10-25
fix(templates): Fix template illustrations via code-backed mapping

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/8713

Fix template illustrations by moving the full mapping from the DB into code, and leaving the `illustrationUrl` column as an override

## Demo
With this PR:
Screen Shot 2023-10-24 at 6 01 02 PM
Screen Shot 2023-10-24 at 6 01 07 PM

On prod today:
Screen Shot 2023-10-24 at 6 01 22 PM
Screen Shot 2023-10-24 at 6 01 28 PM

Copy of a template:
Screen Shot 2023-10-24 at 6 03 08 PM

Copy of the copy:
Screen Shot 2023-10-24 at 6 03 13 PM

## Testing scenarios
- [ ] All custom templates (including copies immediately after migration) have the custom template illustration
- [ ] Pre-/post-mortems have the correct illustrations
- [ ] Copying a template gives the copy the same illustration (and for copies-of-copies)
- [ ] Running the `down` migration and switching to master yields reasonable template image results.

## Final checklist


packages/server/graphql/public/types/helpers/getIllustrationUrlForActivity.ts
```diff
@@ -0,0 +1,122 @@
+import getTemplateIllustrationUrl from '../../../mutations/helpers/getTemplateIllustrationUrl'
+
+const retroIllustrations: Record<string, string> = {
```
mattkrick, line 3: Before figuring out the solution here, I wanted to see what the root cause was for #8713.
It looks like I wrote a migration that overwrote `illustrationUrl` for the pre/post mortem templates.
Given that knowledge, my gut reaction is to write a migration that fixes the values in the DB. That said, I know it can be frustrating to consider that data could be mutated at any time, sometimes incorrectly, and it could just be nice to have something a little more visible.

Your approach is very different! If I understand it correctly, we're nulling out all illustrationUrls in the DB & moving that data from the DB into a code file. As you explain it, this is easier to reason about because we can see the whole table at any given time. I'd like to learn more about this approach! It's nice seeing everything without having to run a query, but I'm having a hard time justifying in my head why this one column should live in code while the other columns live in the DB. For example, why not also have a lookup table for `category`? What makes the illustrationUrl special so it lives outside the DB? Now with this field living in 2 places, how might we prevent devs from forgetting to also add an entry to this table when they write a migration to add a new template?

jmtaber129: > why this one column should live in code while the other columns live in the DB. For example, why not also have a lookup table for category? What makes the illustrationUrl special so it lives outside the DB?

My thinking here is that this field is effectively static for any given template, so it makes sense to keep this as a static mapping that's easy to read and easy to change. Contrast this with `category` or any other template field that's mutable by the user, and couldn't easily be statically mapped.

If/when we allow users to mutate illustrations for templates they own, it might make sense to move this back into the DB, but I don't see us prioritizing that anytime soon.

> Now with this field living in 2 places, how might we prevent devs from forgetting to also add an entry to this table when they write a migration to add a new template?

This was the status quo for a while (forever?) until we moved this field to the DB. The main potential problem I see is that devs might add the illustration to the migration _instead of_ this table. I don't see a solution that would enforce this in code (unless we remove the field from the DB entirely, which would give all copied templates the custom template illustration), so I would announce the change in #engineering, and we would just need to catch this in CR if it comes up.

PR #9049, opened 2023-10-25
feat(isEnterprise): set org as enterprise on new user or org

adaniels-parabol:
# Description

Sets the default org tier to Enterprise if env var IS_ENTERPRISE=true. This is so that all orgs on an instance will be enterprise. This is for our onprem customers that are enterprise.

Also added script that can be ran for existing deployments.

### DIFF
--- a/scripts/toolboxSrc/setIsEnterprise.ts
+++ b/scripts/toolboxSrc/setIsEnterprise.ts
@@ -1,21 +1,18 @@
 import getRethink from '../../packages/server/database/rethinkDriver'
 import getPg from '../../packages/server/postgres/getPg'
+import {defaultTier} from '../../packages/server/utils/defaultTier'
 
 export default async function setIsEnterprise() {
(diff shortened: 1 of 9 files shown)


scripts/toolboxSrc/setIsEnterprise.ts
```diff
@@ -0,0 +1,47 @@
+import getRethink from '../../packages/server/database/rethinkDriver'
+import getPg from '../../packages/server/postgres/getPg'
+
+export default async function setIsEnterprise() {
+  const r = await getRethink()
+
+  console.log('Updating tier to "enterprise" for Organization and OrganizationUser tables in RethinkDB')
```
mattkrick, line 7: +1 before going any further, we should check if `process.env.IS_ENTERPRISE === 'true'`. that way we're sure they've done everything they need to do to achieve success.

adaniels-parabol: I added this check @mattkrick

PR #9050, opened 2023-10-25
chore: build bumps version in gitlab

mattkrick:
# Description

When a release-please PR is merged to `release`, the build.yml will push a docker image to our registry.
Now, it will also add a commit message to gitlab that increments the version, which kicks off the pipeline diff.
This removes 1 more step from the release process. Now all that remains is a dev going to the pipeline & clicking Play.

## Demo

Tested without the IS_RELEASE conditional here: https://github.com/ParabolInc/parabol/actions/runs/6644516359/job/18053839586

You can see that it worked on the staging repo here: https://gitlab.com/parabol1/saas/staging/parabol/-/blob/main/version.yaml?ref_type=heads

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -66,7 +66,7 @@ jobs:
 
       - name: Bump version in GitLab
         run: |
-          curl  "https://gitlab.com/api/v4/projects/${{ env.GITLAB_PROJECT_ID }}/repository/files/version%2Eyaml" \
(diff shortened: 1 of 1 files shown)


rafaelromcar-parabol:
LGTM. Need changes on the Release Playbook thou, to be sure the devs check on if there are any config var changes to do

rafaelromcar-parabol (approved):

PR #9055, opened 2023-10-26
chore: don't lock org in case of failed payment

### LINKED ISSUE #8341: Billing: Don't lock the org immediately if payment fails
Currently, if payment fails we lock the org immediately.

I suggest:

- If payment fails, send `Card rejected` notification and give `30 days` for completing the payment.
- Show persistent UI warning everywhere in the app.
- If it is the first payment that fails (subscription is `incomplete`), do not lock the org at all, downgrade org back to the `team tier`, send a notification.
- After 30 days lock the org, cancel the subscription to stop generating new invoices

We can use lock org mechanism from teams limit for that.

- Is 30 days too long?
- If they decide to resume their subscription later, should we charge them for 30 days additionally?

igorlesnenko:
fixes #8341

**How to test:**

- Run: `stripe listen --forward-to localhost:3000/stripe` (Do `stripe login` if you see an error)
- Upgrade starter org to team plan using "Decline after attaching" test card number: 4000000000000341
- Wait few seconds, refresh page. See that org is not locked
- See that it i upgraded to team tier
- See that card rejected notification is added in the notification center

### DIFF
--- a/packages/server/billing/helpers/terminateSubscription.ts
+++ b/packages/server/billing/helpers/terminateSubscription.ts
@@ -1,6 +1,5 @@
 import getRethink from '../../database/rethinkDriver'
 import Organization from '../../database/types/Organization'
-import updateTeamByOrgId from '../../postgres/queries/updateTeamByOrgId'
 import {getStripeManager} from '../../utils/stripe'
 
 const terminateSubscription = async (orgId: string) => {
@@ -21,8 +20,7 @@ const terminateSubscription = async (orgId: string) => {
           {returnChanges: true}
         )('changes')(0)('old_val')
         .default(null) as unknown as Organization
-    }).run(),
-    updateTeamByOrgId({isPaid: false}, orgId)
+    }).run()
   ])
   const {organization} = rethinkResult
   const {stripeSubscriptionId} = organization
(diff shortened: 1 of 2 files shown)


nickoferrall (approved):
In the new checkout flow, it doesn't upgrade to team with that card number because of the way we're handling 3D secure cards, but it is no longer locking the org after a failed payment

igorlesnenko:
> In the new checkout flow, it doesn't upgrade to team with that card number because of the way we're handling 3D secure cards, but it is no longer locking the org after a failed payment

Ah that is fine too I think!

packages/server/billing/helpers/terminateSubscription.ts
Dschoordsch, line 1: -1 As I understand it, we shouldn't delete the subscription if a payment fails, but we still do on l31. I think also the name `terminateSubscription` could be updated to indicate it's not terminating anymore.

Dschoordsch (changes requested):
Did not test yet.

packages/server/billing/helpers/terminateSubscription.ts
igorlesnenko: @Dschoordsch We still use this function and terminate subscription only in a single case.

If the first payment fails and subscription is in incomplete status.

> After 23 hours subscription updates to incomplete_expired and the invoice becomes void.
> Not to handle this particular case in 23 hours, we do it now

https://github.com/ParabolInc/parabol/pull/9055/files#diff-71929e8c8695ec1c675f137d9b8806bbe9f2b941921f34637902888e0e9c18a4R74-R76

Dschoordsch (approved):
Works

tghanken (approved):

PR #9057, opened 2023-10-26
feat: public teams

nickoferrall:
This PR implements a first version of [public teams](https://www.notion.so/parabol/Public-teams-Paid-Private-Teams-9a48756ff66b47b2a8eca3970e89a43f).

Users with the feature flag:
- Can see all of the teams in the org
- Can click on the team and see a "Request to Join" UI
- Click a CTA button to join the team

See demo: https://www.loom.com/share/00cae6305c93462d90d32b18601688fd

Screenshot 2023-10-26 at 15 14 09
Screenshot 2023-10-26 at 15 14 15

Screenshot 2023-10-26 at 15 14 21

This PR does not include private teams. If any users are unhappy with public teams, we can remove the feature flag.

A concern I have is that orgs with _many_ teams, like Parabol, may find that the sidebar becomes too crowded with all of the public teams. I think this is better than having a scroll bar within a scroll bar - see Loom example [here](https://www.loom.com/share/1288540c4d6548dabfd23e32a91d0551). 

### To test

- [ ] Without the feature flag, see that the UI is the same as production
- [ ] Add the `publicTeams` user feature flag
- [ ] In the sidebar, you can now see teams that you don't belong to
- [ ] The teams you don't belong to have a different UI, as indicated above
- [ ] Click the "Request to Join" CTA button and see that the invitation request is sent to users that do belong to that team


acressall:
Left a couple comments on the Loom! I'd like to run this through design maintainer review as well since there's a navigational change. UI is fine, just the UX questions asked in the Loom to sort out

packages/client/components/DashNavList/DashNavList.tsx
```diff
@@ -41,27 +40,47 @@ const DashHR = styled('div')({
...
+        }
       }
     `,
-    viewerRef
+    organizationsRef
+  )
+  const hasPublicTeamsFlag =
+    organizations?.[0]?.viewerOrganizationUser?.user?.featureFlags?.publicTeams
+  const teams = organizations?.flatMap((org) =>
+    hasPublicTeamsFlag ? org.allTeams : org.viewerTeams
```
igorlesnenko, line 82: If I randomly add this flag to some user, and for example this user has 3 organizations. Now he can see the entire list of teams in those three organizations and send them requests to join without permission from those organizations. 

Moreover, he will also automatically have access to the team list in all future organizations where he can be added afterwards. That is, even if we only select users with one particular organization, he can join new ones after adding the feature flag and we have no control over it. 

Would such requests surprise the organizations and raise security concerns? Should we use the org level feature flag and notify organizations allowing them opt-out?

packages/client/components/RequestToJoin.tsx
```diff
@@ -0,0 +1,68 @@
...
+const RequestToJoinComponent = () => {
+  const atmosphere = useAtmosphere()
+  const match = useRouteMatch<{teamId: string}>('/team/:teamId')
+  const teamId = match?.params.teamId
+  const [isRequestSent, setIsRequestSent] = useState(false)
+  const {onError, onCompleted, error} = useMutationProps()
+
+  const handleRequestJoin = () => {
+    if (!teamId) return
+    SendClientSegmentEventMutation(atmosphere, 'Sent request to join from public teams', {
```
igorlesnenko, line 20: we removed this in master, now there is `SendClientSideEvent`

igorlesnenko (commented):
Not tested yet, have few questions:

packages/client/components/DashNavList/DashNavList.tsx
nickoferrall: Good point - I was thinking it would be inconsistent to see a different UI for some orgs vs others, but I agree that if an org dislikes public teams, it would be easier to to disable an org feature flag

packages/client/components/DashNavList/DashNavList.tsx
```diff
@@ -41,27 +40,47 @@ const DashHR = styled('div')({
...
+            }
+          }
+        }
       }
     `,
-    viewerRef
+    organizationsRef
+  )
+  const hasPublicTeamsFlag =
+    organizations?.[0]?.viewerOrganizationUser?.user?.featureFlags?.publicTeams
```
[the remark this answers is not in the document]
igorlesnenko: In this case the question is what is the reason to explicitly allow billing leader to see all teams in API while we not allowing to see them on frontend anyway? Should we remove this condition?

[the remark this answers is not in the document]
nickoferrall: In the teams sections of the org settings, billing leaders can see all teams within the org: https://action.parabol.co/me/organizations/Skc7syWwNox/teams

This was a feature requested by the sales team a few months ago.

packages/client/mutations/AddTeamMutation.ts
```diff
@@ -19,6 +19,7 @@ graphql`
     team {
       id
       name
+      isViewerOnTeam
```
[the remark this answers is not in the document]
igorlesnenko: Yes, I would ideally add something like this, because otherwise it is not clear why this particular field exists and what will happen if I remove it

igorlesnenko (approved):
Works!

One small bug that could be done in this or next PR:
If I click on a team and click request to join, and then switch to another team to request that too, it still shows that request already sent

image

nickoferrall:
> If I click on a team and click request to join, and then switch to another team to request that too, it still shows that request already sent

Thanks - fixed!

packages/client/modules/teamDashboard/components/EditTeamName/EditableTeamName.tsx
```diff
@@ -29,7 +29,7 @@ const EditableTeamName = (props: Props) => {
         teamId: id
         teamName: name
         organization {
-          teams {
+          viewerTeams {
```
Dschoordsch, line 32: -1 This is used to check for duplicate names in l59, so it should be `allTeams` here.

packages/client/modules/newTeam/components/NewTeamForm/NewTeamForm.tsx
```diff
@@ -117,7 +117,7 @@ const NewTeamForm = (props: Props) => {
         id
         lockedAt
         name
-        teams {
+        viewerTeams {
```
Dschoordsch, line 120: -1 these teams are used for checking for duplicate names in l159

Dschoordsch (approved):
When I request to join a team and the team lead accepts the request, I would like if I would join directly and wouldn't need to accept the invite.

Dschoordsch (changes requested):
There are a couple minor things

nickoferrall:
> When I request to join a team and the team lead accepts the request, I would like if I would join directly and wouldn't need to accept the invite.

Agreed! I'll fix that in another issue: https://github.com/ParabolInc/parabol/issues/9108

Dschoordsch (approved):

acressall:
@nickoferrall this went through design review. Designers requested we change the icon to be the same for all and just use the difference in shade to differentiate between the tow.

They also requested that the "invite request" screen would stay for [x] days before reverting to the "request to join" version, to avoid spamming with requests. I suggested this improvement as a fast follow so it doesn't block this PR.

nickoferrall:
Sounds good @acressall. I've updated the icon and created an issue for the request screen: https://github.com/ParabolInc/parabol/issues/9117

PR #9070, opened 2023-10-27
chore(clean-up): Cleans up all the files used for the old infrastructure and pipelines

rafaelromcar-parabol:
# Description
Removing everything that was used only for the old infrastructure and old pipelines


mattkrick:
Also removed CircleCI from our GH org. I think we're clean 🎉

PR #9071, opened 2023-10-27
chore: update sign up redirect destination

### LINKED ISSUE #8941: Update sign up destination
Design #8639 

AC

- On signup: user lands on team dashboard with "manage team" panel open
- On subsequent signup: user lands on `/meetings` view

igorlesnenko:
Fixes #8941

**How to test:**

- Sign up with password
- See that in 50% cases you redirected to the /team/{ID} page instead of /meetings
- See that login to existing account always redirects to /meetings
- Test the same with sing up with google


packages/client/mutations/handlers/handleAuthenticationRedirect.ts
```diff
@@ -5,20 +5,24 @@ import SendClientSideEvent from '../SendClientSideEvent'
...
+) => {
   SendClientSideEvent(atmosphere, 'User Login')
   const redirectTo = getValidRedirectParam()
   if (redirectTo) {
     history.push(redirectTo)
     return
   }
   if (!acceptTeamInvitation?.team) {
-    history.push('/meetings')
+    history.push(defaultRedirectPath)
```
nickoferrall, line 25: +1 I'd just call this `redirectPath` because the default path is `/meetings`, but `/team/:teamId` isn't the default

nickoferrall (approved):

Dschoordsch (approved):
Does not work with email verification.
I would like to see the redirect to be permanent for the user, at least as long as they have only 1 team

igorlesnenko:
@acressall deploying this will cause the feature to be rollout instantly. Is it ok?

Regarding @Dschoordsch comments:

> Does not work with email verification.
> I would like to see the redirect to be permanent for the user, at least as long as they have only 1 team

How important to have this? Can we test without this?

acressall:
> I would like to see the redirect to be permanent for the user, at least as long as they have only 1 team

I think this is nice to have, but not required

acressall (approved):

igorlesnenko:
> Does not work with email verification.

I fixed it in this PR, otherwise it will give us inaccurate results as we add a feature flag anyway but don't redirect

PR #9072, opened 2023-10-27
feat: adds MoSCoW and RICE prioritization templates

ackernaut:
# Description

Adds MoSCoW and RICE prioritization templates using the poker meeting and scales

## Testing scenarios

- Run a test meeting using each new template (under the strategy category)
- Add Parabol tasks to estimate. Estimate the items. Add a final score per dimension
- End the meeting. See that the CSV has the items with their estimates for each dimension of the template

### DIFF
--- a/packages/server/postgres/migrations/1698263155574_addPrioritizationTemplates.ts
+++ b/packages/server/postgres/migrations/1698263155574_addPrioritizationTemplates.ts
@@ -58,7 +58,9 @@ const MOSCOW_SCALE_CONFIG: TemplateScale = {
     {label: 'M', color: PALETTE.TOMATO_500},
     {label: 'S', color: PALETTE.TERRA_300},
     {label: 'C', color: PALETTE.GOLD_300},
-    {label: 'W', color: PALETTE.GRASS_300}
+    {label: 'W', color: PALETTE.GRASS_300},
+    {label: '?', color: PALETTE.ROSE_500},
+    {label: 'Pass', color: PALETTE.GRAPE_500}
   ]
 }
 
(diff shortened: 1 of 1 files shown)


packages/server/postgres/migrations/1698263155574_addPrioritizationTemplates.ts
```diff
@@ -0,0 +1,212 @@
...
+}
+
+const NEW_TEMPLATE_CONFIGS: Template[] = [
+  {
+    id: 'moscowPrioritizationTemplate',
+    name: 'MoSCoW Prioritization',
+    dimensions: [
+      {
+        id: 'moscowPriorityDimension',
+        name: 'Priority? Must/Should/Could/Won’t Have',
```
jmtaber129, line 74: +1 Most dimension names generally seem to be a simpler label than this. I'd suggest renaming this to just "Priority", since the "MoSCoW" scale label is shown right next to the dimension label in the product:
Screen Shot 2023-10-31 at 1 36 02 PM

jmtaber129 (approved):
LGTM, but with one minor suggestion to simplify the dimension label in the MoSCoW template.

packages/server/postgres/migrations/1698263155574_addPrioritizationTemplates.ts
ackernaut: I agree that I’m stretching the use of this label. However there’s not a place in the template where I can specify what MoSCoW means. I assume most of the time folks reaching for this are familiar with it. However participants may not be as familiar and have to remember ‘what is S again’? Also, we have a good number of folks whose first language isn’t English. This might help there, erring on the side of clarity over brevity. In the future, I imagine we expand on Poker, Dimensions, Scales etc. to support more interaction types and better views into labels, values, prompts, etc. and I wouldn’t abuse this label. Maybe even tooltips with a label e.g. ‘Must Have’ when hovering the cards would help in a case like this. In the meantime it fits in the view, even mobile is okay. Thanks for reviewing!
Screenshot 2023-11-01 at 11 54 14

PR #9080, opened 2023-10-31
fix: gracefully handle reconnects when servers upgrade

### LINKED ISSUE #8903: Shutdown: kick off users from websocket server in batches
When receiving `SIGTERM` we need to properly disconnect all users from the websocket server. We should do it in batches to not overwhelm the replacement with new connections. The default grace period is 30s but should be configurable from the environment

## Acceptance criteria
* ​add a sigterm handler
  * stop accepting new connections
  * cuts websocket connections in batches based on configured grace period (like for 30s grace period kick off 1/25th every 1s)

mattkrick:
# Description

fixes #8903 

## Demo

https://www.loom.com/share/c891b7ca7e7346ba9f6dfbfa8c138157?sid=445095fa-b1a4-4890-b091-3aa1ae7211fa

## Testing scenarios

- [ ] in handleOpen.ts, add the line `console.log('connected', process.pid)` to get the process id of the connected user
- [ ] open a browser & go to the app
- [ ] in a terminal, call `kill -SIGTERM <pid>` where <pid> is what got logged to the console
- [ ] watch the browser reconnect (it'll probably do it twice since it reconnected to the same single server since we have no load balancer locally)

### DIFF
--- a/packages/server/activeClients.ts
+++ b/packages/server/activeClients.ts
@@ -1,6 +1,6 @@
 import ConnectionContext from './socketHelpers/ConnectionContext'
 
-class SSEClients {
+class ActiveClients {
   store = {} as Record<string, ConnectionContext>
   get(connectionId: unknown) {
     return this.store[String(connectionId)]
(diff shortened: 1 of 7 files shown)


nickoferrall (approved):
Tested, and it worked!

PR #9082, opened 2023-10-31
fix: do not fail installing service worker on fetch fail

Dschoordsch:
If we fail to fetch a file for the static cache, we previously failed to install the new service worker. Now we're just catching this error. If the file is missing from the static cache, it will be fetched from the dynamic one later.

## Demo

## Testing scenarios

- run production `yarn build && yarn predeploy && yarn start`
- add a failing file, like 'https://parabol.fun/invalid' to the cache and see installation not failing

## Final checklist

### DIFF
--- a/packages/client/serviceWorker/sw.ts
+++ b/packages/client/serviceWorker/sw.ts
@@ -27,11 +27,12 @@ const onInstall = async (_event: ExtendableEvent) => {
   const cacheNames = await caches.keys()
   const oldStaticCacheName = cacheNames.find((cacheName) => cacheName.startsWith('parabol-static'))
   const newCache = await caches.open(STATIC_CACHE)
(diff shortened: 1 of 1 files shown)


packages/client/serviceWorker/sw.ts
```diff
@@ -31,7 +31,7 @@ const onInstall = async (_event: ExtendableEvent) => {
   // if this is their first service worker, fetch it all
   if (!oldStaticCacheName) {
     console.log('Installing service worker')
-    return newCache.addAll(urls)
+    return newCache.addAll(urls).catch(console.error)
```
Dschoordsch, line 34: `fetchAll` fetches all URLs, even if some fail, but rejects the final promise in that case, see https://w3c.github.io/ServiceWorker/#cache-addAll

igorlesnenko (approved):

Issue #9087, opened 2023-10-31
Implement switching templates during a retro UI


nickoferrall:
See design issue here: https://github.com/ParabolInc/parabol/issues/9047

Figma designs here: https://www.figma.com/file/3mCDNSuJeByj3jiMeGm7yb/Change-retro-template-within-meeting?type=design&node-id=1-1006&mode=design&t=vXi8NP08RgHAYSX4-0

### AC

- [ ] Show options button in retro meeting, just like Standups
- [ ] A dropdown menu opens with "Change template"
- [ ] If a reflection has been added to the meeting, "Change template" is disabled. A tooltip pops-up when hovering over it explaining that you can only change the template before a reflection has been added
- [ ] Clicking "Change template" opens a right drawer with the templates
- [ ] Clicking a template does nothing. The logic will be handled in a subsequent issue.
- [ ] The options button only shows up in the reflect phase as you can't change the template in other phases. This can be changed when we add recurring retros.

jordanh:
Going to icebox this for now

Dschoordsch:
I un-iceboxed the issue. We figured this might be an important part of recurring retros, so the user still has the option to change the template. It should be mostly done, just needs merging of the PR.

PR #9093, opened 2023-11-01
feat: switch template UI

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9087

Demo: https://www.loom.com/share/758f0aef94a943dc9ec28009125b325c

This PR won't be merged until the functionality is added in a subsequent PR.

I'm using the "Options" button from Standups. One small UI thing is the avatar list gets larger and smaller depending on the screen size whereas "options" remains the same height. As that affects Standups, I'll tackle it in a separate PR. 

I created a reusable menu & menu item component in this PR:
Screenshot 2023-11-01 at 18 51 25

### To test

- [ ] Create a retro, go to the "Reflect" phase and see "options"
- [ ] Clicking on the "Change template" menu item opens the right drawer with all the templates
- [ ] This doesn't show up in the group phase


packages/server/graphql/public/types/User.ts
```diff
@@ -174,6 +174,7 @@ const User: UserResolvers = {
         ...activity,
         sortOrder: getScore(activity, teamIds)
       }))
+      .filter((activity) => !type || activity.type === type)
```
nickoferrall, line 177: I could add `type` to the dataloader (or create a second dataloader) so that we only get retrospective templates. But as this query is used in the activity library, which will probably be executed for the org before the user gets to the retro drawer, it should be more efficient to keep it as it is and get the templates from the cache.

igorlesnenko (changes requested):
If we are introducing new radix components here, I would ideally make them similar to our previous radix components:

acressall (approved):

packages/client/components/MeetingOptions.tsx
```diff
@@ -0,0 +1,59 @@
...
+        <MenuItem
+          value='template'
+          label='Change template'
+          onClick={handleClick}
+          isDisabled={hasReflections}
+          icon={<SwapHorizIcon />}
+        />
+      </div>
+      {tooltipPortal('You can only change the template before reflections have been added.')}
+    </Menu>
```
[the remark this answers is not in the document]
nickoferrall: I prefer including `trigger` as a prop because it's always required, and it means the dev just needs to import Menu & MenuItem. I've removed the icon as a prop though

packages/client/ui/Menu/MenuItem.tsx
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export const MenuItem: React.FC<MenuItemProps> = (props: MenuItemProps) => {
+  const {value, label, onClick, isDisabled, icon} = props
+
+  return (
+    <DropdownMenu.Item
+      className={`flex w-full items-center rounded-md px-4 py-3 text-sm text-slate-700 outline-none hover:bg-slate-100 focus:bg-slate-100 ${
+        isDisabled ? 'cursor-not-allowed ' : 'cursor-pointer'
+      }`}
```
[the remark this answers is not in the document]
nickoferrall: I've updated it to include `twMerge` and forward ref. I spent ~1 hour trying to get `forwardRadix` to work, but had problems passing the prop types down and getting the right ref type

packages/client/ui/Menu/Menu.tsx
```diff
@@ -0,0 +1,25 @@
...
+import React from 'react'
+import {twMerge} from 'tailwind-merge'
+
+interface MenuProps {
+  trigger: React.ReactNode
+  children: React.ReactNode
+}
+
+export const Menu = React.forwardRef<HTMLDivElement, MenuProps>(({trigger, children}, ref) => {
+  const contentClass = twMerge(
```
igorlesnenko, line 11: +1 Current usage of `twMerge` is redundant, we are using it to be able to pass our own `className`. 
Same for `MenuItem`.

https://github.com/ParabolInc/parabol/blob/27a9efbbfd680febbe8f06203dab8e53e5e1fb56/packages/client/ui/Dialog/DialogClose.tsx#L13-L15

We also passing down props to support all radix props

https://github.com/ParabolInc/parabol/blob/27a9efbbfd680febbe8f06203dab8e53e5e1fb56/packages/client/ui/Dialog/DialogClose.tsx#L18

We can leave it as is and tweak later if the need arises to not spend more time on this now

igorlesnenko (approved):
A note for the future: I noticed that I can add reflections while drawer is opened, so theoretically I can click on new template with already existing reflections

igorlesnenko:
We don't need a feature flag here?

packages/client/ui/Menu/Menu.tsx
nickoferrall: True - I've added the `className` now so twMerge is used 👍

nickoferrall:
> We don't need a feature flag here?

Nope, I suggested that we don't use a feature flag here and just roll it out to all users. I won't merge this PR until the logic is in place though.

packages/client/components/MeetingOptions.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <div onMouseEnter={handleMouseEnter} onMouseLeave={handleMouseLeave}>
+          <TooltipTrigger asChild>
+            <MenuItem onClick={handleClick} isDisabled={hasReflections}>
+              <div className='mr-3 flex text-slate-700'>{<SwapHorizIcon />}</div>
+              Change template
+            </MenuItem>
+          </TooltipTrigger>
+        </div>
+        <TooltipContent>
+          {'You can only change the template before reflections have been added.'}
```
Dschoordsch, line 55: +1 I would rephrase this slightly to "You can only change the template if no reflections have been added." because it also works if you remove them again.

Dschoordsch (approved):

PR #9097, opened 2023-11-01
chore: fix migration conflicts

mattkrick:
# Description

The most frequent reason for a failed release is a failed predeploy due to conflicting migration names. For example, an older PR gets merged after a newer PR was already merged & pushed to prod, invalidating the migration run order.

This PR stops that by making sure the PR doesn't include any migrations that predate the latest migration on master.

## DEMO

If you have no migrations in your PR, you get this
Screenshot 2023-11-01 at 5 21 35 PM

If you have 1 valid migration in your PR, you get this
Screenshot 2023-11-01 at 5 21 19 PM

If you have 1 migration and it predates the latest migration on master, you get this
Screenshot 2023-11-01 at 5 20 14 PM

If you have 2 migrations in your PR you should consolidate or break apart your PR, so you get this
Screenshot 2023-11-01 at 5 20 06 PM


.github/workflows/build.yml
```diff
@@ -31,9 +31,42 @@ jobs:
...
+          git checkout ${{ github.ref_name }}
+
+          if [ $OLD_MIGRATIONS_COUNT == $NEW_MIGRATIONS_COUNT ]; then
+            echo No new migrations found
+            exit 0
+          elif [[ 'OLD_MIGRATIONS_COUNT + 1' -lt 'NEW_MIGRATIONS_COUNT' ]]; then
+            # Necessary since we only check the newest migration, not each one
+            echo Please only PR 1 Migration at a time
+            exit 1
+          fi
```
Dschoordsch, line 60: +1 I'm not a fan of this restriction. If you know the number of migrations, you can do
```
FIRST_NEW_MIGRATION=$(ls packages/server/postgres/migrations | tail -n $NEW_MIGRATION_COUNT | head -n1)
```

Dschoordsch (approved):
This will only catch some of the cases. If I have 2 PRs with conflicting migrations, both will pass the build pre-merge, but might cause an issue post merge as they're not re-verified after each master change.
I think we could only reliably check this on master post-merge.

Dschoordsch (changes requested):
After thinking about it, I think I would prefer if we could just check master that it's in the correct order. I think something along the lines of this
```bash
#!/bin/bash

PREV_COMMIT=$(git rev-list HEAD | tail -n 1)
PREV_FILE=''
for F in packages/server/postgres/migrations/*.ts; do
  CURR_COMMIT=$( git log --diff-filter=A --pretty=format:"%H" -- "$F")
  git merge-base --is-ancestor $PREV_COMMIT $CURR_COMMIT > /dev/null
  if [ $? -ne 0 ]; then
    echo "$PREV_FILE was added after $F but has a smaller timestamp"
  fi
  PREV_COMMIT=$CURR_COMMIT
  PREV_FILE=$F
done
```
This needs some cleanup though, or we make sure to only consider all files after the last failure

mattkrick:
it'd be cool if a hotfix branch would fail by comparing to master, too. Chances of a hotfix including a migration are pretty high.

We should probably break this out into its own action. That way it could run when the PR gets opened & re-run whenever a commit to master gets pushed. I'll toy around a bit more!

mattkrick:
omg i don't think it's possible to trigger a workflow when the base branch gets updated https://github.com/orgs/community/discussions/64119

PR #9132, opened 2023-11-07
feat: Free trial mutations

jmtaber129:
# Description

Fixes https://github.com/ParabolInc/parabol/issues/9133

[Spec](https://www.notion.so/parabol/Solution-Spec-Trial-Mutations-ba8426be508d41b2a48a1595856a5675) 🔒 

Introduces two mutations, `startTrial` and `endTrial`, which set/reset a `trialStartDate` field on any object with a `tier` field (`Organization`, `Team`, `OrganizationUser`, and `User`) to activate a free trial for a particular org without overwriting `tier` information.

Frontend/GQL:
In the GQL layer, the `trialStartDate` field is used to override the `tier` field, such that any checks in the frontend for whether a user has access to a feature based on tier will treat the object as being on the Team plan. We also introduce an `isTrial` field that indicates that the `tier` is set via free trial, so that billing-related components (e.g. billing panel, component displaying tier, etc.), can show that a free trial is active.

Server:
In server-side mutations + resolvers, anywhere that we check whether `tier` is `starter`, we also check if `trialStartDate` is non-null.

Mutations:
```
mutation {
  startTrial(orgId:"nWW2freL5K") {
    ... on StartTrialSuccess {
      organization {
        trialStartDate
      }
    }
  }
}
mutation {
  endTrial(orgId:"nWW2freL5K") {
    ... on EndTrialSuccess {
      organization {
        trialStartDate
      }
    }
  }
}
```

## Demo
https://www.loom.com/share/e78ac40536d04d8fadd8e92ab97be813?sid=04b7efa6-3a0f-4f76-86f8-865dca95cffd

## Testing scenarios
- [ ] Smoke test feature access + billing-/plan-related pages for each of these scenarios:
  - [ ] Starter plan, never had a trial
  - [ ] Trial activated on starter plan
  - [ ] Trial ended on starter plan
  - [ ] Org upgrade during active trial, then later downgrades
- [ ] Error cases for mutations:
  - [ ] Start trial on paid org
  - [ ] Start trial on org with active trial
  - [ ] End trial on org without active trial

## Final checklist


BartoszJarocki:
@jmtaber129 I'd love to clarify a couple of things.

First, do you see any issues with setting `tier` to `team` when `startTrial` mutation runs? alongside updating `tier` we'd set the information about the trial (start date, whatever else is needed). this way, we wouldn't need to change any business logic related to `tier` handling, except for showing the information about trial and possibly upgrading. Any thoughts? 

Another clarifying question, why do we set the `trialStartDate` in a few places like Org, OrgUser, Team? why not just one place?

jmtaber129:
@BartoszJarocki 

> do you see any issues with setting `tier` to `team` when `startTrial` mutation runs

Nothing specific, but because `tier` is currently tightly coupled with billing, I'm wary about changing that state for a new experimental feature like this one - when starting+ending trials, we'd need to _correctly_ update the `tier` according to billing. By keeping `tier` as-is and adding `trialStartDate`, we can leave billing _state_ untouched, and then use existing state + new state to determine what "feature tier" an org is on.

Effectively, I'm optimizing for "what's easiest to reverse and least likely to cause problems", rather than "what is the ideal long-term state", since this is experimental.

> why do we set the `trialStartDate` in a few places like Org, OrgUser, Team? why not just one place?

This is the same as what we do with `tier` today, so I'm using the same pattern for consistency. Instead of storing `tier` just on the `Org`, we denormalize it onto several different objects that inherit in some way from `Org`, so we don't need to do a DB lookup of `Org` when we check the `tier` of an OrgUser or Team. The `trialStartDate` on the `Org` is still the source of truth.

tianrunhe (approved):
I did a quick test locally and it worked fine. I think the bot added me because of the migration script, which I reviewed and it LGTM. 
Asked a relevant question on Slack [here](https://parabol.slack.com/archives/C042HTVBH8D/p1699912817257269?thread_ts=1698170990.439099&cid=C042HTVBH8D). We might also want to programmatically add/remove feature flags from the `startTrial` and `endTrial` mutations but they are definitely worth a separate PR.
Signing off as-is from data perspective! Great work, looking forward to seeing it in action soon!

BartoszJarocki:
> @BartoszJarocki
> 
> > do you see any issues with setting `tier` to `team` when `startTrial` mutation runs
> 
> Nothing specific, but because `tier` is currently tightly coupled with billing, I'm wary about changing that state for a new experimental feature like this one - when starting+ending trials, we'd need to _correctly_ update the `tier` according to billing. By keeping `tier` as-is and adding `trialStartDate`, we can leave billing _state_ untouched, and then use existing state + new state to determine what "feature tier" an org is on.
> 
> Effectively, I'm optimizing for "what's easiest to reverse and least likely to cause problems", rather than "what is the ideal long-term state", since this is experimental.
> 
> > why do we set the `trialStartDate` in a few places like Org, OrgUser, Team? why not just one place?
> 
> This is the same as what we do with `tier` today, so I'm using the same pattern for consistency. Instead of storing `tier` just on the `Org`, we denormalize it onto several different objects that inherit in some way from `Org`, so we don't need to do a DB lookup of `Org` when we check the `tier` of an OrgUser or Team. The `trialStartDate` on the `Org` is still the source of truth.

got it, makes sense!

BartoszJarocki (approved):

packages/server/graphql/mutations/helpers/createNewMeetingPhases.ts
```diff
@@ -86,7 +87,7 @@ const createNewMeetingPhases = async (
   const facilitatorTeamMemberId = toTeamMemberId(teamId, facilitatorUserId)
   const asyncSideEffects = [] as Promise<any>[]
 
-  const {tier = 'starter'} = team ?? {}
+  const tier = team ? getFeatureTier(team) : 'starter'
```
Dschoordsch, line 90: +1 We should not create meetings without a team, so really l84 should be changed to
```ts
    dataLoader.get('teams').loadNonNull(teamId)
```

packages/server/graphql/private/mutations/endTrial.ts
```diff
@@ -0,0 +1,38 @@
...
+const endTrial: MutationResolvers['endTrial'] = async (_source, {orgId}, {dataLoader}) => {
+  const now = new Date()
+  const r = await getRethink()
+  const pg = getKysely()
+
+  const organization = await dataLoader.get('organizations').load(orgId)
+
+  // VALIDATION
+  if (!organization.trialStartDate) {
+    throw new Error('No trial active for org')
```
Dschoordsch, line 16: -1 You defined an error payload, so you should return an error here instead of throwing

packages/server/graphql/private/mutations/startTrial.ts
```diff
@@ -0,0 +1,46 @@
...
+
+const startTrial: MutationResolvers['startTrial'] = async (_source, {orgId}, {dataLoader}) => {
+  const r = await getRethink()
+  const pg = getKysely()
+  const now = new Date()
+  const organization = await dataLoader.get('organizations').load(orgId)
+
+  // VALIDATION
+  if (organization.tier !== 'starter') {
+    throw new Error('Cannot start trial for organization on paid tier')
```
Dschoordsch, line 17: -1 You've declared an `ErrorPayload` as a valid result, so we should not throw here

packages/server/graphql/private/mutations/startTrial.ts
```diff
@@ -0,0 +1,46 @@
...
+        lockedAt: null,
+        updatedAt: now
+      })
+    }).run(),
+    pg.updateTable('Team').set({trialStartDate: now}).where('orgId', '=', orgId).execute(),
+    removeTeamsLimitObjects(orgId, dataLoader)
+  ])
+  organization.trialStartDate = now
+
+  await Promise.all([setUserTierForOrgId(orgId), setTierForOrgUsers(orgId)])
```
Dschoordsch, line 39: +1 it's not obvious that `setUserTierForOrgId` and `setTierForOrgUsers` can run in parallel as one writes `OrganizationUser` and the other one reads it. A comment on top of `setUserTierForUserIds` that it does not read the `tier` from `OrganizationUser` for this very reason would be nice.

packages/server/utils/setUserTierForUserIds.ts
```diff
@@ -28,14 +28,33 @@ const setUserTierForUserIds = async (userIds: string[]) => {
       )
     }))
     .run()) as {id: string; tier: TierEnum}[]
-  const newUserTiers = userIds.map((userId) => {
+
+  const userTrials = (await r
```
Dschoordsch, line 32: -1 Let's not do a second roundtrip but include this in the `userTiers` query

Dschoordsch (changes requested):
Only did a smoke test

Dschoordsch:
@tianrunhe should we have some analytics events in starttrial and endtrial or is this irrelevant because we're reading the data from the db anyways?

packages/server/utils/setTierForOrgUsers.ts
```diff
@@ -18,7 +18,11 @@ const setTierForOrgUsers = async (orgId: string) => {
     .filter({removedAt: null})
     .update(
       {
-        tier: r.table('Organization').get(orgId).getField('tier') as unknown as TierEnum
+        tier: r.table('Organization').get(orgId).getField('tier') as unknown as TierEnum,
+        trialStartDate: r
+          .table('Organization')
+          .get(orgId)
+          .getField('trialStartDate') as unknown as Date | null
```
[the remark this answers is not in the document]
jmtaber129: Fixed

packages/server/graphql/public/types/Team.ts
```diff
@@ -35,7 +36,13 @@ const Team: TeamResolvers = {
     const teamMember = await dataLoader.get('teamMembers').load(teamMemberId)
     return teamMember
   },
-  isViewerOnTeam: async ({id: teamId}, _args, {authToken}) => isTeamMember(authToken, teamId)
+  isViewerOnTeam: async ({id: teamId}, _args, {authToken}) => isTeamMember(authToken, teamId),
+  tier: ({tier, trialStartDate}) => {
+    return getFeatureTier({tier, trialStartDate})
+  },
+  isTrial: ({tier, trialStartDate}) => {
+    return !!trialStartDate && tier === 'starter'
```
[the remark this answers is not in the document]
jmtaber129: Removed `isTrial` and added `billingTier`, but kept `tier` the same name, since `tier` is used in so many places that it would cause conflicts.

packages/server/graphql/private/mutations/upgradeToTeamTier.ts
```diff
@@ -76,16 +77,19 @@ const upgradeToTeamTier: MutationResolvers['upgradeToTeamTier'] = async (
...
         isPaid: true,
-        tier: 'team'
-      },
-      orgId
-    ),
+        tier: 'team',
+        trialStartDate: null
+      })
+      .where('orgId', '=', orgId)
+      .execute(),
```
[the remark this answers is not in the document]
jmtaber129: Added

packages/server/graphql/mutations/helpers/createNewMeetingPhases.ts
jmtaber129: Fixed

packages/server/graphql/private/mutations/endTrial.ts
jmtaber129: Fixed

packages/server/graphql/private/mutations/startTrial.ts
jmtaber129: Fixed

packages/server/graphql/private/mutations/startTrial.ts
```diff
@@ -0,0 +1,46 @@
...
+  const pg = getKysely()
+  const now = new Date()
+  const organization = await dataLoader.get('organizations').load(orgId)
+
+  // VALIDATION
+  if (organization.tier !== 'starter') {
+    throw new Error('Cannot start trial for organization on paid tier')
+  }
+  if (organization.trialStartDate) {
+    throw new Error(`Trial already started for org. Start date: ${organization.trialStartDate}`)
```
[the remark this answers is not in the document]
jmtaber129: Fixed

packages/server/graphql/private/mutations/startTrial.ts
jmtaber129: Added a comment.

packages/server/utils/setUserTierForUserIds.ts
jmtaber129: Fixed

tianrunhe:
> @tianrunhe should we have some analytics events in starttrial and endtrial or is this irrelevant because we're reading the data from the db anyways?

I think at this time, the intention is to ask our sales to run these mutations internally. Since it's not public (I think?), I have no tension adding metrics atm.

Dschoordsch (approved):
Could you please check what's wrong with the test for `draftEnterpriseInvoice`? It passes locally for me, so it might be a CI setup issue. Querying the `error` field in the test might give more insights.

It would be nice if the trial would show up on the `OrgPlansAndBillingHeading` similar to the team settings, so it would say something along the lines of: "You are on a Free Trial for the Team Tier".

PR #9160, opened 2023-11-09
chore: Users with noTemplateLimit flag can use paid templates

### LINKED ISSUE #8972: Template Limits v2: implement flag for public templates
This issue doesn't include removing the limitation for custom templates. 

### AC

- [ ] If a user has the `noTemplateLimits` flag, they have access to all public templates even if they're on the Starter tier
- [ ] Verify that this works in both the legacy modal and the activity library

Dschoordsch:
# Description

Fixes #8972 

## Demo

https://www.loom.com/share/b9d8fd64a68b4babbcbf13f703052c7c?sid=67d8350a-8811-42b0-af95-e82c34f264e9

## Testing scenarios

- with the `noTemplateLimit` flag applied to your user and a `starter` org
- start a premium retro meeting from the activity library
- repeat with the legacy dialog
- mark a poker template non-free in the database
- start a meeting with a premium poker template from the activity library
- repeat with the legacy dialog

## Final checklist

### DIFF
--- a/packages/server/graphql/queries/helpers/resolveSelectedTemplate.ts
+++ b/packages/server/graphql/queries/helpers/resolveSelectedTemplate.ts
@@ -2,22 +2,40 @@ import getRethink from '../../../database/rethinkDriver'
 import MeetingSettingsPoker from '../../../database/types/MeetingSettingsPoker'
 import MeetingSettingsRetrospective from '../../../database/types/MeetingSettingsRetrospective'
 import {GQLContext} from '../../graphql'
+import {getUserId} from '../../../utils/authorization'
+import isValid from '../../isValid'
 
 const resolveSelectedTemplate =
   (fallbackTemplateId: string) =>
   async (
     source: MeetingSettingsPoker | MeetingSettingsRetrospective,
     _args: unknown,
-    {dataLoader}: GQLContext
+    {authToken, dataLoader}: GQLContext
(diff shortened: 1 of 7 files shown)


packages/server/graphql/queries/helpers/resolveSelectedTemplate.ts
```diff
@@ -2,22 +2,40 @@ import getRethink from '../../../database/rethinkDriver'
...
+    if (template) {
+      if (
+        template.isFree ||
+        template.scope !== 'PUBLIC' ||
+        tier !== 'starter' ||
+        viewer.featureFlags.includes('noTemplateLimit')
+      ) {
+        return template
+      }
+      // if anyone on the team has the noTemplateLimit flag, they might have selected a non-starter template
```
igorlesnenko, line 32: maybe easier to have org level feature flag in this case

igorlesnenko (approved):

packages/server/graphql/queries/helpers/resolveSelectedTemplate.ts
Dschoordsch: Yes, that's true. I think we discussed this at some point, but I cannot remember the argument. I just went with the implementation issue. I'll merge it as is, but might rework this in a separate PR if we change our mind.

PR #9164, opened 2023-11-09
fix: increases integration icon visibility

ackernaut:
## Description

This is intended as a small UX improvement to increase visibility of the integration button on task cards.

## Testing scenarios

- [ ] the task card footer actions show while editing (no need to hover)
- [ ] the task card integration button icon is updated to add weight and represent integration icons even over the push action


enriquesanchez (approved):
Works as expected 👍

PR #9165, opened 2023-11-09
chore: keep release process in GitHub

### LINKED ISSUE #9120: Deployments UI (no GitLab needed)
Goals:
- Deploy in 5 mins
- Don't leave GitHub
- I can see what commit is live on staging/production

AC
- Before a branch can be merged to master, it must pass a `build` check, which runs code quality, tests, etc.
- When release-please gets merged to a branch (hotfix-* or master) it runs a release workflow that versions, dockerizes, and pushes the build to our registry.
  - When a release-please commit is made, it dockerizes the build & pushes it to a development registry. This way all the work is happening before a dev needs it
  - When a release-please commit is merged, it grabs the docker image, tags it with the version, and then copies it to the prod registry. It should also kick of the push to staging, if we want.
- When a commit is pushed to a branch named `staging` or `production`, a deploy script is called that grabs the version number from the package.json & sends that gitlab. it does no dockerizing, it just calls gitlab!

mattkrick:
# Description

fixes #9120

As an app developer, when I push a new version, I shouldn't know or care about the underlying infrastructure. This is how it was with dokku, CI pushed a commit & 60 seconds later that code was live in prod. This is also how it is for any commercial PaaS like Cloudflare pages or  NextJS.  They probably use K8S under the hood, but that layer is hidden from devs because all we're concerned about is application code. We should strive for this simplicity! 

After this PR is merged, the release process will be as follows:
- Merge the release candidate PR from release-please. This will trigger a job that pushes the new version to our staging server and open a new PR with the QA checklist
- Complete the QA checklist & merge the PR. This will push the new version to production

That's it 🎉 

PS, being a GitHub workflow, I don't actually know if this code works 😅 . I'll just have to merge it & fix any hiccups along the way.


.github/workflows/release-to-staging.yml
```diff
@@ -0,0 +1,87 @@
+name: Release to Staging
+on:
+  pull_request:
+    branches:
+      - master
+      - hotfix-*
```
BartoszJarocki, line 6: +1 should we keep the convention of `hotfix/*` here? also, now that I wrote this I'm not sure we have a convention but that's something I'm used to ;)

BartoszJarocki (approved):
this is fantastic. lets see if it works!

.github/workflows/release-to-staging.yml
mattkrick: Good eye! With this change, I'm proposing that we change the hotfix flow to ALWAYS go to staging. If the PR is time critical, you can still merge to prod without waiting for staging to complete. For everything else, it's an extra chance to check your migration or code fix without delay!

.github/workflows/release-to-prod.yml
```diff
@@ -0,0 +1,33 @@
+name: Release to Production
+on:
+  pull_request:
+    branches:
+      - production
```
rafaelromcar-parabol, line 5: Why using the production branch? IMHO using branches for environments in an Open Source repository isn't a good pattern. Our repository hosts the code of our application, and to know what's in production we can know by checking the latest released tag.

We have multiple instances of Parabol (PPMIs and both SaaS) and we might have more in the future. IMHO it doesn't scale. 

If we are going to do this, I would at least use a branch public-saas instead of production. At least it would be clearer.

rafaelromcar-parabol (commented):
I think this is a good idea at the end. I was planning on proposing to automate the staging-release step if the staging-diff worked just fine, but I prefer not to do that and use what you are proposing here, so we DevOps keep that staging-diff to check when doing infra changes. Same goes for production. Also, we keep the view of what was to be changed and we can see if something went wrong.

Questions:
- **How do you plan managing hotfixes?** Do we agree that production branch (or whatever name it ends having) should never be used to push code directly to it and try to release?
- **What's the plan for configuration changes?** IMHO they should for sure go through Gitlab, using a PR to request merging it into Gitlab main even BEFORE the release to Staging happens.

.github/workflows/release-to-prod.yml
mattkrick: oh man, i didn't even think about pushing to PPMIs from github! I say we keep PPMIs out of github, since devs shouldn't even have to mess with them. i don't mind the name, but we call it production everywhere else, so i figured we could keep the name here, too?

.github/workflows/release-to-staging.yml
```diff
@@ -0,0 +1,87 @@
...
+          JOBS=$(curl "https://gitlab.com/api/v4/projects/${{ vars.GITLAB_PROJECT_ID }}/pipelines/$PIPELINE_ID/jobs" \
+            --header 'PRIVATE-TOKEN: ${{ secrets.GITLAB_API_TOKEN }}')
+          JOB_ID=$(echo $JOBS | jq '.[] | select(.name == "${{ env.STAGING_JOB }}")' | jq .id)
+          curl "https://gitlab.com/api/v4/projects/${{ vars.GITLAB_PROJECT_ID }}/jobs/$JOB_ID/play" \
+            --request POST \
+            --header 'PRIVATE-TOKEN: ${{ secrets.GITLAB_API_TOKEN }}'
+          PROD_JOB_ID=$(echo $JOBS | jq '.[] | select(.name == "${{ env.PRODUCTION_JOB}}")' | jq .id)
+          echo "JOB_ID=${JOB_ID}" >> $GITHUB_ENV
+          echo "PROD_JOB_ID=${PROD_JOB_ID}" >> $GITHUB_ENV
+          sleep 90
```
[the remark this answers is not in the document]
mattkrick: i'm using 90s as a minimum here, not a maximum.
After 90 seconds is up, we start polling the staging release. By sleeping for 90s, we remove 30 fetches which are almost guaranteed to fail.

Then again, we don't pay for CPU cycle so i might as well just have it poll so the yaml is more clear! will rewrite

.github/workflows/release-to-prod.yml
rafaelromcar-parabol: Let's keep production then. If we need to change it in the future, we can change it easily.

PR #9166, opened 2023-11-10
chore: Handle migration conflicts

### LINKED ISSUE #9094: build.yml should check to see migration name does not preceed any migration on master
AC
- [ ] build.yml should include a step that looks at the files being merged. for each file being added in packages/server/postgres/migrations, grab the datestamp prefix & compare it to the max value in master. if the max value in master is greater, then throw an error because this migration will preceed the migration that exists in master

mattkrick:
# Description

replaces #9097
fixes #9094

## Demo

If there are no migrations, it won't run

If the migration is correct, it'll look like this
Screenshot 2023-11-09 at 4 59 42 PM

If the migration fails, the CI fails
Screenshot 2023-11-09 at 5 03 15 PM


Dschoordsch (approved):
Looks good

Issue #9112, opened 2023-11-03
Activity library v2: update activity card UI 


nickoferrall:
Figma designs: https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?type=design&node-id=1036-28426&mode=design&t=0hT1JdrobaDPvoKr-0

![image](https://github.com/ParabolInc/parabol/assets/39854876/12fc13b1-e6ee-445e-ab30-ed2b12a59056)

### AC

- [ ] Move template name below the activity card
- [ ] Add the meeting type below the template name
- [ ] Update the background colour on the activity cards

enriquesanchez:
@nickoferrall [I had some early iterations](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?type=design&node-id=1047-29440&mode=design&t=WQPsBnQhR2PWVZg2-4) where I removed the meeting type if you had navigated into the category tab, or more recent ones like [the Favorites view where there are category subsections](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?type=design&node-id=1588-40862&mode=design&t=WQPsBnQhR2PWVZg2-4).

I'm still on the fence about it, the type is a very important piece of info and I worry that despite the user navigating to a specific tab, it still won't be clear enough what they are looking at.

I think the safest thing to try is to keep it for category views, and we can test with the Favorites view if removing it has any negative effects.

How does that sound?

nickoferrall:
Sounds good to me! It'll be easy to implement if we keep the meeting type in all the views too 👍

nickoferrall:
Hey @enriquesanchez, would you mind sending me the background image for the activity card? It looks similar the Perlin swirling lines that Figma offers, but I can't see the image in the Assets folder. I'm probably looking in the wrong place.

Also, for the meeting types, I'm currently showing: retrospective, feedback, poker, and standups. Premortem and post-morten, for example, would count as retrospective. 

I'm mapping our internal `meetingTypeEnum` to these four categories:

```
  retrospective: 'retrospective',
  action: 'feedback',
  poker: 'estimation',
  teamPrompt: 'standup'
```

Let me know if you had something else in mind! 🙏

Issue #9116, opened 2023-11-03
Activity Library v2: add tooltip on card hover


nickoferrall:
Figma designs: https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?type=design&node-id=1583-40584&mode=design&t=0hT1JdrobaDPvoKr-0

![Screenshot 2023-11-03 at 13 20 56](https://github.com/ParabolInc/parabol/assets/39854876/950a8c5c-3bc2-4fb4-becc-1ef7fdceb23d)

### AC

- [ ] Hovering over an activity card opens a tooltip with the prompts and description

PR #9168, opened 2023-11-10
feat: update activity library card UI

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9112

Loom demo: https://www.loom.com/share/d0e7df8917ad47f99ee73c6269d9b5d4

This PR won't be merged until the tooltip PR is also added: https://github.com/ParabolInc/parabol/issues/9116

### To test

- [ ] See that the card UI looks like the figma designs in the linked issue
- [ ] The card sizes are responsive just like production


packages/client/components/ActivityLibrary/Categories.ts
```diff
@@ -35,3 +36,20 @@ export const CATEGORY_ID_TO_NAME: Record<CategoryID | typeof QUICK_START_CATEGOR
...
+  retrospective: 'retrospective',
+  action: 'feedback',
+  poker: 'estimation',
+  teamPrompt: 'standup'
+}
+
+export const CATEGORY_TEXT_COLORS: Record<CategoryID, CardTheme> = {
+  standup: {primary: 'text-aqua-400', secondary: 'text-aqua-100'},
+  estimation: {primary: 'text-tomato-500', secondary: 'text-tomato-100'},
+  retrospective: {primary: 'text-grape-500', secondary: 'text-[#F2E1F7]'},
```
igorlesnenko, line 50: > text-[#F2E1F7]

no grape 100? maybe add?

igorlesnenko (approved):
Previously we showed details of the template on hover, did we decide to remove it?
image

packages/client/components/ActivityLibrary/ActivityCard.tsx
```diff
@@ -1,68 +1,96 @@
...
-  const {className, src} = props
+const backgroundImgMap = {
+  retrospective: retroBackgroundSrc,
+  standup: standupBackgroundSrc,
+  feedback: feedbackBackgroundSrc,
+  estimation: estimationBackgroundSrc,
+  strategy: strategyBackgroundSrc,
+  premortem: premortemBackgroundSrc,
+  postmortem: postmortemBackgroundSrc
+} as const
```
Dschoordsch, line 27: +1 Looks like something that could go into `Categories.ts` to keep it in one place

static/images/illustrations/estimation-background.png
Dschoordsch, line 1: +1 I think the images would be more versatile if these wouldn't have rounded corners

packages/client/components/ActivityLibrary/ActivityCard.tsx
```diff
@@ -1,68 +1,96 @@
...
   children?: React.ReactNode
+  type?: MeetingTypeEnum
 }
 
 export const ActivityCard = (props: ActivityCardProps) => {
-  const {className, theme, title, titleAs, badge, children} = props
-  const Title = titleAs ?? ActivityCardTitle
+  const {className, theme, title, children, type, badge} = props
+  const category = type && MEETING_TYPE_TO_CATEGORY[type]
+  const color = category && CATEGORY_TEXT_COLORS[category].primary
```
Dschoordsch, line 72: -1 This is confusing, we're passing in a `theme: CardTheme` which contains the colours, but then we're also deciding some colour here. We should make it consistent. Maybe this should go into the `theme` as well.

packages/client/components/ActivityLibrary/ActivityLibraryCard.tsx
```diff
@@ -8,7 +8,7 @@ export const ActivityLibraryCard = (props: ActivityCardProps) => {
   return (
     <ActivityCard
       className={clsx(
-        'group/card transition-shadow focus-within:ring-2 focus-within:ring-primary hover:shadow-md motion-reduce:transition-none',
+        'transition-shadow focus-within:ring-2 focus-within:ring-primary hover:shadow-md',
```
Dschoordsch, line 11: +1 I personally am not a big fan of the `transition-shadow` as it feels less snappy. If it would be a different thing that we'd animate, it might be different.

Dschoordsch (changes requested):
Works

enriquesanchez (changes requested):
This is looking nice, Nick! 🙌 

I have some suggestions:
- We have specific hover and focus states for the cards, [they are described here](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?type=design&node-id=1581-36866&mode=design&t=4HN6EI5MWu0ts9Rw-4). Can include them as well on this PR?
- I don't want the cards to be too large regardless of the screen size, [on the mockups I have them all set at fixed size](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?type=design&node-id=1402-34308&mode=design&t=4HN6EI5MWu0ts9Rw-4). How does that feel to you? We don't need to implement side-scrolling for now, we can keep the current flow-to-a-new-row behavior.
- It's hard to tell from the Loom, but it looks like the activity title and activity type labels are about the same text size. The title should be 16px and the type 13.
- Speaking of titles,[ I also have some examples of an activity with a long title](https://www.figma.com/file/PmE9A2ryODVzFPjZFJn7qi/Parabol-Activity-Library?type=design&node-id=1340-32517&mode=design&t=4HN6EI5MWu0ts9Rw-4), I think it looks decent and I prefer displaying the full title, what do you think?

nickoferrall:
I've increased the font size: 
Screenshot 2023-11-16 at 18 03 17




> Ideally we always show a max of 4 cards per row. How does that feel to you? 

I prefer the current flow-to-a-new-row behaviour as I like seeing everything without using the horizontal scroll 


The other two points sound good! I'd like to tackle them in separate issues, though, to get this shipped asap. Here are the issues I've created: https://github.com/ParabolInc/parabol/issues/9201 & https://github.com/ParabolInc/parabol/issues/9200

packages/client/components/ActivityLibrary/ActivityCard.tsx
nickoferrall: Sounds good; I've removed `bg-` from the category themes, so we just have the colours. That way, we can add `bg-` or `text-` wherever they're imported

enriquesanchez:
>I prefer the current flow-to-a-new-row behaviour as I like seeing everything without using the horizontal scroll

@nickoferrall I think we are in agreement, I also don't think we should introduce horizontal scroll. What I meant is that we should keep the cards to a single size, regardless of the browser size. I do wan't to avoid super big cards, they take up so much space and affect discovery. On the screenshot you just shared above, I think 4 cards would fit much more nicely.

Dschoordsch (approved):

Issue #9188, opened 2023-11-14
Unable to AutoJoin with SAML


nickoferrall:
We currently check if `identities` contains an identity with `isEmailVerified`, but we don't update `identities` when users login with SAML.

### AC

- [ ] Users that login with SAML can still autoJoin

PR #9189, opened 2023-11-14
fix: can auto join with saml login

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9188

### To test

- [ ] Add the autoJoin flag to a team that contains a billing leader or founder with a verified email 
- [ ] Login with SAML and automatically join those teams

### DIFF
--- a/packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
+++ b/packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
@@ -18,6 +18,7 @@ import {makeDefaultTeamName} from 'parabol-client/utils/makeDefaultTeamName'
 import {DataLoaderWorker} from '../../graphql'
 import acceptTeamInvitation from '../../../safeMutations/acceptTeamInvitation'
 import isValid from '../../isValid'
+import getSAMLURLFromEmail from '../../../utils/getSAMLURLFromEmail'
 
 const bootstrapNewUser = async (
   newUser: User,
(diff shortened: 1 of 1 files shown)


packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
```diff
@@ -76,10 +77,12 @@ const bootstrapNewUser = async (
   }
 
   const isVerified = identities.some((identity) => identity.isEmailVerified)
+  const hasSAMLURL = !!(await getSAMLURLFromEmail(email, dataLoader, false))
+  const isQualifiedForAutoJoin = (isVerified || hasSAMLURL) && isCompanyDomain
```
Dschoordsch, line 81: -1 I'm fine with the change as is, but only as a temporary workaround. The real solution would be to add an identity when calling `loginWithSAML`. If we have this workaround and the correct identity adding, we can remove the workaround in a month or so

Dschoordsch (changes requested):

packages/server/graphql/mutations/helpers/bootstrapNewUser.ts
Dschoordsch: I'm fine splitting it into 3 parts: this workaround, identity fix, removing the workaround. This should be properly reflected in the issues.

igorlesnenko: We call `bootstrapNewUser` from `loginSAML`. Alternative solution is to pass some additional param here:
```
  const authToken = await bootstrapNewUser(tempUser, !isInvited, dataLoader)
  ```
  
  or check if 
  
  `tier: 'enterprise'` which by default for saml
  
  What do you think?

igorlesnenko: Or check `const userId = `sso|${generateUID()}``

igorlesnenko (changes requested):
We know that we are bootstraping a saml user, I suggest to pass it as an argument to a function, and avoid additional queries

Dschoordsch (approved):
Did not test

PR #9193, opened 2023-11-14
chore(metrics): Only track 'Loaded a Page' event to Amplitude when userId is known

tianrunhe:
# Description
Amplitude bills us on MTU (Monthly Tracking User). Last month our MTU is a little bit above the limit of 100K however our MAU is only at 60K. Upon investigation, I found that there were enough anonymous users triggering the page view events. For example, [this user](https://app.amplitude.com/analytics/parabol/project/224689/search/amplitude_id%3D754404294669) just visited parabol.co, [this user](https://app.amplitude.com/analytics/parabol/project/224689/search/amplitude_id%3D749613875219) run a demo, and [this user](https://app.amplitude.com/analytics/parabol/project/224689/search/amplitude_id%3D746782982176) reset the password. They were not registered or logged in. Amplitude count them as distinct users, causing our MTU spikes.
Eventually we would like to have the anonymous traffic in Amplitude too (majority being from www.parabol.co) so we can do full channel analysis but today is not the date yet (plus we can achieve similar functionality by GA).
This simple change just assure that only when the `userId` is known that we sent `Loaded a Page` event to Amplitude. This way the MTU should be in line with our MAU so we have more buffer below the 100K threshold.

## Testing scenarios
- Open up developer tool on your browser
- Go to a page without logged in: localhost:3000/create-account or localhost:3000/forgot-password and see no Amplitude debug info output
- Logged in, visit any pages and observe the Amplitude debug output, indicating the `Loaded a Page` event is sent

## Final checklist


packages/client/components/AnalyticsPage.tsx
```diff
@@ -172,21 +172,24 @@ const AnalyticsPage = () => {
...
-          path: pathname,
-          url: href,
-          translated,
-          search: location.search
-        },
-        {
-          user_id: atmosphere.viewerId
-        }
-      )
+      const user_id = atmosphere.viewerId
```
Dschoordsch, line 175: +1 It would be nice to keep our naming convention for variables

Dschoordsch (approved):
Did not test it

PR #9199, opened 2023-11-16
feat(kudos): display notification when kudos received

### LINKED ISSUE #9135: Kudos: add receiver notifications
**Designs:**

image

image

**AC:**

- Receiver of kudos should receive a notification + snackbar
- Notification should have some context of where kudos was added

igorlesnenko:
Fixes #9135

image

image

**How to test:**
- Send kudos
- On the receiver side see a snackbar and notification


igorlesnenko:
@acressall I added a link to a meeting in a snackbar too, is it ok?
Do we still need to not show snackbar if I’m in a different meeting?

acressall:
Love the meeting link in the snackbar!

> Do we still need to not show snackbar if I’m in a different meeting?
No, I think it would be great if we could avoid this. Terry made a good point that it would be a distraction if you're in a different meeting

acressall (approved):

igorlesnenko:
> No, I think it would be great if we could avoid this. Terry made a good point that it would be a distraction if you're in a different meeting

I'll create an issue and add it to v2 milestone if that is ok. I feel it on v1 it should be a rare case where you are in one meeting but someone likes your comment in another active meeting

packages/client/components/KudosReceivedNotification.tsx
```diff
@@ -0,0 +1,59 @@
...
+      }
+    `,
+    notificationRef
+  )
+  const {type, name, picture, meetingName, emoji, meetingId} = notification
+
+  const {unicode} = getReactji(emoji)
+
+  useEffect(() => {
+    SendClientSideEvent(atmosphere, 'Notification Viewed', {
```
nickoferrall, line 38: +1 This seems to send a client event for every kudos notification in the dropdown every time the user opens the notification dropdown menu.

Would it be more helpful to only send the client event if it's unread?

packages/client/mutations/toasts/mapKudosReceivedToToast.ts
```diff
@@ -0,0 +1,46 @@
...
+import {OnNextHistoryContext} from '../../types/relayMutations'
+import SendClientSideEvent from '../../utils/SendClientSideEvent'
+import getReactji from '~/utils/getReactji'
+
+graphql`
+  fragment mapKudosReceivedToToast_notification on NotifyKudosReceived {
+    id
+    type
+    name
+    picture
```
nickoferrall, line 14: +1 is picture or type being used?

nickoferrall (approved):

packages/client/components/KudosReceivedNotification.tsx
igorlesnenko: I made it similar to other notifications. @tghanken what do you think?

tghanken: I think this is fine as is.  It could be helpful to add "isRead" ads a boolean parameter if it's easy!

tghanken (approved):
Looks good!

packages/client/components/KudosReceivedNotification.tsx
igorlesnenko: we have notification status, CLICKED, READ etc. I can add it

PR #9222, opened 2023-11-20
fix: fix recreating invite link in case of expiration

### LINKED ISSUE #9062: Invite link on mobile sometimes displays ..... for invite link stub
Invite link on mobile sometimes displays ..... for invite link stub resulting in a link that is not useable. This seems to be random.

image

igorlesnenko:
fix #9062

I'm not exactly sure why #9062 happens.
We show dots if token is invalid, then we are fetching another token from the frontend.
I may suggest that on mobile with slow connection user may see dots for a while while we are fetching a new token.

However, this should rarely happen because if token is expired, we create new on the backend before returning it. 

This mechanism was broken. I fixed it, which possibly fixes the whole issue

**How to test:**
- Create a fresh team
- Open invite dialog, copy invitation ID from the URL for example XSDBR1S0TF0X
- Expire the url by running:

```
r.db('actionDevelopment')
  .table("MassInvitation")
 .get('XSDBR1S0TF0X')
 .update({
   expiration: new Date()
 })
```

- Refresh page, open invite dialog again, see that a new invite URL appears and you didn't see dots while it was loading


packages/server/graphql/types/Team.ts
```diff
@@ -85,7 +85,7 @@ const Team: GraphQLObjectType = new GraphQLObjectType<ITeam, GQLContext>({
           .load(teamMemberId)
         const [newestInvitationToken] = invitationTokens
         // if the token is valid, return it
-        if (newestInvitationToken?.expiration ?? new Date(0) > new Date())
```
igorlesnenko, line 88: this was always `false`

nickoferrall (approved):

Dschoordsch (changes requested):
The change is fine, but does not solve the underlying problem. I could trigger the *** invite by:
* open the "Invite to Team" dialog
* stop the server
* invalidate the token
* start the server

The client reconnects but refreshes the token with ***

igorlesnenko:
@Dschoordsch not able to reproduce this, am I doing this right: https://www.loom.com/share/87e349a8d825420ebacabc5881d087ea ?

Dschoordsch:
Yes, the steps were correct, I think only the slow restart of the server messed it up: https://www.loom.com/share/428542d75b5d4be0aad2d544f9b8b0bd?sid=c859c58b-64f7-4029-878d-ba02dfc8a515

igorlesnenko:
Finally reproduced using offline mode in Network tab

Dschoordsch (approved):
Works!

PR #9233, opened 2023-11-22
fix: author should be nullable when comment was anonymous

### LINKED ISSUE #9197: @ mentions in discussion thread reveal anonymous comments in Notifications
## Issue - Bug
- **How serious is the bug?** P1

@ mentions reveal anonymous discussion comment writer. If someone replies anonymously in a discussion thread and @ mentions another team member. This team member will get a notification with the real account attached to it. It won't stay anonymous, but rather reveal who wrote the anonymous comment. Not great.

Not sharing any screenshots to keep the people anonymous. 

Users can:
   - Notifcations should keep the original user anonymous, be it in the notification center or email. 

BartoszJarocki:
# Description

Fixes #9197 

Main issue was that we didn't check if comment was anonymous in related gql resolvers. keep in mind svg avatars in emails won't work, they _should_ work on prod, see [thread 🔒 ](https://parabol.slack.com/archives/C608QL4RH/p1701167748897159).

## Demo

Screenshot 2023-11-28 at 17 28 08

## Testing scenarios

- [ ] mention someone in discussion thread as anonymous and see if mentioned user see correctly notification about anonymous mention
- [ ] respond as anonymous to someone else standup response and check the notification
- [ ] do the same but for emails

## Final checklist

### DIFF
--- a/packages/server/graphql/public/types/NotifyDiscussionMentioned.ts
+++ b/packages/server/graphql/public/types/NotifyDiscussionMentioned.ts
@@ -7,7 +7,10 @@ const NotifyDiscussionMentioned: NotifyDiscussionMentionedResolvers = {
     const meeting = await dataLoader.get('newMeetings').load(meetingId)
     return meeting as MeetingTeamPrompt
   },
-  author: ({authorId}, _args: unknown, {dataLoader}) => {
+  author: async ({authorId, commentId}, _args: unknown, {dataLoader}) => {
+    const comment = await dataLoader.get('comments').load(commentId)
+    if (comment.isAnonymous) return null
+
     return dataLoader.get('users').loadNonNull(authorId)
   },
   comment: ({commentId}, _args: unknown, {dataLoader}) => {
--- a/packages/server/graphql/public/types/NotifyResponseReplied.ts
+++ b/packages/server/graphql/public/types/NotifyResponseReplied.ts
@@ -13,7 +13,10 @@ const NotifyResponseReplied: NotifyResponseRepliedResolvers = {
     const responses = await getTeamPromptResponsesByMeetingId(meetingId)
     return responses.find(({userId: responseUserId}) => responseUserId === userId)!
   },
-  author: ({authorId}, _args: unknown, {dataLoader}) => {
+  author: async ({authorId, commentId}, _args: unknown, {dataLoader}) => {
+    const comment = await dataLoader.get('comments').load(commentId)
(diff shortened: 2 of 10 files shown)


jmtaber129 (approved):
Didn't test, but code LGTM

packages/server/graphql/public/types/NotifyDiscussionMentioned.ts
```diff
@@ -7,7 +7,10 @@ const NotifyDiscussionMentioned: NotifyDiscussionMentionedResolvers = {
     const meeting = await dataLoader.get('newMeetings').load(meetingId)
     return meeting as MeetingTeamPrompt
   },
-  author: ({authorId}, _args: unknown, {dataLoader}) => {
+  author: async ({authorId, commentId}, _args: unknown, {dataLoader}) => {
+    const comment = await dataLoader.get('comments').load(commentId)
+    if (comment.isAnonymous) return null
```
Dschoordsch, line 12: -1 This is a good quick fix, but we should not store the `author` in the first place for anonymous comment notifications

packages/server/graphql/public/types/NotifyResponseReplied.ts
```diff
@@ -13,7 +13,10 @@ const NotifyResponseReplied: NotifyResponseRepliedResolvers = {
     const responses = await getTeamPromptResponsesByMeetingId(meetingId)
     return responses.find(({userId: responseUserId}) => responseUserId === userId)!
   },
-  author: ({authorId}, _args: unknown, {dataLoader}) => {
+  author: async ({authorId, commentId}, _args: unknown, {dataLoader}) => {
+    const comment = await dataLoader.get('comments').load(commentId)
+    if (comment.isAnonymous) return null
```
Dschoordsch, line 18: -1 See above, we shouldn't store the author in the first place

Dschoordsch (approved):
Tested, works. It would be nice to not store identifiable information for anonymous comments if it's not necessary.
To fix existing notifications this change is fine, but we should change what we store going forward.

Dschoordsch:
I would like to see a follow up PR or issue.

BartoszJarocki:
> Tested, works. It would be nice to not store identifiable information for anonymous comments if it's not necessary. To fix existing notifications this change is fine, but we should change what we store going forward.

I very much agree! the scope would be bigger than this so I think it's better to fix the current implementation and create an issue for changing the implementation and removing the information from the db at all.

PR #9236, opened 2023-11-27
feat: add tooltip to activity library card

nickoferrall:
Fix: https://github.com/ParabolInc/parabol/issues/9116

Demo: https://www.loom.com/share/40363cfc26084c26a20f237358aea811

### To test

- [ ] Hover over the activity library cards and see that it looks like to the linked issue


tailwind.config.js
```diff
@@ -8,6 +8,34 @@ module.exports = {
   },
   presets: [require('./packages/client/tailwind.preset.js')],
   content: ['./packages/client/**/*.{ts,tsx,js,jsx,html}', './template.html'],
+  safelist: [
+    'bg-aqua-400',
```
nickoferrall, line 12: This lets us use these colours in a variable in tailwind. In the parent issue, click on "Retrospective" and see that the background colour doesn't show up in the button, but it does in this PR.

Screenshot 2023-11-27 at 16 32 49

enriquesanchez (commented):
Nice!

What's the logic behind the placement of the tooltip? I see that sometimes it's centered and covers the card behind, and sometimes it's a little more below and only covers part of it.

nickoferrall:
I'm just using the [default](https://github.com/ParabolInc/parabol/blob/2b9a5c9d7f0b7eb0e800264bff999e17a735c18f/packages/client/ui/Tooltip/TooltipContent.tsx) tooltip logic that we have in the codebase for now

enriquesanchez (approved):

packages/client/components/ActivityLibrary/ActivityCard.tsx
```diff
@@ -44,14 +51,44 @@ export interface ActivityCardProps {
...
+        ...ActivityLibraryCardDescription_template
+      }
+    `,
+    templateRef ?? null
+  )
+
+  const handleMouseEnter = () => {
+    const timeoutId = setTimeout(() => {
+      setShowTooltip(true)
+    }, 750)
```
igorlesnenko, line 75: When I trying to read about many templates one by one, personally, I feel it could be faster.

packages/client/components/ActivityLibrary/ActivityCard.tsx
```diff
@@ -72,6 +109,17 @@ export const ActivityCard = (props: ActivityCardProps) => {
           </div>
         </div>
       )}
+      {template && (
+        <Tooltip open={showTooltip}>
```
igorlesnenko, line 113: Sometime position is a bit weird, this might be a separate issue with tooltip component itself or we may need to specify some radix props regarding position

image

packages/client/components/ActivityLibrary/ActivityCard.tsx
```diff
@@ -72,6 +109,17 @@ export const ActivityCard = (props: ActivityCardProps) => {
...
+          <TooltipTrigger asChild>
+            <div className='h-50 w-50 text-sky-500 hover:text-sky-700' />
+          </TooltipTrigger>
+          <TooltipContent className='max-w-md whitespace-normal rounded-lg bg-white p-4 text-left text-slate-700 shadow-lg hover:cursor-pointer sm:max-w-sm'>
```
igorlesnenko, line 117: Sometimes tooltip cover the whole card, if I click on it I go to that template, which might be not obvious behaviour
Maybe clicking should only close tooltip or have pointer cursor

image

igorlesnenko (approved):

packages/client/components/ActivityLibrary/ActivityCard.tsx
nickoferrall: Happy to make this 500ms instead. Let's see how Georg feels about the speed

packages/client/components/ActivityLibrary/ActivityCard.tsx
nickoferrall: Yeah, we had some debate about where the tooltip should go. Maybe in a fast-follow PR I move it below the card at least?

enriquesanchez: Initially we thought the tooltip could hang from and the mouse pointer:
Captura de pantalla 2023-11-28 a la(s) 3 30 42 p m

Dschoordsch (changes requested):
I think it's good enough for now, but it feels a bit messy.

packages/client/components/ActivityLibrary/ActivityCard.tsx
```diff
@@ -44,14 +51,44 @@ export interface ActivityCardProps {
...
   type?: MeetingTypeEnum
+  templateRef?: ActivityCard_template$key
 }
 
 export const ActivityCard = (props: ActivityCardProps) => {
-  const {className, theme, title, children, type, badge} = props
+  const {className, theme, title, children, type, badge, templateRef} = props
   const category = type && MEETING_TYPE_TO_CATEGORY[type]
+  const [showTooltip, setShowTooltip] = useState(false)
+  const [hoverTimeout, setHoverTimeout] = useState<NodeJS.Timeout | null>(null)
```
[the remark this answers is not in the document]
nickoferrall: Sounds good. Updated to `useRef`

tailwind.config.js
nickoferrall: In the previous PR you mentioned [here](https://github.com/ParabolInc/parabol/pull/9168#discussion_r1396147758) that you found it confusing to have both text & bg colour variables. 

Would you like to revert back to that or something else? Happy for me to make that change in a separate PR?

nickoferrall: @Dschoordsch can we merge this PR without updating these colours? I could do a quick release with this, gcal, and the checkout flow

Dschoordsch: Let's just move all the colours into the theme, add a third field `{primary: 'bg-...', secondary:..., text: 'text-...'}` and remove the safelist. That shouldn't take long at all.

nickoferrall: I've updated the `CATEGORY_THEMES` object adding a third `text` key

Issue #9246, opened 2023-11-29
Remove gcal feature flag


nickoferrall:
### AC

- [ ] Remove the gcal feature flag so we can release this to all users

PR #9251, opened 2023-11-29
feat: remove gcal flag

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9246

### To test

- [ ] With a user that does not have the gcal feature flag, create a gcal event


igorlesnenko (commented):
Not fully removed

> modules/teamDashboard/components/ProviderList/ProviderList.tsx:123:74 - error TS2339: Property 'gcal' does not exist on type '{ readonly azureDevOps: boolean; readonly msTeams: boolean; }'.

PR #9256, opened 2023-12-01
chore: efficient webpack bundles

### LINKED ISSUE #9204: Service worker requesting all assets on every new version
As we release more, we want the impact of a new release to decrease. I noticed that the service worker is downloading all 295 assets when a version gets bumped, even though only a handful of files changed. We need to fix this!

AC
- [x] new versions only trigger a fetch of the changed files
- [x] reduction in opaque requests (add crossorigin to requests, etc. i'll do as many of these before i run out of time)
- [x] skeleton.html is removed from the sw manifest
- [x] License files are removed from webpack build (these are in source maps, which we ship. we're also open source!)
- [x] GraphiQL assets are not included in the sw manifest (we exclude GraphQLContainer, but that's not working because the dependencies are still included. not sure why...)

css doesn't support modifiers to url fetches, e.g. `background-image: url(/foo.png) crossorigin=''`. That means any fetches that originate from css can't have CORS added to them. I wonder if I can intercept this fetch request in the service worker & add CORS to the request 🤔

mattkrick:
# Description

Fixes #9204

As we move towards shipping 2x/week, it's important that users are not affected by the increased frequency. That means minimal extra downloading & parsing between versions.

Changes:
- removed ES6 polyfills (unless someone's browser is from 2016, this won't be a problem)
- ensure fetches are not opaque so we can cache more aggressively
- remove `*.LICENSE` files from bundles (technically these have to be at the top of a file anyways, not in a separate file, but since we're open source we don't have to worry about that)
- remove CSS sourcemaps
- minify using SWC (little smaller, little faster)
- output assets (e.g. images) keep their name so we know what they actually are (but have a hashed suffix so if they change it'll invalidate the cache)
- require minimum chunk size, which reduced our bundle size by 550KB gzipped 🤯 
- SERVER: disallow chunks so clients aren't confused with extraneous JS files

Notes:
- It'd still be nice to remove admin tools like `GraphqlContainer` from getting fetched since that's only for parabol employees, but that's hard to do because we require a minimum chunk size & it shares a lot of vendors dependencies. Thankfully, this shouldn't get refetched unless we bump a large vendor package like react 🎉


packages/client/serviceWorker/sw.ts
```diff
@@ -73,7 +74,9 @@ const onFetch = async (event: FetchEvent) => {
       return cachedRes
     }
     try {
-      const networkRes = await fetch(request)
+      // request.mode could be 'no-cors'
+      // By fetching the URL without specifying the mode the response will not be opaque
+      const networkRes = await fetch(request.url)
```
mattkrick, line 79: want to get your thoughts on this! it's the same thing we do up above for all the assets that go into the static cache

Dschoordsch: I have some trouble testing a production build locally, also on master. Would be nice to get this PR running on a on-demand k8s environment, but we could also test on staging.
It seems to  fetch with `cors` for the static content, so I think you're right, it should work for the dynamic ones as well. It probably fails if we fetch stuff from 3rd parties.

mattkrick: oh! that's no good. what's the error you see? I can help debug!

my checklist for testing prod build locally:
- close all but 1 tab of parabol, unregister service worker, hard refresh (cmd + shift + r on mac)
- `yarn build && yarn predeploy && yarn dev` making sure `CDN_BASE_URL` is undefined


to test 3rd party stuff I turned on the new checkout flow & went to the org page so stripe would fetch its stuff. It works! We could scope it to just our assets that we serve from our CDN if you think that's safer. Your call!

Screenshot 2023-12-04 at 3 29 04 PM

Dschoordsch: I'm running `yarn build && yarn predeploy && yarn start` with the CDN set to action-files.parabol.co/development. It used to work a month ago. It appears to be an infrastructure issue. I used to deploy the files to action-files/development, but now I cannot access the data there anymore.
I cannot successfully run `yarn predeploy` without CDN set, also I wonder how accurate testing would be without an external server.

mattkrick: > I cannot successfully run yarn predeploy without CDN set,

what's the error you're seeing?




FWIW here's the output i get when using the following & building locally

```
CDN_BASE_URL='//action-files.parabol.co/development'
FILE_STORE_PROVIDER='s3'
```

Screenshot 2023-12-07 at 3 05 03 PM

Dschoordsch (approved):

packages/client/serviceWorker/sw.ts
mattkrick: switched to only cache things coming from parabol or the parabol CDN ✅ 
true, we don't rotate the dynamic cache! I think that's probably OK, there's no TTL available so it's probably better to flush on version than let it grow

PR #9261, opened 2023-12-01
fix: cork all http write methods

### LINKED ISSUE #9253: cork uWS methods
With Node v20 we had to bump uWS, which now warns when methods are used outside of a cork.

AC
- [ ] cork all uWS methods

mattkrick:
# Description

fix #9253

The latest version of uWS _really_ wants us to cork responses correctly & it's blowing up our logs.
Instead of manually wrapping every HTTP response method with a cork, I wrote it into the monkey patched wrapper so we can keep writing code & not have to worry about forgetting to cork something.
The end result is removing all the one-off corks that i did for high-usage routes. Yay less code!

From uWS:

     * Corking a response is a performance improvement in both CPU and network, as you ready the IO system for writing multiple chunks at once.
     * By default, you're corked in the immediately executing top portion of the route handler. In all other cases, such as when returning from
     * await, or when being called back from an async database request or anything that isn't directly executing in the route handler, you'll want
     * to cork before calling writeStatus, writeHeader or just write. Corking takes a callback in which you execute the writeHeader, writeStatus and
     * such calls, in one atomic IO operation. This is important, not only for TCP but definitely for TLS where each write would otherwise result
     * in one TLS block being sent off, each with one send syscall.
    

## Testing scenarios

- Test it with websockets: click around the app, everything works as expected
- Test it with SSE: remove `trySockets` from this line https://github.com/ParabolInc/parabol/blob/36a9cfc198b4083df88b36985422bc896d08228a/packages/client/Atmosphere.ts#L263 NOTE: There's a bug where SSE doesn't work in dev mode! So you'll have to run `yarn build && yarn predeploy && yarn start`


packages/server/safetyPatchRes.ts
```diff
@@ -69,17 +81,17 @@ const safetyPatchRes = (res: HttpResponse) => {
     if (res.done) {
       console.warn(`uWS: Called writeHeader after done`)
     }
-    if (res.done || res.aborted) return res
-    return res._writeHeader(key, value)
+    res.headers.push([key, value])
```
mattkrick, line 84: This is the crux of it. uWS gets mad when we call write methods outside of a cork, so we capture those writes & save them for later when we're ready to cork.

BartoszJarocki (approved):
everything seems to work well!

PR #9267, opened 2023-12-05
fix: shake some images from server bundle

mattkrick:
# Description

On the server we don't do ES6 modules yet, so tree shaking isn't as good as it can be.
Until that brighter future, we'll just have to keep modules small, knowing that importing a module will import everything it in, whether it gets used or not.
99% of the time, this doesn't matter. In this case, the server was importing from `Categories`, which now includes some images. That meant we were creating those images in the server bundle even though the server has no need for them.

## Demo

Before
Screenshot 2023-12-05 at 10 30 55 AM

After
Screenshot 2023-12-05 at 10 33 50 AM

to reproduce: `yarn build`


nickoferrall (approved):
Ahh this is good to know!

PR #9278, opened 2023-12-06
chore: put server assets on CDN

### LINKED ISSUE #8553: parabol icon in mattermost should be absolute
Our icon in mattermost doesn't show up. I assume this is true when not using a CDN which causes the static asset dir to be relative (/static). The easy fix is to make it absolute by appending the process.env.HOST to the URL to make it absolute

AC
- [ ] parabol icon in mattermost shows up

### LINKED ISSUE #9111: Accessing jira-attachments will result in a 5xx
From Slack:
https://parabol.slack.com/archives/C4JAUUZ9P/p1698966704621919?thread_ts=1698858093.998069&cid=C4JAUUZ9P

> response_sent_by_backend: mostly [action.parabol.co/jira-attachements/HASHES](http://action.parabol.co/jira-attachements/HASHES) urls. 1087 entries, see GCP logs [here](https://console.cloud.google.com/logs/query;query=resource.type%3D%22http_load_balancer%22%20AND%20http_request.status%3E%3D500%0AjsonPayload.statusDetails%3D%22response_sent_by_backend%22;summaryFields=jsonPayload%252FstatusDetails:false:32:beginning;cursorTimestamp=2023-10-31T15:49:42.262323Z;duration=P7D?project=parabol-saas-production). Something must be not working with Jira. This status response_sent_by_backend means that the application is answering with 5XX and the LB is just passing the answer through. We should probably fix how we manage errors coming from JIRA to stop answering with 5XX here.
Latency for the jira-attachements isn't that high, most of them less than 200ms and some almost 1s
Also some POST calls to [action.parabol.co/graphql](http://action.parabol.co/graphql), which have higher latency (over 10s).

AC
- [ ] audit the [handler](https://github.com/ParabolInc/parabol/blob/7c60878ee1740fd9b60a1d62c10b584d04802d11/packages/server/jiraImagesHandler.ts) and see why we don't always return a 200
- [ ] if it's still unclear, gather some recent logs & see if we can repeat the request for a given hash

mattkrick:
# Description

Publicly host server assets so they can be used by downstream services like email, mattermost, etc.

fix #9111
fix #8553

## Testing scenarios

in dev, with no CDN
- [x] http://localhost:3000/jira-attachments/foo.png results in a image not found .png
- [x] emails using anonymous avatar work correctly

in prod, with no CDN (after running preDeploy)
- [x] http://localhost:3000/jira-attachments/foo.png results in a image not found .png
- [x] notifications published to mattermost include an absolute url starting with env.HOST
- [x] images in manifest.json are available
- [x] emails using anonymous avatar work correctly

in prod, with CDN (after running preDeploy)
- [x] http://localhost:3000/jira-attachments/foo.png results in a image not found .png
- [x] notifications published to mattermost include an absolute url starting with CDN_BASE_URL
- [x] images in manifest.json are available
- [x] templates upload to the existing location


packages/server/initPublicPath.ts
```diff
@@ -0,0 +1,16 @@
+import appOrigin from './appOrigin'
+
+declare let __webpack_public_path__: string
```
mattkrick, line 3: we can't set the public path at buildtime because different PPMIs will host their assets in different places.
since the public path is defined during predeploy, we must set it at runtime

packages/server/jiraImagesHandler.ts
```diff
@@ -31,9 +29,12 @@ const getImageFromCache = async (
 let jiraPlaceholderBuffer: Buffer | undefined
 const servePlaceholderImage = async (res: HttpResponse) => {
   if (!jiraPlaceholderBuffer) {
-    jiraPlaceholderBuffer = await fsp.readFile(
-      path.join(__dirname, jiraPlaceholder.slice(__webpack_public_path__.length))
-    )
+    try {
+      const res = await fetch(jiraPlaceholder)
```
mattkrick, line 33: all other assets need a public URL except this one, so instead of special casing this one I just turned it into a fetch vs. a file system read.

packages/server/utils/serveStatic.ts
```diff
@@ -18,9 +18,9 @@ const getProjectRoot = () => {
 const PROJECT_ROOT = getProjectRoot()
 const staticPaths = {
   [path.join(PROJECT_ROOT, 'build')]: true,
-  [path.join(PROJECT_ROOT, 'dist')]: !__PRODUCTION__,
-  [path.join(PROJECT_ROOT, 'static')]: !__PRODUCTION__,
-  [path.join(PROJECT_ROOT, 'dev', 'dll')]: !__PRODUCTION__
+  // publish server assets at /static
+  [path.join(PROJECT_ROOT, 'dist')]: __PRODUCTION__,
```
mattkrick, line 22: path is left intact, e.g. `dist/images/imageNotFound.png` will be published at `/static/images/imageNotFound.png`

scripts/toolboxSrc/applyEnvVarsToClientAssets.ts
```diff
@@ -82,7 +78,7 @@ const rewriteIndexHTML = () => {
   const keys = `<script>window.__ACTION__=${JSON.stringify(clientKeys)}</script>`
   const rawHTML = skeleton
     .replace('<head>', `<head>${noindex}${keys}`)
-    .replaceAll('__PUBLIC_PATH__', getCDNURL())
+    .replaceAll('__PUBLIC_PATH__', __webpack_public_path__.replace(/\/$/, ''))
```
mattkrick, line 81: remove trailing slash

scripts/toolboxSrc/pushToCDN.ts
```diff
@@ -2,6 +2,11 @@ import fs from 'fs'
 import getFileStoreManager from 'parabol-server/fileStorage/getFileStoreManager'
 import path from 'path'
 import getProjectRoot from '../webpack/utils/getProjectRoot'
+;(require as any).context(
```
mattkrick, line 5: since the DB needs these files, we require them here so they get added to the server bundle, e.g. `dist/images/templates`

scripts/toolboxSrc/pushToCDN.ts
```diff
@@ -25,46 +30,78 @@ const pushClientAssetsToCDN = async () => {
   console.log(`⛅️ Uploaded ${dirEnts.length} client assets to CDN`)
 }
 
-const pushTemplatesToCDN = async () => {
+const pushServerAssetsToCDN = async () => {
```
mattkrick, line 33: we were just pushing client assets, now we push server assets, too.
This is useful for things like:
- anonymousAvatar getting referenced in emails
- our logo's URL getting published in mattermost messages

scripts/webpack/dev.servers.config.js
```diff
@@ -25,8 +26,8 @@ module.exports = {
     __dirname: false
   },
   entry: {
-    web: [DOTENV, path.join(SERVER_ROOT, 'server.ts')],
-    gqlExecutor: [DOTENV, path.join(GQL_ROOT, 'gqlExecutor.ts')]
+    web: [DOTENV, INIT_PUBLIC_PATH, path.join(SERVER_ROOT, 'server.ts')],
```
mattkrick, line 29: the public path has to get set before imports use it, so set it here because that way the import order of server.ts doesn't matter

.github/workflows/test.yml
```diff
@@ -92,11 +92,12 @@ jobs:
         run: yarn predeploy
 
       - name: Start testing server in background
-        run: |
-          yarn start &
-
-      - name: Wait for testing server to be healthy
-        run: curl -4 --retry 30 --retry-connrefused --retry-delay 2 http://localhost:3000/graphql
+        uses: JarvusInnovations/background-action@v1
```
mattkrick, line 95: adding this lets us see server logs after the workflow finishes

.github/workflows/test.yml
```diff
@@ -129,5 +130,5 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: test-results
-          path: test-results/
+          path: packages/integration-tests/test-results/
```
mattkrick, line 133: this has been wrong since forever, i just didn't ever need it until today 😞

mattkrick:
@igor would you mind giving this a review?

igorlesnenko (commented):
I do: `yarn build`, `yarn predeploy`, `yarn start`
And images of meetings are not displayed. Am I missing something? 
image

Dschoordsch (approved):

PR #9279, opened 2023-12-07
chore: webpack client and server in parallel

mattkrick:
# Description

I've been doing a lot of testing of the production build & I hate waiting around for 1 minute to watch it build.
By building client & server at the same time, in separate threads, it cuts that time in half 🎉 .

I used `child_process` instead of `worker_threads` because the API is a little cleaner (no need for an extra file). Plus it looks like it's more performant, too, which is a NodeJS problem: https://github.com/orgs/nodejs/discussions/44264

The webpack files hardly changed, i just wrapped the return value in `{}` & normalized the input params. The git diff makes it look much worse!
 
## Demo

Before

```
🙏🙏🙏      Building Production Server      🙏🙏🙏
✔ Parse Configuration
✔ Generate outputs
✨  Done in 58.23s.
```

After

```
🙏🙏🙏      Building Production Server      🙏🙏🙏
✔ Parse Configuration
✔ Generate outputs
✨  Done in 30.16s.
```

## Testing scenarios

`yarn build --no-deps`

### DIFF
--- a/scripts/webpack/prod.client.config.js
+++ b/scripts/webpack/prod.client.config.js
@@ -33,185 +33,189 @@ const babelPresets = [
   ]
 ]
 
-module.exports = ({isDeploy, isStats}) => ({
-  stats: {
-    assets: false
-  },
-  devtool: 'source-map',
-  mode: 'production',
-  entry: {
-    app: [path.join(CLIENT_ROOT, 'polyfills.ts'), path.join(CLIENT_ROOT, 'client.tsx')]
-  },
-  output: {
-    path: buildPath,
-    publicPath: 'auto',
-    filename: '[name]_[contenthash].js',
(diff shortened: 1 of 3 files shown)


scripts/webpack/prod.client.config.js
```diff
@@ -33,185 +33,189 @@ const babelPresets = [
...
-    crossOriginLoading: 'anonymous'
-  },
-  resolve: {
-    alias: {
-      '~': CLIENT_ROOT,
-      'parabol-client': CLIENT_ROOT,
-      static: STATIC_ROOT
+module.exports = (config) => {
+  const isDeploy = config.isDeploy === 'true'
+  const isStats = config.isStats === 'true'
```
mattkrick, line 38: everything after this point is untouched

Dschoordsch (approved):
Works!

Issue #9234, opened 2023-11-22
Kudos: send kudos in reflections


igorlesnenko:
Currently it is not possible to mention people in refections

**AC:**

- Enable mentions in reflections
- Send kudos when meeting ends and summary is generated

PR #9284, opened 2023-12-13
feat(kudos): enable mentions in retro reflections

igorlesnenko:
Related to #9234. We need mentions in reflections to later use it for sending kudos

image

**How to test:**

- Start a retro
- Add few reflections and see that you can mention team users in them
- See that in all other places mention is rendered correctly — vote phase, summary etc.
- Because of some refactoring, play with retro and make sure nothing's broken in general for reflection cards or and reflections editor


packages/client/components/ReflectionGroup/RemoteReflection.tsx
```diff
@@ -273,6 +275,16 @@ const RemoteReflection = (props: Props) => {
             editorState={editorState}
             readOnly
             disableAnonymity={disableAnonymity}
+            teamId={teamId}
+            editorRef={editorRef}
+            placeholder=''
+            setEditorState={() => {
```
nickoferrall, line 281: +1 could this and `handleKeyDownFallback` and `placeholder` be optional props in `ReflectionEditorWrapper`?

nickoferrall (approved):

packages/client/components/RetroReflectPhase/PhaseItemEditor.tsx
```diff
@@ -272,6 +283,9 @@ const PhaseItemEditor = (props: Props) => {
                   editorState={card.editorState}
                   readOnly
                   disableAnonymity={disableAnonymity}
+                  teamId={teamId}
+                  editorRef={editorRef}
```
Dschoordsch, line 287: -1 this part seems wrong. If we have some cards in flight, they should not need a reference to the editor anymore as I should not be able to do any edits anymore

Dschoordsch (changes requested):
Overall looks good. The added functions, like mentions and links work. However there is something weird happening with the editor state: https://www.loom.com/share/4307debd45fb49968183f7c68f58ca94?sid=9cdf50f6-16dc-4d9c-a211-c46e7328e49a

Dschoordsch (approved):
Works.
One thing that's missing: we're not sending any notifications for a mention in a reflection. Will this be part of the next step?

igorlesnenko:
> One thing that's missing: we're not sending any notifications for a mention in a reflection. Will this be part of the next step?

Going to work on it as a next issue: https://github.com/ParabolInc/parabol/issues/9328

PR #9290, opened 2023-12-18
fix: Fix duplicate organizations for teams with auto join

### LINKED ISSUE #9229: Org showing up multiple times after auto join
@acressall shared the bug here: https://parabol.slack.com/archives/C042E8PGPD4/p1700530049859259

### AC

- [ ] Org only shows up once after auto joining multiple teams in the same org

Dschoordsch:
# Description

Fixes #9229 

If an organization had multiple teams with `autoJoin` enabled, multiple `OrganizationUser` objects would be created for a new user.

This fixes the creation of duplicate `OrganizationUser` objects but does not fix the existing duplicates. Ideally we would add a migration for this, but filtering over all `OrganizationUser` entries is expensive and only the Parabol org is affected.

## Testing scenarios

- [ ] create a qualified org (founder has their email verified) with 2 `autoJoin` teams and an active domain `example.com`
- [ ] create a new verified user with `test@example.com`
- [ ] see only 1 `OrganizationUser` is created
  ```graphql
  query User {
    user(email: "test@example.com") {
      organizationUsers {
        id
        userId
        orgId
      }
    }
  }
  ```

## Final checklist

### DIFF
--- a/packages/server/billing/helpers/adjustUserCount.ts
+++ b/packages/server/billing/helpers/adjustUserCount.ts
@@ -137,6 +137,9 @@ const auditEventTypeLookup = {
   [InvoiceItemType.UNPAUSE_USER]: 'activated'
 } as {[key in InvoiceItemType]: OrganizationUserAuditEventTypeEnum}
 
+/**
+ * Also adds the organization user if not present
+ */
 export default async function adjustUserCount(
   userId: string,
   orgInput: string | string[],
--- a/packages/server/billing/helpers/adjustUserCount.ts
+++ b/packages/server/billing/helpers/adjustUserCount.ts
@@ -97,10 +97,16 @@ const addUser = async (orgIds: string[], user: IUser, dataLoader: DataLoaderWork
       (oldOrganizationUser && oldOrganizationUser.newUserUntil) ||
       organization.periodEnd ||
       new Date()
-    return new OrganizationUser({orgId, userId, newUserUntil, tier: organization.tier})
+    return new OrganizationUser({
+      id: oldOrganizationUser?.id,
+      orgId,
+      userId,
+      newUserUntil,
+      tier: organization.tier
+    })
   })
 
-  await r.table('OrganizationUser').insert(docs).run()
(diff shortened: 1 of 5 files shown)


packages/server/dataloader/customLoaderMakers.ts
```diff
@@ -768,13 +768,16 @@ export const autoJoinTeamsByOrgId = (parent: RootDataLoader) => {
...
-        .selectFrom('Team')
-        .where('orgId', 'in', verifiedOrgIds)
-        .where('autoJoin', '=', true)
-        .where('isArchived', '!=', true)
-        .selectAll()
-        .execute()) as unknown as Team[]
+      const teams =
+        verifiedOrgIds.length === 0
+          ? []
+          : ((await pg
```
Dschoordsch, line 774: Found this while debugging, if there were no matching teams we would return a syntax error from PG instead of an empty array. Did not break anything as the caller used `.filter(isValid)`.

packages/server/billing/helpers/adjustUserCount.ts
```diff
@@ -97,10 +97,16 @@ const addUser = async (orgIds: string[], user: IUser, dataLoader: DataLoaderWork
       (oldOrganizationUser && oldOrganizationUser.newUserUntil) ||
       organization.periodEnd ||
       new Date()
-    return new OrganizationUser({orgId, userId, newUserUntil, tier: organization.tier})
+    return new OrganizationUser({
+      id: oldOrganizationUser?.id,
```
Dschoordsch, line 101: If there is an existing organization user, we overwrite it. Previously we would just insert additional ones.

nickoferrall (approved):
Thanks for the fix!

mattkrick:
@Dschoordsch it looks like this produced a flaky test result for me-- failed first test, passed on second try. 

https://github.com/ParabolInc/parabol/actions/runs/7278378730/job/19832437203. Should I make an issue for this?

Dschoordsch:
@mattkrick To me it looks like the test is working fine, but I didn't fix the underlying problem completely resulting in duplicate `OrganizationUser` objects being created in some cases.

PR #9334, opened 2024-01-12
feat(kudos): show snackbar when reflection with kudos created

igorlesnenko:
**How to test:**

- Start a retro and add a reflection with kudos
- See snackbar appearing:

image

### DIFF
--- a/packages/client/mutations/CreateReflectionMutation.ts
+++ b/packages/client/mutations/CreateReflectionMutation.ts
@@ -15,6 +15,7 @@ import SendClientSideEvent from '../utils/SendClientSideEvent'
 
 graphql`
   fragment CreateReflectionMutation_meeting on CreateReflectionPayload {
+    reflectionId
     reflectionGroup {
       meetingId
       sortOrder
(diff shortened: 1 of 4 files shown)


acressall:
I like this solution as the giver of kudos.

If 4 people gave me kudos during a retro, would I see 4 snackbars at the end? Or would we only show them in the notifications menu?

igorlesnenko:
> If 4 people gave me kudos during a retro, would I see 4 snackbars at the end? Or would we only show them in the notifications menu?

Currently, by the end of the retro the receiver will receive 4 different notifications and will see 4 snackbars if he is online

packages/client/mutations/CreateReflectionMutation.ts
```diff
@@ -58,7 +65,30 @@ const CreateReflectionMutation: StandardMutation<TCreateReflectionMutation> = (
...
     variables,
-    onCompleted,
+    onCompleted: (res, errors) => {
+      const draftKudoses = res.createReflection?.draftKudoses
+      if (draftKudoses && draftKudoses.length) {
+        const preferredNames = draftKudoses
+          .map((kudos) => kudos.receiverUser.preferredName)
+          .join(', ')
+        atmosphere.eventEmitter.emit('addSnackbar', {
+          key: 'youGaveKudos',
```
Dschoordsch, line 75: +1 If I give kudos in 2 different reflections within 5s, I will not see the second notification because they share the same key. This is an edge case unlikely to ever occur

Dschoordsch (approved):
Works

acressall:
> > If 4 people gave me kudos during a retro, would I see 4 snackbars at the end? Or would we only show them in the notifications menu?
> 
> Currently, by the end of the retro the receiver will receive 4 different notifications and will see 4 snackbars if he is online

Okay, let's leave it like this and we'll try it out and see how it feels and how often it happens

acressall (approved):

PR #9341, opened 2024-01-17
chore: embedder add pgvector (and fixes)

jordanh:
# Description

chore: add pgvector to development environment Postgres

Also:
   - fix: MailManagerDebug now always creates file paths under the O/S limit
   - fix: bump rethinkdb-ts (fixes development errors on my system)

## Demo

Loom: https://www.loom.com/share/a2291c714c48488eaef7302b7ed365ae?sid=fb726b37-3a51-4785-8d0e-f64c5ae01506

## Testing scenarios

```bash
$ yarn dev
```

## Final checklist


packages/server/postgres/extensions/pgvector/Makefile
```diff
@@ -0,0 +1,20 @@
+.PHONY: all clone build install
```
mattkrick, line 1: instead of building it locally, whatcha think about just using an image with pgvector already installed? https://github.com/ParabolInc/parabol/blob/ec5d15bd851430cc0856a1455ee8c741ec40b26a/.github/workflows/build.yml#L17

packages/server/email/MailManagerDebug.ts
```diff
@@ -9,9 +9,12 @@ export default class MailManagerDebug extends MailManager {
     Subject: ${subject}
     Body: ${body}`)
 
+    let toStr = to
+    if (Array.isArray(to)) toStr = to.join(',')
```
mattkrick, line 13: +1 nit:  we can avoid `let` here, or if you prefer avoid the variable all together by using the method call down below

```ts
const toStr = to.toString()
```

packages/server/postgres/Dockerfile
```diff
@@ -2,9 +2,14 @@ FROM postgres:15.4
 
 ADD extensions /extensions
 
-RUN apt-get update && apt-get install -y build-essential locales
+RUN apt-get update && apt-get install -y \
```
mattkrick, line 5: +1 I found adding --no-install-recommends shaves off some seconds from the install process. I like using v15 moving forward though!

https://github.com/ParabolInc/parabol/blob/ec5d15bd851430cc0856a1455ee8c741ec40b26a/packages/server/postgres/Dockerfile#L5

packages/server/postgres/extensions/pgvector/Makefile
```diff
@@ -0,0 +1,20 @@
...
+
+# Default target
+all: build
+
+# Target for handling the pgvector repository
+clone:
+	if [ ! -d "pgvector" ]; then \
+		git clone https://github.com/ankane/pgvector.git; \
```
mattkrick, line 9: we should install ankane's version on our GH test action otherwise any reference to vector will break

https://github.com/ParabolInc/parabol/blob/ec5d15bd851430cc0856a1455ee8c741ec40b26a/.github/workflows/build.yml#L14-L17

packages/server/postgres/extensions/pgvector/Makefile
jordanh: Why clone the pgvector repo?

packages/server/postgres/extensions/pgvector/Makefile
jordanh: kk, will change

packages/server/postgres/extensions/pgvector/Makefile
jordanh: Yeah, this is cleaner. Originally when I wrote the Makefile I was worried there might be some additional steps to get the extension working. I'll inline it instead

jordanh:
@mattkrick I think this is ready to merge now! After merging, I will begin refactor to divide migration into two pieces

packages/server/postgres/Dockerfile
mattkrick: no `--no-install-recommends`?
perfectly fine if not, should only affect first install

PR #9300, opened 2024-01-02
feat: embedder

jordanh:
# Description – DO NOT MERGE DEMO ONLY

Implements an architecture for generating embedding vectors in our app. Also implements a "Related Discussions" feature that demonstrates using these emebddings.

## Demo

Loom: https://www.loom.com/share/4fe4cc559f5b47f985a3611b0375e9f0?sid=fc15b480-7f8e-4faf-8731-0908a12d3ab6

PR #9372, opened 2024-01-21
chore: add embeddings table migration

jordanh:
# Description

After getting some feedback from @mattkrick, I've split the `EmbeddingsIndex` into two tables with two distinct concerns:

   * `EmbeddingsIndex` – contains all the meta information for which we've generated indexes
   * `EmbeddingsJobQueue` – contains a list of items pending embedding, or the reason(s) why creating the embedding failed

An outline of how these tables will be used:

   * For a particular embedding model, a service called `embedder` will...
      * ...add historical objects to be embedded to the `EmbeddingsIndex`
      * ...add items that lack the target model in the `EmbeddingsIndex`.`model` array to the `EmbeddingsJobQueue`
      * ...shift the state from `EmbeddingsJobQueue`.`queued` -> `embedding` when calculating an embedding for a particular model
      * ...remove the row from the `EmbeddingsJobQueue` if calculating the embedding was successful, and update the `EmbeddingsIndex`.`model` with the appropriate model name (A row will also be created  in `Embeddings_<MODEL>` that can be joined to `EmbeddingsIndex`)
         * else, ...shift the state from `queued` -> `failed` if calculating the embedding should fail
    
## Testing scenarios

- [ ] Migrate up!
- [ ] Migrate down!

## Final checklist


jordanh:
@mattkrick in general, I think separating the two concerns of the table over the original implementation in #9300 is better. 

The only drawback I can think of is, when adding items to the `EmbeddingsJobQueue` from the `EmbeddingsIndex` I'll have to query the `EmbeddingsIndex` and join the `EmbeddingsJobQueue` to filter out previously filtered jobs so we don't keep retrying failed embeddings in perpetuity

Definitely open to suggestions for any improvements/changes

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
```diff
@@ -0,0 +1,107 @@
...
+      EXECUTE 'CREATE INDEX IF NOT EXISTS "idx_EmbeddingsJobQueue_state" ON "EmbeddingsJobQueue"("state")';
+      EXECUTE 'CREATE INDEX IF NOT EXISTS "idx_EmbeddingsJobQueue_stateUpdatedAt" ON "EmbeddingsJobQueue"("stateUpdatedAt")';
+
+      EXECUTE 'CREATE TABLE IF NOT EXISTS "EmbeddingsIndex" (
+        "id" SERIAL PRIMARY KEY,
+        "objectType" "EmbeddingsObjectTypeEnum" NOT NULL,
+        "refId" VARCHAR(100),
+        UNIQUE("objectType", "refId"),
+        "refDateTime" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
+        "models" TEXT[],
```
mattkrick, line 41: +1 i don't have the whole layout baked in my head so I could be misunderstanding this, but denormalizing `models[]` here seems like it'd be difficult to query & keep updated. could we instead join on the `embedding_%` table to determine this, or is there another plan for keeping this in a text array?

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
```diff
@@ -0,0 +1,107 @@
...
+      EXECUTE 'CREATE INDEX IF NOT EXISTS "idx_EmbeddingsJobQueue_refId" ON "EmbeddingsJobQueue"("refId")';
+      EXECUTE 'CREATE INDEX IF NOT EXISTS "idx_EmbeddingsJobQueue_state" ON "EmbeddingsJobQueue"("state")';
+      EXECUTE 'CREATE INDEX IF NOT EXISTS "idx_EmbeddingsJobQueue_stateUpdatedAt" ON "EmbeddingsJobQueue"("stateUpdatedAt")';
+
+      EXECUTE 'CREATE TABLE IF NOT EXISTS "EmbeddingsIndex" (
+        "id" SERIAL PRIMARY KEY,
+        "objectType" "EmbeddingsObjectTypeEnum" NOT NULL,
+        "refId" VARCHAR(100),
+        UNIQUE("objectType", "refId"),
+        "refDateTime" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL,
```
mattkrick, line 40: -1 not sure what this is. if this is the time the row is created could we call this `createdAt`? 

if it's when the reference is created could we call it `refCreatedAt` or leave it normalized on the table related to `objectType`?

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
```diff
@@ -0,0 +1,107 @@
...
+        "models" TEXT[],
+        "teamId" VARCHAR(100) NOT NULL,
+        "orgId" VARCHAR(100) NOT NULL,
```
mattkrick, line 43: -1 this gets really tricky because a team can change orgs. either we need to ditch `orgId` or we need to augment the `moveTeamToOrg` mutation to handle this table, too.

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
```diff
@@ -0,0 +1,107 @@
...
+  DO $$
+  DECLARE
+      r RECORD;
+  BEGIN
+      FOR r IN SELECT tablename FROM pg_tables WHERE schemaname = 'public' AND tablename LIKE 'Embeddings_%'
+      LOOP
+          EXECUTE 'DROP TABLE IF EXISTS public.' || quote_ident(r.tablename) || ' CASCADE';
+      END LOOP;
+
+      EXECUTE 'DROP INDEX IF EXISTS "idx_EmbeddingsJobQueue_objectType"';
```
mattkrick, line 80: -1 no need to drop the indexes individually. dropping the table will suffice!

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
```diff
@@ -0,0 +1,107 @@
...
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+  DO $$
+  DECLARE
+      r RECORD;
+  BEGIN
+      FOR r IN SELECT tablename FROM pg_tables WHERE schemaname = 'public' AND tablename LIKE 'Embeddings_%'
+      LOOP
+          EXECUTE 'DROP TABLE IF EXISTS public.' || quote_ident(r.tablename) || ' CASCADE';
```
mattkrick, line 77: +1 this feels weird because the down migration is undoing something that wasn't done by the up migration. if tables are going to be created at runtime then should we build in functionality that cleans up old ones, too?

mattkrick (changes requested):

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
```diff
@@ -0,0 +1,107 @@
...
+          "stateMessage" TEXT,
+          "stateUpdatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL
+      )';
+
+      EXECUTE 'CREATE INDEX IF NOT EXISTS "idx_EmbeddingsJobQueue_objectType" ON "EmbeddingsJobQueue"("objectType")';
+      EXECUTE 'CREATE INDEX IF NOT EXISTS "idx_EmbeddingsJobQueue_refId" ON "EmbeddingsJobQueue"("refId")';
+      EXECUTE 'CREATE INDEX IF NOT EXISTS "idx_EmbeddingsJobQueue_state" ON "EmbeddingsJobQueue"("state")';
+      EXECUTE 'CREATE INDEX IF NOT EXISTS "idx_EmbeddingsJobQueue_stateUpdatedAt" ON "EmbeddingsJobQueue"("stateUpdatedAt")';
+
+      EXECUTE 'CREATE TABLE IF NOT EXISTS "EmbeddingsIndex" (
```
mattkrick, line 35: +1 this will sound very nitty, but could we call this `EmbeddingsMetadata` or something other than index so we don't confuse it with a DB index?

mattkrick:
> The only drawback I can think of is, when adding items to the EmbeddingsJobQueue from the EmbeddingsIndex I'll have to query the EmbeddingsIndex and join the EmbeddingsJobQueue to filter out previously filtered jobs so we don't keep retrying failed embeddings in perpetuity

I totally agree. Gave this some extra thought today. Since the job queue can be calculated by determining which objects don't have an embedding, it'd be great to leverage that into something simpler. From your PR, I think we can break down the functionality into the following:

- IndexBuilder (populates the EmbeddingsIndex table)
- QueueBuilder (a single worker per model/objectType)
- ObjectBuilder (turns a meeting into many discussion objects)
- LLM Endpoint (could be 1 or many behind a LB)
- Embedder to pop the queue & call the endpoint (could be 1 or many)

The real bottleneck here is going to be the LLM endpoint. There's no easy way to distribute the work of building the queue, but we can break it up by objectType+model to distribute the load a bit. 

Given those constraints, here's my proposal:
- On startup, Embedder queries `EmbeddingsIndex` to see if at least 1 record exists for each `model` for each `objectType`
- If an `objectType` for a `model` isn't found, it seeks a lock on redis key `objectBuilder:${objectType}:${model}`. If it's achieved, that embedder becomes IndexBuilder. If it's not achieved, it listens for redis to tell it the queue is created by subscribing to `objectBuilder:${objectType}:${model}:complete`
- IndexBuilder iterates through `EmbeddingsObjectTypeEnum`. 
  - It has a handler for each, e.g. for `retrospectiveDiscussionTopic` 
  - it batches through NewMeeting extracting the id of each meeting from `r.minval` to `r.now()` (new items ignored, they are handled by GQL Executor). 
  - It creates a row in `EmbeddingsIndex` for each item. on conflict it does nothing. 
- When complete, IndexBuilder becomes QueueBuilder  
  - it left joins to `Embeddings_%` and filters for null to create a list of objects without embeddings. This job queue is pushed to a redis stream via `xadd`
  - On crash, an embedder can skip IndexBuilder and become QueueBuilder if prompted (Out of scope, but this supports redis without persistence if we want to build it later)
  - it publishes `objectBuilder:${objectType}:${model}:complete` to alert all Embedders to begin work. Those embedders do not start work until they exit QueueBuilder mode.
- finally, Embedders subscribe to the queue via `.xreadgroup()`. They use xautoclaim to support failover. they ack when the job is done. they call xclaim if a task is taking longer than expected. see https://cschleiden.dev/blog/2022-04-08-task-queue-with-redis  
 - Once Embedder has an item, it acts as ObjectBuilder & computes embedText & writes it to the DB table
 - After computing embedText, it sends it to the LLM endpoint to get embeddings. on return, it writes that to the Embeddings_% table
 - Finally, it calls ACK to redis to report the job as complete

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
```diff
@@ -0,0 +1,107 @@
...
+
+      EXECUTE 'CREATE TABLE IF NOT EXISTS "EmbeddingsJobQueue" (
+          "id" SERIAL PRIMARY KEY,
+          "objectType" "EmbeddingsObjectTypeEnum" NOT NULL,
+          "refId" VARCHAR(100) NOT NULL,
+          "model" TEXT NOT NULL,
+          UNIQUE("objectType", "refId", "model"),
+          "state" "EmbeddingsStateEnum" DEFAULT ''queued'' NOT NULL,
+          "stateMessage" TEXT,
+          "stateUpdatedAt" TIMESTAMP WITH TIME ZONE DEFAULT CURRENT_TIMESTAMP NOT NULL
```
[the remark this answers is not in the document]
jordanh: I like this change. I will make it

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
```diff
@@ -0,0 +1,107 @@
...
+import connectRethinkDB from '../../database/connectRethinkDB'
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+  DO $$
+  BEGIN
+      IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'EmbeddingsObjectTypeEnum') THEN
+          EXECUTE 'CREATE TYPE "EmbeddingsObjectTypeEnum" AS ENUM (''retrospectiveDiscussionTopic'')';
```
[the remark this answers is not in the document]
jordanh: It's an artifact of me wrapping these statements in a transaction. If we don't mind them being outside a transaction I can get rid of the EXECUTE. Given it's a migration, it would be just as safe...

What's your preference?

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
```diff
@@ -0,0 +1,107 @@
...
+      END IF;
+      IF NOT EXISTS (SELECT 1 FROM pg_type WHERE typname = 'EmbeddingsStateEnum') THEN
+          EXECUTE 'CREATE TYPE "EmbeddingsStateEnum" AS ENUM (''queued'', ''embedding'', ''failed'')';
```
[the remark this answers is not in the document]
jordanh: I was thinking that if it were successful it be purged altogether from the `EmbeddingsJobQueue` and the single source of truth as to whether an EmbeddingsIndex row had been embedded successfully was whether or not it appeared in the `EmbeddingsIndex` with the models array set appropriately. I would be ok to log successful jobs in the JobQueue tho. What is your preference?

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
jordanh: >+1 i don't have the whole layout baked in my head so I could be misunderstanding this, but denormalizing models[] here seems like it'd be difficult to query & keep updated. could we instead join on the embedding_% table to determine this, or is there another plan for keeping this in a text array?

It's actually not super hard. I figured it'd be worse to join N model tables. Here's what updating this array (as a set) looks like:  https://github.com/ParabolInc/parabol/blob/a0272c78e5c471e3b29028197656f2c8c9d4c739/packages/embedder/embedder.ts#L145

Which way would you like to roll?

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
jordanh: The intent is to have a time on the index for "what time the embedded object represents". Use case would be filtering by a date range. This intends to denormalize that contextual time in the index. Job story:

When I want to search for discussions that only occurred between two dates, I want a set of filter controls that return only the results I am interested in.

It's debatable whether we need this now. I'd be ok to drop it until we need it.

What say you?

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
jordanh: blerg...this is true. I would be ok dropping this an just joining through the table instead. I will make this update.

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
jordanh: True? Ok. Then I will update this

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
jordanh: Ok. I will strike this too.

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
jordanh: I like that. I can update this.

jordanh:
What you propose is strikingly similar to what I've written, except for a few details. Let's get work to get this PR merged, and then I'll make some updates to the service I wrote to match new nomenclature...and we can continue. Sound ok?

Awaiting your comments on some of the +1s

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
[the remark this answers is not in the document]
mattkrick: ahhh that explains it! preference is to remove them since other folks will likely look to this file to do something similar in the future, but if time is an issue a comment saying EXECUTE isn't necessary should suffice

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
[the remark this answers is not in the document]
mattkrick: i like removing them after they're complete! As long as we fill it successfully then we don't need to differentiated between `success` and `not-yet-queued`

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
mattkrick: gotcha! yeah if we only add to it that shouldn't be too bad. if we need to query against it we can always do a join then if performance isn't keeping up

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
mattkrick: ah! that's a super useful feature. so this isn't necessarily when the embedding is created, but when the reference was last updated? if that entity is mutable & gets updated, then we update this column + the embed text? if that's the case, i might suggest something like `refUpdatedAt` to make a little more clear, but i really like having the date here

mattkrick:
sounds great! Yep, I think we've largely converged on the embedder layout, i still gotta do a deeper dive on your PR to make sure i understand everything. redis vs. pg's `SKIP LOCKED` is a good area for convo, i haven't read enough to make an informed decision.

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
jordanh: It's an indexed query (via the GIN index) – I think it's going to be good. But let's see!

packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts
jordanh: Ok, will do.

jordanh:
> sounds great! Yep, I think we've largely converged on the embedder layout, i still gotta do a deeper dive on your PR to make sure i understand everything. redis vs. pg's SKIP LOCKED is a good area for convo, i haven't read enough to make an informed decision.

We can definitely kick this conversation to the next PR :)

jordanh:
@mattkrick, I've made all the changes we discussed.

Would you mind giving hits a migrate up/down test? I'd do it, but I don't want to lose all the data I've calculated :)

jordanh:
@dbumblis-parabol this is the PR that is going to land soon, FYI

dbumblis-parabol:
There may be a challenge in the near future where we cannot control pgvector being installed in the ironbank container image used for postgres. Is there any way that the migration itself can be gated? Or, how could we address a situation where the postgresql db does not have pgvector installed?

jordanh:
> Or, how could we address a situation where the postgresql db does not have pgvector installed?

@mattkrick now here's an interesting case. To rephrase what @dbumblis-parabol is saying:

- If we merge this and cut a new IronBank image...
- ...when the migration runs, it may fail because pgvector may not be available in the destination environment

Blarg. That would stink.

What would be a reasonable pattern here? How could we "soft require" pgvecctor?

Dschoordsch:
Do we want to make this all optional, i.e. put the vector data in a different database? Then we could not easily merge with the rest of the data, but I don't see any immediate negative consequences through this.
If it would be a separate DB, it would be easier to gate migrations and ensure fewer accidents if we want to have the option of not using AI in certain offerings.

mattkrick:
if usage of pgvector is dependent on an environment & that decision could change over time then i think an env var is probably the cleanest solution.

```ts
// .env
USE_PGVECTOR=true

// preDeploy
if (process.env.USE_PGVECTOR === 'true') {
  const enabled = await pg.query(`SELECT 1 FROM pg_extension WHERE extname = 'pgvector'`).executeAndTakeFirst()
  if (!enabled) {
    await pg.query(`CREATE EXTENSION vector;`).execute()
  }
}
```

I'll have to hear more about @Dschoordsch's idea. A 2nd DB would mean managing a 2nd pool of PG connections, so I'd want to make sure that we're getting something worthwhile for the complexity.

mattkrick (approved):

jordanh:
```javascript
// .env
USE_PGVECTOR=true

// preDeploy
if (process.env.USE_PGVECTOR === 'true') {
  const enabled = await pg.query(`SELECT 1 FROM pg_extension WHERE extname = 'pgvector'`).executeAndTakeFirst()
  if (!enabled) {
    await pg.query(`CREATE EXTENSION vector;`).execute()
  }
}
```

@mattkrick I like this.

What to do about the migration in `packages/server/postgres/migrations/1703031300000_addEmbeddingTables.ts` tho? The up migration will fail if it is run and the `vector` extension isn't available.

Here's a bad option I can think of:

   1. Rewrite the migration to skip adding the columns that rely on the `vector` type if the extension is not loaded
   2. Update the preDeploy script to add the extension and the column if the `USE_PGVECTOR` env var is `true`

Do you have a better idea than this? If not, I can go ahead and make those changes and we can get this puppy merged

mattkrick:
Interesting problem because 2 different envs will have 2 different DB schemas. this could get pretty hairy!

i like your solution. the alternative i see is having the migration adding the column of type ` bytea` and then changing that column type to `vector` in predeploy. since the table won't have rows yet, changing the column type won't be expensive. It will be a _little_ closer to the truth because the column will exist as the wrong type vs. not existing at all.

I'm OK with either!

jordanh:
@mattkrick and I [discussed via Slack](https://parabol.slack.com/archives/C836NA350/p1707175986660069) and since the migration relies on no `vector` column data types (they are all added by the embedder process dynamically) updating the `preDeploy` script to create the extension if an env var is set will be the way to go for now.

I've create a bool env var called `POSTGRES_USE_PGVECTOR` to gate this behavior

jordanh:
@mattkrick I think this is GTG now

Dschoordsch:
> I'll have to hear more about @Dschoordsch's idea. A 2nd DB would mean managing a 2nd pool of PG connections, so I'd want to make sure that we're getting something worthwhile for the complexity.

That was just an idea I had during our call yesterday. I thought it would be easier to skip the entire db and embedding module if the environment is not present. In that case our normal migrations wouldn't depend on environment variables, despite connection strings. I think it would have no technical benefit, but it might keep developers aware that they have to check for the absence of that data.
It also introduces more complexity in the migrations as we might (probably won't) run into the same ordering issue we had between rethink and postgres.
In both cases the logic to generate and consume the embeddings needs to be guarded by an environment variable, so there is no benefit here.

So I think just changing the type and not populating the data is way easier. We could run into issues when querying with the non-existent type and not catch it before it's deployed to ironbank users.

jordanh:
@Dschoordsch I think as this PR now stands we have the best of all worlds:

   * The `POSTGRES_USE_PGVECTOR` env var gates whether or not the `preDeploy` will try to enable the pgvector extension

And in the next PR:

   * The `embedder`'s `AI_EMBEDDER_ENABLED` env var will gate whether or not the embedder will even run, which  determines whether or not the tables that contain the `vector` type will even be created

This should let us run flexibly in our SaaS or in a self-hosted model with or without pgvector

scripts/toolboxSrc/pgEnsureExtensions.ts
```diff
@@ -0,0 +1,9 @@
+import {sql} from 'kysely'
+import getKysely from '../../packages/server/postgres/getKysely'
+
+export default async () => {
+  console.log('🔩 Postgres Extension Checks Started')
+  const pg = getKysely()
+  await sql`CREATE EXTENSION IF NOT EXISTS "vector";`.execute(pg)
```
mattkrick, line 7: is this where we should use the env var?
i suppose we could also just do a try/catch & swallow the error if we want to avoid using an env var? the env var might be useful at runtime though when deciding if we should use embeddings or not. your call!

jordanh: Wait. Lolololo did I seriously not use the ENV var? Ha! Good catch. I will fix

jordanh:
Ooookay @mattkrick, now I think this is ready :)

scripts/toolboxSrc/pgEnsureExtensions.ts
```diff
@@ -0,0 +1,14 @@
+import {sql} from 'kysely'
+import getKysely from '../../packages/server/postgres/getKysely'
+
+export default async () => {
+  console.log('🔩 Postgres Extension Checks Started')
+  if (process.env.POSTGRES_USE_PGVECTOR) {
```
mattkrick, line 6: you'll have to write `=== 'true'` since our dotenv parser does not parse the string `"true"` to the binary type `true`.

jordanh: Ugh – color me embarrassed. Coming right up sir

PR #9376, opened 2024-01-22
fix: render activity library tooltip

nickoferrall:
In production, the tooltip with the prompt details isn't showing up when you hover over the activity library card.

### To test

- [ ] Hover over the activity library card and see that the tooltip shows up


packages/client/components/ActivityLibrary/ActivityGrid.tsx
```diff
@@ -31,6 +31,7 @@ const ActivityGrid = ({templates, selectedCategory}: ActivityGridProps) => {
               theme={CATEGORY_THEMES[template.category as CategoryID]}
               title={template.name}
               type={template.type}
+              templateRef={template}
```
igorlesnenko, line 34: feels 2 separate components (one with require prop) would be good to avoid cases like this

igorlesnenko (approved):

PR #9377, opened 2024-01-23
feat: add slack notification for mention and kudos in reflections

igorlesnenko:
image

Enabling slack notifications for mentions and kudos in retro reflections

**How to test:**
- You need 2 users in one team for testing
- Add slack integration for the second user
- Start meeting as the first user but do not join as the second user
- As the first user add various reflections with kudos and mentions
- Complete the meeting
- See the second user received slack notifications for mentions and kudos
- Test with non-anonymous reflections
- Smoke test that standup slack notifications still working


Dschoordsch (approved):
Works

tghanken (approved):

PR #9381, opened 2024-01-25
fix: use native dotenv handling

### LINKED ISSUE #8788: Node v20.6.0 Upgrade - Remove dotenv
Node v20.6 adds native support for `.env` files. This should make our DX a bit cleaner.
https://philna.sh/blog/2023/09/05/nodejs-supports-dotenv/

AC
- [ ] dotenv package is gone

mattkrick:
# Description

fix #8788

loading an env file is now possible as a node flag in node v20.7.0.
that means the .env gets loaded before any JS even gets parsed, that's nice because in the past we always had to make sure that it was the very first thing that ever got loaded. this should make building new tooling a lot simpler.

.env files aren't used in prod so no changes should be necessary for PPMIs.


rafaelromcar-parabol:
.env files are used in production in our SaaS actually.

PPMIs use the .env files to mount them as env vars of the container, but in our SaaS the application is receiving the .env file directly and using it

mattkrick:
right you are! should've said .env files aren't used _by the application_ in prod.

rafaelromcar-parabol:
> right you are! should've said .env files aren't used _by the application_ in prod.

I'm not sure if I'm getting it right, but let me explain how the .env is used right now in Production (and Staging and any other Kubernetes based environment).

The container has `/home/node/parabol` as [workdir](https://github.com/ParabolInc/parabol/blob/master/docker/parabol-ubi/docker-build/dockerfiles/basic.dockerfile#L24C1-L24C25) and the content of that folder is:
```
~/parabol$ ls -a
.  ..  .env  .env.example  build  dist  self-hosted  yarn.lock
```
**The .env file IS used by the application in Production.** 

Don't get me wrong, we can change that and pass those parameters in a different way, but as of today, **the application is*using a .env file**. That file has only non-secret parameters and it is the file you all edit in our Gitlab repository.

With the work planned in the following weeks around config and secret management we might stop using .env files, but today we are using it.

mattkrick:
oh! i was thinking k8s set the env vars directly in the shell & didn't write a .env.

to support this PR we'd have to replace every call to `node` with `node --env-file .env`. How big of a pain would that be? If it's too much I'm happy to close this

rafaelromcar-parabol:
> oh! i was thinking k8s set the env vars directly in the shell & didn't write a .env.
> 
> to support this PR we'd have to replace every call to `node` with `node --env-file .env`. How big of a pain would that be? If it's too much I'm happy to close this

It isn't too much I think, but as we have pre-deploy as a pre-upgrade and pre-install hook, it isn't that easy either. If you think this is important in any way, I can spend time on the topic. If not, I would say just close it until we have to do this.

mattkrick:
i'll close for now, we can revisit when we standardize how we deploy & can update all calls to `node` at once

PR #9401, opened 2024-02-05
chore: no force-push to prod

mattkrick:
# Description

currently pushes to prod have to be force pushed, but i want to use the UI to push to prod.
this should solve for that, i hope?


rafaelromcar-parabol:
Wouldn't that addition be a risk in case of hotfix applied directly to production? 🤔 Wouldn't that reset production to master and the changes applied to production would get lost then?

mattkrick:
pushing to master just calls the gitlab JOB ID that is in the PR. if the JOB ID is missing or invalid nothing will happen.

if we do a hotfix, it should be based off of production, otherwise the release-please will include all the stuff that's already on master

PR #9413, opened 2024-02-08
chore: fix misleading `isLead` field name on `Team`

Dschoordsch:
The field indicates whether the viewer is the lead, but when used in a query for a different user, the result could be read wrong.

### DIFF
--- a/packages/server/graphql/types/Team.ts
+++ b/packages/server/graphql/types/Team.ts
@@ -142,7 +142,7 @@ const Team: GraphQLObjectType = new GraphQLObjectType<ITeam, GQLContext>({
         return dataLoader.get('teamInvitationsByTeamId').load(teamId)
       }
     },
-    isLead: {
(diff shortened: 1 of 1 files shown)


mattkrick:
can you remove `isLead` from `Team.graphql`? That should make tsc throw a few errors so you can track the client-side dependencies like: https://github.com/ParabolInc/parabol/blob/fb05fddc99a0ece301e5c072becd52666399c8a4/packages/client/modules/teamDashboard/components/ManageTeam/ManageTeamList.tsx#L39

igorlesnenko (changes requested):
I see it is used on frontend for example here 

https://github.com/ParabolInc/parabol/blob/012ca77c5c050df9bd3711332d21d40026018284/packages/client/components/ReviewRequestToJoinOrgModal.tsx#L24

mattkrick (approved):

PR #9415, opened 2024-02-09
fix: replace lone surrogates in draft-js content

mattkrick:
# Description

Fixed by calling `toWellFormed()` on the string. It still isn't a part of typescript's lib though 😠 
https://github.com/microsoft/TypeScript/issues/55543

We can't validate the entire incoming payload because the lone surrogate is inside quotes inside the string, e.g. the validator is looking for `\udc4d` but it only sees `"\udc4d"` so it doesn't fix it. This makes sense because if someone wanted to say "I have an issue anytime the character \udc4d comes up" we wouldn't want that to be invalid.

I also bumped all the typescript packages & libs thinking that a newer version would have `toWellFormed` in it. Unrelated, but it's a small PR so i left it. 

## Demo

## Testing scenarios

- [ ] start a retro, type in `:+` hit enter to create the reflection, click the thumbs up emoji in the menu, hit backspace to remove the lower surrogate so the emoji turns into a replacement char. submit it
- [ ] backup the reflections table


Dschoordsch (approved):

Dschoordsch:
@mattkrick any reason not to merge this now?

Issue #9374, opened 2024-01-22
limit invites sent from spammers


Dschoordsch:
See [this Slack message](https://parabol.slack.com/archives/C836NA350/p1705690943028989) 🔒
To avoid limiting legit invites, we lifted some of the limitations we had put in place previously for sending invites without having some accepted. One suggestion was to limit unaccepted invites to freemail accounts to 100.

PR #9416, opened 2024-02-09
fix: limit invites from spammers

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9374

If the user has sent 100 or more invites and none have been accepted, prevent them from sending more invites. 

We could be stricter and set a lower limit or prevent them from sending invites even if they have more than one accepted invite if there are many unaccepted invites, but I'm more concerned about preventing real users from sending invites. 

### To test

- [ ] If there are 100 unaccepted invites, no more invites are sent

### DIFF
--- a/packages/server/graphql/mutations/helpers/inviteToTeamHelper.ts
+++ b/packages/server/graphql/mutations/helpers/inviteToTeamHelper.ts
@@ -48,7 +48,7 @@ const inviteToTeamHelper = async (
   ])
   const accepted = total - pending
   // if no one has accepted one of their 100+ invites, don't trust them
-  if (accepted === 0 && total + pending >= 100) {
+  if (accepted === 0 && total + invitees.length >= 100) {
     return standardError(new Error('Exceeded unaccepted invitation limit'), {userId: viewerId})
   }
 
--- a/packages/server/graphql/mutations/helpers/inviteToTeamHelper.ts
+++ b/packages/server/graphql/mutations/helpers/inviteToTeamHelper.ts
@@ -37,6 +37,21 @@ const inviteToTeamHelper = async (
   const operationId = dataLoader.share()
   const subOptions = {mutatorId, operationId}
 
+  const [total, pending] = await Promise.all([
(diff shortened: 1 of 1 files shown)


packages/server/graphql/mutations/helpers/inviteToTeamHelper.ts
```diff
@@ -59,6 +59,21 @@ const inviteToTeamHelper = async (
...
+      .filter({acceptedAt: null})
+      .count()
+      .run()
+  ])
+  const accepted = total - pending
+  // if no one has accepted one of their 100+ invites, don't trust them
+  if (accepted === 0 && total >= 100) {
+    return standardError(new Error('Exceeded unaccepted invitation limit'), {userId: viewerId})
+  }
+
```
Dschoordsch, line 76: -1 We need to do this before we check whether or not an email is valid, otherwise this gives spammers still some utility

Dschoordsch (changes requested):

Dschoordsch (changes requested):
I just tested to send 150 invites straight away and it did not complain. I think we need to check if the total + new invites would be > 100 and fail.

Dschoordsch (changes requested):
Does not work for me:
* create a new team
* from the team dash invite >100 emails

Also we allow invites on team/org creation which are not limited right now.

nickoferrall:
Updated! Here's a Loom demo: https://www.loom.com/share/f091ff7ec2154221bf4b70a9b5b364b8

Emails won't be sent to users with 0 accepted invites and with more than 100 pending invites. If they create a team or org and there's an error sending the emails, they don't get feedback, but the emails won't be sent. You can see in the terminal that there's an error message. Not the ideal UX, but this was a design decision we originally made. 

We could make this stricter, e.g. if there are fewer than 10 accepted invites, you can't invite more than 100 people at a time, but it's probably best to be too lenient than too strict.

Dschoordsch (approved):
Works now!

Issue #9340, opened 2024-01-15
Add subscription to AI Template Search


nickoferrall:
The OpenAI response can take a long time in: https://github.com/ParabolInc/parabol/pull/9223

We should use a subscription to handle timeouts. 

### AC

- [ ] Implement subscription in `GetTemplateSuggestionMutation`

PR #9421, opened 2024-02-13
feat: speed up ai search

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9340

GPT-3.5 turbo now has a large enough context window that it can accept the list of available templates.

I get similar results to GPT-4, and the response time is dramatically faster, so I think we should go with 3.5-turbo.

### To test

- [ ] Add the `aiTemplate` org feature flag
- [ ] Ask the AI Search input questions like "what's a template that helps us celebrate small wins"
- [ ] See that the results are similar to the results we get with the original search and that we no longer get timeouts

### DIFF
--- a/packages/server/utils/OpenAIServerManager.ts
+++ b/packages/server/utils/OpenAIServerManager.ts
@@ -1,4 +1,5 @@
 import OpenAI from 'openai'
+import JSON5 from 'json5'
 import sendToSentry from './sendToSentry'
 import Reflection from '../database/types/Reflection'
 import {ModifyType} from '../graphql/public/resolverTypes'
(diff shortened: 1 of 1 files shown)


packages/server/utils/OpenAIServerManager.ts
```diff
@@ -221,9 +221,11 @@ class OpenAIServerManager {
       })
 
       const templateResponse = (response.choices[0]?.message?.content?.trim() as string) ?? null
-      return JSON.parse(templateResponse) as AITemplateSuggestion
+      const parsedResponse = JSON5.parse(templateResponse)
```
igorlesnenko, line 224: Just curious, why replaced with JSON5?

igorlesnenko (approved):
Works!

packages/server/utils/OpenAIServerManager.ts
nickoferrall: 3.5-turbo doesn't have the JSON parse feature that GPT-4 has, so we need to parse it ourselves

Issue #9395, opened 2024-02-05
Improve error messages in Activity Library


mwermuth:
This came up when we wanted to create more custom templates in our Parabol Org. We have 20 custom templates in the team already, making it impossible to create a new one. In the Activity Library no error or alert message is shown, so the user won't know what is going on.
There is also no good possibility to manage custom templates and delete old ones.

If you revert back to the legacy dialog the message shows.
![image](https://github.com/ParabolInc/parabol/assets/1415538/531de259-7553-404a-94b1-0c1b845c9e95)

## Issue - Bug

- **How serious is the bug?** P2

### Acceptance Criteria (optional)
Users can:
   - see an error/alert message in the Activity Library when there are too many custom templates.
   - (if possible) delete old custom templates in the Activity Library
   - optional: relaxed template limit from 20 templates per team to 100

mwermuth:
Slack Context:

- https://parabol.slack.com/archives/C4JAUUZ9P/p1705939368020659
- https://parabol.slack.com/archives/C4JAUUZ9P/p1707148883612619

jordanh:
While we're in the neighborhood, let's also relax the limit. How about 100?

PR #9424, opened 2024-02-14
feat: remove team template limit

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9395

See Slack discussion here: https://parabol.slack.com/archives/C836NA350/p1707760605806529

### To test

- [ ] Can create and clone templates, regardless of how many you currently have


Dschoordsch (approved):
Looks good, did not test.

PR #9429, opened 2024-02-15
fix: fix prompt to join org when login with saml

### LINKED ISSUE #9411: Prompt to join org does not work with SSO
- Now we require email to be verified to use prompt to join org
- For `loginWithSAML` we don't create `identities` array, which we use to determine if email is verified
- This cause a bug

Solutions:
- Quick solution is to use this workaround like we used for autojoin https://github.com/ParabolInc/parabol/pull/9189
- Good solution, add identities array https://github.com/ParabolInc/parabol/issues/9202

igorlesnenko:
fixes #9411

**How to test**

- Few places related to `autoJoin` have changed, test that `autoJoin` still works correctly
- Test that prompt to join org with SAML works now

### DIFF
--- a/packages/server/dataloader/__tests__/isOrgVerified.test.ts
+++ b/packages/server/dataloader/__tests__/isOrgVerified.test.ts
@@ -12,9 +12,7 @@ jest.mock('../../utils/isUserVerified')
 getRethink.mockImplementation(() => {
   return r
 })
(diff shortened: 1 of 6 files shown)


jordanh:
Since this has been open since February, I'm going to close this PR for now. If we have tension around Prompt to Join org in the future, we can revisit the work here...

PR #9444, opened 2024-02-21
chore: Associate logs with traces

Dschoordsch:
Add trace information to log output for server side log statements. This does not include logging from code exclusively used for debugging, deploying or development.
The environment variable [`DD_LOGS_INJECTION`](https://docs.datadoghq.com/tracing/other_telemetry/connect_logs_and_traces/nodejs/) is used to enable or disable this behaviour.

## Testing scenarios

- run the app and trigger a log output (for example throw in MailManagerDebug and send a summary)

## Final checklist

### DIFF
--- a/packages/server/utils/Logger.ts
+++ b/packages/server/utils/Logger.ts
@@ -11,7 +11,7 @@ const LogFun = {
 } satisfies Record<LogLevel, Console[keyof Console]>
 
 function trace(level: LogLevel, message: any, ...optionalParameters: any[]) {
-  if (!process.env.DD_LOGS_INJECTION) {
+  if (process.env.DD_LOGS_INJECTION !== 'true') {
     return LogFun[level](message, ...optionalParameters)
   }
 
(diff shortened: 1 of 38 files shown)


packages/server/utils/Logger.ts
```diff
@@ -0,0 +1,35 @@
...
+type LogLevel = 'error' | 'warn' | 'info' | 'debug'
+const LogFun = {
+  error: console.error,
+  warn: console.warn,
+  info: console.info,
+  debug: console.debug
+} satisfies Record<LogLevel, Console[keyof Console]>
+
+function trace(level: LogLevel, message: any, ...optionalParameters: any[]) {
+  if (!process.env.DD_LOGS_INJECTION) {
```
mattkrick, line 14: -1 env vars are strings, should we instead say `!== 'true'`?

Dschoordsch: Yes, I wouldn't have checked an env if there wasn't already one defined and set in our environments.

nickoferrall (approved):
LGTM!

mattkrick (approved):

PR #9445, opened 2024-02-21
feat(standalone-deployment): Standalone host deployment improved and documented

rafaelromcar-parabol:
# Description

- Improves the Docker compose stack so it runs pre-deploy and then the rest of the components.
- Documents how to deploy and upgrade the application in the stack.


docker/parabol-ubi/docker-host-st/docker-compose.yaml
```diff
@@ -1,51 +1,144 @@
...
+        condition: service_started
+      postgres:
+        condition: service_healthy
+      redis:
+        condition: service_healthy
     networks:
       - parabol-network
-  app:
-    image: #image:tag
+  # chronos:
```
mattkrick, line 80: +1 could we use profiles instead of commenting this out? https://docs.docker.com/compose/profiles/ then the client could just start 1 or many profiles

rafaelromcar-parabol: Nice that. I've never seen that. I'm a newbie on Docker compose 😸 I'll be checking it. Thanks!

mattkrick:
it's the only way for other containers to communicate with it

> Important note: you must set listen_addresses = '*'so that other containers will be able to access postgres.

sauce: https://hub.docker.com/_/postgres

rafaelromcar-parabol:
@mattkrick I think that only applies if you are using a custom configuration. Please check the "important note" that is inside the section Custom configuration and, inside it, below the section "Use a custom config file.". I think that what happens is that IF you are using a custom configuration file you MUST add `listen_addresses='*'` to it. If you are NOT using a custom configuration file, it works as expected.

In that docker-compose.yaml, both PGAdmin and Parabol are communicating with Postgres without any trouble and I'm not using any custom configuration or passing the `listen_addresses='*'`.

As we are producing here something that is simple and works for the basic cases, I would just remove that almost empty custom postgresql.conf  and any mention to it.

rafaelromcar-parabol:
Actually, if you get into a postgres:15.4 running without any custom configuration and check the file `/usr/share/postgresql/postgresql.conf.sample` you will find the `listen_addresses='*'`. It seems to be the default behavior. But it looks like if you use a custom file, you must add it explicitly.

rafaelromcar-parabol:
- Docker compose profiles added for all: databases, parabol, database-debug and chronos.
- Readme improved with all the how-to run the stack with the profiles.
- Postgres available for all containers (checked with pgadmin and with the application).

WDYT @dbumblis-parabol @mattkrick ? Could you give it a try? Literally, you just need to download one of our containers (or build it) and follow the super small README 😸

mattkrick:
that's awesome that it's in there by default now! yeah, nuke the custom stuff then

docker/parabol-ubi/docker-host-st/README.md
```diff
@@ -1,15 +1,42 @@
...
 
-To run the Parabol UBI in single tenant mode (e.g. simple docker-compose on a docker host).
+To run Parabol in single tenant mode (e.g. simple docker-compose on a docker host).
 
 1. Build your Parabol UBI using instructions in `docker/ubi/docker-build/README.md`
 2. Create a working `.env` from `.env.example`
 3. Update docker-compose.yaml `image: #image:tag` with your built image tag from `step (1.)`
-4. Run `docker-compose up -d` to deploy the local stack. You can run `docker-compose down` to terminate the local stack
-5. Check logs via command `docker logs <name>-app-1` and wait for the following output to appear
+4. Run `docker-compose --profile databases --profile parabol up -d` to deploy the local stack. You can run `docker-compos --profile databases --profile parabol down` to terminate the local stack
```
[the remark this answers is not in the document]
rafaelromcar-parabol: Fixed! Bad copy, paste and edit 😸

mattkrick (approved):

dbumblis-parabol (approved):

PR #9447, opened 2024-02-22
chore(docker-build): simplify the docker build process and reduce docker image size

rafaelromcar-parabol:
# Description

- [x] Reduces size of the image. Down from 795 MB to 445 MB.
- [x] Reduces build time. Down form more than a minute to around 3.5s + the time to download node:VERSION-bookworm-slim (around 201 MB).
- [x] Simplifies the build-env files to have only what is required.
- [ ] Reduce the number of Dockefiles to only one.


docker/parabol-ubi/docker-build/dockerfiles/basic.dockerfile
```diff
@@ -0,0 +1,28 @@
+ARG _NODE_VERSION=${_NODE_VERSION}
+FROM node:${_NODE_VERSION}-bookworm-slim as base
+
+ENV NPM_CONFIG_PREFIX=/home/node/.npm-global
+ENV PORT=3000
+
+# RUN apt update -y && \
```
rafaelromcar-parabol, line 7: @adaniels-parabol you added systemtap in a [PR](https://github.com/ParabolInc/parabol/pull/9241), but there are no details about why it was added. Could you elaborate a bit more and tell us if we can remove it?

docker/parabol-ubi/docker-build/environments/basic-env
```diff
@@ -0,0 +1,17 @@
...
+PORT='3000'
+# Database configurations must be the same used in the docker-build.yml Github workflow
+POSTGRES_PASSWORD='temppassword'
+POSTGRES_USER='tempuser'
+POSTGRES_DB='tempdb'
+POSTGRES_HOST='localhost'
+POSTGRES_PORT='5432'
+REDIS_URL='redis://localhost:6379'
+RETHINKDB_SSL=''
+RETHINKDB_URL='rethinkdb://localhost:28015/buildDB'
```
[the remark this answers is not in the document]
mattkrick: that's probably ok

rafaelromcar-parabol:
One side effect of this PR: as we are using only one layer and just adding our files, a big part of the image will be cached from now on, making our deployment go faster as well 😸

mattkrick (approved):

docker/parabol-ubi/docker-build/dockerfiles/basic.dockerfile
adaniels-parabol: It was used for some of the security analysis (inspection scripts) and for some of the security scripts.

rafaelromcar-parabol:
We cannot reduce the number of dockerfiles to one and we cannot remove all the security scripts until all our instances (PPMIs included) are using the new version. Thus, I'm going to merge this PR with the improvements we have added so far and let the rest for the, I hope, not so far future.

PR #9455, opened 2024-02-26
feat: support env-defined saml issuer for PPMIs

### LINKED ISSUE #9454: Support using SAML login for all domains
From P1 team: domains change so frequently, it's a pain to add new domains to the SAML config.
This is a feature request that is just for PPMIs.

AC
- [ ] if we're in an enterprise environment where SAML is the only way to login, then open a SAML window, regardless of domain.

mattkrick:
# Description

fix #9454

Supports single tenant SAML:
- admins can set `env.SAML_ISSUER` & override what parabol calls itself to the IdP
- admins can use a single SAML record for the entire instance if SSO is the only way to log in to the app & a SAML record already exists. To set up SAML, they can temporarily enable email/password auth & then after it's set up successfully they can disable that.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -14,6 +14,8 @@ SOCKET_PORT='3001'
 # INVITATION_SHORTLINK='example.com'
 # If true, all new orgs will default to being enterprise tier. Use for PPMIs
 # IS_ENTERPRISE=false
+# PPMI single tenant use only. Will set the SAML issuer to this value.
+# SAML_ISSUER=''
 
 # AUTHENTICATION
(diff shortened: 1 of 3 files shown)


dbumblis-parabol:
Does the issuer need any kind of specific syntax / format? Or will it take whatever is defined by the var?

mattkrick:
whatever they tell it to! we won't do any validating on it, it'll be up to them to make sure they get it right

dbumblis-parabol (approved):

Issue #9489, opened 2024-03-01
chore: migrate ScheduledJob to pg (1/1)


jordanh:
## Issue - Enhancement

Let's get `ScheduledJob` migrated to pg. It's small, and it's in the way of migrating `Notifications`.

### Acceptance Criteria (optional)
   - Existing data migrated from rethink to pg
   - All future table operations interact with pg, not rethinkdb

**Estimated effort:** 8 points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

PR #9490, opened 2024-03-01
chore: migrate ScheduledJob from rethinkdb to pg

jordanh:
# Description

Resolves: #9489

I had a bit of extra time last night and decided to try and tackle this small migration. I wanted to see firsthand what kind of effort these migrations (even the simple ones) take.

## Testing scenarios

- [x] Set a meeting timer / let it elapse

- [ ] Check scheduled team limits (not checked, as this is not currently in use)

## Final checklist


packages/server/database/types/ScheduledJob.ts
```diff
@@ -1,11 +1,8 @@
-import generateUID from '../../generateUID'
-
 export type ScheduledJobType =
   | 'MEETING_STAGE_TIME_LIMIT_END'
   | 'LOCK_ORGANIZATION'
   | 'WARN_ORGANIZATION'
 
 export default abstract class ScheduledJob {
-  id = generateUID()
```
jordanh, line 9: Postgres now manages the id

packages/server/graphql/mutations/helpers/removeScheduledJobs.ts
```diff
@@ -1,8 +1,19 @@
...
-  return r.table('ScheduledJob').getAll(runAt, {index: 'runAt'}).filter(filter).delete().run()
+type FilterType = Omit<Updateable<DB['ScheduledJob']>, 'runAt'>
+
+const removeScheduledJobs = async (runAt: Date, filter?: FilterType) => {
+  const pg = getKysely()
+  let query = pg.deleteFrom('ScheduledJob').where('runAt', '=', runAt)
+  if (filter) {
+    Object.keys(filter).forEach((key) => {
+      const value = filter[key as keyof FilterType]
+      if (value) query = query.where(key as keyof FilterType, '=', value)
```
jordanh, line 13: This will "AND ${field} = ${value}" for anything specified in the filter object

packages/server/postgres/migrations/1709312768000_scheduledJobMove.ts
```diff
@@ -0,0 +1,72 @@
...
+    const {id, runAt, type, orgId, meetingId} = row
+    return {
+      runAt,
+      type,
+      orgId,
+      meetingId
+    }
+  }
+
+  const getNextData = async (leftBoundCursor: Date | undefined) => {
```
jordanh, line 29: Note: see https://github.com/ParabolInc/parabol/blob/ba7d7246c8db632f8016ec632c7841ae9a847dd2/packages/server/postgres/migrations/1677272969994_meetingTemplatesMove.ts#L37 for where I pilfered this logic

packages/server/billing/helpers/removeTeamsLimitObjects.ts
```diff
@@ -1,20 +1,22 @@
+import getKysely from '../../postgres/getKysely'
+import {DB} from '../../postgres/pg'
 import {r} from 'rethinkdb-ts'
 import {RValue} from '../../database/stricterR'
 import {DataLoaderWorker} from '../../graphql/graphql'
 import updateNotification from '../../graphql/public/mutations/helpers/updateNotification'
 
 const removeTeamsLimitObjects = async (orgId: string, dataLoader: DataLoaderWorker) => {
-  const removeJobTypes = ['LOCK_ORGANIZATION', 'WARN_ORGANIZATION']
+  const removeJobTypes = ['LOCK_ORGANIZATION', 'WARN_ORGANIZATION'] as DB['ScheduledJob']['type'][]
```
mattkrick, line 9: +1 `as const` is a little narrower here, same can be applied to the line below it or any literal tuple

mattkrick (approved):
code looks good! haven't run a test on it, but can't see any flaws in it from here

packages/server/billing/helpers/removeTeamsLimitObjects.ts
jordanh: I keep forgetting this!

jordanh:
@Dschoordsch, @mattkrick had a look – I yield to you if you want to test and merge

packages/server/postgres/migrations/1709312768000_moveScheduledJob.ts
```diff
@@ -0,0 +1,59 @@
...
+    if (nextBatch.length === 0) return null
+    if (nextBatch.length < batchSize) return nextBatch
+    const lastItem = nextBatch.pop()
+    const lastMatchingRunAt = nextBatch.findLastIndex((item) => item.runAt !== lastItem!.runAt)
+    if (lastMatchingRunAt === -1) {
+      throw new Error(
+        'batchSize is smaller than the number of items that share the same cursor. Increase batchSize'
+      )
+    }
+    return nextBatch.slice(0, lastMatchingRunAt)
```
Dschoordsch, line 35: -1 we should include `lastMatchingRunAt`
```
    return nextBatch.slice(0, lastMatchingRunAt + 1)
```
otherwise we might lose data. Assume a sequence [..., 4, 5, 5, 6], we would return [..., 4, 5] and start with the open (5, ...) next time, omitting this one element.

This sparks the idea that the migration should read the count first and verify it matches in the count in pg

packages/server/graphql/private/mutations/runScheduledJobs.ts
```diff
@@ -60,15 +63,17 @@ const runScheduledJobs: MutationResolvers['runScheduledJobs'] = async (
...
   // RESOLUTION
   const before = new Date(now.getTime() + seconds * 1000)
-  const upcomingJobs = (await r
-    .table('ScheduledJob')
-    .between(r.minval, before, {index: 'runAt'})
-    .run()) as ScheduledJobUnion[]
+  const upcomingJobs = await pg
+    .selectFrom('ScheduledJob')
+    .selectAll()
+    .where('runAt', '>=', new Date(0))
```
Dschoordsch, line 74: +1 I think this line could just be removed

packages/server/postgres/migrations/1709306178000_addScheduledJob.ts
```diff
@@ -0,0 +1,38 @@
...
+      "id" SERIAL PRIMARY KEY,
+      "runAt" TIMESTAMP WITH TIME ZONE DEFAULT NOW() NOT NULL,
+      "type" "ScheduledJobTypeEnum" NOT NULL,
+      "orgId" VARCHAR(100),
+      "meetingId" VARCHAR(100)
+    );
+
+    CREATE INDEX IF NOT EXISTS "idx_ScheduledJob_orgId" ON "ScheduledJob"("orgId");
+    CREATE INDEX IF NOT EXISTS "idx_ScheduledJob_runAt" ON "ScheduledJob"("runAt");
+    CREATE INDEX IF NOT EXISTS "idx_ScheduledJob_type" ON "ScheduledJob"("type");
```
Dschoordsch, line 25: +1 Not sure we need an index here, it's only used after selecting for orgId, so only a few handful  of rows

Dschoordsch (changes requested):
did not test yet

packages/server/postgres/migrations/1709312768000_moveScheduledJob.ts
mattkrick: in future migrations I think I'll probably create a compound index that includes the timestamp + ID so we're guaranteed a unique ID that is also serial so items can still get updated while the migration is running

jordanh: @Dschoordsch this is not correct, see: https://github.com/ParabolInc/parabol/pull/9502#discussion_r1510340193

packages/server/postgres/migrations/1709306178000_addScheduledJob.ts
jordanh: All of these indexes existed on the rethinkdb table...maybe we keep them for that reason?

packages/server/graphql/private/mutations/runScheduledJobs.ts
jordanh: Yeah, that's probably true. I was being a bit too direct in my translation ;)

jordanh:
@Dschoordsch branch has been updated to address one of your code review comments & rebased to master

#9500 should be merged first in order for the migration order to be correct

Dschoordsch (approved):

Issue #9472, opened 2024-02-28
Make all templates free


nickoferrall:
### AC

- [ ] Make all templates available to all users on the Starter plan

PR #9503, opened 2024-03-04
feat: make all templates free

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9472

I suggest that we don't remove all of the template limit logic as it's possible that we will want to implement template limits again in the near future. 

So, we should just set `isFree` to true for all meeting templates.

When I run `yarn pg:migrate up` I get the following error locally:

<details>

> - 1709551949071_makeAllTemplatesFree
> Rolling back attempted migration ...
RethinkDBError [ReqlOpFailedError]: Table `actionDevelopment.MeetingTemplate` does not exist in:
r.db("actionDevelopment").table("TemplateDimension").getAll(r.args(["estimatedEffortTemplate", "p5L84RD2og", "nLmvkxa43S", "wsjfTemplate"]), {
    index: "templateId"
}).merge((var_1) => {
    scale: r.db("actionDevelopment").table("TemplateScale").get(var_1("scaleId"))
}).orderBy("sortOrder").group("templateId").ungroup().merge((var_1) => {
    name: r.db("actionDevelopment").table("MeetingTemplate").get(var_1("group"))("name"),
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                             
    id: var_1("group"),
    dimensions: var_1("reduction")
})
    at Cursor.handleErrors (/Users/nickoferrall/parabol/node_modules/rethinkdb-ts/lib/response/cursor.js:274:23)
    at Cursor.resolve (/Users/nickoferrall/parabol/node_modules/rethinkdb-ts/lib/response/cursor.js:190:18)
    at processTicksAndRejections (node:internal/process/task_queues:95:5)
    at Function.run (/Users/nickoferrall/parabol/node_modules/rethinkdb-ts/lib/query-builder/query-runner.js:28:29) {
  msg: 'Table `actionDevelopment.MeetingTemplate` does not exist.',
  type: 17,
  cause: undefined,
  term: [ 35, [ [Array], [Array] ] ]
}

</details>

This migration is almost identical to [this one](https://github.com/ParabolInc/parabol/blob/7d7c39c10ae016794b2e4c1c8f302467409fc5b9/packages/server/postgres/migrations/1698831891828_moreFreeRetroTemplates.ts), so I assume it's a problem with my local set-up. It's referring to Rethinkdb when the table is in PG. Please let me know if you recognise the error. If not, I'll keep digging into it.

### To test

- [ ] Run the migration and see that all templates are free for all users

### DIFF
--- a/packages/server/postgres/migrations/1709551949071_makeAllTemplatesFree.ts
+++ b/packages/server/postgres/migrations/1709551949071_makeAllTemplatesFree.ts
@@ -15,31 +15,6 @@ export async function up() {
   await client.end()
 }
 
-const PREV_FREE_TEMPLATE_IDS = [
-  'herosJourneyTemplate',
-  'roseThornBudTemplate',
-  'winningStreakTemplate',
-  'starfishTemplate',
-  'hopesAndFearsTemplate',
-  'sixThinkingHatsTemplate',
-  'hotAirBalloonTemplate',
-  'heardSeenRespectedHSRTemplate',
-  'wRAPTemplate',
-  'speedCarTemplate',
-  'surprisedWorriedInspiredTemplate',
-  'marieKondoRetrospectiveTemplate',
-  'highlightsLowlightsTemplate',
-  'saMoLoTemplate',
-  'superheroRetrospectiveTemplate',
-  'questionsCommentsConcernsTemplate',
-  'dreamTeamRetrospectiveTemplate',
-  'handsOnDeckActivityTemplate',
-  'alwaysBeLearningRetrospectiveTemplate',
-  'scrumValuesRetrospectiveTemplate',
-  'estimatedEffortTemplate',
-  'wsjfTemplate'
-]
-
 export async function down() {
   // noop
 }
--- a/packages/server/postgres/migrations/1709551949071_makeAllTemplatesFree.ts
+++ b/packages/server/postgres/migrations/1709551949071_makeAllTemplatesFree.ts
@@ -41,22 +41,5 @@ const PREV_FREE_TEMPLATE_IDS = [
 ]
 
 export async function down() {
-  const client = new Client(getPgConfig())
-  await client.connect()
-  await client.query(
-    `
-    UPDATE "MeetingTemplate"
-      SET "isFree" = false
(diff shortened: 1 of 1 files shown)


Dschoordsch (changes requested):

packages/server/postgres/migrations/1709551949071_makeAllTemplatesFree.ts
```diff
@@ -0,0 +1,65 @@
...
+    `
+    UPDATE "MeetingTemplate"
+      SET "isFree" = true
+   `
+  )
+
+  await client.end()
+}
+
+const PREV_FREE_TEMPLATES = [
```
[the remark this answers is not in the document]
nickoferrall: I think that's because the first query in the down migration didn't check where orgId is `aGhostOrg`, so all of the custom templates were set to non-free. 

I've updated the down migration and checked staging for the list of free template ids.

packages/server/postgres/migrations/1709551949071_makeAllTemplatesFree.ts
```diff
@@ -0,0 +1,62 @@
...
+    UPDATE "MeetingTemplate"
+      SET "isFree" = true
+      WHERE "orgId" = 'aGhostOrg'
+   `
+  )
+
+  await client.end()
+}
+
+const PREV_FREE_TEMPLATE_IDS = [
```
Dschoordsch, line 18: -1 We're still missing more free templates: https://github.com/ParabolInc/parabol/blob/master/packages/server/postgres/migrations/1698831891828_moreFreeRetroTemplates.ts

There are probably others as well. I'd rather skip the down migration altogether than making more templates paid than necessary.

Dschoordsch (changes requested):

packages/server/postgres/migrations/1709551949071_makeAllTemplatesFree.ts
Dschoordsch: +1 This can now be removed

Dschoordsch (approved):

PR #9514, opened 2024-03-06
chore: migration reflection groups to pg

mattkrick:
# Description

migrates reflection groups to pg. tested locally everything migrates well except those titles with lower surrogates.


packages/server/postgres/migrations/1709749330256_retroReflectionGroups2.ts
```diff
@@ -0,0 +1,87 @@
...
+    'sortOrder',
+    'voterIds',
+    'smartTitle',
+    'title',
+    'summary'
+  ] as const
+  type RetroReflectionGroup = {
+    [K in (typeof PG_COLS)[number]]: any
+  }
+  const BATCH_SIZE = ~~(MAX_PG_PARAMS / PG_COLS.length)
```
Dschoordsch, line 40: +1 Had to look it up and the docs https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Bitwise_NO mention `Math.trunc` should be used instead.

Dschoordsch (changes requested):
I don't have enough data locally to test this. Let's make sure we check eq on staging.

PR #9517, opened 2024-03-08
feat: prepare embedder for Production

mattkrick:
# Description

Refactors the embedder to handle production workload.

Out of scope
- [ ] detect when there's not a enough data, or the data is not english
- [ ] consider adversarial models to grade a set of `teamIds`
- [ ] handle priority (e.g. process on feature flag or teamId, etc)

Demo
https://www.loom.com/share/e2d9d43d6d924676bb8a3cce84edb336?sid=82842bf2-68c5-42e5-b4f5-3c7f482b8e0b

How it works:
- 1 worker from the cluster becomes primary and builds the tables, triggers, and historical metadata
- when a meeting ends, the gql executor sends a message to the embedder via redis streams so only 1 worker gets the job to add the discussion to metadata table
- When a row is inserted into the metadata table, it also gets added to the job queue table. This happens via trigger. there is 1 trigger per embedding model 
- The embedder polls the PG job queue table to get a job. it creates `fullText` and embeds that. it then writes that embedding to the model's table.
- Can add items to the metadata by adhoc query passing in startAt and/or endAt
- every 5 mins the primary worker resets stalled embedding jobs


packages/server/postgres/migrations/1709860378296_embeddingsMetadataId.ts
```diff
@@ -0,0 +1,41 @@
+import {Client} from 'pg'
+import getPgConfig from '../getPgConfig'
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+    ALTER TABLE "EmbeddingsMetadata"
+      DROP COLUMN "models",
+      ALTER COLUMN "refId" SET NOT NULL;
```
mattkrick, line 10: @jordanh just to confirm-- was there any reason this was nullable?

packages/embedder/ai_models/AbstractModel.ts
```diff
@@ -42,7 +48,68 @@ export abstract class AbstractEmbeddingsModel extends AbstractModel {
...
+    const vectorDimensions = this.embeddingDimensions
+    console.log(`ModelManager: creating ${this.tableName} with ${vectorDimensions} dimensions`)
+    await sql`
+      DO $$
+        BEGIN
+        CREATE TABLE IF NOT EXISTS ${sql.id(this.tableName)} (
+          "id" INT GENERATED BY DEFAULT AS IDENTITY PRIMARY KEY,
+          "embedText" TEXT,
+          "embedding" vector(${sql.raw(vectorDimensions.toString())}),
+          "embeddingsMetadataId" INTEGER UNIQUE NOT NULL,
```
mattkrick, line 73: @jordanh safe to add the unique constraint here?

packages/server/postgres/migrations/1709860378296_embeddingsMetadataId.ts
jordanh: No, you're right to alter this to NOT NULL

packages/embedder/ai_models/AbstractModel.ts
jordanh: Yes, I think that is ok. It would imply that if we ever have multiple versions of the same embedding models, they'll need to live on their own table which I think is what we'd want

packages/embedder/importHistoricalRetrospectiveDiscussionTopic.ts
```diff
@@ -0,0 +1,36 @@
+import getKysely from 'parabol-server/postgres/getKysely'
+import {Logger} from 'parabol-server/utils/Logger'
+import RedisInstance from 'parabol-server/utils/RedisInstance'
+import {addEmbeddingsMetadataForRetrospectiveDiscussionTopic} from './addEmbeddingsMetadataForRet'
+
+export const importHistoricalRetrospectiveDiscussionTopic = async (redis: RedisInstance) => {
+  const pg = getKysely()
+  const isEarliestMetadataImported = await pg
```
mattkrick, line 9: @jordanh your strategy was best! i'm doing this here on startup. doing this in a migration would require that the embedder service be up at the time that the migration ran, which we can't guarantee

rafaelromcar-parabol:
This PR must upgrade the file [release-please-config.json](https://github.com/ParabolInc/parabol/blob/master/release-please-config.json) to add the Embedder to the list. If not, Release Please will not upgrade the package version with the rest of them.

.github/workflows/test.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       id-token: "write"
     services:
       postgres:
-        image: postgres:15.4
+        image: pgvector/pgvector:pg15
```
rafaelromcar-parabol, line 20: This fixes Postgres to version 15 but it does not fix the version on the pgvector extension. There are tags to fix both, the version of the extension + the version of the database. See [here](https://hub.docker.com/r/pgvector/pgvector/tags?page=1&name=pg15). I strongly recommend we use tags like [0.6.2-pg15](https://hub.docker.com/layers/pgvector/pgvector/0.6.2-pg15/images/sha256-0d3d57e65ca0d75f7d2c5f492f5c06353543495b2766d3998344708c2e8c8a04?context=explore).

CloudSQL uses, for now, the version 0.5.1 as you can see [here](https://cloud.google.com/sql/docs/postgres/release-notes), which isn't available as Docker tag. 

I propose we use 0.6.2-pg15 as I expect Google to upgrade to it sooner than later and it is a fixed version.

rafaelromcar-parabol (changes requested):

docker/images/parabol-ubi/environments/pipeline
```diff
@@ -54,3 +54,7 @@ STRIPE_PUBLISHABLE_KEY='pk_test_MNoKbCzQX0lhktuxxI7M14wd'
 STRIPE_SECRET_KEY=''
 STRIPE_WEBHOOK_SECRET=''
 HUBSPOT_API_KEY=''
+AI_EMBEDDING_MODELS='[{"model": "text-embeddings-inference:llmrails/ember-v1", "url": "http://localhost:3040/"}]'
+AI_GENERATION_MODELS='[{"model": "text-generation-inference:TheBloke/zephyr-7b-beta", "url": "http://localhost:3050/"}]'
```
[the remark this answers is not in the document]
mattkrick: ah yeah, we don't actually use the model in testing ye! We just extract the text `ember-v1` to build the table `Embeddings_ember_1` & then build typings from the PG schema.

.github/workflows/test.yml
mattkrick: nice! i'll fix this in build.yml too

packages/embedder/embedder.ts
```diff
@@ -46,206 +22,81 @@ tracer.init({
...
+  // subscribe to consumer group
+  try {
+    await redis.xgroup('CREATE', 'embedderStream', 'embedderConsumerGroup', '$', 'MKSTREAM')
+  } catch (e) {
+    // stream already exists
+  }
 
-  setTimeout(() => tick(modelManager), POLLING_PERIOD_SEC * 1000)
-}
+  const incomingStream = new RedisStream('embedderStream', 'embedderConsumerGroup', embedderChannel)
```
Dschoordsch, line 83: +1 It might be a bit easier to understand if the name would reflect that this only triggers generation of the metadata (which then triggers the embedding processing), so maybe we could name it 'embedMetadataStream'?

Dschoordsch (approved):
I have only nits. I think you both worked out the overall structure quite well. Happy to see this in production.
I did not get around to play with it, I'll try to do so tomorrow.

packages/embedder/ai_models/RecursiveCharacterTextSplitter.ts
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  private mergeSplits(splits: string[], separator: string) {
+    const docs: string[] = []
+    const currentDoc: string[] = []
+    let total = 0
+    for (const d of splits) {
+      const _len = d.length
+      if (total + _len + (currentDoc.length > 0 ? separator.length : 0) > this.chunkSize) {
+        if (total > this.chunkSize) {
```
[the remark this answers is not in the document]
mattkrick: you know, i'm not sure! i just blindly copied most of this from https://github.com/langchain-ai/langchainjs/blob/2c31e559558d9c2aacc4c1f01a0f7dd29e933ac0/langchain/src/text_splitter.ts#L191

which ported it from python
https://github.com/langchain-ai/langchain/blob/5d220975fc563a92f41aeb0907e8c3819da073f5/libs/text-splitters/langchain_text_splitters/base.py#L117-L118

since the JS matches the python logic i'm assuming it's probably good, but i didn't test it!

packages/embedder/addEmbeddingsMetadataForRet.ts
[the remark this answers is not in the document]
mattkrick: agreed! will fix

packages/embedder/addEmbeddingsMetadataForRet.ts
```diff
@@ -0,0 +1,107 @@
...
+      .where(lessThanTimeOrId)
+      .where('discussionTopicType', '=', 'reflectionGroup')
+      .orderBy('createdAt', 'desc')
+      .orderBy('id')
+      .limit(BATCH_SIZE)
+      .execute()
+    const earliestDiscussionInBatch = discussions.at(-1)
+    if (!earliestDiscussionInBatch) break
+    const {createdAtEpoch, id} = earliestDiscussionInBatch
+    curEndId = curEndAt === createdAtEpoch ? id : ''
```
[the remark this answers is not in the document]
mattkrick: let me benchmark! if i'm thinking about clearly, my solution can yield up to BATCH_SIZE - 1 duplicate results, but means I can run a more efficient query _most_ of the time since count(createdAt) is usually < BATCH_SIZE. will report back soon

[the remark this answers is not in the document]
mattkrick: Both yield the same answer of ~3MM inserts, but by ignoring the ID most of the time we processing time by 50% 🤯

makes me wonder if there's a more efficient way to write this query `eb('createdAt', '<', pgTime).or(eb('createdAt', '=', pgTime).and('id', '>', curEndId))` 
```
Total: 3009208

// Conditional curEndId
11:50:27 1|Embedder  | finished in 108591.671208 ms

// Constant curEndId
11:55:11 1|Embedder  | finished in 205402.91745799998 ms
```

packages/embedder/ai_models/RecursiveCharacterTextSplitter.ts
```diff
@@ -0,0 +1,117 @@
...
+          // - we have a larger chunk than in the chunk overlap
+          // - or if we still have any chunks and the length is long
+          while (total > this.chunkOverlap || (total + _len > this.chunkSize && total > 0)) {
+            total -= currentDoc[0]!.length
+            currentDoc.shift()
+          }
+        }
+      }
+      currentDoc.push(d)
+      total += _len
```
Dschoordsch, line 59: -1 I checked the python code you've linked and they add the separator length here as well which fixes the other places where I said there was a mismatch:

https://github.com/langchain-ai/langchain/blob/5d220975fc563a92f41aeb0907e8c3819da073f5/libs/text-splitters/langchain_text_splitters/base.py#L142

packages/embedder/addEmbeddingsMetadataForRet.ts
```diff
@@ -0,0 +1,107 @@
...
+}
+
+const validateDiscussions = async (discussions: (DiscussionMeta & {meetingId: string})[]) => {
+  const r = await getRethink()
+  if (discussions.length === 0) return discussions
+  // Exclude discussions that belong to an unfinished meeting
+  const meetingIds = discussions.map(({meetingId}) => meetingId)
+  const endedMeetingIds = await r
+    .table('NewMeeting')
+    .getAll(r.args(meetingIds), {index: 'id'})
```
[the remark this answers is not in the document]
mattkrick: not smart at all 😆 if you pass in 3 meetingIds it'll give back 3 docs of the same meeting. i got curious so I de-duped  the input as well as the output. only did 1 run, but it reduced time by 10% 🤯. Key takeaway-- rethinkdb is slooow. 

```
// before
12:12:40 1|Embedder  | done in  107929.862667 ms

// after
12:09:09 1|Embedder  | done in  88822.575084 ms
```

```ts
const meetingIds = [...new Set(discussions.map(({meetingId}) => meetingId))] // new
  const endedMeetingIds = await r
    .table('NewMeeting')
    .getAll(r.args(meetingIds), {index: 'id'})
    .filter((row: RDatum) => row('endedAt').default(null).ne(null))('id')
    .distinct() // new 
    .run()
  const endedMeetingIdsSet = new Set(endedMeetingIds)

```

packages/embedder/JobQueueStream.ts
```diff
@@ -0,0 +1,174 @@
...
+        .updateTable('EmbeddingsJobQueue')
+        .set({state: 'embedding'})
+        .where('id', '=', sql`ANY(SELECT id FROM ids)` as any)
+        .returningAll()
+        .executeTakeFirst()
+      if (!job) {
+        console.log('JobQueueStream: no jobs found')
+        // queue is empty, so sleep for a while
+        await sleep(ms('1m'))
+        return this.next()
```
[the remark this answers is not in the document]
mattkrick: I think about this ALL THE TIME! I wish I had an answer for you. We know it doesn't leak in prod because our gql executor uses this same pattern.
I've also tested this service for mem leaks & it works as expected. How??

If I had to take a guess, I'd say it's because `this.next()` has an arity of 0 so V8 knows there's no params it has to hold onto. We also hint to it because nothing comes after `this.next()` so my guess is when V8 figures out nothing will come after `this.next()` and it has no parameters to hold onto then it can safely GC. I've seen trampolines in the stacktrace when V8 runs out of memory, but i don't know if they use trampolines here.

It'd be really easy to test if we passed a parameter to this.next() or we put a console.log after a call to this.next()!

packages/embedder/JobQueueStream.ts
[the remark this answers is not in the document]
mattkrick: done! called it `EmbeddingsJobQueueStream` to match the table name

packages/embedder/embedder.ts
mattkrick: i like it!

packages/embedder/ai_models/RecursiveCharacterTextSplitter.ts
mattkrick: whoaaa! good find. I'm surprised they don't have a test suite to make sure the JS result is the same as the python result. Would you like to PR langchainjs or shall I?

.github/workflows/test.yml
```diff
@@ -100,6 +96,12 @@ jobs:
           wait-on: |
             http://localhost:3000/graphql
 
+      - name: Kysely Codegen
+        run: yarn pg:generate
+
+      - name: Check Code Quality
+        run: yarn codecheck
```
mattkrick, line 103: i added type safety on the embedding model tables so we can't code check until the embedding service starts up

docker/images/parabol-ubi/environments/pipeline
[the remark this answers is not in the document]
mattkrick: we need em! this env var tells the embedder service what to name the table & then we generate typings against that table name.

packages/embedder/ai_models/RecursiveCharacterTextSplitter.ts
Dschoordsch: https://github.com/langchain-ai/langchainjs/pull/4849

mattkrick:
Updating a readme whenever the version changes isn't great. While looking over that readme again, the build section is basically just `build.yml` in markdown, right? Can we remove that section & just say refer to the `build.yml` to see how to build it?

packages/embedder/mergeAsyncIterators.ts
```diff
@@ -0,0 +1,101 @@
...
+  promise: Promise<T>
+}
+
+// typescript intrinsics for iterables don't work like the do for promises yet, please ignore hackery
+type UnYield<T> = T extends IteratorYieldResult<infer U> ? U : never
+type UnIt<T extends AsyncIterator<any>> = UnYield<Awaited<ReturnType<T['next']>>>
+
+// Promise.race has a memory leak
+// To avoid: https://github.com/tc39/proposal-async-iterator-helpers/issues/15#issuecomment-1937011820
+export function mergeAsyncIterators<
```
jordanh, line 14: +1 

I encourage you to completely ignore this comment but file it away for some future date.


I think this pattern might be useful enough that it warrants becoming its own little library. Rather than work on just two iterables, I think you should extended its logic to operate on n-iterables.

The dance you do between awaiting on the promise (to allow the iterators to run) and adding to the queue is very clever.

The old man in me thinks of:

   * In UNIX, we had select in order to tell us when one or more non-blocking file or socket descriptors was ready for reading, writing (or had errored)
   * In Windows NT, we had `WaitForMultipleObjects` that would await for multiple kinds of handles had an event (such as file handles, event handles, mutexes, and semaphores)
   * Go has a `select { }` statement that does almost exactly what you've created here. Given multiple go channels, it awaits on each of them and runs a code block for the one that has data

packages/embedder/addEmbeddingsMetadataForRetrospectiveDiscussionTopic.ts
```diff
@@ -0,0 +1,108 @@
...
+  if (meetingId) {
+    const discussions = await pg
+      .selectFrom('Discussion')
+      .select(['id', 'teamId', 'createdAt'])
+      .where('meetingId', '=', meetingId)
+      .execute()
+    await insertDiscussionsIntoMetadata(discussions)
+    return
+  }
+  const PG_MAX_PARAMS = 65535
```
jordanh, line 64: +1 

Consider adding a comment on the why behind the BATCH_SIZE, I'd like to understand

packages/embedder/ai_models/TextEmbeddingsInference.ts
```diff
@@ -27,6 +31,28 @@ export class TextEmbeddingsInference extends AbstractEmbeddingsModel {
...
+      body: JSON.stringify({inputs: content}),
+      deadline: new Date(new Date().getTime() + MAX_REQUEST_TIME_S * 1000),
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json; charset=utf-8'
+      },
+      method: 'POST'
+    }
+
+    try {
```
jordanh, line 45: +2

This is very nice. It's so cool to see that we're riding the edge of the hugging face container development and using these endpoints as they ship them.

packages/embedder/indexing/createEmbeddingTextFrom.ts
```diff
@@ -1,15 +1,17 @@
...
-  switch (item.objectType) {
+  embeddingsMetadata: Selectable<DB['EmbeddingsMetadata']>,
+  dataLoader: RootDataLoader
+) => {
+  switch (embeddingsMetadata.objectType) {
     case 'retrospectiveDiscussionTopic':
-      return createTextFromRetrospectiveDiscussionTopic(item, dataLoader)
+      return createTextFromRetrospectiveDiscussionTopic(embeddingsMetadata.refId, dataLoader)
+    default:
+      throw new Error(`Unexcepted objectType: ${embeddingsMetadata.objectType}`)
```
jordanh, line 15: +2

Glad you added this throw

packages/embedder/indexing/countWords.ts
jordanh, line 1: +2 

Good riddance my friend

packages/embedder/indexing/retrospectiveDiscussionTopic.ts
```diff
@@ -251,7 +123,7 @@ export const createTextFromNewMeetingDiscussionStage = async (
   if (discussionSummary) {
     markdown += `Further discussion was made. ` + ` ${discussionSummary}`
   } else {
-    const comments = await dataLoader.get('commentsByDiscussionId').load(stageId)
+    const comments = await dataLoader.get('commentsByDiscussionId').load(discussionId)
```
jordanh, line 126: +2

Oh nice fix!

packages/embedder/resetStalledJobs.ts
```diff
@@ -0,0 +1,13 @@
+import ms from 'ms'
+import getKysely from 'parabol-server/postgres/getKysely'
+
+export const resetStalledJobs = () => {
+  setInterval(async () => {
+    const pg = getKysely()
+    await pg
```
jordanh, line 7: -1

Increment the retry counter here, also consider adding a state message

jordanh (changes requested):
Great work. Super enthused to see this become a thing with a solid foundation we can build on.

Only had one -1 to suggest changing – otherwise we'd been over the general working and names of things so many times there wasn't much to catch/suggest fixing (beyond @Dschoordsch's incredible catch on the text splitter)

rafaelromcar-parabol:
> Updating a readme whenever the version changes isn't great. While looking over that readme again, the build section is basically just `build.yml` in markdown, right? Can we remove that section & just say refer to the `build.yml` to see how to build it?
> 
> PR #9558

I answered the PR. As it is a different PR and we will be discussing it over there, I suggest you do the changes I proposed, just to be consistent with the current state of things, and we then see what to do with the future.

packages/embedder/mergeAsyncIterators.ts
mattkrick: renamed the functions to make reading it a little clearer.
supported n async iterables. that way we can run n job queue workers & still get good types.

packages/embedder/resetStalledJobs.ts
mattkrick: done!

Issue #9473, opened 2024-02-28
Limit Starter tier users to 2 custom templates


nickoferrall:
### AC

- [ ] Starter tier users can create no more than 2 custom templates per meeting type

PR #9518, opened 2024-03-08
feat: allow 2 custom templates for every user

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9473

Demo: https://www.loom.com/share/122c8e9e55db443cb52908a1a6957c07

### To test

- [ ] A user on the Starter tier can create two custom Retro templates and they'll see the upgrade CTA on the third attempt
- [ ] Same goes for Poker
- [ ] If they're on the Team or Enterprise tier, they can create as many custom templates as they like


packages/client/components/ActivityLibrary/CreateNewActivity/CreateNewActivity.tsx
```diff
@@ -95,9 +95,8 @@ const SUPPORTED_CUSTOM_ACTIVITIES: SupportedActivity[] = [
 const query = graphql`
   query CreateNewActivityQuery {
     viewer {
-      featureFlags {
-        noTemplateLimit
-      }
+      freeCustomRetroTemplatesRemaining
```
nickoferrall, line 98: I created two flags, `freeCustomRetroTemplatesRemaining` and `freeCustomPokerTemplatesRemaining`, thinking of our engineering principle that we should repeat the code twice and, on the third time, refactor it into something reusable.

packages/server/graphql/types/AddPokerTemplatePayload.ts
```diff
@@ -15,6 +17,14 @@ const AddPokerTemplatePayload = new GraphQLObjectType<any, GQLContext>({
         if (!templateId) return null
         return dataLoader.get('meetingTemplates').load(templateId)
       }
+    },
+    user: {
```
nickoferrall, line 21: I'll refactor this to the SDL pattern in a follow-up PR. This PR already touches a lot of files, so I didn't want to bloat it further

packages/client/modules/meeting/components/AddNewPokerTemplate.tsx
```diff
@@ -53,16 +53,17 @@ const AddNewPokerTemplate = (props: Props) => {
           id
           user {
             id
-            featureFlags {
-              noTemplateLimit
```
nickoferrall, line 57: Now that we've removed the template limit, we should remove this flag altogether. I've created an issue: https://github.com/ParabolInc/parabol/issues/9523

packages/server/graphql/mutations/addPokerTemplate.ts
```diff
@@ -51,9 +52,14 @@ const addPokerTemplate = {
...
       getFeatureTier(viewerTeam) === 'starter' &&
-      !viewer.featureFlags.includes('noTemplateLimit')
+      viewer.freeCustomPokerTemplatesRemaining === 0
     ) {
-      return standardError(new Error('Creating templates is a premium feature'), {userId: viewerId})
+      return standardError(new Error('You have reached the limit of free custom templates.'), {
+        userId: viewerId
+      })
+    } else {
+      decrementFreeTemplatesRemaining(viewerId, 'poker')
```
mattkrick, line 61: -1 we shouldn't decrement before we could possibly return an error

packages/server/graphql/public/types/User.ts
```diff
@@ -196,7 +196,11 @@ const User: UserResolvers = {
   tier: ({tier, trialStartDate}) => {
     return getFeatureTier({tier, trialStartDate})
   },
-  billingTier: ({tier}) => tier
+  billingTier: ({tier}) => tier,
+  freeCustomRetroTemplatesRemaining: ({freeCustomRetroTemplatesRemaining}) =>
```
mattkrick, line 200: +1 these aren't necessary if they exist on the source object, can remove (billingTier, too)

packages/server/postgres/migrations/1709733510511_addFreeCustomTemplatesRemaining.ts
```diff
@@ -0,0 +1,34 @@
...
+import getPgConfig from '../getPgConfig'
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+    DO $$
+    BEGIN
+      ALTER TABLE "User"
+        ADD COLUMN IF NOT EXISTS "freeCustomRetroTemplatesRemaining" INTEGER DEFAULT 2 NOT NULL;
```
mattkrick, line 11: -1 prefer `SMALLINT` to use 2 bytes instead of 4.

packages/server/postgres/queries/decrementFreeTemplatesRemaining.ts
```diff
@@ -0,0 +1,13 @@
+import getPg from '../getPg'
+import {decrementFreeRetroTemplatesRemainingQuery} from './generated/decrementFreeCustomRetroTemplatesRemainingQuery'
+import {decrementFreePokerTemplatesRemainingQuery} from './generated/decrementFreeCustomPokerTemplatesRemainingQuery'
+
+const decrementFreeTemplatesRemaining = async (userId: string, templateType: 'retro' | 'poker') => {
```
mattkrick, line 5: -1 pg-typed is deprecated. can we use kysely instead here?

mattkrick (changes requested):

nickoferrall:
> Please fix the PR title as it will appear in the changelog

@Dschoordsch it passes the PR title check. What would you like me to change it to?

Dschoordsch:
> > Please fix the PR title as it will appear in the changelog
> 
> @Dschoordsch it passes the PR title check. What would you like me to change it to?

Yes, it passes the check, but if you read
> ### Added
>
> * two custom template limit

would you know what it means?

How about "Allow 2 custom templates for every user"?

packages/server/postgres/queries/decrementFreeTemplatesRemaining.ts
mattkrick: @nickoferrall thoughts on changing this to kysely?

nickoferrall: Ah I've updated the migration to kysely, but not the query. I can do this too

nickoferrall: Updated!

packages/server/postgres/queries/decrementFreeTemplatesRemaining.ts
```diff
@@ -1,13 +1,27 @@
...
+  const pg = getKysely()
+  const customTemplateType =
     templateType === 'retro'
-      ? await decrementFreeRetroTemplatesRemainingQuery.run({userId}, getPg())
-      : await decrementFreePokerTemplatesRemainingQuery.run({userId}, getPg())
-  return res
+      ? 'freeCustomRetroTemplatesRemaining'
+      : 'freeCustomPokerTemplatesRemaining'
+
+  const userBeforeUpdate = await pg
```
mattkrick, line 10: -1 this is slightly worse than the SQL you had because if this function gets called twice then we risk the first query being a stale read. If there's ever a way to do something in a single SQL command, that should be the goal!

See [Playground Link](https://old.kyse.link/?p=b&i=haFkqHBvc3RncmVzoXamMC4yNy4yoXPaAaJpbXBvcnQgeyBHZW5lcmF0ZWQgfSBmcm9tICdreXNlbHknCgpkZWNsYXJlIGdsb2JhbCB7CiAgaW50ZXJmYWNlIERCIHsKICAgIHBlcnNvbjogUGVyc29uVGFibGUKICAgIHBldDogUGV0VGFibGUKICB9CgogIGludGVyZmFjZSBQZXJzb25UYWJsZSB7CiAgICBpZDogR2VuZXJhdGVkPHN0cmluZz4KICAgIGZpcnN0X25hbWU6IHN0cmluZwogICAgbGFzdF9uYW1lOiBzdHJpbmcgfCBudWxsCiAgICBjcmVhdGVkX2F0OiBHZW5lcmF0ZWQ8RGF0ZT4KICAgIGFnZTogbnVtYmVyCiAgfQoKICBpbnRlcmZhY2UgUGV0VGFibGUgewogICAgaWQ6IEdlbmVyYXRlZDxzdHJpbmc-CiAgICBuYW1lOiBzdHJpbmcKICAgIG93bmVyX2lkOiBzdHJpbmcKICAgIHNwZWNpZXM6ICdjYXQnIHwgJ2RvZycKICAgIGlzX2Zhdm9yaXRlOiBib29sZWFuCiAgfQp9oXHZfmNvbnN0IHBlcnNvbiA9IGF3YWl0IGRiCiAgLnNlbGVjdEZyb20oJ3BlcnNvbicpCiAgLnNlbGVjdChbJ2lkJywgJ2ZpcnN0X25hbWUnXSkKICAud2hlcmUoJ2lkJywgJz0nLCAnMScpCiAgLmV4ZWN1dGVUYWtlRmlyc3QoKaFjwg==)


```ts
const person = await db
  .updateTable('User')
  .set((eb) => ({[field]: eb(field, '-', '1')} ))
  .where('id', '=', userId)
  .where(field, '>', 0)
  .executeTakeFirst() 
```

nickoferrall: Makes sense, thanks! Updated

PR #9525, opened 2024-03-11
fix: Korean greeting corrected

jordanh:
feat: added additional greetings
feat: added additinal check-in questions


jordanh:
This is a trivial update, will merge directly without review once tests pass

PR #9530, opened 2024-03-12
chore(repo-structure): Docker images and stacks organized and clarified

rafaelromcar-parabol:
# Description

Deletes all deployment related files that are not used and simplifies that management of all the kept files. This makes it easier to understand how the application can be deployed in different environments and also simplifies how we maintain our repository.

Cleaner things are easier things to maintain.

# Important changes

- build.yml Github Action pointing to https://hub.docker.com/r/pgvector/pgvector/tags?page=1&name=pg15 instead of the unmaintained [ankane/pgvector:latest](https://hub.docker.com/r/ankane/pgvector/tags?page=1).
- Unused files are gone.
- Docker tags `latest` are replaced with specific versions of the applications.
- New folder run-it to host the how-to run the application in different use cases. docker-compose files under it are built with the same format, to ease comparing them and keeping them up to date.
  - Single-tenant: moved from docker/parabol-ubi/docker-host-st, but same content.
  - Development: dev.yml is placed in its own folder with (almost) everything it needs and names of the services are equal to the ones on the single-tenant deployment.

# Questions

- [X] Should we use pgvector:pg15 or use our own built Docker image for Postgres as the one in `docker/postrgres`? We agreed on keeping pgvector:pg15 for now.
- [x] What about the Postgres README? Adapt it so it only shows what relates to the app and nothing about running Postgres or PGAdmin
- [X] What about the Postgres file cacerts? the file is empty and it is used on the readCert.ts. Leave it like it is for now.
- [x] Why keeping two different pm2 files? What's the use case for yarn start right now? that's running it without any Docker. Testing after building? It is uses in test.yml. Do as DevOps like. pm2.config.js is only used in testing and for anyone running the app directly from the repository.
- [x] Postgres json-schema extension is installed on the Docker container built, but it isn't installed on any other environment. What about it? Still required? [Extensions supported by GCP CloudSQL Postgres](https://cloud.google.com/sql/docs/postgres/extensions#postgresql-extensions-supported-by-cloud-sql)

# Validations

- [ ] Github Action build working.
- [x] RethinkDB readme.
- [x] Postgres readme.
- [x] Application works using in development with `yarn && yarn db:start && yarn dev` with the default `.env.example`, following the instructions on the `README.md`.
- [x] Application works using `yarn && yarn build && yarn predeploy && yarn start` with the default `.env.example`.
- [x] Application runs using the single-tenant docker-compose deployment


rafaelromcar-parabol:
Do you think the files `pm2.dev.config.js` and `pm2.config.js` should be on the root folder or inside the folder `run-it/development`? @mattkrick @Dschoordsch

rafaelromcar-parabol:
`yarn start` is found [three times](https://github.com/search?q=repo%3AParabolInc%2Fparabol+%22yarn+start%22&type=code) in our repository:
- On the main README as instructions to deploy the application. Should we remove it from there?
- On the README of the alternative license for the US Department of Defense, also to indicate how to build for production and start the application.
- In our test.yml for Github. We can keep it here no problem.

Should we create a new `pm2.tests.config.js` and have that used for testing and keep `pm2.configs.js` with only stuff that are 100% production ready?

run-it/development/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+> ⚠️ **Parabol does not provide any support on this stack**. Use it under your own resposibility.
+
+## General notes
+
+- **This stack is not meant for production use.** It is our development stack and can change at any moment, have errors and incorporate and remove components we are testing without any notice.
+- A .env file localed in the root of this repository is used as [Docker env_file](https://docs.docker.com/compose/environment-variables/set-environment-variables/#use-the-env_file-attribute). It is the same file used by our command `yarn dev` to run the application in development mode.
+- This stack is designed to be managed using `yarn db:start` and `yarn db:stop` to start the databases and `yarn dev` or `yarn start` to start and manage the application, depending on the use case.
+
+## Components
```
mattkrick, line 11: IMO we should remove this because the docker-compose file provides all the same information. the best documentation is self-documenting code!

run-it/single-tenant-host/README.md
```diff
@@ -1,8 +1,10 @@
 # Docker Host Single Tenant (ST)
 
-To run Parabol in single tenant mode (e.g. simple docker-compose on a docker host).
+> ⚠️ **Parabol does not provide any support on this stack**. Use it under your own resposibility.
```
mattkrick, line 3: Our license already covers this, I'd opt for fewer words here & remove because no one likes reading disclaimers.

run-it/single-tenant-host/README.md
```diff
@@ -31,12 +33,12 @@ This will run `pre-deploy` and thus it will recreate the `web-server` and the `g
...
-- pgadmin
-- redis-commander
+- [pgadmin](https://www.pgadmin.org/)
+- [redis-commander](https://github.com/joeferner/redis-commander)
 
 To operate them use `docker compose up --profile databases --profile database-debug`.
 
 ## Running the whole stack
 
 - Start the whole stack: `docker compose --profile databases --profile parabol --profile database-debug --profile chronos up -d`.
```
mattkrick, line 43: no need to list all the profiles if you're running the whole stack! `docker compose up -d`

docker/parabol-ubi/security/xccdf_org.ssgproject.content_rule_account_disable_post_pw_expiration.sh
mattkrick, line 1: do we still need these now that we're on bookworm?

.env.example
```diff
@@ -74,9 +74,9 @@ POSTGRES_USE_PGVECTOR=true
 # POSTGRES_SSL_DIR='/var/lib/postgresql'
 REDIS_PASSWORD=''
 REDIS_URL='redis://localhost:6379'
-# REDIS_TLS_CERT_FILE=./certs/redis.crt
-# REDIS_TLS_KEY_FILE=./certs/redis.key
-# REDIS_TLS_CA_FILE=./certs/redisCA.crt
+# REDIS_TLS_CERT_FILE=./run-it/development/redis/certs/redis.crt
```
mattkrick, line 77: nit: prefer camelCase to kebab-case for naming convention. Instead of `run-it` perhaps another name like `run, deploy, infra, devOps`? same for `singleTenantHost`.


nit: why have both a `docker` and `run-it` directory? Both are concerned with how the thing is built. `./run-it/single-tenant-host` has a docker-copmose file init. Could we combine those 2 directories? Into a single e.g. `deploy`?

run-it/development/README.md
rafaelromcar-parabol: I agree that the first sentence could be removed. 

On the second sentence, the docker-compose file does not provide information about the `yarn` commands that can be used to manage the stack. I put that text to be explicit about how should it be used.

run-it/single-tenant-host/README.md
rafaelromcar-parabol: Ok!

run-it/single-tenant-host/README.md
rafaelromcar-parabol: Mmmm I'll check that but you might be right!

.env.example
rafaelromcar-parabol: Naming: I do not really care, we can just switch to run instead of run-it. We have more folders using - and I prefer that too, but I do not have a strong opinion on it 😸 we have parabol-ubi (which had inside docker-build and docker-host-st), we have packages/gql-executor and integration-tests

About the different folders:
- Docker is everything that is to build docker images. As you can see, not only Parabol.
- We could host other ways of running the application without Docker inside of it.
- We should not combine single-tenant-host and development. Those are completely different. 
- Single-tenant isn't a meant for development. Single-Tenant is what we use for PPMIs and what we would provide to Self-Hosted clients that would like to see how to make the app run in the simplest way. 
- I would like to keep our development folder so you guys can change it and use it. In the development docker-compose we do not run the application.

run-it/single-tenant-host/README.md
rafaelromcar-parabol: Security scripts are still in use on the `parabol.dockerfile`, which will be removed, along these scripts, as soon as we upgrade all our PPMIs. I hope we can do that sooner than later.

run-it/single-tenant-host/README.md
rafaelromcar-parabol: It doesn't work without adding services 
```
$ docker compose up -d
no service selected
```

That only works to run components without a service, and in this stack all components have services. Keeping those instructions then.

docker/parabol-ubi/security/xccdf_org.ssgproject.content_rule_account_disable_post_pw_expiration.sh
rafaelromcar-parabol: Security scripts are still in use on the parabol.dockerfile, which will be removed, along these scripts, as soon as we upgrade all our PPMIs. I hope we can do that sooner than later.

docker/postgres/Dockerfile
```diff
@@ -9,7 +9,7 @@ RUN apt-get update && apt-get install -y \
     git
 
 RUN cd /extensions/postgres-json-schema && make install && make installcheck
```
rafaelromcar-parabol, line 11: @mattkrick could we remove this, and all the files related to it, as we are not really using it anywhere?

.env.example
rafaelromcar-parabol: `run-it` moved to `run`. Let's keep the rest as it is for now and re-evaluate in the future if there is any tension

run-it/development/README.md
rafaelromcar-parabol: First sentence was removed. Second sentence is staying for now.

docker/postgres/Dockerfile
Dschoordsch: Just remove it and see what happens?

Dschoordsch (approved):
Looks good to me. No issues in dev while testing locally.

run/development/README.md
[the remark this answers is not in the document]
rafaelromcar-parabol: What about deploy?

run/development/README.md
```diff
@@ -0,0 +1,43 @@
+# Development stack
+
+> ⚠️ **Parabol does not provide any support on this stack**. Use it under your own resposibility.
+
+## General notes
+
+- **This stack is not meant for production use.** It is our development stack and can change at any moment, have errors and incorporate and remove components we are testing without any notice.
+- This stack is designed to be managed using `yarn db:start` and `yarn db:stop` to start the databases and `yarn dev` or `yarn start` to start and manage the application, depending on the use case.
```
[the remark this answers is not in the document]
rafaelromcar-parabol: Changed. Now it says that the stack can be used either with `yarn dev` or building the application (there are more than one way to build it) and then running `yarn start`. Actually, it could also be used without any of those yarn options (with the scripts in the dist folder).

docker/postgres/Dockerfile
rafaelromcar-parabol: It should work no trouble because I'm not adding that in any other environment, even when I'm building a Docker image for us. I'm just asking in case he wants to keep it for any reason that I cannot see. IMHO we should just remove it, and all files with it.

run/development/README.md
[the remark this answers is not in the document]
Dschoordsch: It's fine. Deploy sounds like setting up a production environment. I would opt for `start`, but that might also be misleading. I just wanted to rant a bit.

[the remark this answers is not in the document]
Dschoordsch: sounds good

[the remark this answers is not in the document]
rafaelromcar-parabol: Well, I could use `docker/build` as `build` folders are ignored. To avoid confusion and potential important errors, I switched to `docker/images` which is very descriptive.

docker/stacks/development/docker-compose.yml
```diff
@@ -1,10 +1,10 @@
-version: "3.7"
+version: "3.9"
```
mattkrick, line 1: +1 if we give this an explicit name like `name: parabol-dev` then the project name won't change when the folder changes.

mattkrick (approved):

docker/stacks/development/docker-compose.yml
rafaelromcar-parabol: I'll add `name: parabol-dev` tomorrow morning before merging! Good idea!

docker/postgres/Dockerfile
rafaelromcar-parabol: Removed!

PR #9548, opened 2024-03-19
fix: Configure trusted proxies

Dschoordsch:
Our proxy will add the client ip to the x-forwarded-for header correctly and also itself as it does NAT. Configure the number of trusted proxies to read the correct entry from the header to avoid manipulation by clients.

# Description

## Demo

## Testing scenarios

- [ ] Scenario A
  - Step 1
  - Step 2...

- [ ] Scenario B
  - Step 1
  - Step 2....

## Final checklist

### DIFF
--- a/packages/server/utils/uwsGetIP.ts
+++ b/packages/server/utils/uwsGetIP.ts
@@ -1,7 +1,11 @@
 import {HttpRequest, HttpResponse} from 'uWebSockets.js'
 
+const TRUSTED_PROXY_COUNT = Number(process.env.TRUSTED_PROXY_COUNT)
+// if TRUSTED_PROXY_COUNT is not configured correctly we fall back to reading the first IP to avoid rate limiting our proxy
+const CLIENT_IP_POS = isNaN(TRUSTED_PROXY_COUNT) ? 0 : -1 - TRUSTED_PROXY_COUNT
+
 const uwsGetIP = (res: HttpResponse, req: HttpRequest) => {
-  const clientIp = req.getHeader('x-forwarded-for')?.split(',')[0]
+  const clientIp = req.getHeader('x-forwarded-for')?.split(',').at(CLIENT_IP_POS)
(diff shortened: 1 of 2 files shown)


packages/server/utils/uwsGetIP.ts
```diff
@@ -1,7 +1,11 @@
 import {HttpRequest, HttpResponse} from 'uWebSockets.js'
 
+const TRUSTED_PROXY_COUNT = Number(process.env.TRUSTED_PROXY_COUNT)
+// if TRUSTED_PROXY_COUNT is not configured correctly we fall back to reading the first IP to avoid rate limiting our proxy
+const CLIENT_IP_POS = isNaN(TRUSTED_PROXY_COUNT) ? 0 : -1 - TRUSTED_PROXY_COUNT
```
rafaelromcar, line 5: Why setting to `-1 - TRUSTED_PROXY_COUNT` instead of just `TRUSTED_PROXY_COUNT`? I feel that more flexible. In our case we would use a negative value, in other cases it could be used in other way. Some clarification could be added to the `.env.example` saying that the number can be negative to pick stuff starting from the end.

Dschoordsch: When configured, you should always pick from the end. Everything at the beginning of the list is in client control and the proxies will add stuff to the end. I don't see a situation where one would want to read from the start.
The number to configure is really how many proxies are in your stack.

Previously we used the whole array as a key, but now we seem to check that it's a valid ip, so we need to pick a entry from the array. We cannot default to the last entry as this would rate limit the traffic coming from the proxy. Maybe we should add a warning that configuring this the wrong way would do that?

rafaelromcar: Yes, a warning or some clearer instructions would do. That said, LGTM.

rafaelromcar-parabol (approved):

PR #9557, opened 2024-03-21
feat: related discussions refactor

mattkrick:
# Description

Adds support for embedding workflows.
The first workflow we had was an `embed` workflow, but since that workflow was just 1 step long, I didn't really know how to build it.

The `RelatedDiscussions` workflow is a few steps long, which forced me to do a refactor.
The goal here is to be able to add & edit workflows quickly with as many reusable steps as possible.

## Demo

https://www.loom.com/share/d60e3cefb76948589a0ee2e03984e626?sid=bcf00f4f-645d-487e-9984-54b18a483a7f


packages/server/postgres/migrations/1712074131388_priority.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export async function up() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+    ALTER TABLE "EmbeddingsJobQueue"
+      ALTER COLUMN "priority" TYPE INTEGER;
+    ALTER TABLE "EmbeddingsJobQueue"
+      ADD COLUMN "model" VARCHAR(255),
+      ADD COLUMN "embeddingsMetadataId" INTEGER,
```
mattkrick, line 12: i added this & model back in because the foreign key is nice to have for debugging. writing SQL against JSON is just a drag

PR #9558, opened 2024-03-21
chore: remove build steps from readme and reference build.yml

mattkrick:
# Description

The best readme is self-documenting code!
Our build.yml script already outlines exactly how to build our app, so our readme should instruct folks to read the yml instead of creating a second source of truth.

I also renamed `.yaml` to `.yml` so all our yml files have the same extension.


mattkrick:
@rafaelromcar-parabol whatcha think? safe to reference the build yml instead of writing it out?

rafaelromcar-parabol:
Sorry man. I'm into other things and didn't have time to this. 

TBH, I do not like this change. I've used that readme many times to quickly copy and past the commands to build the application locally into a Docker container. I know @Dschoordsch has built it to, but I'm not sure if using this readme.

I agree that removing the readme makes it better in some way, but harder in other. And I like to be able to copy and past if I need to quickly build a Docker image because something is wrong. 

I'll like to get @Dschoordsch and @dbumblis-parabol take on this.

mattkrick:
Ah, I guess I don't know enough about how yall are using this!
@Dschoordsch What's the usecase for building locally? I didn't know devs would ever need to build the app. If we _do_ build, I assume we'd want to make sure we build in the same environment every time, so we'd just let CI do its thing!

Dschoordsch:
@mattkrick I only build the docker image locally if I want to test/debug something related to the docker image.
@rafaelromcar-parabol A link to the build file is enough for me, I can copy & paste from there.

rafaelromcar-parabol:
Just to be clear, all the commands to build the application are NOT just like that on the build CI. You have to be able to understand how to run the databases locally using Docker. Not that hard, but just to be clear about it.

mattkrick:
I think building locally isn't something we should support . We can't guarantee the same architecture, it's easy to have modified node_modules, different env vars, etc. If we want to build an ad-hoc image, we can do that using a GH runner using a workflow:  
Screenshot 2024-04-02 at 8 45 20 AM

right now we've got 2 sources of truth on how to build the app: 1 yml, 1 readme.
what i'd propose is we get rid of the readme & instead, if someone _really_ wants to build locally they can use [Act](https://github.com/nektos/act) to run our github action locally. that way they don't have to copy & paste a bunch of steps!

my goal here is to reduce maintenance overhead. maintaining a readme sucks, so readmes should only cover large fundamental ideas & leave the specifics for the code. whatcha think?

rafaelromcar-parabol:
It is true that now with the workflow we can build images on demand from branches and just pull them from our private dev repository. I didn't know about Act.

I still think this readme is useful but I understand and agree with what you say about only one source of truth. Let's get rid of it. If we feel the need in the future, we can always add something back.

rafaelromcar-parabol (approved):

PR #9586, opened 2024-04-02
fix: Fetch CORS resources from network

Dschoordsch:
If we fetch our own assets we want a CORS response to properly cache it in the service worker. If the response is served from the browser cache however, then there is a high chance that the required response headers are missing and the browser responds with a CORS error instead of refetching the resource. This happens because the Vary header does not include headers not present in the original request. To avoid this error, always fetch from the network for requests we intend to cache in the SW.

## Demo

https://www.loom.com/share/280ee174d1484814a06e4d2a1c9c2674?sid=374d929f-89ef-4ae5-ad9b-b108e5391737

## Testing scenarios

- setup ngrok to get a public URL with https (`ngrok http 3000`) and configure it in `.env` afterwards
- locally run in production mode (`yarn predeploy && yarn start`)
- open the site from the public URL
- clear service worker cache
- close tab and reopen the site
- on master you'll now see broken login images, on PR this is fixed

## Final checklist

### DIFF
--- a/packages/client/serviceWorker/sw.ts
+++ b/packages/client/serviceWorker/sw.ts
@@ -79,8 +79,10 @@ const onFetch = async (event: FetchEvent) => {
       // request.mode could be 'no-cors'
       // By fetching the URL without specifying the mode the response will not be opaque
       const isParabolHosted = url.startsWith(PUBLIC_PATH) || url.startsWith(self.origin)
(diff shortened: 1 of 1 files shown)


rafaelromcar-parabol:
LGTM!

rafaelromcar-parabol (approved):

nickoferrall (approved):

Issue #9589, opened 2024-04-02
Update pricing page to reflect new template limits


nickoferrall:
### AC

- [ ] Pricing page reflects that templates are now free and everyone can create two custom templates per meeting
- [ ] Add Org Admin Role to the Enterprise section

Mohd-Muneeb:
Can I be assigned to this task please? I'd like to know more about this if it's possible so I contribute to it

nickoferrall:
Sorry @Mohd-Muneeb, I've created a PR for this one as we wanted to get it updated quickly. Thank you for offering, though, and if there's another issue you're interested in, we always welcome contributions.

PR #9596, opened 2024-04-03
feat: update pricing page with template changes

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9589

### To test

- [ ] Pricing page includes updates to template limits and org admin role

### DIFF
--- a/packages/client/utils/constants.ts
+++ b/packages/client/utils/constants.ts
@@ -164,7 +164,7 @@ export const PARABOL_AI_USER_ID = 'parabolAIUser'
 export const StarterBenefits = [
   `${Threshold.MAX_STARTER_TIER_TEAMS} teams`,
   'Retrospectives, Sprint Poker, Standups, Check-Ins',
-  'Unlimited templates',
+  'Unlimited meeting templates',
   'Unlimited team members'
 ]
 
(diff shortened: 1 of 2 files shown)


packages/client/utils/constants.ts
```diff
@@ -160,10 +161,16 @@ export const SPOTLIGHT_TOP_SECTION_HEIGHT = 236
 
 export const PARABOL_AI_USER_ID = 'parabolAIUser'
 
+export const StarterBenefits = [
+  `${Threshold.MAX_STARTER_TIER_TEAMS} teams`,
+  'Retrospectives, Sprint Poker, Standups, Check-Ins',
+  'Unlimited templates',
```
Dschoordsch, line 167: +1 I'm not sure if this is self explanatory without knowing the history of this limit. Maybe something like "All predefined templates", "Parabol's premium templates" or just "Meeting templates"?

Dschoordsch (approved):

packages/client/utils/constants.ts
nickoferrall: I think it's similar to the "Unlimited team member" benefit that we also list in the Starter tier. I've updated it to "Unlimited meeting templates" to clarify what type of template it is

PR #9599, opened 2024-04-04
chore(segment): remove Segment

rafaelromcar-parabol:
# Description
Removes all references to Segment


rafaelromcar-parabol:
@tianrunhe you told me one month ago that we could remove segment from the .env.example and I've done that in this PR, but I would like you to validate that Segment isn't used anywhere. I also see some stuff related to Segment on the yarn.lock, but maybe I'm wrong.

Can you do the following?
- Check if there is anything else to remove and, if there is, remove it.
- Merge this PR once you have done that

Thanks

dbumblis-parabol:
LGTM, I will defer to @tianrunhe

tianrunhe (approved):

Issue #9555, opened 2024-03-21
Self-hosted instances have undefined user profile imageUrls


dbumblis-parabol:
### Pre-submission instructions (remove me before creating an issue)

Self hosted instances do not have the default images used on the SaaS version of the application and cannot reach our S3 bucket to retrieve them.

The image url is rendered as: url("https://undefined/static/avatars/da.png")

## Issue - Bug

- **How serious is the bug?** P2
- **OS:** All
- **Browser:** All
- **Node version:** 20.11.1
- **NPM version:** 
- **RethinkDB version:** 

### Acceptance Criteria (optional)
Users can:
   - Have default images for their profile or something other than a broken image that does not appear in interface when using a self-hosted instance

Triage is performed to find the root cause of the bug, timeboxed to ~1 hour. 

**Estimated effort:** 1 hour to triage. More if root cause is already identified.

PR #9603, opened 2024-04-05
fix: fallback for missing avatar images

mattkrick:
# Description

touches on #9555, but doesn't fix it properly.
this adds a fallback image in case their profile pic fails to load, but it only does this on the Avatar component (think user dash).  there may be other places that their avatar fails to load!

not eager to dump 1300 images into our github repo, but not sure what the better alternative is...

### DIFF
--- a/packages/client/components/Avatar/Avatar.tsx
+++ b/packages/client/components/Avatar/Avatar.tsx
@@ -1,5 +1,5 @@
 import styled from '@emotion/styled'
-import React, {forwardRef} from 'react'
+import React, {forwardRef, useState} from 'react'
 import defaultUserAvatar from '../../styles/theme/images/avatar-user.svg'
 import AvatarBadge from '../AvatarBadge/AvatarBadge'
 
@@ -7,7 +7,7 @@ type ImageBlockProps = Pick<Props, 'sansRadius' | 'sansShadow' | 'picture' | 'si
 
 const ImageBlock = styled('div')<ImageBlockProps>(
   ({sansRadius, sansShadow, picture, size, onClick}) => ({
-    backgroundImage: `url(${picture ?? defaultUserAvatar})`,
+    backgroundImage: `url(${picture})`,
     backgroundPosition: 'center center',
     backgroundRepeat: 'no-repeat',
     backgroundSize: 'cover',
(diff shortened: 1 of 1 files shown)


dbumblis-parabol:
@mattkrick What if we just used a single default image?

mattkrick:
that's an option! it just sucks when you're in a meeting & you see a sea of the same avatar so you don't know who is who.
i grabbed the avatars from S3 & the folder comes to 640KB, so i wouldn't feel horrible about including it in the bundle.

Mohd-Muneeb:
An approach would be to make images with random patterns like GitHub does on the fly, and use it with the user's session. It'll  eliminate the drawbacks of single default imaging while being much more user friendly. If you would like me to work on it I can help make an endpoint and raise a PR!

mattkrick:
@Mohd-Muneeb an identicon would be a great idea! Do you have a favorite package? I saw https://www.dicebear.com/styles/initials/ could replicate something very similar to what we have today & it looks like we could even pick a few colors from our [palette](https://github.com/ParabolInc/parabol/blob/3fd27a15aee921512f26360ba2196e8098bbd67c/packages/client/tailwindTheme.ts#L8) to set the background. If you'd like to open a PR I'd love to review it!

I imagine the flow would be something like:
- while bootstrapping a new user, grab the name or email address to use as a seed & generate a png
- upload the png to the CDN using the [FileStoreManager](https://github.com/ParabolInc/parabol/blob/9c330250d00ef30e6439ac089c634300058a46ba/packages/server/fileStorage/getFileStoreManager.ts)
- get the return value as a URL & pass that in for the `avatar` value when creating the User object

Mohd-Muneeb:
@mattkrick Yep, dicebear sounds like a really good idea.

Parabol team should once discuss with the design team to select a theme and we can continue with it. Also, maybe instead of storing them in a CDN, we can probably generate the avatar real time. What I mean to say is, since an email always gives the same image, we can show a identicon generated on client side. This will reduced the cost of a CDN and I don't think it will cause a huge performance reduction on the client side. I'm working on another PR I picked up from Parabol right now and I'm currently also busy with my internship.  I'd love to send a PR by Friday if that's okay! 

Please tag me once if the team decides with a theme. I would also suggest you to look at this [Repoistory](https://github.com/laurentpayot/minidenticons)

mattkrick:
If we only used it on the client that could work. Unfortunately we serve that avatar URL everywhere, including emails, which requires URLs that point to raster images or we risk emails getting truncated and avatars not showing up. A CDN only costs about $0.02/GB, so cost isn't a factor.

mattkrick:
awesome! looking forward to it

Issue #9541, opened 2024-03-15
AI Icebreakers feature completed and rolled out to all users as premium, AI feature


jordanh:
## Issue - Enhancement

Some time ago, Bartosz had created a hack week project that allowed the customization of Icebreaker prompts using AI. Let's make this work production ready and roll it out as a premium feature.

### Acceptance Criteria (optional)
Paying users can:
   - Adjust prompts by using quick action buttons (let's discuss final copy/prompts for these buttons)
      - Do our best in the prompt engineering to make sure generated content is safe for work

**Estimated effort:** X points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

PR #9660, opened 2024-04-19
feat: roll out ai icebreakers

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9541

I find the AI suggestions OK, but I'd probably prefer a feature where the AI creates a brand-new icebreaker question based on a prompt. It _could_ suggest something that's not safe for work, but the facilitator would still need to approve it, so I think that's a tolerable risk. Just an idea for a future enhancement, anyway.

Screenshot 2024-04-19 at 19 03 53

### To test

- [ ] All users that are the facilitator and don't have the `noAISummary` flag now see the AI Icebreaker in the Icebreaker phase
- [ ] There are quick action buttons that let the user change the icebreaker
- [ ] The suggested icebreaker questions are safe for work


Dschoordsch (approved):
Looks good, but I could not test it as we seem to have exceeded our current quota for open ai

PR #9689, opened 2024-04-30
fix: install plex on system

mattkrick:
# Description

to paint an svg to canvas we need a font on the server, so this should install plex.


mattkrick:
tests are breaking, i think they're flaky?

Dschoordsch:
> FAIL __tests__/processRecurrence.test.ts (12.818 s)
>  ● Should not end meetings that are not scheduled to end
>
>    thrown: "Exceeded timeout of 5000 ms for a test.
  
let's observe if we need to increase the timeout.

PR #9693, opened 2024-04-30
chore: Gracefully shutdown the embedder

Dschoordsch:
# Description

Fixes #9693
End all streams on receiving a kill signal to properly shutdown the embedder. Ending just the merged stream does not work in the current situation as it will wait indefinitely if all the workers are idle.

## Demo

Nothing to see here.

## Testing scenarios
 
- run the embedder outside of pm2
  ```
  sh -c 'export SERVER_ID=7 && exec node dev/embedder.js'
  ```
- send `SIGINT` via Ctrl+C
- see the process ending gracefully

## Final checklist

### DIFF
--- a/packages/embedder/logMemoryUse.ts
+++ b/packages/embedder/logMemoryUse.ts
@@ -6,5 +6,5 @@ export const logMemoryUse = () => {
     const {rss} = memoryUsage
     const usedMB = Math.floor(rss / MB)
     console.log('Memory use:', usedMB, 'MB')
-  }, 10000)
+  }, 10000).unref()
 }
--- a/packages/embedder/mergeAsyncIterators.ts
+++ b/packages/embedder/mergeAsyncIterators.ts
@@ -89,8 +89,7 @@ export function mergeAsyncIterators<T extends AsyncIterator<any>[] | []>(
       // Unwind remaining iterators on failure
       await Promise.allSettled(iterators.map((iterator) => iterator.return?.()))
       throw err
-    }
-    finally {
+    } finally {
       // Unwind remaining iterators on success
       await Promise.allSettled(iterators.map((iterator) => iterator.return?.()))
     }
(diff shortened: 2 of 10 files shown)


packages/embedder/mergeAsyncIterators.ts
```diff
@@ -15,82 +15,79 @@ type Result<T extends AsyncIterator<any>> = UnYield<Awaited<ReturnType<T['next']
...
       }
-    } catch (err) {
-      // Unwind remaining iterators on failure
-      try {
-        await Promise.all(iterators.map((iterator) => iterator.return?.()))
-      } catch {}
-      throw err
+    },
+    return: async () => {
+      await Promise.allSettled(iterators.map((iterator) => iterator.return?.()))
```
Dschoordsch, line 84: This refactoring was done so we can define this return function. This is necessary so all streams are notified of the return. The async generator function we had previously would hang in `await capability.promise` and handle the return only afterwards. If all embedders are idle, the promise would never resolve.

packages/embedder/logMemoryUse.ts
```diff
@@ -6,5 +6,5 @@ export const logMemoryUse = () => {
     const {rss} = memoryUsage
     const usedMB = Math.floor(rss / MB)
     console.log('Memory use:', usedMB, 'MB')
-  }, 10000)
+  }, 10000).unref()
```
Dschoordsch, line 9: This was an attempt to regularly end the event loop. It will continue if there are still timers running unless these are `unref`ed. This did not allow proper shutdown as there are many things in dependencies still running, but it's good practice nonetheless if we're not keeping a reference to the timer to allow cleanup.

packages/embedder/mergeAsyncIterators.ts
mattkrick: interesting! really good find having to define our own `return` here

Issue #8858, opened 2023-09-20
Remove AI Summary from discussion thread


nickoferrall:
Since upgrading to 3.5 turbo, the AI Summaries have become longer. Now that we have the prompt suggestions in the discussion thread (and they're good!), we should remove the AI Summary and just have the discussion prompt so that there's less text in the thread.

We will need to have released discussion prompts to all users before removing the AI Summary. 

### AC

- [ ] Remove AI Summary from discussion thread once discussion prompts have been rolled out

jordanh:
Just saw this. I'll tie this one to the AI Discussion thread issue, too

PR #9708, opened 2024-05-02
chore: remove ai summary from discussion thread

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/8858

This PR doesn't remove the discussion summary from the meeting summary. This will be handled in a separate issue: https://github.com/ParabolInc/parabol/issues/9710

### To test

- [ ] Create a retro and have a group with two or more reflections
- [ ] See the AI discussion suggestion in the discussion thread, but not the AI Summary


Dschoordsch (approved):
Now we only have the discussion prompt questions, also on the summary. I think we've discussed once that it might make sense to keep the questions in the meeting, but the summary for the groups in the meeting summary.

nickoferrall:
For sure! This PR (https://github.com/ParabolInc/parabol/pull/9711) removes the discussion prompts from the meeting summary

Issue #9710, opened 2024-05-03
Remove AI Discussion prompt from meeting summary


nickoferrall:
See Slack [message](https://parabol.slack.com/archives/C836NA350/p1714675308715089). We currently show the AI discussion summary and discussion question. Just show the AI Summary.

### To test

- [ ] Remove AI discussion question from the meeting summary

PR #9711, opened 2024-05-03
chore: remove discussion prompt from summary

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9710

Demo: https://www.loom.com/share/728f893c2e054fbbad82f849e89ffeaa

The AI Summary for the reflection group is no longer being used but I don't want to remove it now as it looks like we're migrating that table from RethinkDB to PG. I've created an issue: https://github.com/ParabolInc/parabol/issues/9712

### To test

- [ ] Add comments to a discussion thread 
- [ ] Go to the meeting summary and see that there's an AI Summary of the discussion thread and no discussion prompt

### DIFF
--- a/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/WholeMeetingSummary.tsx
+++ b/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/WholeMeetingSummary.tsx
@@ -26,16 +26,8 @@ const WholeMeetingSummary = (props: Props) => {
         }
         ... on RetrospectiveMeeting {
           reflectionGroups(sortBy: voteCount) {
-            summary
-          }
-          phases {
-            phaseType
-            ... on DiscussPhase {
-              stages {
-                discussion {
-                  summary
-                }
-              }
+            reflections {
+              id
             }
           }
         }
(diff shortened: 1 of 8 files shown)


packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/WholeMeetingSummary.tsx
```diff
@@ -49,14 +41,12 @@ const WholeMeetingSummary = (props: Props) => {
...
-    const hasTopicSummary = reflectionGroups!.some((group) => group.summary)
-    const hasDiscussionSummary = !!stages?.some((stage) => stage.discussion?.summary)
-    const hasOpenAISummary = hasTopicSummary || hasDiscussionSummary
-    if (!hasOpenAISummary) return null
-    if (hasOpenAISummary && !wholeMeetingSummary) return <WholeMeetingSummaryLoading />
+    const {summary: wholeMeetingSummary, reflectionGroups, organization} = meeting
+    const hasMoreThanOneReflection =
+      (reflectionGroups?.length && reflectionGroups.length > 1) ||
+      reflectionGroups?.some((group) => group.reflections.length > 1)
+    if (!hasMoreThanOneReflection || organization.featureFlags.noAISummary) return null
```
nickoferrall, line 48: If there's more than one reflection, we'll always create the meeting summary unless `noAISummary` is true, so we don't need to check if there's a discussion summary or reflection group summary (which is now removed anyway)

packages/client/mutations/EndRetrospectiveMutation.ts
```diff
@@ -122,12 +126,9 @@ export const endRetrospectiveTeamOnNext: OnNextHandler<
...
       popEndMeetingToast(atmosphere, meetingId)
     } else {
-      const discussPhase = phases.find((phase) => phase.phaseType === 'discuss')
-      const {stages} = discussPhase ?? {}
-      const hasTopicSummary = reflectionGroups.some((group) => group.summary)
       const reflections = reflectionGroups.flatMap((group) => group.reflections) // reflectionCount hasn't been calculated yet so check reflections length
-      const hasDiscussionSummary = !!stages?.some((stage) => stage.discussion?.summary)
-      const hasOpenAISummary = hasTopicSummary || hasDiscussionSummary
+      const hasMoreThanOneReflection = reflections.length > 1
+      const hasOpenAISummary = hasMoreThanOneReflection || !organization.featureFlags.noAISummary
```
Dschoordsch, line 131: -1 this should be `&&`, shouldn't it?

packages/server/graphql/mutations/helpers/generateDiscussionPrompt.ts
```diff
@@ -40,30 +40,21 @@ const generateGroupSummaries = async (
         ({reflectionGroupId}) => reflectionGroupId === group.id
       )
       if (reflectionsByGroupId.length <= 1) return
-      const reflectionTextByGroupId = reflectionsByGroupId.map(
-        ({plaintextContent}) => plaintextContent
-      )
-      const [fullSummary, fullQuestion] = await Promise.all([
-        manager.getSummary(reflectionTextByGroupId),
+      const [fullQuestion] = await Promise.all([
```
Dschoordsch, line 43: -1 No need to wrap a single call in `Promise.all([...])`

Dschoordsch (changes requested):
Did not test yet

Dschoordsch (approved):

Issue #9115, opened 2023-11-03
Activity Library v2: favourite activities logic


nickoferrall:
Predecessor: https://github.com/ParabolInc/parabol/issues/9114

### AC

- [ ] Clicking the heart button in an activity card adds the template to the favourites section
- [ ] Clicking the heart button in an activity card in the favourites section removes the card from the favourites section

PR #9713, opened 2024-05-03
feat: add logic that lets users favorite a template

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9115

Loom demo: https://www.loom.com/share/d59162967e244522aa40b73290a9c9a8

### AC

- [ ] Can favourite a template and see it show up in your favourites section
- [ ] Can un-favourite a template
- [ ] Can favourite and unfavourite a template from the template details view


packages/server/graphql/public/types/User.ts
```diff
@@ -85,6 +85,12 @@ const User: UserResolvers = {
     }
     return request
   },
+  favoriteTemplates: async ({favoriteTemplateIds}, _args, {dataLoader}) => {
+    const templates = await Promise.all(
+      favoriteTemplateIds.map((templateId) => dataLoader.get('meetingTemplates').load(templateId))
```
Dschoordsch, line 90: -1 Why not use `dataLoader.get('meetingTemplates').loadMany(favouriteTemplateIds).filter(isValid)`?

packages/server/dataloader/customLoaderMakers.ts
```diff
@@ -840,3 +840,24 @@ export const isCompanyDomain = (parent: RootDataLoader) => {
...
+      const users = await pg
+        .selectFrom('User')
+        .select(['id', 'favoriteTemplateIds'])
+        .where('id', 'in', userIds)
+        .execute()
+
+      const userIdToFavoriteTemplateIds = new Map(
+        users.map((user) => [user.id, user.favoriteTemplateIds])
+      )
+      return userIds.map((userId) => userIdToFavoriteTemplateIds.get(userId) || [])
```
Dschoordsch, line 857: +1 you could use `normalizeResults`:
```
return normalizeResults(userIds, users).map((user) => user?.favouriteTemplateIds || [])
```

packages/server/graphql/public/typeDefs/toggleFavoriteTemplate.graphql
```diff
@@ -0,0 +1,20 @@
...
+  ): ToggleFavoriteTemplatePayload!
+}
+
+union ToggleFavoriteTemplatePayload = ErrorPayload | ToggleFavoriteTemplateSuccess
+
+type ToggleFavoriteTemplateSuccess {
+  """
+  The user who's favorite templates were updated
+  """
+  user: User!
```
Dschoordsch, line 19: +1 Seems excessive to return the full user. I think it would be enough to return the updated template and favourite state.

Dschoordsch (changes requested):
The heart seems a bit misaligned for me (Safari, Mac), slightly too high on the card and more so on the category. It also conflicts with the "custom" badge on master, but we could remove that one again.
image

nickoferrall:
I've moved the heart in the card to be slightly lower and to the right. For the custom label, we have a separate [discussion](https://parabol.slack.com/archives/C836NA350/p1715104125865839?thread_ts=1715094125.441169&cid=C836NA350) in Slack around that.

In the category labels, I think it is centred:

Screenshot 2024-05-07 at 17 46 18

Screenshot 2024-05-07 at 17 44 49 (2)

Dschoordsch (changes requested):
I don't want to be picky, but the category heart is offset 2px to the top
image
image

We also need to do something about the custom label first before merging.

nickoferrall:
For the custom label, once you and Terry have reached a decision in the Slack [discussion](https://parabol.slack.com/archives/C836NA350/p1715176323724199?thread_ts=1715094125.441169&cid=C836NA350), I can update this PR or wait for the "Custom" label to be reverted/changed. 

> I don't want to be picky, but the category heart is offset 2px to the top

On Chrome, it was centered for me, but I can see the 2px offset on Safari. I've pushed an update and now it's centered on Safari for me, too:

Screenshot 2024-05-08 at 18 00 38

Screenshot 2024-05-08 at 18 00 25

Dschoordsch (approved):

Issue #9738, opened 2024-05-08
Team lead can't view teams page


nickoferrall:
Only the billing lead can view the Teams page in the org settings. This means that if someone is promoted to Team Lead who is not the billing lead, they can't view the Teams page.

<img width="1289" alt="Screenshot 2024-05-08 at 12 46 56" src="https://github.com/ParabolInc/parabol/assets/39854876/8efa66f8-d33d-4cca-b5e1-1e921f0bb4b4">

### AC

- [ ] Team lead can view Teams in org settings

PR #9739, opened 2024-05-08
fix: team lead can view teams in org settings

nickoferrall:
Fix: https://github.com/ParabolInc/parabol/issues/9738

image

### To test

- [ ] Team lead can view the teams page even if they're not the billing lead

### DIFF
--- a/packages/server/graphql/types/Organization.ts
+++ b/packages/server/graphql/types/Organization.ts
@@ -10,6 +10,7 @@ import {
 import {
   getUserId,
   isSuperUser,
+  isTeamLead,
   isUserBillingLeader,
   isUserOrgAdmin
 } from '../../utils/authorization'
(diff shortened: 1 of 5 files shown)


packages/server/graphql/types/Organization.ts
```diff
@@ -106,6 +107,14 @@ const Organization: GraphQLObjectType<any, GQLContext> = new GraphQLObjectType<a
           .sort((a, b) => a.name.localeCompare(b.name))
       }
     },
+    isTeamLead: {
+      type: new GraphQLNonNull(GraphQLBoolean),
+      description: 'true if the viewer is a team lead for a team in the organization',
+      resolve: async (_, _args: unknown, {authToken, dataLoader}) => {
+        const viewerId = getUserId(authToken)
+        return authToken.tms.some((teamId) => isTeamLead(viewerId, teamId, dataLoader))
```
Dschoordsch, line 115: -1 Did not test, but Array.some would call isTeamLead synchronous which would always be a truthy result (a Promise)

Dschoordsch (changes requested):

nickoferrall:
Updated! 

In production, only billing leads can view the Teams page. So a user that is not the billing lead but has been promoted to team lead, can't see the Teams page. 

This PR fixes that. But a user could just create a new team in the org, be team lead of that new team, and then have access to the Teams page. 

The Teams page contains email information, which is why we restricted it before:

Screenshot 2024-05-09 at 14 37 21


Some options:
1. Keep things as they are in prod. Only billing leads can see the Teams page.
2. Use this PR. Team leads can see the Teams page, too. We know that users could join an org, create a new team, and get access to the Teams page. 
3. Use this PR, but in the Teams page, only let team leads who are not the billing lead see the teams that they are the lead of. 
4. Let everyone in the org see the Teams page.

I lean towards options 1 or 3. What do you think?

Dschoordsch (approved):

Issue #9733, opened 2024-05-07
Org Admin is unable to change the Team Lead


charlesmathias:
Second item in this [Slack thread](https://parabol.slack.com/archives/C4JAUUZ9P/p1714403545560899) 🔒 

An org admin tries to update the team lead of a team and nothing changes on the team after confirming.

PR #9742, opened 2024-05-08
fix: org admin can change team lead

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9733

Currently throws an [error](https://parabol.sentry.io/issues/4780883754/?query=not+team+lead&referrer=issue-stream&statsPeriod=14d&stream_index=0) if the viewer isn't the team lead. This PR enables both team leads and org admins to change team lead, remove team members, and archive teams.

### To test

- [ ] Org admin can change the team lead


Dschoordsch (approved):
Looks good, did not test

PR #9744, opened 2024-05-08
fix: close websocket with reason on invalid token

### LINKED ISSUE #8370: RelayNetwork: No data returned for operation `TeamContainerQuery`, got error(s):
https://parabol.sentry.io/issues/4145374621/?referrer=github_plugin

```
RelayNetwork: No data returned for operation `TeamContainerQuery`, got error(s):
Server error

See the error `source` property for more information.
  at Array.forEach (<anonymous>)
  at Set.forEach (<anonymous>)
  at Set.forEach (<anonymous>)
  at errorCheckerSink.next (./packages/client/Atmosphere.ts:314:25)
...
(24 additional frame(s) were not displayed)
```

21 events 16 users

mattkrick:
# Description

fix #8370 

The server wasn't closing websockets with a `reason` when the session was invalidated.
This is important because if a `reason` is included, the client nukes the authToken.
Either I missed an API change in uws.js, or it hasn't been working for awhile.
This also rolls back to trebuchet-client 3.0.1 because once `canConnect` is true, it should not turn to false when it encounters an error. It should just try again.

The downside to this pattern is that the server has to upgrade the HTTP1.1 request to a websocket before it can close the websocket. This means the server tells the client that the websocket is open (fires `WebSocket.onopen`), the client sends pending messages (e.g. subscription requests) and then the server immediately closes it. It's extra work on the client & server, but sending a 401 during a websocket upgrade does not make it to the client. Per the spec: https://websockets.spec.whatwg.org/#eventdef-websocket-error 

```
User agents must not convey any failure information to scripts in a way that would allow a script to distinguish the following situations:

A server whose host name could not be resolved.

A server to which packets could not successfully be routed.

A server that refused the connection on the specified port.

A server that failed to correctly perform a TLS handshake (e.g., the server certificate can’t be verified).

A server that did not complete the opening handshake (e.g. because it was not a WebSocket server).

A WebSocket server that sent a correct opening handshake, but that specified options that caused the client to drop the connection (e.g. the server specified a subprotocol that the client did not offer).

A WebSocket server that abruptly closed the connection after successfully completing the opening handshake.

In all of these cases, [the WebSocket connection close code](https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.5) would be 1006, as required by WebSocket Protocol. [[WSP]](https://websockets.spec.whatwg.org/#biblio-wsp)

Allowing a script to distinguish these cases would allow a script to probe the user’s local network in preparation for an attack.
```

### TEST

- [ ] be logged in, change the SERVER_SECRET and restart the server. Upon reconnect, the client logs out automatically with a snack that says the session has expired.

### DIFF
--- a/packages/server/socketHandlers/handleUpgrade.ts
+++ b/packages/server/socketHandlers/handleUpgrade.ts
@@ -1,38 +1,14 @@
 import {WebSocketBehavior} from 'uWebSockets.js'
-import {TrebuchetCloseReason} from '../../client/types/constEnums'
-import safetyPatchRes from '../safetyPatchRes'
-import {isAuthenticated} from '../utils/authorization'
-import checkBlacklistJWT from '../utils/checkBlacklistJWT'
 import getQueryToken from '../utils/getQueryToken'
-import sendToSentry from '../utils/sendToSentry'
 import uwsGetIP from '../utils/uwsGetIP'
 
-const handleUpgrade: WebSocketBehavior<void>['upgrade'] = async (res, req, context) => {
-  safetyPatchRes(res)
-  const protocol = req.getHeader('sec-websocket-protocol')
-  if (protocol !== 'trebuchet-ws') {
-    sendToSentry(new Error(`WebSocket error: invalid protocol: ${protocol}`))
-    // WS Error 1002 is roughly HTTP 412 Precondition Failed because we can't support the req header
-    res.writeStatus('412').end()
-    return
-  }
-  const authToken = getQueryToken(req)
-  if (!isAuthenticated(authToken)) {
-    res.writeStatus('401').end()
-    return
-  }
-
+const handleUpgrade: WebSocketBehavior<void>['upgrade'] = (res, req, context) => {
   const key = req.getHeader('sec-websocket-key')
+  const protocol = req.getHeader('sec-websocket-protocol')
   const extensions = req.getHeader('sec-websocket-extensions')
   const ip = uwsGetIP(res, req)
-  const {sub: userId, iat} = authToken
-  // ALL async calls must come after the message listener, or we'll skip out on messages (e.g. resub after server restart)
-  const isBlacklistedJWT = await checkBlacklistJWT(userId, iat)
-  if (isBlacklistedJWT) {
-    res.writeStatus('401').end(TrebuchetCloseReason.EXPIRED_SESSION)
-    return
-  }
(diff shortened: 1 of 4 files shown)


jordanh (approved):
Looks good! No comments

packages/server/socketHandlers/handleUpgrade.ts
```diff
@@ -1,38 +1,14 @@
...
 
-const handleUpgrade: WebSocketBehavior<void>['upgrade'] = async (res, req, context) => {
-  safetyPatchRes(res)
-  const protocol = req.getHeader('sec-websocket-protocol')
-  if (protocol !== 'trebuchet-ws') {
-    sendToSentry(new Error(`WebSocket error: invalid protocol: ${protocol}`))
-    // WS Error 1002 is roughly HTTP 412 Precondition Failed because we can't support the req header
-    res.writeStatus('412').end()
-    return
-  }
```
Dschoordsch, line 18: -1 I think we need to distinguish 2 cases here: The client speaks our language but says the wrong thing (right protocol, wrong auth) and we don't understand it at all (wrong protocol). I think in the latter case, we should not move forward with the upgrade which should also fail the client permanently because it could never connect. Only when the protocol matches, we should upgrade and do the other checks. In that case the client can talk websocket with the server, so there is no need to try other connection methods.

packages/server/socketHandlers/handleUpgrade.ts
```diff
@@ -1,38 +1,14 @@
...
   const ip = uwsGetIP(res, req)
-  const {sub: userId, iat} = authToken
-  // ALL async calls must come after the message listener, or we'll skip out on messages (e.g. resub after server restart)
-  const isBlacklistedJWT = await checkBlacklistJWT(userId, iat)
-  if (isBlacklistedJWT) {
-    res.writeStatus('401').end(TrebuchetCloseReason.EXPIRED_SESSION)
-    return
-  }
-  res.upgrade({ip, authToken}, key, protocol, extensions, context)
+  const authToken = getQueryToken(req)
```
Dschoordsch, line 10: -2 we throw here now and the client tries reconnecting endlessly

mattkrick: whoa! could you provide some steps to reproduce? This function has a try/catch inside it so it shouldn't throw, it should just return a `{}`, which would then be used to send the `ws.close` signal, but I may have missed something!

PR #9750, opened 2024-05-10
feat: saml upload

### LINKED ISSUE #9254: SAML: Allow manually uploading the metadata as XML
Some IDPs (e.g. Google) don't provide a URL to the SAML metadata, but instead the user can only download it. We should allow to upload the XML file manually.

# Acceptance criteria
* allow uploading the metadata XML as alternative to the metadata URL
* upload and verify the file, validate it's not too big and a valid XML and store it on S3
* do the login verification check with a link to S3

mattkrick:
# Description

fix #9254

Supports uploading an XML file, which can then be used for the metadata URL

## Demo

https://www.loom.com/share/23e34f4943ca4650803c9faad05bc488?sid=9d52a92c-093b-4b88-8bc3-8f37c31db1ca

### DIFF
--- a/packages/server/graphql/public/permissions.ts
+++ b/packages/server/graphql/public/permissions.ts
@@ -4,10 +4,11 @@ import {Resolvers} from './resolverTypes'
 import getTeamIdFromArgTemplateId from './rules/getTeamIdFromArgTemplateId'
 import isAuthenticated from './rules/isAuthenticated'
 import isEnvVarTrue from './rules/isEnvVarTrue'
-import {isOrgTier, isOrgTierSource} from './rules/isOrgTier'
+import {isOrgTier} from './rules/isOrgTier'
 import isSuperUser from './rules/isSuperUser'
 import isUserViewer from './rules/isUserViewer'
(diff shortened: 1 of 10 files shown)


packages/server/graphql/public/permissions.ts
```diff
@@ -61,7 +63,7 @@ const permissionMap: PermissionMap<Resolvers> = {
     SAMLIdP: rateLimit({perMinute: 120, perHour: 3600})
   },
   Organization: {
-    saml: and(isViewerBillingLeaderSource, isOrgTierSource('enterprise'))
+    saml: and(isViewerBillingLeader('source.id'), isOrgTier('source.id', 'enterprise'))
```
mattkrick, line 66: curious on your thoughts about this new pattern. 
I want to reuse rules as much as I can. For example, this rule requires an orgId to determine if the viewer is on the org. That `orgId` may exist in an arg, or in the source, and it could have any name. I think a dot notation path is the cleanest way to do this, but wonder what you think?

nickoferrall: I like it! Not having type safety is my main concern, but that's still probably safer that accidentally mixing `isOrgTierSource` for `isOrgTier`

nickoferrall (approved):

PR #9776, opened 2024-05-23
feat: Create embeddings for meeting templates

### LINKED ISSUE #9772: AL template search: Create embeddings

Dschoordsch:
# Description

Fixes #9772 

## Demo

## Testing scenarios

- run the migration
- start the embedder
- wait for a bit
- see the embeddings being generated in PG

## Final checklist


Dschoordsch:
This one should wait until I verified or implemented the logic to update the `updatedAt` timestamp of meeting templates on prompt changes.

packages/server/postgres/migrations/1717685812677_addMeetingTemplateEmbedding.ts
```diff
@@ -0,0 +1,37 @@
...
+  await client.end()
+}
+
+export async function down() {
+  const client = new Client(getPgConfig())
+  await client.connect()
+  await client.query(`
+  DO $$
+  BEGIN
+    DELETE FROM "EmbeddingsMetadata" WHERE "objectType" = 'meetingTemplate';
```
nickoferrall, line 22: +1 would using kysely work here instead?

nickoferrall (approved):

packages/server/postgres/migrations/1717685812677_addMeetingTemplateEmbedding.ts
Dschoordsch: I think it would. I'm not sure what I would gain in that case. I guess typesafety after the up migration was run, but that seems a bit odd 🤔 
Any particular reason we should use kysely in migrations?

Dschoordsch: I don't see any examples of this elsewhere and I have some concerns regarding the typecheck. The typecheck would fail before it was run. In theory we could end up with migrations which never pass typecheck if the types differ between `up` and `down` too much.

PR #9779, opened 2024-05-23
feat: Use ChatGPT for reflection group title

tianrunhe:
# Description

In practice, I found `gpt-4o` produces better titles than `gpt-3.5-turbo` but given the price difference as much as 10x, I sticked with `gpt-3.5-turbo`. After all this is already much better than the salience extraction method.

## Testing scenarios

- [ ] AI generated group title
  - Drag some reflection cards into a group
  - See "🤖🤔 Thinking..." as a temporary title while we are waiting for OpenAI
  - See the AI generated title for the group

- [ ] Custom group title
  - Drag some reflection cards into a group
  - Manually edit the group title to something else
  - Drag more cards to the group
  - Verify the custom title does not change

## Final checklist


jordanh:
Any estimate on what the cost impact might be?

(Pardon me for not having reviewed the code yet): if OPENAI isn't available will this fall back to the previous behavior(s)?

packages/server/utils/OpenAIServerManager.ts
```diff
@@ -34,6 +34,36 @@ class OpenAIServerManager {
     })
   }
 
+  async getReflectionGroupTitle(reflections: Reflection[]) {
+    if (!this.openAIApi) return null
+    const prompt = `Below are lines of texts possibly talking about the same topic. Summarize all of the content in less than 3 simple words, with a leading emoji if possible. If you cannot generate a meaningful summary title, simply return "NO_RESPONSE" with nothing else instead. You will be penalized if your return exceeds 3 words of length, or the language of your output does not match the majority language of the input.
```
nickoferrall, line 39: +1 have you seen prompts written using "penalized" before? I'd say it'd be a clearer instruction to just say "do not exceed 3 words"

nickoferrall (changes requested):
Two small things in this Loom: https://www.loom.com/share/f5d8c072a0854b3797178b294a92021a

Do you have a rough idea of cost? If it's cheap enough, maybe Starter tier users could get access to it, too, as the title are way better

tianrunhe:
Quick response for the potential cost incurred after this change (with GPT-3.5-turbo): from $6.96 (April actual number) to ~$11 per month. I can show detailed numbers offline if you feel that feels fishy.
Given this magnitude, I agree with Nick: we can just enable it for all tiers.

packages/server/utils/OpenAIServerManager.ts
tianrunhe: This prompt has been "polished" by ChatGPT itself 😂

Dschoordsch: Isn't there an off-by-one error in the prompt?
> less than 3 simple words

vs.
> exceeds 3 words of length

tianrunhe:
My shower thought last night 😂: my previous calculation was wrong. What I did was: average number of retro meetings per month * average number of reflection groups per retro meeting * average number of chars per reflection group => tokens => cost. However it's actually an O(N^2) operation than O(N): let's say you have 3 reflections A, B, C. You first drag B into A, that's a call with token length = A + B. That forms a group & generates a title. Now you drag C into the group AB: here we call OpenAI with token length = A + B + C. So that's an O(N^2) operation.

Therefore I'm just gonna revert back to paid tier only for now. We run it for a while to monitor the actual costs and then make a decision if we want to release to normal users too

tianrunhe:
@nickoferrall, regarding two of your suggestions:
1. My ideal case would be: if the mutation is not calling OpenAI, do not show 🤖🤔 otherwise show them. However, I'm having difficulty distinguishing them from the client side so I just remove them altogether. There will be some cases when the latency is a bit higher so it looks like the app "hangs" after merging two cards. I'm not sure which way is better honestly.
2. I couldn't consistently reproduce this one. I think it only happens for shorter words when GPT-3.5 fails to generate a meaningful title so we fall back to the old title computing algorithm which I don't fully comprehend 😂. I have a stat though: the average number of characters per reflection group is ~90 so this should be pretty rare in real world.

nickoferrall (changes requested):
When I drag and drop reflections, it hangs for a split second for me: https://www.loom.com/share/56907fd5c1bc4b04b8f89f27f81a7c1d

tianrunhe:
@nickoferrall The issue you showed is actually also present in production today. It's just that the mutation mostly runs faster than this version (because we don't need to call OpenAI) so to human's eye, it's not quite noticeable. You can see that pretty clearly if you intentionally slow down the mutation by adding a sleep promise.
It took me sometime to fix that issue. TLDR: we need to set some additional values in the `optimisticUpdater` for the component to "mimic" the end state after we hear back from the mutation.
Now there's no delay or hang for the component, I think we need to bring back the 🤖🤔. Otherwise, the reflection group will show the old title first (because there's no way to compute the new title in `optimisticUpdater` from client side) and then switch back to the real title, which feels even more confusing I think.

nickoferrall (changes requested):
I see the "thinking..." loading feedback has been re-added. It feels like it's flashing to me because the loading state is so brief, so I think we should remove it. Would an empty string work? Maybe I'm being too picky. You could always ask Terry what he thinks and I'll defer to his UX taste. 

I also noticed that when you ungroup a reflection, it doesn't end up in exactly the right position atm. I'm not sure if that's caused by this PR, but it is working as expected for me on prod - see Loom: 
https://www.loom.com/share/a31bdf8c54b54b26a7f32beedb0bdb29

tianrunhe:
OK I spent some time this week looking into these issues, only to find myself falling into rabbit holes. So, I reverted most of my changes so that the current version becomes the minimum set of changes needed to introduce GPT for the reflection title. Theoretically the behavior should align with mainline.
Can we agree that the GPT generated group title brings much more value for our customers that they would overlook, if not completely ignore the nitty stuff like cards popping up for a few millisec? I'm trying to be less dogmatic here but citing the rubric from the T2 strategy:
> Shipping valuable units even over shipping low-risk units

I would argue that this is a definitely valuable unit that we can bear the relatively low risk. I can create issues for tracking for sure, if you agree with me on moving forward with this version. Thanks again for taking the time reviewing!

tianrunhe:
Oh and also, recall that [this issue](https://github.com/ParabolInc/parabol/issues/9833) is present today in production

tianrunhe:
@nickoferrall One of my previous versions (kind of) fixed the hanging but the side effect is the little "popping" thingy. So... which pill would you take? 😅

nickoferrall:
I don't think either are good options. If reflections hang or pop when ungrouping, it'll feel buggy. 

I think we should only use GPT for the group titles if grouping/ungrouping feels the same as production.

tianrunhe:
OK I'm gonna close this PR for now until I have spare time to revisit the performance issues. Apologize for wasting your time and thanks for testing this with me!

nickoferrall:
No time wasted at all - thanks for creating the PR. I think the GPT titles are way better, and if we can get the grouping/ungrouping working the same as prod, it'll be a great improvement.

PR #9780, opened 2024-05-23
feat: security banner concept

### LINKED ISSUE #9778: CUI banner for pubsec self-hosted deployments
## Issue - Enhancement

For government software it is a requirement to display a banner with the current CUI classification so users know which IL they're interacting on.

### **Implementation**
Ideally there is an easy way for the operator of the instance to configure which IL the deployment is in and the banner displayed would change based on that value

For ex: 
IL2 = Green Banner with specific CUI unclassified copy
IL4 = Purple banner with specific CUI classified information copy

### **Design**
Here's a reference doc:
https://www.astrouxds.com/components/classification-markings/#overall-marking

Also here is a current example for IL2 or unclassified:
Screenshot 2024-05-23 at 9 55 19 AM

The text displayed must match that exactly, with the correct color values for the banner

### Acceptance Criteria
Users can:
   - ability to configure which incident level the current application is running in
   - display a banner that conforms to DoD standards according to the configured IL

**Estimated effort:** 12 points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

ackernaut:
## Concept

Adds a banner with color and message options. Fixes #9778 

## Demo

[Loom 1 of 3 — Proof of Concept](https://www.loom.com/share/96f1e12836524843b296750fef2212b8)
[Loom 2 of 3 — Fixed Banner in All App Views](https://www.loom.com/share/c4382f5a23b74daba3b06fd5ea39282c)
[Loom 3 of 3 — Layout Accounts for Banner](https://www.loom.com/share/5c5ca6b556474d208a70e33b1e7006e1)

## Todo
- [x] Does the banner cause any overflow issues in the dashboard viewports when views are overflowing with content?
- [x] Add .env vars to provide message and style (change vars to generic banner naming convention)
- [x] Make sure banner shows in all views (e.g. move to Action.ts)
- [x] Tested: remote reflection drag and drop has accurate coordinates

## Test
- [x] The banner is fixed at the top of all app views — it doesn’t scroll away
- [x] The banner is set using `.env` variables including options for text, text color, and background colors
- [x] Remote reflection visual feedback (other users are dragging a reflection and you see it) has accurate coordinates (use a second browser and user to test this)
- [x] Check all views when banner is disabled for desktop
- [x] Check all views when banner is enabled for desktop
- [x] Check all views when banner is disabled for mobile
- [x] Check all views when banner is enabled for mobile
- Views to check:
  - auth page
  - `/invitation-link/12345678` view
  - 404
  - dash timeline, meetings, tasks
  - dash team, team tasks
  - dash team drawers (manage team, agenda)
  - dash team invite dialog
  - dash meeting summary
  - org admin views
  - org admin billing drawer
  - activity library
  - activity library details
  - meeting nav
  - meeting layout
  - poker discussion drawer
  - stand-up response thread drawer
  - stand-up work drawer
  - recurring retro choose template drawer
  - retro demo


packages/client/components/Dashboard.tsx
```diff
@@ -126,13 +127,26 @@ const Dashboard = (props: Props) => {
...
   const overLimitCopy = viewer?.overLimitCopy
   const meetingsDashRef = useRef<HTMLDivElement>(null)
+
+  // security banner concept WIP
+  const showSecBanner = true
+  const SECURITY_BANNER_TEXT = 'UNCLASSIFIED CUI (IL4)'
+  const SECURITY_BANNER_STYLE = {
+    backgroundColor: 'blue',
+    color: 'white'
+  }
```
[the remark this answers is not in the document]
ackernaut: I just stubbed these in for now. Agreed on changing the naming convention. I’m not sure how to access `.env` variables when using a client component, but I see the use of `process.env` in server files. What is the magic?

packages/client/components/Dashboard.tsx
```diff
@@ -126,13 +127,26 @@ const Dashboard = (props: Props) => {
...
+
   useSnackNag(overLimitCopy)
   useUsageSnackNag(insights)
   useSnacksForNewMeetings(activeMeetings)
   useNewFeatureSnackbar(viewer)
 
   return (
     <DashLayout>
+      {showSecBanner && (
+        <SecurityBanner text={SECURITY_BANNER_TEXT} style={SECURITY_BANNER_STYLE} />
```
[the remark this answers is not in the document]
ackernaut: Yep, on my to-do list. Thank you!

packages/client/components/Dashboard.tsx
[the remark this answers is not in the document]
Dschoordsch: Easiest is to follow an existing example like `AUTH_GOOGLE_ENABLED`: https://github.com/search?q=repo%3AParabolInc%2Fparabol%20AUTH_GOOGLE_ENABLED&type=code

ackernaut:
> For QA, we may have a spot or 2 in the app that calculates coordinates assuming a banner doesn't exist, e.g. remotely dragged reflections & oauth popup windows. Chances are any calculation that adds a hardcoded number to `innerHeight` from `window` could be vulnerable

Good to know @mattkrick — if I set a height to the banner, can we add or subtract that amount to the `innerHeight` (have to see how it works)?

packages/client/components/Dashboard.tsx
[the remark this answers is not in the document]
ackernaut: Thanks @Dschoordsch I’ll take a look

mattkrick:
I think it'll just be a case-by-case basis, if thing look good with the banner, we're good!

scripts/webpack/dev.client.config.js
```diff
@@ -131,6 +131,10 @@ module.exports = {
         AMPLITUDE_WRITE_KEY: process.env.AMPLITUDE_WRITE_KEY,
         microsoftTenantId: process.env.MICROSOFT_TENANT_ID,
         microsoft: process.env.MICROSOFT_CLIENT_ID,
+        GLOBAL_BANNER_ENABLED: process.env.GLOBAL_BANNER_ENABLED,
+        GLOBAL_BANNER_TEXT: process.env.GLOBAL_BANNER_TEXT,
+        GLOBAL_BANNER_BG_COLOR: process.env.GLOBAL_BANNER_BG_COLOR,
+        GLOBAL_BANNER_COLOR: process.env.GLOBAL_BANNER_COLOR
```
mattkrick, line 137: could you add these to .env.example with some sample text?

packages/client/components/ActivityLibrary/ActivityDetails/ActivityDetails.tsx
```diff
@@ -88,10 +88,12 @@ const ActivityDetails = (props: Props) => {
 
   const isOwner = viewerLowestScope === 'TEAM'
 
+  const isGlobalBannerEnabled = window.__ACTION__.GLOBAL_BANNER_ENABLED
+
   return (
     <div className='flex h-full w-full flex-col bg-white'>
       <div className='flex grow'>
-        <div className='mt-4 w-full grow'>
+        <div className={clsx('mt-4 w-full grow', isGlobalBannerEnabled && 'pt-6')}>
```
mattkrick, line 96: -1 as we build out our app, it'd be pretty easy to forget to include this logic! Instead of making sure this is included in every child component, could we just keep it in `<Action/>`? We can wrap the `<Switch/>` with <div className={isGlobalBannerEnabled ? 'pt-6' : ''}>

mattkrick (changes requested):
it looks good! the only downfall is that this adds quite a bit of complexity to our whole app. before merging, can we make sure that there's no simpler way to do this? I think we could probably keep the logic constrained to the parent `<Action/>` component

packages/client/components/ActivityLibrary/ActivityDetails/ActivityDetails.tsx
ackernaut: Maybe, I can try this and see if layout breaks.

ackernaut: The problem is we have some components that use `100vh` so we’d probably at least have to touch those. I can see about reducing the scope to `<Action />` and any components using `100vh`

devTemplate.html
```diff
@@ -1,4 +1,4 @@
-<!DOCTYPE html>
+<!doctype html>
```
ackernaut, line 1: This was due to auto-formatting

scripts/webpack/dev.client.config.js
ackernaut: Done! Added these to `.env.example`

ackernaut:
@mattkrick I added a wrapper around the `<Switch />` in `<Action />` and one around both the Banner and the inner wrapper to control sizing more consistently from a single place. I still had to apply some conditional styling to side sheets that use `100vh` — still haven’t found a good way to do that since the swipeable/transition parents are generated by the library.
```
<div>
  <Banner />
  <div>
    <Switch/>
  </div>
</div>
```

ackernaut (commented):
Ready for additional review and testing!

dbumblis-parabol:
LGTM

dbumblis-parabol:
Testing logic changes and smoke tested the views. LGTM! @mattkrick if things look ok to you let's ship it

packages/client/components/Action/Action.tsx
```diff
@@ -33,53 +35,75 @@ const Action = memo(() => {
...
-              component={DemoMeeting}
-            />
-            <Route path='/retrospective-demo-summary/:urlAction?' component={DemoSummary} />
-            {isInternalAuthEnabled && (
-              <Route
-                exact
-                path={`/forgot-password`}
-                render={(p) => <AuthenticationPage {...p} page={'forgot-password'} />}
-              />
+          <div className='flex h-full max-h-screen flex-col'>
```
mattkrick, line 49: in testing, i deleted this div & everything still worked OK, is there a use case where this is required?

packages/client/components/Action/Action.tsx
```diff
@@ -33,53 +35,75 @@ const Action = memo(() => {
...
-            />
-            <Route path='/reset-password/:token' component={SetNewPassword} />
-            <Route path='/team-invitation/:token' component={TeamInvitation} />
-            <Route path='/invitation-link/:token' component={InvitationLink} />
-            <Route component={PrivateRoutes} />
-          </Switch>
+            {/* Set up a container for the global banner and the main app area */}
+            <div
+              className={clsx(
+                'h-100 flex flex-1 flex-col overflow-hidden',
```
mattkrick, line 56: can we get rid of everything here except the `pt-6`? having overflow-hidden as a parent could make things difficult because that means no component will be able to use the viewport scroll

ackernaut: I think it was hard to control height of inner components on all the main views. In other words, I can do it on this one view and then tweak a few things here in there, or I have to go to most of the main views and adjust heights and what not. I’m not how much that would have to be conditional based on the `GLOBAL_BANNER_ENABLED` which I was trying to use less. So much of the app has to control inner scrolling. Maybe only mobile is relying on true viewport scrolling?

packages/client/components/Action/Action.tsx
ackernaut: This one allows for better control of the main height by wrapping both the banner and the main view. I think there may be one other way I could try. Let me test.

ackernaut:
Trying to remove those new outermost containers, I run into the same problems I fixed with them, go figure. I’m not sure how to get containers to behave. We have many, many containers where we are controlling the scrolling, and we have many views that layout to fit the viewport with inner scrolling. Introducing a global banner affects all views. Not sure if or how long it will take to try to find a way to achieve the tight views without odd scrolling etc due to the banner. We could ship this as it is timely, and then maybe clean up the container later. I’ll see if I can get somewhere with another approach that plays nice with all views, but until then I think it’s shippable. Good over perfect.

packages/client/components/Action/Action.tsx
ackernaut: I’ve got something I want to try but I think we should consider moving forward and making an additional set of improvements in a later PR.

packages/client/components/GlobalBanner.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+// https://www.astrouxds.com/components/classification-markings/#banner-examples
+
+// Note: banner height is 24px per Tailwind settings (Scale of 6: h-6, leading-6, p-6, etc.)
+// This height value is also in constEnums GlobalBanner
+
+const GlobalBanner = (props: Props) => {
+  const {bgColor, color, text} = props
+  return (
+    <div className='w-100 h-6 max-h-6 pt-6'>
```
ackernaut, line 18: @mattkrick I added a container here (reducing a container from Action.tsx). This pushes the container down that is in Action and use height instead of padding which makes the inner containers that are based on height behave nicely.

mattkrick (approved):

ackernaut:
@mattkrick thanks for reviewing, do I merge now or do you or @Dschoordsch have to merge?

PR #9798, opened 2024-05-29
feat: type safety for gql perms

mattkrick:
# Description

Adds type safety to the GQL permission rules.

Addresses comment here: https://github.com/ParabolInc/parabol/pull/9750#discussion_r1607070301

I'm not sure I love this solution. It fixes the problem, but it's a little verbose. Looking for thoughts!

## Demo

https://www.loom.com/share/6e708950dd4344e5beb7c6e6366cdd80?sid=36830171-f926-4c86-96b4-e70cce2f1c76


nickoferrall (approved):
I really like this! The type safety feels so much better, and I don't think it's too verbose

PR #9800, opened 2024-05-29
feat: nav style refinements

ackernaut:
Easier than noting small paddings and margins, colors, etc. (which I would think is annoying), I made a few refinements to get closer to the Figma. If you like them feel free to merge them!

[See the Loom](https://www.loom.com/share/b2bec95ef2af4407ae037f00f9033c62)

### DIFF
--- a/packages/client/components/Dashboard/DashSidebar.tsx
+++ b/packages/client/components/Dashboard/DashSidebar.tsx
@@ -45,7 +45,7 @@ const DashHR = styled('div')({
 
 const NavItem = styled(LeftDashNavItem)({
   borderRadius: 44,
-  paddingLeft: 16
+  paddingLeft: 15
 })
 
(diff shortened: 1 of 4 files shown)


nickoferrall (approved):
Love it! Thanks for making those changes. I like the subtle vertical margin so that the active & hovered nav items don't touch. 

The `isOpen` bit I commented out as I was going to update those styles in a later PR, but thanks for making those changes and including it now. 

I'll merge this and crack on with the next part.

ackernaut:
Thanks @nickoferrall ! I hope making a small PR doesn’t feel like stepping on toes but it’s hard to give a bunch of nits on pixel-level style details as feedback. This way it’s up to you if you want the changes. :)

nickoferrall:
Oh no - not at all, I think that's the right way to do it! Thanks for making the PR

PR #9801, opened 2024-05-29
chore: read ReflectionGroups from PG

mattkrick:
# Description

Removes all reads from RethinkDB#RetroReflectionGroup.

Note: this doesn't delete the table! We can't do that until after this has been deployed


Dschoordsch (approved):
Looks good to me.

For future reference: this change also includes stopping writes to rethink.

Issue #9791, opened 2024-05-29
Admin Nav: org nav section


nickoferrall:
From: https://github.com/ParabolInc/parabol/issues/9781

### Org Nav Sections
- [Note the header styles](https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=39-2965&t=XlnNDmMHL9H69kaK-11)
- The tier tag is spaced to the right and only wraps if the name is long. It seems most orgs have shorter names so this is an okay compromise.
- There are two nav items in the header
  - Settings & Members: opens a menu (see Nav Menu notes)
  - Add a Team: goes to New Team form (I'll add this in https://github.com/ParabolInc/parabol/issues/9792)
- There’s a divider
- There are teams I’m a member of

PR #9815, opened 2024-06-03
feat: update org nav dashboard styles

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9791

Loom demo: https://www.loom.com/share/cc181f0c48fc437880aa260bfce9ede1

### Questions for Terry (outlined in the Loom above)

1. Are you happy with the active nav item styles? There's the vertical space and rounded edges. I think the vertical space is fine, but the edges probably shouldn't be rounded. 
Screenshot 2024-06-03 at 16 58 57

2. Should the organisation settings page sidebar be updated to look like the dash sidebar? So this section: 
Screenshot 2024-06-03 at 17 00 39

I lean towards updating it, but only once the dashboard nav is complete. For now, we could leave it as it is in production. 

3. How do you feel about the "Team" tag? I didn't see this in designs. Happy to change the colour, of course. 
Screenshot 2024-06-03 at 16 02 56

4. Should we remove the links to org and settings here? I think we probably should as it's visible enough from the sidebar now
Screenshot 2024-06-03 at 17 03 04

The parent PR won't be merged until the mobile styles and finished.


ackernaut:
@nickoferrall here are numbered responses
1) Re: active team nav items, they should be rounded and the container with a border should have padding ([see Figma](https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=214-2347&t=UKiC0gYG7bVGXFDu-11)). Note: ‘Settings & Members’ would also have a similar hover state, rounded and inset
2) Re: org admin nav, we should update the org admin nav, too, since this is meant to be the same nav component. They should be rounded like the new main nav items (Meetings, History, etc.) and just need some padding. I’m happy to make a refinement PR if I need to.
3) Re: tier tag colors, it seems the Team tag color we use is gold. See the mobile nav for an idea of the tag colors used. The layout looks good. 
Screenshot 2024-06-04 at 13 23 04.
4) Re: team settings links, we can remove the org link. We need to keep the ‘Settings’ link as that’s the only way a Team Lead can delete their team. Once all functionality is moved to the Org Admin > Teams > Team view for a Team Lead we could consider removing both the Settings view and that link to it.
5) Nit: (don’t use blue for dash nav items) all the nav items including Add a Team and Settings & Members will use the same icon and text styles for both default and hover/active ([see Figma](https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=214-2347&t=UKiC0gYG7bVGXFDu-11))

ackernaut (changes requested):
See my other comment with numbered responses! Once it’s ready I can take a refinement pass if needed so that I’m not backseat driving on pixel values.

nickoferrall:
Thanks for the responses! That all makes sense to me.

For number 5, I had been working off the last design from [here](https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=39-2965&t=a5yvYAkO8MY5T9FV-0) with the blue links, but I'll work from [this design](https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=214-2347&t=4tCOf2zx4FTHRbjT-0) from now onwards

nickoferrall:
I've pushed an update with:

1. The active team nav bar
2. I've created an issue for this: https://github.com/ParabolInc/parabol/issues/9826
3. Tier tag updated 
4. I think we should probably wait until we can delete "Settings" too, otherwise it looks strange to me:
Screenshot 2024-06-05 at 12 57 11

5. Swapped blue for slate

Issue #9792, opened 2024-05-29
Admin Nav: nav menu & add team


nickoferrall:
From https://github.com/ParabolInc/parabol/issues/9781

### Nav Menu
- The nav menu mirrors the navigation of the org admin area
- Optional: for the Starter tier we can add a blue Upgrade text following a `•` character after Plans & Billing
- When tapping Settings & Members it uses the active/hover state like any other nav item.
### Add a Team
- Each org section header has a Add a Team link. The link routes the user to [the New Team form](https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=39-2965&t=XlnNDmMHL9H69kaK-11) with the correct Organization selected in the dropdown

Issue #9814, opened 2024-06-03
Admin Nav: update mobile styles


nickoferrall:
### AC

- [ ] Now that the admin styles have been updated on desktop, make sure they look good on mobile too

PR #9818, opened 2024-06-04
feat: promote Add Team in org nav dash

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9792

This PR promotes "Add a Team" to the org nav and removes it from the bottom. 

Screenshot 2024-06-04 at 16 02 01


ackernaut (changes requested):
I think we’ll like placing the Add a Team item as the last main nav item. We probably need to update how scrolling works. Here’s what I had in mind.
1) Mobile: the entire mobile menu scrolls, the footer is fixed ([see Figma](https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=39-2628&t=UKiC0gYG7bVGXFDu-11))
2) Desktop: the primary action and main nav items are static/fixed to the top by the logo depending on the view; now there is 1 scrollable section which contains all the org/team sections ([see Figma](https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=214-2347&t=UKiC0gYG7bVGXFDu-11))

Nit: the Add a Team is styled like all the other nav items including hover/active state. Not blue ([see Figma](https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=214-2347&t=UKiC0gYG7bVGXFDu-11))

nickoferrall:
Screenshot 2024-06-05 at 14 52 20
Screenshot 2024-06-05 at 14 52 26

1. I just wanted to confirm the differences in the mobile view (I'll tackle this in https://github.com/ParabolInc/parabol/issues/9814). In the new view, on the right, we'll remove the "Upgrade" bit, and add "My Settings", "Organizations", and "Sign Out" to the top section. "Add a Team" and "Sign Out" are then removed from the bottom. What page would "Organizations" take users to?

2. I think the scrolling behaviour is working as expected. See here: https://www.loom.com/share/63e1035b083f462bbcc106d0cc754879

3. I've pushed an update so that "Add a Team" is no longer blue and is the same as the other nav items.

ackernaut:
@nickoferrall this is the right direction.
- ‘Organizations’ would go to `/me/organizations`
- ‘Add a Team’ is now in the top section after ‘Tasks’, so yes it is removed from the bottom
- Also remove ‘Activity Library’ to save space. It’s the main activity button nowadays 
- Yes, the scrolling behavior at the laptop size is what I had in mind. Per the notes, on mobile the whole menu is scrollable with a pinned brand footer.
- Thanks for updating styles and I understand we can refine mobile styles as we wrap up. Happy to help with that

I’ll pull this down and take a closer look

ackernaut:
I ran this branch and #9827 PR’s branch locally and everything is headed the right direction.

Eventually we’ll want to refine dash nav, org admin nav, and mobile nav. Maybe that is a final PR, happy to help.

All items are essentially a nav item. Although it looks different, the View 12 Public Teams has the same footprint of a nav item.

Here I’ve shown how the nav items are all aligned and the same width regardless of if they are in a box with a border or without. The inner padding was meant to be 8px, but the inner borders ended up making it 7px. We can see which looks better before we ship. The Settings & Members has padding around it. The divider is full width. Here are some examples:

- Alignment and inner padding of **dashboard** [at desktop](https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=294-2899&t=eC0XfdDyJB2smK7U-11), and [at mobile](https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=39-2628&t=eC0XfdDyJB2smK7U-11)
- Alignment and inner padding of **org admin** [at desktop](https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=294-3059&t=eC0XfdDyJB2smK7U-11), and [at mobile](https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=294-3211&t=eC0XfdDyJB2smK7U-11)
Today the org nav menu takes over mobile, what do you think about something [like this](https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=294-3211&t=eC0XfdDyJB2smK7U-11), where the user hub nav items are persistent regardless of where you are (it’s either user hub + dash nav or user hub + org admin nav)?

ackernaut (approved):

nickoferrall:
> Eventually we’ll want to refine dash nav, org admin nav, and mobile nav. Maybe that is a final PR, happy to help.

Sounds good! Perhaps I could work on the issues remaining in the milestone and then before merging to production, you could create a PR refining the styles? 

> Today the org nav menu takes over mobile, what do you think about something [like this](https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=294-3211&t=eC0XfdDyJB2smK7U-11), where the user hub nav items are persistent regardless of where you are (it’s either user hub + dash nav or user hub + org admin nav)?

I like including "My Settings" and "Sign Out" in the nav, but we could remove "Organizations" as that's already featured below?

ackernaut:
Re: removing organizations, I like that idea. At some point if we have it so that Org Admins and maybe Billing Admins can be members without being on a team we’ll need to decide if they just get org nav sections with no teams in them, just an org name and settings link I suppose.

PR #9820, opened 2024-06-04
chore(rethinkdb): RetroReflection: Phase 1

mattkrick:
# Description

Phase 1 inserts & updates all rows to PG.

I also refactored some reads to using the dataloader so Phase 3 will go more smoothly.

To test, just run a retrospective meeting & make sure you can CRUD the reflections like normal


nickoferrall (approved):
Works! I had to wipe my dbs clean as I was getting some errors running migrations, but with a fresh db, it's all working swimmingly

packages/server/database/types/Reflection.ts
```diff
@@ -1,8 +1,14 @@
+import {sql} from 'kysely'
 import extractTextFromDraftString from 'parabol-client/utils/draftjs/extractTextFromDraftString'
 import generateUID from '../../generateUID'
 import GoogleAnalyzedEntity from './GoogleAnalyzedEntity'
 import Reactji from './Reactji'
 
+export const toGoogleAnalyzedEntityPG = (entities: GoogleAnalyzedEntity[]) =>
+  sql<
+    string[]
+  >`(select coalesce(array_agg((name, salience, lemma)::"GoogleAnalyzedEntity"), '{}') from json_populate_recordset(null::"GoogleAnalyzedEntity", ${JSON.stringify(entities)}))`
```
mattkrick, line 10: we have to be complex here to handle entities like `Google, Inc.` since that name is stored in a composite type & it has a comma in it.

Issue #9821, opened 2024-06-04
Remove AI Icebreaker UI for users that don't have access to OpenAI


nickoferrall:
### AC

- [ ] AI Icebreaker UI doesn't show up if you don't have access to OpenAI

PR #9824, opened 2024-06-04
fix: hide ai icebreaker ui for non ai users

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9821

The `noAISummary` feature flag is useful for disabling AI features for users who don't want them (it should probably be renamed to `noAI`). 

But we should also have an endpoint that lets the client know if the server doesn't have access to OpenAI.

### AC

- [ ] Remove access to OpenAI and see that the AI Icebreaker UI doesn't show up


packages/server/graphql/public/typeDefs/Organization.graphql
```diff
@@ -156,6 +156,11 @@ type Organization {
   """
   featureFlags: OrganizationFeatureFlags!
 
+  """
+  Has access to AI features
+  """
+  hasAI: Boolean!
```
mattkrick, line 162: curious, why put this on the organization when it is defined at the environment level?
could we instead send this down as property on `window.__ACTION__` like we do with other environment vars?

nickoferrall: Ahh yes, of course! I like that approach. A lot less code. Updated!

scripts/toolboxSrc/applyEnvVarsToClientAssets.ts
```diff
@@ -60,6 +60,7 @@ const rewriteIndexHTML = () => {
     stripe: process.env.STRIPE_PUBLISHABLE_KEY,
     publicPath: __webpack_public_path__,
     oauth2Redirect: process.env.OAUTH2_REDIRECT,
+    openAI: process.env.OPEN_AI_API_KEY,
```
mattkrick, line 63: -2 please correct me if i'm wrong, but isn't this a secret key?

nickoferrall: It is - we could make `openAI` a boolean and check if the key exists?

packages/client/types/modules.d.ts
```diff
@@ -40,6 +40,7 @@ interface Window {
     slack: string
     stripe: string
     oauth2Redirect: string
+    openAI: boolean
```
Dschoordsch, line 43: -1 hasOpenAI would fit the type better

Dschoordsch (approved):

Issue #9826, opened 2024-06-05
Admin Nav: org dash nav matches the dash nav


nickoferrall:
<img width="277" alt="Screenshot 2024-06-05 at 14 19 09" src="https://github.com/ParabolInc/parabol/assets/39854876/7cfbc7da-55a7-4579-a646-1379f09f33d0">

### AC

- [ ] Update org dash nav to reflect the new dashboard sidebar nav

PR #9827, opened 2024-06-05
feat: org dash matches dash nav

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9826

Screenshot 2024-06-05 at 15 23 46
Screenshot 2024-06-05 at 15 23 55

I don't think the org dash should mirror the dash nav exactly. Two things that I think should be different:

1. We don't need the "Settings & Members" item as it's redundant on the org page
2. I don't think we need the border. That works well on the main dashboard because it separates orgs from each other, but when we're viewing one org, I don't think we need it.

Curious to know what you think, @ackernaut! 🙏


ackernaut:
@nickoferrall we had the same thing in mind. I wanted the org admin nav items to look like the main nav items from the dashboard (Meetings, History, Tasks, Add a Team). No border, no new links, etc. so that is good! I will look it over!

ackernaut (approved):

PR #9834, opened 2024-06-06
chore(rethinkdb): RetroReflection: Phase 2

mattkrick:
# Description

Migrates RetroReflection to PG.
Note: this takes about 40 mins! Lets do this after hours if possible.


rafaelromcar-parabol:
> Note: this takes about 40 mins! Lets do this after hours if possible.

Let's then do it either
- Any week day between 11pm and 2am the next day UK time - 3pm to 6pm Pacific Time. The time were there are less writes to RethinkDB (and I use that as less traffic into our app) is around 1am UK / 5pm Pacific.
- Weekends after Fridays 11pm UK / 3pm Pacific and until Sunday are just perfect, almost no writes. For these big tables I'm not against doing the release on weekends. I think this one, TimeLineEvent, MeetingMember and maybe some more would benefit from doing this.

Up to you. I can make myself available just in case

Issue #9830, opened 2024-06-06
Admin Nav: add org menu 


nickoferrall:
### See https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=39-2965&t=ql2fzl5PH064XymP-0

![Screenshot 2024-06-06 at 15 00 40](https://github.com/ParabolInc/parabol/assets/39854876/02e693f1-891d-475c-aad6-f0c7b677264a)

### AC

- [ ] When hovering over nav items, the menu shows up looking like the designs

nickoferrall:
Hey @ackernaut, I just want to confirm that the menu should open onHover and only when hovering over Settings & Members? It doesn't open when hovering over the team names as it's org specific?

ackernaut:
@nickoferrall I think we should make it open on click. I want to say all of our menus open on click in the app. The Settings & Members item has the same footprint and hover state of the other nav items, but instead of going somewhere it opens a menu. How does that sound?

nickoferrall:
Keeping it consistent makes sense to me! I'll update this issue so it opens on click instead of on hover 👍

PR #9844, opened 2024-06-12
feat: add menu to dashboard nav

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9830

Loom demo: https://www.loom.com/share/ba5521f4b6464fd8bb5f6331ec5eab2e

Screenshot 2024-06-12 at 19 01 51

### To test

- [x] Click on "Settings & Members" and see the menu pops open, looking like the designs
- [x] On mobile, the menu doesn't open. It just takes you to the organization settings page

### DIFF
--- a/packages/client/components/DashNavList/DashNavMenu.tsx
+++ b/packages/client/components/DashNavList/DashNavMenu.tsx
@@ -1,6 +1,9 @@
 import styled from '@emotion/styled'
+import graphql from 'babel-plugin-relay/macro'
 import React from 'react'
+import {useFragment} from 'react-relay'
 import {useHistory} from 'react-router'
+import {DashNavMenu_organization$key} from '../../__generated__/DashNavMenu_organization.graphql'
 import {PALETTE} from '../../styles/paletteV3'
 import {Menu} from '../../ui/Menu/Menu'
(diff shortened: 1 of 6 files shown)


packages/client/components/DashNavList/DashNavMenu.tsx
```diff
@@ -0,0 +1,80 @@
...
+}
+
+const DashNavMenu = (props: Props) => {
+  const {orgId} = props
+  const history = useHistory()
+  const menuItems = [
+    {
+      label: (
+        <>
+          Plans & Billing •&nbsp;<span className='text-sky-500'>Upgrade</span>
```
ackernaut, line 29: I think the Upgrade label should be conditional. I’d either remove it altogether or make the `• Upgrade` bit conditional. It shows if the org is on the Starter tier.

ackernaut (changes requested):
This looks good. I like the onClick menu behavior for desktop, and the direct routing on mobile! Just one detail to look at for me. Note I didn’t look super closely at all code patterns but did go over once, seems straightforward.

packages/client/components/DashNavList/DashNavMenu.tsx
nickoferrall: Thanks, updated!

Issue #9858, opened 2024-06-18
Clicking Jira link affects the checkbox


nickoferrall:
@drewhousman1010 reported [here](https://app.hubspot.com/live-messages/3888472/inbox/7379923271#email):

> In the Scope screen of the Sprint Poker meeting type, clicking on a Jira link for an item affects the checkbox selection to add/remove it to the meeting scope
Message: Hi, just wanted to mention this as it has caused me to inadvertently drop an item from the meeting scope a few times. Any time I click on the Jira hyperlink for an item in the meeting scope to go view it in Jira, it changes the checkbox selection for that item.

### AC

- [ ] Clicking on the jira link doesn't affect the checkbox selection

PR #9859, opened 2024-06-18
fix: clicking scope link does not affect checkbox

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9858

Loom demo: https://www.loom.com/share/d47bb7b4bdce4ad49cc43cbbd7e38714

### To test

- [ ] Click a link in the scope phase of Poker and see that it does not check/uncheck the checkbox

### DIFF
--- a/packages/client/components/ScopingSearchResultItem.tsx
+++ b/packages/client/components/ScopingSearchResultItem.tsx
@@ -66,7 +66,11 @@ const ScopingSearchResultItem = (props: Props) => {
   const {onCompleted, onError, submitMutation} = useMutationProps()
   const disabled = !isSelected && usedServiceTaskIds.size >= Threshold.MAX_POKER_STORIES
   const isTemp = isTempId(serviceTaskId)
-  const onClick = () => {
+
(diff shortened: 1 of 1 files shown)


packages/client/components/ScopingSearchResultItem.tsx
```diff
@@ -66,7 +66,10 @@ const ScopingSearchResultItem = (props: Props) => {
   const {onCompleted, onError, submitMutation} = useMutationProps()
   const disabled = !isSelected && usedServiceTaskIds.size >= Threshold.MAX_POKER_STORIES
   const isTemp = isTempId(serviceTaskId)
-  const onClick = () => {
+
+  const handleClick = (e: React.MouseEvent) => {
+    e.stopPropagation()
```
Dschoordsch, line 71: -1 We have quite a few places where we stop event propagation in the app and to me it's never clear why it was put there, so it's impossible to remove or retest. Please add a comment.

Dschoordsch (changes requested):

PR #9865, opened 2024-06-25
fix: ensure dashboard sidebar animates smoothly

nickoferrall:
Fix issues @mattkrick shared [here](https://github.com/ParabolInc/parabol/pull/9863#issuecomment-2187305336)

Loom demo: https://www.loom.com/share/7ec1e767949b4c7fbee1fff3afe2ed09

This PR ensures that:

- Opening and closing the sidebar animates smoothly
- If an org has no teams, there's no divider
- The settings menu doesn't conflict with the text on the task cards

Things to note:
- I reduced the width of the Add Meeting button slightly. It's still wider than prod, but narrower than it was. 
- The three-line long team name is an edge case. A team name with the max characters allowed is usually 2 lines, but if the characters are particularly wide, e.g. `3` instead of `i`, it can take up three lines. I'm hesitant about changing the padding without a review from @ackernaut, so I'd prefer to merge this.
- I also like your point about removing the menu. I think Terry could see how it feels in production and decide whether to keep/remove it.

### DIFF
--- a/packages/client/components/DashNavList/DashNavList.tsx
+++ b/packages/client/components/DashNavList/DashNavList.tsx
@@ -67,7 +67,11 @@ const DashNavList = (props: Props) => {
     <div className='w-full p-2 lg:pt-4'>
       {organizations?.map((org) => (
         <div key={org.id} className='mb-3 w-full rounded-lg border border-solid border-slate-400'>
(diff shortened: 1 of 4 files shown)


mattkrick:
awesome! yeah, team name length is an edge case for sure, i'll get it on staging & give it a go!

PR #9867, opened 2024-06-25
chore(rethinkdb): RetroReflection: Phase 3

mattkrick:
# Description

Stop reads/writes to RethinkDB


packages/server/dataloader/primaryKeyLoaderMakers.ts
```diff
@@ -27,3 +27,29 @@ export const embeddingsMetadata = primaryKeyLoaderMaker((ids: readonly number[])
 export const retroReflectionGroups = primaryKeyLoaderMaker((ids: readonly string[]) => {
   return getKysely().selectFrom('RetroReflectionGroup').selectAll().where('id', 'in', ids).execute()
 })
+
+export const selectRetroReflections = () =>
```
mattkrick, line 31: I broke this out because the foreign key loaders also need this & wanted to keep it DRY.
Can't use `.selectAll()` here because `reactjis` is a `string[]` and we're exporting it as JSON, so there's a type conflict.

PR #9876, opened 2024-06-25
chore(rethinkdb): TimelineEvent: Phase 3

mattkrick:
# Description

Moves all reads to PG


packages/server/postgres/migrations/1719351990570_TimelineEvent-phase2.ts
```diff
@@ -0,0 +1,103 @@
+import {Kysely, PostgresDialect} from 'kysely'
```
rafaelromcar-parabol, line 1: Why is this file shown here as diff if it was already merged in the PR of the phase 2? 🤔 https://github.com/ParabolInc/parabol/pull/9875/files there should not be any diff between them.

mattkrick: gotta merge master into this PR & then it'll go away

PR #9881, opened 2024-06-26
fix: Avoid adding embedding jobs without metadata id

Dschoordsch:
# Description

When no new metadata was inserted, the resulting empty array would be fully outer joined with the models, resulting in [undefined, 'ember_1'] pairs.

## Testing scenarios

- [ ] start the server and see no jobs with `embed:start` and `embeddingsMetadataId === null` is added to the EmbeddingsJobQueue table

## Final checklist


mattkrick:
nice find! what's the culprit on why the metadata is undefined? did it previously get inserted & hit the onConflict clause?

Dschoordsch:
It happens when nothing is inserted. The full outer join of `[]` and `['ember1']` is `null, "ember1"`.

mattkrick (approved):

PR #9883, opened 2024-06-26
chore(rethinkdb): Organization: Phase 1

mattkrick:
# Description

Writes to PG

Things to test:
- [x] terminateSubscription
- [x] stripeUpdateCreditCard
- [x] upgradeToTeamTier


packages/server/postgres/migrations/1719351990570_TimelineEvent-phase2.ts
```diff
@@ -0,0 +1,103 @@
+import {Kysely, PostgresDialect} from 'kysely'
```
rafaelromcar-parabol, line 1: This file should not be here, isn't it? 🤔 I wonder if there are other files in the PR that shouldn't be here too 🤔

Issue #9862, opened 2024-06-24
PDF Downloads are only showing the first page of a summary


kendradixon:
When selecting the Download PDF from the meeting summary page, only the first page is getting added to the PDF.

PR #9889, opened 2024-06-27
fix: download pdf shows multiple pages

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/9862

In production, download to pdf just downloads the first page, rather than all pages. This is because `window.print()` just shows the visible part of the web page in a SPA.

In this PR, I'm opening a new window and rending the HTML as a static snapshot so that all of the content is visible. 

The pdf shows "about:blank" at the bottom left of the page. I don't think this can be removed as the browser checks the URL. And there's some box shadow at the end of the summary in the PDF, but I think that's fine. 

### To test

- [ ] Create a retro with a lot of content
- [ ] In the meeting summary, click "Download to PDF"
- [ ] See that it downloads every page, not just the first

### DIFF
--- a/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/EmailReflectionCard.tsx
+++ b/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/EmailReflectionCard.tsx
@@ -18,6 +18,7 @@ const contentStyle = {
   borderStyle: 'solid',
   borderWidth: '1px',
   boxSizing: 'content-box',
+  breakInside: 'avoid',
   color: PALETTE.SLATE_700,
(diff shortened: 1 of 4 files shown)


packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/EmailReflectionCard.tsx
```diff
@@ -18,6 +18,7 @@ const contentStyle = {
   borderStyle: 'solid',
   borderWidth: '1px',
   boxSizing: 'content-box',
+  breakInside: 'avoid',
```
Dschoordsch, line 21: I snuck in this to avoid having a page break in a reflection card.

Dschoordsch (approved):
Looks good

PR #9902, opened 2024-06-28
fix: speed up team upgrade

mattkrick:
# Description

Upgrading via stripe takes too long!
It's still too long, but this shaves off about 2 seconds (in the video it was 6!) & now you don't need to listen to webhooks to test locally.

## Demo

https://www.loom.com/share/a2482e91d4b940e19dd31f66fb7fa119?sid=67a06848-bc0e-490d-85ac-3d230073d975

### DIFF
--- a/packages/server/graphql/public/mutations/createStripeSubscription.ts
+++ b/packages/server/graphql/public/mutations/createStripeSubscription.ts
@@ -1,7 +1,6 @@
 import Stripe from 'stripe'
 import getRethink from '../../../database/rethinkDriver'
 import {getUserId} from '../../../utils/authorization'
-import {fromEpochSeconds} from '../../../utils/epochTime'
 import standardError from '../../../utils/standardError'
 import {getStripeManager} from '../../../utils/stripe'
 import {MutationResolvers} from '../resolverTypes'
@@ -12,7 +11,6 @@ const createStripeSubscription: MutationResolvers['createStripeSubscription'] = 
   {authToken, dataLoader}
 ) => {
   const viewerId = getUserId(authToken)
-  const now = new Date()
   const r = await getRethink()
 
   const [viewer, organization, orgUsersCount, organizationUser] = await Promise.all([
@@ -44,7 +42,7 @@ const createStripeSubscription: MutationResolvers['createStripeSubscription'] = 
     const {id: customerId} = customer
     const res = await manager.attachPaymentToCustomer(customerId, paymentMethodId)
     if (res instanceof Error) return standardError(res, {userId: viewerId})
-    // wait until the payment is attached to the customer before updating the default payment method
+    // cannot updateDefaultPaymentMethod until it is attached to the customer
     await manager.updateDefaultPaymentMethod(customerId, paymentMethodId)
   } else {
     customer = await manager.createCustomer(orgId, email, paymentMethodId)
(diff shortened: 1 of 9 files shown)


packages/server/graphql/public/mutations/createStripeSubscription.ts
```diff
@@ -38,14 +38,18 @@ const createStripeSubscription: MutationResolvers['createStripeSubscription'] =
...
-  // wait until the payment is attached to the customer before updating the default payment method
-  await manager.updateDefaultPaymentMethod(customerId, paymentMethodId)
+  let customer: Stripe.Response<Stripe.Customer | Stripe.DeletedCustomer>
+  if (stripeId) {
+    customer = await manager.retrieveCustomer(stripeId)
+    const {id: customerId} = customer
+    const res = await manager.attachPaymentToCustomer(customerId, paymentMethodId)
+    if (res instanceof Error) return standardError(res, {userId: viewerId})
+    // wait until the payment is attached to the customer before updating the default payment method
+    await manager.updateDefaultPaymentMethod(customerId, paymentMethodId)
```
nickoferrall, line 48: For the edge case where a user is upgrading for a second time, we could use a `Promise.all` to speed things up a bit here?  Something like:

```
  const [attachPaymentResult, updateDefaultPaymentResult] = await Promise.all([
    manager.attachPaymentToCustomer(customerId, paymentMethodId),
    manager.updateDefaultPaymentMethod(customerId, paymentMethodId)
  ])
  if (attachPaymentResult instanceof Error) return standardError(attachPaymentResult, {userId: viewerId})
```

packages/server/graphql/public/mutations/createStripeSubscription.ts
```diff
@@ -38,14 +38,18 @@ const createStripeSubscription: MutationResolvers['createStripeSubscription'] =
...
+  } else {
+    customer = await manager.createCustomer(orgId, email, paymentMethodId)
+  }
+
```
nickoferrall, line 52: And we could remove the `await` on line 65 when updating rethinkdb?

nickoferrall (approved):
Looks great! With this PR, we don't need to listen to webhooks to test the card payment, but the org isn't upgraded to the `team` tier as `upgradeToTeamTier` isn't called via the webhook. 

Another possible way of speeding up the Stripe API calls might be upgrading from versions 2020-08-27 to 2024-06-20, but I don't see anything about relevant performance improvements in their changelog: https://docs.stripe.com/upgrades#api-changelog

packages/server/graphql/public/mutations/createStripeSubscription.ts
mattkrick: i was thinking about that! but since our DB is colocated it's usually only a couple ms so I figured probably better to be safe than sorry in case the write failed for some reason.

mattkrick: On second look, we should completely get rid of the write to the DB here! 
We should only update our DB when the payment succeeds, otherwise if we update the stripeSubscriptionId & then the 3D payment fails, the org gets stuck in a weird state where it can't upgrade because it has a subscription, but it can't downgrade because it's still on the starter tier 😬

packages/server/graphql/public/mutations/createStripeSubscription.ts
mattkrick: i tried this...
nope, gotta attach first. ah well, not too bad since re-upgrade is such an edge case!

PR #9908, opened 2024-07-01
chore: Show only available integrations

### LINKED ISSUE #9879: Env var to disable integration providers
## Issue - Enhancement

Most deployments at elevated security levels disallow the usage of Jira Cloud vs. Jira Server. Removing this integrations from the product would be beneficial to the user experience. Folks are already clicking Jira Cloud and getting an error when they should be clicking "Jira Server" instead.

Admin/SREs can:
   - Set an environment variable that blacklists certain integrations
   - Refer to documentation to tell them what the env var is and how to set it

**Estimated effort:** 8 points ([see CONTRIBUTING.md](https://github.com/ParabolInc/parabol/blob/master/CONTRIBUTING.md#points-and-sizes))

Dschoordsch:
# Description

Fixes #9879 
Hide Github, Atlassian and Slack integrations if the oauth environment variables are not set.

If a user has integrated with any of these in the past, then this is not changed. They will be able to use the integration until their token expires. This was already the case before. This PR only affects the visibility of the provider rows.

## Demo

https://www.loom.com/share/881a214cb5c042aba2f9df3a96f6f190?sid=658c14b3-8343-4d08-ac48-6cec0b67ed6d 🔒 
 
## Testing scenarios

- [ ] with the oauth variables for Github, Atlassian and Slack unset check
  - [ ] Poker scope
  - [ ] Standup Your Work drawer
  - [ ] Team integrations
  - [ ] Meeting timer
- [ ] repeat with variables set

## Final checklist


rafaelromcar-parabol:
LGTM!

Just one question: what happens if no integration is available? Shouldn't we show something? I think on the video you say some integrations will always be shown there but I think we should allow to disable those too.

rafaelromcar-parabol:
Also, IMHO this PR would require a full testing round before being released to production.

Dschoordsch:
> what happens if no integration is available? Shouldn't we show something?

I think we should ignore this case. This is only an edge case for private instances. These get support from us in any case, so it's not an issue.

> some integrations will always be shown there but I think we should allow to disable those too.

I can add 2 variables to disable Mattermost and MS Teams.

> Also, IMHO this PR would require a full testing round before being released to production.

No objection against testing, but it just touches some UI elements, no logic changed. Production should look exactly the same.

rafaelromcar-parabol:
> I think we should ignore this case. 

Agreed. I was just curious.

> I can add 2 variables to disable Mattermost and MS Teams.

If it's easy and quick, I think you should, but I'll approve the PR without that, as it's not a requirement.

> No objection against testing, but it just touches some UI elements, no logic changed. Production should look exactly the same.

I thought it might risk something other than the UI. In that case, go ahead @Dschoordsch !

rafaelromcar-parabol (approved):

PR #9910, opened 2024-07-01
chore: Read Gitlab server URL from env for prime integrations

Dschoordsch:
## Description

Right now it is hardcoded, but private Gitlab instances behave like the public one, so for some Parabol instances it might make sense to point it to a different Gitlab instance.

## Demo

https://www.loom.com/share/bd9d06ef81b74f74b46ca2e80af04383?sid=3e3369ab-3e94-4d7e-8fb2-7d960cffe032 🔒 

## Testing scenarios

- [ ] Set the URL
- [ ] run `yarn build && yarn predeploy`
- [ ] check the correct value is set in PG

## Final checklist


scripts/toolboxSrc/primeIntegrations.ts
```diff
@@ -8,7 +8,7 @@ const upsertGlobalIntegrationProvidersFromEnv = async () => {
       authStrategy: 'oauth2',
       scope: 'global',
       teamId: 'aGhostTeam',
-      serverBaseUrl: 'https://gitlab.com',
+      serverBaseUrl: process.env.GITLAB_SERVER_URL ?? 'https://gitlab.com',
```
rafaelromcar-parabol, line 11: Why a default value? IMHO we should just use the env GITLAB_SERVER_URL and on the .env.example we can have our default value.

rafaelromcar-parabol (approved):
Other than my inline comment, LGTM.

This would be a great use case for you to test the process of adding a new environment variable to the application 😸 You have Loom videos and documentation explaining how to do it, but please, let me know if you need any help with that 😃

scripts/toolboxSrc/primeIntegrations.ts
rafaelromcar-parabol: It's true that removing that default value is a breaking change. Ok, let's keep it there. 

I'm just against default hardcoded values in general 😸

PR #9936, opened 2024-07-04
fix: Read embedder URL from env

Dschoordsch:
I had hardcoded it before 🤦

## Demo

## Testing scenarios

- [ ] Scenario A
  - Step 1
  - Step 2...

- [ ] Scenario B
  - Step 1
  - Step 2....

## Final checklist

### DIFF
--- a/packages/server/graphql/public/types/User.ts
+++ b/packages/server/graphql/public/types/User.ts
@@ -28,6 +28,20 @@ import {UserResolvers} from '../resolverTypes'
 
 declare const __PRODUCTION__: string
 
+const MODEL = 'Embeddings_ember_1'
+const EMBED_URL = (() => {
+  try {
+    const availableModels =
+      process.env.AI_EMBEDDING_MODELS && JSON.parse(process.env.AI_EMBEDDING_MODELS)
+    return availableModels.find(
+      ({model}: {model?: string}) => model?.split(':')[1] === 'llmrails/ember-v1'
+    )?.url
+  } catch {
+    return undefined
+  }
+})()
+const SIMILARITY_THRESHOLD = 0.5
+
 const User: UserResolvers = {
   activity: async (_source, {activityId}, {dataLoader}) => {
     const activity = await dataLoader.get('meetingTemplates').load(activityId)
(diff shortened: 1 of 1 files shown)


packages/server/graphql/public/types/User.ts
```diff
@@ -28,6 +28,20 @@ import {UserResolvers} from '../resolverTypes'
 
 declare const __PRODUCTION__: string
 
+const MODEL = 'Embeddings_ember_1'
+const EMBED_URL = (() => {
+  try {
+    const availableModels =
+      process.env.AI_EMBEDDING_MODELS && JSON.parse(process.env.AI_EMBEDDING_MODELS)
+    return availableModels.find(
+      ({model}: {model?: string}) => model?.split(':')[1] === 'llmrails/ember-v1'
```
rafaelromcar-parabol, line 37: @Dschoordsch why is `llmrails/ember-v1` hardcoded? That's just a possibility but in the future we might have other thing there. I might be lacking of JS knowledge to read this expression 😺

Dschoordsch: You're right, it's hardcoded. It's also hardcoded here: https://github.com/ParabolInc/parabol/blob/46770a048bfd4cd21c0e1cab89abdc01f8a79af3/packages/embedder/ai_models/TextEmbeddingsInference.ts#L14

I tried to strike the balance between getting stuff out now vs. making it future proof. Even if we add more models, we would need to test which one to use here and adjust accordingly. Having it configurable via environment for example add some complexity which might never be needed.

rafaelromcar-parabol: If I'm not mistaking, on that file TextEmbeddingsInference.ts you have that model but there is [another one there too](https://github.com/ParabolInc/parabol/blob/46770a048bfd4cd21c0e1cab89abdc01f8a79af3/packages/embedder/ai_models/TextEmbeddingsInference.ts#L8).

rafaelromcar-parabol: If I'm not mistaking, on that file TextEmbeddingsInference.ts you have that model but there is [another one there too](https://github.com/ParabolInc/parabol/blob/46770a048bfd4cd21c0e1cab89abdc01f8a79af3/packages/embedder/ai_models/TextEmbeddingsInference.ts#L8).

PR #9943, opened 2024-07-09
chore(rethinkdb): phase 4 of RetroReflection, RetroReflectionGroup and TimelineEvent

rafaelromcar-parabol:
# Description

These three tables represent almost 50% of the storage and almost 55% of the documents for RethinkDB. They are not used anymore since almost two weeks for TimelineEvent and RetroReflection, and a month for RetroReflectionGroup. We are not writing new data to them and if we find any problem we should fix it in PostgreSQL.

This PR allows us to reduce backup and restoration time. We shouldn't need to delete any more tables until we are done with the migration, but these three are good candidates. 

## Summary by CodeRabbit

- **Chores**
  - Migrated database to drop the `TimelineEvent`, `RetroReflectionGroup`, and `RetroReflection` tables using RethinkDB.

### DIFF
--- a/packages/server/postgres/migrations/1720517256158_TimelineEvent-phase4.ts
+++ b/packages/server/postgres/migrations/1720517256158_TimelineEvent-phase4.ts
@@ -1,0 +1,15 @@
+import {r} from 'rethinkdb-ts'
+import connectRethinkDB from '../../database/connectRethinkDB'
+
+export async function up() {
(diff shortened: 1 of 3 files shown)


Dschoordsch (approved):
I see no issues merging it, but we should wait for @mattkrick to approve it.

tianrunhe (approved):

PR #9956, opened 2024-07-11
fix: colors of the prompts for the threat level retro match now the prompts' names

### LINKED ISSUE #9852: Fix pre-mortem template colours
Jordan reported [here](https://parabol.slack.com/archives/C4JAUUZ9P/p1718648952828739) that the colours for the pre-mortem template are off:

image

### AC

- [ ] Prompt colours should match the prompt name emoji colours

rafaelromcar-parabol:
# Description

Fixes #9852

https://www.loom.com/share/7f5ddf56c4594e1692e266d556d95298

## Summary by CodeRabbit

- **Bug Fixes**
	- Updated group colours in the ReflectPrompt table to correctly reflect threat levels.

### DIFF
--- a/packages/server/postgres/migrations/1720689742934_fixThreatLevelPromptColors.ts
+++ b/packages/server/postgres/migrations/1720689742934_fixThreatLevelPromptColors.ts
@@ -1,0 +1,50 @@
+import {r} from 'rethinkdb-ts'
+import connectRethinkDB from '../../database/connectRethinkDB'
+
+export async function up() {
+  try {
+    await connectRethinkDB()
+
+    await r
+      .table('ReflectPrompt')
+      .filter({id: 'threatLevelPremortemTemplateSeverePrompt'})
+      .update({groupColor: '#FD6157'})
+      .run()
+    await r
+      .table('ReflectPrompt')
+      .filter({id: 'threatLevelPremortemTemplateElevatedPrompt'})
+      .update({groupColor: '#FFCC63'})
+      .run()
+    await r
+      .table('ReflectPrompt')
+      .filter({id: 'threatLevelPremortemTemplateLowPrompt'})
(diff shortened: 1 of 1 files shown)


packages/server/postgres/migrations/1720689742934_fixThreatLevelPromptColors.ts
```diff
@@ -0,0 +1,42 @@
+import {r} from 'rethinkdb-ts'
+import connectRethinkDB from '../../database/connectRethinkDB'
+
+export async function up() {
+  await connectRethinkDB()
+
+  await r
+    .table('ReflectPrompt')
+    .filter({id: 'threatLevelPremortemTemplateSeverePrompt'})
```
mattkrick, line 9: -1 use `get` instead of `filter`, otherwise it won't use the index

rafaelromcar-parabol: Done and tested locally again both migrating up and down! Thanks for the correction!

rafaelromcar-parabol:
Question @mattkrick: what about the migrations and their order? This fix adds a migration, but there have been new migrations added to master since then 🤔 what do you usually do when that happens? rebase, merge master into the branch? recreate the migration after rebasing/merging? just merge the branch into master like it is?

Dschoordsch:
@rafaelromcar-parabol you need to rename the migration so it is newer than anything on master before you merge. I usually just create a new migration and copy and paste, but there might be a more elegant solution.

mattkrick:
i usually down-migrate, rename, then up-migrate. no perfect way to do it, distributed state management is hard!

rafaelromcar-parabol:
Thanks for the tips! ❤️ and thanks for merging the branch!

PR #9975, opened 2024-07-15
fix: add ClearAll dataloader method

mattkrick:
# Description

Curious for your feedback on this pattern @Dschoordsch @nickoferrall.

The problem:
After the DB is mutated, devs have to call `clearAll` on every dataloader that could have been used & is now stale.  Adding a new dataloader requires auditing all existing DB mutations to make sure calls to clear it exist.

```ts
// imagine this line gets called early on in the resolver, nested a few functions deep, so it's not obvious it was called
const orgUsersByOrg = await dataLoader.get('organizationUsersByOrgId').load(orgId)

// resolver.ts
const orgUsersByUserId = await dataLoader.get('organizationUsersByUserId').load(userId)
await pg.updateTable('OrganizationUser').set({foo: 2}).where('userId', '=', userId).where('orgId', '=', orgId).execute()

dataLoader.get('organizationUsersByUserId').clear(userId) // does not clear ByOrgId
dataLoader.get('organizationUsers').updateCache({orgId, userId}, {foo: 2}) // does not update the object in ByOrgId
const orgUsersByOrg = await dataLoader.get('organizationUsersByOrgId').load(orgId) // ERROR! Stale data
```

Proposed solution:
Immediately after a database mutation, wipe out any dataloaders that depend on that table
 
```ts
dataLoader.clearAll('organizationUsers') // clears any used dataloader that registers as a dependent of `organizationUsers`
const orgUsersByOrg = await dataLoader.get('organizationUsersByOrgId').load(orgId) // Fresh hit to the DB
```

## Summary by CodeRabbit

- **New Features**
  - Introduced a generic `saml` function for fetching SAML-related data.
  
- **Improvements**
  - Enhanced dataLoader functionality with a more flexible and extensible design.
  - Improved performance and consistency in data loading by clearing all `organizationUsers` entries.
  
- **Refactor**
  - Unified and simplified loader dependency management.
  - Refactored `normalizeResults` to improve performance.

- **Bug Fixes**
  - Corrected return type in `rethinkForeignKeyLoader` to ensure accurate data loading and caching.


packages/server/billing/helpers/adjustUserCount.ts
```diff
@@ -97,6 +96,7 @@ const addUser = async (orgIds: string[], user: IUser, dataLoader: DataLoaderWork
       tier: organization.tier
     })
   })
+  dataLoader.clearAll('organizationUsers')
```
mattkrick, line 99: now we don't have to wipe every single dataloader that could have been used.
we just wipe the primary key datalaoder & it wipes all dependents.

packages/server/dataloader/RootDataLoader.ts
```diff
@@ -41,79 +39,81 @@ const loaderMakers = {
...
-  typeof gcalLoaders
-type CustomLoaders = keyof CustomLoaderMakers
-type Uncustom<T> = T extends (parent: RootDataLoader) => infer U ? U : never
-type TypeFromCustom<T extends CustomLoaders> = Uncustom<CustomLoaderMakers[T]>
-
-// Use this if you don't need the dataloader to be shareable
-export interface DataLoaderInstance {
-  get<LoaderName extends Loaders>(loaderName: LoaderName): TypedDataLoader<LoaderName>
+// The RethinkDB logic is a leaky abstraction! It will be gone soon & this will be generic enough to put in its own package
+interface GenericDataLoader<TLoaders, TPrimaryLoaderNames> {
```
mattkrick, line 50: this is basically `TypedDataLoader`, but the old version took 350ms to load types, this takes ~150ms

packages/server/dataloader/customLoaderMakers.ts
```diff
@@ -85,7 +85,11 @@ export const serializeUserTasksKey = (key: UserTasksKey) => {
   return parts.join(':')
 }
 
-export const commentCountByDiscussionId = (parent: RootDataLoader) => {
+export const commentCountByDiscussionId = (
+  parent: RootDataLoader,
+  dependsOn: RegisterDependsOn
+) => {
+  dependsOn('comments')
```
mattkrick, line 92: custom dataloaders can register with a primary key loader.
when a mutation calls `dataLoader.clearAll('comments')` this dataloader, if initialized, will get wiped, too.

packages/server/dataloader/foreignKeyLoaderMaker.ts
```diff
@@ -18,19 +18,17 @@ type LoaderType<LoaderName extends LoaderKeys> =
...
-  ) => Promise<(LoaderType<LoaderName> & {id: string | number})[]>
+  fetchFn: (keys: readonly T[KeyName][]) => Promise<(T & {id?: string | number})[]>
 ) {
-  type T = LoaderType<LoaderName>
-  type PrimaryKeyT = string | number
   type KeyValue = T[KeyName]
-  return (parent: RootDataLoader) => {
-    const primaryLoader = parent.get(primaryLoaderKey) as DataLoader<PrimaryKeyT, T, PrimaryKeyT>
+  return (parent: RootDataLoader, dependsOn: RegisterDependsOn) => {
+    dependsOn(primaryLoaderKey)
```
mattkrick, line 30: foreignKey dataloaders are automatically tagged as dependents of the pk loaders

nickoferrall (approved):
I really like this! Not clearing the right dataloader is such a classic bug. This will help a lot!

PR #9997, opened 2024-07-18
fix: Filipino checkin greeting

jordanh:
# Description

The `content` and `language` fields of the Filipino checkin question were reversed


jordanh:
Yikes. I made this via the GitHub editor and it reformatted the file! I'm going to checkout locally and fix instead. Standby

PR #10016, opened 2024-07-23
fix(build): build prod exits with code 1 if anything goes wrong

### LINKED ISSUE #9995: Build process failing on Docker build instead on app build
Build Github Action does not fail on build step if it does not work correctly, it fails on the Docker build because it's missing stuff. [Example](https://github.com/ParabolInc/parabol/actions/runs/9912071743/job/27386120662).

rafaelromcar-parabol:
# Description

Fixes #9995

The script `prod.js` would not exit with code different than 0. I tested adding a `runChild("exit 1")` and it wouldn't fail or show anything at all, as the script completed successfully.

Changes on this PR do two things:
- Make `prod.js` exit with code 1 if anything goes wrong in any of the try-catch.
- Throw an exception if any of the webpack builds ends with a code > 0.

In the issue that started this work, at least one of the runChild had ended with exit code 1, the code only captured but no used anywhere. 

## Summary by CodeRabbit

- **Bug Fixes**
	- Enhanced error handling in the build process to ensure immediate termination on encountering errors.
	- Improved feedback for build failures with clearer exit code reporting.

- **New Features**
	- Introduced robust error handling and reporting during process execution, providing clearer feedback on execution status.

### DIFF
--- a/scripts/prod.js
+++ b/scripts/prod.js
@@ -2,8 +2,16 @@ const generateGraphQLArtifacts = require('./generateGraphQLArtifacts')
 const cp = require('child_process')
 
 const runChild = (cmd) => {
-  return new Promise((resolve) => {
-    const build = cp.exec(cmd).on('exit', resolve)
+  return new Promise((resolve, reject) => {
+    const build = cp.exec(cmd).on('exit', (code, signal) => {
+      if (code !== 0) {
+        reject(new Error(`Received non-zero exit code ${code}`))
+      } else if (signal) {
+        reject(new Error(`Received signal ${signal}`))
+      } else {
+        resolve()
+      }
+    })
     build.stderr.pipe(process.stderr)
   })
 }
@@ -19,7 +27,7 @@ const prod = async (isDeploy, noDeps) => {
 
   console.log('starting webpack build')
   try {
-    const webpacks = await Promise.all([
+    await Promise.all([
       runChild(
         `yarn webpack --config ./scripts/webpack/prod.servers.config.js --no-stats --env=noDeps=${noDeps}`
       ),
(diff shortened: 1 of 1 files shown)


scripts/prod.js
```diff
@@ -14,20 +14,28 @@ const prod = async (isDeploy, noDeps) => {
...
       runChild(
         `yarn webpack --config ./scripts/webpack/prod.client.config.js --no-stats --env=minimize=${isDeploy}`
       )
     ])
+    webpacks.forEach(
+      function (exitCode) {
+        if (exitCode > 0)
+          throw new Error(`Error during one of the webpack build processes. Webpacks error codes ${webpacks}`)
+      }
+    )
```
Dschoordsch, line 35: -1 I think we should fix `runChild` instead to reject the promise if the process did not successfully exit. Something along the lines of
```ts
const runChild = (cmd) => {
  return new Promise((resolve, reject) => {
    const build = cp.exec(cmd).on('exit', (code, signal) => {
      if (code) {
        reject(new Error(`Received non-zero exit code ${code}`))
      }
      if (signal) {
        reject(new Error(`Received signal ${signal}`))
      }
      resolve()
    })
    build.stderr.pipe(process.stderr)
  })
}
```

Dschoordsch (changes requested):

scripts/prod.js
mattkrick: this feels right! only nit is to use an `else if` so we don't resolve after we reject

rafaelromcar-parabol:
Sorry I couldn't work on this before Thursday. I'm going to spend some time on it tomorrow 😃

mattkrick:
no worries! I've got a breaking build in another PR & figured I should merge this one to get to the bottom of it.

PR #10019, opened 2024-07-23
chore(SDL): refactor _legacy.graphql into individual typeDefs

mattkrick:
# Description

Refactoring into the SDL pattern has been a pain because we need to remove the typeDef from the _legacy.graphql file & put it into its own typeDef. We do this because it's 8000 LOC & ripe for merge conflicts.

I wrote some code that did this for us for every type.
Now refactoring to the SDL pattern just requires transforming the resolve statements, no need to touch .graphql files 🎉

How is it broken out?
- Every type has its own file, even one-liners like unions & scalars
- Mutations & Queries are types too, so you'll see a `Mutation.graphql`
- Additionally, we have _xGitHub and _xGitLab.graphql, which override their integrations


packages/server/graphql/public/typeDefs/Mutation.graphql
```diff
@@ -0,0 +1,2018 @@
+type Mutation {
```
Dschoordsch, line 1: I liked the previous way it was split into multiple files better: put each mutation in its own file and add the related return and input types there too. The second part of this might be a bit harder to automate, but then you have all related types in one file and don't need to jump between them too often.

nickoferrall (approved):
Love this!

packages/server/graphql/public/typeDefs/Mutation.graphql
mattkrick: It was definitely nice grouping them together & having fewer files!
The question I kept asking myself is, "why are we treating the Mutation object differently from other objects?"

If the mutation returned a `User` object, we wouldn't put the `User` definition in there, we'd give it its own file because it's shared across definitions.
Using that same logic, `UpdateMeetingPromptSuccess` type is shared with the `Subscription` object, so why does that get grouped with in the same file as the mutation field?

Moving forward, I'm down to keep using out `extend` pattern! TBH I don't really care where the defs are stored, I just use search anyways. I just wanted to get rid of the stale definitions in `_legacy.graphql` that were holding us back.

How should we proceed? Could we merge this now, or should we refactor manually?

Dschoordsch: It's just a preference. I don't see a reason why we shouldn't merge it as is. Better than manually moving types out of `_legacy.graphql`.

Issue #10040, opened 2024-07-26
Insights: update meeting summary UI


nickoferrall:
If we go with the summary of summaries approach, the new meeting summaries (created here https://github.com/ParabolInc/parabol/issues/10038) will include links to discussions. We'll need to update the UI to show the links as [links](https://parabol.co/)  rather than `[links](https://parabol.co/)`.

### AC

- [ ] Update meeting summary UI to handle links

PR #10081, opened 2024-08-06
feat: update meeting summary UI

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/10040

### To test

- [ ] Replace `wholeMeetingSummary` [here](https://github.com/ParabolInc/parabol/blob/0f98d26c304e749181adae271b99f6bfc567843f/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/WholeMeetingSummaryResult.tsx#L68) with dummy text that has markdown, e.g.: `The budget work is recognized for its future value in improving focus and communication within the team ([link](https://action.parabol.co/meet/twbP2qPXNK/discuss/3](https://action.parabol.co/meet/twbP2qPXNK/discuss/3) )).'`
- [ ] See that the link shows up as a formatted link


packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/WholeMeetingSummaryResult.tsx
```diff
@@ -78,7 +94,7 @@ const WholeMeetingSummaryResult = (props: Props) => {
             </td>
           </tr>
           <tr>
-            <td style={textStyle}>{wholeMeetingSummary}</td>
+            <td style={textStyle} dangerouslySetInnerHTML={{__html: formattedSummary}} />
```
Dschoordsch, line 97: -1 We should sanitize the html as marked also [recommends](https://marked.js.org). We already use DOMPurify, so let's use it here as well.

packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/WholeMeetingSummaryResult.tsx
```diff
@@ -55,16 +57,30 @@ const WholeMeetingSummaryResult = (props: Props) => {
...
+
+  const {summary: wholeMeetingSummary, team} = meeting
+
+  if (!wholeMeetingSummary) return null
+
+  const renderedSummary = marked(wholeMeetingSummary, {
+    gfm: true,
+    breaks: true
+  }) as string
+  // replace links with styled links
```
Dschoordsch, line 76: -1 this is obvious, but instead we should explain why we do this instead of doing a cleaner approach so it serves as a warning not to copy this.

Dschoordsch (changes requested):
Works, but I don't feel good merging it as is. At the very least we should sanitize it because I don't trust it. People might find a way to sneak html past ChatGPT somehow.

mattkrick:
look sanitary to me!

mattkrick (approved):

PR #10162, opened 2024-08-26
fix: consolidate org user menus

### LINKED ISSUE #10159: Org Admins cannot remove Org Admin role from user in the Plans & Billing page
[Slack thread](https://parabol.slack.com/archives/C4JAUUZ9P/p1724421028271749
) 🔒 

In order to remove an Org Admin role, an existing Org Admin needs to go the Members page. However, on the Plans & Billing page, they're prompted to contact love@parabol.co

We should allow users to remove Org Admin role from this page, too

Screenshot 2024-08-23 at 14 46 28

mattkrick:
# Description

fix #10159 

We've had a mix of logic for org administration.
We could stand to make this simpler, but for now I wanted to get all the logic in 1 place.

TEST
- [ ] make sure actions that appear in the menu are as expected (I couldn't find a spec for this business logic. this _should_ exist on the server, but since it didn't already exist there, I didn't add it)


packages/server/graphql/public/fields/invoices.ts
```diff
@@ -30,14 +30,30 @@ export const invoices: NonNullable<UserResolvers['invoices']> = async (
...
-  const [session, upcomingInvoice, invoices] = await Promise.all([
+  const [sessionRes, upcomingInvoiceRes, invoicesRes] = await Promise.allSettled([
     manager.stripe.billingPortal.sessions.create({
       customer: stripeId,
       return_url: makeAppURL(appOrigin, `me/organizations/${orgId}/billing`)
     }),
     manager.retrieveUpcomingInvoice(stripeId),
     manager.listInvoices(stripeId)
   ])
+  if (
```
nickoferrall, line 41: Is this invoices logic related to the org user logic?

nickoferrall (approved):
Tested and it LGTM

packages/server/graphql/public/fields/invoices.ts
mattkrick: 100% invoices, i wanted to test it on the parabol org & it kept breaking in dev mode with prod data, so this prevents it from throwing a fatal error

PR #10181, opened 2024-08-30
fix: move to rrule-rust

### LINKED ISSUE #9722: Bug: processRecurrence is slow
The mutation takes >15s, see [logs](https://app.datadoghq.com/apm/traces?query=%40_top_level%3A1%20resource_name%3A%22mutation%20ProcessRecurrence%22&agg_m=count&agg_m_source=base&agg_t=count&cols=core_service%2Ccore_resource_name%2Clog_duration%2Clog_http.method%2Clog_http.status_code&fromUser=false&graphType=span_list&historicalData=true&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=time&spanType=service-entry&traceQuery=&view=spans&start=1714472919314&end=1715077719314&paused=false) 🔒 We only run it every 5min so it's not an issue at the moment, but it should be faster.
Also testing processRecurrence times out sometimes https://github.com/ParabolInc/parabol/actions/runs/8982173873/job/24669192689?pr=9720

mattkrick:
# Description

fix #9722

rrule is stalling our gql executor whenever it runs. since it is sychronous, that means all pending requests get timed out.
image

there's a rust version without the bugs, so this moves to that.
The API is different, primarily dtstart and tzid only exist on RRuleSet.
Additionally, I found a ton of bugs with how we were handling timezones, so I fixed the ones I could find.

Most notably is how we handle timezones on the client.
Before, when you edited a rrule that was created in a different timezone, it kept the time the same, so it reported 11PM eastern as 11PM pacific.
Now it reports that accurately.

TEST
- [x] create gcal event
- [x] process
- [x] update


packages/client/shared/rruleUtil.ts
```diff
@@ -1,21 +1,33 @@
...
-    date.getMonth() + 1,
-    date.getDate(),
-    date.getHours(),
-    date.getMinutes()
-  )
+dayjs.extend(customParsePlugin)
+dayjs.extend(utcPlugin)
+dayjs.extend(timezonePlugin)
+
+// the RRule package requires dstart to be a date object set to a negative UTC offset. It's ugly!
```
Dschoordsch, line 11: +1 I don't see where we set a negative UTC offset here or what that means.

Dschoordsch (approved):
Works as far as I can tell. It would be nice to have fewer different date representations in the code base, like getting rid of the RRule package on the client side, but it's good for now.

packages/client/shared/rruleUtil.ts
mattkrick: Same as the old helper functions in this file, the rrule package stores `dtstart` as a Date, but interprets the local time as UTC time. e.g. it'll create a Date of 1/1/2020 8:00AM PST and then interpret that as 1/1/2020 8:00AM UTC. It's pants-on-head stupid. https://github.com/jkbrzt/rrule/issues/501#issuecomment-1146513557.

Issue #10099, opened 2024-08-08
Refactor featureFlags proposal


nickoferrall:
### Problems

Today, featureFlags are an array of enums in both the organization and user tables. Soon, we may want to add a team feature flag too, e.g. Team Insights.

Some flags exist in both tables, e.g. `noAISummary`. 

Feature flags stay in the app until we decide to remove them. `zoomTranscription` and `standupAISummary`, for example, have been in the app for over a year. This leads to our experience using the Parabol org to be quite different to the experience of all other users.

### Proposal

Create a `featureFlag` table with the following properties:

```
id (string)
featureName (enum)
isEnabled (boolean)
description (string, optional)
userId (string, optional)
orgId (string, optional)
teamId (string, optional)
expirationDate (timestamp, optional)
createdAt (timestamp)
updatedAt (timestamp)
```

Devs are encouraged to set an expiration date for the feature flag to force us to make a decision on a feature, but it's an optional field, so we don't have to include it if we're not sure. 

In the server, we can check if the feature flag exists like so:

```
const Organization: OrganizationResolvers = {
  featureFlag: async ({ id: orgId }, { name }, { dataLoader }) => {
    const flags = await dataLoader.get('featureFlagsByOrgId').load(orgId);
    const flag = flags.find((flag) => flag.featureName === name);
    return !!(flag && flag.isEnabled && (!flag.expirationDate || new Date(flag.expirationDate) > new Date()));
  },
```

And in the client:

```
const organization = useFragment(
  graphql`
    fragment RetroDiscussionThreadHeader_organization on Organization {
      publicTeams: featureFlag(name: PUBLIC_TEAMS)
      relatedDiscussions: featureFlag(name: RELATED_DISCUSSIONS)
    }
  `,
);

const hasPublicTeamsFlag = organization.publicTeams;
const hasRelatedDiscussionsFlag = organization.relatedDiscussions;
```

Let me know what you think!

Dschoordsch:
Why have userId and teamId if the interface only allows querying for orgId?
Having `featureName` being an enum is a pain if we want to remove a flag again at some point.

Dschoordsch:
> Yeah, would you prefer it to be a string? My concern is that it'd be easy to make a typo

I think we need to run a migration in any case, so we could add a check on a string rather than a full blown enum type. But maybe there are cleaner ways?

mattkrick:
- I really like querying a single flag!
[...]

PR #10184, opened 2024-09-03
feat: add feature flag tables

nickoferrall:
Related issue: https://github.com/ParabolInc/parabol/issues/10099

This PR was the first in a series of PRs. I've merged all child PRs into this one.

It includes:

- Adding featureFlag and featureFlagOwner tables
- refactoring the array of featureFlag enums on the Org & User objects to the new feature flag tables
- featureFlag can now exist in the Org, User, or Team
- migrates `noAISummary`. This is now only scoped to the Org rather than the Org and User


nickoferrall:
Hey @Dschoordsch, would you mind reviewing just the migration in this PR? And not the rest of the code.

That's the building block for the rest of this refactor.

My plan:

- In this PR, I create the migration and `addFeatureFlag`, which is where we can add new feature flags like `Insights`.
- In a follow-up PR, I'm working on adding `applyFeatureFlag`, which applies the feature flag to the feature flag owner, so we can apply the `Insights` feature flag to a specific team, org, or user, for example. I'll also and `updateFeatureFlag`, and `deleteFeatureFlag`
- And then in another PR, I'll refactor the application to use the new feature flag

packages/server/postgres/migrations/1724949259979_addFeatureFlagTables.ts
```diff
@@ -0,0 +1,73 @@
...
+    .on('FeatureFlagOwner')
+    .columns(['teamId', 'featureFlagId'])
+    .execute()
+
+  await pg.schema
+    .createIndex('idx_feature_flag_owner_org')
+    .on('FeatureFlagOwner')
+    .columns(['orgId', 'featureFlagId'])
+    .execute()
+}
```
Dschoordsch, line 61: -1 I don't see how we would use these indexes.
I think what we want though is the same set of columns `['userId', 'featureFlagId']`, ... as unique constraints. Thanks to Matt I just learned that Postgres is treating `null` values as distinct by default[^1]

[^1]: https://www.postgresql.org/about/featurematrix/detail/392/

packages/server/postgres/migrations/1724949259979_addFeatureFlagTables.ts
```diff
@@ -0,0 +1,73 @@
...
+    .execute()
+
+  await pg.schema
+    .createTable('FeatureFlagOwner')
+    .addColumn('featureFlagId', 'uuid', (col) =>
+      col.notNull().references('FeatureFlag.id').onDelete('cascade')
+    )
+    .addColumn('userId', 'varchar(255)')
+    .addColumn('teamId', 'varchar(255)')
+    .addColumn('orgId', 'varchar(255)')
```
Dschoordsch, line 32: -1 Why don't we make these foreign keys?

packages/server/postgres/migrations/1724949259979_addFeatureFlagTables.ts
```diff
@@ -0,0 +1,73 @@
...
+  await pg.schema
+    .createTable('FeatureFlag')
+    .addColumn('id', 'uuid', (col) => col.primaryKey().defaultTo(sql`gen_random_uuid()`))
+    .addColumn('featureName', 'varchar(255)', (col) => col.notNull())
+    .addColumn('scope', sql`"ScopeEnum"`, (col) => col.notNull())
+    .addColumn('description', 'text')
+    .addColumn('expiresAt', 'timestamptz', (col) => col.notNull())
+    .addColumn('createdAt', 'timestamptz', (col) => col.notNull().defaultTo(sql`CURRENT_TIMESTAMP`))
+    .addColumn('updatedAt', 'timestamptz', (col) => col.notNull().defaultTo(sql`CURRENT_TIMESTAMP`))
+    .addUniqueConstraint('unique_featureName_scope', ['featureName', 'scope'])
```
Dschoordsch, line 22: +1 What's the point of having one feature flag applied to different scopes?

Dschoordsch (changes requested):
Just a few smaller things, looks good otherwise.

packages/server/postgres/migrations/1724949259979_addFeatureFlagTables.ts
nickoferrall: Maybe I can remove `Scope` altogether? That way one `FeatureFlag` could apply to users, orgs, and teams?

Dschoordsch: Then we would need to handle conflicts if the feature flags for a user, their team and org differ. But it would make it more versatile. We could apply a flag to just one team if they want to test something even though we normally apply it to an org.
On the other hand, how would we know how to apply the feature flag to new users if it has no scope?

nickoferrall: I think it would be more common that we would apply the same feature flag to multiple scopes. For example, `aiSummary` could be applied to the org and the user.

With the `Scope` field, we'd have to create two `aiSummary` feature flags, one with `Scope` `Org` and one with `User`, even though they'd have the same expiry date and description.

> On the other hand, how would we know how to apply the feature flag to new users if it has no scope?

I don't think that adding the scope would change how that would work. When the new user is created, we'd get the `FeatureFlag` by the `featureName` and apply it to the new user.

I lean towards removing `Scope` so we have more versatility, less repeated `FeatureFlag`s, and we can have one less column. What do you think?

Dschoordsch: How do you know which feature flags to apply to new users?

nickoferrall: Wouldn't it be the same way we're adding feature flags here: https://github.com/ParabolInc/parabol/blob/d00da10595e85bd6cd7004b621e17efb2522b995/packages/server/graphql/mutations/helpers/bootstrapNewUser.ts#L50 to new users?

Dschoordsch: But that would only apply to user feature flags, otherwise we would call the same logic on new team or new org depending on scope.

nickoferrall: I don't understand your concern. Could you share an example of the problem and why the `Scope` field would help?

Dschoordsch: If we apply the feature flag `aiSummary` to orgs, then let's say Org A has it set and we have the entry (orgA, aiSummary) in the cross table. Now user Bob signs up. If we just apply it in `bootstrapNewUser` we might add the relation (userBob, aiSummary). This would be ok in this case, but what if we now remove it from Org A, Bob would still have the flag. Could also be that Org A does not have it and only Bob gets it. We would also need to check in all places where we use it, if Bob has the user feature flag, or the team we're looking at right now has it or if the org the team belongs to has the flag.
We could just accept this case and say a flag is set if it's true for either the user, team or org depending on the current scope. That's ok so long we don't introduce conflicting feature flags.

nickoferrall: I _think_ I understand. I'll go ahead and drop the `Scope` field and make sure we don't have conflicting feature flags. 

I plan on querying it on the client like so:

```
const organization = useFragment(
  graphql`
    fragment RetroDiscussionThreadHeader_organization on Organization {
      publicTeams: featureFlag(name: "publicTeams")
      relatedDiscussions: featureFlag(name: "relatedDiscussions")
    }
  `,
);

const hasPublicTeamsFlag = organization.publicTeams;
const hasRelatedDiscussionsFlag = organization.relatedDiscussions;
```

Dschoordsch: Still curious how your assignment logic for new users will look like or if we don't want to add any.
It's easy enough to add the column later, so it's not a blocker.

mattkrick: Reviewing #10234 led me here.
Removing the constraint that 1 feature flag has 1 scope will result in inconsistent results.
For example, say a user has `aiInsights` turned on for user, but not the org.
if we query `organization { featureFlag(featureName: "aiInsights")}` we will get false, but `user { featureFlag(featureName: "aiInsights")}` is true, yikes!

What we're essentially doing is creating 1 feature flag per name _per scope_. Is that the intended result?

nickoferrall: Hey @mattkrick, I was thinking that with my current approach (without `Scope`), you only need one entry into the `FeatureFlag` table. 

With the `Scope` column, we'd need to insert `aiInsight` twice, once with `Organization` as the `Scope` and once with `User`. `aiInsight` could then potentially have different expiry dates, which might be confusing. 


Getting inconsistent results for `organization { featureFlag(featureName: "aiInsights")` and the user query isn't great, although we did previously have `featureFlags` returning different enums depending on whether it was queried from the user or org.

Adding `Scope` feels a bit redundant to me when we're querying within the organization or user, e.g. `organization { featureFlag(featureName: "aiInsights", scope: "Organization")`, if that's what you have in mind?

I'm happy to add the `Scope` column and update the migration and queries, I'd just like to get a better understanding of why this isn't a good approach.

nickoferrall: > we could remove featureFlag from team & organization & query all the feature flags from the User GQL object

With that approach, if a user has `Insights` available for one team but not other teams and we query `user { featureFlag(featureName: "insights" }` wouldn't it always return true? Even for teams that shouldn't have access to Insights?

>  so the question we gotta consider is should 2 different feature flags share the same name, description, and expiry?

I was thinking that they probably would. We might want to turn on `aiSummary`, for instance, for some friendly users and orgs with the same expiry date, but I can also see why we'd want to have the option of having different expiry dates. 

I like your idea of including the prefix `user`, `org` or `team` if there are two flags with the same name. If we do that, do you think we need the `Scope` column in `FeatureFlag` as `viewer.userAIInsights` !== `orgnization.orgAIInsights`?

The `User` scope could feel redundant in: `user { featureFlag(featureName: "userAISummary" , scope: "User"}`?

But...I guess it does add extra clarity. Maybe we'd add the flag `Insights` to team, do `team { featureFlag(featureName: "Insights"}`, but later want to add `Insights` to org. Then we'd have to refactor all of the old queries to `team { featureFlag(featureName: "TeamInsights"}`.

In conclusion, I'll go with the original plan 😄 I'll add `Scope` to `FeatureFlag`. That way, we could add a flag `team { featureFlag(featureName: "Insights", scope: "Team"}`, and then if we add an org flag later, it'd be simple and clear to add: `organization { featureFlag(featureName: "Insights", scope: "Org"}`.

packages/server/graphql/public/typeDefs/FeatureFlag.graphql
```diff
@@ -0,0 +1,16 @@
+type FeatureFlagNew {
```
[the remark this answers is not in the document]
nickoferrall: I added `FeatureFlagNew` as a temporary name to minimise conflicts with the old `FeatureFlag`. Sorry, I should have added a comment to clarify that. It's `FeatureFlag` in the subsequent PR.

PR #10189, opened 2024-09-06
feat: implement apply, update, and delete feature flag

nickoferrall:


packages/server/graphql/private/typeDefs/DeleteFeatureFlagSuccess.graphql
```diff
@@ -0,0 +1,6 @@
+type DeleteFeatureFlagSuccess {
+  """
+  The deleted feature flag
+  """
+  featureFlag: FeatureFlagNew!
```
nickoferrall, line 5: Once I remove the existing `FeatureFlag`, I'll change this from `FeatureFlagNew` to `FeatureFlag`

nickoferrall:
Hey @Dschoordsch, would you mind giving this a light PR review to double check it's on the right track?

 In the next PR, I'll use these mutations and types and refactor the code in the application. In that PR, I'll clean this up, pass the tests, and request a more thorough review.

packages/server/graphql/private/typeDefs/Mutation.graphql
```diff
@@ -14,10 +14,54 @@ type Mutation {
...
+    """
+    The subjects to which the feature flag will be applied
+    """
+    subjects: SubjectsInput!
+  ): ApplyFeatureFlagPayload!
+
+  """
+  Delete an existing feature flag
+  """
+  deleteFeatureFlag(
```
Dschoordsch, line 39: -1 I'm not sure what this mutation is supposed to do. Does it remove the feature flag from all subjects or does it delete the entry from the feature flag table? What happens to the static types in this case?
I think we probably don't need static types for feature flags but rather a field on Org, Team and User: `hasFeatureFlag(flagName: String!): Boolean!`. But make sure first that this actually would work.

packages/server/graphql/private/mutations/applyFeatureFlag.ts
```diff
@@ -0,0 +1,93 @@
...
+  const userIds: string[] = []
+  const teamIds: string[] = []
+  const orgIds: string[] = []
+
+  for (const subjectType of subjectKeys) {
+    const subjectValue = subjects[subjectType as keyof typeof subjects]
+    if (!subjectValue || subjectValue.length === 0) continue
+
+    switch (subjectType) {
+      case 'emails':
```
Dschoordsch, line 48: +1 Instead of looping over the keys, we could just do
```
if (subjects.emails) {
  ...
}
if (subjects.domains) {
  ...
}
```

packages/server/graphql/private/mutations/applyFeatureFlag.ts
```diff
@@ -0,0 +1,93 @@
...
+  }
+
+  const targetIds = [...userIds, ...teamIds, ...orgIds]
+
+  const values = targetIds.map((targetId) => ({
+    featureFlagId,
+    userId: userIds.includes(targetId) ? targetId : null,
+    teamId: teamIds.includes(targetId) ? targetId : null,
+    orgId: orgIds.includes(targetId) ? targetId : null
+  }))
```
Dschoordsch, line 77: +1 If we refactor this slightly
```ts
const targetIds = [
  ...userIds.map(userId => ({userId}),
  ...teamIds.map(teamId => ({teamId}),
  ...orgIds.map(orgId => ({orgId})
].map(subject => ({...subject, featureFlagId})
```
we wouldn't need to go O(n^2)

Dschoordsch (changes requested):
There is some inconsistency whether the feature flag is just a string or an enum in the mutations.

packages/server/graphql/private/typeDefs/Mutation.graphql
nickoferrall: It removes the flag from the `featureFlag` table. 

It's not strictly necessary as we can just update the expiry date of the flag so that it no longer applies. 

I thought it'd be helpful to include so that we can clean up flag names in the db that we're no longer using, but I can remove this if it's confusing.

packages/server/graphql/private/typeDefs/Mutation.graphql
```diff
@@ -14,10 +14,54 @@ type Mutation {
...
   ): AddFeatureFlagPayload!
+
+  """
+  Apply a feature flag to specified subjects (users, teams, or organizations)
+  """
+  applyFeatureFlag(
+    """
+    The name of the feature flag to apply
+    """
+    flagName: String!
```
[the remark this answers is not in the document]
nickoferrall: `DeleteFeatureFlagSuccess` returns the type `featureFlag: FeatureFlagNew!` (after the refactor it'll be `FeatureFlag` instead of `FeatureFlagNew`), which is the `FeatureFlag` object, not an enum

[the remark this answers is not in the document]
Dschoordsch: Sorry, missed this type because I only checked this PR. All good then.

packages/server/graphql/private/typeDefs/Mutation.graphql
Dschoordsch: I think that's fine then. It will also remove it from all owners. Maybe update the description to match it?

Dschoordsch (approved):

PR #10198, opened 2024-09-09
fix: Remove duplicate org users

### LINKED ISSUE #9302: Bug: duplicate OrganizationUser objects are created in some cases
A newly added test to check that we're not creating duplicated `OrganizationUser` objects is failing randomly. It appears there are still duplicates created in certain cases.

> @Dschoordsch it looks like this produced a flaky test result for me-- failed first test, passed on second try. 
>
> https://github.com/ParabolInc/parabol/actions/runs/7278378730/job/19832437203. Should I make an issue for this?
>
>edit: happened again, so likely not a fluke: https://github.com/ParabolInc/parabol/actions/runs/7278798261/attempts/1?pr=9295

_Originally posted by @mattkrick in https://github.com/ParabolInc/parabol/issues/9290#issuecomment-1864849230_

Dschoordsch:
# Description

Fixes #9302
Add a constraint that forbids duplicate organization users. There is no need to have multiple entries as the join/leave events are already captured in OrganizationUserAudit.

## Demo

## Testing scenarios

- [ ] Remove a user Joe from an organization
- [ ] Add Joe again
- [ ] check that Joe can see the org correctly

## Final checklist

### DIFF
--- a/packages/server/billing/helpers/adjustUserCount.ts
+++ b/packages/server/billing/helpers/adjustUserCount.ts
@@ -78,16 +78,14 @@ const addUser = async (orgIds: string[], user: IUser, dataLoader: DataLoaderWork
     .insertInto('OrganizationUser')
     .values(docs)
     .onConflict((oc) =>
-      oc
-        .constraint('unique_org_user')
-        .doUpdateSet({
-          joinedAt: sql`CURRENT_TIMESTAMP`,
-          removedAt: null,
-          inactive: false,
-          role: null,
-          suggestedTier: null,
-          tier: (eb) => eb.ref('excluded.tier')
-        })
+      oc.constraint('unique_org_user').doUpdateSet({
+        joinedAt: sql`CURRENT_TIMESTAMP`,
+        removedAt: null,
+        inactive: false,
+        role: null,
+        suggestedTier: null,
+        tier: (eb) => eb.ref('excluded.tier')
+      })
     )
     .execute()
   await Promise.all(
--- a/packages/server/billing/helpers/adjustUserCount.ts
+++ b/packages/server/billing/helpers/adjustUserCount.ts
@@ -61,19 +61,13 @@ const changePause = (inactive: boolean) => async (_orgIds: string[], user: IUser
 
 const addUser = async (orgIds: string[], user: IUser, dataLoader: DataLoaderWorker) => {
   const {id: userId} = user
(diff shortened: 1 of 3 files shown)


packages/server/billing/helpers/adjustUserCount.ts
```diff
@@ -83,7 +77,16 @@ const addUser = async (orgIds: string[], user: IUser, dataLoader: DataLoaderWork
...
+    .onConflict((oc) =>
+      oc.constraint('unique_org_user').doUpdateSet({
+        joinedAt: sql`CURRENT_TIMESTAMP`,
+        removedAt: null,
+        inactive: false,
+        role: null,
+        suggestedTier: null,
+        tier: (eb) => eb.ref('excluded.tier')
+      })
+    )
```
Dschoordsch, line 89: I'm mostly concerned with this part. Previously we just created a new org user when a new user joined, but now we have to reset all the old fields that should be reset.

packages/server/postgres/migrations/1725438336254_addOrganizationUserUniqueConstraint.ts
```diff
@@ -0,0 +1,45 @@
...
+           )
+        )
+        -- or newer non-removed
+        OR (k."removedAt" IS NULL AND d."removedAt" IS NULL AND k.id > d.id)
+      );
+  `.execute(pg)
+
+  await pg.schema
+    .alterTable('OrganizationUser')
+    .addUniqueConstraint('unique_org_user', ['orgId', 'userId'])
```
mattkrick, line 34: back in the days of proration, we needed to keep track of `removedAt` since some users would join an org, then leave it, then join again & we didn't want to treat them as a new user with a grace period.

I see the unique constraint doesn't include `removedAt`, which means only 1 row per user per org, so if a user joins, then leaves, then joins again, what happens?
The logic here would say `removedAt` doesn't get reset: https://github.com/ParabolInc/parabol/blob/63bf930989f1493f54d36407a990863519b80528/packages/server/billing/helpers/adjustUserCount.ts#L86. At least I think, I didn't actually test this yet

packages/server/billing/helpers/adjustUserCount.ts
mattkrick: LGTM! i don't think we use trialStartDate anymore so we can ignore that column.

mattkrick (approved):

PR #10234, opened 2024-09-16
feat: refactor feature flags 

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/10099

This PR refactors feature flags from the array of enums to the new table approach.

The parent PRs added the new feature flag tables and mutations. Here I'm updating all of the references.

This PR can't be merged until we migrate old necessary feature flags over to the new tables, see: https://github.com/ParabolInc/parabol/issues/10248

I'll also create a subsequent PR to notify the team via cron job when a feature flag is expiring: https://github.com/ParabolInc/parabol/issues/10249. 

### To test

- [ ] Add a feature flag with the `addFeatureFlag` mutation
- [ ] Apply a feature flag to a user or org with `applyFeatureFlag`
- [ ] See that the client reads the feature flag as expected


nickoferrall:
Sorry the PR is so large! I tried to break it down by splitting it into a few PRs, but I don't think there's a good way to make this one smaller.

packages/server/dataloader/customLoaderMakers.ts
```diff
@@ -843,58 +843,81 @@ export const meetingCount = (parent: RootDataLoader, dependsOn: RegisterDependsO
...
-//     },
-//     {
-//       ...parent.dataLoaderOptions,
-//       cacheKeyFn: (key) => `${key.ownerId}:${key.scope}`
-//     }
-//   )
-// }
+export const featureFlagByOwnerId = (parent: RootDataLoader) => {
+  return new DataLoader<
+    {ownerId: string; ownerType: 'User' | 'Team' | 'Organization'; featureName: string},
```
mattkrick, line 848: `ownerType` looks like a footgun,
see convo here: https://github.com/ParabolInc/parabol/pull/10184#discussion_r1763838801

packages/client/components/ActivityLibrary/ActivityLibrary.tsx
```diff
@@ -118,7 +118,7 @@ const query = graphql`
         }
       }
       organizations {
-        aiTemplate: featureFlag(featureName: "aiTemplate")
```
mattkrick: Help me understand this change here, why would someone request a scope of anything other than Organization in this example?  Couldn't it be inferred since this is a field under Organization?

packages/server/graphql/public/typeDefs/FeatureFlagScope.graphql
```diff
@@ -0,0 +1,8 @@
+"""
+The list of scopes available for feature flags
+"""
+enum FeatureFlagScope {
```
mattkrick, line 4: +1 can remove?

packages/client/components/RetroGroupPhase.tsx
```diff
@@ -64,9 +64,7 @@ const RetroGroupPhase = (props: Props) => {
         }
         organization {
           tier
-          featureFlags {
-            suggestGroups
-          }
+          hasSuggestGroupsFlag: featureFlag(featureName: "suggestGroups")
```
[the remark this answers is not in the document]
nickoferrall: Makes sense! I like the most explicit option of `has[name]Flag`, so I've updated all of them to follow that convention

packages/server/graphql/public/typeDefs/FeatureFlagScope.graphql
nickoferrall: We need this for the `addFeatureFlag` mutation

packages/server/dataloader/customLoaderMakers.ts
```diff
@@ -843,58 +843,62 @@ export const meetingCount = (parent: RootDataLoader, dependsOn: RegisterDependsO
...
+        if (missingFeatureNames.length > 0) {
+          throw new Error(`Feature flag name(s) not found: ${missingFeatureNames.join(', ')}`)
+        }
+      }
+
+      const results = await pg
+        .selectFrom('FeatureFlag')
+        .innerJoin('FeatureFlagOwner', 'FeatureFlag.id', 'FeatureFlagOwner.featureFlagId')
+        .where((eb) =>
+          eb.and([
```
mattkrick, line 873: +1 you could simplify this a little because FeatureFlag tells you what column to look in so you don't have to look in all 3 columns.

You also don't have to destructure the `(featureName, ownerId)` tuple. You could do something like:

```
WHERE 
  ("featureName", CASE "FeatureFlag"."scope"
    WHEN 'User' THEN "FeatureFlagOwner"."userId"
    WHEN 'Team' THEN "FeatureFlagOwner"."teamId"
    WHEN 'Organization' THEN "FeatureFlagOwner"."orgId"
  END) in :keys;
``` 

This will execute as something like `WHERE ("featureName", "userId") in (("insights", "user123"))`. That way you won't overfetch (albeit the overfetch is pretty minimal in this case)

mattkrick (commented):
Looks good! 1 little change on the SQL piece, other than that good to go!

Issue #10261, opened 2024-09-23
Bug: unpaid invoices on self-hosted Enterprise deployment


jordanh:
Platform one reported that a number of their organizations began displaying messages such as:

![image](https://github.com/user-attachments/assets/65623dd3-9fd3-433c-9c71-5613bbdd6161)

and

![image](https://github.com/user-attachments/assets/15f1db65-012e-45a7-ba9f-3312fc91713f)

## Issue - Bug

- **How serious is the bug?** P0

### Acceptance Criteria (optional)
Self-hosted Enterprise users:
   - Never see unpaid invoice messages

Triage is performed to find the root cause of the bug, timeboxed to ~1 hour. 

**Estimated effort:** 1 hour to triage. More if root cause is already identified.

dbumblis-parabol:
The affected team had the `isPaid` flag set to false. This team had been manually moved to an org via a mutation.

dbumblis-parabol:
Here's how the flag was most likely set to false
- A new org was created
- A new team was created within that org mistakenly
- The new team was then archived (deleted)
- The mutation was run to move the original team to the new org
- The mutation to move the teams contained logic that would set the flag to false if the org contained 0 unarchived teams (this was fixed in this [PR ](https://github.com/ParabolInc/parabol/pull/10263))

PR #10263, opened 2024-09-24
fix: isPaid flag when moving teams to 0-team org

mattkrick:
# Description

handles orgs without an unarchived team


dbumblis-parabol:
Fixes https://github.com/ParabolInc/parabol/issues/10261

Issue #10267, opened 2024-09-24
Remove team insights


nickoferrall:
See Slack thread here: https://parabol.slack.com/archives/C836NA350/p1727109432387389?thread_ts=1726736384.635569&cid=C836NA350

It'd be confusing to have Team Insights in the Activity tab and another Insight tab. Let's remove team insights in favour of the new Insights feature.

### AC

- [ ] Team insights safely removed

jordanh:
Done!

PR #10270, opened 2024-09-24
chore: remove team insights

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/10267

I've added the git tag remove-team-insights if we want to re-add the code later.


Dschoordsch (changes requested):
We should also remove the columns if we don't update them anymore, i.e. undo the migration: https://github.com/ParabolInc/parabol/blob/master/packages/server/postgres/migrations/1689949187848_addTeamInsights.ts

Dschoordsch (approved):

Issue #10239, opened 2024-09-17
Insights: show default insight


nickoferrall:
Predecessor: https://github.com/ParabolInc/parabol/issues/10238

### AC

- [ ] In the white sheet of paper, show a default insight with the teams wins & challenges over a period of time
- [ ] This issue does not include changing the date to see different insights
- [ ] This issue does not include reference topics

Issue #10281, opened 2024-10-01
Insights: show empty state


nickoferrall:
See designs: https://www.figma.com/design/eGGewFjqfNpfU3TUgaBSDo/Insights-UI-Views?node-id=0-1&node-type=canvas&t=pdJBEVrMqAUKcdO7-0

### AC

- [ ] Show empty state if Insight doesn't exist
- [ ] Include button to generate insight

PR #10283, opened 2024-10-01
feat: show default insight

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/10239

Screenshot 2024-10-01 at 16 46 05

You could test this PR by connecting to staging and running `generateInsight` from graphiql. However, I suggest just reviewing the code as we'll only release this to Parabol to start with once the team feature flag is ready.

In the next PR, I'll create an empty state where you can generate an insight if one doesn't exist.


packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/WholeMeetingSummaryResult.tsx
```diff
@@ -93,7 +93,7 @@ const WholeMeetingSummaryResult = ({meetingRef}: Props) => {
             <td
               align='center'
               style={textStyle}
-              className='summary-link-style'
```
nickoferrall, line 96: I updated this as link-style is being used in two places now, not just the summary

packages/client/modules/teamDashboard/components/TeamDashHeader/TeamDashHeader.tsx
```diff
@@ -109,15 +110,15 @@ const TeamDashHeader = (props: Props) => {
     `,
     teamRef
   )
-  const {organization, id: teamId, name: teamName, teamMembers} = team
+  const {organization, id: teamId, name: teamName, teamMembers, isViewerTeamLead} = team
```
nickoferrall, line 113: When the team feature flag is ready, I'll add a check for this too

packages/client/modules/teamDashboard/components/TeamDashInsightsTab/TeamInsights.tsx
```diff
@@ -22,23 +32,101 @@ const query = graphql`
...
           Give us feedback
         </a>
       </p>
       <div className='mx-auto aspect-[1/1.414] w-[640px] max-w-3xl overflow-y-auto rounded-lg bg-white px-[56px] pt-8 shadow-md'>
         <h2 className='mb-4 mt-0 flex items-center pt-0 text-2xl font-semibold leading-9'>
           <AutoAwesomeIcon className='mr-2 h-9 w-9 text-grape-500' />
-          <span>Insights - Aug to Sep 2024</span>
+          <span>Insights - {dateRange}</span>
         </h2>
+        <p className='mb-6 text-sm text-slate-600'>Summarized {insight?.meetingsCount} meetings</p>
```
nickoferrall, line 99: I'll add empty state soon (https://github.com/ParabolInc/parabol/issues/10281) so there's no need to check if insights is null

packages/client/modules/teamDashboard/components/TeamDashInsightsTab/TeamInsights.tsx
```diff
@@ -22,23 +32,101 @@ const query = graphql`
...
+                className='link-style'
+                dangerouslySetInnerHTML={{__html: renderMarkdown(win)}}
+              />
+            </li>
+          ))}
+        </ul>
+
+        <h3 className='mb-0 text-lg font-semibold text-slate-700'>Challenges</h3>
+        <p className='mb-2 mt-0 text-sm italic text-slate-600'>
+          What challenges has "{name}" faced during this timeframe?
```
nickoferrall, line 118: The design shows "this team" but I changed that to "name" as it's more personalised

packages/server/graphql/public/types/Team.ts
```diff
@@ -51,6 +51,16 @@ const Team: TeamResolvers = {
...
     return teamMembers.find((teamMember) => teamMember.isLead)!
+  },
+  isViewerTeamLead: async ({id: teamId}, _args, {authToken, dataLoader}) => {
+    const viewerId = getUserId(authToken)
+    const teamMemberId = toTeamMemberId(teamId, viewerId)
+    const teamMember = await dataLoader.get('teamMembers').load(teamMemberId)
+    return teamMember?.isLead || false
+  },
+  insight: async ({id: teamId}, _args, {dataLoader}) => {
+    const insight = await dataLoader.get('latestInsightByTeamId').load(teamId)
```
nickoferrall, line 62: I'm getting a type error. The problem seems to be that the id is an Int because I'm using SERIAL [here](https://github.com/ParabolInc/parabol/blob/41fc26856c2217bd7b4035d09b6d04b8e9d60ef6/packages/server/postgres/migrations/1722638208553_addInsight.ts). 

Insight.id is defined as type ID [here](https://github.com/ParabolInc/parabol/blob/fc6400af5e00ad56f3b7f8dbd19b38889782ff79/packages/server/graphql/public/typeDefs/Insight.graphql). 

But I get a type error, that's resolved when I do:

```
  insight: async ({id: teamId}, _args, {dataLoader}) => {
    const insight = await dataLoader.get('latestInsightByTeamId').load(teamId)
    if (!insight) return null
    return {
      ...insight,
      id: 'hey'
    }
  }
```

So the problem seems to be the id is an int, but it's expecting a string. I could change the type Insight.id to Int [here](https://github.com/ParabolInc/parabol/blob/fc6400af5e00ad56f3b7f8dbd19b38889782ff79/packages/server/graphql/public/typeDefs/Insight.graphql), but that doesn't seem clean when it's an ID. 

Any ideas?

nickoferrall: Ah that makes sense, thanks!

packages/server/graphql/public/types/Team.ts
```diff
@@ -52,15 +53,13 @@ const Team: TeamResolvers = {
...
-    const teamMember = await dataLoader.get('teamMembers').load(teamMemberId)
-    return teamMember?.isLead || false
-  },
   insight: async ({id: teamId}, _args, {dataLoader}) => {
     const insight = await dataLoader.get('latestInsightByTeamId').load(teamId)
-    return insight || null
+    if (!insight) return null
+    return {
+      ...insight,
+      id: InsightId.join(teamId, insight.id)
```
Dschoordsch, line 61: +1 The nicer solution would be to have a resolver for the insights type. Define the DB type as source for it and then have a single id field doing this conversion. This way it will work in all occurrences where we would add this field.

Dschoordsch (approved):

Issue #10248, opened 2024-09-17
Migrate old feature flags to new table


nickoferrall:
### AC

- [ ] Migrate necessary feature flags like `noAISummary` to new feature flag table

nickoferrall:
Complete by https://github.com/ParabolInc/parabol/pull/10293

PR #10293, opened 2024-10-03
chore: migrate noaisummary flag

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/10248

I think noAISummary should just exist on the org and not the org and user. That's a simpler approach. 

This PR migrates the noAISummary flag to the new featureFlag table and creates an org scoped flag. 

This is the only flag that I think we need to migrate.


mattkrick:
neat! didn't run the code, but it looks good from a distance!

mattkrick (approved):

Issue #10243, opened 2024-09-17
Insights: send feedback functionality


nickoferrall:
Predecessor: https://github.com/ParabolInc/parabol/issues/10242

### AC

- [ ] Track feedback for Insights

PR #10342, opened 2024-10-15
feat: insights feedback functionality

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/10243

### To test

- [ ] Add [insights feature flag](https://www.notion.so/parabol/How-to-enable-or-remove-a-feature-flag-4b4be190ed674bf8903340d6b5fa50e4), and go to a team where you're the team lead
- [ ] Go to the Insights tab and click "Give us feedback"
- [ ] See that it opens a modal
- [ ] Enter info and click submit. See it stores it in amplitude

### DIFF
--- a/packages/client/hooks/useManualClientSideTrack.ts
+++ b/packages/client/hooks/useManualClientSideTrack.ts
@@ -2,16 +2,13 @@ import {useCallback} from 'react'
 import SendClientSideEvent from '../utils/SendClientSideEvent'
 import useAtmosphere from './useAtmosphere'
 
-let eventId = 0
-
 const useManualClientSideTrack = () => {
   const atmosphere = useAtmosphere()
 
   const trackEvent = useCallback(
     (event: string, options: Record<string, any>) => {
       SendClientSideEvent(atmosphere, event, {
-        ...options,
-        eventId: ++eventId
+        ...options
       })
     },
     [atmosphere]
--- a/packages/client/hooks/useManualClientSideTrack.ts
+++ b/packages/client/hooks/useManualClientSideTrack.ts
@@ -1,0 +1,23 @@
+import {useCallback} from 'react'
+import SendClientSideEvent from '../utils/SendClientSideEvent'
+import useAtmosphere from './useAtmosphere'
+
+let eventId = 0
+
+const useManualClientSideTrack = () => {
+  const atmosphere = useAtmosphere()
+
+  const trackEvent = useCallback(
+    (event: string, options: Record<string, any>) => {
+      SendClientSideEvent(atmosphere, event, {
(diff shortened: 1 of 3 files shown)


nickoferrall:
Hey @tianrunhe, would you mind reviewing this one? 🙏  Curious to get your thoughts on the `useManualClientSideTrack` hook. I couldn't use the `useClientSideTrack` hook because we don't want to send the data when the modal opens, but it also seemed excessive to create a mutation when we're not updating the db.

Hopefully, `useManualClientSideTrack` could be a helpful pattern.

tianrunhe:
Sorry I didn't quite follow why you need a new hook? Can you explain a little bit more? Thank you!

nickoferrall:
@tianrunhe the `useClientSideTrack` tracks the event when the component loads, so when a modal is opened, for instance, we could track `Modal Opened`. 

But what if we don't want to track the event when the component loads, but when a user does an action, like submitting a form. 

At the moment, I think we can only handle this on the server, but creating a server mutation seems like overkill if you're not updating the db. So I created this hook so that we can track the action once an action happens, like submitting a form.

packages/client/hooks/useManualClientSideTrack.ts
```diff
@@ -0,0 +1,23 @@
...
+let eventId = 0
+
+const useManualClientSideTrack = () => {
+  const atmosphere = useAtmosphere()
+
+  const trackEvent = useCallback(
+    (event: string, options: Record<string, any>) => {
+      SendClientSideEvent(atmosphere, event, {
+        ...options,
+        eventId: ++eventId
```
tianrunhe, line 14: Why do we need `eventId` here?

packages/server/utils/analytics/analytics.ts
```diff
@@ -709,6 +710,19 @@ class Analytics {
     })
   }
 
+  insightsFeedbackSubmitted = (
+    user: AnalyticsUser,
+    isUseful: boolean | null,
```
tianrunhe, line 715: -1: it seems `isUseful` would never be `null`?

tianrunhe (changes requested):

packages/client/hooks/useManualClientSideTrack.ts
nickoferrall: We have it in the other hook as there seemed to be a bug where some users were sending it non-stop, and we were trying to debug whether it was happening from that hook or elsewhere. 

If it is happening in the other hook, it's more likely to be related to the `useEffect` which isn't in this hook. I'll remove it, and if the event fires non-stop too, I'll add eventId back.

PR #10391, opened 2024-10-21
chore(rethinkdb): Initialize DB

mattkrick:
# Description

creates a new baseline for migrations. One that doesn't include RethinkDB 🎉


packages/server/postgres/migrations/2024-10-21T21:21:21.212Z_init.ts
rafaelromcar-parabol, line 1: Isn't that filename problematic? if we use dates, we might have some problems with time zones

scripts/flushRedis.js
rafaelromcar-parabol, line 1: Is this script supposed to be used in production?

rafaelromcar-parabol (commented):
After this migration, would RethinkDB be needed during build process?

packages/server/postgres/migrations/2024-10-21T21:21:21.212Z_init.ts
mattkrick: the Z at the end denotes UTC & it's generated automatically

mattkrick:
> After this migration, would RethinkDB be needed during build process?

nope, build script has been amended, but i'll clean up the rest in the next PR

packages/server/postgres/migrations/2024-10-21T21:21:21.212Z_init.ts
rafaelromcar-parabol: Ouch, read too fast 😸 thanks!

scripts/flushRedis.js
rafaelromcar-parabol: I supposed, but just wanted to be 100% sure

rafaelromcar-parabol (approved):

mattkrick:
closing in favor of #10395 so we only have 1 PR

PR #10392, opened 2024-10-22
chore(rethinkdb): Remove RethinkDB usage from everything except migrations files

mattkrick:
# Description

Removes all references to RethinkDB


rafaelromcar-parabol:
Files to include in this PR:
- `.github/workflows/`: `build.yml` and `test.yml`
- `docker/images/parabol-ubi/`: `README.md` and both environment files.
- `docker/stacks`: in both stacks remove RethinkDB container and all references to it on the documentation.
- `.env.example`: all `RETHINKDB parameters`
- `.gitignore`: warning about checking with developers in case some very old data is still on any of those ignored folders.

Also: this migration must be scheduled, as it might make the RethinkDB ping alarm go off, as the application won't be able to connect to RethinkDB. The monitors on RethinkDB should be disabled right before the release.

Issue #10244, opened 2024-09-17
Insights: show Insights tab by default if viewer hasn't seen it


nickoferrall:
### AC

- [ ] If the viewer hasn't seen the Insights tab, make it the default
- [ ] Once they have seen it, make "Activity" the default again

PR #10405, opened 2024-10-23
feat: show insights by default

nickoferrall:
Fix https://github.com/ParabolInc/parabol/issues/10244

### To test

- [ ] Apply the insights feature flag to a team that you're the team lead of
- [ ] When you visit the team dash, the Insights tab shows by default
- [ ] When you return for a second time, the Activity tab is shown by default

### DIFF
--- a/packages/client/modules/teamDashboard/components/TeamDashHeader/TeamDashHeader.tsx
+++ b/packages/client/modules/teamDashboard/components/TeamDashHeader/TeamDashHeader.tsx
@@ -1,25 +1,26 @@
 import {ClassNames} from '@emotion/core'
 import styled from '@emotion/styled'
 import graphql from 'babel-plugin-relay/macro'
-import React from 'react'
+import React, {useEffect} from 'react'
 import {useFragment} from 'react-relay'
+import {useLocation} from 'react-router'
 import {NavLink} from 'react-router-dom'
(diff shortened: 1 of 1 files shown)


packages/client/modules/teamDashboard/components/TeamDashHeader/TeamDashHeader.tsx
```diff
@@ -132,42 +136,55 @@ const TeamDashHeader = (props: Props) => {
     ...(canViewInsights ? [{label: 'Insights', path: 'insights'}] : [])
   ]
 
+  const insightsSeenKey = `insightsSeen_${teamId}`
+  const isLocalStorageAvailable = typeof window !== 'undefined' && window.localStorage
```
Dschoordsch, line 140: +1 Any reason we explicitly check this here? I don't see this checked in other places

Dschoordsch (approved):

packages/client/modules/teamDashboard/components/TeamDashHeader/TeamDashHeader.tsx
nickoferrall: Local storage might not be available, like in Safari's incognito mode. It could throw an error without the check

PR #10437, opened 2024-10-31
feat: add use ai to org

nickoferrall:
This PR replaces `noAISummary` flag with `useAI` on the org table. 

`noAISummary` is confusing because:

1. It removes all AI features, not just the AI Summary 
2. It's not a feature flag, but it's being used as a feature flag
3. The negative form of `noAISummary` instead of `aiSummary` is confusing 

This PR adds `useAI` to the org table instead. By default, it's true. If an org has `noAISummary` flag turned on, this migration will make sure that `useAI` is false.

This PR won't be merged until a subsequent PR replaces the `noAISummary` checks with the `useAI` check.

### To test

- [ ] Run the migration and see that `useAI` is on by default and off for orgs that had `noAISummary` turned on

### DIFF
--- a/packages/server/postgres/migrations/2024-10-31T14:42:16.120Z_add-use-ai-to-orgs.ts
+++ b/packages/server/postgres/migrations/2024-10-31T14:42:16.120Z_add-use-ai-to-orgs.ts
@@ -1,0 +1,59 @@
+import {Kysely, sql} from 'kysely'
+
+export async function up(db: Kysely<any>): Promise<void> {
+  await db.schema
+    .alterTable('Organization')
+    .addColumn('useAI', 'boolean', (col) => col.notNull().defaultTo(true))
+    .execute()
+
+  await db
+    .updateTable('Organization')
+    .set({useAI: false})
+    .where(
+      'id',
+      'in',
+      db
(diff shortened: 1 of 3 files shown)


nickoferrall:
Merged this PR into https://github.com/ParabolInc/parabol/pull/10446

packages/server/postgres/migrations/2024-10-31T14:42:16.120Z_add-use-ai-to-orgs.ts
```diff
@@ -0,0 +1,59 @@
...
+  await db
+    .deleteFrom('FeatureFlagOwner')
+    .where(
+      'featureFlagId',
+      'in',
+      db.selectFrom('FeatureFlag').select('id').where('featureName', '=', 'noAISummary')
+    )
+    .execute()
+
+  await db.deleteFrom('FeatureFlag').where('featureName', '=', 'noAISummary').execute()
```
Dschoordsch, line 32: +1 I would expect the delete to cascade, so we wouldn't need ll. 23-30

Dschoordsch (commented):
It looks good, I did not test.

PR #10446, opened 2024-11-01
feat: update org feature flags backend

nickoferrall:
This PR shows all of the org feature flags in the organization settings. Org admins can view the feature flags and toggle them, updating the backend.

This PR includes migrating `noAISummary` feature flag to `useAI` on the `Organization` object. It does not include toggling the `useAI` boolean in the db. I'll add that in a subsequent PR.

Screenshot 2024-11-06 at 15 10 08

### To test

- [ ] Go to the Organization settings of an org where you're the org admin
- [ ] Toggle the feature flags


packages/server/postgres/migrations/2024-10-31T14:42:16.120Z_add-use-ai-to-orgs.ts
```diff
@@ -0,0 +1,59 @@
...
+    .where(
+      'id',
+      'in',
+      db
+        .selectFrom('FeatureFlagOwner')
+        .innerJoin('FeatureFlag', 'FeatureFlag.id', 'FeatureFlagOwner.featureFlagId')
+        .where('FeatureFlag.featureName', '=', 'noAISummary')
+        .select('FeatureFlagOwner.orgId')
+    )
+    .execute()
```
Dschoordsch, line 21: +1 Why do we have to do this? Shouldn't delete cascade for the 'FeatureFlagOwner' take care of this already?

packages/server/postgres/migrations/2024-10-31T14:42:16.120Z_add-use-ai-to-orgs.ts
```diff
@@ -0,0 +1,59 @@
...
+  await db
+    .deleteFrom('FeatureFlagOwner')
+    .where(
+      'featureFlagId',
+      'in',
+      db.selectFrom('FeatureFlag').select('id').where('featureName', '=', 'noAISummary')
+    )
+    .execute()
+
+  await db.deleteFrom('FeatureFlag').where('featureName', '=', 'noAISummary').execute()
```
Dschoordsch, line 32: +1 shouldn't this also delete cascade the 'FeatureFlagOwner' entries above?

nickoferrall: True! It does delete cascade so I can remove that FeatureFlagOwner deletion

nickoferrall:
Hey @mattkrick,  bumping this review request 🙏

packages/server/dataloader/customLoaderMakers.ts
```diff
@@ -955,3 +964,59 @@ export const publicTemplatesByType = (parent: RootDataLoader) => {
...
+
+export const allFeatureFlags = (parent: RootDataLoader) => {
+  return new DataLoader<'Organization' | 'Team' | 'User' | 'all', FeatureFlag[], string>(
+    async (scopes) => {
+      const pg = getKysely()
+      return await Promise.all(
+        scopes.map(async (scope) => {
+          let query = pg.selectFrom('FeatureFlag').selectAll().where('expiresAt', '>', new Date())
+          if (scope !== 'all') {
+            query = query.where('scope', '=', scope)
```
mattkrick, line 976: +1 you can use the `.$if` method to make this a const if you want, but end of the day it's all the same

packages/server/dataloader/customLoaderMakers.ts
```diff
@@ -955,3 +964,59 @@ export const publicTemplatesByType = (parent: RootDataLoader) => {
     }
   )
 }
+
+export const allFeatureFlags = (parent: RootDataLoader) => {
```
mattkrick, line 968: +1 if this will only be used by dataloaders you could make the PG query a little more efficient by doing it all in 1 go in the `allFeatureFlagsByOwner`, but probably not a big deal until we benchmark

packages/server/postgres/types/pg.d.ts
mattkrick, line 1: -1 this should not exist. isn't it in the gitignore?

packages/server/postgres/pg.d.ts
mattkrick, line 1: why is this here?

packages/server/graphql/public/types/ToggleFeatureFlagSuccess.ts
```diff
@@ -0,0 +1,18 @@
...
+  featureFlagId: string
+  enabled: boolean
+}
+
+const ToggleFeatureFlagSuccess: ToggleFeatureFlagSuccessResolvers = {
+  featureFlag: async ({featureFlagId, enabled}, _, {dataLoader}) => {
+    const flag = await dataLoader.get('featureFlags').loadNonNull(featureFlagId)
+    return {
+      ...flag,
+      enabled
```
mattkrick, line 13: +1 is this the only place where `enabled` is put onto the feature flag object? if so, it might make more sense to put `enabled` on its own field

packages/server/graphql/public/typeDefs/FeatureFlag.graphql
```diff
@@ -18,4 +23,9 @@ type FeatureFlag {
   The scope of the feature flag
   """
   scope: FeatureFlagScope!
+
+  """
+  Whether the flag is enabled for an owner or not
```
mattkrick, line 28: -1 `enabled` is a property of FeatureFlagOwner, so a type like `UpdateFeatureFlagSuccess` that returns a FeatureFlag won't have enabled as a field. This is technically OK since `enabled` is optional, but it's not very expressive because we're basically using the same `FeatureFlag` type for 2 different types of objects (a FeatureFlag, and a FeatureFlagWithOwner). i would keep `enabled` off this object or make a second object like `OwnedFeatureFlag` to express that they're different

packages/server/postgres/types/index.d.ts
```diff
@@ -80,6 +81,7 @@ export type TemplateScaleRef = ExtractTypeFromQueryBuilderSelect<typeof selectTe
 export type MeetingSettings = ExtractTypeFromQueryBuilderSelect<typeof selectMeetingSettings>
 export type PokerMeetingSettings = MeetingSettings & {meetingType: 'poker'}
 export type RetrospectiveMeetingSettings = MeetingSettings & {meetingType: 'retrospective'}
+export type FeatureFlag = Selectable<FeatureFlagPG> & {isEnabled: boolean}
```
mattkrick, line 84: -1 this is no longer true, right?

packages/server/postgres/types/pg.d.ts
mattkrick, line 1: this shouldn't be in version control, right?

packages/server/graphql/public/typeDefs/ToggleFeatureFlagSuccess.graphql
```diff
@@ -0,0 +1,6 @@
+type ToggleFeatureFlagSuccess {
+  """
+  The feature flag that was toggled
+  """
+  featureFlag: OwnedFeatureFlag!
```
mattkrick, line 5: curious, why create a new type vs. having `{enabled, featureFlag}`?

packages/server/postgres/types/pg.d.ts
nickoferrall: Removed!

packages/server/graphql/public/typeDefs/ToggleFeatureFlagSuccess.graphql
nickoferrall: I think we need it [here](https://github.com/ParabolInc/parabol/blob/93fa3cf3231c560b7b455ce4b6d68408711cc11a/packages/client/modules/userDashboard/components/OrgBilling/OrgFeatureFlags.tsx). 

In `ToggleFeatureFlagSuccess`, we could do what you're suggesting, but in `Organization`, I think we do have to return a new type? I went with your suggestion [here](https://github.com/ParabolInc/parabol/pull/10446#discussion_r1842984643).

mattkrick: ah! that makes sense. LGTM

mattkrick (approved):

PR #10457, opened 2024-11-06
feat: implement toggle ai features mutation

nickoferrall:
This PR enables org admins to toggle on or off AI features. 

Loom demo: https://www.loom.com/share/9294e9363fa94c098bed98d161229115

### To Test

- [ ] Ensure AI Features toggle is turned on and see AI summaries, icebreaker, etc
- [ ] Toggle it off and you don't see any AI features

### DIFF
--- a/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/WholeMeetingSummary.tsx
+++ b/packages/client/modules/email/components/SummaryEmail/MeetingSummaryEmail/WholeMeetingSummary.tsx
@@ -10,7 +10,7 @@ interface Props {
 }
 
 const isServer = typeof window === 'undefined'
-const hasAI = isServer
+const hasAiApiKey = isServer
   ? !!process.env.OPEN_AI_API_KEY
   : !!window.__ACTION__ && !!window.__ACTION__.hasOpenAI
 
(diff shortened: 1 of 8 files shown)


packages/server/postgres/select.ts
```diff
@@ -138,6 +138,7 @@ export const selectOrganizations = () =>
       'trialStartDate',
       'scheduledLockAt',
       'lockedAt',
+      'useAI',
```
nickoferrall, line 141: Adding this was a gotcha for me. Curious why we don't just do `selectAll`?

packages/client/modules/userDashboard/components/OrgBilling/OrgFeatures.tsx
```diff
@@ -34,23 +40,49 @@ const FeatureNameGroup = styled('div')({
...
+        useAI
+      }
+    `,
+    organizationRef
+  )
+  const {id: orgId, isOrgAdmin, useAI} = organization
+
+  const handleToggle = () => {
+    const variables = {orgId}
+    ToggleAIFeaturesMutation(atmosphere, variables, {
```
mattkrick, line 65: +1 if you use the `useMutation` pattern instead of the older pattern then you don't need `useMutationProps` in this component & it'll be a little cleaner

